PR #1535, opened 2024-01-14
Verify close() works in ws close handler

MellowYarker:
It looks like we're failing to send websocket `close()` events in the `webSocketClose()` handler for hibernatable websockets. Print debugging suggests we queue a close message to be sent but then never get past waiting for the `GatedMessage::outputLock` promise.

Strangely, the test seems to pass if a 1ms wait is added between a websocket send and a websocket close.

### DIFF
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -728,7 +728,21 @@ void WebSocket::ensurePumping(jsg::Lock& js) {
           reportError(js, KJ_EXCEPTION(DISCONNECTED, "WebSocket peer disconnected"));
         }
       } else if (native.closedIncoming && native.closedOutgoing) {
-        if (native.state.is<Accepted>()) {
+        if (awaitingHibernatableRelease()) {
+          // TODO(someday): These async races can be pretty complicated, and while it's good to have
+          // tests to make sure we're not broken, it would be nice to refactor this code eventually.
+
+          // Hibernatable WebSockets had a subtle race condition where one pump() promise would
+          // start right after a previous pump() completed, but before this continuation ran.
+          //
+          // This race prevented close messages from being sent from inside the webSocketClose()
+          // handler because prior to the CLOSE getting sent in the second pump(), the promise
+          // continuation following the first pump() would transition us from Accepted to Released,
+          // triggering the canceler and cancelling the outgoing CLOSE of the second pump() promise.
+          //
+          // For a more detailed explanation, see https://github.com/cloudflare/workerd/pull/1535.
+          tryReleaseNative(js);
+        } else if (native.state.is<Accepted>()) {
           // Native WebSocket no longer needed; release.
           native.state.init<Released>();
         } else if (native.state.is<Released>()) {
--- a/src/workerd/api/tests/websocket-hibernation.js
+++ b/src/workerd/api/tests/websocket-hibernation.js
@@ -1,0 +1,86 @@
+// Copyright (c) 2023 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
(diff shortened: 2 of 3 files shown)


MellowYarker:
Confirmed that the test fails in CI across all builds without the very brief `await` in the test. Checking if it succeeds with the `await`.

MellowYarker:
All builds passed in CI with that 1 ms `await` between the `send()` and `close()`. This seems to only be an issue for WebSockets using the Hibernation API. Not sure why `await`ing in the test would make any difference to outgoing messages from the DOs `api::WebSocket`, will need to investigate further.

jqmmes (approved):
LGTM!

That was hard to find, but good job fixing this :slightly_smiling_face:

src/workerd/api/web-socket.c++
```diff
@@ -728,7 +728,18 @@ void WebSocket::ensurePumping(jsg::Lock& js) {
...
+        if (awaitingHibernatableRelease()) {
+          // Because Hibernatable WebSocket events are dispatched at different times,
+          // we may have a situation where one event's pump() promise finishes (setting
+          // isPumping to false), but then this continuation doesn't get scheduled until after we
+          // start handling the next websocket event, which may call send()/close(), and result in
+          // another pump() promise being registered.
+          //
+          // When the first event continues execution, if we release the WebSocket without
+          // confirming it is not currently pumping, we can transition the WebSockets state from
+          // Accepted to Released. This will cancel the second events pump() promise.
```
MellowYarker, line 740: Wasn't exactly sure how to phrase this comment because its a pretty tricky situation.

Suppose a client does `ws.send()` followed immediately by `ws.close()`. The server (a DO in this case) has set up a message handler that replies with a message, and a close handler that replies with a close.

#### Regular WebSockets
The server receives a MESSAGE event, so it replies with a message. This inserts a new outgoing messsage to our `outgoingMessages`, then while we're `pump()`ing (specifically while `co_await`ing the `kj::WebSocket::send()`) we dispatch the CLOSE event, which inserts another message into our `outgoingMessage` map.
- Both messages are sent before we exit `pump()`.

#### Hibernatable WebSockets
The server receives a MESSAGE event, so it replies with a message. This inserts a new outgoing messsage to our `outgoingMessages`, but this time we complete the `co_await` of the `kj::WebSocket::send()` and are out of messages to send, so we finish with `pump()`.

Before we continue with `ensurePumping()`s continuation(following the completion of `pump()`, it's line 719 above), we dispatch the CLOSE event, which inserts a message into our `outgoingMessage` map and calls `ensurePumping()`. Since we had literally just set `native.isPumping = false` when we finished the pump for the previous message, we create another `pump()` promise. 

This second promise gets to the point where it `co_await`s the `GatedMessage::outputLock`, but we never get past that because we start executing the first `ensurePumping()` call's pump continuation, which sees that we closed our incoming and outgoing channels and transitions the websocket from `Accepted -> Released`. This state transition triggers the `canceler` that we wrapped the second `pump()` promise in and the message never gets sent out.

jasnell (approved):

justin-mp (approved):

src/workerd/api/web-socket.c++
bcaimano: Oooof, yeah, that makes sense but it's hard to phrase. For now, I'd probably just dump what you have here into the comment wholesale. I must admit it feels like we did something wrong that this all is this complicated.

bcaimano (approved):

src/workerd/api/web-socket.c++
MellowYarker: My gut says big comments get ignored/treated like artifacts (unless they're documentation), but I think [this](https://github.com/cloudflare/workerd/compare/a0a979b5e4686569934b57fde826b4a1c7e1aff9..a9e101b3a4f157dbbbb192ed30118a8442f255a4) is a decent summary w/ link to this PR for more context.

PR #1542, opened 2024-01-16
Reorder JS RPC message size exception

MellowYarker:
Addresses small typo.

### DIFF
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -52,9 +52,9 @@ kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> WorkerRpc::sendWorke
   // Note that we may fail to serialize some element, in which case this will throw back to JS.
   if (argv.size() > 0) {
     auto ser = serializeV8(js, js.arr(argv.asPtr()));
-    // TODO(soon): We will drop this requirement once we support streaming.
-    JSG_ASSERT(ser.size() < MAX_JS_RPC_MESSAGE_SIZE, Error,
-        "Serialized RPC request is too large: ", ser.size(), " <= ", MAX_JS_RPC_MESSAGE_SIZE);
+    JSG_ASSERT(ser.size() <= MAX_JS_RPC_MESSAGE_SIZE, Error,
+        "Serialized RPC requests are limited to 1MiB, but the size of this request was: ",
+        ser.size(), " bytes.");
     builder.initSerializedArgs().setV8Serialized(kj::mv(ser));
   }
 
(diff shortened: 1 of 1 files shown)


justin-mp (approved):

src/workerd/api/worker-rpc.c++
```diff
@@ -54,7 +54,7 @@ kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> WorkerRpc::sendWorke
     auto ser = serializeV8(js, js.arr(argv.asPtr()));
     // TODO(soon): We will drop this requirement once we support streaming.
     JSG_ASSERT(ser.size() < MAX_JS_RPC_MESSAGE_SIZE, Error,
-        "Serialized RPC request is too large: ", ser.size(), " <= ", MAX_JS_RPC_MESSAGE_SIZE);
+        "Serialized RPC request is too large: ", MAX_JS_RPC_MESSAGE_SIZE, " <= ", ser.size());
```
kentonv, line 57: This message formatting is probably still unclear to users, since it doesn't really say which number is the limit vs. the actual. Maybe needs to be something like:

> Serialized RPC requests are limited to 1MB, but the size of this request was: 12345678

src/workerd/api/worker-rpc.c++
```diff
@@ -129,8 +130,8 @@ public:
         auto result = serializeV8(js, jsg::JsValue(value.getHandle(js)));
         // TODO(soon): We will drop this requirement once we support streaming.
         JSG_ASSERT(result.size() < MAX_JS_RPC_MESSAGE_SIZE, Error,
-            "Serialized RPC response is too large: ", result.size(),
-            " <= ", MAX_JS_RPC_MESSAGE_SIZE);
+            "Serialized RPC responses are limited to 1MiB but the size of this request was: ",
```
MellowYarker, line 133: Oops, should be `size of this response`.

jasnell (approved):

bcaimano (approved):

src/workerd/api/worker-rpc.c++
```diff
@@ -54,7 +54,8 @@ kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> WorkerRpc::sendWorke
     auto ser = serializeV8(js, js.arr(argv.asPtr()));
     // TODO(soon): We will drop this requirement once we support streaming.
     JSG_ASSERT(ser.size() < MAX_JS_RPC_MESSAGE_SIZE, Error,
-        "Serialized RPC request is too large: ", ser.size(), " <= ", MAX_JS_RPC_MESSAGE_SIZE);
+        "Serialized RPC requests are limited to 1MiB, but the size of this request was: ",
+        ser.size(), " bytes.");
```
jclee, line 58: Probably not important, but just happened to notice that the message implies "<=" while the assert is checking "<".

MellowYarker: No this was a good catch, thanks Joe!

PR #1543, opened 2024-01-16
Implements embedding of Python packages in workerd.

dom96:
Embeds a tarball containing a Python `site-packages` which includes all the packages necessary to run a langchain + openai demo without the need to download anything at runtime.

Tested upstream via `python-langchain.ew-test`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -111,6 +111,14 @@ http_archive(
     build_file = "//:build/BUILD.pyodide",
 )
 
+http_archive(
+    name = "pyodide_packages",
+    sha256 = "f8fca2c4ecc09a57c86a2ed8d217e955d7599a83dad73fb989c86206706fdff6",
+    type = "zip",
+    urls = ["https://github.com/dom96/pyodide_packages/releases/download/v0.1/pyodide_packages_unzipped_0.1.tar.zip"],
+    build_file = "//:build/BUILD.pyodide_packages",
+)
+
 # ========================================================================================
 # Dawn
 #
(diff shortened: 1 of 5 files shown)


WORKSPACE
```diff
@@ -111,6 +111,14 @@ http_archive(
...
 )
 
+http_archive(
+    name = "pyodide_packages",
+    sha256 = "f8fca2c4ecc09a57c86a2ed8d217e955d7599a83dad73fb989c86206706fdff6",
+    type = "zip",
+    urls = ["https://github.com/dom96/pyodide_packages/releases/download/v0.1/pyodide_packages_unzipped_0.1.tar.zip"],
+    build_file = "//:build/BUILD.pyodide_packages",
+)
+
```
hoodmane, line 121: We'll want to add more packages almost immediately so it'd be good to add a bit more flexibility. Probably it'd be better to have it as a build step. But I suppose it's good enough for this PR.

src/pyodide/BUILD.bazel
```diff
@@ -14,6 +15,12 @@ copy_file(
     out = "generated/python_stdlib.zip",
 )
 
+copy_file(
+    name = "pyodide_packages_file.zip@rule",
```
hoodmane, line 19: There's a few places where it says zip that should probably say tar
```suggestion
    name = "pyodide_packages_file.tar@rule",
```

WORKSPACE
dom96: Yeah, I think it's easy enough to modify the tarball ourselves if we need to. I don't think it's worth spending time creating a build step for this for now.

hoodmane: You're not in the mood for a bazel learning experience?

hoodmane (approved):

src/pyodide/BUILD.bazel
dom96: for the sake of CI expediency, I'll fix this in another PR (where I will add fastapi packages too)

PR #1544, opened 2024-01-16
Update capnproto to get new WebSocket closing behavior

smerritt:
When WebSocket clients send frames we can't accept (too large, invalid UTF-8, etc.), we'll now reply with a Close frame explaining the error. Previously, we'd just disconnect the client without any hint as to what they did wrong.

This is analogous to an HTTP server responding with 400 when a client sends bad headers (or something) instead of just disconnecting.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -39,10 +39,10 @@ apple_support_dependencies()
 
 http_archive(
     name = "capnp-cpp",
-    sha256 = "d2cac87fabbb9dfd38a20903158b533eb38cd369f8d5fe89b65fef3f75099cad",
-    strip_prefix = "capnproto-capnproto-9084594/c++",
+    integrity = "sha256-kVnOrg/h0eq37kBsFEHTUZax20JTR4/1eDy+kRY7wD8=",
+    strip_prefix = "capnproto-capnproto-bafc222/c++",
     type = "tgz",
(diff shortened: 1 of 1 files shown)


justin-mp (approved):

WORKSPACE
```diff
@@ -39,10 +39,10 @@ apple_support_dependencies()
 
 http_archive(
     name = "capnp-cpp",
-    sha256 = "d2cac87fabbb9dfd38a20903158b533eb38cd369f8d5fe89b65fef3f75099cad",
-    strip_prefix = "capnproto-capnproto-9084594/c++",
+    integrity = "sha256-kVnOrg/h0eq37kBsFEHTUZax20JTR4/1eDy+kRY7wD8=",
```
MellowYarker, line 42: @smerritt just curious about this change, all the other deps still use `sha256 = `. Is this better, and if so should we switch the other deps as well?

smerritt: It looks equivalent, like it's a sha256 but base64 encoded (and with a type specifier).

I think this is different because we upgraded to a new Bazel version. The old version, if no checksum was specified, would say "hey, put in `sha256 = "abc123..."` for assurance that it hasn't changed", and then I would put that line in. The new Bazel version says "hey, put in `integrity = "sha256-asdfZr+blah"`", so I did that.

PR #1545, opened 2024-01-17
Improve streams ReadPendingScope cleanup

jasnell:
Changes the `ReadPendingScope` from a struct to a function that includes (hopefully) appropriate error handling logic.

### DIFF
--- a/src/workerd/api/streams/standard.c++
+++ b/src/workerd/api/streams/standard.c++
@@ -524,22 +524,28 @@ int getHighWaterMark(const UnderlyingSource& underlyingSource,
 
 // It is possible for the controller state to be released synchronously while
 // we are in the middle of a read. When that happens we need to defer the actual
-// close/error state change until the read call is complete. ReadPendingScope
-// handles this for us. It is constructed when we start a read and destructed
-// synchronously when the call to read completes.
-template <typename C>
-struct ReadPendingScope {
-  jsg::Lock& js;
-  C& controller;
-  ReadPendingScope(jsg::Lock& js, C& controller)
-      : js(js), controller(controller) {
+// close/error state change until the read call is complete. deferControllerStateChange
+// handles this for us by making sure that pendingReadCount is
+// incremented until the read operation completes and deferring
+// a state change until it is 0
+template <typename Controller>
+jsg::Promise<ReadResult> deferControllerStateChange(
+    jsg::Lock& js,
+    Controller& controller,
+    kj::FunctionParam<jsg::Promise<ReadResult>()> readCallback) {
+  bool decrementCount = true;
+  // The readCallback and the controller.doClose(..) and controller.doError(...)
+  // methods, as well as the methods can trigger JavaScript errors to be thrown
+  // synchronously in some cases. We want to make sure non-fatal errors cause the
+  // stream to error and only fatal cases bubble up.
+  return js.tryCatch([&] {
     controller.pendingReadCount++;
-  }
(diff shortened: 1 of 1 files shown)


src/workerd/api/streams/standard.c++
```diff
@@ -524,22 +524,26 @@ int getHighWaterMark(const UnderlyingSource& underlyingSource,
...
-// handles this for us. It is constructed when we start a read and destructed
-// synchronously when the call to read completes.
-template <typename C>
-struct ReadPendingScope {
-  jsg::Lock& js;
-  C& controller;
-  ReadPendingScope(jsg::Lock& js, C& controller)
-      : js(js), controller(controller) {
+// close/error state change until the read call is complete. readPendingScope
+// handles this for us.
```
mikea, line 528: this is a tricky code, I would very much like this sentence to continue, e.g.: `handles this for us by making sure that pendingReadCount is incremented until the read operation completes and deferring a state change until it is 0`. (did I get the code right btw?)

src/workerd/api/streams/standard.c++
```diff
@@ -2467,12 +2482,14 @@ kj::Maybe<jsg::Promise<ReadResult>> ReadableStreamJsController::read(
       // The ReadableStreamDefaultController does not support ByobOptions.
       // It should never happen, but let's make sure.
       KJ_ASSERT(maybeByobOptions == kj::none);
-      ReadPendingScope readPendingScope(js, *this);
-      return consumer->read(js);
+      return readPendingScope(js, *this, [&]() mutable {
```
mikea, line 2485: maybe a better name is possible? how about `ensureControllerAlive` or something like that?

mikea (approved):
LG but would really like expanded comment and slightly better function name if possible. thanks.

src/workerd/api/streams/standard.c++
```diff
@@ -524,22 +524,26 @@ int getHighWaterMark(const UnderlyingSource& underlyingSource,
...
-template <typename C>
-struct ReadPendingScope {
-  jsg::Lock& js;
-  C& controller;
-  ReadPendingScope(jsg::Lock& js, C& controller)
-      : js(js), controller(controller) {
+// close/error state change until the read call is complete. readPendingScope
+// handles this for us.
+template <typename Controller>
+jsg::Promise<ReadResult> readPendingScope(
```
mikea, line 530: Also, can't this be `jsg::Promise<T> ensureAlive(..., <Promise<T>()> callback)`?

hoodmane (commented):
So is the point of this that previously if `readable->read` threw a C++ error, it wouldn't be caught whereas now it gets caught and used to reject the promise?

Looks reasonable to me, but I'm not great at reviewing C++ patches.

src/workerd/api/streams/standard.c++
jasnell, line 528:
```suggestion
// handles this for us by making sure that pendingReadCount is incremented until the read
// operation completes and deferring a state change until it is 0
```

src/workerd/api/streams/standard.c++
harrishancock: `runDeferringCloseAndErrorStateChanges()`? :)

harrishancock (approved):

src/workerd/api/streams/standard.c++
jasnell: Perhaps, but I think keeping this as is should be fine.

PR #1552, opened 2024-01-18
Update to v8 12.1

jasnell:
Includes a new patch to the v8 build that ensures that continuation preserved data is always enabled in the bazel build.


fhanau:
Was going to comment here about `new_git_repository` being deprecated, but I was on mobile and the comment didn't actually get submitted since Github treats it as part of a not yet completed review... – I'll add this to a cleanup PR.

Issue #854, opened 2023-05-02
🐛 BUG: Issue with nodejs_compat + dynamic require of node:events


dja:
### Which Cloudflare product(s) does this pertain to?

Workers/Other, Wrangler

### What version of `Wrangler` are you using?

2.18.0

### What operating system are you using?

Mac

### Describe the Bug

Attempting to upload a worker with Wrangler using `compatibility_flags = ['nodejs_compat']` or command line `--compatibility-flags="nodejs_compat"` fails with this error message:

```
  Uncaught Error: Dynamic require of "node:events" is not supported
    at worker.js:13:11
    at worker.js:13348:28
    at worker.js:13631:3
   [code: 10021]
```

The package using `node:events` is this LaunchDarkly SDK: https://github.com/launchdarkly/js-core/tree/main/packages/sdk/cloudflare

dcrodman:
I'm using `import { Buffer } from "node:buffer";` (as opposed to `require()`) as the docs recommended and am still encountering the issue

jasnell:
Which workers syntax are y'all using? The older "server worker" syntax (using `addEventListener`) or the new ESM syntax e.g. `export default { async fetch .... }` ? If you have a simple repro script handy that doesn't use any specific dependencies that'll be helpful too.

dcrodman:
@jasnell here's useless but functional barebones example where I see the problem:

worker.ts
```
import { Buffer } from 'node:buffer';

addEventListener('fetch', (event) => {
	let fetchEvent: FetchEvent = event as FetchEvent;
	fetchEvent.respondWith(handle(fetchEvent));
});

async function handle(event: FetchEvent): Promise<Response> {
	let b = Buffer.from(event.request.body);
	console.log(b);
	return new Response(event.request.body);
}
```
wrangler.toml
```
name = "cloudflare-broken-node-demo"
main = "src/worker.ts"
compatibility_date = "2023-07-10"
compatibility_flags = [ "nodejs_compat" ]
```
package.json
```
{
  "name": "cloudflare-broken-node-demo",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "deploy": "wrangler deploy",
    "start": "wrangler dev"
  },
  "devDependencies": {
    "@cloudflare/workers-types": "^4.20230419.0",
    "typescript": "^5.0.4",
    "wrangler": "^3.0.0"
  }
}
```
I haven't deployed this to the actual worker instance, but I see the issue I was seeing before with miniflare locally:
[...]

PR #1553, opened 2024-01-18
Initial implementation of new jsg module registry

jasnell:
RM-17318

First step of the new module registry implementation (the jsg only bits). Expect several more follow up PRs.

This PR provides the initial core of the new module registry implementation. It DOES NOT enable it in workerd at all, this is just the core implementation in workerd/jsg plus tests. Note that this will be gated with a compatibility flag because there are observable changes in behavior. The old module registry implementation will continue to exist but it will be incrementally refactored a bit later to reduce some complexities and eliminate some redundancies.

The new tests should illustrate fairly clearly how the new registry implementation works. I recommend spending a fair amount of time reading through the test code.

Here's a tl;dr with some highlights.

Every Worker will have a `ModuleRegistry` that is built when the worker is created. The `ModuleRegistry` is composed of zero or more `ModuleBundle`s. It is the reponsibility of the `ModuleBundle` to actually manage and maintain the `Module`s.

A `ModuleRegistry::Builder` is used to create the `ModuleRegistry`. Multiples types of `ModuleBundle`s can be added:

* The `BUNDLE` type `ModuleBundle` contains modules from the worker configuration bundle itself.
* The `BUILTIN` type `ModuleBundle` contains modules from the runtime that can be imported by user code
* The `BUILTIN_ONLY` type `ModuleBundle` contains modules from the runtime that can only be imported by other built-ins
* The `FALLBACK` type `ModuleBundle` is used to provide dynamic resolution using a fallback service (for local dev)

Importantly, the `ModuleBundle` is a virtual type that can be implemented in multiple ways. This will give us long term flexibility to add new kinds of modules from different kinds of sources without having to modify `workerd/jsg` directly to support them.

Once a `ModuleRegistry` is created it is entirely immutable other than any internal caching a `ModuleBundle` might do to optimize performance or ensure consistent resolution.

Importantly, a `ModuleRegistry` only needs to be built once. While `workerd` will only ever use a `ModuleRegistry` instance with a single Worker, the production environment may share a single instance across multiple replicas. The `ModuleRegistry` may be bound to any number of `v8::Isolate`s. 

All import specifiers are `URL`s. For `BUNDLE` Modules, the URL base is always `file:///`. For `BUILTIN` and `BUILTIN_ONLY` Modules, the base is always some built-in prefix like `node:` or `cloudflare:` or `workerd:`. `FALLBACK` type modules have a bit more flexibility on the base URL.

ESM modules support `import.meta.url` and `import.meta.resolve(...)`.

New module types can be added without making any modifications to `workerd/jsg`.

All modules are resolved and evaluated lazily (when they are first imported). Dynamic imports and sync requires run within the current `IoContext` (if any) so they can schedule i/o (e.g. `fetch`, `scheduler.wait(...)`, `setTimeout(...)`) if there is a current `IoContext`.

Expanded metrics are supported (via a new `ResolveObserver`) so we can measure resolution times and track which built-ins are actually being used.

All ESM modules (and eventually all CommonJS modules) will use v8 compile cache.

So... why do this at all? There are a number of intended goals:

* Treat all import specifiers as URLs (per web platform standards)
* Implement web platform standard `import.meta.url`, `import.meta.resolve(...)`, and `import.meta.main`
* Soon, support TC-39 import attributes
* Resolve and evaluate *all* modules lazily
* Allow all replicas of a worker to use the same `ModuleRegistry` (avoid rebuilding it for every replica)
* Support compile cache for all ESM/CJS modules
* Improve impl of fallback resolution service for local dev
* Allow dynamic imports/requires to be IoContext aware
* Enable `nodeJsCompatModule` to use node.js built-ins without needing the `nodejs-compat` flag (eventually, not covered in this PR)
* Allow workerd internals to use node.js built-ins even when legacy service worker syntax is used
* Allow easier introduction of new module types (can be added without modifying workerd/jsg)
* Improve observability and metrics
* Improved maintainability

Some details on handling import specifiers as URL:

* All modules contained in a worker bundle use `file:///` as the default base URL. We don't actually expose a filesystem in workerd so these aren't *actual* file paths but using `file:///` is consistent with other runtimes, such as Node.js
* Worker bundles CAN specify full absolute URLs other than `file:///`. For instance, they can shadow built-ins like `node:buffer`, or even use `http://` URLs. These are used only as identifiers and never actually are used for http requests.
* Single and double dot segments are normalized in the specifier. For instance, the import specifiers `foo/../bar` and `bar` are equivalent.
* Relative import specifiers are resolved relative to the module that is importing them. For instance, if the importing module has a base URL of `file:///foo`, then the import specifier `bar` will resolve to `file:///bar`. However, if the importing module has a base URL of `file:///foo/` (with a trailing slash), then the import specifier `bar` will resolve to `file:///foo/bar`.
* Percent-encoding in the specifier path is normalized. For instance `%66oo` is normalized to `foo`.
* Query strings and fragments in the specifier are significant in that they will resolve to the same underlying module but will result in a new evaluated instance. For example, `foo?x` and `foo?y` will cause the module registry to resolve the module `file:///foo` *only once* but the specifiers will evaluate into two separate instances of the module. This is consistent with the behavior we see in other runtimes.

Few other bits and pieecs:

* Import attributes are explicitly *unsupported* right now. This is per the recommendation of the TC-39 proposal that asks embedders to explicitly reject import attributes that are not understood in order to preserve backwards compatibility.
* Alias specifiers are supported. This means the module registry can map one import specifier as an alias of another import-maps style. If `file:///foo` is registered as an alias of `file:///bar`, then both will resolve the same underlying `Module` but will be instantiated as two different instances.
* WASM modules are compiled once and cached. When multiple instances are evaulated, the cache will be used, speeding initialization and reducing memory.


src/workerd/jsg/observer.h
```diff
@@ -16,6 +17,58 @@ namespace v8 {
...
+    // The resolve originated from a dynamic import statement.
+    DYNAMIC_IMPORT,
+    // The resolve originated from a CommonJS require() call.
+    REQUIRE,
+    // The resolve originated from an internal direct call to
+    // the ModuleRegistry.
+    INTERNAL,
+    // The resolve originated from some other source (to be defined).
+    OTHER,
+  };
```
jasnell, line 41: Review note: The source is generally informational. It is included largely to support the fallback service use case where they need to know how exactly a module is being resolved.

jasnell:
TODOs...

* [ ] Review Asan failures reported in github CI

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1656 @@
...
+  ResolveObserverImpl observer;
+  CompilationObserver compilationObserver;
+
+  ModuleBundle::BundleBuilder bundleBuilder;
+
+  // A recursive import with an ESM works just fine...
+  auto foo = kj::str("import foo from 'foo'; export default 123;");
+  bundleBuilder.addEsmModule("foo", foo.slice(0, foo.size()).attach(kj::mv(foo)));
+
+  // A CommonJS-style module, however, does not allow recursive evaluation.
```
mrbbot, line 1399: Are circular `require()`s supported? If so, could a test be added for those?

mrbbot (commented):
This looks great! 😃 Added some questions from a supporting Vitest perspective.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1154 @@
...
+    .referrer = referrerUrl,
+  };
+  // TODO(soon): Add import assertions to the context.
+
+  return registry.resolve(js, resolveContext);
+}
+
+// The fallback module bundle calls a single resolve callback to resolve all modules
+// it is asked to resolve. Instances must be thread-safe.
+class FallbackModuleBundle final : public ModuleBundle {
```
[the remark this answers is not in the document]
jasnell: I'll make sure it does. We would likely want both specifiers to serve as aliases for the returned module in the internal cache.

src/workerd/jsg/modules-new-test.c++
jasnell: No, they are not. A test will be added for that.

src/workerd/jsg/modules-new.h
```diff
@@ -0,0 +1,556 @@
...
+
+// The ResolveContext identifies the module that is being resolved along with
+// other key bits of information that may be used to resolve the module.
+struct ResolveContext {
+  using Source = ResolveObserver::Source;
+  using Type = ResolveObserver::Context;
+
+  Type type;
+  Source source;
+  const Url& specifier;
```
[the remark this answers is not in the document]
jasnell: Noted, will get that added.

src/workerd/jsg/modules-new-test.c++
jasnell: I was pretty sure there was already a test in `modules-new-test.c++` for this and just verified it. When we add the `CommonJs` and `NodeJs` module specific tests for this new impl later I'll add further verifications.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1382 @@
...
+          return ModuleBundle::BundleBuilder::BASE.clone();
+        }
+        auto str = js.toString(resource_name);
+        return KJ_ASSERT_NONNULL(Url::tryParse(str.asPtr()));
+      })();
+
+      KJ_IF_SOME(url, referrer.tryResolve(spec.asPtr())) {
+        auto normalized = url.clone(Url::EquivalenceOption::NORMALIZE_PATH);
+        auto& registry = IsolateModuleRegistry::from(isolate);
+        return registry.dynamicResolve(js, kj::mv(normalized), kj::mv(referrer), spec);
```
jasnell, line 619: Note for reviewers: As discussed in the overview call, the new module registry is *currently* implemented such that dynamic imports are resolved and evaluated within the active request context (`IoContext`) if there is one. This is where that happens. If we decide that dynamic imports should NOT be aware of the request context, then we should change this here to ensure that we pop out of the context, perform the evaluation, then come back into it the way we currently do with the old registry.

irvinebroque:
Does this address https://github.com/cloudflare/workerd/issues/854#issuecomment-1675324298 ?

src/workerd/jsg/modules-new.h
```diff
@@ -0,0 +1,575 @@
...
+#include <kj/common.h>
+#include <kj/function.h>
+#include <kj/refcount.h>
+#include <kj/map.h>
+#include <kj/table.h>
+#include <v8.h>
+
+namespace workerd::jsg::modules {
+
+// Defines and implements workerd's (new) module loader subsystem.
```
garrettgu10, line 21: I'm unfamiliar with the old module system, and I think it would be nice to include a sentence or two here about how this new system is different from the old one. 

My impression is that this registry is primarily designed to be more modular and extensible, with the motivation being supporting more diverse specifiers including URL, dynamic imports, and sharing between replicas. Is that correct?

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1382 @@
...
+
+  kj::Array<const char> source;
+  kj::ArrayPtr<const char> ptr;
+  // When externed is true, the source buffer is passed into the isolate as an externalized
+  // string. This is only appropriate for built-in modules that are compiled into the binary.
+  bool externed = false;
+  kj::MutexGuarded<kj::Maybe<kj::Array<kj::byte>>> cachedData;
+};
+
+class SyntheticModule final : public Module {
```
garrettgu10, line 168: What is a synthetic module?

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1382 @@
...
+    bool matches(const Entry& entry, const Url& specifier) const {
+      return entry.specifier.specifier == specifier;
+    }
+    uint hashCode(const Url& specifier) const {
+      return specifier.hashCode();
+    }
+  };
+
+  // Resolves the module from the inner ModuleRegistry, caching the results.
+  kj::Maybe<Entry&> resolveWithCaching(Lock& js, const ResolveContext& context)
```
garrettgu10, line 451: Name seems a bit misleading, I suggest resolveAndCache maybe?

garrettgu10 (commented):
Admittedly I'm still missing a little context, but hopefully my second pass will be more effective once I get these questions answered :-)

jasnell:
Separated out a couple of the smaller changes to separate PRs in #1823 and #1824

src/workerd/jsg/modules-new.c++
jasnell: There are generally two kinds of modules in v8: ESM and Synthetic. An ESM module is one whose module namespace is defined by the `exports` in the module code. A Synthetic module is one that is not based on ESM syntax, where the runtime essentially synthetically creates the module namespace programmatically.

src/workerd/jsg/url.c++
```diff
@@ -37,6 +40,57 @@ T getInner(const kj::Own<void>& inner) {
...
 
+kj::Array<const char> normalizePathEncoding(kj::ArrayPtr<const char> pathname) {
+  // Sadly, this is a bit tricky because we do not want to decode %2f as a slash.
+  // we want to keep those as is. So we'll split the input around those bits.
+  // Unfortunately we need to split on either %2f or %2F, so we'll need to search
+  // through ourselves rather than using something like kj::delimited. This is
+  // simple enough, tho. We'll percent decode as we go, re-encode the pieces
+  // and then join them back together with %2F.
+
+  static constexpr auto findNext = [](std::string_view input)
```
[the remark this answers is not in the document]
jasnell: Handled in the separate PR #1823 ... will rebase this once that PR lands

src/workerd/jsg/modules-new.h
```diff
@@ -0,0 +1,575 @@
...
+// the ModuleRegistry will not actually generate the v8::Local<v8::Module>,
+// compile scripts, or evaluate it until the module is actually imported.
+// This means that any modules that are never actually imported by a worker
+// will never actually be compiled or evaluated.
+//
+// A ModuleBundle will have one of three basic types: Bundle, Builtin,
+// or Builtin-Only. A Bundle ModuleBundle provides access to modules
+// that are defined in the worker bundle configuration. A Builtin ModuleBundle
+// provides access to modules that are compiled into the runtime and are
+// importable by bundle scripts. A Builtin-Only ModuleBundle provides access
```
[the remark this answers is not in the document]
jasnell: The names come from the current module system. Will look at renaming these to something better in a separate follow-on PR

src/workerd/jsg/modules-new.h
```diff
@@ -0,0 +1,575 @@
...
+// qualified, absolute URL whose protocol is `node:` and whose pathname is
+// `buffer`.
+//
+// Specifieers for modules that come from the worker bundle config are always
+// relative to `file:///`.
+//
+// For ESM modules, the specifier is accessible using import.meta.url.
+// All ESM modules support import.meta.resolve(...)
+//
+// If the first module in the module worker bundle is an ESM it will specify
```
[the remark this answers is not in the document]
jasnell: This has been the convention we've always followed and is the way wrangler and EWC will bundle workers.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1382 @@
...
+  })();
+
+  return [ref=js.v8Ref(fn)](Lock& js) mutable {
+    js.withinHandleScope([&] {
+      auto fn = ref.getHandle(js);
+      // Any return value is explicitly ignored.
+      auto context = js.v8Context();
+      auto val = JsValue(check(fn->Call(context, context->Global(), 0, nullptr)));
+      JSG_REQUIRE(!val.isPromise(), TypeError,
+          kj::str("Module evaluation function must not return a promise"));
```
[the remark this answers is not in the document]
jasnell: Largely just because of compatible implementation with other runtimes. Technically it could but there's nothing that can reasonably be expected to be done with it.

jasnell:
Separated a bit more of this out into a separate PR https://github.com/cloudflare/workerd/pull/1977 ... just trying to shrink the size a bit more.

src/workerd/jsg/modules-new.c++
jasnell: Hmm, I'm not entirely sure how that's different ;-)

src/workerd/jsg/modules-new.h
```diff
@@ -0,0 +1,580 @@
...
+    kj::Own<ModuleRegistry> finish() KJ_WARN_UNUSED_RESULT;
+
+  private:
+    bool allowsFallback() const;
+
+    // One slot for each of ModuleBundle::Type
+    const ResolveObserver& observer;
+    kj::Maybe<ModuleRegistry&> maybeParent;
+    const Options options;
+    kj::Vector<kj::Own<ModuleBundle>> bundles_[4];
```
fhanau, line 527: Use `kBundleCount` here too if possible

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "jsg-test.h"
```
fhanau, line 5: Nit: use the more lightweight `#include <kj/test.h>` when not using Evaluator etc. (this was applied to existing code in 48f1af4e5bb9)

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+  ModuleBundle::BundleBuilder builder;
+
+  kj::String source = kj::str("export const foo = 123;");
+  builder.addEsmModule("foo", source.releaseArray(), Module::Flags::MAIN);
+
+  auto bundle = builder.finish();
+
+  const auto specifier = "file:///foo"_url;
+
+  // Resolve should return nothing.
```
fhanau, line 260: Resolve does return the ESM module here.

src/workerd/jsg/modules-new.h
```diff
@@ -0,0 +1,580 @@
...
+      Lock& js,
+      kj::StringPtr code,
+      kj::StringPtr name,
+      kj::Maybe<JsObject> compileExtensions,
+      const CompilationObserver& observer) KJ_WARN_UNUSED_RESULT;
+
+  // Some modules may need to protect against being evaluated recursively. The
+  // EvaluatingScope class makes it possible to guard against that, throwing an
+  // error if there's already an active evaluation happening.
+  class EvaluatingScope {
```
fhanau, line 294: nit: We can use more `final` statements here, `EvaluatingScope` and several subclasses come to mind.

src/workerd/jsg/setup.h
```diff
@@ -485,12 +485,19 @@ class Isolate: public IsolateBase {
...
       // TODO(soon): Requiring move semantics for the global object is awkward. This should instead
       //   allocate the object (forwarding arguments to the constructor) and return something like
       //   a Ref.
+      return jsgIsolate.wrapper->newContext(*this, options, jsgIsolate.getObserver(), (T*)nullptr, kj::fwd<Args>(args)...);
+    }
 
-      return jsgIsolate.wrapper->newContext(*this, jsgIsolate.getObserver(), (T*)nullptr, kj::fwd<Args>(args)...);
+    // Creates a new JavaScript "context", i.e. the global object. This is the first step to
+    // executing JavaScript code. T should be one of your API types which you want to use as the
+    // global object. `args...` are passed to the type's constructor.
```
fhanau, line 497: Nit: comment is repeated from above and perhaps redundant, this can stay as-is if you think that this is more clear.

fhanau:
[build] Note that as of #2012 we no longer use a glob for the JSG source files in `src/workerd/jsg/BUILD.bazel` – you'll need to explicitly add `modules-new.c++` to the set, linking the new test will fail otherwise after rebasing.

fhanau:
I went through the header and some of the tests, the API design looks good! Will continue reviewing this tomorrow.

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+      KJ_ASSERT(str == "Error: The module evaluation did not complete synchronously. "
+                       "This is not permitted for synchronous require(...). "
+                       "Use await import(...) instead.");
+    });
+  });
+}
+
+// ======================================================================================
+
+KJ_TEST("Recursively require ESM from CJS required from ESM fails as expected (2)") {
```
fhanau, line 1437: It took me a while to find the difference from the previous test... Let's rename them to e.g. `<test name> (await import)` and `<test name> (synchronous import)` to highlight the difference

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+
+KJ_TEST("Building a bundle from a capnp description works") {
+
+  capnp::MallocMessageBuilder builder;
+  auto bundle = builder.initRoot<workerd::jsg::Bundle>();
+
+  auto modules = bundle.initModules(3);
+
+  auto str = kj::str("export default 1+1;");
+  auto wasm = kj::heapArray<kj::byte>({
```
fhanau, line 1528: Does KJ have something to avoid heap allocation here? Not important within a test of course, but I would find it good to know about.

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+
+    js.tryCatch([&] {
+      ModuleRegistry::resolve(js, "file:///foo", "default"_kjc);
+      JSG_FAIL_REQUIRE(Error, "Should have thrown");
+    }, [&](Value exception) {
+      auto str = kj::str(exception.getHandle(js));
+      KJ_ASSERT(str == "Error: foo");
+    });
+
+    js.tryCatch([&] {
```
fhanau, line 1252: Nit: If this block is repeated on purpose, add a comment stating this. This will make it clear from the start that the blocks are identical. Would a 2-iteration loop be an option here?

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+  ResolveContext context {
+    .type = ResolveContext::Type::BUNDLE,
+    .source = ResolveContext::Source::OTHER,
+    .specifier = foo,
+    .referrer = ModuleBundle::BundleBuilder::BASE,
+  };
+
+  auto& module = KJ_ASSERT_NONNULL(registry->resolve(context));
+  KJ_ASSERT(module.specifier() == foo);
+  KJ_ASSERT(module.type() == Module::Type::BUNDLE);
```
fhanau, line 776: How do these asserts show that the built-in has been shadowed? Is it through the isMain() check?

src/workerd/jsg/modules-new-test.c++
jasnell: Yeah the comment is out of date
```suggestion
```

src/workerd/jsg/modules-new-test.c++
jasnell: The `Module::Type::BUNDLE` check

src/workerd/jsg/modules-new-test.c++
jasnell: Actually,..... I'm not sure why it ended up being repeated lol

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+      KJ_ASSERT(str == "Error: foo");
+    });
+
+    js.tryCatch([&] {
+      ModuleRegistry::resolve(js, "file:///foo", "default"_kjc);
+      JSG_FAIL_REQUIRE(Error, "Should have thrown");
+    }, [&](Value exception) {
+      auto str = kj::str(exception.getHandle(js));
+      KJ_ASSERT(str == "Error: foo");
+    });
```
jasnell, line 1258:
```suggestion
```

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+    }, [&](Value exception) {
+      auto str = kj::str(exception.getHandle(js));
+      KJ_ASSERT(str == "Error: Module cannot be recursively evaluated: file:///bar");
+    });
+  });
+}
+
+// ======================================================================================
+
+KJ_TEST("Recursively require ESM from CJS required from ESM fails as expected") {
```
jasnell, line 1399:
```suggestion
KJ_TEST("Recursively require ESM from CJS required from ESM fails as expected (dynamic import)") {
```

src/workerd/jsg/modules-new-test.c++
jasnell:
```suggestion
KJ_TEST("Recursively require ESM from CJS required from ESM fails as expected (static import)") {
```

src/workerd/jsg/modules-new-test.c++
jasnell: No, this could likely just be a `kj::ArrayPtr` but it's just a test so didn't bother

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+      .type = ResolveContext::Type::BUNDLE,
+      .source = ResolveContext::Source::OTHER,
+      .specifier = specifier,
+      .referrer = ModuleBundle::BundleBuilder::BASE,
+    };
+    KJ_ASSERT_NONNULL(registry->resolve(resolveContext));
+
+    registry->attachToIsolate(js, compilationObserver);
+
+    js.tryCatch([&] {
```
fhanau, line 901: Merging some of the tryCatch blocks would make this test more concise

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+      auto val = ModuleRegistry::resolve(js, "file:///worker", "xyz"_kjc);
+      KJ_ASSERT(val.isArrayBuffer());
+    }, [&](Value exception) {
+      js.throwException(kj::mv(exception));
+    });
+
+    js.tryCatch([&] {
+      auto val1 = ModuleRegistry::resolve(js, "file:///worker", "json"_kjc);
+      auto val2 = ModuleRegistry::resolve(js, "file:///json", "default"_kjc);
```
fhanau, line 918: Nit: the default value for `exportName` already is "default", it can be omitted here.

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+
+    auto bar = kj::str("export default 123;");
+    bundleBuilder.addEsmModule("bar", bar.slice(0, bar.size()).attach(kj::mv(bar)));
+
+    auto registry = ModuleRegistry::Builder(observer).add(bundleBuilder.finish()).finish();
+
+    registry->attachToIsolate(js, compilationObserver);
+
+    js.tryCatch([&] {
+      // This is a synchronous resolve that should successfully wait on the async
```
fhanau, line 1216: This comment is misleading and the bar variable above is redundant, bar is not being referenced anywhere.

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+      KJ_ASSERT(!wasm1.strictEquals(wasm3));
+    }, [&](Value exception) {
+      js.throwException(kj::mv(exception));
+    });
+  }));
+}
+
+// ======================================================================================
+
+KJ_TEST("compileEvalFunction in synthetic module works") {
```
fhanau, line 945: I unfortunately didn't fully understand what this test does, can you add more comments?

src/workerd/jsg/modules-new-test.c++
```diff
@@ -0,0 +1,1889 @@
...
+
+KJ_TEST("Attaching a module registry works") {
+  PREAMBLE(([&](Lock& js) {
+    ResolveObserver resolveObserver;
+    CompilationObserver compilationObserver;
+    ModuleRegistry::Builder registryBuilder(resolveObserver);
+
+    ModuleBundle::BundleBuilder bundleBuilder;
+    kj::String source = kj::str("export default 123; export const m = 'abc';");
+    // Done this way to avoid including the nullptr at the end...
```
fhanau, line 791: Using slice here looks correct, but do we not have syntactic sugar for this? Would directly declaring the variable as an Array be an option?

src/workerd/jsg/setup.h
jasnell: Since the comments are documenting two different methods, I think there's value in keep them, even if redundant

src/workerd/jsg/modules-new-test.c++
jasnell: Copy'n'paste error

src/workerd/jsg/modules-new-test.c++
jasnell: I agree but I prefer keeping them separate as it helps keep each individual test more self-contained.

src/workerd/jsg/modules-new-test.c++
jasnell: I know, I just prefer to be more explicit

src/workerd/jsg/modules-new-test.c++
jasnell: Probably? Given that it's just a test I didn't really want to worry about that too much.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+    return kj::none;
+  }
+
+  kj::Table<Entry, kj::HashIndex<EntryCallbacks>,
+                   kj::HashIndex<ContextCallbacks>,
+                   kj::HashIndex<UrlCallbacks>> lookupCache;
+  friend class SyntheticModule;
+};
+
+v8::MaybeLocal<v8::Value> SyntheticModule::evaluationSteps(
```
fhanau, line 503: Nit: Would moving evaluationSteps into the class definition be cleaner? This can stay as-is if you think having it around the other big functions here makes more sense.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+}
+
+class EsModule final : public Module {
+public:
+  explicit EsModule(Url specifier, Type type, Flags flags, kj::Array<const char> source)
+      : Module(kj::mv(specifier), type, flags | Flags::ESM),
+        source(kj::mv(source)),
+        ptr(this->source),
+        externed(false),
+        cachedData(nullptr) {
```
fhanau, line 58: Nit: `cachedData` is set to nullptr in one constructor but not the other one – with it also being set to nullptr in its definition I guess doing so here is not needed.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+      // embedders reject import attributes and types they do not understand/implement.
+      // This is because import attributes can alter the interpretation of a module.
+      // Throwing an error for things we do not understand is the safest thing to do
+      // for backwards compatibility.
+      //
+      // For now, we do not support any import attributes, so if there are any at all
+      // we will reject the import.
+      if (!import_assertions.IsEmpty()) {
+        if (import_assertions->Length() > 0) {
+          auto error = js.typeError("Import attributes are not supported");
```
fhanau, line 615: Nit: This pattern to return a rejected promise with an error message appears 3 times in this function – would it make sense to define a lambda for it?

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+      KJ_IF_SOME(found, lock->aliases.find(context.specifier)) {
+        return kj::Maybe<Module&>(*found);
+      }
+    }
+
+    {
+      auto lock = cache.lockExclusive();
+      KJ_IF_SOME(resolved, callback(context)) {
+        Module& module = *resolved;
+        lock->storage.upsert(context.specifier.clone(), kj::mv(resolved));
```
fhanau, line 784: Conceptual question: Is the clone() call needed for memory safety here or is it just based on how URL works? I noticed that Url::clone takes an option parameter that influences cloning, but it is not being used here.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+  };
+
+  auto filter = ([&] {
+    switch (builder.type()) {
+      case Module::Type::BUILTIN: return ModuleType::BUILTIN;
+      case Module::Type::BUILTIN_ONLY: return ModuleType::INTERNAL;
+      case Module::Type::BUNDLE: break;
+      case Module::Type::FALLBACK: break;
+    }
+    KJ_UNREACHABLE;
```
fhanau, line 889: So... we get to unreachable for non-builtin modules? And we check that `module.getType()` is equal to the lambda? Can't we just check if it is equal to BUILTIN or BUILTIN_ONLY?

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+    Lock& js,
+    kj::StringPtr code,
+    kj::StringPtr name,
+    kj::Maybe<JsObject> compileExtensions,
+    const CompilationObserver& observer) {
+  auto metrics = observer.onScriptCompilationStart(js.v8Isolate, name);
+  v8::ScriptOrigin origin(js.str(name));
+  v8::ScriptCompiler::Source source(js.str(code), origin);
+  auto fn = ([&] {
+    KJ_IF_SOME(ext, compileExtensions) {
```
fhanau, line 1356: Off-topic: I wish kj::Maybe had something like `unwrap_or` that would allow us to avoid having two check() calls here. Probably worth adding in the future.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+      }
+
+      KJ_IF_SOME(parent, maybeParent) {
+        return parent.resolve(context);
+      }
+
+      metrics->notFound();
+      return kj::none;
+    }
+    case ResolveContext::Type::BUILTIN_ONLY: {
```
fhanau, line 1142: How frowned upon are case fallthroughs nowadays? The steps for BUILTIN_ONLY are a suffix of those for BUILTIN, so we could merge them if we find a fallthrough acceptable and add a comment stating that it's intentional.

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+
+JsValue ModuleRegistry::resolve(
+    Lock& js,
+    kj::StringPtr specifier,
+    kj::StringPtr exportName,
+    ResolveContext::Type type,
+    ResolveContext::Source source,
+    kj::Maybe<const Url&> maybeReferrer) {
+  auto& bound = IsolateModuleRegistry::from(js.v8Isolate);
+  auto url = ([&]{
```
fhanau, line 1170: Based on my question about recursive aliasing above: Is that needed here too?

src/workerd/jsg/modules-new.c++
```diff
@@ -0,0 +1,1401 @@
...
+  bundles_[kFallback] = builder->bundles_[kFallback].releaseAsArray();
+}
+
+void ModuleRegistry::attachToIsolate(
+    Lock& js,
+    const CompilationObserver& observer) {
+  // We use new here because it is required for the v8 integration.
+  // Responsibility for cleaning up the type is passed to v8 when
+  // the context is cleaned up.
+  new IsolateModuleRegistry(js, *this, observer);
```
fhanau, line 1082: The unwrapped `new` here feels scary, but we keep track of this through the `SetAlignedPointerInEmbedderData` call in the `IsolateModuleRegistry` constructor right?

jasnell: Yes, this is a v8 requirement. V8 takes ownership of the pointer and destroys it when necessary.

fhanau:
LGTM otherwise! 😌 
I didn't really see correctness issues in the implementation, perhaps the prior comments were already useful for that although we might find some later on. Looks well-designed and coherent overall.

My only ask would be to clean up the `foo.slice(0, foo.size())` pattern in the tests I mentioned by using kj::Array/ArrayPtr directly if possible. It appears around 20 times and should be somewhat amenable to Find-Replace. If it feels like something that would slow down the shipping process let's add a TODO for now.

src/workerd/jsg/modules-new.c++
jasnell: It's here because the implementation depends on the definition of `IsolateModuleRegistry` which comes after `SyntheticModule`

src/workerd/jsg/modules-new.c++
jasnell: A bit of both. The lifetimes of these things are a bit different so cloning is the safest option.

src/workerd/jsg/modules-new.c++
jasnell: we could. I kind of prefer this one tho. Works either way I suppose. For now I'll keep it as is.

src/workerd/jsg/modules-new.c++
jasnell: For now, I think I'd rather keep these separate even if there's a bit of duplication. We can come back and de-dup in a later pass.

src/workerd/jsg/modules-new.c++
jasnell: Not here. We'll leave it to individual `ModuleBundle`s to deal with aliases.

jasnell:
@fhanau .. I believe I've addressed all the feedback thus far.

fhanau (approved):
LGTM ✅

PR #1554, opened 2024-01-18
Update following a handful of v8 pending deprecations

jasnell:


fhanau:
Can you also create an internal PR for this? Some deprecation warnings were difficult to fix without breaking things in the past (see https://github.com/cloudflare/workerd/pull/716, although that change is not included here).

src/workerd/jsg/modules.c++
```diff
@@ -433,7 +432,7 @@ v8::Local<v8::Module> createSyntheticModule(
   return v8::Module::CreateSyntheticModule(
       js.v8Isolate,
       v8StrIntern(js.v8Isolate, name),
-      exportNames,
```
fhanau, line 436: Since we no longer need to pass a std::vector to the V8 interface, have you considered eliminating it here or replacing it with a KJ equivalent?

fhanau (approved):
LGTM otherwise, thank you for fixing these warnings.

src/workerd/jsg/modules.c++
jasnell: yeah, we probably can. Will do that separately tho.

jasnell:
@fhanau :
> Can you also create an internal PR for this...

Internal PR CI is green.

PR #1555, opened 2024-01-19
Python: Set up read only file system backed by packages tar

hoodmane:
This is a startup performance optimization. Instead of copying the data around, this uses the tar file in place as the backing for the files.


samples/pyodide/worker.py
```diff
@@ -3,3 +3,7 @@
 
 def fetch(request):
   return Response.new("hello world")
+
+def test():
+  import aiohttp
+  print("hi")
```
dom96, line 9:
```suggestion
```

src/pyodide/BUILD.bazel
```diff
@@ -1,7 +1,42 @@
...
+
+cc_capnp_library(
+    name = "pyodide_packages_capnp",
+    srcs = ["generated/pyodide_packages.capnp"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":packages_capnp_file_embed",
+        ":pyodide_packages_archive_embed",
+    ],
+)
```
dom96, line 39: Are these changes necessary? From what I can tell you should be able to call `getPyodideEmbeddedPackages` in `PackagesTarReader`. Would avoid the need for defining a new capnp library manually.

src/pyodide/internal/pyodide-bootstrap.js
```diff
@@ -175,15 +158,30 @@ export default {
...
+            canOwn: true,
+          });
+        }
+      }
+      await pyodide.pyimport(metadata.mainModule).test();
+    } catch (e) {
+      console.log(e);
+      console.log(e.stack);
+    }
+  },
```
dom96, line 186: I guess this is just testing code that can be removed.

```suggestion

```

src/pyodide/internal/tar.js
```diff
@@ -0,0 +1,83 @@
+import { default as Reader } from "pyodide-internal:packages_tar_reader";
+
```
dom96, line 2: Is the code in this file based on any libraries/specs we could link to for future reference?

dom96: Also it may be worth writing about the limitations of this tar parser (if there are any).

src/pyodide/internal/tarfs.js
```diff
@@ -0,0 +1,86 @@
...
+
+export function createTarFS(Module) {
+  const FS = Module.FS;
+  const TARFS = {
+    mount(mount) {
+      return TARFS.createNode(null, "/", mount.opts.info);
+    },
+    createNode(parent, name, info) {
+      let mode = info.mode;
+      const isDir = !!info.children;
```
dom96, line 11: Nitpicking: could you be more specific with your check here to make it clearer? i.e. `info.children !== undefined`.

src/pyodide/internal/tar.js
```diff
@@ -0,0 +1,83 @@
...
+    mode,
+    size,
+    modtime,
+    type,
+    parts: [],
+    children: undefined,
+  };
+}
+
+export function tarInfo() {
```
dom96, line 38: Nitpicking so feel free to ignore.
```suggestion
export function parseTarInfo() {
```

src/workerd/api/pyodide/pyodide.h
```diff
@@ -2,15 +2,39 @@
 
 #include <kj/common.h>
 #include <pyodide/pyodide.capnp.h>
+#include <pyodide/generated/pyodide_packages.capnp.h>
 #include <workerd/util/autogate.h>
 
 namespace workerd::api::pyodide {
 
+// A special binding object that allows for dynamic evaluation.
```
dom96, line 10: Update docs

dom96 (approved):
This is really cool!

How much performance improvement does it give us?

Before merging make sure you create an upstream PR to run our test suite against these changes. Also see my comments for some suggestions/questions.

src/pyodide/internal/tar.js
hoodmane: I think I just read the wikipedia page on the tar format:
https://en.wikipedia.org/wiki/Tar_(computing)#File_format
I will add a comment with a link to it.

src/pyodide/BUILD.bazel
hoodmane: I didn't really like the vibes of sticking it in a list of packages with a `moduleName` when it isn't intended to be imported as a module, and then we have to iterate the BUNDLE to fish it back out. It seems like a hack. Why not just put it in its own separate capnp variable?

src/workerd/api/pyodide/pyodide.h
```diff
@@ -2,15 +2,40 @@
...
+class PackagesTarReader: public jsg::Object {
+public:
+  PackagesTarReader() = default;
+
+  void read(jsg::Lock& js, int offset, kj::Array<kj::byte> buf) {
+    auto x = *PYODIDE_PACKAGES_TAR;
+    kj::byte* start = buf.begin();
+    for (int i = 0; i < buf.size(); i++) {
+      start[i] = x[offset + i];
+    }
```
jasnell, line 21: Why not use `memcpy`? And for safety should this check that `offset` is within range?

hoodmane: How's it look now?

src/pyodide/internal/pyodide-bootstrap.js
hoodmane: I think it's kind of useful to have something like this to handle a `test` entrypoint (I've certainly been using it a lot). But I'll  remove this here and add something more carefully considered in #1566.

Issue #1521, opened 2024-01-04
Consider adding `self.location` to the worker global scope.


johnspurlock:
A common way for JS libraries to check if they are running in a worker is: `typeof importScripts == 'function'`.

Once this passes, they assume other web-standard API like [`self.location`](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/location) is defined.

Perhaps workerd could use a blob url as the `location` here instead of leaving it undefined?

jasnell:
Not sure there's really sufficient use case justification for this. How would you expect to use it? I'll note that none of the other primary non-browser runtimes (node.js, deno, bun, etc) implement `self.location`.

johnspurlock:
I ran into emscripten js boilerplate that does a "am I running in a web worker" check using `typeof importScripts = 'function'`.  That actually passes on CF workers, but later on when it attempts to use `self.location` (unchecked, assuming it's there since `importScripts` was), it crashes.  It just uses the `.href` for some default location probing logic.

I have a PR out for that on their side - honestly I was surprised `importScripts` was defined at all in CF workers.  Does anyone use it?  Removing `importScripts` would also fix this particular issue.

hoodmane:
For Emscripten feature detection I've been doing:
```js
    // Force Emscripten to feature detect the way we want
    // They used to have an `environment` setting that did this but it has been
    // removed =(
    globalThis.window = {};       // makes ENVIRONMENT_IS_WEB    = true
    globalThis.importScripts = 1; // makes ENVIRONMENT_IS_WORKER = false
    const p = createEmscriptenModule(emscriptenSettings);
    delete globalThis.window;
    delete globalThis.importScripts;
    const emscriptenModule = await p;
```

johnspurlock:
Here's the PR over there in case anyone wants to +1: https://github.com/emscripten-core/emscripten/pull/21010

For workerd tho, I'm still curious why `importScripts` exists - never seen a single cloudflare worker example where it is utilized

hoodmane:
It's also not implemented:
```
jsg::Unimplemented importScripts(kj::String s) { return {}; };
```
> Use Unimplemented as the return type of a method to mark the whole method unimplemented.
> Attempts to use the API will throw an exception.

So removing it should be pretty harmless...

jasnell:
I'd be fine with removing `importScripts(...)` /cc @kentonv

jasnell:
[...]

PR #1556, opened 2024-01-19
Remove explicitly unimplemented importScripts

jasnell:
We have no plans to ever implement `importScripts()` and having it defined but marked explicitly non-implemented is causing issues with using certain third-party libraries in workers. So let's just remove it entirely.

Refs: https://github.com/cloudflare/workerd/issues/1521

### DIFF
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -133,12 +133,14 @@ class WorkerGlobalScope: public EventTarget, public jsg::ContextGlobal {
 public:
   jsg::Unimplemented importScripts(kj::String s) { return {}; };
 
-  JSG_RESOURCE_TYPE(WorkerGlobalScope) {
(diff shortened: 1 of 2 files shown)


jasnell:
@hoodmane ... once this lands it'll be nice to see if we can remove the workaround you added to the pyodide stuff.

kentonv:
Hmm given that we know there exist libraries out there that explicitly check for the existence of this and change behavior based on it, does this need a compat flag? If _all_ such libraries end up broken as a result, then I don't think we need a compat flag, since we generally don't consider it a "breaking change" to break something that was already broken. But if there's some code out there which might work now (maybe my accident) and this causes it to change to non-working, then we need a flag...

jasnell:
Updated to add the compat flag.

kentonv (approved):

kentonv:
Don't forget to update the docs.

jasnell:
Docs: https://github.com/cloudflare/cloudflare-docs/pull/12618

PR #1559, opened 2024-01-19
[nfc] Update bazel to 7.0.1, clean up git-based dependencies and related documentation

fhanau:
Upstream PR to follow. I don't feel entirely sure about dropping shallow_since but if I interpret https://github.com/bazelbuild/bazel/issues/12857 correctly it will only be faster in edge cases while being slower otherwise as `--depth 1` is not being used.


fhanau:
CC @jasnell here's the documentation cleanup based on v8 no longer being fetched using git.

WORKSPACE
```diff
@@ -172,11 +172,10 @@ http_archive(
 #   old for tcmalloc. Absurdly, Bazel simply ignores later attempts to define the same repo name,
 #   rather than erroring out. Thus this leads to confusing compiler errors in tcmalloc complaining
 #   that ABSL_ATTRIBUTE_PURE_FUNCTION is not defined.
-new_git_repository(
```
Warfields, line 175: What's the difference between these sources?

fhanau: There was a difference in the past – `new_git_repository()` supports a `build_file` argument for importing a repository that does not have a pre-existing bazel build file. However, `git_repository()` has been updated as of bazel 6.1.0 to also have that argument so that only one function is needed. `new_git_repository()` is set to be removed in a future version, see https://github.com/bazelbuild/bazel/pull/16064 for details. Currently both versions are supported, but only using one of them looks cleaner.

jasnell (approved):

PR #1560, opened 2024-01-20
Refactor jsg::V8StackScope into callback-based API

jasnell:
Defining `jsg::V8StackScope` as a stack is problematic as it does not guarantee that the stack start will be set correctly in v8. This commit refactors the API to use a callback pattern instead.

```
jsg::runInV8Stack([&](jsg::V8StackScope& stackScope) {
  // ...
});
```

See the primary change in `jsg.h` .. the rest is updating the usage sites.


src/workerd/jsg/jsg.h
```diff
@@ -2271,27 +2271,40 @@ class Lock {
...
-  // oddities of our internal environment which do not apply to workerd. For workerd, V8's default
-  // behavior is just fine.
-  v8::StackStartMarker v8Marker;
+    // This currently depends on a V8 patch which hasn't been upstreamed. Note that workerd does
+    // not use this patch; it's only used internally. The patch is needed in order to work around
+    // oddities of our internal environment which do not apply to workerd. For workerd, V8's default
+    // behavior is just fine.
+    v8::StackStartMarker marker(pos);
 #endif
+    // We create a V8StackStack only as proof that we are running in the V8 stack.
```
harrishancock, line 2293:
```suggestion
    // We create a V8StackScope only as proof that we are running in the V8 stack.
```

src/workerd/jsg/jsg.h
```diff
@@ -2271,27 +2271,40 @@ class Lock {
...
+    V8StackScope stackScope;
+    return callback(stackScope);
+  }
+
+  friend auto runInV8Stack(auto callback);
+};
+
+// Ensures that av8::StackStartMarker is allocated on the stack before calling the callback.
```
harrishancock, line 2301:
```suggestion
// Ensures that a v8::StackStartMarker is allocated on the stack before calling the callback.
```

harrishancock (approved):

src/workerd/io/worker.c++
```diff
@@ -758,26 +759,22 @@ struct Worker::Script::Impl {
...
-      Worker::Lock lock(*worker, asyncLock);
+  struct DynamicImportResult {
+    jsg::Value value;
+    bool isException = false;
+  };
+  using DynamicImportHandler = kj::Function<jsg::Value()>;
+
+  static DynamicImportResult getDynamicImportResult(
+      Worker::AsyncLock& asyncLock,
+      auto& worker,
```
kentonv, line 770: Does this `auto&` always resolve to `const Worker&`? If so we should really just write out the type. `auto`-typed function arguments turn the function into a template which, among other things, breaks IDE auto-completion, so should generally be avoided.

src/workerd/io/worker.c++
```diff
@@ -758,26 +759,22 @@ struct Worker::Script::Impl {
...
-      auto asyncLock = co_await worker->takeAsyncLockWithoutRequest(nullptr);
-
-      Worker::Lock lock(*worker, asyncLock);
+  struct DynamicImportResult {
+    jsg::Value value;
+    bool isException = false;
+  };
+  using DynamicImportHandler = kj::Function<jsg::Value()>;
+
+  static DynamicImportResult getDynamicImportResult(
```
kentonv, line 768: I don't understand why this was refactored. Is it to avoid allocating `Worker::Lock` as a local variable in a coroutine? That was indeed a problem, but it looks like the change to use `jsg::runInV8Stack()` actually fixes the problem already -- the `Worker::Lock` is now in the inner callback, which is never a coroutine. So it seems like refactoring these functions wasn't really necessary? I worry that the refactoring makes the code a bit harder to read and adds some risk of breakage.

src/workerd/io/worker.c++
```diff
@@ -2217,30 +2235,36 @@ public:
     return ioHandler.messagePump();
   }
 
+  void handleDispatchProtocolMessage(
```
kentonv, line 2238: Here too, it seems like this was refactored to avoid allocating `jsg::V8StackScope` on a coroutine frame -- but changing to `jsg::runInV8Stack` actually solves that problem on its own, so the surrounding refactoring seems unnecessary?

src/workerd/server/server.c++
```diff
@@ -1532,8 +1532,18 @@ public:
             // promise. If a new request comes in while we're waiting to get the lock then we will
             // cancel this promise.
             Worker::AsyncLock asyncLock = co_await worker.takeAsyncLockWithoutRequest(nullptr);
-            Worker::Lock lock(*workerStrongRef, asyncLock);
-            m->hibernateWebSockets(lock);
+
+            static constexpr auto handle = [](
```
kentonv, line 1536: Unnecessary refactoring?

src/workerd/io/worker.c++
jasnell: I disagree. For myself, the slight refactoring on these makes it clearer that something is going on that I need to pay extra attention to, which I think is ideal. Unless there is something technically wrong with the refactor I prefer to keep these as is.

PR #1566, opened 2024-01-22
Get fastapi working the rest of the way

hoodmane:
I removed the C++ functions `getPyodideLock()`, `getPyodidePatches()`, and `generatePyodidePatches()` in favor of using the import system normally.


jasnell (approved):

src/pyodide/pyodide-bootstrap.js
```diff
@@ -0,0 +1,19 @@
...
+  },
+  async test() {
+    try {
+      const pyodide = await loadPyodide();
+      const mainModule = await setupPackages(pyodide, getMetadata());
+      return await mainModule.test();
+    } catch (e) {
+      console.warn(e);
+    }
+  },
```
dom96, line 18: is this something you want to keep? if so curious why

src/pyodide/pyodide-bootstrap-helper.js
```diff
@@ -120,68 +104,59 @@ function transformMetadata(metadata) {
...
-      }
-      await pyodide.pyimport(metadata.mainModule).test();
-    } catch (e) {
-      console.log(e);
-      console.log(e.stack);
-    }
-  },
-};
+  if (micropipRequirements.length > 0) {
+    const micropip = pyodide.pyimport("micropip");
```
dom96, line 138:
```suggestion
    // Micropip and ssl packages are pre-loaded via the packages tarball. This means
    // we should be able to load micropip directly now.
    const micropip = pyodide.pyimport("micropip");
```

src/pyodide/pyodide-bootstrap-helper.js
```diff
@@ -50,39 +48,25 @@ function initializePackageIndex(pyodide, lockfile) {
 // These packages are currently embedded inside workerd and so don't need to
 // be separately installed.
 const EMBEDDED_PYTHON_PACKAGES = [
-  "tqdm",
-  "openai",
```
dom96, line 54: I'm not sure about other packages but this one is critical. Did you remove these packages from the tarball? If so the langchain test won't work.

src/pyodide/pyodide-bootstrap-helper.js
```diff
@@ -1,7 +1,5 @@
-import { loadPyodide } from "pyodide:python";
-import { getMetadata } from "pyodide:current-bundle";
-import { lockFile } from "pyodide:package-lock.json";
-import { getPatches } from "pyodide:patches";
+export { loadPyodide } from "pyodide-internal:python";
+import { lockFile } from "pyodide-internal:pyodide-lock";
```
dom96, line 2: Since you're changing this anyway: i've been wondering why we include this file separately at all, it's part of the pyodide tarball, so why don't we just grab it from there?

src/pyodide/internal/asgi.py
```diff
@@ -0,0 +1,176 @@
...
+    finally:
+        buf.release()
+
+
+def request_to_scope(req, ws=False):
+    from js import URL
+
+    headers = [tuple(x) for x in req.headers]
+    url = URL.new(req.url)
+    scheme = url.protocol[:-1]
```
dom96, line 26: I'd add a couple asserts in here

```suggestion
    assert url.protocol[-1] == ':'
    scheme = url.protocol[:-1]
```

src/pyodide/internal/asgi.py
```diff
@@ -0,0 +1,176 @@
...
+    path = url.pathname
+    query_string = url.search[1:].encode()
```
dom96, line 28:
```suggestion
    assert url.search[0] == '?'
    query_string = url.search[1:].encode()
```

src/pyodide/internal/asgi.py
```diff
@@ -0,0 +1,176 @@
...
+    return Response.new(None, status=101, webSocket=client)
+
+
+async def fetch(app, req):
+    await start_application(app)
+    return await process_request(app, req)
+
+
+async def websocket(app, req):
+    # await start_application(app)
```
dom96, line 175: Should this be uncommented?

samples/pyodide/worker.py
```diff
@@ -1,5 +1,60 @@
-from js import Response
+async def fetch(request):
```
dom96, line 1: Can you move this into a separate samples folder? something like `pyodide_fastapi`. It would be nice to keep the main sample as simple as possible.

src/pyodide/pyodide-bootstrap.js
hoodmane: Yeah I think this is useful to keep, for the same reason that the ES6 test entrypoint is useful: it is really helpful to be able to test functionality you use in your `fetch` handler without having to restart the server and refresh `localhost:8080` every time.

src/pyodide/pyodide-bootstrap-helper.js
```diff
@@ -120,68 +104,59 @@ function transformMetadata(metadata) {
...
+    await micropip.install(micropipRequirements);
+  }
+
+  // Apply patches that enable some packages to work.
+  if (requirements.includes("aiohttp")) {
+    const mod = await import("pyodide-internal:patches/aiohttp_fetch_patch.py");
+    pyodide.FS.writeFile(
+      "/lib/python3.11/site-packages/aiohttp_fetch_patch.py",
+      new Uint8Array(mod.default),
+      { canOwn: true }
```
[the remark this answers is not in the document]
hoodmane: Emscripten keeps the file data in a buffer. If `canOwn` is `true`, then Emscripten will just mount the buffer you give it as a file so this operation will not actually copy the data. In all of these cases, we don't use the buffer again after calling `writeFile` so setting `canOwn` is appropriate.

src/pyodide/pyodide-bootstrap-helper.js
hoodmane: Yeah since langchain doesn't have a direct requirement for openai `pip install langchain fastapi` misses `openai`. So I guess you need `pip install langchain openai fastapi` to get all of these.

src/pyodide/pyodide-bootstrap-helper.js
hoodmane: At an earlier point I had some hot patches for Pyodide bugs and they required changes to this file. All the changes got upstreamed into v0.25.0a2 though so it's no longer necessary.

hoodmane: I'll plan to remove this in a followup (can probably be combined with update to Pyodide 0.25.0)

src/pyodide/python-entrypoint-helper.js
```diff
@@ -120,51 +115,61 @@ function transformMetadata(metadata) {
...
+  if (requirements.includes("aiohttp")) {
+    const mod = await import("pyodide-internal:patches/aiohttp_fetch_patch.py");
+    pyodide.FS.writeFile(
+      "/lib/python3.11/site-packages/aiohttp_fetch_patch.py",
+      new Uint8Array(mod.default),
+      { canOwn: true }
+    );
+    pyodide.pyimport("aiohttp_fetch_patch");
+  }
+  if (requirements.includes("fastapi")) {
```
hoodmane, line 165: @dom96 these changes may have messed up your langchain demo depending on whether they have `aiohttp` as a `pythonRequirement`.

I think `micropip.freeze().has("aiohttp")` will always return `true` because we are unconditionally exposing it to the filesystem. We could for now just unconditionally load the aiohttp patch but that will be a performance hit when it isn't needed.

Before we have any sort of beta we'll definitely have to ensure that we don't expose any extra packages that user didn't explicitly request. Another thing to discuss is whether the `pythonRequirement` list needs to include all transitive dependencies or just leafs.

dom96: `Aiohttp` is not one of the `pythonRequirement`s, it is just `langchain` and `openai`, so it definitely expects the transitive dependencies to be resolved and installed implicitly.

I think that for the demo listing all the transitive dependencies would be too painful, but for the long-term it may make sense to do so. How about for now you check for `langchain` or `openai` in the dependencies and if they are there include the aiohttp patch? You can add a TODO in there so we remember to remove this later.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -50,39 +52,32 @@ function initializePackageIndex(pyodide, lockfile) {
...
-  "langsmith",
-  "dataclasses_json",
-  "jsonpatch",
-  "requests",
-  "sniffio",
-  "marshmallow",
-  "urllib3",
-  "typing_inspect",
-  "jsonpointer",
-  "mypy_extensions",
```
dom96, line 83: I'm still concerned about these, I guess you stripped these to save space? are they definitely not necessary for the demo?

hoodmane: I think these are actually in the `pyodide_packages.tar.zip` file now (so they can be imported) I just forgot to update the list here. 

The reason they went away originally was actually because of a divergence in behavior between pip and micropip because `platform.machine()` was returning the host value in pip:
https://github.com/pyodide/pyodide/pull/4412/files

langchain 0.0.4 added a dependency on sqlalchemy >= 2 which has a dependency on greenlet on x86_64 but not on wasm32. So when `platform.machine()` returned `x86_64` instead of `wasm32` it wanted greenlet, but greenlet cannot be supported without stack switching and has no available candidate on Emscripten, so instead it installed langchain 0.0.3. Anyways with that fixed it now successfully installs  langchain 0.1.3.

But langchain does download a whole ton of dependencies that are needed for only a specific subpackage that we don't use.

PR #1569, opened 2024-01-23
Automatically run webgpu tests if enable_experimental_webgpu is enabled

fhanau:
The WebGPU tests were previously not included in the bazel build system. With this change we have 5 more tests that will run on CI and achieve better code coverage.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,6 +1,7 @@
 common --enable_platform_specific_config
 build --verbose_failures
 build --build_tag_filters=-off-by-default
+test --test_tag_filters=-off-by-default
 
 # Not using bzlmod for dependencies yet
 common --noenable_bzlmod
--- a/build/wd_test.bzl
+++ b/build/wd_test.bzl
@@ -9,7 +9,8 @@ def wd_test(
     Args:
      src: A .capnp config file defining the test. (`name` will be derived from this if not
         specified.) The extension `.wd-test` is also permitted instead of `.capnp`, in order to
-        avoid confusing other build systems that may assume a `.capnp` file should be complied.
+        avoid confusing other build systems that may assume a `.capnp` file should be compiled. As
+        an extension, `.gpu-wd-test` is supported to enable special handling for GPU tests.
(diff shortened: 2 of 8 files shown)


fhanau:
@edevil Looks like all tests are failing on Linux and one is failing on macOS. I was hoping these would work out of the box since all tests are passing locally on mac... If you can take a look at the Linux issues we can probably enable all/some of these.

edevil:
I haven't looked at the mac one yet, but the linux ones fail because we couldn't find a compatible Vulkan driver installed in the system:
```
Warning: vkCreateInstance failed with VK_ERROR_INCOMPATIBLE_DRIVER
```
Are there any CI Linux images that have a working Vulkan installation? In my Linux dev VM I can use a software-only rasterizer via LLVMpipe and mesa-vulkan-drivers.

fhanau:
> Vulkan

Looks like this is possible, but perhaps prohibitively slow: https://github.com/actions/runner-images/issues/2998
I can probably add a `needs-webgpu` tag here that would allow us to only run this on CI on Windows for now, so we can run the tests on at least some CI jobs.

edevil:
The mac failure is a bit weird. We render an image to a texture, copy it to a buffer, hash the content and we compare the hash value with a known good value previously calculated in Chrome.

The hash value is correct in my Linux VM (vulkan), which uses LLVMpipe, and in my Mac (metal) laptop that uses the M2 chip. It also seems to be calculated correctly in the windows (D3D) CI. But in the MacOS (metal) CI it is using a virtual GPU "Apple Paravirtual device" and seems to calculate the wrong value. I'll have to see if I can somehow simulate that environment in my laptop but whatever it is should take some time to debug.

edevil:
> Looks like this is possible, but perhaps prohibitively slow: [actions/runner-images#2998](https://github.com/actions

Do you mean slow to install the tools, or slow to run the gpu workload? Because this test workload is very simple, it should run fast even on emulated GPUs.

dom96 (approved):
Nice! lgtm but based on the discussion we probably want to update this to only run on Windows for now

edevil (approved):

PR #1573, opened 2024-01-23
Increase size limit for inspector protocol messages.

### LINKED ISSUE #1563: devtools/vscode debugging fails kj/compat/http.c with Message is too large (>1MB)
Using either the remix cloudflare workers or pages template out of box on WSL2 linux.  Trying to debug locally, setting a breakpoint.  Breakpoint is hit, but then emits the error listed below, after which the session is broken.  Chrome devtools and vscode run into the same issue.

After investigating, it seems to have to have to do with the size of certain chunks that remix emits, as well as map files, which are clearly greater than 1MB.  Apparently these are sent via WebSocket?  Minifying the build reduced that file to under 300K, and debugging breakpoints works again.

Can't find a way to ask remix to chunk that piece any smaller.  Would it make sense when running locally to allow larger messages to improve the worker/pages developer experience?  Even if behind an option/flag?

✘ [ERROR] workerd/jsg/util.c++:281: error: e = kj/compat/http.c++:5618: failed: WebSocket protocol error; protocolError.statusCode = 1009; protocolError.description = Message is too large: **2157260** > **1048576**

  stack: /home/david/github/plg-perf-app-cfp/node_modules/@cloudflare/workerd-linux-64/bin/workerd@35bf930 /home/david/github/plg-perf-app-cfp/node_modules/@cloudflare/workerd-linux-64/bin/workerd@35bfb7a /home/david/github/plg-perf-app-cfp/node_modules/@cloudflare/workerd-linux-64/bin/workerd@48f24cf /home/david/github/plg-perf-app-cfp/node_modules/@cloudflare/workerd-linux-64/bin/workerd@2c10ba0 /home/david/github/plg-perf-app-cfp/node_modules/@cloudflare/workerd-linux-64/bin/workerd@35dddc0 /home/david/github/plg-perf-app-cfp/node_modules/@cloudflare/workerd-linux-64/bin/workerd@2c10ba0 /home/david/github/plg-perf-app-cfp/node_modules/@cloudflare/workerd-linux-64/bin/workerd@2e1932c; sentryErrorContext = jsgInternalError

kentonv:
Since this is for debugging only, we don't really need to be so strict.

Fixes #1563.

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -2395,7 +2395,7 @@ private:
 
     kj::Promise<void> receiveLoop() {
      for (;;) {
-        auto message = co_await webSocket.receive();
+        auto message = co_await webSocket.receive(MAX_MESSAGE_SIZE);
         KJ_SWITCH_ONEOF(message) {
           KJ_CASE_ONEOF(text, kj::String) {
             incomingQueue.lockExclusive()->messages.add(kj::mv(text));
@@ -2455,6 +2455,11 @@ private:
     kj::WebSocket& webSocket;                 // only accessed on the InspectorService thread.
     std::atomic_bool receivedClose;           // accessed on any thread (only transitions false -> true).
     kj::Maybe<InspectorChannelImpl&> channel; // only accessed on the isolate thread.
+
+    // Sometimes the inspector protocol sends large messages. KJ defaults to a 1MB size limit
+    // for WebSocket messages, which makes sense for production use cases, but for debug we should
+    // be OK to go larger. So, we'll accept 128MB.
+    static constexpr size_t MAX_MESSAGE_SIZE = 128u << 20;
   };
 
   WebSocketIoHandler ioHandler;


justin-mp (approved):

jasnell (approved):

tforster:
I am getting the same error albeit with a much larger message size

> Message is too large: 1050289 > 1048576

Is there something I can do in my code to reduce the message size? I can't tell from the workerd error where this occurs.

kentonv:
@tforster You are probably writing a Worker that uses a WebSocket, and you are trying to send or receive a message on that WebSocket which is larger than 1MB. You will need to find a way to split your message up into multiple smaller messages. This is different, though, from the problem being addressed by this PR.

tforster:
Nope, no websockets. But there are outbound HTTP requests.

kentonv:
@tforster This error is specific to WebSockets. It does not apply to other HTTP requests.

Please open a new issue, this isn't related to this PR thread.

PR #1576, opened 2024-01-24
Update pyodide to 0.25.0

hoodmane:
And tidy up loading process:

1. switch Python metadata from a user js module to an internal json module
2. switch lockfile.js to a json file and add capability for builtin json modules


src/workerd/jsg/modules.capnp
```diff
@@ -18,6 +18,7 @@ struct Module {
     src @1 :Data; # JS / TS code
     wasm @4 :Data; # Wasm module
     data @5 :Data; # Binary data module
+    json @6 :Text; # Json module
```
dom96, line 21: This is nice, but I wonder if there is a reason we haven't introduced this in the past. We'll have to add this in edgeworker too I suppose and I want to make sure the team is happy with this direction (just calling it out here so hopefully if someone has objections they can call them out)

src/pyodide/internal/asgi.py
```diff
@@ -23,10 +23,10 @@ def request_to_scope(req, ws=False):
 
     headers = [tuple(x) for x in req.headers]
     url = URL.new(req.url)
-    assert url.protocol[-1] == ':'
+    assert url.protocol[-1] == ":"
     scheme = url.protocol[:-1]
     path = url.pathname
-    assert url.search[0] == '?'
+    assert "?".startswith(url.search[0:1])
```
dom96, line 29: curious about this, the former is far clearer, why the change?

dom96 (approved):

src/pyodide/internal/asgi.py
hoodmane: `url.search` can also be emtpty.

src/workerd/jsg/modules.capnp
hoodmane: Well I didn't introduce this in the wasm/data PRs because I wasn't including `pyodide-package.json` at all then. It seems much preferable to include them as `json` than to try to do some brittle rewrite on the source.

PR #1579, opened 2024-01-26
Add expectedLength for JS streams

jasnell:
Allows setting an `expectedLength` option for JS-backed readable and transform streams. When the readable stream is used as the body of a request or response, the expectedLength is used as the value of the content-length header. When set, chunked encoding will not be used.


kentonv:
I see this is a non-standard extension. Was this discussed somewhere? Any design doc discussing alternatives?

We also have the non-standard `FixedLengthStream` that has a similar use case.

kentonv:
Discussed directly. Seems like a reasonable extension to me. Streams spec really ought to define something like this, but it doesn't.

dom96 (approved):

PR #1580, opened 2024-01-26
Don't embed Python packages in workerd.

dom96:
Just a simple way to disable embedding of Python packages only in workerd to reduce its binary size.

### DIFF
--- a/src/pyodide/python-entrypoint-helper.js
+++ b/src/pyodide/python-entrypoint-helper.js
@@ -12,8 +12,8 @@ function initializePackageIndex(pyodide) {
     );
   }
   const API = pyodide._api;
-  API.config.indexURL = "https://cdn.jsdelivr.net/pyodide/v0.25.0a2/full/";
-  globalThis.location = "https://cdn.jsdelivr.net/pyodide/v0.25.0a2/full/";
+  API.config.indexURL = "https://cdn.jsdelivr.net/pyodide/v0.25.0/full/";
+  globalThis.location = "https://cdn.jsdelivr.net/pyodide/v0.25.0/full/";
   API.lockfile_info = lockfile.info;
   API.lockfile_packages = lockfile.packages;
   API.repodata_packages = lockfile.packages;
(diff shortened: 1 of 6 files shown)


fhanau (approved):

src/pyodide/python-entrypoint-helper.js
```diff
@@ -150,7 +151,8 @@ async function setupPackages(pyodide) {
 
     if (value.pythonRequirement !== undefined) {
       requirements.push(name);
-      if (!EMBEDDED_PYTHON_PACKAGES.includes(name)) {
+      // Packages are not embedded in workerd.
```
hoodmane, line 154: Maybe:
```suggestion
      // Packages are not embedded in workerd.
      // TODO: Improve package loading in workerd
```

hoodmane (approved):
LGTM

hoodmane:
When I run this, it seems that it entirely breaks package loading in workerd. But this is probably acceptable in the short term to get the binary size down. I can manually revert this to do my tests...

dom96:
@hoodmane can you try again?

src/pyodide/python-entrypoint-helper.js
```diff
@@ -153,16 +154,26 @@ async function setupPackages(pyodide) {
...
+      // TODO: Improve package loading in workerd.
+      if (isWorkerd) {
+        micropipRequirements.push(name);
+      }
     }
   }
 
+  if (isWorkerd) {
+    pythonRequirements.push("micropip");
+  }
```
hoodmane, line 167: Perhaps instead of `empty` we should stick micropip and packaging in there. It would significantly improve load time when micropip is needed and I don't think they are very big. But this can happen in a followup.

hoodmane:
I added an extra commit which fixes all the examples.

jasnell (approved):

samples/pyodide-fastapi/config.capnp
```diff
@@ -23,7 +23,8 @@ const config :Workerd.Config = (
 const mainWorker :Workerd.Worker = (
   modules = [
     (name = "worker", pythonModule = embed "./worker.py"),
-    (name = "fastapi", pythonRequirement = "fastapi"),
+    (name = "fastapi==0.103.2", pythonRequirement = ""),
```
dom96, line 26: will merge this, but I don't think we want to support version numbers and operators in here

PR #1581, opened 2024-01-26
Set backpressure by default on JS-backed TransformStream

jasnell:
Sets up backpressure on the readable-side of a JS-backed `TransformStream` by default.


dom96 (approved):
LGTM but is this a breaking change? Something we should put behind a compat flag?

PR #1582, opened 2024-01-27
Implement climem memory monitor service

jasnell:
[`climem`](https://www.npmjs.com/package/climem) is a tool that I've used extensively in the past for diagnosing memory leaks in node.js applications. It provides simple monitoring of v8 memory usage and rss for the application and displays it in a chart in the terminal.

This PR implements support for climem directly within workerd. The workerd portion of it is implemented as a simple TCP server (per isolate) that, when connected, generates a json-serialized log of the current memory stats for the process/isolate. 

To use, specify the `-m localhost` option when running workerd serve. Then install `climem` from npm and run it, pointing at the tcp port that worker tells you the climem server is running on.

It's extremely rudimentary but that's the goal. Keep it simple.

/cc  @mcollina who wrote `climem`

### DIFF
--- a/src/workerd/io/worker.h
+++ b/src/workerd/io/worker.h
@@ -334,6 +334,8 @@ public:
 
   kj::Own<const WeakIsolateRef> getWeakRef() const;
 
+  const jsg::MemStats getCurrentMemStats(jsg::Lock&) const;
+
 private:
   kj::Promise<AsyncLock> takeAsyncLockImpl(
       kj::Maybe<kj::Own<IsolateObserver::LockTiming>> lockTiming) const;
@@ -440,6 +442,10 @@ public:
   // WebCrypto algorithms supported.
   virtual kj::Maybe<const api::CryptoAlgorithm&> getCryptoAlgorithm(kj::StringPtr name) const {
     return kj::none;
+  }
+
+  virtual const jsg::MemStats getCurrentMemStats(jsg::Lock&) const {
+    return {};
   }
 
   // Set the module fallback service callback, if any.
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -18,6 +18,7 @@
 #include <workerd/io/compatibility-date.h>
 #include <workerd/io/io-context.h>
 #include <workerd/io/worker.h>
+#include <workerd/util/mem-utils.h>
 #include <time.h>
 #include <openssl/bio.h>
 #include <openssl/pem.h>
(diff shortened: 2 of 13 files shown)


irvinebroque:
@admah — we should figure out the interface for Wrangler, so that you can say "turn on memory monitoring" and show this output in the terminal

src/workerd/util/mem_utils.c++
```diff
@@ -0,0 +1,37 @@
...
+
+#if !_WIN32
+kj::Maybe<size_t> tryGetResidentSetMemory() {
+  int fd_;
+  KJ_SYSCALL(fd_ = open("/proc/self/stat", O_RDONLY));
+  kj::AutoCloseFd fd(fd_);
+  auto text = kj::FdInputStream(kj::mv(fd)).readAllText();
+  auto p = text.begin();
+  for (kj::uint i = 0; i < 23; i++) {
+    p = strchr(p, ' ');
```
justin-mp, line 22: Drive-by: `kj::String` has a `find` method now, though I don't know if it's any easier to use in this case.

src/workerd/io/worker.h
```diff
@@ -334,6 +334,24 @@ class Worker::Isolate: public kj::AtomicRefcounted {
 
   kj::Own<const WeakIsolateRef> getWeakRef() const;
 
+  struct MemStats {
```
kentonv, line 337: Why is this type declared separately in `Worker` and `jsg::IsolateBase`? Can both places share one type declaration?

In fact, could both just use `v8::HeapStatistics` directly instead of defining our own struct(s)? These stats are very specific to V8 anyway; it's not like a different JS engine is going to export the same statistics. So defining our own parallel structs seems to just add boilerplate IMO.

src/workerd/util/mem_utils.c++
```diff
@@ -0,0 +1,37 @@
...
+
+#if !_WIN32
+#include <fcntl.h>
+#include <stdlib.h>
+#include <unistd.h>
+#endif
+
+namespace workerd::util {
+
+#if !_WIN32
```
kentonv, line 14: This code is Linux-specific since it depends on procfs.

src/workerd/util/mem_utils.c++
kentonv, line 1: Nitpick: We usually use hyphens rather than underscores in file names.

src/workerd/server/server.c++
```diff
@@ -1075,6 +1076,95 @@ private:
...
+                                     KJ_ASSERT_NONNULL(worker), name));
+        }
+
+        co_return;
+      }))();
+
+      kj::NEVER_DONE.wait(io.waitScope);
+    });
+    thread->detach();
+    threads.add(kj::mv(thread));
```
kentonv, line 1149: Since you detached the thread, you can just discard the thread object.

jasnell:
Meant to open this as a draft PR as it's not going to ready to land for a bit. Reviews are welcome but it's probably not going to be a super high priority to finish up.

src/workerd/util/mem_utils.c++
jasnell: Yes, I know... this bit will need to be fully finished to work on macos/windows before the PR lands.

src/workerd/util/mem_utils.c++
jasnell: Will look at that in a future iteration.

PR #1583, opened 2024-01-29
[nfc] dependency updates, expand usage of bazel implementation_deps

fhanau:


build/BUILD.ada-url
```diff
@@ -1,12 +1,9 @@
...
     srcs = ["ada.cpp"],
+    hdrs = [
+        "ada.h",
+        "ada_c.h",
+    ],
     visibility = ["//visibility:public"],
-    include_prefix = ".",
-    copts = ["-w"],
-    defines = [
-      "ADA_SSE2=1"
```
jasnell, line 10: Just need some context... Why are you removing this define?

fhanau: ada-url auto-detects support for SSE2, see https://github.com/ada-url/ada/blob/7ed703c4fddeaaa0178383a86df9f40d34ff2881/include/ada/common_defs.h#L292. SSE2 support is required for x86_64 machines and detected as such, so there is no need to enable it explicitly. I previously thought that this could also cause problems when compiling on arm64 as SSE2 is an x86-specific instruction set, but luckily ada checks for the presence of ADA_NEON before checking for SSE2 (Neon instructions are required for arm64 and also autodetected here).
Note that this also enables warnings for ada-url – disabling warnings can be useful for 3rd-Party dependencies and we do it manually for a few in .bazelrc but I didn't encounter any warnings locally on macOS so it might not be needed here.

dom96 (approved):

build/BUILD.ada-url
jasnell: ah nice. I added this originally because the build had failed originally locally without it but appears to be working with this now. Interesting.

PR #1584, opened 2024-01-29
Implement MemoryTracker for Heap Snapshots

jasnell:
This is adapted from Node.js' implementation.

This extends workerd's v8::HeapSnapshot collection process to allow including more detailed information about internal/native objects.  Best illustrated with a screenshot... 
image

This PR includes the core part of the implementation. It provides a straightforward implementation of `jsg::Object`s / `jsg::Wrappable` snapshot tracing along with a handful of other types. It does *NOT add all of the instrumentation to all of the various objects.

Those familiar with the GC tracing mechanism should find this familiar.

Look at the memory-test.c++ for an example of how it's used.

To keep things as simple as possible, to instrument your typical `jsg::Object` type (anything using `JSG_RESOURCE_TYPE`, you only need to provide an implementation of the `visitForMemoryInfo(...)` method:

```cpp
struct Foo: public Object {
  kj::String bar = kj::str("test");

  JSG_RESOURCE_TYPE(Foo) {}
  
void visitForMemoryInfo(MemoryTracker& tracker) const {
    tracker.trackField("bar", bar);
  }
};
```

This should be familiar to anyone who has implemented support for GC tracing (using the `visitForGc(...)` method).

Any fields you want to include in the heap snapshot are registered using `trackField(...)` or `trackFieldWithSize(...)`. 

If you want anything else to be trackable, then it will need to implement all three of the following methods:

```cpp
kj::StringPtr getMemoryInfoName() const { return "TheName"_kjc; }
size_t getMemoryInfoSelfSize() const { return sizeof(ThisThing); }
void getMemoryInfo(MemoryTracker& tracker) const {
   // register the fields here
}
```

To check to see if a type if trackable, you can test it using something like:

```
static constexpr bool IS_TESTABLE = MemoryRetainer<T>;
```

From the typical day-to-day use there wouldn't be much more than that. There a few more advanced cases to consider and we might want to expand the `trackField(...)` implementations in `memory.h` to add support for various types, but for the most part this was designed to be as minimally intrusive as possible.

### DIFF
--- a/src/workerd/jsg/memory.h
+++ b/src/workerd/jsg/memory.h
@@ -1,0 +1,810 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+// Implements mechanism for incorporating details about the native (c++) objects
+// in a v8 heap snapshot. The design of the API and implementation were heavily
+// influenced by Node.js' implementation of the same feature.
+
+#pragma once
+
+#include <kj/common.h>
+#include <kj/debug.h>
+#include <kj/exception.h>
+#include <kj/hash.h>
+#include <kj/table.h>
+#include <kj/string.h>
+#include <v8-profiler.h>
+#include <v8.h>
+#include <stack>
+#include <string>
+
+namespace workerd::jsg {
+
+class MemoryTracker;
+class MemoryRetainerNode;
+
+template <typename T> class V8Ref;
+template <typename T> class Ref;
+
+enum class MemoryInfoDetachedState {
+  UNKNOWN,
+  ATTACHED,
+  DETACHED,
+};
+
+template <typename T>
+concept MemoryRetainer = requires(const T* a) {
+  std::is_member_function_pointer_v<decltype(&T::jsgGetMemoryInfo)>;
+  std::is_member_function_pointer_v<decltype(&T::jsgGetMemoryName)>;
+  std::is_member_function_pointer_v<decltype(&T::jsgGetMemorySelfSize)>;
+};
+
+template <typename T>
+concept MemoryRetainerObject = requires(T a) {
+  MemoryRetainer<T>;
+  std::is_member_function_pointer_v<decltype(&T::jsgGetMemoryInfoWrapperObject)>;
+};
+
+template <typename T>
+concept MemoryRetainerDetachedState = requires(T a) {
+  MemoryRetainer<T>;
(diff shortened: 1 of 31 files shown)


src/workerd/jsg/memory.h
```diff
@@ -0,0 +1,810 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+// Implements mechanism for incorporating details about the native (c++) objects
+// in a v8 heap snapshot. The design of the API and implementation were heavily
+// influenced by Node.js' implementation of the same feature.
+
```
dom96, line 8: This header is pretty large, can/should a lot of the code in here be moved into a .c++ file?

I wouldn't let that block the merging of this PR though.

src/workerd/jsg/memory.h
```diff
@@ -0,0 +1,810 @@
...
+template <typename T>
+MemoryTracker& MemoryTracker::trackField(
+    kj::StringPtr edgeName,
+    const std::basic_string<T>& value,
+    kj::Maybe<kj::StringPtr> nodeName) {
+  return trackFieldWithSize(edgeName, value.size() * sizeof(T), "std::basic_string"_kjc);
+}
+
+template <typename T, typename test, typename dummy>
```
dom96, line 478: Unused?

```suggestion
template <typename T>
```

src/workerd/jsg/memory.h
```diff
@@ -0,0 +1,810 @@
...
+    } else {
+      graph_->AddEdge(nullptr, found.node, edgeName.cStr());
+    }
+    return *this;
+  }
+
+  return track(value, edgeName);
+}
+
+template <typename T, typename test_for_number, typename dummy>
```
dom96, line 600: Also unused?

```suggestion
template <typename T>
```

dom96 (approved):
This is super cool!

One concern: what overhead does this add? Is it something that we disable by default so it shouldn't affect our performance upstream?

jasnell:
> One concern: what overhead does this add? Is it something that we disable by default so it shouldn't affect our performance upstream?

This code is only engaged when a heap snapshot is taken so the overhead will be minimal.

src/workerd/jsg/memory.h
jasnell: It is used. The additional template parameters create a specialization for when `T` is numeric. Will keep these as is for now.

src/workerd/jsg/memory.h
jasnell: Most of this are inline template method definitions. Will explore ways of simplifying in another iteration.

PR #1585, opened 2024-01-29
Simplify "named intercept"

kentonv:
This implements the change we discussed previously, to allow the named property getter to return an arbitrary wrappable value.

I also came up with what I think is more intuitive name for this feature: "Wildcard property"

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -477,9 +477,9 @@ using HasGetTemplateOverload = decltype(
 
 // Configures the resource type to implement named property interception.
 // @see the definition of jsg::NamedIntercept in resource.h for more information.
-#define JSG_NAMED_INTERCEPT() \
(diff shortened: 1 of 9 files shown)


src/workerd/jsg/jsg.h
```diff
@@ -475,11 +475,24 @@ using HasGetTemplateOverload = decltype(
     wrapper.initReflection(this, __VA_ARGS__); \
   }
 
-// Configures the resource type to implement named property interception.
-// @see the definition of jsg::NamedIntercept in resource.h for more information.
-#define JSG_NAMED_INTERCEPT(method) \
+// Declares a wildcart property getter. If a property is requested that isn't already present on
```
justin-mp, line 478: nit: s/wildcart/wildcard/

jasnell (approved):
Nice!

PR #1594, opened 2024-01-30
sqlite: don't count rows written to temporary (ephemeral) tables

smerritt:
SQLite may generate temporary tables while executing a SELECT statement. Rows written to these tables shouldn't count as they're not under control of the user.

This change is essentially
https://github.com/tursodatabase/libsql/commit/5d6c79e8fe493232111cf9ea85ff46765e7a23a9, which does this same thing in libsql; libsql is also where the original row-counting logic comes from.

The test was written by glen@cloudflare.com.


rozenmd (approved):
LGTM!

smerritt:
Fixup: 3044e117

smerritt:
An existing test was broken; it was updating 2 of 4 rows in an indexed table and asserting `rows_written == 4`, which becomes 2 with this change.

PR #1595, opened 2024-01-31
Python: Make capnp field for pythonRequirement void

hoodmane:
The requirement is in the name of the "module" and the field value is unused. Marking the field void makes that clear.

### DIFF
--- a/src/workerd/server/workerd.capnp
+++ b/src/workerd/server/workerd.capnp
@@ -267,7 +267,7 @@ struct Worker {
       # A Python module. All bundles containing this value type are converted into a JS/WASM Worker
       # Bundle prior to execution.
 
-      pythonRequirement @9 :Text;
+      pythonRequirement @9 :Void;
       # A Python package that is required by this bundle. The package must be supported by
       # Pyodide (https://pyodide.org/en/stable/usage/packages-in-pyodide.html). All packages listed
       # will be installed prior to the execution of the worker.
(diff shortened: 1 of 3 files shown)


dom96 (changes requested):
I think we should keep this: it allows us to introduce version information into the requirements in the future without changing the capnp schema.

I also don't think the `name` should contain the version + operator. Ideally it should be consistent with the names of the JS modules.

src/workerd/server/workerd.capnp
```diff
@@ -267,7 +267,7 @@ struct Worker {
       # A Python module. All bundles containing this value type are converted into a JS/WASM Worker
       # Bundle prior to execution.
 
-      pythonRequirement @9 :Text;
+      pythonRequirement @9 :Void;
```
kentonv, line 270: In general, you cannot change the type of an existing field in a Cap'n Proto schema without potentially breaking binary compatibility -- even for messages that do not use the specific field, since the sequence of field affects the layout of the struct. The types of changes that are safe are documented here:

https://capnproto.org/language.html#evolving-your-protocol

We do at least sort of care about binary compatibility for workerd config, as we allow people to compile a config into binary format and then load that in production. Whether or not anyone actually does this, I'm honestly not sure, but it is a thing we support via the CLI.

With all that said, it _just happens_ that this particular change will not break binary compatibility. This is essentially by coincidence, due to details of the encoding algorithm and the particular sequence of fields defined before this one. But in general this is not something you should try to guess about. You can always ask me to check.

PR #1596, opened 2024-01-31
[bazel] Mark param_extractor{_bin} as no-downstream

fhanau:
Dependent repositories will not have the required rust crates for them by default, so disable them there.


fhanau:
@mrbbot For context we're adding the tag so that we can compile all of workerd (`@workerd//...`) from dependent targets like the internal repository without having to exclude the tools folder manually.

mrbbot (approved):
Cool! ✅

PR #1597, opened 2024-01-31
Accept module names with and without a .py extension.

dom96:
It looks like both are supported by JS Workers (both in EWC and in workerd). So this PR ensures the same is the case for Python Workers.

### Test Plan

```
$ bazel run @workerd//src/workerd/server:workerd -- serve $PWD/deps/workerd/samples/pyodide/config.capnp --experimental
```

+ test upstream

### DIFF
--- a/src/pyodide/python-entrypoint-helper.js
+++ b/src/pyodide/python-entrypoint-helper.js
@@ -144,7 +144,9 @@ async function setupPackages(pyodide) {
   const micropipRequirements = [];
   for (const { name, value } of metadata.globals) {
     if (value.pythonModule !== undefined) {
-      pyodide.FS.writeFile(`/session/${name}.py`, value.pythonModule, {
+      // Support both modules names with the `.py` extension as well as without.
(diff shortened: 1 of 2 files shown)


src/pyodide/python-entrypoint-helper.js
```diff
@@ -201,7 +203,11 @@ async function setupPackages(pyodide) {
     );
   }
 
-  return pyodide.pyimport(metadata.mainModule);
+  // The main module can have a `.py` extension, strip it if it exists.
+  const mainName = metadata.mainModule;
+  const mainModule = mainName.endsWith(".py") ? mainName.slice(0, mainName.length - 3) : mainName;
```
hoodmane, line 208:
```suggestion
  const mainModule = mainName.endsWith(".py") ? mainName.slice(0, -3) : mainName;
```

hoodmane (approved):

harrishancock (approved):

jasnell (approved):

PR #1598, opened 2024-01-31
Fixup gc visitation in streams readers and writers

jasnell:
PR does a couple of things:

1. Ensures the the Reader/Writer is gc visiting the strong ref to the ReadableStream/WritableStream
2. Switches to uses a WeakRef rather than a Maybe bare ref from the ReadableStream/WritableStream to the Reader

### DIFF
--- a/src/workerd/api/streams/common.h
+++ b/src/workerd/api/streams/common.h
@@ -318,9 +318,9 @@ public:
     // passing along the closed promise that will be used to communicate state to the
     // user code.
     //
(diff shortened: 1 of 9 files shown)


jasnell:
There are a number of relevant failures on internal CI. Marking this as a draft while I investigate.

jasnell:
> Does the change to WeakRef fix a problem or is it just aesthetics? ...

Fixes a problem in the cleanup ordering that was masked by the memory leak.

jasnell:
Superseded by #1606

PR #1599, opened 2024-01-31
Move python-entrypoint.js into a top level Capnp constant

hoodmane:
So we don't have to search for it in the module list. It's not meant to be imported by the user anyways.

### DIFF
--- a/src/pyodide/BUILD.bazel
+++ b/src/pyodide/BUILD.bazel
@@ -17,24 +17,37 @@ capnp_embed(
 )
 
 copy_file(
-    name = "packages_capnp_file",
-    src = "pyodide_packages.capnp",
-    out = "generated/pyodide_packages.capnp",
+    name = "python_entrypoint_file",
+    src = "python-entrypoint.js",
+    out = "generated/python-entrypoint.js"
 )
 
 capnp_embed(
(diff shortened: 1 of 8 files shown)


src/pyodide/BUILD.bazel
```diff
@@ -17,24 +17,37 @@ capnp_embed(
...
+copy_file(
+    name = "pyodide_extra_capnp_file",
+    src = "pyodide_extra.capnp",
+    out = "generated/pyodide_extra.capnp",
+)
+
+capnp_embed(
+    name = "pyodide_extra_file_embed",
+    src = "generated/pyodide_extra.capnp",
+    deps = ["pyodide_extra_capnp_file"],
```
dom96, line 40: A lot of duplication here with the other files that are embedded. Can we refactor this into a function to make it easier to see which files are getting embedded and with what names?

dom96 (approved):

src/pyodide/BUILD.bazel
hoodmane: Sure, how about in a followup?

dom96: Sure. Big fan of leaving things like this to other PRs so we can move fast :)

PR #1600, opened 2024-01-31
Python: Only pass metadata that we need

hoodmane:
Rather than serializing the entire workerd config into a JSON string and then deserializing it at runtime this makes another readonly file system so we can copy only the data that Python wants to look at. This will copy all Data, Text, and Python files once at startup into the MetadataReader since I am not sure about the lifespan of the capnp config and it seems easier to understand if we make the module own its data.


src/workerd/api/pyodide/pyodide.h
```diff
@@ -1,45 +1,185 @@
...
+    JSG_METHOD(getNames);
+    JSG_METHOD(getSizes);
+    JSG_METHOD(read);
+  }
+};
+
+using Worker = server::config::Worker;
+
+inline jsg::Ref<PyodideMetadataReader> makePyodideMetadataReader(Worker::Reader conf) {
+  auto modules = conf.getModules();
```
jasnell, line 118: This method is rather too large to be inlined. Would much rather this be moved out into a `pyodide.c++`

hoodmane: Okay I did this.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -93,89 +94,19 @@ const EMBEDDED_PYTHON_PACKAGES = [
...
 
   if (pythonRequirements.length > 0) {
-    await pyodide.loadPackage(pythonRequirements);
-  }
-
-  if (micropipRequirements.length > 0) {
-    // Micropip and ssl packages are pre-loaded via the packages tarball. This means
-    // we should be able to load micropip directly now.
-
     const micropip = pyodide.pyimport("micropip");
```
dom96, line 108:
```suggestion
    // Our embedded packages tarball always contains at least `micropip` so we can load it here safely.
    const micropip = pyodide.pyimport("micropip");
```

src/workerd/api/BUILD.bazel
```diff
@@ -92,16 +96,22 @@ wd_cc_library(
 
 wd_cc_library(
     name = "pyodide",
-    srcs = ["pyodide.c++"],
+    srcs = [
+        "pyodide.c++",
+        "pyodide/pyodide.c++",
```
dom96, line 101: These should be merged together, or the second one given a more descriptive name.

dom96 (approved):

src/workerd/api/BUILD.bazel
hoodmane: I'll leave that to a followup. But it would also be good to merge `pyodide.h` and `pyodide/pyodide.h`

PR #1601, opened 2024-02-01
Await actor alarm deletion cache flush outputGate when alarm succeeds

jqmmes:
When an alarm handler successfully completes its execution, the alarm is marked ready for deletion in actor-cache (unless it's been updated since the start of handler execution). This alarm change will only be reflected in storage, once cache flushes and changes are written.

If we have access to actor-cache, we must await for the alarm deletion or update to flush(), meaning the changes were written to storage before we return the alarm result.

If we don't await, it's possible for alarm manager to pull the wrong alarm value (the same alarm that just completed) from storage before these changes are actually made, rerunning it, when it shouldn't.

### DIFF
--- a/src/workerd/io/worker-entrypoint.c++
+++ b/src/workerd/io/worker-entrypoint.c++
@@ -558,6 +558,20 @@ kj::Promise<WorkerInterface::AlarmResult> WorkerEntrypoint::runAlarmImpl(
           return lock.getGlobalScope().runAlarm(scheduledTime, lock, handler);
         });
 
+        // The alarm handler was successfully complete. We must guarantee this same alarm does not
+        // run again.
+        if (result.outcome == EventOutcome::OK){
+          // When an alarm handler completes its execution, the alarm is marked ready for deletion in
+          // actor-cache. This alarm change will only be reflected in the alarmsXX table, once cache
+          // flushes and changes are written to CRDB.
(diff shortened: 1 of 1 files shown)


src/workerd/io/worker-entrypoint.c++
```diff
@@ -558,6 +558,25 @@ kj::Promise<WorkerInterface::AlarmResult> WorkerEntrypoint::runAlarmImpl(
...
+          // actor-cache. This alarm change will only be reflected in the alarmsXX table, once cache
+          // flushes and changes are written to CRDB.
+          // If we have access to actor-cache (which we should have at this point), we must await
+          // for the alarm deletion or update to flush(), meaning the changes were written to CRDB
+          // before we return the alarm result.
+          // If we don't await, it's possible for alarm manager to pull the wrong alarm value (the
+          // same alarm that just completed) from CRDB before these changes are actually made,
+          // rerunning it, when it shouldn't.
+          KJ_IF_SOME(persistent, actor.getPersistent()) {
+            KJ_IF_SOME(pendingFlush, persistent.onNoPendingFlush()) {
```
kentonv, line 574: Should this maybe be waiting on the output gate, instead of directly on underlying storage? They are usually the same thing but if there's a case where they don't match, I would hope that breaking the output gate will prevent an alarm from being treated as successful.

kentonv (approved):

PR #1602, opened 2024-02-01
Fix unchecked malloc

garrettgu10:
The original code did not check if the allocation succeeded, leading to possible issues if malloc() returns null. 

This PR should fix that by simply replacing the malloc with a stack allocation.

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -720,8 +720,8 @@ static void stopProfiling(jsg::Lock& js,
       }
 
       auto hitLineCount = allNodes[i]->GetHitLineCount();
-      v8::CpuProfileNode::LineTick lineBuffer[hitLineCount];
-      allNodes[i]->GetLineTicks(lineBuffer, hitLineCount);
+      auto lineBuffer = kj::heapArray<v8::CpuProfileNode::LineTick>(hitLineCount);
+      allNodes[i]->GetLineTicks(lineBuffer.begin(), lineBuffer.size());
 
       auto positionTicks = nodeBuilder.initPositionTicks(hitLineCount);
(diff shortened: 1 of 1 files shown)


garrettgu10:

garrettgu10:
recheck

jclee:
Just naively wondering... If the malloc is sometimes failing, is it just due to running out of heap space when trying to allocate a large block of memory?  If so, wouldn't allocating on the stack just introduce a different failure mode where stack space is exhausted?

kentonv:
@jclee The malloc is not ever failing. This was flagged by a security scanner tool. In practice we have overcommit enabled so mallocs never fail. But we also shouldn't be using malloc anywhere in KJ code.

src/workerd/io/worker.c++
```diff
@@ -720,10 +720,7 @@ static void stopProfiling(jsg::Lock& js,
       }
 
       auto hitLineCount = allNodes[i]->GetHitLineCount();
-      v8::CpuProfileNode::LineTick* lineBuffer =
-          (v8::CpuProfileNode::LineTick*)malloc(
-          hitLineCount * sizeof(v8::CpuProfileNode::LineTick));
-      KJ_DEFER(free(lineBuffer));
+      v8::CpuProfileNode::LineTick lineBuffer[hitLineCount];
```
kentonv, line 723: This creates a variable-length stack array, which may be dangerous depending on how large `hitLineCount` can get. There is a macro `KJ_STACK_ARRAY` which helps with allocating small arrays on the stack and big ones on the heap. Alternatively, you could just use `kj::heapArray` here to always allocate on the heap.

jasnell (approved):

PR #1603, opened 2024-02-01
Check lower bound returned by atoi()

garrettgu10:
A potential negative value returned by atoi() may cause an excessive memory allocation. This PR adds a check that the returned value is nonnegative.


Warfields (approved):

kentonv:
We should really remove the `atoi()` call entirely and instead use `metadata.parseAs<uint>()`.

PR #1606, opened 2024-02-01
Fix GC visitation in stream readers and writers

kentonv:
Fixes a memory leak in certain uses of streams.

This is a simpler version of #1598. I was able to directly fix the problem that the WeakRef refactor also solved.


byule (approved):

kentonv:
Missed some instances of the teardown bug, discovered when I ran more tests. Updated.

PR #1607, opened 2024-02-01
Add toStringTag value on JSG resource types

garrettgu10:
This PR ensures that Object.prototype.toString() returns the correct values for built-in objects (i.e. ResourceTypes).

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -47,7 +47,7 @@ namespace workerd::jsg {
       ::workerd::jsg::JsgKind::RESOURCE; \
   using jsgSuper = jsgThis; \
   using jsgThis = Type; \
-  inline kj::StringPtr jsgGetName() const { return #Type##_kjc; } \
+  static inline kj::StringPtr jsgGetName() { return #Type##_kjc; } \
   inline kj::StringPtr jsgGetMemoryName() const override { return #Type##_kjc; } \
   inline size_t jsgGetMemorySelfSize() const override { return sizeof(Type); } \
   inline void jsgGetMemoryInfo(jsg::MemoryTracker& tracker) const override { \
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -47,6 +47,7 @@ namespace workerd::jsg {
       ::workerd::jsg::JsgKind::RESOURCE; \
   using jsgSuper = jsgThis; \
   using jsgThis = Type; \
+  inline kj::StringPtr jsgGetName() const { return #Type##_kjc; } \
   inline kj::StringPtr jsgGetMemoryName() const override { return #Type##_kjc; } \
   inline size_t jsgGetMemorySelfSize() const override { return sizeof(Type); } \
   inline void jsgGetMemoryInfo(jsg::MemoryTracker& tracker) const override { \
--- a/src/workerd/jsg/resource.h
+++ b/src/workerd/jsg/resource.h
@@ -681,7 +681,7 @@ struct ResourceTypeBuilder {
       v8::PropertyAttribute::ReadOnly | v8::PropertyAttribute::DontEnum));
 
     auto toStringTagSymbol = v8::Symbol::GetToStringTag(isolate);
-    prototype->Set(toStringTagSymbol, v8StrIntern(isolate, typeWrapper.jsgGetName()), static_cast<v8::PropertyAttribute>(0));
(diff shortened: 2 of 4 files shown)


src/workerd/jsg/resource.h
```diff
@@ -653,6 +653,16 @@ struct WildcardPropertyCallbacks<
   }
 };
 
+template<typename TypeWrapper>
+decltype(std::declval<TypeWrapper&>().jsgGetMemoryName()) typeWrapperName(TypeWrapper &typeWrapper) {
+  return typeWrapper.jsgGetMemoryName();
```
jasnell, line 658: `jsgGetMemoryName` is the wrong thing to use. Likely what we'll want is to integrate this somehow into `JSG_RESOURCE_TYPE` using the `Type` parameter to set the name.

src/workerd/jsg/resource.h
```diff
@@ -679,6 +689,9 @@ struct ResourceTypeBuilder {
     inspectProperties = v8::ObjectTemplate::New(isolate);
     prototype->Set(symbol, inspectProperties, static_cast<v8::PropertyAttribute>(
       v8::PropertyAttribute::ReadOnly | v8::PropertyAttribute::DontEnum));
+
+    auto toStringTagSymbol = v8::Symbol::GetToStringTag(isolate);
+    prototype->Set(toStringTagSymbol, v8StrIntern(isolate, typeWrapperName(typeWrapper)), v8::PropertyAttribute::ReadOnly);
```
jasnell, line 694: The `ToStringTag` symbol should not really be `v8::PropertyAttribute::ReadOnly`

src/workerd/jsg/jsg.h
```diff
@@ -47,6 +47,7 @@ namespace workerd::jsg {
       ::workerd::jsg::JsgKind::RESOURCE; \
   using jsgSuper = jsgThis; \
   using jsgThis = Type; \
+  inline kj::StringPtr jsgGetName() const { return #Type##_kjc; } \
```
garrettgu10, line 50: @jasnell I'm not sure what to do here. The build fails when I don't include it, but the `Type` here is the name of the isolate wrapper, not the wrapped types, so we get weird results in test cases.

dom96:
what's the difference between getMemoryName and getName?

jasnell:
> what's the difference bretween `getMemoryName` and `getName`

`jsgGetMemoryName(...)` is used while generating the heap snapshot. While for `jsg::Object` instances its value is *currently* the same name as the class, it does not need to be and could change in the future. We should not rely on its value for stable user-facing API.

src/workerd/jsg/resource.h
```diff
@@ -679,6 +679,9 @@ struct ResourceTypeBuilder {
     inspectProperties = v8::ObjectTemplate::New(isolate);
     prototype->Set(symbol, inspectProperties, static_cast<v8::PropertyAttribute>(
       v8::PropertyAttribute::ReadOnly | v8::PropertyAttribute::DontEnum));
+
+    auto toStringTagSymbol = v8::Symbol::GetToStringTag(isolate);
+    prototype->Set(toStringTagSymbol, v8StrIntern(isolate, Self::jsgGetName()), static_cast<v8::PropertyAttribute>(0));
```
jasnell, line 684:
```suggestion
    prototype->Set(toStringTagSymbol, v8StrIntern(isolate, Self::jsgGetName()), v8::PropertyAttribute::None);
```

src/workerd/jsg/resource.h
```diff
@@ -679,6 +679,9 @@ struct ResourceTypeBuilder {
     inspectProperties = v8::ObjectTemplate::New(isolate);
     prototype->Set(symbol, inspectProperties, static_cast<v8::PropertyAttribute>(
       v8::PropertyAttribute::ReadOnly | v8::PropertyAttribute::DontEnum));
+
+    auto toStringTagSymbol = v8::Symbol::GetToStringTag(isolate);
```
jasnell, line 683: As discussed separately on a call, there's a slight risk that adding this property automatically to all `jsg::Object`s *could* be breaking, at least in theory.  If there is code out there that checks for the presence of the `toStringTag` currently and takes action if it is not there, this can break that. I think the risk is probably really quite low but it's non-zero. Do we need a compatibility flag? Interested in what others think @irvinebroque @kentonv @harrishancock 

Specifically, this changes the current default result of calling `obj[Symbol.toStringTag]` on any of our API objects.

garrettgu10: I think the consensus is that the risk is acceptable, and we are free to make the change without a compatibility flag.

jasnell (approved):

PR #1608, opened 2024-02-02
Implement memory tracking for Blob/File/Streams

jasnell:


src/workerd/api/blob.h
```diff
@@ -60,6 +60,18 @@ class Blob: public jsg::Object {
...
 
+  void visitForMemoryInfo(jsg::MemoryTracker& tracker) const {
+    KJ_SWITCH_ONEOF(ownData) {
+      KJ_CASE_ONEOF(data, kj::Array<byte>) {
+        tracker.trackField("ownData", data);
+      }
+      KJ_CASE_ONEOF(data, jsg::Ref<Blob>) {
+        tracker.trackField("ownData", data);
+      }
+    }
```
hoodmane, line 71: It would be convenient if there were a way to call `trackField` with a `kj::OneOf<S, T, ...>` where there is an impl for `trackField` for each of them. Perhaps it's not worth the hassle to save a little bit of boilerplate.

But maybe you could make a concept `Trackable` with `T::trackField(tracker, name, value)` and then define `tracker.trackField<Trackable T>` to call `T::trackField`. Then maybe you can define `OneOf<Trackable S, Trackable T, Trackable U>::trackField` or something?

hoodmane (approved):
Looks good to me.

src/workerd/api/blob.h
jasnell: I'd considered adding that but I think I prefer separating it out like this in order to give a bit more control over how things are tracked. In some cases, it won't need to be tracked at all, in other cases it make use `trackField(...)` for one and `trackFieldWithSize(...)` for others.

PR #1609, opened 2024-02-02
Python: Share Python interpreter across requests

hoodmane:
Before this PR we instantiate the Python interpreter from scratch on every request. This is slow and uses up a lot of memory. This switches to sharing an interpreter across requests. This makes per-request non-coldstart execution time and memory usage much lower.

### DIFF
--- a/src/pyodide/python-entrypoint-helper.js
+++ b/src/pyodide/python-entrypoint-helper.js
@@ -194,7 +194,7 @@ async function setupPackages(pyodide) {
     );
     pyodide.pyimport("aiohttp_fetch_patch");
   }
-  if (requirements.includes("fastapi")) {
+  if (requirements.some(req => req.startsWith("fastapi"))) {
     const mod = await import("pyodide-internal:asgi.py");
     pyodide.FS.writeFile(
       "/lib/python3.11/site-packages/asgi.py",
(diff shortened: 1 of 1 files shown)


src/pyodide/python-entrypoint-helper.js
```diff
@@ -210,16 +210,26 @@ async function setupPackages(pyodide) {
   return pyodide.pyimport(mainModule);
 }
 
+let mainModule;
+async function getMainModule() {
+  if (mainModule) {
```
jasnell, line 215:
```suggestion
  if (mainModule === undefined) {
```

jasnell: If this method ends up getting called multiple times without `await` we'll end up with multiple calls to `loadPyodide` running. If the call is already inflight, in would be good to just attach to the existing promise.

hoodmane: How's it look now?

src/pyodide/python-entrypoint-helper.js
```diff
@@ -210,16 +210,28 @@ async function setupPackages(pyodide) {
...
+  if (mainModulePromise !== undefined) {
+    return mainModulePromise;
+  }
+  mainModulePromise = (async function() {
+    // TODO: investigate whether it is possible to run part of loadPyodide in top level scope
+    // When we do it in top level scope we seem to get a broken file system.
+    const pyodide = await loadPyodide();
+    return await setupPackages(pyodide);
+  })();
+  return mainModulePromise;
```
jasnell, line 224: Bit easier with at least one fewer Promise allocated...

```suggestion
  mainModulePromise = setupPackages(await loadPyodide());
  return mainModulePromise;
```

hoodmane: Is this still guaranteed to only execute once? As I understand it, 
```js
mainModulePromise = setupPackages(await loadPyodide());
```
means the same thing as:
```js
const tmp = await loadPyodide();
mainModulePromise = setupPackages(tmp);
```
and so there is an await before assigning to `mainModulePromise` meaning that a second request could come in and observe `mainModulePromise` as not null. The way I wrote it with the outside function being synchronous and assigning to the promise guarantees that we don't yield to the event loop when `mainModulePromise` is still `undefined.

jasnell: hah... doh you're right lol. that's what I get for doing code review at dinner time.

jasnell (approved):

dom96 (approved):

hoodmane:
My concern with rebasing is that I thought github would delete review comments on the old commits? It's fine if you've read and responded to them as long as nobody but it makes the PR history difficult to understand and you can lose comments if you force push before reading the PR. Or is this not right? I'm not really sure what the other tradeoffs are with rebase to understand under which circumstances one or the other is preferred. With stacked PRs merge is really annoying, but stacked PRs are really annoying in general.

dom96:
> My concern with rebasing is that I thought github would delete review comments on the old commits?

I don't believe that happens

hoodmane:
Okay, probably I should just try it for a while and see. Thanks!

PR #1610, opened 2024-02-02
Python: Pass the env parameter into handlers

hoodmane:
This allows Python workers to access secrets and such.

This adds a relaxed_call function and uses it to pass the env argument to the fetch/test handlers.
The relaxed_call implementation is from here:
https://github.com/pyodide/pyodide/pull/4392


src/pyodide/python-entrypoint-helper.js
```diff
@@ -190,47 +199,51 @@ async function setupPackages(pyodide) {
...
+  const relaxed_call = pyodide.pyimport("relaxed_call").relaxed_call;
+  result = { mainModule, relaxed_call };
+  return result;
 }
 
 export default {
-  async fetch(request, env) {
-    const mainModule = await getMainModule();
-    return await mainModule.fetch(request);
+  async fetch(ctx, env) {
```
jasnell, line 239: Not sure we should rename `request` to `ctx` here. With `fetch(...)` handlers, the third argument is the `ExecutionContext` that, by convention, we've typically named `ctx`. Would be good to avoid confusion here.

jasnell (approved):

src/pyodide/python-entrypoint-helper.js
hoodmane: What about for `test`? Should the first argument always be called `request`? And presumably if `fetch` takes a third argument I should pass that along too?

jasnell: No, for `test` it should be `ctx`. I'm not a big fan of the difference but that first argument to `fetch` is a `Request` object so the name should match.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -190,22 +199,24 @@ async function setupPackages(pyodide) {
...
+      { canOwn: true },
     );
   }
 
   // The main module can have a `.py` extension, strip it if it exists.
   const mainName = metadata.mainModule;
-  const mainModule = mainName.endsWith(".py") ? mainName.slice(0, -3) : mainName;
+  const mainModule = mainName.endsWith(".py")
+    ? mainName.slice(0, -3)
+    : mainName;
```
dom96, line 219: Might be good to disable autoformat, you'll get lots of conflicts because you touch these lines in other PRs :)

src/pyodide/python-entrypoint-helper.js
```diff
@@ -219,20 +230,23 @@ function getMainModule() {
...
   })();
   return mainModulePromise;
 }
 
 export default {
-  async fetch(request, env) {
-    const mainModule = await getMainModule();
-    return await mainModule.fetch(request);
+  async fetch(request, env, ctx) {
+    const { relaxed_call, mainModule } = await getMainModule();
```
dom96, line 243:
```suggestion
    const { relaxed_call, mainModule } = await getPyodide();
```

feel free to choose a better name than my suggestion :)

dom96 (approved):
Awesome!

src/pyodide/python-entrypoint-helper.js
kentonv: Actually, `test`'s arguments are `test(controller, env, ctx)`. `ctx` is still the third argument. `controller` is equivalent to `request`.

PR #1612, opened 2024-02-03
Python: include data and text modules into the file system

hoodmane:
Previously there wasn't a good way to access these from Python.

### DIFF
--- a/src/pyodide/python-entrypoint-helper.js
+++ b/src/pyodide/python-entrypoint-helper.js
@@ -3,7 +3,7 @@
 
 import { loadPyodide } from "pyodide-internal:python";
 import { default as lockfile } from "pyodide-internal:generated/pyodide-lock.json";
-import { default as origMetadata } from "pyodide-internal:runtime-generated/current-bundle";
+import { default as metadata } from "pyodide-internal:runtime-generated/current-bundle";
 
 function initializePackageIndex(pyodide) {
   if (!lockfile.packages) {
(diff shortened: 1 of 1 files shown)


src/pyodide/python-entrypoint-helper.js
```diff
@@ -98,42 +98,19 @@ const EMBEDDED_PYTHON_PACKAGES = [
...
-      metadata.globals.push({
-        name: module.name,
-        value: {
-          pythonRequirement: module.pythonRequirement,
-        },
-      });
-    }
-  }
-  return metadata;
+  metadata.globals = metadata.modules.map((mod) => ({
```
fhanau, line 103: This looks cleaner than before, nice! One concern I had: Is it possible for modules to be metadata.modules to be empty? In that case accessing metadata.modules[0].name may cause issues.

fhanau (approved):
LGTM overall but I haven't worked with Pyodide before, would be good if someone else can take a look too.

src/pyodide/python-entrypoint-helper.js
hoodmane: Thanks for the review @fhanau!

I think Workerd is required to have at least one module. Otherwise we get an error here:
https://github.com/cloudflare/workerd/blob/main/src/workerd/server/workerd-api.c++?plain=1#L208
Might be worth adding a comment though.

dom96 (commented):
I guess we don't need this if we merge https://github.com/cloudflare/workerd/pull/1600?

hoodmane:
Closed as superseded by #1600.

PR #1617, opened 2024-02-04
Refactor JSRPC to prepare for capability-based RPC

kentonv:
This is a series of small refactorings that prepares for the future where RPC params or return values can contain new RPC objects. For this to work, we need to disentangle the sending of one RPC call from the creation of RPC sessions, since one session may encompass many calls in both directions.

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1869,13 +1869,13 @@ jsg::Promise<jsg::Ref<Response>> Fetcher::fetch(
 }
 
 kj::Maybe<Fetcher::RpcFunction> Fetcher::getRpcMethod(jsg::Lock& js, kj::StringPtr name) {
-  // This is like WorkerRpc::getRpcMethod(), but we also initiate a whole new JS RPC session each
-  // time the method is called.
+  // This is like JsRpcCapability::getRpcMethod(), but we also initiate a whole new JS RPC session
+  // each time the method is called.
   return RpcFunction(JSG_VISITABLE_LAMBDA(
       (methodName = kj::str(name), self = JSG_THIS),
       (self),
       (jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>& args) -> jsg::Promise<jsg::Value> {
-    // Same as WorkerRpc::getRpcMethod(), we want to prevent calling on the wrong `this`.
+    // Same as JsRpcCapability::getRpcMethod(), we want to prevent calling on the wrong `this`.
     JSG_REQUIRE(args.This() == KJ_ASSERT_NONNULL(self.tryGetHandle(js)), TypeError,
         "Illegal invocation");
 
@@ -1884,7 +1884,7 @@ kj::Maybe<Fetcher::RpcFunction> Fetcher::getRpcMethod(jsg::Lock& js, kj::StringP
     auto event = kj::heap<api::JsRpcSessionCustomEventImpl>(
         JsRpcSessionCustomEventImpl::WORKER_RPC_EVENT_TYPE);
 
-    auto result = WorkerRpc::sendWorkerRpc(js, event->getCap(), methodName, args);
+    auto result = JsRpcCapability::sendJsRpc(js, event->getCap(), methodName, args);
 
     // Arrange to cancel the CustomEvent if our I/O context is destroyed. But otherwise, we don't
     // actually care about the result of the event. If it throws, the membrane will already have
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -7,6 +7,7 @@
 #include "system-streams.h"
 #include "util.h"
 #include "queue.h"
+#include "worker-rpc.h"
 #include <kj/encoding.h>
 #include <kj/compat/url.h>
 #include <kj/memory.h>
@@ -1865,6 +1866,36 @@ jsg::Promise<jsg::Ref<Response>> Fetcher::fetch(
     jsg::Lock& js, kj::OneOf<jsg::Ref<Request>, kj::String> requestOrUrl,
     jsg::Optional<kj::OneOf<RequestInitializerDict, jsg::Ref<Request>>> requestInit) {
   return fetchImpl(js, JSG_THIS, kj::mv(requestOrUrl), kj::mv(requestInit));
+}
+
+kj::Maybe<Fetcher::RpcFunction> Fetcher::getRpcMethod(jsg::Lock& js, kj::StringPtr name) {
+  // This is like WorkerRpc::getRpcMethod(), but we also initiate a whole new JS RPC session each
+  // time the method is called.
+  return RpcFunction(JSG_VISITABLE_LAMBDA(
+      (methodName = kj::str(name), self = JSG_THIS),
+      (self),
+      (jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>& args) -> jsg::Promise<jsg::Value> {
+    // Same as WorkerRpc::getRpcMethod(), we want to prevent calling on the wrong `this`.
+    JSG_REQUIRE(args.This() == KJ_ASSERT_NONNULL(self.tryGetHandle(js)), TypeError,
+        "Illegal invocation");
+
+    auto& ioContext = IoContext::current();
+    auto worker = self->getClient(ioContext, kj::none, "jsRpcSession"_kjc);
+    auto event = kj::heap<api::JsRpcSessionCustomEventImpl>(
+        JsRpcSessionCustomEventImpl::WORKER_RPC_EVENT_TYPE);
+
+    auto result = WorkerRpc::sendWorkerRpc(js, event->getCap(), methodName, args);
+
+    // Arrange to cancel the CustomEvent if our I/O context is destroyed. But otherwise, we don't
+    // actually care about the result of the event. If it throws, the membrane will already have
+    // propagated the exception to any RPC calls that we're waiting on, so we even ignore errors
+    // here -- otherwise they'll end up logged as "uncaught exceptions" even if they were, in fact,
+    // caught elsewhere.
+    ioContext.addTask(worker->customEvent(kj::mv(event)).attach(kj::mv(worker))
+        .then([](auto&&) {}, [](kj::Exception&&) {}));
+
+    return result;
+  }));
 }
 
 static jsg::Promise<void> throwOnError(jsg::Lock& js,
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -95,14 +95,11 @@ private:
 
 jsg::Promise<jsg::Value> WorkerRpc::sendWorkerRpc(
     jsg::Lock& js,
+    rpc::JsRpcTarget::Client client,
     kj::StringPtr name,
     const v8::FunctionCallbackInfo<v8::Value>& args) {
-
   auto& ioContext = IoContext::current();
-  auto worker = getClient(ioContext, kj::none, "jsRpcSession"_kjc);
-  auto event = kj::heap<api::JsRpcSessionCustomEventImpl>(WORKER_RPC_EVENT_TYPE);
 
-  rpc::JsRpcTarget::Client client = event->getCap();
   auto builder = client.callRequest();
   builder.setMethodName(name);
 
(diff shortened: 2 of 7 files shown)


src/workerd/api/worker-rpc.c++
```diff
@@ -62,50 +122,47 @@ kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> WorkerRpc::sendWorke
...
-      return jsg::Value(js.v8Isolate, deserializeV8(js, result.getResult().getV8Serialized()));
-    });
-  });
+    // Let's make sure that the returned function wasn't invoked using a different `this` than the
+    // one intended. We don't really _have to_ check this, but if we don't, then we have to forever
+    // support tearing RPC methods off their owning objects and invoking them elsewhere. Enforcing
+    // it might allow future implementation flexibility.
+    //
+    // TODO(cleanup): Ideally, we'd actually unwrap `args.This()` instead of capturing `self` at
+    // all, but there isn't a JSG-bselled way to do that.
```
jasnell, line 144:
```suggestion
    // all, but there isn't a JSG-based way to do that.
```

kentonv: wat

jasnell: Given that this needs `self` anyway for the call below (`*self->capnpClient`) is this comment entirely accurate? It would seem this needs to hold a strong reference to self either way? Unless you had something else in mind for that bit also?

src/workerd/api/worker-rpc.c++
```diff
@@ -62,50 +122,47 @@ kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> WorkerRpc::sendWorke
...
+    JSG_REQUIRE(args.This() == KJ_ASSERT_NONNULL(self.tryGetHandle(js)), TypeError,
+        "Method '", methodName, "' called with wrong 'this'.");
```
jasnell, line 146: nit: the canonical error message for this case (elsewhere in our code and JS in general) is "Illegal invocation"  (which has always been a terrible error message but it is most common.

src/workerd/api/worker-rpc.c++
kentonv: The comment is saying we'd use `unwrap<JsRpcCapability>(args.This())->capnpClient` instead of `self->capnpClient`, if such an `unwrap` function existed.

jasnell: I figured it had to be something along those lines. The comment can be clarified. As currently worded it wouldn't be clear enough to someone else coming into the codebase later trying to figure out how to make the TODO actionable.

src/workerd/api/worker-rpc.c++
```diff
@@ -95,14 +95,11 @@ private:
 
 jsg::Promise<jsg::Value> WorkerRpc::sendWorkerRpc(
     jsg::Lock& js,
+    rpc::JsRpcTarget::Client client,
     kj::StringPtr name,
     const v8::FunctionCallbackInfo<v8::Value>& args) {
-
   auto& ioContext = IoContext::current();
```
MellowYarker, line 101: nit: Should we just pass the `IoContext&` in via method param since we already called this in the caller in `http.c++`?

src/workerd/api/worker-rpc.h
```diff
@@ -28,30 +28,30 @@ namespace workerd::api {
...
   // Serializes the method name and arguments, calls customEvent to get the capability, and uses
   // the capability to send our request to the remote Worker. This resolves once the RPC promise
   // resolves.
   //
-  // This is a static helper, no `WorkerRpc` object is needed. This is the shared implementation
-  // between `WorkerRpc::getRpcMethod()` and `Fetcher::getRpcMethod()`.
-  static jsg::Promise<jsg::Value> sendWorkerRpc(
+  // This is a static helper, no `JsRpcCapability` object is needed. This is the shared
+  // implementation between `JsRpcCapability::getRpcMethod()` and `Fetcher::getRpcMethod()`.
+  static jsg::Promise<jsg::Value> sendRpc(
```
MellowYarker, line 54: nit: This is probably fine, but I hesitated for a moment when I saw this method rename. It makes sense of course, but now we have `JsRpcCapability::sendRpc()` _and_ `JsRpcSessionCustomEventImpl::sendRpc()`. For anyone who looks at this code for the first time and isn't already familiar with how we dispatch `CustomEvent`s, this could be pretty confusing (a quick search for `sendRpc()` brings up a ton of results.

Maybe it's not an issue given we've got clangd and can find references, but figured I might as well mention it since overloading of names has been definitely been a pain point in the past. Maybe `sendJsRpc()`?

src/workerd/api/worker-rpc.c++
```diff
@@ -200,39 +260,107 @@ private:
...
-  auto req = dispatcher.getJsRpcTargetRequest();
+  // We arrange to revoke all capabilities in this session as soon as `sendRpc()` completes or is
+  // canceled. Normally, the server side doesn't return if any capabilities still exist, so this
+  // only makes a difference in the case that some sort of an error occurred. We don't strictly
+  // have to revoke the capabilities as they are probably already broken anyway, but revoking them
+  // helps to ensure that the underlying transport isn't "held open" waiting for the JS garbage
+  // collector to actually collect the JsRpcCapability objects.
+  auto [revokePromise, revokeFulfiller] = kj::newPromiseAndFulfiller<void>();
+
+  KJ_DEFER({
```
MellowYarker, line 340: Just a guess regarding the linux build, but perhaps this should be

```
  kj::defer([revokeFulfiller = kj::mv(revokeFulfiller)]() mutable {
    // ... stuff
  }
```

src/workerd/api/worker-rpc.c++
```diff
@@ -200,39 +260,107 @@ private:
...
+    if (revokeFulfiller->isWaiting()) {
+      revokeFulfiller->reject(KJ_EXCEPTION(DISCONNECTED, "JS-RPC session canceled"));
+    }
```
jasnell, line 343: Compiler in CI doesn't like this for some reason. If I'm recalling correctly, I think there was an old clang bug with capturing a reference to a structured binding, can't remember the details tho (https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding)

MellowYarker (commented):
This is a great refactor! I want to spend a bit more time looking at the membrane related commits since it's still fairly new to me but generally this PR makes sense.

src/workerd/api/worker-rpc.c++
```diff
@@ -244,11 +343,38 @@ kj::Promise<WorkerInterface::CustomEvent::Result>
...
-  this->capFulfiller->fulfill(sent.getServer());
 
-  auto resp = co_await sent;
+  this->capFulfiller->fulfill(
+      capnp::membrane(
+          sent.getServer(),
+          kj::refcounted<RevokerMembrane>(kj::mv(revokePromise))));
+
+  try {
+    co_await sent;
```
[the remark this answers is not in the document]
kentonv: Nope, membrane revocation promises must only reject, never resolve. This is in the docs for `MembranePolicy::onRevoked()`. The reason is, on revocation, any RPCs still in-flight over the membrane (or new RPCs initiated later) need to throw something; the exception that the revocation promise rejected with becomes the thing they throw.

Note that the KJ_DEFER block above ensures that we always reject the revocation promise, if anyone is actually listening. (In most cases, though, no one should be listening.)

src/workerd/api/worker-rpc.c++
kentonv: We could, but `IoContext::current()` is not expensive to call, so I didn't think it was worth the API change. It's sort of a coincidence that the call site happens to have its own copy of this reference.

src/workerd/api/worker-rpc.h
kentonv: Yeah that's a good call.

src/workerd/api/worker-rpc.c++
kentonv: I'm going to go with removing the destructuring so I can keep the more concise KJ_DEFER.

kentonv:
Fixups:

[fixup! JSRPC: Rename: getJsRpcTarget() -> jsRpcSession()](https://github.com/cloudflare/workerd/pull/1617/commits/b4efada488188abb850c88158b2672aa718cd9ab)
[fixup! JSRPC: Rename: WorkerRpc -> JsRpcCapability](https://github.com/cloudflare/workerd/pull/1617/commits/1e79a0006c3929de330d970f8a091ae84c96e07d)
[fixup! JSRPC: Refactor: Allow open capabilities to extend session.](https://github.com/cloudflare/workerd/pull/1617/commits/a32e7a65fc78342043a9d63457a8bb1a5f53b545)

src/workerd/api/worker-rpc.c++
kentonv: Huh, I guess this "bug in the standard" must have been fixed in the newer clang we're using for our internal build, but not in the default clang used in CI? Fixed by removing the destructuring.

src/workerd/api/http.c++
```diff
@@ -1867,6 +1868,36 @@ jsg::Promise<jsg::Ref<Response>> Fetcher::fetch(
...
+kj::Maybe<Fetcher::RpcFunction> Fetcher::getRpcMethod(jsg::Lock& js, kj::StringPtr name) {
+  // This is like JsRpcCapability::getRpcMethod(), but we also initiate a whole new JS RPC session
+  // each time the method is called.
+  return RpcFunction(JSG_VISITABLE_LAMBDA(
+      (methodName = kj::str(name), self = JSG_THIS),
+      (self),
+      (jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>& args) -> jsg::Promise<jsg::Value> {
+    // Same as JsRpcCapability::getRpcMethod(), we want to prevent calling on the wrong `this`.
+    JSG_REQUIRE(args.This() == KJ_ASSERT_NONNULL(self.tryGetHandle(js)), TypeError,
+        "Method '", methodName, "' called with wrong 'this'.");
```
jasnell, line 1880: Nit: `Invalid invocation` for consistency or just say screw it and go with a better error message? Either way is fine but let's make sure the decision is intentional.

src/workerd/api/worker-rpc.c++
```diff
@@ -62,50 +122,50 @@ kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> WorkerRpc::sendWorke
...
-  // If customEvent throws, we'll cancel callResult and propagate the exception. Otherwise, we'll
-  // just wait until callResult finishes.
-  return callResult.exclusiveJoin(customEventResult
-      .then([](auto&&) -> kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> {
-        return kj::NEVER_DONE;
-      }));
+
+  return ioContext.awaitIo(js, kj::mv(callResult),
+      [](jsg::Lock& js, auto result) -> jsg::Value {
+    return jsg::Value(js.v8Isolate, deserializeV8(js, result.getResult().getV8Serialized()));
```
jasnell, line 128: nit:

```suggestion
    return js.v8Ref(deserializeV8(js, result.getResult().getV8Serialized());
```

src/workerd/api/worker-rpc.c++
```diff
@@ -62,50 +122,50 @@ kj::Promise<capnp::Response<rpc::JsRpcTarget::CallResults>> WorkerRpc::sendWorke
...
+  });
 }
 
-kj::Maybe<WorkerRpc::RpcFunction> WorkerRpc::getRpcMethod(jsg::Lock& js, kj::StringPtr name) {
-  // Named intercept is enabled, this means we won't default to legacy behavior.
-  // The return value of the function is a promise that resolves once the remote returns the result
-  // of the RPC call.
-  return RpcFunction([this, methodName=kj::str(name)]
+kj::Maybe<JsRpcCapability::RpcFunction> JsRpcCapability::getRpcMethod(
```
MellowYarker, line 132: May be missing something, but is this method actually needed? AFAICT we only call `Fetcher::getRpcMethod()`, and never actually construct a `JsRpcCapability`, we just call the static `JsRpcCapability::sendJsRpc()`. It seems like all we actually need is the static method, which could just be a function in `worker-rpc.c++` instead, no?

Judging by some comments in `worker-rpc.h` I figure it may be relevant in the future when we're able to pass capabilities around, but I'm wondering if the intention is to use this class then, or just put more stuff on `Fetcher`? If we're going to use `Fetcher` then this class isn't needed anymore.

MellowYarker (approved):

src/workerd/api/worker-rpc.c++
kentonv: This actually doesn't work, unless I explicitly specify the type, i.e. `js.v8Ref<v8::Value>(`. This looks like a bug, `v8Ref()` needs overloads specifically for `JsValue` and friends, or something. Or maybe I should be using `JsRef`?

Anyway, this code is not new, it was just moved from below. I think this should be cleaned up in a separate change.

kentonv:
Fixup: 059e0d89

jasnell (approved):

src/workerd/api/worker-rpc.c++
kentonv: Sorry, started replying to this earlier then got distracted.

Currently, there's no way to construct a `JsRpcCapability`, but that will change soon. This is the type we will use when people pass an object with methods as one of the parameters or results of some other RPC call. The receiving end will get a `JsRpcCapability` that makes RPCs back to the sender.

PR #1619, opened 2024-02-05
[CI] Reduce Windows debug binary size

fhanau:
This resolves CI failures due to running out of disk space, at least for
the near term. Also drops a flag modifying features for windows_no_dbg,
disabling optimizations when building for the host configuration is not
necessary.

I picked common MSVC flags that are safe to use with debug info here. When I last measured this, the bazel output + cache takes up 60GB with this on a fully cached Windows-debug build whereas the build previously ran out of space upon reaching 70GB. Based on those numbers this is just a short-term fix unfortunately, a larger improvement would be possible by changing cache behavior (70GB is only reached when binaries are duplicated between disk cache and regular bazel output) or by patching bazel to make opting out of PDB generation possible.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -207,8 +207,13 @@ build:windows_no_dbg -c opt
 build:windows_no_dbg --copt='-O0' --host_copt='-O0'
 build:windows_no_dbg --copt='/Od' --host_copt='/Od'
 build:windows_no_dbg --copt='/INCREMENTAL:NO' --host_copt='/INCREMENTAL:NO'
-build:windows_no_dbg --noincompatible_use_host_features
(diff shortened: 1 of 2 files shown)


fhanau:
When a PR to reduce disk space usage has failing CI jobs due to still having too much disk space 😭😭😭
I think the recent additions of WebGPU and Python support with its large capnp embeds really did a number on the Windows runner since it is already encumbered with so much debug info and it is duplicated in a PDB file. I will merge this as the build fails later in the process than on main which is consistent with the 70GB => 60GB disk space reduction encountered when I first made this PR, but this is insufficient to resolve the problem unfortunately.

PR #1621, opened 2024-02-05
Cleanup EventTarget::NativeHandler

jasnell:
Make the connection between EventTarget and NativeHandler safer.

Simplify/Cleanup the implementation a bit

### DIFF
--- a/src/workerd/api/basics.h
+++ b/src/workerd/api/basics.h
@@ -224,7 +224,33 @@ class EventTarget: public jsg::Object {
 public:
 
   // RAII-style listener that can be attached to an EventTarget.
-  class NativeHandler;
+  class NativeHandler {
+  public:
+    using Signature = void(jsg::Ref<Event>);
+    NativeHandler(jsg::Lock& js, jsg::Ref<EventTarget> target, kj::String type,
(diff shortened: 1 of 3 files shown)


harrishancock (approved):

src/workerd/api/basics.h
```diff
@@ -224,7 +224,33 @@ class EventTarget: public jsg::Object {
...
+
+    uint hashCode() const;
+
+    void visitForGc(jsg::GcVisitor& visitor);
+  private:
+    void detach();
+
+    kj::String type;
+    struct State {
+      jsg::Ref<EventTarget> target;
```
mikea, line 245: so, lumping target & func under Maybe together seems to be the change to NativeHandler?

src/workerd/api/basics.h
```diff
@@ -224,7 +224,33 @@ class EventTarget: public jsg::Object {
...
+      jsg::Function<Signature> func;
+    };
+    kj::Maybe<State> state;
```
mikea, line 248: maybe one-liner comment about when it is none? E.g. // is never empty during operations, gets cleared in detach()

mikea (approved):

PR #1625, opened 2024-02-06
Implement memory tracking for handful of classes

jasnell:
* cache
* cf-property
* form-data
* hyperdrive
* kv
* queue

### DIFF
--- a/src/workerd/jsg/memory.h
+++ b/src/workerd/jsg/memory.h
@@ -536,7 +536,7 @@ void MemoryTracker::trackField(
     decCurrentNodeSize(sizeof(T));
   }
   pushNode(nodeName.orDefault(edgeName), sizeof(T), edgeName);
-  for (const auto item : value) {
+  for (const auto& item : value) {
     trackField(nullptr, item, elementName);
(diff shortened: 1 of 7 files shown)


src/workerd/jsg/memory.h
```diff
@@ -536,7 +536,7 @@ void MemoryTracker::trackField(
     decCurrentNodeSize(sizeof(T));
   }
   pushNode(nodeName.orDefault(edgeName), sizeof(T), edgeName);
-  for (const auto item : value) {
+  for (const auto& item : value) {
```
hoodmane, line 539: What does this change do? Originally were we copying the value and now we don't? So:
* The original code trigger a compile error if T doesn't have a compile error?
* If T was a large value before, we would copy each entry just to calculate it's size?

hoodmane (approved):

src/workerd/jsg/memory.h
jasnell: Original was a bug. This template previously wasn't being used yet so it hadn't been caught.

PR #1627, opened 2024-02-06
Configurable local v8 heap limits

jasnell:
Experimental....

This is intended to provide a mechanism for testing the memory heap limits of a worker locally. It is intentionally more flexible than the production limits in order to give more flexibility in local testing. Heap snapshots will be generated whenever the limit is reached if the heapSnapshotNearHeapLimit is set to a positive, non-zero value. No more than the number of snapshots specified will be generated.

```
  modules = [
  ],
  compatibilityDate = "2023-02-28",
  limits = (
    # Emit a warning if the worker's heap size exceeds 128MB.
    heapSoftLimitMb = 128,

    # Generate a heap snapshot when the limit is reached.
    # Generate no more than this number of snapshots...
    heapSnapshotNearHeapLimit = 10,

    # Whenever the heap limit is reached, multiply the limit
    # by this factor and continue.
    heapLimitMultiplier = 2,

    # The number of times the heap limit can be exceeded before
    # the isolate is terminated.
    heapLimitExceedsMax = 1,
  )
);
```


src/workerd/server/isolate-limit-enforcer.c++
```diff
@@ -0,0 +1,227 @@
...
+        heapSnapshotNearHeapLimit(limits.getHeapSnapshotNearHeapLimit()),
+        heapLimitMultiplier(limits.getHeapLimitMultiplier()),
+        heapLimitExceedsMax(limits.getHeapLimitExceedsMax()) {}
+
+  v8::Isolate::CreateParams getCreateParams() override {
+    v8::Isolate::CreateParams params;
+    uint64_t softLimit = softHeapLimitMb * 1024 * 1024;
+    if (softLimit > 0) {
+      params.constraints.set_max_young_generation_size_in_bytes(
+          kj::min(softLimit, 2 * 1024 * 1024));
```
harrishancock, line 88: Should we make the `2` be configurable, too?

src/workerd/server/isolate-limit-enforcer.c++
```diff
@@ -0,0 +1,227 @@
...
+    KJ_DEFER(self.inNearLimitCallback = false);
+
+    if (self.exceededCounter >= self.heapLimitExceedsMax) {
+      KJ_LOG(ERROR, "Exceeded the configured hard heap limit.",
+             currentHeapLimit,
+             self.exceededCounter);
+      self.maybeGenerateHeapshot();
+      self.v8Isolate->TerminateExecution();
+    } else {
+      size_t newLimit = currentHeapLimit * self.heapLimitMultiplier;
```
harrishancock, line 112: Every code path calculates this, so might as well hoist up top.

harrishancock (commented):
It seems very useful to me. We could bikeshed the API, I suppose, as the new `limits` config field is sort of a tiny DSL.

Perhaps we could rename `limits` to `experimentalLimits` so it's clear it might evolve, merge, and play with it?

jasnell:
> ... We could bikeshed the API, I suppose, as the new limits config field is sort of a tiny DSL.

Yeah, I didn't bother putting a ton of effort into the design on this bit yet. Renaming to clearly mark it as experimental makes sens

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,40 @@ struct Worker {
 
   moduleFallback @13 :Text;
 
+  struct Limits {
```
kentonv, line 604: I'm concerned that this limits structure offers too much configurability, which has two problems:
* We'll be obliged to support it forever, which may be difficult if the way heap enforcement works in V8 changes. Like, what if they deprecate the near-heap-limit callback and it's no longer possible to apply a multiplier to the limit there? What if the GC algorithm changes and "young generation size" is no longer an option? (In fact, when the workers runtime was first written, the GC algorithm was different and had entirely different options. It can change again.)
* It's probably confusing for developers who don't actually want to control things in so much detail. Why would someone want to change the multiplier or the number of times it is applied? Who would ever want to tune the young generation size?

Can we make this more abstract? I'd suggest two settings:

* The heap limit.
* A boolean or enum that chooses between warn and terminate.
* A boolean or enum that chooses whether to dump a heap snapshot.

src/workerd/server/isolate-limit-enforcer.c++
```diff
@@ -0,0 +1,236 @@
...
+
+    if ((self.exceededCounter >= self.heapLimitExceedsMax) ||
+        (self.heapHardLimitMb > self.softHeapLimitMb &&
+         newLimit > self.heapHardLimitMb * 1024 * 1024)) {
+      KJ_LOG(ERROR, "Exceeded the configured hard heap limit.",
+             currentHeapLimit,
+             self.exceededCounter,
+             self.heapHardLimitMb);
+      self.maybeGenerateHeapshot();
+      self.v8Isolate->TerminateExecution();
```
kentonv, line 116: This will only terminate any currently-executing code, but will not cause the isolate to shut down or be replaced. For workerd, it probably makes more sense to kill the whole process here.

src/workerd/server/isolate-limit-enforcer.c++
kentonv: I actually think we shouldn't. We shouldn't offer config options that no one actually wants or needs. Doing so only overwhelms the user and locks us into supporting those options forever.

src/workerd/server/workerd.capnp
kentonv: I'd suggest if the user chooses "warn" mode, when the near-heap-limit callback runs and writes the warning, it should then change the limit to unlimited.

kentonv: To head off the obvious question:

"What if people want to emulate our production behavior exactly, where an isolate that has gone over the limit is permitted to complete in-flight requests, as long as it doesn't go excessively over?"

I don't think that's a behavior people should be testing against. This behavior is designed to be lenient to code that is already deployed. But when testing, developers should enforce the limit strictly, and design their app so it never exceeds that limit in the first place. Moreover, we might change our production behavior in the future, probably making it a bit more strict, but probably in a nuanced way.

src/workerd/server/isolate-limit-enforcer.c++
jasnell: Yeah I was considering different ways of handling this and shutting down the whole process after the final snapshot is taken would make sense.

src/workerd/server/isolate-limit-enforcer.c++
jasnell: Well, I'm not sure we can yet say no one actually wants or needs it yet ;-) ... there's no rush to land this so we can look at various options. I'd like to get some feedback from the devprod folks before we lock down any of the configuration options here, as well as see if we can solicit some input from users.

src/workerd/server/workerd.capnp
kentonv: Can you explain the use case for allowing the heap limit to be exceeded multiple times?

jasnell: > Can you explain the use case for allowing the heap limit to be exceeded multiple times?

For memory leak analysis I want to be able to generate 2 or more heap-snaphots at various points in order to generate a comparison view. In most cases I likely only need 2 (which is why the defaults are set the way they are here now) but generating a third or fourth snapshot can be help in some cases.

kentonv: Oh I see. But is generating these on near-heap-limit callbacks actually the best approach? I've more commonly seen time-based, that is, take a snapshot, wait 10 seconds, take another one. It might also be nice if the snapshots can be taken between requests, to reduce the amount of temporary garbage that shows up in them...

What about having an API exposed to JavaScript for taking a snapshot, similar to `--expose-gc`?

src/workerd/server/isolate-limit-enforcer.c++
```diff
@@ -0,0 +1,236 @@
...
+    // and continue, even if it exceeds the configured hard limit
+    if (self.inNearLimitCallback) return newLimit;
+    self.inNearLimitCallback = true;
+    KJ_DEFER(self.inNearLimitCallback = false);
+
+    if ((self.exceededCounter >= self.heapLimitExceedsMax) ||
+        (self.heapHardLimitMb > self.softHeapLimitMb &&
+         newLimit > self.heapHardLimitMb * 1024 * 1024)) {
+      self.maybeGenerateHeapshot();
+      kj::throwFatalException(KJ_EXCEPTION(FAILED,
```
kentonv, line 112: Maybe wrap this in a `noexcept` IIFE, to make clear it is intended to crash the process?

I think it will anyway since all of V8 is effectively `noexcept`, and this is a callback from V8. But as written it _looks_ like this is trying to throw an exception rather than crash.

(On another note, I should really rename `throwFatalException` to just `throwException` to avoid confusion about what "fatal" means.)

src/workerd/server/workerd.capnp
jasnell: For time-based or snapshots between requests, using the devtools debugger is likely still the better approach or we can add an API for it similar to what Node.js provides.

I can see us allowing for multiple approaches here.

For reference to what Node.js allows...

* nearlimit snapshot via cli option: https://nodejs.org/docs/latest/api/cli.html#--heapsnapshot-near-heap-limitmax_count
* Generate snapshot when signal received: https://nodejs.org/docs/latest/api/cli.html#--heapsnapshot-signalsignal
* API for taking snapshots - https://nodejs.org/docs/latest/api/v8.html#v8writeheapsnapshotfilenameoptions
* API for setting nearlimit snapshot: https://nodejs.org/docs/latest/api/v8.html#v8setheapsnapshotnearheaplimitlimit

jasnell: For now, let's defer discussion of a possible API-based approach for generating snapshots to a separate PR.

I have no particularly strong opinions on exactly what the config options here should be beyond making it possible to define some reasonable threshold and getting 1+ heap snapshots generated automatically when we hit the threshold. So I'm open to specific suggestions on which set of config options to use.

jasnell:
I've added the requirement that this can only be used when using `--experimental` and put clear indications in the comments in the capnp that it is experimental and subject to change. Notably, I've indicated that there are no guarantees that several of the fields will continue to be supported long term. I think it's worth going ahead and getting this landed so that we can at least get some experience with it. We can tweak things as we go.

@mrbbot ... it would be great to get some feedback from the wrangler side on this.

The tl;dr is: set memory limits in the worker config, if the worker hits the limit, a heap snapshot is generated and the worker is allowed to keep going up until a given hard limit, at which time another heap snapshot will be generated and the process will exit. Warnings will be printed to the console whenever the limit is hit and whenever a snapshot is generated.

I want to make sure this is something that would be useful for local dev for more than just us.

In a follow up PR, I plan to experiment with an API for  taking heapsnapshots. I might end up just implementing the same API that Node.js exposes (requiring nodejs-compat) rather than reinventing the wheel.

jasnell:
Interesting, ok, just so I understand... rather than crashing the process when we hit the hard limit, we'd effectively terminate the worker and just start returning 500s? Like in production, the soft limit would "condemn" the worker and the hard limit would permanently terminate it so that the process is still running but 500s are returned. We'd still generate the snapshots. Am I understanding that correctly? As in the case in the production environment, would you expect workerd to automatically restart the worker with a fresh isolate?

jasnell:
Putting this back into draft status while I work on @mrbbot's feedback

PR #1512, opened 2023-12-29
[bazel] Do not compile wd_cc_library by default

fhanau:
Mark library targets as off-by-default. We do not ship any libraries as part of workerd releases or use them outside of bazel, so there is generally no need to build them explicitly. This allows us to not build library targets by default. Note that on macOS (and likely Windows) library targets will still be built as dependencies of other targets such as tests and workerd itself as `start_stop_lib` is not supported – on Linux binaries are built directly from object files so there's no need to build the library files.
This reduces the size of the build output folder from 6.1GB to 5.9GB on Linux CI builds.

PR #1628, opened 2024-02-06
Revert "[bazel] Do not compile wd_cc_library by default"

mikea:
this is very incovenient. `bazel build` commands pass without doing anything, while everything is broken.

Reverts cloudflare/workerd#1512

### DIFF
--- a/build/wd_cc_library.bzl
+++ b/build/wd_cc_library.bzl
@@ -1,10 +1,9 @@
 """wd_cc_library definition"""
 
-def wd_cc_library(strip_include_prefix = "/src", tags = ["off-by-default"], **kwargs):
+def wd_cc_library(strip_include_prefix = "/src", **kwargs):
     """Wrapper for cc_library that sets common attributes
     """
     native.cc_library(
         strip_include_prefix = strip_include_prefix,
-        tags = tags,
         **kwargs
     )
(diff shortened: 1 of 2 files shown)


fhanau:
It is inconvenient – I noticed this issue after `bazel build //src/workerd/io` became a no-op, so I didn't apply this to the equivalent function in the downstream repo. I didn't look at reverting the workerd change since `--build_tag_filters=` represents a workaround and I felt unsure if others were also building like this.
At the same time, building the archive files themselves is clearly redundant as bazel prefers to use object files directly when building binaries unless `start_end_library` is enabled. On CI, not building the archives feels valuable due to space constraints. Maybe we can introduce something like a `no-ci` tag to handle this?

jasnell (approved):

mikea:
> Maybe we can introduce something like a `no-ci` tag to handle this?

would `tags = ["manual"]` on archives solve the issue?

fhanau:
manual would solve this for some use cases: `bazel build //src/workerd/io` would work but `bazel build //src/workerd/io/...` only builds the io library because some tests in the directory depend on it, so in cases where only a library target is defined in a directory the //... syntax would fail to build it.
TBF there might not be an elegant solution here, I'm open to reverting it.

fhanau (approved):

PR #1631, opened 2024-02-07
Implement heap snapshot memorytracker for http types

jasnell:


src/workerd/api/http.h
```diff
@@ -1096,6 +1159,9 @@ class FetchEvent: public ExtendableEvent {
 
   void visitForGc(jsg::GcVisitor& visitor) {
     visitor.visit(request);
+    KJ_IF_SOME(respondWithCalled, state.tryGet<RespondWithCalled>()) {
+      visitor.visit(respondWithCalled.promise);
+    }
```
hoodmane, line 1164: Does this fix a memory leak?

hoodmane (approved):

src/workerd/api/http.h
jasnell: It might. If the `promise` is left dangling and any closures attached to it happen to create a cycle then yes, it could have leaked. A big part of the reason I'm adding this memorytracker mechanism is to help find anything potentially like this.

PR #1632, opened 2024-02-07
Start migrating to kj::Rc/kj::rc

jasnell:
The new kj::Rc<T> has landed in capnp/kj... it is intended to clarify usage of refcounted types. This change is a first step in migrating away from kj::Own<T>/kj::refcounted, largely to start helping to identify if there are any issues to address before a larger-scale migration.

Question: For `IoOwn<T>` ... should we continue to just have the single `IoOwn<T>` or should we consider introducing an `IoRc<T>` to mirror `kj::Rc<T>`?


kentonv:
Given the number of bugs and incidents that have been caused by past aesthetic refactorings (e.g. coroutine cleanup), do we really want to embark on another? Can we just say that it's up to engineers making changes to particular parts of the code in the future to decide if converting it over to Rc at that time helps them understand it better?

PR #1634, opened 2024-02-07
Python: add some tests

hoodmane:


src/workerd/server/tests/python/hello/worker.py
```diff
@@ -0,0 +1,4 @@
+def test():
+    # This just tests that nothing raises when we run this. It isn't great though
+    # because we don't test whether we printed anything.
+    print("Does this work?")
```
dom96, line 4: Does wd-test not give us the ability to check the response like we have with ew-test?

src/workerd/server/tests/python/langchain/langchain.wd-test
```diff
@@ -0,0 +1,19 @@
...
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "python-hello",
+      worker = (
+        modules = [
+          (name = "worker.py", pythonModule = embed "./worker.py"),
+          (name = "aiohttp", pythonRequirement = "aiohttp"),
+          (name = "ssl", pythonRequirement = "ssl"),
+          (name = "langchain==0.0.339", pythonRequirement = ""),
+          (name = "openai==0.28.1", pythonRequirement = ""),
```
dom96, line 12:
```suggestion
          (name = "langchain", pythonRequirement = ""),
          (name = "openai", pythonRequirement = ""),
```

src/workerd/server/tests/python/langchain/worker.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test():
+    from langchain.chat_models import ChatOpenAI
+    import openai
+
+    API_KEY = "sk-abcdefgh"
+
+    ChatOpenAI(openai_api_key=API_KEY)
+    print("OK?")
```
dom96, line 15: it would be nice to have a way to verify that we got to this part of the code... I guess no exceptions is good enough for now though

Probably want to remove this `print`

dom96 (approved):

src/workerd/server/tests/python/hello/worker.py
jasnell: Not directly. The way I've done this with other tests is to set it up so that the test itself runs in a subrequest binding, then use the test handler itself to invoke the subrequest. Look, for instance, at the abortable-fetch tests as an example.

src/workerd/server/tests/python/langchain/langchain.wd-test
hoodmane: I think this breaks it doesn't it?

src/workerd/server/tests/python/hello/worker.py
hoodmane: Thanks for the tip!

hoodmane: I added some TODOs for this.

src/workerd/server/tests/python/langchain/langchain.wd-test
dom96: I suppose in workerd it does. But then we should pin the requirements in workerd's code, not here. If we don't then workerd will have different behaviour to edgeworker (since in edgeworker it is pinned via the packages we embed)

hoodmane: Well I just wanted to get some tests into workerd so that I have a place to stick more things. I agree that this should be fixed.

PR #1636, opened 2024-02-08
Implement heap snapshot memory tracker for r2

jasnell:


src/workerd/jsg/jsg.h
```diff
@@ -945,15 +945,17 @@ struct Dict {
     Value value;
 
     JSG_MEMORY_INFO(Field) {
-      tracker.trackField(name);
-      tracker.trackField(value);
+      tracker.trackField("name", name);
+      tracker.trackField("value", value);
```
hoodmane, line 949: Maybe "key" and "value" would be slightly more standard?

hoodmane (approved):

src/workerd/jsg/jsg.h
jasnell: possibly but I want to keep the names aligned with the actual fields as much as possible.

PR #1638, opened 2024-02-08
Use `IncomingRequestCfProperties` in Pages `EventContext`

### LINKED ISSUE #1327: EventContext.request don't support the cf object inside a PagesFuncion
In a pages function when the handlers are typed `PagesFunction<ENV>`, it throws a type error because EventContext.request doesn't have `cf` object.

https://github.com/cloudflare/workerd/blob/1be127453b0cc6e2e0c14351f9a694eb60d38a41/types/defines/pages.d.ts#L3-L12

mrbbot:
Whilst I wasn't able to reproduce the error from #1327, it did surface that the incoming request doesn't have the correct `cf` type. Currently, it's implicitly typed with `CfProperties` which includes both incoming and outgoing types, whereas we know this is an incoming request, so only `IncomingRequestCfProperties` are valid. This PR fixes that.

Closes #1327

### DIFF
--- a/types/defines/pages.d.ts
+++ b/types/defines/pages.d.ts
@@ -1,7 +1,7 @@
 type Params<P extends string = any> = Record<P, string | string[]>;
 
 type EventContext<Env, P extends string, Data> = {
-  request: Request;
+  request: Request<unknown, IncomingRequestCfProperties<unknown>>;
   functionPath: string;
   waitUntil: (promise: Promise<any>) => void;
   passThroughOnException: () => void;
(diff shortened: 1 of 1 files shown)


jasnell (approved):

types/defines/pages.d.ts
```diff
@@ -1,7 +1,7 @@
 type Params<P extends string = any> = Record<P, string | string[]>;
 
 type EventContext<Env, P extends string, Data> = {
-  request: Request;
+  request: Request<unknown, IncomingRequestCfProperties<unknown>>;
```
penalosa, line 4: What's this `unknown`? Should it be a type parameter?

mrbbot: `CfHostMetadata`, not sure if you can have Workers for Platforms with Pages so don't think it applies.

penalosa (approved):

PR #1642, opened 2024-02-08
Explicitly avoid propagating the async context into DO constructors

jasnell:
This is implicit now due to the fact that the async context does not propagate in certain cases involving custom thenables but as we move forward in fixing that limitation for custom thenables we need to make it explicit.


jasnell:
hmm, ok, looks like more needs to be investigated. The key issue this PR is looking to address, btw, is the fact that if we update v8 to propagate async context via custom thenables then the DO constructor can end up receiving an async context with a trace span bound to the wrong `IoContext`, causing an `IoOwn`-dereferencing exception when anything in the DO constructor tries to perform i/o (like a fetch).

jasnell:
Going to close this PR for now. Need to dig into the v8 patch a bit more

PR #1643, opened 2024-02-08
Implement heap snapshot memory tracking for scheduled, sockets, sql

jasnell:


src/workerd/api/sockets.h
```diff
@@ -167,7 +189,10 @@ class Socket: public jsg::Object {
   };
 
   void visitForGc(jsg::GcVisitor& visitor) {
-    visitor.visit(readable, writable);
+    visitor.visit(readable, writable, closedResolver,
+                  closedPromiseCopy, closedPromise,
+                  openedResolver, openedPromiseCopy,
+                  openedPromise);
```
jasnell, line 195: Entirely possible that there was a memory leak here previous with the promises not being traced.

src/workerd/api/sockets.h
```diff
@@ -113,6 +116,25 @@ class Socket: public jsg::Object {
     JSG_METHOD(startTls);
   }
 
+  void visitForMemoryInfo(jsg::MemoryTracker& tracker) const {
+    tracker.trackFieldWithSize("connectionStream",
+        sizeof(IoOwn<kj::RefcountedWrapper<kj::Own<kj::AsyncIoStream>>>));
```
hoodmane, line 121: Another question: would it work to say `sizeof(connectionStream)` here so you don't have to duplicate the type? If so, is there a reason to prefer explicitly writing the type here?

hoodmane (approved):

src/workerd/api/sockets.h
jasnell: It might. I prefer this approach since it's visually clearer what I'm measuring and that it's distinct from the other fields. But happy with whatever works.

PR #1644, opened 2024-02-08
Python/fastapi: Update asgi.py to pass request environment

hoodmane:
This makes the env parameter available in fastapi, though in a somewhat obscure way. Will have to investigate how to improve the user interface for this. Tested downstream.

cc @garrettgu10 @dom96


dom96 (approved):

irvinebroque:
@hoodmane — if we're going to do this — can we document it in https://github.com/cloudflare/cloudflare-docs/pull/12887 ?

PR #1645, opened 2024-02-08
Python: update to Pyodide 0.26.0a1 / Python 3.12

hoodmane:


fhanau:
This would be a good opportunity to bump the python version for rules_python too – I think it's primarily used for build scripts, but it would be nice to keep things consistent. See `python_register_toolchains` in WORKSPACE.

dom96 (approved):

PR #1648, opened 2024-02-09
Fixup gc tracing in EventTarget

jasnell:
No need to visit the native handler ref here.

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -658,7 +658,7 @@ void EventTarget::visitForGc(jsg::GcVisitor& visitor) {
           visitor.visit(js);
         }
         KJ_CASE_ONEOF(native, EventHandler::NativeHandlerRef) {
-          visitor.visit(native.handler);
+          // This is just a ref to the native handler, no need to visit.
         }
       }
     }


fhanau (approved):

kentonv:
@jasnell @fhanau This looks wrong. The native handler is a reference because there is some C++ type holding ownership such that when it is dropped, the NativeHandler unregisters itself. However, from the JavaScript perspective, the contents of the NativeHandler are reachable only from JS; the C++ owner does not intend to invoke them. Therefore, we should still be visiting the handles from JavaScript, so that they can be collected if the EventTarget itself goes away.

That said it appears `NativeHandler` is not used anywhere in our code today. Perhaps it should be removed?

kentonv:
Oh, `newNativeHandler()` is never used, but there is a direct use of `NativeHandler` within the same files, to implement `JavaScriptHandler::abortHandler`.

This is all very convoluted.

There are two EventTargets: The EventTarget on which an abortable event is being registered, and the AbortSignal that is used to abort the other event. We internally register a NativeHandler on the AbortSignal to signal us when it is aborted, at which point we unregister the abortable event on the other EventTarget.

Now, the NativeHandler will be visited via the `abortHandler` reference whenever the main EventTarget is visited. And when the main EventTarget is destroyed, then the NativeHandler itself is unregistered from the AbortSignal, so is no longer reachable through it. You might argue, therefore, that we don't need to visit the NativeHandler when the AbortSignal is visited (which is the change made by this PR).

This is, unfortunately, not correct. The problem is, objects are not destroyed immediately upon becoming unreachable. The GC may defer destruction until some time later. However, you are prohibited from accessing handles that have already been determined to be unreachable, even if their owning objects aren't destroyed yet. So, the main EventTarget could become unreachable while the AbortSignal still exists. But, before the EventTarget is actually destroyed, the AbortSignal might fire. It will then attempt to dereference the callback function, which is unreachable. This will crash.

PR #1651, opened 2024-02-09
Remove the non-implemented waitUntil() method of DurableObjectState

joshthoward:


src/workerd/api/actor-state.h
```diff
@@ -475,7 +475,9 @@ class DurableObjectState: public jsg::Object {
   kj::Array<kj::StringPtr> getTags(jsg::Lock& js, jsg::Ref<api::WebSocket> ws);
 
   JSG_RESOURCE_TYPE(DurableObjectState, CompatibilityFlags::Reader flags) {
-    JSG_METHOD(waitUntil);
+    if (!flags.getNoDurableObjectWaitUntil()) {
```
jasnell, line 478: nit: it would probably be helpful to also add a comment to the implementation of `waitUntil` in `actor-state.c++` indicating that it is deprecated

kentonv:
FWIW it's not so much that the method isn't implemented as that its effect is moot: waitUntil() can extend an event context up to 30 seconds after the client disconnects, but Durable Objects already run for at least 60 seconds after the last client disconnects. So, waitUntil() makes no difference.

We can still remove it. But, just pointing out that as-is it is in fact fulfilling its contract (trivially).

a-robinson:
I kind of wonder if we ever should have bothered trying to hide/remove waitUntil at all. It's arguably more work to explain to people that they don't need it than it is to just let folks who are familiar with it use it and move on with their lives. People who are familiar with it will use it and be happy, and people who aren't will be fine because it's not needed.

But trying to explain the subversion of expectations that it isn't there because you don't need it just adds mental overhead for unclear benefit.

joshthoward:
I added James's requested comment and updated the text to indicate that it is unnecessary. 

I am in favor of removing it because we would like to build out an API reference for Durable Object objects, e.g. `DurableObjectState`. In that API reference it would be strange to document a method and then communicate that it isn't necessary. In that context I'd feel obligated to explain that it isn't necessary. 

However, I am not a Javascript developer, so I don't know who would be familiar with it or what their expectations are.

DaniFoldi:
Hi :wave:,

I've got a question about this change. My understanding is that I/O from Durable Objects isn't cancelled immediately when a handler finishes execution, as also stated above by @kentonv. Are these 30/60 seconds guaranteed, even if waitUntil wasn't used?

The documentation states, in limits (https://developers.cloudflare.com/durable-objects/platform/limits/#limits)
> Each incoming HTTP request or WebSocket message resets the remaining available CPU time to 30 seconds. This allows the Durable Object to consume up to 30 seconds of compute after each incoming network request, with each new network request resetting the timer. If you consume more than 30 seconds of compute between incoming network requests, there is a heightened chance that the individual Durable Object is evicted and reset.

If the wording of this paragraph is intentional, there is no strict guarantee that objects wouldn't be evicted and reset prior to 30 seconds of CPU time (which is the same as walltime for DOs anyway), is that correct? What if a response was sent to a request, but there is still a promise in a waitUntil executing? How does this interact with runtime updates, or code updates? 

My observation here was that requests already in the queue are given to the old object, which has 30 seconds total remaining to handle them all, and new requests are queued and delivered to the new version once the old one has been evicted. In theory, would this mean that a code update and inefficient code (with a 30sec setTimeout, or any I/O) causes a 30-second "gap" in DOs handling requests?

Additionally, I'd like to provide a counterargument, hopefully showing that having a waitUntil method (even if it doesn't influence the eviction rules) is useful, in certain cases.

Imagine a logging library for Workers that gathers logs in an array and has an async method to flush them all - very straightforward to put into a waitUntil in a regular Worker. In DOs, the same can be done, providing waitUntil from `state` instead of `ctx`.

With this change, the usage of such functions would change, and differ between DOs and Workers, unfortunately. In terms of documentation, users would still have to be told _why_ waitUntil isn't necessary in DOs, so the absence of the method in my opinion doesn't decrease the complexity (in fact, the difference mentioned above might increase).

kentonv:
Hi @DaniFoldi, sorry but you seem to have introduced quite a few questions here, this would be better asked in a separate discussion thread. One thing I wanted to correct, though:

> 30 seconds of CPU time (which is the same as walltime for DOs anyway)

CPU time is not the same as wall time (for DOs or otherwise). CPU time is the time spent actively executing JavaScript code, not counting time when you are merely waiting for a response from the network or for a timer to expire. The 30-second limit on CPU time and the 30-second limit on waitUntil() wall time are entirely different things that happen to share the same time value.

kentonv:
@joshthoward I think I agree with @a-robinson that we might as well just leave this. @DaniFoldi provides another point here: that people might have common library code that they want to call from both stateless handlers and Durable Objects, and that library code might want to be given a `ctx` object that implements `waitUntil()`. It's sort of convenient that you can just pass the `state` object.

More generally, the DO `state` object and the stateless handler `ctx` object are similar in many ways, and this becomes even more obvious with the new class-based syntax for stateless handlers proposed here: https://github.com/cloudflare/workerd/pull/1658/commits/7f3c728171d2b1b04af37a5333fa228bf12689ca

People will be able to use exported classes to define regular stateless services (not DOs) using syntax like:

```
export class MyService extends StatelessService {
  constructor(ctx, env) { ... }
  async fetch(req) { ... }
}
```

You can see that `ctx` shows up here in exactly the place that `state` shows up for Durable Objects. Although these two types will never be exactly the same, it may make sense to try to converge them where possible.

Yes, the docs would indeed point out that `waitUntil()` doesn't actually have any effect for DO.

joshthoward:
Okay. I don't feel particularly strong about this one. Library code is a fair point.

PR #1652, opened 2024-02-09
Add heap snapshot memory tracker to api/url.h

jasnell:


src/workerd/api/url.h
```diff
@@ -106,6 +106,11 @@ class URL: public jsg::Object {
   // Treat as private -- needs to be public for jsg::alloc<T>()...
   explicit URL(kj::Url&& u);
 
+  void visitForMemoryInfo(jsg::MemoryTracker& tracker) const {
+    tracker.trackFieldWithSize("url", url->toString().size());
```
jclee, line 110: I guess technically these `toString()` calls are going to do some allocation, just to get the size.  Not sure if that's a potential issue.

jasnell: not ideal but also not too much of an issue. I can play with this more to avoid it.

jclee (approved):

PR #1656, opened 2024-02-10
Rough sketch of jsg::TracedRef

jasnell:
This is an extremely rough early sketch of a jsg::TracedRef<T> alternative to jsg::Ref<T> that includes debug time post-gc checks to see if the reference was actually traced. This is based on a quick conversation with Kenton about finding better ways of tracking down refs that should be traced but aren't. This is still very early just to start playing with the idea.

There are a few questions outstanding still...

1. Does this really need to be a separate type from Ref<T>? These essentially need to act like Ref<T> in the common case but it would also be nice to simplify the complexity around the sometimes strong, sometimes weak Ref<T> semantics.
2. Should these always be considered weak? Not sure it actually can be in case the parent has no strong refs.
3. This currently just uses a static thread local registry to track the trace state for warnings but is that ok? Can these end up being traced across multiple threads?

This PR is just intended to give us something to kick around.

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -1077,6 +1077,8 @@ private:
   friend void visitSubclassForMemoryInfo(const T* obj, MemoryTracker& visitor);
   template <typename T>
   friend class Ref;
+  template <typename T>
+  friend class TracedRef;
   friend class kj::Refcounted;
   template <typename T>
   friend kj::Own<T> kj::addRef(T& object);
@@ -1220,6 +1222,8 @@ private:
   template <typename>
   friend class ObjectWrapper;
   friend class GcVisitor;
+  template <typename>
+  friend class TracedRef;
 };
 
 template <MemoryRetainer T>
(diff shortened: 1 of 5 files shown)


src/workerd/jsg/jsg.h
```diff
@@ -1230,6 +1234,135 @@ void MemoryTracker::trackField(
   trackField(edgeName, value.get(), nodeName);
 }
 
+// A TraceRef is essentially a Ref<T> that is always weak and must be traced during GC.
+template <typename T>
+class TracedRef {
+public:
```
jasnell, line 1240: This initial iteration was literally just Do Something Quick That demonstrates the concept... and even then it's largely just a copy of `Ref<T>` with the additional debug book keeping.  We need to clarify what the semantics of this alternate-Ref would be and whether it's sufficiently different to warrant a separate class.

FWIW, my goal here is really not to do anything fundamentally different than Ref<T> but to improve our ability to detect when things that should be traced aren't. Something like this is one possible idea.

kentonv:
> Does this really need to be a separate type from Ref?

I think they have to be a different type because the point here is that the person declaring the ref is saying "This is a ref that is intended to be traced." Without that, we don't know if the ref is intended to be traced or not. Maybe it's intended to be a root ref, that is, held by C++ objects that aren't subject to GC, which is the default state of `Ref`.

But this calls into question: Why would someone remember to declare something as `TracedRef`, but forget to actually trace it?

Perhaps this needs to be paired with a rule -- checked by a linter -- that JSG objects are not allowed to contain bare `Ref`, they are required to hold only `TracedRef`. That way, the compiler warns you if you forget to use `TracedRef`, and then the extra checking tells you if your forget to actually trace it.

Then again, maybe such a linter could go ahead and examine your `visitForGc()` function to see if it traces your refs.

> Should these always be considered weak?

Note that the strong/weak terminology is kind of misleading here. "Strong" means "not traced", whereas "weak" actually means "traced". So a traced ref should always be considered "weak" by definition.

(This weird terminology comes from back when there was no explicit TracedReference type in V8, but instead you had a persistent handle, you marked it weak, and you traced the object it pointed to...)

> Not sure it actually can be in case the parent has no strong refs.

Not sure I understand this sentence. But keep in mind if the ref only knows its "parent" if it is traced.

> This currently just uses a static thread local registry to track the trace state for warnings but is that ok? Can these end up being traced across multiple threads?

Shouldn't these be associated with the isolate, not with a thread?

jasnell:
> But this calls into question: Why would someone remember to declare something as TracedRef, but forget to actually trace it?

I'm actually more concerned about what the semantics of `jsg::TracedRef<T>` are if the type holding it is actually never traced. For instance, let's say..

```
public:
  Foo(jsg::Ref<Bar> bar) : bar(kj::mv(bar)) {}
  JSG_RESOURCE_TYPE(Foo) {}

private:
  jsg::TracedRef<Bar> bar;
};

// ...

const bar = jsg::alloc<Bar>();
const foo = jsg::alloc<Foo>(kj::mv(bar));

// foo gets passed into a regular macro... but it's never passed out to JavaScript
```

The example is a bit silly but it's certainly possible for us to create and use `jsg::Object` types without ever attaching a wrapper to those. In this case, are the semantics of `jsg::TracedRef` just identical to `jsg::Ref`? ... or perhaps more broadly, are the semantics of `jsg::TraceRef<T>` different *at all* from `jsg::Ref<T>` when running in a release build?

jasnell:
Won't be moving forward with this right now. Switching everything over to use a new ref type to achieve the goal is going to be too disruptive

PR #1658, opened 2024-02-11
JSRPC: Service bindings and stateless class syntax

kentonv:
This fully supports RPC to service bindings, as opposed to Durable Objects.

Actually, service bindings already worked, by accident, as of #1617.

But this change does the following:
* Makes sure `env` and `ctx` get passed to RPC methods.
* Introduces an alternative class-based syntax for named entrypoints. If you export a class that extends `StatelessService`, we now understand it to be a stateless entrypoint, not a Durable Object. Meanwhile, you can extend `DurableObject` to explicitly mark an exported class as being a Durable Object class. Both of these are imported from the new module "cloudflare:entrypoints".
* Enables RPC server-side for classes that extend these new base classes, without the need for the `js_rpc` compat flag. (`experimental` is still needed for now.)
* Implements the new policy that only class methods (prototype properties) are exposed. Instance properties are not available over RPC. This is to defend people who forget to use private properties from accidentally exposing things over RPC that weren't intended.
* Adds "constructor" to the reserved name list. I was wrong about this before, it needs to be reserved. Sorry.


src/workerd/api/worker-rpc.h
```diff
@@ -115,7 +115,74 @@ class JsRpcSessionCustomEventImpl final: public WorkerInterface::CustomEvent {
...
+  JSG_RESOURCE_TYPE(DurableObjectBase) {}
+};
+
+// The "cloudflare:entrypoints" module, which exposes the StatelessService and DurableObject types
+// for extending.
+class EntrypointsModule: public jsg::Object {
+public:
+  void foo() {}
+  JSG_RESOURCE_TYPE(EntrypointsModule) {
+    JSG_METHOD(foo);
```
jasnell, line 168: is `foo` meant to be here?

src/workerd/jsg/setup.h
```diff
@@ -389,6 +389,25 @@ class Isolate: public IsolateBase {
...
+    // Returns the constructor function for a given type declared as JSG_RESOURCE_TYPE.
+    //
+    // Note there's a useful property of class constructor functions: A constructor's __proto__
+    // is set to the parent type's constructor. Thus you can discover whether one class is a
+    // subclass of another by following the __proto__ chain.
+    //
+    // TODO(cleanup): This should return `JsFunction`, but there is no such type. We only have
+    //   `jsg::Function<...>` (or perhaps more appropriately, `jsg::Constructor<...>`), but we
+    //   don't actually know the function signature so that's not useful here. Should we add a
+    //   `JsFunction` that has no signature?
```
jasnell, line 401: Yeah, I intentionally didn't add `JsFunction` because I wanted make sure we thought through the relationship with `jsg::Function`. We could, for instance have something like `jsg::Function<JsValue(kj::ArrayPtr<JsValue> args)>` or we could have a generic `jsg::JsFunction`. Not sure which is the best approach. We don't need to solve that now, tho.

src/workerd/api/worker-rpc.h
kentonv: Oops, this was me trying to figure out why named exports weren't working, and then I forgot to delete it. :facepalm:

kentonv:
(sorry, bad rebase, fixing)

kentonv:
Fixups:

* 0811199b
* 2fd5f22b
* c38465b4

kentonv:
Windows fixup:
* 8bfb1577

Dunno why only needed on Windows.

kentonv:
Fixup for missing return, not sure why only the internal build caught it.
* 07bc414c

kentonv:
Uhh, another fixup that undoes two previous fixups and tries a different way... :sweat_smile: 
* e32defb7

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -0,0 +1,180 @@
...
+    assert.strictEqual(x, undefined);
+    return new Response("method = " + req.method + ", url = " + req.url);
+  }
+}
+
+export class MyActor extends DurableObject {
+  #counter = 0;
+
+  constructor(state, env) {
+    super(state, env);
```
geelen, line 53: Is the superclass making properties with these names? I feel like it should for convenience, and because there's no risk of exposing them over RPC accidentally.

Either way, for this class, you can omit the constructor entirely

src/workerd/api/tests/js-rpc-flag.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2023 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import assert from 'node:assert';
+
+// This class does not extend `DurableObject`, but because we have the `js_rpc` compat flag on,
+// it'll still accept RPC.
```
geelen, line 8: You deleted the "js-rpc-disabled" test, should something like that still exist?

src/cloudflare/internal/entrypoints.d.ts
```diff
@@ -0,0 +1,7 @@
+export class DurableObject {
+  public constructor(state: unknown, env: unknown);
+}
+
+export class StatelessService {
+  public constructor(ctx: unknown, env: unknown);
```
geelen, line 6: The order of these arguments makes me deeply uncomfortable, but I see why you've done it.

Btw I don't think this is doing what you expect, TS is weird around asserting constructor arguments. However we might want to specify that both of these attach properties to their instance, if that's what they're going to do (otherwise why are they passed to super?)

geelen (approved):
I only have bikesheddy things to suggest, like not liking `cloudflare:entrypoints` (I think it should be `cloudflare:workers`) and thinking `Service` is fine rather than `StatelessService` (since it's being imported, not a global), but I don't feel all that strongly.

src/workerd/api/tests/js-rpc-test.js
kentonv: I think we still don't want to:
* We _are_ going to allow exposing properties as long as they are on the class, not instance properties. So we'd have to make special exceptions for these which is weird.
* I don't like them being automatically public to local code, even if they aren't exposed over RPC.
* I am not sure the names `state` and `env` (and `ctx` for stateless) are the best names. We've always used them in docs but nothing in the API actually mandates these specific names today. I'd rather not have to think about this.

src/workerd/api/tests/js-rpc-flag.js
kentonv: As of this commit, the main `js-rpc-test` no longer has the flag set, and tests all the behavior around that, making `js-rpc-disabled` redundant. I did however introduce `js-rpc-flag` which is the opposite test: it tests behavior when the flag is _enabled_.

src/cloudflare/internal/entrypoints.d.ts
kentonv: Yes, the ordering is awkward but it seems ore important to be consistent with `DurableObject` than to be consistent with the freestanding functions.

The superclass constructor requires these arguments (despite doing nothing with them) for a couple reasons:
* If the subclass doesn't declare a constructor, the superclass constructor will receive these params by defalut, so it seems like the right thing to do.
* I want to reserve the right to do something with them in the future even if we don't today.

jasnell (approved):

src/cloudflare/internal/workers.d.ts
geelen, line 1: I actually kinda think the internal file could still be `entrypoints`, and the public `cloudflare:workers` file could export a bunch of stuff from a few `cloudflare-internal` files?

src/workerd/api/worker-rpc.c++
```diff
@@ -173,14 +173,20 @@ public:
...
-          "The receiving Worker does not allow its methods to be called over RPC.");
 
       auto handler = KJ_REQUIRE_NONNULL(lock.getExportedHandler(entrypointName, ctx.getActor()),
                                         "Failed to get handler to worker.");
       auto handle = handler->self.getHandle(lock);
 
+      if (handler->missingSuperclass) {
+        // JS RPC is not enabled on the server side, we cannot call any methods.
+        JSG_REQUIRE(FeatureFlags::get(js).getJsRpc(), TypeError,
+            "The reciving Durable Object does not support RPC, because its class was not declared "
```
MellowYarker, line 184: `reciving` -> `receiving` since this is a user facing error.

src/cloudflare/internal/workers.d.ts
kentonv: Maybe but it doesn't matter, the file is internal and ideally I'd like to remove it entirely when built-in modules (defined in C++) start supporting named exports properly.

kentonv:
^ Fixed test failure

src/workerd/api/worker-rpc.c++
kentonv: fixup: 496e33c6

kentonv:
Added commits covering recent discussions, this should now be up to date with the current consensus.

src/cloudflare/internal/workers.d.ts
```diff
@@ -0,0 +1,13 @@
+export class DurableObject {
+  public constructor(ctx: unknown, env: unknown);
+
+  public ctx: unknown;
+  public env: unknown;
```
jasnell, line 5: Are these readonly?

src/workerd/api/worker-rpc.c++
```diff
@@ -365,4 +485,35 @@ kj::Promise<WorkerInterface::CustomEvent::Result>
...
+    const v8::FunctionCallbackInfo<v8::Value>& args,
+    jsg::Ref<ExecutionContext> ctx, jsg::JsObject env) {
+  // HACK: We take `FunctionCallbackInfo` mostly so that we can set properties directly on
+  //   `This()`. There ought to be a better way to get access to `this` in a constructor.
+  //   We *also* delcare `ctx` and `env` params more explicitly just for the sake of type checking.
+  jsg::Lock& js = jsg::Lock::from(args.GetIsolate());
+
+  jsg::JsObject self(args.This());
+  self.set(js, "ctx", jsg::JsValue(args[0]));
+  self.set(js, "env", jsg::JsValue(args[1]));
```
jasnell, line 501: I'm wondering if these shouldn't be defined as `readonly` instance properties on the instance. I also wonder if these should be marked non-enumerable. The former being more important.

jasnell (approved):
LGTM with one question relative to whether the fields should be readonly.

src/workerd/api/worker-rpc.c++
kentonv: Eh, I don't feel the need for these to be anything other than plain old properties as you'd get by assigning them in JS. They exist for the benefit of the subclass, so if the subclass decides to overwrite them, whatever.

MellowYarker (approved):

PR #1659, opened 2024-02-11
Reduce server-test runtime from >3min to <1.5s.

kentonv:
server-test has always featured a mock timer. The idea behind this is that the test can simulate the passage of time by advancing the timer explicitly, rather than waiting for real time to pass.

Apparently, there was some confusion about this, and we ended up with code in the test which tried to pass time by busy-looping while watching the real clock. This is completley unnecessary.


jasnell (approved):

kentonv:
(Simplified this change a bit.)

kentonv:
Windows failure seems unrelated?

MellowYarker (approved):

PR #1665, opened 2024-02-13
Add v8 patch to preserve async context for custom thenables

jasnell:
Patches v8 to allow async context to propgate to async thenables

@qard ... this probably doesn't tick all of the boxes v8 would need to land this (I skipped tests and there might be some additional debug asserts they'd want in a couple places) but this patch should cover the basics. Could I bother you to give it a once over? ... specifically this patch: `patches/v8/0015-Attach-continuation-context-to-Promise-thenable-task.patch`

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -434,6 +434,7 @@ http_archive(
         "//:patches/v8/0012-Fix-V8-ICU-build.patch",
         "//:patches/v8/0013-Randomize-the-initial-ExecutionContextId-used-by-the.patch",
         "//:patches/v8/0014-Always-enable-continuation-preserved-data-in-the-bui.patch",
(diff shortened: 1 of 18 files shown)


Qard (commented):
I believe storing the CPED data may also be required in `Factory::NewPromiseResolveThenableJobTask`. If I recall correctly, there were some weird cases like `new Promise((resolve) => resolve(thenable))` where the CSA changes weren't enough.

jasnell:
@Qard ... updated to include the factory.cc changes and extended the test to cover that case.

Qard (commented):
That should do it. LGTM. 👍🏻

jasnell:
Thank you @Qard !

kentonv (approved):

PR #1666, opened 2024-02-13
Move volatile-cache to workerd

jasnell:
First steps to moving volatile-cache to workerd. This will remain draft until the tests are also ported and the internal PR switching over to this is opened.

This is effectively a straightforward port of the version in the internal repo with a callback hook replacing some of the internal-specific additional limits we enforce when resizing.

Key questions to answer:

* [x] ~~Should we rename the API before it lands here? What should the name be?~~ Renamed to `MemoryCache` (at least for now)
* [x] ~~Should any changes be made to the API before it lands here? What changes?~~
  * ~~getInstance(...) method changed to also accept the `ownerId`~~
  * ~~virtual interface added to abstract the actual implementation of the cache map~~
* [x] ~~Should any changes be made to the binding configuration before it lands here? What changes?~~
  * ~~id field is now optional~~
* [x] ~~Should the implementation be updated to use memcache now or later?~~


src/workerd/server/workerd.capnp
```diff
@@ -386,6 +386,17 @@ struct Worker {
       unsafeEval @23 :Void;
       # A simple binding that enables access to the UnsafeEval API.
 
+      volatileCache :group {
+        # A binding representing access to an in-memory cache.
+
+        id @24 :Text;
```
kentonv, line 392: Can this be optional, if you don't set it you get a cache that isn't shared with anyone?

src/workerd/server/workerd.capnp
```diff
@@ -386,6 +386,17 @@ struct Worker {
       unsafeEval @23 :Void;
       # A simple binding that enables access to the UnsafeEval API.
 
+      volatileCache :group {
```
kentonv, line 389: I wonder if we want to think of a better name for this. "Volatile cache" is technically correct but sounds scary. Kind of tempted to just call it "memcache" (especially given it may eventually be backed by memcache) but I suppose that's a bit misleading at the moment...

src/workerd/server/workerd.capnp
jasnell: Probably. At least initially I wanted to make this a straightforward port of the internal impl without any changes. We could mark this experimental for now, make any changes we want to improve, then unflag?

src/workerd/server/workerd.capnp
irvinebroque: Agree — what seems most important to get across is that it is an in-memory cache — this answers "why use this instead of Cache API"?

And moment you say `memcached` I immediately grok what this is. And I understand that can be distributed, vs. volatileCache or just "in memory cache" I think volatile memory, think about cache for particular isolate, etc.

jasnell: How about `MemoryCache` as an alternative name for now. Later, we can hook this up to `memcached`.

src/workerd/server/workerd.capnp
jasnell: Made optional now.

jasnell:
Do we want to mark this experimental initially until we're ready to enable this in production or is landing it without the experimental flag ok here?

irvinebroque:
Really seems like needs to be experimental flag.

jasnell:
Given that this is renamed, it will no longer conflict with `VolatileCache` in the internal repo and shouldn't need to wait for the internal repo to be updated to land.

src/workerd/api/memory-cache.h
```diff
@@ -428,30 +428,16 @@ class MemoryCache: public jsg::Object {
...
+// bindings as needed. The default implementation (created and returned by createDefault())
+// uses a simple in-memory map to store the SharedMemoryCache instances.
+class MemoryCacheProvider {
 public:
-  MemoryCacheMap(
-      kj::Maybe<SharedMemoryCache::AdditionalResizeMemoryLimitHandler>
-          additionalResizeMemoryLimitHandler = kj::none)
-      : additionalResizeMemoryLimitHandler(kj::mv(additionalResizeMemoryLimitHandler)) {}
-  KJ_DISALLOW_COPY_AND_MOVE(MemoryCacheMap);
+  virtual SharedMemoryCache& getInstance(kj::StringPtr cacheId) const = 0;
```
kentonv: I know this isn't new, but I'm surprised to see this return a reference. It looks like memory caches are never wholly evicted today? That's going to be a big blocker to actually rolling out this feature. This really ought to return an atomic-refcounted object, and when the last reference is dropped (the last `Worker` using it is destroyed) then the cache should be evicted as well.

src/workerd/api/memory-cache.h
```diff
@@ -428,30 +428,16 @@ class MemoryCache: public jsg::Object {
   SharedMemoryCache::Use cacheUse;
 };
 
-// Data structure that maps unique cache identifiers to cache instances.
-// This allows separate isolates to access the same in-memory caches.
-class MemoryCacheMap {
+// The MemoryCacheProvider provides the internal implementation of the MemoryCache mechanism.
```
kentonv, line 431: What's the benefit of virtualizing `MemoryCacheMap` into `MemoryCacheProvider`? All this type implements is the mapping of cache IDs to `SharedMemoryCache` objects. If we introduced a new memcache-backed cache, this lookup would presumably still work the same? It's the `SharedMemoryCache` object itself which needs to be extensible. And even there, I don't think it would be accomplished by virtualization, but rather extending it to accept some sort of fallback cache storage that it checks before giving up and calling the callback.

src/workerd/api/memory-cache.h
```diff
@@ -434,7 +434,8 @@ class MemoryCache: public jsg::Object {
 // uses a simple in-memory map to store the SharedMemoryCache instances.
 class MemoryCacheProvider {
 public:
-  virtual SharedMemoryCache& getInstance(kj::StringPtr cacheId) const = 0;
+  virtual ~MemoryCacheProvider() noexcept(false) = default;
+  virtual SharedMemoryCache& getInstance(kj::StringPtr cacheId, uint32_t ownerId) const = 0;
```
kentonv, line 438: I believe that the way this already works internally is that the cache IDs are globally unique. There's no need to add `ownerId`, and in fact this might break cases where a cache is intentionally shared between owners (I don't know if any such cases exist today).

src/workerd/server/server.c++
```diff
@@ -2423,9 +2424,17 @@ static kj::Maybe<WorkerdApi::Global> createBinding(
...
+      // be optional?
+      KJ_REQUIRE(cache.hasLimits());
       Global::MemoryCache cacheCopy;
-      cacheCopy.cacheId = kj::str(cache.getId());
+      // The id is optional. If provided, then multiple bindings with the same id will
+      // share the same cache. Otherwise, a unique id is generated for the cache.
+      if (cache.hasId()) {
+        cacheCopy.cacheId = kj::str(cache.getId());
+      } else {
+        cacheCopy.cacheId = randomUUID(kj::none);
```
kentonv, line 2441: Instead of randomizing the ID in order to guarantee uniqueness, how about we just don't put the object into the map in the first place? We should construct a `SharedMemoryCache` object and give it directly to the specific Worker.

This does of course require that the Worker holds a strong reference to the `SharedMemoryCache`, which goes back to my other comment about making it atomic-refcounted.

samples/volatile-cache/README.md
kentonv, line 1: The directory name still says "volatile".

src/workerd/api/memory-cache.h
jasnell: Yep, I noted that also. Quite happy to make that change.

jasnell:
Marking draft until the feedback can be addressed!

src/workerd/api/memory-cache.h
jasnell: The keys are globally unique, yes, but it doesn't appear to be any form of validation. That is, if I'm lucky enough to guess an existing key in the binding configuration then I can get access to that existing cache. But this is something that needs to be evaluated before this lands... it's likely there's a very different way we should handle this.

src/workerd/server/server.c++
jasnell: Yep, was thinking exactly the same :-) ... Only went with this simply because of how the map *currently* works. If the cache is not shared, there's no reason for an id at all.

src/workerd/api/memory-cache.h
kentonv: I'd say, let's not make any movement in that direction until we know where we're going. Keep `MemoryCacheMap` as a concrete class for now.

src/workerd/api/memory-cache.h
kentonv: You mean the control plane API lets you specify any ID you want today? That sounds like a control plane issue which we would indeed need to fix before this is open to the public.

src/workerd/api/memory-cache.h
jasnell: Aside from this... I think we need to determine if the way limits are specified/suggested for the shared class via the binding configuration continues to make sense with the memcache-based implementation. If not, we might want to refactor that a bit here also.

src/workerd/api/memory-cache.h
jasnell: No idea what the control plane allows and I want to make sure that we are resilient that. If we are reasonably certain that the control plane ensures that these values are not arbitrarily passed through then we might be ok, but I'd want some verification of that to be on the safe side.

jasnell: ok, in offline communication we decide that it is ultimately the role of the control plane to ensure the uniqueness guarantee of a cache id. That's something we will follow up with separately. For now, I'll be changing this back to remove the additional `ownerId` arg.

src/workerd/api/memory-cache.c++
```diff
@@ -34,6 +35,15 @@ static bool hasExpired(const kj::Maybe<double>& expiration, bool allowOutsideIoC
...
 }
 
+SharedMemoryCache::SharedMemoryCache(
+    kj::Maybe<AdditionalResizeMemoryLimitHandler&> additionalResizeMemoryLimitHandler)
+    : data(),
+      additionalResizeMemoryLimitHandler(additionalResizeMemoryLimitHandler),
+      self(kj::refcounted<WeakRef<SharedMemoryCache>>(
+          kj::Badge<SharedMemoryCache> {}, *this)) {}
+
+SharedMemoryCache::~SharedMemoryCache() noexcept(false) { self->invalidate(); }
```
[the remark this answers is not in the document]
jasnell: Yeah I was afraid of that. Will rework this bit.

src/workerd/api/memory-cache.c++
```diff
@@ -0,0 +1,506 @@
...
+    return kj::atomicRefcounted<SharedMemoryCache>(provider, id, handler);
+  };
+
+  KJ_IF_SOME(cid, cacheId) {
+    auto lock = caches.lockExclusive();
+
+    // First, let's see if the cache already exists. If it does, we'll just return
+    // a strong reference to it.
+    KJ_IF_SOME(weak, lock->find(cid)) {
+      return weak->addRef();
```
kentonv, line 483: Again, this needs to be `kj::atomicAddRefWeak()`, because the destructor could already be running in another thread (blocked waiting for the `caches` mutex), in which case we cannot reuse the object.

Annoyingly if you find an entry in the map, but `atomicAddRefWeak()` fails, I suppose you have to replace the map entry with the new object. That in turn means that `removeInstance()`, below, actually has to check whether the map entry still points to the instance being removed, and avoid changing anything if not.

Also, it's theoretically possible that the new object is created and immediately destroyed, fully removing the map entry before the old instance even gets a chance to try to remove itself. That means that the old instance's `id` is now a dangling pointer. So you need to actually make a copy of the ID, it would seem. (But these objects are not constructed often, so it's fine.)

src/workerd/api/memory-cache.h
```diff
@@ -0,0 +1,497 @@
...
+// It is responsible for owning the SharedMemoryCache instances and providing them to the
+// bindings as needed. The default implementation (created and returned by createDefault())
+// uses a simple in-memory map to store the SharedMemoryCache instances.
+//
+// This class is virtualized for a number of reasons, the most important of which is that
+// we may want to replace the default implementation in production with one that includes
+// more metrics and monitoring, or that uses a different cache implementation. Ultimately,
+// however, if we do end up having additional implementations, it will be the SharedMemoryCache
+// class that will need to be updated.
+class MemoryCacheProvider {
```
kentonv, line 477: I still don't feel that this abstraction is adding value. I think we should just have `MemoryCacheMap` as a concrete type. I don't see how you could have any other implementation of this interface anyway.

jasnell:
Updated in 1ee1e0f

src/workerd/api/memory-cache.c++
```diff
@@ -0,0 +1,506 @@
...
+  // what we expect it to be.
+  auto lock = caches.lockExclusive();
+  KJ_IF_SOME(found, lock->find(instance.getId())) {
+    if (found != &instance) {
+      // Not the instance we expected it to be. Cache instance was likely replaced
+      // by a new instance with the same id. Do nothing.
+      return;
+    }
+  }
+  KJ_ASSERT(lock->erase(instance.getId()));
```
kentonv, line 503: You're doing a double lookup of the same key here, can avoid it like this:

```suggestion
  KJ_IF_SOME(entry, lock->findEntry(instance.getId())) {
    if (entry.value != &instance) {
      // Not the instance we expected it to be. Cache instance was likely replaced
      // by a new instance with the same id. Do nothing.
      return;
    }
    lock->erase(entry);
  }
```

src/workerd/api/memory-cache.h
```diff
@@ -0,0 +1,512 @@
...
+  ~SharedMemoryCache() noexcept(false);
+
+  kj::StringPtr getId() const { return id; }
+
+  static kj::Own<const SharedMemoryCache> create(
+    kj::Maybe<const MemoryCacheProvider&> provider,
+    kj::StringPtr id,
+    kj::Maybe<AdditionalResizeMemoryLimitHandler&> additionalResizeMemoryLimitHandler);
+
+  kj::Maybe<kj::Own<const SharedMemoryCache>> tryAddRef() const;
```
kentonv, line 160: Why not just inline this into the one call site?

It seems like providing `tryAddRef()` as a helper method isn't that useful, because the only place that would want to use the `tryAddRef()` version is the case where were trying to upgrade the weak reference in the map to a strong ref. All other places that want to addRef() a SharedMemoryCache will already have a strong ref and so won't need or want the "try" part.

kentonv (approved):

src/workerd/server/server.c++
```diff
@@ -2494,6 +2496,30 @@ static kj::Maybe<WorkerdApi::Global> createBinding(
       }
       return makeGlobal(Global::UnsafeEval {});
     }
+    case config::Worker::Binding::MEMORY_CACHE: {
+      if (!experimental) {
+        errorReporter.addError(kj::str(
+          "MemoryCache bindings are an experimental feature which may change or go away "
+          "in the future. You must run workerd with `--experimental` to use this feature."));
```
mikea, line 2503: `return kj::none`

src/workerd/server/server.c++
```diff
@@ -2494,6 +2496,30 @@ static kj::Maybe<WorkerdApi::Global> createBinding(
...
+      }
+      auto cache = binding.getMemoryCache();
+      // TODO(cleanup): Should we have some reasonable default for these so they can
+      // be optional?
+      KJ_REQUIRE(cache.hasLimits());
```
mikea, line 2508: let's add nicely formatted error when there are no limits too. it is a common complaint that workerd doesn't have nice error messages :)

mikea (approved):

PR #1668, opened 2024-02-14
Revert "Revert "Add implementations for getPreferredExtensions""

MellowYarker:
This reverts commit 25172d0c0a8415c284cdae0933b11a70f9813d58, and includes a bugfix in a caller of `getPreferredExtensions()`.

Capnp PR: https://github.com/capnproto/capnproto/pull/1943

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -39,10 +39,10 @@ apple_support_dependencies()
 
 http_archive(
     name = "capnp-cpp",
-    sha256 = "3b07a3b3f685ea5a8b0f4e27c336e26035c6efe7e23ef787a870fb978a633c24",
-    strip_prefix = "capnproto-capnproto-20be408/c++",
+    sha256 = "bed3c01caad985c91b58c8123cedfe46c4135d1698303d89c89bc49d3deb3697",
+    strip_prefix = "capnproto-capnproto-6480280/c++",
     type = "tgz",
-    urls = ["https://github.com/capnproto/capnproto/tarball/20be408666207623fbfd5002628455c41d15e3b9"],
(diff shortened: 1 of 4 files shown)


MellowYarker:
Need to fix the capnp dep, but will do that once the capnp PR is merged.

kentonv:
Normally it's best to separate the bugfix into a separate commit after the un-revert. Though in this case it's pretty easy to see.

kentonv (approved):

MellowYarker:
> Normally it's best to separate the bugfix into a separate commit after the un-revert. Though in this case it's pretty easy to see.

I agree, but was recently convinced it's better to not introduce breaking commits at all. In a situation like this, I guess the best of both worlds would be have separate commits during development and then squash into one before merging?

MellowYarker:
https://github.com/cloudflare/workerd/compare/5d650309f454b52ab2a79f02a982660b828a241f..cef4d968e89cbcc5aea46d7f0ab70097fcb2bfe4 updates the Cap'n Proto dep.

https://github.com/cloudflare/workerd/compare/cef4d968e89cbcc5aea46d7f0ab70097fcb2bfe4..ca2bb50963da82a489fe81747ca6995c3444efab is a rebase

kentonv:
> I agree, but was recently convinced it's better to not introduce breaking commits at all. In a situation like this, I guess the best of both worlds would be have separate commits during development and then squash into one before merging?

I normally make that argument too, but I think having clear reviewability outweighs the desire to have commits that aren't broken. And anyway, if you are reverting a previous rollback, that implies that equivalently broken commits existed in the past.

PR #1672, opened 2024-02-15
D1: Adding `.raw({columnNames})` types and tests for `.run()`

geelen:
Now https://github.com/cloudflare/workerd/pull/1586 is shipped, I wanted to update the type definitions for `.raw()` to include `{columnNames: true}` which retrun as a special case. I believe this is the best way to express it:

```ts
  raw<T = unknown[]>(options: {columnNames: true}): Promise<[string[], ...T[]]>;
  raw<T = unknown[]>(options?: {columnNames?: false}): Promise<T[]>;
```

This means if you pass anything other than `{columnNames: true}` you get the old type signature, i.e. just the rows. But it validates that the following syntax is correctly typed:

```ts
const [columns, ...rows] = await stmt
  .bind(1)
  .raw<[string, number, boolean]>({ columnNames: true });
expectType<string[]>(columns);
expectType<[string, number, boolean][]>(rows);
```

I also added an assertion that `.run()` doesn't return a `.results` field as per the docs and the recent bugfix.

### DIFF
--- a/types/defines/d1.d.ts
+++ b/types/defines/d1.d.ts
@@ -36,5 +36,6 @@ declare abstract class D1PreparedStatement {
   first<T = Record<string, unknown>>(): Promise<T | null>;
   run(): Promise<D1Response>;
   all<T = Record<string, unknown>>(): Promise<D1Result<T>>;
-  raw<T = unknown[]>(): Promise<T[]>;
+  raw<T = unknown[]>(options: {columnNames: true}): Promise<[string[], ...T[]]>;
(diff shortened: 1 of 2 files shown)


Jolg42:
I tried using directly `.raw({columnNames: true})` on a deployment, and it worked, but the types are indeed not in sync, so I was going to file an issue, and then I found this.
Many thanks @geelen for creating this PR 🙌🏼

mrbbot (approved):
LGTM! ✅

PR #1673, opened 2024-02-15
Distinguish initial alarm invocations from retries.

jqmmes:
This PR adds a new `runAlarm` parameter to keep track of alarm invocation retries.

### DIFF
--- a/src/workerd/io/worker-interface.capnp
+++ b/src/workerd/io/worker-interface.capnp
@@ -245,9 +245,10 @@ interface EventDispatcher @0xf20697475ec1752d {
   # the outcome and whether the run should be retried. This does not complete immediately.
 
 
-  runAlarm @4 (scheduledTime :Int64) -> (result :AlarmRun);
+  runAlarm @4 (scheduledTime :Int64, retryCount :UInt32) -> (result :AlarmRun);
   # Runs a worker's alarm.
   # scheduledTime is a unix timestamp in milliseconds for when the alarm should be run
+  # retryCount indicates the retry count, if it's a retry. Else it'll be 0.
   # Returns an AlarmRun, detailing information about the run such as
   # the outcome and whether the run should be retried. This does not complete immediately.
   #
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -3216,7 +3216,7 @@ void Worker::Actor::Impl::HooksImpl::updateAlarmInMemory(kj::Maybe<kj::Date> new
     for (auto i : kj::zeroTo(WorkerInterface::ALARM_RETRY_MAX_TRIES)) {
       co_await timerChannel.atTime(scheduledTime);
       auto result = co_await loopback->getWorker(IoChannelFactory::SubrequestMetadata{})
-          ->runAlarm(originalTime);
+          ->runAlarm(originalTime, i);
 
       if (result.outcome == EventOutcome::OK || !result.retry) {
         break;
@@ -3585,7 +3585,7 @@ public:
       kj::HttpConnectSettings settings) override;
   void prewarm(kj::StringPtr url) override;
   kj::Promise<ScheduledResult> runScheduled(kj::Date scheduledTime, kj::StringPtr cron) override;
-  kj::Promise<AlarmResult> runAlarm(kj::Date scheduledTime) override;
+  kj::Promise<AlarmResult> runAlarm(kj::Date scheduledTime, uint32_t retryCount) override;
   kj::Promise<CustomEvent::Result> customEvent(kj::Own<CustomEvent> event) override;
 
 private:
@@ -3810,8 +3810,8 @@ kj::Promise<WorkerInterface::ScheduledResult> Worker::Isolate::SubrequestClient:
   return inner->runScheduled(scheduledTime, cron);
 }
 kj::Promise<WorkerInterface::AlarmResult> Worker::Isolate::SubrequestClient::runAlarm(
-    kj::Date scheduledTime) {
-  return inner->runAlarm(scheduledTime);
+    kj::Date scheduledTime, uint32_t retryCount) {
+  return inner->runAlarm(scheduledTime, retryCount);
 }
 kj::Promise<WorkerInterface::CustomEvent::Result>
     Worker::Isolate::SubrequestClient::customEvent(kj::Own<CustomEvent> event) {
--- a/src/workerd/server/alarm-scheduler.c++
+++ b/src/workerd/server/alarm-scheduler.c++
@@ -151,9 +151,9 @@ bool AlarmScheduler::deleteAlarm(ActorKey actor) {
 }
 
 kj::Promise<AlarmScheduler::RetryInfo> AlarmScheduler::runAlarm(
-    const ActorKey& actor, kj::Date scheduledTime) {
+    const ActorKey& actor, kj::Date scheduledTime, uint32_t retryCount) {
   KJ_IF_SOME(ns, namespaces.find(actor.uniqueKey)) {
-    auto result = co_await ns.getActor(kj::str(actor.actorId))->runAlarm(scheduledTime);
+    auto result = co_await ns.getActor(kj::str(actor.actorId))->runAlarm(scheduledTime, retryCount);
 
     co_return RetryInfo {
       .retry = result.outcome != EventOutcome::OK && result.retry,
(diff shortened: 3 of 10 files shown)


src/workerd/io/worker-interface.capnp
```diff
@@ -214,9 +214,10 @@ interface EventDispatcher @0xf20697475ec1752d {
   # the outcome and whether the run should be retried. This does not complete immediately.
 
 
-  runAlarm @4 (scheduledTime :Int64) -> (result :AlarmRun);
+  runAlarm @4 (scheduledTime :Int64, isRetry :Bool) -> (result :AlarmRun);
   # Runs a worker's alarm.
   # scheduledTime is a unix timestamp in milliseconds for when the alarm should be run
+  # isRetry indicates wether this runAlarm is a retry or not.
```
a-robinson, line 220:
```suggestion
  # isRetry indicates whether this runAlarm is a retry or not.
```

src/workerd/io/worker.c++
```diff
@@ -3105,8 +3105,9 @@ void Worker::Actor::Impl::HooksImpl::updateAlarmInMemory(kj::Maybe<kj::Date> new
 
     for (auto i : kj::zeroTo(WorkerInterface::ALARM_RETRY_MAX_TRIES)) {
       co_await timerChannel.atTime(scheduledTime);
+      // TODO(now): Set correct retry info here.
```
a-robinson, line 3108: Leaving a comment so this is less likely to be forgotten.

a-robinson (commented):
This code does what it says, but what's the purpose? What's this value going to be used for?

a-robinson:
> This code does what it says, but what's the purpose? What's this value going to be used for?

And is there a reason for choosing a bool over a count of the number of past attempts?

DaniFoldi:
Hi :wave:,

With my somewhat limited c++ knowledge, I couldn't find any references to the js invocation of the alarm handler - this would be useful to have exposed to the handler for observability purposes. Is that something that could be added perhaps? It's the user code throwing that causes a retry anyways, so in theory it could be tracked by user code (try/catch, store and rethrow), but a native counter would be more reliable.

src/workerd/io/worker-interface.capnp
jqmmes: Currently, each retry is counted as a different alarm invocation, and we have no way to differentiate them. We want to be able to distinguish retries from regular invocation.

jqmmes:
> > This code does what it says, but what's the purpose? What's this value going to be used for?
> 
> And is there a reason for choosing a bool over a count of the number of past attempts?

We'll still record each and every invocation internally, I can make it as a counter if you think it's best. Then we would have something like:
```
invocation X retry 0
invocation X retry 1
invocation X retry 2
invocation X retry 3
...
```
Instead of 
```
invocation X retry false
invocation X retry true
invocation X retry true
...
```

src/workerd/server/alarm-scheduler.c++
```diff
@@ -166,7 +166,7 @@ kj::Promise<AlarmScheduler::RetryInfo> AlarmScheduler::runAlarm(
 
 AlarmScheduler::ScheduledAlarm AlarmScheduler::scheduleAlarm(
     kj::Date now, kj::Own<ActorKey> actor, kj::Date scheduledTime) {
-  auto task = makeAlarmTask(scheduledTime - now, *actor, scheduledTime);
+  auto task = makeAlarmTask(scheduledTime - now, *actor, scheduledTime, false);
```
jasnell, line 169:
```suggestion
  auto task = makeAlarmTask(scheduledTime - now, *actor, scheduledTime, false /* isRetry */);
```

src/workerd/server/alarm-scheduler.c++
```diff
@@ -265,7 +266,8 @@ kj::Promise<void> AlarmScheduler::makeAlarmTask(kj::Duration delay,
       entry.value.backoff++;
       entry.value.retry++;
 
-      entry.value.task = makeAlarmTask(delay, actorRef, scheduledTime);
+      // This alarm task is a retry. As such, we set isRetry to true.
+      entry.value.task = makeAlarmTask(delay, actorRef, scheduledTime, true);
```
jasnell, line 270:
```suggestion
      entry.value.task = makeAlarmTask(delay, actorRef, scheduledTime, true /* isRetry */);
```

jqmmes:
PR Updated:
- Applied reviewer suggestions.
- Converted retry info from `bool` to `uint32_t` counting the current retry number (0 if it's not a retry).

src/workerd/io/worker-interface.capnp
a-robinson: I was just trying to point out the spelling typo in the comment ("whether", not "wether")

src/workerd/server/alarm-scheduler.c++
```diff
@@ -265,7 +266,8 @@ kj::Promise<void> AlarmScheduler::makeAlarmTask(kj::Duration delay,
       entry.value.backoff++;
       entry.value.retry++;
 
-      entry.value.task = makeAlarmTask(delay, actorRef, scheduledTime);
+      // This alarm task is a retry. As such, we set isRetry to true.
+      entry.value.task = makeAlarmTask(delay, actorRef, scheduledTime, retryNumber++);
```
a-robinson, line 270: Why not just use `entry.value.countedRetry` for this rather than maintaining a new separate variable?

But if there's some reason to use `retryNumber` instead, don't we want to increment it for the retry *before* passing it in (i.e. `++retryNumber`) rather than after? If I'm reading this right it'd result in both the first attempt of the alarm and the first retry being sent with a retryNumber of 0

MellowYarker:
> With my somewhat limited c++ knowledge, I couldn't find any references to the js invocation of the alarm handler - this would be useful to have exposed to the handler for observability purposes. Is that something that could be added perhaps? It's the user code throwing that causes a retry anyways, so in theory it could be tracked by user code (try/catch, store and rethrow), but a native counter would be more reliable.

@DaniFoldi the `alarm()` handler is declared in [global-scope.h](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/global-scope.h#L205-L207) and we invoke it in [global-scope.c++](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/global-scope.c%2B%2B#L365-L469), but yes this is a good idea.

src/workerd/io/worker-interface.capnp
```diff
@@ -245,9 +245,10 @@ interface EventDispatcher @0xf20697475ec1752d {
   # the outcome and whether the run should be retried. This does not complete immediately.
 
 
-  runAlarm @4 (scheduledTime :Int64) -> (result :AlarmRun);
+  runAlarm @4 (scheduledTime :Int64, retryNumber :UInt32) -> (result :AlarmRun);
   # Runs a worker's alarm.
   # scheduledTime is a unix timestamp in milliseconds for when the alarm should be run
+  # isRetry indicates whether this runAlarm is a retry or not.
```
a-robinson, line 251: This comment needs to be updated to refer to `retryNumber` rather than `isRetry`

src/workerd/server/alarm-scheduler.c++
```diff
@@ -265,7 +266,8 @@ kj::Promise<void> AlarmScheduler::makeAlarmTask(kj::Duration delay,
       entry.value.backoff++;
       entry.value.retry++;
 
-      entry.value.task = makeAlarmTask(delay, actorRef, scheduledTime);
+      // This alarm task is a retry. As such, we set isRetry to true.
```
a-robinson, line 269: Another reference to `isRetry` that should be updated

src/workerd/server/alarm-scheduler.c++
```diff
@@ -197,7 +198,7 @@ kj::Promise<void> AlarmScheduler::makeAlarmTask(kj::Duration delay,
 
   auto retryInfo = co_await ([&]() -> kj::Promise<RetryInfo> {
     try {
-      co_return co_await runAlarm(actorRef, scheduledTime);
+      co_return co_await runAlarm(actorRef, scheduledTime, retryNumber);
```
a-robinson, line 201: Why pass in `retryNumber` vs using `entry.value.countedRetry` here as well? To be clear, I'm ok with it, but is it just a matter of convenience or is there a more meaningful reason?

PR #1677, opened 2024-02-16
D1: allow binding 'true' and 'false' by converting to 1 or 0

geelen:
Sadly there's no way when pulling the results out to stop SQLite returning 0 or 1 (since to it, a "boolean" is just a number: https://www.sqlite.org/datatype3.html), but at least this lets you use `true` and `false` to mean 1 or 0 (the way SQLite will allow you to use `TRUE` and `FALSE`).

This fixes https://github.com/cloudflare/workers-sdk/issues/3070

### DIFF
--- a/src/cloudflare/internal/d1-api.ts
+++ b/src/cloudflare/internal/d1-api.ts
@@ -214,9 +214,12 @@ class D1PreparedStatement {
   public bind(...values: unknown[]): D1PreparedStatement {
     // Validate value types
     const transformedValues = values.map((r: unknown): unknown => {
-      if (typeof r === 'number' || typeof r === 'string') {
+      const rType = typeof r;
+      if (rType === 'number' || rType === 'string') {
         return r
-      } else if (typeof r === 'object') {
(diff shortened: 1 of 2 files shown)


rozenmd (approved):

justin-mp (approved):

FlareLine:
Hey @geelen, for my own selfish benefit, is there any associated timeline on this one? This is unfortunately preventing me from using `boolean` types in my models do to difficulties during deserialization (field is returned as a `number` and not a `boolean`, causing some incosistencies / incompatibilities with my frontend). Thanks! 😁

FlareLine:
Is there any way of translating the results returned from SQLite based on the binding? Is the binding stored somewhere and could we potentially look the type up for translation? Primitive values like a `b oolean` seem fairly trivial to do this with, but obviously `Date`s might prove more difficult.

This kinda seems like a feature of an ORM on top of D1, but I feel like native boolean handling should be included in the native D1 implementation?

Keen to hear your thoughts.

jasnell (approved):

PR #1682, opened 2024-02-16
Fixup importing built-ins from module "subdirectories"

jasnell:
there's probably more elegant additional cleanups that can be made in here but this fixes the immediate issue.

### DIFF
--- a/src/workerd/api/tests/module-test.js
+++ b/src/workerd/api/tests/module-test.js
@@ -1,7 +1,7 @@
 import * as assert from 'a/b/c';
 import * as assert3 from 'node:assert';
 
-export const abortcontroller = {
+export const basics = {
   async test() {
     const assert2 = await import('a/b/c');
     if (assert !== assert2 && assert !== assert3) {
--- a/src/workerd/api/tests/module-test.js
+++ b/src/workerd/api/tests/module-test.js
@@ -1,0 +1,11 @@
+import * as assert from 'a/b/c';
+import * as assert3 from 'node:assert';
+
+export const abortcontroller = {
+  async test() {
+    const assert2 = await import('a/b/c');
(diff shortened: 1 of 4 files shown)


src/workerd/jsg/modules.c++
```diff
@@ -587,7 +605,16 @@ v8::Local<v8::Value> NodeJsModuleContext::require(jsg::Lock& js, kj::String spec
   auto modulesForResolveCallback = jsg::getModulesForResolveCallback(js.v8Isolate);
   KJ_REQUIRE(modulesForResolveCallback != nullptr, "didn't expect resolveCallback() now");
 
-  kj::Path targetPath = path.parent().eval(specifier);
+  kj::Path targetPath = ([&] {
+    // If the specifier begins with one of our known prefixes, let's not resolve
+    // it against the referrer.
+    if (specifier.startsWith("node:") ||
+        specifier.startsWith("cloudflare:") ||
+        specifier.startsWith("workerd:")) {
```
dom96, line 613: Non-blocking: would be nice to refactor this out into a separate function to avoid duplication.

dom96 (approved):
Nice!

src/workerd/jsg/modules.c++
jasnell: As a follow on to the module registry refactor I expect much of this to be cleaned up anyway so I'm not that concerned about the duplication right now.

src/workerd/api/tests/module-test.js
```diff
@@ -0,0 +1,11 @@
+import * as assert from 'a/b/c';
+import * as assert3 from 'node:assert';
+
+export const abortcontroller = {
```
jasnell, line 4:
```suggestion
export const basics = {
```

PR #1690, opened 2024-02-19
Cancel pending reads and writes on releaseLock

jasnell:
In the standard streams impl only (for now) cancel pending reads and writes when the reader/writer lock is released per the standard.


jasnell:
This will need an internal PR to fixup a test before this lands here.

harrishancock (approved):

PR #1692, opened 2024-02-19
JSRPC: Support passing RPC stubs across other RPCs

kentonv:
The first four commits (mainly the third commit) add a framework to allow JSG types to declare themselves serializable. @jasnell I know you were also working on something here so sorry if this overlaps. I did take some inspiration from your design doc. But, it occurred to me that instead of writing a tag and a version, we could combine these: whenever a type decides it needs a new serialization version, it allocates a whole new tag for it. This way we save some bytes on the wire.

The remaining commits, after some bug fixes and refactoring, introduce `RpcTarget` as a new type which classes can extend to mark themselves as being RPC objects -- without being a top-level entrypoint. Instances of these types can then be sent in RPC messages. The receiving side gets an `RpcStub` instead, which allows calls back to the original object.

Example:

```
import {WorkerEntrypoint,RpcTarget} from 'cloudflare:workers';

class Counter extends RpcTarget {
  i = 0;
  async increment(j) {
    this.i += j;
    return this.i;
  }
}

export class CounterService extends WorkerEntrypoint {
  async makeCounter() {
    return new Counter();
  }
}
```

```
let counter = await env.COUNTER_SERVICE.makeCounter();
await counter.increment(5);  // returns 5
await counter.increment(3);  // returns 8
```

NOT done in this PR, but needed before this is production-ready:
* resource management
  * close() methods
  * auto-close parameter caps after call returns
  * inject close method on returned object if it contains nested stubs
* promise pipelining
* streams, etc.

### DIFF
--- a/src/cloudflare/internal/workers.d.ts
+++ b/src/cloudflare/internal/workers.d.ts
@@ -11,3 +11,10 @@ export class WorkerEntrypoint {
   public ctx: unknown;
   public env: unknown;
 }
+
+export class RpcStub {
+  public constructor(server: object);
+}
+
+export class RpcTarget {
+}
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -174,14 +174,33 @@ kj::Maybe<JsRpcStub::RpcFunction> JsRpcStub::getRpcMethod(
   }));
 }
 
-// The capability that lets us call remote methods over RPC.
-// The client capability is dropped after each callRequest().
-class JsRpcTargetImpl final : public rpc::JsRpcTarget::Server {
+// Callee-side implementation of JsRpcTarget.
+//
+// Most of the implementation is in this base class. There are subclasses specializing for the case
+// of a top-level entrypoint vs. a transient object introduced by a previous RPC in the same
+// session.
+class JsRpcTargetBase: public rpc::JsRpcTarget::Server {
 public:
-  JsRpcTargetImpl(
-      IoContext& ctx,
-      kj::Maybe<kj::StringPtr> entrypointName)
-      : weakIoContext(ctx.getWeakRef()), entrypointName(entrypointName) {}
+  JsRpcTargetBase(IoContext& ctx)
+      : weakIoContext(ctx.getWeakRef()) {}
+
+  struct EnvCtx {
+    v8::Local<v8::Value> env;
+    jsg::JsObject ctx;
+  };
+
+  struct TargetInfo {
+    // The object on which the RPC method should be invoked.
+    v8::Local<v8::Object> target;
+
+    // If `env` and `ctx` need to be delivered as arguments to the method, these are the values
+    // to deliver.
+    kj::Maybe<EnvCtx> envCtx;
+  };
+
+  // Get the object on which the method is to be invoked. This is virtual so that we can have
+  // separate subclasses handling the case of an entrypoint vs. a transient RPC object.
+  virtual TargetInfo getTargetInfo(Worker::Lock& lock, IoContext& ioCtx) = 0;
 
   // Handles the delivery of JS RPC method calls.
   kj::Promise<void> call(CallContext callContext) override {
@@ -213,37 +232,25 @@ public:
 
       jsg::Lock& js = lock;
 
-      auto handler = KJ_REQUIRE_NONNULL(lock.getExportedHandler(entrypointName, ctx.getActor()),
-                                        "Failed to get handler to worker.");
-      auto handle = handler->self.getHandle(lock);
+      auto targetInfo = getTargetInfo(lock, ctx);
 
-      if (handler->missingSuperclass) {
-        // JS RPC is not enabled on the server side, we cannot call any methods.
-        JSG_REQUIRE(FeatureFlags::get(js).getJsRpc(), TypeError,
-            "The receiving Durable Object does not support RPC, because its class was not declared "
-            "with `extends DurableObject`. In order to enable RPC, make sure your class "
-            "extends the special class `DurableObject`, which can be imported from the module "
-            "\"cloudflare:workers\".");
-      }
-
-      // `handler->ctx` is present when we're invoking a freestanding function, and therefore
+      // `targetInfo.envCtx` is present when we're invoking a freestanding function, and therefore
       // `env` and `ctx` need to be passed as parameters. In that case, we our method lookup
       // should obviously permit instance properties, since we expect the export is a plain object.
       // Otherwise, though, the export is a class. In that case, we have set the rule that we will
       // only allow class properties (aka prototype properties) to be accessed, to avoid
       // programmers shooting themselves in the foot by forgetting to make their members private.
-      bool allowInstanceProperties = handler->ctx != kj::none;
+      bool allowInstanceProperties = targetInfo.envCtx != kj::none;
 
       // We will try to get the function, if we can't we'll throw an error to the client.
-      auto fn = tryGetFn(lock, ctx, handle, methodName, allowInstanceProperties);
+      auto fn = tryGetFn(lock, ctx, targetInfo.target, methodName, allowInstanceProperties);
 
       v8::Local<v8::Value> invocationResult;
-      KJ_IF_SOME(execCtx, handler->ctx) {
-        invocationResult = invokeFnInsertingEnvCtx(js, methodName, fn, handle, args,
-            handler->env.getHandle(js),
-            lock.getWorker().getIsolate().getApi().wrapExecutionContext(js, execCtx.addRef()));
+      KJ_IF_SOME(envCtx, targetInfo.envCtx) {
+        invocationResult = invokeFnInsertingEnvCtx(js, methodName, fn, targetInfo.target, args,
+            envCtx.env, envCtx.ctx);
       } else {
-        invocationResult = invokeFn(js, fn, handle, args);
+        invocationResult = invokeFn(js, fn, targetInfo.target, args);
       }
 
       // We have a function, so let's call it and serialize the result for RPC.
@@ -281,7 +288,7 @@ public:
     return kj::mv(paf.promise);
   }
 
-  KJ_DISALLOW_COPY_AND_MOVE(JsRpcTargetImpl);
+  KJ_DISALLOW_COPY_AND_MOVE(JsRpcTargetBase);
 
 private:
   // The following names are reserved by the Workers Runtime and cannot be called over RPC.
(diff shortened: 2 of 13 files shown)


src/workerd/jsg/ser.c++
```diff
@@ -43,6 +47,42 @@ void Serializer::ThrowDataCloneError(v8::Local<v8::String> message) {
...
 }
 
+v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::Object> object) {
+  try {
+    if (object->InternalFieldCount() != Wrappable::INTERNAL_FIELD_COUNT ||
+        object->GetAlignedPointerFromInternalField(Wrappable::WRAPPABLE_TAG_FIELD_INDEX) !=
+            &Wrappable::WRAPPABLE_TAG) {
+      // v8::ValueSerializer by default will send us anything that has internal fields, but this
+      // object doesn't appear to match the internal fields expected on a JSG object.
+      JSG_FAIL_REQUIRE(DOMDataCloneError, "Foreign host object could not be serialized.");
```
jasnell, line 57: non-blocking nit: Hmm... not sure if this is a great user-facing error message. They might not be aware of what a "Foreign host object" is. Fortunately this case shouldn't ever actually happen so it's probably ok.

src/workerd/jsg/ser-test.c++
```diff
@@ -0,0 +1,226 @@
...
+    kj::String getText() {
+      return kj::str(text);
+    }
+
+    JSG_RESOURCE_TYPE(Bar) {
+      JSG_READONLY_PROTOTYPE_PROPERTY(text, getText);
+    }
+
+    void serialize(jsg::Lock& js, jsg::Serializer& serializer) {
+      serializer.writeRawUint64(text.size());
```
jasnell, line 67: Is the assumption here that the serialize method always serializes to the most recent version `Tag` for this type?

src/workerd/jsg/ser-test.c++
```diff
@@ -0,0 +1,226 @@
...
+    JsRef<JsValue> getVal(Lock& js) {
+      return val.addRef(js);
+    }
+
+    JSG_RESOURCE_TYPE(Bar) {
+      JSG_READONLY_PROTOTYPE_PROPERTY(val, getVal);
+    }
+
+    void serialize(jsg::Lock& js, jsg::Serializer& serializer) {
+      // V2 just writes a value!
```
jasnell, line 133: Nit: a test variation that throws an exception in the `serialize(...)` and `deserialize(...)` methods and validates those are propagated correctly would be helpful.

src/workerd/jsg/ser.c++
kentonv: I also don't know what a "foreign host object" is, unfortunately! (I don't know what kinds of objects V8 puts internal fields on, though I'm pretty sure there is at least one type that it does this with.)

src/workerd/jsg/ser-test.c++
kentonv: The first argument to `JSG_SERIALIZABLE(...)` is the tag that `serialize()` produces, yes.

src/workerd/jsg/ser.h
```diff
@@ -13,18 +13,83 @@ namespace workerd::jsg {
...
+// * `MyTag` is some enum type declared in the application which enumerates all known serializeable
+//   types. This can be any enum, but it is suggested that all types in the application use the
+//   same enum type, and the numeric values of the tags must never change. A Cap'n Proto enum is
+//   suggested.
+// * `JSG_SERIALIZABLE`'s parameters are a list of tags that encode this type. There may be more
+//   than one tag listed to support versioning. The first tag listed is the current version, which
+//   is the format that `serialize()` will write. The other tags are non-current versions that
+//   `deserialize()` accepts in addition to the current version. These are usually old versions,
+//   but could also include a new version that hasn't fully rolled out yet -- it will be necessary
+//   to fully roll out support for parsing a new version before anyone can start generating it.
```
jasnell, line 45: nit: worth mentioning the `ONEWAY` variant of the macro here?

src/workerd/jsg/ser.h
```diff
@@ -13,18 +13,83 @@ namespace workerd::jsg {
...
+    virtual ~ExternalHandler() noexcept(false) = 0;
+  };
+
   struct Options {
     // When set, overrides the default wire format version with the one provided.
     kj::Maybe<uint32_t> version;
     // When set to true, the serialization header is not written to the output buffer.
     bool omitHeader = false;
+
+    kj::Maybe<ExternalHandler&> externalHandler;
```
jasnell, line 81: nit: a short note here referencing where the ownership of the `ExternalHandler&` ref is expected to be held would be helpful.

src/workerd/jsg/ser.h
```diff
@@ -33,20 +98,41 @@ class Serializer final: v8::ValueSerializer::Delegate {
...
   void transfer(Lock& js, const JsValue& value);
 
   Released release();
 
+  void writeRawUint32(uint32_t i) { ser.WriteUint32(i); }
+  void writeRawUint64(uint64_t i) { ser.WriteUint64(i); }
+
+  void writeRawBytes(kj::ArrayPtr<const kj::byte> bytes) {
+    ser.WriteRawBytes(bytes.begin(), bytes.size());
+  }
```
jasnell, line 123: non-blocking nit: as a convenience, it may be nice to have this automatically write the size as a prefix.

Also, a `writeString(kj::StringPtr str)` variation would probably be helpful at some point. Both can be done later tho.

src/workerd/jsg/ser.h
```diff
@@ -60,9 +146,16 @@ class Serializer final: v8::ValueSerializer::Delegate {
...
+  // Exactly like Serializer::ExternalHandler, but for Deserializer.
+  class ExternalHandler {
+  public:
+    virtual ~ExternalHandler() noexcept(false) = 0;
+  };
+
   struct Options {
     kj::Maybe<uint32_t> version;
     bool readHeader = true;
+    kj::Maybe<ExternalHandler&> externalHandler;
```
jasnell, line 158: nit: a comment here identifying the owner of the `ExternalHandler` instance would be helpful

src/workerd/api/worker-rpc.c++
```diff
@@ -153,80 +208,151 @@ kj::Maybe<JsRpcCapability::RpcFunction> JsRpcCapability::getRpcMethod(
   }));
 }
 
-// The capability that lets us call remote methods over RPC.
-// The client capability is dropped after each callRequest().
-class JsRpcTargetImpl final : public rpc::JsRpcTarget::Server {
+void JsRpcStub::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
+  auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
```
jasnell, line 212: side note: In general, I think `DOMDataCloneError` is the only error type we should be throwing from these. I wonder if there's a reasonable way of enforcing that..

jasnell:
Generally LGTM. Left some notes that can safely be handled later if you'd like.

src/workerd/jsg/ser.c++
jasnell: heh, maybe just leave the `Foreign host` part of the message off then? non-blocking either way.

src/workerd/jsg/ser-test.c++
jasnell: fwiw, I know this is covered a bit in the documentation in ser.h already, I just think it might need to be made more prominent. The note is a bit easy to miss.

jasnell (approved):

src/workerd/jsg/ser.c++
kentonv: I figure if someone reports the error to us, the words can at least help us figure out where to look.

src/workerd/jsg/ser-test.c++
kentonv: This should get a section in jsg/README.md once we've settled on the interface, but I suspect this first iteration won't actually be what we settle on, so I didn't want to over-document it just yet.

Having the serializer write its own tag would also eliminate the need for JSG_SERIALIZABLE_ONEWAY -- you'd just have JSG_SERIALIZABLE() instead. So maybe that is the best way. I'm just a little sad about the asymmetry -- the serializer writes its own tag but the parser doesn't read its own tag. But maybe that's fine.

I'd like to think about this a bit more before changing it, though. We can easily change this in a follow-up PR.

src/workerd/jsg/ser.h
kentonv: I don't think it should write the size as a prefix. You might be writing a fixed-size array, or something where the size is known based on some other property read before the bytes.

We could add a separate `writeLengthDelimited()` that writes a length first. (But these can easily be added as needed, as you say.)

src/workerd/api/worker-rpc.c++
kentonv: Hmm, I'm not sure I could say confidently that all serialization errors are DataCloneError. Probably not worth trying to enforce.

src/workerd/jsg/ser-test.c++
kentonv: 1780da8f

src/workerd/jsg/ser.h
kentonv: f1ee030f

geelen (approved):
LGTM. Only really reviewed the surface-level JS stuff & left a few comments, but nothing major.

src/cloudflare/internal/workers.d.ts
```diff
@@ -11,3 +11,10 @@ export class WorkerEntrypoint {
   public ctx: unknown;
   public env: unknown;
 }
+
+export class RpcStub {
+  public constructor(server: object);
+}
+
+export class RpcTarget {
```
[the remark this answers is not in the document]
kentonv: I thought about it a little bit but it's not clear to me. I think we may want to handle these in different ways. Consider that if you have a DO _stub_ (you're a client of it) and you want to send that stub to someone else over a separate RPC, we probably want to implement that by sending the actor ID, not the live capability. The receiving worker can connect to that actor ID themselves. Similarly if you send a service binding then we can send the worker ID instead of a live capability.

But what happens if you are not a client, but you're doing `return this` inside the object implementation? If we're going to say these things behave like RpcTarget, then `return this` should return a capability to the _specific_ instantiation? But it's weird that this is _different_ from what happens if you send a copy of the binding / DO stub?

This needs more thought and I think we should punt for now.

src/cloudflare/internal/workers.d.ts
```diff
@@ -11,3 +11,10 @@ export class WorkerEntrypoint {
   public ctx: unknown;
   public env: unknown;
```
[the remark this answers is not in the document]
kentonv: @irvinebroque did get feedback and the feedback was people weren't excited about a change, so we're not changing it.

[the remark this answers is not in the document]
kentonv: > Oh and don't these constructors need to set this.context and this.bindings here?

This is a `.d.ts` file, the constructors are not defined here.

kentonv:
Ugh botched a rebase again (my fixup script doesn't work correctly if `main` isn't synced, oops).

This is the change I actually made:
* 6cfa2fbf

This restores the default error messages when serializing a non-serializable object. Otherwise some tests broke in the internal build.

PR #1693, opened 2024-02-19
Log a warning message for long running storage operations when flushing

Frederik-Baetens:


src/workerd/io/actor-cache.c++
```diff
@@ -2634,6 +2641,7 @@ kj::Promise<void> ActorCache::flushImplUsingSingleCountedDelete(CountedDeleteFlu
   // reads before actually sending the write. The same exact logic applies here.
   co_await waitForPastReads();
   try {
+    durationAlert(clock, kj::Duration(kj::SECONDS*1), "storage operation took longer than 1s: counted delete");
```
Frederik-Baetens, line 2644: didn't add extra scope here because I figured the tine added by the fulfill call would be minimal.

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,26 @@
...
+public:
+  durationAlert(const kj::MonotonicClock& clock, kj::Duration logWarningTime, kj::StringPtr logMessage)
+    : logWarningTime(logWarningTime),
+      logMessage(logMessage),
+      start(clock.now()),
+      clock(clock) {}
+
+  ~durationAlert() {
+    if (start - clock.now() >= logWarningTime) {
+      KJ_LOG(WARNING, "NOSENTRY", logMessage);
```
jqmmes, line 17: nit: You can use `LOG_NOSENTRY` here.

jqmmes:
This LGTM, but I guess you need runtime approval. I find it strange that touching `actor-cache` doesn't have `Workers Durable Objects` as CODEOWNERS.

src/workerd/util/duration-alert.h
Frederik-Baetens: Hmm, I tried, but it doesn't seem to play nice with passing the logMessage as an argument rather than a string literal. Don't really know why that is. I'd assume b/c of it being a macro defined in terms of another macro perhaps?

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include "kj/string.h"
+#include "kj/time.h"
+#include <kj/debug.h>
+
+class durationAlert {
```
jclee, line 7: Class names are typically in PascalCase?

src/workerd/io/actor-cache.c++
```diff
@@ -2581,7 +2582,10 @@ kj::Promise<void> ActorCache::flushImplUsingSinglePut(PutFlush putFlush) {
   // See the comment in flushImplUsingTxn for why we need to construct our RPC and then wait on
   // reads before actually sending the write. The same exact logic applies here.
   co_await waitForPastReads();
-  co_await request.send().ignoreResult();
+  {
+    durationAlert(clock, kj::Duration(kj::SECONDS*1), "storage operation took longer than 1s: single put");
```
jclee, line 2586: I'm not sure this does what you're expecting...  I suspect the compiler is allowed to instantiate a temporary "durationAlert" value for the scope of the statement, then free it before doing the `co_await`.  If you want the value to exist for the scope of the code block, I think some further action is needed, like putting it in a named variable?

I guess one way to test this would be to add a `usleep(5000 * 1000);` statement in the block and see if the log fires.  Or alternately, to put in several `KJ_DBG()` statements to verify that the destructor runs after the end of the code block.

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,31 @@
...
+// thereby timing everything after the initialization of the timer, until the end of the scope where it was
+// instantiated.
+public:
+  durationAlert(const kj::MonotonicClock& clock, kj::Duration logWarningTime, kj::StringPtr logMessage)
+    : logWarningTime(logWarningTime),
+      logMessage(logMessage),
+      start(clock.now()),
+      clock(clock) {}
+
+  ~durationAlert() {
```
jclee, line 20: Per the style guide, explicit destructors should typically be declared `noexcept(false)`.

https://github.com/capnproto/capnproto/blob/master/style-guide.md#exceptions-usage

src/workerd/util/duration-alert.h
jclee: Yeah, `LOG_NOSENTRY` does its concatenation by putting two string literals next to each other to make a new string literal -- e.g. `("a" "b")` -- so it would only be expected to work when the first macro argument is itself a string literal, not a variable.

I suppose it also might be worth considering using some sort of limiting mechanism like `LOG_ERROR_PERIODICALLY` does, if we expect a delay would cause a flood of these messages at once in syslog.  But maybe it's hard to judge up front how necessary that might be.

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,31 @@
...
+    if (start - clock.now() >= logWarningTime) {
```
jclee, line 21: Might have meant `clock.now() - start` ?

jclee: If you wanted to write a `KJ_TEST` unit test to confirm this class works as expected, you might be able to use the same mechanism as the kj debug test to capture the log output (declare an instance of `ExceptionCallback` on the stack implementing `logMessage()`):

https://github.com/capnproto/capnproto/blob/958ccfe/c%2B%2B/src/kj/debug-test.c%2B%2B#L168-L175

...along with using `kj::TimerImpl::advanceTo()` to simulate advancing the clock without actually waiting, e.g.:

https://github.com/capnproto/capnproto/blob/958ccfe4/c%2B%2B/src/kj/compat/http-test.c%2B%2B#L3843

src/workerd/io/actor-cache.c++
Frederik-Baetens: This indeed doesn't do what I want it to do. Tried to do something similar in a separate c++ experiment, and I do indeed need to bind it to a named variable. Thanks for the feedback!

src/workerd/util/duration-alert.h
Frederik-Baetens: Thanks! Marked it as noexcept(false), but I didn't include the UnwindDetector. I think that's fine?

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#pragma once
+
+#include "kj/string.h"
+#include "kj/time.h"
```
jclee, line 8: I think the preferred style is to use quotes for relative includes (usually just files in the same directory) and angle brackets for absolute includes.

src/workerd/io/actor-cache.c++
```diff
@@ -2581,7 +2582,10 @@ kj::Promise<void> ActorCache::flushImplUsingSinglePut(PutFlush putFlush) {
   // See the comment in flushImplUsingTxn for why we need to construct our RPC and then wait on
   // reads before actually sending the write. The same exact logic applies here.
   co_await waitForPastReads();
-  co_await request.send().ignoreResult();
+  {
+    util::DurationAlert _ = util::DurationAlert(clock, kj::Duration(kj::SECONDS*1), "storage operation took longer than 1s: single put");
```
jclee, line 2586: I think typically we'd either declare the variable like
```
util::DurationAlert alert(clock, [...]);
```
or
```
auto alert = util::DurationAlert(clock, [...]);
```
just to avoid the repetition of writing the type twice.  Pretty sure the underscore variable name would be OK, but it's infrequently used.

src/workerd/util/duration-alert.h
jclee: Yeah, I think that's fine.  It seems like UnwindDetector is mostly used in places that either explicitly throw exceptions or could be reasonably anticipated to throw exceptions.

src/workerd/io/actor-cache.c++
joshthoward: For reference https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-list. I'd like to know generally whether we prefer list initialization or auto/assignment.

src/workerd/util/duration-alert.h
Frederik-Baetens: I added a unit test based on KJ_EXPECT_LOG :slightly_smiling_face: .

I wanted to test the negative case as well (no log emitted when the duration was not exceeded), but that'll require some KJ changes, and doesn't seem quite as critical to have a test for. I might add that after making a change to KJ

src/workerd/util/duration-alert-test.c++
```diff
@@ -0,0 +1,25 @@
...
+    // we don't check the actual duration emitted to avoid making the test flaky.
+    // this is okay because KJ_EXPECT_LOG just checks for substring occurences
+    {
+        DurationAlert duration(timer, 10*kj::SECONDS, "durationAlert Test Message");
+        timer.advanceTo(timer.now() + 100*kj::SECONDS);
+    }
+}
+
+}
+}
```
jasnell, line 25:
```suggestion
}  // namespace
}  // namespace workerd::util
```

src/workerd/io/actor-cache.c++
```diff
@@ -2581,7 +2582,10 @@ kj::Promise<void> ActorCache::flushImplUsingSinglePut(PutFlush putFlush) {
   // See the comment in flushImplUsingTxn for why we need to construct our RPC and then wait on
   // reads before actually sending the write. The same exact logic applies here.
   co_await waitForPastReads();
-  co_await request.send().ignoreResult();
+  {
+    util::DurationAlert alert(clock, kj::Duration(kj::SECONDS*1), "storage operation took longer than 1s: single put");
```
jasnell, line 2586: Does this actually need the `kj::Duration(...)` around the calculation. Would `1 * kj::SECONDS` also yield a `kj::Duration`?

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,40 @@
...
+// This works by relying on RAII (Scope-Bound Resource Management) to check how much time has elapsed
+// when the object is destructed. This ensures that the time is checked when the timer object goes out of scope
+// thereby timing everything after the initialization of the timer, until the end of the scope where it was
+// instantiated.
+public:
+  DurationAlert(const kj::MonotonicClock& clock, kj::Duration logWarningTime, kj::StringPtr logMessage)
+    : logWarningTime(logWarningTime),
+      logMessage(logMessage),
+      start(clock.now()),
+      clock(clock) {}
```
jasnell, line 24:
```suggestion
      clock(clock) {}
      
    KJ_DISALLOW_COPY_AND_MOVE(DurationAlert);
```

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,40 @@
...
+
+  ~DurationAlert() noexcept(false) {
+    kj::Duration duration = clock.now() - start;
+    if (duration >= logWarningTime) {
+      KJ_LOG(WARNING, "NOSENTRY", logMessage, duration);
+    }
+  }
+
+  private:
```
jasnell, line 33:
```suggestion
private:
```

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,40 @@
...
+  DurationAlert(const kj::MonotonicClock& clock, kj::Duration logWarningTime, kj::StringPtr logMessage)
+    : logWarningTime(logWarningTime),
+      logMessage(logMessage),
+      start(clock.now()),
+      clock(clock) {}
+
+  ~DurationAlert() noexcept(false) {
+    kj::Duration duration = clock.now() - start;
+    if (duration >= logWarningTime) {
+      KJ_LOG(WARNING, "NOSENTRY", logMessage, duration);
```
jasnell, line 29:
```suggestion
      KJ_LOG(WARNING, kj::str("NOSENTRY ", logMessage), duration);
```

src/workerd/util/duration-alert.h
jclee: The test looks good, thanks!

src/workerd/util/duration-alert.h
jasnell: Actually, we already have a macro for handling the `NOSENTRY` bit here, don't we?

src/workerd/io/actor-cache.c++
jclee: FWIW, in situations where either parens or braces would work, I've just stuck to using parens to match the existing kj-style code for consistency.  But I think this is not strictly enforced, and that there's some variation throughout the code base.

IIRC, there were situations for both parens and braces that were not totally covered by the other, and similarly for "almost always auto" style.  But with C++ making breaking-change improvements in recent years, maybe this has changed?

src/workerd/util/duration-alert.h
jclee: Heh... per earlier discussion, the current implementation of that macro requires a literal string as the first argument, so you'd have to do something like this anyway, to get it to work with the logMessage member variable.

https://github.com/cloudflare/workerd/pull/1693#discussion_r1495678123

src/workerd/io/actor-cache.c++
Frederik-Baetens: That does indeed also work, thanks!

src/workerd/util/duration-alert.h
Frederik-Baetens: Added this to the PR, but what's the rationale behind this?

Frederik-Baetens: I suppose having a move constructor for a raii object like this would lead to confusing behavior, as it'd be weird to track what's actually being timed. Copying probably doesn't make sense either for similar reasons, since you'd always want to re-initialize the start time rather than blindly copying.

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,42 @@
...
+#include <kj/debug.h>
+
+namespace workerd::util {
+
+// This is a utility class for instantiating a timer, which will log if it's destructed after a specified time
+// This works by relying on RAII (Scope-Bound Resource Management) to check how much time has elapsed
+// when the object is destructed. This ensures that the time is checked when the timer object goes out of scope
+// thereby timing everything after the initialization of the timer, until the end of the scope where it was
+// instantiated.
+class DurationAlert {
```
justin-mp, line 18: Optional, non-blocking, nit: consider changing the name of this to DurationLogger or LongDurationLogger since that better describes what the class does.  (Alerts mean different things.)

justin-mp (changes requested):

src/workerd/util/duration-alert.h
```diff
@@ -0,0 +1,42 @@
...
+      logMessage(logMessage),
+      start(clock.now()),
+      clock(clock) {}
+      
+    KJ_DISALLOW_COPY_AND_MOVE(DurationAlert);
+
+  ~DurationAlert() noexcept(false) {
+    kj::Duration duration = clock.now() - start;
+    if (duration >= logWarningTime) {
+      KJ_LOG(WARNING, kj::str("NOSENTRY ", logMessage), duration);
```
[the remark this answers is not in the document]
Frederik-Baetens: Thanks Justin! I've incorporated this, and also renamed it to "durationExceededLogger". I haven't gone as far as taking an operation description, since this might be useful in places where you're not specifically timing operations.

src/workerd/io/actor-cache.c++
```diff
@@ -2581,7 +2582,10 @@ kj::Promise<void> ActorCache::flushImplUsingSinglePut(PutFlush putFlush) {
   // See the comment in flushImplUsingTxn for why we need to construct our RPC and then wait on
   // reads before actually sending the write. The same exact logic applies here.
   co_await waitForPastReads();
-  co_await request.send().ignoreResult();
+  {
+    util::DurationExceededLogger logger(clock, kj::SECONDS*1, "storage operation took longer than expected: single put");
```
justin-mp, line 2586: nit: in this file, `1 * kj::SECONDS` is much easier to read than `kj::SECONDS * 1`.

justin-mp (approved):

jasnell (approved):

PR #1694, opened 2024-02-20
v8 12.2 update

mikea:


kentonv (approved):

fhanau (approved):
There's some more cleanup potential (Patch 11 can be cleaned up following a change I landed), this is non-blocking though and I can do it in a later cleanup PR. Should be ready to go from my side.

jasnell (approved):

PR #1696, opened 2024-02-20
Add TransformStream cancel algorithm

jasnell:
The cancel algorithm for TransformStreams was recently added to the spec. This implements it. See the new test for example use.

Refs: https://streams.spec.whatwg.org/#callbackdef-transformercancelcallback

### DIFF
--- a/src/workerd/api/streams/standard.c++
+++ b/src/workerd/api/streams/standard.c++
@@ -2044,6 +2044,13 @@ kj::Own<WeakRef<ReadableStreamDefaultController>> ReadableStreamDefaultControlle
   return kj::addRef(*weakRef);
 }
 
+kj::Maybe<StreamStates::Errored> ReadableStreamDefaultController::getMaybeErrorState(
+    jsg::Lock& js) {
+  KJ_IF_SOME(errored, impl.state.tryGet<StreamStates::Errored>()) {
+    return errored.addRef(js);
+  }
+  return kj::none;
+}
 
 void ReadableStreamDefaultController::start(jsg::Lock& js) {
   impl.start(js);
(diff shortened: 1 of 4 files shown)


jasnell:
There's a relevant ASAN failure in internal CI in the added test to investigate before this can land.

src/workerd/api/streams/standard.c++
```diff
@@ -3971,8 +3978,32 @@ jsg::Promise<void> TransformStreamDefaultController::write(
...
     v8::Local<v8::Value> reason) {
-  error(js, reason);
-  return js.resolvedPromise();
+  KJ_IF_SOME(finish, algorithms.maybeFinish) {
+    return finish.whenResolved(js);
+  }
+  auto& promise = algorithms.maybeFinish.emplace(maybeRunAlgorithm(js, algorithms.cancel,
+      JSG_VISITABLE_LAMBDA(
+          (this, ref=JSG_THIS, reason = jsg::JsRef(js, jsg::JsValue(reason))),
+          (ref, reason),
```
fhanau, line 3987: Just noticed that you moved this back to draft state, but LGTM overall. One thing I didn't understand were the JSG_VISITABLE_LAMBDA arguments here, even after looking at existing usage examples (why does the closure include two arguments here?)

jasnell: The `JSG_VISITABLE_LAMBA`  is a bit weird. The first set `(this, ref=JSG_THIS, reason = jsg::JsRef(...))` defines the captures for the lamba. The second set `(ref, reason)` defines the bits that are gc visited, and the third is the actual body of the lambda. The point is to ensure that `ref` and `reason` (in this case) are visible to the v8 gc visitation mechanism when the promise they are attached to here is gc visited. It's not the easiest thing to get your head around but it works.

fhanau (approved):

PR #1697, opened 2024-02-20
disable webgpu until build issues are resolved

mikea:


dom96 (approved):
Not sure about the context here, but accepting to help unblock if you're stuck on this

kentonv:
I think we decided that we can't just disable webgpu but #1701 fixes the issue. Closing to avoid confusion, but feel free to reopen if I got it wrong.

PR #1700, opened 2024-02-20
Revert "Add toStringTag value on JSG resource types"

mikea:
Reverts cloudflare/workerd#1607

Breaks downstream trace test

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -47,7 +47,6 @@ namespace workerd::jsg {
       ::workerd::jsg::JsgKind::RESOURCE; \
   using jsgSuper = jsgThis; \
   using jsgThis = Type; \
-  static inline kj::StringPtr jsgGetName() { return #Type##_kjc; } \
   inline kj::StringPtr jsgGetMemoryName() const override { return #Type##_kjc; } \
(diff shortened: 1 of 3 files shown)


garrettgu10 (approved):

jasnell (approved):

jasnell:
@mikea ... how was the internal repo test broken? Does the test assume a particular output when an object to stringified? e.g. `[object Object]` vs. `[object Foo]`? Depending on exactly how that test was broken this change might end up having to go behind a compat flag.

mikea:
> @mikea ... how was the internal repo test broken? Does the test assume a particular output when an object to stringified? e.g. `[object Object]` vs. `[object Foo]`? Depending on exactly how that test was broken this change might end up having to go behind a compat flag.

hard for me to say what's wrong: there's so much output in the test. All I know is that `Can log JSON edge cases` subtest fails with this commit and passes without.

garrettgu10:
@jasnell It looks like the test does assume a particular output when an object is stringified:
```
Expected: ["[object Object]"], actual: ["[object Headers]"]
```

kentonv:
It sounds like the test just needs to be updated, but it's important to merge those downstream updates at the same time as the workerd changes.

PR #1702, opened 2024-02-20
Remove log line about addEventListener in modules scripts.

kentonv:
It logs frequently in production. We're stuck supporting this. Sadness.

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -1732,16 +1732,11 @@ kj::Maybe<kj::Own<api::ExportedHandler>> Worker::Lock::getExportedHandler(
   } else if (name == kj::none) {
     // If the default export was requested, and we didn't find a handler for it, we'll fall back
     // to addEventListener().
-    // TODO(cleanup): The intention has always been that we only use addEventListener() for
+    //
+    // Note: The original intention was that we only use addEventListener() for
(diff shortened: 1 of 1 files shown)


src/workerd/io/worker.c++
```diff
@@ -1732,16 +1732,11 @@ kj::Maybe<kj::Own<api::ExportedHandler>> Worker::Lock::getExportedHandler(
...
-    //   modules-based script too, if they lacked an `export default`. Yikes! However, it looks
-    //   to me like the validator would have rejected modules-based scripts that lacked a default
-    //   export, so perhaps this is not a real problem in production. We'd better find out by
-    //   logging, though...
-    if (worker.impl->context == kj::none) {
-      LOG_ERROR_PERIODICALLY(
-          "modules-based script has no default export; falling back to addEventListener()");
-    }
+    //   modules-based script too, if they lacked an `export default`. Yikes! Sadly, there are
+    //   Workers in production relying on this so we are stuck with it.
```
jasnell, line 1739: non-blocking for this pr .. but just noting... for things like this, I almost wonder if it would be worthwhile to implement a metrics counter so we can avoid spamming logs but still have some sense of how often this happens. If the count ever drops persistently to zero then maybe we could fix this in the future.

jasnell (approved):

jp4a50 (approved):

PR #1703, opened 2024-02-21
Cleanup standard streams definitions

jasnell:
Following the memleak fixup, I wanted to go through an audit the streams code for any remaining possible issues. This is the first pass. Much of this is cleanup along the way but there's at least one bug fix in here addressing a UAF. This is ready for review but I'll be doing some extensive testing on it before landing.


src/workerd/api/streams/standard.c++
```diff
@@ -1074,6 +1075,13 @@ void ReadableImpl<Self>::doError(jsg::Lock& js, jsg::Value reason) {
       return;
     }
     KJ_CASE_ONEOF(queue, Queue) {
+      // When we call queue.error, it will cause all of the consumers attached to
+      // be errored as well. This may cause the consumer to drop it's strong reference
+      // to the controller, which would cause this impl to be dropped. Let's make we we
+      // hold a strong reference at least until we're done with the call to doError here
+      // so that if the controller is to be freed, it will be after we're done with it.
+      auto tempRef = self.addRef();
```
jasnell, line 1083: This is a bug fix. In certain causes the call to `queue.error(...)` can (in a somewhat roundabout way) cause the `ReadableStreamDefaultController` or `ReadableByteStreamController` that own this `ReadableImpl` to be freed if there are no other `jsg::Ref`s holding it. We want to make sure that it stays alive at least through the end of this method so that the subsequent calls to update the state and clear the algorithms are cleared. We grab the temporary strong ref here to ensure that the owning controller is kept alive at least through the end of this scope.

src/workerd/api/streams/standard.c++
```diff
@@ -1808,10 +1819,13 @@ struct ValueReadable final: public api::ValueQueue::ConsumerImpl::StateListener
     // Here, we rely on the controller implementing the correct behavior since it owns
     // the queue that knows about all of the attached consumers.
     KJ_IF_SOME(s, state) {
-      KJ_DEFER({
-        state = kj::none;
-      });
-      return s.cancel(js, kj::mv(maybeReason));
+      auto promise = s.cancel(js, kj::mv(maybeReason));
```
jasnell, line 1822: These defers just visually made this code more difficult to reason about.

fhanau (approved):
LGTM. `controller` was added to the GC visitor, was it just missing there previously?

jasnell:
Few internal CI failures to investigate before this progresses

PR #1704, opened 2024-02-21
MQ-498: Add support for delivery delay in Queues

sesteves:
This patch updates the queues api so that a delay value can be specified when sending messages to a queue. The delay is sent as an header in the request and also as the message payload in case of a batch.

### DIFF
--- a/src/workerd/api/queue.c++
+++ b/src/workerd/api/queue.c++
@@ -15,6 +15,13 @@
 namespace workerd::api {
 
 namespace {
+
+// Header for the message format.
+static constexpr kj::StringPtr HDR_MSG_FORMAT = "X-Msg-Fmt"_kj;
+
+// Header for the message delivery delay.
+static constexpr kj::StringPtr HDR_MSG_DELAY = "X-Msg-Delay-Secs"_kj;
+
 kj::StringPtr validateContentType(kj::StringPtr contentType) {
   auto lowerCase = toLower(contentType);
   if (lowerCase == IncomingQueueMessage::ContentType::TEXT) {
@@ -127,9 +134,10 @@ Serialized serialize(jsg::Lock& js,
   }
 }
 
-struct SerializedWithContentType {
+struct SerializedWithOptions {
   Serialized body;
   kj::Maybe<kj::StringPtr> contentType;
+  kj::Maybe<int> delaySeconds;
 };
 
 jsg::JsValue deserialize(jsg::Lock& js,
@@ -179,19 +187,23 @@ kj::Promise<void> WorkerQueue::send(jsg::Lock& js,
 
   JSG_REQUIRE(!body.isUndefined(), TypeError, "Message body cannot be undefined");
 
-  kj::Maybe<kj::StringPtr> contentType;
-  KJ_IF_SOME(opts, options) {
-    KJ_IF_SOME(type, opts.contentType) {
-      contentType = validateContentType(type);
-    }
-  }
-
   auto headers = kj::HttpHeaders(context.getHeaderTable());
   headers.set(kj::HttpHeaderId::CONTENT_TYPE, MimeType::OCTET_STREAM.toString());
 
+  kj::Maybe<kj::StringPtr> contentType;
+  KJ_IF_SOME(opts, options) {
+    KJ_IF_SOME(type, opts.contentType) {
+      auto validatedType = validateContentType(type);
+      headers.add(HDR_MSG_FORMAT, validatedType);
+      contentType = validatedType;
+    }
+    KJ_IF_SOME(secs, opts.delaySeconds) {
+      headers.add(HDR_MSG_DELAY, kj::str(secs));
+    }
+  }
+
   Serialized serialized;
   KJ_IF_SOME(type, contentType) {
-    headers.add("X-Msg-Fmt", type);
     serialized = serialize(js, body, type, SerializeArrayBufferBehavior::DEEP_COPY);
   } else {
     // TODO(cleanup) send message format header (v8) by default
@@ -223,7 +235,8 @@ kj::Promise<void> WorkerQueue::send(jsg::Lock& js,
       .attach(context.registerPendingEvent());
 };
 
-kj::Promise<void> WorkerQueue::sendBatch(jsg::Lock& js, jsg::Sequence<MessageSendRequest> batch) {
+kj::Promise<void> WorkerQueue::sendBatch(jsg::Lock& js, jsg::Sequence<MessageSendRequest> batch,
+                                         jsg::Optional<SendBatchOptions> options) {
   auto& context = IoContext::current();
 
   JSG_REQUIRE(batch.size() > 0, TypeError, "sendBatch() requires at least one message");
@@ -231,13 +244,17 @@ kj::Promise<void> WorkerQueue::sendBatch(jsg::Lock& js, jsg::Sequence<MessageSen
   size_t totalSize = 0;
   size_t largestMessage = 0;
   auto messageCount = batch.size();
-  auto builder = kj::heapArrayBuilder<SerializedWithContentType>(messageCount);
+  auto builder = kj::heapArrayBuilder<SerializedWithOptions>(messageCount);
   for (auto& message: batch) {
     auto body = message.body.getHandle(js);
     JSG_REQUIRE(!body.isUndefined(), TypeError,
                 "Message body cannot be undefined");
 
-    SerializedWithContentType item;
+    SerializedWithOptions item;
+    KJ_IF_SOME(secs, message.delaySeconds) {
+      item.delaySeconds = secs;
+    }
+
     KJ_IF_SOME(contentType, message.contentType) {
       item.contentType = validateContentType(contentType);
       item.body = serialize(js, body, contentType,
@@ -255,7 +272,7 @@ kj::Promise<void> WorkerQueue::sendBatch(jsg::Lock& js, jsg::Sequence<MessageSen
   // Construct the request body by concatenating the messages together into a JSON message.
   // Done manually to minimize copies, although it'd be nice to make this safer.
   // (totalSize + 2) / 3 * 4 is equivalent to ceil(totalSize / 3) * 4 for base64 encoding overhead.
-  auto estimatedSize = (totalSize + 2) / 3 * 4 + messageCount * 32 + 32;
+  auto estimatedSize = (totalSize + 2) / 3 * 4 + messageCount * 64 + 32;
   kj::Vector<char> bodyBuilder(estimatedSize);
   bodyBuilder.addAll("{\"messages\":["_kj);
   for (size_t i = 0; i < messageCount; ++i) {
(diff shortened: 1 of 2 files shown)


sesteves:

sesteves:
recheck

jbwcloudflare:
@a-robinson and I will take a look

src/workerd/api/queue.h
```diff
@@ -32,30 +32,48 @@ class WorkerQueue: public jsg::Object {
...
     jsg::Optional<kj::String> contentType;
 
-    JSG_STRUCT(contentType);
+    // The number of seconds to delay the delivery of the message being sent.
+    jsg::Optional<uint> delaySecs;
+
+    JSG_STRUCT(contentType, delaySecs);
     JSG_STRUCT_TS_OVERRIDE(QueueSendOptions {
       contentType?: QueueContentType;
+      delaySecs ?: number;
```
a-robinson, line 41: I don't imagine the override is needed here given that we aren't change the type at all. `contentType` only needed an override due to wanting to advertise that it only accepts the `QueueContentType` type

But you should be able to confirm by building the types locally using the instructions in https://github.com/cloudflare/workerd/tree/main/types

src/workerd/api/queue.h
```diff
@@ -32,30 +32,48 @@ class WorkerQueue: public jsg::Object {
...
 
-    JSG_STRUCT(body, contentType);
+    // The number of seconds to delay the delivery of the message being sent.
+    jsg::Optional<uint> delaySecs;
+
+    JSG_STRUCT(body, contentType, delaySecs);
     JSG_STRUCT_TS_OVERRIDE(MessageSendRequest<Body = unknown> {
       body: Body;
       contentType?: QueueContentType;
+      delaySecs ?: number;
```
a-robinson, line 68: Same comments as in the above `JSG_STRUCT_TS_OVERRIDE`

src/workerd/api/queue.h
a-robinson, line 41: And why the the extra space between `delaySecs` and the question mark?

src/workerd/api/queue.h
```diff
@@ -64,7 +82,9 @@ class WorkerQueue: public jsg::Object {
     JSG_TS_ROOT();
     JSG_TS_OVERRIDE(Queue<Body = unknown> {
       send(message: Body, options?: QueueSendOptions): Promise<void>;
-      sendBatch(messages: Iterable<MessageSendRequest<Body>>): Promise<void>;
+      sendBatch(messages
+                : Iterable<MessageSendRequest<Body>>, options ?: QueueSendBatchOptions)
```
a-robinson, line 86: What's going on with the whitespace in here? Have you managed to get `make format` working?

src/workerd/api/queue.c++
```diff
@@ -179,21 +190,30 @@ kj::Promise<void> WorkerQueue::send(jsg::Lock& js,
...
   JSG_REQUIRE(!body.isUndefined(), TypeError, "Message body cannot be undefined");
 
+  auto headers = kj::HttpHeaders(context.getHeaderTable());
+  headers.set(kj::HttpHeaderId::CONTENT_TYPE, MimeType::OCTET_STREAM.toString());
+
   kj::Maybe<kj::StringPtr> contentType;
   KJ_IF_SOME(opts, options) {
     KJ_IF_SOME(type, opts.contentType) {
       contentType = validateContentType(type);
+      headers.add(HDR_MSG_FORMAT, type);
```
a-robinson, line 200: We should continue using `contentType` for this (as before) so that we always send the standardized lower-case version rather than whatever the user gave us.

In this case you may need to use `KJ_ASSERT_NONNULL(contentType)` in order to make it compile since the `contentType` variable is a `kj::Maybe`. Or just do something like:

```
auto validatedType = validateContentType(type);
contentType = validatedContentType;
headers.add(HDR_MSG_FORMAT, validatedContentType);
```

Either is fine.

src/workerd/api/queue.c++
```diff
@@ -179,21 +190,30 @@ kj::Promise<void> WorkerQueue::send(jsg::Lock& js,
...
-  auto headers = kj::HttpHeaders(context.getHeaderTable());
-  headers.set(kj::HttpHeaderId::CONTENT_TYPE, MimeType::OCTET_STREAM.toString());
-
   Serialized serialized;
   KJ_IF_SOME(type, contentType) {
-    headers.add("X-Msg-Fmt", type);
     serialized = serialize(js, body, type, SerializeArrayBufferBehavior::DEEP_COPY);
-  } else {
+  }
+  else {
```
a-robinson, line 216: We always keep `else` on the same line as the prior closing brace, so I'd suggest getting your auto-formatter (`make format`) working properly :)

src/workerd/api/queue.c++
```diff
@@ -272,6 +302,11 @@ kj::Promise<void> WorkerQueue::sendBatch(jsg::Lock& js, jsg::Sequence<MessageSen
       bodyBuilder.add('"');
     }
 
+    KJ_IF_SOME(delaySecs, serializedBodies[i].delaySecs) {
+      bodyBuilder.addAll(",\"delaySecs\": "_kj);
```
a-robinson, line 306: If we're adding additional fields into the JSON, do we need to allocate a larger buffer up above? (see the `estimatedSize` calculation)

src/workerd/api/queue.h
```diff
@@ -32,30 +32,48 @@ class WorkerQueue: public jsg::Object {
...
+
   struct MessageSendRequest {
     jsg::JsRef<jsg::JsValue> body;
 
     // contentType determines the serialization format of the message.
     jsg::Optional<kj::String> contentType;
 
-    JSG_STRUCT(body, contentType);
+    // The number of seconds to delay the delivery of the message being sent.
+    jsg::Optional<uint> delaySecs;
```
a-robinson, line 62: Do we even want to support setting the delay on each message in a batch individually vs just on an entire batch? I thought the discussion on the PRD was that we were only going to support it on the entire batch. I'll tag you in the thread.

src/workerd/api/queue.c++
```diff
@@ -294,6 +329,12 @@ kj::Promise<void> WorkerQueue::sendBatch(jsg::Lock& js, jsg::Sequence<MessageSen
   headers.add("CF-Queue-Largest-Msg"_kj, kj::str(largestMessage));
   headers.set(kj::HttpHeaderId::CONTENT_TYPE, MimeType::JSON.toString());
 
+  KJ_IF_SOME(opts, options) {
+    KJ_IF_SOME(secs, opts.delaySecs) {
```
a-robinson, line 333: 1. Are we ever validating that this is beneath the max delay?
2. Pending the result of the question around whether we want to support this option both on individual messages within a batch and within a batch, what happens if the user specifies a delay for the entire batch *and* for messages within the batch? Is that meant to be allowed?

src/workerd/api/queue.h
```diff
@@ -32,30 +32,48 @@ class WorkerQueue: public jsg::Object {
     // contentType determines the serialization format of the message.
     jsg::Optional<kj::String> contentType;
 
-    JSG_STRUCT(contentType);
+    // The number of seconds to delay the delivery of the message being sent.
+    jsg::Optional<uint> delaySecs;
```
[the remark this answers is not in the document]
jasnell: This is fine. I would probably just cut it down to `delay` and document it as being seconds. Or, avoid the abbreviation and make it `delaySeconds`

src/workerd/api/queue.h
```diff
@@ -285,15 +305,9 @@ class QueueCustomEventImpl final: public WorkerInterface::CustomEvent, public kj
...
-  api::IncomingQueueMessage,            \
-  api::QueueResponse,                   \
-  api::QueueMessage,                    \
-  api::QueueEvent,                      \
-  api::QueueController,                 \
-  api::QueueExportedHandler
+#define EW_QUEUE_ISOLATE_TYPES                                                                     \
+  api::WorkerQueue, api::WorkerQueue::SendOptions, api::WorkerQueue::SendBatchOptions,             \
+      api::WorkerQueue::MessageSendRequest, api::IncomingQueueMessage, api::QueueResponse,         \
+      api::QueueMessage, api::QueueEvent, api::QueueController, api::QueueExportedHandler
```
jasnell, line 311: Nit: generally prefer to keep this in the original one-item-per-line format as it makes reviewing additions/subtractions easier.

src/workerd/api/queue.h
sesteves: I even tried `clang-format` with the style file (.clang-format) used in `edgeworker`, and it always adds me that space there. I removed it anyway to match the original style.

src/workerd/api/queue.h
sesteves: There is no `make format` in this project. In the `edgeworker` repo, `make format` uses `clang-format` with a custom code style; when I apply it to this file, I get the same result (same line breaking and space before `?`.

src/workerd/api/queue.c++
a-robinson: Huh, that's weird. I don't know much about our clang-format setup since it arose after I did most of my runtime work, but here's what I know:

* `make format` in edgeworker doesn't appear to touch files in workerd
* When I change a line in edgeworker to move the `else` to the following line like this, running `make format` moves it back to the same line
* The style guide we follow is [pretty clear about the intended format here](https://github.com/capnproto/capnproto/blob/v2/style-guide.md#spacing-and-bracing) and all the code I can recall reading in the runtime code follows that style

I don't know much about our clang-format config for workerd, though. Maybe @jasnell has a pointer here, but no worries if you don't.

src/workerd/api/queue.h
a-robinson: Looks like the issue is that these `JSG_*` blocks are explicitly excluded from clang-format: https://github.com/cloudflare/workerd/blob/main/.clang-format

So it's just not changing anything regardless of what you put here.

src/workerd/api/queue.c++
a-robinson: FWIW I'm fine with just doing the validation in the queue service rather than here. Feel free to rip it out of here. It just didn't look right to do some validation here but to miss some cases.

src/workerd/api/queue.h
a-robinson: Did you manage to get this building to see how the resulting types look? My machine unfortunately complains about a dependency issue when I try, otherwise I'd do it for you.

I'm fine skipping it for now and just validating them after merge if needed, so long as you do come back and make sure the published types look ok afterwards.

a-robinson (approved):

jbwcloudflare:
nice job, LGTM

src/workerd/api/queue.c++
```diff
@@ -179,19 +187,23 @@ kj::Promise<void> WorkerQueue::send(jsg::Lock& js,
...
   kj::Maybe<kj::StringPtr> contentType;
   KJ_IF_SOME(opts, options) {
     KJ_IF_SOME(type, opts.contentType) {
-      contentType = validateContentType(type);
+      auto validatedType = validateContentType(type);
+      headers.add(HDR_MSG_FORMAT, validatedType);
+      contentType = validatedType;
+    }
+    KJ_IF_SOME(secs, opts.delaySeconds) {
+      headers.add(HDR_MSG_DELAY, kj::str(secs));
```
jasnell, line 201: Since `delaySeconds` is defined as an `int`, what happens in a negative value is provided?

jasnell (approved):

src/workerd/api/queue.c++
sesteves: We agreed on doing that validation on the server side. I had it as a `uint` before and the error that was thrown was not super descriptive and user-friendly, that's why I changed it to an `int`.

PR #1706, opened 2024-02-21
Add compat flag to enable AsyncLocalStorage separately

jasnell:


irvinebroque (commented):
Will let someone else review code but support doing this to unblock use cases we have been discussing

src/workerd/jsg/modules.h
```diff
@@ -437,70 +437,70 @@ class ModuleRegistryImpl final: public ModuleRegistry {
     entries.insert(kj::heap<Entry>(specifier, Type::BUNDLE, kj::fwd<ModuleInfo>(info)));
   }
 
+  void addBuiltinModule(Module::Reader module) {
```
kentonv, line 440: What's this refactor? Did anything change in this code?

kentonv (approved):

src/workerd/jsg/modules.h
jasnell: Nope nothing changed, separated it out so that we could call this directly for the one `Module` in the bundle that is the `node:async_hooks`

PR #1707, opened 2024-02-21
Pin workerd local dev python package version numbers to match deployment

garrettgu10:
During local development, we would like micropip to fetch the same version numbers as what's currently deployed on edgeworker. This PR hardcodes some version numbers used by edgeworker according to (https://github.com/dom96/pyodide_packages/releases) and makes sure that they are passed to micropip when installing the packages.


jasnell (approved):

hoodmane (approved):
Thanks @garrettgu10 this looks good to me for now. Eventually it would be nice to extract this from `pyodide_packages.tar` or something like that though since these will take some effort to update.

hoodmane:
> I couldn't find the version numbers for async_timeout or setuptools.

As far as I can tell we aren't including them anymore and that was just out of date.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -54,49 +54,63 @@ function initializePackageIndex(pyodide) {
...
-]);
+// If a version number is specified (i.e. non-empty), it is passed to micropip
+// when running in local dev.
+const EMBEDDED_PYTHON_PACKAGES = {
+  "aiohttp": "3.9.3",
+  "aiosignal": "1.3.1",
+  "anyio": "3.7.1",
+  "attrs": "23.2.0",
+  "certifi": "2024.2.2",
+  "charset_normalizer": "3.3.2",
```
garrettgu10, line 65: @dom96 @hoodmane 

I kept the underscores since that's what the original code said, but I think it might be necessary to [normalize package names](https://packaging.python.org/en/latest/specifications/name-normalization/) in this list as well as during lookup. But that might be better to put in another PR.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -54,49 +54,65 @@ function initializePackageIndex(pyodide) {
...
+  "tenacity": "8.2.3",
+  "tqdm": "4.66.2",
+  "typing_extensions": "4.9.0",
+  "typing_inspect": "0.9.0",
+  "urllib3": "2.2.0",
+  "yarl": "1.9.4"
+}
+
+function addPinnedVersionToRequirement(requirement) {
+  if (requirement.includes("==")) {
```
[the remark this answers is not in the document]
hoodmane: I think it would be better to check for presence of the `requirement` against the lock file. That would also catch package names that we don't have / have a typo.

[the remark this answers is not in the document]
garrettgu10: @hoodmane that would come after we disable micropip on edgeworker, right?

src/pyodide/python-entrypoint-helper.js
hoodmane: We probably want to have pretty similar logic to `recursiveDependencies` here:
https://github.com/pyodide/pyodide/blob/main/src/js/load-package.ts?plain=1#L186

In fact there's a pretty good chance we could use that same logic unchanged.

dom96: Good point, yeah, normalization is a must.

PR #1709, opened 2024-02-22
Update Bazel to 7.1.1, slight cleanup and compatibility improvements

fhanau:
This should improve performance and stability, 7.1 includes a number of performance and correctness improvements for remote caching, sandboxed execution and fetching dependencies.
- Re. the incompatible flags: This is based on bazel's [compatibility policy](https://bazel.build/release/backward-compatibility), I'm mostly flipping these early to avoid headaches when updating bazel in the future, but we also avoid some footguns with globs. Context is available at https://github.com/bazelbuild/bazel/issues/8195 and https://github.com/bazelbuild/bazel/issues/12821 but not required for reviewing this.
- I previously didn't find any quantifiable improvements from `--legacy_external_runfiles=false` but `bazel aquery  --include_file_write_contents=true "deps(//src/workerd/io/...)"` shrinks by ~3MB (10%) with the flag so there's certainly less work that bazel needs to do.

Commit message:
> 7.1 (https://github.com/bazelbuild/bazel/releases/tag/7.1.0) introduces
a number of bug fixes and performance improvements, 7.1.1 has some
additional bug fixes.
>
>- Pass all parameters in kj_test to ensure tags will actually be
  applied, avoiding a considerable footgun.
>- Clean up a few globs, allowing us to enable incompatible_disallow_empty_glob
>- Flip --legacy_external_runfiles early to avoid generating superfluous
      runfile trees.
>- Update wd_ts_bundle to fix --legacy_external_runfiles=false for
      dependents.


src/cloudflare/BUILD.bazel
```diff
@@ -9,14 +9,15 @@ wd_ts_bundle(
             "internal/*.ts",
             "internal/*.js",
         ],
-        exclude = [
-            "internal/test/*",
```
hoodmane, line 13: Is there a reason you don't want to exclude this anymore?

hoodmane (approved):

src/cloudflare/BUILD.bazel
fhanau: Excluding this is not needed – `internal/*.{js,ts}` covers files in the internal directory, but not the subdirectories.

PR #1712, opened 2024-02-22
Bump supported compatibility date.

jp4a50:
For weekly release.


fhanau (approved):

fhanau:
Approved, but it looks like we need a review from Wrangler folks @mrbbot @penalosa? We really should update those ownership settings to have co-ownership.

mrbbot (approved):

PR #1715, opened 2024-02-22
Separate JsObject and JsMap

jasnell:


jasnell:
there are existing tests

kentonv (approved):
I audited all call sites of get(), set(), delete_(), and has(). It looks like there's no case that could have been accidentally operating on a Map when not intending to, nor vice versa, and so there's no risk of breaking anything with this change.

Specifically, the call sites I found are:
* Setting `ctx` and `env` on RPC objects in their constructor, can't possibly be a Map.
* Manipulating the request.cf blob immediately after parsing it from JSON; can't possibly be a Map.
* Reading `.prototype` from a `Function`, can't possibly be a Map.
* Explicitly constructing a Map as the result of DO `storage.list()` and `storage.getMultiple()`. In this case it's _always_ a Map (returned by `JsMap`) so should be fine with the new code.
* parseListMetadata() in `kv.c++`, operates on the result of parsing JSON, can't be a Map.
* `R2Admin::list()`, operating on a map just constructed with `js.map()`, always a Map.
* `R2Error::getStack()`, this appears to be dead (unreachable) code, and the handle it operates on is never set to anything.

PR #1717, opened 2024-02-23
Clean up V8 patches after 12.2 update

fhanau:
https://chromium-review.googlesource.com/c/v8/v8/+/5201460 and https://chromium-review.googlesource.com/c/v8/v8/+/5173468 landed with V8 12.2, allowing us to do some cleanup.


hoodmane (approved):
The `absl` changes are mysterious to me but everything else looks good.

fhanau:
Thank you for approving!
The abseil changes are based on dropping the patch: V8 marks `absl_optional` as a dependency despite not actually requiring it, so far we used a patch to remove that dependency. Now that other abseil components are being required and the other half of the patch became redundant, I figured that dropping the patch and adding another binding in WORKSPACE would be cleaner. I also sorted the absl bindings alphabetically.
I'll wait a but before merging this in case anything else V8-related comes up.

WORKSPACE
```diff
@@ -181,8 +181,8 @@ git_repository(
 )
 
 bind(
```
mikea, line 183: I don't remember, can we do list comprehension in WORKSPACE? Lets dedup if we can?

mikea (approved):

WORKSPACE
fhanau: Turns out we can. Updated this, at least the three `absl/container:*` dependencies are deduplicated now.

PR #1723, opened 2024-02-24
Implement v8::Exception::CaptureStackTrace for DOMException

jasnell:
To support adding the stack property to DOMException, we used to create a dummy internal Error object when the the DOMException was created. This is wasteful and requires an additional object allocation. Instead, let's tap into v8's built-in stack capture mechanism and and add the stack directly to the DOMException wrapper object.

### DIFF
--- a/src/workerd/jsg/dom-exception.h
+++ b/src/workerd/jsg/dom-exception.h
@@ -49,24 +49,19 @@ namespace workerd::jsg {
 // "DOMException".
 class DOMException: public Object {
 public:
-  DOMException(kj::String message,
-               kj::String name,
-               V8Ref<v8::Object> errorForStack)
+  DOMException(kj::String message, kj::String name)
       : message(kj::mv(message)),
-        name(kj::mv(name)),
-        errorForStack(kj::mv(errorForStack)) {}
+        name(kj::mv(name)) {}
 
   // JS API
 
-  static Ref<DOMException> constructor(Lock& js,
+  static Ref<DOMException> constructor(const v8::FunctionCallbackInfo<v8::Value>& args,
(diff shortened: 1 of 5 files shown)


src/workerd/jsg/dom-exception.h
```diff
@@ -109,17 +93,12 @@ class DOMException: public Object {
...
     tracker.trackField("message", message);
     tracker.trackField("name", name);
-    tracker.trackField("errorForStack", errorForStack);
   }
 
 private:
   kj::String message;
   kj::String name;
 
-  // We implement the `stack` property in a similarly hacky way as Chrome: store an Error object
```
kentonv, line 119: If this is good enough for Chrome, why isn't it good enough for us? I'm nervous about adding new patches to V8, don't think it's really worth it just to avoid an allocation on the error path. Though if V8 would accept upstreaming the patch, then maybe it makes sense?

jasnell: I've already got an open thread in the v8-dev group to ask if they would accept this change. Given how convoluted chrome's approach to handling this is, even compared to ours, I would have to imagine that it would be of benefit there also.

jasnell:
https://chromium-review.googlesource.com/c/v8/v8/+/5444318

jasnell:
Happy days.. the upstream patch has been approved. Should be landing upstream soonish. Once we've received the updated version of v8 with the API change I'll be able to update this PR to drop the patch and move forward.

kentonv:
I'd be fine with floating a cherry-pick in the meantime.

jasnell:
Superseded by https://github.com/cloudflare/workerd/pull/2139

PR #1724, opened 2024-02-24
V8 patch for v8::Date::ToUTCString and v8::Date::Parse

jasnell:


kentonv:
I'm not a big fan of this, especially the part where the implementation is copied. We won't get bugfixes from the original. Anyway, we should try to keep a light touch with V8 patches if at all possible.

I don't think it's so bad to read the properties off the JS objects, can we just leave it that way?

patches/v8/0017-Add-v8-Date-ToUTCString-and-v8-Date-Parse-APIs.patch
```diff
@@ -0,0 +1,124 @@
...
++#include "src/date/dateparser.h"
+ #include "src/debug/debug.h"
+ #include "src/deoptimizer/deoptimizer.h"
+ #include "src/execution/embedder-state.h"
+@@ -7917,6 +7918,57 @@ MaybeLocal<v8::Value> v8::Date::New(Local<Context> context, double time) {
+   RETURN_ESCAPED(result);
+ }
+ 
++namespace {
++// This is copied from builtins-date.cc... a cleaner solution would be to make
```
kentonv, line 51: I don't think we should be copying this. This code could easily change over time, either to fix bugs or because the internal interfaces it calls have changed. This doesn't seem worth it to maintain this just to avoid having to call the method through JS.

jasnell:
I will be contributing these changes directly to v8. Once that lands there, we can either float that version of the patch until we pick it up in the regular v8 update or keep this PR pending until the update.v

https://groups.google.com/g/v8-dev/c/Ezua_w2NH_M

kentonv:
> I will be contributing these changes directly to v8.

That sounds great!

jasnell:
https://chromium-review.googlesource.com/c/v8/v8/+/5444005

jasnell:
Happy days... the v8 change has been accepted. Will just be waiting on us to receive the v8 update that includes the new APIs and this PR can be updated to drop the patch and move forward

jasnell:
Looks like it'll be v8 12.6 before the upstream change will land

jasnell:
Superseded by https://github.com/cloudflare/workerd/pull/2137

PR #1726, opened 2024-02-25
Just some minor cleanups/clarifications on the JSG readme

jasnell:


src/workerd/jsg/README.md
```diff
@@ -396,13 +421,13 @@ There are generally three ways of using `jsg::Function`:
...
 
-We have a C++ function that takes a callback argument. The callback argument takes a single `int`
+We have a C++ function that takes a callback argument. The callback argument takes an `int`
 as an argument (Ignore the `jsg::Lock& js` part for now, we'll explain that in a bit). As soon
 as we get the callback, we invoke it as a function.
 
 ```cpp
 void doSomething(jsg::Lock& js, jsg::Function<void(int)> callback) {
-  callback(1);
+  callback(js, 1);
```
justin-mp, line 430: Should the declaration of `callback` be `jsg::Function<void(jsg::Lock&,int)> callback`?

justin-mp (approved):

src/workerd/jsg/README.md
jasnell: No, the `jsg::Lock&` argument is not included in the declaration bit, just required when it is actually called.

src/workerd/jsg/README.md
```diff
@@ -427,12 +452,6 @@ const func = getFunction();
 func(1);  // prints 1
 ```
 
-For the third case, because it is possible to create a `jsg::Function` from either a JS function
```
fhanau, line 430: If we're removing this, we should also update the section above where it says `generally three ways of using jsg::Function` right?

jasnell: Maybe? not sure. I want to go back and take another pass through this section.

PR #1728, opened 2024-02-25
Pass CppHeap using CreateParams

jasnell:
Using Attach/Detach to pass has been deprecated.

### DIFF
--- a/src/workerd/jsg/setup.c++
+++ b/src/workerd/jsg/setup.c++
@@ -281,8 +281,33 @@ bool HeapTracer::TryResetRoot(const v8::TracedReference<v8::Value>& handle) {
 }
 
 namespace {
-  static v8::Isolate* newIsolate(v8::Isolate::CreateParams&& params) {
+  static v8::Isolate* newIsolate(
+      V8PlatformWrapper* system,
+      v8::Isolate::CreateParams&& params) {
     return jsg::runInV8Stack([&](jsg::V8StackScope& stackScope) -> v8::Isolate* {
+      v8::CppHeapCreateParams heapParams {
+        {},
+        v8::WrapperDescriptor(
+            Wrappable::WRAPPABLE_TAG_FIELD_INDEX,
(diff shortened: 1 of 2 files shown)


src/workerd/jsg/setup.c++
```diff
@@ -281,8 +281,33 @@ bool HeapTracer::TryResetRoot(const v8::TracedReference<v8::Value>& handle) {
...
+      // this would be a win anyway, since Worker heaps are relatively small and therefore doing a
+      // full atomic mark-sweep usually doesn't require much of a pause.
+      //
+      // We probably won't ever support concurrent marking or sweeping because concurrent GC is
+      // only expected to be a win if there are idle CPU cores available. Workers normally run on
+      // servers that are handling many requests at once, thus it's expected CPU cores will be
+      // fully utilized. This differs from browser environments, where a user is typically doing
+      // only one thing at a time and thus likely has CPU cores to spare.
+
+      // v8 takes ownership of the v8::CppHeap when passed this way.
```
kentonv, line 308: I see the doc comments on CreateParams::cpp_heap say it takes ownership, but it sure seems strange that they don't take a unique_ptr, especially given that CppHeap::Create() returns a unique_ptr.

Can we verify somehow that the CppHeap really is being destroyed when the isolate is destroyed?

jasnell: Take look in `heap.cc` ... in the `Heap::ConfigureHeap` method... it takes the given `CppHeap*` and passes it off to a `std::unique_ptr` (`owning_cpp_heap_` defined in `heap.h`)

kentonv (approved):

src/workerd/jsg/setup.c++
kentonv: Weird!

jasnell: Yeah, when I first saw that I had to go hunting for it also. I wouldn't be surprised if they change that.

PR #1729, opened 2024-02-25
JSRPC: Promise Pipelining (and property access)

kentonv:
You can now access properties on an RPC object by just awaiting them:

    let value = await stub.someProperty;

And you can access properties on a _promise_ returned by an RPC call, in order to make speculative calls on stubs that you expect to be returned:

    await stub.someMethod().someOtherMethod();

Both of these are achieved via custom thenables rather than regular promises. This allows us to return a value that can be awaited, but which also has a wildcard property.

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1866,40 +1866,35 @@ jsg::Promise<jsg::Ref<Response>> Fetcher::fetch(
   return fetchImpl(js, JSG_THIS, kj::mv(requestOrUrl), kj::mv(requestInit));
 }
 
-kj::Maybe<Fetcher::RpcFunction> Fetcher::getRpcMethod(jsg::Lock& js, kj::StringPtr name) {
+kj::Maybe<jsg::Ref<JsRpcProperty>> Fetcher::getRpcMethod(jsg::Lock& js, kj::StringPtr name) {
   // This is like JsRpcStub::getRpcMethod(), but we also initiate a whole new JS RPC session
-  // each time the method is called.
+  // each time the method is called (handled by `getClientForOneCall()`, below).
 
   // Do not return a method for `then`, otherwise JavaScript decides this is a thenable, i.e. a
   // custom Promise, which will mean a Promise that resolves to this object will attempt to chain
   // with it, which is not what you want!
   if (name == "then"_kj) return kj::none;
 
-  return RpcFunction(JSG_VISITABLE_LAMBDA(
-      (methodName = kj::str(name), self = JSG_THIS),
-      (self),
-      (jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>& args) -> jsg::Promise<jsg::Value> {
-    // Same as JsRpcStub::getRpcMethod(), we want to prevent calling on the wrong `this`.
-    JSG_REQUIRE(args.This() == KJ_ASSERT_NONNULL(self.tryGetHandle(js)), TypeError,
-        "Illegal invocation");
+  return jsg::alloc<JsRpcProperty>(JSG_THIS, kj::str(name));
+}
 
-    auto& ioContext = IoContext::current();
-    auto worker = self->getClient(ioContext, kj::none, "jsRpcSession"_kjc);
-    auto event = kj::heap<api::JsRpcSessionCustomEventImpl>(
-        JsRpcSessionCustomEventImpl::WORKER_RPC_EVENT_TYPE);
+rpc::JsRpcTarget::Client Fetcher::getClientForOneCall(jsg::Lock& js) {
+  auto& ioContext = IoContext::current();
+  auto worker = getClient(ioContext, kj::none, "jsRpcSession"_kjc);
+  auto event = kj::heap<api::JsRpcSessionCustomEventImpl>(
+      JsRpcSessionCustomEventImpl::WORKER_RPC_EVENT_TYPE);
 
-    auto result = JsRpcStub::sendJsRpc(js, event->getCap(), methodName, args);
+  auto result = event->getCap();
 
-    // Arrange to cancel the CustomEvent if our I/O context is destroyed. But otherwise, we don't
-    // actually care about the result of the event. If it throws, the membrane will already have
-    // propagated the exception to any RPC calls that we're waiting on, so we even ignore errors
-    // here -- otherwise they'll end up logged as "uncaught exceptions" even if they were, in fact,
-    // caught elsewhere.
-    ioContext.addTask(worker->customEvent(kj::mv(event)).attach(kj::mv(worker))
-        .then([](auto&&) {}, [](kj::Exception&&) {}));
+  // Arrange to cancel the CustomEvent if our I/O context is destroyed. But otherwise, we don't
+  // actually care about the result of the event. If it throws, the membrane will already have
+  // propagated the exception to any RPC calls that we're waiting on, so we even ignore errors
+  // here -- otherwise they'll end up logged as "uncaught exceptions" even if they were, in fact,
+  // caught elsewhere.
+  ioContext.addTask(worker->customEvent(kj::mv(event)).attach(kj::mv(worker))
+      .then([](auto&&) {}, [](kj::Exception&&) {}));
 
-    return result;
-  }));
+  return result;
 }
 
 static jsg::Promise<void> throwOnError(jsg::Lock& js,
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1878,7 +1878,8 @@ kj::Maybe<jsg::Ref<JsRpcProperty>> Fetcher::getRpcMethod(jsg::Lock& js, kj::Stri
   return jsg::alloc<JsRpcProperty>(JSG_THIS, kj::mv(name));
 }
 
-rpc::JsRpcTarget::Client Fetcher::getClientForOneCall(jsg::Lock& js) {
+rpc::JsRpcTarget::Client Fetcher::getClientForOneCall(
+    jsg::Lock& js, kj::Vector<kj::StringPtr>& path) {
   auto& ioContext = IoContext::current();
   auto worker = getClient(ioContext, kj::none, "jsRpcSession"_kjc);
   auto event = kj::heap<api::JsRpcSessionCustomEventImpl>(
@@ -1893,6 +1894,8 @@ rpc::JsRpcTarget::Client Fetcher::getClientForOneCall(jsg::Lock& js) {
   // caught elsewhere.
   ioContext.addTask(worker->customEvent(kj::mv(event)).attach(kj::mv(worker))
       .then([](auto&&) {}, [](kj::Exception&&) {}));
+
+  // (Don't extend `path` because we're the root.)
 
   return result;
 }
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1866,7 +1866,7 @@ jsg::Promise<jsg::Ref<Response>> Fetcher::fetch(
   return fetchImpl(js, JSG_THIS, kj::mv(requestOrUrl), kj::mv(requestInit));
 }
 
-kj::Maybe<jsg::Ref<JsRpcProperty>> Fetcher::getRpcMethod(jsg::Lock& js, kj::StringPtr name) {
+kj::Maybe<jsg::Ref<JsRpcProperty>> Fetcher::getRpcMethod(jsg::Lock& js, kj::String name) {
   // This is like JsRpcStub::getRpcMethod(), but we also initiate a whole new JS RPC session
   // each time the method is called (handled by `getClientForOneCall()`, below).
 
@@ -1875,7 +1875,7 @@ kj::Maybe<jsg::Ref<JsRpcProperty>> Fetcher::getRpcMethod(jsg::Lock& js, kj::Stri
   // with it, which is not what you want!
   if (name == "then"_kj) return kj::none;
 
-  return jsg::alloc<JsRpcProperty>(JSG_THIS, kj::str(name));
+  return jsg::alloc<JsRpcProperty>(JSG_THIS, kj::mv(name));
 }
 
 rpc::JsRpcTarget::Client Fetcher::getClientForOneCall(jsg::Lock& js) {
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -562,7 +562,8 @@ public:
     return getRpcMethod(js, kj::mv(name));
   }
 
-  rpc::JsRpcTarget::Client getClientForOneCall(jsg::Lock& js) override;
+  rpc::JsRpcTarget::Client getClientForOneCall(
+      jsg::Lock& js, kj::Vector<kj::StringPtr>& path) override;
 
   JSG_RESOURCE_TYPE(Fetcher, CompatibilityFlags::Reader flags) {
     // WARNING: New JSG_METHODs on Fetcher must be gated via compatibility flag to prevent
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -557,9 +557,9 @@ public:
 
   jsg::Promise<ScheduledResult> scheduled(jsg::Lock& js, jsg::Optional<ScheduledOptions> options);
 
-  kj::Maybe<jsg::Ref<JsRpcProperty>> getRpcMethod(jsg::Lock& js, kj::StringPtr name);
+  kj::Maybe<jsg::Ref<JsRpcProperty>> getRpcMethod(jsg::Lock& js, kj::String name);
   kj::Maybe<jsg::Ref<JsRpcProperty>> getRpcMethodForTestOnly(jsg::Lock& js, kj::String name) {
-    return getRpcMethod(js, name);
+    return getRpcMethod(js, kj::mv(name));
   }
 
   rpc::JsRpcTarget::Client getClientForOneCall(jsg::Lock& js) override;
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -18,6 +18,7 @@
 #include "url-standard.h"
 #include "blob.h"
 #include <workerd/io/compatibility-date.capnp.h>
+#include "worker-rpc.h"
 
 namespace workerd::api {
 
@@ -419,7 +420,7 @@ typedef kj::OneOf<SocketAddress, kj::String> AnySocketAddress;
 //
 // TODO(cleanup): This probably doesn't belong in `http.h` anymore. And perhaps it should be
 //   renamed, though I haven't heard any great suggestions for what the name should be.
-class Fetcher: public jsg::Object {
+class Fetcher: public JsRpcClientProvider {
 public:
   // Should we use a fake https base url if we lack a scheme+authority?
   enum class RequiresHostAndProtocol {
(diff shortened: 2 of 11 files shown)


src/workerd/api/http.c++
```diff
@@ -1868,40 +1868,36 @@ jsg::Promise<jsg::Ref<Response>> Fetcher::fetch(
...
-    ioContext.addTask(worker->customEvent(kj::mv(event)).attach(kj::mv(worker))
-        .then([](auto&&) {}, [](kj::Exception&&) {}));
-
-    return result;
-  }));
+  return jsg::alloc<JsRpcProperty>(JSG_THIS, kj::mv(name));
+}
+
+rpc::JsRpcTarget::Client Fetcher::getClientForOneCall(
+    jsg::Lock& js, kj::Vector<kj::StringPtr>& path) {
```
jasnell, line 1884: Non-blocking nit: the unused `path` here is... interesting. It might be worth renaming this to `unused` or `_`, or leaving it unnamed entirely (but if I recall correctly I don't think you're a fan of leaving it unnamed).

src/workerd/api/worker-rpc.c++
```diff
@@ -148,64 +148,182 @@ private:
...
+  // switched to RPC methods being implemented as callable objects (JsRpcProperty), this became
+  // impossible, because V8's SetCallAsFunctionHandler() arranges that `this` is bound to the
+  // callable object itself, regardless of how it was invoked. So now we cannot detect the
+  // situation above, because V8 never tells us about `obj` at all.
+  //
+  // Oh well. It's not a big deal. Just annoying that we have to forever support tearing RPC
+  // methods off their source object, even if we change implementations to something where that's
+  // less convenient.
+
+  kj::Vector<kj::StringPtr> path;
```
jasnell, line 182: A comment here very briefly explaining what `path` is would be helpful.

src/workerd/api/worker-rpc.c++
```diff
@@ -148,64 +148,182 @@ private:
...
+    }))),
+    .pipeline = kj::mv(callResult),
+  };
+}
+
+jsg::Ref<JsRpcPromise> JsRpcProperty::call(const v8::FunctionCallbackInfo<v8::Value>& args) {
+  jsg::Lock& js = jsg::Lock::from(args.GetIsolate());
+
+  auto [promise, pipeline] = callImpl(js, [&](rpc::JsRpcTarget::CallParams::Operation::Builder op) {
+    kj::Vector<jsg::JsValue> argv(args.Length());
```
jasnell, line 216: non-blocking nit.. would KJ_STACK_ARRAY be ok here?

src/workerd/api/worker-rpc.c++
```diff
@@ -148,64 +148,182 @@ private:
...
+      jsg::JsRef<jsg::JsPromise>(js, promise),
+      IoContext::current().addObject(kj::heap(kj::mv(pipeline))));
+}
+
+namespace {
+
+jsg::JsValue thenImpl(jsg::Lock& js, v8::Local<v8::Promise> promise,
+      v8::Local<v8::Function> handler, jsg::Optional<v8::Local<v8::Function>> errorHandler) {
+  KJ_IF_SOME(e, errorHandler) {
+    return jsg::JsPromise(jsg::check(promise->Then(js.v8Context(), handler, e)));
```
jasnell, line 241: The use of `jsg::check(...)` here will call these to throw synchronously... which I think is what we probably want in the cases that would normally cause `->Then(...)` to fail. But it's worth double checking, and likely worth a comment in here since normally we would expect `then()`, `catch()` and `finally()` to catch exceptions and turn them into rejections.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -165,6 +204,28 @@ export let namedServiceBinding = {
     // Properties that return a function can actually be called.
     assert.strictEqual(await env.MyService.functionProperty(19, 6), 13);
 
+    // Members of an object-typed property can be invoked.
+    assert.strictEqual(await env.MyService.objectProperty.func(6, 4), 24);
+    assert.strictEqual(await env.MyService.objectProperty.deeper.deepFunc(6, 3), 2);
+    assert.strictEqual(await env.MyService.objectProperty.counter5.increment(3), 8);
```
jasnell, line 210: non-blocking nit: for completeness, let's have a test that uses `Promise.all(...)`, `Promise.any`, `Promise.race`, and `Promise.allSettled` with these to ensure the concurrent outstanding requests have no problem.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -80,6 +99,26 @@ export class MyService extends WorkerEntrypoint {
...
+      someText: "hello",
+    };
+  }
+
+  get promiseProperty() {
+    return scheduler.wait(10).then(() => 123);
+  }
+
+  get rejectingPromiseProperty() {
+    return Promise.reject(new Error("REJECTED"));
```
jasnell, line 120: I might just be overlooking it here, but is there also a test case with a synchronous throw?

jasnell: Also, a test verifying the behavior when a non-typical exception is thrown (e.g. an `AggegateError` triggered by `Promise.any([...])`. Our tunneled exception mechanism currently doesn't handle these directly so we likely should test the behavior.

src/workerd/api/http.c++
kentonv: I don't like removing the name since it removes information that's useful to the reader.

I can add a comment, though.

src/workerd/api/worker-rpc.c++
kentonv: Looks like it, yeah. (Not new code in this PR but I'll go ahead and improve it.)

kentonv: Actually it doesn't work since JsValue doesn't have a default constructor. Will have to revisit later.

src/workerd/api/worker-rpc.c++
kentonv: If this throws here, it's because calling `.then()` on the underlying V8 promise threw synchronously. If the V8 promise threw synchronously then our wrapper presumably should too.

But I should make sure KJ exceptions thrown synchronously from capnp are caught and turned async...

src/workerd/api/http.c++
justin-mp: There's always the idiom where the unused name is a comment:
```
rpc::JsRpcTarget::Client Fetcher::getClientForOneCall(
    jsg::Lock& js, kj::Vector<kj::StringPtr>& /*path*/) 
```

src/workerd/api/worker-rpc.c++
jasnell: > If this throws here, it's because calling .then() on the underlyuing V8 promise threw...

Yep. As I said, likely worth a brief comment here.

src/workerd/api/tests/js-rpc-test.js
kentonv: I've added a test of concurrent calls which uses `Promise.all()`. I don't see a need to test all the different `Promise` helpers since there's no reason to expect any of them behave unusually on thenables.

src/workerd/api/tests/js-rpc-test.js
kentonv: Added tests for throwing exceptions.

Not sure what you are asking for regarding non-typical exceptions. We know that these do not work, but it's because we're relying on the general exception tunneling behavior, which is tested elsewhere.

kentonv:
Had to rebase over the JsMap change in order to resolve merge conflicts so that tests would run, but this means this branch is running ahead of what's merged downstream.

src/workerd/api/worker-rpc.h
```diff
@@ -81,6 +81,115 @@ class JsRpcTarget: public jsg::Object {
...
+  jsg::JsValue finally(jsg::Lock& js, v8::Local<v8::Function> onFinally);
+
+  // Get a nested property, using pipelining.
+  kj::Maybe<jsg::Ref<JsRpcProperty>> getProperty(jsg::Lock& js, kj::String name);
+
+  JSG_RESOURCE_TYPE(JsRpcPromise) {
+    JSG_WILDCARD_PROPERTY(getProperty);
+    JSG_METHOD(then);
+    JSG_METHOD_NAMED(catch, catch_);
+    JSG_METHOD(finally);
```
jasnell, line 131: how should type generation metadata be handled for these?

src/workerd/api/worker-rpc.h
```diff
@@ -81,6 +81,115 @@ class JsRpcTarget: public jsg::Object {
...
+  }
+
+private:
+  jsg::JsRef<jsg::JsPromise> inner;
```
jasnell, line 135: Should this implement gc visitation for the promise held here?

src/workerd/api/worker-rpc.h
```diff
@@ -81,6 +81,115 @@ class JsRpcTarget: public jsg::Object {
...
+  // Get a nested property, using pipelining.
+  kj::Maybe<jsg::Ref<JsRpcProperty>> getProperty(jsg::Lock& js, kj::String name);
+
+  JSG_RESOURCE_TYPE(JsRpcPromise) {
+    JSG_WILDCARD_PROPERTY(getProperty);
+    JSG_METHOD(then);
+    JSG_METHOD_NAMED(catch, catch_);
+    JSG_METHOD(finally);
+  }
+
```
jasnell, line 133: This should implement the `visitForMemoryInfo(...)` tracking the held promise

src/workerd/api/worker-rpc.h
```diff
@@ -81,6 +81,115 @@ class JsRpcTarget: public jsg::Object {
...
+  kj::Maybe<jsg::Ref<JsRpcProperty>> getProperty(jsg::Lock& js, kj::String name);
+
+  JSG_RESOURCE_TYPE(JsRpcProperty) {
+    JSG_CALLABLE(call);
+    JSG_WILDCARD_PROPERTY(getProperty);
+    JSG_METHOD(then);
+    JSG_METHOD_NAMED(catch, catch_);
+    JSG_METHOD(finally);
+  }
+
```
jasnell, line 172: This should implement the `visitForMemoryInfo(...)` tracking the held `parent`

src/workerd/api/worker-rpc.h
```diff
@@ -81,6 +81,115 @@ class JsRpcTarget: public jsg::Object {
...
+  // Get a nested property, using pipelining.
+  kj::Maybe<jsg::Ref<JsRpcProperty>> getProperty(jsg::Lock& js, kj::String name);
+
+  JSG_RESOURCE_TYPE(JsRpcProperty) {
+    JSG_CALLABLE(call);
+    JSG_WILDCARD_PROPERTY(getProperty);
+    JSG_METHOD(then);
+    JSG_METHOD_NAMED(catch, catch_);
+    JSG_METHOD(finally);
```
jasnell, line 170: How should type metadata be handled here?

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -286,5 +405,20 @@ export let receiveStubOverRpc = {
...
+  },
+}
+
+export let promisePipelining = {
+  async test(controller, env, ctx) {
+    assert.strictEqual(await env.MyService.makeCounter(12).increment(3), 15);
+
+    assert.strictEqual(await env.MyService.getAnObject(5).foo, 128);
+    assert.strictEqual(await env.MyService.getAnObject(5).counter.increment(7), 12);
   },
```
jasnell, line 423: A test verifying the expected behavior if one of these promises is shared across IoContext's and awaited would be good.

src/workerd/api/http.h
```diff
@@ -556,14 +557,14 @@ class Fetcher: public jsg::Object {
...
   jsg::Promise<ScheduledResult> scheduled(jsg::Lock& js, jsg::Optional<ScheduledOptions> options);
 
-  using RpcFunction = jsg::Function<jsg::Promise<jsg::Value>(
-      const v8::FunctionCallbackInfo<v8::Value>& info)>;
-
-  kj::Maybe<RpcFunction> getRpcMethod(jsg::Lock& js, kj::StringPtr name);
-  kj::Maybe<RpcFunction> getRpcMethodForTestOnly(jsg::Lock& js, kj::String name) {
-    return getRpcMethod(js, name);
+  kj::Maybe<jsg::Ref<JsRpcProperty>> getRpcMethod(jsg::Lock& js, kj::String name);
+  kj::Maybe<jsg::Ref<JsRpcProperty>> getRpcMethodForTestOnly(jsg::Lock& js, kj::String name) {
```
jasnell, line 561: A comment here explaining the difference would be helpful

src/workerd/api/worker-rpc.h
```diff
@@ -81,6 +81,115 @@ class JsRpcTarget: public jsg::Object {
...
+// This class is NOT part of the JavaScript class heirarchy (it has no JSG_RESOURCE_TYPE block),
+// it's only a C++ class used to abstract how to get a capnp client out of the object.
+class JsRpcClientProvider: public jsg::Object {
+public:
+  // Get a capnp client that can be used to dispatch one call.
+  //
+  // If this isn't the root object (i.e. this is a JsRpcProperty), the property path starting from
+  // the root object will be appended to `path`.
+  virtual rpc::JsRpcTarget::Client getClientForOneCall(
+      jsg::Lock& js, kj::Vector<kj::StringPtr>& path) = 0;
```
jasnell, line 96: Non-blocking but I'm not a huge fan of using a `kj::Vector` for the path here. Something like a simple stack allocated struct would probably suffice?

```cpp
struct Path {
  kj::Maybe<Path&> parent = kj::none;
  kj::StringPtr name;
  kj::String toString();
}
```

So that this could be called simply as...

```
getClientForOneCall(js, Path { path, "foo"_kjc" })
```

But not that critical and don't feel too strongly about it.

src/workerd/api/worker-rpc.h
```diff
@@ -96,6 +96,46 @@ class JsRpcClientProvider: public jsg::Object {
...
+  rpc::JsRpcTarget::Client getClientForOneCall(
+      jsg::Lock& js, kj::Vector<kj::StringPtr>& path) override;
+
+  // Implement standard Promise interface, especially `then()` so that this works as a custom
+  // thenable.
+  //
+  // Note that we intentionally return jsg::JsValue rather than jsg::JsPromise because we actually
+  // do not want the JSG glue to recognize we're returning a promise triggering behavior that pins
+  // the JsRpcPromise in memory until it resolves. It's actually fine if the JsRpcPromise is GC'd
+  // before the promise resolves.
```
MellowYarker, line 118: > // It's actually fine if the JsRpcPromise is GC'd before the promise resolves.

nit: I'm a bit confused by this, why can the object representing the RPC promise be GC'd before the result is ready? I'm assuming it's because calling `.then()` ensures we aren't dropping the `inner` promise, but in that case it might be helpful to mention that in this comment.

geelen:
Just a general comment here, how are we going to represent the return type of both an RpcTarget or a property in Typescript wrt promise pipelining? Both doing multiple awaits and relying on pipeline need to be valid:

```ts
let value = await stub.someProperty
let x = value.inner
let y = await stub.someProperty.inner

await stub.someMethod().someOtherMethod();
await (await stub.someMethod()).someOtherMethod();
```

@mrbbot @GregBrimble thoughts?

src/workerd/api/worker-rpc.h
kentonv: I have it on my TODO list to circle back and add proper GC visitation a little later.

src/workerd/api/worker-rpc.h
kentonv: That's a really interesting question that is probably going to require help from some people who know more about typescript than me. I'll start a chat thread.

src/workerd/api/worker-rpc.h
kentonv: Will do that at the same time as visitForGc.

src/workerd/api/worker-rpc.h
kentonv: Oh @mrbbot already has suggestions below, hah!

src/workerd/api/worker-rpc.h
kentonv: `path` is an "out" parameter. The callee fills it in, for the caller to use.

If we wanted to avoid any heap allocation here we could perhaps do it by tracking the "depth" of each property, so that we can pre-allocate the capnp list where the path is to be written, and then have `getClientForOneCall` directly fill that in. But, I don't think it's worth the complexity.

src/workerd/io/worker-interface.capnp
```diff
@@ -220,11 +220,33 @@ struct JsValue {
...
+  struct CallParams {
+    union {
+      methodName @0 :Text;
+      # Equivalent to `methodPath` where the list has only one element equal to this.
+
+      methodPath @2 :List(Text);
+      # Path of properties to follow from the JsRpcTarget itself to find the method being called.
+      # E.g. if the application does:
+      #
+      #     myRpcTarget.foo.bar.baz()
```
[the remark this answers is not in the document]
kentonv: Yep, that works by extension.

src/workerd/api/worker-rpc.h
kentonv: I've extended the comments slightly to try to clarify. But essentially the JsRpcPromise is just an API wrapper to provide a JavaScirpt interface, it's not the actual promise. The actual promise keeps running even if the JsRpcPromise goes away.

* e771128b
* ab35cef1

src/workerd/api/tests/js-rpc-test.js
kentonv: Although it's not intended to work, there are some cases that do work, and I suppose it's important to make sure they don't break later. Test added:

ce1f3f8607d2468cc7e6eccc3e845329a4f2b7ff

src/workerd/api/worker-rpc.h
jasnell: ah, right, overlooked that part. ok.

src/workerd/api/worker-rpc.c++
```diff
@@ -148,11 +148,28 @@ private:
 
 } // namespace
 
-jsg::Promise<jsg::Value> JsRpcStub::sendJsRpc(
-    jsg::Lock& js,
-    rpc::JsRpcTarget::Client client,
-    kj::StringPtr name,
-    const v8::FunctionCallbackInfo<v8::Value>& args) {
+jsg::Promise<jsg::Value> JsRpcProperty::call(const v8::FunctionCallbackInfo<v8::Value>& args) {
```
MellowYarker: I was reading through the code locally and realized I'm actually a bit confused how we get from `Fetcher::getRpcMethod()` to sending the rpc. 

When we return a `JsRpcProperty`, does the `JSG_CALLABLE(call)` in `JsRpcProperty` mean that in JS if I actually use `()` after the property name it will run this `JsRpcProperty::call()` method? This might be worth a comment somewhere.

kentonv: Yes, that's what JSG_CALLABLE means... the object can be called as a function.

kentonv:
9babd0d is new, the rest of the above is a rebase.

src/workerd/api/worker-rpc.c++
kentonv: 0bf67e4

jasnell (approved):

MellowYarker (approved):

PR #1732, opened 2024-02-26
Remove use of deprecated V8 WrapperClassId.

kentonv:


jasnell (dismissed):

kentonv:
OK I dug into it and it turns out V8 introduced a notion of "droppable" references exactly to cover our use case here. Annoying  that the deprecation notice didn't mention this, I had to trawl through issue history to find it! But the new API is definitely better.

jasnell (approved):

PR #1733, opened 2024-02-26
Add compatibility flag for queues JSON format by default

jbwcloudflare:
Adds a new compat flag + date to switch to JSON as the default message format for Queues.

### DIFF
--- a/src/workerd/api/queue.c++
+++ b/src/workerd/api/queue.c++
@@ -5,6 +5,7 @@
 #include "queue.h"
 #include "util.h"
 
+#include <workerd/io/features.h>
 #include <workerd/jsg/buffersource.h>
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/ser.h>
@@ -193,6 +194,9 @@ kj::Promise<void> WorkerQueue::send(jsg::Lock& js,
   KJ_IF_SOME(type, contentType) {
     headers.add("X-Msg-Fmt", type);
     serialized = serialize(js, body, type, SerializeArrayBufferBehavior::DEEP_COPY);
+  } else if (workerd::FeatureFlags::get(js).getQueuesJsonMessages()) {
+    headers.add("X-Msg-Fmt", IncomingQueueMessage::ContentType::JSON);
+    serialized = serialize(js, body, IncomingQueueMessage::ContentType::JSON, SerializeArrayBufferBehavior::DEEP_COPY);
   } else {
     // TODO(cleanup) send message format header (v8) by default
     serialized = serializeV8(js, body);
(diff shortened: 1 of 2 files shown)


src/workerd/api/queue.c++
```diff
@@ -174,7 +174,8 @@ jsg::JsValue deserialize(jsg::Lock& js, rpc::QueueMessage::Reader message) {
 
 kj::Promise<void> WorkerQueue::send(jsg::Lock& js,
                                     jsg::JsValue body,
-                                    jsg::Optional<SendOptions> options) {
+                                    jsg::Optional<SendOptions> options,
+                                    CompatibilityFlags::Reader flags) {
```
jasnell, line 178: Note that this is no longer really necessary. You can grab the feature flags also using the `FeatureFlags` type (see `/workerd/io/features.h`)

jbwcloudflare: thanks for the tip! updated to use this approach

jbwcloudflare:
@a-robinson could you take a look please? 🙏

a-robinson (approved):
It makes me sad to start silently dropping fields that don't serialize to json properly, but I take it the UX improvements in HTTP pull and the Dash are too important at this point?

One thing I'd like to see before checking this in is updating  some of the tests in the internal repo appropriately, but this code LGTM.

jbwcloudflare:
> It makes me sad to start silently dropping fields that don't serialize to json properly, but I take it the UX improvements in HTTP pull and the Dash are too important at this point?

Yeah, @elithrar requested the change, given the addition of HTTP pull consumers. Using V8 format by default makes it too easy for folks to get stuck putting messages into a Queue that they don't know how to pull back out.

> One thing I'd like to see before checking this in is updating some of the tests in the internal repo appropriately, but this code LGTM.

👍 I've got a PR started with some added tests.

src/workerd/io/compatibility-date.capnp
```diff
@@ -377,4 +377,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("no_nodejs_als");
   # Enables the availability of the Node.js AsyncLocalStorage API independently of the full
   # node.js compatibility option.
+
+  queuesJsonMessages @42 :Bool
+      $compatEnableFlag("queues_json_messages")
+      $compatDisableFlag("no_queues_json_messages")
+      $compatEnableDate("2024-03-04");
```
kentonv, line 384: This date is too soon: This will not be rolled out to production, much less to the validator, by Monday. Please push this back by two weeks.

If any release has already been built with this compat flag, that release will need to be patched to fix this, otherwise when the validator is updated (presumably after March 4) it will retroactively change the meaning of past compat dates.

jasnell: https://github.com/cloudflare/workerd/pull/1744

PR #1735, opened 2024-02-27
add LOG_PERIODICALLY and LOG_ONCE

Frederik-Baetens:
Maybe a general LOG_PERIODICALLY would be useful too, so users could just pass in whatever warning level they'd like?

### DIFF
--- a/src/workerd/util/sentry.h
+++ b/src/workerd/util/sentry.h
@@ -48,28 +48,27 @@ inline kj::StringPtr maybeOmitColoFromSentry(uint32_t coloId) {
     KJ_LOG(severity, __VA_ARGS__); \
   }
 
-// Log this to Sentry once ever per process. Typically will be better to use LOG_ERROR_PERIODICALLY (or a future LOG_WARNING_PERIODICALLY level equivalent).
-#define LOG_WARNING_ONCE(...)                                                  \
+// Log this to Sentry once ever per process. Typically will be better to use LOG_PERIODICALLY.
+#define LOG_ONCE(severity, ...)                                                \
   do {                                                                         \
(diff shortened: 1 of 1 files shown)


justin-mp (changes requested):
As you suggest, we should refactor LOG_WARNING_PERIODICALLY and LOG_ERROR_PERIODICALLY to use the same LOG_PERIODICALLY code.  The periodic stuff is gnarly enough without having two nearly-identical copies of it.

src/workerd/util/sentry.h
```diff
@@ -48,39 +48,44 @@ inline kj::StringPtr maybeOmitColoFromSentry(uint32_t coloId) {
...
       KJ_UNIQUE_NAME(lastLogged) = now;                                        \
-      KJ_LOG(ERROR, __VA_ARGS__);                                              \
+      KJ_LOG(severity, __VA_ARGS__);                                           \
     }                                                                          \
   } while (0)
 
+// Slightly more expensive than LOG_ERROR_ONCE. Avoid putting into a hot path (e.g. within a loop)
+// where an overhead of ~hundreds of nanoseconds per evaluation to retrieve the current time would
+// be prohibitive.
+#define LOG_ERROR_PERIODICALLY(...)                                            \
```
justin-mp, line 86: For symmetry with LOG_ONCE, you should probably have LOG_WARNING_PERIODICALLY.

Frederik-Baetens: :+1: added it. I initially left it out, because I think that people should just use LOG_PERIODICALLY directly, and think that the specific LOG_WARNING/LOG_ERROR variants are regrettable, but I'm not eager to go refactor them throughout the codebase.

justin-mp (approved):

PR #1738, opened 2024-02-27
MQ-498: Add support for delayed retries in Queues

sesteves:
This patch updates the queues api so that a delay value can be specified (from the consumer side) when retrying a batch or a message. Existing retry fields in capnp structs were maintained to prevent breaking changes. These fields are consolidated with the new ones on the broker side.

### DIFF
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -19,6 +19,7 @@
 #include "blob.h"
 #include <workerd/io/compatibility-date.capnp.h>
 #include "worker-rpc.h"
+#include "queue.h"
 
 namespace workerd::api {
 
@@ -530,16 +531,15 @@ public:
 
   struct QueueResult {
     kj::String outcome;
-    bool retryAll;
     bool ackAll;
-    kj::Array<kj::String> explicitRetries;
+    QueueRetryBatch retryBatch;
     kj::Array<kj::String> explicitAcks;
-
-    JSG_STRUCT(outcome, retryAll, ackAll, explicitRetries, explicitAcks);
+    kj::Array<QueueRetryMessage> retryMessages;
+    JSG_STRUCT(outcome, ackAll, retryBatch, explicitAcks, retryMessages);
   };
 
-  jsg::Promise<QueueResult> queue(
-      jsg::Lock& js, kj::String queueName, kj::Array<ServiceBindingQueueMessage> messages);
+  jsg::Promise<QueueResult> queue(jsg::Lock& js, kj::String queueName,
+                                  kj::Array<ServiceBindingQueueMessage> messages);
 
   struct ScheduledOptions {
     jsg::Optional<kj::Date> scheduledTime;
--- a/src/workerd/api/queue-test.js
+++ b/src/workerd/api/queue-test.js
@@ -15,12 +15,13 @@ export default {
     if (pathname === "/message") {
       const format = request.headers.get("X-Msg-Fmt") ?? "v8";
       if (format === "text") {
+        assert.strictEqual(request.headers.get("X-Msg-Delay-Secs"), "2")
         assert.strictEqual(await request.text(), "abc");
       } else if (format === "bytes") {
         const array = new Uint16Array(await request.arrayBuffer());
         assert.deepStrictEqual(array, new Uint16Array([1, 2, 3]));
       } else if (format === "json") {
-        assert.deepStrictEqual(await request.json(), {a: 1});
+        assert.deepStrictEqual(await request.json(), { a: 1 });
       } else if (format === "v8") {
         // workerd doesn't provide V8 deserialization APIs, so just look for expected strings
         const buffer = Buffer.from(await request.arrayBuffer());
@@ -31,6 +32,8 @@ export default {
         assert.fail(`Unexpected format: ${JSON.stringify(format)}`);
       }
     } else if (pathname === "/batch") {
+      assert.strictEqual(request.headers.get("X-Msg-Delay-Secs"), "2")
+
       const body = await request.json();
 
       assert.strictEqual(typeof body, "object");
@@ -44,10 +47,11 @@ export default {
       assert.deepStrictEqual(Buffer.from(body.messages[1].body, "base64"), Buffer.from([4, 5, 6]));
 
       assert.strictEqual(body.messages[2].contentType, "json");
-      assert.deepStrictEqual(JSON.parse(Buffer.from(body.messages[2].body, "base64")), [7, 8, {b: 9}]);
+      assert.deepStrictEqual(JSON.parse(Buffer.from(body.messages[2].body, "base64")), [7, 8, { b: 9 }]);
 
       assert.strictEqual(body.messages[3].contentType, "v8");
       assert(Buffer.from(body.messages[3].body, "base64").includes("value"));
+      assert.strictEqual(body.messages[3].delaySecs, 1);
     } else {
       assert.fail(`Unexpected pathname: ${JSON.stringify(pathname)}`);
     }
@@ -66,11 +70,12 @@ export default {
     assert.deepStrictEqual(batch.messages[1].body, new Uint8Array([7, 8, 9]));
 
     assert.strictEqual(batch.messages[2].id, "#2");
-    assert.deepStrictEqual(batch.messages[2].body, { c: {d: 10 } });
+    assert.deepStrictEqual(batch.messages[2].body, { c: { d: 10 } });
     batch.messages[2].retry();
 
     assert.strictEqual(batch.messages[3].id, "#3");
     assert.deepStrictEqual(batch.messages[3].body, batch.messages[3].timestamp);
+    batch.messages[3].retry({ delaySeconds: 2 });
 
     assert.strictEqual(batch.messages[4].id, "#4");
     assert.deepStrictEqual(batch.messages[4].body, new Map([["key", "value"]]));
@@ -79,17 +84,17 @@ export default {
   },
 
   async test(ctrl, env, ctx) {
-    await env.QUEUE.send("abc", { contentType: "text" });
+    await env.QUEUE.send("abc", { contentType: "text", delaySeconds: 2 });
     await env.QUEUE.send(new Uint16Array([1, 2, 3]), { contentType: "bytes" });
-    await env.QUEUE.send({a: 1}, { contentType: "json" });
+    await env.QUEUE.send({ a: 1 }, { contentType: "json" });
     await env.QUEUE.send(new Map([["key", "value"]]), { contentType: "v8" });
 
     await env.QUEUE.sendBatch([
       { body: "def", contentType: "text" },
-      { body: new Uint8Array([4,5,6]), contentType: "bytes" },
-      { body: [7, 8, {b: 9}], contentType: "json" },
-      { body: new Set(["value"]), contentType: "v8" },
-    ]);
+      { body: new Uint8Array([4, 5, 6]), contentType: "bytes" },
+      { body: [7, 8, { b: 9 }], contentType: "json" },
+      { body: new Set(["value"]), contentType: "v8", delaySeconds: 1 },
+    ], { delaySeconds: 2 });
 
     const timestamp = new Date();
     const response = await env.SERVICE.queue("test-queue", [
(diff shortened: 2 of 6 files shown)


sesteves:
@jbwcloudflare and @a-robinson will take a look.

jasnell (approved):
Change LGTM but should this also include tests?

src/workerd/io/worker-interface.capnp
```diff
@@ -147,6 +147,16 @@ struct QueueMessage @0x944adb18c0352295 {
   contentType @3 :Text;
 }
 
+struct QueueRetryBatch {
+  value @0 :Bool;
```
a-robinson, line 151: What does `value` mean here? It's very unclear from just looking at this file what this is for, and ideally capnp schemas are the best place to start reading when trying to understand part of the system.

Could you rename this and/or add a comment?

src/workerd/api/queue.h
```diff
@@ -108,24 +108,52 @@ struct IncomingQueueMessage {
...
+struct QueueRetryMessage {
+  kj::String msgId;
+  jsg::Optional<int> delaySeconds;
+  JSG_STRUCT(msgId, delaySeconds);
+};
+
 struct QueueResponse {
   uint16_t outcome;
+  QueueRetryBatch retryBatch;
+  // Deprecated, retryBatch should be used instead.
```
a-robinson, line 126: I assume you'll come back to remove the old deprecated fields after the queues team's code that uses this has been updated?

jbwcloudflare:
This looks pretty solid to me too -- still need to review the tests on the internal PR

src/workerd/io/worker-interface.capnp
```diff
@@ -155,6 +165,8 @@ struct QueueResponse @0x90e98932c0bfc0de {
   # List of Message IDs that were explicitly marked for retry
   explicitAcks @4 :List(Text);
   # List of Message IDs that were explicitly marked as acknowledged
+  retryBatch @5 :QueueRetryBatch;
```
a-robinson, line 168: Given that this struct is never used for communication outside of just the local process tree, we should be able to rip out the old fields (`retryAll` and `explicitRetries`). Although perhaps even better would be to use the name `explicitRetries` for the `:List(QueueRetryMessage)`, but that's up to you.

Although ripping them out now will require more code changes compared to doing it later (after the JS interfaces can also be cleaned up), so delaying it probably makes sense. But if you don't rip them out now, at least leave a comment in here that they're deprecated and should be removed later.

src/workerd/api/http.h
```diff
@@ -529,16 +530,19 @@ class Fetcher: public jsg::Object {
 
   struct QueueResult {
     kj::String outcome;
+    QueueRetryBatch retryBatch;
+    // Deprecated, retryBatch should be used instead.
     bool retryAll;
     bool ackAll;
+    kj::Array<QueueRetryMessage> retryMessages;
+    // Deprecated, retry should be used instead.
```
a-robinson, line 538:
```suggestion
    // Deprecated, retryMessages should be used instead.
```

sesteves:
Please do not merge yet. Planned changes:

1. Remove deprecated retry fields
2. Make `delayedSeconds` an optional field in capnp (add a boolean or assign bogus value for this purpose), as we need to distinguish between 0 (the default value) and "undefined".

src/workerd/api/queue.c++
```diff
@@ -386,23 +382,29 @@ void QueueMessage::retry() {
     IoContext::current().logWarning(msg);
     return;
   }
-  result->explicitRetries.findOrCreate(id, [this]() { return kj::heapString(id); } );
```
a-robinson, line 389: Don't we need to insert results into both `explicitRetries` and the new `retries` field until the consumers have this data have been switched over to use the new field? Unless you've already updated the queues code to read both fields and merge the results, this looks like it could cause bugs during the transition.

src/workerd/api/queue.c++
```diff
@@ -386,23 +382,29 @@ void QueueMessage::retry() {
...
+  if (result->retryAll.value) {
     auto msg = kj::str(
         "Received a call to ack() on message ", id, " after retryAll() was already called. "
         "Calling ack() on a message after calling retryAll() has no effect.");
     IoContext::current().logWarning(msg);
     return;
   }
 
-  if (result->explicitRetries.contains(id)) {
+  if (result->retries.find(id) != kj::none) {
```
a-robinson, line 407: Hm, I'm actually not 100% sure how this will behave if the `id` is found but it doesn't have a corresponding `delaySeconds` set. Will the `jsg::Optional` returned by `retries.find` be implicitly coerced into a `kj::none`, causing this to be false when we expect it to be true?

It's probably ok since it'd be odd if a `Maybe<Maybe<>>` with a none purely in the inner `Maybe` compared as equal to `none`, but this seems worth double checking if there isn't already a test that exercises this code.

src/workerd/api/queue.c++
```diff
@@ -433,18 +435,24 @@ QueueEvent::QueueEvent(jsg::Lock& js, Params params, IoPtr<QueueEventResult> res
...
 
-void QueueEvent::retryAll() {
+void QueueEvent::retryAll(jsg::Optional<QueueRetryOptions> options) {
   if (result->ackAll) {
     IoContext::current().logWarning(
         "Received a call to retryAll() after ackAll() was already called. "
         "Calling retryAll() after calling ackAll() has no effect.");
     return;
   }
-  result->retryAll = true;
```
a-robinson, line 443: Similar to above, shouldn't we continue setting the old field until we know that no code is still around that consumes from it?

a-robinson:
> as we need to distinguish between 0 (the default value) and "undefined".

Why? Doesn't `undefined` just mean no delay, which is the same thing as 0 delay?

src/workerd/io/worker-interface.capnp
sesteves: `value` indicates whether the batch should be retried. I will make it clear.

src/workerd/api/queue.h
sesteves: Yes, I am making these fields optional on the queue service side, and I will come back to remove them from here (in this PR).

src/workerd/io/worker-interface.capnp
sesteves: I will remove them in this PR, just need to update queues code first.

src/workerd/api/queue.c++
sesteves: Correct, queue code is already consolidating/merging both fields. In any case, I planning to remove deprecated fields in this PR.

src/workerd/api/queue.c++
sesteves: I will check this and get back to you later on.

src/workerd/api/queue.c++
sesteves: Queues code currently merges both fields.

sesteves:
> > as we need to distinguish between 0 (the default value) and "undefined".
> 
> Why? Doesn't `undefined` just mean no delay, which is the same thing as 0 delay?

-0 delay means no delay
-`undefined` means "inherit upper level delay"
E.g.
* Batch delay: 10
* 2 messages in the batch where one of them explicitly sets the delay to 0
* one message in the batch will have no delay, while the other needs to inherit the delay that was set at batch level 
* since capnp sends both messages with delay = 0 we cannot distinguish between the cases

src/workerd/api/queue.c++
```diff
@@ -394,23 +390,29 @@ void QueueMessage::retry() {
     IoContext::current().logWarning(msg);
     return;
   }
-  result->explicitRetries.findOrCreate(id, [this]() { return kj::heapString(id); } );
+
+  result->retries.upsert(kj::heapString(id), {});
+  KJ_IF_SOME(opts, options) {
+    KJ_IF_SOME(secs, opts.delaySeconds) {
+      result->retries.upsert(kj::heapString(id), {.delaySeconds = secs});
```
a-robinson, line 397: There's no need to allocate another copy of `id` and do an upsert here. The `upsert` call above returns a reference to the entry, so this code would be more efficient (and not really any more complicated) as something like:

```
  auto& entry = result->retries.upsert(kj::heapString(id), {});
  KJ_IF_SOME(opts, options) {
    KJ_IF_SOME(secs, opts.delaySeconds) {
      entry.value.delaySeconds = secs;
    }
  }
```

src/workerd/api/queue.c++
a-robinson: Thanks for confirming offline that this is ok 👍

src/workerd/io/worker-interface.capnp
```diff
@@ -147,14 +147,31 @@ struct QueueMessage @0x944adb18c0352295 {
...
 struct QueueResponse @0x90e98932c0bfc0de {
   outcome @0 :EventOutcome;
-  retryAll @1 :Bool;
-  ackAll @2 :Bool;
-  explicitRetries @3 :List(Text);
-  # List of Message IDs that were explicitly marked for retry
-  explicitAcks @4 :List(Text);
-  # List of Message IDs that were explicitly marked as acknowledged
+  ackAll @1 :Bool;
+  retryBatch @2 :QueueRetryBatch;
```
a-robinson, line 169: Feel free to ignore, but the naming difference in ack*All* vs retry*Batch* feels unnecessary vs aligning them both on either "all" or "batch".

src/workerd/io/worker-interface.capnp
```diff
@@ -147,14 +147,31 @@ struct QueueMessage @0x944adb18c0352295 {
   contentType @3 :Text;
 }
 
+struct QueueRetryBatch {
+  retry @0 :Bool;
+  union {
```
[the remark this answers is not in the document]
sesteves: My initial preference was to provide a bogus value or simply use a boolean. However, other team members were more inclined to use a union, especially since there were already some examples in the codebase utilizing unions with void fields. The decision to use the union appeared to be cleaner (the field does not show up in the resulting object) and less error-prone (queues code does not need to be aware of any special meanings associated with the `delaySeconds` field) and the overhead does not seem to be that big.

src/workerd/io/worker-interface.capnp
sesteves: The naming difference was introduced to deal with the breaking change. Queues code is expecting `retryAll` to be a boolean (runtimes in different versions can send either `retryAll` or `retryBatch` for some period of time, I presume). I know that we can make `retryAll` to accept one of two types, but the resulting code to deal with that would be less elegant than simply maintaining this naming difference. Queues code is already implemented following this approach. I propose we revisit this under MQ-435 (cleanup task to remove deprecated fields from queue code).

a-robinson (approved):
Looks good! Sorry about the slow back and forth.

src/workerd/api/queue-test.js
```diff
@@ -15,12 +15,13 @@ export default {
     if (pathname === "/message") {
       const format = request.headers.get("X-Msg-Fmt") ?? "v8";
       if (format === "text") {
+        assert.strictEqual(request.headers.get("X-Msg-Delay-Secs"), "2")
```
vickykont, line 18: This part of this test looks a little fragile to me, but I also don't know if there's an easy way to refactor it.

src/workerd/api/queue.h
```diff
@@ -108,24 +108,48 @@ struct IncomingQueueMessage {
...
+    jsg::Optional<int> delaySeconds;
+  };
+  struct RetryBatch {
+    bool retry;
+    jsg::Optional<int> delaySeconds;
+  };
+  RetryBatch retryBatch = {.retry = false};
   bool ackAll = false;
-  kj::HashSet<kj::String> explicitRetries;
+  kj::HashMap<kj::String, RetryOptions> retries;
```
vickykont, line 144: Are these `retryMessages`, `retryBatches` or both? Does it make sense to make this a little more explicit?

src/workerd/api/queue.c++
```diff
@@ -394,23 +390,29 @@ void QueueMessage::retry() {
     IoContext::current().logWarning(msg);
     return;
   }
-  result->explicitRetries.findOrCreate(id, [this]() { return kj::heapString(id); } );
+
+  auto& entry = result->retries.upsert(kj::heapString(id), {});
```
vickykont, line 394: Should this be moved after the `KJ_IF_SOME`s?

src/workerd/api/queue.c++
```diff
@@ -624,28 +632,38 @@ kj::Promise<WorkerInterface::CustomEvent::Result> QueueCustomEventImpl::sendRpc(
...
+      this->result.retryBatch.delaySeconds = retryBatch.getDelaySeconds();
     }
+
     this->result.explicitAcks.clear();
     for (const auto& msgId : respResult.getExplicitAcks()) {
       this->result.explicitAcks.insert(kj::heapString(msgId));
     }
+    this->result.retries.clear();
+    for (const auto& retry : respResult.getRetryMessages()) {
+      auto& entry = this->result.retries.upsert(kj::heapString(retry.getMsgId()), {});
```
vickykont, line 648: Same here. Should this be inside the `if`?

vickykont (approved):

PR #1739, opened 2024-02-27
Adapt /site-packages directory according to requirements.txt

garrettgu10:
RM-17870

This PR ensures that a worker's view of site-packages is limited only to the packages they specified. In pyodide-packages.tar we split each package wheel into its own folder named according to the package name, then at runtime we merge the virtual filesystem such that only the required wheels are seen within /site-packages. 

Here's how to run tests since they are disabled in CI:
```
for t in samples/pyodide*; do echo $t; ./bazel-bin/src/workerd/server/workerd test $t/config.capnp; done
```


src/pyodide/internal/python.js
```diff
@@ -277,16 +277,45 @@ function simpleRunPython(emscriptenModule, code) {
...
+      res.children.set(key, val);
+    });
+  }
+
+  return res;
+}
+
+export function mountLib(pyodide, requirements) {
   const [info, _] = parseTarInfo();
+  const sitePackagesInfo = buildSitePackages(info, requirements);
```
hoodmane, line 310: Maybe: 
```suggestion
  const [origTarInfo, _] = parseTarInfo();
  const info = buildSitePackages(origTarInfo, requirements);
```

src/pyodide/internal/python.js
```diff
@@ -277,16 +277,45 @@ function simpleRunPython(emscriptenModule, code) {
...
+
+  for(const req of ["stdlib", ...requirements]) {
+    const child = tarInfo.children.get(req);
+    if(!child) {
+      throw new Error(`Requirement ${req} not found in pyodide packages tar`);
+    }
+    child.children.forEach((val, key) => {
+      if(res.children.has(key)) {
+        throw new Error(`Folder ${key} being written by multiple packages`);
+      }
```
hoodmane, line 300: Not particularly important but this could also be a nonfolder file.

src/pyodide/internal/python.js
```diff
@@ -277,16 +277,45 @@ function simpleRunPython(emscriptenModule, code) {
...
+      res.children.set(key, val);
+    });
+  }
+
+  return res;
+}
+
+export function mountLib(pyodide, requirements) {
```
hoodmane, line 308: Rather than exporting `mountLib` it looks like we could just call it at the start of `setupPackages`?

src/pyodide/python-entrypoint-helper.js
```diff
@@ -61,10 +61,19 @@ function disabledLoadPackage() {
...
+
+/**
+ * Normalize a package name. Port of Python's packaging.utils.canonicalize_name.
+ * @param name The package name to normalize.
+ * @returns The normalized package name.
+ * @private
+ */
+function canonicalizePackageName(name) {
+  return name.replace(canonicalizeNameRegex, "-").toLowerCase();
+}
```
hoodmane, line 74: I think `getTransitiveRequirements` already does this so we could leave it out here.

src/pyodide/internal/python.js
garrettgu10: setupPackages is in another file so we would still need to export it, but I agree it makes sense to call it from there.

src/pyodide/python-entrypoint-helper.js
garrettgu10: I had to keep it to canonicalize the stdlib package names but it now lives in python.js

garrettgu10:
Looking at the failed CI builds, it looks like the windows build is just being flaky.

dom96:
> Looking at the failed CI builds, it looks like the windows build is just being flaky.

Yeah, I saw it being flaky previously

src/pyodide/BUILD.bazel
```diff
@@ -115,6 +116,18 @@ REPLACEMENTS = [
     ]
 ]
 
+load("//:build/pyodide_bucket.bzl", "PYODIDE_PACKAGE_BUCKET_URL")
+
+PYODIDE_BUCKET_MODULE = json.encode({
+    "PYODIDE_PACKAGE_BUCKET_URL": PYODIDE_PACKAGE_BUCKET_URL,
+})
```
dom96, line 123: Nice, awesome way to solve this.

build/pyodide_bucket.bzl
```diff
@@ -0,0 +1,3 @@
+PYODIDE_PACKAGE_BUCKET_URL = "https://pub-45d734c4145d4285b343833ee450ef38.r2.dev/20240305/"
```
hoodmane, line 1: Maybe comment here that it's factored out b/c it's shared between WORKSPACE and src/pyodide/BUILD.bazel

src/pyodide/internal/python.js
```diff
@@ -2,6 +2,23 @@ import { parseTarInfo } from "pyodide-internal:tar";
...
+
+/**
+ * Normalize a package name. Port of Python's packaging.utils.canonicalize_name.
+ * @param name The package name to normalize.
+ * @returns The normalized package name.
+ * @private
+ */
+export function canonicalizePackageName(name) {
+  return name.replace(canonicalizeNameRegex, "-").toLowerCase();
+}
```
hoodmane, line 17: I still don't think you need to do this, Pyodide is already doing it:
https://github.com/pyodide/pyodide/blob/main/src/js/load-package.ts#L150

src/pyodide/python-entrypoint-helper.js
```diff
@@ -76,7 +73,11 @@ async function setupPackages(pyodide) {
       .values()
       .map(({ packageData }) => packageData);
   }
-  const transitiveRequirements = new Set(packageDatas.map(({ name }) => name));
+  return new Set(packageDatas.map(({ name }) => name));
```
hoodmane, line 76: We should probably grab the `normalizedName` here so we don't have to renormalize it ourselves:
```suggestion
  return new Set(packageDatas.map(({ normalizedName }) => normalizedName));
```

src/pyodide/internal/python.js
```diff
@@ -2,6 +2,23 @@ import { parseTarInfo } from "pyodide-internal:tar";
...
+
+const STDLIB_PACKAGES = Object.values(LOCKFILE.packages)
+  .filter(({install_dir}) => install_dir === "stdlib")
+  .map(({name}) => canonicalizePackageName(name));
```
hoodmane, line 21: The name field should already be normalized. If this actually is necessary for some reason, we should also open an upstream bug.

hoodmane (approved):
Again I'd like to see `canonicalizePackageName` removed or if it is necessary there should be a Pyodide bug report. Otherwise it looks good to me. Thanks @garrettgu10!

src/pyodide/internal/python.js
garrettgu10: It's necessary because the name is normalized in the key but not in the name field for some reason.

src/pyodide/internal/python.js
```diff
@@ -277,8 +294,44 @@ function simpleRunPython(emscriptenModule, code) {
...
+    mode: 0o777,
+    type: 5,
+    modtime: 0,
+    size: 0,
+    path: "",
+    name: "",
+    parts: [],
+  };
+
+  for(const req of new Set([...STDLIB_PACKAGES, ...requirements])) {
```
dom96, line 309: Nit
```suggestion
  for (const req of new Set([...STDLIB_PACKAGES, ...requirements])) {
```

src/pyodide/internal/python.js
```diff
@@ -277,8 +294,44 @@ function simpleRunPython(emscriptenModule, code) {
...
+    const child = tarInfo.children.get(req);
+    if(!child) {
```
dom96, line 311: Nit
```suggestion
    if (!child) {
```

src/pyodide/internal/python.js
garrettgu10: I can rewrite it to use the name in the object key

src/pyodide/internal/python.js
```diff
@@ -277,8 +294,44 @@ function simpleRunPython(emscriptenModule, code) {
...
+      throw new Error(`Requirement ${req} not found in pyodide packages tar`);
+    }
+    child.children.forEach((val, key) => {
+      if(res.children.has(key)) {
```
dom96, line 315:
```suggestion
      if (res.children.has(key)) {
```

dom96 (approved):
Brilliant, looks good!

src/pyodide/internal/python.js
hoodmane: Run `npx prettier -w .` to format everything.

src/pyodide/internal/python.js
hoodmane: Interesting. Well that's fine then, maybe just put a comment saying that and leave as-is. And open a bug on Pyodide asking if this is intentional.

src/pyodide/python-entrypoint-helper.js
garrettgu10: Hm, that fails on edgeworker since normalizedName doesn't seem to be defined there. Imma just leave it the way it is if that's ok

src/pyodide/internal/python.js
garrettgu10: We need the names to be normalized on edgeworker since we're not calling loadPackage and the names are normalized in the tar file.

src/pyodide/python-entrypoint-helper.js
hoodmane: We can clean up later is fine.

src/pyodide/internal/python.js
hoodmane: We are calling `recursiveDependencies` which should handle this. But it's fine for now, we should fix it upstream and remove later but it's pretty low priority.

PR #1750, opened 2024-03-01
Adds python_workers compatibility flag

dom96:
This is necessary in order to allow upload for non-cloudflare users. This is safe because EWC has custom logic for blocking the upload of bundles containing Python modules, so only certain users will be allowed to upload.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -383,4 +383,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("no_queues_json_messages")
       $compatEnableDate("2024-03-18");
   # Queues bindings serialize messages to JSON format by default (the previous default was v8 format)
+
+  pythonWorkers @43 :Bool
+      $compatEnableFlag("python_workers");
+  # Enables Python Workers. Access to this flag is not restricted, instead bundles containing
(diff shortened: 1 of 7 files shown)


src/workerd/io/compatibility-date.capnp
```diff
@@ -383,4 +383,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("no_queues_json_messages")
       $compatEnableDate("2024-03-18");
   # Queues bindings serialize messages to JSON format by default (the previous default was v8 format)
+
+  pythonWorkers @43 :Bool
+      $compatEnableFlag("python_workers");
+  # Enables Python Workers. Access to this flag is not restricted, instead bundles containing
+  # Python modules are restricted in EWC.
+  #
+  # WARNING: Python Workers are still an experimental feature and thus subject to change.
```
harrishancock, line 392: Should we annotate it with `$experimental`?

harrishancock (approved):

src/workerd/io/compatibility-date.capnp
dom96: No, because doing so would restrict it to only cloudflare.com internal accounts (we want to restrict to those + account IDs we have added to a Flipper feature)

hoodmane:
Thanks!

PR #1752, opened 2024-03-01
Use on_fetch for Python Worker handler.

dom96:
We cannot use `fetch` for the handler name because it conflicts with the `fetch` function imported from `js`. This works out fine in JS, but not in Python where you end up calling the `fetch` handler recursively. Renaming this to `on_fetch` solves this problem.


src/pyodide/python-entrypoint-helper.js
```diff
@@ -124,7 +124,10 @@ export default {
   async fetch(request, env, ctx) {
     try {
       const { mainModule } = await getPyodide(ctx);
-      return await mainModule.fetch.callRelaxed(request, env, ctx);
+      if (mainModule.fetch) {
+        throw new Error("Python Worker should define an on_fetch method");
+      }
```
hoodmane, line 129: Definitely don't do this. If they do:
```
from js import fetch

def on_fetch():
   fetch(...)
```
it will throw an error.

hoodmane (approved):
LGTM if you remove the check for `main_module.fetch`.

hoodmane:
Actually, let's update all the samples `pyodide-xxx` too.

src/pyodide/python-entrypoint-helper.js
dom96: hmmm. That's unfortunate, is there a more reliable way to detect this?

dom96: What I'll do is show this error if `on_fetch` also isn't defined. I just want to make sure we get a nice error message for this case.

garrettgu10: It seems a little strange that there's a slight mismatch between what the error is saying vs. what the if statement is checking. 

I would have expected something more like if (no on_fetch) throw error("No on_fetch")

hoodmane: What if the user does:
```py
from js import fetch

async def on_alarm():
   resp = await fetch("http://example.com")
```
Presumably they are allowed to have an alarm handler but no fetch handler and they are allowed to fetch inside their alarm handler? Would they even see this in a clear way? Maybe wrangler should warn in the case that there is a `fetch` but no `on_fetch`, but I don't think it should be a hard error, and I don't think workerd should check for it.

dom96: @garrettgu10 yeah, you're totally right, I should just be throwing a nicer error when `on_fetch` isn't there

hoodmane: But people aren't required to have a fetch handler, right? What if the worker is just a cron job?

dom96: Yeah, but I will only throw it in this function (which won't be called unless a `fetch` handler is needed)

hoodmane: Ah so this is just trapping bugs in wrangler then?

dom96: It's just so folks who are using the old syntax get a nice error message.

dom96:
hmm, I wonder: should we change `test` to `on_test` too?

hoodmane:
Thanks @dom96!

jasnell:
I'm going to remove myself as a reviewer on this as I don't have enough context on the python stuff to do an adequate review.

garrettgu10:
> hmm, I wonder: should we change `test` to `on_test` too?

Do we use `test` for Python workers at the moment? Assuming it's the same semantics as in JavaScript workers, I think `test` is a much better descriptor than `on_test`.

PR #1755, opened 2024-03-01
Greatly simplify JS-backed ReadableStreams AllReader

jasnell:
The internal implementation of the JS-backed streams is rather overly complicated and has a few design flaws that make it difficult to debug and maintain. This, coupled with tests scattered across multiple repos with incomplete coverage. It's time to simplify the implementation and improve test coverage. This is the first of several clean up steps that will greatly simplify a number of cases. This commit focuses on one specific internal bit, follow on commits will touch other pieces. The plan is to do this as incrementally as possible spread over some time to ensure things don't break. The incomplete test coverage makes things a bit difficult so that will be a key focus.

Currently in a JS-backed `ReadableStream`, we have an underlying `ReadableStreamJsController` that has either a `kj::Own<ValueReadable>` or `kj::Own<ByteReadable>`. This owns a ref to the underlying `ReadableStreamDefaultController` or `ReadableByteStreamController` that actually provides the data along with a `kj::Own<Queue::Consumer>` that consumes the data from the underlying controller. In the existing design, we were rather needlessly passing ownership of the `ValueReadable` and `ByteReadable` to the `AllReader`. This proves to be problematic and unnecessary. Problematic because these are gc visitable types and changing the ownership ends up making the gc visitation semantics unnecessarily complicated and easy to get wrong (as a few recent fixed bugs demonstrate). Unnecessary because we can simply read directly from the `ReadableStream` instance itself without actually changing ownership of anything. The prior design was the result of a premature attempt at optimization that was completely unnecessary -- especially given the way v8 is able to optimistically gc the `ReadableStream` JS wrapper class.

This commit simplifies the `AllReader` to directly read from the `ReadableStream`. Rather than moving the `kj::Own<ValueReadable>` or `kj::Own<ByteReadable>` into the `AllReader`, we simply pass a jsg::Ref to the `ReadableStream` and read directly from it. We also remove some unnecessary book keeping around pending state changes since we only do serialized reads and can detect the state changes reliably when those complete. The commit also eliminates some duplication of the code to further simplify. Soon I will make a similar change to the `PumpToReader` and the couple of other places where we currently move the `kj::Own<ValueReadable>` and `kj::Own<ByteReadable>` out of the readable stream -- essentially the idea is through a couple of incremental PRs, completely eliminate moving that out.

The majority of added lines in this commit are in the form of new tests. The new tests (a set of c++ tests and new js tests) demonstrate the area of the api that is impacted by this commit.

A future new set of tests will seek to concretely and reliably verify the gc characteristics of the entire streams implementation.

### DIFF
--- a/src/workerd/api/streams/standard.c++
+++ b/src/workerd/api/streams/standard.c++
@@ -768,6 +768,9 @@ private:
   size_t pendingReadCount = 0;
   kj::Maybe<kj::OneOf<StreamStates::Closed, StreamStates::Errored>> maybePendingState = kj::none;
 
+  template <typename T>
+  jsg::Promise<T> readAll(jsg::Lock& js, uint64_t limit);
+
   void setPendingState(kj::OneOf<StreamStates::Closed, StreamStates::Errored> pending) {
     if (maybePendingState == kj::none) {
       maybePendingState = kj::mv(pending);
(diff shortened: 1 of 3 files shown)


src/workerd/api/streams/standard.c++
```diff
@@ -2905,8 +2858,11 @@ private:
           auto handle = KJ_ASSERT_NONNULL(result.value).getHandle(js);
           if (!handle->IsArrayBufferView() && !handle->IsArrayBuffer()) {
             auto error = js.v8TypeError("This ReadableStream did not return bytes.");
-            doError(js, error);
-            return loop(js);
+            auto rs = kj::mv(readable);
+            state.template init<StreamStates::Errored>(js.v8Ref(error));
```
mikea, line 2862: (not a review comment): I wonder if there is a `swap` std function that replaces the value? Maybe in kj? I often see this pattern of `kj:mv`ing things to temp.

src/workerd/api/streams/standard.c++
```diff
@@ -3267,102 +3223,74 @@ private:
...
-    return js.rejectedPromise<kj::Array<byte>>(KJ_EXCEPTION(FAILED,
+    return js.rejectedPromise<T>(KJ_EXCEPTION(FAILED,
         "jsg.TypeError: This ReadableStream is currently locked to a reader."));
   }
   disturbed = true;
+
+  // This operation leaves the stream locked and disturbed. The loop will read until
+  // the stream is closed or errored. If the limit is reached, the loop will error.
+
+  const auto readAll = [&]() -> jsg::Promise<T> {
```
mikea, line 3239: this def. deserves a separate method: it has two nested lambdas!

mikea:
The core change makes sense.

The c++ test is very dense with a lot of almost-duplication going around. Maybe a utility or two could cut every test by few lines? That would go a long way! Maybe custom asserts or additional setup functions/macros?

src/workerd/api/streams/standard.c++
jasnell: A swap method on `kj::OneOf`<...>` would be helpful, I think. We've had problems with this kind of pattern in the past where the value was moved but not replaced with something and it caused issues, The follow would be a whole lot nicer/safer

```
auto rs = state.template swap<StreamStates::Errored>(js.v8Ref(error));
```

src/workerd/api/streams/standard.c++
```diff
@@ -2790,113 +2793,63 @@ ReadableStreamJsController::getController() {
...
   }
 
 private:
-  kj::Maybe<IoContext&> ioContext;
-  kj::OneOf<StreamStates::Closed, StreamStates::Errored, Readable> state;
+  kj::OneOf<StreamStates::Closed, StreamStates::Errored, jsg::Ref<ReadableStream>> state;
   uint64_t limit;
   kj::Vector<jsg::BackingStore> parts;
   uint64_t runningTotal = 0;
+  kj::Maybe<kj::OneOf<StreamStates::Closed, StreamStates::Errored>> maybePendingState = kj::none;
```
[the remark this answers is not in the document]
jasnell: Given that it's only used here, I'd prefer to avoid using a typedef or using here. If I had to use this in a couple of places then I'd be fine with it.

src/workerd/api/streams/standard.c++
```diff
@@ -2790,113 +2793,63 @@ ReadableStreamJsController::getController() {
...
-              return readable->read(js, kj::none);
-            } else {
-              return readable->read(js);
-            }
-          });
-        };
-
-        auto onSuccess = [this](auto& js, ReadResult result) {
+      KJ_CASE_ONEOF(readable, jsg::Ref<ReadableStream>) {
+        auto onSuccess = [&](jsg::Lock& js, ReadResult result) -> jsg::Promise<PartList>{
```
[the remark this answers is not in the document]
jasnell: I'll list out the captures. I'd rather not split these out into separate functions as I think doing so spreads the code out a bit too much and can make it harder to follow.

src/workerd/api/streams/standard.c++
jasnell: Given that this already spread out over several other methods (e.g. `readAllText(...)` and `readAllBytes(...)` each call into this... My fear is that things are already spread out a bit too much, making understanding the code flow a bit difficult. I prefer to keep this as is for readability. I will explicitly list out the captures for sure tho.

src/workerd/api/streams/standard.c++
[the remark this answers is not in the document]
jasnell: will list out the captures

jasnell:
> The c++ test is very dense with a lot of almost-duplication going around. Maybe a utility or two could cut every test by few lines? That would go a long way! Maybe custom asserts or additional setup functions/macros?

I'll see about this in a separate PR. I actually want to see if we can introduce a new new test fixture utilities that would make a number of things easier. For instance, in node.js we have a utility like `mustCall(...)` that wraps a JS lambda and asserts with a failure if the function is not called by the end of the test. That said, for right now I don't want to hide too much of the detail in the tests and want each to stand on their own while I go through and audit the test coverage that we have.

mikea (approved):

PR #1756, opened 2024-03-02
JSRPC: Support "serializing" functions by turning them into stubs

kentonv:
It's simple: If you send someone a function, they get an RpcStub. The stub is callable. Calling it makes an RPC back to the isolate where the function was created, calling the original object there.

To reiterate, because this keeps confusing people: The function's code is NOT serialized. We are NOT magically moving code around. Closures strictly execute in the isolate where they were created.

I also went ahead and said that functions are allowed to have properties, but these work like properties on an RpcTarget. That is, they are not marshalled proactively when sending the function, but you can access them by accessing properties of the RpcStub representing the function.

There are also some other commits thrown in this PR because I did them at the same time:
* First two commits: Prevent sending instances of application-defined classes over RPC, and improve error messages around inability to serialize.
* Last two commits: Implement `visitForGc()` and `visitForMemoryInfo()`.

There are a lot of commits here but they are each pretty light...

### DIFF
--- a/src/node/internal/internal_assert.ts
+++ b/src/node/internal/internal_assert.ts
@@ -846,10 +846,7 @@ function isValidThenable(maybeThennable: any): boolean {
     return true;
   }
 
-  const isThenable = typeof maybeThennable.then === "function" &&
-    typeof maybeThennable.catch === "function";
-
-  return isThenable && typeof maybeThennable !== "function";
+  return typeof maybeThennable.then === "function";
 }
 
 export { AssertionError };
--- a/src/workerd/jsg/ser.c++
+++ b/src/workerd/jsg/ser.c++
@@ -44,10 +44,29 @@ v8::Maybe<uint32_t> Serializer::GetSharedArrayBufferId(
   return v8::Just(n);
 }
 
+void Serializer::throwDataCloneErrorForObject(jsg::Lock& js, v8::Local<v8::Object> obj) {
+  // The default error that V8 would generate is "#<TypeName> could not be cloned." -- for some
+  // reason, it surrounds the type name in "#<>", which seems bizarre? Let's generate a better
+  // error.
+  v8::Local<v8::String> message = js.str(kj::str(
+      "Could not serialize object of type \"", obj->GetConstructorName(), "\". This type does "
+      "not support serialization."));
+  js.throwException(jsg::JsValue(makeDOMException(js.v8Isolate, message, "DataCloneError")));
+}
+
 void Serializer::ThrowDataCloneError(v8::Local<v8::String> message) {
-  // makeDOMException could throw an exception. If it does, we'll end up crashing but that's ok?
   auto isolate = v8::Isolate::GetCurrent();
-  isolate->ThrowException(makeDOMException(isolate, message, "DataCloneError"));
+  try {
+    isolate->ThrowException(makeDOMException(isolate, message, "DataCloneError"));
+  } catch (JsExceptionThrown&) {
+    // Apparently an exception was thrown during the construction of the DOMException. Most likely
+    // we were terminated. In any case, we'll let that exception stay scheduled and propagate back
+    // to V8.
+  } catch (...) {
+    // A KJ exception was thrown, we'll have to convert it to JavaScript and propagate that
+    // exception instead.
+    throwInternalError(isolate, kj::getCaughtExceptionAsKj());
+  }
 }
 
 bool Serializer::HasCustomHostObject(v8::Isolate* isolate) {
@@ -71,6 +90,8 @@ v8::Maybe<bool> Serializer::IsHostObject(v8::Isolate* isolate, v8::Local<v8::Obj
 
 v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::Object> object) {
   try {
+    jsg::Lock& js = jsg::Lock::from(isolate);
+
     if (object->InternalFieldCount() != Wrappable::INTERNAL_FIELD_COUNT ||
         object->GetAlignedPointerFromInternalField(Wrappable::WRAPPABLE_TAG_FIELD_INDEX) !=
             &Wrappable::WRAPPABLE_TAG) {
@@ -79,9 +100,7 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
       //
       // We also get here if treatClassInstancesAsPlainObjects is false, and the object is an
       // application-defined class. We don't currently support serializing class instances.
-
-      // Call the default implementation to get the default error message.
-      return v8::ValueSerializer::Delegate::WriteHostObject(isolate, object);
+      throwDataCloneErrorForObject(js, object);
     }
 
     Wrappable* wrappable = reinterpret_cast<Wrappable*>(
@@ -99,9 +118,7 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
     if (!IsolateBase::from(isolate).serialize(
           Lock::from(isolate), typeid(*wrappable), *obj, *this)) {
       // This type is not serializable.
-
-      // Call the default implementation to get the default error message.
-      return v8::ValueSerializer::Delegate::WriteHostObject(isolate, object);
+      throwDataCloneErrorForObject(js, object);
     }
 
     return v8::Just(true);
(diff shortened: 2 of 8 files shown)


src/node/internal/internal_assert.ts
```diff
@@ -846,10 +846,7 @@ function isValidThenable(maybeThennable: any): boolean {
     return true;
   }
 
-  const isThenable = typeof maybeThennable.then === "function" &&
-    typeof maybeThennable.catch === "function";
-
-  return isThenable && typeof maybeThennable !== "function";
```
kentonv, line 852: I couldn't tell if there was a specific reason for the "not a function" check here. If so, I can remove this commit and keep my ugly monkey-patch instead.

jasnell: Looking at it again refreshed my memory, this bit is taken directly from the polyfill that deno is using (https://github.com/denoland/deno/blob/main/ext/node/polyfills/assert.ts#L864) ... looking at it I'm not entirely sure what the intent originally was but I think it's fine to fix this up.

src/workerd/jsg/ser.c++
```diff
@@ -41,22 +51,73 @@ v8::Maybe<uint32_t> Serializer::GetSharedArrayBufferId(
   return v8::Just(n);
 }
 
+void Serializer::scheduleDataCloneErrorForObject(v8::Isolate* isolate, v8::Local<v8::Object> obj) {
+  jsg::Lock& js = jsg::Lock::from(isolate);
+
+  try {
+    // The default error that V8 would generate is "#<TypeName> could not be cloned." -- for some
+    // reason, it surrounds the type name in "#<>", which seems bizarre? Let's generate a better
+    // error.
```
jasnell, line 60: +1 ... the default error has always been super weird.

jasnell (approved):

jasnell:
One additional thought ... It would be good to include some additional tests for the  structuredClone method to ensure it still maintains the expected behavior for things like functions as we're enabling these new bits. Just to make sure we don't accidentally introduce regressions... It might be that the existing structuredClone tests are already adequate but let's double check.

src/workerd/jsg/ser.c++
```diff
@@ -7,15 +7,25 @@
...
+      ser(js.v8Isolate, this) {
 #ifdef KJ_DEBUG
   kj::requireOnStack(this, "jsg::Serializer must be allocated on the stack");
 #endif
-  auto options = kj::mv(maybeOptions).orDefault({});
-  externalHandler = options.externalHandler;
+  if (!treatClassInstancesAsPlainObjects) {
+    prototypeOfObject = js.obj().getPrototype();
+  }
+  if (externalHandler != kj::none) {
```
jasnell, line 25: Non-blocking for now, but I wonder if there shouldn't be a separate config `Option` for handling functions this way. I imagine we might want to use an external handler for `structuredClone` cases where we may not actually want to handle functions this way.

src/workerd/jsg/ser.c++
```diff
@@ -41,22 +51,73 @@ v8::Maybe<uint32_t> Serializer::GetSharedArrayBufferId(
...
+  jsg::Lock& js = jsg::Lock::from(isolate);
+
+  try {
+    // The default error that V8 would generate is "#<TypeName> could not be cloned." -- for some
+    // reason, it surrounds the type name in "#<>", which seems bizarre? Let's generate a better
+    // error.
+    v8::Local<v8::String> message = js.str(kj::str(
+        "Could not serialize object of type \"", obj->GetConstructorName(), "\". This type does "
+        "not support serialization."));
+    isolate->ThrowException(makeDOMException(isolate, message, "DataCloneError"));
```
jasnell, line 64: non-blocking for later: having a `js.throwDomException(...)` would probably be nice.

src/workerd/jsg/ser.c++
```diff
@@ -41,22 +51,73 @@ v8::Maybe<uint32_t> Serializer::GetSharedArrayBufferId(
...
+  } catch (JsExceptionThrown&) {
+    // Apparently an exception was thrown during the construction of the DOMException. Most likely
+    // we were terminated. In any case, we'll let that exception stay scheduled and propagate back
+    // to V8.
+  }
```
jasnell, line 69: non-blocking for latre: Any chance of a `kj::Exception` being thrown? Should we handle it just in case?

src/workerd/jsg/ser.c++
```diff
@@ -41,22 +51,73 @@ v8::Maybe<uint32_t> Serializer::GetSharedArrayBufferId(
...
+v8::Maybe<bool> Serializer::IsHostObject(v8::Isolate* isolate, v8::Local<v8::Object> object) {
+  // This is only called if HasCustomHostObject() returned true.
+  KJ_ASSERT(!treatClassInstancesAsPlainObjects);
+  KJ_ASSERT(!prototypeOfObject.IsEmpty());
+
+  // If the object's prototype is Object.prototype, then it is a plain object, which we'll allow
+  // to be serialized normally. Otherwise, it is a class instance, which we should treat as a host
+  // object. Inside `WriteHostObject()` we will throw DataCloneError due to the object not having
+  // internal fields.
+  return v8::Just(object->GetPrototype() != prototypeOfObject);
```
jasnell, line 99: non-blocking for later: including a test to ensure that objects with null prototypes are properly handled would likely be good.

src/workerd/jsg/ser.c++
jasnell: Also, why not just have this defer to `ThrowDataCloneError`?

src/workerd/jsg/ser.h
```diff
@@ -95,7 +115,8 @@ class Serializer final: v8::ValueSerializer::Delegate {
     kj::Array<std::shared_ptr<v8::BackingStore>> transferredArrayBuffers;
   };
 
-  explicit Serializer(Lock& js, kj::Maybe<Options> maybeOptions = kj::none);
+  explicit Serializer(Lock& js): Serializer(js, {}) {}
+  explicit Serializer(Lock& js, Options options);
```
jasnell, line 119: Non-blocking nit: why not `explicit Serializer(Lock& js, Options options = {});` ?

src/workerd/jsg/ser.c++
kentonv: The default implementation of `ExternalHandler::serializeFunction()` produces behavior identical to if you didn't enable treating functions as host objects, so I think a separate option isn't needed. You just don't override the method if you don't want functions.

src/workerd/jsg/ser.h
kentonv: C++ will not let you specify a default value for a parameter whose type is declared nested within the same class. TBH I never understood why but it seems to be a basic language limitation. In the past I've sometimes solved this by pulling the `Options` struct out of the class, but that's ugly.

jasnell: ah, right, forgot about that

src/workerd/jsg/ser.c++
jasnell: Yeah, I spotted that which is why I definitely don't think this is blocking. I'd just prefer not to engage any of this at all in cases where it's not needed. But, definitely not critical in any way.

src/workerd/jsg/ser.h
```diff
@@ -78,6 +78,26 @@ class Serializer final: v8::ValueSerializer::Delegate {
...
+    // (they throw DataCloneError).
+    //
+    // TODO(someday): Perhaps we could create a framework for application-defined classes to define
+    //   their own serializers. However, we would need to be extremely careful about this when
+    //   deserializing data from a possibly-malicious source. Such serialization frameworks have
+    //   a history of creating security bugs as people declare various classes serializable without
+    //   fully thinking through what an attacker could do by sending them an instance of that class
+    //   when it isn't expected. Probably, we just shouldn't support this over RPC at all. For DO
+    //   storage, it could be OK since the application would only be deserializing objects it wrote
+    //   itself, but it may not be worth it to support for only that use case.
```
irvinebroque, line 98: this comment helped me a lot, thank you

MellowYarker (approved):

src/workerd/jsg/ser.c++
kentonv: > Also, why not just have this defer to ThrowDataCloneError?

I guess because `ThrowDataCloneError()` contains its own try/catch similar to this function, and does not throw JsExceptionThrown. While calling it would work fine in this case, that's only true because it would be the last thing we call before returning, which seems precarious.

src/workerd/jsg/ser.c++
kentonv: Cleaned this up: 0a3a2719

src/workerd/jsg/ser.c++
kentonv: 89d996ab

PR #1757, opened 2024-03-03
JSRPC: Support capnp-over-http for external server bindings

kentonv:
This PR makes it possible to configure incoming sockets and outgoing ExternalServers to allow negotiating a Cap'n Proto connection over HTTP. Miniflare will need this to allow RPC over service bindings to be tested, since miniflare currently implements service bindings by running each worker in a separate workerd instance.

Currently, this works by using an HTTP CONNECT request specifying a magic hostname (hostname chosen in the config). In theory it might be better to use `Upgrade`, but that would require extensions to KJ HTTP. It could also make sense to use a WebSocket endpoint, but that would have been more work. We can support those as alternatives in the future if there is demand.

I also went ahead and used this new code path to enhance js-rpc-test, so that it actually tests running over a real socket. This actually caught a bug!

### DIFF
--- a/src/workerd/api/tests/js-rpc-test.js
+++ b/src/workerd/api/tests/js-rpc-test.js
@@ -309,8 +309,8 @@ export let namedServiceBinding = {
     assert.strictEqual(sawFinally, true);
 
     // `fetch()` is special, the params get converted into a Request.
-    let resp = await env.MyService.fetch("http://foo", {method: "POST"});
-    assert.strictEqual(await resp.text(), "method = POST, url = http://foo");
+    let resp = await env.MyService.fetch("http://foo/", {method: "POST"});
+    assert.strictEqual(await resp.text(), "method = POST, url = http://foo/");
 
     await assert.rejects(() => env.MyService.instanceMethod(), {
       name: "TypeError",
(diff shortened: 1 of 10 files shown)


src/workerd/api/tests/js-rpc-test.js
```diff
@@ -246,8 +246,8 @@ export let namedServiceBinding = {
     assert.strictEqual(sawFinally, true);
 
     // `fetch()` is special, the params get converted into a Request.
-    let resp = await env.MyService.fetch("http://foo", {method: "POST"});
-    assert.strictEqual(await resp.text(), "method = POST, url = http://foo");
+    let resp = await env.MyService.fetch("http://foo/", {method: "POST"});
```
jasnell, line 249: Out of curiosity, why is adding the trailing `/` here necessary?

src/workerd/server/workerd.c++
```diff
@@ -486,11 +487,125 @@ kj::Maybe<kj::Own<capnp::SchemaFile>> tryImportBulitin(kj::StringPtr name) {
...
+    return kj::heap<NetworkWithLoopback>(
+        inner.restrictPeers(allow, deny), ioProvider, loopbackEnabled);
+  }
+
+private:
+  kj::Network& inner;
+  kj::Own<kj::Network> ownInner;
+  kj::AsyncIoProvider& ioProvider KJ_UNUSED;
+  bool rootLoopbackEnabled = false;
+  bool& loopbackEnabled;
```
jasnell, line 535: A brief comment indicating who owns this `bool&` would be helpful.

jasnell (approved):

src/workerd/api/tests/js-rpc-test.wd-test
```diff
@@ -26,5 +27,18 @@ const unitTests :Workerd.Config = (
...
+        http = (capnpConnectHost = "cappy")
+      )
+    )
+  ],
+  sockets = [
+    ( name = "loop",
+      address = "loopback:loop",
+      service = (name = "js-rpc-test", entrypoint = "MyService"),
+      http = (capnpConnectHost = "cappy")
+    )
```
mrbbot, line 42: This should work well for us! Thanks! 😃

mrbbot (approved):

src/workerd/api/tests/js-rpc-test.js
kentonv: When sending the request over actual HTTP, the trailing `/` gets added if it is missing. So in order to make the test consistent between when it runs over a real socket vs. local service binding, I had to add the `/`.

src/workerd/server/workerd.c++
kentonv: 6719d3db

kentonv:
Changed this to be based on #1756 so I don't have to rebase again later...

src/workerd/api/tests/js-rpc-test.js
jasnell: hmm, ok. Per fetch spec the trailing slash is supposed to be implicit in this case so it technically should be fine to pass in without it (since the `url` arg here is technically supposed to be passed through the whatwg url parser if we're following spec)... so it just feels a bit odd to have to modify the test to make it work correctly. But I think this is fine for now. I do think at some point we need to revisit how the input url is handled on fetch calls tho.

kentonv: The problem is if I don't add the `/` here, then the assert on the next line fails either when using real HTTP (if it doesn't expect a `/`) or when not using real HTTP (if it does expect a `/`). In order to be able to write a single assert which works in both cases, I had to add the `/` on this line.

Arguably this is a bug in service bindings, that they do not implicitly add the `/` as regular HTTP would, but that bug has been around a while, not introduced in this PR.

PR #1758, opened 2024-03-03
Fix stack traces for native async functions that throw.

kentonv:
I can't believe this was sitting under our nose this entire time!

Until now, when a native function returned a Promise that eventually rejected, the rejection Error would have an empty stack trace, making it very hard to figure out where the problem came from.

For years, we've known this is a big problem, but I thought it would be much harder to solve. Something like four years ago I was trying to figure out how to make async stack traces work, and I learned that native async stack traces require using the inspector API. But, we don't enable the inspector in production; only when debugging. So it seemed like we couldn't use this.

I'd learned this from a former V8 team member, but that person probably wasn't aware than in 2019 (after he had left) V8 added a new built-in async tracing mechanism, documented here:

https://docs.google.com/document/d/13Sy_kBIJGP0XT34V1CV3nkWya4TwYx9L3Yv45LdGB6Q/edit

This mechanism will automatically add an async trace to errors _as long as_ the Error is constructed inside a JavaScript promise continuation.

The problem for us is, native async code doesn't necessarily run in JS promise continuations. It runs in KJ promises. We were reporting errors by creating a resolver at the beginning of the operation, and then calling `resolver.reject()` at the end. `reject()` would convert the KJ exception into a JS error, but it was not being called from a JS promise continuation.

The solution is to have the resolver be for a value that is `OneOf<Result, kj::Exception>`. Then, add a `.then()` to the promise which pulls out the `kj::Exception` and converts it into a rejection. This way, the Error is constructed inside a promise continuation, and the tracing mechanism works.

### DIFF
--- a/src/workerd/api/tests/js-rpc-test.js
+++ b/src/workerd/api/tests/js-rpc-test.js
@@ -483,3 +483,20 @@ export let crossContextSharingDoesntWork = {
     await assert.rejects(() => env.MyService.tryUseGlobalRpcPromisePipeline(), expectedError);
   },
 }
+
+// Test that exceptions thrown from async native functions have a proper stack trace. (This is
+// not specific to RPC but RPC is a convenient place to test it since we can easily define the
+// callee to throw an exception.)
+//
+// Note that it's only a *local* stack trace of the client-side stack leading up to the call. The
+// stack on the server side is not, at present, transmitted to the client.
+export let testAsyncStackTrace = {
+  async test(controller, env, ctx) {
+    try {
+      await env.MyService.throwingMethod();
+    } catch (e) {
+      // verify stack trace was produced
+      assert.strictEqual(e.stack.includes("at async Object.test"), true);
+    }
+  }
+}
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -1009,14 +1009,14 @@ jsg::Promise<IoContext::MaybeIoOwn<addIoOwn, T>> IoContext::awaitIoImpl(
   auto& lock = getCurrentLock();
   jsg::Lock& js = lock;
 
-  auto [ jsPromise, resolver ] = js.newPromiseAndResolver<Result>();
-
   // It is necessary for us to grab a reference to the jsg::AsyncContextFrame here
   // and pass it into the then(). If the promise is rejected, and there is no rejection
   // handler attached to it, an unhandledrejection event will be scheduled, and scheduling
   // that event needs to be done within the appropriate frame to propagate the correct context.
 
   if constexpr (jsg::isVoid<T>()) {
+    auto [ jsPromise, resolver ] = js.newPromiseAndResolver<kj::Maybe<kj::Exception>>();
+
     addTask(promise.then([]() -> kj::Maybe<kj::Exception> {
       return kj::none;
     }, [](kj::Exception&& exception) -> kj::Maybe<kj::Exception> {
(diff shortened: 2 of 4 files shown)


src/workerd/api/tests/js-rpc-test.js
```diff
@@ -483,3 +483,17 @@ export let crossContextSharingDoesntWork = {
     await assert.rejects(() => env.MyService.tryUseGlobalRpcPromisePipeline(), expectedError);
   },
 }
+
+// Test that exceptions thrown from async native functions have a proper stack trace. (This is
+// not specific to RPC but RPC is a convenient place to test it since we can easily define the
+// callee to throw an exception.)
```
jasnell, line 489: A comment here clearly indicating that this is the local stack would be helpful. Just as I briefly made the mistake thinking that the stack would include the remote frames, it would be good to set expectations for folks coming into the code later.

src/workerd/io/io-context.h
```diff
@@ -1029,15 +1029,20 @@ jsg::Promise<IoContext::MaybeIoOwn<addIoOwn, T>> IoContext::awaitIoImpl(
                   maybeException = kj::mv(maybeException),
                   maybeAsyncContext = kj::mv(maybeAsyncContext)](Worker::Lock& lock) mutable {
         jsg::AsyncContextFrame::Scope asyncScope(lock, maybeAsyncContext);
-        KJ_IF_SOME(exception, maybeException) {
-          resolver.reject(lock, kj::mv(exception));
-        } else {
-          resolver.resolve(lock);
-        }
+        resolver.resolve(kj::mv(maybeException));
```
jasnell, line 1032: A comment here explaining this would be helpful for folks coming into this later.

src/workerd/io/io-context.h
```diff
@@ -1029,15 +1029,20 @@ jsg::Promise<IoContext::MaybeIoOwn<addIoOwn, T>> IoContext::awaitIoImpl(
...
     }));
+
+    return jsPromise.then(js, [](jsg::Lock& js, kj::Maybe<kj::Exception>&& maybeEx) -> Result {
+      KJ_IF_SOME(e, maybeEx) {
+        js.throwException(kj::mv(e));
+      }
+    });
   } else {
     typedef kj::OneOf<T, kj::Exception> ResultOrException;
+    typedef kj::OneOf<Result, kj::Exception> IoResultOrException;
```
jasnell, line 1043: non-blocking nit... naming here (`ResultOrException` and `IoResultOrException`) is a bit unfortunate, but no better suggestions.

src/workerd/io/io-context.h
```diff
@@ -1059,14 +1064,24 @@ jsg::Promise<IoContext::MaybeIoOwn<addIoOwn, T>> IoContext::awaitIoImpl(
             }
           }
           KJ_CASE_ONEOF(exception, kj::Exception) {
-            resolver.reject(lock, kj::mv(exception));
+            resolver.resolve(kj::mv(exception));
```
jasnell, line 1067: Here, as well, a comment explaining why we're resolving with an exception would be helpful for future generations.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -483,3 +483,17 @@ export let crossContextSharingDoesntWork = {
...
+export let testAsyncStackTrace = {
+  async test(controller, env, ctx) {
+    try {
+      await env.MyService.throwingMethod();
```
jasnell, line 493: Non-blocking: Not sure if this is tested elsewhere, but a test verifying the behavior when the remote throws an arbitrary JS value that is not an error object, and making sure it comes through here correctly would be also be good.

kentonv: I believe non-error values get stringified and wrapped in `Error`. Of course, this behavior long predates RPC, and is also not really relevant to this particular PR.

kentonv:
Ugh my fixup script barfed again.

kentonv:
Here was the fixup (adding comments): b2d6bf6f

jasnell (approved):

kentonv:
Ugh, can't merge this yet because a bunch of tests in the internal codebase are string-matching the full `error.stack` and now they fail. :facepalm:

Issue #1760, opened 2024-03-04
🐛 Bug Report — Runtime APIs - FTS5 authorization needed


rozenmd:
According to https://github.com/cloudflare/workerd/pull/619/files#diff-3009a3da17d76a94673aa828ce6c364fc24e34d889b4995adfc249a0e2a3c754 we're looking for the exact string `fts5` when working out if a command is allow-listed. 

It should probably be case-insensitive?

geelen:
Should be as simple as making this comparison case-insensitive: https://github.com/cloudflare/workerd/blob/d8ffd3de075ad5dba5b5d072165c9e1830586695/src/workerd/util/sqlite.c++#L658

I'll jump in and fix it next time I'm in there, unless someone gets to it first.

PR #1762, opened 2024-03-04
Make SQLite VTable module name comparison case-insensitive

KianNH:
Fixes https://github.com/cloudflare/workerd/issues/1760


rozenmd (commented):
I don't know C++ but seems legit to me

KianNH:
Windows tests are failing due to a 500 from the remote cache from the looks of it, but I suspect it's since we're only adding a define for `strncasecmp` -> `_strnicmp` and not `strcasecmp` -> `_stricmp`

kentonv (approved):
Not sure why Mac build failed, I'm re-running it.

Whoever merges this: Don't forget to choose squash-merge.

PR #1764, opened 2024-03-04
Add dispatch metrics

Warfields:
Allows RequestObservers to receive metrics on dynamic dispatch.j

### DIFF
--- a/src/workerd/io/observer.h
+++ b/src/workerd/io/observer.h
@@ -62,6 +62,9 @@ public:
     return kj::mv(client);
   }
 
+  // Used to record when a worker has a dynamic dispatch binding (Called on the dispatching side)
+  virtual void setHasDispatchBinding() {};
+
   virtual SpanParent getSpan() { return nullptr; }
 
   virtual void addedContextTask() {}


src/workerd/io/observer.h
```diff
@@ -62,6 +62,12 @@ class RequestObserver: public kj::Refcounted {
     return kj::mv(client);
   }
 
+  // Used to record dynamic dispatch workers.
```
jasnell, line 65: This comment is a bit vague. Would this be called on the side doing the dispatching? Or the target side?

src/workerd/io/observer.h
```diff
@@ -62,6 +62,12 @@ class RequestObserver: public kj::Refcounted {
     return kj::mv(client);
   }
 
+  // Used to record when a worker has a dynamic dispatch binding (Called on the dispatching side)
+  virtual void setHasDispatchBinding() {};
+
+  // Used to record what worker dispatched the current worker (Called by the dispatch target).
+  virtual void setDispatcherId(kj::Maybe<kj::String> id) {};
```
jasnell, line 69: non-blocking nit: does it need to be `kj::String` or can it be `kj::StringPtr`?

jasnell (approved):

src/workerd/io/observer.h
Warfields: It did need to be a `kj::String`, the object that it's coming from does not outlive the metrics object and is dropped before the worker starts executing.

PR #1765, opened 2024-03-05
Refactor: awaitIo() captures callback in KJ promise, not JSG.

kentonv:
Previously, awaitIoImpl() would return a jsg::Promise. If awaitIo() had been called with a continuation callback, then this was appended to the promise chain by calling `.then()` on the JSG promise.

This required a lot of hackery. Many such callbacks capture KJ I/O types, so the callback had to be wrapped in an IoOwn. Similarly, the *parameter* to the callback could contain KJ I/O types, so had to be wrapped in IoOwn. Also lots of JavaScript allocation and V8 API back-and-forth is needed to schedule JSG promises like this. Finally, it could become a bit difficult to reason about whether the callback would actually outlive the KJ promise -- but many call sites require this.

Instead, we can pass the callback down and capture it via the KJ promise chain, so it never has to be "owned" by JSG and we don't need any implicit IoOwns. This also makes it very clear that the callback outlives the KJ promise, as it's directly part of the KJ promise chain.

### DIFF
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -757,12 +757,6 @@ private:
 
   void runFinalizers(Worker::AsyncLock& asyncLock);
 
-  template <bool addIoOwn, typename T> struct MaybeIoOwn_;
-  template <typename T> struct MaybeIoOwn_<false, T> { typedef T Type; };
-  template <typename T> struct MaybeIoOwn_<true, T> { typedef IoOwn<T> Type; };
-  template <bool addIoOwn, typename T>
-  using MaybeIoOwn = typename MaybeIoOwn_<addIoOwn, T>::Type;
-
   template <typename T>
   struct IdentityFunc {
     inline T operator()(jsg::Lock&, T&& value) const {
@@ -788,15 +782,6 @@ private:
   template <typename T, typename InputLockOrMaybeCriticalSection, typename Func>
   jsg::PromiseForResult<Func, T, true> awaitIoImpl(
       jsg::Lock& js, kj::Promise<T> promise, InputLockOrMaybeCriticalSection ilOrCs, Func&& func);
-
-  // Same as addFunctor() but expects that the returned function will receive a IoOwn<T> as the
-  // parameter, and instead passes the underlying `T` as the parameter to `func`. This is needed
-  // in the implementation of awaitIo().
-  //
-  // If `passLock` is true, then `func()`'s first parameter is `jsg::Lock&`, and it's the second
-  // parameter that needs to be wrapped.
-  template <typename T, bool passLock, typename Func>
-  auto addFunctorIoOwnParam(Func&& func);
 
   // The IncomingRequest that is currently considered "current". This is always the
   // latest-starting request that hasn't yet completed.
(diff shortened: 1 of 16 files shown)


src/workerd/io/io-context.h
```diff
@@ -901,200 +892,166 @@ kj::PromiseForResult<Func, Worker::Lock&> IoContext::run(
...
-        jsg::AsyncContextFrame::Scope asyncScope(lock, maybeAsyncContext);
-        KJ_SWITCH_ONEOF(resultOrException) {
-          KJ_CASE_ONEOF(result, T) {
-            if constexpr (addIoOwn) {
-              resolver.resolve(addObject(kj::heap(kj::mv(result))));
-            } else {
-              (void)this; // used only in one branch of constexpr expression
-              resolver.resolve(kj::mv(result));
-            }
+  auto [ jsPromise, resolver ] = js.newPromiseAndResolver<ExceptionOr<Result>>();
```
jasnell, line 945: There is a possibility that `newPromiseAndResolver(...)` here could throw `JsExceptionThrown` before the `func` is attached to anything, Is that ok here? It should only ever happen in fatal cases, of course, but if part of the intent is to ensure that `func` outlives `promise`, then I think maybe this might be a condition where it may not?

kentonv: Technically at this point, both `func` and `promise` are on the stack, but since `func` was passed in by rvalue reference whereas `promise` was by value, `promise`'s destructor will run first. But I that is rather... precarious. I should make it more explicit.

src/workerd/io/io-context.h
```diff
@@ -773,24 +757,31 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
...
+  struct IdentityFunc<void> {
+    inline void operator()(jsg::Lock&) const {}
+  };
 
-  template <bool addIoOwn, typename T, typename InputLockOrMaybeCriticalSection>
-  jsg::Promise<MaybeIoOwn<addIoOwn, T>> awaitIoImpl(
-      kj::Promise<T> promise, InputLockOrMaybeCriticalSection ilOrCs);
+  template <typename T>
+  struct ExceptionOr_ {
+    using Type = kj::OneOf<T, kj::Exception>;
```
jasnell, line 773: Does the `T` here need to be `typename T` or something? Just looks a bit off. Might run into some compile errors in CI I think.

kentonv: There is no ambiguity here about whether `T` is a type, since it is a direct template parameter and was declared as `typename` there. So there's no need to add `typename` on this line.

kentonv: The `using` declaration a few lines later does appear to need `typename` though. Weird that it works locally.

jasnell: heh, I was close it seems lol ..

```
bazel-out/k8-fastbuild/bin/src/workerd/io/_virtual_includes/io/workerd/io/io-context.h:780:23: error: missing 'typename' prior to dependent type name 'ExceptionOr_<T>::Type'
  using ExceptionOr = ExceptionOr_<T>::Type;
                      ^~~~~~~~~~~~~~~~~~~~~
                      typename 
```

src/workerd/io/io-context.h
```diff
@@ -901,200 +892,166 @@ kj::PromiseForResult<Func, Worker::Lock&> IoContext::run(
...
         }
-      }, kj::mv(ilOrCs));
-    }));
+      }
+    }, kj::mv(ilOrCs));
+  }));
 
-    return jsPromise.then(js, [](jsg::Lock& js, IoResultOrException&& resultOrException) -> Result {
-      KJ_SWITCH_ONEOF(resultOrException) {
+  return jsPromise.then(js, [](jsg::Lock& js, ExceptionOr<Result>&& exceptionOrResult) -> Result {
```
jasnell, line 1033: non-blocking: it might be worthwhile adding a short comment here reminding folks that the `then` call here could throw a `JsExceptionThrown` but that it's ok if it does since `func` is already attached. Completely optional tho ... I just want to see about taking opportunities to remind folks that the `jsg::Promise` `then`/`catch_` calls can throw and need to handled appropriately.

src/workerd/io/io-context.h
kentonv: Fixed: 07f8cb10

src/workerd/io/io-context.h
kentonv: Decided to just comment on it instead of change it: 07f8cb10

Also moved `newPromiseAndResolver` down, though, since it wasn't needed yet.

src/workerd/io/io-context.h
kentonv: 07f8cb10

jasnell (approved):

kentonv:
I guess I missed stuff in GPU land: 074adf18

PR #1759, opened 2024-03-03
fix: CustomEvent generic type

Cherry:
Fixes https://github.com/cloudflare/workerd/issues/1504

This overrides the `CustomEvent` types to better match other implementations.

This is my first time contributing to `workerd`, and really using JSG at all, so apologies if there's anything weird with this change. If there's a better way to do this, please just let me know.

Old output:
```typescript
export declare class CustomEvent extends Event {
  constructor(type: string, init?: CustomEventCustomEventInit);
  get detail(): any | undefined;
}
```

New output:
```typescript
export declare class CustomEvent<T = any> extends Event {
  constructor(type: string, init?: CustomEventCustomEventInit);
  get detail(): T;
}
```

TypeScript `lib.dom.d.ts` reference: https://github.com/microsoft/TypeScript/blob/1d6d962d3132a901f5fb48be4a389c45faf5a74e/src/lib/dom.generated.d.ts#L5935

PR #1766, opened 2024-03-05
chore: simplify CustomEvent type override

Cherry:
This simplifies this type override as reported by @mrbbot in https://github.com/cloudflare/workerd/pull/1759#discussion_r1513493487.

It produces the same output:
```typescript
export class CustomEvent<T = any> extends Event {
  constructor(type: string, init?: CustomEventCustomEventInit);
  get detail(): T;
}
```


jasnell (approved):

mrbbot (approved):
Thanks for updating this! 😃

PR #1767, opened 2024-03-05
Initialize Pyodide in the `StartupJs` limitEnforcer

hoodmane:
We're concerned that free tier workers might time out due to very slow cold start times. We have various plans to improve cold starts which should ideally help a lot. But in the meantime, it also doesn't make sense to count runtime startup time against the workers.

In the future we may want a separate StartupPython limit enforcer.

Because the Python startup code is async, we give ownership of the limitEnforcer RAII guard to a closure that we run when the startup code resolves. No actual user code can be executed while we hold the guard since the first thing that the Python entrypoint code does is await for the completion of the setup. But whatever other startup code will also be run under the StartupJs LimitEnforcer.


src/workerd/io/worker-entrypoint.c++
```diff
@@ -213,6 +213,63 @@ void WorkerEntrypoint::init(
...
+ * Ideally we should be able to wrap a JS task so that only it is executed with a different limiter,
+ * which is switched on when entering the task and back off when exiting. But I don't know how to do
+ * that so maybe this works okay for now.
+ */
+void
+initPythonIfNeeded(kj::Maybe<kj::StringPtr>& entrypointName, kj::Own<workerd::IoContext_IncomingRequest>& incomingRequest) {
+  auto& context = incomingRequest->getContext();
+  auto& script = context.getWorker().getScript();
+  if (!script.isPython) {
+    return;
```
dom96, line 233: Might be good to throw here

src/workerd/io/worker-entrypoint.c++
```diff
@@ -213,6 +213,63 @@ void WorkerEntrypoint::init(
...
+  }
+  context.addWaitUntil(context.run(
+      [entrypointName=entrypointName, &context, &metrics = incomingRequest->getMetrics()]
+      (Worker::Lock& lock) mutable -> kj::Promise<void> {
+    kj::Maybe<kj::Exception> maybeLimitError;
+
+    // Raise limit
```
dom96, line 240: How does the below raise the limit? It just seems to create a limit enforcer.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -213,6 +213,63 @@ void WorkerEntrypoint::init(
...
+ */
+void
+initPythonIfNeeded(kj::Maybe<kj::StringPtr>& entrypointName, kj::Own<workerd::IoContext_IncomingRequest>& incomingRequest) {
+  auto& context = incomingRequest->getContext();
+  auto& script = context.getWorker().getScript();
+  if (!script.isPython) {
+    return;
+  }
+  context.addWaitUntil(context.run(
+      [entrypointName=entrypointName, &context, &metrics = incomingRequest->getMetrics()]
```
dom96, line 236: Are we certain the `StringPtr` will stay alive for the duration of this? Might be best to copy it just in case (you can do so via `kj::str()`)

src/workerd/io/worker-entrypoint.c++
```diff
@@ -213,6 +213,63 @@ void WorkerEntrypoint::init(
...
+    auto& isolate = Worker::Isolate::from(lock);
+    kj::Own<void> limitScope = isolate.getLimitEnforcer().enterStartupJs(lock, maybeLimitError);
+
+    // call initPython
+    auto handler = KJ_REQUIRE_NONNULL(lock.getExportedHandler(entrypointName, context.getActor()), "Failed to get worker handler.");
+    jsg::Promise<void> res = nullptr;
+    KJ_IF_SOME(initPython, handler->initPython) {
+      res = initPython(lock, handler->getCtx());
+    } else {
+      KJ_ASSERT(false, "Internal error, expected to see initPython on handler.");
```
dom96, line 250:
```suggestion
      KJ_FAIL_REQUIRE("Internal error, expected to see initPython on handler.");
```

src/workerd/io/worker-entrypoint.c++
```diff
@@ -213,6 +213,63 @@ void WorkerEntrypoint::init(
...
+ * Setup Python with a more generous limit enforcer, currently enterStartupJs.
+ *
+ * The Pythons setup is async so we hold the limitScope until the setup work is done. Whatever else
+ * happens between now and when we finish initPython will be done with the higher limit. User code
+ * cannot be entered with the higher limit because the first thing the main entrypoints do is wait
+ * on this initialization work being completed.
+ *
+ * Ideally we should be able to wrap a JS task so that only it is executed with a different limiter,
+ * which is switched on when entering the task and back off when exiting. But I don't know how to do
+ * that so maybe this works okay for now.
```
dom96, line 226: Might be worth asking in our chat if this is possible.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -213,6 +213,63 @@ void WorkerEntrypoint::init(
...
+
+    // Goal: attach a callback to the promise so that limitScope isn't dropped until the callback
+    // returns. We need the owner of limitScope to be an IoOwn since we seem to need the lock to
+    // drop limitScope.
+    //
+    // Unfortunately the type of limitScope is kj::Own<void> and IoContext::addObject does not work
+    // on a kj::Own<void>. Work around this by making a LimitScopeHolder class that holds the
+    // limitScope and then attach the limitScopeHolder to the IoContext.
+    class LimitScopeHolder {
+      public:
```
dom96, line 261: Nit:
```suggestion
    public:
```

dom96 (commented):
LGTM but I'm not super familiar with the limit enforcer personally. Ideally I want another pair of eyes on this.

@jasnell can you have a look?

src/workerd/io/worker-entrypoint.c++
hoodmane: I think the return guard of `enterStartupJs` is an RAII guard that maintains a higher limit while we hold it. Is this wrong?

src/workerd/io/worker-entrypoint.c++
hoodmane: We call this function unconditionally hence the `IfNeeded` in `initPythonIfNeeded`. If it's **not** needed, we return without doing anything. If we had to move these 5 lines of code into each of the 5 call sites it would be an extra (5-1)*5 = 20 lines of boilerplate. In the interest of minimizing code duplication I figured it's better to do the check in here.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -213,6 +213,63 @@ void WorkerEntrypoint::init(
...
+    //
+    // Unfortunately the type of limitScope is kj::Own<void> and IoContext::addObject does not work
+    // on a kj::Own<void>. Work around this by making a LimitScopeHolder class that holds the
+    // limitScope and then attach the limitScopeHolder to the IoContext.
+    class LimitScopeHolder {
+    public:
+      kj::Own<void> limitScope;
+      LimitScopeHolder(kj::Own<void> limitScope) : limitScope(kj::mv(limitScope)) {}
+    };
+    auto limitScopeHolder = IoContext::current().addObject(kj::heap<LimitScopeHolder>(kj::mv(limitScope)));
```
kentonv, line 265: Regardless of whether overlapping CPU limits work, extending the CPU limit's lifetime beyond the scope of a (synchronous) function call is also going to create trouble. We can't be returning to the event loop with a CPU limit still in force.

kentonv (changes requested):
As discussed, we really should move init to happen at actual worker init time, not on first request.

Additionally, I don't think what you're doing here will actually work with our CPU limiter implementation. It's not designed to support this kind of usage and will likely lead to unexpected behavior.

PR #1768, opened 2024-03-05
Simplify PumpToReader, eliminate AllReaderBase

jasnell:
Further necessary simplification of JS-backed readable stream.

Like the previous change with `AllReader`, this modifies `PumpToReader` so that we do not move the underlying `kj::Own<ValueReadable>` or `kj::Own<ByteReadable>` out of the stream into the `PumpToReader`. This allows us to remove `AllReaderBase` and eliminate the buggy logic around changing the ownership of the `ValueReadable` and `ByteReadable`.

This allows us to further simplify other bits in`ReaddableStreamJsController` while we're at it.

### DIFF
--- a/src/workerd/api/streams/standard.c++
+++ b/src/workerd/api/streams/standard.c++
@@ -664,8 +664,6 @@ public:
   explicit ReadableStreamJsController(StreamStates::Errored errored);
   explicit ReadableStreamJsController(jsg::Lock& js, ValueReadable& consumer);
   explicit ReadableStreamJsController(jsg::Lock& js, ByteReadable& consumer);
-  explicit ReadableStreamJsController(kj::Own<ValueReadable> consumer);
-  explicit ReadableStreamJsController(kj::Own<ByteReadable> consumer);
 
   jsg::Ref<ReadableStream> addRef() override;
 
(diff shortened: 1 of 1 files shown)


src/workerd/api/streams/standard.c++
```diff
@@ -2918,57 +2847,41 @@ private:
...
-  PumpToReader(Readable readable, kj::Own<WritableStreamSink> sink, bool end)
+  PumpToReader(jsg::Ref<ReadableStream> stream, kj::Own<WritableStreamSink> sink, bool end)
     : ioContext(IoContext::current()),
-      state(kj::mv(readable)),
-      sink(kj::mv(sink)),
-      end(end) {
-    state.template get<Readable>()->setOwner(addWeakRef());
-  }
+      state(kj::mv(stream)),
+      sink(ioContext.addObject(kj::mv(sink))),
```
fhanau, line 2855: Can you explain what addObject being added here does?

src/workerd/api/streams/standard.c++
```diff
@@ -2918,57 +2847,41 @@ private:
   }
 };
 
-template <typename T>
-class PumpToReader final: public AllReaderBase {
+class PumpToReader {
```
fhanau, line 2850: Nit: PumpToReader should still be final.

src/workerd/api/streams/standard.c++
```diff
@@ -3333,6 +3236,18 @@ kj::Promise<DeferredProxy<void>> ReadableStreamJsController::pumpTo(
...
+  // This operation will leave the ReadableStream locked and disturbed. It will consume
+  // the stream until it either closed or errors. If the deferred proxy promise or it's
+  // inner promise is dropped, the PumpToReader (and sink) will be dropped and the stream
+  // will be canceled. If the PumpToReader is dropped while there is a pending write on
+  // the sink, the pending write will be canceled.
+
+  const auto handlePump = [&] {
+    KJ_ASSERT(lock.lock());
+    auto reader = kj::heap<PumpToReader>(addRef(), kj::mv(sink), end);
+    return addNoopDeferredProxy(reader->pumpTo(js).attach(kj::mv(reader)));
```
fhanau, line 3248: If I interpret the comment for handlePump() correctly the doClose() can be dropped here becauce the stream is being closed in pumpTo(), right?

fhanau (approved):
Glad we're doing this. I haven't worked with the standard streams before so it's entirely possible I overlooked some subtle details here, but the change makes sense to me as a whole.

src/workerd/api/streams/standard.c++
jasnell: Yes, that's it exactly

src/workerd/api/streams/standard.c++
jasnell: The `sink` is a kj heap i/o object that ends up getting passed into a v8 heap object, which means it needs to be wrapped in an `IoOwn` to be used safely. That's what `addObject` does

PR #1770, opened 2024-03-05
Remove Fetcher methods get(), put(), delete() with compat flag.

kentonv:
These were intended to be convenient wrappers around `fetch()` that perform the respective HTTP method. It was a bad idea (my fault). Luckily it was never documented and probably few people depend on it.

Note that a long time ago, KV namespace bindings were Fetchers rather than being the specific KvNamespace type. This had lots of problems, but sort of worked as long as your key names were URL-safe. It's possible very old scripts are in production that still depend on these old-style bindings, but they won't be broken since they obviously have ancient compat dates as well. (In fact... they predate compat dates...)

### DIFF
--- a/src/workerd/api/tests/js-rpc-flag.js
+++ b/src/workerd/api/tests/js-rpc-flag.js
@@ -12,6 +12,10 @@ export class DurableObjectExample {
   foo() {
     return 123;
   }
+
+  fetch(req) {
+    return new Response(req.method + " " + req.url);
+  }
 }
 
 export default {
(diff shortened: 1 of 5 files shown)


src/workerd/api/tests/js-rpc-flag.js
```diff
@@ -12,13 +12,21 @@ export class DurableObjectExample {
   foo() {
     return 123;
   }
+
+  fetch(req) {
+    return new Response(req.method + " " + req.url);
```
jasnell, line 17:
```suggestion
    return new Response(`${req.method} ${req.url}`);
```

jasnell (approved):

src/workerd/api/tests/js-rpc-flag.js
kentonv: Agreed that's an improvement. Don't think it's worth restarting CI, though.

irvinebroque:
This is actually a pretty cool idea and TIL it existed! Very similar to Pages Functions.

kentonv:
The "linux" build seems to have hung on mksnapshot, but "linux-debug" passed and all internal builds passed, and GitHub apparently cannot cancel the hung workflow, it's just gonna run until it times out or something. I'm just going to merge.

PR #1773, opened 2024-03-06
Update queues debug logging to ignore aborted requests

jbwcloudflare:
WIP

- Ignore "aborted" requests and only log info on timeouts
- TODO: Add more contextual info (user account ID)

CC: @a-robinson

### DIFF
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -127,8 +127,8 @@ public:
   kj::Promise<void> drain();
 
   // Waits for all "waitUntil" tasks to finish, up to the time limit for scheduled events, as
-  // defined by `scheduledTimeoutMs` in `WorkerLimits`. Returns a bool indicating `true` if the
-  // event completed successfully, or `false`, if it was canceled early.
+  // defined by `scheduledTimeoutMs` in `WorkerLimits`. Returns an enum indicating if the
+  // event completed successfully, hit a timeout, or was aborted.
   //
   // Note that, while this is similar in some ways to `drain()`, `finishScheduled()` is intended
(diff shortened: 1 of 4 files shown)


src/workerd/io/io-context.h
```diff
@@ -91,7 +91,8 @@ class IoContext_IncomingRequest {
   // This method is also used by some custom event handlers (see WorkerInterface::CustomEvent) that
   // need similar behavior, as well as the test handler. TODO(cleanup): Rename to something more
   // generic?
-  kj::Promise<bool> finishScheduled();
+  enum class FinishScheduledResult { COMPLETED, ABORTED, TIMEOUT };
+  kj::Promise<FinishScheduledResult> finishScheduled();
```
a-robinson, line 95: Changing the return value here will also need a parallel change to the internal code since there are a couple callers of this method.

a-robinson (commented):
As mentioned in chat I believe we can get the script ID using `incomingRequest->getContext().getWorker().getScript().getId()`, but I'm not seeing an existing way to get the owner ID other than either plumbing it into the `QueueEvent` or returning the error back to the caller in some way for it to do the logging.

a-robinson:
> I'm not seeing an existing way to get the owner ID 

To be clear, hopefully I've just missed something, but it doesn't look promising.

src/workerd/io/io-context.h
jbwcloudflare: 👍 I've got that all wired up, just haven't put it up for PR yet.

a-robinson:
Two force pushes:
1. Rebase onto main
2. Add scriptId to the log (tested locally to work as intended)

a-robinson (approved):

PR #1776, opened 2024-03-07
Disable Windows-debug CI job

fhanau:


jasnell (approved):

fhanau:
@jasnell friendly note – when I mark a PR `[EXP]`, `[WIP]` or `[RFC]`, I make no guarantees about its correctness or behavior. It may or may not do what the description says, and it may turn Earth into a Black Hole if used incorrectly. In this case, it turns out that while `-gline-tables-only` is supported by clang-cl, it causes it to emit DWARF-based debug information instead of the CodeView format generally used on Windows, which probably isn't what we want.

fhanau:
> @jasnell friendly note – when I mark a PR `[EXP]`, `[WIP]` or `[RFC]`, I make no guarantees about its correctness or behavior. It may or may not do what the description says, and it may turn Earth into a Black Hole if used incorrectly. In this case, it turns out that while `-gline-tables-only` is supported by clang-cl, it causes it to emit DWARF-based debug information instead of the CodeView format generally used on Windows, which probably isn't what we want.

I ended up redoing this PR to disable Windows-debug until the underlying issue in bazel has been addressed, finally created a [tracking issue](https://github.com/bazelbuild/bazel/issues/21615) and [PR](https://github.com/bazelbuild/bazel/pull/21616) for that.

mrbbot (approved):

PR #1779, opened 2024-03-07
Simplify ValueReadable/ByteReadable

jasnell:
Continuing on with necessary simplifications of the js-backed streams implementation. This carries on from the previous `AllReader` and `PumpToReader` change to further simplify


fhanau (approved):
LGTM.

PR #1780, opened 2024-03-07
Set up Pyodide in the top level scope

hoodmane:
This ensures that Pyodide instantiation and package loading happens inside of the enterStartupJs limitEnforcer. In workerd, we can't yet load packages outside of an IoContext because we make actual requests to load the packages. So in workerd, only run loadPyodide at top level and run setupPackages at first request. In the internal runtime, we can do the whole thing at top level.

As a side effect, we have to disable hash randomization because normally the hash comes from `crypto.getRandomValues` but this crashes at top level. Likewise, Python initializes the `random` module's random seed using `crypto.getRandomValues`. We replace `crypto.getRandomValues` with a function that tries to call `crypto.getRandomValues` but if it throws an Error saying `"Disallowed operation called within global scope"` then we fall back to doing nothing. We reseed the rng in the first request context. Note however that hash randomization can only happen at startup, if we don't have access to crypographic rng at startup we can't enable hash randomization later.

### DIFF
--- a/src/pyodide/internal/builtin_wrappers.js
+++ b/src/pyodide/internal/builtin_wrappers.js
@@ -19,6 +19,23 @@ export function monotonicDateNow() {
 }
 
 /**
+ * We initialize Python at top level, but it tries to initialize the random seed with
+ * crypto.getRandomValues which will fail at top level. So we don't produce any entropy the first
+ * time around and we reseed the rng in the first request context before executing user code.
+ */
+export function getRandomValues(arr) {
+  try {
+    return crypto.getRandomValues(arr);
+  } catch (e) {
+    if (e.message.includes("Disallowed operation called within global scope")) {
+      // random.seed() can't work at startup. We'll seed again under the request scope.
+      return arr;
+    }
+    throw e;
+  }
+}
+
+/**
  * First check that the callee is what we expect, then use `UnsafeEval` to
  * construct a `WasmModule`.
  *
--- a/src/pyodide/internal/python.js
+++ b/src/pyodide/internal/python.js
@@ -70,39 +70,6 @@ import stdlib from "pyodide-internal:generated/python_stdlib.zip";
 let MEMORY = undefined;
 
 /**
- * This is passed as a preRun hook in EmscriptenSettings, run just before
- * main(). It ensures that the file system includes the stuff that main() needs,
- * most importantly the Python standard library.
- *
- * Put the Python + Pyodide standard libraries into a zip file in the
- * appropriate location /lib/python311.zip . Python will import stuff directly
- * from this zip file using ZipImporter.
- *
- * ZipImporter is quite useful here -- the Python runtime knows how to unpack a
- * bunch of different archive formats but it is not possible to use these until
- * the runtime state is initialized. So ZipImporter breaks this bootstrapping
- * knot for us.
- *
- * We also make an empty home directory and an empty global site-packages
- * directory `/lib/pythonv.vv/site-packages`.
- *
- * This is a simpified version of the `prepareFileSystem` function here:
- * https://github.com/pyodide/pyodide/blob/main/src/js/module.ts
- */
-function prepareFileSystem(Module) {
-  const pymajor = Module._py_version_major();
-  const pyminor = Module._py_version_minor();
-  Module.FS.mkdirTree("/lib");
-  Module.FS.mkdirTree(`/lib/python${pymajor}.${pyminor}/site-packages`);
-  Module.FS.writeFile(
-    `/lib/python${pymajor}${pyminor}.zip`,
-    new Uint8Array(stdlib),
-    { canOwn: true },
-  );
-  Module.FS.mkdir(Module.API.config.env.HOME);
-}
-
-/**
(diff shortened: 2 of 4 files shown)


src/pyodide/internal/builtin_wrappers.js
```diff
@@ -18,6 +18,23 @@ export function monotonicDateNow() {
   return now + lastDelta;
 }
 
+/**
+ * We initialize Python at top level, but it tries to initialize the random seed with
+ * crypto.getRandomValues which will fail at top level. So we don't produce any entropy the first
+ * time around and we reseed the rng in the first request context before executing user code.
+ */
```
hoodmane, line 25: This won't happen if we are restoring from a memory snapshot since it occurs in the call to `main()` and we skip calling `main()` in that case.

src/pyodide/internal/python.js
```diff
@@ -139,7 +146,12 @@ function getEmscriptenSettings(lockfile, indexURL) {
     // jsglobals is used for the js module.
     jsglobals: globalThis,
     // environment variables go here
-    env: { HOME: "/session" },
+    env: {
+      HOME: "/session",
+      // We don't have access to cryptographic rng at startup so we cannot support hash
+      // randomization. Setting `PYTHONHASHSEED` disables it.
+      PYTHONHASHSEED: "111",
```
hoodmane, line 153: Likewise, not necessary if using a memory snapshot.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -109,35 +119,66 @@ function pyimportMainModule(pyodide) {
...
+  await getPyodide();
+} else {
+  // If we're not in workerd, setupPackages doesn't require IO so we can do it all here.
+  await getMainModule();
+}
+
+/**
+ * Have to reseed randomness in the IoContext of the first request since we gave a low quality seed
+ * when it was seeded at top level.
+ */
```
hoodmane, line 156: It's probably a good idea to reseed the randomness for memory snapshots too, otherwise I think everyone gets the same sequence of random values.

garrettgu10 (approved):
LGTM!

src/pyodide/internal/builtin_wrappers.js
```diff
@@ -18,6 +18,23 @@ export function monotonicDateNow() {
...
+export function getRandomValues(arr) {
+  try {
+    return crypto.getRandomValues(arr);
+  } catch (e) {
+    if (e.message.includes("Disallowed operation called within global scope")) {
+      // random.seed() can't work at startup. We'll seed again under the request scope.
+      return arr;
+    }
```
dom96, line 33: This is fragile and will break as soon as someone changes the error message (which presumably would cause Python Workers to break).

Can we do better here without too much work? If not are we sure that our tests will fail reliably when this code breaks (would be good to test this case)?

src/pyodide/internal/python.js
dom96: Does this mean we can avoid the error handling in our custom `getRandomValues`?

src/pyodide/internal/python.js
```diff
@@ -350,7 +362,7 @@ export function mountLib(pyodide, requirements, isWorkerd) {
   sys.destroy();
 }
 
-export async function loadPyodide(context, lockfile, indexURL) {
```
dom96, line 353: This will be used for snapshot upload, I know it's not used but I'd rather you left it in here so I can avoid dealing with merge conflicts later.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -109,35 +119,66 @@ function pyimportMainModule(pyodide) {
...
+  `);
+}
+
+async function preparePython() {
+  const pyodide = await getPyodide();
+  reseedRandom(pyodide);
+  return await getMainModule();
+}
+
+// Error.stackTraceLimit = Infinity;
```
dom96, line 176:
```suggestion

```

dom96 (approved):
> As a side effect, we have to disable hash randomization because normally the hash comes from crypto.getRandomValues but this crashes at top level. Likewise, Python initializes the random module's random seed using crypto.getRandomValues

Do we have anything else we can use at the top-level that would work fine as a random seed? I guess we can't the current time at the top-level either?

Just wondering whether it's possible to avoid the need to reseed later.

Anyway, overall looks good. But check out my comments.

src/pyodide/internal/builtin_wrappers.js
hoodmane: Python will never start without this.

src/pyodide/internal/python.js
hoodmane: It won't be available when calling `loadPyodide` at top level, so I think the snapshot upload will need to go somewhere else.

hoodmane:
Supposedly at top level the time shows 0 so you can't get entropy that way. You can use `Math.random` to get non cryptographic rng. But I think reseeding later with cryptographic rng is the right thing to do. Otoh any source of startup rng would be good for the hash seed. On the other other hand, memory snapshots imply a fixed hash seed anyways so we can't really win there unless we periodically regenerate the memory snapshots. This will have to be a longer term discussion I think.

PR #1782, opened 2024-03-08
Implement EventSource

jasnell:
SHIP-9206

~~This is just a draft proposal at the moment, but it should be functional for folks who want to give it a try.~~ Had a couple of folks ask me about this over the past few months so figured I'd go ahead and see about an actual implementation. The format is simple enough that it really doesn't require bringing in any dependencies.

Implements the web platform standard `EventSource` API (https://developer.mozilla.org/en-US/docs/Web/API/EventSource).

There are three non-standard extensions implemented to the API:

1. Passing in a custom fetcher

```js
const ev = new EventSource('https://...', { fetcher: env.myfetcher });
```

2. Creating an `EventSource` from an existing `ReadableStream`

```js
const ev = EventSource.from(readableStream);
```

3. The `EventSource` implements ERM `Symbol.dispose` (which simply defers to calling `eventsource.close()`

See the sample for an example.

Note that each time `EventStream` connects it counts as a subrequest and is subject to all of the same characteristics/limitations as other subrequests. Concurrent subrequest limits will apply.

~~This PR will need tests before it could land.~~

### DIFF
--- a/samples/eventsource/README.md
+++ b/samples/eventsource/README.md
@@ -1,0 +1,13 @@
+# EventSource Example
+
+To run the example on http://localhost:8080
+
+```sh
+$ ./workerd serve config.capnp
+```
+
+To run using bazel
+
+```sh
+$ bazel run //src/workerd/server:workerd -- serve ~/cloudflare/workerd/samples/eventsource/config.capnp
+```
--- a/samples/eventsource/config.capnp
+++ b/samples/eventsource/config.capnp
@@ -1,0 +1,22 @@
+# Copyright (c) 2017-2024 Cloudflare, Inc.
+# Licensed under the Apache 2.0 license found in the LICENSE file or at:
+#     https://opensource.org/licenses/Apache-2.0
+
+using Workerd = import "/workerd/workerd.capnp";
+
+const eventSourceExample :Workerd.Config = (
+
+  services = [
+    (name = "main", worker = .eventsource),
+    (name = "internet", network = (allow = ["private"]))
+  ],
+
+  sockets = [ ( name = "http", address = "*:8080", http = (), service = "main" ) ]
+);
+
+const eventsource :Workerd.Worker = (
+  modules = [
+    (name = "worker", esModule = embed "worker.js")
+  ],
+  compatibilityDate = "2024-05-31",
+);
--- a/src/workerd/api/eventsource.c++
+++ b/src/workerd/api/eventsource.c++
@@ -1,0 +1,527 @@
+// Copyright (c) 2017-2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "eventsource.h"
+#include "http.h"
+#include "streams/common.h"
+#include <workerd/io/features.h>
+#include <workerd/jsg/exception.h>
+#include <workerd/util/mimetype.h>
+
+namespace workerd::api {
+
+namespace {
+class EventSourceSink final: public WritableStreamSink {
+public:
+  EventSourceSink(EventSource& eventSource) : eventSource(eventSource) {}
+
+  kj::Promise<void> write(kj::ArrayPtr<const kj::byte> buffer) override {
+    // The event stream is a new-line delimited format where each line represents an event.
+    // We need to scan the buffer for end-of-line characters. When we find one, everything before
+    // it is pushed into the event queue and we keep scanning. If we do not find an end-of-line
+    // sequence in the remaining input, we buffer it and wait for the next write to continue
+    // scanning, or until the stream is ended or aborted.
+
+    if (eventSource == kj::none) {
+      // Write was received after end() or abort() was called.
+      // We'll just ignore the write.
+      return kj::READY_NOW;
+    }
+
+    auto input = buffer.asChars();
+
+    // The stream may or may not begin with the UTF-8 BOM (%xFEFF). If this is the
+    // first write, we need to check for it and skip it if it is present.
+    // The BOM is a 3-byte sequence (0xEF, 0xBB, 0xBF) that encodes the Unicode
+    // codepoint U+FEFF. We only want to check for this once.
+    if (!bomChecked) {
+      bomChecked = true;
+      if (input.size() >= 3 &&
+          input[0] == '\xEF' &&
+          input[1] == '\xBB' &&
+          input[2] == '\xBF') {
+        input = input.slice(3);
+      }
+    }
+
+    while (input != nullptr) {
+      KJ_IF_SOME(found, findEndOfLine(input)) {
+          auto prefix = kept.releaseAsArray();
+          // Feed the line into the processor.
+          feed(kj::str(prefix, input.slice(0, found.pos)));
+          input = found.remaining;
+          // If we've reached the end of the input, input will == nullptr here.
+      } else {
+        // No end-of-line found, buffer the input.
+        kept.addAll(input.begin(), input.end());
+        input = nullptr;
+      }
+    }
+
+    // Release any buffered events to the EventSource
+    release();
+
+    return kj::READY_NOW;
+  }
+
+  kj::Promise<void> write(kj::ArrayPtr<const kj::ArrayPtr<const kj::byte>> pieces) override {
+    for (auto& piece : pieces) {
+      co_await write(piece);
+    }
+    co_return;
+  }
+
+  kj::Promise<void> end() override {
+    // The stream has finished. There's really nothing left to do here. Any partially
+    // filled data will be dropped on the floor.
+    clear();
+    return kj::READY_NOW;
+  }
+
+  void abort(kj::Exception reason) override {
+    // There's really nothing to do here.
+    clear();
+  }
+
+private:
+  kj::Maybe<EventSource&> eventSource;
+
+  // Retained bytes to be processed in the next write.
+  kj::Vector<char> kept;
+
+  // The collected messages that are pending to be dispatched as events
+  kj::Vector<EventSource::PendingMessage> pendingMessages;
+
+  // The message that is currently being processed.
+  kj::Maybe<EventSource::PendingMessage> currentPendingMessage;
+
+  // Set to true once the byte-order-mark has been checked
+  bool bomChecked = false;
+
+  EventSource::PendingMessage& getPendingMessage() {
+    KJ_IF_SOME(pending, currentPendingMessage) {
+      return pending;
+    }
+    return currentPendingMessage.emplace();
+  }
+
+  void feed(kj::String line) {
+    // Parse line according to the event stream format and dispatch the event.
+
+    // stream        = [ bom ] *event
+    // event         = *( comment / field ) end-of-line
+    // comment       = colon *any-char end-of-line
+    // field         = 1*name-char [ colon [ space ] *any-char ] end-of-line
+    // end-of-line   = ( cr lf / cr / lf )
+
+    // ; characters
+    // lf            = %x000A ; U+000A LINE FEED (LF)
+    // cr            = %x000D ; U+000D CARRIAGE RETURN (CR)
+    // space         = %x0020 ; U+0020 SPACE
+    // colon         = %x003A ; U+003A COLON (:)
+    // bom           = %xFEFF ; U+FEFF BYTE ORDER MARK
+    // name-char     = %x0000-0009 / %x000B-000C / %x000E-0039 / %x003B-10FFFF
+    //                 ; a scalar value other than U+000A LINE FEED (LF), U+000D CARRIAGE RETURN
+    //                   (CR), or U+003A COLON (:)
+    // any-char      = %x0000-0009 / %x000B-000C / %x000E-10FFFF
+    //                 ; a scalar value other than U+000A LINE FEED (LF) or U+000D CARRIAGE
+    //                   RETURN (CR)
+
+    // Note that the BOM (if present) is filtered out in the write() method.
+
+    if (line.size() == 0) {
+      // Dispatch the current pending message and clear it. If there is no
+      // pending message, we'll just ignore the line.
+      KJ_IF_SOME(pending, currentPendingMessage) {
+        // This message is done and ready to be dispatched. Add it to the
+        // pendingMessages list. The next time release() is called, it will
+        // be passed off to the EventSource.
+        pending.id = kj::str(KJ_ASSERT_NONNULL(eventSource).getLastEventId());
+        pendingMessages.add(kj::mv(pending));
+        currentPendingMessage = kj::none;
+      }
+    } else if (line[0] == ':') {
+      // Ignore the line.
+    } else {
+      static constexpr auto handle =
+          [](auto& self, kj::ArrayPtr<const char> field, kj::ArrayPtr<const char> value) {
+        auto& pending = self.getPendingMessage();
+        auto& ev = KJ_ASSERT_NONNULL(self.eventSource);
+        // Per the spec, only one space after the colon is optional and trimmed.
+        // Any other whitespace, or additional spaces aren't accounted for so would
+        // be part of the value.
+        if (value.size() > 0 && value[0] == ' ') {
+          value = value.slice(1);
+        }
+        if (field == "data"_kjc) {
+          pending.data.add(kj::str(value));
+        } else if (field == "event"_kjc) {
+          pending.event = kj::str(value);
+        } else if (field == "id"_kjc) {
+          ev.setLastEventId(kj::str(value));
+        } else if (field == "retry"_kjc) {
+          KJ_IF_SOME(time, kj::str(value).tryParseAs<uint32_t>()) {
+            KJ_ASSERT_NONNULL(self.eventSource).setReconnectionTime(time);
+          }
+          // Ignore the line if it cannot be successfully parsed as a uint32_t
+        }
+      };
+
+      KJ_IF_SOME(pos, line.findFirst(':')) {
+        handle(*this, line.slice(0, pos), line.slice(pos + 1));
+      } else {
+        handle(*this, line, ""_kjc);
+      }
+    }
+  }
+
+  void release() {
+    if (pendingMessages.size() == 0) return;
+    auto pending = pendingMessages.releaseAsArray();
+    // If the event source is gone, just drop the messages on the floor.
+    KJ_IF_SOME(es, eventSource) {
+      es.enqueueMessages(kj::mv(pending));
+    }
+  }
+
+  void clear() {
+    eventSource = kj::none;
+    kept.clear();
+    pendingMessages.clear();
+    currentPendingMessage = kj::none;
+  }
+
+  struct EndOfLine {
+    size_t pos;
+    kj::ArrayPtr<const char> remaining;
+  };
+  kj::Maybe<EndOfLine> findEndOfLine(kj::ArrayPtr<const char> input) {
+    // The end-of-line marker is either \n, \r, or \r\n
+    size_t pos = 0;
+    while (pos < input.size()) {
+      if (input[pos] == '\n') {
+        return EndOfLine{pos, input.slice(pos + 1)};
+      } else if (input[pos] == '\r') {
+        if (pos + 1 < input.size() && input[pos + 1] == '\n') {
+          return EndOfLine{pos, input.slice(pos + 2)};
+        }
+        return EndOfLine{pos, input.slice(pos + 1)};
+      }
+      pos++;
+    }
+    return kj::none;
+  }
+};
+
+kj::Promise<void> processBody(IoContext& context, kj::Promise<DeferredProxy<void>> promise) {
+  try {
+    co_await context.waitForDeferredProxy(kj::mv(promise));
+  } catch (...) {
+    auto ex = kj::getCaughtExceptionAsKj();
+    // We would see a disconnection exception if the eventstream is closed for
+    // multiple kinds of reasons. If it's a network error and we know we can
+    // reconnect, we should try to reconnect.
+    if (ex.getType() == kj::Exception::Type::DISCONNECTED) {
+      co_return;
+    }
+    // Propagate the exception up.
+    throw;
+  }
+}
+}  // namespace
+
+jsg::Ref<EventSource> EventSource::constructor(
+    jsg::Lock& js,
+    kj::String url,
+    jsg::Optional<EventSourceInit> init) {
+  JSG_REQUIRE(IoContext::hasCurrent(), DOMNotSupportedError,
+      "An EventSource can only be created within the context of a worker request.");
+
+  KJ_IF_SOME(i, init) {
+    KJ_IF_SOME(withCredentials, i.withCredentials) {
+      JSG_REQUIRE(!withCredentials, DOMNotSupportedError,
+          "The init.withCredentials option is not supported. It must be false or undefined.");
+    }
+  }
+
+  auto eventsource = jsg::alloc<EventSource>(js,
+      JSG_REQUIRE_NONNULL(jsg::Url::tryParse(url.asPtr()),
+          DOMSyntaxError,
+          kj::str("Cannot open an EventSource to '", url ,"'. The URL is invalid.")),
+      kj::mv(init));
+  eventsource->start(js);
+  return kj::mv(eventsource);
+}
+
+jsg::Ref<EventSource> EventSource::from(jsg::Lock& js, jsg::Ref<ReadableStream> readable) {
+  JSG_REQUIRE(IoContext::hasCurrent(), DOMNotSupportedError,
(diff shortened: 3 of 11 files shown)


irvinebroque:
refs https://github.com/cloudflare/workerd/discussions/1640

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,176 @@
...
+    kj::Maybe<kj::StringPtr> getLastEventId() {
+      return lastEventId.map([](kj::String& s) { return s.asPtr(); });
+    }
+    kj::ArrayPtr<const char> getOrigin() { return origin; }
+  };
+
+  struct EventSourceInit {
+    // We don't actually make use of the standard withCredentials option. If this is set to
+    // any truthy value, we'll throw.
+    jsg::Optional<bool> withCredentials;
```
kentonv, line 75: `RequestInitializerDict::credentials` is defined as `jsg::WontImplement`. I think the two should be consistent, but I don't have a strong opinion about which one should change.

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,176 @@
...
+
+    // This is a non-standard workers-specific extension that allows the EventSource to
```
kentonv, line 77: I wonder if the better extension would be to allow an `EventSource` to be constructed directly from a `Response`. It feels very weird that this interface is reproducing a lot of the `fetch()` interface.

kentonv (commented):
Sorry to ask this after all this code is written, but why not implement this in JavaScript? Seems like a good candidate: not widely used, not too complicated, contains hairy parsing code that would benefit from memory safety.

(I haven't fully reviewed the code, just looked a bit at the interface.)

jasnell:
> Sorry to ask this after all this code is written, but why not implement this in JavaScript? Seems like a good candidate: not widely used, not too complicated, contains hairy parsing code that would benefit from memory safety.

Could be done either way. One bit I like about the c++ implementation is that it makes use of the optimized pumpTo when going from the fetch response stream to the `EventSourceSink` that actually performs the parsing, only popping back into JS to schedule emitting the message events. In other words, this approach should be pretty low overhead. But doing the implementation in JavaScript is also feasible. Implementation time is about the same either way.

src/workerd/api/eventsource.h
jasnell: I think I prefer this approach ... that is, I think we should change the `jsg::WontImplement` in `RequestInitializerDict` and handle it equivalently as here. Don't have strong feelings on it tho.

src/workerd/api/eventsource.h
jasnell: you mean like `new EventSource(response, ...)` ? or something else?

I view `EventSource` largely as just a convenience wrapper around fetch. One additional extension I was considering was making it so that we could wrap an `EventSource` around any `ReadableStream` so that we could use it even without performing a fetch at all... but I also don't want to stray too far from the standard API.

```
// I wouldn't want to diverge the EventSource constructor too far from the standard...
const eventSource = EventSource.from(readableStream);
```

kentonv: Yeah if EventSource doesn't actually depend on anything in the Response except for the body stream, then it really ought to be possible to create one from an arbitrary ReadableStream.

The fact that this API wants to make a fetch on your behalf makes it rather non-composable. Supporting fetcher bindings is one thing. But what about if you receive a Response (or a ReadableStream) over RPC, and want to parse that?

Also, I guess this API doesn't let you specify any custom headers on the request? Or see the headers on the response? This all seems very limiting for no good reason.

(I understand this is how the standard was written, just seems like a poorly-designed standard TBH.)

jasnell: heh, it's certainly not my favorite..

The most sophisticated bit is the auto-reconnection. This implementation does not currently do so, but there's a header that's supposed to be set in the reconnection fetch request to specify the last known event id provided by the server in a previous connection. Beyond that, the only thing it uses in the fetch is checking the response status code and verifying that the content-type is correct.

I think I will go ahead and add the non-standard `EventSource.from(...)` static method to make it possible to create an instance from a `ReadableStream` (or any async iterable). Then likely keep the rest as is.

jasnell: Added `EventSource.from(...)`

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,510 @@
...
+    // colon         = %x003A ; U+003A COLON (:)
+    // bom           = %xFEFF ; U+FEFF BYTE ORDER MARK
+    // name-char     = %x0000-0009 / %x000B-000C / %x000E-0039 / %x003B-10FFFF
+    //                 ; a scalar value other than U+000A LINE FEED (LF), U+000D CARRIAGE RETURN
+    //                   (CR), or U+003A COLON (:)
+    // any-char      = %x0000-0009 / %x000B-000C / %x000E-10FFFF
+    //                 ; a scalar value other than U+000A LINE FEED (LF) or U+000D CARRIAGE
+    //                   RETURN (CR)
+
+    if (line.size() == 0) {
```
jasnell, line 104: Parser TODOs:

* [ ] Needs to take the BOM into consideration.

src/workerd/jsg/exception.h
```diff
@@ -24,6 +24,7 @@ namespace workerd::jsg {
 #define JSG_ERROR_DOMTypeMismatchError JSG_DOM_EXCEPTION("TypeMismatchError")
 #define JSG_ERROR_DOMQuotaExceededError JSG_DOM_EXCEPTION("QuotaExceededError")
 #define JSG_ERROR_DOMAbortError JSG_DOM_EXCEPTION("AbortError")
+#define JSG_ERROR_DOMSyntaxError JSG_DOM_EXCEPTION("SyntaxError")
```
vickykont, line 27: Seems like this is already defined above.

jasnell, line 27:
```suggestion
```

samples/eventsource/config.capnp
```diff
@@ -0,0 +1,19 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const eventSourceExample :Workerd.Config = (
+
+  services = [
+    (name = "main", worker = .eventsource),
```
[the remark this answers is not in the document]
jasnell: Not sure it matters much.

samples/eventsource/README.md
```diff
@@ -0,0 +1,28 @@
+# Hello World Example
```
jasnell, line 1:
```suggestion
# EventSource Example
```

samples/eventsource/README.md
```diff
@@ -0,0 +1,28 @@
...
+To run the example on http://localhost:8080
+
+```sh
+$ ./workerd serve config.capnp
+```
+
+To run using bazel
+
+```sh
+$ bazel run //src/workerd/server:workerd -- serve ~/cloudflare/workerd/samples/helloworld_esm/config.capnp
```
jasnell, line 12:
```suggestion
$ bazel run //src/workerd/server:workerd -- serve ~/cloudflare/workerd/samples/eventsource/config.capnp
```

samples/eventsource/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+$ ./helloworld
+```
+
+To test:
+
+```sh
+% curl http://localhost:8080
+Hello World
+```
```
jasnell, line 28:
```suggestion
```

vickykont (commented):
Leaving a few comments and I'll take one more look!

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,257 @@
+#pragma once
+#include <workerd/jsg/jsg.h>
```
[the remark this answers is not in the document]
jasnell: Yes it's imported there but I think our goal is to move more towards IWYU (include-what-you-use)

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,257 @@
...
+
+    // EventSource is not defined by the spec as being disposable using ERM, but
+    // it makes sense to do so. The dispose operation simply defers to close().
+    // This will enable `using eventsource = new EventSource(...)`
+    JSG_DISPOSE(close);
+  }
+
+  struct PendingMessage {
+    kj::Vector<kj::String> data;
+    kj::Maybe<kj::String> event;
```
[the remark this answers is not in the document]
jasnell: The `type` comes from the `EventTarget` naming while `event` is how `EventSource` refers to it. Yes it's a bit confusing but I think it's preferable to keep the language as is.

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,257 @@
...
+    kj::String id;
+  };
+
+  // Called by the internal implementation to notify the EventSource about messages
+  // received from the server.
+  void enqueueMessages(kj::Array<PendingMessage> messages);
```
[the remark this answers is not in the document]
jasnell: They could be but I don't think it matters much

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,257 @@
...
+
+  // Called by the internal implementation to retrieve the last event id that was
+  // specified by the server.
+  kj::StringPtr getLastEventId();
+
+  // Called by the internal implementation to set the last event id that was specified
+  // by the server.
+  void setLastEventId(kj::String id);
+
+  void visitForGc(jsg::GcVisitor& visitor);
```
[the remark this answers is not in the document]
jasnell: It can be but also really doesn't matter much if it is.

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,257 @@
...
+  // The EventSource spec defines onopen, onmessage, and onerror as prototype
+  // properties on the class.
+  kj::Maybe<jsg::JsRef<jsg::JsValue>> onopenValue;
+  kj::Maybe<jsg::JsRef<jsg::JsValue>> onmessageValue;
+  kj::Maybe<jsg::JsRef<jsg::JsValue>> onerrorValue;
+
+  // The default reconnection wait time. This is fairly arbitrary and is left
+  // entirely up to the implementation. The event stream can provide a new value
+  static constexpr auto DEFAULT_RECONNECTION_TIME = 2 * kj::SECONDS;
+  static constexpr uint32_t MIN_RECONNECTION_TIME = 1000;
```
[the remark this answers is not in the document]
jasnell: Largely to simplify the arithmetic in setReconnectionTime a bit.

src/workerd/api/tests/eventsource-test.wd-test
```diff
@@ -0,0 +1,18 @@
...
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "eventsource-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "eventsource-test.js")
+        ],
+        compatibilityDate = "2023-01-15",
+        compatibilityFlags = ["nodejs_compat", "experimental"],
```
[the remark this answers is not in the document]
jasnell: Yes, the test uses `node:assert`

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,526 @@
...
+  readyState = State::OPEN;
+  dispatchEventImpl(js, jsg::alloc<OpenEvent>());
+}
+
+void EventSource::notifyMessages(jsg::Lock& js, kj::Array<PendingMessage> messages) {
+  if (readyState == State::CLOSED) return;
+  js.tryCatch([&] {
+    for (auto& message : messages) {
+        auto data = kj::str(kj::delimited(kj::mv(message.data), "\n"_kjc));
+        if (data.size() == 0) continue;
```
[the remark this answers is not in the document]
jasnell: it can if `message.data` is empty

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,526 @@
...
+    }
+  }, [&](jsg::Value exception) {
+    // If we end up with an exception being thrown in one of the event handlers, we will
+    // stop trying to process the messages and instead just error the EventSource.
+    notifyError(js, jsg::JsValue(exception.getHandle(js)));
+  });
+}
+
+void EventSource::reconnect(jsg::Lock& js) {
+  KJ_ASSERT(impl != kj::none);
```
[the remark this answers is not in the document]
jasnell: Simply because we're not doing anything with the return value.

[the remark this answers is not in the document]
jasnell: I don't think it would be necessary to do so.

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,526 @@
...
+    self->notifyError(js, jsg::JsValue(exception.getHandle(js)));
+    return js.resolvedPromise();
+  });
+
+  auto headers = jsg::alloc<Headers>();
+  headers->set(jsg::ByteString(kj::str("accept")),
+               jsg::ByteString(MimeType::EVENT_STREAM.essence()));
+  headers->set(jsg::ByteString(kj::str("cache-control")),
+               jsg::ByteString(kj::str("no-cache")));
+  if (lastEventId != ""_kjc) {
```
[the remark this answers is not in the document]
jasnell: yes, per the spec

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,526 @@
...
+      auto handle = KJ_ASSERT_NONNULL(ex.tryGetHandle(js));
+      self->notifyError(js, jsg::JsValue(handle));
+      return js.resolvedPromise();
+    }
+
+    // If the request was redirected, update the URL to the new location.
+    if (response->getRedirected()) {
+      KJ_IF_SOME(newUrl, jsg::Url::tryParse(response->getUrl())) {
+        impl.url = kj::mv(newUrl);
+      } else {}  // Extra else block to squash compiler warning
```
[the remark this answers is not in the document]
jasnell: In this case there's likely always a URL. This could throw an error but at this point there's no harm in just skipping if there's no url. (the `Response` object will always have a URL).

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,526 @@
...
+                 jsg::ByteString(kj::str(lastEventId)));
+  }
+
+  fetchImpl(js, kj::mv(fetcher), kj::str(i.url), RequestInitializerDict {
+    .headers = kj::mv(headers),
+    .signal = abortController->getSignal(),
+  }).then(js, kj::mv(onSuccess), kj::mv(onFailed));
+}
+
+namespace {
```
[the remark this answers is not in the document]
jasnell: Perhaps but not critical in this PR

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+
+void EventSource::reconnect(jsg::Lock& js) {
+  KJ_ASSERT(impl != kj::none);
+  readyState = State::CONNECTING;
+  abortController = jsg::alloc<AbortController>();
+  auto signal = abortController->getSignal();
+  context.awaitIo(js, signal->wrap(context.afterLimitTimeout(reconnectionTime)))
+      .then(js, JSG_VISITABLE_LAMBDA((self=JSG_THIS), (self), (jsg::Lock& js) mutable {
+    self->start(js);
+  }), JSG_VISITABLE_LAMBDA((self=JSG_THIS),(self),(jsg::Lock& js, jsg::Value exception) {
```
vickykont, line 327: Syntax seems a little off here.

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+  }
+
+private:
+  kj::Maybe<EventSource&> eventSource;
+  // Retained bytes to be processed in the next write.
+  kj::Vector<char> kept;
+  // The collected messages that are pending to be dispatched as events
+  kj::Vector<EventSource::PendingMessage> pendingMessages;
+
+  // The current message that is currently being processed.
```
vickykont, line 91:
```suggestion
  // The message that is currently being processed.
```

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+    clear();
+  }
+
+private:
+  kj::Maybe<EventSource&> eventSource;
+  // Retained bytes to be processed in the next write.
+  kj::Vector<char> kept;
+  // The collected messages that are pending to be dispatched as events
+  kj::Vector<EventSource::PendingMessage> pendingMessages;
+
```
vickykont, line 90: Super nit; New lines are separating some of the fields but not all. Maybe we don't need them?

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+  void release() {
+    if (pendingMessages.size() == 0) return;
+    auto pending = pendingMessages.releaseAsArray();
+    // If the event source is gone, just drop the messages on the floor.
+    KJ_IF_SOME(es, eventSource) {
+      es.enqueueMessages(kj::mv(pending));
+    }
+  }
+
+  void clear() {
```
vickykont, line 183: Do we need to do anything for the `pendingMessage`?

vickykont (commented):
I did one more pass. Mostly nits and questions, but we also need to address the copyrights.

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,261 @@
...
+                 kj::Maybe<jsg::Url&> url)
+        : Event(kj::mv(type).orDefault([] { return kj::str("message");})),
+          data(kj::mv(data)),
+          lastEventId(kj::mv(lastEventId)),
+          origin(url.map([](auto& url) { return url.getOrigin(); })) {}
+
+    static jsg::Ref<MessageEvent> constructor() = delete;
+    JSG_RESOURCE_TYPE(MessageEvent) {
+      JSG_INHERIT(Event);
+      JSG_LAZY_READONLY_INSTANCE_PROPERTY(data, getData);
```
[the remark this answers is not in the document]
jasnell: The `MessageEvent` type is used by multiple mechanisms in the browser. The `source` and `ports` are fields used by one of the mechanisms we do not implement support for. These aren't relevant to `EventSource`

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,261 @@
...
+  static jsg::Ref<EventSource> from(jsg::Lock& js, jsg::Ref<ReadableStream> stream);
+
+  kj::Maybe<jsg::JsValue> getOnOpen(jsg::Lock& js) {
+    return onopenValue.map([&](jsg::JsRef<jsg::JsValue>& ref) -> jsg::JsValue {
+      return ref.getHandle(js);
+    });
+  }
+  void setOnOpen(jsg::Lock& js, jsg::JsValue value) {
+    if (!value.isObject() && !value.isFunction()) {
+      onopenValue = kj::none;
```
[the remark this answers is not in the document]
jasnell: No, the standard behavior is to clear the value and ignore.

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+  JSG_REQUIRE(!readable->isLocked(), TypeError,
+               "This ReadableStream is locked");
+  JSG_REQUIRE(!readable->isDisturbed(), TypeError,
+              "This ReadableStream has already been read from");
+  auto eventsource = jsg::alloc<EventSource>(js);
+  eventsource->run(js, kj::mv(readable), false /* No reconnection attempts */);
+  return kj::mv(eventsource);
+}
+
+EventSource::EventSource(jsg::Lock& js, jsg::Url url, kj::Maybe<EventSourceInit> init)
```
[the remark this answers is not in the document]
jasnell: By convention I'll typically try to use `jsg::Optional<T>` on the JavaScript-facing constructor, `kj::Maybe<T>` on the internal C++ constructor

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+  }, [&](jsg::Value exception) {
+    // If we end up with an exception being thrown in one of the event handlers, we will
+    // stop trying to process the messages and instead just error the EventSource.
+    notifyError(js, jsg::JsValue(exception.getHandle(js)));
+  });
+}
+
+void EventSource::reconnect(jsg::Lock& js) {
+  KJ_ASSERT(impl != kj::none);
+  readyState = State::CONNECTING;
```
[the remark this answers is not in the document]
jasnell: Not really

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+    return js.resolvedPromise();
+  };
+
+  auto onSuccess = JSG_VISITABLE_LAMBDA(
+      (self=JSG_THIS,
+       fetcher = fetcher.map([](jsg::Ref<Fetcher>& f) -> jsg::Ref<Fetcher> { return f.addRef(); })),
+      (self, fetcher),
+      (jsg::Lock& js, jsg::Ref<Response> response) {
+    if (self->readyState == State::CLOSED) return js.resolvedPromise();
+    auto& impl = KJ_ASSERT_NONNULL(self->impl);
```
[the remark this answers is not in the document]
jasnell: The state could have changed while we were waiting for the promise to resolve.

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+       fetcher=addRef(fetcher)),
+      (self, readable, response, fetcher),
+      (jsg::Lock& js) {
+    // The pump finished. Did the server disconnect? If so, try reconnecting if we can.
+    self->notifyError(js, js.error("The server disconnected."), withReconnection);
+    if (withReconnection) self->reconnect(js);
+  });
+
+  auto onFailed = JSG_VISITABLE_LAMBDA(
+      (self=JSG_THIS, response=addRef(response), fetcher=addRef(fetcher)),
```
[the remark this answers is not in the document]
jasnell: To defensively maintain strong references to they can't be garbage collected while these operations are pending. Might not be strictly necessary but also fairly harmless.

src/workerd/api/eventsource.c++
jasnell: Not sure what you mean. We are clearing the `pendingMessages`

src/workerd/api/eventsource.c++
jasnell: Not too worried about the formatting here.

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+
+    // If the request was redirected, update the URL to the new location.
+    if (response->getRedirected()) {
+      KJ_IF_SOME(newUrl, jsg::Url::tryParse(response->getUrl())) {
+        impl.url = kj::mv(newUrl);
+      } else {}  // Extra else block to squash compiler warning
+    }
+
+    KJ_IF_SOME(body, response->getBody()) {
+      // Well, ok! We're ready to start trying to process the stream! We do so by
```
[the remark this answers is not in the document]
jasnell: Yes it is intended.

src/workerd/api/eventsource.c++
```diff
@@ -0,0 +1,520 @@
...
+
+  KJ_IF_SOME(i, init) {
+    KJ_IF_SOME(withCredentials, i.withCredentials) {
+      JSG_REQUIRE(!withCredentials, DOMNotSupportedError,
+          "The init.withCredentials option is not supported. It must be false or undefined.");
+    }
+  }
+
+  auto eventsource = jsg::alloc<EventSource>(js,
+      JSG_REQUIRE_NONNULL(jsg::Url::tryParse(url.asPtr()),
```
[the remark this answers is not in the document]
jasnell: For now I think I'll leave it as is. If it becomes confusing for folks we can move it later

src/workerd/api/eventsource.c++
vickykont: Sorry, I meant the `currentPendingMessage`.

jasnell: ah right. added

src/workerd/api/eventsource.h
```diff
@@ -0,0 +1,264 @@
...
+  static constexpr uint32_t MIN_RECONNECTION_TIME = 1000;
+  static constexpr uint32_t MAX_RECONNECTION_TIME = 10 * 1000;
+
+  kj::Duration reconnectionTime = DEFAULT_RECONNECTION_TIME;
+
+  void notifyOpen(jsg::Lock& js);
+  void notifyError(jsg::Lock& js, const jsg::JsValue& error, bool reconnecting = false);
+  void notifyMessages(jsg::Lock& js, kj::Array<PendingMessage> messages);
+
+  // The run() method handles the actually processing of the stream.
```
vickykont, line 245:
```suggestion
  // The run() method handles the actual processing of the stream.
```

PR #1783, opened 2024-03-08
SQLite: add streaming input SQL endpoint `sql.ingest`

geelen:
This PR adds the required APIs to enable ingesting a ReadableStream of data from a Request and, in a single transaction, applying all the statements to the DB. For example:

```js
const reader = request.body.getReader()
const decoder = new TextDecoder()

await storage.transaction(async () => {
  let buffer = ''
  while (true) {
    const { value, done } = await reader.read()
    if (done) break
    const chunk = decoder.decode(value, { stream: true })

    // Append the new chunk to the existing buffer
    buffer += chunk

    // Ingest any complete statements and snip those chars off the buffer
    buffer = sql.ingest(buffer)
  }
})
```

This is made possible by a new API, `sql.ingest(string)` that immediately executes all valid SQL statements in the string, returning the substring of the input that is yet to be processed (which may be ''). Using this, and an outer transaction, you can safely consume the input a chunk at a time, enabling ingestion of SQL inputs way larger than the Worker could keep in memory.

To do this, we needed to alter the behaviour of SQLite's internal `sqlite3_complete` function to return true if there was at least one valid statement in the given input (rather than the entire input being complete). To keep the diff as small as possible, we extracted a new internal `sqlite3_complete_length` with an optional `firstOnly` param and wired the existing `sqlite3_complete` function up.

Note that `exec` will continue to happily compile & execute a statement without a trailing `;` (i.e. `SELECT 123`), but `ingest` will see no "complete" statements in that string and return the string unchanged. This matches the behaviour of `sqlite_complete` versus `sqlite3_prepare_v3`, as that's what those functions are calling under the hood.

### DIFF
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -96,15 +96,19 @@ async function test(storage) {
   {
     sql.exec(`CREATE TABLE streaming(val TEXT);`)
 
-    // Use a chunk size 1, 2, 4, 8, 16, ... characters
-    for (let length = 1; length < INSERT_36_ROWS.length; length = length * 2) {
+    // Convert to binary otherwise .split can cause corruption for multi-byte chars
+    const inputBytes = new TextEncoder().encode(INSERT_36_ROWS)
+    const decoder = new TextDecoder()
+
+    // Use a chunk size 1, 3, 9, 27, 81, ... bytes
+    for (let length = 1; length < inputBytes.length; length = length * 3) {
       let buffer = ''
-      for (let offset = 0; offset < INSERT_36_ROWS.length; offset += length) {
+      for (let offset = 0; offset < inputBytes.length; offset += length) {
         // Simulate a single "chunk" arriving
-        const chunk = INSERT_36_ROWS.substring(offset, offset + length)
+        const chunk = inputBytes.slice(offset, offset + length)
 
         // Append the new chunk to the existing buffer
-        buffer += chunk
+        buffer += decoder.decode(chunk, { stream: true })
 
         // Ingest any complete statements and snip those chars off the buffer
         buffer = sql.ingest(buffer)
@@ -118,6 +122,18 @@ async function test(storage) {
         { 'count(*)': 36 },
       ])
 
+      // Ensure our precious emoji were preserved, even if their bytes occur across split points
+      assert.deepEqual(
+        Array.from(sql.exec(`SELECT * FROM streaming WHERE val LIKE 'f%'`)),
+        [
+          { val: 'f: 😳' },
+          { val: 'f: 🫠' },
+          { val: 'f: 🙃' },
+          { val: 'f: 🤡' },
+          { val: 'f: 🥺' },
+          { val: 'f: 🔥😎🔥' },
+        ]
+      )
       sql.exec(`DELETE FROM streaming`)
       await scheduler.wait(1)
     }
@@ -1016,6 +1032,17 @@ async function testStreamingIngestion(request, storage) {
   assert.deepEqual(Array.from(sql.exec(`SELECT count(*) FROM streaming`)), [
     { 'count(*)': 36 },
   ])
+  assert.deepEqual(
+    Array.from(sql.exec(`SELECT * FROM streaming WHERE val LIKE 'f%'`)),
+    [
+      { val: 'f: 😳' },
+      { val: 'f: 🫠' },
+      { val: 'f: 🙃' },
+      { val: 'f: 🤡' },
+      { val: 'f: 🥺' },
+      { val: 'f: 🔥😎🔥' },
+    ]
+  )
 }
 
 export class DurableObjectExample {
@@ -1079,11 +1106,19 @@ export default {
         method: 'POST',
         body: new ReadableStream({
           async start(controller) {
-            const data = INSERT_36_ROWS.match(new RegExp('.{1,100}', 'g')) // splits into 5 chunks
+            const data = new TextEncoder().encode(INSERT_36_ROWS)
+
+            // Pick a value for chunkSize that splits the first emoji in half
+            const chunkSize = INSERT_36_ROWS.indexOf('😳') + 1
+            assert.equal(chunkSize, 35) // Validate we're getting the value we expect
 
             // Send each chunk with a wait of 1ms in between
-            for (const chunk of data) {
-              controller.enqueue(new TextEncoder().encode(chunk))
+            for (
+              let offset = 0;
+              offset < data.length - 1;
+              offset += chunkSize
+            ) {
+              controller.enqueue(data.slice(offset, offset + chunkSize))
               await scheduler.wait(1)
             }
 
@@ -1122,7 +1157,7 @@ export default {
 const INSERT_36_ROWS = ['a', 'b', 'c', 'd', 'e', 'f']
   .map(
     (prefix) =>
-      `INSERT INTO streaming VALUES ${['u', 'v', 'w', 'x', 'y', 'z']
+      `INSERT INTO streaming VALUES ${['😳', '🫠', '🙃', '🤡', '🥺', '🔥😎🔥']
         .map((suffix) => `('${prefix}: ${suffix}')`)
         .join(',')};`
   )
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -90,6 +90,40 @@ async function test(storage) {
     sql.ingest(`SELECT 123; SELECT 456; -- trailing comment`),
     ' -- trailing comment'
   )
+
+  // Test execution of ingested queries by taking an input of 6 INSERT statements, that all
+  // add 6 rows of data, then splitting that into a bunch of chunks, then ingesting them all
+  {
+    sql.exec(`CREATE TABLE streaming(val TEXT);`)
+
+    // Use a chunk size 1, 2, 4, 8, 16, ... characters
+    for (let length = 1; length < INSERT_36_ROWS.length; length = length * 2) {
+      let buffer = ''
+      for (let offset = 0; offset < INSERT_36_ROWS.length; offset += length) {
+        // Simulate a single "chunk" arriving
+        const chunk = INSERT_36_ROWS.substring(offset, offset + length)
+
+        // Append the new chunk to the existing buffer
+        buffer += chunk
+
+        // Ingest any complete statements and snip those chars off the buffer
+        buffer = sql.ingest(buffer)
+
+        // Simulate awaiting next chunk
+        await scheduler.wait(1)
+      }
+
+      // Verify exactly 36 rows were added
+      assert.deepEqual(Array.from(sql.exec(`SELECT count(*) FROM streaming`)), [
+        { 'count(*)': 36 },
+      ])
+
+      sql.exec(`DELETE FROM streaming`)
+      await scheduler.wait(1)
+    }
+    sql.exec(`DROP TABLE streaming;`)
+  }
+
   // Test count
   {
     const result = [
@@ -1037,3 +1071,12 @@ export default {
     assert.equal(await doReq('increment'), 3)
   },
 }
+
+const INSERT_36_ROWS = ['a', 'b', 'c', 'd', 'e', 'f']
+  .map(
+    (prefix) =>
+      `INSERT INTO streaming VALUES ${['u', 'v', 'w', 'x', 'y', 'z']
+        .map((suffix) => `('${prefix}: ${suffix}')`)
+        .join(',')};`
+  )
+  .join(' ')
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -60,6 +60,36 @@ async function test(storage) {
     assert.equal(result[2]['value'], 3)
   }
 
+
+  // Test partial query ingestion
+  assert.deepEqual(sql.ingest(`SELECT 123; SELECT 456;    `), '    ')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELECT 456;`), '')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELECT 456`), ' SELECT 456')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELECT 45`), ' SELECT 45')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELECT 4`), ' SELECT 4')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELECT `), ' SELECT ')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELECT`), ' SELECT')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELEC`), ' SELEC')
+  assert.deepEqual(sql.ingest(`SELECT 123; SELE`), ' SELE')
+  assert.deepEqual(sql.ingest(`SELECT 123; SEL`), ' SEL')
+  assert.deepEqual(sql.ingest(`SELECT 123; SE`), ' SE')
+  assert.deepEqual(sql.ingest(`SELECT 123; S`), ' S')
+  assert.deepEqual(sql.ingest(`SELECT 123; `), ' ')
+  assert.deepEqual(sql.ingest(`SELECT 123;`), '')
+  assert.deepEqual(sql.ingest(`SELECT 123`), 'SELECT 123')
+  assert.deepEqual(sql.ingest(`SELECT 12`), 'SELECT 12')
+  assert.deepEqual(sql.ingest(`SELECT 1`), 'SELECT 1')
+
+  // Exec throws with trailing comments
+  assert.throws(
+    () => sql.exec('SELECT 123; SELECT 456; -- trailing comment'),
+    /SQL code did not contain a statement/
+  )
+  // Ingest does not
+  assert.deepEqual(
+    sql.ingest(`SELECT 123; SELECT 456; -- trailing comment`),
+    ' -- trailing comment'
+  )
   // Test count
   {
     const result = [
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -91,6 +91,11 @@ async function test(storage) {
     ' -- trailing comment'
   )
 
+  // Ingest throws if statement looks "complete" but is actually a syntax error:
+  assert.throws(() => sql.ingest(`SELECT * bunk;`), /Error: near "bunk": syntax error at offset/)
+  assert.throws(() => sql.ingest(`INSER INTO xyz VALUES ('a'),('b');`), /Error: near "INSER": syntax error/)
+  assert.throws(() => sql.ingest(`INSERT INTO xyz VALUES ('a')('b');`), /Error: near "\(": syntax error/)
+
   // Test execution of ingested queries by taking an input of 6 INSERT statements, that all
   // add 6 rows of data, then splitting that into a bunch of chunks, then ingesting them all
   {
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -995,6 +995,29 @@ async function testForeignKeys(storage) {
   }
 }
 
+async function testStreamingIngestion(request, storage) {
+  const { sql } = storage
+
+  sql.exec(`CREATE TABLE streaming(val TEXT);`)
+
+  await storage.transaction(async () => {
+    const stream = request.body.pipeThrough(new TextDecoderStream())
+    let buffer = ''
+
+    for await (const chunk of stream) {
+      // Append the new chunk to the existing buffer
+      buffer += chunk
+
+      // Ingest any complete statements and snip those chars off the buffer
+      buffer = sql.ingest(buffer)
+    }
+  })
+  // Verify exactly 36 rows were added
+  assert.deepEqual(Array.from(sql.exec(`SELECT count(*) FROM streaming`)), [
+    { 'count(*)': 36 },
+  ])
+}
+
 export class DurableObjectExample {
   constructor(state, env) {
     this.state = state
@@ -1024,6 +1047,9 @@ export class DurableObjectExample {
     } else if (req.url.endsWith('/sql-test-io-stats')) {
       await testIoStats(this.state.storage)
       return Response.json({ ok: true })
+    } else if (req.url.endsWith('/streaming-ingestion')) {
+      await testStreamingIngestion(req, this.state.storage)
+      return Response.json({ ok: true })
     }
 
     throw new Error('unknown url: ' + req.url)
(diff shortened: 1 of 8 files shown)


src/workerd/util/sqlite.c++
```diff
@@ -438,16 +447,62 @@ kj::Own<sqlite3_stmt> SqliteDatabase::prepareSql(
...
+int SqliteDatabase::ingestSql(Regulator& regulator, kj::StringPtr sqlCode) {
+  // Set up the regulator that will be used for authorizer callbacks while preparing this
+  // statement.
+  KJ_ASSERT(currentRegulator == nullptr, "recursive prepareSql()?");
+  KJ_DEFER(currentRegulator = nullptr);
+  currentRegulator = regulator;
+  const char* start = sqlCode.begin();
+  const char* tail = start;
+
+  SQLITE_REQUIRE(sqlite3_get_autocommit(db) == 0, "sql.ingest must be run inside a transaction");
```
geelen, line 475: Some debate here whether this should be a real constraint of the new API or just a convention we suggest users follow.

Note that if we remove this constraint we must add the code to call the `onWriteCallback` to set up an implicit transaction if not already present (see line 431 above).

src/workerd/util/sqlite.c++
```diff
@@ -386,37 +386,46 @@ kj::StringPtr SqliteDatabase::getCurrentQueryForDebug() {
   }
 }
 
-// Set up the regulator that will be used for authorizer callbacks while preparing this
-// statement.
 kj::Own<sqlite3_stmt> SqliteDatabase::prepareSql(
     Regulator& regulator, kj::StringPtr sqlCode, uint prepFlags, Multi multi) {
+  // Set up the regulator that will be used for authorizer callbacks while preparing this
```
kentonv, line 391: Is this refactoring necessary?

The structure might be a bit more straightforward, but a bunch of code is now duplicated, which makes it easy for future changes to accidentally only edit one of the duplicates.

And reviewing this isn't easy.

I see at least one behavior change here: In multi mode, we'll now accept the presence of empty statements, as long as there is at least one complete statement. Was this change intended? It is not mentioned in the commit description.

I'd prefer to drop this commit if it isn't needed.

src/workerd/util/sqlite.c++
```diff
@@ -438,16 +447,62 @@ kj::Own<sqlite3_stmt> SqliteDatabase::prepareSql(
...
+  const char* start = sqlCode.begin();
+  const char* tail = start;
+
+  SQLITE_REQUIRE(sqlite3_get_autocommit(db) == 0, "sql.ingest must be run inside a transaction");
+
+  // While there are still valid
+  while (tail != sqlCode.end() && sqlite3_complete_inner(tail, 1)) {
+    // Compile the next statement in the SQL
+    sqlite3_stmt* result = nullptr;
+    SQLITE_CALL(sqlite3_prepare_v3(db, sqlCode.begin(), sqlCode.size(), 0, &result, &tail));
```
kentonv, line 481: There's a bunch of code here duplicating stuff in `prepareSql()`. Could sqlite3_complete_inner() instead return the split point so we can split off exactly one statement and pass it to prepareSql()? Or could we add a new `Multi` enum for `prepareSql()` which instructs it to parse a single statement and then not fail if there's data left over, but instead maybe modify `sqlCode` in-place to remove the prefix?

src/workerd/util/sqlite.c++
```diff
@@ -438,16 +447,62 @@ kj::Own<sqlite3_stmt> SqliteDatabase::prepareSql(
...
+      } else {
+        SQLITE_CALL_FAILED("sqlite3_step()", err);
+      }
     }
 
-    return ownResult;
+    sqlCode = kj::StringPtr(tail, sqlCode.end());
   }
+
+  // Return how many characters were processed
```
kentonv, line 504: Unfortunately, you are returning a number of UTF-8 bytes, but JavaScript will of course need a measure of UTF-16 code units in order to compute a substring.

Perhaps the function should actually return the leftover text, to avoid ambiguity?

Speaking of, when streaming in SQL, how will you deal with multi-byte characters being split across a buffer boundary?

src/workerd/api/sql-test.js
```diff
@@ -60,6 +61,67 @@ async function test(storage) {
...
+  assert.deepEqual(sql.ingest(`SELECT 123; SELEC`), 11)
+  assert.deepEqual(sql.ingest(`SELECT 123; SELE`), 11)
+  assert.deepEqual(sql.ingest(`SELECT 123; SEL`), 11)
+  assert.deepEqual(sql.ingest(`SELECT 123; SE`), 11)
+  assert.deepEqual(sql.ingest(`SELECT 123; S`), 11)
+  assert.deepEqual(sql.ingest(`SELECT 123; `), 11)
+  assert.deepEqual(sql.ingest(`SELECT 123;`), 11)
+  assert.deepEqual(sql.ingest(`SELECT 123`), 0)
+  assert.deepEqual(sql.ingest(`SELECT 12`), 0)
+  assert.deepEqual(sql.ingest(`SELECT 1`), 0)
```
justin-mp: It would probably be good to have a test case for just a semicolon (";") and for something where there's a semicolon inside another part of the statement (e.g., a string) just as a confidence test.

src/workerd/api/sql-test.js
```diff
@@ -99,28 +99,76 @@ async function test(storage) {
...
+  {
+    sql.exec(`CREATE TABLE streaming(val TEXT);`)
+    const inserts = ['a', 'b', 'c', 'd', 'e', 'f']
+      .map(
+        (c) =>
+          `INSERT INTO streaming VALUES ('${c}a'),('${c}b'),('${c}c'),('${c}d'),('${c}e'),('${c}f');`
+      )
+      .join('\n')
+
+    // Use a chunk size 2, 4, 8, 16, ... characters
```
justin-mp, line 137: The gold standard for this kind of test is getting 1 byte at a time (so the minimum here should be 1, not 2).

Do we have to worry about encodings here or is something ensuring that we get complete characters?  For example, if we're using UTF-8, what ensures that we're seeing everything in a multi-byte sequence.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#utf-16_characters_unicode_code_points_and_grapheme_clusters tells me that JavaScript uses UTF-16.  How do we ensure that we don't split up surrogate pairs (as one might have for an emoji in a string)?

src/workerd/util/sqlite.c++
kentonv: I don't think requiring a transaction is warranted:

* It seems like the primary intended use case for ingest() is for uploading a database dump in a streaming way. But a database which is being restored from a dump is almost certainly not serving live traffic. So while it is arguably a good idea to wrap it in a transaction, it probably doesn't matter for this use case.
* We don't know what the other use cases might be, but if there are others, I am not at all convinced that wrapping in a transaction is always the right thing to do. Forcing the developer to do it feels like a nanny-state type of thing to do.
* I think developers are plenty capable of reasoning for themselves about transactions. Or if they aren't, forgetting to apply one to ingest() is likely not going to be their biggest mistake.
* I think requiring a transaction makes it harder to explain what ingest() does. The explanation should be as simple as: "It's like exec(), except that it only parses complete statements ending in a semicolon. Any incomplete statement at the end is returned to you."

src/workerd/api/sql-test.js
kentonv: It's somewhat more accurate to say JavaScript uses WTF-16, that is, it does care if surrogate pairs are split, it just treats strings as arrays of 16-bit values. It's the font renderer's problem to decide what to do about a split surrogate pair.

Meanwhile in KJ land, we use UTF-8 -- or more accurately, WTF-8. If we see properly paired surrogates we will translate them into a single UTF-8 codepoint but if we see unpaired surrogates we'll just translate each one into a WTF-8 codepoint and carry on.

For the purpose of this specific PR, I think none of this is a concern because the only place we might potentially see the input get split in the middle of a code point is at the end of the buffer, but in every such case, we'll simply report that the statement is incomplete.

But as I commented above, it is a problem that the API returns the offset of the end of the last statement as a UTF-8 byte offset, which won't be useful for JavaScript operating on a UTF-16 string.

src/workerd/util/sqlite.c++
geelen: > It's like exec(), except that it only parses complete statements ending in a semicolon. Any incomplete statement at the end is returned to you.

Thanks for this, that's a better API and removes the need to even talk about transactions.

src/workerd/util/sqlite.c++
geelen: Have refactored, `ingestSql` now calls `prepareSql` and it's a bit nicer. It could be better with a bit more refactoring but I'll clean that up once I'm confident this version works re: string encoding.

src/workerd/util/sqlite.c++
geelen: I've dropped this commit for now but I _do_ actually want to change this as I've identified some test cases it'll fail on (plus I find the current implementation pretty confusing). But I'll do that as a follow-up rather than confusing the issue now.

src/workerd/util/sqlite.c++
geelen: > Perhaps the function should actually return the leftover text, to avoid ambiguity?

Implemented this

> Speaking of, when streaming in SQL, how will you deal with multi-byte characters being split across a buffer boundary?

Would appreciate some suggestions here, I've pushed a commit with a failing test case:

image

Maybe instead of `substring` I should be converting to an array buffer and concatenating that way?

src/workerd/api/sql-test.js
geelen: Will do, although with my latest refactoring `';'` will fail, and so will `SELECT 123;; SELECT 456;` because `prepareSql` explodes on all empty statements (which for things like trailing comments has been raised as a bug in the past).

Will fix and push test cases for handling semicolons/comments before merging this PR

src/workerd/api/sql-test.js
geelen: Have refactored this to no longer use character counts at all.

> The gold standard for this kind of test is getting 1 byte at a time (so the minimum here should be 1, not 2).

I've added a test-case for single byte chunks and it's corrupting emojis. See discussion here: https://github.com/cloudflare/workerd/pull/1783#discussion_r1521252427

src/workerd/util/sqlite.c++
kentonv: Hmm, I'd expect your use of `TextDecoder` in streaming mode to solve this... presumably it should buffer any incomplete UTF-8 sequences?

src/workerd/api/sql.c++
```diff
@@ -19,6 +19,11 @@ jsg::Ref<SqlStorage::Cursor> SqlStorage::exec(jsg::Lock& js, kj::String querySql
   return jsg::alloc<Cursor>(*sqlite, regulator, querySql, kj::mv(bindings));
 }
 
+kj::StringPtr SqlStorage::ingest(jsg::Lock& js, kj::String querySql) {
```
kentonv, line 22: Returning `StringPtr` here is not safe since it's going to be pointing into `querySql` which is deleted at the end of this function, before the caller can do anything with the return value.

src/workerd/api/sql.c++
```diff
@@ -19,6 +19,11 @@ jsg::Ref<SqlStorage::Cursor> SqlStorage::exec(jsg::Lock& js, kj::String querySql
...
+  SqliteDatabase::Regulator& regulator = *this;
+  return sqlite->ingestSql(regulator, querySql);
```
kentonv, line 24: Perhaps in the case that nothing is consumed, we should return `null`. Otherwise the string is being copied back and forth from JS (converted UTF-16 -> UTF-8 -> UTF-16) for no reason. Could be expensive if a long SQL statement arrives in many small chunks.

src/workerd/api/sql.c++
justin-mp: For my own education, is there any reason why we need to take querySql as a `kj::String`?  If I were doing this just in KJ-land, it would make sense to take in `kj::StringPtr` and return a `kj::StringPtr` (with the documented restriction that the returned `kj::StringPtr` has the same lifetime as the input `kj::StringPtr`), but I don't know if the runtime has other restrictions.

kentonv: This is a limitation of the JSG layer. We don't support receiving a StringPtr from JavaScript because there's nothing for it to point to: the original JS string is in UTF-16. Since we have to make a copy anyway, we make you accept ownership of that copy.

I _suppose_ we could change that and _not_ make you take ownership, but it'd be precarious in this particular situation, where you're trying to return a substring of that string... you have to count on the JSG layer converting that _back_ to JS before the input string is freed.

src/workerd/api/sql.c++
geelen: If `sqlite->ingestSql` returned null could `SqlStorage::ingest` return the input string unchanged (i.e. without converting back)? So in JS land you can still write:

```js
bufferAfter = sql.ingest(bufferBefore)
```

And you'll just find `bufferAfter === bufferBefore` if no complete statements were present?

geelen: Alternatively, would it make sense for `sql.ingest` to actually operate on an `ArrayBuffer`, in which case it could go back to returning the _number_ of elements in the input that can now be dropped, but we side-step JS's string encoding?

Is an Response.body's ReadableStream encoded as UTF-16 or UTF-8?

src/workerd/util/sqlite.c++
geelen: Yes it turns out it was me doing `.substring` before calling `TextEncoder.encode` to send the binary. So I was mangling it before encoding it.

geelen: Yes it turns out it was me doing `.substring` before calling `TextEncoder.encode` to send the binary. So I was mangling it before encoding it.

src/workerd/api/sql.c++
geelen: This feels like it'd complicate the API quite a bit for a less than common use-case (long SQL statement, small chunks). `SQLITE_LIMIT_SQL_LENGTH` is set to 100k anyway so at worst it'd be a chunk size of, what, 4kb? So it'd go back-and-forth across the JS/C++ boundary 25 times in the absolute worst case?

src/workerd/api/sql-test.js
```diff
@@ -922,6 +1002,45 @@ async function testForeignKeys(storage) {
...
+
+  sql.exec(`CREATE TABLE streaming(val TEXT);`)
+
+  const reader = request.body.getReader()
+  const decoder = new TextDecoder()
+
+  await storage.transaction(async () => {
+    let buffer = ''
+    while (true) {
+      const { value, done } = await reader.read()
```
jasnell, line 1016: non-blocking... but just reminder.... Keep in mind that this is going to give you a new 4096-byte chunk allocated for each read. For reading a `request.body`, it's way more efficient to use a `byob` reader with a single allocated buffer up front.

src/workerd/api/sql-test.js
```diff
@@ -60,6 +60,86 @@ async function test(storage) {
...
+      let buffer = ''
+      for (let offset = 0; offset < inputBytes.length; offset += length) {
+        // Simulate a single "chunk" arriving
+        const chunk = inputBytes.slice(offset, offset + length)
+
+        // Append the new chunk to the existing buffer
+        buffer += decoder.decode(chunk, { stream: true })
+
+        // Ingest any complete statements and snip those chars off the buffer
+        buffer = sql.ingest(buffer)
```
jasnell, line 114: Nit: I wonder if it would be reasonable to have `sql.ingest(...)` also optionally accept the `TypedArray` here and perform the equivalent to `decoder.decode(...)` internally. It would save a bit of boilerplate in this pattern.

src/workerd/api/sql.c++
jasnell: Might I suggest a different approach entirely? Define the API as a `WritableStream`, this way you could pipe into it, then await the results separately...

```js
const ingester = sql.ingester();
ctx.waitUntil(response.body.pipeTo(ingester));
for await (const results of ingester) {
  // handle results of each distinct sql operation piped in...
}
```

This way there would be no need for the additional book keeping boilerplate or the additional copying.

src/workerd/api/sql-test.js
geelen: Potentially, and this would avoid copying strings around, but I think what we have is good enough for a v1. Imo it'd be more likely to introduce something like `response.body.pipeTo(sql.ingester())` instead...

src/workerd/api/sql-test.js
geelen: Moved to `TextDecoderStream` as suggested in chat.

src/workerd/api/sql.c++
kentonv: > If sqlite->ingestSql returned null could SqlStorage::ingest return the input string unchanged (i.e. without converting back)? So in JS land you can still write:

Theoretically I prefer that as an API, but every time you cross over between JS and C++, the string is being converted between UTF-16 and UTF-8, respectively. So simply returning the same string is not O(1).

Then again, even if `ingest()` returns null when there's nothing to consume, you're still doing the JS -> C++ conversion with every call, only avoiding the conversion back from C++ -> JS. So maybe it's not worth it.

Using ArrayBuffers everywhere could avoid this issue, they are zero-copy. In fact, maybe that ends up being more convenient anyway, given you are receiving the bytes from a stream on the JS side... you can skip the whole TextDecoder step and make things more efficient...

patches/sqlite/0002-sqlite-complete-early-exit.patch
```diff
@@ -0,0 +1,64 @@
...
+   static const u8 trans[8][8] = {
+@@ -252,14 +259,18 @@
+           token = tkOTHER;
+         }
+         break;
+       }
+     }
+-    state = trans[state][token];
+     zSql++;
++    state = trans[state][token];
```
kentonv, line 34: Reversing these two lines has no effect?

patches/sqlite/0002-sqlite-complete-early-exit.patch
```diff
@@ -0,0 +1,64 @@
...
++
++    // Once we have found a single complete statement, return its length
++    if (firstOnly && state==1) return zSql - initialZSql;
+   }
+-  return state==1;
++  // Otherwise only return the length if the entire input is valid
++  return state==1 && zSql - initialZSql;
```
kentonv, line 41: In C/C++, an `&&` expression can evaluate only to 0 or 1. It does not evaluate to the right-side value when the left side is truthy.

Is there a missing test which should have caught this?

src/workerd/util/sqlite.c++
```diff
@@ -445,6 +445,43 @@ kj::Own<sqlite3_stmt> SqliteDatabase::prepareSql(
...
+kj::StringPtr SqliteDatabase::ingestSql(Regulator& regulator, kj::StringPtr sqlCode) {
+  // While there's still some input SQL to process
+  while (sqlCode.begin() != sqlCode.end()) {
+    // And there are still valid statements:
+    auto statementLength = sqlite3_complete_length(sqlCode.begin(), 1);
+    if (!statementLength) break;
+
+    // Slice off the next valid statement and prepare it
+    auto nextStatement = kj::str(sqlCode.slice(0, statementLength));
+    auto result = prepareSql(regulator, nextStatement, 0, SINGLE);
```
kentonv, line 457: Instead of calling `prepareSql()` directly, I think you should construct a `Query` here and then call it.

* That will take care of the write callback for you.
* It'll also take care of a few things you are missing here, like setting `currentRegulator` correctly. It is actually necessary to set the regulator when _executing_ the query, in addition to preparing it, since various things can cause the query to be rebuilt at runtime.

patches/sqlite/0002-sqlite-complete-early-exit.patch
geelen: Unintentional change, reverted.

patches/sqlite/0002-sqlite-complete-early-exit.patch
geelen: That's weird, it's passing my tests. If this was returning `1` it would definitely be breaking them. But I changed it to this form because of build errors in MacOS, so I've switched it back to explicitly return `zSql - initialZSql` or `0`.

src/workerd/util/sqlite.c++
geelen: This is now done in commit https://github.com/cloudflare/workerd/commit/624db515d1ea53d947cb0e7f2b8a4c4f9916e383

Interestingly creating the `Query` is sufficient since it calls `init` which calls `nextRow` internally, so I don't actually have to do anything with the query object for the statement to be executed. Seems a bit weird to write:

```cpp
// Create a Query object, which will prepare & execute it
Query(*this, regulator, nextStatement);
```

without assigning it to a variable. Is there a different way I should express that (like `auto _query = Query(...)`?)

src/workerd/api/sql.c++
geelen: Yeah I'm happy with this slight inefficiency for now. I do think I'll add a separate API for consuming a ReadableStream. I really like `for await (const results of ingester)` because we could start reporting back progress as the import is running.

I'm not sure how that would work with the output gate so I think I'd like to punt on making this more optimised/explanatory until I have a full end-to-end example working from Wrangler <-> R2 <-> D1.

patches/sqlite/0002-sqlite-complete-early-exit.patch
```diff
@@ -0,0 +1,61 @@
...
+     }
+     state = trans[state][token];
+     zSql++;
++
++    // Once we have found a single complete statement, return its length
++    if (firstOnly && state==1) return zSql - initialZSql;
+   }
+-  return state==1;
++  // Otherwise only return the length if the entire input is valid
++  return state==1 ? zSql - initialZSql : 0;
```
kentonv, line 38: I suspect your tests don't cover this line because they always take the earlier return. In fact, I suppose when `firstOnly` is true, then this line can only be hit when `state` is 0?

I'm not sure there's any action to take here, it would be weird to write tests for code in sqlite that we don't use. But just checking that this matches your understanding...

kentonv (approved):

patches/sqlite/0002-sqlite-complete-early-exit.patch
geelen: Yeah we don't use `sqlite3_complete` in its original version anywhere. I have been manually running the SQLite suite on the code with this patch though and it's passing those tests.

src/workerd/util/sqlite.c++
smerritt: Casting to void is the C++ idiom for "I am intentionally ignoring this result", so something like

```
// Create a Query object, which will prepare & execute it
(void)Query(*this, regulator, nextStatement);
```

This only matters to the compiler if `Query` is marked `[[nodiscard]]`, which it isn't, but it will communicate to human readers of the code that yes, we're making and discarding this Query, and no, we didn't forget to do something with it.

PR #1784, opened 2024-03-08
JSRPC: Support resource management for RPC stubs

kentonv:
* RpcStubs and RpcPromises now have `Symbol.dispose`.
* Stubs passed in call parameters are automatically disposed upon return.
* If the result of a call is an object, a disposer will be added to it, which disposes all stubs therein. (If it's not an object, it can't possibly contain any stubs...)
* Stubs also have a `dup()` method which creates a separately-disposed copy of the stub, in case you want to keep a stub past when it would otherwise be automatically disposed.
* Of course, when the I/O context is destroyed, everything is implicitly disposed.

Note: The first few commits use a method named "dispose" but the last commit changes it to `Symbol.dispose`. Maybe I should have done that first, oh well.

### DIFF
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -177,9 +177,27 @@ public:
   void waitUntil(kj::Promise<void> promise);
   void passThroughOnException();
 
-  JSG_RESOURCE_TYPE(ExecutionContext) {
+  // Cancels the current execution context with the given exception, causing all execution to stop
+  // and throwing an error at the client.
+  void abort(jsg::Lock& js, jsg::Optional<jsg::Value> reason);
+
+  JSG_RESOURCE_TYPE(ExecutionContext, CompatibilityFlags::Reader flags) {
     JSG_METHOD(waitUntil);
     JSG_METHOD(passThroughOnException);
+
+    if (flags.getWorkerdExperimental()) {
+      // TODO(soon): Before making this generally available we need to:
+      // * Consider whether to use TerminateExecution() instead of throwing.
+      // * Make sure it's really not possible for more code to run in the context after abort().
+      //   Currently, abort() triggers in a partially async way so there's an opportunity for some
+      //   other event in the event queue to squeeze in.
+      // * Try to ensure that the provided error is actually the one that propagates out of event
+      //   handlers. Currently this is not consistently true.
+      // * Make sure all event handlers actually honor onAbort().
+      // * Enable the Durable Object version at the same time -- and make sure they're suitably
+      //   consistent with each other.
+      JSG_METHOD(abort);
+    }
   }
 };
 
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -619,11 +619,15 @@ public:
                  paramDisposalGroup = kj::mv(invocationResult.paramDisposalGroup)]
                 (jsg::Lock& js, jsg::Value value) mutable {
           jsg::JsValue resultValue(value.getHandle(js));
+
+          // Call makeCallPipeline before serializing becaues it may need to extract the disposer.
+          auto maybePipeline = makeCallPipeline(js, resultValue);
+
           serializeJsValue(js, resultValue, [&](capnp::MessageSize hint) {
             hint.wordCount += capnp::sizeInWords<CallResults>();
             hint.capCount += 1;
             auto results = callContext.initResults(hint);
-            KJ_IF_SOME(p, makeCallPipeline(js, resultValue)) {
+            KJ_IF_SOME(p, maybePipeline) {
               results.setCallPipeline(kj::mv(p));
             }
             return results.initResult();
@@ -972,6 +976,17 @@ public:
     }
   }
 
+  // Use this version of the constructor to pass the dispose function separately.
+  TransientJsRpcTarget(jsg::Lock& js, IoContext& ioCtx, jsg::JsObject object,
+                       kj::Maybe<v8::Local<v8::Function>> dispose,
+                       bool allowInstanceProperties = false)
+      : JsRpcTargetBase(ioCtx), object(js, object),
+        dispose(dispose.map([&](v8::Local<v8::Function> func) {
+          return jsg::V8Ref<v8::Function>(js.v8Isolate, func);
+        })),
+        allowInstanceProperties(allowInstanceProperties),
+        pendingEvent(ioCtx.registerPendingEvent()) {}
+
   ~TransientJsRpcTarget() noexcept(false) {
     // If we have a disposer, and the I/O context is not already destroyed, arrange to call the
     // disposer.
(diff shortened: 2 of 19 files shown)


src/workerd/jsg/setup.c++
```diff
@@ -360,15 +360,24 @@ IsolateBase::IsolateBase(const V8System& system, v8::Isolate::CreateParams&& cre
...
       opaqueTemplate->InstanceTemplate()->SetInternalFieldCount(Wrappable::INTERNAL_FIELD_COUNT);
       this->opaqueTemplate.Reset(ptr, opaqueTemplate);
+
+      // Create Symbol.dispose and Symbol.asyncDispose.
+      symbolDispose.Reset(ptr, v8::Symbol::New(ptr,
+          v8::String::NewFromUtf8(ptr, "dispose",
+              v8::NewStringType::kInternalized).ToLocalChecked()));
+      symbolAsyncDispose.Reset(ptr, v8::Symbol::New(ptr,
+          v8::String::NewFromUtf8(ptr, "asyncDispose",
+              v8::NewStringType::kInternalized).ToLocalChecked()));
```
jasnell, line 380: `v8::Symbol::For(...)` is probably better here. And we should probably namespace the description in the polyfill, essentially like `Symbol.for('cloudflare.dispose')`.

Thinking about it, I guess there's a non-zero chance this might break someone if they happen to already be polyfilling `Symbol.dispose` in their worker. It's probably unlikely enough to not worry about, but it's something we'll want to keep an eye out for.

src/workerd/jsg/rtti.capnp
```diff
@@ -212,6 +212,16 @@ struct Structure {
   asyncIterator @7 :Method;
   # Method returning async iterator if the structure is async iterable
 
+  disposable @13 :Bool;
+  # true if the structure is disposable
+  dispose @14 :Method;
+  # dispose method
```
jasnell, line 218: Just curious, why two separate fields for each? Could we not assume `disposable` is true if `dispose` has a value?

src/workerd/jsg/resource.c++
```diff
@@ -24,6 +25,28 @@ void exposeGlobalScopeType(v8::Isolate* isolate, v8::Local<v8::Context> context)
   KJ_ASSERT(check(global->Set(context, name, constructor)));
 }
 
+void polyfillSymbols(jsg::Lock& js, v8::Local<v8::Context> context) {
+  js.withinHandleScope([&]() {
+    JsObject obj(context->Global());
+
+    auto symbol = KJ_ASSERT_NONNULL(obj.get(js, "Symbol").tryCast<JsObject>());
+
+    KJ_ASSERT(!symbol.has(js, "dispose") && !symbol.has(js, "asyncDispose"),
```
jasnell, line 34: Probably only needs to be checked in debug builds?

```suggestion
    KJ_DASSERT(!symbol.has(js, "dispose") && !symbol.has(js, "asyncDispose"),
```

src/workerd/api/worker-rpc.h
```diff
@@ -259,7 +316,39 @@ class JsRpcStub: public JsRpcClientProvider {
...
+  // Nulled out upon dispose().
+  kj::Maybe<IoOwn<rpc::JsRpcTarget::Client>> capnpClient;
+
+  kj::Maybe<RpcStubDisposalGroup&> disposalGroup;
+  kj::ListLink<JsRpcStub> disposalGroupLink;
+
+  friend class RpcStubDisposalGroup;
+};
+
+class RpcStubDisposalGroup {
```
jasnell, line 328: non-blocking: I wonder if this is a pattern we might end up re-using for other disposable things.. might be worth abstracting a bit to have a generic `DisposalGroup` type.

src/workerd/api/global-scope.h
```diff
@@ -176,10 +176,15 @@ class ExecutionContext: public jsg::Object {
 public:
   void waitUntil(kj::Promise<void> promise);
   void passThroughOnException();
+  void abort(jsg::Lock& js, jsg::Optional<jsg::Value> reason);
```
jasnell, line 179: nit: some documentation comments here explaining the semantics expected would be helpful

src/workerd/api/worker-rpc.c++
```diff
@@ -147,12 +186,107 @@ private:
   kj::ForkedPromise<void> promise;
 };
 
+// Given a value, check if it has a dispose method and, if so, invoke it.
+void tryCallDisposeMethod(jsg::Lock& js, jsg::JsValue value) {
+  js.withinHandleScope([&]() {
```
jasnell, line 191: non-blocking: Hopefully v8 will give us an api to call this directly on a v8::Object

src/workerd/api/worker-rpc.c++
```diff
@@ -147,12 +186,107 @@ private:
...
+    KJ_IF_SOME(obj, value.tryCast<jsg::JsObject>()) {
+      auto dispose = obj.get(js, js.symbolDispose());
+      if (dispose.isFunction()) {
+        jsg::check(v8::Local<v8::Value>(dispose).As<v8::Function>()->Call(
+            js.v8Context(), value, 0, nullptr));
```
jasnell, line 196: The `Symbol.dispose` method is not supposed to return a value, but if is does, I wonder if we should emit a warning here... especially true if user code gets it wrong and has `Symbol.dispose` return a promise.

src/workerd/jsg/setup.c++
kentonv: > v8::Symbol::For(...) is probably better here.

Hmm, why would we want that? Seems like that could only lead to foot-shooting: someone might depend on the fact that `Symbol.dispose == Symbol.for("cloudflare.dispose")`, and then this assumption will break as soon as V8 starts providing this symbol itself.

> Thinking about it, I guess there's a non-zero chance this might break someone if they happen to already be polyfilling Symbol.dispose in their worker. It's probably unlikely enough to not worry about, but it's something we'll want to keep an eye out for.

How could it break someone?

If they are polyfilling carefully (only assigning the symbol if it doesn't already exist) they should be fine.

If they are just overwriting the symbol indiscriminately, they won't be able to use RPC correctly, but existing code should continue to work fine.

src/workerd/jsg/rtti.capnp
kentonv: I was just copying the pattern from iterable and asyncIterable. No idea why it's done this way but assume I should stay consistent.

src/workerd/api/worker-rpc.h
kentonv: Eh I don't want to create premature abstractions. Let's wait until we have a use case.

src/workerd/api/worker-rpc.c++
kentonv: Well, it's actually fine if it returns a promise. The promise will still run.

Note that all JSRPC disposal is async, that is, it always runs after the caller has already moved on, and any exceptions thrown by the disposer do not propagate to the caller. Given that, there's not actually anything inherently wrong with a disposer that performs async tasks.

src/workerd/api/worker-rpc.c++
```diff
@@ -366,10 +525,130 @@ kj::Maybe<jsg::Ref<JsRpcProperty>> JsRpcPromise::getProperty(jsg::Lock& js, kj::
...
+    // itself shuts down.
+    kj::mv(c).deferGcToContext();
+
+    // In preview, let's try to warn the developer about the problem.
+    //
+    // TODO(cleanup): Instead of logging this warning at GC time, it would be better if we logged
+    //   it at the time that the client is destroyed, i.e. when the IoContext is torn down,
+    //   which is usually sooner (and more deterministic). But logging a warning during
+    //   IoContext tear-down is problematic since logWarningOnce() is a method on
+    //   IoContext...
```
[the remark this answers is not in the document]
kentonv: I think we just need to refactor a bit so that it's a method on Worker::Lock or something like that, instead of IoContext.

src/workerd/api/global-scope.h
kentonv: b17aa75c

src/workerd/jsg/resource.c++
kentonv: f8ded118

src/workerd/api/worker-rpc.c++
jasnell: Oh, I know. My concern is largely catching coding bugs.. like exceptions that potentially end up being swallowed and just going nowhere.

```
async [Symbol.dispose]() { throw new Error('boom'); }
```

Certainly not a big deal, just being overly cautious.

src/workerd/jsg/setup.c++
jasnell: Sorry, I had meant `ForApi(...)`. My brain got stuck on `Symbol.for` when I was writing the comment. If we used `ForApi(...)` we wouldn't need to cache the symbol ref anywhere and could just use `ForApi(...)` when we wanted to grab it again.... So if I took another pass at my comment, it would have been...

```
v8::Symbol::ForApi(...) is probably better....`
```

and

```
v8::Symbol::ForApi(js.strIntern("cloudflare.dispose"));
```

But as it is currently is fine also.

> How could it break someone

As I said, it's probably unlikely enough not to worry about, but theoretically someone could have a branch the executes only if `Symbol.dispose` is not present, or they are already depending on `Symbol.dispose` being identity equal to their own polyfilled `Symbol.for('whatever')` variant. 

I don't think anyone will actually be broken by this, just acknowledging that the risk is non-zero (even if super close to zero)

src/workerd/api/worker-rpc.c++
kentonv: I believe these exceptions will be reported in the inspector as uncaught exceptions FWIW.

src/workerd/jsg/setup.c++
kentonv: Ah, I suppose `ForApi()` wouldn't have the problem of allowing applications to become dependent on it -- but still not clear if it provides any real advantages. The cached symbol ref is faster to access anyway (no need to look up an interned string, etc.) and more type-safe.

PR #1785, opened 2024-03-08
WIP: add expiry to state observing/modifying functions in actor-storage.capnp

Frederik-Baetens:


src/workerd/io/actor-storage.capnp
```diff
@@ -51,7 +54,7 @@ interface ActorStorage @0xd7759d7fc87c08e4 {
     txn @0 (settings :DbSettings) -> (transaction :Transaction);
 
     interface Transaction extends(Operations) {
-      commit @0 ();
+      commit @0 (ExpiryMsSinceUnixEpoch :UInt64);
       rollback @1 ();
```
justin-mp, line 58: Why doesn't `rollback` need a expiration timestamp?  Is the idea that something that's rolled back hasn't actually hit storage yet so we don't care about the session?

src/workerd/io/actor-storage.capnp
```diff
@@ -23,19 +23,22 @@ interface ActorStorage @0xd7759d7fc87c08e4 {
...
 
   interface Operations @0xb512f2ce1f544439 {
-    get @0 (key :Data) -> (value :Data);
-    list @3 (start :Data, end :Data, limit :Int32, reverse :Bool, stream :ListStream, prefix :Data);
-    put @1 (entries :List(KeyValue));
-    delete @2 (keys :List(Data)) -> (numDeleted :Int32);
+    # These operations are tagged with an ExpiryMsSinceUnixEpoch timestamp.
+    # This optional timestamp can be used to set a timeout after which the storage operation
+    # should no longer be executed, and an exception should be thrown instead.
+    get @0 (key :Data, ExpiryMsSinceUnixEpoch :UInt64) -> (value :Data);
```
justin-mp, line 29: Internally, a `kj::Date` is nanoseconds from the UNIX epoch, so I wonder if we're better off using nanoseconds.  (Nanoseconds will work for dates until 2292; milliseconds work for a million more years if that's what you need.) 

 An even better approach would be the one in https://groups.google.com/g/capnproto/c/YaeLDnShN_c/m/g4XoL04QS1gJ  where the *type* tells you the properties you care about and the name tells you what the particular value represents.

In this case, you would define a type like:

```
using Date = UInt64;
# Nanoseconds since the UNIX epoch
```
and then define your methods like so:

```
get @0 (key :Data, expiration :Date) -> (value :Data);
list @3 (start :Data, end :Data, limit :Int32, reverse :Bool, stream :ListStream, prefix :Data, expiration :Date);
...
```

justin-mp (commented):
Dumb question: where are the expiration timestamps going to come from?

src/workerd/io/actor-storage.capnp
Frederik-Baetens: Yes, I don't think rollback will be able to affect anything. It won't be able to affect a new session, since that new session shouldn't have a way of knowing about previous uncommitted transactions.

src/workerd/io/actor-storage.capnp
Frederik-Baetens: Thanks. I'll switch to this. I'm indeed having to do a little bit of ns -> ms conversions while continuing work on my ew pr, and just switching to ns will make that easier.

Frederik-Baetens:
> Dumb question: where are the expiration timestamps going to come from?

Haha that's what I've been spending the past week trying to figure out. Will dm you about it.

PR #1788, opened 2024-03-08
Determine handler exports from Python module

hoodmane:
After #1780, we now instantiate mainModule at top level (if not in workerd). In that case we can determine the handler exports directly from the functions that are defined on the Python module.

### DIFF
--- a/src/pyodide/python-entrypoint-helper.js
+++ b/src/pyodide/python-entrypoint-helper.js
@@ -173,27 +173,38 @@ async function preparePython() {
   return await getMainModule();
 }
 
+function makeHandler(pyHandlerName) {
+  return async function (...args) {
+    try {
+      const mainModule = await preparePython();
+      return await mainModule[pyHandlerName].callRelaxed(...args);
+    } catch (e) {
+      console.warn(e.stack);
+      throw e;
+    }
+  };
+}
 
-export default {
-  async fetch(request, env, ctx) {
-    try {
-      const mainModule = await preparePython();
-      if (mainModule.on_fetch === undefined) {
-        throw new Error("Python Worker should define an on_fetch method");
-      }
-      return await mainModule.on_fetch.callRelaxed(request, env, ctx);
-    } catch (e) {
-      console.warn(e.stack);
-      throw e;
+const handlers = {};
+
+if (IS_WORKERD) {
+  handlers.fetch = makeHandler("on_fetch");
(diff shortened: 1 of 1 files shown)


src/pyodide/python-entrypoint-helper.js
```diff
@@ -173,27 +173,37 @@ async function preparePython() {
...
+}
+
+const handlers = {};
+
+if (IS_WORKERD) {
+  handlers.fetch = makeHandler("on_fetch");
+  handlers.test = makeHandler("test");
+} else {
+  for (const handlerName of [
+    "fetch",
```
garrettgu10, line 195: Why is it fetch instead of on_fetch here?

hoodmane: Oops meant to do `"on_" +` instead of `"py_" +`  a couple lines down. These are the javascript names.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -173,27 +173,38 @@ async function preparePython() {
...
-      throw e;
+  };
+}
+
+const handlers = {};
+
+if (IS_WORKERD) {
+  handlers.fetch = makeHandler("on_fetch");
+  handlers.test = makeHandler("test");
+} else {
```
dom96, line 193: why not do the same for workerd as below?

dom96 (approved):

src/pyodide/python-entrypoint-helper.js
hoodmane: It's the same reason we have the `IS_WORKERD` check above:
```
  // If we're in workerd, we have to do setupPackages in the IoContext, so don't start it yet.
  // TODO: fix this.
```
can't call `setupPackages` in workerd without an IoContext, can't import the main module without calling `setupPackages` first.

dom96: not a blocker for this PR but we'll need to fix this to support other handlers in local dev mode

hoodmane: Indeed.

venu-mallik: How long before we have on_scheduled in local mode along with triggers supported for c3 and deployment with reqs.txt supported for python workers ? @dom96  @hoodmane 

I need a tentative timeline for this. thanks in advance.

PR #1794, opened 2024-03-11
Return errors for mutations on read-only files in the in-memory VFS (rather than crashing)

justin-mp:
There are two commits here:

1. Return error when writing a read-only file in the in-memory VFS
2. Return error when truncating a read-only file in the in-memory VFS

The commit message for the write commit has a bunch of detail on why we want that particular fix.  I'm ambivalent on the truncate commit and could see the wisdom of dropping that commit (especially since I don't know how to test it at a high level), but it does seem weird to fix write, but not fix truncate.

### DIFF
--- a/src/workerd/util/sqlite-test.c++
+++ b/src/workerd/util/sqlite-test.c++
@@ -271,6 +271,37 @@ KJ_TEST("Read-only database picks up on changes from mutable database (on-disk)"
   doReadOnlyUpdateTest(*dir);
 }
 
+KJ_TEST("In-memory read-only crash regression") {
+  auto dir = kj::newInMemoryDirectory(kj::nullClock());
+  SqliteDatabase::Vfs vfs(*dir);
+
+  {
+    SqliteDatabase db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+    setupSql(db);
+    checkSql(db);
+  }
+
+  // When using the in-memory file system, if we first create a read-only database
+  kj::Maybe<SqliteDatabase> rodb;
+  rodb.emplace(vfs, kj::Path({"foo"}));
+  checkSql(KJ_ASSERT_NONNULL(rodb));
+
+  // then create a read/write database
+  kj::Maybe<SqliteDatabase> db;
+  db.emplace(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+  checkSql(KJ_ASSERT_NONNULL(db));
+
+  // then write into the read/write database:
+  KJ_ASSERT_NONNULL(db).run("INSERT INTO people (id, name, email) VALUES (?, ?, ?);", 234,
+                            "Carol"_kj, "carol@example.com");
+
+  // we can destroy the read/write database with no problems,
+  db = kj::none;
+
+  // but we would crash when destroying the read-only database:
+  rodb = kj::none;
+}
+
 // Tests that concurrent database clients don't clobber each other. This verifies that the
 // LockManager interface is able to protect concurrent access and that our default implementation
 // works.
--- a/src/workerd/util/sqlite.c++
+++ b/src/workerd/util/sqlite.c++
@@ -1300,9 +1300,13 @@ const sqlite3_io_methods SqliteDatabase::Vfs::FileImpl::FILE_METHOD_TABLE = {
   .xWrite = [](sqlite3_file* file, const void* buffer, int iAmt,
                sqlite3_int64 iOfst) noexcept -> int {
     WRAP_METHOD(SQLITE_IOERR_WRITE, {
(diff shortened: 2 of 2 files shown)


jqmmes (approved):

kentonv (approved):
This change is clearly strictly an improvement regardless, but I have a couple questions:

* Why is sqlite trying to write to a read-only database, exactly? And if we return SQLITE_READONLY, does it do something reasonable?
* You describe the previous behavior as "crashing" but it should be catching the exception and returning SQLITE_IOERR_WRITE. Where does the crash happen?

justin-mp:
> * You describe the previous behavior as "crashing" but it should be catching the exception and returning SQLITE_IOERR_WRITE. Where does the crash happen?

"crashing" is  the wrong word.  What I see is that the sqlite-test prints a bunch of exceptions and then exits with status code 3.  
[test-run.txt](https://github.com/cloudflare/workerd/files/14562665/test-run.txt) has the full details of what I see.  In the middle of all of that debug spew is `sentryErrorContext = sqliteVfsError`, so I believe the wrapper that should return SQLITE_IOERR_WRITE should be running.

It could be that all this change is doing is shutting up the tests.

> * Why is sqlite trying to write to a read-only database, exactly? And if we return SQLITE_READONLY, does it do something reasonable?

I see the following stack trace in the test:
```
workerd::SqliteDatabase::Vfs::FileImpl::$_16::operator()(sqlite3_file*, void const*, int, long long) const::'lambda'(kj::Exception const&)::operator()(kj::Exception const&) const at external/workerd/src/workerd/util/sqlite.c++:1302:5
 (inlined by) workerd::SqliteDatabase::Vfs::FileImpl::$_16::operator()(sqlite3_file*, void const*, int, long long) const at external/workerd/src/workerd/util/sqlite.c++:1302:5
 (inlined by) workerd::SqliteDatabase::Vfs::FileImpl::$_16::__invoke(sqlite3_file*, void const*, int, long long) at external/workerd/src/workerd/util/sqlite.c++:1300:13
sqlite3OsWrite at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:25698:10
 (inlined by) walCheckpoint at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:66496:14 **1**
 (inlined by) sqlite3WalCheckpoint at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:68464:14
sqlite3WalClose at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:66706:12
sqlite3PagerClose at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:60569:5
sqlite3BtreeClose at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:72654:5
sqlite3LeaveMutexAndCloseZombie at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:178939:7
sqlite3Close at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:178852:3
sqlite3_close at bazel-out/k8-dbg/bin/external/sqlite3/sqlite3.c:178895:51
workerd::SqliteDatabase::~SqliteDatabase() at external/workerd/src/workerd/util/sqlite.c++:364:14
```

The line marked by `**1**` in `walCheckpoint` is in this code:
```
     /* Iterate through the contents of the WAL, copying data to the db file */
      while( rc==SQLITE_OK && 0==walIteratorNext(pIter, &iDbpage, &iFrame) ){
        i64 iOffset;
        assert( walFramePgno(pWal, iFrame)==iDbpage );
        SEH_INJECT_FAULT;
        if( AtomicLoad(&db->u1.isInterrupted) ){
          rc = db->mallocFailed ? SQLITE_NOMEM_BKPT : SQLITE_INTERRUPT;
          break;
        }
        if( iFrame<=nBackfill || iFrame>mxSafeFrame || iDbpage>mxPage ){
          continue;
        }
        iOffset = walFrameOffset(iFrame, szPage) + WAL_FRAME_HDRSIZE;
        /* testcase( IS_BIG_INT(iOffset) ); // requires a 4GiB WAL file */
        rc = sqlite3OsRead(pWal->pWalFd, zBuf, szPage, iOffset);
        if( rc!=SQLITE_OK ) break;
        iOffset = (iDbpage-1)*(i64)szPage;
        testcase( IS_BIG_INT(iOffset) );
        rc = sqlite3OsWrite(pWal->pDbFd, zBuf, szPage, iOffset); // **2**
        if( rc!=SQLITE_OK ) break;
      }
 ```

where the line marked by `**2**` uses the error return about whether to continue iterating in the WAL.  Both the previous SQLITE_IOERR_WRITE and SQLITE_READONLY errors *should* trigger here.

Looking through the entire stack trace, SQLite doesn't whether or not you've opened the database in read-only mode on cleanup.  SQLite optimistically tries to clean things up (using SQLITE_CHECKPOINT_PASSIVE) and if it can't clean up anything, it gives up.

Long story short, this change is just shutting up the tests.  It's probably still a good idea to do though.

PR #1795, opened 2024-03-11
Extend Actor::shutdown() to allow shutting down in `HARD` mode

jp4a50:
Doing a `hard` shutdown aborts the IO context and therefore stops execution of all in-flight requests.


jasnell (approved):
only looking at the code change itself, LGTM. Should have someone more familiar with DO in general review also tho.

src/workerd/io/worker.h
```diff
@@ -673,21 +673,42 @@ class Worker::Actor final: public kj::Refcounted {
   // support publicly).
   void ensureConstructed(IoContext&);
 
-  // Forces cancellation of all "background work" this actor is executing, i.e. work that is not
-  // happening on behalf of an active request. Note that this is not a part of the dtor because
-  // IoContext objects prolong the lifetime of their Actor.
+
+  // Revokes access to storage and forces cancellation of all "background work" this actor is
```
MellowYarker, line 677: nit: Maybe this comment should be immediately above `shutdown()` below?

MellowYarker (approved):

src/workerd/io/worker.c++
```diff
@@ -3380,10 +3385,19 @@ void Worker::Actor::setIoContext(kj::Own<IoContext> context) {
...
   auto& limitEnforcer = context->getLimitEnforcer();
   impl->ioContext = kj::mv(context);
-  impl->metricsFlushLoopTask = impl->metrics->flushLoop(impl->timerChannel, limitEnforcer)
-      .eagerlyEvaluate([](kj::Exception&& e) {
-    LOG_EXCEPTION("actorMetricsFlushLoop", e);
+  impl->onShutdownTask = onShutdown().then(
+      [&context = KJ_ASSERT_NONNULL(this->impl->ioContext)](ShutdownContext shutdown) mutable {
+    if (shutdown.mode == ShutdownMode::HARD) {
+      context->abort(kj::mv(shutdown.error).orDefault(KJ_EXCEPTION(
+          FAILED, "broken.ignored; jsg.internal-error; actor was "
```
justin-mp, line 3392: Dumb question: is this the right error string to be using?  The string `internal-error` only appears here and src/workerd/jsg/exception.h doesn't seem to list internal-error as one of the `JSG_ERROR_` types.

justin-mp (approved):

src/workerd/io/worker.h
```diff
@@ -673,21 +673,42 @@ class Worker::Actor final: public kj::Refcounted {
...
 
+  struct ShutdownContext {
+    // The mode passed to `shutdown()`.
+    ShutdownMode mode;
+    // The error passed to `shutdown()`.
+    kj::Maybe<kj::Exception> error;
+  };
   // Get a promise that resolves when `shutdown()` has been called.
-  kj::Promise<void> onShutdown();
+  kj::Promise<ShutdownContext> onShutdown();
```
kentonv, line 711: I guess this is a little weird now -- what if shutdown() is called multiple times, first SOFT and then HARD? Maybe we need separate onSoftShutdown() and onHardShutdown()?

jp4a50: I think I'm going to shelve this PR now that we've decided we don't need to abort in-flight requests in any of the cases we discussed yesterday.

But it might be nice to revive it later since having a minimal set of shutdown modes that always leave the DO in a well-defined state regardless of what triggered the shutdown would probably make this stuff easier to reason about. And yeah I agree that, if I do revive it, I should probably handle hard shutdown coming after soft shutdown.

jp4a50:
Closing this since there's no immediate need for it.

PR #1797, opened 2024-03-11
Increase the timeout for storage operations before resetting an actor

a-robinson:
The lower timeout could plausibly be hit purely due to a busy CPU loop after issuing a storage operation. The new number matches the CPU limit per request. There's no need for this to be exactly the same number as that, but it seemed just as good a reason as I had for any other larger number.


MellowYarker (approved):

jasnell (approved):

a-robinson:
Rebased on latest main in order to ease merging into the internal repo.

PR #1799, opened 2024-03-11
Support for custom thenables in PromiseWrapper with compat flag

jasnell:
Adds a new compat flag that enables support for accepting custom thenables in the jsg `PromiseWrapper`.

Note that this is on received values only (e.g. `tryUnwrap(...)`.

### DIFF
--- a/src/workerd/jsg/promise.h
+++ b/src/workerd/jsg/promise.h
@@ -659,6 +659,12 @@ void thenUnwrap(const v8::FunctionCallbackInfo<v8::Value>& args) {
 template <typename TypeWrapper>
 class PromiseWrapper {
 public:
+  // The constructor here is a bit of a hack. The config is optional and might not be a JsgConfig
+  // object (or convertible to a JsgConfig) if is provided. However, because of the way TypeWrapper
+  // inherits PromiseWrapper, we always end up passing a config option (which might be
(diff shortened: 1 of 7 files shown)


src/workerd/jsg/promise.h
```diff
@@ -755,6 +756,17 @@ class PromiseWrapper {
...
     }
   }
+
+private:
+  const JsgConfig config;
+
+  static bool isThenable(v8::Local<v8::Context> context, v8::Local<v8::Value> handle) {
+    if (handle->IsObject()) {
+      auto obj = handle.As<v8::Object>();
+      return check(obj->Has(context, v8StrIntern(context->GetIsolate(), "then")));
```
kentonv, line 766: Probably not worth solving in this PR, but I worry about the cost of repeatedly looking up interned strings like this all over the place. I wonder if we should be caching them somewhere, similar to how we now cache `Symbol.dispose`.

kentonv (approved):

src/workerd/jsg/promise.h
jasnell: Yep, I've been wanting to do that for a while. In Node.js we maintain a registry of handles to commonly used strings and it has a big impact. We absolutely should be caching these.

jasnell:
Documentation PR: https://github.com/cloudflare/cloudflare-docs/pull/13411

PR #1800, opened 2024-03-11
Tweak awaitIoImpl some more.

kentonv:


src/workerd/io/io-context.h
```diff
@@ -983,15 +983,22 @@ jsg::PromiseForResult<Func, T, true> IoContext::awaitIoImpl(
...
+      // destroyed in unspecified order, so we wrap them in a struct to make it explicit.
+      Func func;
+      ExceptionOr<T> exceptionOrT;
+    };
+
     return run([resolver = kj::mv(resolver),
-                exceptionOrT = kj::mv(exceptionOrT),
-                maybeAsyncContext = kj::mv(maybeAsyncContext),
-                func = kj::fwd<Func>(func)](Worker::Lock& lock) mutable {
+                funcResultPair = FuncResultPair { kj::fwd<Func>(func), kj::mv(exceptionOrT) },
```
jasnell, line 994: A comment explaining this briefly would be helpful

jasnell: ha! nevermind... I wrote this then immediately spotted the comment immediately above. Sigh. Need more coffee.

jasnell (approved):

src/workerd/io/io-context.h
ObsidianMinor, line 994: I suppose this could also be done by explicitly moving the func and the value into the body of the function, assuming the function doesn't run twice.

ObsidianMinor: Oh but wait if the function is never called the func and value *won't* be moved into the body and so they could still fail to destruct properly.

ObsidianMinor (approved):

src/workerd/io/io-context.h
kentonv: Right, it's the never-called case that's an issue here.

PR #1801, opened 2024-03-11
Add separate enterStartupPython limiter specifically for Python

hoodmane:
Currently it isn't very different but this gives us a bit of additional flexibility for configuring Python startup limits.


dom96 (approved):
The `isPython` passing feels a bit fragile, but I suppose it's necessary.

Otherwise LGTM.

PR #1802, opened 2024-03-12
Add support for a storageClass option for R2 objects

bthwaites:
Adds an string option called storageClass for the R2 binding PutObject and CreateMultipartUpload methods. R2 will interpret this string to determine the storage class of the newly created object.

Adds a field to HeadResult, and by extension GetResult and ListResult, that R2 will use to return the storage class to the binding.

There is a corresponding internal PR for this change.

### DIFF
--- a/src/workerd/api/r2-api.capnp
+++ b/src/workerd/api/r2-api.capnp
@@ -105,12 +105,14 @@ struct R2PutRequest {
   sha256 @6 :Data $Json.hex;
   sha384 @7 :Data $Json.hex;
   sha512 @8 :Data $Json.hex;
+  storageClass @9 :Text;
 }
 
 struct R2CreateMultipartUploadRequest {
   object @0 :Text;
   customFields @1 :List(Record);
   httpFields @2 :R2HttpFields;
+  storageClass @3 :Text;
 }
 
 struct R2UploadPartRequest {
@@ -214,6 +216,10 @@ struct R2HeadResponse {
 
   checksums @8 :R2Checksums;
   # If set, the available checksums for this object
+
+  storageClass @9 :Text;
+  # The storage class of the object. Standard or Infrequent Access.
+  # Provided on object creation to specify which storage tier R2 should use for this object.
 }
 
 using R2GetResponse = R2HeadResponse;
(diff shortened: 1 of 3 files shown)


bthwaites:

src/workerd/api/r2-api.capnp
```diff
@@ -214,6 +216,9 @@ struct R2HeadResponse {
 
   checksums @8 :R2Checksums;
   # If set, the available checksums for this object
+
+  storageClass @9 :Text;
+  # The storage class of the object. Standard or Infrequent Access.
```
jasnell, line 221: nit: the comment could be expanded a bit more to explain what a "storage class" is

src/workerd/api/r2-bucket.h
```diff
@@ -171,27 +171,29 @@ class R2Bucket: public jsg::Object {
...
+    JSG_STRUCT(httpMetadata, customMetadata, storageClass);
     JSG_STRUCT_TS_OVERRIDE(R2MultipartOptions);
   };
 
   class HeadResult: public jsg::Object {
   public:
     HeadResult(kj::String name, kj::String version, double size,
                kj::String etag, jsg::Ref<Checksums> checksums, kj::Date uploaded, jsg::Optional<HttpMetadata> httpMetadata,
-               jsg::Optional<jsg::Dict<kj::String>> customMetadata, jsg::Optional<Range> range):
+               jsg::Optional<jsg::Dict<kj::String>> customMetadata, jsg::Optional<Range> range, kj::String storageClass):
```
jasnell, line 193: nit: can this be formatted so the lines do not exceed 100 chars?

src/workerd/api/r2-bucket.h
```diff
@@ -200,6 +202,7 @@ class R2Bucket: public jsg::Object {
     kj::String getHttpEtag() const { return kj::str('"', etag, '"'); }
     jsg::Ref<Checksums> getChecksums() { return checksums.addRef();}
     kj::Date getUploaded() const { return uploaded; }
+    kj::String getStorageClass() const { return kj::str(storageClass); }
```
jasnell, line 205: This can return `kj::StringPtr` instead to avoid the additional copying.

src/workerd/api/r2-api.capnp
```diff
@@ -214,6 +216,10 @@ struct R2HeadResponse {
 
   checksums @8 :R2Checksums;
   # If set, the available checksums for this object
+
+  storageClass @9 :Text;
```
jasnell, line 220: non-blocking: Are these expected to remain fixed? Just wondering if an enum wouldn't be better. not critical tho

bthwaites: We don't expect to remove any, but we might add more in the future, so a Text field makes that a little easier. I could see a case for validating the strings on the workerd side, though. For now I think we're ok with letting R2 interpret and validate the string.

jasnell:
Given that it's been about three weeks since CI ran on this, I'm going to running it again just to double check

jasnell:
@bthwaites ... can I ask you to rebase this PR on latest workerd/main? Just want to double check that nothing changed in the three weeks since reviewed. Haven't looked into why the macOS tests failed but maybe a rebase will fix it?

PR #1804, opened 2024-03-12
Fill out more of the node:util implementation

jasnell:
Handful of additional `node:util` APIs

### DIFF
--- a/src/node/util.ts
+++ b/src/node/util.ts
@@ -15,6 +15,12 @@ import {
 } from 'node-internal:validators';
 
 import {
+  debuglog,
+} from 'node-internal:debuglog';
+export const debug = debuglog;
+export { debuglog };
+
+import {
   ERR_FALSY_VALUE_REJECTION,
   ERR_INVALID_ARG_TYPE,
 } from 'node-internal:internal_errors';
@@ -249,6 +255,8 @@ export default {
   toUSVString,
   log,
   aborted,
+  debuglog,
+  debug,
   // Node.js originally exposed TextEncoder and TextDecoder off the util
   // module originally, so let's just go ahead and do the same.
   TextEncoder,
@@ -277,7 +285,6 @@ export default {
 
 // util._errnoException
 // util._exceptionWithHostPort
-// util.aborted
 // util.debug
 // util.debuglog
 // util.deprecate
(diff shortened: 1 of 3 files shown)


src/node/util.ts
```diff
@@ -183,18 +259,42 @@ export default {
...
-// getSystemErrorMap/getSystemErrorName -- libuv specific. No use in workerd?
-// is{Type} variants -- these are deprecated in Node. Use util.types
-// toUSVString -- Not clear how broadly this is used in the ecosystem outside of node.js.
-//                also this is soon to be obsoleted by toWellFormed in the language.
-// transferableAbortSignal/transferableAbortController -- postMessage and worker threads
-//      are not implemented in workerd. No use case for these.
+//   * util._errnoException
+//   * util._exceptionWithHostPort
+//   * util.getSystemErrorMap
+//   * util.getSystemErrorName
```
jasnell, line 284: The four above here relate closely to the `uv` module. There's likely no valid use case for these in workers outside of Just Make It Work compat.

src/node/util.ts
```diff
@@ -183,18 +259,42 @@ export default {
...
+//   * util.isFunction
+//   * util.isNull
+//   * util.isNullOrUndefined
+//   * util.isNumber
+//   * util.isObject
+//   * util.isPrimitive
+//   * util.isRegExp
+//   * util.isString
+//   * util.isSymbol
+//   * util.isUndefined
```
jasnell, line 300: These are all deprecated. I think we should only add them if someone specifically asks for it or we find a module we want to support that requires them.

fhanau:
Code looks good to me. Some observations:
- util.log() is deprecated.
- util.toUSVString is doc-only deprecated as of https://github.com/nodejs/node/pull/49725. Not sure if we want to keep those, but since it's not marked as deprecated in the main documentation we can implement it.
- deprecations.md in the Node docs also lists util.debug() as deprecated, but that's not reflected in https://nodejs.org/api/util.html#utildebugsection so adding it makes sense to me.
- deprecate() is a no-op as implemented here, glad we're having a TODO for it.
- Are we adding tests for this?

jasnell:
`util.log()`, `util.toUSVString()` are both still fairly common in the wild, so even with deprecation they make sense.

I'll see about adding some tests.

fhanau (approved):
Tests LGTM, having more would be nice to have but the attack surface of these functions is small so not required. Based on the number of commits consider combining some of them/using squash-merge.

PR #1805, opened 2024-03-12
Implement node:fs stubs

jasnell:
This just stubs out the node:fs module with implementations that throw. The intent here is to allow imports/requires to work without actually implementing the fs module.


jasnell:
Not sure if we're going to go forward with this or not. There's more to figure out. But if we do move forward with this approach, the idea is to have throwing stubs for *all* Node.js APIs we don't implement. Stay tuned.

jasnell:
We've decided not to go this direction, thankfully :-)

PR #1806, opened 2024-03-12
JSRPC: Support sending ReadableStream and WritableStream over RPC

kentonv:


src/workerd/api/streams/writable.c++
```diff
@@ -280,6 +282,137 @@ jsg::Ref<WritableStream> WritableStream::constructor(
...
+
+  auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
+      "WritableStream can only be serialized for RPC.");
+  auto externalHandler = dynamic_cast<RpcSerializerExternalHander*>(&handler);
+  JSG_REQUIRE(externalHandler != nullptr, DOMDataCloneError,
+      "WritableStream can only be serialized for RPC.");
+
+  IoContext& ioctx = IoContext::current();
+
+  // TODO(soon): Support JS-backed WritableStreams. Currently this only supports native streams
```
kentonv, line 372: @jasnell Is there an easy way to support this? Basically I need a `WritableStreamSink` representing the `WritableStream` so I can easily adapt it to a KJ stream. I think there are two options, when the `WritableStream` is not backed by a `WritableStreamSink`:

1. Create an `IdentityTransformStreamImpl`, which is always backed by native source/sink. Use the writable side of that, and separately pump the readable side into the JS stream.
2. Create an alternate version of `WritableStreamRpcAdapter`, above, which locks the isolate on every call and then calls into the JS Writer API.

Is there a better option?

src/workerd/api/streams/writable.c++
```diff
@@ -280,6 +282,137 @@ jsg::Ref<WritableStream> WritableStream::constructor(
...
+    doneFulfiller->fulfill();
+  }
+
+  // Returns a promise that resolves when the stream is dropped. If the promise is canceled before
+  // that, the stream is revoked.
+  kj::Promise<void> waitForCompletionOrRevoke() {
+    auto paf = kj::newPromiseAndFulfiller<void>();
+    doneFulfiller = kj::mv(paf.fulfiller);
+
+    return paf.promise.attach([weakRef = weakRef->addRef()]() mutable {
```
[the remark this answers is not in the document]
kentonv: Oh shoot, using `.attach()` here actually just attaches the lambda object but it will never be called. This is a bug, I meant to use `kj::defer()`. :facepalm: 

Ben C even tried to make this an error at one point but that change never got merged...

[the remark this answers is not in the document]
kentonv: 90847efe

src/workerd/api/streams/readable.c++
```diff
@@ -526,6 +529,169 @@ jsg::Optional<uint32_t> ByteLengthQueuingStrategy::size(
...
+  kj::Maybe<uint64_t> expectedLength;
+};
+
+}  // namespace
+
+void ReadableStream::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
+  // Serialize by effectively creating a `JsRpcStub` around this object and serializing that.
+  // Except we don't actually want to do _exactly_ that, because we do not want to actually create
+  // a `JsRpcStub` locally. So do the important parts of `JsRpcStub::constructor()` followed by
+  // `JsRpcStub::serialize()`.
```
jasnell, line 620: If the stream is already locked or disturbed, we shouldn't serialize it. Both of those conditions should be checked in here. I know that the `pumpTo(...)` method below will also check for that (and will leave the `ReadableStream` locked and disturbed) but these conditions need to be checked before setting up the rest of this.

src/workerd/api/streams/writable.c++
```diff
@@ -280,6 +282,137 @@ jsg::Ref<WritableStream> WritableStream::constructor(
...
+  }
+};
+
+}  // namespace
+
+void WritableStream::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
+  // Serialize by effectively creating a `JsRpcStub` around this object and serializing that.
+  // Except we don't actually want to do _exactly_ that, because we do not want to actually create
+  // a `JsRpcStub` locally. So do the important parts of `JsRpcStub::constructor()` followed by
+  // `JsRpcStub::serialize()`.
```
jasnell, line 362: Likewise here, this should check to make sure the `WritableStream` is not locked. If it is, this should throw before doing anything else here.

src/workerd/api/streams/writable.c++
jasnell: For short term, I would probably just have a `WritableStreamSink` implementation that wraps the `WritableStream` and writes directly to it's underlying controller, acquiring the isolate lock on every write. Using the `write(...)` that passes `kj::ArrayPtr<kj::Array<...>> pieces` would end up being the most performant way to write if you have multiple chunks. Longer term I'd like to explore more efficient ways of optimizing these kind of KJ-to-JS and JS-to-KJ pipelines for both this and the `PumpToAdapter` in JS-backed ReadableStream

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -259,6 +268,25 @@ export class MyService extends WorkerEntrypoint {
...
+  }
+
+  async readFromStream(stream) {
+    return await new Response(stream).text();
+  }
+
+  async returnReadableStream() {
+    let { readable, writable } = new IdentityTransformStream();
+    this.ctx.waitUntil(this.writeToStream(writable));
+    return readable;
```
jasnell, line 288: We need a test that verifies that a `readable` that is already locked and disturbed cannot be passed.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -832,6 +860,106 @@ export let serializeRpcPromiseOrProprety = {
...
+      let promise = env.MyService.writeToStream(writable);
+      let text = await new Response(readable).text();
+      assert.strictEqual(text, "foo, bar, baz!");
+      await promise;
+    }
+
+    // TODO(someday): Test JS-backed WritableStream, when it actually works.
+
+    // TODO(someday): Is there any way to construct an encoded WritableStream? Only system
+    //   streams can be encoded, but there's no API that returns an encoded WritableStream I think.
```
jasnell, line 877: What do you mean by "encoded WritableStream"?

src/workerd/api/streams/readable.c++
jasnell: Discussed in chat: instead of moving up the checks, comments should be added that we're depending on the existing checks in `pumpTo(...)` and `removeSink(...)` to catch the locked/disturbed states.

src/workerd/api/streams/writable.h
```diff
@@ -158,6 +158,12 @@ class WritableStream: public jsg::Object {
     });
   }
 
+  void serialize(jsg::Lock& js, jsg::Serializer& serializer);
+  static jsg::Ref<WritableStream> deserialize(
+      jsg::Lock& js, rpc::SerializationTag tag, jsg::Deserializer& deserializer);
+
+  JSG_SERIALIZABLE(rpc::SerializationTag::WRITABLE_STREAM);
```
jasnell, line 165: Non-blocking nit... can the `JSG_SERIALIZABLE` macro handle adding the `serialize` and `deserialize` boilerplate methods too... just something like `JSG_SERIALIZABLE(WritableStream, rpc::SerializationTag::WRITABLE_STREAM)`

src/workerd/api/tests/js-rpc-test.js
kentonv: I mean a `WritableStream` which wraps a `WritableStreamSink` which is `EncodedAsyncOutputStream` with a `StreamEncoding` other than `IDENTITY`.

The only place I know of where our API even returns a `WritableStream` is when constructing a transform stream and those never wrap `EncodedAsyncOuputStream`.

src/workerd/api/streams/writable.h
kentonv: Eh, I think it would be confusing if the method declarations are hidden in a macro, given that the definitions are not hidden.

Also, an upcoming change will make it so that these methods can have additional parameters (to request TypeHandlers).

src/workerd/api/tests/js-rpc-test.js
kentonv: 4f34626f

src/workerd/api/streams/writable.c++
kentonv: ff1c6fed

src/workerd/api/streams/common.h
```diff
@@ -216,6 +220,10 @@ class ReadableStreamSource {
   // directly might attempt to use the `IoContext` to call `registerPendingEvent()`.
   virtual kj::Promise<DeferredProxy<void>> pumpTo(WritableStreamSink& output, bool end);
 
+  // If pumpTo() pumps to a system stream, what is the best encoding for that system steram to
+  // use? This is just a hint.
+  virtual StreamEncoding getPrerferredEncoding() { return StreamEncoding::IDENTITY; };
```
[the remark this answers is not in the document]
kentonv: Wow I really copied that everywhere. 5e635f87

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -832,6 +860,118 @@ export let serializeRpcPromiseOrProprety = {
...
+    // Send a ReadableStream that is locked.
+    {
+      let { readable, writable } = new IdentityTransformStream();
+
+      let reader = readable.getReader();
+
+      assert.rejects(env.MyService.readFromStream(readable), {
+        name: "TypeError",
+        message: "The ReadableStream has been locked to a reader."
+      });
```
jasnell, line 920: Might be good to perform a `reader.read()` after this check to ensure that the state of the stream has not changed.

jasnell: Also... similar test for `writable`?

src/workerd/api/streams/readable.c++
```diff
@@ -526,6 +529,173 @@ jsg::Optional<uint32_t> ByteLengthQueuingStrategy::size(
...
+
+  auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
+      "ReadableStream can only be serialized for RPC.");
+  auto externalHandler = dynamic_cast<RpcSerializerExternalHander*>(&handler);
+  JSG_REQUIRE(externalHandler != nullptr, DOMDataCloneError,
+      "ReadableStream can only be serialized for RPC.");
```
jasnell, line 626: Is there any concern about serializing a `ReadableStream` (or `WritableStream`) that was received on the other side... basically we'd end up with multiple layers of these adapters in that case, which seems fine on the surface but I want to make sure we're not missing anything.

In particular, if my RPC server-side does something silly like:

```
function roundTrip(writable) {
  return writable;
}
```

Awfully silly, yes, but possible. Should it be?

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -832,6 +860,118 @@ export let serializeRpcPromiseOrProprety = {
...
+      await assert.rejects(promise, {
+        name: "Error",
+        // TODO(someday): Propagate the actual error.
+        message: "ReadableStream received over RPC disconnected prematurely."
+      });
+    }
+
+    // Send fixed-length ReadableStream.
+    {
+      let { readable, writable } = new FixedLengthStream("foo, bar, baz!".length);
```
jasnell, line 951: non-blocking: a test that passes the `readable` or `writable` as a nested property of some other object would be good also here. This would demonstrate that the value serializer / deserializer delegate is properly handling nested custom api types. If it didn't, that would more signal an issue in v8 than our code, but certainly a good thing to verify.

Also, it strikes me that someone could create a `TransformStream` on one side of the RPC channel, and pass both the `readable` and `writable` to the other side such that transformations are occurring across the RPC boundary but the main stream operations are happening locally... e.g.

```
const { readable, writable } = env.MyService.GetTransformer();
```

That's actually pretty nice.

src/workerd/api/streams/readable.c++
kentonv: It should be fine. Worst-case, bytes have to proxy through the remote end. Best case, Cap'n Proto path shortening actually kicks in to avoid that.

src/workerd/api/tests/js-rpc-test.js
jasnell: Ah, I think your test would have to construct it. We do have `CompressionStream` which would make for a good test, but not quite what you're asking for here.

src/workerd/api/tests/js-rpc-test.js
kentonv: 39ae6632

src/workerd/api/streams/readable.c++
kentonv: Added test, caught bug: 15227332

src/workerd/api/tests/js-rpc-test.js
kentonv: There are a number of existing test cases already for nested properties (using RpcTarget, functions, etc.). However, a test that passes multiple ReadableStreams at the same time is a good idea, to make sure the right indexes are passed to StreamSink. And indeed this caught a bug: 26f332f0

jasnell:
Overall this is LGTM so far. Looks like there's a relevant test failure to investigate. Do you want to try to get the JS-backed WritableStream support included in this PR? If so, I'll hold off on a final review check and sign off until that's in. Otherwise, I'll give it a final once over once the failing test is figured out.

kentonv:
I'm going to punt on pure-JS WritableStream for now. I think ReadableStream is the far more common use case anyway.

The test failure is frustrating, there's no logs and it doesn't repro locally. :/ Will try to figure it out tomorrow I guess.

kentonv:
Why is the build suddenly taking hours? Argh.

kentonv:
The last two commits are new, the rest is a rebase.

ce0682f - Refactor `StreamSinkImpl` to avoid using a promise for `resultStreamSink`, which should perform better.
901a3b4 - Move the real-socket tests into a separate test where info logging is turned off. I think that the test failures may be related to the excessive stack trace symbolication for info logs, hoping this helps with that.

kentonv:
Yay the failing test passed! But I broke something else. Two last fixups:

* a05bec8f
* 4879b037

jasnell (approved):

PR #1809, opened 2024-03-12
Make SqliteDatabase write mode a kj::Maybe

justin-mp:
I have a use case for SqliteDatabase where I need to parameterize whether the database is read-only or read/write.  Having two constructors for these cases and no default constructor means that I cannot parameterize the read/write-mode of the SqliteDatabase in the constructor of another class.

It’s much easier to use this if we can indicate we want read-only mode using a parameter.  I considered using a WriteMode of 0 to mean “read-only”, but the kj::WriteMode documentation says:

> (To open a file or directory read-only, do not specify a mode.)

The best way to “not specify a mode” as a parameter is to make it a kj::Maybe.

I made the write mode an optional parameter so that to make the new interface compatible the previous one.


src/workerd/util/sqlite.c++
```diff
@@ -316,43 +316,41 @@ static constexpr PragmaInfo ALLOWED_PRAGMAS[] = {
...
-    if (kj::has(mode, kj::WriteMode::CREATE_PARENT) && path.size() > 1) {
-      // SQLite isn't going to try to create the parent directory so let's try to create it now.
-      vfs.directory.openSubdir(path.parent(),
-          kj::WriteMode::CREATE | kj::WriteMode::MODIFY | kj::WriteMode::CREATE_PARENT);
+    KJ_IF_SOME(rootedPath, vfs.tryAppend(path)) {
+      // If we can get the path rooted in the VFS's directory, use the system's default VFS instead
+      // TODO(bug): This doesn't honor vfs.options. (This branch is only used on Windows.)
+      SQLITE_CALL_NODB(sqlite3_open_v2(rootedPath.toString().cStr(), &db, flags, nullptr));
+    }
+    else {
```
kentonv, line 339:
```suggestion
    } else {
```

src/workerd/util/sqlite.c++
```diff
@@ -316,43 +316,41 @@ static constexpr PragmaInfo ALLOWED_PRAGMAS[] = {
...
-    SQLITE_CALL_NODB(sqlite3_open_v2(path.toString().cStr(), &db,
-                                    flags, vfs.getName().cStr()));
+  else {
+    KJ_IF_SOME(rootedPath, vfs.tryAppend(path)) {
+      // If we can get the path rooted in the VFS's directory, use the system's default VFS instead
+      // TODO(bug): This doesn't honor vfs.options. (This branch is only used on Windows.)
+      SQLITE_CALL_NODB(
+          sqlite3_open_v2(rootedPath.toString().cStr(), &db, SQLITE_OPEN_READONLY, nullptr));
+    }
+    else {
```
kentonv, line 350:
```suggestion
    } else {
```

src/workerd/util/sqlite.c++
```diff
@@ -316,43 +316,41 @@ static constexpr PragmaInfo ALLOWED_PRAGMAS[] = {
...
-
-  KJ_IF_SOME(rootedPath, vfs.tryAppend(path)) {
-    // If we can get the path rooted in the VFS's directory, use the system's default VFS instead
-    // TODO(bug): This doesn't honor vfs.options. (This branch is only used on Windows.)
-    SQLITE_CALL_NODB(sqlite3_open_v2(rootedPath.toString().cStr(), &db,
-                                    flags, nullptr));
-  } else {
-    SQLITE_CALL_NODB(sqlite3_open_v2(path.toString().cStr(), &db,
-                                    flags, vfs.getName().cStr()));
+  else {
```
kentonv, line 343: Need to fix this too. (Github won't let me propose a fix here since the `}` on the previous line is now considered new.)

kentonv (approved):
I guess that makes sense. FWIW, I had done it this way for consistency with the KJ filesystem APIs, which similarly have methods like `openFile()` which have read-only and writable overloads, based on whether you provide a `WriteMode`. But in the case of `openFile`, the two overloads actually return different types (`ReadableFile` vs `File`), so it really does have to be an overload in that case. Here, using `Maybe` is fine.

Whitespace nitpick: It looks like you have several instances of `}\n  else {` that should be `} else {` instead.

src/workerd/util/sqlite.c++
justin-mp: Fixed in d415067a8fa3833862aaecc6d31b7e2a680f9479.

src/workerd/util/sqlite.c++
justin-mp: Fixed in d415067a8fa3833862aaecc6d31b7e2a680f9479.

src/workerd/util/sqlite.c++
justin-mp: Fixed in d415067a8fa3833862aaecc6d31b7e2a680f9479.

justin-mp:
> Whitespace nitpick: It looks like you have several instances of `}\n else {` that should be `} else {` instead.

That'll teach me to run `clang-format` on my changes in workerd.

PR #1810, opened 2024-03-12
Add discussion category and template for Python packages

irvinebroque:
Creates a template for submitting requests to support additional Python packages in the Workers Runtime.

- Will show up here: https://github.com/cloudflare/workerd/discussions/new/choose
- This will be directly linkable, so we can direct people to it from error states cc @garrettgu10 
- Will show up as its own discussion category so that we can easily filter for just this.


garrettgu10:
Maybe also add an optional field for version?

dom96 (approved):
Nice!

garrettgu10 (approved):

PR #1811, opened 2024-03-12
New Pyodide package bundle including aiohttp

garrettgu10:
I generated a new package bundle using a GitHub action and this one includes aiohttp. I will make an upstream PR to make sure this doesn't break anything.


garrettgu10:
oops, this is on my fork and I think I need to make it on the main repo

PR #1814, opened 2024-03-12
Coalesce reads in Actor Cache flush

MellowYarker:


MellowYarker:
Huh... [my branch](https://github.com/cloudflare/workerd/compare/main...milan/actor-cache-batch-reads) has 13 commits but the PR pointing to the same branch has 12...

> Exciting! Let me know if/when you'd like me to take a look at this.

There's still 1 failing test (`ActorCache list stream cancellation`) but definitely feel free to read through @a-robinson. Admittedly my commits could probably just be fixups on Ben's but I felt it was better to append than overwrite since I wasn't familiar with the code.

I just tossed in [one more commit](https://github.com/cloudflare/workerd/commit/035450664ec894cc6d26540e096d7c4f612652fa) to try and (at least partially) fix that failing test, but it doesn't seem to be doing much. Other than that, I think I may try to add some more comments and do a bit more refactoring.

MellowYarker:
https://github.com/cloudflare/workerd/compare/035450664ec894cc6d26540e096d7c4f612652fa..a529139f3b54dacedd1b6b77b0b3bebe2babc63d 

^ tries to make it a bit easier to track when `Entry`s are added to the lists and when we're transitioning `EntrySyncStatus` states.

MellowYarker:
https://github.com/cloudflare/workerd/compare/a529139f3b54dacedd1b6b77b0b3bebe2babc63d..4feb0f11e12a52065c9f38936ff165d4af9c622e was a rebase to fix conflicts. I accidentally pulled 1 extra commit so the second force push just gets us back to this feature branch on top of  `main`.

src/workerd/io/actor-cache.h
```diff
@@ -319,23 +319,10 @@ class ActorCache final: public ActorCacheInterface {
...
-    // DIRTY. If the write completes successfully and the state is still FLUSHING, then the state
-    // can progress to CLEAN.
-    //
-    // The entry remains in `dirtyList` while in this state. Since newly-dirty entries are always
-    // appended to `dirtyList`, it's guaranteed that all `FLUSHING` entries in `dirtyList` come
-    // before all `DIRTY` entries.
-    //
-    // Next state: CLEAN (if the flush completes) or NOT_IN_CACHE (if a new put()/delete()
-    // overwrites the entry first).
-    FLUSHING,
```
a-robinson, line 337: Do you understand the motivation for this committ, @MellowYarker ? The commit message gives no explanation of why removing the `FLUSHING` state makes entry removal easier or why the `FLUSHING` state isn't needed anymore, and I don't understand why we'd make this kind of substantive change without a clear reason -- it's the sort of thing that leads to nasty regressions if not very carefully reviewed.

src/workerd/io/actor-cache.c++
```diff
@@ -2427,10 +2419,25 @@ kj::Promise<void> ActorCache::flushImpl(uint retryCount) {
...
+  // disconnect).
+  //
+  // In practice, most code probably will not have any reads in flight when a flush occurs.
+  //
+  // Note that we have cached strong references to all entries we intend to mutate above. This means
+  // that we can be confident that flushing the cached set will not conflict with future reads
+  // because:
+  // - All our cached entries are dirty.
+  // - Dirty entries can only be removed from the cache map if replaced by a new dirty entry.
+  // - Thus all new read requests for our cached entries keys will be served from cache.
```
a-robinson, line 2533: Is this actually true? Are list requests necessarily served from cache? What if there's a scenario like:

1. Some read starts
2. Write key b = 1
3. A read attempts to list a range including b
4. Write key b= 2
5. The write makes it to cockroach prior to the read and commits

Is it possible for the list to return key b=2? To be clear, I don't know without spending 10+ minutes reading the code, but my understanding is that that sort of thing is exactly what the logic here was meant to prevent (as explained by [the comment that was modified and moved up here](https://github.com/cloudflare/workerd/pull/1814/commits/b3de4131f0b52695c6d0a44a1336a522983c9f29#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eL2838)).

a-robinson: Ideally the answer to this question would be a test case proving this case wrong. But failing that, how confident are you in it? Do I need to check? Is this commit actually important to the broad goal?

Sorry for being picky here, but we've been burned many times now by the practice of including a bunch of random refactors/cleanups that aren't really needed in an unrelated change.

src/workerd/io/actor-cache-test.c++
```diff
@@ -950,7 +948,7 @@ KJ_TEST("ActorCache coalescing") {
         .withParams(CAPNP(keys = ["grault"]))
         .thenReturn(CAPNP(numDeleted = 0));
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["garply", "fred", "waldo"]))
```
a-robinson, line 953: Any idea what's with all these random reorderings of keys, both in this commit and in bb39bf72713bf13b7d781864de3a43ea014ade55?

src/workerd/io/actor-cache.h
a-robinson: And as a TODO for myself, I need to come back and review this commit for real if it really is important, since I haven't yet.

src/workerd/io/actor-cache-test.c++
```diff
@@ -1166,9 +1164,6 @@ KJ_TEST("ActorCache flush retry") {
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("corge"))) == "555");
   KJ_ASSERT(expectCached(test.get("grault")) == nullptr);
 
-  // Although the transaction didn't complete, the delete did, and so it resolves.
```
a-robinson, line 1176: If this promise definitely hasn't resolved in the new version of the code, can we encode that via a `poll()` like we've done below for `promise2`?

Frankly it looks worrying to me that `promise1` used to resolve and return 1 here -- we shouldn't have been trusting the results from a non-committed delete within a transaction, although perhaps we were cheating a bit and trusting it due to our knowledge that there's only ever one writer to a given actor's data at a time.

src/workerd/io/actor-cache-test.c++
```diff
@@ -1182,10 +1177,10 @@ KJ_TEST("ActorCache flush retry") {
     // last time, because it hasn't been further overwritten, and that delete from last time
     // wasn't actually committed.
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["grault", "corge"]))
+        .withParams(CAPNP(keys = ["corge", "grault"]))
         .thenReturn(CAPNP(numDeleted = 2));
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["quux", "baz"]))
```
a-robinson, line 1195: Hm, why aren't we including `quux` here anymore? The logic in the comment above still seems correct to me -- we need to ensure the delete for it is still sent because the previous attempt at deleting it never committed.

This looks like a bug, unless I'm misunderstanding.

a-robinson: TODO for me -- review the rest of this commit more closely after this question has been sorted out

a-robinson (commented):
Just a partial review of the first handful of commits for now, sorry. I'll continue reviewing the rest, although have a couple questions that could use a response as well.

MellowYarker:
> Just a partial review of the first handful of commits for now, sorry. I'll continue reviewing the rest, although have a couple questions that could use a response as well.

Thanks Alex, was hoping to get to some of the comments today but I noticed there was a new segfault. Thought something broke during the rebase, but it was just a bad commit. This took way longer to figure out than it should've :upside_down_face: 

https://github.com/cloudflare/workerd/compare/d673ea074b7970362c48ea38034359757727c6c1..70a2d4637607cc2dc5028d723acd4dcaa975c20e

a-robinson (commented):
And it looks like you're aware of it, but there are a bunch of TODO(now) comments to work out :)

src/workerd/io/actor-cache-test.c++
MellowYarker: This is a great call out, I think you're right.

Both before and after this change, the test is supposed to have the delete "succeed" (returning the count) and the transaction fail. We save the delete count [here](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2828), then because there's nothing left to do we mark the `CountedDelete` as having finished [right below](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2834).

Since the flush fails, we try again, but as we go through our `countedDeletes` we [check for those that have already finished](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2318-R2322) and skip them since they don't need to be deleted anymore. However, I suppose this violates the transaction assumptions, so instead we should probably be marking all the `CountedDelete`s as having been successfully deleted only after all the promises associated with the transaction [have completed](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2895).

Will give it a try in the morning, I'm also curious to see if this would have consequences for `qux`, since we overwrite it with a `put` after the delete and I'm not totally sure what the expected behavior is. It looks like we shouldn't try to include it in the retried `CountedDelete` (judging by the test), but I suspect the current code doesn't handle that properly.

MellowYarker: https://github.com/cloudflare/workerd/pull/1814/commits/6d15b89f9ba8ec8395582196db46dd54a2260f44

I think this should address this particular issue, `quux` is now retried, though `qux` is not (since we overwrote it _and_ the first delete succeeded in giving us the count).

I've left a few additional TODOs for myself to think a bit more about, but I just want to get this off my local machine for now.

src/workerd/io/actor-cache-test.c++
MellowYarker: The "make entry removal easier" commit slightly changed `deleteAll()` https://github.com/cloudflare/workerd/pull/1814/commits/bb39bf72713bf13b7d781864de3a43ea014ade55#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR1989-R1993 so that now we just pick `Entry`s off the dirty list.

The reordering as seen from mock storage makes sense, because our dirtyList should look something like:
`foo, bar, grault, baz, garply` (where `foo` is least recent) prior to the `deleteAll` [here](https://github.com/cloudflare/workerd/pull/1814/commits/bb39bf72713bf13b7d781864de3a43ea014ade55#diff-80fa4859170178cc35404a2937001fc86b9430cada8aca5fe1cc3e3494509622R729).

We then switch it back in a later commit ([Make CountedDelete Simpler](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-80fa4859170178cc35404a2937001fc86b9430cada8aca5fe1cc3e3494509622R753)) because rather than [giving an `Entry`](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eL2129) a `CountedDelete` and adding it to the `CountedDelete`s entries [when we're flushing](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eL2370), we just [immediately assign](https://github.com/cloudflare/workerd/pull/1814/commits/1669f0d0e8c8beceeac378adfbc6f1d164a7504b#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2111-R2112) the `Entry` to the `CountedDelete`s entries in `putImpl()`. This essentially preserves the ordering as seen from the JS API, rather than whatever order things are in within the dirtyList.

src/workerd/io/actor-cache.h
```diff
@@ -728,7 +728,9 @@ class ActorCache final: public ActorCacheInterface {
 
   // If the LRU is currently over the soft limit, returns a promise that resolves when it is
   // back under the limit.
-  kj::Maybe<kj::Promise<void>> getBackpressure();
+  kj::Maybe<kj::Promise<void>> flushIfNeeded();
```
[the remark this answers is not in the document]
MellowYarker: I agree with you

https://github.com/cloudflare/workerd/pull/1814/commits/d18886e8a9781d7f0a1810f16432584c065a0b2c
https://github.com/cloudflare/workerd/pull/1814/commits/c90d9f97bada482967f26c55aeba64a7bd8a3e29

src/workerd/io/actor-cache.c++
```diff
@@ -1868,7 +1868,7 @@ kj::OneOf<bool, kj::Promise<bool>> ActorCache::delete_(Key key, WriteOptions opt
     // fulfiller.
     auto paf = kj::newPromiseAndFulfiller<uint>();
     countedDelete->resultFulfiller = kj::mv(paf.fulfiller);
-    KJ_IF_SOME(p, getBackpressure()) {
+    KJ_IF_SOME(p, flushIfNeeded()) {
```
MellowYarker: Oops, looks like this evaded me because it changed in the counted delete commit. Will fix.

src/workerd/io/actor-cache.h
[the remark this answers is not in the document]
MellowYarker: I've dropped the first commit entirely now.

src/workerd/io/actor-cache.h
MellowYarker: > Do you understand the motivation for this commit

No, just seems like a cleanup to consolidate removing from clean and dirtylist in one method.

I'm not too familiar with the `FLUSHING` state but it looks like its not really checked much, maybe only as part of `isDirty()` calls. That said, I notice we aren't ever setting `flushStarted` to false after its been set to true (only on instantiation of an `Entry`). Not clear yet if that's intentional, though I suspect it's not since we'd previously move from `FLUSHING` to `CLEAN/NOT_IN_CACHE`. It may be the case that it's worth keeping the `FLUSHING` state, although I know that uncached `get()`s are set to `DIRTY` and put in the `dirtyList` now, so it might be tricky. Will look into it.

src/workerd/io/actor-cache-test.c++
```diff
@@ -4653,18 +4805,24 @@ KJ_TEST("ActorCache transaction multiple counted delete batches") {
...
+      .withParams(CAPNP(keys = ["count1", "count4"]), "stream"_kj)
       .useCallback("stream", [&](MockClient stream) {
     // Pretend that "count1" already exists but was not in the cache.
     stream.call("values", CAPNP(list = [(key = "count1", value = "1")]))
         .expectReturns(CAPNP(), ws);
     stream.call("end", CAPNP()).expectReturns(CAPNP(), ws);
-  }).expectCanceled();
+  }).thenReturn(CAPNP());
+
+  // TODO(now): I'm not fully sure I understand why this is a separate getMultiple?
```
a-robinson, line 4816: Yeah this is very odd. Although it's interesting that it isn't an entirely new thing either.

a-robinson (commented):
Finished reviewing all test changes, will look at the code changes now.

Sorry for some of the comments being more questions and/or things I could poke at myself once my devspace is usable again, feel free to breeze past any such comments that aren't useful.

src/workerd/io/actor-cache.h
a-robinson: Hm, removing `FLUSHING` from the list of states but adding a new `flushStarted` bool that has to be checked separately from the state doesn't really feel like a simplification, but I guess I should reserve judgment until I see how it's used by the later commits.

Although after reading through this commit in its entirety, I'm pretty suspicious of the risks that it introduces given that the benefit isn't obvious yet. And regarding the benefits, we still reference "flushing" entries in a bunch of places, it's just that now it's lower-case "flushing" rather than an explicit FLUSHING state.

src/workerd/io/actor-cache.c++
```diff
@@ -1960,17 +1983,15 @@ ActorCache::DeleteAllResults ActorCache::deleteAll(WriteOptions options) {
     auto& map = currentValues.get(lock);
 
     kj::Vector<kj::Own<Entry>> deletedDirty;
+    for (auto& entry: dirtyList) {
+      // Preserve the dirty list so we can run it before our delete all. Note that it's not a given
```
a-robinson, line 2073: > Note that it's not a given that all dirty list entries are in the map.

So is this a bug fix given that we previously only iterated over the map? If so, what was the bug and why didn't we add a test reproducing it? Or was the code fine before and this "note" only applies to the code *after* the changes in this commit?

src/workerd/io/actor-cache.c++
```diff
@@ -2047,6 +2068,14 @@ void ActorCache::putImpl(Lock& lock, kj::Own<Entry> newEntry,
           return;
         }
 
+        // Entry may have `countedDelete` indicating we're still waiting to get a count from a
```
a-robinson: Huh, I'm pretty suspicous of this code movement -- there's nothing saying that a `PRESENT` value can't have a `countedDelete` attached. In fact, this comment covers one such scenario for how that's possible -- a counted delete followed by a `put()` will leave the entry in the `PRESENT` state but with a non-null `countedDelete` attached. And the if another `put` came in for the same key, it looks like we'd hit the `PRESENT` case above and fail to carry forward the counted delete?

I also just don't understand the motivation for this code movement. What does it have to do with the rest of the commit?

src/workerd/io/actor-cache.c++
```diff
@@ -291,8 +301,21 @@ void ActorCache::touchEntry(Lock& lock, Entry& entry) {
 }
 
 void ActorCache::removeEntry(Lock& lock, Entry& entry) {
-  KJ_IASSERT(!entry.isDirty());
```
a-robinson, line 296: Hm, I kind of appreciated that we had some protections against accidentally removing dirty entries (and ensuring their `countedDelete` is carried forward) rather than happily doing so silently here.

src/workerd/io/actor-cache.c++
a-robinson: Never mind, I take this line of reasoning back now that I've refreshed myself on the code again. The `list()` implementation basically grabs a snapshot of the current state of the cache before sending the list() request to storage, and when it gets results back it merges those results with the saved values from when it was first called, which should avoid this problem.

In other words, this commit (`Refactor: Streamline actor cache flush transaction usage`) looks fine.

src/workerd/io/actor-cache.h
MellowYarker: So I tried the naive approach of switching back to `FLUSHING` and dropping `flushStarted`, and ran into issues with the **ActorCache deleteAll() during transaction commit** test. Specifically, following `put(bar, 456)` we do a `deleteAll()`, and that `deleteAll()` is registered before we get around to flushing that `put(bar, 456)`. This means when we get around to attempting the `startFlushTransaction()`, we have already `requestedDeleteAll`.

In fact, the prior flush hasn't even finished yet so [we modify](https://github.com/cloudflare/workerd/blob/9651cc9b814b20d4674f1b0cba2e83ebbfbc8c38/src/workerd/io/actor-cache.c%2B%2B#L2484-L2496) our `deletedDirty` collection (so it only contains `bar`), which ensures we'll flush that `put(bar)` before doing the `deleteAll()`.

This all kind of breaks when we switch back to having a `FLUSHING` state, since `deleteAll()` calls `ActorCache::clear()`, which calls `removeEntry()` on all cached entries in the map, and `removeEntry()` (now) sets the `EntrySyncStatus` to `NOT_IN_CACHE`. Since we're no longer tracking the entry as waiting to be flushed (via `flushStarted`), we assume it has already been flushed and the `put()` is abandoned in favour of the `deleteAll()`.

This should be fixable, although it might make tracking `NOT_IN_CACHE` a bit less clear.

src/workerd/io/actor-cache.c++
```diff
@@ -2843,28 +2853,25 @@ kj::Promise<void> ActorCache::flushImplUsingTxn(
     };
 
     for (auto& promise : promises) {
-      // Reuse `countDeleted` since it's already in a state object anyway.
-      rpcCountedDelete.countedDelete->countDeleted += co_await promise;
+      // TODO(now): Note to self - Are we absolutely, positively certain that we will always come
```
a-robinson, line 2978: I agree this looks likely to be buggy. Specifically, it looks most suspicous in the case where there are multiple `rpcDeletes` attached to the `rpcCountedDelete` and the first one succeeds but a latter one fails. There may be other scenarios where a retry could cause problems too, but that one jumps out.

It'd be interesting to try to write a test case to break this. I think it could be fixed by accumulating the counts from all the promises in a local variable first, and only adding them to `rpcCountedDelete.countedDelete->countDeleted` when they've all resolved successfuly. And then we could add them to the count at the same time that we set `rpcCountedDelete.countedDelete->completedInTransaction` (I don't actually see why we need to set that in callback below rather than immediately before returning from this lambda), which would make this look much less scary.

src/workerd/io/actor-cache.h
a-robinson: We can go either way on this. Stylistically it doesn't seem like a clear improvement, but it's not a fundamentally bad approach. And it comes with some risks (as called out in my other comments on the commit), but if it simplifies the later changes in this PR then that's a solid point in its favor.

At this point I see it mostly as a matter of picking which battle to take on -- either making sure the things I commented on aren't actually an issue (and/or more surgically undoing the parts that are) or figuring out how to rip out this commit from the stack of changes.

> and removeEntry() (now) sets the EntrySyncStatus to NOT_IN_CACHE

It doesn't if we were to just rip out the whole commit, but I'm assuming that's difficult due to how later commits build on it?

src/workerd/io/actor-cache.h
```diff
@@ -489,19 +487,69 @@ class ActorCache final: public ActorCacheInterface {
   // This object can only be manipulated in the thread that owns the specific actor that made
   // the request. That works out fine since CountedDelete only ever exists for dirty entries,
   // which won't be touched cross-thread by the LRU.
-  struct CountedDelete: public kj::Refcounted {
+  class CountedDeleteWaiter;
```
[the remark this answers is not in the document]
MellowYarker: Should've done a fixup on the commit directly but as part of [a cleanup commit](https://github.com/cloudflare/workerd/pull/1814/commits/adde09a4e80ee42a740dfdb9426931ff9e21b909#diff-a40aec54cc202afdf5b713427d52537f7835238e1d035e334b66d0c93c8d96aaL495) I actually fixed this and https://github.com/cloudflare/workerd/pull/1814#discussion_r1540390140 below. Will just fix the original commits.

src/workerd/io/actor-cache-test.c++
```diff
@@ -1427,70 +1413,122 @@ KJ_TEST("ActorCache read retry on flush containing only puts") {
   KJ_ASSERT(KJ_ASSERT_NONNULL(promise.wait(ws)) == "123");
 }
 
-KJ_TEST("ActorCache read hard fail") {
+// KJ_TEST("ActorCache read hard fail") {
```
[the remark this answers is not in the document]
MellowYarker: Since we've moved reads to flush along with writes, if the read fails, it looks like we're actually [breaking the output gate](https://github.com/cloudflare/workerd/pull/1814/files#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2751-R2752) :grimacing:.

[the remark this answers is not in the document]
a-robinson: 😬

src/workerd/io/actor-cache.c++
a-robinson: > I would expect at least a clear example where a dirty entry isn't in `currentValues`

Going by the comments in the header file, that shouldn't be possible:

https://github.com/cloudflare/workerd/blob/2bf2d26a9ba5dd78868a25edb96f50b73c2740f9/src/workerd/io/actor-cache.h#L345-L346

a-robinson: But your explanation makes sense, it seems that we're ok here and just need to clarify the comment.

src/workerd/io/actor-cache.c++
```diff
@@ -2055,8 +2042,19 @@ void ActorCache::putImpl(Lock& lock, kj::Own<Entry> newEntry,
...
-        KJ_IF_SOME(c, countedDelete) {
-          // Overwrote an entry that was in cache, so we can count it now.
+        KJ_IF_SOME(c, maybeCountedDelete) {
+          // TODO(now): It's not clear to me that there's any benefit in counting the delete early,
+          // we have to go to storage regardless, and the counted delete promise would not resolve
+          // until we flush successfully anyways. I don't think we should do this, but maybe I just
+          // don't fully understand it?
+          //
+          // TODO(now): Is the implication of the comment below that we're not considering this key
+          // anymore in the CountedDelete's `entries`? It's not clear to me that this is actually
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1915/commits/694b9515a38e8b8ac2a330c216362cc5aa76fa5d#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2044-R2046

src/workerd/io/actor-cache.c++
MellowYarker: https://github.com/cloudflare/workerd/pull/1915/commits/694b9515a38e8b8ac2a330c216362cc5aa76fa5d#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2841

src/workerd/io/actor-cache-test.c++
```diff
@@ -1215,6 +1194,9 @@ KJ_TEST("ActorCache flush retry") {
 
   // Second delete finished this time.
   KJ_ASSERT(promise2.wait(ws) == 2);
+
+  // Although the transaction didn't complete, the delete did, and so it resolves.
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1915/commits/d9e483993affc8381c1f33c4844d5cfd2bf92b5d

src/workerd/io/actor-cache.h
```diff
@@ -489,19 +487,69 @@ class ActorCache final: public ActorCacheInterface {
...
-    // When `countOutstanding` reaches zero, fulfill this with `countDeleted`.
-    kj::Own<kj::PromiseFulfiller<uint>> resultFulfiller;
+    // Did this particular counted delete succeed within a transaction? In other words, did we
+    // already get the count? Even if we got the count, we may need to retry if the transaction
+    // itself failed, though we won't need to get the count again.
+    bool completedInTransaction = false;
+
+    // Did this particular counted delete succeed? Note that this can be true even if the flush
+    // failed on a different batch of operations.
+    bool isFinished = false;
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1915/commits/91cd1147b2fd6098095981234882b9ad043d38ab#diff-a40aec54cc202afdf5b713427d52537f7835238e1d035e334b66d0c93c8d96aaL517

We did need the default ctor, but not the dtor.

src/workerd/io/actor-cache.h
```diff
@@ -489,19 +487,69 @@ class ActorCache final: public ActorCacheInterface {
...
+    kj::Promise<void> forgiveIfFinished(kj::Promise<void> promise) try {
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1915/commits/91cd1147b2fd6098095981234882b9ad043d38ab#diff-a40aec54cc202afdf5b713427d52537f7835238e1d035e334b66d0c93c8d96aaR518-R529

I started seeing this `fn() try {} catch (...) {}` a few months ago, honestly felt a bit weird to me too, I prefer seeing the entire function contained in a block.

src/workerd/io/actor-cache.h
```diff
@@ -431,6 +431,12 @@ class ActorCache final: public ActorCacheInterface {
       }
     }
 
+    // This Entry is part of a CountedDelete, but has since been overwritten via a put().
+    // This is really only useful in determining if we need to retry the deletion of this entry from
+    // storage, since we're interested in the number of deleted records. If we already got the count,
+    // we won't include this entry as part of our retried delete.
+    bool overwritingCountedDelete = false;
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1915/commits/cb9cff084f7d9c00b530d0e57f540d27f96843f1#diff-a40aec54cc202afdf5b713427d52537f7835238e1d035e334b66d0c93c8d96aaR481

src/workerd/io/actor-cache.h
```diff
@@ -431,6 +431,12 @@ class ActorCache final: public ActorCacheInterface {
       }
     }
 
+    // This Entry is part of a CountedDelete, but has since been overwritten via a put().
+    // This is really only useful in determining if we need to retry the deletion of this entry from
```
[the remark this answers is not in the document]
MellowYarker: We're only checking it [here](https://github.com/cloudflare/workerd/pull/1814/commits/8f85f7a07c9c464fa6e0f0ed5ee9343817ebb7c8#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR2345) when deciding if this particular entry should be part of the counted delete this flush. We're using it to check if we're in a retry _and_ we've since updated the key.

src/workerd/io/actor-cache-test.c++
```diff
@@ -1374,24 +1356,18 @@ KJ_TEST("ActorCache read retry") {
   auto mockGet = mockStorage->expectCall("get", ws)
       .withParams(CAPNP(key = "foo"));
 
-  // No activity on the transaction yet, because reads are outstanding.
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1915/commits/e2fa4e2fb4bdf186961b86ec346117d2b90be714

src/workerd/io/actor-cache-test.c++
```diff
@@ -1031,24 +1025,32 @@ KJ_TEST("ActorCache get-put ordering") {
...
+      .withParams(CAPNP(keys = ["bar", "baz", "foo"]), "stream"_kj)
+      .useCallback("stream", [&](MockClient stream) {
+    stream.call("values", CAPNP(list = [(key = "bar", value = "654"),
+                                        (key = "baz", value = "987"),
+                                        (key = "foo", value = "321")]))
+        .expectReturns(CAPNP(), ws);
+    stream.call("end", CAPNP()).expectReturns(CAPNP(), ws);
+  }).thenReturn(CAPNP());
 
-  // The flush transaction's capability will be created but no writes will be done on it until our
```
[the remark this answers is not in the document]
MellowYarker: I think https://github.com/cloudflare/workerd/pull/1814/commits/f6d3815fad3a11be427d3ba23fceaef3654f6b58#r1543760319 covers this too?

src/workerd/io/actor-cache-test.c++
```diff
@@ -1442,76 +1418,122 @@ KJ_TEST("ActorCache read retry on flush containing only puts") {
...
-  auto mockGet = mockStorage->expectCall("get", ws)
-      .withParams(CAPNP(key = "foo"));
+  test.put("foo", "456");
+  test.put("bar", "789");
+  test.put("baz", "123");
 
-  // The transaction won't write anything until the read completes.
-  auto mockTxn = mockStorage->expectCall("txn", ws).returnMock("transaction");
-  mockTxn->expectNoActivity(ws);
+  // Since we still have the promise for foo and bar, we do send a get for them. But baz is not in
```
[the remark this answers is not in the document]
MellowYarker: This can be phrased better. When we drop the promise, we drop our `GetWaiter`, which owns a strong reference to the Entry. `~GetWaiter()` will remove the Entry for `baz` from the `dirtyList`, but not from `currentValues`. When we do a `put("baz", "123")` after this, we add the `newEntry` to `dirtyList` and it gets flushed as a put. There is no `Entry` to get from storage since we removed it from the dirtyList earlier.

src/workerd/io/actor-cache-test.c++
```diff
@@ -1031,24 +1025,32 @@ KJ_TEST("ActorCache get-put ordering") {
 
   // Expect to receive the storage gets. But, don't return from them yet!
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/b53df5d5a759ea419c9cfd99544268aba56b4226

src/workerd/io/actor-cache-test.c++
```diff
@@ -1442,76 +1418,122 @@ KJ_TEST("ActorCache read retry on flush containing only puts") {
...
-  auto promise = test.get("foo").get<kj::Promise<kj::Maybe<kj::String>>>();
+  (void)expectUncached(test.get("corge"));
+  auto promise = expectUncached(test.get("foo"));
   test.put("bar", "456");
   test.delete_("baz");
 
-  // Expect the get, but don't resolve yet.
-  auto mockGet = mockStorage->expectCall("get", ws)
-      .withParams(CAPNP(key = "foo"));
+  // Expect the get, but cancel the promise before we finish it.
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/2531d9cd09115b4cff48a98c6af4dae15a970dc2

src/workerd/io/actor-cache-test.c++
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/2089ca41f41da23bd40f198bea269ef2d7ebba38

src/workerd/io/actor-cache-test.c++
```diff
@@ -1571,16 +1593,18 @@ KJ_TEST("ActorCache get-multiple partial retry") {
...
       .withParams(CAPNP(keys = ["foo", "qux"]), "stream"_kj)
       .useCallback("stream", [&](MockClient stream) {
     stream.call("values", CAPNP(list = [(key = "qux", value = "789")]))
         .expectReturns(CAPNP(), ws);
     stream.call("end", CAPNP()).expectReturns(CAPNP(), ws);
-  }).expectCanceled();
+  }).thenReturn(CAPNP());
 
-  KJ_ASSERT(promise.wait(ws) == kvs({{"baz", "456"}, {"qux", "789"}}));
+  // KJ_ASSERT(promise.wait(ws) == kvs({{"baz", "456"}, {"qux", "789"}}));
```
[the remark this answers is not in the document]
MellowYarker: Thanks for catching, didn't seem to be a problem https://github.com/cloudflare/workerd/pull/1916/commits/f1fe435cc43adb80af42fcc08418d3ba1d539f74

src/workerd/io/actor-cache-test.c++
```diff
@@ -1744,6 +1768,9 @@ KJ_TEST("ActorCache list() with limit") {
   // Cached if we try it again though.
   KJ_ASSERT(expectCached(test.list("bar", "qux", 4)) ==
       kvs({{"bar", "456"}, {"baz", "789"}, {"foo", "123"}, {"garply", "54321"}}));
+
+  // Return our uncached get from earlier.
+  mockStorage->expectCall("get", ws).withParams(CAPNP(key = "fooa")).thenReturn(CAPNP());
```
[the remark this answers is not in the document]
MellowYarker: Moved this expect closer to the actual `get()` https://github.com/cloudflare/workerd/pull/1916/commits/b1b503ccbbe8d451029b99fee87b1e5e69d83715

> Huh, so now we issue a get for this even though its promise was dropped?
> I don't understand this :/

That makes two of us :sweat_smile:, it's not clear to me why our GetWaiter isn't dtor'd here when it is [above](https://github.com/cloudflare/workerd/pull/1814/commits/4a28176ca4d53331308bdfb49d76452d8041d4ca#r1540071138).

src/workerd/io/actor-cache-test.c++
```diff
@@ -3555,11 +3618,25 @@ KJ_TEST("ActorCache LRU purge ordering") {
...
   (void)expectUncached(test.get("bar"));
   (void)expectUncached(test.get("baz"));
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("qux"))) == "555");
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("xxx"))) == "aaa");
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("yyy"))) == "bbb");
+
+  // Now that we've confirmed `qux`, `xxx`, and `yyy` are in cache, we can let `bar` and `baz` be
+  // inserted into cache. If we did individual `get()`s, then `bar` and `baz` would actually
+  // evict `qux` and `xxx` before we got to verify if they were in cache!
+  mockStorage->expectCall("getMultiple", ws)
```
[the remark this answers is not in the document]
MellowYarker: Yes, especially because in some cases we drop it, ex. `ActorCache read overwrite`
```
  // Make some gets but overwrite them in the cache with puts.
  auto promise1 = expectUncached(test.get("foo"));
  auto promise2 = expectUncached(test.get("bar"));
  (void)expectUncached(test.get("baz"));

  test.put("foo", "456");
  test.put("bar", "789");
  test.put("baz", "123");
```

In this case, we don't do the `get()` for `baz`, but in other cases, ex. [this comment](https://github.com/cloudflare/workerd/pull/1814/commits/4a28176ca4d53331308bdfb49d76452d8041d4ca#r1543935758) we still do the get.

It's not yet clear to me what's causing this, but I think it's worth figuring out.

src/workerd/io/actor-cache-test.c++
```diff
@@ -1442,76 +1418,122 @@ KJ_TEST("ActorCache read retry on flush containing only puts") {
...
+
+KJ_TEST("ActorCache read cancel") {
   ActorCacheTest test;
   auto& ws = test.ws;
   auto& mockStorage = test.mockStorage;
 
-  // Don't use expectCached() this time because we don't want eagerlyReportExceptions(), because
-  // we actually expect an exception.
-  auto promise = test.get("foo").get<kj::Promise<kj::Maybe<kj::String>>>();
+  (void)expectUncached(test.get("corge"));
```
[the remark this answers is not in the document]
MellowYarker: This has since been modified [in the other PR](https://github.com/cloudflare/workerd/pull/1916/commits/2531d9cd09115b4cff48a98c6af4dae15a970dc2) to test dropping before we yield the event loop, but as mentioned in another comment, that behavior looks a bit weird at the moment...

src/workerd/io/actor-cache.c++
MellowYarker: https://github.com/cloudflare/workerd/compare/ccaf59b65229b2520ad8e6f3913856471947d4ab..9bb1e4d593bf4e07382f72bdf28b2343716a8ab9

src/workerd/io/actor-cache.h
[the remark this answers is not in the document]
MellowYarker: This code has moved to https://github.com/cloudflare/workerd/pull/1915 but to address your real concern:

**_Suppose we have an unresolved counted delete_**, then the Entry could not have been `PRESENT` before as we would have already counted it. We also know the prior state couldn't have been `ABSENT`, as we would have done an early return when trying to `delete()` a known negative entry.

Okay, so we have 2 possible cases:

1. The Entry was in cache, but it was `UNKNOWN`. In this case, we give the `CountedDelete` a strong reference to the newly created `ABSENT` entry (representing our delete attempt). Our `Entry` goes in the `dirtyList`.
2. The Entry was not in cache. Here we also give the `CountedDelete` a strong reference to the newly created `ABSENT` entry. Our `Entry` goes in the `dirtyList`.

Now assume we're overwriting the entry with a `put()` before it has resolved.
- **Note**: We know we can't overwrite with a delete because in `putImpl()` we would see that we're in cache _and_ the existing entry is `ABSENT` _and_ the existing entry and the new entry are the same (non-existent), so we would early return.

Alright, so we're definitely dealing with a `put()` and the existing entry is `ABSENT`. We mark the _existing Entry_ as having been overwritten, remove it from the `dirtyList`, and replace it in cache with the new `PRESENT` Entry (which is also added to the `dirtyList`. The `CountedDelete` still hasn't ran, but it's fine because the `CountedDelete` actually has a strong reference to the `ABSENT` Entry via `CountedDelete::entries`!

When we eventually go to flush, we'll look at all our `CountedDelete`s and find the Entry which has `overwritingCountedDelete == true`, but it won't matter in this case because this isn't a retry following a successful counted delete in a failed transaction (i.e. we haven't already obtained the count previously).

---
To summarize, we only need `overwritingCountedDelete` to help us filter out `Entry`s that have already been counted as part of a counted delete. All `Entry`s associated with a CountedDelete are tracked by the `CountedDelete` object itself.

src/workerd/io/actor-cache.c++
MellowYarker: Removed in this commit https://github.com/cloudflare/workerd/pull/1915/commits/534ac7f9a9abab840d9030c1e0b9ca859a03c083

PR #1817, opened 2024-03-12
[WIP][bazel] disable creating duplicate runfile trees

fhanau:


fhanau:
Closing this – as I understand this helps with tests/genrules in external repositories, so the workerd build is not affected much as there are only a few build scripts being invoked in its dependencies.

PR #1821, opened 2024-03-13
WIP: slightly simplify io bazel build file

Frederik-Baetens:
separate actor-id, io-channels, actor-storage.capnp and worker-interface.capnp

WIP, this doesn't work with internal EW yet.


jasnell (approved):

src/workerd/io/BUILD.bazel
```diff
@@ -62,8 +64,11 @@ wd_cc_library(
     visibility = ["//visibility:public"],
     deps = [
         ":capnp",
+        ":actor-id",
```
justin-mp, line 67: nit: In general, bazel dependencies should be in alphabetical order.  (There's a tool called buildifier that will do this for you automatically, but I don't know if we apply it.)

src/workerd/io/BUILD.bazel
```diff
@@ -48,6 +48,8 @@ wd_cc_library(
         # TODO(soon): Fix this
         exclude = [
             "observer.h",
+            "io-channels.h",
```
justin-mp, line 51: nit: we might as well keep the excludes in alphabetical order too.

justin-mp (approved):

src/workerd/io/BUILD.bazel
kentonv: My personal opinion: A human should never spend time alphabetizing things like this, the time they spend will outweigh any possible benefits.

Having a computer do it seems fine if it's all automated.

PR #1823, opened 2024-03-13
Separate out some jsg::Url mods from the module registry refactor

jasnell:
Separates the `jsg::Url` modifications out from the module registry refactor PR to ever-so-slightly reduce the size of that PR... addresses review comments also.

Refs: https://github.com/cloudflare/workerd/pull/1553

### DIFF
--- a/src/workerd/jsg/url-test.c++
+++ b/src/workerd/jsg/url-test.c++
@@ -3145,6 +3145,47 @@ KJ_TEST("Can parse") {
   }
 }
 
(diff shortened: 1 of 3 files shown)


garrettgu10:
LGTM!

One small question: Is it valid to percent-encode a byte with mixed-case? like %bA

garrettgu10 (approved):

jasnell:
Yes, the percent encoding is case-insensitive. Generally, as a best practice, you're always supposed to *produce* upper-case sequences but accept any mix of cases.

PR #1827, opened 2024-03-13
More JS stream simplifications

jasnell:
More simplification of the JS-backed streams... 

Previously, the code was needlessly jumping through a few weakref hoops to try to avoid a hard circular ref. These can be handled just fine via tracing since v8 now more optimistically gc'ing wrapper objects. Long overdue change.


fhanau (approved):

jasnell:
There's a relevant internal ASAN test failure to investigate before this can land

PR #1828, opened 2024-03-13
Add ActorCache hooks for recording storage read/write latencies

a-robinson:


src/workerd/io/actor-cache.c++
```diff
@@ -2794,6 +2815,8 @@ kj::Promise<void> ActorCache::flushImplUsingTxn(
   // avoid the data changing out from under us while we wait.
   co_await waitForPastReads();
 
+  auto writeObserver = recordStorageWrite(hooks, clock);
```
justin-mp, line 2818: The is the only time that the `recordStorage*` function isn't measuring the same time as a `util::DurationExceededLogger`.  (The corresponding logger starts recording on line 2920.)  Should they be coincident?  If not, can you comment why they're not the same?

src/workerd/io/actor-cache.c++
```diff
@@ -2657,6 +2675,9 @@ kj::Promise<void> ActorCache::flushImplUsingSingleCountedDelete(CountedDeleteFlu
 kj::Promise<void> ActorCache::flushImplAlarmOnly(DirtyAlarm dirty) {
   co_await waitForPastReads();
 
+    auto writeObserver = recordStorageWrite(hooks, clock);
```
justin-mp, line 2678: nit: there's weird whitespace here

```suggestion
  auto writeObserver = recordStorageWrite(hooks, clock);
```

src/workerd/io/actor-cache.c++
a-robinson: No important reason. I was in some sense future-proofing against any awaits being added between `waitForPastReads` and when we wait on our response, and in some sense attempting to account for the time spent sending the storage operations as opposed to just waiting on their responses. But that time should be negligible, so I'll just stick with consistency I guess.

justin-mp (approved):

a-robinson:
Rebased for merge.

src/workerd/io/actor-cache.c++
```diff
@@ -25,6 +25,21 @@ static constexpr size_t MAX_ACTOR_STORAGE_RPC_WORDS = (16u << 20) / sizeof(capnp
 
 ActorCache::Hooks ActorCache::Hooks::DEFAULT;
 
+namespace {
+
+// Utility functions for recording latency metrics via a one-liner in the callers below.
+auto recordStorageRead(ActorCache::Hooks& hooks, const kj::MonotonicClock& clock) {
+  auto start = clock.now();
```
MellowYarker, line 32: Sorry for asking late, but I'm a bit confused by how this works. I figure we start running immediately since we invoke when `attach()`ing, but then wouldn't the `kj::defer` run almost immediately after?

justin-mp: Exactly!  We return the `kj::defer()`, so it becomes the responsibility of the caller to destroy it.

PR #1830, opened 2024-03-13
Give a friendly message when the user specifies an unsupported package in local dev

garrettgu10:
Test plan:

add an unsupported package to a sample like `samples/pyodide-fastapi/config.capnp`, then build and run 
```
./bazel-bin/src/workerd/server/workerd test samples/pyodide-fastapi/config.capnp
```

TODO: Currently blocked on getting a permalink to wherever we want people to go to to request packages

### DIFF
--- a/src/node/internal/internal_inspect.ts
+++ b/src/node/internal/internal_inspect.ts
@@ -29,6 +29,7 @@
 /* todo: the following is adopted code, enabling linting one day */
 /* eslint-disable */
 
+import internalWorkers from "cloudflare-internal:workers";
 import internal from "node-internal:util";
 
 import { Buffer } from "node-internal:internal_buffer";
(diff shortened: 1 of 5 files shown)


hoodmane (approved):
Looks reasonable. What does this look like when you hit it in local dev? Presumably you also get an error if you attempt to upload it / remote development?

garrettgu10:
> Looks reasonable. What does this look like when you hit it in local dev? Presumably you also get an error if you attempt to upload it / remote development?

I tested with remote development and I think the other (non-IS_WORKERD) branch can throw the same error. So I moved the try/catch so that the friendly message now gets added in both remote and local contexts. I don't think we can do this at upload time since I don't think workerd even spins up then.

hoodmane:
> since I don't think workerd even spins up then.

Well the validator does run Python setup I believe.

irvinebroque:
@garrettgu10 can link to https://github.com/cloudflare/workerd/discussions/new?category=python-packages

src/pyodide/python-entrypoint-helper.js
```diff
@@ -85,11 +85,22 @@ async function setupPackages(pyodide) {
...
+    transitiveRequirements = new Set(
+      Array.from(await getTransitiveRequirements(pyodide, requirements)).map(
+        canonicalizePackageName,
+      ),
+    );
+  } catch (e) {
+    if (e.message.includes("No known package")) {
+      e.message +=
+        ". \nIt appears that a package you requested is not available yet on workerd. \n" +
+        "If you would like this package to be included, please open an issue at TODO.";
```
irvinebroque, line 100:
```suggestion
        "If you would like this package to be included, please open an issue at https://github.com/cloudflare/workerd/discussions/new?category=python-packages.";
```

src/pyodide/python-entrypoint-helper.js
```diff
@@ -85,11 +85,22 @@ async function setupPackages(pyodide) {
   const requirements = MetadataReader.getRequirements().map(
     canonicalizePackageName,
   );
-  const transitiveRequirements = new Set(
-    Array.from(await getTransitiveRequirements(pyodide, requirements)).map(
-      canonicalizePackageName,
-    ),
-  );
+
+  let transiteRequirements = [];
```
dom96, line 89:
```suggestion
  let transitiveRequirements = [];
```

PR #1831, opened 2024-03-13
Implements Jaeger tracing for Python.

dom96:
This assumes that we disallow access to `pyodide-internal` from the Python code.

TODOs left:
- [x] Currently this only adds a few spans, will add more before merging but want to get this to review asap.
- [x] Need to set up a disabled internal jaeger tracer in workerd-api.

Tested upstream.

### DIFF
--- a/src/pyodide/internal/python.js
+++ b/src/pyodide/internal/python.js
@@ -3,6 +3,7 @@ import { createTarFS } from "pyodide-internal:tarfs";
 import { createMetadataFS } from "pyodide-internal:metadatafs";
 import { default as ArtifactBundler } from "pyodide-internal:artifacts";
 import { default as LOCKFILE } from "pyodide-internal:generated/pyodide-lock.json";
+import { default as internalJaeger } from "pyodide-internal:jaeger";
 
 const canonicalizeNameRegex = /[-_.]+/g;
 
@@ -68,6 +69,19 @@ import stdlib from "pyodide-internal:generated/python_stdlib.zip";
  * decent performance/memory usage tradeoff.
  */
 let MEMORY = undefined;
+
+/**
+ * Used for tracing via Jaeger.
+ */
+
+export async function enterJaegerSpan(span, callback) {
+  if (!internalJaeger.traceId) {
+    // Jaeger tracing not enabled or traceId is not present in request.
+    return callback();
+  }
+
+  return internalJaeger.enterSpan(span, callback);
+}
 
 /**
  * A hook that the Emscripten runtime calls to perform the WebAssembly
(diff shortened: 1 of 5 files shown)


src/pyodide/internal/python.js
```diff
@@ -69,6 +70,22 @@ import stdlib from "pyodide-internal:generated/python_stdlib.zip";
...
+export async function enterJaegerSpan(span, callback) {
+  console.log(internalJaeger.traceId);
+  if (!internalJaeger.traceId) {
+    // Jaeger tracing not enabled or traceId is not present in request.
+    return await callback();
+  }
+
+  return await internalJaeger.enterSpan(span, async (span) => {
+    return await callback();
+  });
```
hoodmane, line 86: any reason not to do:
```suggestion
  return await internalJaeger.enterSpan(span, callback);
```

src/pyodide/python-entrypoint-helper.js
```diff
@@ -176,7 +182,7 @@ async function preparePython() {
 function makeHandler(pyHandlerName) {
   return async function (...args) {
     try {
-      const mainModule = await preparePython();
+      const mainModule = await enterJaegerSpan("prep_python", async () => await preparePython());
```
hoodmane, line 185: Should break this down into a span for `getPyodide` and a span for `setupPackages`.

src/pyodide/internal/python.js
```diff
@@ -69,6 +70,22 @@ import stdlib from "pyodide-internal:generated/python_stdlib.zip";
  */
 let MEMORY = undefined;
 
+/**
+ * Used for tracing via Jaeger.
+ */
+
+export async function enterJaegerSpan(span, callback) {
+  console.log(internalJaeger.traceId);
```
hoodmane, line 78:
```suggestion
```

src/pyodide/internal/python.js
```diff
@@ -3,6 +3,7 @@ import { createTarFS } from "pyodide-internal:tarfs";
 import { createMetadataFS } from "pyodide-internal:metadatafs";
 import { default as ArtifactBundler } from "pyodide-internal:artifacts";
 import { default as LOCKFILE } from "pyodide-internal:generated/pyodide-lock.json";
+import { default as internalJaeger } from "pyodide-internal:jaeger";
```
hoodmane, line 6: Looks like we need to fill this in with a `DisabledInternalJaeger` in order to avoid breaking Python in workerd?

hoodmane: Ah I see your checklist item for this.

src/pyodide/internal/python.js
dom96: I wanted to make sure the callback is awaited inside the span scope, I was worried that `enterSpan` may not await the callback for me. But I see in its code that it does so I'll change this.

src/pyodide/python-entrypoint-helper.js
dom96: We can have sub-spans, so I'll just add spans inside `preparePython` too.

hoodmane (approved):

PR #1833, opened 2024-03-14
Add PreventIoScope

jasnell:
This implements a stack-allocated guard that prevents acquiring the current IoContext and prevents the scope from detecting if there is a current IoContext (or if an IoContext reference is the current).

The goal is to provide a simple means of preventing some block of code from acquiring the IoContext to schedule i/o without having to drop and then reacquire the isolate lock just to prevent from bit of code from accessing i/o.

This is intended to be used primarily with dynamic imports (for both the old and new module registry impls), where the import may be running within an I/O context but we want to prevent the modules being imported from accessing the IoContext while evaluating their top level scope. In the current implementation, we schedule a kj::Promise to run on the event loop to evaluate the module, which has to reacquire the isolate lock, evaluate the module, etc.

Currently, this guard only:

* Prevents calling `IoContext::current()` (it will throw)
* Causes `IoContext::hasCurrent()` to return false
* Causes `ioContext.isCurrent()` to return false

It does not prevent calls to various methods on an IoContext you already have a reference to unless those methods would call any of the three above operations.

### DIFF
--- a/src/workerd/io/io-context.c++
+++ b/src/workerd/io/io-context.c++
@@ -16,6 +16,13 @@ namespace workerd {
 static thread_local IoContext* threadLocalRequest = nullptr;
 static thread_local void* threadId = nullptr;
 
+PreventIoScope::PreventIoScope() : current(threadLocalRequest) {
+  threadLocalRequest = nullptr;
+}
+PreventIoScope::~PreventIoScope() {
+  threadLocalRequest = current;
+}
+
 static void* getThreadId() {
   if (threadId == nullptr) threadId = new int;
   return threadId;
@@ -1117,22 +1124,18 @@ void IoContext::runImpl(Runnable& runnable, bool takePendingEvent,
   });
 }
 
-static constexpr auto kAsyncIoErrorMessage =
-    "Disallowed operation called within global scope. Asynchronous I/O "
-    "(ex: fetch() or connect()), setting a timeout, and generating random "
-    "values are not allowed within global scope. To fix this error, perform this "
-    "operation within a handler. "
-    "https://developers.cloudflare.com/workers/runtime-apis/handlers/";
-
 IoContext& IoContext::current() {
   if (threadLocalRequest == nullptr) {
     v8::Isolate* isolate = v8::Isolate::TryGetCurrent();
-    if (isolate == nullptr) {
-      KJ_FAIL_REQUIRE("there is no current request on this thread");
-    } else {
-      isolate->ThrowError(jsg::v8StrIntern(isolate, kAsyncIoErrorMessage));
-      throw jsg::JsExceptionThrown();
-    }
+    KJ_REQUIRE(isolate != nullptr, "there is no current request on this thread");
+    jsg::Lock& js = jsg::Lock::from(isolate);
+    js.throwException(js.error(
+        "Disallowed operation called within global scope. Asynchronous I/O "
+        "(ex: fetch() or connect()), setting a timeout, and generating random "
+        "values are not allowed within global scope. To fix this error, perform this "
+        "operation within a handler. "
+        "https://developers.cloudflare.com/workers/runtime-apis/handlers/"_kj
+    ));
   } else {
     return *threadLocalRequest;
   }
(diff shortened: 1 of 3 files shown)


src/workerd/io/io-context.c++
```diff
@@ -1109,33 +1120,35 @@ void IoContext::runImpl(Runnable& runnable, bool takePendingEvent,
...
-      throw jsg::JsExceptionThrown();
-    }
+    KJ_REQUIRE(isolate != nullptr, "there is no current request on this thread");
+    jsg::Lock& js = jsg::Lock::from(isolate);
+    js.throwException(js.strIntern(
+        "Disallowed operation called within global scope. Asynchronous I/O "
+        "(ex: fetch() or connect()), setting a timeout, and generating random "
+        "values are not allowed within global scope. To fix this error, perform this "
+        "operation within a handler. "
+        "https://developers.cloudflare.com/workers/runtime-apis/handlers/"_kjc
```
kentonv, line 1133: it's better to use `_kj` unless you have a very specific reason that you need `_kjc`.

src/workerd/io/io-context.c++
```diff
@@ -1109,33 +1120,35 @@ void IoContext::runImpl(Runnable& runnable, bool takePendingEvent,
...
+    ));
+  } else if (PreventIoScope::inPreventIoScope()) {
+    jsg::Lock& js = threadLocalRequest->getCurrentLock();
+    js.throwException(js.strIntern(
+        "Disallowed operation called within this scope. Asynchronous I/O "
+        "(ex: fetch() or connect(), setting a timeout, and generating random "
```
jasnell, line 1139:
```suggestion
        "(ex: fetch() or connect()), setting a timeout, and generating random "
```

src/workerd/io/io-context.c++
jasnell, line 1133:
```suggestion
        "https://developers.cloudflare.com/workers/runtime-apis/handlers/"_kj
```

src/workerd/io/io-context.c++
```diff
@@ -1109,33 +1120,35 @@ void IoContext::runImpl(Runnable& runnable, bool takePendingEvent,
...
+        "values are not allowed within this scope."_kjc));
```
jasnell, line 1140:
```suggestion
        "values are not allowed within this scope."_kj));
```

src/workerd/io/io-context.c++
```diff
@@ -1109,22 +1116,18 @@ void IoContext::runImpl(Runnable& runnable, bool takePendingEvent,
...
-    }
+    KJ_REQUIRE(isolate != nullptr, "there is no current request on this thread");
+    jsg::Lock& js = jsg::Lock::from(isolate);
+    js.throwException(js.strIntern(
+        "Disallowed operation called within global scope. Asynchronous I/O "
+        "(ex: fetch() or connect()), setting a timeout, and generating random "
+        "values are not allowed within global scope. To fix this error, perform this "
+        "operation within a handler. "
+        "https://developers.cloudflare.com/workers/runtime-apis/handlers/"_kj
+    ));
```
jasnell, line 1130: This part of the change is not technically necessary but it makes this bit more compact/cleaner so kept it in.

jasnell:
Really unsure why the python related tests here are failing @hoodmane @dom96

hoodmane:
Looks like it's because we attempt to test whether `crypto.getRandomValues` fails because we're in global scope:
`if (e.message.includes("Disallowed operation called within global scope"))`

jasnell:
Wait... doh, I think I goofed one of the changes here... sigh.

kentonv (approved):
While I have nothing against this API, it occurs to me that it might not actually satisfy the intended use case of supporting dynamic imports without dropping the isolate lock.

A dynamic import can presumably have top-level awaits. That means we will need to pump the microtask queue in order to complete the import. We cannot pump the queue synchronously while JavaScript is on the stack. So we have to evaluate the whole import asynchronously anyway. It's also important that we make sure the microtask queue is clear of any request-specific tasks before we begin the dynamic import, so that we don't end up with request tasks running without the IoContext.

So PreventIoScope doesn't end up being useful here after all?

jasnell:
hmm... yeah, sadly I think you're right. I really wanted to try to avoid having to drop and reacquire the lock but I don't think there's really going to be a way to do it reasonably. Ah well, will keep things working the way they are now then.

PR #1834, opened 2024-03-14
Fixes memory snapshot upload for Python Workers.

dom96:
This PR:

* Moves upload outside the global scope
* Makes certain methods in IoContext  public so they can be used in EW.

Tested in EW.

### DIFF
--- a/src/pyodide/python-entrypoint-helper.js
+++ b/src/pyodide/python-entrypoint-helper.js
@@ -6,6 +6,7 @@ import {
   mountLib,
   canonicalizePackageName,
   enterJaegerSpan,
+  uploadArtifacts,
 } from "pyodide-internal:python";
 import { default as LOCKFILE } from "pyodide-internal:generated/pyodide-lock.json";
 import { default as MetadataReader } from "pyodide-internal:runtime-generated/metadata";
@@ -193,6 +194,8 @@ function makeHandler(pyHandlerName) {
     } catch (e) {
       console.warn(e.stack);
       throw e;
+    } finally {
+      args[2].waitUntil(uploadArtifacts());
     }
   };
 }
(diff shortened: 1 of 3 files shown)


src/workerd/io/io-context.h
```diff
@@ -726,6 +726,28 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
...
+  // request for logging and tracing. This is a hack, we don't actually know which request
+  // "caused" any particular resource usage, so this is merely our best guess.
+  //
+  // The IoChannelFactory must also be accessed through the currentIncomingRequest because it has
+  // some tracing context built in.
+  IncomingRequest& getCurrentIncomingRequest() {
+    KJ_REQUIRE(!incomingRequests.empty(), "the IoContext has no current IncomingRequest");
+    return incomingRequests.front();
+  }
+  IoChannelFactory& getIoChannelFactory() {
```
harrishancock, line 747: I'd prefer if we could keep this private, too -- I describe a possible way to do so in the counterpart internal PR. However, I can live with it being public if time is of the essence, so I'm approving.

harrishancock (approved):

src/pyodide/python-entrypoint-helper.js
```diff
@@ -193,6 +194,8 @@ function makeHandler(pyHandlerName) {
     } catch (e) {
       console.warn(e.stack);
       throw e;
+    } finally {
+      args[2].waitUntil(uploadArtifacts());
```
hoodmane, line 198: Maybe we should give the arguments names again. Though this is okay for now.

src/workerd/io/io-context.h
dom96: Added a todo for now

PR #1835, opened 2024-03-14
Sanitize globalThis for Python Workers.

dom96:
Hopefully fairly self explanatory.

### DIFF
--- a/src/pyodide/internal/python.js
+++ b/src/pyodide/internal/python.js
@@ -437,5 +437,12 @@ export async function loadPyodide(lockfile, indexURL) {
       snapshot: snapshotString,
     });
   }
+
+  // Sanitize `globalThis`. This is not strictly doing much right now, but may be necessary in
+  // the future if we enable one or more of these.
+  globalThis.eval = undefined;
+  globalThis.import = undefined;
+  globalThis.require = undefined;
+
   return pyodide;
 }


garrettgu10 (approved):

src/pyodide/internal/python.js
```diff
@@ -429,5 +429,12 @@ export async function loadPyodide(lockfile, indexURL) {
       snapshot: snapshotString,
     });
   }
+
+  // Sanitize `globalThis`. This is not strictly doing much right now, but may be necessary in
+  // the future if we enable one or more of these.
+  globalThis.eval = undefined;
```
jasnell, line 435: Thinking through what the ramifications of this would be. Technically there's no reason to null out `globalThis.eval` since we block all calls to `eval(...)` on the backend. There are only a handful of cases where we temporarily bypass that restriction and none of those would apply here.

As for `globalThis.import`, does that actually do anything? The `import` statement and `import()` functions are language level constructs. I don't believe that `import` is ever actually installed on `globalThis` so by nulling this out you might be breaking something the user code intended (via use of some polyfill). Even if we set `globalThis.import` to `undefined`, `await import('foo')` and `import * from ...` should both still work.

For `globalThis.require(...)`, I also wonder if this is needed. The `require(...)` method is only ever injected specifically into `commonJs` and `nodeJsCompat` type modules, which means in the common case this won't ever be set. The `python.js` module is loaded as an ESM module, so `require(...)` shouldn;'t be there.

hoodmane: I agree that none of this is strictly necessary. But I feel like it's also harmless and if it helps @dom96 sleep at night it's maybe even helpful...

jasnell: Well, the biggest thing I'm worried about is a false sense of security. `globalThis.import = undefined` does not actually do what it is intended to do here... so I'm not even sure I would call it harmless

hoodmane: Well in Python the only way to access `import` is via `Reflect.get(globalThis, "someString")`. Which is to say, it's not possible to access `import`.

dom96: We don't allow `eval` _right now_, but I think it's possible we will enable it in the future, hence the extra safety here. But perhaps even if it is enabled it will have a different scope for imports anyway so won't be dangerous. My worry is that some time in the future it'll get enabled without considering Python and Python Code will suddenly be able to do `eval('import { default as internalJaeger } from "pyodide-internal:jaeger";')`.

I agree the others are much less practical, since it sounds like they may cause harm I'll just remove them.

hoodmane: I think `eval('import { default as internalJaeger } from "pyodide-internal:jaeger";')` will raise: `SyntaxError: Cannot use import statement outside a module`. But `eval('import("pyodide-internal:jaeger");')` is more worrisome...

jasnell: Honestly, it's unlikely that we'll ever support `eval(...)` outside of the `unsafe` binding. @kentonv might have other opinions on that tho. We have talked about possibly supporting it for certain known, specifically allow-listed operations but that's never been settled and I think `import(...)` wouldn't be on that allow-list.

> My worry is that some time in the future it'll get enabled without considering Python and Python Code will suddenly be able to do eval('import { default as internalJaeger } from "pyodide-internal:jaeger";').

This could be caught with a test tho, right?

dom96: Yeah, fair, I'll add a test.

PR #1836, opened 2024-03-14
Update `docs/v8-update.md`

vickykont:
Just a few updates to our V8 update process. Also, I added the `--no-stat` flag to the `git format-patch` command, since when I attempted to use the current version of the command, I also got some summaries of the changes (files and additions/deletions) at the end of the generated patches.

### DIFF
--- a/docs/v8-updates.md
+++ b/docs/v8-updates.md
@@ -2,9 +2,9 @@
 
 To update the version of V8 used by workerd, the steps are:
 
-1. Check <https://omahaproxy.appspot.com/> and identify the latest version of V8 used by the beta versions of Chrome beta.
+1. Check <https://chromiumdash.appspot.com/> and identify the latest version of V8 used by the beta versions of Chrome beta.
 
-2. Install depot_tools if it is not already present on your machine.
+2. Install `depot_tools` if it is not already present on your machine.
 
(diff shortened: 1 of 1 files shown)


docs/v8-updates.md
```diff
@@ -66,16 +66,19 @@ To update the version of V8 used by workerd, the steps are:
...
+    workerd using the newer V8 version.
+
     See [V8 http_archive in WORKSPACE](https://github.com/cloudflare/workerd/blob/587ad90dd1e91d2660c271018056f4189fca3501/WORKSPACE#L408)
 
 10. Update V8's dependencies in `workerd/WORKSPACE`.
 
-    The `v8/.gclient_entries` file contains the commit versions for V8's dependencies.
-
-    You can find V8's dependencies that are carried through to workerd in the `WORKSPACE` file.
+    You can find the commit versions for V8's dependencies under `v8/.gclient_entries` and the ones
```
fhanau, line 77: I know that this was already the case so far/you didn't add this but: I would prefer pointing to V8's `DEPS` here to get the dependency commits since it's part of the repository and tied to a given V8 commit – if someone forgets to run `gclient sync` after changing the V8 commit the `v8/.gclient_entries` entry will be outdated, i.e. may point to the wrong version.

vickykont: Ah, I didn't know. True, this seems to make more sense, although the other file is a little more compact. Updated!

fhanau (approved):
Thank you!

PR #1838, opened 2024-03-14
Add pyodide-lock command to cli

garrettgu10:
This command simply prints the pyodide-lock.json file onto stdout and exits. 

I have no idea whether or not I'm actually supposed to print stuff out like this, someone please check my work in workerd.c++.

### DIFF
--- a/src/pyodide/BUILD.bazel
+++ b/src/pyodide/BUILD.bazel
@@ -41,6 +41,12 @@ capnp_embed(
     deps = ["pyodide_extra_capnp_file"],
 )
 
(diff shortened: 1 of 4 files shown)


hoodmane (commented):
It looks very reasonable to me but let's ask others to approve before merging since I don't know how we do this sort of thing either. For instance, I think we might prefer some `kj` method to stdio.

dom96 (approved):
Looks good to me. I think `printf` is fine.

hoodmane:
Thanks @garrettgu10!

PR #1840, opened 2024-03-15
JSRPC: Support sending Headers, Request, and Response over RPC

kentonv:
(Based on top of #1806 which should be ready to merge very soon.)

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1137,6 +1137,64 @@ kj::Maybe<kj::String> Request::serializeCfBlobJson(jsg::Lock& js) {
   return cf.serialize(js);
 }
 
+void Request::serialize(
+    jsg::Lock& js, jsg::Serializer& serializer,
+    const jsg::TypeHandler<RequestInitializerDict>& initDictHandler) {
+  serializer.writeLengthDelimited(url);
+
+  // Our strategy is to construct an initializer dict object and serialize that as a JS object.
+  // This makes the deserialization end really simple (just call the constructor), and it also
+  // gives us extensibility: we can add new fields without having to bump the serialization tag.
+  serializer.write(js, jsg::JsValue(initDictHandler.wrap(js, RequestInitializerDict {
+    // GET is the default, so only serialize the method if it's something else.
+    .method = method == kj::HttpMethod::GET ? jsg::Optional<kj::String>() : kj::str(method),
+
+    .headers = headers.addRef(),
+
+    .body = getBody().map([](jsg::Ref<ReadableStream> stream) -> Body::Initializer {
+      // jsg::Ref<ReadableStream> is one of the possible variants of Body::Initializer.
+      return kj::mv(stream);
+    }),
+
+    // "manual" is the default for `redirect`, so only encode if it's not that.
+    .redirect = redirect == Redirect::MANUAL
+        ? kj::str(getRedirect()) : kj::Maybe<kj::String>(kj::none),
+
+    // We have to ignore .fetcher for serialization. We can't simply fail if a fetcher is present
+    // because requests received by the top-level fetch handler actually have .fetcher set to
+    // the hidden "next" binding, which historically could be different from null (although in
+    // practice these days it is always the same). We obviously want to be able to serialize
+    // requests received by the top-level fetch handler so... we have to ignore this. This
+    // property should probably go away in any case.
+
+    .cf = cf.getRef(js),
+
+    // .mode is unimplemented
+    // .credentials is unimplemented
+    // .cache is unimplemented
+    // .referrer is unimplemented
+    // .referrerPolicy is unimplemented
+    // .integrity is required to be empty
+
+    // If an AbortSignal is present, we'll try to serialize it. As of this writing, AbortSignal
+    // is not serializable, but we could add support for sending it over RPC in the future.
+    //
+    // Note we have to double-Maybe this, so that if no signal is present, the property is absent
+    // instead of `null`.
+    .signal = signal.map([](jsg::Ref<AbortSignal>& s) -> kj::Maybe<jsg::Ref<AbortSignal>> {
+      return s.addRef();
+    })
+  })));
+}
+
+jsg::Ref<Request> Request::deserialize(
+    jsg::Lock& js, rpc::SerializationTag tag, jsg::Deserializer& deserializer,
+    const jsg::TypeHandler<RequestInitializerDict>& initDictHandler) {
+  auto url = deserializer.readLengthDelimitedString();
+  auto init = KJ_ASSERT_NONNULL(initDictHandler.tryUnwrap(js, deserializer.readValue(js)));
+  return Request::constructor(js, kj::mv(url), kj::mv(init));
+}
+
 // =======================================================================================
 
 Response::Response(
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1583,6 +1583,43 @@ jsg::Optional<jsg::JsObject> Response::getCf(jsg::Lock& js) {
   return cf.get(js);
 }
 
+void Response::serialize(
+    jsg::Lock& js, jsg::Serializer& serializer,
+    const jsg::TypeHandler<InitializerDict>& initDictHandler,
+    const jsg::TypeHandler<kj::Maybe<jsg::Ref<ReadableStream>>>& streamHandler) {
+  serializer.write(js, jsg::JsValue(streamHandler.wrap(js, getBody())));
+
+  // As with Request, we serialize the initializer dict as a JS object.
+  serializer.write(js, jsg::JsValue(initDictHandler.wrap(js, InitializerDict {
+    .status = statusCode == 200 ? jsg::Optional<int>() : statusCode,
+    .statusText = statusText == defaultStatusText(statusCode)
+        ? jsg::Optional<kj::String>() : kj::str(statusText),
+    .headers = headers.addRef(),
+    .cf = cf.getRef(js),
+
+    // If a WebSocket is present, we'll try to serialize it. As of this writing, WebSocket
+    // is not serializable, but we could add support for sending it over RPC in the future.
+    //
+    // Note we have to double-Maybe this, so that if no signal is present, the property is absent
+    // instead of `null`.
+    .webSocket = webSocket.map([](jsg::Ref<WebSocket>& s) -> kj::Maybe<jsg::Ref<WebSocket>> {
+      return s.addRef();
+    }),
+
+    .encodeBody = bodyEncoding == BodyEncoding::AUTO
+        ? jsg::Optional<kj::String>() : kj::str("manual"),
+  })));
+}
+
+jsg::Ref<Response> Response::deserialize(
+    jsg::Lock& js, rpc::SerializationTag tag, jsg::Deserializer& deserializer,
+    const jsg::TypeHandler<InitializerDict>& initDictHandler,
+    const jsg::TypeHandler<kj::Maybe<jsg::Ref<ReadableStream>>>& streamHandler) {
+  auto body = KJ_ASSERT_NONNULL(streamHandler.tryUnwrap(js, deserializer.readValue(js)));
+  auto init = KJ_ASSERT_NONNULL(initDictHandler.tryUnwrap(js, deserializer.readValue(js)));
+  return Response::constructor(js, kj::mv(body), kj::mv(init));
+}
+
 // =======================================================================================
 
 jsg::Ref<Request> FetchEvent::getRequest() {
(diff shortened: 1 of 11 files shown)


src/workerd/api/http.c++
```diff
@@ -417,6 +418,143 @@ bool Headers::inspectImmutable() {
...
+// http-over-capnp.capnp has a nice list of common header names, taken from the HTTP/2 standard.
+// We'll use it as an optimization.
+//
+// Note that using numeric IDs for headers implies we lose the original capitalization. However,
+// the JS Headers API doesn't actually give the application any way to observe the capitalization
+// of header names -- it only becomes relevant when serializing over HTTP/1.1. And at that point,
+// we are actually free to change the capitalization anyway, and we commonly do (KJ itself will
+// normalize capitalization of all registered headers, and http-over-capnp also loses
+// capitalization). So, it's certainly not worth it to try to keep the original capitalization
+// across serialization.
```
jasnell, line 433: Probably easier just to note that http defines header names as case-insensitive and ignoring case here is perfectly fine.

src/workerd/api/http.c++
```diff
@@ -417,6 +418,143 @@ bool Headers::inspectImmutable() {
...
+      serializer.writeLengthDelimited(value);
+    }
+  }
+}
+
+jsg::Ref<Headers> Headers::deserialize(
+    jsg::Lock& js, rpc::SerializationTag tag, jsg::Deserializer& deserializer) {
+  auto result = jsg::alloc<Headers>();
+  uint guard = deserializer.readRawUint32();
+  KJ_REQUIRE(guard <= static_cast<uint>(Guard::NONE), "unknown Guald value");
```
jasnell, line 532:
```suggestion
  KJ_REQUIRE(guard <= static_cast<uint>(Guard::NONE), "unknown guard value");
```

src/workerd/jsg/setup.h
```diff
@@ -365,7 +365,7 @@ class Isolate: public IsolateBase {
   bool serialize(
       Lock& js, std::type_index type, jsg::Object& instance, Serializer& serializer) override {
     KJ_IF_SOME(func, wrapper->serializerMap.find(type)) {
-      func(js, instance, serializer);
+      func(*wrapper, js, instance, serializer);
```
jasnell, line 368: minor nit: can we keep the `js` argument first per our usual convention. It's fine if not but I definitely prefer the consistency

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -302,6 +302,38 @@ export class MyService extends WorkerEntrypoint {
...
     return value;
   }
+
+  async returnHeaders() {
+    let result = new Headers();
+    result.append("foo", "bar");
+    result.append("Set-Cookie", "abc");
+    result.append("set-cookie", "def");
+    result.append("corge", "!@#");
+    result.append("Content-Length", "123");
```
jasnell, line 312: Might be a bit pedantic but a test that serializes an empty `Headers()` might be worthwhile.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1031,6 +1063,48 @@ export let streams = {
...
+      let headers = await env.MyService.returnHeaders();
+      assert.strictEqual(headers instanceof Headers, true);
+
+      // Awkwardly, there's actually no API to get the non-lowercased header names.
+      assert.strictEqual([...headers].map(pair => pair.join(": ")).join("\n"),
+          "content-length: 123\n" +
+          "corge: !@#\n" +
+          "foo: bar\n" +
+          "set-cookie: abc\n" +
+          "set-cookie: def");
```
jasnell, line 1078: does `assert.deepStrictEqual(...)` not work here? e.g. 

```js
const check = new Headers([[ ... expected headers ]]);
assert.deepStrictEqual(check, headers);
```

I know it works in Node.js to compare map-like objects like `Headers` but I can't remember if it works with our polyfills.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -302,6 +302,38 @@ export class MyService extends WorkerEntrypoint {
...
+    return new Request("http://my-url.com", {
+      method: "PUT",
+      headers: {
+        "Accept-Encoding": "bazzip",
+        "Foo": "Bar"
+      },
+      redirect: "manual",
+      body: "Hello every body!",
+    });
+  }
```
jasnell, line 326: A test that verifies that a `Request` with an `AbortSignal` fails as expected would be a good idea.

src/workerd/api/http.c++
```diff
@@ -989,6 +1127,64 @@ kj::Maybe<kj::String> Request::serializeCfBlobJson(jsg::Lock& js) {
...
+        ? kj::str(getRedirect()) : kj::Maybe<kj::String>(kj::none),
+
+    // We have to ignore .fetcher for serialization. We can't simply fail if a fetcher is present
+    // because requests received by the top-level fetch handler actually have .fetcher set to
+    // the hidden "next" binding, which historically could be different from null (although in
+    // practice these days it is always the same). We obviously want to be able to serialize
+    // requests received by the top-level fetch handler so... we have to ignore this. This
+    // property should probably go away in any case.
+
+    .cf = cf.getRef(js),
```
jasnell, line 1160: We've accidentally broken the lazy parsing of the `cf` property before. It would probably be a good idea to add a test that ensures that the `request.cf` property continues to act as expected after cloning (both on the sending end and the receiving end)

src/workerd/api/http.c++
```diff
@@ -417,6 +418,143 @@ bool Headers::inspectImmutable() {
...
+    static_cast<uint>(capnp::CommonHeaderName::WWW_AUTHENTICATE);
+
+static kj::Array<kj::StringPtr> makeCommonHeaderList() {
+  auto enums = capnp::Schema::from<capnp::CommonHeaderName>().getEnumerants();
+  auto builder = kj::heapArrayBuilder<kj::StringPtr>(enums.size());
+  bool first = true;
+  for (auto e: enums) {
+    if (first) {
+      // Value zero is invalid, skip it.
+      static_assert(static_cast<uint>(capnp::CommonHeaderName::INVALID) == 0);
```
jasnell, line 449: Minor nit: Having this `static_assert` inside the for loop when it doesn't seem to be testing `e` is... a bit odd.

src/workerd/api/http.c++
```diff
@@ -417,6 +418,143 @@ bool Headers::inspectImmutable() {
...
+      builder.add(nullptr);
```
jasnell, line 450: Why add `nullptr` at all?

src/workerd/api/http.c++
```diff
@@ -417,6 +418,143 @@ bool Headers::inspectImmutable() {
...
+      first = false;
+      continue;
+    }
+
+    kj::Maybe<kj::StringPtr> name;
+
+    // Look for $commonText annotation.
+    for (auto ann: e.getProto().getAnnotations()) {
+      if (ann.getId() == 0x857745131db6fc83) {
```
jasnell, line 459: non-blocking nit: would prefer the id value to be defined in a constant rather than used inline, but obvious not a big deal

src/workerd/api/http.c++
```diff
@@ -417,6 +418,143 @@ bool Headers::inspectImmutable() {
...
+// across serialization.
+
+// If any more headers are added to the CommonHeaderName enum later, we should be careful about
+// introducing them into serialization. We need to roll out a change that recognizes the new IDs
+// before rolling out a change that sends them. MAX_COMMON_HEADER_ID is the max value we're willing
+// to send.
+static constexpr uint MAX_COMMON_HEADER_ID =
+    static_cast<uint>(capnp::CommonHeaderName::WWW_AUTHENTICATE);
+
+static kj::Array<kj::StringPtr> makeCommonHeaderList() {
```
jasnell, line 442: non-blocking nit: `api/http.c++~ is already fairly large... I'd prefer if these kinds of utility methods could be moved out to a separate `util/http.h` or something but obviously not critical.

jasnell (approved):
Number of comments. None that are blocking I believe.

jasnell (approved):
Number of comments. None that are blocking I believe.

src/workerd/api/http.c++
kentonv: Well, `Headers` does go out of its way to remember the original capitalization. Though given it's not visible at all in the API, maybe we should change that?

jasnell:
weird, I have no idea why github is showing my approval comment twice.

src/workerd/api/http.c++
jasnell: I would. The spec is pretty clear on them being case-insensitive https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1

src/workerd/jsg/setup.h
kentonv: Well, in many internal parts of JSG, the convention is that `TypeWrapper&` is the first argument, so there's a bit of a conflict. `DeserializeFunc` actually already had `TypeWrapper&` coming first, so in this commit when I added a similar `TypeWraper&` argument to `SerializeFunc`, I made it match.

These are entirely internal functions so I don't think it's a big deal one way or the other.

src/workerd/api/http.c++
```diff
@@ -417,6 +418,143 @@ bool Headers::inspectImmutable() {
...
+//
+// Note that using numeric IDs for headers implies we lose the original capitalization. However,
+// the JS Headers API doesn't actually give the application any way to observe the capitalization
+// of header names -- it only becomes relevant when serializing over HTTP/1.1. And at that point,
+// we are actually free to change the capitalization anyway, and we commonly do (KJ itself will
+// normalize capitalization of all registered headers, and http-over-capnp also loses
+// capitalization). So, it's certainly not worth it to try to keep the original capitalization
+// across serialization.
+
+// If any more headers are added to the CommonHeaderName enum later, we should be careful about
```
jasnell, line 435: Might be a good idea to reconcile this against the most recent qpack spec used for http/3 https://www.rfc-editor.org/rfc/rfc9204.html#appendix-A

src/workerd/api/http.c++
kentonv: I kind of prefer to keep the code next to where it is used. That said, it would make a lot of sense (in a separate PR) to break up http.h/c++ into separate files for Headers, Request, Response, etc.

src/workerd/api/http.c++
kentonv: Weird, this list seems different from the HTTP/2 list that I based http-over-capnp on. But if we're going to change anything I'd prefer to update http-over-capnp itself first.

src/workerd/api/http.c++
kentonv: 6382755c

src/workerd/api/tests/js-rpc-test.js
kentonv: f64d6252

src/workerd/api/tests/js-rpc-test.js
kentonv: Unfortunately, `assert.deepStrictEqual(a, b)` appears to pass if the parameters are merely both of type `Headers`, regardless of whether their content matches.

That said, using deep-equal to compare `[...headers]` against an array does work and is a lot nicer than my weird joins here. 5615cde4

src/workerd/api/tests/js-rpc-test.js
kentonv: ec2b08c2

src/workerd/api/http.c++
kentonv: 807678e5

FWIW if the cf property isn't structured clonable, sending it over RPC will fail here. We know that non-clonable cf properties exist in the wild, but I think this is OK because we're not forcing RPC into any existing scripts.

src/workerd/api/http.c++
kentonv: This `static_assert` is showing that the first enum value (zero) has the name INVALID, just to prove that the comment above is correct. I like to write `static_assert`s as close as possible to the code that is relying on whatever is asserted.

src/workerd/api/http.c++
kentonv: 4207edd3

src/workerd/api/http.c++
kentonv: bb6225bf

kentonv:
Decided to consolidate #1844 into this PR to reduce self-conflicts for the internal repo.

src/workerd/api/http.c++
jasnell: Yep, it just looks a bit out of place since it's not testing `e` directly here. Minor issue but I can see someone coming into the code later and being a bit confused.

jasnell:
Still LGTM

src/workerd/api/http.c++
kentonv: Of course. Our `Headers` implementation is case-insensitive, but also case-preserving, kind of like the Windows filesystem. The question is whether we should abandon case preservation.

PR #1842, opened 2024-03-15
Reland TransformStream cancel algorithm

jasnell:
Previously landed but reverted while tracking down some other issue. Adds support for the new TransformStream cancel algorithm


fhanau (approved):
I reviewed the previous PR and this looks identical so stamping this too. Only difference I noticed is that `expectedLength` within `Transformer` is no longer being exposed to the  JSG interface, just wanted to make sure that's intentional.

PR #1848, opened 2024-03-15
Implement AbortSignal.protototype.onabort

jasnell:
The spec defines `onabort` as a prototype property, implement it as such.

Technically speaking, this *could* be considered a breaking change :-/ ... I'm hoping we can get away with it without requiring a compat flag. It's a potentially breaking change because prior to this, the following operations would have different results:

* `'onabort' in AbortSignal.prototype`  ... previously, false, now true
* `delete signal.onabort` ... previously deletes the own property if set, now has no change
* `Object.keys(signal)` ... previously would include `onabort` if it was set, now always excludes it

One thing this PR doesn't do (that we can do later in a separate PR) is define the type mapping override for `onabort`.

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -496,6 +496,23 @@ AbortSignal::AbortSignal(kj::Maybe<kj::Exception> exception,
     flag(flag),
     reason(kj::mv(maybeReason)) {}
 
+kj::Maybe<jsg::JsValue> AbortSignal::getOnAbort(jsg::Lock& js) {
+  return onAbortHandler.map([&](jsg::JsRef<jsg::JsValue>& ref)
+      -> jsg::JsValue { return ref.getHandle(js); });
+}
+
+void AbortSignal::setOnAbort(jsg::Lock& js, jsg::Optional<jsg::JsValue> handler) {
+  // We only want to accept the handler if it's a valid handler... For anything
+  // else, set it to null.
+  KJ_IF_SOME(h, handler) {
+    if (h.isFunction() || h.isObject()) {
+      onAbortHandler = jsg::JsRef(js, h);
+      return;
+    }
+  }
+  onAbortHandler = kj::none;
+}
+
 jsg::JsValue AbortSignal::getReason(jsg::Lock& js) {
   KJ_IF_SOME(r, reason) {
     return r.getHandle(js);
@@ -612,7 +629,7 @@ jsg::Ref<AbortSignal> AbortSignal::any(
 }
 
 void AbortSignal::visitForGc(jsg::GcVisitor& visitor) {
-  visitor.visit(reason);
+  visitor.visit(reason, onAbortHandler);
 }
 
 RefcountedCanceler& AbortSignal::getCanceler() {
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -508,6 +508,16 @@ void AbortSignal::setOnAbort(jsg::Lock& js, jsg::Optional<jsg::JsValue> handler)
     if (h.isFunction() || h.isObject()) {
       onAbortHandler = jsg::JsRef(js, h);
       return;
+    } else {
+      // TODO(soon): Per the spec we are supposed o set the handler to null if it is not
+      // a function or an object. However, there's an ever so slight change that would
+      // be breaking. So let's go ahead and set the value in this case and log a warning.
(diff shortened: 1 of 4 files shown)


src/workerd/api/basics.c++
```diff
@@ -496,6 +496,23 @@ AbortSignal::AbortSignal(kj::Maybe<kj::Exception> exception,
     flag(flag),
     reason(kj::mv(maybeReason)) {}
 
+kj::Maybe<jsg::JsValue> AbortSignal::getOnAbort(jsg::Lock& js) {
+  return onAbortHandler.map([&](jsg::JsRef<jsg::JsValue>& ref)
+      -> jsg::JsValue { return ref.getHandle(js); });
+}
+
+void AbortSignal::setOnAbort(jsg::Lock& js, jsg::Optional<jsg::JsValue> handler) {
+  // We only want to accept the handler if it's a valid handler... For anything
```
kentonv, line 505: I feel like this is the most likely part of this PR to break someone, if anything. Previously people could set `onabort` to any value.

Do we really need this type check?

jasnell: To be spec compliant yes. The entire change potentially breaks someone. I do believe it's highly unlikely that someone would be using a non-event-handler value for this property anyway.

kentonv: To be clear, is this type check specifically necessary to pass the test we're trying to pass?

kentonv: I agree it's not likely.

What I'm saying is that, in my mind, the value of this type check is less than the cost of a compat flag. So if it came down to it and we decided we needed a compat flag for this change, I would be inclined instead to say, just remove the type check and don't bother with a compat flag.

But if we believe no compat flag is needed then, fine.

One specific question, though: `null` is considered an object. Does this code do the right thing when the handler is assigned to `null`? It looks like we'll save the value `null`, rather than actually null out the `kj::Maybe`. Is that OK?

jasnell: If I'm remembering correctly, the c++ `IsObject()` check returns false on `null` so this should just do the right thing. But even if it did store `null`, that would be fine.

IgorMinar: It would be preferable to stick to the spec here and keep the check in order to stop bad code from seeping in. I think it's unlikely that anyone reads the value back in in the case of this API, except when manually debugging code.

jasnell: I'd say let's go with it as is. *IF* someone is broken, the entire change can be reverted and landed again behind a compat flag.

kentonv: Honestly these arguments are actually making me less comfortable with the change.

> IF someone is broken, the entire change can be reverted and landed again behind a compat flag.

This isn't the right mindset. Breaking people is not OK even if it's temporary, and we certainly cannot assume that a breakage will actually be visible in our metrics or reported to us. We have to be confident that we will not break anyone in the first place.

> It would be preferable to stick to the spec here and keep the check in order to stop bad code from seeping in. I think it's unlikely that anyone reads the value back in in the case of this API, except when manually debugging code.

This seems self-contradictory: If we think people are likely to write such code, we can't introduce the type check because it will break them. If we don't think people are likely to write such code, then the type check doesn't provide value so it doesn't matter if we introduce it or not.

My opinion is that this change won't break anyone -- with the corollary that it doesn't provide any value. I will approve the change as-is because I don't think it actually matters, but my recommendation would be to remove the type check.

kentonv (approved):

src/workerd/api/basics.c++
IgorMinar: I totally agree about the mindset. I don't want to speak for James, but I think he meant that he also doesn't believe that this will break anyone and that's why we should land it as is.

In any case, I agree that we should move on.

IgorMinar (approved):

src/workerd/api/basics.c++
jasnell: Going ahead with a change that temporarily accepts whatever value is given and logs if it's the wrong type. If we don't get any hits on the new log, we can keep the new type check. If we do get hits on it, then we either drop the check entirely or introduce a compat flag.

jasnell:
I'm not going to do it in this PR, but I think because this pattern is common to all whatwg `EventTarget` types, it would make sense to wrap this in a JSG helper macro, like `JSG_EVENT(abort)` that wraps the boilerplate.

PR #1849, opened 2024-03-15
The second argument to CloseEvent is supposed to be optional

jasnell:


fhanau (approved):
A pointer to [documentation](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/CloseEvent) for CloseEvent would have been helpful, but it was easy to find in this case.

PR #1850, opened 2024-03-16
Convert the standard properties on streams objects to prototype

jasnell:
Per the standard, these properties are really supposed to be prototype properties, not instance properties.

Technically this *could* be considered breaking... I think it's extremely unlikely to actually break anyone.


fhanau (approved):

jasnell:
There's an internal CI repo test failure that will need to be investigated before this lands

PR #1851, opened 2024-03-16
Disk Cache for Pyodide Wheels

garrettgu10:
Test plan:
```
$ mkdir ~/.cache/workerd
$ ./bazel-bin/src/workerd/server/workerd test --disk-cache-dir ~/.cache/workerd samples/pyodide-fastapi/config.capnp
# disable internet (either by blocking pub-45d734c4145d4285b343833ee450ef38.r2.dev in /etc/hosts or by switching off all internet access)
$ ./bazel-bin/src/workerd/server/workerd test --disk-cache-dir ~/.cache/workerd samples/pyodide-fastapi/config.capnp
# it should still work
```


garrettgu10:
I just found out KJ does in fact have file I/O, so I'm in the process of rewriting that section of the codebase

garrettgu10:
Only thing left to do is to get the cache root from the command line or from the capnp config and edit cacheRoot in server.c++ to make it dynamic

garrettgu10:
Having trouble testing since devspace is down, but the test plan is at the top and everything should be complete at this point

hoodmane (commented):
Generally looks good to me. Is there a downstream PR? Looks to me like you'll need to add a DisabledDiskCache or something like that so that the disk-cache import won't fail.

garrettgu10:
Relies on #1919

src/workerd/server/workerd.c++
```diff
@@ -927,6 +929,12 @@ public:
     server.enableControl(fd);
   }
 
+  void diskCacheDir(kj::StringPtr pathStr) {
+    kj::Path path = fs->getCurrentPath().eval(pathStr);
+    kj::Maybe<kj::Own<const kj::Directory>> dir = fs->getRoot().tryOpenSubdir(path, kj::WriteMode::MODIFY);
+    server.setDiskCacheRoot(kj::mv(dir));
+  }
```
dom96, line 936: Should this parameter have a default value pointing at `$TMPDIR/workerd-python-package-cache`?

hoodmane: On linux and mac I'd like to see it default to:
* `$XDG_CACHE_HOME/workerd-python-package-cache` if `XDG_CACHE_HOME` is defined
* `$HOME/.cache/workerd-python-package-cache` otherwise.

Unless you have a strong opinion that it should go in `/tmp` instead.

dom96: True, better to follow the XDG spec here.

For macOS I think we want `~/Library/Caches/` and I guess `%LOCALAPPDATA%` for Windows.

PR #1855, opened 2024-03-18
Implement node:test MockTracker

jasnell:
Requires the `nodejs_compat` compatibility flag to be enabled.

Does not yet support mock timers.

Does not implement the full `node:test` module, just the mocks

Refs: https://nodejs.org/docs/latest/api/test.html#class-mocktracker

Implementation was adapted directly from Node.js implementation. Only changes necessary to make it work in workers were made. Majority of new lines are the tests (which were also adapted directly from Node.js' tests). The API and implementation here are intended to match Node.js'.

While this could likely be polyfilled fairly easily, building it into the runtime allows us to take advantage of it for our own tests, much like we've started to do with `node:assert`

### DIFF
--- a/src/node/internal/mock.js
+++ b/src/node/internal/mock.js
@@ -1,0 +1,485 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+//
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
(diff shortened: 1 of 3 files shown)


src/node/internal/mock.js
```diff
@@ -0,0 +1,480 @@
...
+  validateBoolean,
+  validateFunction,
+  validateInteger,
+  validateObject,
+} from 'node-internal:validators';
+
+const kEmptyObject = Object.create(null);
+function kDefaultFunction() {}
+
+//const { MockTimers } = require('internal/test_runner/mock/mock_timers');
```
fhanau, line 44: I understand that this is not implemented yet – we should have a comment mentioning this here and where timers() and the timers?.reset() call are commented out.

src/workerd/api/node/assert-test.js
```diff
@@ -1,3 +1,31 @@
...
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT ORs
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+/* todo: the following is adopted code, enabling linting one day */
+/* eslint-disable */
+
```
fhanau, line 28: Since much of this test is adopted code – do you think pointing to the file path in node (`test/parallel/test-runner-mocking.js`) would help for maintaining this?

fhanau:
- Does MockTimers need TS bindings?
- MockTimers is `Stability: 1 - Experimental` and may have an unstable API – does this need to be guarded by e.g. an experimental flag or does `nodejs_compat` cover cases like this?
- Tests look good, `local mocks are auto restored after the test finishes` wasn't adopted but I assume that's intentional.

LGTM otherwise, seems pretty helpful for developers to me.

jasnell:
I wasn't planning on implementing `MockTimers` for a while. I think just omitting those bits makes the most sense. I'll add todo comments.

Not porting that one test was intentional. The way it is structured it would need to be rewritten since wd-test has no concept of subtests/before/after etc so I wanted to look at it separately.

src/workerd/api/node/assert-test.js
jasnell: Node.js implementation could move around, files could be renamed, etc. Not really sure a link over would be too helpful here.

fhanau (approved):

PR #1858, opened 2024-03-18
Pass scriptId in Trace

zebp:
Passes `scriptId` and `sourcemapsAvailable` through various usages of `Trace`.

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -243,12 +243,13 @@ Trace::Exception::Exception(kj::Date timestamp, kj::String name, kj::String mess
 
 Trace::Trace(kj::Maybe<kj::String> stableId, kj::Maybe<kj::String> scriptName,
     kj::Maybe<kj::Own<ScriptVersion::Reader>> scriptVersion,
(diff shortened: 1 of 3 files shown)


src/workerd/io/trace.c++
```diff
@@ -259,6 +260,7 @@ void Trace::copyTo(rpc::Trace::Builder builder) {
     }
   }
 
+  builder.setSourcemapsAvailable(sourcemapsAvailable);
```
mikea, line 263: let's remove sourcemapsAvailable from here, since we solved this problem downstream.

jasnell (approved):

jclee:
I'm not sure exactly how the new field is intended to be consumed, but it might be worth noting that `scriptId` is an internal ID, which isn't intended to be exposed to users.

zebp:
> I'm not sure exactly how the new field is intended to be consumed, but it might be worth noting that `scriptId` is an internal ID, which isn't intended to be exposed to users.

The new field is consumed by a the sourcemap stacktrace remapping system within the internal runtime and is not exposed to users at any point

PR #1861, opened 2024-03-19
Prevent queuing writes while internal writable is being piped

jasnell:
In some uses we end up enqueuing a write event to a `WritableStreamInternalController` while it is being piped to, leading to an assertion given that we expect the `Pipe` event to be the last event in the queue while it is being piped. To match the behavior seen when using the `WritableStreamDefaultWriter`, we should reject attempts to write, close, or flush while the pipe is still going.

Also improves the logging around the assertion that was being hit and fixes another apparent latent bug in the pipe source close check.

### DIFF
--- a/src/workerd/api/streams/internal-test.c++
+++ b/src/workerd/api/streams/internal-test.c++
@@ -8,6 +8,7 @@
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/jsg-test.h>
 #include <openssl/rand.h>
+#include <workerd/tests/test-fixture.h>
 
 namespace workerd::api {
 namespace {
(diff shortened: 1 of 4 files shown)


src/workerd/api/streams/internal-test.c++
```diff
@@ -185,5 +186,72 @@ KJ_TEST("honest small stream") {
   KJ_ASSERT(stream.maxMaxBytesSeen(), 100);
 }
 
+KJ_TEST("WirtableStreamInternalController queue size assertion") {
```
jasnell, line 189: Just confirming that without the fix using the `isPiping()` checks in this PR, this test will fail with the queue.size assertion

src/workerd/api/streams/internal.c++
```diff
@@ -1409,6 +1431,23 @@ jsg::Promise<void> WritableStreamInternalController::writeLoopAfterFrontOutputLo
...
+          KJ_CASE_ONEOF(flush, Flush) { events.add(kj::str("Flush")); }
+          KJ_CASE_ONEOF(close, Close) { events.add(kj::str("Close")); }
+          KJ_CASE_ONEOF(pipe, Pipe) { events.add(kj::str("Pipe")); }
+        }
+      }
+      return kj::str("Too many events in internal writablestream queue: ",
+                     kj::delimited(kj::mv(events), ", "));
+    }
+    return kj::String();
+  };
```
jasnell, line 1449: This method just enhances the logging for the assertion.

src/workerd/api/streams/internal.c++
```diff
@@ -1520,14 +1562,12 @@ jsg::Promise<void> WritableStreamInternalController::writeLoopAfterFrontOutputLo
...
-          } else {
-            writeState.init<Unlocked>();
-          }
-          return js.resolvedPromise();
+        if (!request.preventClose && !isClosedOrClosing()) {
+          doClose(js);
+        } else {
+          writeState.init<Unlocked>();
         }
+        return js.resolvedPromise();
```
jasnell, line 1570: This changed block here appears to be a latent bug in the handling of the source being closed. Hitting this is very much an edge case so shouldn't have much impact but definitely worth fixing.

jasnell:
hmm, handful of test failures that appear to be related to me adding the `TestFixture` to `internal-test.c++`... weird. I'll fix that up in the morning.

dom96 (approved):

PR #1862, opened 2024-03-19
Implements ReadableStream.from(gen)

jasnell:
Web platform standard API for creating a ReadableStream from an async or sync generator/iterable.

https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/from_static

```js
async function* gen() {
  yield 1;
  yield 2;
  yield 3;
}

const readable = ReadableStream.from(gen());
for await (const chunk of readable) {
  console.log(chunk);  // 1, 2, 3
}
```

```js
const readable = ReadableStream.from([1, 2, 3]);
for await (const chunk of readable) {
  console.log(chunk);  // 1, 2, 3
}
```


jasnell:
Looks like there's an ASAN issue to investigate.

src/workerd/api/tests/streams-test.js
```diff
@@ -316,6 +316,104 @@ export const writableStreamGcTraceFinishes = {
...
+    const chunks = [];
+    for await (const chunk of rs) {
+      chunks.push(chunk);
+      return;
+    }
+    deepStrictEqual(chunks, ['hello']);
+  }
+};
+
+export const readableStreamFromThrowingAsyncGen = {
```
harrishancock, line 385: Does it make sense to test the throwing sync case too?

src/workerd/jsg/iterator.h
```diff
@@ -445,6 +451,92 @@ class AsyncGenerator final {
   friend class GeneratorWrapper;
 };
 
+template <typename T>
+Promise<kj::Maybe<T>> AsyncGenerator<T>::next(Lock& js) {
+  KJ_IF_SOME(i, impl) {
+    KJ_SWITCH_ONEOF(i->state) {
+      KJ_CASE_ONEOF(finsihed, typename Impl::Finished) {
```
harrishancock, line 458: Nit
```suggestion
      KJ_CASE_ONEOF(finished, typename Impl::Finished) {
```
Or maybe just `_`.

harrishancock (approved):

src/workerd/api/tests/streams-test.js
jasnell: We could but it's going to end up exercising exactly the same code paths so I'm not sure it's actually worth it.

PR #1863, opened 2024-03-19
Remove Cache API PUT limit

fhanau:
After further discussions within the team, we will be lifting the limit entirely as limits defined in the Cache backend will be sufficient. This enables significant cleanup, including the limit functions defined in #1772 and the existing cache stream setup.

Also see the corresponding downstream PR.

### DIFF
--- a/src/workerd/io/io-context.c++
+++ b/src/workerd/io/io-context.c++
@@ -122,16 +122,9 @@ IoContext::IoContext(ThreadContext& thread,
       limitEnforcer(kj::mv(limitEnforcerParam)),
       threadId(getThreadId()),
       deleteQueue(kj::atomicRefcounted<DeleteQueue>()),
-      cachePutQuota(nullptr),
+      cachePutSerializer(kj::READY_NOW),
       waitUntilTasks(*this),
       timeoutManager(kj::heap<TimeoutManagerImpl>()) {
-  KJ_IF_SOME(putLimit, limitEnforcer->getCachePUTLimitMB()) {
-    initialPutQuota = putLimit * MB;
-  } else {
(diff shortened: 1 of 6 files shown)


src/workerd/io/io-context.c++
```diff
@@ -1222,68 +1213,38 @@ public:
...
+  return awaitIo(js, cachePutSerializer.then(
+      [fulfiller = kj::mv(paf.fulfiller), stream = kj::mv(stream)]() mutable
+          -> kj::Own<kj::AsyncInputStream> {
+    if (stream->tryGetLength() != kj::none) {
+      // PUT with Content-Length. We can just return immediately, allowing the next PUT to start.
+      KJ_DEFER(fulfiller->fulfill());
+      return kj::mv(stream);
     } else {
+      // TODO(later): With Cache streams no longer having a size limit enforced by the runtime,
+      // explore if we can clean up stream serialization too.
```
harrishancock, line 1241: Yes, I think we can absolutely get rid of stream serialization. Its only purpose was to guarantee a limit on the total number of bytes in a conservative way. Since we no longer impose a total limit on Content-Length requests, we can treat Transfer-Encoding: chunked requests the same.

harrishancock (approved):

src/workerd/io/io-context.c++
fhanau: Sounds good – removing serialization looks non-trivial based on the amount of code used for it in api/cache.c++, but that also means it would be pretty helpful, seems to reduce code size for cache significantly too. Internal tests that check for serialization will also need to be updated. For this PR I just changed the code comments to refer to serialization and added the TODOs to keep track of this.

PR #1865, opened 2024-03-19
Implements changes needed for Python mem snapshot in validator.

dom96:
Some small changes to make it possible for validator to get the memory snapshot.

Tested upstream.

### DIFF
--- a/src/pyodide/internal/python.js
+++ b/src/pyodide/internal/python.js
@@ -37,7 +37,7 @@ import pyodideWasmModule from "pyodide-internal:generated/pyodide.asm.wasm";
  */
 import stdlib from "pyodide-internal:generated/python_stdlib.zip";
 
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/python.js
```diff
@@ -82,6 +82,10 @@ export async function uploadArtifacts() {
   }
 }
 
+export function getMemorySnapshot() {
+  return new Uint8Array(MEMORY).slice();
```
hoodmane, line 86:
```suggestion
  return MEMORY;
```
Otherwise, this does two unnecessary copies:
1. we copied the heap when we assigned to it; we can do whatever we want with it now don't have to copy again
2. `MEMORY` is already a `Uint8Array` so `new Uint8Array(MEMORY)` copies it once and `slice()` copies it again.

There's even further copy reductions available now but I can do them in a different PR =)

hoodmane (approved):
Looks great, thanks @dom96! See my one comment.

PR #1868, opened 2024-03-20
Python perf: Don't report undefined symbols

hoodmane:
In flame graphs this single call frame takes up a lot of loading time. It also is not necessary except for debugging purposes. This gets rid of it.


dom96 (approved):

garrettgu10:
I think this broke local dev, I'll make a PR to fix

PR #1872, opened 2024-03-20
Python snapshots: Only load dynamic libraries that are needed

hoodmane:
Rather than actually preloading all libraries, we just preallocate space for them. There is a function called `getMemory` that determines the location of the dynamic libraries and nothing else so we can patch this to ensure that the libraries get allocated in their dedicated location if they are loaded at all. This allows us to ensure that their metadata always lands in the right spot.

We still have to make sure to load all the libraries in the correct order. There is also the possibility that someone could use ctypes and mess everything up. Though ctypes doesn't work with our snapshots before this PR. I guess it could be fixed by patching libffi to record the trampoline address and function table slot into the DSO_METADATA and then recreate them the same way when restoring the snapshot. We'd also need to record the function table base for each loaded library. Once we do all this, we should be safe again...


dom96:
I also wonder if we could use capnp for the memory snapshot in JS

PR #1874, opened 2024-03-21
MQ-576: Add attempts to QueueMessage

sesteves:
This patch updates the queue message capnp schema so that the number of attempts is included within each message, allowing queue consumers to implement backoff logic based on that. Note this is a breaking change and we are making sure that all clients are updated accordingly before this change gets released.

### DIFF
--- a/src/workerd/api/queue.h
+++ b/src/workerd/api/queue.h
@@ -98,7 +98,8 @@ struct IncomingQueueMessage {
   kj::Date timestamp;
   kj::Array<kj::byte> body;
   kj::Maybe<kj::String> contentType;
-  JSG_STRUCT(id, timestamp, body, contentType);
+  uint16_t attempts;
+  JSG_STRUCT(id, timestamp, body, contentType, attempts);
 
   struct ContentType {
     static constexpr kj::StringPtr TEXT = "text"_kj;
@@ -158,6 +159,7 @@ public:
   kj::StringPtr getId() { return id; }
   kj::Date getTimestamp() { return timestamp; }
   jsg::JsValue getBody(jsg::Lock& js);
+  uint16_t getAttempts() { return attempts; };
 
   void retry(jsg::Optional<QueueRetryOptions> options);
   void ack();
@@ -168,6 +170,7 @@ public:
     JSG_READONLY_INSTANCE_PROPERTY(id, getId);
     JSG_READONLY_INSTANCE_PROPERTY(timestamp, getTimestamp);
     JSG_READONLY_INSTANCE_PROPERTY(body, getBody);
+    JSG_READONLY_INSTANCE_PROPERTY(attempts, getAttempts);
     JSG_METHOD(retry);
     JSG_METHOD(ack);
 
@@ -186,6 +189,7 @@ private:
   kj::String id;
   kj::Date timestamp;
   jsg::JsRef<jsg::JsValue> body;
+  uint16_t attempts;
   IoPtr<QueueEventResult> result;
 
   void visitForGc(jsg::GcVisitor& visitor) {
--- a/src/workerd/io/worker-interface.capnp
+++ b/src/workerd/io/worker-interface.capnp
@@ -146,6 +146,7 @@ struct QueueMessage @0x944adb18c0352295 {
   timestampNs @1 :Int64;
   data @2 :Data;
   contentType @3 :Text;
+  attempts @4 :UInt16;
 }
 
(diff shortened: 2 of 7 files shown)


src/workerd/io/worker-interface.capnp
```diff
@@ -146,6 +146,7 @@ struct QueueMessage @0x944adb18c0352295 {
   timestampNs @1 :Int64;
```
sesteves, line 146: Need some guidance regarding the id @0x944adb18c0352295

a-robinson: What guidance do you need? We shouldn't need to do anything with the ID, all we're doing is adding a field to the struct.

src/workerd/api/queue.h
```diff
@@ -98,7 +98,8 @@ struct IncomingQueueMessage {
   kj::Date timestamp;
   kj::Array<kj::byte> body;
   kj::Maybe<kj::String> contentType;
-  JSG_STRUCT(id, timestamp, body, contentType);
+  uint16_t attempts;
```
a-robinson, line 101: I assume this will be documented in the docs as well, but does this start at 0 for the first attempt or at 1? It'd be worth noting here so that it's in the code as well as in the docs.

Perhaps the ambiguity suggests a different name would be better (`previousAttempts`, `pastAttempts`, etc.), but those come with their own downside in terms of wordiness, and they assume that this starts at 0, but judging by your tests I'm guessing this actually starts at 1 on the first attempt.

a-robinson (approved):
> Note this is a breaking change and we are making sure that all clients are updated accordingly before this change gets released.

What about this is a breaking change?

1. If the client doesn't send the field but the server expects it, it'll just default to 0
2. If the client does send the field but the server doesn't check for it, it'll behave as if the client hadn't sent the field.

Both of which seem fine to me?

src/workerd/io/worker-interface.capnp
```diff
@@ -146,6 +146,7 @@ struct QueueMessage @0x944adb18c0352295 {
...
   data @2 :Data;
   contentType @3 :Text;
+  attempts @4 :UInt16;
```
a-robinson, line 149: If the answer to https://github.com/cloudflare/workerd/pull/1874/files#r1534621760 is that the first attempt of a message is meant to be delivered with an `attempts` of 1, then perhaps we should just default this field to 1 when it isn't explicitly set by the client? By doing:

```
  attempts @4 :UInt16 = 1;
```

sesteves:
> > Note this is a breaking change and we are making sure that all clients are updated accordingly before this change gets released.
> 
> What about this is a breaking change?
> 
>     1. If the client doesn't send the field but the server expects it, it'll just default to 0
> 
>     2. If the client does send the field but the server doesn't check for it, it'll behave as if the client hadn't sent the field.
> 
> 
> Both of which seem fine to me?

Regarding (1), I tested it and the attempts field does not default to zero, it throws an error, something like:  `TypeError: The value cannot be converted because it is not an integer.`. In any case, this will be a non-optional field on the queues side of things, and it should be always sent.

src/workerd/io/worker-interface.capnp
sesteves: It must be always set by the client (we already made the changes), o/w we may be opening the door to bugs.

src/workerd/api/queue.h
a-robinson: I thought that comments here got automatically picked up and put into the typescript types for public consumption, but now that I look at https://www.npmjs.com/package/@cloudflare/workers-types?activeTab=code I see that most of the types don't have comments on their fields, so I guess it's not a big deal. Plus it'd be nice to merge these before conflicts start happening, so I'll get this in now.

a-robinson:
> Regarding (1), I tested it and the attempts field does not default to zero, it throws an error, something like: TypeError: The value cannot be converted because it is not an integer.

That could be changed by making the `attempts` field into a `jsg::Optional<uint16_t>` here instead of just a `uint16_t`: https://github.com/cloudflare/workerd/pull/1874/files#diff-281d27b2a052c66055b1152a6c2b33c9baf601c47aa164482f4fbbc1f094f77cR528

But it sounds like that doesn't matter if the queues code is already updated, so let's get this merged before any new conflicts creep in.

irvinebroque:
Think this change needs Miniflare changes in order to be able to land in local dev, see test failures in https://github.com/cloudflare/workers-sdk/pull/5451

PR #1875, opened 2024-03-21
Python: Reduce memory usage when restoring snapshot

hoodmane:
Previously we paid for two copies of the snapshot memory: one copy in the wasm linear memory itself and a second copy in `BUNDLE_MEMORY_SNAPSHOT`. This ensures that we never have more memory than one copy of the linear memory heap by copying the memory directly from the snapshot to the linear memory. We also release the C++ memory when we are done with it.


src/workerd/api/pyodide/pyodide.h
```diff
@@ -33,16 +41,16 @@ class PyodideMetadataReader : public jsg::Object {
...
 public:
   PyodideMetadataReader(kj::String mainModule, kj::Array<kj::String> names,
                         kj::Array<kj::Array<kj::byte>> contents, kj::Array<kj::String> requirements,
                         bool isWorkerd, bool isTracing,
-                        kj::Maybe<kj::Array<kj::byte>> memorySnapshot)
+                        kj::Maybe<kj::Array<kj::byte>> snapshot)
       : mainModule(kj::mv(mainModule)), names(kj::mv(names)), contents(kj::mv(contents)),
         requirements(kj::mv(requirements)), isWorkerdFlag(isWorkerd), isTracingFlag(isTracing),
-        memorySnapshot(kj::mv(memorySnapshot)) {}
+        memorySnapshot(maybeArrayToArray(kj::mv(memorySnapshot))) {}
```
dom96, line 53: you need to s/memorySnapshot/snapshot/ here I think

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -56,6 +58,32 @@ int PyodideMetadataReader::read(jsg::Lock& js, int index, int offset, kj::Array<
...
+  if (offset >= snapshotSize || offset < 0) {
+    return 0;
+  }
+  int toCopy = buf.size();
+  if (snapshotSize - offset < toCopy) {
+    toCopy = snapshotSize - offset;
+  }
+  memcpy(buf.begin(), &existingSnapshot[0] + offset, toCopy);
+  return toCopy;
+}
```
dom96, line 85: These two are practically identical, can we refactor to avoid duplication fairly easily here?

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -56,6 +58,32 @@ int PyodideMetadataReader::read(jsg::Lock& js, int index, int offset, kj::Array<
...
+int PyodideMetadataReader::readMemorySnapshot(int offset, kj::Array<kj::byte> buf) {
+  int snapshotSize = memorySnapshot.size();
+  if (offset >= snapshotSize || offset < 0) {
+    return 0;
+  }
+  int toCopy = buf.size();
+  if (snapshotSize - offset < toCopy) {
+    toCopy = snapshotSize - offset;
+  }
+  memcpy(buf.begin(), &memorySnapshot[0] + offset, toCopy);
```
dom96, line 70: Hmm, I'm concerned this may be unsafe.

@jasnell the `buf` here comes from JS, can you verify that doing a `memcpy` like this is safe?

src/pyodide/internal/python.js
```diff
@@ -440,13 +433,22 @@ function encodeSnapshot(heap, dsoJSON) {
...
+  MEMORY_SNAPSHOT_READER.readMemorySnapshot(0, buf);
+  const snapshotOffset = buf[0];
+  const snapshotSize = MEMORY_SNAPSHOT_READER.getMemorySnapshotSize() - snapshotOffset;
+  const jsonLength = buf[1];
+  const jsonBuf = new Uint8Array(jsonLength);
+  MEMORY_SNAPSHOT_READER.readMemorySnapshot(8, jsonBuf);
+  DSO_METADATA = JSON.parse(new TextDecoder().decode(jsonBuf));
+  READ_MEMORY = function(Module) {
+    // resize linear memory to fit our snapshot.
+    Module.growMemory(snapshotSize);
```
dom96, line 447: This allocates `snapshotSize` of memory, right? So we'll have 2*snapshotSize of memory allocated here (though not all of it will be attributed to the Worker I guess)

Is there a way we could avoid that? Could we directly assign the snapshot to `Module.HEAP8`?

src/workerd/api/pyodide/pyodide.h
```diff
@@ -68,6 +72,18 @@ class PyodideMetadataReader : public jsg::Object {
...
 
+  bool hasMemorySnapshot() {
+    return memorySnapshot.size() > 0;
+  }
+  int getMemorySnapshotSize() {
+    return memorySnapshot.size();
+  }
+
+  void disposeMemorySnapshot() {
+    memorySnapshot = kj::heapArray<kj::byte>(0);
```
dom96, line 83: I believe just doing `kj::mv(memorySnapshot)` may be slightly more efficient.

But really this makes me wonder why you got rid of the `kj::Maybe`. We might get a memory snapshot of size 0 at some point, it would be good to be able to disambiguate this vs. the snapshot being actually missing. And then you'll be able to just assign `kj::none` here.

src/workerd/api/pyodide/pyodide.c++
hoodmane: We already have an almost identical copy of this code elsewhere in this same file which James reviewed =)

src/workerd/api/pyodide/pyodide.h
hoodmane: We can't have a memory snapshot of size zero, what would that even mean? The heap is currently always at least 20 mb.

dom96: I'm thinking if one of the many layers that does the writing and storing of the memory snapshot has a bug that causes the snapshot to be set to 0 bytes.

It would just give us a clue that it's not that the snapshot isn't being generated/is missing, but rather that it's corrupted.

In general `kj::Maybe` here just seems more ideal, you can get rid of the `maybeArrayToArray` function then too.

hoodmane: Well `kj::Maybe` is kind of annoying to use. I need extra branches and `KJ_IF_SOME` and stuff. My problem before was I couldn't remember how to `.unwap()` the optional. Arrays already have empty arrays as a natural "not present" value and it's more ergonomic to use that way. I guess I can do
```C++
if (val == kj::none) {
   return;
}
auto & blah = KJ_REQUIRE_NONNULL(val);
```
but there is a lot of api surface here for me to figure out.

hoodmane: As it is though, even if we can distinguish "empty snapshot" from "no snapshot" we'll treat the empty snapshot as a test. It doesn't really make sense to have empty snapshots, it's not a valid value, so I think it makes the most sense to just use "empty array" as the missing value here. Makes all the code a bit shorter and I don't really see any downside.

dom96: > Well kj::Maybe is kind of annoying to use.

It is something we use throughout the code base though, so avoiding its use makes the code less consistent with the rest of the codebase (and thus harder for others understand), and I would say is actually less ergonomic.

Plus with your change, the code now has `storedSnapshot` which is a `kj::Maybe<Array>`, but `existingSnapshot` is a `kj::Array` where a size of 0 means the snapshot is missing.

I think the extra code you'll need to handle the `kj::none` explicitly isn't particularly long.

But anyway, I don't think it's worth blocking the change over this, I just think we should in general use `kj::Maybe` in situations like this.

dom96 (approved):

src/pyodide/internal/python.js
hoodmane: It appears that `INITIAL_MEMORY` is an Emscripten runtime setting that controls the size of the memory at startup. If we set this we can avoid the call to `growMemory` which is going to save some work.

src/workerd/api/pyodide/pyodide.h
```diff
@@ -116,7 +135,7 @@ class ArtifactBundler : public jsg::Object {
 
   ArtifactBundler(bool isValidating = false)
       : storedSnapshot(kj::none),
-        existingSnapshot(kj::none),
+        existingSnapshot(kj::heapArray<kj::byte>(0)),
```
dom96, line 138:
```suggestion
        existingSnapshot(kj::none),
```

hoodmane: Thanks!

PR #1876, opened 2024-03-21
Make DOMException serializable/cloneable

jasnell:
Makes DOMException cloneable for use with jsrpc and `structuredClone()`

### DIFF
--- a/src/workerd/io/worker-interface.capnp
+++ b/src/workerd/io/worker-interface.capnp
@@ -225,6 +225,11 @@ enum SerializationTag {
   headers @4;
   request @5;
   response @6;
+
+  domException @7;
+  # Keep this value in sync with the DOMException::SERIALIZATION_TAG in
+  # /src/workerd/jsg/dom-exception (but we can't actually change this value
+  # without breaking things).
 }
 
 enum StreamEncoding {
--- a/src/workerd/jsg/dom-exception.c++
+++ b/src/workerd/jsg/dom-exception.c++
@@ -3,6 +3,7 @@
 //     https://opensource.org/licenses/Apache-2.0
 
 #include "dom-exception.h"
+#include "ser.h"
 #include <workerd/jsg/memory.h>
 #include <kj/string.h>
 #include <map>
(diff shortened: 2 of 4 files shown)


jasnell:
Doh.. forgot to run the whole test suite locally lol... there's a build error to fix up...

fhanau (approved):
LGTM.

src/workerd/jsg/dom-exception.h
```diff
@@ -112,6 +113,15 @@ class DOMException: public Object {
     tracker.trackField("errorForStack", errorForStack);
   }
 
+  // The value is taken from worker-interface.capnp, which we can't depend
+  // on directly here so we have to set it manually.
+  static const uint SERIALIZATION_TAG = 0x7;
```
kentonv, line 118: Nitpick: I don't think there's any reason for this to be hex -- the `@N` tags in Cap'n Proto are decimal.

```suggestion
  static const uint SERIALIZATION_TAG = 7;
```

src/workerd/io/worker-interface.capnp
```diff
@@ -216,6 +216,10 @@ enum SerializationTag {
   headers @4;
   request @5;
   response @6;
+
+  domException @7;
+  # Keep this value in sync with the DOMException::SERIALIZATION_TAG in
```
kentonv, line 221: Not that the value can ever change anyway...

src/workerd/jsg/dom-exception.h
```diff
@@ -112,6 +113,15 @@ class DOMException: public Object {
     tracker.trackField("errorForStack", errorForStack);
   }
 
+  // The value is taken from worker-interface.capnp, which we can't depend
```
kentonv, line 116: I guess the reason we can't depend on it directly is because `DomException` is defined in `jsg` instead of under `api`?

Why is that? It kind of feels like it actually belongs under `api`, TBH. (Not that I'm suggesting changing it in this PR.)

kentonv (approved):

src/workerd/jsg/dom-exception.c++
```diff
@@ -49,4 +50,20 @@ void DOMException::visitForGc(GcVisitor& visitor) {
   visitor.visit(errorForStack);
 }
 
+void DOMException::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
+  serializer.writeLengthDelimited(message);
+  serializer.writeLengthDelimited(name);
+  serializer.write(js, JsValue(errorForStack.getHandle(js)));
```
kentonv, line 56: Note that `errorForStack` contains a copy of `message`, so that string will be duplicated here. Should we avoid that? Maybe don't bother writing `message` separately, just have it come from `errorForStack`?

src/workerd/jsg/dom-exception.h
jasnell: It really *should* be in `DOMException` and I just haven't had enough justification to move it. We certainly can tho.

src/workerd/jsg/dom-exception.c++
jasnell: Well, not sure it matters too much. In the coming weeks I'm going to be pursuing the contribution to v8 that will eliminate the need to have `errorForStack` and attach the `stack` property directly to `DOMException`'s JS wrapper, in which case we'd want to preserve `message`. For now, I think we should just keep this as is and we can tweak later.

src/workerd/io/worker-interface.capnp
jasnell: heh, true. clarified in a comment

src/workerd/jsg/dom-exception.h
jasnell: Added a comment. Will look at moving `DOMException` later.

PR #1877, opened 2024-03-22
Add Workers Ai fetcher api

G4brym:


G4brym:

G4brym:
recheck

src/cloudflare/internal/ai-api.ts
```diff
@@ -0,0 +1,124 @@
...
+      },
+    });
+
+    const fetchOptions = {
+      method: "POST",
+      body: body,
+      headers: {
+        ...(this.options?.sessionOptions?.extraHeaders || {}),
+        ...(this.options?.extraHeaders || {}),
+        "content-encoding": "application/json",
```
jasnell, line 66: `application/json` is not a valid `content-encoding` value. Do you mean `content-type`?

src/cloudflare/internal/ai-api.ts
```diff
@@ -0,0 +1,124 @@
...
+    const res = await this.fetcher.fetch("http://workers-binding.ai/run?version=2", fetchOptions);
+
+    this.lastRequestId = res.headers.get("cf-ai-req-id");
+
+    if (inputs['stream']) {
+      if (!res.ok) {
+        throw new InferenceUpstreamError(await res.text(), res.status);
+      }
+
+      return res.body;
```
jasnell, line 82: Should this be checking the status of the response?

src/cloudflare/internal/ai-api.ts
```diff
@@ -0,0 +1,124 @@
...
+        this.logs = parsedLogs;
+      }
+
+      if (!res.ok || !res.body) {
+        throw new InferenceUpstreamError(await res.text(), res.status);
+      }
+
+      // Non streaming responses are always in gzip
+      // eslint-disable-next-line
+      const decompressed = new Response(res.body.pipeThrough(new DecompressionStream("gzip")));
```
jasnell, line 105: Why create a separate `Response` here? And is passing it through `DecompressionStream` necessary? If the transfer encoding is identified correctly does the `res.body` already automatically decompress the gzip?

src/cloudflare/internal/ai-api.ts
G4brym: I ment `content-type`, fixed

src/cloudflare/internal/ai-api.ts
G4brym: didn't know fetch would automatically decompress the gzip based on that header, but tried it now locally with the header on the server and it worked fine 👍 
fixed

irvinebroque:
More tests seem worthwhile here?

src/cloudflare/internal/test/ai/ai-api-test.wd-test
```diff
@@ -0,0 +1,36 @@
...
+        modules = [
+          (name = "worker", esModule = embed "ai-api-test.js")
+        ],
+        compatibilityDate = "2023-01-15",
+        compatibilityFlags = ["nodejs_compat"],
+        bindings = [
+        (
+          name = "ai",
+          wrapped = (
+            moduleName = "cloudflare-internal:ai-api",
```
dom96, line 16: hmm, `internal` implies it shouldn't be used by end users. But this isn't the case here, right? Maybe this shouldn't be an internal module?

G4brym: agreed, i've just moved this module to outside the internal folder

src/cloudflare/internal/ai-api.ts
```diff
@@ -0,0 +1,124 @@
...
+};
+
+export class InferenceUpstreamError extends Error {
+  public httpCode: number;
+
+  public constructor(message: string, httpCode: number) {
+    super(message);
+    this.name = "InferenceUpstreamError";
+    this.httpCode = httpCode;
+  }
```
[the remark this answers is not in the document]
G4brym: We are fine to lose `httpCode` and `InferenceUpstreamError` class as long as the exception message is still passed on

src/cloudflare/internal/test/ai/ai-api-test.wd-test
G4brym: humm, but end users shouldn't be able to initialize the Ai class on themself, it should be done by wrangler automatically when they define the AI binding
I've kept the code in the internal folder, but added a mapping outside to end users could import the options and Ai class from `cloudflare:ai`

G4brym:
@irvinebroque i've just added more tests now

src/cloudflare/internal/test/ai/ai-api-test.wd-test
irvinebroque: What do other APIs do? What is the pattern we want people to use for this @mikea ?

mikea (approved):
code (and test) structure LG to me. I don't have much opinion on the API itself or its implementation detail.

src/cloudflare/internal/test/ai/ai-api-test.wd-test
G4brym: i'm currently having some issues importing this module from wrangler iv'e tried with `cloudflare:ai` and `cloudflare-internal:ai-api`
image

G4brym:
Was able to successfully test this pull request together with the [wrangler pull request here](https://github.com/cloudflare/workers-sdk/pull/5371) and i've confirmed that it is working as expected

irvinebroque:
Can we merge this so that it can go out in a workerd release for Wrangler, and then EW?

G4brym:
Everything ready from my side

PR #1878, opened 2024-03-22
Python snapshots: Add workerd cli options to save memory snapshots to disk

hoodmane:


kentonv:
Since this doesn't affect production, is an autogate actually necessary? Autogates exist for the purpose of rolling things out to production.

Maybe this behavior should be enabled by something in workerd/server/server.c++ instead?

hoodmane:
@kentonv are there other development-only features that I could look at for inspiration?

kentonv:
> are there other development-only features that I could look at for inspiration?

Perhaps `durableObjectStorage` in `workerd.capnp`? Configures local disk storage for durable objects.

garrettgu10:
Seems like there's some overlap with #1851. Can you take a look?

hoodmane:
Yeah.

src/pyodide/internal/snapshot.js
```diff
@@ -412,3 +417,15 @@ export function finishSnapshotSetup(pyodide) {
...
+
+/**
+ * Store the memory snapshot in the ArtifactBundler so that the validator can read it out.
+ * Needs to happen at the top level because the validator does not perform requests.
+ */
+export function maybeStoreMemorySnapshot() {
+  if (ArtifactBundler.isEwValidating()) {
+    ArtifactBundler.storeMemorySnapshot(getMemoryToUpload());
+  } else if (SHOULD_SNAPSHOT_TO_DISK) {
+    DiskCache.put("snapshot.bin", getMemoryToUpload());
```
garrettgu10, line 429: Not blocking, but I think we should name the snapshot something different for each worker, since eventually we want to load the snapshot from disk cache in addition to storing it. I think it may make sense to turn the `python-save-snapshot` flag into a string where the string is the file name, or use some other unique identifier for the worker.

garrettgu10 (approved):

src/pyodide/internal/snapshot.js
hoodmane: Yeah for now I just want some way to get snapshots onto disk that I can use for tests. Note that this doesn't even implement a way to load the snapshot in workerd, only to save it. Which is probably much higher priority than giving the files distinct names...

src/pyodide/internal/snapshot.js
```diff
@@ -412,3 +417,15 @@ export function finishSnapshotSetup(pyodide) {
     });
   }
 }
+
+/**
+ * Store the memory snapshot in the ArtifactBundler so that the validator can read it out.
+ * Needs to happen at the top level because the validator does not perform requests.
```
dom96, line 423: This comment needs to be adjusted to mention the saving to disk.

src/pyodide/python-entrypoint-helper.js
```diff
@@ -201,12 +200,7 @@ try {
       }
     }
   }
-
-  // Store the memory snapshot in the ArtifactBundler so that the validator can read it out.
-  // Needs to happen at the top level because the validator does not perform requests.
-  if (ArtifactBundler.isEwValidating()) {
-    ArtifactBundler.storeMemorySnapshot(getMemoryToUpload());
-  }
+  maybeStoreMemorySnapshot();
```
dom96, line 203: Actually I wonder if it wouldn't be better to keep the comment here. Just to make this function call more visible.

src/workerd/server/server.h
```diff
@@ -93,7 +102,11 @@ class Server: private kj::TaskSet::ErrorHandler {
   kj::Network& network;
   kj::EntropySource& entropySource;
   kj::Function<void(kj::String)> reportConfigError;
-  kj::Maybe<kj::Own<const kj::Directory>> diskCacheRoot;
+  PythonConfig pythonConfig = PythonConfig {
+    .diskCacheRoot = kj::none,
+    .createSnapshot =false,
```
dom96, line 107:
```suggestion
    .createSnapshot = false,
```

src/workerd/server/workerd-api.c++
```diff
@@ -105,11 +107,18 @@ JSG_DECLARE_ISOLATE_TYPE(JsgWorkerdIsolate,
   jsg::NodeJsModuleObject,
   jsg::NodeJsModuleContext);
 
+
+static PythonConfig defaultConfig {
+  .diskCacheRoot=kj::none,
+  .createSnapshot=false,
+  .createBaselineSnapshot=false,
```
dom96, line 114:
```suggestion
  .diskCacheRoot = kj::none,
  .createSnapshot = false,
  .createBaselineSnapshot = false,
```

src/workerd/server/workerd.c++
```diff
@@ -727,8 +727,12 @@ public:
         .addOption({"experimental"}, [this]() { server.allowExperimental(); return true; },
                    "Permit the use of experimental features which may break backwards "
                    "compatibility in a future release.")
-        .addOptionWithArg({"disk-cache-dir"}, CLI_METHOD(diskCacheDir), "<path>",
-                  "Use <path> as a disk cache to avoid repeatedly fetching packages from the internet. ");
+        .addOptionWithArg({"disk-cache-dir"}, CLI_METHOD(setPythonDiskCacheDir), "<path>",
+                  "Use <path> as a disk cache to avoid repeatedly fetching packages from the internet. ")
+        .addOption({"python-save-snapshot"}, [this]() { server.setPythonCreateSnapshot();  return true; },
+                  "Save a snapshot to the disk cache")
```
dom96, line 733:
```suggestion
                  "Save a dedicated snapshot to the disk cache")
```

dom96 (approved):

PR #1882, opened 2024-03-23
JSRPC: Ensure logging `RpcStub`s doesn't dispose them (skip getting `entries()` for RPC types)

mrbbot:
Previously, code like this would fail:

```js
const counter = new Counter();
const stub = new RpcStub(counter);
console.log(await stub.increment()); // 1
console.log(await stub.increment()); // 2
console.log(stub);
console.log(await stub.increment()); // Fails with "Error: RPC stub used after being disposed."
```

The `util.inspect()` function has special handling for JSG resource types. In particular, it tries to call the `entries()` function if it's defined for `Map`-like formatting. Because `RpcStub`s and other JSRPC types define wildcard property handlers, `entries()` existed, but attempting to call it and iterate through the resulting value ended up triggering this line:

https://github.com/cloudflare/workerd/blob/bfcfe9a7b2518a93d6c6d176be2f45d27e7c0913/src/workerd/api/worker-rpc.c%2B%2B#L836

This change ensures we never call `entries()` if a resource type defines a wildcard handler. It's unlikely this will return the expected value here.

The solution defines a well-known symbol on the prototype of types with wildcard handlers, similar to custom inspect properties. Other solutions considered were checking whether something like `"crazyUnguessablePropertyThatNoOneWouldEverAdd" + Math.random()` was defined, or adding a method to check if a value had a defined property handler.

### DIFF
--- a/src/node/internal/internal_inspect.ts
+++ b/src/node/internal/internal_inspect.ts
@@ -29,6 +29,7 @@
 /* todo: the following is adopted code, enabling linting one day */
 /* eslint-disable */
 
+import internalWorkers from "cloudflare-internal:workers";
 import internal from "node-internal:util";
 
 import { Buffer } from "node-internal:internal_buffer";
@@ -2414,10 +2415,23 @@ function isBuiltinPrototype(proto: unknown) {
   );
 }
 
+function isRpcWildcardType(value: unknown) {
+  return (
+    value instanceof internalWorkers.RpcStub ||
+    value instanceof internalWorkers.RpcPromise ||
+    value instanceof internalWorkers.RpcPromise
+  );
+}
+
 function isEntry(value: unknown): value is [unknown, unknown] {
   return Array.isArray(value) && value.length === 2;
 }
 function maybeGetEntries(value: Record<PropertyKey, unknown>): [unknown, unknown][] | undefined {
+  // If this value is an RPC type with a wildcard property handler (e.g. `RpcStub`), don't try to
+  // call `entries()` on it. This won't be an `entries()` function, and calling it with `.call()`
+  // would dispose the stub.
+  if (isRpcWildcardType(value)) return;
+
   const entriesFunction = value["entries"];
   if (typeof entriesFunction !== "function") return;
   const entriesIterator: unknown = entriesFunction.call(value);
--- a/src/node/internal/workers.d.ts
+++ b/src/node/internal/workers.d.ts
@@ -1,0 +1,9 @@
+declare namespace _default {
+  class WorkerEntrypoint {}
+  class DurableObject {}
+  class RpcPromise {}
+  class RpcProperty {}
+  class RpcStub {}
+  class RpcTarget {}
+}
+export default _default;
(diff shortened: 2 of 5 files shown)


src/node/internal/internal_inspect.ts
```diff
@@ -2414,10 +2414,19 @@ function isBuiltinPrototype(proto: unknown) {
   );
 }
 
+function hasWildcardPropertyHandler(value: Record<PropertyKey, unknown>) {
+  return value[internal.kResourceTypeWildcard] === true;
```
kentonv, line 2418: Wacky idea, but what if we checked for the presence of a weird name like `"extremely unlikely method name1^%@!*$&!@17512anhsttiuesoau,.bj"`... it should only be present if the type has a wildcard property (or if someone is intentionally trying to trick us I guess, but who case). This would also detect types that are "wildcards" by virtue of being pure-JS Proxies -- but it seems like in that case, too, we probably shouldn't assume that the property `"entries"` is what we think it is?

kentonv (commented):
Whoa. Some extreme duck tying going on here!

Out of curiosity, why is the value passed as a parameter to its own `entries()` method, rather than just relying on `this`?

kentonv (approved):

src/node/internal/internal_inspect.ts
mrbbot: > Wacky idea, but what if we checked for the presence of a weird name

This is something I considered. My main concern was that accessing a wildcard property might trigger side effects. Symbols aren't intercepted by `JSG_WILDCARD_PROPERTY`s so this seemed safer. I was also thinking it would be harder for someone to trick us, but this is JavaScript, so they could just do `Object.getOwnPropertySymbols()` or wrap the type with a `Proxy` before formatting and exfiltrate the `Symbol` from the access... 😅 

> we probably shouldn't assume that the property `"entries"` is what we think it is?

In any case, we're checking the type of `entries` matches what we expect it to be before we use it. Using the `unknown` type forces us to add runtime type validation before we can do anything with the value.

kentonv:
Ah, OK, that explains how the stub is getting disposed: when a stub is passed to an RPC, ownership is transferred away.

Hmm. I _suppose_ you could do `Function.prototype.call.call(a.b, a, c)`, if you want correctness even when `a.b.call` is overridden (not that this solves the problem since we still don't want to call `entries` on RPC objects).

But I wonder if we're asking for trouble by effectively overriding `call` here. It's kind of hard to avoid, though: `call` feels like too generic a name to disallow as an RPC method. We'd only really want to block it specifically on stubs that represent functions. But we don't actually _know_ whether any particular RPC value is a function or not.

On another note, I'm wondering now if detecting wildcard properties in general is too generic here. The existence of a wildcard property itself does not necessarily mean that it actually returns values for all possible property names -- the callback is allowed to return `kj::none` for "not present". In the case of a wildcard property callback that actually proxies a local object -- returning `kj::none` for properties that don't exist on the local object -- then calling `.entries()` may still be correct.

So now I think the correct answer here is either:
1. Detect RPC proxies specifically. (Perhaps we need a new base class for `RpcStub`, `RpcPromise`, and `RpcProperty` which can be checked using `instanceof`?)
2. Check for an unlikely name, as suggested above. This distinguishes between wildcard properties that actually return every possible name, vs. ones that only actually match a restricted list.

kentonv (approved):

PR #1883, opened 2024-03-23
Implement URL.parse

jasnell:
~~Should wait to land until after https://github.com/whatwg/url/pull/825 lands..~~ spec change landed!

Refs: https://github.com/whatwg/url/pull/825


jasnell:
Actually, this is going to need a bit more work. The method is not supposed to throw the way constructor does. Please do not merge until it can be updated

PR #1889, opened 2024-03-25
Additional logging when alarm throws during execution

jqmmes:
We were logging when an alarm broke its output gate after running. 

This PR will also log when output gate is broken during the alarm execution. 

This is disabling `retryAgainstLimits`. We want to be aware when, and why this happens.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -455,6 +455,13 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
       KJ_IF_SOME(status, context.getLimitEnforcer().getLimitsExceeded()) {
         outcome = status;
       }
+
+      // We want to alert if we aren't going to count this alarm retry against limits
+      if (auto desc = e.getDescription();
+          !jsg::isTunneledException(desc) && !jsg::isDoNotLogException(desc)
+          && context.isOutputGateBroken()) {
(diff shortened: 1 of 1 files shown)


src/workerd/api/global-scope.c++
```diff
@@ -455,6 +455,17 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
       KJ_IF_SOME(status, context.getLimitEnforcer().getLimitsExceeded()) {
         outcome = status;
       }
+
+      // We want to alert if we aren't going to count this alarm retry against limits
+      if (auto desc = e.getDescription();
+          !jsg::isTunneledException(desc) && !jsg::isDoNotLogException(desc)
+          && context.isOutputGateBroken()) {
+        if (isInterestingException(e)) {
+          LOG_EXCEPTION("alarmOutputLock"_kj, e);
```
justin-mp, line 464: Do you have a sense of how often this will happen?  Will we be spamming Sentry?

justin-mp (commented):
The commit message has typos (e.g., outpugate).  You probably should copy the PR's details into the commit message.

src/workerd/api/global-scope.c++
jqmmes: Hopefully it won't happen much. I can edit and make it always NO_SENTRY. We mostly want this for debugging purposes

justin-mp (approved):

PR #1894, opened 2024-03-25
Python memory snapshot: add magic number and version

hoodmane:


src/pyodide/internal/python.js
```diff
@@ -442,8 +449,13 @@ function encodeSnapshot(heap, dsoJSON) {
  */
 function decodeSnapshot(memorySnapshot) {
   const uint32View = new Uint32Array(memorySnapshot);
-  const snapshotOffset = uint32View[0];
-  const jsonLength = uint32View[1];
+  if (uint32View[0] !== SNAPSHOT_MAGIC) {
+    throw new Error("Memory snapshot doesn't start with snapshot magic number");
```
hoodmane, line 453: Do we want to have a fallback here to make the existing snapshots work?
@kflansburg @dom96

dom96: yeah, I'd say so

src/pyodide/internal/python.js
```diff
@@ -419,20 +419,27 @@ function setUploadFunction(toUpload) {
   };
 }
 
+// "\x00snp"
+const SNAPSHOT_MAGIC = 0x6d736100;
```
dom96, line 423: Maybe instead of using the wasm magic constant we should use something random?

```
$ capnp id
@0xa6a71b7b00626e55
```

You can truncate the above.

src/pyodide/internal/python.js
```diff
@@ -419,20 +419,27 @@ function setUploadFunction(toUpload) {
...
-  let sx = 8 + 2 * dsoString.length;
+  let snapshotOffset = 8 + 2 * dsoString.length;
   // align to 8 bytes
-  sx = Math.ceil(sx / 8) * 8;
-  const toUpload = new Uint8Array(sx + heap.length);
+  snapshotOffset = Math.ceil(snapshotOffset / 8) * 8;
+  const toUpload = new Uint8Array(snapshotOffset + heap.length);
   const encoder = new TextEncoder();
   const { written } = encoder.encodeInto(dsoString, toUpload.subarray(8));
+  const jsonLength = written;
```
dom96, line 437:
```suggestion
  const { jsonLength } = encoder.encodeInto(dsoString, toUpload.subarray(8));
```

dom96 (approved):

src/pyodide/internal/python.js
hoodmane: Well this change won't work, but you can do `written: jsonLength`.

src/pyodide/internal/python.js
hoodmane: This isn't the wasm magic number, that's `"\x00asm"` this is `"\x00snp"`. Things that are not valid webassembly should not start with the wasm magic number. I think there is an advantage to having the magic number be something one can actually read. But either way is probably fine.

dom96: Hm, are you sure? I see it here: https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/wasm/wasm-constants.h#22

hoodmane: Fixed. Thanks!

src/pyodide/internal/python.js
```diff
@@ -186,9 +186,13 @@ function loadDynlib(Module, path, wasmModuleData) {
  * there.
  */
 function preloadDynamicLibs(Module) {
+  let SO_FILES_TO_LOAD = SITE_PACKAGES_SO_FILES;
+  if (DSO_METADATA?.settings?.baselineSnapshot) {
+    SO_FILES_TO_LOAD = [[ '_lzma.so' ]];
```
dom96, line 191: is this definitely safe? Should we maybe not preload anything just to be extra cautious?

dom96 (approved):

src/pyodide/internal/python.js
hoodmane: This gets imported by something in `SNAPSHOT_IMPORTS` I'm not really sure why. But because we use it in the snapshot, we'll crash without it.

PR #1895, opened 2024-03-26
JSRPC: Fill in trace (tail) events better.

kentonv:
eventInfo was being left null, causing wrangler tail not to report the event at all. Let's fill it in and specify the top-level method name.

Also added entrypoint name, which seems highly relevant to JSRPC.

### DIFF
--- a/src/workerd/api/trace.h
+++ b/src/workerd/api/trace.h
@@ -67,6 +67,7 @@ struct ScriptVersion {
 class TraceItem final: public jsg::Object {
 public:
   class FetchEventInfo;
+  class JsRpcEventInfo;
   class ScheduledEventInfo;
   class AlarmEventInfo;
   class QueueEventInfo;
@@ -78,6 +79,7 @@ public:
   explicit TraceItem(jsg::Lock& js, const Trace& trace);
 
   typedef kj::OneOf<jsg::Ref<FetchEventInfo>,
+                    jsg::Ref<JsRpcEventInfo>,
                     jsg::Ref<ScheduledEventInfo>,
                     jsg::Ref<AlarmEventInfo>,
                     jsg::Ref<QueueEventInfo>,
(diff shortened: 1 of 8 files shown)


src/workerd/api/worker-rpc.c++
```diff
@@ -1167,13 +1173,13 @@ private:
...
     kj::Maybe<jsg::JsValue> result;
-    kj::StringPtr methodNameForErrors;
+    kj::ConstString methodNameForTrace;
 
     switch (callParams.which()) {
       case rpc::JsRpcTarget::CallParams::METHOD_NAME: {
         kj::StringPtr methodName = callParams.getMethodName();
         result = getProperty(methodName);
-        methodNameForErrors = methodName;
+        methodNameForTrace = methodName.attach();
```
jclee, line 1182: If I'm understanding correctly, this `StringPtr::attach()` call produces a `ConstString` that does not own its own memory, but instead relies on the caller to keep `callParams` alive.

...which I assume is OK, since the only caller of `tryGetProperty()` uses the returned `methodNameForTrace` value synchronously in the same scope as the passed `params` value.

src/workerd/io/trace.h
```diff
@@ -251,8 +261,9 @@ class Trace final : public kj::Refcounted {
   // We treat the origin value as "unset".
   kj::Date eventTimestamp = kj::UNIX_EPOCH;
 
-  typedef kj::OneOf<FetchEventInfo, ScheduledEventInfo, AlarmEventInfo, QueueEventInfo,
-          EmailEventInfo, TraceEventInfo, HibernatableWebSocketEventInfo, CustomEventInfo> EventInfo;
+  typedef kj::OneOf<FetchEventInfo, JsRpcEventInfo, ScheduledEventInfo, AlarmEventInfo,
+          QueueEventInfo, EmailEventInfo, TraceEventInfo, HibernatableWebSocketEventInfo,
+          CustomEventInfo> EventInfo;
   kj::Maybe<EventInfo> eventInfo;
   // TODO(someday): Support more event types.
```
jclee, line 269: Heh... Guess we can remove this comment, since we now support lots of event types.

jclee (approved):

src/workerd/api/trace.h
```diff
@@ -240,6 +242,29 @@ class TraceItem::FetchEventInfo::Response final: public jsg::Object {
   uint16_t status;
 };
 
+class TraceItem::JsRpcEventInfo final: public jsg::Object {
+public:
+  explicit JsRpcEventInfo(const Trace& trace, const Trace::JsRpcEventInfo& eventInfo);
+
+  // We call this `rpcMethod` to make clear this is an RPC event, since some tail workers rely
+  // on duck-typing EventInfo based on the properties present. (`methodName` might be ambiguous
+  // since HTTP also has methods.)
```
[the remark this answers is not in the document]
kentonv: Well, I had written all the code before I realized that the tail worker distinguishes event types based on what properties are present. At that point I decided only to change the JS API. The other places could be renamed in the future without breaking any kind of compatibility, but I think they are fine as-is.

src/workerd/api/worker-rpc.c++
kentonv: Right.

PR #1900, opened 2024-03-26
Add platform-specific release configuration

fhanau:
- Match macOS wd_benchmark link options with wd_binary options
- Correct spelling of WIN32_LEAN_AND_MEAN
- /await and /GF are ignored by clang-cl, no need to pass them in the
    Windows build.


jasnell (approved):

fhanau:
Thank you for reviewing James!
CC @penalosa @mrbbot @jp4a50 this has nothing that should cause breakage per se, but there are release configuration changes that did not go through CI before. If problems pop up with releases, please shout!

PR #1901, opened 2024-03-26
JSRPC: Fix waitUntil() and response streams

kentonv:
This fixes two bugs @jasnell discovered (but he's OOO this week so can't review).


MellowYarker (approved):
Seems reasonable, though we should wait for internal CI before merging.

PR #1902, opened 2024-03-26
Give a friendly message when the user specifies an unsupported package in local dev

garrettgu10:
Remake of #1830 since too much stuff changed since then

### DIFF
--- a/src/pyodide/internal/setupPackages.js
+++ b/src/pyodide/internal/setupPackages.js
@@ -173,7 +173,9 @@ function addPackageToLoad(lockfile, name, toLoad) {
   }
   const pkgInfo = lockfile.packages[normalizedName];
   if (!pkgInfo) {
-    throw new Error(`No known package with name '${name}'`);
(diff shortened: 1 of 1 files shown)


src/pyodide/internal/setupPackages.js
```diff
@@ -174,7 +174,9 @@ function addPackageToLoad(lockfile, name, toLoad) {
   }
   const pkgInfo = lockfile.packages[normalizedName];
   if (!pkgInfo) {
-    throw new Error(`No known package with name '${name}'`);
+    throw new Error("It appears that a package (" + normalizedName + ") you requested is not available yet on workerd. \n" +
```
hoodmane, line 177: Maybe use `name` instead of `normalizedName` in the message:
```suggestion
    throw new Error(`It appears that a package ("${name}") you requested is not available yet in workerd. \n` +
```

hoodmane (approved):

PR #1904, opened 2024-03-27
Add 2024-03-26 entry point to workers-types

IgorMinar:


IgorMinar:

IgorMinar:
recheck

fhanau (approved):

kentonv:
Note that RPC is scheduled for `2024-04-03`. Will we be able to add another entrypoint next week?

IgorMinar:
I see, I suggest that we then wait and release the 2024-04-03 entrypoint instead. (There is a cost to having too many entrypoints as it bloats the npm package.)

kentonv:
Oh good point... Yeah, the "rpc" compat flag doesn't actually impact the types at all.

PR #1906, opened 2024-03-27
Implements limiter for startup of Python user code.

dom96:
Right now we have a limiter around the full Python Worker startup (which includes the time it takes to start up Pyodide). This PR implements a simple limiter to have a separate limit for the Python user code specifically.


hoodmane (approved):
Thanks!

PR #1910, opened 2024-03-27
Add exception stack traces to trace/tail events.

kentonv:
A new property, `stack`, contains the stack trace.

To support this in the case of async exceptions (which have already been converted to KJ), exception tunneling now adds a serialized copy of the V8 error to the exception itself, as a "detail". For now, this is _only_ used by `logUncaughtException()`. In the future, though, when the exception propagates back into JavaScript, we could de-tunnel it by deserializing this detail, rather than based on the description string. However, there are several complexities that need to be solved first, so I have not tried to do that in this PR.

Depends on: https://github.com/capnproto/capnproto/pull/1978

### DIFF
--- a/src/workerd/io/trace.h
+++ b/src/workerd/io/trace.h
@@ -240,7 +240,8 @@ public:
 
   class Exception {
   public:
-    explicit Exception(kj::Date timestamp, kj::String name, kj::String message);
+    explicit Exception(kj::Date timestamp, kj::String name, kj::String message,
+                       kj::Maybe<kj::String> stack);
     Exception(rpc::Trace::Exception::Reader reader);
     Exception(Exception&&) = default;
     KJ_DISALLOW_COPY(Exception);
@@ -250,8 +251,9 @@ public:
     kj::Date timestamp;
 
     kj::String name;
-    // TODO(someday): record exception source, line/column number, stack trace?
     kj::String message;
+
+    kj::Maybe<kj::String> stack;
 
     void copyTo(rpc::Trace::Exception::Builder builder);
   };
@@ -366,7 +368,8 @@ public:
   // TODO(soon): Eventually:
   //void setMetrics(...) // Or get from MetricsCollector::Request directly?
 
-  void addException(kj::Date timestamp, kj::String name, kj::String message);
+  void addException(kj::Date timestamp, kj::String name, kj::String message,
+                    kj::Maybe<kj::String> stack);
 
   void addDiagnosticChannelEvent(kj::Date timestamp, kj::String channel,
                                  kj::Array<kj::byte> message);
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -148,8 +148,12 @@ void addExceptionToTrace(jsg::Lock& js,
   }
 
   auto timestamp = ioContext.now();
-  auto error = KJ_REQUIRE_NONNULL(errorTypeHandler.tryUnwrap(js, exception),
-      "Should always be possible to unwrap error interface from an object.");
+  Worker::Api::ErrorInterface error;
+
+  if (exception.isObject()) {
+    error = KJ_REQUIRE_NONNULL(errorTypeHandler.tryUnwrap(js, exception),
+        "Should always be possible to unwrap error interface from an object.");
+  }
 
   kj::String name;
   KJ_IF_SOME(n, error.name) {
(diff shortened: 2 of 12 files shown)


src/workerd/io/trace.h
```diff
@@ -242,6 +243,8 @@ class Trace final : public kj::Refcounted {
     // TODO(someday): record exception source, line/column number, stack trace?
```
jclee, line 243: I guess we could remove this TODO.

src/workerd/jsg/util.c++
```diff
@@ -341,6 +351,16 @@ void throwInternalError(v8::Isolate* isolate, kj::Exception&& exception) {
   }
 }
 
+void addExceptionDetail(Lock& js, kj::Exception& exception, v8::Local<v8::Value> handle) {
+  js.tryCatch([&]() {
+    Serializer ser(js);
```
jclee, line 356: I assume we don't need to worry about the Serializer wire format version.

src/workerd/jsg/util.c++
```diff
@@ -341,6 +351,16 @@ void throwInternalError(v8::Isolate* isolate, kj::Exception&& exception) {
...
+    ser.write(js, JsValue(handle));
```
jclee, line 357: Should we be doing anything to bound the amount of data added to an exception?

src/workerd/io/worker.c++
```diff
@@ -148,8 +148,12 @@ void addExceptionToTrace(jsg::Lock& js,
   }
 
   auto timestamp = ioContext.now();
-  auto error = KJ_REQUIRE_NONNULL(errorTypeHandler.tryUnwrap(js, exception),
-      "Should always be possible to unwrap error interface from an object.");
+  Worker::Api::ErrorInterface error;
+
+  if (exception.isObject()) {
```
jclee, line 153: I guess the `isObject()` check is needed here because the deserialized result isn't necessarily an Object?  (But would typically be an Object?)

src/workerd/jsg/util.c++
kentonv: At present, this isn't actually sent over the wire regardless. But I should probably specify a version explicitly, as we do elsewhere, so that V8 updates don't create temporary backward-incompatibility.

src/workerd/io/worker.c++
kentonv: Annoyingly, JavaScript permits you to throw any value, not just objects. I think there was a bug here that caused us to fail to log non-object exceptions, but it wasn't noticed before because our exception tunneling would turn non-objects into objects. (Exception tunneling today always produces `Error` values even if the original wasn't.)

kentonv:
[Changes to address review comments.](https://github.com/cloudflare/workerd/compare/a015f0cafe6e3b4a8e04c9dfa03b3554abd15efb..e57f5bb8e5d53b4bffd2991fbf66e9d639d4d09e)

src/workerd/io/worker.c++
```diff
@@ -1839,6 +1883,42 @@ void Worker::Lock::logUncaughtException(UncaughtExceptionSource source,
...
+        jsg::Deserializer deser(js, serializedJsError);
+        deserialized = deser.readValue(js);
+      } catch (jsg::JsExceptionThrown&) {
+        // Failed to deserialize, we'll continue with exceptionToJsValue() instead.
+        //
+        // Note that we're intentionally not checking tryCatch.CanContinue() here, because we still
+        // want to log the exception even if the isolate has been terminated.
+      }
+
+      KJ_IF_SOME(d, deserialized) {
```
jclee, line 1910: Looks like `exceptionToJsValue()` does some additional property population, filtering, and logging of internal tunneled exceptions that would be skipped when deserialization is successful.  I guess the assumption here is that if a valid serialized value is present, it didn't come from an internal tunneled exception?  And that the other populated properties are not currently consumed by `logUncaughtException()`?

src/workerd/io/worker.c++
jclee: Ah, OK... that would explain why the current test cases for `throw "string"` were working without the Object check.

jclee (approved):

src/workerd/io/worker.c++
kentonv: Yeah, `logUncaughtException()` doesn't inspect any of those extra properties today. Only type, message, and now stack.

src/workerd/io/worker.c++
```diff
@@ -160,9 +164,49 @@ void addExceptionToTrace(jsg::Lock& js,
...
+        // want to continue even if the isolate has been terminated.
+      }
+    }
+  }
+
+  kj::Maybe<kj::String> stack;
+  KJ_IF_SOME(s, error.stack) {
+    kj::StringPtr slice = s;
+
+    // Normally `error.stack` repeats the error type and message first. We don't want send two
```
irvinebroque, line 187: @kentonv — see internal thread on this, I think by stripping out here to be efficient in what gets sent across, we ended up making it such that the `stack` property in the Tail Worker ends up not including full stack trace?

kentonv: I don't know which internal thread you mean, and I don't see how the code here could possibly be stripping any part of the actual stack trace.

PR #1912, opened 2024-03-27
Minor Python cleanup

hoodmane:
Nonfunctional change. Rename reportError.js into util.js and move simpleRunPython there. Also run formatter.


fhanau (approved):

hoodmane:
Thanks for the review @fhanau!

PR #1914, opened 2024-03-28
Quick ActorCache refactor

MellowYarker:
These commits were scattered before and after the larger read coalesce changes, and since that change is really big and complicated, I figured it would be easier to yank these quick refactors out into a separate PR so we can focus on the scary stuff elsewhere.

### DIFF
--- a/src/workerd/io/actor-cache.c++
+++ b/src/workerd/io/actor-cache.c++
@@ -277,19 +277,19 @@ bool ActorCache::SharedLru::evictIfNeeded(Lock& lock) const {
   }
 }
 
-void ActorCache::touchEntry(Lock& lock, Entry& entry, const ReadOptions& options) {
-  if (!options.noCache) {
-    if (!entry.isDirty()) {
-      entry.isStale = false;
-      lock->remove(entry);
-      lock->add(entry);
-    }
-
-    // If this is a dirty entry previously marked no-cache, remove that mark. This results in the
-    // same end state as if the entry had been flushed and evicted before the read -- it would have
-    // been read back, and then into cache.
-    entry.noCache = false;
+void ActorCache::touchEntry(Lock& lock, Entry& entry) {
+  if (!entry.isDirty()) {
+    entry.isStale = false;
+    lock->remove(entry);
+    lock->add(entry);
   }
+
+  // We only call `touchEntry` when the operation or the LRU has !noCache, so we want to cache this.
+  //
+  // If this is a dirty entry previously marked no-cache, remove that mark. This results in the
+  // same end state as if the entry had been flushed and evicted before the read -- it would have
+  // been read back, and then into cache.
+  entry.noCache = false;
 }
 
 void ActorCache::removeEntry(Lock& lock, Entry& entry) {
@@ -954,7 +954,9 @@ kj::OneOf<ActorCache::GetResultList, kj::Promise<ActorCache::GetResultList>>
          positiveCount < limit.orDefault(kj::maxValue); ++iter) {
     Entry& entry = **iter;
 
-    touchEntry(lock, entry, options);
+    if (!options.noCache) {
+      touchEntry(lock, entry);
+    }
 
     switch(entry.valueStatus) {
       case EntryValueStatus::ABSENT: {
@@ -990,7 +992,9 @@ kj::OneOf<ActorCache::GetResultList, kj::Promise<ActorCache::GetResultList>>
   if (iter != ordered.end() && iter->get()->key == endKey) {
     // We have an entry exactly at our end, it might even be a previously inserted UNKNOWN. Let's
     // touch it for freshness.
-    touchEntry(lock, **iter, options);
+    if (!options.noCache) {
+      touchEntry(lock, **iter);
+    }
   }
 
   if (storageListStart == kj::none || knownPrefixSize >= limit.orDefault(kj::maxValue)) {
@@ -1256,7 +1260,9 @@ kj::OneOf<ActorCache::GetResultList, kj::Promise<ActorCache::GetResultList>>
   if (iter != map.ordered().end() && iter->get()->key == endKey) {
     // We have an entry exactly at our end, it might even be a previously inserted UNKNOWN. Let's
     // touch it for freshness.
-    touchEntry(lock, **iter, options);
+    if (!options.noCache) {
+      touchEntry(lock, **iter);
+    }
   }
   for (; positiveCount < limit.orDefault(kj::maxValue); ) {
     if (iter == ordered.begin()) {
@@ -1281,7 +1287,9 @@ kj::OneOf<ActorCache::GetResultList, kj::Promise<ActorCache::GetResultList>>
       break;
     }
 
-    touchEntry(lock, entry, options);
+    if (!options.noCache) {
+      touchEntry(lock, entry);
+    }
 
     // Note that we need to add even negative entries to `cachedEntries` so that they override
     // whatever we read from storage later. However, they should not count against the limit.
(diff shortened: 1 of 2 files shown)


src/workerd/io/actor-cache.h
```diff
@@ -407,11 +407,21 @@ class ActorCache final: public ActorCacheInterface {
...
     EntryValueStatus valueStatus;
+
+    // This enum indicates how synchronized this entry is with storage.
+    EntrySyncStatus syncStatus = EntrySyncStatus::NOT_IN_CACHE;
   public:
     auto getValueStatus() const {
       return valueStatus;
     }
 
+    // TODO(now): We have `isDirty()`, which tells me we might also want methods for other states,
```
a-robinson, line 418: Only if we want to keep separate DIRTY and FLUSHING states. If they're going to merge, I don't see much point in having `isFoo` getters for each possible status.

src/workerd/io/actor-cache.h
```diff
@@ -407,11 +407,21 @@ class ActorCache final: public ActorCacheInterface {
...
+    // but the KJ_FAIL_ASSERT is throwing me off a bit. Either we should drop isDirty() and use this
```
a-robinson, line 419: The `KJ_FAIL_ASSERT` in `isDirty` is awkwardly encoding knowledge about its caller(s). It's nice to have that assertion there since it does narrow the space of possible states we need to worry about, but it does entangle this type with the rest of `ActorCache` a bit.

a-robinson (approved):
This set of changes LGTM modulo the TODO(now) in the `Make Entry::syncStatus private` commit. If you want to get it in ahead of the other PRs, feel free.

fhanau (approved):
LGTM

src/workerd/io/actor-cache.c++
```diff
@@ -106,7 +106,7 @@ ActorCache::Entry::~Entry() noexcept(false) {
   }
 }
 
-ActorCache::SharedLru::SharedLru(Options options): options(options) {}
+ActorCache::SharedLru::SharedLru(const Options options): options(options) {}
```
justin-mp, line 109: While it can make sense to have a const value parameter in a function definition, I don't think it makes sense to do so here.  Two reasons:
1. The fact that you've declared `this->options` as const means that you can only use `this->options` as const.  The `const` in the parameter declaration doesn't do anything for you here.
2. The `const` in the parameter declaration prevents you from moving the value, e.g., `options(kj::mv(options))`.

src/workerd/io/actor-cache.h
```diff
@@ -805,40 +805,40 @@ class ActorCacheOps::GetResultList {
 // forward-declared elsewhere.
 struct ActorCacheSharedLruOptions {
   // Memory usage that the LRU will try to stay under by evicting clean values.
-  size_t softLimit;
+  const size_t softLimit;
```
justin-mp, line 845: I don't think these const declarations buy you anything.  Two reasons:
1. You can only set the values on construction, which seems like an unreasonable constraint on users of the struct.
2. It makes this struct non-moveable.

src/workerd/io/limit-enforcer.h
```diff
@@ -25,7 +25,7 @@ class IsolateLimitEnforcer {
   // Further customize the isolate immediately after startup.
   virtual void customizeIsolate(v8::Isolate* isolate) = 0;
 
-  virtual ActorCacheSharedLruOptions getActorCacheLruOptions() = 0;
+  virtual const ActorCacheSharedLruOptions getActorCacheLruOptions() = 0;
```
justin-mp, line 28: This doesn't make sense because it restricts what users of the return value can do with the struct.  For example, users cannot `kj::mv` them.

src/workerd/io/actor-cache.h
```diff
@@ -847,7 +847,7 @@ class ActorCache::SharedLru {
   size_t currentSize() const { return size.load(std::memory_order_relaxed); }
 
 private:
-  Options options;
+  const Options options;
```
justin-mp, line 887: This one makes sense.  Let's keep this.  :-)

src/workerd/io/actor-cache.c++
justin-mp, line 1: FYI, note that KJ uses const to mean thread-safe too.  See https://github.com/capnproto/capnproto/blob/v2/style-guide.md#constness

src/workerd/io/actor-cache.h
```diff
@@ -406,8 +406,12 @@ class ActorCache final: public ActorCacheInterface {
     // from cache and needs to remember the original cached values even if they are overwritten
     // before the read completes.
     const Value value;
-  public:
     EntryValueStatus valueStatus;
+  public:
+    auto getValueStatus() const {
```
justin-mp, line 414: nit: please put the return type here instead of `auto`.  It saves the reader trying to figure out what the return type is.

src/workerd/io/actor-cache.h
```diff
@@ -630,6 +630,20 @@ class ActorCache final: public ActorCacheInterface {
   // Type of a lock on `SharedLru::cleanList`. We use the same lock to protect `currentValues`.
   typedef kj::Locked<kj::List<Entry, &Entry::link>> Lock;
 
+  // Add this entry to the clean list and set its status to CLEAN.
+  // This doesn't do much, but it makes it easier to track what's going on.
+  void addToCleanList(Lock& listLock, Entry& entryRef) {
```
justin-mp, line 654: I love these kinds of refactors!  Making something easier to track is awesome!

src/workerd/io/actor-cache.h
```diff
@@ -427,11 +434,23 @@ class ActorCache final: public ActorCacheInterface {
...
-    EntrySyncStatus syncStatus = EntrySyncStatus::NOT_IN_CACHE;
+    void setFlushing() {
+      syncStatus = EntrySyncStatus::FLUSHING;
+    }
+
+    void setNotInCache() {
+      syncStatus = EntrySyncStatus::NOT_IN_CACHE;
+    }
+
+    // Avoid using this directly. If you want to set the status to CLEAN or DIRTY, consider using
```
justin-mp, line 445: If we have this function, but we're not supposed to use it, that's an indication of something buggy in our abstraction layers.

A better approach might be to remove `setSyncStatus` and have setClean and setDirty instead.

justin-mp (changes requested):

src/workerd/io/actor-cache.c++
MellowYarker: https://github.com/cloudflare/workerd/pull/1914/commits/a8da46b64c7cd1c429527277c1f1236beda426ed

Yeah, this makes sense. Thanks!

src/workerd/io/actor-cache.h
MellowYarker: I think I overshot a bit here :sweat_smile:, I wanted to see more `const` in the other parts of actor cache to make it easier to reason about, but that was tricky, so all that energy went into this struct https://github.com/cloudflare/workerd/pull/1914/commits/bdbae9bfac75544a05c4945c94ac33b5bd8640b1

src/workerd/io/actor-cache.h
MellowYarker: https://github.com/cloudflare/workerd/pull/1914/commits/fcef6f68058edfed70c7d00ebcb1efc1ae55752c

src/workerd/io/limit-enforcer.h
MellowYarker: Don't think they're needed here either: https://github.com/cloudflare/workerd/pull/1914/commits/f8b58b5fa86d655035ca4b271362b9bcdb82c149

src/workerd/io/actor-cache.h
MellowYarker: What we really want is to prevent directly modifying the `syncStatus` unless it's part of some action (like adding to a list), but the methods that add to the respective lists don't have direct access to the `syncStatus`. The same issue would exist with `setClean` and `setDirty` -- though I suppose it would make it more obvious that you'd have to use `setNotInCache()` explicitly too...

a-robinson (approved):
Still LGTM, but it looks like merge on this will be blocked until @justin-mp removes his "Request changes" bit.

justin-mp (approved):

jasnell (approved):

PR #1915, opened 2024-03-28
Refactor: ActorCache flush transaction and CountedDeletes

MellowYarker:
This is some additional refactoring work that the read coalesce change relies on.

I specifically wanted to separate this so we can think about whether we want to retain the `FLUSHING` state or continue forward with `flushStarted`. It was too difficult to untangle that change when the smaller commits were all over the large single PR, but now that they're all [in front of us](https://github.com/cloudflare/workerd/pull/1914) we should be able to make changes with less effort.

### DIFF
--- a/src/workerd/io/actor-cache-test.c++
+++ b/src/workerd/io/actor-cache-test.c++
@@ -750,7 +750,7 @@ KJ_TEST("ActorCache deleteAll()") {
   {
     auto mockTxn = mockStorage->expectCall("txn", ws).returnMock("transaction");
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["bar", "baz", "grault"]))
+        .withParams(CAPNP(keys = ["bar", "grault", "baz"]))
         .thenReturn(CAPNP(numDeleted = 2));
     mockTxn->expectCall("delete", ws)
         .withParams(CAPNP(keys = ["garply"]))
@@ -862,7 +862,7 @@ KJ_TEST("ActorCache deleteAll() again when previous one isn't done yet") {
   {
     auto mockTxn = mockStorage->expectCall("txn", ws).returnMock("transaction");
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["bar", "baz", "grault"]))
+        .withParams(CAPNP(keys = ["bar", "grault", "baz"]))
         .thenReturn(CAPNP(numDeleted = 2));
     mockTxn->expectCall("delete", ws)
         .withParams(CAPNP(keys = ["garply"]))
--- a/src/workerd/io/actor-cache-test.c++
+++ b/src/workerd/io/actor-cache-test.c++
@@ -750,7 +750,7 @@ KJ_TEST("ActorCache deleteAll()") {
   {
     auto mockTxn = mockStorage->expectCall("txn", ws).returnMock("transaction");
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["bar", "grault", "baz"]))
+        .withParams(CAPNP(keys = ["bar", "baz", "grault"]))
         .thenReturn(CAPNP(numDeleted = 2));
     mockTxn->expectCall("delete", ws)
         .withParams(CAPNP(keys = ["garply"]))
@@ -766,8 +766,6 @@ KJ_TEST("ActorCache deleteAll()") {
     mockTxn->expectDropped(ws);
   }
 
-  KJ_ASSERT(deletePromise.wait(ws) == 2);
-
   mockStorage->expectCall("deleteAll", ws)
       .thenReturn(CAPNP(numDeleted = 2));
 
@@ -781,6 +779,8 @@ KJ_TEST("ActorCache deleteAll()") {
                                      (key = "waldo", value = "99999")]))
         .thenReturn(CAPNP());
   }
+
+  KJ_ASSERT(deletePromise.wait(ws) == 2);
 
   KJ_ASSERT(expectCached(test.get("foo")) == nullptr);
   KJ_ASSERT(expectCached(test.get("baz")) == nullptr);
@@ -862,7 +862,7 @@ KJ_TEST("ActorCache deleteAll() again when previous one isn't done yet") {
   {
     auto mockTxn = mockStorage->expectCall("txn", ws).returnMock("transaction");
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["bar", "grault", "baz"]))
+        .withParams(CAPNP(keys = ["bar", "baz", "grault"]))
         .thenReturn(CAPNP(numDeleted = 2));
     mockTxn->expectCall("delete", ws)
         .withParams(CAPNP(keys = ["garply"]))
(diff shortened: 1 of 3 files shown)


src/workerd/io/actor-cache.c++
```diff
@@ -2086,16 +2083,6 @@ void ActorCache::putImpl(Lock& lock, kj::Own<Entry> newEntry,
     // Inherit gap state.
     newEntry->gapIsKnownEmpty = slot->gapIsKnownEmpty;
 
-    if (slot->isDirty()) {
```
MellowYarker, line 2066: The reason we can get rid of this (i.e. not inherit the counted delete) is because the `CountedDelete` actually takes a strong reference to the Entry.

In the case the `CountedDelete` _doesn't_ take a strong reference (i.e. Entry is `PRESENT`), we count the delete early, then the new `ABSENT` entry ends up in the dirtyList and we do a regular delete. If we subsequently overwrite with a `put()` before the entry is flushed, then we don't need to bother doing the delete and we replace the `ABSENT` entry (removing from the dirtyList) with the new `PRESENT` one. Since we already contributed to the count, it's effectively the same as if we had done the delete followed by the put.

MellowYarker: @a-robinson this explains why https://github.com/cloudflare/workerd/pull/1814#discussion_r1540158815 shouldn't be much of a concern.

src/workerd/io/actor-cache.c++
```diff
@@ -2325,56 +2321,78 @@ kj::Promise<void> ActorCache::startFlushTransaction() {
...
+
+    auto& countedDeleteFlush = countedDeleteFlushes.add(CountedDeleteFlush{
+      .countedDelete = kj::addRef(*countedDelete),
+    });
+    // We might have successfully deleted these entries, but had the broader transaction fail.
+    // In that case, we might have entries that have since been overwritten, and which no longer
+    // need to be scheduled for deletion.
+    kj::Vector<kj::Own<Entry>> entriesToDelete(countedDelete->entries.size());
+    for (auto& entry : countedDelete->entries) {
+      if (entry->overwritingCountedDelete && countedDelete->completedInTransaction) {
```
a-robinson, line 2341: Given this check, is it possible for a `CountedDeleteFlush` to end up with zero entries in `entriesToDelete`? If so, will anything bad happen when we later go to use it?

Might be better to delay the `countedDeleteFlushes.add()` call from above until after we've confirmed `entriesToDelete` is non-empty?

Also, if we do that, a test case covering it would be nice, assuming it doesn't require anything terribly convoluted.

src/workerd/io/actor-cache.c++
```diff
@@ -2325,56 +2321,78 @@ kj::Promise<void> ActorCache::startFlushTransaction() {
...
+      // Note that a subsequent put() call could have set entry.value to non-null, but we still
+      // have to perform the delete first in order to determine the count that the delete() call
+      // should return.
+      //
+      // There is a minor quirk here because the counted delete set does not distinguish between
+      // before and after a delete all. That's actually okay because we should be able to
+      // immediately resolve counted deletes requested after a delete all (either the values are
+      // absent or they have a dirty put). This might also be an issue if we respected noCache for
+      // delete all's dummy value, but we do not.
+      entry->flushStarted = true;
```
a-robinson, line 2369: Why this over `entry->setFlushing()`?

a-robinson (approved):
Whew that last commit is a lot to process, but I think it looks good and it does appear to be a net improvement. Thanks for your persistence on this, @MellowYarker !

src/workerd/io/actor-cache.c++
```diff
@@ -2313,17 +2321,37 @@ kj::Promise<void> ActorCache::flushImpl(uint retryCount) {
...
+        // Since we already have the delete count, we don't need to delete this entry again.
+        continue;
+      }
+      entriesToDelete.add(kj::mv(entry));
+    }
 
-    KJ_IF_SOME(c, entry.countedDelete) {
-      if (c->resultFulfiller->isWaiting()) {
+    // We will only include this CountedDelete if there are entries that need to be deleted.
+    if (entriesToDelete.size() > 0) {
```
a-robinson, line 2347: Minor nit, but it's generally easier to read if you return/continue early rather than indenting the main code path further. Which would in this case mean:

```
if (entriesToDelete.size() == 0) {
  continue;
}

// previously indented code goes here
```

src/workerd/io/actor-cache-test.c++
```diff
@@ -1188,11 +1185,14 @@ KJ_TEST("ActorCache flush retry") {
...
+        .withParams(CAPNP(keys = ["quux"]))
+        .thenReturn(CAPNP());  // count ignored because we got it on the first try!
+    mockTxn->expectCall("delete", ws)
+        .withParams(CAPNP(keys = ["corge", "grault"]))
         .thenReturn(CAPNP(numDeleted = 2));
     mockTxn->expectCall("delete", ws)
-        .withParams(CAPNP(keys = ["quux", "baz"]))
-        .thenReturn(CAPNP(numDeleted = 1234));  // count ignored
+        .withParams(CAPNP(keys = ["baz"]))
+        .thenReturn(CAPNP(numDeleted = 1234));
```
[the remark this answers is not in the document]
MellowYarker: I'm confused by the number regardless of the comment, what is the significance of numDeleted = 1234? I see baz is counted immediately so this would do a regular delete, but in that case why not exclude numDeleted entirely? Does including numDeleted with a clearly unexpected value imply the value is not checked better than excluding it completely?

src/workerd/io/actor-cache-test.c++
```diff
@@ -1172,10 +1170,9 @@ KJ_TEST("ActorCache flush retry") {
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("corge"))) == "555");
   KJ_ASSERT(expectCached(test.get("grault")) == nullptr);
 
-  // Although the transaction didn't complete, the delete did, and so it resolves.
-  KJ_ASSERT(promise1.wait(ws) == 1);
-
-  // The second delete had failed, though, so is still outstanding.
+  // Although the counted delete succeeded, the promise will not resolve until our flush succeeds!
+  KJ_ASSERT(!promise1.poll(ws));
+  // The second delete had failed, though; it is still outstanding.
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/compare/047bf09588153127db3c029914a55e182c426390..f284b1704defed396b1eb8d17e5ff0928b0a9518

src/workerd/io/actor-cache-test.c++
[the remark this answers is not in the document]
a-robinson: > Does including numDeleted with a clearly unexpected value imply the value is not checked better than excluding it completely?

¯\_(ツ)_/¯ that's a fair question. In cases like this where we've already verified that the `delete()` call returned true, it's serving no purpose. In a case where perhaps the delete call didn't return immediately (e.g. because multiple keys were provided) there's a bit more of an argument for returning an unexpected number rather than not returning a count, since not returning a count is equivalent to returning 0 and we might fail to notice if a count of 0 is being used incorrectly.

But in either case, I think your understanding here is correct.

PR #1916, opened 2024-03-28
ActorCache Read Coalesce

MellowYarker:


src/workerd/io/actor-cache-test.c++
```diff
@@ -1444,76 +1424,131 @@ KJ_TEST("ActorCache read retry on flush containing only puts") {
   KJ_ASSERT(KJ_ASSERT_NONNULL(promise.wait(ws)) == "123");
 }
 
-KJ_TEST("ActorCache read hard fail") {
+// KJ_TEST("ActorCache read hard fail") {
```
MellowYarker, line 1427: This test is currently commented out because:
1. When we `kj::mv(mockGet).thenThrow(KJ_EXCEPTION(FAILED, "read failed"));` the read promise, we cause the `syncGet()` storage operation to throw.
2. This results in `startFlushTransaction()` to throw, which causes the `flushProm` in `flushImpl()` to throw.
3. We end up in the `catch_()` handler of the `flushProm`, and since our exception cannot be retried, we go into the last `else` and throw another exception and break the output gate.

Prior to the coalesced read implementation, we would expect the failure of a single storage get() to propagate the exception to the JS caller. This was easy because storage gets() were completely independent operations from writes. Now that `get()` goes through the same flush as writes, we've got a few more things to worry about.

---
### Propagating the Exception to the caller
We'll focus on the single key `get()` for now. [Currently](https://github.com/cloudflare/workerd/pull/1916/commits/516c7ce7fe28e264f494f9c9c9b62938d3547bb6#diff-52b552cc9d6477516b49e970aa23de236c7a1f7327b09c243c336445cbb2e18eR439-R444), the JS get caller must wait for the flush to complete, and upon success, the caller will receive the result of the `get()`.

This breaks when we force a storage `get()` to throw, because instead of succeeding, the flush will fail with `broken.outputGateBroken; jsg.Error: Internal error in Durable Object storage write caused object to be reset`. 

I believe we can fix this by waiting on a joint promise in `ActorCache::get()`. We can keep the successful flush case (linked above), and add another promise for when the storage get (in `syncGet()`) fails. To make things explicit, we can wrap much of `syncGet()` in a `try...catch`, and upon entering the catch branch, propagate the exception to the waiting `ActorCache::get()`. This would effectively ensure that the JS caller receives the exception. 

I've already tested this and confirmed it works, and it doesn't look _too_ daunting. In short, the constructor for `GetWaiter` takes all the entries we care about for this `get()`. For each `Entry`, we create a PaF, give the fulfiller to the `Entry`, and add the promise to a single joint "_the read failed_" promise owned by the `GetWaiter`.

- At this point, the `GetWaiter` has a promise that resolves when any of the associated `Entry`s use their fulfiller, indicating the entire get has failed.

Back in the `try...catch` for `syncGet()`, if we're in the catch branch, we can just use the fulfiller on the `Entry` to signal that the storage `get()` failed.

This approach should extend to the multi-key get() fairly easily, since we can just iterate over all the `Entry`s in a `GetFlush` and fulfill each of their fulfillers. **That said, I don't know if this is what we want** -- I would have to look into it more. Not sure if we have a precedent for this.
### What to do with the flush
Assuming we've propagated the exception to `ActorCache::get()`, we now need to decide what to do about this ongoing flush. Should we throw a fatal exception from `syncGet()`? Maybe only if it's a `DISCONNECTED` exception?

At least in the case of `FAILED`, if we still throw the exception, then this `ActorCache read hard fail` test will still fail because the output gate will break. Perhaps there are certain types of exceptions that should be propagated to the get() caller (`FAILED/OVERLOADED`), and certain types that should only be thrown from `syncGet()` (`DISCONNECTED` so we retry).
### Do Entry's need to be removed from dirtyList/cache?
If we don't propagate `FAILED` exceptions from `syncGet()`, do we need to worry about the `Entry`s associated with the failed get() in the `dirtyList`? 

For example, let's say we fail the storage get(), send the exception to `ActorCache::get()`, and then don't throw from `syncGet()`. Assuming there are no other storage operations, we will continue down our flush path to the successful flush case, and iterate over the `dirtyList`, moving values into the `cleanList`. This seems wrong, although perhaps if the `GetWaiter` is destroyed these entries will also be removed from the `dirtyList` before the `flushProm` continuation runs?

I would think we would need to remove the entries from `dirtyList` inside `syncGet()`, but may need to think about this some more.

src/workerd/io/actor-cache.h
```diff
@@ -573,6 +583,35 @@ class ActorCache final: public ActorCacheInterface {
...
+  // A convenient object for holding all our `Entry`s that are waiting on the result of a get() or
+  // getMultiple() call. `get()` will only ever have 1 element in `entries`, unlike `getMultiple()`.
+  class GetWaiter {
+  public:
+    explicit GetWaiter(ActorCache& cache, kj::Array<kj::Own<Entry>> entries)
+      : cache(cache), entries(kj::mv(entries)) {}
+    KJ_DISALLOW_COPY_AND_MOVE(GetWaiter);
+    ~GetWaiter() noexcept(false) {
+      for (auto& entry : entries) {
+        if (!entry->isShared()) {
```
MellowYarker, line 595: This "cancel if the `GetWaiter` is destroyed" approach doesn't really work as is now. 

If the `GetWaiter` is holding the last reference to the `Entry`, then upon its destruction (because JS dropped the get() promise) we would remove the `Entry` from the `dirtyList` and we wouldn't attempt to do the get() when we get around to flushing. The thing is, we almost always give the `currentValues` map a strong reference to the `Entry`, then `addRef` that and put it in the `dirtyList` (see `findInCache()`). This means by the time the `GetWaiter` has been initialized, we will (in almost all cases) have at least 2 strong references to the `Entry` -- one owned by the `GetWaiter` and one owned by `currentValues`. When `GetWaiter`'s dtor runs, we won't even try to remove the Entry from `dirtyList` because it looks like it's shared.

I think one way to get around this is to change the check from "is shared" to "our Entry matches what's in `currentValues` + the status is `UNKNOWN`". This would imply we haven't changed the Entry since we initiated the `get()`.

---
There's an open question on if we want to try to cancel the storage get() during the flush. I think this might be more complicated than it's worth, since we would have already put `Entry`s into a `GetFlush`, and the storage operation might have already run. What if the values have already been retrieved from storage, and the waiter is just waiting for the flush promise to finish?

src/workerd/io/actor-cache-test.c++
```diff
@@ -4655,18 +4825,24 @@ KJ_TEST("ActorCache transaction multiple counted delete batches") {
...
+      .withParams(CAPNP(keys = ["count1", "count4"]), "stream"_kj)
       .useCallback("stream", [&](MockClient stream) {
     // Pretend that "count1" already exists but was not in the cache.
     stream.call("values", CAPNP(list = [(key = "count1", value = "1")]))
         .expectReturns(CAPNP(), ws);
     stream.call("end", CAPNP()).expectReturns(CAPNP(), ws);
-  }).expectCanceled();
+  }).thenReturn(CAPNP());
+
+  // TODO(now): I'm not fully sure I understand why this is a separate getMultiple?
```
a-robinson, line 4836: Isn't it a consequence of the `{.maxKeysPerRpc = 2}` config up at the start of this test? Or are you referencing something deeper than that?

src/workerd/io/actor-cache.h
```diff
@@ -319,6 +324,9 @@ class ActorCache final: public ActorCacheInterface {
     // The value was set by the app via put() or delete(), and we have not yet initiated a write
     // to disk. The entry is appended to `dirtyList` whenever entering this state.
     //
+    // Alternatively, the app requested a get, and while we aren't "writing" anything to storage,
```
a-robinson, line 327: Hmm, this is a pretty gross overloading of what it means to be `DIRTY` -- there's nothing here that needs to be written to storage, which is what dirty traditionally means in similar systems.

I'll get a better idea of how important this is after reading the rest of the code, but this really feels like it merits a new Status, or reusing one of the others like `CLEAN` or `NOT_IN_CACHE`.

src/workerd/io/actor-cache.h
```diff
@@ -384,15 +393,8 @@ class ActorCache final: public ActorCacheInterface {
 
   private:
     // The value associated with this key. If our `valueStatus` below is `ABSENT` or `UNKNOWN`, it
-    // will have size 0.
-    //
-    // `value` cannot change after the `Entry` is constructed. When a key is overwritten, the
```
a-robinson, line 389: The old comment here still provided valuable context IMO. It'd be better to clarify it with some special case (e.g. "`value` cannot change after the `Entry` is constructed and moved out of the FOO status" or something like that, which again points at the value of a different `EntrySyncStatus` for this sort of `Entry`; or of using `EntryValueStatus::UNKNOWN`, perhaps?).

src/workerd/io/actor-cache.h
a-robinson: > I think one way to get around this is to change the check from "is shared" to "our Entry matches what's in `currentValues` + the status is `UNKNOWN`". This would imply we haven't changed the Entry since we initiated the `get()`.

But what if some other `get()` request has been made for the same key? Couldn't that logic cause a false positive that would inadvertently cancel the read when it shouldn't?

(again, sorry for commenting on this without having yet read the full implementation -- I've gone from the test file to the header file so far).

---

> There's an open question on if we want to try to cancel the storage get() during the flush.

If the storage operation has already been sent (or packed into a `GetFlush` in preparation for sending), then yeah I see little marginal return in trying to cancel it. But it does (to me at least) seem meaningfully more valuable to attempt to avoid sending a get to storage if it's canceled before we get around to the flush call.

src/workerd/io/actor-cache.h
```diff
@@ -599,6 +638,18 @@ class ActorCache final: public ActorCacheInterface {
     auto begin() { return inner.begin(); }
     auto end() { return inner.end(); }
 
+    // The way we coalesce reads means we update `Entry`s while they're still in the `dirtyList`.
+    // This means the size of the list may change, without us adding or removing an Entry.
+    // To ensure proper accounting, we have to modify the size when we modify an Entry.
```
a-robinson, line 643: Heh, this seems like the kind of logic that will lead to a bug that breaks proper accounting. Again, my fault for not yet having read the implementation (so this is largely a note for me to come back to), but why are we updating the size of the entry *while* it's in the dirty list rather than when it's being removed from the dirty list (because the read is done and thus it no longer needs to be in the dirty list)?

src/workerd/io/actor-cache.h
```diff
@@ -599,6 +638,18 @@ class ActorCache final: public ActorCacheInterface {
...
+    void decreaseSize(size_t bytes) {
+      innerSize -= bytes;
```
a-robinson, line 645: Some kind of assertion that `innerSize >= bytes` seems important here if we need to keep this.

src/workerd/io/actor-cache.h
```diff
@@ -710,14 +775,25 @@ class ActorCache final: public ActorCacheInterface {
...
   //
   // Note that the returned entry could have `EntryValueStatus::UNKNOWN` which means we do not know
   // if it is in storage or `EntryValueStatus::ABSENT` which means we know it is not in storage.
   kj::Own<Entry> findInCache(Lock& lock, KeyPtr key, const ReadOptions& options);
 
+  // We skip flushing in preview sessions (we don't write to storage), so we use getForPreview()
+  // for reads instead.
+  kj::Promise<kj::Maybe<Value>> getForPreview(kj::Own<Entry> entry, ReadOptions options);
+
+  // TODO(now): Need to implement getMultiple for preview sessions.
```
a-robinson, line 792: Leaving a comment as a reminder.

src/workerd/io/actor-cache.c++
```diff
@@ -573,61 +617,84 @@ kj::OneOf<ActorCache::GetResultList, kj::Promise<ActorCache::GetResultList>>
...
-    } else {
-      // We'll be canceled momentarily...
-      return kj::NEVER_DONE;
-    }
+  // After this point, the only strong references to `Entry`s should be in the individual
+  // GetMultiStreamImpl objects.
+  getFlush.batches.clear();
+  getFlush.entries.clear();
+
+  // TODO(now): This doesn't seem to be firing...
```
a-robinson, line 690: The `readCancellation` promise only gets resolved during the `ActorCache` destructor. That `cancelReadFlush->fulfill()` call there adds an event to the kj event loop to call any callbacks that were waiting on the `readCancellation` promise, but it doesn't immediately call them. The `ActorCache` finishes getting destructed (and whatever code path caused the destruction of the `ActorCache` would also continue running up until it yields the event loop) before any of those callbacks would run.

And, importantly, `syncGets()` is only called from within the `flushImpl()` code path, which is a chain of async work owned by the `lastFlush` promise on the `ActorCache` class. So when the `ActorCache` instance gets destructed, that `lastFlush` promise will also get canceled.

So it would surprise me greatly (and probably be a bad thing) if there was some code path in which this promise fired. Which means we may need a different approach to more synchronously do whatever needs doing here during destruction (assuming it can't be handled gracefully via destructors/RAII).

a-robinson (commented):
Now that I make it to the implementation file I see enough TODO(now) comments that I suppose I may not be intended to review that yet, but if there's more that you want me to comment on beyond the couple big questions you left, let me know. I'll get to those big questions next, at least.

src/workerd/io/actor-cache.c++
```diff
@@ -2446,6 +2535,18 @@ kj::Promise<void> ActorCache::startFlushTransaction() {
...
   // - Thus all new read requests for our cached entries keys will be served from cache.
+  // TODO(now) We actually want a separate forked promise chain for reads so that we don't block on
+  // write flushes.
+  if (getFlush.entries.size() == 1) {
+    auto entry = kj::mv(getFlush.entries.front());
+    getFlush.batches.clear();
+    getFlush.entries.clear();
+    co_await syncGet(storage, kj::mv(entry));
+  } else if (getFlush.entries.size() > 1) {
+    co_await syncGets(storage, kj::mv(getFlush));
```
a-robinson, line 2574: Does this structuring of the code -- where after doing our coalesced reads we have to go ahead and finish executing this `startFlushTransaction` method, including successfully flushing all of our writes to storage, before the `lastFlush` promise will resolve -- mean that our `storage.get()`/`storage.getMultiple()` promises no longer resolve until *after* any dirty writes have been flushed to disk?

If so, that feels like a meaningful latency regression for reads given that writes take much longer than reads.

If not, I'm not seeing what mechanism would cause the reads to resolve prior to this method finishing.

src/workerd/io/actor-cache-test.c++
a-robinson: This all feels like the natural consequences of reusing `lastFlush` and the `dirtyList`. Can we keep the storage reads segregated onto a separate promise rather than integrating them so tightly into `lastFlush`?  e.g. a `coalescedReads` or `lastBatchedGet` promise or something.

It seems like that would simplify some of this desire to treat read failures separately from write failures (and, as asked about in my other comment, it'd allow reads to resolve prior to needing to wait on writes to flush).

---

As for removing entries from the `dirtyList`, I'm not sure I totally understand the question unfortunately. If we're throwing an exception to the caller for a given `get`, then I'd think we should remove it from the list (although presumably not by adding it to the `cleanList` given that we don't know its value). If we're still going to try to get the key's value and return it, then we shouldn't. But I suspect that isn't an answer to your question.

src/workerd/io/actor-cache.h
MellowYarker: Would you want it to happen as part of the actual removal from the list (i.e. `removeEntry()` or `DirtyList::remove()`?

I think I would find it less difficult to track a down a bug if we explicitly modified the size _prior_ to attempting to remove from `dirtyList` because we would know it's limited to get() operations rather than any arbitrary removal from `dirtyList`.

MellowYarker:
I turned the 2nd commit into a fixup because I resolved the TODOs I had and the rest of it was very small.

src/workerd/io/actor-cache.c++
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/497ded47c634e0505299694de82f7cef780c8864

This seems to get the test passing locally, but some of our internal tests are seeing segfaults now (need to look into them more, this is weirdly tricky). Any idea why destroying the joint promise wouldn't also cause the `GetMultiStreamImpl` to be dropped? I had to explicitly clear out the `GetMultiStreamImpl::entries`, was hoping to just dtor the object but it seems like its lifetime is extended further, but I'm not sure what's keeping it alive.

src/workerd/io/actor-cache-test.c++
```diff
@@ -2716,7 +2773,13 @@ KJ_TEST("ActorCache listReverse() with limit") {
   KJ_ASSERT(expectCached(test.get("fon")) == nullptr);
 
   // Stuff before the first key is not in cache.
+  // TODO(now): Maybe it'd be good to return a value here, because currently we are ABSENT
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/51b8f0f1b04bcf495a99243ede476c172968c85b

src/workerd/io/actor-cache-test.c++
MellowYarker: Thanks :) https://github.com/cloudflare/workerd/pull/1916/commits/6db7310a1d04ea381b6c0469970f9bb0b33b9208

src/workerd/io/actor-cache-test.c++
```diff
@@ -3623,6 +3718,19 @@ KJ_TEST("ActorCache LRU purge larger") {
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("corge"))) == kilobyte);
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("grault"))) == kilobyte);
   KJ_ASSERT(KJ_ASSERT_NONNULL(expectCached(test.get("garply"))) == kilobyte);
+
+  mockStorage->expectCall("getMultiple", ws)
+      .withParams(CAPNP(keys = ["bar", "baz", "qux"]), "stream"_kj)
+      .useCallback("stream", [&](MockClient stream) {
+    // TODO(now): Using `kilobyte` for value results in the error "External constants not allowed."
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/9ab5599ddf2d76bf1543668467b0f09817514a9a

src/workerd/io/actor-cache.h
a-robinson: That's what I would have expected, but you have a point, and at this point I don't remember the context enough to have a strong opinion. It's fine to leave as is.

src/workerd/io/actor-cache.c++
a-robinson: > Any idea why destroying the joint promise wouldn't also cause the GetMultiStreamImpl to be dropped?

If the call to `syncGets()` is still being awaited on somewhere, then the `clients` vector would still be getting held on the relevant coroutine stack. We could presumably fix that by either (A) moving the `clients` vector into an attachment on the joint promise, or (B) making sure that whatever promise that's keeping the `syncGets` call alive gets canceled.

And I'd guess what's keeping it alive is the `lastFlush` promise, which is held as a member variable on `ActorCache`.

But would you like me to try poking at this locally? It's always easier to figure cancellation / destruction order stuff out by testing stuff out than just by looking at the code and guessing.

src/workerd/io/actor-cache.c++
a-robinson: > I think the most obvious (potential) issue here is that although reads can now resolve before the flush finishes, the next batch of reads will have to wait for the next flush

That isn't the end of the world, although I also don't see why it would be necessary (aside from it being more convenient in the code).

a-robinson:
Sorry I waited so long to look at this, I assumed there was still more work to be done. And TBH I'm still not sure if this was waiting on me or if you were planning on doing more.

So to be clearer, what is the actual state of this and what can I do to help right now?

src/workerd/io/actor-cache.c++
MellowYarker: So, the test that would fail is `bazel test @workerd//src/workerd/io:actor-cache-test`, specifically `ActorCache list stream cancellation`. It seems like the joint promise does get cancelled (and the individual promises in the loop also get cancelled), but the `GetMultiStreamImpl`, which we move into a `rpc::ActorStorage::ListStream::Client`, doesn't seem to get destroyed.

I'm not sure if it's because the `ListStream::Client` is living long, or if the `req` object which takes a reference to the `ListStream::Client` is living long. I suspect we're somehow creating a strong reference to the capability (briefly looking at `req.setStream()`, I see we eventually call `Capability::Client::hook->addRef()` in `PointerHelpers<>::set()`), preventing it from being dtor'd even though the promise is done.

MellowYarker:
FYI @a-robinson I'm going to push up some new work as a third commit today, with the eventual goal of squashing the commits into 1 before we finally merge this.

Just want to keep this one separate for now since it's giving me some trouble... This third commit focuses on having reads intercept exceptions from RPC calls, and then either propagating them to the `GetWaiters`, or allowing the exception trigger a retry (via the normal flush retry mechanism).

This fixes the test `"ActorCache read hard fail"` test, but this test only checks that single key gets() receive the exception. I've added another (currently broken) test that extends `"ActorCache read hard fail"` to see if we get the same behavior when we have multiple keys awaiting a response from storage. I'm not sure if the mock framework actually lets us test this or not though.

It seems this change has also broken 9 other tests as well, I haven't had time to look into why just yet because: tl;dr other tests were broken due to a very subtle issue with my initial implementation and I was chasing that bug the last day and a half :grimacing:. Anyways, I suspect these are all breaking for one or 2 reasons that shouldn't be too hard to clear up.

src/workerd/io/actor-cache.h
MellowYarker: > But what if some other get() request has been made for the same key? Couldn't that logic cause a false positive that would inadvertently cancel the read when it shouldn't?

Yeah, that's a solid point. If we have 2 outstanding gets() for one key (i.e. two separate `GetWaiter`s), then they'll each hold a strong ref (via `addAtomicRef()`), and so if one of the `GetWaiter`s is destroyed, we still can't assume with that a match within `currentValue` & status `UNKNOWN` means we're the only outstanding get.

Maybe we can put a counter on the `Entry` which we increment/decrement whenever a GetWaiter with that Entry is created/destroyed? If it goes down to 1 we know we're the last waiter?

src/workerd/io/actor-cache.c++
a-robinson: It looks like this is fixed now? At least it's not failing for me when I pull this branch down and try to repro it.

MellowYarker: Yup! This test is fine, but one of our internal tests is failing because of a segfault that was introduced when fixing `ActorCache list stream cancellation`. I'll let you know which one some time this week, not able to connect to wifi on my laptop for some reason.

When we destroy `ActorCache`, we go through and destroy the promises in `syncGets()`, which results in us calling `GetMultiStreamImpl::clear()`. However, sometimes this object has already been destroyed by the time we call `clear()`. There's probably a reasonable fix, but this already feels a bit hacky, which is why I mentioned that it'd be better to find a more robust approach.

src/workerd/io/actor-cache-test.c++
```diff
@@ -3806,6 +3929,14 @@ KJ_TEST("ActorCache lru evict entry with known-empty gaps") {
   (void)expectUncached(test.get("baza"));
   (void)expectUncached(test.get("corge"));
   (void)expectUncached(test.get("fo"));
+  // We coalesce these gets.
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/13bf180bd92857f7ebe5dedf9a1b42a2d530af81

src/workerd/io/actor-cache-test.c++
```diff
@@ -2864,8 +2927,17 @@ KJ_TEST("ActorCache listReverse() multiple ranges") {
   KJ_ASSERT(expectCached(test.listReverse("x", "z")) == kvs({{"y", "9"}}));
 
   (void)expectUncached(test.get("w"));
+  mockStorage->expectCall("get", ws)
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/13bf180bd92857f7ebe5dedf9a1b42a2d530af81

src/workerd/io/actor-cache-test.c++
```diff
@@ -1715,6 +1752,8 @@ KJ_TEST("ActorCache list() with limit") {
 
   // Stuff after the last key is not in cache.
   (void)expectUncached(test.get("fooa"));
+  // Return our uncached get.
+  mockStorage->expectCall("get", ws).withParams(CAPNP(key = "fooa")).thenReturn(CAPNP());
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/13bf180bd92857f7ebe5dedf9a1b42a2d530af81

src/workerd/io/actor-cache-test.c++
```diff
@@ -1444,76 +1424,131 @@ KJ_TEST("ActorCache read retry on flush containing only puts") {
...
-  // Expect the get, but intentionally don't resolve it.
-  auto mockGet = mockStorage->expectCall("get", ws)
-      .withParams(CAPNP(key = "foo"));
+  test.put("foo", "456");
+  test.put("bar", "789");
+  test.put("baz", "123");
 
-  // We won't write anything until the read completes.
-  mockStorage->expectNoActivity(ws);
+  // Since we still have the promise for foo and bar, we do send a get for them. However, 'baz' will
```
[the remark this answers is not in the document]
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/13bf180bd92857f7ebe5dedf9a1b42a2d530af81

src/workerd/io/actor-cache.c++
MellowYarker: https://github.com/cloudflare/workerd/pull/1916/commits/875e4aca6dfc652578cf2c02b04762b4071d658a fixes the segfault and I think is sufficiently commented on.

MellowYarker:
The latest commit https://github.com/cloudflare/workerd/pull/1916/commits/00ba98560d35f0f7fe216b989ec17f4d07eec522 uncomments the "getMultiple hard fail" test and attempts to address the reason for the test failure, but for some reason the exception seems to be causing more problems in this case than in the single-key get() test.

We do manage to propagate it to the caller (the test in this case), but the exception is getting logged and the test is failing.

I thought it might have to do with how we are catching the exception over RPC, or maybe that we give copies to more than 1 promise fulfiller (which ends up in a single joined promise)?

MellowYarker:
https://github.com/cloudflare/workerd/compare/00ba98560d35f0f7fe216b989ec17f4d07eec522..7f5b18fac3bab8fc6901ba1345a82b4ad0df709d is a rebase near or on `main` since this work might be paused for a bit.

PR #1917, opened 2024-03-28
add WorkerVersionMetadata binding type

RamIdeas:
This PR adds a type for the "version_metadata" binding

It's not urgent to get this merged as I will make `wrangler types` output this type literally (since it's very simply) and can amend that command to use this built-in type when available


types/defines/versions.ts
mrbbot, line 1: This should be a `.d.ts` file instead of `.ts`.

mrbbot (approved):

PR #1918, opened 2024-03-28
Delete all alarms on `abortAllDurableObjects()`

mrbbot:
Fixes https://github.com/cloudflare/workers-sdk/issues/5388

In the Workers Vitest integration, we call the `abortAllDurableObjects()` function after each test to abort all Durable Objects  and close all SQLite databases. If the isolated storage option is enabled, we backup/restore `.sqlite` files to automatically undo writes performed in tests. Right now, this doesn't cancel Durable Object alarms. This means that an alarm scheduled in one test might end up firing when another later test is running. This could lead to unexpected test results, and non-determinism in tests.

This PR addresses this issue by automatically deleting all Durable Object alarms when `abortAllDurableObjects()` is called. This is a local-only internal API that's currently only used in the Vitest integration, so making this breaking change is fine. 👍

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -2548,6 +2548,8 @@ void Server::abortAllActors() {
       }
     }
   }
+
+  alarmScheduler->deleteAllAlarms();
 }
 
 kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name, config::Worker::Reader conf,
--- a/src/workerd/server/alarm-scheduler.c++
+++ b/src/workerd/server/alarm-scheduler.c++
@@ -150,6 +150,25 @@ bool AlarmScheduler::deleteAlarm(ActorKey actor) {
   return query.changeCount() > 0;
 }
 
+void AlarmScheduler::deleteAllAlarms() {
+  stmtDeleteAllAlarms.run();
+
+  alarms.eraseAll([this](ActorKey& key, ScheduledAlarm& value) {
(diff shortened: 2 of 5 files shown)


jasnell (approved):

allan-simon:
is there a reason why this PR was not merged, it would help solve the known issue of vitest integration ? 

If necessary I can reopen a similar PR rebased on latest master if @mrbbot has not the time  ?

src/workerd/server/server.c++
```diff
@@ -3871,6 +3871,8 @@ void Server::abortAllActors(kj::Maybe<const kj::Exception&> reason) {
       }
     }
   }
+
+  alarmScheduler->deleteAllAlarms();
```
anonrig, line 3905: Can you add a comment to why we delete all alarms when `abortAllActors` are called? It might not be obvious for others.

anonrig (approved):

kentonv:
Ironically, a week after this PR was merged, vitest-pool-workers actually removed its usage of this API. After two years of waiting...

And after some discussion, we feel that it doesn't make sense for abortAllDurableObjects() to cancel alarms. Cancelling alarms is essentially _deleting data_ from the DOs, but since it doesn't delete _all_ data, it leaves them in an inconsistent state.

It could make sense to add a new deleteAllDurableObjects() API which aborts DOs, cancels alarms, and deletes underlying data, but that should be a separate API.

In any case I will be rolling this back as it is blocking #6104.

PR #1919, opened 2024-03-28
Fix bug with reportUndefinedSymbols shim

garrettgu10:
Looks like #1868 also replaces a call to Module.reportUndefinedSymbols, which throws an error in local dev because this value is assigned in two other places like so:

```
Module.reportUndefinedSymbols = () => {}
...
Module["reportUndefinedSymbols"] = ...
```

Which are not matched by the rule. This PR simply refuses to replace `Module.reportUndefinedSymbols()` with `Module.reportUndefinedSymbolsNoOp()`.

### DIFF
--- a/src/pyodide/BUILD.bazel
+++ b/src/pyodide/BUILD.bazel
@@ -124,6 +124,10 @@ REPLACEMENTS = [
     [
         "reportUndefinedSymbols()",
         "reportUndefinedSymbolsNoOp()"
+    ],
+    [
+        "Module.reportUndefinedSymbolsNoOp()",
+        "Module.reportUndefinedSymbols()"
     ]
 ]
 


hoodmane (approved):
This change confuses me but the tests pass so I'll approve. Thanks @garrettgu10!

garrettgu10:
@hoodmane

I'm also confused about why emscripten has both reportUndefinedSymbols and Module.reportUndefinedSymbols, maybe it would be smart to get to the bottom of that before merging...

another alternative fix would be to replace "reportUndefinedSymbols" instead of "reportUndefinedSymbols()". I went with this approach because I was concerned about changing the API of Module, but I think that might be more consistent with other substitutions we make. What do you think?

hoodmane:
> both reportUndefinedSymbols and Module.reportUndefinedSymbols

Generally `Module.blah` is the "public" version of things, so anything that exposed to user code is put there. Possibly the `Module.reportUndefinedSymbols` stuff comes from Pyodide.

Replacing `reportUndefinedSymbols` by itself doesn't work because then we also replace the definition and we get a `reportUndefinedSymbolsNoOp` that is not a no-op.

PR #1920, opened 2024-03-28
Fix universal snapshots again

hoodmane:


src/pyodide/internal/python.js
```diff
@@ -397,6 +420,9 @@ function memorySnapshotDoImports(Module) {
 function makeLinearMemorySnapshot(Module) {
   memorySnapshotDoImports(Module);
   const dsoJSON = recordDsoHandles(Module);
+  if (IS_CREATING_BASELINE_SNAPSHOT) {
+    // checkLoadedSoFiles(dsoJSON);
```
dom96, line 424: Should this be commented out?

dom96 (approved):

src/pyodide/internal/python.js
hoodmane: It doesn't work with the intentionally messed up creation logic in `preloadDynamicLibs` so we can leave a todo to enable a check like that.

PR #1922, opened 2024-03-29
Different fix for reportUndefinedSymbolsNoOp bug

garrettgu10:
See #1919 

Instead of keeping Modules.reportUndefinedSymbols, this PR actually defines Modules.reportUndefinedSymbolsNoOp and performs the substitution as before.


hoodmane (approved):
Thanks @garrettgu10!

PR #1924, opened 2024-03-29
Update Release docs

Warfields:
Added explanation on why release actions are done on main, not on the release tag.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -21,7 +21,7 @@ Once all binaries (5 platforms) have been automatically added to the release (fe
 
 - navigate to "Actions" github tab
 - pick "Publish to NPM" workflow
-- *keep* main branch
+- select the tag created for the release. (Ex v1.20240327.0 where the middle version is the new compat date)
 - *keep* patch 0 unless there was a problem with release and you need to publish a patch version (increment in that case).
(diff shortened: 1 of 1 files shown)


hoodmane (approved):

RELEASE.md
```diff
@@ -22,6 +22,9 @@ Once all binaries (5 platforms) have been automatically added to the release (fe
 - navigate to "Actions" github tab
 - pick "Publish to NPM" workflow
 - *keep* main branch
+  - This must be main, not the version tag. The workflow is privileged and requires access to github
```
garrettgu10, line 25:
```suggestion
  - This must be main, not the version tag. The workflow requires access to GitHub
```

RELEASE.md
```diff
@@ -22,6 +22,9 @@ Once all binaries (5 platforms) have been automatically added to the release (fe
...
+  repo secrets, so it can only be executed from a protected branch like main, otherwise it can't
+  get the required API keys
```
garrettgu10, line 27:
```suggestion
  get the required npm API key.
```

fhanau:
I actually don't think this is correct – at least the `Publish to NPM` action has successful runs from a tag. I think avoiding secrets being misused is intended to already be ensured by only the team being allowed to run these actions manually.

fhanau:
In fact, maybe we **should** recommend to use the release tag in case there are changes on main that add new APIs that aren't part of the actual release yet.

fhanau (approved):

garrettgu10:
Shoild we also update guidance for workers-types?

Warfields:
I agree with @fhanau.

> Shoild we also update guidance for workers-types?

Yes.

PR #1926, opened 2024-03-31
Revert "MQ-576: Add attempts to QueueMessage"

garrettgu10:
Reverts cloudflare/workerd#1874


mrbbot (changes requested):
I don't think we need to revert this. Should be able to bump `workerd` in `workers-sdk` with something like: https://github.com/cloudflare/workers-sdk/pull/5455.

PR #1933, opened 2024-04-02
Support long names in tarfs

garrettgu10:
Previously file names (which are actually file paths + names) could not exceed 100 bytes. This supports the `gnu` style of long names, which is the default behavior of `tar` on my debian system. We will probably need to support this eventually since some file names are already up to 97 bytes.


hoodmane:
Too bad we don't have unit tests for this file... I wonder how hard it would be to add some. Maybe not worth blocking this change for it though.

hoodmane (approved):

hoodmane:
Thanks @garrettgu10!

PR #1952, opened 2024-04-03
Python: Make top level random() raise

hoodmane:
We don't have access to entropy outside of request scope, so `random()` will give poor results. Raise an error if it is called.

### DIFF
--- a/src/pyodide/internal/python.js
+++ b/src/pyodide/internal/python.js
@@ -1,3 +1,4 @@
+Error.stackTraceLimit = Infinity;
 import { enterJaegerSpan } from "pyodide-internal:jaeger";
 import {
   SITE_PACKAGES_INFO,
@@ -12,6 +13,11 @@ import {
   maybeSetupSnapshotUpload,
   restoreSnapshot,
 } from "pyodide-internal:snapshot";
+import {
+  entropyMountFiles,
+  entropyAfterRuntimeInit,
+  entropyBeforeTopLevel,
+} from "pyodide-internal:topLevelEntropy/lib";
 
 /**
  * This file is a simplified version of the Pyodide loader:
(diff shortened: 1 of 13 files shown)


src/workerd/server/tests/python/random/worker.py
```diff
@@ -0,0 +1,41 @@
...
+  pass
+else:
+  assert False
+
+try:
+  randbytes(5)
+except RuntimeError:
+  pass
+else:
+  print("randbytes() no")
```
garrettgu10, line 15: Probably not worth restarting CI for this, but Is this print() supposed to be here?

garrettgu10 (approved):

src/pyodide/internal/patches/random.py
```diff
@@ -0,0 +1,27 @@
+import random as _orig_random
```
dom96, line 1:
```suggestion
# Wraps the `random` module so that calls to any methods within it raise
# unless they are called as part of a request.
import random as _orig_random
```

src/pyodide/internal/patches/random.py
```diff
@@ -0,0 +1,27 @@
...
+from functools import wraps as _wraps
+import sys as _sys
+
+_IS_READY = False
```
dom96, line 5:
```suggestion
_IN_REQUEST_CONTEXT = False
```

src/pyodide/internal/python.js
```diff
@@ -206,5 +207,19 @@ export async function loadPyodide(lockfile, indexURL) {
...
+      exec(modStr, module.__dict__)
+      import sys
+      sys.modules["random"] = module
+  `);
+  const initRandom = pyodide.globals.pop("_tmp");
+  const randomPatch = new TextDecoder().decode(
+    new Uint8Array(randomPatchBuffer),
+  );
+  initRandom(randomPatch);
+  initRandom.destroy();
```
dom96, line 223: Refactor this into a `wrapRandomModule` method.

src/workerd/server/tests/python/random/worker.py
```diff
@@ -0,0 +1,41 @@
+from random import random, randbytes, choice
```
dom96, line 1:
```suggestion
# Verifies that calling `random` at the top-level throws. Calls to random should only work inside a request context.
from random import random, randbytes, choice
```

dom96 (approved):
Just a few suggestions to make the code easier to understand, otherwise LGTM.

hoodmane:
Okay, now there is a problem that rust uses randomness for hash seeds:
https://github.com/rust-lang/rust/blob/master/library/std/src/sys/pal/unix/rand.rs#L235-L247
https://github.com/rust-lang/rust/blob/master/library/std/src/hash/random.rs?plain=1#L68

PR #1964, opened 2024-04-04
[build] Fix Unix release configuration

fhanau:
When adding the release configuration, `release_unix` failed to actually enable `release` and `opt`. Performance should not be affected significantly as release_unix sets -O3, but there is a sizable regression in macOS binary size based on different linker flags being used in fastbuild resulting in worse dead code elimination (113MB vs. 147MB).


jasnell (approved):

jp4a50 (approved):

fhanau:
Superseded by #1986.

PR #1970, opened 2024-04-04
Use disk cache to run wd tests

garrettgu10:
This PR does a few things:
1. It updates the pyodide bucket to one that includes a file called `all_wheels.zip` in its GitHub release.
2. It uses this file to create a bazel repo that contains all the wheels
3. It passes a path to the bazel repo as `--disk-cache-dir` when the test is being run.

Open questions:
1. Does this work on Windows
2. Does this break tests on edgeworker somehow

Why not just download the wheels from R2? The main reason is that the R2 bucket is not accessible from our build infrastructure.


dom96:
> The main reason is that the R2 bucket is not accessible from our build infrastructure.

Why is that? Can't we make the R2 bucket accessible from there?

WORKSPACE
```diff
@@ -126,14 +126,28 @@ http_archive(
...
 
 http_file(
     name = "pyodide-lock.json",
     sha256 = PYODIDE_LOCK_SHA256,
     url = PYODIDE_GITHUB_RELEASE_URL + "pyodide-lock.json",
 )
 
+http_archive(
+    name = "all_pyodide_wheels",
+    sha256 = PYODIDE_ALL_WHEELS_ZIP_SHA256,
```
dom96, line 139: This variable is only used here. Why not just stick the value directly here?

build/pyodide_bucket.bzl
```diff
@@ -1,7 +1,8 @@
 # Do not edit this file by hand. See docs/pyodide.md for info on how to generate it.
 # These variables are factored out here because they are being shared by the WORKSPACE files in
 # both edgeworker and workerd, as well as src/pyodide/BUILD.bazel
-PYODIDE_PACKAGE_BUCKET_URL = "https://pub-45d734c4145d4285b343833ee450ef38.r2.dev/20240320/"
-PYODIDE_GITHUB_RELEASE_URL = "https://github.com/cloudflare/pyodide-build-scripts/releases/download/20240320/"
-PYODIDE_LOCK_SHA256 = "a176311d4c449aac4ef7a333977af8b6e08224c115a9a6d05c04592c841b8a58"
-PYODIDE_PACKAGES_TAR_ZIP_SHA256 = "e191bae60aad75f6a9c33fac5c0ff1ad2b0e564bdd7a07fbdc848df4b62c60a1"
+PYODIDE_PACKAGE_BUCKET_URL = "https://pub-45d734c4145d4285b343833ee450ef38.r2.dev/20240404.1/"
```
dom96, line 4: Why was this updated?

samples/pyodide-langchain/worker.py
```diff
@@ -1,6 +1,7 @@
 from js import Response
 from langchain_core.prompts import PromptTemplate
 from langchain_openai import OpenAI
+import requests
```
dom96, line 4: This unused so can be removed, right?

```suggestion
```

WORKSPACE
hoodmane: I'm pretty sure the point is that `pyodide_bucket.bzl` is a generated file.

build/pyodide_bucket.bzl
garrettgu10: We generated a new package bundle at `pyodide-build-scripts`. The only difference should be the new added zip file.

hoodmane (approved):
Thanks @garrettgu10!

PR #1976, opened 2024-04-05
[build] Add -fvisibility-inlines-hidden

fhanau:
Mark C++ inline methods as hidden, causing them to not be exported when creating shared libraries.

This significantly reduces binary sizes and possibly link times on Linux, where shared linkage is used for everything except the workerd binary (build folder size shrinks 7011MB => 6954 MB based on CI statistics). There is a smaller binary size improvement for macOS, where shared linkage is not enabled by default.
I want to also enable actual hidden visibility (i.e. `-fvisibility=hidden`) for V8, but that likely requires significant patching as it is only supported in the V8 GN build so far.
See https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-fvisibility-inlines-hidden for background information.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -80,6 +80,10 @@ build:debug -c dbg
 
 # Using simple template names saves around 5% of binary size of workerd.
 build:unix --cxxopt='-gsimple-template-names' --host_cxxopt='-gsimple-template-names'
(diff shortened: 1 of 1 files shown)


hoodmane (approved):
Sounds reasonable to me.

PR #1977, opened 2024-04-05
Add ResolveObserver to jsg/observer.h

jasnell:
This separates out the new ResolveObserver interface from the new module registry PR in order to make it easier to review and further reduce the size of that other PR. This only adds the definition and does not introduce any code that uses it. That'll come with the module registry PR.

The idea is to improve observability of the module resolution process, including tracking the failure cases. For workerd, these observers are non-ops by default but can be set up by embedders to have actual backing implementations.

Refs: https://github.com/cloudflare/workerd/pull/1553

### DIFF
--- a/src/workerd/jsg/observer.h
+++ b/src/workerd/jsg/observer.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <workerd/jsg/url.h>
 #include <kj/common.h>
 #include <kj/string.h>
 #include <kj/exception.h>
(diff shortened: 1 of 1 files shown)


src/workerd/jsg/observer.h
```diff
@@ -24,18 +96,40 @@ struct CompilationObserver {
...
-  // It is guaranteed that isolate lock is held during both invocations.
+  // It is guaranteed that isolate lock is held during invocation.
   virtual kj::Own<void> onWasmCompilationStart(v8::Isolate* isolate, size_t codeSize) const {
     return kj::Own<void>();
   }
+
+  // Variation that is called at the start of wasm compilation from cache.
+  // Returned value will be destroyed when module compilation finishes.
+  // It is guaranteed that isolate lock is held during invocation.
+  virtual kj::Own<void> onWasmCompilationStart(v8::Isolate* isolate) const {
```
dom96, line 123: Nit: but should this maybe have "cache" in the name?

dom96 (approved):

PR #1979, opened 2024-04-06
Implements the web platform standard reportError API

jasnell:
Refs: https://developer.mozilla.org/en-US/docs/Web/API/reportError

/cc @irvinebroque

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -1025,6 +1025,26 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
         }
         KJ_LOG(INFO, "console warning", message);
       });
+      lock->setErrorReporterCallback([this](jsg::Lock& js, kj::String desc,
+                                            const jsg::JsValue& error,
+                                            const jsg::JsMessage& message) {
+        // Only add exception to trace when running within an I/O context with a tracer.
+        if (IoContext::hasCurrent()) {
+          auto& ioContext = IoContext::current();
+          KJ_IF_SOME(tracer, ioContext.getWorkerTracer()) {
+            addExceptionToTrace(js, ioContext, tracer,
+                UncaughtExceptionSource::REQUEST_HANDLER, error,
+                api->getErrorInterfaceTypeHandler(js));
+          }
+        }
+
+        KJ_IF_SOME(i, impl->inspector) {
+          jsg::sendExceptionToInspector(js, *i.get(), kj::str(desc), error, message);
+        }
+
+        // Run with --verbose to log JS exceptions to stderr. Useful when running tests.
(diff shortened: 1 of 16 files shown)


samples/helloworld_esm/worker.js
```diff
@@ -2,8 +2,17 @@
...
+  console.log('reported error:', event.error);
+//  event.preventDefault();
+});
+
+reportError(new DOMException('boom'));
+
 export default {
   async fetch(req, env) {
+    reportError(new DOMException('boom'));
+
```
jasnell, line 15: The changes in this file are temporary... being used for testing... proper tests and sample files will be added.

src/workerd/io/worker.c++
```diff
@@ -1025,6 +1025,12 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
         }
         KJ_LOG(INFO, "console warning", message);
       });
+      lock->setErrorReporterCallback([this](jsg::Lock& js, kj::StringPtr message) {
+        if (impl->inspector != kj::none) {
+          logMessage(js, static_cast<uint16_t>(cdp::LogType::ERROR), message);
+        }
+        KJ_LOG(INFO, "unhandled error", message);
```
jasnell, line 1032: Need to think about this a bit. When the error is reported to the inspector, the stack is properly reported. But the stack is dropped from the `KJ_LOG` line. Is that ok? Probably not. what else can we do here?

irvinebroque:
If I call `reportError` — what happens here with [Tail Workers](https://developers.cloudflare.com/workers/observability/logging/tail-workers/)? Will the error will show up in the `exceptions` field or the `logs` field?

https://github.com/wintercg/proposal-common-minimum-api/issues/41
https://github.com/nodejs/node/issues/38947

jasnell:
> ... what happens here with Tail Workers?

I think that still needs to be determined. The reporting is the same as the existing `logUncaughtException(...)` but I don't believe that ends up directly in the tail worker's `exceptions` field. That'll be something we want to determine, I think.

irvinebroque:
Yeah we should figure out before merging any change, since would be weird if this ended up showing as logs. @zebp and @maxwellpeterson may know and can point to internal tail worker code

jasnell:
@irvinebroque ... ok, the PR has been updated to publish the reported error to the worker tracer as an exception. Unfortunately we can't actually have a test to verify it in workerd since workerd does not initialize the worker tracer at all. Manually verified in an internal test, however, that the reported error does, in fact, show up in the worker tracer as an exception, so I think we're good on that front.

irvinebroque (commented):
Supportive of landing this

harrishancock (approved):

PR #1986, opened 2024-04-07
Reland PR #1900 "Add platform-specific release configuration"

fhanau:
#1900 failed to apply the `release` configuration to `release_unix`, leading to a binary size regression, especially on macOS. This was caused by bazel using fastbuild mode instead of opt mode, leading to Rust code not being optimized and -DDEBUG being defined on macOS through the bazel toolchain defaults.

- Fix a typo and clarify Windows release config comment

Original commit message:
- Match macOS wd_benchmark link options with wd_binary options
- Correct spelling of WIN32_LEAN_AND_MEAN
- /await and /GF are ignored by clang-cl, no need to pass them in the Windows build.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -101,10 +101,8 @@ build --@rules_rust//:extra_exec_rustc_flags=-C,panic=abort,-C,debug-assertions=
 build:fastdbg --@rules_rust//:extra_rustc_flags=-C,panic=unwind,-C,debug-assertions=y,-C,debuginfo=1,-C,embed-bitcode=n
 build:fastdbg --@rules_rust//:extra_exec_rustc_flags=-C,panic=unwind,-C,debug-assertions=y,-C,debuginfo=1,-C,embed-bitcode=n
 
-# TODO(later): -C codegen-units=1 improves code size and quality, should be enabled in a future
-# release configuration even if lto is off. Similarly, adding -C lto=thin would improve binary size
-# significantly – disable it for now due to compile errors and wrong code generation when bazel and
-# rust use different LLVM versions.
+# Adding -C lto=thin here would improve binary size significantly – disable it for now due to
+# compile errors and wrong code generation when bazel and rust use different LLVM versions.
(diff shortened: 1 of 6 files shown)


fhanau:
Relevant diff compared to initial PR: https://github.com/cloudflare/workerd/compare/3eb728b0...felix/bazel-release-cfg-reland#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832f

I initially assumed that linker flag issues were responsible for the large impact for macOS, but that turned out not to be an issue with the new linker introduced with Xcode 15. It's rather unintuitive that the macOS toolchain [defines `-DDEBUG`](https://github.com/bazelbuild/apple_support/blob/d59a3e0fa67ff097182165d3f3b6910c450ea92a/crosstool/cc_toolchain_config.bzl#L1070) for fastbuild while the Linux one doesn't – might be useful to fix this later on.

dom96 (commented):
Is there no way in bazel to detect which platform we're on and apply the platform-specific release flags appropriately?

It would be nice not having to worry about changing the config depending on the platform.

fhanau:
> Is there no way in bazel to detect which platform we're on and apply the platform-specific release flags appropriately?
> 
> It would be nice not having to worry about changing the config depending on the platform.

That's a valid concern, but unfortunately bazel is not very helpful here. We can select specific flags based on the build configuration for individual bazel targets, for example `build/wd_cc_binary.bzl` enables macOS-specific flags for the workerd binary in non-debug builds. Unfortunately it is difficult to do so for all build targets, including those in external dependencies like V8, this would require defining a custom toolchain (highly complex although we use this approach in the downstream repo).
With `--enable_platform_specific_config` enabled we can define platform specific flags in .bazelrc, and can use `build:macos --copt=[...]` but it doesn't allow enabling flags for the intersection of a platform and a build configuration. It would be great if setting `build:macos_opt` in .bazelrc allowed setting flags for builds using `-c opt` on macOS, but bazel is not very flexible here. The PR tries to emulate this by defining `release_<platform>` configurations and enabling them in the CI configuration.

mikea (approved):

PR #1993, opened 2024-04-08
call syncTime for hibernatable websocket autoresponses

Frederik-Baetens:
I believe it should be ok to only do this for the autoresponse because I think the non-autoresponse path already calls synctime.


jasnell (approved):
Any way to test this?

Frederik-Baetens:
> Any way to test this?

There likely is. I'll have a look. I'm not really up to speed on what the tests for this code look like, but I don't expect this to be too difficult to test.

nvie:
> Any way to test this?

I made a [minimal replication](https://github.com/nvie/cloudflare-repros/tree/gwsart-bug) for this bug. One way to test this would be to deploy this worker to an environment with this fixed applied and run the client script (see the video).

PR #2005, opened 2024-04-10
update v8 to 12.4

garrettgu10:


WORKSPACE
```diff
@@ -447,15 +447,13 @@ http_archive(
         "//:patches/v8/0010-Enable-V8-shared-linkage.patch",
         "//:patches/v8/0011-Randomize-the-initial-ExecutionContextId-used-by-the.patch",
         "//:patches/v8/0012-Always-enable-continuation-preserved-data-in-the-bui.patch",
-        "//:patches/v8/0013-Attach-continuation-context-to-Promise-thenable-task.patch",
-        "//:patches/v8/0014-increase-visibility-of-virtual-method.patch",
-        "//:patches/v8/0015-Add-ValueSerializer-SetTreatFunctionsAsHostObjects.patch",
-        "//:patches/v8/0016-wasm-liftoff-arm64-Fix-LoadTaggedPointer.patch",
+        "//:patches/v8/0013-increase-visibility-of-virtual-method.patch",
```
jasnell, line 450: I assume the continuation context patch is removed because 12.4 has the upstream fix available?

garrettgu10: Yep! this test was already removed in upstream

fhanau:
Looks like this is failing because the bazel rule for building torque is not working properly – perhaps https://chromium-review.googlesource.com/c/v8/v8/+/5339896 is responsible – not changing `v8root` based on whether V8 is built from another bazel workspace might be responsible for bazel not finding the torque output files.

fhanau:
@garrettgu10 for the patch to be applied make sure to add it to the list of v8 patches in WORKSPACE.

jasnell (approved):

PR #2007, opened 2024-04-11
Implements Python AI binding API test.

dom96:


src/cloudflare/internal/test/ai/ai-api-test.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2023 Cloudflare, Inc.
+# Licensed under the Apache 2.0 license found in the LICENSE file or at:
+#     https://opensource.org/licenses/Apache-2.0
+
+async def test(context, env):
+  resp = await env.ai.run('testModel', {"prompt": 'test'})
```
hoodmane, line 6: I would think this wouldn't work as-is because of the type conversions problems. Does the mock validate the second parameter?

dom96: Yeah, the mock doesn't validate the second parameter. It's a very simple test mock.

dom96:
@hoodmane @garrettgu10 review/stamp appreciated

hoodmane (approved):

Issue #1422, opened 2023-11-21
🐛 wrangler dev error: Received signal #11: Segmentation fault: 11


ds300:
Hello!

Recently we've been seeing this issue with one of our worker scripts (we have three, the other two seem fine) where after some time being used in dev mode it will segfault and become unresponsive. This seems to happen anywhere from 10 seconds to 10 minutes after booting. I have not been able to find a way to reproduce it on demand, but it will happen _every_ time, at some point.

This is on MacOS, using M1 chips.

The full error that is printed when this happens is **(`--log-level debug`)**

```
*** Received signal #11: Segmentation fault: 11
stack: 
```

There are always other errors logged out above, but they seem to change every time and are maybe unrelated since they seem to happen all the time regardless of how long it takes to hit the segfault. Here's an example (again `--log-level debug`)

```
workerd/io/worker-entrypoint.c++:218: info: exception = kj/compat/http.c++:2673: disconnected: WebSocket disconnected between frames without sending `Close`.
stack: 105822c40 10689dd88 105824bd8 10529a004 10506a108 10506a517 104f40d78
workerd/io/worker-entrypoint.c++:218: info: exception = kj/compat/http.c++:2673: disconnected: WebSocket disconnected between frames without sending `Close`.
stack: 105822c40 10689dd88 105824bd8 105824d97 10689dd88 105842d8c 105824bd8 10529a004 10506a108 10506a517 104f40d78
workerd/io/worker-entrypoint.c++:218: info: exception = kj/compat/http.c++:2673: disconnected: WebSocket disconnected between frames without sending `Close`.
stack: 105822c40 10689dd88 105824bd8 105824d97 10689dd88 105842d8c 105824bd8 105824d97 10689dd88 105842d8c 10529a004 10506a108 10506a517 104f40d78
workerd/server/server.c++:2838: error: Uncaught exception: kj/compat/http.c++:2673: disconnected: worker_do_not_log; Request failed due to internal error
stack: 105822c40 10689dd88 105824bd8 105824d97 10689dd88 105842d8c 105824bd8 105824d97 10689dd88 105842d8c 10529a004 10506a108 10506a517 104f40d78 104a85c38 10584ebdc 10584fb84 10

workerd/io/worker-entrypoint.c++:218: info: exception = kj/async.c++:220: disconnected: other end of WebSocketPipe was destroyed

[...]

PR #2011, opened 2024-04-12
Reuse ActorContainerRef if it exists

MellowYarker:
Prior to this commit, if a Durable Object in Workerd (local dev/miniflare) hibernated (i.e. Worker::Actor was destroyed), but the stub wasn't dropped by the worker (i.e. PromisedWorkerInterface still alive), then we would incorrectly create an entirely new refcounted ActorContainerRef upon a new request coming in.

This is incorrect because the old ActorContainerRef is still owned, and by creating an entirely separate one, we break our request tracking.

Specifically, the ActorContainer only tracks a weak reference to an ActorContainerRef, and so if we have two separate ActorContainerRef's one of them will not be tracked.

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -1901,6 +1901,18 @@ public:
             entry.value = onActorBroken(actorRef->onBroken(), *actorContainer)
                 .eagerlyEvaluate([](kj::Exception&& e) { KJ_LOG(ERROR, e); });
 
+            KJ_IF_SOME(ref, actorContainer->getContainerRef()) {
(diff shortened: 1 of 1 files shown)


MellowYarker:
I suspect this may fix https://github.com/cloudflare/workerd/issues/1422, although I haven't been able to force a segfault in `server-test.c++` yet. At the very least, it's clear the code prior to this commit is dropping a live `ActorContainerRef`, and if the `ActorContainer` goes away before the untracked `ActorContainerRef`, then the untracked `ActorContainerRef` will attempt to access invalid memory in its dtor (which is what we see happening in that issue).

jasnell (approved):

PR #2012, opened 2024-04-12
[nfc] replace glob with explicit file list

mikea:
Also remove url.c++ from "jsg" library since it is part of "url". Glob before didn't make it obvious that url.c++ is included twice.


fhanau:
Looks like `src/workerd/jsg/jsg.h` expects to have url.h in the same target, causing tests to fail 🙄. Changing the include to `workerd/jsg/url.h` should be sufficient to fix that.
Change LGTM, I was actually working on something similar. There's more things going wrong here (e.g. `jsg-test.h` being part of the jsg target), but that can be addressed in another cleanup PR.

fhanau (approved):

PR #2023, opened 2024-04-15
update capnp

mikea:


WORKSPACE
```diff
@@ -39,10 +39,10 @@ apple_support_dependencies()
 
 http_archive(
     name = "capnp-cpp",
-    sha256 = "21ae72a199a995b72e9bd359d4815539158d93a15cf36e284ef201fde7338c3c",
-    strip_prefix = "capnproto-capnproto-8ba7a6e/c++",
+    integrity = "sha256-I/jF25bRniNfmXR4KdeXTnwetKPKEm6fkYfibpZFYoU=",
```
kentonv, line 42: How do I generate this new format?

The old format, I would just do: `curl -L $URL | sha256sum`

But off the top of my head I'm not sure how I'd generate this base64.

Can we maybe just stick with the old format? Or is there some meaningful advantage to this change?

mikea: > How do I generate this new format?

you just run bazel without any integrity attribute and it tells you what to add:

```
INFO: Invocation ID: a7c49738-0fe1-47d0-b3aa-2ea74014898c
DEBUG: Rule 'capnp-cpp' indicated that a canonical reproducible form can be obtained by modifying arguments integrity = "sha256-I/jF25bRniNfmXR4KdeXTnwetKPKEm6fkYfibpZFYoU="
DEBUG: Repository capnp-cpp instantiated at:
  /home/maizatskyi/src/github.com/cloudflare/workerd/WORKSPACE:40:13: in <toplevel>
```

> The old format, I would just do: curl -L $URL | sha256sum

I have never actually downloaded any archive and was always relying on bazel to tell me :)


> Can we maybe just stick with the old format?

I personally don't care. Up to you.

>  Or is there some meaningful advantage to this change?

This is a "standard" thing: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

jasnell (approved):

PR #2024, opened 2024-04-15
switch to new package index, decompress into .tar in javascript

garrettgu10:
how to test: `bazel build //... && time ./bazel-bin/src/workerd/server/workerd test samples/pyodide-langchain/config.capnp | ts`
Without this change: 27s loading packages, 155s CPU time, 63s total wall time
With this change: 1s loading packages, 100s CPU time, 40s total wall time


src/pyodide/internal/builtin_wrappers.js
```diff
@@ -147,7 +147,10 @@ export function patchFetch(origin) {
 
     // we didn't find it in the disk cache, continue with original fetch
     const response = await origFetch(url, options);
-    const arrayBuffer = await response.arrayBuffer();
+
+    const arrayBuffer = await new Response(response.body.pipeThrough(new DecompressionStream("gzip"))).arrayBuffer();
+
+    console.log("decompressed", fileName, arrayBuffer.byteLength, "bytes");
```
jasnell, line 153: Is the `console.log(...)` intentional here?

garrettgu10: Yeah, it's intentional, this whole section is not the final design and is just meant for testing for now.

src/pyodide/internal/builtin_wrappers.js
```diff
@@ -147,7 +147,10 @@ export function patchFetch(origin) {
 
     // we didn't find it in the disk cache, continue with original fetch
     const response = await origFetch(url, options);
-    const arrayBuffer = await response.arrayBuffer();
+
+    const arrayBuffer = await new Response(response.body.pipeThrough(new DecompressionStream("gzip"))).arrayBuffer();
+
```
jasnell, line 152: hmm... I always get nervous when I see decompression immediately going into an in-memory structure, especially given the potential number of copy operations involved in constructing the `arrayBuffer()`. Is there any chance that `DiskCache.put(...)` could just receive a teed `ReadbleStream` branch rather than an ArrayBuffer here? .. something like:

```
const [b1, b2] = response.body.pipeThrough(new DecompressionStream('gzip')).tee();
DiskCache.put(fileName, b1);
return new Response(b2);
```

This would likely be a lot faster and use far less memory.

garrettgu10: Sounds good! We'll also have to make a call on whether to keep the tars stored on disk compressed or decompressed. 

It's also notable that this code only executes in local dev mode, so memory usage is less of a priority.

src/pyodide/internal/loadPackage.js
```diff
@@ -0,0 +1,87 @@
...
+      return 0;
+    }
+    let toCopy = buf.length;
+    if (size - offset < toCopy) {
+      toCopy = size - offset;
+    }
+    buf.set(new Uint8Array(this.arrayBuffer, offset, toCopy));
+    return toCopy;
+  }
+}
```
dom96, line 58: Should tar.js instead be changed to expect an ArrayBuffer?

src/pyodide/internal/loadPackage.js
```diff
@@ -0,0 +1,87 @@
...
+
+  let loadPromises = [];
+  let loading = [];
+  for (const req of requirements) {
+    if (SITE_PACKAGES.loadedRequirements.has(req)) continue;
+    loadPromises.push(loadBundle(req));
+    loading.push(req);
+  }
+
+  console.log("Loading " + loading.join(", "));
```
dom96, line 71: Should these console.logs be removed?

src/pyodide/internal/loadPackage.js
```diff
@@ -0,0 +1,87 @@
...
+
+  await Promise.all(loadPromises).then((buffers) => {
```
dom96, line 73: Nit

```suggestion
  const buffers = await Promise.all(loadPromises); 
```

src/pyodide/internal/python.js
```diff
@@ -194,6 +195,7 @@ async function prepareWasmLinearMemory(Module) {
 }
 
 export async function loadPyodide(lockfile, indexURL) {
+  console.log("loading pyodide");
```
dom96, line 198:
```suggestion

```

src/pyodide/internal/loadPackage.js
```diff
@@ -0,0 +1,87 @@
...
+    for (const [requirement, buffer] of buffers) {
+      const reader = new ArrayBufferReader(buffer);
+      const [tarInfo, soFiles] = parseTarInfo(reader);
+      SITE_PACKAGES.addSmallBundle(tarInfo, soFiles, requirement);
+    }
+  });
+
+  console.log("Loaded " + loading.join(", "));
```
dom96, line 81:
```suggestion
```

src/pyodide/internal/loadPackage.js
dom96, line 71:
```suggestion
```

dom96 (approved):
See my comments, also can you describe how you've tested this in the PR description for future reference?

Some changes needed but accepting to unblock.

src/pyodide/internal/setupPackages.js
```diff
@@ -23,6 +22,86 @@ const STDLIB_PACKAGES = Object.values(LOCKFILE.packages)
...
+  }
+
+  /**
+   * A big bundle contains multiple packages, each package contained in a folder whose name is the canonicalized package name.
+   * This function overlays the requested packages onto the site-packages directory.
+   * @param {TarInfo} tarInfo The root tarInfo for the big bundle (See tar.js)
+   * @param {List<String>} soFiles A list of .so files contained in the big bundle
+   * @param {List<String>} requirements canonicalized list of packages to pick from the big bundle
+   */
+  addBigBundle(tarInfo, soFiles, requirements) {
```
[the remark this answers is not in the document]
garrettgu10: The structure of a big bundle is different from a small bundle. A big bundle contains one folder for each package, with each folder containing the structure of a wheel. A small bundle contains just the wheel for a single package, not contained within a folder.

src/pyodide/internal/loadPackage.js
garrettgu10: hmm, then we would have to expose an ArrayBuffer instead of a PackagesTarReader, not sure if we want that big of a refactor. This interface is meant to match exactly what PackagesTarReader does in JSG.

src/pyodide/internal/loadPackage.js
garrettgu10: I kept them there to match the original prints from the pyodide package loader. I'd prefer to keep them for now so we have more visibility on performance in testing.

PR #2034, opened 2024-04-18
Remove unique symbol from RPC types

### LINKED ISSUE #1996: Durable Object RPC Type Errors
I ran into an issue with typescript in Durable Objects x RPC following these docs - https://developers.cloudflare.com/durable-objects/best-practices/create-durable-object-stubs-and-send-requests/#call-rpc-methods

Posted to discord a few days ago and sounds like I'm not the only one with this problem.

In the worker where I'm calling the RPC, I'm defining the env:

```
export interface Env {
// other vars
 MERCHANT_INSTANCE: DurableObjectNamespace<MerchantInstance>
}
```
typescript throws this error:

```
TS2344: Type MerchantInstance does not satisfy the constraint DurableObjectBranded
Property [__DURABLE_OBJECT_BRAND] is missing in type MerchantInstance but required in type DurableObjectBranded
```

I'm importing the types in the worker and using most recent versions of all packages in both the DO and separate worker where I'm calling the DO:

```
{
        "devDependencies": {
        "@cloudflare/workers-types": "^4.20240405.0",
        "typescript": "^5.0.4",
        "wrangler": "^3.41.0"
    }
}
```

DO class is defined as:

```
import { DurableObject } from "cloudflare:workers";

export interface Env {
}


export class MerchantInstance extends DurableObject {
    constructor(state: DurableObjectState, env: Env) {
        super(state, env)
    }

    async increment(amount = 1) {
        let value: number = (await this.ctx.storage.get("value")) || 0;
        value += amount;
        await this.ctx.storage.put("value", value);
        return value;
    }
}
```

in addition to the typescript error in the Env type definition, I'm also not getting type completion on the 'increment method'.  When calling the .increment() method on a stub, I get the TS error:

`TS2339: Property increment does not exist on type DurableObjectStub<MerchantInstance>`

penalosa:
Fixes #1996

Previously, the RPC types used `unique symbol` to enforce nominal typing rather than structural typing for Durable Object and Service binding entrypoints. While this works, it's slightly _too_ strict. It meant that different versions of `DurableObject` and `WorkerEntrypoint` were not compatible (e.g. across workers-types entrypoints, and across ambient and exported types. This PR relaxes that constraint to use a well-known string instead, which will preserve nominal typing unless someone goes out of their way to add a specific property.

cc @Skye-31

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -6,10 +6,10 @@ declare namespace Rpc {
   // TypeScript uses *structural* typing meaning anything with the same shape as type `T` is a `T`.
   // For the classes exported by `cloudflare:workers` we want *nominal* typing (i.e. we only want to
   // accept `WorkerEntrypoint` from `cloudflare:workers`, not any other class with the same shape)
-  export const __RPC_STUB_BRAND: unique symbol;
-  export const __RPC_TARGET_BRAND: unique symbol;
-  export const __WORKER_ENTRYPOINT_BRAND: unique symbol;
-  export const __DURABLE_OBJECT_BRAND: unique symbol;
+  export const __RPC_STUB_BRAND: "__RPC_STUB_BRAND";
+  export const __RPC_TARGET_BRAND: "__RPC_TARGET_BRAND";
+  export const __WORKER_ENTRYPOINT_BRAND: "__WORKER_ENTRYPOINT_BRAND";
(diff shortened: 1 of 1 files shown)


penalosa:
@mrbbot you may have some thoughts here...

CarmenPopoviciu (approved):

Skye-31:
This change does fix the issue I was seeing here!

PR #2035, opened 2024-04-18
Improve new IoOwn logging

jasnell:
Avoid allocating the new string on every dereference. The previous iteration was done pretty quickly just to get something we can test with. I imagine we'll keep it so let's make it more performant.


kentonv:
I think allocating a string on every IoOwn is too heavy. And allocating one on every deference is _really_ too heavy.

A much better approach would be to pass `const std::type_info&` as a parameter to `DeleteQueue::checkFarGet()`, then only turn it into a string if and when an error message needs to be generated.

kentonv (approved):

PR #2040, opened 2024-04-20
Fix returning interface types from rpc methods

### LINKED ISSUE #2003: rpc can only return `type` not `interface`/`class`
[Repro](https://github.com/helloimalastair/rpc-collapse)

When calling a DO RPC method, return type collapses to `Promise<undefined>`/`never` if using an `interface`. When using a `type`, return values are as expected.

alpertuna:
Fixes #2003

`interface` does not extend `Serializable` but `type` does.
```typescript
interface R1 {
  field: string;
}
type R2 = {
  field: string;
}
```
```R1 extends Serializable``` fails.
```R2 extends Serializable``` ok.

After the patch both `interface` and `type` match `Serializable`.

This fix is a starting point for the issue, there may be a better solution.

### DIFF
--- a/types/test/types/rpc.ts
+++ b/types/test/types/rpc.ts
@@ -10,7 +10,40 @@ import {
 } from "cloudflare:workers";
 import { expectTypeOf } from "expect-type";
 
-class BoringClass {}
+type TestType = {
+  fieldString: string;
+  fieldCallback: (p: string) => number;
+  fieldBasicMap: Map<string, number>;
+  fieldComplexMap: Map<string, {
+    fieldString: string;
+    fieldCallback: (p: string) => number;
+  }>;
+  fieldSet: Set<string>;
+  fieldSubLevel: {
+    fieldString: string;
+    fieldCallback: (p: string) => number;
+  };
+};
+
+interface ABasicInterface {
+  fieldString: string;
+  fieldCallback: (p: string) => number;
+};
+
+interface TestInterface extends ABasicInterface {
+  fieldBasicMap: Map<string, number>;
+  fieldComplexMap: Map<string, ABasicInterface>;
+  fieldSet: Set<string>;
+  fieldSubLevelInline: {
+    fieldString: string;
+    fieldCallback: (p: string) => number;
+  };
+  fieldSubLevelInterface: ABasicInterface;
+};
+
+interface NonSerializableInterface {
+  field: ReadableStream<string>;
+};
 
 class TestCounter extends RpcTarget {
   constructor(private val = 0) {
@@ -181,14 +214,52 @@ class TestEntrypoint extends WorkerEntrypoint<Env> {
       Object: { a: { b: new TestCounter() } },
     };
   }
-  get nonSerializable1() {
-    return new BoringClass();
+
+  methodReturnsTypeObject(): TestType {
+    return {
+      fieldString: "a",
+      fieldCallback: (p: string) => 1,
+      fieldBasicMap: new Map([["b", 2]]),
+      fieldComplexMap: new Map([["c", {
+        fieldString: "d",
+        fieldCallback: (p: string) => 3,
+      }]]),
+      fieldSet: new Set(["e"]),
+      fieldSubLevel: {
+        fieldString: "f",
+        fieldCallback: (p: string) => 4,
+      },
+    };
+  }
+  methodReturnsInterfaceObject(): TestInterface {
+    return {
+      fieldString: "a",
+      fieldCallback: (p: string) => 1,
+      fieldBasicMap: new Map([["b", 2]]),
+      fieldComplexMap: new Map([["c", {
+        fieldString: "d",
+        fieldCallback: (p: string) => 3,
+      }]]),
+      fieldSet: new Set(["e"]),
+      fieldSubLevelInline: {
+        fieldString: "f",
+        fieldCallback: (p: string) => 4,
+      },
+      fieldSubLevelInterface: {
+        fieldString: "e",
+        fieldCallback: (p: string) => 5,
+      },
+    };
+  }
+
+  nonSerializable1() {
+    return new ReadableStream<string>();
   }
   nonSerializable2() {
-    return { a: new BoringClass() };
+    return { field: new ReadableStream<string>() };
   }
-  async nonSerializable3() {
-    return new ReadableStream<string>();
+  nonSerializable3(): NonSerializableInterface {
+    return { field: new ReadableStream<string>() };
   }
 
   [Symbol.dispose]() {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by alpertuna** [comment]

**Comment by alpertuna** [comment]
Better fix may be done like this;

Here is `Serializable` type
https://github.com/cloudflare/workerd/blob/4098ac7c038d5097c470b3b5413b598b52730b98/types/defines/rpc.d.ts#L30-L58

This line does not match interfaces
```
    | { [key: string | number]: Serializable }
```

To make it work, `Serializable` should take the interface as a generic type like this;
```
type Serializable<T> =
...
    | { [K in keyof T]: Serializable<T[K]> }
...
```

But this solution make things more complicated.

**Comment by kentonv** [comment]
The trouble is, we don't want user-defined classes to be considered serializable _unless_ they extend `RpcTarget`.

Does this change keep that property, or will it make all user-defined classes be considered serializable?

**Comment by alpertuna** [comment]
@kentonv  I see the trouble, but the point is if developers believe that if the value is serializable and the type check fails, they will use some workarounds or will manuplate types, like here;
```typescript
class SomeClass {
  member: number = 0;
}

type WorkAround = {
  member: number
};

class SomeDurableObject {
  async someRpc(): Promise<WorkAround> {
    return new SomeClass();
  }
}

// This will pass the check
const res = await someDoStub.someRpc();
console.log(res.member);
```
There is no way to avoid the example above.
So the real responsible part is the business logic that serializes.

I think that if the value is serializable, types should let it without workarounds that look unnecessary.

I also improved the fix and added a new constraint; object member cannot be a function.

Here is an example that shows the before and after the fix;

```typescript
type RpcReturnType1 = {
  a: number;
  sub: {
    b: number;
  },
}
// Before fix: Ok
// After fix: Ok

interface SomeSubInterface {
  b: number;
}
interface RpcReturnType2 {
  a: number;
  sub: SomeSubInterface;
}
// Before fix: Fails
// After fix: Ok

type RpcReturnType3 = {
  cb: () => number
}
// Before fix: Ok (functions are considered objects, that's why they pass the check)
// After fix: Fails (added new constraint in the check)
```

Actually, it looks like returning callbacks are ok as I tested. (I didn't know rpc was capable of this)
As you said `We don't want user-defined classes` I just added the constraint, so developers use  `RpcTarget` for advanced usage. But I'm not sure if this constraint should be there.

By the way, my work here is to speed up things, and it is ok if you don't like it as I understand your concerns : )

alpertuna:
I read the documentation about RpcTarget, and my fix conflicted with the documentation since using callback as rpc return type is valid usage.
I just reverted the function check in case of that you may use this pull request.

Now all the fix does is let class instances and interfaces as return types of an rpc method.

kentonv:
Can someone on the team who knows typescript better than me review this? @petebacondarwin @penalosa ?

RamIdeas:
LGTM

Just needed to use the generic within the recursive parts of the type too

RamIdeas (approved):

RamIdeas:
The tests are failing because of the requirement Kenton metioned above: "we don't want user-defined classes to be considered serializable _unless_ they extend RpcTarget"

I'll take another look at this now

alpertuna:
> I think the way forward here is to allow classes to be serializable but make clear to users that the serialized form is a plain-object, not the original class instance

Since the returned object is stubified, return types of all methods of the object will be Promise, and that will make a visible difference that it is not the original class instance. (Unless all methods of the original class are already returning Promise.)

kentonv:
Regardless of what the type system says, attempting to serialize an instance of a class (that does not extend `RpcTarget`) will fail at runtime. This is an intentional design choice which we won't be changing.

If there's no way for TypeScript to enforce the same at type-checking time, then it seems we should make TypeScript more lenient, and just live with the fact that this problem won't be diagnosed until runtime.

alpertuna:
If class instances cause failure at runtime, I respect and agree that types should not let this usage.

Btw I missed fixing other `Structured cloneable composites` types, that's my fault and a proper fix should be like this;
```typescript
type Serializable<T> =
...
  | Map<
      Serializable<T extends Map<infer U, unknown> ? U : never>,
      Serializable<T extends Map<unknown, infer U> ? U : never>
    >
  | Set<Serializable<T extends Set<infer U> ? U : never>>
  | ReadonlyArray<Serializable<T extends ReadonlyArray<infer U> ? U : never>>
...
```

Now my main concern is `Serializable` type is getting spagettier and may be harder to maintain future improvements, so I'm not confident with this fix.

RamIdeas:
> Regardless of what the type system says, attempting to serialize an instance of a class (that does not extend RpcTarget) will fail at runtime. This is an intentional design choice which we won't be changing.

To clarify, I did mean to allow it in the typings, not to change the runtime behaviour to allow it.

I admit though I made an assumption that, since plain objects are allowed, classes would work in some form at runtime as if they were plain objects – eg. instance properties and methods work but prototype methods would not. But I see [from the docs](https://developers.cloudflare.com/workers/runtime-apis/rpc/#:~:text=This%20behavior%20is%20rarely%20useful%20in%20practice%2C%20and%20could%20be%20confusing%20if%20the%20developer%20had%20intended%20the%20class%20to%20be%20treated%20as%20an%20RpcTarget.%20So%2C%20Workers%20RPC%20has%20chosen%20to%20disallow%20classes%20that%20are%20not%20RpcTargets%2C%20to%20avoid%20any%20confusion) that we have decided that behaviour is "rarely useful in practice" and to avoid the footgun.

> If there's no way for TypeScript to enforce the same at type-checking time, then it seems we should make TypeScript more lenient, and just live with the fact that this problem won't be diagnosed until runtime.

Essentially, what I was saying. And if we deem it useful to catch these anti-patterns at dev-time, we can implement a lint rule.

jfsiii:
Linking [this comment](https://github.com/cloudflare/workerd/issues/2003#issuecomment-2267516475) for visibility

kentonv:
Sorry I lost track of this.

Honestly I don't understand how this works. It might be nice to add a comment, unless this should just be obvious to typescript wizards.

Otherwise @penalosa if you could get the test passing (looks like CI doesn't like it right now) and click your approval then I'll be happy to merge.

alpertuna:
Some little further fixes need to be done before merging. Let me do it.

alpertuna:
My 2 main fixes are;

- Fixed `Serializable` to properly resolve structured composite types.
```diff
-    | Map<Serializable<T>, Serializable<T>>
-    | Set<Serializable<T>>
-    | ReadonlyArray<Serializable<T>>
+    | Map<
+        T extends Map<infer U, unknown> ? Serializable<U> : never,
+        T extends Map<unknown, infer U> ? Serializable<U> : never
+      >
+    | Set<T extends Set<infer U> ? Serializable<U> : never>
+    | ReadonlyArray<T extends ReadonlyArray<infer U> ? Serializable<U> : never>
```

- I realized that when using interfaces, callbacks were not being stubified (would not return a promise). I fixed this in the `Stubify<T>` type with a small change:
```diff
- : T extends { [key: string | number]: unknown } ? { [K in keyof T]: Stubify<T[K]> }
+ : T extends { [key: string | number]: any } ? { [K in keyof T]: Stubify<T[K]> }
```

I also added a comment at the top of the `Serializable` type.

See the commit: https://github.com/cloudflare/workerd/pull/2040/commits/09429df522cb1d43c5ead365035b959e9d035c6b

types/test/types/rpc.ts
```diff
@@ -440,8 +440,14 @@ export default <ExportedHandler<Env>>{
...
 
       expectTypeOf(s.everySerializable).not.toBeNever();
-      expectTypeOf(s.nonSerializable1).toBeNever();
-      expectTypeOf(s.nonSerializable2).returns.toBeNever();
+
+      // These are _not_ serializable, but will nevertheless not be typed as `never`. This
+      // is because TS can't distinguish between a class instance and an interface. At runtime,
+      // using an RPC method like this will throw:
+      //   "Could not serialize object. This type does not support serialization.""
+      expectTypeOf(s.nonSerializable1).not.toBeNever();
```
kentonv, line 448: These new lines of test aren't actually testing what this PR is meant to fix, right? Could we add such a test?

alpertuna: I agree, the variable names in the current test are also misleading. Types defined with "type" and "interface" should be tested separately. Additionally, stubified callbacks for interfaces should be tested—I realized by chance that they weren't stubified before the fix.
I'll update the tests accordingly.

alpertuna: Just added new tests for Serializable objects, covering "type" and "interface" definitions individually. Including multi-level objects, objects with stubable callback members, and objects with non-serializable members.
Removed redundant old tests.
Commit: https://github.com/cloudflare/workerd/pull/2040/commits/25cfd1a4ebe1fd33292b0b35fd34ec5215982147

kentonv (approved):

kentonv:
Thanks so much @alpertuna!

This is looking pretty good to me but I'd appreciate if @penalosa could sign off before we merge.

kentonv:
Doh, there's now a merge conflict...

alpertuna:
> @alpertuna ... rather than using a merge commit here, can I ask you to rebase the PR on `main`?

Sorry about that. Everything should be fine now.

penalosa (approved):

PR #2044, opened 2024-04-22
Deprecate getLogs method from Ai binding

G4brym:
This was a testing method added to debug the Ai binding, and is now being deprecated.
It no longer returns any data from the server

### DIFF
--- a/src/cloudflare/internal/ai-api.ts
+++ b/src/cloudflare/internal/ai-api.ts
@@ -31,7 +31,6 @@ export class Ai {
   private readonly fetcher: Fetcher
 
   private options: AiOptions = {};
-  private logs: Array<string> = [];
   public lastRequestId: string | null = null;
 
   public constructor(fetcher: Fetcher) {
@@ -80,22 +79,8 @@ export class Ai {
       }
 
       return res.body;
+
     } else {
-      // load logs
-      if (this.options.debug) {
-        let parsedLogs: string[] = [];
-        try {
-          const logHeader = res.headers.get("cf-ai-logs")
-          if (logHeader) {
-            parsedLogs = (JSON.parse(atob(logHeader)) as string[]);
-          }
-        } catch {
-          /* empty */
-        }
-
-        this.logs = parsedLogs;
-      }
-
       if (!res.ok || !res.body) {
         throw new InferenceUpstreamError(await res.text());
       }
@@ -110,9 +95,12 @@ export class Ai {
     }
   }
 
-  public getLogs(): Array<string> {
-    return this.logs;
-  }
+    /*
+     * @deprecated this method is deprecated, do not use this
+     */
+    public getLogs(): Array<string> {
+      return []
+    }
 }
 
 export default function makeBinding(env: { fetcher: Fetcher }): Ai {
--- a/src/cloudflare/internal/test/ai/ai-api-test.js
+++ b/src/cloudflare/internal/test/ai/ai-api-test.js
@@ -11,9 +11,6 @@ export const tests = {
             const resp = await env.ai.run('testModel', {prompt: 'test'})
             assert.deepStrictEqual(resp, { response: 'model response' });
 
-            // Test logs is empty
-            assert.deepStrictEqual(env.ai.getLogs(), []);
-
             // Test request id is present
             assert.deepStrictEqual(env.ai.lastRequestId, '3a1983d7-1ddd-453a-ab75-c4358c91b582');
         }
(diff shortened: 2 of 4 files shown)


src/cloudflare/internal/ai-api.ts
```diff
@@ -109,10 +94,6 @@ export class Ai {
       return res.body;
     }
   }
-
-  public getLogs(): Array<string> {
```
Cherry, line 113: Is this safe to remove without causing userland runtime issues? Although not officially documented, it's been listed and available in the types via `@cloudflare/ai` for a while: 
image

G4brym: From my point of view, end users don't get any benefit from what's being returned here
And this method is going to stop returning results from server side soon, so the client method will have nothing to show

Cherry: Agreed, but returning an empty array likely won't break applications, whereas a method being removed that was previously there will cause a `TypeError` and potentially break userland code if called and expected to be present.

Cherry: >  so provably very few users have seen this method

Are there any stats for this? This is a breaking change and should probably be behind a compat flag imo. Workers AI is GA.

celso: I can't think of any reason why userland code outside our teams would use the undocumented getLogs() method. Even if they used it, it wouldn't return any data. I understand that the types declarations exposed it, and I'd rather they didn't, but really I think the risk is non-existent.

zegevlier: > Also the new Ai binding is like 2 weeks old, so provably very few users have seen this method

I don't think that's very relevant here either. The blog post from when compatibility dates were introduced worded it well in my opinion:

> This means that if a runtime upgrade breaks someone's application, it's really bad. The developer didn't make any change, so won't be watching for problems. They may be asleep, or on vacation. If we want people to trust serverless, we can't let this happen.

This change is a breaking change if you're using the method, which was in the public types. If _anyone_ is using this, I would expect Cloudflare to put the change behind a compatibility date.

CyebSloth: Hey @G4brym ,

I think I have approximately ~6 Workers that use workers-ai, some of them use `getLogs` both to debug/log and to get RAW outputs from the model (or part of them) since that information is not exposed anywhere and post-processing isn't always wanted, it's a useful method albeit hacky, but it is what customers have available now.

Some of these workers are going to hit prod in 2-3 days and having this pushed potentially risks breaking the product after going live, deployments breaking without previous notice happened already in January and we were "okay" with it due to the BETA status of the product (+ usage happened in the background, no user was directly affected by the APIs breaking).

When a managed product like workers ai goes GA, I expect to be able to deploy models and be done with it for as long as possible, I shouldn't be on the look for changes or be surprised with notifications that ask me whether X PR will break our workers; The potential for our API to go down in production due to these types of changes is a serious concern.

In our four years of operation (with AI services), we've never experienced a service provider making a breaking change like this without providing advance notice of the upcoming changes (+ multiple reminders).

For now, we will be removing our `getLogs` usage.

G4brym: @CyebSloth what Information are you looking to get from the getLogs
Maybe we can workout an official solution

CyebSloth: Hey @G4brym ,
> what Information are you looking to get from the getLogs

The result before any `postProcessing` is executed, the raw output can be used as input for other models or could be a requirement when the customer wants to apply a different post-processing technique. 

On an ideal scenario, we'd be able to "hook" intermediate activations and gradients, potentially allowing customers to use existing models for more advanced purposes (feature extraction with resnet and interpretation with llama + loras, for example).

Thank you for the follow-up!

G4brym: I will forward your feature suggestion internally 👍

G4brym: In another note, I've come to realize this getLogs method actually never worked when using the new binding, because the run method was throwing an error when parsing the logs from the server.
So for anyone to be using this method it's guaranteed that you are using the old npm package.
So in reality this is not a breaking change, because it never worked on the new binding.
--
With this in mind, in my opinion a good compromise resolution for this is to:
Merge this PR that removes this method as it never returned anything besides an empty array, and keep supporting the getLogs from the npm package for the foreseeable future 
Please fell free to comment, I'm open for suggestions

Cherry: It doesn't break anything though. Code like this works in production today, using the new binding:
```ts
export default {
	async fetch(request: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
		const messages = [
			{ role: "system", content: "You are a friendly assistant" },
			{
				role: "user",
				content: "What is the origin of the phrase Hello, World",
			},
		];
		const response = await env.AI.run("@cf/meta/llama-2-7b-chat-fp16", { messages, debug: true });
		const logs = env.AI.getLogs();

		return Response.json({response, logs})
	},
};
```
Even with `logs` being empty, it's not breaking anything, after migrating from the old `@cloudflare/ai` package. With the change in this PR, that code won't work anymore with no change of my own, as `getLogs` will now throw an error in my code.

G4brym: just updated the pr to deprecate the getLogs, instead of removing it completely 👍

fhanau (approved):

PR #2045, opened 2024-04-22
Have internal streams return empty Uint8Array on end of byob stream

jasnell:
Per the spec, a BYOB read is supposed to return an empty ArrayBufferView when the stream is closed (done = true) in order to allow the code reading to retake ownership of the underlying buffer. This change makes the internal streams implementation return an empty Uint8Array in this case, matching the behavior of the JS-backed standard streams impl.

~~It's unlikely but this may be considered a breaking change if anyone has code that depends on the fact that `value` would be `undefined` if `done === true` when using the old streams implementation. We need to decide if this case needs a compat flag or not. I'd prefer not to have to introduce Yet Another One.~~ Updated to add the compat flag :-(

### DIFF
--- a/src/workerd/api/streams/internal.c++
+++ b/src/workerd/api/streams/internal.c++
@@ -9,6 +9,7 @@
 #include <kj/vector.h>
 #include <workerd/api/util.h>
 #include <workerd/util/string-buffer.h>
+#include <workerd/io/features.h>
 
 namespace workerd::api {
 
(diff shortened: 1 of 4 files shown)


kentonv:
Unfortunately, I could pretty easily imagine someone checking the truthiness of `value` instead of checking for `done`.

jasnell:
Hmm... all of the documentated examples pretty much everywhere show checking the `done = true` as the correct way to determine completion, especially given that in a non-BYOB read, `value: undefined` (or `value: false`, etc) is a perfectly valid read result while the stream is still active. *Unfortunately*, however, we don't really have any way of know if someone is checking the truthiness of `value` instead of doing it the right way. :-( ... So yeah... sadly this will probably require a compat flag and that makes me very sad.

kentonv:
I feel like I personally have written code which checked the truthiness, and it worked. Some time later I think I did realize it should check `done` instead. But it would have been easy for me to not have noticed the problem -- so I'm sure someone out there has gotten it "wrong". ("Wrong" in quotes because Hyrum's Law says it's not wrong.)

kentonv (approved):

PR #2046, opened 2024-04-22
Hold strong refs in certain streams Writer/Reader methods

jasnell:
In some edge cases, if the Writer/Reader is the only thing remaining holding a strong reference to a WritableStream or ReadableStream, the calls to `close()`, `abort()`, `cancel()`, etc can cause the Writer/Reader to drop it's strong reference in mid operation, leading to issues. Let's defensively hold an additional strong reference while the operations are pending


jasnell:
Failure on internal CI needs to be investigated...

harrishancock (approved):

PR #2048, opened 2024-04-23
Cleanup deprecated jsg/promise methods

jasnell:
Remove last few uses of the deprecated jsg/promise method variants that do not take a jsg::Lock as the first param.

Just some drive by general cleanup while investigating some other issues.


fhanau (approved):
LGTM

PR #2054, opened 2024-04-24
Add SqliteDatabase::executeWithRegulator

justin-mp:
There are `sqlite3` API functions that may prepare and execute their own internal statements.  We need to call those functions with a regulator or we’ll summarily reject them.  Adding a method to do this seemed like a clean way of handling these cases.

### DIFF
--- a/src/workerd/util/sqlite.h
+++ b/src/workerd/util/sqlite.h
@@ -94,6 +94,9 @@ public:
   // that was not processed. This is used for streaming SQL ingestion.
   kj::StringPtr ingestSql(Regulator& regulator, kj::StringPtr sqlCode);
 
+  // Execute a function with the given regulator.
+  void executeWithRegulator(Regulator& regulator, kj::FunctionParam<void()> func);
+
 private:
   sqlite3* db;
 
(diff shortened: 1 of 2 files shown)


justin-mp:
https://github.com/cloudflare/workerd/compare/1942089694234ce7772b84225b714dc7df3c1a09..5ff0c599283a7aeb93b5f4a7f4c6c019c83f7fa5 is a rebase on master.

src/workerd/util/sqlite.h
```diff
@@ -94,6 +94,9 @@ class SqliteDatabase {
   // that was not processed. This is used for streaming SQL ingestion.
   kj::StringPtr ingestSql(Regulator& regulator, kj::StringPtr sqlCode);
 
+  // Execute a function with the given regulator.
+  void executeWithRegulator(Regulator& regulator, kj::FunctionParam<void()> func);
```
smerritt, line 98: This looks good, but I've got one brain cell that won't quit screaming about something like

```
template <typename Ret, typename... Params>
Ret executeWithRegulator(Regulator& regulator, kj::FunctionParam<Ret(Params...)> func, Params... params) {
// ...
return func(kj::fwd<Params>(params)...);
```

Let's ignore that brain cell, though, since we can go back and generalize this later if we need to. YAGNI, blah blah blah.

smerritt (approved):

kentonv (approved):

PR #2057, opened 2024-04-24
add duration logging to flushImpl and to waiting for prev flushes

Frederik-Baetens:
This attempts to differentiate single flushes taking a long time vs the effect of multiple queued flushes causing issues.

### DIFF
--- a/src/workerd/io/actor-cache.c++
+++ b/src/workerd/io/actor-cache.c++
@@ -2155,10 +2155,17 @@ void ActorCache::ensureFlushScheduled(const WriteOptions& options) {
 
   if (!flushScheduled) {
     flushScheduled = true;
-    auto flushPromise = lastFlush.addBranch().attach(kj::defer([this]() {
-      flushScheduled = false;
-      flushScheduledWithOutputGate = false;
-    })).then([this]() {
+    auto flushPromise = lastFlush.addBranch().attach(
+      util::DurationExceededLogger(
+        clock,
+        5 * kj::SECONDS,
+        kj::str("waiting on previous flushes took longer than expected", flushesEnqueued)
+      ),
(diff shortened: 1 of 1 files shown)


src/workerd/io/actor-cache.c++
```diff
@@ -2155,10 +2155,17 @@ void ActorCache::ensureFlushScheduled(const WriteOptions& options) {
 
   if (!flushScheduled) {
     flushScheduled = true;
-    auto flushPromise = lastFlush.addBranch().attach(kj::defer([this]() {
-      flushScheduled = false;
-      flushScheduledWithOutputGate = false;
-    })).then([this]() {
+    auto flushPromise = lastFlush.addBranch().attach(
+      util::DurationExceededLogger(
```
justin-mp, line 2159: I don't think this works.  `Own<T>` has an attachVal that would work like this, but kj::Promise doesn't have an attachVal.  Maybe you can just heap allocate it and then move it here.

src/workerd/io/actor-cache.c++
```diff
@@ -2492,6 +2499,11 @@ kj::Promise<void> ActorCache::startFlushTransaction() {
 }
 
 kj::Promise<void> ActorCache::flushImpl(uint retryCount) {
+  auto durationAlert = util::DurationExceededLogger(
+    clock,
+    5 * kj::SECONDS,
+    "flushImpl took longer than expected"
```
a-robinson, line 2505: This would be much more useful if it included some summary of the operations included in the flush. Or if it allowed us to distinguish between the actual flushing being slow and the `waitForPastReads` step being slow.

Frederik-Baetens:
Closing this too, seems like most slowness here is usually driven by cpu usage.

PR #2059, opened 2024-04-25
Return row counts for SQL ingestion (storage.sql.ingest()).

smerritt:
D1 bills based on rows read and written, but sql.ingest() didn't return those. Now it does.

This required changing the return type of sql.ingest(), but it's an experimental API so that should be okay.

  let sqlCode = "INSERT INTO tbl (col) VALUES (123); INSERT I";
  let result = sql.ingest(sqlCode);

Old:

  result == " INSERT I"

New:

  result.remainder == " INSERT I"
  result.meta.rowsRead == 0
  result.meta.rowsWritten == 1

This "meta" attribute also gives us a convenient spot to stick additional information. For example, if we wanted to count how many SQL statements ingest() executed, we could easily add "result.meta.statementsExecuted" without breaking any existing users.

Implementation-wise, there might be an optimization opportunity around SqlStorage::IngestResult::getMeta(). Right now, it allocates a new Meta every time it's called, so if you do the obvious thing like so:

  totalRead += result.meta.rowsRead
  totalWritten += result.meta.rowsWritten
  totalRedFish += result.meta.redFish  // and so on

then that might end up allocating and discarding multiple Meta objects, and some caching could fix that. On the other hand, I'm not very familiar[1] with JSG internals, so maybe that's already happening somehow. I have no idea.

[1] This is a terrible understatement.

### DIFF
--- a/src/workerd/api/sql.h
+++ b/src/workerd/api/sql.h
@@ -22,9 +22,10 @@ public:
 
   class Cursor;
   class Statement;
+  class IngestResult;
 
   jsg::Ref<Cursor> exec(jsg::Lock& js, kj::String query, jsg::Arguments<BindingValue> bindings);
-  kj::String ingest(jsg::Lock& js, kj::String query);
+  jsg::Ref<IngestResult> ingest(jsg::Lock& js, kj::String query);
 
   jsg::Ref<Statement> prepare(jsg::Lock& js, kj::String query);
 
@@ -249,10 +250,35 @@ private:
   friend class Cursor;
 };
 
+
+
+class SqlStorage::IngestResult final : public jsg::Object {
+public:
+
+  IngestResult(kj::String remainder, uint64_t rowsRead, uint64_t rowsWritten);
+
+  JSG_RESOURCE_TYPE(IngestResult) {
+    JSG_READONLY_PROTOTYPE_PROPERTY(rowsRead, getRowsRead);
+    JSG_READONLY_PROTOTYPE_PROPERTY(rowsWritten, getRowsWritten);
+    JSG_READONLY_PROTOTYPE_PROPERTY(remainder, getRemainder);
+  }
+
+  kj::StringPtr getRemainder();
+  double getRowsRead();
+  double getRowsWritten();
+
+private:
+  kj::String remainder;
+  uint64_t rowsRead;
+  uint64_t rowsWritten;
+};
+
+
 #define EW_SQL_ISOLATE_TYPES                    \
   api::SqlStorage,                              \
   api::SqlStorage::Statement,                   \
   api::SqlStorage::Cursor,                      \
+  api::SqlStorage::IngestResult,                \
   api::SqlStorage::Cursor::RowIterator,         \
   api::SqlStorage::Cursor::RowIterator::Next,   \
   api::SqlStorage::Cursor::RawIterator,         \
--- a/src/workerd/api/sql.h
+++ b/src/workerd/api/sql.h
@@ -255,15 +255,17 @@ private:
 class SqlStorage::IngestResult final : public jsg::Object {
 public:
 
-  IngestResult(kj::String remainder, uint64_t rowsRead, uint64_t rowsWritten);
+  IngestResult(kj::String remainder, uint64_t rowsRead, uint64_t rowsWritten, uint64_t statementCount);
 
   JSG_RESOURCE_TYPE(IngestResult) {
+    JSG_READONLY_PROTOTYPE_PROPERTY(statementCount, getStatementCount);
     JSG_READONLY_PROTOTYPE_PROPERTY(rowsRead, getRowsRead);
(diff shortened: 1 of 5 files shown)


geelen:
Thanks for looking at this @smerritt! I think this would actually work fine, but interested in following up on this idea (from internal chat):

> At the SQLite level, we've got a pair of DB-wide row counters (read and written), which would be really handy here.
workerd's API has `SqliteDatabase::Query()` resetting the row counters, though, which sort of makes sense and sort of doesn't. I'd almost prefer if the DB-wide counters just kept on going up† and `Query` computed deltas.
> 
> Part of me wants DB-wide non-resetting counters to make it easier to do stuff like this, and part of me thinks that it'd also make it easy to compute garbage values, like if you had a function that was synchronous but got an async step wedged in the middle later on, your before/after row counts could end up garbage. 

We already make some fairly strict assumptions around the SQL api being complete synchronous, in particular calling `sql.databaseSize` and executing `SELECT total_changes() as changes, last_insert_rowid() as last_row_id` before and after each D1 query (which could be a batch with multiple SQL queries inside).

I don't think there's any real worry in adding `sql.totalRowsRead` and `sql.totalRowsWritten` as ever-incrementing counters (accessible from JS), then making `Query` read the value before and after to produce the value for `getRowsRead()` and `getRowsWritten()`. Then APIs like `ingest` (and whatever a future DO + SQL user might need) can simply read the value before and after each logical operation and return it to the user in whatever form makes sense.

At the very least, this removes any concept of `meta` from the `.ingest` API and makes the change fully backwards-compatible.

kentonv:
I think we intentionally didn't add DB-wide counters because queries can be running concurrently (you can have multiple cursors) and so these counters felt like a footgun.

Regarding the interface change to `ingest()`, my nitpick is I would not create a `meta` property, just flatten the counters into `result` itself. This is closer to what `exec()` returns -- a cursor which has `rowsRead` and `rowsWritten` as top-level properties.

smerritt:
Flattening the counters into `result` sounds good to me. It's still the case that we can add more properties and methods later on, so I'm happy.

smerritt:
Fixup: f5838657

smerritt:
This is ready for review.

`ingest`'s return type is now flat: it's got three properties: `remainder`, `rowsRead`, and `rowsWritten`.

As far as the DB-wide counters, Kenton's right: they're a footgun. If you compute deltas around a chunk of purely synchronous code, you'll get correct numbers, but if any asynchrony creeps in between the queries across which you are computing deltas, your deltas can be too large. Summing row counts across queries is much less error-prone. I think it's better to take the interface change now, while it's still experimental, and just deal with a bit of jank in the D1 worker for a few weeks.

geelen:
@smerritt ok, I'm happy with this. I can tweak the d1-worker stuff as necessary. Can I ask to add one more counter: `numQueries`? It turns out that was the _one_ bit we used to show to users in Wrangler, that we've now lost due to no longer parsing the SQL locally:

image

src/workerd/api/sql.h
```diff
@@ -22,9 +22,10 @@ class SqlStorage final: public jsg::Object, private SqliteDatabase::Regulator {
 
   class Cursor;
   class Statement;
+  class IngestResult;
 
   jsg::Ref<Cursor> exec(jsg::Lock& js, kj::String query, jsg::Arguments<BindingValue> bindings);
-  kj::String ingest(jsg::Lock& js, kj::String query);
+  jsg::Ref<IngestResult> ingest(jsg::Lock& js, kj::String query);
```
jasnell, line 28: I know that `ingest` is still experimental so this should be fine, but I just want to confirm that we're sure changing the return value here won't break anyone that we know of?

src/workerd/api/sql.h
```diff
@@ -249,10 +250,38 @@ class SqlStorage::Statement final: public jsg::Object {
   friend class Cursor;
 };
 
+
+
+class SqlStorage::IngestResult final : public jsg::Object {
+public:
```
jasnell, line 256: Does this really need to be a `jsg::Object`? Could `IngestResult` just be a `JSG_STRUCT`?

smerritt: Sure! I only went with `jsg::Object` since that's what the other classes in this header file inherit from. I've never heard of JSG_STRUCT, but I can take a look at using it.

src/workerd/api/sql.h
smerritt: The only thing it'll break is the D1 worker, and I've already merged (and am working on deploying) code that handles both the old and the new type.

src/workerd/api/sql.h
jasnell: The key difference between the two is that when a `JSG_STRUCT` type is passed out to JavaScript, it comes in the form of an ordinary JavaScript object with properties as opposed to an instance of a class. There is also no persistent c++ object behind the object. The struct is thrown away as soon as the JavaScript object is created.

jasnell (approved):
Code changes LGTM

PR #2061, opened 2024-04-26
Avoid unnecessary allocations in HTMLRewriter

jasnell:
The original code was allocating an `IdentityTransformStream` and `WritableStream` just to end up throwing both away immediately after. This simplifies things by making it possible to create a the identity pipeline that underlies `IdentityTransformStream` directly, simplifying things a bit by avoiding the unnecessary additional allocations and removing an unnecessary use of the `removeSink()` API.

Internal CI is green.

### DIFF
--- a/src/workerd/api/html-rewriter.c++
+++ b/src/workerd/api/html-rewriter.c++
@@ -1063,11 +1063,10 @@ jsg::Ref<Response> HTMLRewriter::transform(jsg::Lock& js, jsg::Ref<Response> res
 
   auto& ioContext = IoContext::current();
 
-  // lol-html writes to a pipe, the other end of which is our transformed response body.
-  auto ts = IdentityTransformStream::constructor(js);
-  response = Response::constructor(js, kj::Maybe(ts->getReadable()), kj::mv(response));
-
-  auto outputSink = ts->getWritable()->removeSink(js);
+  auto pipe = newIdentityPipe();
+  response = Response::constructor(js,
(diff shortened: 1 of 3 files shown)


src/workerd/api/html-rewriter.c++
```diff
@@ -1063,11 +1063,10 @@ jsg::Ref<Response> HTMLRewriter::transform(jsg::Lock& js, jsg::Ref<Response> res
 
   auto& ioContext = IoContext::current();
 
-  // lol-html writes to a pipe, the other end of which is our transformed response body.
-  auto ts = IdentityTransformStream::constructor(js);
-  response = Response::constructor(js, kj::Maybe(ts->getReadable()), kj::mv(response));
-
-  auto outputSink = ts->getWritable()->removeSink(js);
```
jasnell, line 1070: For reviewers: If you don't have the context on this change, what is happening in the original code here is that we allocate an `IdentityTransformStream`, which creates an internal refcounted `IdentityTransformStreamImpl` instance, then uses that to create one `WritableStream` instance and one `ReadableStream` instance. We then immediately extract the reference to the `IdentityTransformStreamImpl` out of the `WritableStream`. In this case, creating both the `IdentityTransformStream` and the `WritableStream` are unnecessary since both are immediately thrown away.

In the new code, we create only the `IdentityTransformStreamImpl` and the `ReadableStream` that we are keeping.

fhanau (approved):
LGTM

PR #2064, opened 2024-04-26
Deprecate WritableStream::removeSink

jasnell:
removeSink is used to extract the WritableStreamSink that is owned by a WritableStream using the original implementation. It is currently used in only three places in the codebase, two of which are unnecessary. There are two pending changes refactoring those two places to avoid using `removeSink()` as currently defined. Replacing the version of `removeSink` that returns the `WritableStreamSink` with a `detach()` method that returns nothing covers the sockets use case where we call the method then immediately throw away the sink.

Opening this to broadcast the intent in case folks have reason to keep removeSink around.

Once this and the two other PRs land, we hopefully move towards removing removeSink entirely. Note that ultimately the goal is to entirely eliminate the need for (and remove entirely) the `WritableStreamInternalController` as a separate implementation, making it such that all `WritableStream` use cases can be served by the `WritableStreamJsController` implementation. This, however, raises an issue with the JSRPC implementation which uses `removeSink()` in the implementation of `WritableStream::serialize()`. JS-backed WritableStreams are currently not supported by JSRPC. We will need to address that limitation before being able to proceed here. FWIW, the use of `removeSink()` in JSRPC does not actually require the public method on `WritableStream` to work, so that's not a blocker.

Refs: https://github.com/cloudflare/workerd/pull/2061
Refs: https://github.com/cloudflare/workerd/pull/2050

### DIFF
--- a/src/workerd/api/sockets.c++
+++ b/src/workerd/api/sockets.c++
@@ -318,7 +318,7 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
   auto secureStreamPromise = context.awaitJs(js, writable->flush(js).then(js,
       [this, domain = kj::heapString(domain), tlsOptions = kj::mv(tlsOptions),
       tlsStarter = kj::mv(tlsStarter)](jsg::Lock& js) mutable {
-    writable->removeSink(js);
+    writable->detach(js);
     readable = readable->detach(js, true);
     closedResolver.resolve(js);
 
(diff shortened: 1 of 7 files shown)


kentonv:
> FWIW, the use of removeSink() in JSRPC does not actually require the public method on WritableStream to work, so that's not a blocker.

How so? It does call that method. With this PR, doesn't it now raise a warning?

jasnell:
The `WritableStream::Serialize` can use the internal version of `removeSink()` (which this PR deprecates currently deprecates but doesn't need to). It's really just the one on `WritableStream` that really ought to be gone.

dom96 (approved):

kentonv (approved):

PR #2066, opened 2024-04-26
Implement support for JS-backed WritableStreams in JSRPC

jasnell:
First pass at implementing support for JS-backed WritableStreams with JSRPC.

Provides an alternative adapter that will acquire the isolate lock on each write/close call, passing the operation off to the JS interface. This is obviously not as efficient as the original adapter used for `WritableStreamSink` but not yet sure if there's a better approach.

### DIFF
--- a/src/workerd/api/streams/writable.c++
+++ b/src/workerd/api/streams/writable.c++
@@ -369,6 +369,181 @@ private:
   }
 };
 
+// In order to support JavaScript-backed WritableStreams that do not have a backing
+// WritableStreamSink, we need an alternative version of the WritableStreamRpcAdapter
+// that will arrange to acquire the isolate lock when necessary to perform writes
+// directly on the WritableStreamController. Note that this approach is necessarily
+// a lot slower
+class WritableStreamJsRpcAdapter final: public capnp::ExplicitEndOutputStream {
+public:
+  WritableStreamJsRpcAdapter(IoContext& context, jsg::Ref<WritableStreamDefaultWriter> writer)
+      : context(context), writer(kj::mv(writer)) {}
+
+  ~WritableStreamJsRpcAdapter() noexcept(false) {
+    weakRef->invalidate();
+    doneFulfiller->fulfill();
+
+    // If the stream was not explicitly ended and the writer still exists at this point,
+    // then we should trigger calling the abort algorithm on the stream. Sadly, there's a
+    // bit of an incompetibility with kj::AsyncOutputStream and the standard definition of
+    // WritableStream in that AsyncOutputStream has no specific way to explicitly signal that
+    // the stream is being aborted due to a particular reason.
+    //
+    // On the remote side, because it is using a WritableStreamSink implementation, when that
+    // side is aborted, all it does is record the reason and drop the stream. It does not
+    // propagate the reason back to this side. So, we have to do the best we can here. Our
+    // assumption is that once the stream is dropped, if it has not been explicitly ended and
+    // the writer still exists, then the writer should be aborted. This is not perfect because
+    // we cannot propgate the actual reason why it was aborted.
+    //
+    // Note also that there is no guarantee that the abort will actually run if the context
+    // is being torn down. Some WritableStream implementations might use the abort algorithm
+    // to clean things up or perform logging in the case of an error. Care needs to be taken
+    // in this situation or the user code might end up with bugs. Need to see if there's a
+    // better solution.
+    //
+    // TODO(someday): If the remote end can be updated to propagate the abort, then we can
+    // hopefully improve the situation here.
+    if (!ended) {
+      KJ_IF_SOME(writer, this->writer) {
+        context.addTask(context.run(
+            [writer=kj::mv(writer), exception=cancellationException()]
+            (Worker::Lock& lock) mutable {
+          jsg::Lock& js = lock;
+          auto ex = js.exceptionToJs(kj::mv(exception));
+          return IoContext::current().awaitJs(lock,
+              writer->abort(lock, ex.getHandle(js)));
+        }));
+      }
+    }
+  }
+
+  // Returns a promise that resolves when the stream is dropped. If the promise is canceled before
+  // that, the stream is revoked.
+  kj::Promise<void> waitForCompletionOrRevoke() {
+    auto paf = kj::newPromiseAndFulfiller<void>();
+    doneFulfiller = kj::mv(paf.fulfiller);
+
+    return paf.promise.attach(kj::defer([weakRef = weakRef->addRef()]() mutable {
+      KJ_IF_SOME(obj, weakRef->tryGet()) {
+        // Stream is still alive, revoke it.
+        if (!obj.canceler.isEmpty()) {
+          obj.canceler.cancel(cancellationException());
+        }
+        auto w = kj::mv(obj.writer);
+        KJ_IF_SOME(writer, w) {
+          obj.context.addTask(obj.context.run(
+              [writer=kj::mv(writer), exception=cancellationException()]
+              (Worker::Lock& lock) mutable {
+            jsg::Lock& js = lock;
+            auto ex = js.exceptionToJs(kj::mv(exception));
+            return IoContext::current().awaitJs(lock,
+                writer->abort(lock, ex.getHandle(js)));
+          }));
+        }
+      }
+    }));
+  }
+
+  kj::Promise<void> write(const void* buffer, size_t size) override {
+    if (writer == kj::none) {
+      return KJ_EXCEPTION(FAILED, "Write after stream has been closed.");
+    }
+    if (size == 0) return kj::READY_NOW;
+    kj::ArrayPtr<const kj::byte> ptr(reinterpret_cast<const kj::byte*>(buffer), size);
+    return canceler.wrap(context.run([this, ptr](Worker::Lock& lock) mutable {
+      auto& writer = getInner();
+      auto source = KJ_ASSERT_NONNULL(jsg::BufferSource::tryAlloc(lock, ptr.size()));
+      source.asArrayPtr().copyFrom(ptr);
+      return context.awaitJs(lock, writer.write(lock, source.getHandle(lock)));
+    }));
+  }
+
+  kj::Promise<void> write(kj::ArrayPtr<const kj::ArrayPtr<const byte>> pieces) override {
+    if (writer == kj::none) {
+      return KJ_EXCEPTION(FAILED, "Write after stream has been closed.");
+    }
+    auto amount = 0;
+    for (auto& piece : pieces) {
+      amount += piece.size();
+    }
+    if (amount == 0) return kj::READY_NOW;
+    return canceler.wrap(context.run([this, amount, pieces](Worker::Lock& lock) mutable {
+      auto& writer = getInner();
+      // Sadly, we have to allocate and copy here. Our received set of buffers are only
+      // guaranteed to live until the returned promise is resolved, but the application code
+      // may hold onto the ArrayBuffer for longer. We need to make sure that the backing store
+      // for the ArrayBuffer remains valid.
+      auto source = KJ_ASSERT_NONNULL(jsg::BufferSource::tryAlloc(lock, amount));
+      auto ptr = source.asArrayPtr();
+      for (auto& piece : pieces) {
+        KJ_DASSERT(ptr.size() > 0);
+        KJ_DASSERT(piece.size() <= ptr.size());
+        if (piece.size() == 0) continue;
+        ptr.slice(0, piece.size()).copyFrom(piece);
+        ptr = ptr.slice(piece.size());
+      }
+
+      return context.awaitJs(lock, writer.write(lock, source.getHandle(lock)));
+    }));
+  }
+
+  // TODO(perf): We can't properly implement tryPumpFrom(), which means that Cap'n Proto will
+  //   be unable to perform path shortening if the underlying stream turns out to be another capnp
+  //   stream. This isn't a huge deal, but might be nice to enable someday. It may require
+  //   significant refactoring of streams.
+
+  kj::Promise<void> whenWriteDisconnected() override {
+    // TODO(soon): We might be able to support this by following the writer.closed promise,
+    // which becomes resolved when the writer is used to close the stream, or rejects when
+    // the stream has errored. However, currently, we don't have an easy way to do this.
+    //
+    // The Writer's getClosed() method returns a jsg::MemoizedIdentity<jsg::Promise<void>>.
+    // jsg::MemoizedIdentity lazily converts the jsg::Promise into a v8::Promise once it
+    // passes through the type wrapper. It does not give us any way to consistently get
+    // at the underlying jsg::Promise<void> or the mapped v8::Promise. We would need to
+    // capture a TypeHandler in here and convert each time to one or the other, then
+    // attach our continuation. It's doable but a bit of a pain.
+    //
+    // For now, let's handle this the same as WritableStreamRpcAdapter and just return a
+    // never done.
+    return kj::NEVER_DONE;
+  }
+
+  kj::Promise<void> end() override {
+    if (writer == kj::none) {
+      return KJ_EXCEPTION(FAILED, "End after stream has been closed.");
+    }
+    ended = true;
+    return canceler.wrap(context.run([this](Worker::Lock& lock) mutable {
+      return context.awaitJs(lock, getInner().close(lock));
+    }));
+  }
+
+private:
+  IoContext& context;
+  kj::Maybe<jsg::Ref<WritableStreamDefaultWriter>> writer;
+  kj::Canceler canceler;
+  kj::Own<kj::PromiseFulfiller<void>> doneFulfiller;
+  kj::Own<WeakRef<WritableStreamJsRpcAdapter>> weakRef =
+      kj::refcounted<WeakRef<WritableStreamJsRpcAdapter>>(
+          kj::Badge<WritableStreamJsRpcAdapter>(), *this);
+  bool ended = false;
+
+  WritableStreamDefaultWriter& getInner() {
+    KJ_IF_SOME(inner, writer) {
+      return *inner;
+    }
+    kj::throwFatalException(cancellationException());
+  }
+
+  static kj::Exception cancellationException() {
+    return JSG_KJ_EXCEPTION(DISCONNECTED, Error,
+        "WritableStream received over RPC was disconnected because the remote execution context "
+        "has endeded.");
+  }
+};
+
 }  // namespace
 
 void WritableStream::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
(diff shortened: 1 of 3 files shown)


src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+        }
+        KJ_DEFER(obj.writer = kj::none);
+        KJ_IF_SOME(writer, obj.writer) {
+          obj.context.addTask(obj.context.run(
+              [writer=kj::mv(writer), exception=kj::mv(exception)](Worker::Lock& lock) mutable {
+            jsg::Lock& js = lock;
+            auto ex = js.exceptionToJs(kj::mv(exception));
+            return IoContext::current().awaitJs(lock,
+                writer->abort(lock, ex.getHandle(js)));
+          }));
```
jasnell, line 404: For review: This part is a bit weird currently. We need to be able to trigger the abort algorithm on the WritableStream if we're still actually able to. At this point, however, the context might not exist anymore I think? Does the adapter need to have an explicit mechanism for triggering the `abort()` within the lock?

src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+// a lot slower
+class WritableStreamJsRpcAdapter final: public capnp::ExplicitEndOutputStream {
+public:
+  WritableStreamJsRpcAdapter(IoContext& context, jsg::Ref<WritableStreamDefaultWriter> writer)
+      : context(context), writer(kj::mv(writer)) {}
+
+  ~WritableStreamJsRpcAdapter() noexcept(false) {
+    weakRef->invalidate();
+    doneFulfiller->fulfill();
+  }
```
jasnell, line 381: For review... there's a fair amount of duplication in this class currently. Once we're sure this works the way it's supposed to, we can look to eliminate the duplicated code.

src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+        [ref=weakRef->addRef(), pieces](Worker::Lock& lock) mutable
+            -> kj::Promise<void> {
+      jsg::Lock& js = lock;
+      auto& context = IoContext::current();
+      KJ_IF_SOME(obj, ref->tryGet()) {
+        // This is rather unfortunate because we need to write each piece individually,
+        // which is a separate JS promise :-( ... what we need here is the equivalent of
+        // Promise.all but there is no C++ API provided by v8 to implement that... so we
+        // need to be creative. Still, it's more efficient than individual writes since
+        // we do not need to drop the isolate lock between each piece.
```
jasnell, line 445: Side note: Boy I wish `WritableStream` had a `writev` equivalent :-/

src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+
+        kj::Maybe<jsg::Promise<void>> lastPromise = kj::none;
+        for (auto& piece : pieces) {
+          if (piece.size() == 0) continue;
+          auto source = jsg::BufferSource::wrap(js, const_cast<byte*>(piece.begin()), piece.size(),
+                                                [](void*,size_t,void*) {}, nullptr);
+          v8::Local<v8::Value> value = source.getHandle(js);
+          auto write = obj.getInner().write(js, value);
+          KJ_IF_SOME(promise, lastPromise) {
```
jasnell, line 454: For review: this is the second place in the code that we have code approximating `Promise.all` (here and in socket.c++). We might consider moving it into a jsg utility... or maybe even submitting a patch to v8 to implement a `v8::Promise::All(...)` API.

src/workerd/api/streams/writable.c++
```diff
@@ -384,24 +532,44 @@ void WritableStream::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
...
-    ws.setEncoding(encoding);
-  });
+    externalHandler->write(
+        [capnpStream = kj::mv(capnpStream), encoding]
+        (rpc::JsValue::External::Builder builder) mutable {
+      auto ws = builder.initWritableStream();
+      ws.setByteStream(kj::mv(capnpStream));
+      ws.setEncoding(encoding);
+    });
+  } else {
```
jasnell, line 553: For review: there's obviously a fair amount of duplication here to with the previous branch. Once we're sure this is the approach that works, then we can eliminate the duplication.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1006,7 +1006,24 @@ export let streams = {
...
+      let result = '';
+      const { promise, resolve } = Promise.withResolvers();
+      const writable = new WritableStream({
+        write(chunk) {
+          result += dec.decode(chunk, {stream: true});
+        },
+        close() {
+          result += dec.decode();
+          resolve();
+        },
```
jasnell, line 1020: For review... once we figure out the right approach for aborting the writable stream such that the abort algorithm is invoked reliably, we need to add a test that verifies it.

src/workerd/api/streams/writable.c++
kentonv: I think that it is _always_ the case, in this branch, that the IoContext doesn't exist anymore:

* In the case where `waitForCompletionOrRevoke()` completes normally, then `weakRef` is already invalidated before this `kj::defer` callback runs, so none of this code executes.
* So, this block only executes in the case that `waitForCompletionOrRevoke()` was actually canceled prematurely.
* The promise is passed directly to `ioctx.addTask()`, so it would only be canceled if the `IoContext` is destroyed.
* Therefore, when executing this code, the `IoContext` is always destroyed (or in the process of being destroyed).

So I think it's unsafe to call `obj.context.run()` here. But also, you don't need to! Since the context is shutting down anyway, all I/O is canceled, including the abort algorithm.

src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+    }));
+  }
+
+  kj::Promise<void> write(const void* buffer, size_t size) override {
+    if (size == 0) return kj::READY_NOW;
+    return canceler.wrap(context.run(
+        [ref=weakRef->addRef(), buffer, size](Worker::Lock& lock) mutable
+            -> kj::Promise<void> {
+      jsg::Lock& js = lock;
+      auto& context = IoContext::current();
```
kentonv, line 416: You could capture `this` or `context` and then you wouldn't need to call `IoContext::current()` here.

(Same with other methods below.)

src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+      KJ_IF_SOME(obj, ref->tryGet()) {
+        auto source = jsg::BufferSource::wrap(js, const_cast<void*>(buffer), size,
+                                [](void*,size_t,void*) {}, nullptr);
```
kentonv, line 419: This doesn't seem safe -- you're wrapping the buffer passed into `write()`, which is no longer valid as soon as `write()` completes, but the application could hold on to the ArrayBuffer for arbitrarily long, couldn't it? I think we need to make a copy?

src/workerd/api/streams/writable.c++
kentonv: Since we need to copy the bytes into a GC'd buffer anyway, perhaps we can concatenate them into a single buffer in the process and avoid the need for multiple promises here?

src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+    }));
+  }
+
+  // TODO(perf): We can't properly implement tryPumpFrom(), which means that Cap'n Proto will
+  //   be unable to perform path shortening if the underlying stream turns out to be another capnp
+  //   stream. This isn't a huge deal, but might be nice to enable someday. It may require
+  //   significant refactoring of streams.
+
+  kj::Promise<void> whenWriteDisconnected() override {
+    return kj::NEVER_DONE;
```
kentonv, line 476: Is there any way to listen for the stream entering an error state?

src/workerd/api/streams/writable.c++
```diff
@@ -365,6 +365,154 @@ private:
...
+  }
+
+  kj::Promise<void> write(const void* buffer, size_t size) override {
+    if (size == 0) return kj::READY_NOW;
+    return canceler.wrap(context.run(
+        [ref=weakRef->addRef(), buffer, size](Worker::Lock& lock) mutable
+            -> kj::Promise<void> {
+      jsg::Lock& js = lock;
+      auto& context = IoContext::current();
+      KJ_IF_SOME(obj, ref->tryGet()) {
```
kentonv, line 417: `tryGet()` here can't possibly return null because the caller of `write()` must cancel the promise before allowing the stream to be destroyed. So we know the stream is still valid here (or we already have a UAF anyway).

So you can just capture `this` rather than capturing `weakRef->addRef()` above.

Same for other methods below.

src/workerd/api/streams/writable.c++
kentonv: I think the place where you actually want to invoke the abort algorithm is in the destructor `~WritableStreamJsRpcAdapter()`, if `writer` is non-null at that point and `end()` was never called.

src/workerd/api/streams/writable.c++
jasnell: umm.. we may be able to attach to the `writer.closed` promise? Or maybe the `writer.ready` promise. Will have to experiment to see.

jasnell: I added some comments in the code about this

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -293,6 +293,30 @@ export class MyService extends WorkerEntrypoint {
...
+    // TODO(now): In this test, we write to a stream that we expect to error
+    // immediately. Ideally, the writes would throw or at least there would be
+    // some signal that thge remote stream has failed. Currently it does not
+    // appear that the error signal is propagating back here correctly. Both
+    // writer.write() calls resolve successfully and we get no signal that the
+    // stream failed.
+    let writer = stream.getWriter();
+    let enc = new TextEncoder();
+    await writer.write(enc.encode("foo, "));
+    await writer.write(enc.encode("bar, "));
```
jasnell, line 306: @kentonv ... could use some help here. This is the remote side that receives the writable stream. On the first call to write, on the other side, the JS-backed writable stream will error. Unfortunately, that error does not appear to be making its way to this side, and I'm not sure why exactly. Both of these write awaits end up succeeding as if the stream is working just fine. See the additional comments below...

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1015,7 +1039,79 @@ export let streams = {
...
+      // (js-rpc-socket-test). Need to understand why.
+      // The writeToStreamExpectingError calls write twice, The first call to write should
+      // error the stream and should cause the subsequent write to fail also. However, both
+      // writes on the remote side succeed. What we should end up with is
+      // writeToStreamExpectingError throwing an error that matches the error thrown in the
+      // write algorithm.
+      // The error thrown in the write IS being propagated up to the capnp stream created
+      // in WritableStream::serialize, but it does not get propagated back to the remote
+      // side as far as I can tell. At least, the writes happening on that side are never
+      // errored and the remote stream is not errored.
```
jasnell, line 1084: In the `js-rpc-test` variant, this test case "passes", the `js-rpc-socket-test` variant, it fails, but the test log gives literally no information why it failed (there are no helpful log messages). That said, even in the `js-rpc-test` case, I don't believe the test is actually working correctly since on the remote end (in the `writeToStreamExpectingError()` method, both write awaits succeed even tho the first one should have triggerd an error. It's just not clear to me what's breaking down here.

kentonv: `js-rpc-socket-test.wd-test` had to be configured to disable info logs because otherwise the test takes 20x longer just due to symbolizing all the stack traces for logs of expected errors. This is configured in the BUILD file. You could temporarily change it while debugging. (You may also want to temporarily delete all the test cases except the one you care about.)

One reason why js-rpc-socket-test may behave differently is because when writing messages over a real socket, there is much more room for the kind of buffering I mentioned above. That is, you'll be able to queue even more write()s before one of them actually blocks.

src/workerd/api/streams/writable.c++
```diff
@@ -369,6 +369,173 @@ private:
...
+  //   stream. This isn't a huge deal, but might be nice to enable someday. It may require
+  //   significant refactoring of streams.
+
+  kj::Promise<void> whenWriteDisconnected() override {
+    // TODO(soon): We might be able to support this by following the writer.closed promise,
+    // which becomes resolved when the writer is used to close the stream, or rejects when
+    // the stream has errored. However, currently, we don't have an easy way to do this for
+    // a couple of reasons.
+    //
+    // 1. The Writer's getClosed() method returns a jsg::MemoizedIdentity<jsg::Promise<void>>.
```
kentonv, line 488: This is actually somewhat of an advantage since `jsg::Promise<T>` is single-use but a true V8 promise can have multiple waiters. The only problem is that we don't have a convenient JSG wrapper for a raw V8 promise -- that's actually something I found myself wanting a bunch while writing the RPC code. Essentially a `JsValue` variant for promises, probably called `JsPromise`?

src/workerd/api/streams/writable.c++
```diff
@@ -369,6 +369,173 @@ private:
...
+    //    jsg::MemoizedIdentity lazily converts the jsg::Promise into a v8::Promise once it
+    //    passes through the type wrapper. It does not give us any way to consistently get
+    //    at the underlying jsg::Promise<void> or the mapped v8::Promise. We would need to
+    //    capture a TypeHandler in here and convert each time to one or the other, then
+    //    attach our continuation. It's doable but a bit of a pain.
+    //
+    // 2. The contract of whenWritedDisconnected says that it indicates when the stream will
+    //    start to reject with DISCONNETED exceptions. There is no distinction with WritableStream.
```
kentonv, line 496: I think it's fine if it resolves upon the stream being broken with any permanent error. You don't actually have to distinguish DISCONNECTED errors. The doc comment is a bit over-fitted to the raw socket use case.

src/workerd/api/streams/writable.c++
jasnell: Sounds reasonable.

jasnell: For now, I'll leave this method unimplemented and will tackle this problem in a follow up PR.

jasnell:
@kentonv ...this should be ready for review!

src/workerd/api/streams/writable.c++
```diff
@@ -369,6 +369,173 @@ private:
...
+        if (!obj.canceler.isEmpty()) {
+          obj.canceler.cancel(cancellationException());
+        }
+      }
+    }));
+  }
+
+  kj::Promise<void> write(const void* buffer, size_t size) override {
+    if (size == 0) return kj::READY_NOW;
+    return canceler.wrap(context.run([this, buffer, size](Worker::Lock& lock) mutable {
```
jasnell, line 438: For review: note that this approach does mean that we end up having to reacquire the lock on *every write*. Which is painful. Another approach we could take here is to push the writes onto a buffer and separately fire up the `context.run(...)` such that once the lock is acquired that buffer is drained. We could potentially coalesce multiple writes within a single lock. Went with the less sophisticated approach here for now but likely a good idea to optimize here as a next step.

src/workerd/api/streams/writable.c++
```diff
@@ -369,6 +369,173 @@ private:
...
+  // that, the stream is revoked.
+  kj::Promise<void> waitForCompletionOrRevoke() {
+    auto paf = kj::newPromiseAndFulfiller<void>();
+    doneFulfiller = kj::mv(paf.fulfiller);
+
+    return paf.promise.attach(kj::defer([weakRef = weakRef->addRef()]() mutable {
+      KJ_IF_SOME(obj, weakRef->tryGet()) {
+        // Stream is still alive, revoke it.
+        if (!obj.canceler.isEmpty()) {
+          obj.canceler.cancel(cancellationException());
```
kentonv, line 430: Note that `canceler.cancel()` cancels _current_ promises but not future ones. You also need to set some state which is checked at the top of each method below. Maybe set `writer = kj::none`? But note that the calls to `getInner()` below happen too late -- you may need to check that `writer` is non-null _before_ calling `context.run()`.

src/workerd/api/streams/writable.c++
kentonv: We should probably optimize `context.run` itself so that it coalesces jobs rather than repeatedly re-acquiring the lock. (Not in this PR.)

src/workerd/api/streams/writable.c++
```diff
@@ -369,6 +369,173 @@ private:
...
+    if (size == 0) return kj::READY_NOW;
+    return canceler.wrap(context.run([this, buffer, size](Worker::Lock& lock) mutable {
+      jsg::Lock& js = lock;
+      // Sadly, we have to allocate and copy here. Our received buffer is only guaranteed
+      // to live until the returned promise is resolved, but the application code may
+      // hold onto the ArrayBuffer for longer. We need to make sure that the backing store
+      // for the ArrayBuffer remains valid.
+      auto source = KJ_ASSERT_NONNULL(jsg::BufferSource::tryAlloc(js, size));
+      auto ptr = source.asArrayPtr();
+      memcpy(ptr.begin(), buffer, size);
```
kentonv, line 446: Use Mike's new helpers?

kentonv (commented):
Sorry for slow review, this slipped past my inbox somehow.

src/workerd/api/streams/writable.c++
jasnell: Definitely will now that they exist :-)

src/workerd/api/streams/writable.c++
jasnell: Ah right, good point. Will work that up.

kentonv:
Reminder that GitHub loses information about force-pushes when you do two in a row which makes it hard to incrementally review. I think I've figured out that [this diff](https://github.com/cloudflare/workerd/compare/9a34e23c4e155492c4ac914d912741a20ce09187..3c989ac2786856938a5c83311a89fbb8ed7f87d5) represents the changes since I last reviewed, but in the future can you please post a list of fixup commits? The `fixup.sh` script in the internal repo generates such a list automatically if you use it.

kentonv (commented):
Ugh I wrote these comments two weeks ago but apparently forgot to submit the review.

src/workerd/api/streams/writable.c++
```diff
@@ -369,6 +369,171 @@ private:
...
+    // attach our continuation. It's doable but a bit of a pain.
+    //
+    // For now, let's handle this the same as WritableStreamRpcAdapter and just return a
+    // never done.
+    return kj::NEVER_DONE;
+  }
+
+  kj::Promise<void> end() override {
+    ended = true;
+    if (writer == kj::none) return kj::READY_NOW;
```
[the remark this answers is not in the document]
jasnell: Will make it throw. I think I prefer just checking to see if the writer is `kj::none` here directly and throwing a more specific error message rather than the one thrown by `getInner()` but the effect will be the same.

src/workerd/api/streams/writable.c++
```diff
@@ -369,6 +369,171 @@ private:
...
+        if (!obj.canceler.isEmpty()) {
+          obj.canceler.cancel(cancellationException());
+        }
+      }
+    }));
+  }
+
+  kj::Promise<void> write(const void* buffer, size_t size) override {
+    auto pieces = kj::arr(kj::arrayPtr(static_cast<const kj::byte*>(buffer), size));
+    return write(pieces).attach(kj::mv(pieces));
```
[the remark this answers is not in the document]
jasnell: Fair

kentonv (approved):

PR #2073, opened 2024-04-30
Update Request to support cache option

jasnell:
RM-18292

Updates the Request object to support specifying a cache mode per the standard. Only `undefined`, `'no-store'` and `'no-cache'` may be specified, and the fetch implementation specifically does not implement the semantics for either `'no-store'` or `'no-cache'` yet. This just lays the ground work for the additional changes.

The next step is to start adding the support for adding the appropriate header fields in the fetch implementation behind an autogate.

### DIFF
--- a/src/workerd/api/http-test.js
+++ b/src/workerd/api/http-test.js
@@ -126,6 +126,7 @@ export const inspect = {
     });
     assert.strictEqual(util.inspect(request),
 `Request {
+  cache: undefined,
   keepalive: false,
   integrity: '',
   cf: undefined,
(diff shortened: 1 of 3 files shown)


src/workerd/api/http-test.js
```diff
@@ -205,3 +206,36 @@ export const inspect = {
...
+      assert.strictEqual(req.cache, 'no-cache');
+    }
+    assert.throws(() => {
+      new Request('https://example.org', { cache: 'unsupported' });
+    }, {
+      name: 'TypeError',
+      message: 'Unsupported cache mode: unsupported',
+    });
+
+    // Any value other that undefined is currently not supported
```
dom96, line 231:
```suggestion
    // Any value other than undefined is currently not supported
```

src/workerd/api/http.c++
```diff
@@ -1739,6 +1772,15 @@ jsg::Promise<jsg::Ref<Response>> fetchImplNoOutputLock(
   kj::HttpHeaders headers(ioContext.getHeaderTable());
   jsRequest->shallowCopyHeadersTo(headers);
 
+  // If the jsRequest has a CacheMode, we need to handle that here.
+  // Currently, the ony cache mode we support is undefined, but we will soon support
```
dom96, line 1776:
```suggestion
  // Currently, the only cache mode we support is undefined, but we will soon support
```

dom96 (approved):

PR #2082, opened 2024-05-02
Continue work on new module registry, deferred eval callback

jasnell:
Iterating on the implementatin to add the mechanism for deferring evaluation outside the request context.

### DIFF
--- a/src/workerd/jsg/modules-new.c++
+++ b/src/workerd/jsg/modules-new.c++
@@ -51,7 +51,7 @@ ModuleBundle::Type toModuleBuilderType(ModuleBundle::BuiltinBuilder::Type type) 
 class EsModule final : public Module {
 public:
   explicit EsModule(Url specifier, Type type, Flags flags, kj::Array<const char> source)
-      : Module(kj::mv(specifier), type, flags | Flags::ESM),
+      : Module(kj::mv(specifier), type, flags | Flags::ESM | Flags::EVAL),
         source(kj::mv(source)),
         ptr(this->source),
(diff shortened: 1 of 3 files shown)


src/workerd/jsg/modules-new.c++
```diff
@@ -1217,7 +1262,8 @@ kj::Own<Module> Module::newSynthetic(
     Url specifier,
     Type type,
     EvaluateCallback callback,
-    kj::Array<kj::String> namedExports) {
+    kj::Array<kj::String> namedExports,
+    Flags flags) {
```
fhanau, line 1266: - [ ] Use flags in the SyntheticModule constructor

src/workerd/jsg/modules-new.h
```diff
@@ -171,9 +171,25 @@ class Module {
     // This is generally only suitable for worker-bundle entry point modules,
     // but could in theory be applied to any module.
     MAIN = 1 << 0,
+    // A Module with the ESM flag set is interpreted as an ECMAScript module.
     ESM = 1 << 1,
+    // A Module with the EVAL flag set is interpreted as a module that requires
```
fhanau, line 176: The description here is a bit confusing – with the current implementation `EVAL` being set only means that we use `maybeEvalCallback` if present, there is no error if there is no callback available. Maybe describing it with something like "if Flag::EVAL is not set, then the EvalCallback is ignored" (i.e. the converse of what we state below) would be more precise.

fhanau (approved):

PR #2083, opened 2024-05-02
Add temporary module logging

jasnell:
We need to determine two things:

1. Are worker bundles with multiple modules all that common and
2. Are worker bundles using module names that are known to be Node.js bare builtin specifiers.

The idea is for this logging to be temporary.

### DIFF
--- a/src/workerd/jsg/modules.c++
+++ b/src/workerd/jsg/modules.c++
@@ -9,8 +9,47 @@
 #include <set>
 
 namespace workerd::jsg {
-namespace {
 
+namespace {
+// This list must be kept in sync with the list of builtins from Node.js.
+// It should be unlikely that anything is ever removed from this list, and
+// adding items to it is considered a semver-major change in Node.js.
+static const std::set<kj::StringPtr> NODEJS_BUILTINS {
+  "_http_agent",         "_http_client",        "_http_common",
+  "_http_incoming",      "_http_outgoing",      "_http_server",
+  "_stream_duplex",      "_stream_passthrough", "_stream_readable",
+  "_stream_transform",   "_stream_wrap",        "_stream_writable",
+  "_tls_common",         "_tls_wrap",           "assert",
+  "assert/strict",       "async_hooks",         "buffer",
+  "child_process",       "cluster",             "console",
+  "constants",           "crypto",              "dgram",
+  "diagnostics_channel", "dns",                 "dns/promises",
+  "domain",              "events",              "fs",
+  "fs/promises",         "http",                "http2",
+  "https",               "inspector",           "inspector/promises",
+  "module",              "net",                 "os",
+  "path",                "path/posix",          "path/win32",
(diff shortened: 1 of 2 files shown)


src/workerd/jsg/modules.c++
```diff
@@ -9,8 +9,47 @@
...
+  "url",                 "util",                "util/types",
+  "v8",                  "vm",                  "worker_threads",
+  "zlib"
+};
+
+}  // namespace
+
+void logIfNodeSpecifier(kj::StringPtr specifier) {
+  if (NODEJS_BUILTINS.contains(specifier)) {
+    LOG_NOSENTRY(WARNING, "Using require() to import a module with a known Node.js built-in name",
```
dom96, line 47: Won't this be very noisy? Should this be a periodic log? or possibly a metric in prometheus?

jasnell: Part of the idea here is to get a sense of just how often folks are using modules that use the same name as Node.js builtins. Because of the way module imports work, it should happen at most once per isolate, not once per request unless they are using `require(...)` which also gives us good information. The intent is to keep this in for only about two weeks then revert it back out. Also the `NOSENTRY` should limit noise impact.

dom96 (approved):

src/workerd/jsg/modules.h
```diff
@@ -434,6 +436,12 @@ class ModuleRegistryImpl final: public ModuleRegistry {
 
   void add(kj::Path& specifier, ModuleInfo&& info) {
     entries.insert(kj::heap<Entry>(specifier, Type::BUNDLE, kj::fwd<ModuleInfo>(info)));
+
+    if (entries.size() > 1) {
+      // This is just to give us some sense of how many workers make use of multiple modules
+      // in the bundle...
+      LOG_PERIODICALLY(WARNING, "NOSENTRY Worker using multiple modules in a single bundle.");
```
IgorMinar, line 443: will this happen once per isolate start or how frequently? would it be possible to aggregate this at a worker or an account level?

the main answer I'm trying to answer is: how many unique workers with multiple modules are active in a given period of time.

src/workerd/jsg/modules.c++
```diff
@@ -9,8 +9,47 @@
...
+  "tls",                 "trace_events",        "tty",
+  "url",                 "util",                "util/types",
+  "v8",                  "vm",                  "worker_threads",
+  "zlib"
+};
+
+}  // namespace
+
+void logIfNodeSpecifier(kj::StringPtr specifier) {
+  if (NODEJS_BUILTINS.contains(specifier)) {
```
IgorMinar, line 46: similar to my comment below - would it be possible to log enough info here so that we can aggregate the signal at a worker level?

kentonv:
Should this have been a metric rather than a syslog? This seems like it'll be a very noisy log.

Issue #1957, opened 2024-04-04
Fetch API is inconsistent with the browser/web API for URL starting with whitespaces


SamyPesse:
I'm opening this issue to report a difference in implementation between `fetch` in the browser and in Cloudflare for an edge case. We are not really expecting a fix, as I'm not sure the browser is doing the right thing there; but I've suggested one at the end.

If you fetch a URL that starts with a whitespace, like:

```js
const res = await fetch(' https://raw.githubusercontent.com/cloudflare/workerd/main/README.md');
console.log(await res.text());
```

It will throw an error `TypeError: Fetch API cannot load` in Cloudflare Workers, while this code will work in the browser (at least Chrome).

We encountered a situation in which we needed to fetch a user-inputted URL, but our validation for the URL was not great.

A workaround we implemented was to wrap the url in a `new URL()`: `fetch(new URL(' https://raw.githubusercontent.com/cloudflare/workerd/main/README.md'))` works in Cloudflare. 

A solution in `workerd` could be to normalize / trim these urls in the fetch function, the same way `new URL` is doing it.

PR #2088, opened 2024-05-03
Add UrlWrapper/UrlList abstractions

jasnell:
In preparation for introducing a compatibility flag to change the handling of URLs in the fetch implementation to conform to standards, we need an abstraction over both kj::Url and jsg::Url. This adds that abstraction. It does not change anything to use it yet, but it provides the basics. The next step (in a follow up PR) will be to update all of the fetch API code to use it without with the compatibility flag, then finally add the flag and that switches things over to use jsg::Url

This is part of a refactor meant to improve standards compliance and works towards fixing a bug.

Refs: https://github.com/cloudflare/workerd/issues/1957

### DIFF
--- a/src/workerd/util/url-wrapper.c++
+++ b/src/workerd/util/url-wrapper.c++
@@ -1,0 +1,139 @@
+#include <workerd/util/url-wrapper.h>
+#include <workerd/jsg/jsg.h>
+
+namespace workerd {
+
+namespace {
+kj::OneOf<kj::String, kj::Url, jsg::Url> initialize(UrlWrapper::Mode mode, kj::StringPtr input) {
+  switch (mode) {
+    case UrlWrapper::Mode::LEGACY: {
+      return kj::str(input);
+    }
+    case UrlWrapper::Mode::STANDARD: {
+      return JSG_REQUIRE_NONNULL(jsg::Url::tryParse(input), TypeError, "Invalid URL");
+    }
+  }
+  KJ_UNREACHABLE;
+}
+}  // namespace
+
+UrlWrapper UrlWrapper::kLegacyFake = UrlWrapper::legacy("https://fake-host/");
+UrlWrapper UrlWrapper::kStandardFake = UrlWrapper::standard("https://fake-host/");
+
+UrlWrapper::UrlWrapper(Mode mode, kj::StringPtr input)
+   : inner(initialize(mode, input)) {}
+
+UrlWrapper::UrlWrapper(kj::OneOf<kj::Url, jsg::Url> inner)
+    : inner(kj::mv(inner)) {}
+
+void UrlWrapper::ensureParsed() {
+  KJ_SWITCH_ONEOF(inner) {
+    KJ_CASE_ONEOF(url, jsg::Url) {
+      // Parsed!
+    }
+    KJ_CASE_ONEOF(url, kj::Url) {
+      // Parsed!
+    }
+    KJ_CASE_ONEOF(str, kj::String) {
+      inner = JSG_REQUIRE_NONNULL(kj::Url::tryParse(str), TypeError, "Invalid URL");
+    }
+  }
+}
+
+UrlWrapper UrlWrapper::resolve(kj::StringPtr other) {
+  ensureParsed();
+  KJ_SWITCH_ONEOF(inner) {
+    KJ_CASE_ONEOF(url, jsg::Url) {
+      auto resolved = JSG_REQUIRE_NONNULL(url.resolve(other), TypeError, "Invalid URL");
+      return UrlWrapper(kj::mv(resolved));
+    }
+    KJ_CASE_ONEOF(url, kj::Url) {
+      auto resolved = JSG_REQUIRE_NONNULL(url.tryParseRelative(other), TypeError, "Invalid URL");
+      return UrlWrapper(kj::mv(resolved));
+    }
+    KJ_CASE_ONEOF(str, kj::String) {
+      // The call to ensureParsed above should have already parsed the URL.
+      KJ_UNREACHABLE;
+    }
+  }
+  KJ_UNREACHABLE;
+}
+
+kj::String UrlWrapper::toString(kj::Maybe<ToStringOptions> maybeOptions) {
+  auto options = maybeOptions.orDefault({});
+  ensureParsed();
+  auto context = options.context.orDefault(kj::Url::Context::REMOTE_HREF);
+  KJ_SWITCH_ONEOF(inner) {
+    KJ_CASE_ONEOF(url, jsg::Url) {
+      switch (context) {
+        case kj::Url::Context::REMOTE_HREF: {
+          if (options.includeFragments) {
+            return kj::str(url.getHref());
+          } else {
+            auto cloned = url.clone(jsg::Url::EquivalenceOption::IGNORE_FRAGMENTS);
+            return kj::str(cloned.getHref());
+          }
+        }
+        case kj::Url::Context::HTTP_PROXY_REQUEST: {
+          return kj::str(url.getOrigin(), url.getPathname(), url.getSearch());
+        }
+        case kj::Url::Context::HTTP_REQUEST: {
+          return kj::str(url.getPathname(), url.getSearch());
+        }
+      }
+    }
+    KJ_CASE_ONEOF(url, kj::Url) {
+      if (options.includeFragments) {
+        return url.toString(context);
+      } else {
+        auto cloned = url.clone();
+        cloned.fragment = kj::none;
+        return cloned.toString(context);
+      }
+    }
+    KJ_CASE_ONEOF(str, kj::String) {
+      // The call to ensureParsed above should have already parsed the URL.
+      KJ_UNREACHABLE;
+    }
+  }
+  KJ_UNREACHABLE;
+}
+
+UrlWrapper UrlWrapper::clone() {
+  KJ_SWITCH_ONEOF(inner) {
+    KJ_CASE_ONEOF(url, jsg::Url) {
+      return UrlWrapper(url.clone());
+    }
+    KJ_CASE_ONEOF(url, kj::Url) {
+      return UrlWrapper(url.clone());
+    }
+    KJ_CASE_ONEOF(str, kj::String) {
+      // The only reason this would be a string is if we're in LEGACY mode.
+      return UrlWrapper(Mode::LEGACY, str);
+    }
+  }
+  KJ_UNREACHABLE;
+}
+
+UrlWrapper UrlWrapper::legacy(kj::StringPtr input) {
+  return UrlWrapper(Mode::LEGACY, input);
+}
+
+UrlWrapper UrlWrapper::standard(kj::StringPtr input) {
+  return UrlWrapper(Mode::STANDARD, input);
+}
+
+size_t UrlList::size() const {
+  return urls.size();
+}
+
+void UrlList::add(UrlWrapper&& url) {
+  urls.add(kj::mv(url));
+}
+
+kj::Maybe<UrlWrapper&> UrlList::back() {
+  if (urls.size() == 0) return kj::none;
+  return urls.back();
+}
+
+}  // namespace workerd
--- a/src/workerd/util/url-wrapper.h
+++ b/src/workerd/util/url-wrapper.h
@@ -1,0 +1,70 @@
+#pragma once
+
+#include <workerd/jsg/url.h>
+#include <kj/compat/url.h>
+#include <kj/one-of.h>
+#include <kj/vector.h>
+
+namespace workerd {
+
+class UrlWrapper final {
+public:
+  // Determines the mode under which the UrlWrapper should operate.
+  enum class Mode {
+    // Legacy mode indicates that UrlWrapper will use the kj::Url implementation.
+    // This includes lazy parsing and a few other quirks.
+    LEGACY,
+    // Standard mode indicates that UrlWrapper will use the jsg::Url implementation.
+    STANDARD,
+  };
+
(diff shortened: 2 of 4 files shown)


src/workerd/util/url-wrapper.c++
```diff
@@ -0,0 +1,139 @@
...
+void UrlWrapper::ensureParsed() {
+  KJ_SWITCH_ONEOF(inner) {
+    KJ_CASE_ONEOF(url, jsg::Url) {
+      // Parsed!
+    }
+    KJ_CASE_ONEOF(url, kj::Url) {
+      // Parsed!
+    }
+    KJ_CASE_ONEOF(str, kj::String) {
+      inner = JSG_REQUIRE_NONNULL(kj::Url::tryParse(str), TypeError, "Invalid URL");
```
kentonv, line 38: Aren't you going to need to pass some options to `tryParse()`? All of the uses in `workerd/api/http.c++` pass options.

kentonv (commented):
I think before we embark on this we need a better idea of what we're fixing.

The linked issue is about failing to trim whitespace around the URL. That could, of course, be solved trivially. What other bugs are there?

The fetch API implementation does very little with `kj::Url`. It mainly only parses the URL to verify the scheme is in fact `http://` or `https://`. It also uses `kj::Url` to compute relative redirects, but is careful to set the parsing options to avoid interpreting percent encodes, which is otherwise the usual source of conformance problems.

`kj::Url` is also used elsewhere in our codebase and in KJ itself, but in order for this new `UrlWrapper` class to be able to help, it seems like we have to change a lot of interfaces to pass `UrlWrapper` instead of a string. In particular, it seems like `kj::HttpService` and `kj::HttpClient` interfaces need to be changed to pass this instead of strings. But that would be a very invasive change, and indeed would require putting this code in KJ rather than here... is that what you had in mind?

But again, I'm not sure that `kj::Url` is actually causing any problems in those places, since just about everywhere we are careful to use it with the options that avoid problems.

src/workerd/util/url-wrapper.h
```diff
@@ -0,0 +1,70 @@
...
+  UrlWrapper clone();
+
+  // Returns a legacy mode UrlWrapper that represents the "https://fake-host/" URL.
+  static UrlWrapper kLegacyFake;
+
+  // Returns a standard mode UrlWrapper that represents the "https://fake-host/" URL.
+  static UrlWrapper kStandardFake;
+
+private:
+  kj::OneOf<kj::String, kj::Url, jsg::Url> inner;
```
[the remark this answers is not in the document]
jasnell: Yeah, can definitely add a comment. The string is there largely because of our current behavior of lazily parsing the `Request` URL into the `kj::Url`. With new compat flag, the intent when using the `jsg::Url` variant is to not parse lazily (so it will be parsed when the `Request` object is created per the spec).

src/workerd/util/url-wrapper.c++
jasnell: Probably. I was actually hoping to see if I could avoid it :-) Once I start plugging in the actual use I was planning to tweak this as necessary.

jasnell:
>  think before we embark on this we need a better idea of what we're fixing....

There are a handful of subtle ways we are currently non-comformant with the fetch spec and other runtimes.

1. The leading and trailing whitespace issue described in #1957 is only part of the story there. The spec calls for the input URL to be parsed using a WHATWG compliant URL parser, which means not just handling leading and trailing whitespace a certain way, but ignoring interspersed tabs and newlines, ignoring invalid percent encodings, 
2. Per the spec, the URL is supposed to be parsed in the `Request` object constructor, meaning that any parse errors are supposed to be processed then. We currently (intentionally) violate that part of the spec and defer parsing till later. As part of aligning our behavior to the spec, browsers, and other runtimes, with the compat flag the URL will start to be parsed in the `Request` constructor.
3. On a redirect response, the spec requires that the `Location` header be parsed using WHATWG URL rules. We currently parse those using kj::Url. With the compat flag it would be updated to follow the spec.

> The fetch API implementation does very little with kj::Url. It mainly only parses the URL to verify the scheme is in fact http:// or https://. It also uses kj::Url to compute relative redirects, but is careful to set the parsing options to avoid interpreting percent encodes, which is otherwise the usual source of conformance problems.

Yep.. and the goal here to keep the interaction with that (and jsg::Url) as minimal as possible here. That minimism doesn't change.

> kj::Url is also used elsewhere in our codebase and in KJ itself, but in order for this new UrlWrapper class to be able to help, it seems like we have to change a lot of interfaces to pass UrlWrapper instead of a string. In particular, it seems like kj::HttpService and kj::HttpClient interfaces need to be changed to pass this instead of strings. But that would be a very invasive change, and indeed would require putting this code in KJ rather than here... is that what you had in mind?

Certainly not. What I do plan on is adding a method to `jsg::Url` that will return a `kj::Url` without the parsing. Something that would allow trivial conversion from one to the other, bypassing parsing. The intent is to use `UrlWrapper` only within http.h/c++ as a way of abstracting whether or not the new compat flag for fetch spec compliance is in use or not. 

> But again, I'm not sure that kj::Url is actually causing any problems in those places, since just about everywhere we are careful to use it with the options that avoid problems.

I don't think we really need to change the vast majority of cases where `kj::Url` is used. This is specifically just to align our handling of the `url` for `fetch`, `Request`, and `Response`.

src/workerd/util/url-wrapper.h
jasnell, line 50:
```suggestion
  // The UrlWrapper is either a kj::Url or a jsg::Url. When the UrlWrapper is
  // created initially, and the url is going to be a kj::Url, then the inner will
  // initially be a kj::String that is lazily parsed into the kj::Url. This is to
  // preserve an existing pattern we've had that defers parsing of the url
  // until it is used. When the url is going to a jsg::Url, then it is parsed when
  // the UrlWrapper is created.
  kj::OneOf<kj::String, kj::Url, jsg::Url> inner;
```

kentonv:
> The leading and trailing whitespace issue described in Fetch API is inconsistent with the browser/web API for URL starting with whitespaces #1957 is only part of the story there. The spec calls for the input URL to be parsed using a WHATWG compliant URL parser, which means not just handling leading and trailing whitespace a certain way, but ignoring interspersed tabs and newlines, ignoring invalid percent encodings, etc.

We already use `options.percentDecode = false` which causes the KJ parser to ignore percent encodings.

I was hoping to get some specific examples of what we do wrong here.

> Per the spec, the URL is supposed to be parsed in the Request object constructor, meaning that any parse errors are supposed to be processed then. We currently (intentionally) violate that part of the spec and defer parsing till later. As part of aligning our behavior to the spec, browsers, and other runtimes, with the compat flag the URL will start to be parsed in the Request constructor.

This seems like it could be solved by parsing the input URL and re-serializing it in the request constructor itself (if the compat flag is enabled). I'm fine with that, but it seems like a localized change that doesn't require this wrapper class.

> On a redirect response, the spec requires that the Location header be parsed using WHATWG URL rules. We currently parse those using kj::Url. With the compat flag it would be updated to follow the spec.

Again, would like to know specifically what kinds of things we're getting wrong here. I ask because historically there were two things we were getting wrong: interpreting percent escapes (standards say you're not supposed to attempt to decode them) and collapsing consecutive slashes. But, we added options to the KJ URL parser to fix both of these specific problems.

Note that I'm pushing on this in part because we've also noticed real-world use cases where using a _standard_ parser would actually have broken things, and we've tweaked our handling over the years to accommodate these. For example, there are real instances in the wild of HTTP requests whose path contains a `#` character, and such requests need to be able to proxy through Workers correctly even though passing such a URL to `fetch()` would not work. I'm worried that making a big change to URL handling could re-introduce some of these problems which we'll then have to re-fix. Whereas if the problems with the KJ URL parser are small and fixable, it might be easier to just do that.

> Then, of course, there are old outstanding issues like https://community.cloudflare.com/t/worker-fetch-mangles-location-header-url-on-redirect-from-origin/311984 ... in which using kj::Url for redirect resolution introduces bugs.

Thanks, this is a specific example. It appears that `tryParseRelative()` specifically is not correctly honoring the `allowEmpty` option here. This should be a one-liner fix. I wish someone had pointed this out to me before...

jasnell:
> I was hoping to get some specific examples of what we do wrong here.

I don't understand. I *gave* specific examples.

1. We do not handle leading, trailing, or interspersed whitespace correctly
2. We're not parsing the URL at the correct time (in the `Request` constructor)
3. We're not handling redirect resolution correctly...

Even with the ignoring percent decoding option we still aren't completely correct in the handling. For instance, there are odd cases found in the URL web platform tests like `https://a%C2%ADb/` that should be parsed and handled as `https://ab`  (the specific cases like this from the web platform tests are simplified repros of real world examples). In another example from the Web platform tests, the input url `https://example.com/\"quoted\"` is expected to be parsed as `https://example.com/%22quoted%22`, kj::Url parses it incorrectly as `https://example.com/"quoted"`.  Weirder examples from the tests like `https://x/?#\uFFFFy` is likewise mishandled by kj::Urls parsing. It should be `https://x/?#%EF%BF%BFy` but kj::Url produces, `https://x/?#￿y` ... the point here is that with just a cursory scan through only a subset of the standard web platform test URL samples it's trivial to find examples of cases that the kj::Url parser does not handle correctly. I'm certain there are absolutely more. (test cases duplicated here in ada-url test repo: https://github.com/ada-url/ada/blob/main/tests/wpt/urltestdata.json).

> This seems like it could be solved by parsing the input URL and re-serializing it in the request constructor itself (if the compat flag is enabled). I'm fine with that, but it seems like a localized change that doesn't require this wrapper class.

That would end up requiring us to parse the URL twice, once with `jsg::Url` then again with `kj::Url` ... and that doesn't resolve the redirect `location` URL handling which is expected to allow follow the WHATWG URL parsing rules. Double parsing doesn't seem to make sense, especially if our intent is to make the standard URL handling the default behavior.

> ... I ask because historically there were two things we were getting wrong: interpreting percent escapes (standards say you're not supposed to attempt to decode them) and collapsing consecutive slashes. But, we added options to the KJ URL parser to fix both of these specific problems.

As the above specific examples show we're also getting whitespace handling wrong, required percent-encoding of certain characters, and more. These rules apply both during the initial parse in the `Request` constructor and resolving the redirect `location` header. That's more than just making sure `tryParseRelative`supports `allowEmpty` since all of the other issues (whitespace, special characters, etc) would still be an issue.

> ...For example, there are real instances in the wild of HTTP requests whose path contains a # character, and such requests need to be able to proxy through Workers correctly even though passing such a URL to fetch() would not work. I'm worried that making a big change to URL handling could re-introduce some of these problems which we'll then have to re-fix. Whereas if the problems with the KJ URL parser are small and fixable, it might be easier to just do that.

But that's exactly why we have compatibility flags/dates, isn't it? The same issue exists between using the old `kj::Url` based implementation of `URL` verses the new compliant implementation. The compat flag in that case seems to be working rather well. We haven't had anyone complain about `URL` using standard parsing by default now -- but we have had multiple folks complain about the old non-standard behavior and have advised them to update their compat date to get the correct behavior. Existing workers that are handling an invalid `#` in the URL path today would continue to work so long as their compat date is set before the as-yet-undetermined default on date of the standard parsing. That seems consistent with our overall strategy. For folks with the newer compat date, the `#` in the path would be handled just like it would be in any other runtime that is implementing fetch, which seems to be the most desirable option. There would be no need to "re-fix" the `# in the path` issue.

*Could* we double parse the URLs (once using `jsg::Url`, followed by serialization, followed by reparsing with `kj::Url`). Sure. But that doesn't seem ideal. The `jsg::Url` parser (backed by ada-url) is a bit faster than kj::Url's implementation in most cases, correctly handles 100% of the web platform test URLs, correctly handles all of the various cases defined by the spec, fixes bugs reported by users, aligns our behavior with the browsers, Node.js, deno, etc. Double parsing just so we can keep the existing `kj::Url` bits in place seems silly to me.

Oh, side note: another thing I noticed that we're not doing correctly currently in the handling of the redirect `location`. Our handling logic does not validate that the new resulting URL is `http` or `https` the way we do in the initial request (that check is located in `parseUrl(...)` which is not used in the redirect case). We end up with an "internal error" because of a failed assert in `kj/compat/http.cc` rather than a more useful response code.

kentonv:
> But that's exactly why we have compatibility flags/dates, isn't it?

No. The issue I mentioned is not an issue of compatibility with past Worker versions. It's an issue of whether enabling a simple pass-through Worker on your site might break it. It is important that the answer is "no", regardless of the compat date used.

In the wild, we receive HTTP requests with URLs that contain `#` characters. These are clearly invalid, but nginx doesn't seem to have a problem passing them through. When a simple pass-through worker is applied, it needs to do the same. Note that if the application tries to actually parse/rewrite the URL, that's a different matter -- we can then throw an exception. But if the application simply passes the request on to origin without inspecting the URL, we need to make sure that we don't mess with `#` characters.

Now, there are a couple things true of such pass-through workers that we can take advantage of here:
- They never call the request constructor passing a string URL. The URL value is always inherited from an existing Request object.
- They always have redirect: manual, so I suppose we don't have to interpret redirections.

I think the following strategy would work and would keep the code as simple as possible:
- Parse/reserialize the URL in the Request constructor in order to catch invalid URLs, normalize escapes and whitespace, etc.
    - Use old or new parser based on compat flag.
    - Don't store the parsed URL in the Request object. Just the string.
    - For requests received off the wire, don't do any parsing of the URL.
- Change the code that checks for http/https schema to just do a prefix match, which should always work if we can assume the URL was sanitized first.
- If and only if we need to follow a redirect later (rare!), parse the URL again there for that purpose. I'd be OK with us just always using the new parser for this, but we could also choose new or old parser based on compat flag.

PR #2091, opened 2024-05-06
Fix package installation in workerd with top level entropy restrictions

hoodmane:
Package installation uses tempfile which tries to create entropy. This patches it to use a deterministic sequence of file names.

@garrettgu10


garrettgu10 (approved):
Amazing, LGTM 🚀

PR #2095, opened 2024-05-08
Fix ai binding logs property

G4brym:
This re-adds a removed property as backwards compatibility, removed in the last release

### DIFF
--- a/src/cloudflare/internal/ai-api.ts
+++ b/src/cloudflare/internal/ai-api.ts
@@ -30,6 +30,11 @@ export class InferenceUpstreamError extends Error {
 export class Ai {
   private readonly fetcher: Fetcher
 
+  /*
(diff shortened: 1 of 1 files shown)


DaniFoldi:
Hi, I saw the workers-sdk test broken by the removal of this property. Just curious, if you can share, is there any particular reason the logs were removed/hidden in the first place?

Something like AI is much more complicated than say KV - there's not much to log in the case of KV (put or delete, that's it), but with AI, many things can happen, and logs were useful for me (and probably other folks who found them).

RamIdeas (approved):

RamIdeas:
FYI we've decided to hold off upgrading workerd in wrangler until this PR lands and is released https://github.com/cloudflare/workers-sdk/pull/5752#issuecomment-2102288398

irvinebroque (approved):

PR #2096, opened 2024-05-08
Generate python import tests in bazel

garrettgu10:
Prerequisite: #1970 

Includes: #2108

This PR adds dozens of wd-tests generated by Bazel, based off a list generated by pyodide-build-scripts from the pyodide `meta.yaml` files and added onto `pyodide_bucket.bzl`. 

Each test simply has the tested package listed as a requirement and a `worker.py` script that just imports a bunch of stuff from that package. I expect it will take a while to get all these tests passing but we do need them to pass eventually.


build/pyodide_bucket.bzl
```diff
@@ -1,8 +1,178 @@
...
-PYODIDE_PACKAGES_TAR_ZIP_SHA256 = "749967941204154e7ae866fe08f1216a3e5ee58ba6a3757231a5be0d9d4430f8"
-PYODIDE_ALL_WHEELS_ZIP_SHA256 = "9e7c330ee93d81d0356cc2d585f217dfee58b623ad4535282baa6e82bd063eee"
+PYODIDE_PACKAGE_BUCKET_URL = "https://pub-45d734c4145d4285b343833ee450ef38.r2.dev/20240508/"
+PYODIDE_GITHUB_RELEASE_URL = "https://github.com/cloudflare/pyodide-build-scripts/releases/download/20240508/"
+PYODIDE_LOCK_SHA256 = "697b8917a8dcedbb78dd21fa1927ee310604106579e93b1eac425344d9212a73"
+PYODIDE_PACKAGES_TAR_ZIP_SHA256 = "12238b283ab2c2b996fcec73aa9e61f057c86aad5d612c2d44f4b2764f7ea03a"
+PYODIDE_ALL_WHEELS_ZIP_SHA256 = "36f980ae303357410bc4a34b430bcf82631e80525ac144888c696bcf92d03504"
+
+# IMPORTANT: when updating this file in git, check the diff to make sure none of the imports below are being removed unexpectedly
+PYODIDE_IMPORTS_TO_TEST = {
```
dom96, line 11: Do we have a script that can regenerate this? I think it would be useful and it could also verify that no existing packages are being removed.

dom96 (commented):
I'd love @mikea and/or @fhanau to have a look and verify the way we're using bazel to generate these tests makes sense.

But otherwise LGTM. Thanks for implementing this, it makes our test coverage so much better!

build/pyodide_bucket.bzl
garrettgu10: It's generated by `pyodide-build-scripts`. This whole file is part of the release that's generated by the GitHub action.

garrettgu10: For reference here is the script: https://github.com/cloudflare/pyodide-build-scripts/pull/2

src/workerd/server/tests/python/import_tests.bzl
```diff
@@ -0,0 +1,45 @@
...
+load("//:build/pyodide_bucket.bzl", "PYODIDE_IMPORTS_TO_TEST")
+
+def generate_import_py_file(imports):
+  res = "def test():\n"
+  for imp in imports:
+    res += "   import "+imp+"\n"
+  return res
+
+def generate_wd_test_file(requirement):
+  return """
```
mikea, line 12: for readability define this constant as top-level variable:

```
WD_FILE_TEMPLATE = """
...
```

src/workerd/server/tests/python/import_tests.bzl
```diff
@@ -0,0 +1,45 @@
...
+        ],
+        compatibilityDate = "2024-05-02",
+        compatibilityFlags = ["python_workers"],
+      )
+    ),
+  ],
+);""".format(requirement, requirement)
+
+def gen_import_tests():
+  for lib in PYODIDE_IMPORTS_TO_TEST.keys():
```
mikea, line 31: maybe pass PYODIDE_IMPORTS_TO_TEST as a parameter? from BUILD file? It is better if rules don't depend on setup details.

mikea (approved):
build changes make sense mod couple nits

WORKSPACE
```diff
@@ -126,14 +126,28 @@ http_archive(
...
 http_file(
     name = "pyodide-lock.json",
     sha256 = PYODIDE_LOCK_SHA256,
     url = PYODIDE_GITHUB_RELEASE_URL + "pyodide-lock.json",
 )
 
+http_archive(
+    name = "all_pyodide_wheels",
+    sha256 = PYODIDE_ALL_WHEELS_ZIP_SHA256,
+    type="zip",
```
fhanau, line 140: Nit: this line is not needed, bazel can infer the type from the extension.

fhanau:
Code LGTM, I think auto-generating tests here will be really useful in improving python test coverage.
I hate to always bring up resource constraints but the ~50 tests being added all take between 20 - 50 seconds each – with the GitHub Actions runners being limited to 4 vCPUs each, that will slow down CI on every run making a code change. Is there any potential to optimize python-based tests?

hoodmane:
Could we gate the tests to only run on changes inside the pyodide folders?

garrettgu10:
> Could we gate the tests to only run on changes inside the pyodide folders?

Good idea, anyone know how to do that? 🫣

Maybe Bazel's cache could be useful if someone could point me to how to configure it 😮‍💨

garrettgu10:
Is it possible to run the tests off a baseline snapshot as well? @hoodmane @dom96

dom96:
> Is it possible to run the tests off a baseline snapshot as well? @hoodmane @dom96

It's not something that we have set up, but it could be done. We have lots of tests using different kinds of snapshots in the EW repo, might be easier to do there.

garrettgu10:
@fhanau This PR is ready for review again after changes -- the tests are now configured to only run in the (linux) configuration, which was chosen because it was the fastest-running test pipeline. 

Will require similar CI changes upstream.

.bazelrc
```diff
@@ -47,6 +47,17 @@ build --host_per_file_copt='external/zlib@-Wno-unknown-warning-option,-Wnodeprec
 # for some reason tcmalloc uses internal msan-enabled annotation in the code, rather than public one
 build --per_file_copt=external/com_google_tcmalloc@-DANNOTATE_MEMORY_IS_INITIALIZED=ABSL_ANNOTATE_MEMORY_IS_INITIALIZED
 
+# Increasing the optimization level of some portions of V8 significantly speeds up Python tests
```
fhanau, line 50: I know I suggested this approach, but it should be explained well here – it should be clear that this is an optional "performance hack" and we could explore if applying this to only a few folders is sufficient to get most of the speedup for python. This is not a blocking issue however, V8 is cached for the vast majority of builds so build speeds should be acceptable. – I hope that we'll reduce the amount of code we need to compile for python soon so that there's less of a need for manual intervention here.

We can't apply this by default unfortunately – the release configuration uses -O3 so using -O2 here reduces the optimization level for these files. One workaround would be to define a build:v8-codegen-opt configuration and add --config=v8-codegen-opt to `bazelisk build [...]` and `bazelisk test [...]` in .github/workflows/test.yml – it is safe to apply this to all debug builds.

Before merging, make sure that the downstream branch has tests passing to avoid any breakage there.

fhanau (approved):

PR #2103, opened 2024-05-09
Add better error handling to ai binding

G4brym:


fhanau (commented):
The changes make sense to me. Since the provided error includes the `name` and `description` fields, is `message` less relevant here?

jasnell (approved):

src/cloudflare/internal/ai-api.ts
```diff
@@ -98,9 +104,18 @@ export class Ai {
...
     }
+
+    private async _parseError(res: Response): Promise<InferenceUpstreamError> {
+      try {
+        const content = (await res.json()) as AiError
+        return new InferenceUpstreamError(`${content.internalCode}: ${content.description}`, content.name);
+      } catch {
+        return new InferenceUpstreamError(await res.text());
+      }
+    }
```
Cherry, line 118: Will this work? If you call `.json()` and consume the response body, you can't then call `.text()` on it if that throws, can you?

src/cloudflare/internal/ai-api.ts
```diff
@@ -52,37 +59,36 @@ export class Ai {
...
 
-    const res = await this.fetcher.fetch("http://workers-binding.ai/run?version=2", fetchOptions);
+    const res = await this.fetcher.fetch("http://workers-binding.ai/run?version=3", fetchOptions);
 
     this.lastRequestId = res.headers.get("cf-ai-req-id");
 
     if (inputs['stream']) {
       if (!res.ok) {
-        throw new InferenceUpstreamError(await res.text());
+        throw await this._parseError(res)
```
Cherry, line 84: Has backwards compat been considered here? Are there any possible cases where people would previously be expecting a specific error string (to retry, log, etc.) and this might now change?

src/cloudflare/internal/ai-api.ts
```diff
@@ -52,37 +59,36 @@ export class Ai {
...
       }
 
       return res.body;
 
     } else {
       if (!res.ok || !res.body) {
-        throw new InferenceUpstreamError(await res.text());
+        throw await this._parseError(res)
```
Cherry, line 91: Has backwards compat been considered here? Are there any possible cases where people would previously be expecting a specific error string (to retry, log, etc.) and this might now change?

src/cloudflare/internal/d1-api.ts
```diff
@@ -6,7 +6,7 @@ interface Fetcher {
   fetch: typeof fetch
 }
 
-type D1Response = {
+interface D1Response {
```
Cherry, line 9: unrelated changes in this file?

src/cloudflare/internal/sockets.d.ts
```diff
@@ -12,16 +12,16 @@ export class Socket {
   public startTls(options: TlsOptions): Socket
 }
 
-export type TlsOptions = {
+export interface TlsOptions {
```
Cherry, line 15: unrelated changes in this file?

src/cloudflare/internal/vectorize.d.ts
```diff
@@ -38,17 +38,14 @@ type VectorizeVectorMetadataFilterOp = "$eq" | "$ne";
 /**
  * Filter criteria for vector metadata used to limit the retrieved query result set.
  */
-type VectorizeVectorMetadataFilter = {
-  [field: string]:
-    | Exclude<VectorizeVectorMetadataValue, string[]>
+type VectorizeVectorMetadataFilter = Record<string, | Exclude<VectorizeVectorMetadataValue, string[]>
```
Cherry, line 41: unrelated changes in this file?

src/cloudflare/internal/ai-api.ts
```diff
@@ -98,9 +104,18 @@ export class Ai {
...
+    public getLogs(): string[] {
       return []
     }
+
+    private async _parseError(res: Response): Promise<InferenceUpstreamError> {
+      try {
+        const content = (await res.json()) as AiError
+        return new InferenceUpstreamError(`${content.internalCode}: ${content.description}`, content.name);
+      } catch {
+        return new InferenceUpstreamError(await res.text());
```
DaniFoldi, line 116: Hey :wave:,

@Cherry sent me this PR, and while browsing through I noticed that this wouldn't actually work. Even if the JSON parsing fails, such as on an empty response body, the body would be already consumed and this would throw again.

src/cloudflare/internal/d1-api.ts
G4brym: eslint formated the file by mistake, fixed

src/cloudflare/internal/ai-api.ts
G4brym: Thanks for pointing that out, i've fixed the issue and added a unit test to cover it

src/cloudflare/internal/ai-api.ts
G4brym: It depends on how the customer is matching the strings, the main texts didn't change, for example if you are checking a match with `err.message.includes('timeout')` this will continue to work
Since this is including new information on the errors and not excluding, customers shouldn't be affected

Cherry: That sounds like a breaking change then. If people are checking for a specific string error and that changes, their application could behave in different ways after this update.

celso: I don't think this is a breaking change. Workers AI errors were largely undocumented and flaky so far, developers should def not be matching strings, we can't support that. This PR is part of larger effort to fix errors end to end, 
 one of the goals is to document the errors and create error codes, which then developer can match agaisn't.

Cherry: > developers should def not be matching strings, we can't support that

Was this ever called out in any documentation for Workers AI? Up to this point, people (as far as I know), haven't had any guidance on how best to catch and handle errors from Workers AI. So as per https://www.hyrumslaw.com/ and other Developer Platform products, people are going to rely on strings.

Hopefully this doesn't break anyone's applications.

kentonv: > Workers AI errors were largely undocumented and flaky so far, developers should def not be matching strings, we can't support that.

"Should" doesn't matter. Documentation doesn't matter. The only thing that matters is what developers are actually doing in the wild. It is always our (Cloudflare's) fault if an update in the runtime breaks live workers -- it always requires an immediate rollback of the runtime release.

Now, I don't necessarily think that changing error message text must always be considered a breaking change. It really depends on whether the API is such that people have a need to match error strings. If we suspect that such a need exists, then we have to assume some people are doing it, so we can't change the text without a compat flag. (I don't know anything about this specific API so I'm not the right person to judge this.)

But also, if we think that people have a need to match error text, it suggests a flaw in the API design. In general, exceptions should only be thrown for things that the application isn't expected to handle. Situations which apps are expected to handle should be communicated without using exceptions. Of course, if it's already the case that the API has forced people to match error strings then it's too late to fix that now.

Cherry: For some additional context about this API, all kinds of errors can be thrown when you call `env.AI.run(...)`.

Some examples:
- `Error with the model repository`
- `ERROR 3015: Error with the model repository`
- `ERROR 3025: Unknown internal error`
- `ERROR 3001: Unknown internal error`
- `could not route request to AI model`
- `ERROR 3034: Unknown internal error`
- `<!DOCTYPE html> <html la...` (HTML exceptions)

Some of these are transient, and I think it would be pretty reasonable to expect folks to be expecting some/many of them and retrying requests. Especially since there's no other error story today, and catching strings is basically all users can do.

If I'm understanding the change here, these errors on `err.message` now might instead look like:

`3025: Unknown internal error`

Which is a subtle but meaningful difference from the previous `ERROR 3025: Unknown internal error` if anyone is explicitly checking `if (err.message === '...')`.

I'm not sure what errors like `could not route request to AI model` or `Error with the model repository` would look like now, since I don't know how/if they'd return JSON and be caught in the new JSON handling of errors here.

kentonv: @Cherry TBH, none of the errors you list seem to me like things that an application could reasonably respond to in any way other than report an error to the user. Maybe they could influence the app's decision on whether to retry, but I find it pretty hard to say which ones would be retryable based on the text... Personally I would probably just retry all of them or nothing depending on how common errors are...

G4brym: All the errors you listed above, are errors not caused by the user, and in this case the customers probably isn't matching strings against it, because he will probably want to retry the request regardless of the error message.

For example, and lets be reasonable, why would a customer match exactly this string `err.message === 'ERROR 3034: Unknown internal error'`?

Cherry: I would argue that `ERROR 3007: Request timeout` is a pretty good indication to retry, and `ERROR 3015 Error with the model repository` also seems to be one that's particularly transient and clears up after a retry.  Obviously, none of this is documented, and therefore mostly anecdotal, but if folks have experimented with the API and hit these things, it's possible they're catching specific errors they have seen occur more than others.

Or, my code might specifically check for `ERROR 3010: Invalid or incomplete input for the model: failed to decode JSON: Request is too large`, and then give the user a nicer message about what they uploaded.

> For example, and lets be reasonable, why would a customer match exactly this string err.message === 'ERROR 3034: Unknown internal error'?

In a reasonable scenario, I don't think they would catch that specific error alone if they understand its cause. But they might do something like this:
```js
const errorsToRety = [
	"ERROR 3025: Unknown internal error",
	"ERROR 3034: Unknown internal error",
	"ERROR 3040: Unknown internal error",
	"could not route request to AI model",
]

if (errorsToRety.includes(err.message)) {
	// Retry the request
}
```
after observing these as transient errors when calling the Workers AI API, and want to limit the errors their own users see. Of course you could argue they should retry all errors, but if they've observed `ERROR 3010: Invalid or incomplete input for the model: failed to decode JSON: Request is too large`, or other non-transient errors, they might specifically be handling this case to report a problem to their own users.

I don't think this is an unlikely scenario if anyone has been using a product for a few months and responding to different errors they've observed in their own logging. But, I trust that you all are going to be most familiar with what your customers are doing on your API and will make the right call here.

kentonv: Retryability in particular is an interesting case.

Frankly, we should design our APIs to retry automatically if it's safe and effective to do so. If we do that, then apps can assume any API error is not retryable.

Of course, that's not the case today. But I would say that any retry logic like this is obviously a best-effort thing anyway. Breaking someone's retry heuristic doesn't really break their app. It might increase their rate of serving errors to end users slightly. I'd say if the only way that people are inspecting errors is to decide whether to retry, there's still room for some judgement call about how important it is to keep that logic working. (In making this judgment I would factor in how common errors are today and how commonly retries actually work.)

Cherry: For reference, errors are relatively common with Workers AI: https://grafana.chaika.me/public-dashboards/82c159675de54d62b42e60e5238e28c5?orgId=1. Examples from Cloudflare's own devrel often includes retry logic with how common they are, in fact.

> Frankly, we should design our APIs to retry automatically if it's safe and effective to do so. If we do that, then apps can assume any API error is not retryable.

Agreed, this would definitely be the most ideal scenario.

PR #2108, opened 2024-05-09
Fix top level randomness in aiohttp

hoodmane:
aiohttp instantiates a `random.Random()` in order to be a default argument for `WebSocketWriter`. Instead, use the `random` module itself for this which implements the same interface as `Random`.

Should fix one of the test failures in #2096.


hoodmane:
It works:
```
2024-05-09T20:58:32.9101316Z //src/workerd/server/tests/python:import/aiohttp/import    
PASSED in 26.5s
```

garrettgu10 (approved):

garrettgu10:
Merged as part of #2096 🥳

Issue #2099, opened 2024-05-09
🐛 Bug Report — Body.prototype.blob() incorrectly extracting content-type


jasnell:
Reproduction:

```js
export default {
  async fetch(req, env) {
    if (req.url.endsWith('/sub')) {
      return new Response("Hello World\n", {
        status: 200, headers: {
          "content-type": ['text/plain;charset=xyz,a=b', 'x/y', 'text/plain;c=d']
        }
      });
    } else {
      const resp = await fetch('http://localhost:8080/sub');
      console.log(resp.headers.get('content-type'));
      const blob = await resp.blob();
      console.log(blob.type);
      return new Response("ok");
    }
  }
};
```

Expected Result:

Per the [spec](https://fetch.spec.whatwg.org/#concept-header-extract-mime-type) the `blob.type` property extracted should be `text/plain;c=d`

Actual Result:

The `console.log(blob.type)` outputs `text/plain;charset=xyz;a=b,x/y,text/plain;c=d`

PR #2109, opened 2024-05-10
Fix content type determination for blob from fetch

jasnell:
Using a compatibility flag, selects the correct content type for a blob from the fetch-provided content-type header(s) as defined by the fetch spec: https://fetch.spec.whatwg.org/#concept-header-extract-mime-type

Sadly it's potentially breaking change so the compat flag is needed.

Fixes: https://github.com/cloudflare/workerd/issues/2099

Goal: Web standards compliance, bug fix

### DIFF
--- a/src/workerd/util/mimetype.c++
+++ b/src/workerd/util/mimetype.c++
@@ -26,7 +26,7 @@ bool isQuotedStringTokenChar(const unsigned char c) {
   return (c == '\t') || (c >= 0x20 && c <= 0x7e) || (c >= 0x80);
 }
 
-kj::StringPtr skipWhitespace(kj::StringPtr str) {
+kj::ArrayPtr<const char> skipWhitespace(kj::ArrayPtr<const char> str) {
   auto ptr = str.begin();
   auto end = str.end();
   while (ptr != end && isWhitespace(*ptr)) { ptr++; }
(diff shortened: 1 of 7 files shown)


src/workerd/util/mimetype.c++
```diff
@@ -378,4 +383,64 @@ bool MimeType::isVideo(const MimeType& mimeType) {
...
+  kj::Maybe<MimeType> mimeType;
+
+  constexpr static auto findNextSeparator = [](auto& input) -> kj::Maybe<size_t> {
+    // Scans input to find the next comma (,) that is not contained within
+    // a quoted section, returning the position of the comma or kj::none
+    // if not found.
+    for (size_t i = 0; i < input.size(); ++i) {
+      if (input[i] == '"') {
+        // Skip to the end of the quoted section
+        while (++i < input.size() && (input[i] != '"' || input[i-1] == '\\')) {}
```
dom96, line 397: I'm confused by the handling of inputs like `\"foobar` here. Doesn't that indicate that it's not a quoted section in which case nothing should be skipped? But this code as it is right now will skip to the end of the "foobar", am I misreading something or is this incorrect?

dom96 (approved):
Nice! Looks good overall but have one question.

src/workerd/util/mimetype.c++
jasnell: ah, likely need an additional condition on the previous line to ensure the `"` we're on currently isn't escaped. Will update and add a test (that edge case is apparently not covered by the existing web platform tests)

PR #2110, opened 2024-05-10
Fix setTimeout ID collisions.

jp4a50:
`workerd::api::gpu::AsyncRunner` was using a different instance of `TimeoutId::Generator` than is used for `setTimeout` calls from JS.

I fix this by moving ownership of the `TimeoutIdGenerator` into the `TimeoutManagerImpl` itself.

In the case of `gpu::AsyncRunner`, it doesn't look like there's any need to generate an ID for the timeout, so we could alternatively approach this by re-using the timeout logic but not generating an ID in order to avoid silently using up timeout ID values under the hood. Not sure whether it's worth it.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -687,7 +687,6 @@ TimeoutId::NumberType ServiceWorkerGlobalScope::setTimeoutInternal(
     jsg::Function<void()> function,
     double msDelay) {
   auto timeoutId = IoContext::current().setTimeoutImpl(
-      timeoutIdGenerator,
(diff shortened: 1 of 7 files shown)


jp4a50:
Hmm, actually, this does change the semantics of `setTimeout/clearTimeout` across multiple requests that execute within the same global scope.

Currently, I think it would be possible to do clear a timeout set by a previous request.

Is this something which we would want to continue to support, or are we OK just preventing this? @jasnell

jasnell:
The WebGPU API has been removed. Presumably this is no longer an issue.

Issue #2101, opened 2024-05-09
🐛 Bug Report — fetch response with multiple content-length values yields "internal error"


jasnell:
When fetch receives a response with multiple content-length headers, the runtime throws an unhelpful "internal error".

Reproduction:

```
// Node.js script because workers does not let us set the content-length manually:
import { createServer } from 'node:http';
createServer((req, res) => {
  res.writeHead(200, {
    'content-length': [4,5,6],
  });
}).listen(8081);
```

```
// Then in workers
const resp = await fetch('http://localhost:8081');
```

The runtime will log an internal error indicating "invalid Content-Length header value" but does not propagate a useful error up to the user. 

Per the spec, fetch should be tolerant of multiple content-length headers without failing and the spec includes an algorithm to extract the content length (https://fetch.spec.whatwg.org/#content-length-header) .. *ideally* we whould be using that to determine if the content-length is invalid. In the repro example case above, the content-length *would* be invalid given the inconsistent values but a more useful error should be reported.

In the case there are multiple `content-length` headers with the *same* value, e.g. `content-length': [5,5,5]`, the runtime will  report this as an invalid content-length and an opaque "internal error". Per the fetch spec, the extracted valid content length here is `5`.

Note that per all the specs, having multiple content-length headers with different values is absolutely invalid and should result in an error (it just needs to be a useful one). RFC 9110 indicates that it is acceptable to have multiple content-length headers with the *same* value but that implementations have latitude to decide how to handle it. The fetch spec algorithm is tolerant of it and will normalize those duplicates away. Other implementations do, in fact, throw an error (like kj).  If we are going to continue treating this case as an error rather than follow the fetch spec algorithm, we need to at least make it a useful error rather than a generic "internal error"

This *might* need to be addressed at the kj-http level more than in the runtime but opening the issue here to track.

PR #2111, opened 2024-05-10
Improve user facing error message on invalid content-length

jasnell:
Previously an invalid content-length would result in an unhelpful "internal error" message to the user. This intercepts that error and replaces it with a more helpful message.

Longer term, it would be idea to align this more with the fetch spec but for now, improving the error message is a good first step.

Fixes: https://github.com/cloudflare/workerd/issues/2101

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1887,7 +1887,13 @@ jsg::Promise<jsg::Ref<Response>> fetchImplNoOutputLock(
       nativeRequest = client->request(jsRequest->getMethodEnum(), url, headers, uint64_t(0));
     }
     return ioContext.awaitIo(js,
-        AbortSignal::maybeCancelWrap(signal, kj::mv(KJ_ASSERT_NONNULL(nativeRequest).response)),
+        AbortSignal::maybeCancelWrap(signal, kj::mv(KJ_ASSERT_NONNULL(nativeRequest).response))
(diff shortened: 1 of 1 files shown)


fhanau (approved):

src/workerd/api/http.c++
```diff
@@ -1887,7 +1887,13 @@ jsg::Promise<jsg::Ref<Response>> fetchImplNoOutputLock(
       nativeRequest = client->request(jsRequest->getMethodEnum(), url, headers, uint64_t(0));
     }
     return ioContext.awaitIo(js,
-        AbortSignal::maybeCancelWrap(signal, kj::mv(KJ_ASSERT_NONNULL(nativeRequest).response)),
+        AbortSignal::maybeCancelWrap(signal, kj::mv(KJ_ASSERT_NONNULL(nativeRequest).response))
+            .catch_([](kj::Exception&& exception) -> kj::Promise<kj::HttpClient::Response> {
+          if (exception.getDescription().startsWith("invalid Content-Length header value")) {
+            return JSG_KJ_EXCEPTION(FAILED, Error, exception.getDescription());
```
kentonv, line 1893: @jasnell Minor point, but JavaScript error message style is usually more formal than KJ error message style, e.g. expecting complete sentences with a capitalized first letter and terminating punctuation.

On another note, I don't think it's possible for this error to happen in production since the invalid value would never get past FL.

PR #2114, opened 2024-05-11
Add fetch_standard_url compat flag

jasnell:
Adds a compat flag that makes it so that the fetch('...') method variant that takes a URL string, the `Request` object constructor, and redirects will use the standard URL parser.

This is not really how I would prefer to address this since, unfortunately, this introduces a double parsing and serialization of the URL in order to minimize changes to the existing implementation. It would be nicer (and very possible) to refactor things to avoid the double parsing at some point.

Fixes: https://github.com/cloudflare/workerd/issues/1957

### DIFF
--- a/src/workerd/api/http-standard-test.js
+++ b/src/workerd/api/http-standard-test.js
@@ -1,0 +1,27 @@
+import { strictEqual } from 'node:assert';
+
+export default {
+  async fetch(req) {
+    if (req.url.endsWith('/')) {
+      return new Response(null, {
+        status: 302,
+        headers: {
+          location: ' /\t2  '
+        }
+      });
+    } else if (req.url.endsWith('/2')) {
+      return new Response("ok");
+    }
+  }
+};
+
+export const test = {
+  async test(ctrl, env) {
+
+    const resp = await env.SERVICE.fetch(' http://p\tl\na\tc\ne\th\no\tl\nd\te\nr\t/ ');
+
+    console.log(resp.url);
+
+    strictEqual(await resp.text(), "ok");
+  }
+};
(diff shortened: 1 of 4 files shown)


src/workerd/api/http-standard-test.js
```diff
@@ -0,0 +1,27 @@
+import { strictEqual } from 'node:assert';
+
+export default {
+  async fetch(req) {
+    if (req.url.endsWith('/')) {
+      return new Response(null, {
+        status: 302,
+        headers: {
+          location: ' /\t2  '
```
dom96, line 9: Am I reading this right? The standards-based parser strips tabulator characters?

dom96 (approved):

src/workerd/api/http.c++
```diff
@@ -1965,7 +1981,29 @@ jsg::Promise<jsg::Ref<Response>> handleHttpRedirectResponse(
...
+  auto redirectedLocation = ([&]() -> kj::Maybe<kj::Url> {
+    // TODO(later): This is a bit unfortunate. Per the fetch spec, we're supposed to be
+    // using standard WHATWG URL parsing to resolve the redirect URL. However, changing it
+    // now requires a compat flag. In order to minimize changes to the rest of the impl
+    // we end up double parsing the URL here, once with the standard parser to produce the
+    // correct result, and again with kj::Url in order to produce something that works with
+    // the existing code. Fortunately the standard parser is fast but it would be nice to
+    // be able to avoid the double parse at some point.
+    if (FeatureFlags::get(js).getFetchStandardUrl()) {
+      auto base = urlList.back().toString();
```
kentonv, line 1993: Should we change `urlList` to store strings rather than `kj::Url` objects?

src/workerd/api/http-standard-test.js
jasnell: Yep. tabs and newlines that are interspersed through the input are removed, as is any leading/trailing whitespace. Spaces within the input are handled by converting to `%20`, but they can also be represented at `+`

src/workerd/api/http.c++
jasnell: will make this change in a separate follow up PR

PR #2124, opened 2024-05-14
Cleanup in preparation for a v8 12.6 change

jasnell:
An API change in v8-profiler.h causes the build to complain about ambiguous API usage. This fixes things up so it's not ambiguous.

Needed for the v8 12.6 update

### DIFF
--- a/src/workerd/jsg/modules-new-test.c++
+++ b/src/workerd/jsg/modules-new-test.c++
@@ -1836,32 +1836,40 @@ KJ_TEST("Aliased modules (import maps) work") {
 
 // ======================================================================================
 
+// TODO(soon): Temporarily disabling this test. TC-39 recently changed the syntax
+// for import attributes to move away from the "assert". Prior to v8 12.6 the
+// assert keyword was still supporteed, with 12.6 it is not. Since we currently
+// update the v8 version in the internal repo separately than we do the public
+// workerd repo, there's a small period of time in which the versions will be out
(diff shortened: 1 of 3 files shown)


a-robinson (approved):

jp4a50 (approved):

src/workerd/jsg/modules-new-test.c++
```diff
@@ -1836,32 +1836,40 @@ KJ_TEST("Aliased modules (import maps) work") {
 
 // ======================================================================================
 
+// TODO(soon): Temporarily disabling this test. TC-39 recently changed the syntax
+// for import attributes to move away from the "assert". Prior to v8 12.6 the
+// assert keyword was still supporteed, with 12.6 it is not. Since we currently
```
fhanau, line 1841: FYI we have used V8_MAJOR_VERSION/V8_MINOR_VERSION before to support both a newer and an older V8 version before; this can go ahead as-is though with the test not being used for production code.

fhanau (approved):

src/workerd/jsg/modules-new-test.c++
jasnell: yep, I intentionally wanted to keep this simple, especially since it's still considered an experimental language feature and the test is non-critical. The test will need to be updated anyway so I'll do it all as part of the 12.6 update specifically for workerd.

PR #2125, opened 2024-05-14
Python: Fix header conversion to byte-pair on scope building

morgan9e:
> headers: These are byte strings of the exact byte sequences sent by the client/to be sent by the server. While modern HTTP standards say that headers should be ASCII, older ones did not and allowed a wider range of characters. Frameworks/applications should decode headers as they deem appropriate.
(https://asgi.readthedocs.io/en/latest/specs/www.html#wsgi-compatibility)

Workerd uses Pyodide JSProxy object when passing `fetch(request, env)` to Python, but JSProxy objects translates into string, not byte-string.

In order to match ASGI specification, it needs to be translated into byte-pair when building scopes.

### DIFF
--- a/src/pyodide/internal/asgi.py
+++ b/src/pyodide/internal/asgi.py
@@ -24,8 +24,10 @@ def acquire_js_buffer(pybuffer):
 
 def request_to_scope(req, env, ws=False):
     from js import URL
-
-    headers = [tuple(x) for x in req.headers]
+    # @app.get("/example")
+    # async def example(request: Request):
+    #     request.headers.get("content-type") # this will error if header is not "bytes" as in ASGI spec.
+    headers = [(k.lower().encode(), v.encode()) for k, v in req.headers]
     url = URL.new(req.url)
     assert url.protocol[-1] == ":"
     scheme = url.protocol[:-1]
@@ -111,7 +113,8 @@ async def process_request(app, req, env):
         nonlocal headers
         if got["type"] == "http.response.start":
             status = got["status"]
-            headers = got["headers"]
+            # Like above, we need to convert byte-pairs into string explicitly.
+            headers = [(k.decode(), v.decode()) for k, v in got["headers"]]
         if got["type"] == "http.response.body":
             # intentionally leak body to avoid a copy
             #
--- a/src/workerd/server/tests/python/BUILD.bazel
+++ b/src/workerd/server/tests/python/BUILD.bazel
@@ -33,6 +33,17 @@ wd_test(
     ),
 )
 
+wd_test(
+    src = "asgi/asgi.wd-test",
+    args = ["--experimental"],
+    data = glob(
+        [
+            "asgi/*",
+        ],
+        exclude = ["**/*.wd-test"],
+    ),
+)
+
 # langchain test: disabled for now because it's flaky
 # TODO: reenable this?
 #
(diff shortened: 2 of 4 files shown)


morgan9e:

hoodmane (commented):
Code change makes sense. Ideally we ought to add a test.

hoodmane (approved):

hoodmane (approved):

src/pyodide/internal/asgi.py
```diff
@@ -25,7 +25,7 @@ def acquire_js_buffer(pybuffer):
 def request_to_scope(req, env, ws=False):
     from js import URL
 
-    headers = [tuple(x) for x in req.headers]
+    headers = [(k.lower().encode(), v.encode()) for k, v in req.headers]
```
hoodmane, line 28: In lieu of adding a test, can we add a comment here with an example that cares about this being bytes so that we can add a test ourselves later?

morgan9e: It assumes FastAPI so that would be a main concern, but its for ASGI spec though so should I add for FastAPI or just asserting for spec?
 
Like, anything for 
```
@app.get("/example")
  async def example(request: Request):
    request.headers.get("content-type") # this will error if header is not bytes.
```
perhaps?

morgan9e: Also is there way to test this locally? I want to build miniflare with master workerd, but building workers-sdk doesnt works..

morgan9e: Oh Thanks, now I can test few things.

spikerwork:
> That's not the only trouble w/ ASGI and binary strings. We also need to decode resulting headers in `asgi.py`, if I understand correctly:
> 
> ```
> async def send(got):
>         ...
>         if got['type'] == 'http.response.start':
>             status = got['status']
>             headers = [(k.decode(), v.decode()) for k, v in got['headers']]
> ```

Confirmed! This patch fixed the problem with b'' I've seen before

Screenshot from 2024-06-05 11-03-09

src/pyodide/internal/asgi.py
```diff
@@ -111,7 +115,7 @@ async def send(got):
         nonlocal headers
         if got["type"] == "http.response.start":
             status = got["status"]
-            headers = got["headers"]
+            headers = [(k.decode(), v.decode()) for k, v in got['headers']]
```
leobuskin, line 118: @morgan9e I'd use double quotes here to be compliant w/ folks' formatting standards

morgan9e: Was working on it!

hoodmane:
Since #2096 merged, we're actually running Python tests on workerd so it should be possible to add a test now.

morgan9e:
> Since #2096 merged, we're actually running Python tests on workerd so it should be possible to add a test now.

Should I add test for ASGI only or if its working with Starlette?

morgan9e:
> @hoodmane: Since #2096 merged, we're actually running Python tests on workerd so it should be possible to add a test now.

Should I add test for ASGI only or if its working with Starlette? I think it might be better if we make asgi.py independant to FastAPI

hoodmane:
Agreed we should make it independent. Currently the only thing in there that is dependent on fastapi is `Env`. Perhaps we can just delete that for now and we can figure out the appropriate place to add it later.

dom96:
Hey @morgan9e, would love to get this merged in, are you still planning to create a test for this?

morgan9e:
> Hey @morgan9e, would love to get this merged in, are you still planning to create a test for this?

Yep, thanks I'll do it ASAP.

morgan9e:
Sorry, I accidently merged and removed upstream branch, I reopened it. 3142

PR #2128, opened 2024-05-16
Update to v8 12.6

jasnell:


fhanau:
Looks like the breaking bazel changes related to Google3 only affect a few python tests, so this is easier than expected 😅 . Code changes LGTM. I'll try to compile this locally to rule out ARM64 issues but will stamp it once that's done.

fhanau:
Note that V8 replaces the `trace_event_common` dependency with an internal header in this version. `trace_event_common` can be cleaned up in `WORKSPACE`, `compile_flags.txt` and `docs/v8-updates.md`

fhanau (approved):

Issue #2112, opened 2024-05-10
🚀 Feature Request — Provide the `raw-specifier` in the Module fallback service requests


dario-piotrowicz:
It would be great if the requests that the [module fallback server](https://github.com/cloudflare/workerd/pull/1423) receive could provide the original/raw/unaltered specifier exactly as it was defined in the referrer code.

___

To give more context we've been experimenting with using the module fallback service alongside the [Vite environment api](https://github.com/vitejs/vite/pull/16471) in order to lazily import cjs modules (which we could otherwise not import since vite does not perform any transformation on such code).

Meaning that, in workerd (specifically in the [`ModuleEvaluator`](https://deploy-preview-16471--vite-docs-main.netlify.app/guide/api-vite-environment.html#moduleevaluator)'s `runExternalModule` method), we receive a path such as the following from Vite:
```
/Users/me/project/node_modules/.pnpm/react@18.3.1/node_modules/react/cjs/react-jsx-dev-runtime.development.js
```
that we need to import and run.

The above works fine initially, but can easily break when the module in question includes `require`/`import` calls, such for example as:
```
var React = require('react');
```
and
```
var util = require('./utils.js');
```

The problem being that the above `require` calls produce respectively the following `specifier`s and `referrer`s:
```
 specifier: /Users/me/project/node_modules/.pnpm/react@18.3.1/node_modules/react/cjs/react
 referrer: /Users/me/project/node_modules/.pnpm/react@18.3.1/node_modules/react/cjs/react-jsx-dev-runtime.development.js
```
and
```
 specifier: /Users/me/project/node_modules/.pnpm/react@18.3.1/node_modules/react/cjs/utils
 referrer: /Users/me/project/node_modules/.pnpm/react@18.3.1/node_modules/react/cjs/react-jsx-dev-runtime.development.js
```
and from these we are unable to distinguish/differentiate the fact that `react` is referring to a dependency package and `utils` a local file (which for obvious reasons is fundamental for us).

So it would be necessary for us to also have the actual import specifier being used (i.e. `"react"` and `"./utils.js"` respectively)
(this would not introduce any breaking change if such specifier would be added to the request's search params or headers potentially as a `raw-specifier`/`original-specifier` or something along those lines).

> [!Note]
> We could check the file extension to discern which is which but that is quite brittle as libraries can also end with `.js` etc...

___

[...]

PR #2131, opened 2024-05-16
Provide the raw specifier for the module fallback service

jasnell:
Fixes: https://github.com/cloudflare/workerd/issues/2112

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -2715,7 +2715,8 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name, config::Worker::
          kj::StringPtr specifier,
          kj::Maybe<kj::String> referrer,
          jsg::CompilationObserver& observer,
-         jsg::ModuleRegistry::ResolveMethod method) mutable
+         jsg::ModuleRegistry::ResolveMethod method,
+         kj::Maybe<kj::StringPtr> rawSpecifier) mutable
             -> kj::Maybe<kj::OneOf<kj::String, jsg::ModuleRegistry::ModuleInfo>> {
       kj::Maybe<kj::String> jsonPayload;
       bool redirect = false;
@@ -2751,6 +2752,11 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name, config::Worker::
       KJ_IF_SOME(ref, referrer) {
         url.query.add(
           kj::Url::QueryParam { kj::str("referrer"), kj::mv(ref) }
+        );
+      }
+      KJ_IF_SOME(raw, rawSpecifier) {
+        url.query.add(
+          kj::Url::QueryParam { kj::str("raw"), kj::str(raw) }
         );
       }
 
--- a/src/workerd/io/worker.h
+++ b/src/workerd/io/worker.h
@@ -478,7 +478,8 @@ public:
           kj::StringPtr,
           kj::Maybe<kj::String>,
           jsg::CompilationObserver&,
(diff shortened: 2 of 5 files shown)


IgorMinar:
I took a stab at testing this using a local build from this PR and I can confirm that the `raw` parameter works as expected.

I modified @dario-piotrowicz's app to test this. You can see my changes here: https://github.com/dario-piotrowicz/miniflare-ModuleFallbackService-demo/compare/main...IgorMinar:miniflare-ModuleFallbackService-demo:workerd-pr-2131-test?expand=1

Overall, while I succeeded in getting the fallback service to work, I had to hack around a lot and I don't have confidence that this will fly in the real world, without further changes in workerd. Specifically workerd's inability to support bare specifiers and specifiers ending with `/` seems troublesome - see my `TODO(discuss)` comments in the PR above.

 Btw I do find `raw` as the param name to be too ambiguous and as you can see in my code, I resorted to calling the value `rawSpecifier` in my code — we should consider renaming it.

Interestingly, I found that workerd resolved specifier is [still useful to detect when a redirect is needed](https://github.com/dario-piotrowicz/miniflare-ModuleFallbackService-demo/pull/1/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R68), which implies that we actually need both the workerd-resolved specifier as well as the raw specifier. This surprised me, but it's likely a good news for us as we don't need to deprecate and remove the original `specifier`.

Let's discuss this a bit before this change is merged.

side note: I had to employ 301 redirects to get everything to work, but I think that is expected, as this is exactly why the fallback service supports the redirects at all.

src/workerd/server/server.c++
```diff
@@ -2753,6 +2754,11 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name, config::Worker::
           kj::Url::QueryParam { kj::str("referrer"), kj::mv(ref) }
         );
       }
+      KJ_IF_SOME(raw, rawSpecifier) {
+        url.query.add(
+          kj::Url::QueryParam { kj::str("raw"), kj::str(raw) }
```
jasnell, line 2759:
```suggestion
          kj::Url::QueryParam { kj::str("rawSpecifier"), kj::str(raw) }
```

IgorMinar (approved):
Dario and I hacked on the module fallback service with this PR and we were able to get it all working.

We came across some new bugs and ultimately failed to use : prefixed modules with the fallback service, but this is not a show stopper for us, and I'm not quite sure if the issue is worth fixing before your module registry lands. (the issue is that internally in workerd node_modules:foo when imported from /my-app/index.js is internally resolved to /my-app/node_modules:foo because the current specifier resolver doesn't treat node_modules:foo as an absolute specifier. I do recommend that you try to add this as a test to your module registry PR so that we avoid reintroducing this issue in your rewrite.

But ultimately the PR is good enough for us, and I think we should merge your PR and release it, so that we can do more through testing of this code outside of our small demos.

Our test code can be seen here: https://github.com/dario-piotrowicz/miniflare-ModuleFallbackService-demo/pull/1/files

IgorMinar:
btw I can't give you review on the c++ code as I'm not familiar with the code base, so my approval is simply just stating that from the external point of view the code does what we asked for it to do. It would be good to have someone from the runtime team to confirm the correctness of the implementation.

jasnell:
@fhanau @dom96 @garrettgu10 ... would like to get this landed soon to unblock @IgorMinar's team. Can I ask one of y'all to do a quick review on this?

fhanau (approved):
LGTM

PR #2140, opened 2024-05-19
Implement data URL support in fetch

jasnell:
As part of the effort to make our fetch impl more spec compliant and align with other runtimes... provides an implementation of `data:` URL support in fetch. The implementation here is based on the fetch spec algorithm https://fetch.spec.whatwg.org/#data-urls ... tho it is a bit more forgiving of invalid base64 inputs than the spec, which should be ok but we can tighten it up later if necessary.

We *might* want to put this behind an experimental flag for a bit. Open to opinions.

(~670 of new lines of code added of the `+838 - 6` are tests)


jasnell:
Needs a few small tweaks still to deal with test breaks.

jasnell:
This is ready for an initial review but I'm likely to tweak it a bit more before landing.

src/workerd/api/http.c++
```diff
@@ -2111,6 +2112,20 @@ jsg::Promise<jsg::Ref<Response>> fetchImplNoOutputLock(
           IoContext::NULL_CLIENT_CHANNEL, Fetcher::RequiresHostAndProtocol::YES);
     }
 
+    KJ_IF_SOME(dataUrl, DataUrl::tryParse(jsRequest->getUrl())) {
+      // If the URL is a data URL, we need to handle it specially.
+      auto data = dataUrl.getData();
+      auto headers = jsg::alloc<Headers>();
+      headers->set(jsg::ByteString(kj::str("content-type")),
+                   jsg::ByteString(dataUrl.getMimeType().toString()));
+      return js.resolvedPromise(Response::constructor(js, kj::Maybe(kj::heapArray<kj::byte>(data)),
```
jasnell, line 2121: TODO(now)

* [ ] Given that we're not reusing the `DataUrl` structure here, we can refactor this to move the underlying `data` rather than copying it.

jasnell:
> Why is this the case?

Largely because our existing base64 implementation in kj is more forgiving and not sure it's worth the effort to make it stricter for this one case. If we need to we can introduce a compat flag and add a stricter implementation

src/workerd/api/data-url.c++
```diff
@@ -0,0 +1,104 @@
...
+      return DataUrl(MimeType::PLAINTEXT_ASCII.clone(), kj::mv(decoded));
+    }
+  };
+
+  KJ_IF_SOME(pos, href.findFirst(',')) {
+    auto unparsed = href.first(pos);
+    auto data = href.slice(pos + 1);
+
+    // We need to trim leading and trailing whitespace from the mimetype
+    unparsed = trim(unparsed);
```
garrettgu10, line 78: Nit: Maybe it's better to name these variables the same as what they're in the fetch standard, just to make it easy to ensure that we're following the spec to a T.

src/workerd/api/data-url-test.c++
```diff
@@ -0,0 +1,632 @@
...
+
+    auto url = KJ_ASSERT_NONNULL(DataUrl::tryParse(input));
+
+    KJ_ASSERT(url.getData() == test.expected, test.input);
+  }
+
+}
+
+KJ_TEST("Large Data URL") {
+  auto str = kj::str("a b c d e f g h i j k l m n o p q r s t u v w x y z ",
```
garrettgu10, line 553: Nit: can we have less repetition here?

src/workerd/api/data-url-test.c++
```diff
@@ -0,0 +1,632 @@
...
+    }
+  }
+}
+
+struct Base64Test {
+  kj::StringPtr input;
+  kj::Array<kj::byte> expected;
+};
+
+KJ_TEST("DataUrl base64") {
```
garrettgu10, line 453: Did we get these test cases from somewhere or did we write them ourselves? Can we add a comment or test name that specifies this?

garrettgu10 (approved):
LGTM, I'm convinced the implementation follows the spec perfectly & testing appears robust. Thanks!

src/workerd/api/data-url.c++
jasnell: I tend to prefer not, to be honest. I'm not always a big fan of how the specs name these things and prefer the names here.

PR #2143, opened 2024-05-20
Fix JsRpcPromise::Resolved context checking.

kentonv:
`IoContext::WeakRef` is a KJ I/O object, so `kj::Own<IoContext::WeakRef>` is not safe to hold from a JSG object. Instead, it would need to be `kj::IoOwn<IoContext::WeakRef>`. But, at that point we don't even need the `WeakRef` anymore because dereferencing the `IoOwn` itself will do exactly the check that we were using the `WeakRef` to do.

Actually, all we really need here is an `IoPtr<T>` (the type of `T` is irrelevant) which we can attempt to dereference in order to effect the context check we want.

### DIFF
--- a/src/workerd/api/worker-rpc.h
+++ b/src/workerd/api/worker-rpc.h
@@ -213,8 +213,11 @@ private:
   struct Resolved {
     jsg::Value result;
 
-    // We only use this to prohibit use from the wrong context.
-    kj::Own<IoContext::WeakRef> ioCtx;
+    // Dummy IoPtr to self, used only to verify that we're running in the correct context.
+    // (Dereferencing from the wrong context would throw an exception.)
+    // Note: Can't use IoContext::WeakRef here because it's not thread-safe (it's only intended to
+    //   be helf from KJ I/O objects, but this is a JSG object).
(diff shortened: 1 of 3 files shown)


src/workerd/api/worker-rpc.h
```diff
@@ -213,8 +213,11 @@ class JsRpcPromise: public JsRpcClientProvider {
   struct Resolved {
     jsg::Value result;
 
-    // We only use this to prohibit use from the wrong context.
-    kj::Own<IoContext::WeakRef> ioCtx;
+    // Dummy IoPtr to self, used only to verify that we're running in the correct context.
+    // (Dereferencing from the wrong context would throw an exception.)
+    // Note: Can't use IoContext::WeakRef here because it's not thread-safe (it's only intended to
+    //   be helf from KJ I/O objects, but this is a JSG object).
```
jasnell, line 219:
```suggestion
    //   be held from KJ I/O objects, but this is a JSG object).
```

jasnell (approved):
Oh nice.. I kind of expected it would be something along these lines.

PR #2144, opened 2024-05-20
V8 Updates: Migrate to v8::Object::Wrap/Unwrap

jasnell:
v8 is simplifying the way embedder apis connect to wrapper JS objects.

This attempts to keep the necessary changes as minimal as possible.

Note that with the move to `v8::Object::Wrap/Unwrap`, we ought to be able to eliminate the special handling of the context global wrapper and simplify things a bit further, but that's bit of a larger change. Will save that for a separate PR.

### DIFF
--- a/src/workerd/jsg/ser.c++
+++ b/src/workerd/jsg/ser.c++
@@ -103,8 +103,7 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
     jsg::Lock& js = jsg::Lock::from(isolate);
 
     if (object->InternalFieldCount() != Wrappable::INTERNAL_FIELD_COUNT ||
-        object->GetAlignedPointerFromInternalField(Wrappable::WRAPPABLE_TAG_FIELD_INDEX) !=
-            &Wrappable::WRAPPABLE_TAG) {
+        !Wrappable::isWorkerdApiObject(object)) {
       KJ_IF_SOME(eh, externalHandler) {
         if (object->IsFunction()) {
           eh.serializeFunction(js, *this, object.As<v8::Function>());
--- a/src/workerd/jsg/wrappable.h
+++ b/src/workerd/jsg/wrappable.h
@@ -47,29 +47,31 @@ class HeapTracer;
 // Wrappable and are not visible to GC tracing.
 class Wrappable: public kj::Refcounted {
 public:
-  // Number of internal fields in a wrapper object.
-  static constexpr uint INTERNAL_FIELD_COUNT = 3;
 
-  // Index of the internal field that points back to the `Wrappable`.
-  static constexpr uint WRAPPED_OBJECT_FIELD_INDEX = 2;
+  enum InternalFields : int {
+    // Field must contain a pointer to `WORKERD_WRAPPABLE_TAG`. This is a workerd-specific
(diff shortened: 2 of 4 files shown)


src/workerd/jsg/ser.c++
```diff
@@ -100,8 +100,7 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
     jsg::Lock& js = jsg::Lock::from(isolate);
 
     if (object->InternalFieldCount() != Wrappable::INTERNAL_FIELD_COUNT ||
-        object->GetAlignedPointerFromInternalField(Wrappable::WRAPPABLE_TAG_FIELD_INDEX) !=
-            &Wrappable::WRAPPABLE_TAG) {
+        !object->IsApiWrapper()) {
```
kentonv, line 103: Hmm, this is only equivalent to the old code if _all_ API objects are `Wrappable` instances, that is, JSG is the only one creating API wrappers. But I'm not sure if this is true -- I think V8 itself might implement some types as API wrappers internally? Even if not, I don't want to preclude being able to bring in other libraries that implement V8 APIs in their own way.

Can we somehow check not just that it's a wrapper, but that it's _our_ wrapper? Probably by checking the tag?

src/workerd/jsg/wrappable.h
```diff
@@ -47,29 +47,16 @@ class HeapTracer;
...
-  // objects of its own which have internal fields, and then GC doesn't check that the index is
-  // in-bounds before reading it...
-  static constexpr uint WRAPPABLE_TAG_FIELD_INDEX = 0;
-
-  // The value pointed to by the internal field field `WRAPPABLE_TAG_FIELD_INDEX`.
-  //
-  // This value was chosen randomly.
-  static constexpr uint16_t WRAPPABLE_TAG = 0xeb04;
+  static constexpr v8::CppHeapPointerTag WRAPPABLE_TAG =
+      v8::CppHeapPointerTag::kDefaultTag;
```
kentonv, line 59: We should probably have our own tag so that we can check if objects are our wrappables (see other comment).

src/workerd/jsg/ser.c++
jasnell: We might have to re-introduce an API api for that as the tag is not exposed in a way that is queryable. Also, the tag is now actually from the `CppHeapPointerTag` enum class which can also used internally by v8.

src/workerd/jsg/wrappable.h
jasnell: Doing so would require a patch to v8 to modify the `CppHeapPointerTag` enum class. Certainly possible to do but I was rather hoping to avoid adding another floated patch

kentonv: Oh, I see the internal enum `v8::internal::ExternalPointerTag` actually has quite a few different values and the integer values actually encode structured information in the bits... huh...

So obviously we _can't_ actually extend this enum.

src/workerd/jsg/ser.c++
kentonv: Perhaps we need to continue to have two internal pointer fields, one of which continues to work like `WRAPPABLE_TAG`, but we only use it to identify our own objects...

jasnell: Done

Issue #2133, opened 2024-05-16
Introduce --nodejs_compat_unleashed runtime flag


IgorMinar:
As we improve our node.js compat story, it would be great have a runtime flag that offers a more "aggressive" and compatible version of `--nodejs_compat` runtime flag. We could call this flag `--node_compat_unleashed` or similar.

This new flag should expose supported node builtin modules without the `node:` prefix, and should also expose globals like `process`, `Buffer` on `globalThis`.

We can then test with this flag, and at some point in the near future, this flag would be activated via a compat date as long as `nodejs_compat` is present (essentially making the existing `nodejs_compat` become `nodejs_compat_unleashed` once the compatibility date is reached or exceeded.

Originally proposed in https://chat.google.com/room/AAAAZwUYa48/4rD9Yjauj9w/yE5EU56IaRg?cls=10

kentonv:
Bikeshed comment: The name "unleashed" sounds a little, uh, silly to me. Maybe `nodejs_compat_no_prefix` or `_no_namespace` or something?

IgorMinar:
I'm not attached to the name. The flag should in addition to unprefixing modules also introduce node's globals, so `no_prefix` seems not comprehensive enough.

jasnell:
heh, yeah, I wasn't going to go with "unleashed" ;-) ... I'll be working on this in the coming couple of weeks, want to get the `Request.cache` stuff done first, then will do this next.

jasnell:
Initial draft implementation in #2147

PR #2147, opened 2024-05-21
Node.js Compat: Implement the nodejs_compat_v2 compat flag

jasnell:
The `nodejs_compat_v2` flag supersedes the 'nodejs_compat' flag. When enabled...

1. Node.js built-ins are available for import/require
2. Unlike the original `nodejs_compat` flag, Node.js imports do not require the 'node:' specifier prefix. Internally, the implementation will detect a Node.js raw specifier and convert it into the appropriate prefixed specifier, e.g. 'fs' becomes 'node:fs'
3. The `Buffer`, `process`, and `global` are exposed on the global
4. A user worker bundle can still provide its own implementations of all `node:` modules which will take precedence over the built-ins
5. The new `process.getBuiltinModule(...)` API is implemented. See https://github.com/nodejs/node/pull/52762

A worker can replace the implementation of `node:process` if they choose, which may mean that `getBuiltinModule(...)` is not available.

Refs: https://github.com/cloudflare/workerd/issues/2133

### DIFF
--- a/src/node/internal/process.ts
+++ b/src/node/internal/process.ts
@@ -16,6 +16,8 @@ import {
   ERR_INVALID_ARG_VALUE,
 } from 'node-internal:internal_errors'
 
+import { default as utilImpl } from 'node-internal:util';
+
 export function nextTick(cb: Function, ...args: unknown[]) {
   queueMicrotask(() => { cb(...args); });
 };
@@ -62,7 +64,12 @@ export const env = new Proxy({}, {
   }
 });
 
+export function getBuiltinModule(id: string) : any {
+  return utilImpl.getBuiltinModule(id);
+}
+
 export default {
   nextTick,
   env,
+  getBuiltinModule,
 };
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -777,6 +777,56 @@ void ServiceWorkerGlobalScope::reportError(jsg::Lock& js, jsg::JsValue error) {
   }
 }
 
+namespace {
+jsg::JsValue resolveFromRegistry(jsg::Lock& js, kj::StringPtr specifier) {
+  auto moduleRegistry = jsg::ModuleRegistry::from(js);
+  if (moduleRegistry == nullptr) {
+    // TODO: Return the known built-in instead? This gets a bit tricky as we currently
+    // have no mechanism defined for accessing and caching the built-in module without
+    // the module registry. Should we even support this case at all? Without the module
+    // registry the user can't access the other importable modules anyway. For now, just
+    // returning undefined in this case seems the most appropriate.
+    return js.undefined();
+  }
+
+  auto spec = kj::Path::parse(specifier);
+  auto& info = JSG_REQUIRE_NONNULL(moduleRegistry->resolve(js, spec), Error,
+                                   kj::str("No such module: ", specifier));
+  auto module = info.module.getHandle(js);
+  jsg::instantiateModule(js, module);
+
+  auto handle = jsg::check(module->Evaluate(js.v8Context()));
+  KJ_ASSERT(handle->IsPromise());
+  auto prom = handle.As<v8::Promise>();
+  KJ_ASSERT(prom->State() != v8::Promise::PromiseState::kPending);
+  if (module->GetStatus() == v8::Module::kErrored) {
+    jsg::throwTunneledException(js.v8Isolate, module->GetException());
+  }
+  return jsg::JsValue(js.v8Get(module->GetModuleNamespace().As<v8::Object>(), "default"_kj));
+}
+}  // namespace
+
+jsg::JsValue ServiceWorkerGlobalScope::getBuffer(jsg::Lock& js) {
+  KJ_ASSERT(FeatureFlags::get(js).getNodeJsCompatV2());
+  auto value = resolveFromRegistry(js, "node:buffer"_kj);
+  auto obj = JSG_REQUIRE_NONNULL(value.tryCast<jsg::JsObject>(), TypeError,
+                                 "Invalid node:buffer implementation");
+  // Unlike the getProcess() case below, this getter is returning an object that
+  // is exported by the node:buffer module and not the module itself, so we need
+  // this additional get to the grab the reference to the thing we're actually
+  // returning.
+  auto buffer = obj.get(js, "Buffer"_kj);
+  JSG_REQUIRE(buffer.isFunction(), TypeError, "Invalid node:buffer implementation");
+  return buffer;
+}
+
+jsg::JsValue ServiceWorkerGlobalScope::getProcess(jsg::Lock& js) {
+  KJ_ASSERT(FeatureFlags::get(js).getNodeJsCompatV2());
+  auto value = resolveFromRegistry(js, "node:process"_kj);
+  JSG_REQUIRE(value.isObject(), TypeError, "Invalid node:process implementation");
+  return value;
+}
+
 double Performance::now() {
   // We define performance.now() for compatibility purposes, but due to spectre concerns it
   // returns exactly what Date.now() returns.
--- a/src/workerd/api/node/node-compat-v2-test.js
+++ b/src/workerd/api/node/node-compat-v2-test.js
@@ -1,0 +1,64 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+//
+
+// Imports of Node.js built-ins should work both with and without
+// the 'node:' prefix.
+import { default as assert } from 'node:assert';
+import { default as assert2 } from 'assert';
+const assert3 = (await import('node:assert')).default;
+const assert4 = (await import('assert')).default;
+
+assert.strictEqual(assert, assert2);
+assert.strictEqual(assert, assert3);
+assert.strictEqual(assert, assert4);
+
+export const nodeJsExpectedGlobals = {
+  async test() {
+    // Expected Node.js globals Buffer, process, and global should be present.
+    const { Buffer } = await import('node:buffer');
+    assert.strictEqual(Buffer, globalThis.Buffer);
+
+    const { default: process } = await import('node:process');
+    assert.strictEqual(process, globalThis.process);
+
+    assert.strictEqual(global, globalThis);
+  }
+};
+
+export const nodeJsGetBuiltins = {
+  async test() {
+    // node:* modules in the worker bundle should override the built-in modules...
+    const { default: fs } = await import('node:fs');
+    const { default: path } = await import('node:path');
+
+    await import ('node:path');
+
+    // But process.getBuiltinModule should always return the built-in module.
+    const builtInPath = process.getBuiltinModule('node:path');
+    const builtInFs = process.getBuiltinModule('node:fs');
+
+    // These are from the worker bundle....
+    assert.strictEqual(fs, 1);
+    assert.strictEqual(path, 2);
+
+    // But these are from the built-ins...
+    // node:fs is not implemented currently so it should be undefined here.
+    assert.strictEqual(builtInFs, undefined);
+
+    // node:path is implemented tho...
+    assert.notStrictEqual(path, builtInPath);
+    assert.strictEqual(typeof builtInPath, 'object');
+    assert.strictEqual(typeof builtInPath.join, 'function');
+
+    // While process.getBuiltinModule(...) in Node.js only returns Node.js
+    // built-ins, our impl will return cloudflare: and workerd: built-ins
+    // also, for completeness. A key difference, however, is that for non-Node.js
+    // built-ins, the return value is the module namespace rather than the default
+    // export.
+
+    const socket = await import('cloudflare:sockets');
+    assert.strictEqual(process.getBuiltinModule('cloudflare:sockets'), socket);
+  }
+};
--- a/src/workerd/api/node/node-compat-v2-test.wd-test
+++ b/src/workerd/api/node/node-compat-v2-test.wd-test
@@ -1,0 +1,17 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "node-compat-v2-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "node-compat-v2-test.js"),
+          (name = "node:fs", esModule = "export default 1"),
+          (name = "node:path", esModule = "export default 2"),
+        ],
+        compatibilityDate = "2024-05-01",
+        compatibilityFlags = ["nodejs_compat_v2", "experimental"]
+      )
+    ),
+  ],
+);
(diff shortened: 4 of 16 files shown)


IgorMinar:
@jasnell could you also expose `global` on `globalThis` as an alias of `globalThis`?

Essentially:
```js
globalThis.global = globalThis;
```

IgorMinar (approved):
I can't judge the implementation soundness, but I was able to play with this and can confirm that with the `nodejs_compat_v2` flag `Buffer` and `process` are now exposed, however `global` is not. Could you please add that?

I was also able to do some basic experiments with `process.getBuiltinModule` which also seems to work as expected. I did not test if overwriting a built-in using a user-provided module results in no change for the getBuiltinModule - this would be a good thing to add to an automated test.

Can you please add a test or sample for this functionality?

Otherwise this change looks good to me and will unblock our work on hybrid polyfills (https://github.com/unjs/unenv/issues/181). Thank you for a speedy turnaround.

To recap the todos:
- [ ] expose `global`
- [ ] add tests

FYI here is my test code:
```js
import path from 'path';
import npath from 'node:path';

export default {
  async fetch(req, env) {
    return new Response("Hello World\n" +
        (('Buffer' in globalThis) ? Buffer : "xBufferx") + '\n' +
        (('process' in globalThis) ? JSON.stringify(Object.keys(process)) : 'xprocessx') + '\n' +
        (JSON.stringify(Object.keys(process.getBuiltinModule('path')))) + '\n' +
        (JSON.stringify(Object.keys(path))) + '\n' +
        (JSON.stringify(Object.keys(npath))) + '\n'+
        (('global' in globalThis) ? global : "xglobalx") + '\n'
    );
  }
};
```

jasnell:
The `getBuiltinModule(...)` implementation is not actually working correctly yet as verified by a test. Going to require just a bit more work to get right.

jasnell:
@fhanau @harrishancock @dom96 ... this change is needed by the frameworks/wrangler folks to move forward on evolving the node.js compat story.

@irvinebroque @mikenovitch ... heads up from the product perspective this changes the way node.js compat is enabled. There's the existing `nodejs_compat` flag for the original version, and then a new `nodejs_compat_v2` flag to enable this variation. Whee!

IgorMinar:
> @irvinebroque @mikenomitch ... heads up from the product perspective this changes the way node.js compat is enabled. There's the existing nodejs_compat flag for the original version, and then a new nodejs_compat_v2 flag to enable this variation. Whee!

Is the plan to make `nodejs_compat_v2` become equivalent with `nodejs_compat` after a certain compat date? That would be my preference as I want to avoid a proliferation of compat flags. I see `nodejs_compat_v2` just as a temporary vehicle to test the changes we plan to make to `nodejs_compat` which will be enabled at a set compat date.

src/workerd/api/node/node-compat-v2-test.wd-test
```diff
@@ -0,0 +1,17 @@
...
+  services = [
+    ( name = "node-compat-v2-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "node-compat-v2-test.js"),
+          (name = "node:fs", esModule = "export default 1"),
+          (name = "node:path", esModule = "export default 2"),
+        ],
+        compatibilityDate = "2024-05-01",
+        compatibilityFlags = ["nodejs_compat_v2", "experimental"]
```
IgorMinar, line 13: is it sufficient to just add `"experimental"` into copatiblityFlags array? I tried this a modified version of one of the existing samples and got the following:

```
bazel-bin/src/workerd/server/workerd serve ./samples/helloworld_esm/config.capnp 
service main: The compatibility flag experimental is experimental and may break or be removed in a future version of workerd. To use this flag, you must pass --experimental on the command line.
```

only `bazel-bin/src/workerd/server/workerd serve ./samples/helloworld_esm/config.capnp --experimental` worked for me.

IgorMinar: my config.capnp looked like this:
```
  modules = [
  ],
  compatibilityDate = "2024-05-12",
  compatibilityFlags = ["nodejs_compat", "nodejs_compat_v2", "experimental"]
);

```

jasnell:
> Is the plan to make nodejs_compat_v2 become equivalent with nodejs_compat after a certain compat date? That would be my preference as I want to avoid a proliferation of compat flags. I see nodejs_compat_v2 just as a temporary vehicle to test the changes we plan to make to nodejs_compat which will be enabled at a set compat date.

Unfortunately I don't think so. Given the way compat dates work and the guarantees they carry, we can't really combine these at any point and we had made the decision long ago that Node.js compat would not be on-by-default ever. If we move forward with adding this, it'll mean a second node.js compat mode going forward indefinitely. This is something we should discuss tho to see if there's a reasonable path forward.

src/workerd/api/node/node-compat-v2-test.wd-test
jasnell: The `--experimental` label is there only while this is being evaluated. I plan to remove it before landing if we do decide to move forward on this. I'd like to make sure we get input from @irvinebroque and @mikenomitch

IgorMinar:
@jasnell sorry if I was not clear, I'm not suggesting that `nodejs_compat_v2` becomes activated on by default after some compat date. What I'm suggesting instead is that after some compat date `nodejs_compat` and `nodejs_compat_v2` mean the same thing (and `nodejs_compat_v2` becomes unnecessary). This would reduce the number of flags we need to document and explain to developers as well as number of branches in our codebase.

src/workerd/api/node/node-compat-v2-test.wd-test
IgorMinar: I think I didn't make myself clear, what I'm puzzled about is that you are able to enable the experimental mode via a runtime flag in the config, but for me this didn't work and I had to explicitly use `--experimental` command line flag 
instead. Is that expected?

I don't know how this stuff works, and I tried doing what you did here at first but it didn't work, so when I later saw you using this approach in this PR I was surprised.

jasnell:
Even then, I'm not sure we can combine `nodejs_compat` and `nodejs_compat_v2` unfortunately. This PR introduces a few changes that could be technically breaking even in the `nodejs_compat` case currently. We'll need to evaluate carefully.

jasnell:
> ... is the main breaking change the new globals that are added?

I'll say "potentially" breaking change are the new globals and the fact that statements like `import('buffer')` will suddenly start working where they fail now. As we've seen with other recent changes, folks like to do feature-sniffing by trying an action and revering to some fall back logic when it doesn't work. Having it so that the non-prefixed specifiers suddenly start working could cause unexpected side effects that we need to be careful of.

src/workerd/io/compatibility-date.capnp
```diff
@@ -441,4 +441,13 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
   # Ensures that WHATWG standard URL parsing is used in the fetch API implementation.
+
+  nodeJsCompatV2 @50 :Bool
+      $compatEnableFlag("nodejs_compat_v2")
+      $compatDisableFlag("no_nodejs_compat_v2")
+      $experimental;
+  # Implies nodeJSCompat with the following additional modifications:
+  # * Node.js Compat built-ins may be imported/required with or without the node: prefix
+  # * Node.js Compat common globals are available everywhere (e.g. Buffer, process)
+  # * node:process cannot be replaced by the worker bundle
```
fhanau, line 452: Where is this implemented?

src/workerd/api/node/node-compat-v2-test.js
```diff
@@ -0,0 +1,64 @@
...
+    // But process.getBuiltinModule should always return the built-in module.
+    const builtInPath = process.getBuiltinModule('node:path');
+    const builtInFs = process.getBuiltinModule('node:fs');
+
+    // These are from the worker bundle....
+    assert.strictEqual(fs, 1);
+    assert.strictEqual(path, 2);
+
+    // But these are from the built-ins...
+    // node:fs is not implementated currently so it should be undefined here.
```
fhanau, line 47: nit: spelling

src/node/internal/process.ts
```diff
@@ -62,7 +64,12 @@ export const env = new Proxy({}, {
   }
 });
 
+export function getBuiltinModule(id : string) : any {
```
fhanau, line 67: nit: fix spaces

src/workerd/io/compatibility-date.capnp
```diff
@@ -441,4 +441,13 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
     $compatDisableFlag("fetch_legacy_url")
     $compatEnableDate("2024-06-03");
   # Ensures that WHATWG standard URL parsing is used in the fetch API implementation.
+
+  nodeJsCompatV2 @50 :Bool
```
fhanau, line 445: For visibility, should we place this next to the existing `nodeJsCompat` flag?

src/workerd/api/global-scope.c++
```diff
@@ -777,6 +777,48 @@ void ServiceWorkerGlobalScope::reportError(jsg::Lock& js, jsg::JsValue error) {
   }
 }
 
+namespace {
+jsg::JsValue resolveFromRegistry(jsg::Lock& js, kj::StringPtr specifier) {
+  auto moduleRegistry = jsg::ModuleRegistry::from(js);
+  if (moduleRegistry == nullptr ) {
+    // TODO: Return the known built-in instead?
```
fhanau, line 784: Decide how to approach this TODO, there's also minor lint in the line above.

fhanau:
LGTM otherwise

src/workerd/io/compatibility-date.capnp
jasnell: oh right... decided not to go with this. Just need to remove this bullet point

src/workerd/io/compatibility-date.capnp
jasnell: I'd rather keep the ordinal values in order to make it easier to know which is the next available value.

src/node/internal/process.ts
jasnell: For stuff like this... using the "Add a suggestion" option in the github editor is easier.

src/workerd/api/global-scope.c++
```diff
@@ -777,6 +777,48 @@ void ServiceWorkerGlobalScope::reportError(jsg::Lock& js, jsg::JsValue error) {
   }
 }
 
+namespace {
+jsg::JsValue resolveFromRegistry(jsg::Lock& js, kj::StringPtr specifier) {
+  auto moduleRegistry = jsg::ModuleRegistry::from(js);
+  if (moduleRegistry == nullptr ) {
```
dom96:
```suggestion
  if (moduleRegistry == nullptr) {
```

src/workerd/jsg/modules.h
```diff
@@ -441,12 +444,8 @@ class ModuleRegistryImpl final: public ModuleRegistry {
 
   void addBuiltinModule(Module::Reader module) {
     if (module.which() != Module::SRC) {
-      using Key = typename Entry::Key;
       auto specifier = module.getName();
       auto path = kj::Path::parse(specifier);
-      if (module.getType() == Type::BUILTIN && entries.find(Key(path, Type::BUNDLE)) != kj::none) {
```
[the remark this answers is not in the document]
jasnell: Previously, we would skip registering the built-in module at all if the worker bundle itself included module with the same name. That was really just an optional optimization. With the introduction of `process.getBuiltinModule(...)`, we want the built-ins to be registered and accessible even if the worker bundle shadows it with its own implementation.

src/workerd/api/node/node-compat-v2-test.js
```diff
@@ -0,0 +1,64 @@
...
+    const { default: process } = await import('node:process');
+    assert.strictEqual(process, globalThis.process);
+
+    assert.strictEqual(global, globalThis);
+  }
+};
+
+export const nodeJsGetBuiltins = {
+  async test() {
+    // node:* modules in the worker bundle should override the built-in modules...
```
dom96, line 32: Does the same apply to non-`node:` prefixed modules? Do we want tests for it too?

src/workerd/io/compatibility-date.capnp
```diff
@@ -441,4 +441,12 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatEnableFlag("nodejs_compat_v2")
+      $compatDisableFlag("no_nodejs_compat_v2")
+      $experimental;
+  # Implies nodeJSCompat with the following additional modifications:
+  # * Node.js Compat built-ins may be imported/required with or without the node: prefix
+  # * Node.js Compat the globals Buffer and process are available everywhere
```
dom96, line 451:
```suggestion
  # * The Node.js globals Buffer and process are available everywhere
```

src/workerd/io/worker.c++
```diff
@@ -1018,6 +1018,9 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
 
     lock->setCaptureThrowsAsRejections(features.getCaptureThrowsAsRejections());
     lock->setCommonJsExportDefault(features.getExportCommonJsDefaultNamespace());
+    if (features.getNodeJsCompatV2()) {
+      lock->setNodeJsCompatEnabled();
```
dom96, line 1022: Should `setNodeJsCompatEnabled` mention the "v2"?

src/workerd/jsg/jsg.h
```diff
@@ -2296,6 +2296,8 @@ class Lock {
   void setCaptureThrowsAsRejections(bool capture);
   void setCommonJsExportDefault(bool exportDefault);
 
+  void setNodeJsCompatEnabled();
```
dom96, line 2299: A comment on what this enables (v1. vs. v2) would be nice here

src/workerd/io/compatibility-date.capnp
dom96: Also, why is `Buffer` and `process` only made available? Are we planning to make more of the builtins available too?

dom96 (approved):

fhanau (approved):

src/workerd/api/node/node-compat-v2-test.js
jasnell: It does but tests for that should be added as a separate PR.

src/workerd/io/compatibility-date.capnp
jasnell: These are the only Node.js specific global properties that we implement. Most of the remaining globals supported by Node.js are either web platform standard APIs we already implement/export, are ones we don't implement (e.g. `BroadcastChannel`), or that don't make sense in our environment, e.g. `__filename` and `__dirname`.

src/workerd/io/worker.c++
jasnell: No, eventually the goal is to combine the two compat flags if we can. In either case, there's no reason to version this particular API.

src/workerd/jsg/jsg.h
jasnell: At this level it doesn't matter as jsg really has (and should not have) any information directly relating to compatibility flags. This is just a JSG specific configuration option that the compat flag happens to enable. We don't want any specific versioning on this.

PR #2153, opened 2024-05-22
[CI][build] Improve CI configuration to mitigate resource constraints

fhanau:
- Disable the macOS-debug runner for now to avoid macOS build queueing unless we have a proper way to approach this.
- For macOS fastbuild, don't define DEBUG to match Linux configuration
- Drop the nonsensical experimental_objc_fastbuild_options flag.
- Don't try to re-enable assertions for Windows fastbuild – this never took effect as the feature was spelled wrong (needs to be just "disable_assertions"), actually using assertions would likely make the Windows build slower (it is already slower than the macOS/Linux jobs).
- Drop flags to disable optimization in the host configuration for Windows fastbuild. Host tools (e.g. capnpc for generating .capnp.h files) use 'opt' mode by default and we want them to run quickly.

Only apply -Wl,-O2 under release configuration
We turned on linker string tail merging as enabled by lld's -O2 flag by default after experiencing binary size constraints on CI, but this can slow down linking. Only enable it for release builds.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -176,6 +176,14 @@ build:macos --macos_minimum_os=11.5 --host_macos_minimum_os=11.5
 # information on the flag.
 build:macos --copt='-femit-dwarf-unwind=no-compact-unwind'
 
+# Bazel doesn't allow setting flags only for a specific build mode like fastbuild, but somehow has
+# a flag that does this for ObjC to set "-O0 -DDEBUG=1". Worse yet, options set by this flag cannot
+# be overwritten on the command line, which confusingly makes enabling optimization there not have
+# an effect. On macOS we build some ObjC as part of Dawn, set this flag to an empty string so we
+# can set the right configuration flags.
+# TODO(soon): File a bazel issue for the flag so we can drop this.
+build:macos --experimental_objc_fastbuild_options=''
+
 # Unfortunately experimental_objc_fastbuild_options does not respect options set before and
 # overwrites them. This causes mysterious binary size bloat for fastbuild-based configurations on
 # macOS when enabling some optimizations. Turn it in into a no-op.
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -129,6 +133,16 @@ jobs:
           build:asan --config=limit-storage
           build:debug --config=limit-storage
           EOF
+      - name: Fix macOS fastbuild configuration
+        # Unlike the bazel Unix toolchain the macOS toolchain sets "-O0 -DDEBUG" for fastbuild by
+        # default. This is unhelpful for compile speeds and test performance, remove the DEBUG
+        # define.
+        if: ${{ !(contains(matrix.config.suffix, 'debug') || contains(matrix.config.suffix, 'asan')) }}
```
garrettgu10, line 140: Why is this condition necessary? I thought we were making it so that macos didn't have debug or asan configs?

.bazelrc
```diff
@@ -194,7 +202,7 @@ build:linux --force_pic
 
 # On Linux, garbage collection sections and optimize binary size. These do not apply to the macOS
 # toolchain.
-build:linux --linkopt="-Wl,--gc-sections" --linkopt="-Wl,-O2"
+build:linux --linkopt="-Wl,--gc-sections"
```
garrettgu10, line 205: Is this change mentioned in the PR description?

garrettgu10 (approved):
LGTM, thanks for optimizing the CI config, any improvement here can have a huge impact on dev productivity. 🚀

.github/workflows/test.yml
fhanau: You're right, with this PR both the debug and asan macOS jobs will be disabled. I think it's future proof in case we enable these runners again in the future though. The flags themselves are being added here https://github.com/bazelbuild/apple_support/blob/facfb550e423d2a8b8c7835bd31e9d1d6af51f14/crosstool/cc_toolchain_config.bzl#L2024

.bazelrc
fhanau: Good catch. It was mentioned in the description of the second commit, which I added after the original PR. Updated the PR description too.

PR #2154, opened 2024-05-23
Add `highWaterMark` option to `Socket` for backpressure signaling

jasnell:
Adds a `highWaterMark` to `SocketOptions` that is used to configure the high water mark for the socket `WritableStream`. When the size of the internal pending write queue reaches this value, the stream will signal backpressure to the user.


vickykont (approved):
Looks good to me!

PR #2155, opened 2024-05-23
Bug Fix: Fixup WritableStream backpressure signaling, add warning

jasnell:
Fixes a bug in the internal WritableStream impl that would cause it to signal backpressure too often, also add a warning if a WritableStream excessively exceeds the configured highwatermark (if any). "Excessive" here is defined as a multiple of the high water mark.

### DIFF
--- a/src/workerd/api/streams/internal.c++
+++ b/src/workerd/api/streams/internal.c++
@@ -958,7 +958,28 @@ void WritableStreamInternalController::increaseCurrentWriteBufferSize(
     uint64_t amount) {
   currentWriteBufferSize += amount;
   KJ_IF_SOME(highWaterMark, maybeHighWaterMark) {
-    updateBackpressure(js, highWaterMark - currentWriteBufferSize <= 0);
+    int64_t amount = highWaterMark - currentWriteBufferSize;
+    updateBackpressure(js, amount <= 0);
+    // If the current buffer size is greater than or equal to double the high water mark,
+    // let's emit a warning about excessive backpressure.
+    // TODO(later): For the standard stream, we use a variable multiplier if the highWaterMark
+    // is < 10 because the default high water mark is 1 and we don't want to emit the warning
+    // too often. For internal streams, tho, there is no default high water mark and the user
+    // would have to provide one... and since these are always bytes it would make sense
+    // for the user to specify a larger value here in the typical case... so I decided to go with
+    // the fixed 2x multiplier. However, I can make this variable too if folks feel the consistency
+    // is important.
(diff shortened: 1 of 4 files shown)


src/workerd/api/streams/internal.c++
```diff
@@ -958,7 +958,21 @@ void WritableStreamInternalController::increaseCurrentWriteBufferSize(
     uint64_t amount) {
   currentWriteBufferSize += amount;
   KJ_IF_SOME(highWaterMark, maybeHighWaterMark) {
-    updateBackpressure(js, highWaterMark - currentWriteBufferSize <= 0);
+    int64_t amount = highWaterMark - currentWriteBufferSize;
+    updateBackpressure(js, amount <= 0);
+    // If the current buffer size is greater than or equal to double the high water mark,
+    // let's emit a warning about excessive backpressure.
+    if (warnAboutExcessiveBackpressure && (currentWriteBufferSize >= 2 * highWaterMark)) {
```
jasnell, line 965: For the standard stream, we use a variable multiplier if the highWaterMark is < 10 because the default high water mark is 1 and we don't want to emit the warning too often. For internal streams, tho, there is no default high water mark and the user would have to provide one... and since these are always bytes it would make sense for the user to specify a larger value here in the typical case... so I decided to go with the fixed 2x multiplier. However, I can make this variable too if folks feel the consistency is important.

src/workerd/api/streams/internal.c++
```diff
@@ -958,7 +958,21 @@ void WritableStreamInternalController::increaseCurrentWriteBufferSize(
     uint64_t amount) {
   currentWriteBufferSize += amount;
   KJ_IF_SOME(highWaterMark, maybeHighWaterMark) {
-    updateBackpressure(js, highWaterMark - currentWriteBufferSize <= 0);
+    int64_t amount = highWaterMark - currentWriteBufferSize;
+    updateBackpressure(js, amount <= 0);
+    // If the current buffer size is greater than or equal to double the high water mark,
+    // let's emit a warning about excessive backpressure.
```
jasnell, line 964:
```suggestion
    // let's emit a warning about excessive backpressure.
    // TODO(later): For the standard stream, we use a variable multiplier if the highWaterMark
    // is < 10 because the default high water mark is 1 and we don't want to emit the warning
    // too often. For internal streams, tho, there is no default high water mark and the user
    // would have to provide one... and since these are always bytes it would make sense
    // for the user to specify a larger value here in the typical case... so I decided to go with
    // the fixed 2x multiplier. However, I can make this variable too if folks feel the consistency
    // is important.
```

PR #2156, opened 2024-05-23
V8 Updates: Replace deprecated SetAccessor with SetAccessorProperty

jasnell:
V8 has deprecated the SetAccessor method. Per the deprecation message, we're supposed to replace it with SetNativeDataProperty. Unfortunately, SetNativeDataProperty does not work with setters specified on the prototype. Doh! Per feedback from the v8 team, we should use SetAccessorProperty for this case instead.

Unfortunately, SetAccessorProperty is not a drop in replacement for the deprecated SetAccessor and requires more extensive changes than simply calling a different method.

Can't say that I'm entirely happy with this PR as is given the duplication is ends up introducing. We may want to revisit this whole bit to determine if there's a way of simplifying things.

### DIFF
--- a/src/workerd/api/http-test.js
+++ b/src/workerd/api/http-test.js
@@ -90,17 +90,17 @@ export const inspect = {
     assert.strictEqual(util.inspect(url),
 `URL {
   searchParams: URLSearchParams(3) { 'a' => '1', 'a' => '2', 'b' => '3' },
-  hash: '',
-  search: '?a=1&a=2&b=3',
-  pathname: '/path',
-  port: '8787',
-  hostname: 'placeholder',
-  host: 'placeholder:8787',
-  password: 'pass',
-  username: 'user',
-  protocol: 'http:',
+  origin: 'http://placeholder:8787',
   href: 'http://user:pass@placeholder:8787/path?a=1&a=2&b=3',
-  origin: 'http://placeholder:8787'
+  protocol: 'http:',
+  username: 'user',
+  password: 'pass',
+  host: 'placeholder:8787',
+  hostname: 'placeholder',
+  port: '8787',
+  pathname: '/path',
+  search: '?a=1&a=2&b=3',
+  hash: ''
 }`
     );
 
@@ -129,7 +129,7 @@ export const inspect = {
   keepalive: false,
   integrity: '',
   cf: undefined,
-  signal: AbortSignal { onabort: null, reason: undefined, aborted: false },
+  signal: AbortSignal { reason: undefined, aborted: false, onabort: null },
   fetcher: null,
   redirect: 'follow',
   headers: Headers(1) { 'content-type' => 'text/plain', [immutable]: false },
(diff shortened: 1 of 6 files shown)


src/workerd/api/http-test.js
```diff
@@ -90,17 +90,17 @@ export const inspect = {
...
-  hash: '',
-  search: '?a=1&a=2&b=3',
-  pathname: '/path',
-  port: '8787',
-  hostname: 'placeholder',
-  host: 'placeholder:8787',
-  password: 'pass',
-  username: 'user',
-  protocol: 'http:',
+  origin: 'http://placeholder:8787',
```
jasnell, line 93: Note for reviewers: Using `SetAccessorProperty` changes the iteration order of these properties... doh!

src/workerd/jsg/resource.h
```diff
@@ -458,6 +457,100 @@ JSG_DEFINE_GETTER_CALLBACK_STRUCTS(const)
...
+    }; \
+    \
+    /* Specialization for methods that take `const v8::PropertyCallbackInfo<v8::Value>&` as \
+      * their first parameter. */ \
+    template <typename TypeWrapper, const char* methodName, typename T, typename Ret, \
+              typename... Args, \
+              Ret (T::*method)(const v8::PropertyCallbackInfo<v8::Value>&, Args...) __VA_ARGS__, \
+              bool isContext> \
+    struct PropertyGetterCallback<TypeWrapper, methodName, \
+                          Ret (T::*)(const v8::FunctionCallbackInfo<v8::Value>&, Args...) __VA_ARGS__, \
```
jasnell, line 520: Note for reviewers: with the change to url.h included in this PR, I don't believe we actually have any more cases where a property getter takes either a `v8::FunctionCallbackInfo` or a `v8::PropertyCallbackInfo`. So perhaps we can simplify things by removing these specializations?

src/workerd/jsg/resource.h
```diff
@@ -531,6 +624,75 @@ struct SetterCallback<TypeWrapper, methodName,
...
+      auto& self = extractInternalPointer<T, isContext>(context, obj);
+      (self.*method)(Lock::from(isolate), wrapper.template unwrap<Arg>(context, info[0],
+          TypeErrorContext::setterArgument(typeid(T), methodName)));
+    });
+  }
+};
+
+// Specialization for methods that take `const v8::PropertyCallbackInfo<void>&` as their
+// first parameter.
+template <typename TypeWrapper, const char* methodName, typename T, typename Arg,
```
jasnell, line 675: Note for reviewers: likewise here... we don't have anything that's actually using this specialization. We can simplify by removing this an the `v8::PropertyCallbackInfo` variation

src/workerd/jsg/resource.h
harrishancock: I'd be supportive of this!

harrishancock (approved):

PR #2157, opened 2024-05-24
Pass through the exception type of several durable object internal errors

jclee:


justin-mp (approved):

jasnell (approved):

jclee:
(Per internal feedback, updated PR to move `WD_EXPECT_THROW()` test macro to a more accessible location.)

src/workerd/util/test.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+#include <kj/test.h>
+
+namespace workerd {
+
+// Checks that code throws an exception matching the type and message of the given exception.
+#define WD_EXPECT_THROW(expException, code, ...)                                                   \
+  do {                                                                                             \
+    auto expExcObj = expException;                                                                 \
+    KJ_IF_SOME(e, ::kj::runCatchingExceptions([&]() { (void)({ code; }); })) {                     \
```
jasnell, line 14: Nit: what about exceptions like `jsg::JsExceptionThrown`? If those aren't important to capture here, all good... but might be good to specify that in a comment

justin-mp (approved):

src/workerd/util/test.h
jclee: Good point...  The function does not currently provide special handling for `jsg::JsExceptionThrown`, and I hadn't been fully aware that those exceptions might need to be handled differently.

However, I think the function as-is is sufficient for a test to ensure that a thrown `kj::Exception` has the expected type and message, which is what it is currently used for.  If a `JsExceptionThrown` is thrown instead, it would be captured via `getCaughtExceptionAsKj()` as a `kj::Exception` of type "failed", with an extra `std::exception` in the description.  So I think it would hard at least for a `JsExceptionThrown` to accidentally match an expected non-`JsExceptionThrown` exception.

I've added a comment advising that the macro is intended for matching `kj::Exception`s, and that someone who wants to assert on a JS exception should probably use a separate macro.

PR #2161, opened 2024-05-24
Use weak refs between WebSockets in a WebSocketPair

jasnell:
Breaks the strong ref cycle between the two WebSockets in a WebSocketPair.

Aso, while at it, add a missing visitForGc impl to WebSocket to make sure the stored error, if any, is traced.

### DIFF
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -48,7 +48,8 @@ WebSocket::WebSocket(jsg::Lock& js,
     IoContext& ioContext,
     kj::WebSocket& ws,
     HibernationPackage package)
-    : url(kj::mv(package.url)),
+    : weakRef(kj::refcounted<WeakRef<WebSocket>>(kj::Badge<WebSocket> {}, *this)),
+      url(kj::mv(package.url)),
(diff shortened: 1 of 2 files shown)


src/workerd/api/web-socket.c++
```diff
@@ -1068,13 +1071,20 @@ bool WebSocket::awaitingHibernatableRelease() {
 }
 
 void WebSocket::setRemoteOnPair() {
-  JSG_REQUIRE_NONNULL(maybePair, Error,
-      "this WebSocket is not one end of a WebSocketPair")->locality = REMOTE;
+  auto& ref = JSG_REQUIRE_NONNULL(maybePair, Error,
+      "this WebSocket is not one end of a WebSocketPair");
+  if (!ref->runIfAlive([](WebSocket& ref) { ref.locality = REMOTE; })) {
+    JSG_FAIL_REQUIRE(Error, "the other end of this WebSocketPair has been released.");
```
kentonv, line 1077: Do we actually want to throw in this case? Or would the right behavior be to ignore it, because if the pair has been discarded then setting the flag is irrelevant?

mikea (approved):

PR #2163, opened 2024-05-24
[CI] Reduce set of build dependencies

fhanau:
Be more intentional about what dependencies we use. This should make CI run a bit faster.

===========

Was inspired to do this while debugging an internal CI issue. We have long resolved our CI space constraints by not building everything twice and using shared linkage, but this should still speed up the build process slightly.


petebacondarwin (approved):

mikea (approved):

fhanau:
Rebased this and updated `.github/workflows/npm-types.yml`, where we also install clang manually as of #2173.

Issue #2051, opened 2024-04-24
devtools/vscode debugging still fails with Message is too large (>1MB)


erkkah:
As mentioned in #1563, this still happens for large devtools messages that are passed between workerd instances.
Since these messages get read here:

https://github.com/cloudflare/workerd/blob/ce1f7c3e745ae3e17796ba853a911b301171e9da/src/workerd/api/web-socket.c%2B%2B#L913

They are also limited by the default 1M limit.

As an experiment (https://github.com/cloudflare/workerd/compare/main...erkkah:workerd:fix/hanging-debugsession) I increased the limit, which made the problem go away.

However, the limit should probable only be changed for the debugging case, and I don't know the code enough to suggest a solution.

In my setup, I hit this limit as soon as a breakpoint hits. I use 15 durable object namespaces. If I reduce the number of loaded namespaces, the problem goes away.

kentonv:
Hmm, this suggests that the devtools WebSocket connection is being proxied through a worker. Wrangler normally creates a hidden proxy worker around the application itself, but is it also proxying devtools inspector connections? It seems like this shouldn't be needed -- can wrangler connect directly to the inspector port instead?

kyle-pena-nlp:
I am also seeing this behavior.  Would be great if team could fix sooner rather than later, I have downgraded wrangler (which depends on workerd) to 3.18.0 as a workaround.

PR #2164, opened 2024-05-24
Add a compat flag for increasing the WS message size limit

penalosa:
This is one step towards addressing #2051 (Wrangler changes will also be needed)

### DIFF
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -259,7 +259,7 @@ jsg::Ref<WebSocket> WebSocket::constructor(
     headers.unset(kj::HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS);
   }
 
-  if (!FeatureFlags::get(js).getIncreaseWebsocketMessageSize()) {
+  if (FeatureFlags::get(js).getIncreaseWebsocketMessageSize()) {
     ws->maxMessageSize = 128u << 20;
   }
 
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -259,6 +259,10 @@ jsg::Ref<WebSocket> WebSocket::constructor(
     headers.unset(kj::HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS);
   }
 
+  if (!FeatureFlags::get(js).getIncreaseWebsocketMessageSize()) {
+    ws->maxMessageSize = 128u << 20;
+  }
+
   auto prom = ([](auto& context, auto connUrl, auto headers, auto client)
       -> kj::Promise<PackedWebSocket> {
     auto response = co_await client->openWebSocket(connUrl, headers);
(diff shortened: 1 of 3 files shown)


src/workerd/api/web-socket.c++
```diff
@@ -259,6 +259,10 @@ jsg::Ref<WebSocket> WebSocket::constructor(
     headers.unset(kj::HttpHeaderId::SEC_WEBSOCKET_EXTENSIONS);
   }
 
+  if (!FeatureFlags::get(js).getIncreaseWebsocketMessageSize()) {
+    ws->maxMessageSize = 128u << 20;
+  }
```
jasnell, line 264: Nit: Given that the flag is marked experimental, this check feels a bit backwards. The `size_t maxMessageSize` default value should probably be `128u <<20` and this check should be inverted to set `1u << 20` when the flag is set.

jasnell (approved):

src/workerd/io/compatibility-date.capnp
```diff
@@ -441,4 +441,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
     $compatDisableFlag("fetch_legacy_url")
     $compatEnableDate("2024-06-03");
   # Ensures that WHATWG standard URL parsing is used in the fetch API implementation.
+
+  increaseWebsocketMessageSize @50 :Bool
+      $compatEnableFlag("increase_websocket_message_size")
+      $experimental;
+  # Experimental, increase the WebSocket message size limit to 128MB
```
kentonv, line 448: In order to avoid misleading people about the intention here, can we please say:

> For local development purposes only, increase the message size limit to 128MB. This is not expected ever to be made available in production, as large messages are inefficient.

src/workerd/api/web-socket.c++
kentonv: Shouldn't the `if` condition be reversed? Currently it's `if (!flag)` but it should be `if (flag)`? Then the rest of the logic is correct.

I'd prefer that `1u << 20` is the default, and the flag changes it.

src/workerd/api/web-socket.h
```diff
@@ -425,6 +425,9 @@ class WebSocket: public EventTarget {
   // `close()`, thereby preventing calls to `send()` even after we wake from hibernation.
   bool closedOutgoingForHib = false;
 
+  // Maximum allowed size for WebSocket messages
+  size_t maxMessageSize = 1u << 20;
```
kentonv, line 429: Use the constant `kj::WebSocket::SUGGESTED_MAX_MESSAGE_SIZE`.

kentonv, line 429: Instead of having this be a member of `webSocket`, could you pass it as a parameter to `readLoop()` (reading the feature flag in `startReadLoop()`)?

petebacondarwin:
Probably related: https://github.com/cloudflare/workers-sdk/issues/6219

src/workerd/api/web-socket.h
RamIdeas: https://github.com/cloudflare/workerd/pull/2164/commits/e1fcf20a31aa5617c3e2a253634e84dd7c82fa08

RamIdeas: https://github.com/cloudflare/workerd/pull/2164/commits/bfbfda1c06ba3934725d1312f2e0a0fdd1546472

PR #2167, opened 2024-05-25
Add FeatureObserver to metrics

jasnell:
Too often we end up adding temporary logging output to determine if a particular feature, API, etc is being used. This PR introduces a new "Feature Observer" that is essentially a process-wide set of counters that can be used to determine how much a particular feature is used. For workerd this is a non-op by default. Internally, the idea is to collect these metrics process-wide via a prometheus endpoint.

Features themselves are just simple identifiers defined in a capnp enum.  In code, we can call `FeatureObserver::maybeRecordUse(feature)` to increment the counter for a particular feature. This will make it easier for us to add in tracking to determine if and how much a particular feature is used.

There is an internal PR that builds on this. Discussion on how this is expected to be consumed should happen there.

### DIFF
--- a/src/workerd/io/observer.h
+++ b/src/workerd/io/observer.h
@@ -12,6 +12,7 @@
 #include <kj/time.h>
 #include <kj/compat/http.h>
 #include <workerd/io/trace.h>
+#include <workerd/io/features.capnp.h>
 #include <workerd/jsg/observer.h>
 
 namespace workerd {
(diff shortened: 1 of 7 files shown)


src/workerd/io/observer.h
```diff
@@ -247,4 +248,33 @@ class TeardownFinishedGuard {
   Observer& ref;
 };
 
+// Provides counters/observers for various features. The intent it to
```
harrishancock, line 251:
```suggestion
// Provides counters/observers for various features. The intent is to
```

src/workerd/io/observer.h
```diff
@@ -247,4 +248,33 @@ class TeardownFinishedGuard {
...
+
+  // A "Feature" is just an opaque identifier defines in the features.capnp
+  // file.
+  using Feature = workerd::Features;
+
+  // Called to increment the usage counter for a feature.
+  virtual void use(Feature feature) const {}
+
+  using CollectCallback = kj::Function<void(Feature, const uint64_t)>;
+  virtual void collect(CollectCallback&& callback) const {}
```
harrishancock, line 270: Doc comment? It's not clear from the interface what this function is for.

src/workerd/io/observer.h
```diff
@@ -247,4 +248,33 @@ class TeardownFinishedGuard {
...
+// used and how often.
+//
+// There is exactly one instance of this class per worker process.
+class FeatureObserver {
+public:
+  static kj::Own<FeatureObserver> createDefault();
+  static void init(kj::Own<FeatureObserver> instance);
+  static kj::Maybe<FeatureObserver&> get();
+
+  // A "Feature" is just an opaque identifier defines in the features.capnp
```
harrishancock, line 262:
```suggestion
  // A "Feature" is just an opaque identifier defined in the features.capnp
```

harrishancock (approved):

src/workerd/io/observer.h
jasnell, line 270:
```suggestion
  using CollectCallback = kj::Function<void(Feature, const uint64_t)>;
  // This method is called from the internal metrics collection mechanisn to harvest the
  // current features and counts that have been recorded by the observer.
  virtual void collect(CollectCallback&& callback) const {}
```

PR #2170, opened 2024-05-26
Add type definition of Rate Limit bindings

DaniFoldi:
Hi :wave:,

This PR adds a simple type definition for Rate Limit bindings.

As it is in beta, I expect this will evolve over time, so I didn't copy most of the documentation into jsdoc.

Happy to take suggestions for the names of the interfaces and anything else, my other idea was `RateLimitBinding`, but not entirely sure if it would fit in with other bindings.

cc @elithrar

### DIFF
--- a/types/defines/ratelimit.d.ts
+++ b/types/defines/ratelimit.d.ts
@@ -1,0 +1,16 @@
+interface RateLimitOptions {
+  key: string
+}
+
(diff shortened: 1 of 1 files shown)


dom96:
CC @oliy

jasnell (approved):

PR #2175, opened 2024-05-28
convert memcmp == 0 to arrayPtr equality

tewaro:


AdityaAtulTewari:
This PR will be broken up and instead accompanied by tests.

PR #2176, opened 2024-05-28
Updated types to seperate models for composability and reusability

Dhravya:
As discussed in the internal chat, I feel it's better to output types where the models are detached from the Ai class, so I slightly updated the type generator to generate types in this format
```ts
export type BaseAiTextClassificationModels = ...;
  

export declare abstract class Ai {
  fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response>;
  run(
    model: BaseAiTextClassificationModels,
    inputs: BaseAiTextClassification["inputs"],
    options?: AiOptions
  ): Promise<BaseAiTextClassification["postProcessedOutputs"]>;
  ```
  
This will not only expose the types out for the users but also for our new features in workersAI where having TextModel types will be good for the SDK.
I have also created a PR in types generator in AI SDK

### DIFF
--- a/types/defines/ai.d.ts
+++ b/types/defines/ai.d.ts
@@ -14,6 +14,12 @@ export type AiImageToTextInput = {
   prompt?: string;
   max_tokens?: number;
   temperature?: number;
+  top_p?: number;
+  top_k?: number;
+  seed?: number;
+  repetition_penalty?: number;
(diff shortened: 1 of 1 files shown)


Dhravya:

mikea (approved):

PR #2179, opened 2024-05-28
avoid backing store allocation for 0-sized arrays

mikea:
There are plenty memory leaks reported with clean shutdown.

I think the fundamental reason is this line: https://github.com/v8/v8/blob/main/src/objects/backing-store.cc#L180
and our array returning nullptr for begin()


jp4a50 (approved):

jasnell (approved):
Ha! Great timing. I spotted this today also on the flight back home 😁

PR #2182, opened 2024-05-29
Replacing memcmp == 0 with ArrayPtr == in tests only

AdityaAtulTewari:


mikea:
nit for title: something like `replacing memcpy with ==`

mikea (approved):

PR #2185, opened 2024-05-29
Account for external memory usage internal streams

jasnell:
These changes improve the account for external memory usage in internal readable and writable streams. JS-backed streams will be handled in a separate PR.

When a `v8::BackingStore` is detached from the `v8::ArrayBuffer`, it is no longer accounted for in the external memory reporting for a given isolate. Here we implement manual adjustments for both pending read into buffers and enqueued data.

### DIFF
--- a/src/workerd/api/streams/internal.c++
+++ b/src/workerd/api/streams/internal.c++
@@ -515,31 +515,35 @@ kj::Maybe<jsg::Promise<ReadResult>> ReadableStreamInternalController::read(
         js.v8TypeError("This ReadableStream belongs to an object that is closing."_kj));
   }
 
-  std::shared_ptr<v8::BackingStore> store;
+  v8::Local<v8::ArrayBuffer> store;
   size_t byteLength = 0;
   size_t byteOffset = 0;
   size_t atLeast = 1;
 
   KJ_IF_SOME(byobOptions, maybeByobOptions) {
-    auto buffer = byobOptions.bufferView.getHandle(js)->Buffer();
-    store = buffer->GetBackingStore();
+    store = byobOptions.bufferView.getHandle(js)->Buffer();
     byteOffset = byobOptions.byteOffset;
     byteLength = byobOptions.byteLength;
     atLeast = byobOptions.atLeast.orDefault(atLeast);
     if (byobOptions.detachBuffer) {
-      if (!buffer->IsDetachable()) {
+      if (!store->IsDetachable()) {
         return js.rejectedPromise<ReadResult>(
             js.v8TypeError("Unable to use non-detachable ArrayBuffer"_kj));
       }
-      jsg::check(buffer->Detach(v8::Local<v8::Value>()));
+      auto backing = store->GetBackingStore();
+      jsg::check(store->Detach(v8::Local<v8::Value>()));
+      store = v8::ArrayBuffer::New(js.v8Isolate, kj::mv(backing));
     }
   }
 
   auto getOrInitStore = [&](bool errorCase = false) {
-    if (!store) {
+    if (store.IsEmpty()) {
       // In an error case, where store is not provided, we can use zero length
       byteLength = errorCase ? 0 : UnderlyingSource::DEFAULT_AUTO_ALLOCATE_CHUNK_SIZE;
-      store = v8::ArrayBuffer::NewBackingStore(js.v8Isolate, byteLength);
+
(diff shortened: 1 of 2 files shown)


src/workerd/api/streams/queue.c++
```diff
@@ -251,16 +251,21 @@ void maybeInvalidateByobRequest(kj::Maybe<ByteQueue::ByobRequest&>& req) {
...
 ByteQueue::ReadRequest::ReadRequest(
+    jsg::Lock& js,
     jsg::Promise<ReadResult>::Resolver resolver,
     ByteQueue::ReadRequest::PullInto pullInto)
     : resolver(kj::mv(resolver)),
-      pullInto(kj::mv(pullInto)) {}
+      pullInto(kj::mv(pullInto)) {
+  // The store needs to be accounted for as external memory in the isolate.
+  js.v8Isolate->AdjustAmountOfExternalAllocatedMemory(this->pullInto.store.size());
+}
```
jasnell, line 261: For reviewers: There is arguably an easier approach we can take here. The only reason we need to call `AdjustAmountOfExternalAllocatedMemory` here and when the entry is popped from the queue is because at this level we are currently only storing and handling the detached `v8::BackingStore` instances. If we changed this queue to store `v8::ArrayBuffer` instances then the memory accounting would be automatically tracked. That comes at the cost of minor additional allocations on the v8 heap but that may be ok? I'm not in a rush to land this so let's consider the options.

jasnell: The more I think about it... the more I think taking this alternative approach is the better path. I'm going to move this back into draft PR status until I can work that up.

jp4a50: Depends how confident you are you can test all the edge cases vs. the potential overhead of the `v8::ArrayBuffer`s I guess. If we think the overhead is relatively low (hopefully is for typical read sizes?) then I'd definitely agree that it makes sense to opt for something which eliminates our need to manually account for the memory usage.

Are you proposing to also use `v8::ArrayBuffer` to account for the period of untracked memory in `ReadableStreamInternalController::read` too? If so, sounds good. If not, I was going to propose at least wrapping the symmetric calls to `AdjustAmountOfExternalAllocatedMemory` in some kind of RAII helper to avoid the possibility of incrementing but not decrementing or vice versa.

jasnell: I plan to split the effort in two parts. First, focusing only on what is stored in the queues, then look separately at the buffers as they are in motion via the various promises. It likely makes sense to use the `v8::ArrayBuffer` wrapper in every case as it'll just make things easier, but I need to verify. The `v8::ArrayBuffer` itself is only going to be a handful of additional bytes, they aren't that big and will be tracked against the heap limit so it's safe.

jasnell:
Ok, this is ready for review. I've limited this PR to only touch the `api/streams/internal.*` bits. I'll handle the JS-backed streams separately just to make things easier to review.

src/workerd/api/streams/internal.c++
```diff
@@ -515,31 +515,35 @@ kj::Maybe<jsg::Promise<ReadResult>> ReadableStreamInternalController::read(
...
   }
 
   auto getOrInitStore = [&](bool errorCase = false) {
-    if (!store) {
+    if (store.IsEmpty()) {
       // In an error case, where store is not provided, we can use zero length
       byteLength = errorCase ? 0 : UnderlyingSource::DEFAULT_AUTO_ALLOCATE_CHUNK_SIZE;
-      store = v8::ArrayBuffer::NewBackingStore(js.v8Isolate, byteLength);
+
+      if (!v8::ArrayBuffer::MaybeNew(js.v8Isolate, byteLength).ToLocal(&store)) {
```
jp4a50, line 544: Why do we use `MaybeNew` here but `New` [here](https://github.com/cloudflare/workerd/pull/2185/files#diff-745dea5c7b6cdcab0023d80592ab9318c5a8adbf6e40e83e73e66b7d86241f74R535)?

jp4a50 (approved):

src/workerd/api/streams/internal.c++
jasnell: If I'm following the link there correctly, this one is actually allocating new space, the other is just wrapping around the existing allocation.

jp4a50: Ah, of course, so the one that wraps an existing allocation can't fail which is why it's `New` and not `MaybeNew`. Makes sense.

PR #2188, opened 2024-05-29
IMAGES-1177: Implement wrapped Images binding

ns476:
This creates a new TS binding that allows image transformations and makes it available as an internal Cloudflare binding.

### DIFF
--- a/src/cloudflare/internal/images-api.ts
+++ b/src/cloudflare/internal/images-api.ts
@@ -1,0 +1,179 @@
+// Copyright (c) 2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+type Fetcher = {
+  fetch: typeof fetch;
+};
+
+type RawInfoResponse =
+  | { format: "image/svg+xml" }
+  | {
+      format: string;
+      file_size: number;
+      width: number;
+      height: number;
+    };
+
+class TransformationResultImpl implements TransformationResult {
+  public constructor(private readonly bindingsResponse: Response) {}
+
+  public contentType(): string {
+    const contentType = this.bindingsResponse.headers.get("content-type");
+    if (!contentType) {
+      throw new ImagesErrorImpl(
+        "IMAGES_TRANSFORM_ERROR 9523: No content-type on bindings response",
+        9523
+      );
+    }
+
+    return contentType;
+  }
+
+  public image(): ReadableStream<Uint8Array> {
+    return this.bindingsResponse.body || new ReadableStream();
+  }
+
+  public response(): Response {
+    return new Response(this.image(), {
+      headers: {
+        "content-type": this.contentType(),
+      },
+    });
+  }
+}
+
+// Read input into memory for now, as it's difficult to stream multipart
+// requests and this is what the backend uses.
+//
+// The interface takes streams to allow streaming later.
+async function streamToBlob(stream: ReadableStream<Uint8Array>): Promise<Blob> {
+  return new Response(stream).blob();
+}
+
+class ImageTransformerImpl implements ImageTransformer {
+  private transforms: Transform[];
+  private consumed: boolean;
+
+  public constructor(
+    private readonly fetcher: Fetcher,
+    private readonly stream: ReadableStream<Uint8Array>
+  ) {
+    this.transforms = [];
+    this.consumed = false;
+  }
+
+  public transform(transform: Transform): ImageTransformerImpl {
+    this.transforms.push(transform);
+    return this;
+  }
+
+  public async output(options: OutputOptions): Promise<TransformationResult> {
+    if (this.consumed) {
+      throw new ImagesErrorImpl(
+        "IMAGES_TRANSFORM_ERROR 9525: ImageTransformer consumed; you may only call .output() once",
+        9525
+      );
+    }
+    this.consumed = true;
+
+    const body = new FormData();
+    body.append("image", await streamToBlob(this.stream));
+    body.append("output_format", options.format);
+    if (options.quality !== undefined) {
+      body.append("output_quality", options.quality.toString());
+    }
+
+    if (options.background !== undefined) {
+      body.append("background", options.background);
+    }
+
+    body.append("transforms", JSON.stringify(this.transforms));
+
+    const response = await this.fetcher.fetch(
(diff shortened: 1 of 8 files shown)


xjewer:
lgtm

types/defines/images.d.ts
```diff
@@ -0,0 +1,105 @@
...
+  input(blob: Blob): ImageTransformer;
+}
+
+interface ImageTransformer {
+  transform(transform: Transform): ImageTransformer;
+  output(options: OutputOptions): Promise<TransformationResult>;
+}
+
+interface TransformationResult {
+  response(): Response;
```
IgorMinar, line 98: Could you avoid making the result of type `Response` as that leaks the underlying transport layer. You already expose `image` as a readable stream which is perfect. Do we need to expose more?

types/defines/images.d.ts
```diff
@@ -0,0 +1,105 @@
...
+  contentType(): string;
+  image(): ReadableStream<Uint8Array> | null;
+}
+
+interface ImagesError {
+  code: number;
```
IgorMinar, line 104: I wonder if there is a precedence for error handling in bindings. It seems that R2 and D1 as preexisting APIs take vastly different approach to this.

Could we think about which way we want to go in this area, and start consolidating the error APIs to feel consistent?

IgorMinar: at minimum this property should be readonly. For reference this is what R2Error looks like:

```

declare interface R2Error extends Error {
  readonly name: string;
  readonly code: number;
  readonly message: string;
  readonly action: string;
  readonly stack: any;
}
```
source: https://unpkg.com/browse/@cloudflare/workers-types@4.20240614.0/experimental/index.d.ts

types/defines/images.d.ts
```diff
@@ -0,0 +1,105 @@
...
+    | "image/avif"
+    | "rgb"
+    | "rgba";
+  quality?: number;
+  background?: string;
+};
+
+interface ImagesBinding {
+  info(blob: Blob): Promise<InfoResponse>;
+  input(blob: Blob): ImageTransformer;
```
IgorMinar, line 89: could the input be a stream instead? It's easy to turn a blob to a stream, the opposite is problematic.

IgorMinar: my main goal here is to enable streaming transformations/use-cases by default, because enabling that ad-hoc will be hard/awkward.

types/defines/images.d.ts
```diff
@@ -0,0 +1,105 @@
...
+
+interface ImageTransformer {
+  transform(transform: Transform): ImageTransformer;
+  output(options: OutputOptions): Promise<TransformationResult>;
+}
+
+interface TransformationResult {
+  response(): Response;
+  contentType(): string;
+  image(): ReadableStream<Uint8Array> | null;
```
IgorMinar, line 100: do we need to support `null`?

IgorMinar (commented):
I'm excited about this addition! I have some API design feedback, please take a look...

IgorMinar:
one more thing... this feature is really cool, but it will not be very useful if we don't come up with a local development story for it. Have you thought about how can one develop a worker locally with wrangler and miniflare and still use this new binding?

types/defines/images.d.ts
ns476: This is just a convenience that does:
```ts
new Response(this.image(), {
      headers: {
        "content-type": this.contentType(),
      },
    });
```
so I don't think it leaks anything, but happy to remove it if you don't think it's useful

types/defines/images.d.ts
ns476: I've made the interfaces take streams and then buffer internally for now as it's a bit tricky to stream properly with multipart, which is what the backend uses

types/defines/images.d.ts
ns476: Made it this for now:
```ts
interface ImagesError extends Error {
  readonly code: number;
  readonly message: string;
  readonly stack?: string;
}
```

It's fairly similar to the Vectorize error (which I cribbed from shamelessly)

ns476:
Thanks for the review!

> one more thing... this feature is really cool, but it will not be very useful if we don't come up with a local development story for it. Have you thought about how can one develop a worker locally with wrangler and miniflare and still use this new binding?

I think the [approach taken with the AI binding](https://github.com/cloudflare/workers-sdk/commit/c10bf0fd2f24681cfcb78c6bf700a2e4acf41f30) will work nicely here: provide an endpoint on api.cloudflare.com with OAuth that proxies to a worker that invokes the binding.

GregBrimble:
Like Igor, very excited for this! I've personally needed it many many times :)

src/cloudflare/internal/images-api.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (value) {
+      chunks.push(value);
+    }
+
+    if (done) {
+      break;
+    }
+  }
+
+  return new Blob(chunks);
```
RamIdeas, line 61: Worth measuring but in theory it would be more efficient to avoid bouncing back and forth between C++ and JS:

```suggestion
  return new Response(stream).blob();
```

ns476: Made this change, also much more obviously correct which is a plus for me!

src/cloudflare/internal/images-api.ts
```diff
@@ -0,0 +1,178 @@
...
+// requests and this is what the backend uses.
+//
+// The interface takes streams to allow streaming later.
+async function streamToBlob(stream: ReadableStream<Uint8Array>): Promise<Blob> {
+  return new Response(stream).blob();
+}
+
+export class ImageTransformerImpl implements ImageTransformer {
+  transforms: Transform[];
+  consumed: boolean;
```
anonrig, line 55: I believe this should be private.

src/cloudflare/internal/images-api.ts
```diff
@@ -0,0 +1,178 @@
...
+    if ("file_size" in r) {
+      return {
+        fileSize: r.file_size,
+        width: r.width,
+        height: r.height,
+        format: r.format,
+      };
+    } else {
+      return r;
+    }
```
anonrig, line 134:
```suggestion
    }
    
    return r;
```

jasnell (approved):

anonrig (approved):

RamIdeas (approved):

Issue #2174, opened 2024-05-27
Unexpected change in `DOMException.stack` type


Cherry:
In `workers-types` `<=4.20240512.0`, the following code worked without issue, causing no type errors to be raised at build time.

But since `4.20240524.0`, an error is now raised:

```
Argument of type 'string | undefined' is not assignable to parameter of type 'string | Record<string, unknown>'.
  Type 'undefined' is not assignable to type 'string | Record<string, unknown>'.
```

```typescript
class CustomError extends Error {
	data: Record<string, unknown>;
	constructor(message: string, data: Record<string, any>) {
		super(message);
		this.data = data;
	}
}

function handleError(err: Record<string, unknown> | string) {
	// do something
}

export default {
	async fetch(request: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
		try {
			// some code that might throw a custom error
			throw new CustomError('example', { code: 123 });

			// or some code that might throw a DOMException
			throw new DOMException('test');
		} catch (err) {
			if (err instanceof CustomError) {
				handleError(err.data);
			} else if (err instanceof DOMException) {

				// previously, `.stack` was defined as `any` by DOMException, so it passed the type check in handleError
				// this is no longer the case, and it inherits `string | undefined` from `Error` as its not defined explicitly anymore
				// so the following line will cause a TS build error
				handleError(err.stack);
			} else {
				// handle other errors
			}
		}
		return new Response('Hello World');
	},
};

```

This appears to be due to `stack` being omitted from `DOMException`, so it's now just inherited from the base `Error` class. Which results in it being `stack?: string`, so effectively `string | undefined`.

Previously, `workers-types` defined `readonly stack: any` in `DOMException` but that's no longer present as of `4.20240524.0`, resulting in not only a type change, but also a property that was previously no longer optional, now being optional. I suspect this is due to changes in https://github.com/cloudflare/workerd/pull/2139.

## `DOMException` reference

I've included a reference diff of what `DOMException` looks like as output in `workers-types` between these two versions.

### `workers-types` `4.20240512.0`
```typescript
declare class DOMException extends Error {
  constructor(message?: string, name?: string);
  /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/DOMException/message) */
  readonly message: string;
  /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/DOMException/name) */
  readonly name: string;
  /**
   * @deprecated
   *
[...]

PR #2190, opened 2024-05-29
Add type override for DOMException stack property

jasnell:
Fixes: https://github.com/cloudflare/workerd/issues/2174


fhanau (approved):
Change looks sane. Since you're adding a set method here, is stack supposed to be writable? It was readonly before the refactor PR, but based on discussion it sounds like it might be writable now.

jasnell:
Yes, originally it was readonly but with this change it becomes mutable.

PR #2193, opened 2024-05-30
Avoid unnecessary allocation in internal stream read

jasnell:
The implementation was allocating a new 4096 byte buffer for each read even if the stream was errored or closed. This change avoids the unnecessary allocation.

I'll likely take another cleanup pass in this area soon to simplify a few bits and pieces since this was all written before `jsg::BufferSource` existed, but for now just making this allocation lazy will make things more efficient.

### DIFF
--- a/src/workerd/api/streams/internal.c++
+++ b/src/workerd/api/streams/internal.c++
@@ -533,13 +533,17 @@ kj::Maybe<jsg::Promise<ReadResult>> ReadableStreamInternalController::read(
       }
       jsg::check(buffer->Detach(v8::Local<v8::Value>()));
     }
-  } else {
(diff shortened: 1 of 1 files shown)


src/workerd/api/streams/internal.c++
```diff
@@ -572,8 +577,8 @@ kj::Maybe<jsg::Promise<ReadResult>> ReadableStreamInternalController::read(
       }
       readPending = true;
 
-      // TODO(now): This is the implementation we should be able to move to except there's a bug
-      // somewhere in the awaitIo implementation that allows IdentityTransformStream to hang.
```
jasnell, line 576: FWIW, this comment is just an old obsolete comment that should have been deleted before.

Issue #2189, opened 2024-05-29
🐛 import destructuring doesn't work for CJS modules imported from an ESM module


IgorMinar:
## Repro:

`worker.js`
```js
import {parse} from 'cookie';

console.log(parse());

export default {
  async fetch(request) {
    return new Response();
  }
};
```

`cookie.js`
```js
exports.parse = function () {
  return "yay!";
}
```

```capnp
using Workerd = import "/workerd/workerd.capnp";
const helloWorldExample :Workerd.Config = (
  services = [ (name = "main", worker = .helloWorld) ],
  sockets = [ ( name = "http", address = "*:8080", http = (), service = "main" ) ]
);

const helloWorld :Workerd.Worker = (
  modules = [
    (name = "worker", esModule = embed "worker.js"),
    (name = "cookie", commonJsModule = embed "cookie.js"),
  ],
  compatibilityDate = "2024-05-29",
  compatibilityFlags = ["nodejs_compat"]
);
```

## Expected result
console output:
```
yay!
```

## Actual result
console output:
```
service main: Uncaught SyntaxError: The requested module 'cookie' does not provide an export named 'parse'
  at worker:1:8
```

## Workaround
Rewrite imports in the ES module as follows:

from:
```js
import {parse} from 'cookie';
```
to:
```js
import cookie from 'cookie'; const { parse } = cookie;
```

IgorMinar:
Thanks for the assessment, James!

What if the fallback service returned the list of known exports for CJS modules? We could get that info to you easily.

If we did, would it still be a lot of work in workerd to make it all work?

jasnell:
That would certainly decrease the complexity of the change but I think there would still be a bit of effort involved. Specifically, in or current implementation of the module registry, for CommonJS modules, we'd have to make sure the necessary plumbing is there to allow the fallback service to pass the list of exports in and have them be used. I *think* it's there but would need to verify. It's something I ought to be able to look at this week but given that it's already wednesday I'm not confident that there's enough time to both implement *and* review to get it done this week. Let me evaluate and I'll get back to you.

IgorMinar:
sounds good. thanks James.

We are trying to come up with a workaround that would work well enough for our Vite experimentation because otherwise the Vite effort is blocked and can't progress until this bug is resolved in workerd.

jasnell:
[...]

PR #2194, opened 2024-05-30
Implement named export support for CJS modules

jasnell:
Adds the ability for a worker module definition to define a list of named exports. So if the CJS module exports something like...

```
exports.foo = 42;
```

Then the worker can import `foo` directly from the module:

```
import { foo } from 'cjs-module';
```

This just enables the exports in workerd, a separate internal PR will be required to enable it's use in production.

The runtime does very little lifting here. It will be up to whomever is constructing the worker module configuration to identify the exports that should be provided. The runtime will not try to automatically detect the exports. if an export is named but not actually defined by the CJS module, then it's value will be `undefined`. Only exports that are defined when the initial evaluation of the module completes will be supported (CJS generally allows exports to be modified dynamically), and the exports will only work if the `module.exports` value is a non-null object.

This feature was requested by the frameworks team.

Fixes: #2189

### DIFF
--- a/src/workerd/api/tests/commonjs-module-test.js
+++ b/src/workerd/api/tests/commonjs-module-test.js
@@ -1,0 +1,14 @@
+import { foo, bar, default as baz } from 'foo';
+import * as fooModule from 'foo';
+import { strictEqual } from 'node:assert';
+
+export const test = {
+  test() {
+    strictEqual(fooModule.default, baz);
+    strictEqual(fooModule.foo, foo);
+    strictEqual(fooModule.bar, undefined);
+    strictEqual(foo, 1);
+    strictEqual(bar, undefined);
+    strictEqual(baz.foo, foo);
+  }
+};
--- a/src/workerd/api/tests/commonjs-module-test.wd-test
+++ b/src/workerd/api/tests/commonjs-module-test.wd-test
@@ -1,0 +1,18 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "commonjs-module-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "commonjs-module-test.js"),
+          (name = "foo",
+           commonJsModule = "exports.foo = 1",
+           namedExports = ["foo", "bar"])
+        ],
+        compatibilityDate = "2023-01-15",
+        compatibilityFlags = ["nodejs_compat"],
+      )
+    ),
+  ],
+);
--- a/src/workerd/jsg/modules.c++
+++ b/src/workerd/jsg/modules.c++
@@ -178,7 +178,7 @@ v8::MaybeLocal<v8::Value> evaluateSyntheticModuleCallback(
     };
 
     auto& synthetic = KJ_REQUIRE_NONNULL(ref.module.maybeSynthetic, "Not a synthetic module.");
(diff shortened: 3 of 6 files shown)


src/workerd/api/tests/commonjs-module-test.js
```diff
@@ -0,0 +1,10 @@
+import { foo, bar, default as baz } from 'foo';
```
IgorMinar, line 1: could you please also test star imports?

```
import * as foo from 'foo';
```

and ensure that foo is defined with they foo, bar, baz keys?

src/workerd/api/tests/commonjs-module-test.wd-test
```diff
@@ -0,0 +1,18 @@
...
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "commonjs-module-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "commonjs-module-test.js"),
+          (name = "foo",
+           commonJsModule = "exports.foo = 1",
+           namedExports = ["foo", "bar"])
```
IgorMinar, line 11: interesting. I think this will work well enough. 99.999% of developers will never need this as we'll generate these on the fly and provide it via fallback service.

IgorMinar (commented):
this looks very promising! once we have this plumbed through via the fallback service we can test it end to end on real world packages.

src/workerd/api/tests/commonjs-module-test.js
jasnell: Done.

IgorMinar (approved):
LGTM for the JS bits (tests)

IgorMinar:
I thought about this a bit this morning and I was wondering if we could make this an internal feature similar to the fallback service to limit the exposure.

We don't need this functionality in the prod where users can easily work around the lack of this feature by using namespace imports rather than named/destructured or star imports.

We only need this functionality in combination with the module fallback service which deals with 3rd party code that we don't control and don't want to augment.

jasnell:
I think it makes sense to expose it and implementing this way is the easiest thing to do for the fallback service as well.

IgorMinar:
I'll leave it to y'all to decide. I'm just suggesting reducing the scope so that we don't needlessly expand the API surface we need to support in edgeworker/prod.

While this feature is critical for the Vite integration that uses the fallback service, it is not very useful for normal users as they control the code in the bundle, so adding it to edgeworker is just a liability IMO especially given the potential to misconfigure the named experts, etc. But again, I'll defer to y'alls decision.

jasnell:
> ... given the potential to misconfigure the named experts, etc

There's really not much risk here. At worse they specify a non-existent export and the value gets set to `undefined`. There's very little actual risk in the runtime for this.

dom96 (approved):
LGTM but I think it would be helpful to have a test for the case when the expected namedExports are not exported. We're expecting an exception to be thrown, right?

jasnell:
>... to have a test for the case when the expected namedExports are not exports...

That's already covered by the test here https://github.com/cloudflare/workerd/pull/2194/files#diff-d670df980d58c3fa6e9288fc2791fcd96c65a684afe41d98dbb36c7749227fcc ... specifically, the `namedExports` defines a `bar` export that is not actually exported by the module. It's value ends up being `undefined` on import. I would not expect there to be an exception thrown.

fhanau (approved):
LGTM.

PR #2199, opened 2024-05-31
fix(node:buffer): add missing atob and btoa exports

### LINKED ISSUE #2196: 🐛 expose btoa and atob via node:buffer
This should be a simple alias and there is no reason not to do expose it to improve the native compatibility.

Test case:
```
import {btoa, atob} from "node:buffer";

assert(globalThis.btoa === btoa);
assert(globalThis.atob === atob);
```

IgorMinar:
https://nodejs.org/api/buffer.html#bufferatobdata
https://nodejs.org/api/buffer.html#bufferbtoadata

Fixes #2196

### DIFF
--- a/src/node/buffer.ts
+++ b/src/node/buffer.ts
@@ -10,18 +10,34 @@ import {
   SlowBuffer,
 } from 'node-internal:internal_buffer';
 
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+const atob = globalThis.atob;
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+const btoa = globalThis.btoa;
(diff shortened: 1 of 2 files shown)


jasnell (approved):
Might be worth going ahead and including Blob here too

IgorMinar:
@jasnell good catch. added.

jasnell:
@IgorMinar ... can I ask you to rebase to deal with the merge conflict?

IgorMinar:
@jasnell done. please note that I had to add eslint-disable-next-line bailouts into the code because our current typescript setup is not great, or at least I couldn't figure out how to make it all work with our custom d.ts files.

I suspect that we should really be using @types/node instead of our custom typings here but that was a much bigger change that I was willing to bite off in this PR.

PR #2205, opened 2024-05-31
Disable slow tests by default

garrettgu10:
Test plan:
```
bazel test /src/...
```
Observe that Python import tests are not run.

Furthermore, Python import tests should be run in:
1. Linux nondebug on workerd CI
2. Linux debug on Edgeworker CI


jasnell (approved):

hoodmane (approved):
Thanks!

PR #2209, opened 2024-06-03
Improve Blob memory accounting

jasnell:
Updates the `Blob` storage (`ownData`) to use a `jsg::BufferSource` rather than a raw `kj::Array<byte>` to allow the isolate to better track the external memory allocation.

While in here, also updates the `concat(...)` method to use the new safer `copyFrom(...)` API in kj replacing the less safe `memcpy` and manual pointer arithmetic.

We also avoid an extraneous copy when creating a File from a Blob

### DIFF
--- a/src/workerd/api/form-data.h
+++ b/src/workerd/api/form-data.h
@@ -49,8 +49,12 @@ public:
   // `convertFilesToStrings` is for backwards-compatibility. The first implementation of this
   // class in Workers incorrectly represented files as strings (of their content). Changing this
   // could break deployed code, so this has to be controlled by a compatibility flag.
-  void parse(kj::ArrayPtr<const char> rawText, kj::StringPtr contentType,
-             bool convertFilesToStrings);
+  //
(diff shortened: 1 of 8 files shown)


src/workerd/api/form-data.h
```diff
@@ -49,8 +49,12 @@ class FormData: public jsg::Object {
   // `convertFilesToStrings` is for backwards-compatibility. The first implementation of this
   // class in Workers incorrectly represented files as strings (of their content). Changing this
   // could break deployed code, so this has to be controlled by a compatibility flag.
-  void parse(kj::ArrayPtr<const char> rawText, kj::StringPtr contentType,
-             bool convertFilesToStrings);
+  //
+  // Parsing may or may pass a jsg::Lock. If a lock is passed, any File objects created will
```
harrishancock, line 53:
```suggestion
  // Parsing may or may not pass a jsg::Lock. If a lock is passed, any File objects created will
```

harrishancock (approved):

PR #2210, opened 2024-06-03
[build] Update python build dependencies to silence dependabot warning

fhanau:
- Add-on: Update typescript Node.js types and bazelisk versions


package.json
```diff
@@ -10,9 +10,9 @@
     "typescript": "~5.4.2"
   },
   "devDependencies": {
-    "@bazel/bazelisk": "~1.12.1",
```
fhanau, line 13: I tend to do omnibus PRs when updating versions, so this also updates bazelisk and Node.js types – Bazelisk is highly backwards-compatible and Nodejs types also don't change much within a major release, both should be plenty safe. There is a lot more to update here, but that'll require some testing to avoid breakage.

fhanau: "Plenty safe" – there were type errors with the types/node update and how node crypto is declared, I made it a partial update for now.

fhanau:
Updated to include some more minor NPM package updates, PTAL

dom96 (approved):

Issue #2172, opened 2024-05-27
🐛 [bug] `crypto.timingSafeEqual` === `undefined`


helloimalastair:
## Versions
`wrangler@3.57.1`
`miniflare@3.20240512.0`
`workerd@1.20240512.0`
## Issue
According to [Misc | Node Crypto](https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/#misc), the `timingSafeEqual` function should be supported by `nodejs_compat`.

However, when logging `import crypto from "node:crypto";`, you see
```js
{
	DiffieHellman: [Function: DiffieHellman],
	DiffieHellmanGroup: [Function: DiffieHellmanGroup],
	createDiffieHellman: [Function: createDiffieHellman],
	createDiffieHellmanGroup: [Function: createDiffieHellmanGroup],
	getDiffieHellman: [Function: getDiffieHellman],
	KeyObject: [class KeyObject],
	PublicKeyObject: [class PublicKeyObject extends AsymmetricKeyObject],
	PrivateKeyObject: [class PrivateKeyObject extends AsymmetricKeyObject],
	SecretKeyObject: [class SecretKeyObject extends KeyObject],
	generateKey: [Function: generateKey],
	generateKeyPair: [Function: generateKeyPair],
	generateKeyPairSync: [Function: generateKeyPairSync],
	generateKeySync: [Function: generateKeySync],
	createPrivateKey: [Function: createPrivateKey],
	createPublicKey: [Function: createPublicKey],
	createSecretKey: [Function: createSecretKey],
	getRandomValues: [Function: getRandomValues],
	randomBytes: [Function: randomBytes],
	randomFillSync: [Function: randomFillSync],
	randomFill: [Function: randomFill],
	randomInt: [Function: randomInt],
	randomUUID: [Function: randomUUID],
	generatePrime: [Function: generatePrime],
	generatePrimeSync: [Function: generatePrimeSync],
	checkPrime: [Function: checkPrime],
	checkPrimeSync: [Function: checkPrimeSync],
	Hash: [Function: Hash],
	Hmac: [Function: Hmac],
	createHash: [Function: createHash],
	createHmac: [Function: createHmac],
	getHashes: [Function: getHashes],
	hkdf: [Function: hkdf],
	hkdfSync: [Function: hkdfSync],
	pbkdf2: [Function: pbkdf2],
	pbkdf2Sync: [Function: pbkdf2Sync],
	getCiphers: [Function: getCiphers],
[...]

PR #2216, opened 2024-06-04
Fixup timingSafeEqual export for node crypto

jasnell:
Fixes: https://github.com/cloudflare/workerd/issues/2172

### DIFF
--- a/src/node/crypto.ts
+++ b/src/node/crypto.ts
@@ -10,7 +10,7 @@ import {
 
 export const getRandomValues = crypto.getRandomValues;
 export const subtle = crypto.subtle;
-export const timingSafeEqual = (crypto as any).timingSafeEqual;
+export const timingSafeEqual = (subtle as any).timingSafeEqual;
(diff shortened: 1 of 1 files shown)


fhanau (approved):

src/node/crypto.ts
```diff
@@ -10,7 +10,7 @@ import {
 
 export const getRandomValues = crypto.getRandomValues;
 export const subtle = crypto.subtle;
-export const timingSafeEqual = (crypto as any).timingSafeEqual;
+export const timingSafeEqual = (subtle as any).timingSafeEqual;
```
Skye-31, line 13: Is the `as any` still needed? I imagine it was previously needed because it wasn't under the correct object

jasnell: It might not be needed. Not 100% certain ;-)

jasnell: Quick test and the build blows up without it ... for now, will land it like this an d we can revisit

PR #2219, opened 2024-06-05
VS-197: Add new type for Vectorize v2 insert response

garvit-gupta:
Add the new type definition for Vectorize v2 insert/upsert response. This would be used to enable v2 operations in wrangler by making the new type available in workers-sdk.


garvit-gupta:
recheck

garvit-gupta:

jasnell (approved):

PR #2222, opened 2024-06-05
Update AI binding gateway property to be an object

G4brym:


src/cloudflare/internal/ai-api.ts
```diff
@@ -17,8 +17,14 @@ export type SessionOptions = {  // Deprecated, do not use this
...
 
+export type GatewayOptions = {
+  id: string;
+  cacheTtl?: number
+  skipCache?: boolean
+}
+
 export type AiOptions = {
-  gatewayId?: string;
+  gateway?: GatewayOptions;
```
Cherry, line 27: What's the usage look like on `gateway` right now? This is a pretty big breaking change moving from a `string` to an `object`, especially when it's been public in the types as a `string` now for a while.

G4brym: The gateway field isn't released nor working for us internally, so I think it's safe to assume users aren't relying on it 😄

dom96 (approved):

PR #710, opened 2023-05-26
Support for breakpoint debugging

ohodson:

PR #2227, opened 2024-06-06
Add V8 cross-thread profiling patches

### LINKED ISSUE #1754: 🐛 Bug Report — CPU profiling reports entire execution as `(program)`
Hey! 👋 When investigating https://github.com/cloudflare/workers-sdk/issues/4409, I observed that CPU profiles from `workerd` contain a single `(program)` entry for the entire duration of the profile. To reproduce this...

```capnp
# config.capnp

const config :Workerd.Config = (
  services = [
    ( name = "main", worker = .worker ),
  ],
  sockets = [
    ( name = "http", address = "*:8080", http = (), service = "main" ),
  ]
);

const worker :Workerd.Worker = (
  modules = [
    ( name = "./index.mjs", esModule = embed "index.mjs" )
  ],
  compatibilityDate = "2024-01-01",
  compatibilityFlags = ["nodejs_compat"],
);
```

```js
// index.mjs
import { Buffer } from "node:buffer";

async function pbkdf2Derive(password) {
  const passwordArray = encoder.encode(password);
  const passwordKey = await crypto.subtle.importKey(
    "raw", passwordArray, "PBKDF2", false, ["deriveBits"]
  );
  const saltArray = crypto.getRandomValues(new Uint8Array(16));
  const keyBuffer = await crypto.subtle.deriveBits(
    { name: "PBKDF2", hash: "SHA-256", salt: saltArray, iterations: 1_000_000 },
    passwordKey, 256
  );
  return Buffer.from(keyBuffer).toString("base64");
}

export default {
    return new Response(await pbkdf2Derive("hello!"));
  }
}
```

```js
// stress.mjs
const total = 1_000;
for (let i = 0; i < total; i++) {
  const response = await fetch("http://127.0.0.1:8080");
  await response.arrayBuffer();
  if (i % 10 === 0) console.log(`${(i * 100 / total).toFixed(0)}%`);
}
```

...then run...

```shell
$ npx workerd serve config.capnp --verbose --control-fd=1 --watch --inspector-addr=127.0.0.1:9229 &
$ node stress.mjs
```

...then visit `chrome://inspect` to open DevTools and start a CPU profile for a few seconds. For me, the result looks something like...

Screenshot 2024-03-01 at 14 33 30

If I run the following similar code in Node directly with `node --inspect-brk index-node.mjs`, visit `chrome://inspect` and start a CPU profile I see something like...

```js
// index-node.mjs
import { Buffer } from "node:buffer";
import { webcrypto as crypto } from "node:crypto";

async function pbkdf2Derive(password) {
  const passwordArray = encoder.encode(password);
  const passwordKey = await crypto.subtle.importKey(
    "raw", passwordArray, "PBKDF2", false, ["deriveBits"]
  );
  const saltArray = crypto.getRandomValues(new Uint8Array(16));
  const keyBuffer = await crypto.subtle.deriveBits(
    { name: "PBKDF2", hash: "SHA-256", salt: saltArray, iterations: 1_000_000 },
    passwordKey, 256
  );
  return Buffer.from(keyBuffer).toString("base64");
}

const total = 1_000;
for (let i = 0; i < total; i++) {
  console.log(await pbkdf2Derive("hello!"));
  if (i % 10 === 0) console.log(`${(i * 100 / total).toFixed(0)}%`);
}
```

Screenshot 2024-03-01 at 14 33 15

I get that Node probably implements more of these APIs in JavaScript, hence the deeper flame graphs, but I'd still expect to see `pbkdf2Dervie()` in the `workerd` profile's flame graph.

harrishancock:
V8's profiler does not expect isolates to be locked from different threads. Pull request #710 moved inspector protocol handling to a separate thread, meaning profiling was started and stopped on the inspector thread, not the Worker's request connection thread. If I remember right, the result of this is that the profiler ends up sampling the inspector thread, which doesn't normally run JavaScript, and causes a bunch of "(program)" placeholder spans to show up in the profile.

We previously discovered this issue with our internal runtime, which has always been heavily multi-threaded, and patched around the problem. This commit copies the two relevant patches from our internal repo.

Fixes #1754.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -479,6 +479,8 @@ http_archive(
         "//:patches/v8/0015-Modify-where-to-look-for-fp16-dependency.-This-depen.patch",
         "//:patches/v8/0016-Expose-v8-Symbol-GetDispose.patch",
         "//:patches/v8/0017-Rename-V8_COMPRESS_POINTERS_IN_ISOLATE_CAGE-V8_COMPR.patch",
+        "//:patches/v8/0018-EW-3483-CPU-profiler-patches.patch",
+        "//:patches/v8/0019-EW-6624-Profiler-fix-for-V8-versions-10.0-and-above.patch",
     ],
     integrity = "sha256-5gW+N4R+oVxzMZexrqgp14gU42bsFBc5RHPN2oOrTTw=",
     strip_prefix = "v8-12.6.228.9",
--- a/patches/v8/0018-EW-3483-CPU-profiler-patches.patch
+++ b/patches/v8/0018-EW-3483-CPU-profiler-patches.patch
@@ -1,0 +1,178 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Bradley Thwaites <bradley@cloudflare.com>
+Date: Mon, 5 Oct 2020 10:38:59 -0500
+Subject: EW-3483: CPU profiler patches
+
+- Avoids use of pthread_self(), which uses an incorrect cached id.
+- Sampler::PlatformData now calls SYS_getpid and saves tgid.
+- SamplerManager calls SYS_getpid to set a thread_local tgid.
+- SamplerManager::DoSample uses tgid to look up in the sample map.
+- New samplers are added to SamplerManager using tgid, not vm_tid.
+- ThreadManager records the tid of the last thread to lock the
+isolate.
+- Sampler::DoSample sends SIGPROF to that tid.
+- Add libsampler dependencies to BUILD.gn
+
+diff --git a/src/execution/v8threads.cc b/src/execution/v8threads.cc
+index 4205817b73f3d954be4109d91721037b3aa2d4ee..326cb6bcd0769646575f8b30d5327dfa6f747fda 100644
+--- a/src/execution/v8threads.cc
++++ b/src/execution/v8threads.cc
+@@ -2,6 +2,9 @@
+ // Use of this source code is governed by a BSD-style license that can be
+ // found in the LICENSE file.
+
++#include <unistd.h>
++#include <sys/syscall.h>
++
+ #include "src/execution/v8threads.h"
+
+ #include "include/v8-locker.h"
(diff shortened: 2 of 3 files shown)


jasnell (approved):

kentonv:
This seems a bit surprising -- the inspector protocol seems explicitly designed to run from a different thread, in order to support breakpoint debugging (which suspends the execution thread?).

(But hey if it works, ship it.)

harrishancock:
I guess our patches are Linux-specific:

Mac:
```
external/v8/src/libsampler/sampler.cc:604:11: error: use of undeclared identifier 'SYS_tgkill'
  syscall(SYS_tgkill, platform_data()->tgid(), lockedTid, SIGPROF);
          ^
```

Windows failed for a separate, probably spuriuos reason about an unexpected clang version, which I think @jasnell mentioned in chat. Presumably it won't appreciate `SYS_tgkill`, too, though.
```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\include\yvals_core.h(898,1): error: static assertion failed: error STL1000: Unexpected compiler version, expected Clang 17.0.0 or newer.
```

harrishancock:
> This seems a bit surprising -- the inspector protocol seems explicitly designed to run from a different thread, in order to support breakpoint debugging (which suspends the execution thread?).

It looks like it suspends execution by calling this `runMessageLoopOnPause()` function, which then explicitly pumps the inspector message loop until V8 tells it to stop:
https://github.com/cloudflare/workerd/blob/6a57e25fd87f8ce498e4cf293cf5074959ea30bb/src/workerd/io/worker.c%2B%2B#L406-L411

kentonv:
Could we perhaps actually deliver the profiler start/stop messages over to the isolate thread, rather than trying to start the profiler cross-thread?

harrishancock:
> Could we perhaps actually deliver the profiler start/stop messages over to the isolate thread, rather than trying to start the profiler cross-thread?

This is done in #2497.

PR #2234, opened 2024-06-07
Add logging if r2-bucket metadata hash sizes are not correct

jasnell:
Recently tried to make an optimization here but ran into some errors. Some of the hash sizes in some cases were not correct. Need to understand how common this is and determine if there's another problem somewhere.


Frederik-Baetens (approved):
Looks good to me. Parsing user generated checksums generally have a good amount of validation done against them, but we were intentionally lax when returning them. I'm curious how this issue is cropping up.

jasnell:
Yeah I was rather surprised to see the asserts. Hoping this gives us some good insight. Will merge now but likely won't get pulled into a release until next week sometime.

PR #2237, opened 2024-06-07
Bump supported compatibility date 06-10

garrettgu10:
Now that remote package loading was fixed, we should bump workerd so we can update the dependency from workers-sdk. This will also be an opportunity to actually implement disk caching in miniflare.


jasnell:
Can this be rebased on main again? I just landed a regression fix that would be nice to make sure is in this update

kentonv:
@jasnell Hmm, arguably we don't want to release that regression fix in workerd until it is actually live in prod, otherwise people are going to think their code works when they test locally only to see it break in prod...

harrishancock (approved):

PR #2243, opened 2024-06-08
Implement new bytes() method for Blob and Body objects

jasnell:
Refs: https://fetch.spec.whatwg.org/#body-mixin
Refs: https://w3c.github.io/FileAPI/#dfn-Blob

The `bytes()` method was recently added to `Blob` and the `Body` mixin.

### DIFF
--- a/src/workerd/api/blob.h
+++ b/src/workerd/api/blob.h
@@ -41,6 +41,7 @@ public:
                         jsg::Optional<kj::String> type);
 
   jsg::Promise<kj::Array<kj::byte>> arrayBuffer(jsg::Lock& js);
+  jsg::Promise<jsg::BufferSource> bytes(jsg::Lock& js);
   jsg::Promise<kj::String> text(jsg::Lock& js);
   jsg::Ref<ReadableStream> stream();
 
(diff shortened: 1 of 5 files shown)


src/workerd/api/blob.h
```diff
@@ -55,8 +56,13 @@ class Blob: public jsg::Object {
 
     JSG_METHOD(slice);
     JSG_METHOD(arrayBuffer);
+    JSG_METHOD(bytes);
     JSG_METHOD(text);
     JSG_METHOD(stream);
+
+    JSG_TS_OVERRIDE({
+      bytes(): Uint8Array;
```
a-robinson, line 64: It's way more likely you're right than I am, but why shouldn't this be a `Promise<Uint8Array>`?

Same question for the TS override in http.h as well

a-robinson (approved):

src/workerd/api/blob.h
jasnell: Doh. Yes lol

jasnell, line 64:
```suggestion
      bytes(): Promise<Uint8Array>;
```

PR #2244, opened 2024-06-09
Fix connect() SocketOptions type

DaniFoldi:
Hi :wave:,

I noticed that while the internal sockets type definitions https://github.com/cloudflare/workerd/blob/main/src/cloudflare/internal/sockets.d.ts and docs https://developers.cloudflare.com/workers/runtime-apis/tcp-sockets/#socketoptions
both type `secureTransport` as a proper enum, and say `allowHalfOpen` is optional, the generated types don't reflect this: https://workers-types.pages.dev/4.20240605.0/#SocketOptions

This PR is my attempt to fix this, although unlike my previous types fixes, this one was one bit more complicated so if it's using the wrong macro or missing a semicolon, apologies, I'll happily fix - compiling didn't work locally so my only source of confidence is the GH Action run on my fork.

Additionally, I found that somehow the copyright and license comment from only `vectorize.d.ts` is included: see just below https://workers-types.pages.dev/4.20240605.0/#webSocketError.connect, and while this is by no means important, it's annoying me that I couldn't yet find the reason.

### DIFF
--- a/src/workerd/api/sockets.h
+++ b/src/workerd/api/sockets.h
@@ -51,6 +51,9 @@ struct SocketOptions {
   JSG_MEMORY_INFO(SocketOptions) {
     tracker.trackField("secureTransport", secureTransport);
   }
+  JSG_STRUCT_TS_OVERRIDE({
+    secureTransport?: "off" | "on" | "starttls";
+  });
 };
 
 struct TlsOptions {


jasnell (approved):

src/workerd/api/sockets.h
```diff
@@ -43,6 +43,10 @@ struct SocketOptions {
   JSG_MEMORY_INFO(SocketOptions) {
     tracker.trackField("secureTransport", secureTransport);
   }
```
dom96, line 45:
```suggestion
  };
```

dom96 (approved):
Possibly just a missing semi-colon and it'll pass https://github.com/cloudflare/workerd/actions/runs/9439473980/job/26032713772?pr=2244#step:11:2809

src/workerd/api/sockets.h
DaniFoldi: Added a semicolon in the latest force push [04dae05](https://github.com/cloudflare/workerd/pull/2244/commits/04dae0534cbb42c6ebb81c2e0a76c81d8a84705f)

I read the docs for the JSG_TS macros (whew), but didn't check what the surrounding ones do 😁

DaniFoldi: Hmmm, seems like it's still failing, with my very limited knowledge about the macros, maybe putting the type override before the memory info one would help? Not sure if that would fit the overall code style either, but a quick ctrl+f showed that `JSG_MEMORY_INFO` is always last in a struct definition

DaniFoldi:
Hi, I completely forgot about this PR; I've rebased it and I think I found the reason why it couldn't compile last time.

PR #2245, opened 2024-06-09
Experimental implementation of the Observable API

jasnell:
The `Observable` API (https://wicg.github.io/observable) is an upcoming Web Platform Standard API that extends EventTarget with a flexible subscription model that is similar to that found in a number of popular frameworks.

This provides a WIP experimental implementation of the API in workerd.

The wintercg has been asked to consider a position on the `Observable` API and the best way to evaluate it is with running code. This would need a lot more testing and evaluation and the spec is still a work in progress so this isn't expected to land any time soon and will remain in draft state.

TODOs

* [ ] Tests tests and more tests (Likely ports of the Web platform tests)
* [ ] Verify that `Observable::from(...)` behavior is correct for sync iterables
* [ ] Documentation PR
* [ ] Determine if a JS/TS based implementation would be better (it's a more natural fit given the way the spec is defined and would be a smaller implementation). The `Observable` and `Subscriber` classes are expected to be available as globals but we can handle it the same way we handle exposure of the Node.js `Buffer` and `process` globals. There are limitations with that approach, however, in that it would mean the API is not available for the old service worker syntax workers.

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -5,6 +5,7 @@
 #include "actor-state.h"
 #include "basics.h"
 #include "global-scope.h"
+#include "observable.h"
 #include <kj/async.h>
 #include <kj/vector.h>
 #include <workerd/io/io-context.h>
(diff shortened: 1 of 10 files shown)


jasnell:
closing the PR for now as finishing this up is not an immediate priority.

PR #2250, opened 2024-06-10
include object name & version in log for r2 hash length mismatch

Frederik-Baetens:


src/workerd/api/r2-bucket.c++
```diff
@@ -116,39 +116,44 @@ static jsg::Ref<T> parseObjectMetadata(R2HeadResponse::Reader responseReader,
       auto md5 = checksumsBuilder.getMd5();
       if (md5.size() != 16) {
         KJ_LOG(WARNING, "NOSENTRY MD5 checksum is not 16 bytes long",
-               md5.size(), md5.asBytes());
+               md5.size(), md5.asBytes(),
+               responseReader.getName(), responseReader.getVersion());
```
jasnell, line 120: I guess if we're expanding the logging here moving it out to a separate utility method here to eliminate the duplication would be useful but I'm hoping this logging can be temporary so likely not worth changing further now :-)

jasnell (approved):

src/workerd/api/r2-bucket.c++
Frederik-Baetens: Haha, that's how a lot of code like this has stuck around in the past :sweat_smile: . Pushed a commit wrapping this in a little utility function.

Frederik-Baetens:
Merged, but let me know if you have any comments around the utility function I added.

PR #2256, opened 2024-06-11
[CI] Fix miniflare build

fhanau:
The issue is a re-edition of #2168, no longer defining the version directly
should stop this from reappearing.


fhanau:
@petebacondarwin @penalosa Was about to ask for ideas on coupling CI better to avoid issues like this in the future, but looks like there's a simple solution 😌.

petebacondarwin (approved):
🎉

PR #2262, opened 2024-06-12
Implements cloudflare:workers compatFlags API

jasnell:
A simple built-in module and API for determining if a given compat flag is set.

```
import { compatFlags } from 'cloudflare:workers'

// Any of the enable/disable flag names can be passed here.
// Invalid values come out as `undefined`.
console.log(compatFlags['url_standard']);
console.log(compatFlags['url_original']);
```

Having an API for this is something we've discussed a few times.

Note that this only works with ESM syntax workers given that it requires the module registry.

### DIFF
--- a/src/workerd/api/node/node.h
+++ b/src/workerd/api/node/node.h
@@ -9,6 +9,7 @@
 #include <workerd/jsg/modules.h>
 #include <capnp/dynamic.h>
 #include <node/node.capnp.h>
+#include <workerd/io/compatibility-date.h>
 
 namespace workerd::api::node {
 
@@ -17,7 +18,28 @@ namespace workerd::api::node {
 // built-ins
 class CompatibilityFlags : public jsg::Object {
 public:
+
+  jsg::JsObject getCompatFlags(jsg::Lock& js, workerd::CompatibilityFlags::Reader flags) {
+    auto obj = js.obj();
+    auto dynamic  = capnp::toDynamic(flags);
+    auto schema = dynamic.getSchema();
+    for (auto field : schema.getFields()) {
+      bool value = dynamic.get(field).as<bool>();
+      for (auto annotation : field.getProto().getAnnotations()) {
+        if (annotation.getId() == COMPAT_ENABLE_FLAG_ANNOTATION_ID) {
+          obj.setReadOnly(js, annotation.getValue().getText(), js.boolean(value));
+        }
+        else if (annotation.getId() == COMPAT_DISABLE_FLAG_ANNOTATION_ID) {
+          obj.setReadOnly(js, annotation.getValue().getText(), js.boolean(!value));
+        }
+      }
+    }
+    return obj;
+  }
+
   JSG_RESOURCE_TYPE(CompatibilityFlags, workerd::CompatibilityFlags::Reader flags) {
+    JSG_LAZY_READONLY_INSTANCE_PROPERTY(compatFlags, getCompatFlags);
+
     // Not your typical JSG_RESOURCE_TYPE definition.. here we are iterating
     // through all of the compatibility flags and registering each as read-only
     // literal values on the instance...
--- a/src/workerd/api/tests/compat-flags-test.js
+++ b/src/workerd/api/tests/compat-flags-test.js
@@ -1,0 +1,23 @@
+import {
+  ok,
+  throws,
+} from 'node:assert';
+
+import { compatFlags } from 'cloudflare:workers';
+
+export const compatFlagsTest = {
+  test() {
+    throws(() => compatFlags.no_nodejs_compat_v2 = true);
+    throws(() => compatFlags.not_a_real_compat_flag = true);
+    ok(compatFlags['nodejs_compat_v2']);
+    ok(!compatFlags['no_nodejs_compat_v2']);
+    ok(compatFlags['url_standard']);
+    ok(!compatFlags['url_original']);
+    ok(!compatFlags['not-a-real-compat-flag']);
+    const keys = Object.keys(compatFlags);
+    ok(keys.includes('nodejs_compat_v2'));
+    ok(keys.includes('url_standard'));
+    ok(keys.includes('url_original'));
+    ok(!keys.includes('not-a-real-compat-flag'));
+  }
+}
--- a/src/cloudflare/workers.ts
+++ b/src/cloudflare/workers.ts
@@ -11,3 +11,9 @@ export const WorkerEntrypoint = entrypoints.WorkerEntrypoint;
 export const DurableObject = entrypoints.DurableObject;
 export const RpcStub = entrypoints.RpcStub;
 export const RpcTarget = entrypoints.RpcTarget;
+
+/* eslint-disable */
+import { default as flags } from 'workerd:compatibility-flags';
+export const compatFlags = (flags as any).compatFlags;
+Object.freeze(compatFlags);
+/* eslint-enable */
--- a/src/workerd/api/tests/compat-flags-test.wd-test
+++ b/src/workerd/api/tests/compat-flags-test.wd-test
@@ -1,0 +1,15 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "compat-flags-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "compat-flags-test.js")
+        ],
+        compatibilityDate = "2023-01-15",
+        compatibilityFlags = ["nodejs_compat_v2"],
+      )
+    ),
+  ],
+);
--- a/src/workerd/jsg/jsvalue.c++
+++ b/src/workerd/jsg/jsvalue.c++
@@ -58,6 +58,11 @@ void JsObject::set(Lock& js, kj::StringPtr name, const JsValue& value) {
   set(js, js.strIntern(name), value);
 }
 
+void JsObject::setReadOnly(Lock& js, kj::StringPtr name, const JsValue& value) {
+  v8::Local<v8::String> nameStr = js.strIntern(name);
+  check(inner->DefineOwnProperty(js.v8Context(), nameStr, value, v8::ReadOnly));
+}
+
 JsValue JsObject::get(Lock& js, const JsValue& name) {
   return JsValue(check(inner->Get(js.v8Context(), name.inner)));
 }
--- a/src/workerd/jsg/jsvalue.h
+++ b/src/workerd/jsg/jsvalue.h
@@ -329,6 +329,7 @@ public:
 
   void set(Lock& js, const JsValue& name, const JsValue& value);
   void set(Lock& js, kj::StringPtr name, const JsValue& value);
+  void setReadOnly(Lock& js, kj::StringPtr name, const JsValue& value);
   JsValue get(Lock& js, const JsValue& name) KJ_WARN_UNUSED_RESULT;
   JsValue get(Lock& js, kj::StringPtr name) KJ_WARN_UNUSED_RESULT;
 


garrettgu10 (approved):
Nice!

kentonv:
Maybe we want a more specific name than `isEnabled`? Like `usingCompatFlag`?

I know it's scoped to the module but most people will import it into their module's top-level scope, like in your example.

I guess relatedly... with a more-specific name, we could put the export into the `cloudflare:workers` module rather than having a whole module just for this.

jasnell:
> Maybe we want a more specific name than isEnabled? Like usingCompatFlag?..... , we could put the export into the cloudflare:workers module rather than having a whole module just for this.

Both done.

src/workerd/api/node/node.h
```diff
@@ -17,7 +18,37 @@ namespace workerd::api::node {
...
+        else if (annotation.getId() == COMPAT_DISABLE_FLAG_ANNOTATION_ID) {
+          if (flag == annotation.getValue().getText()) {
+            bool enabled = !dynamic.get(field).as<bool>();
+            cachedFlags.upsert(kj::mv(flag), enabled);
+            return enabled;
+          }
+        }
+      }
+    }
+    return false;
```
IgorMinar, line 46: why is defaulting to `false` a good thing? wouldn't that mean that I can easily have a typo in the flag and get a false negative response?

For example: `nodejs_compattt` would yield `false` even when `node_compat` is turned on.

src/workerd/api/node/node.h
```diff
@@ -17,7 +18,37 @@ namespace workerd::api::node {
 // built-ins
 class CompatibilityFlags : public jsg::Object {
 public:
+
+  bool isEnabled(kj::String flag, workerd::CompatibilityFlags::Reader flags) {
```
IgorMinar, line 22: rename to `usingCompatFlag` for consistency?

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,15 @@
+import {
+  ok,
+} from 'node:assert';
+
+import { usingCompatFlag } from 'cloudflare:workers';
+
+export const compatFlags = {
+  test() {
+    ok(usingCompatFlag('nodejs_compat_v2'));
```
IgorMinar, line 9: if I use this as `usingCompatFlag('no_nodejs_compat_v2')` will `false` result mean that `node_compat_v2` is not active or that the user didn't configure workerd with `no_nodejs_compat_v2` (which are the exact opposites).

Could we tweak the API such that this ambiguity doesn't exist?

IgorMinar: have you considered to using imported boolean constants instead of `usingCompatFlag` and string arg here? For example:

```
import { NODEJS_COMPAT_V2 } from 'cloudflare:workers';

...
if (NODEJS_COMPAT_V2) {
  console.log('yay! you are using nodejs_compat_v2 mode!!');
}
```

IgorMinar:
@jasnell this is a cool feature! I left some suggestions, but I think this feature could be useful in some scenarios where generated code or 3rd party library depends on a particular compat flag.

src/workerd/api/node/node.h
jasnell: Yep. What would you rather it do? My preference would be not to have it throw. Returning `undefined` instead of `false` allows falsy checks to still happen but if someone really needs to differentiate they could.

src/workerd/api/tests/compat-flags-test.js
jasnell: > if I use this as usingCompatFlag('no_nodejs_compat_v2') will false result mean that node_compat_v2 is not active or that the user didn't configure workerd with no_nodejs_compat_v2 (which are the exact opposites).

If you specify the `nodejs_compat_v2` flag... then `usingCompatFlag('no_nodejs_compat_v2')` would return `false` and `usingCompatFlag('nodejs_compat_v2')` would return true. Inversely, if you specify `no_nodejs_compat_v2` then `usingCompatFlag('nodejs_compat_v2')` would be false and `usingCompatFlag('no_nodejs_compat_v2')` would be true.

> have you considered to using imported boolean constants instead of usingCompatFlag and string arg here?

Yeah, that's basically what we do for the internal modules. I didn't go that route here to avoid adding too many new symbols exported on the module. I can go either way with it but I tend to prefer this approach. If someone wants const values they can easily define those themselves. Then again, constants would perform better since those would be initialized only once and there's no need for the caching.

jasnell: Ok, updated to replace the method with an object full of constants...

```js
import { compatFlags } from 'cloudflare:workers';

console.log(compatFlags['nodejs_compat_v2']);
```

The `compatFlags` is an object like...

```
{
  formdata_parser_supports_files: true,
  formdata_parser_converts_files_to_strings: false,
  fetch_refuses_unknown_protocols: true,
  fetch_treats_unknown_protocols_as_http: false,
  html_rewriter_treats_esi_include_as_void_tag: false,
  durable_object_fetch_requires_full_url: true,
  durable_object_fetch_allows_relative_url: false,
  streams_byob_reader_detaches_buffer: true,
  streams_byob_reader_does_not_detach_buffer: false,
  streams_enable_constructors: true,
  streams_disable_constructors: false,
  workers_api_getters_setters_on_prototype: true,
  workers_api_getters_setters_on_instance: false,
  minimal_subrequests: true,
  no_minimal_subrequests: false,
  no_cots_on_external_fetch: true,
  cots_on_external_fetch: false,
  url_standard: true,
  url_original: false,
  global_navigator: true,
  no_global_navigator: false,
  capture_async_api_throws: true,
  do_not_capture_async_api_throws: false,
  r2_public_beta_bindings: true,
  r2_internal_beta_bindings: false,
  durable_object_alarms: false,
  dont_substitute_null_on_type_error: true,
  substitute_null_on_type_error: false,
  transformstream_enable_standard_constructor: true,
  transformstream_disable_standard_constructor: false,
  r2_list_honor_include: true,
  export_commonjs_default: true,
  export_commonjs_namespace: false,
  durable_object_rename: false,
  web_socket_compression: false,
  no_web_socket_compression: true,
  nodejs_compat: false,
  no_nodejs_compat: true,
  tcp_sockets_support: false,
  response_redirect_url_standard: false,
  response_redirect_url_original: true,
  experimental: false,
  durable_object_get_existing: false,
  http_headers_getsetcookie: false,
  no_http_headers_getsetcookie: true,
  dynamic_dispatch_tunnel_exceptions: false,
  dynamic_dispatch_treat_exceptions_as_500: true,
  service_binding_extra_handlers: false,
  no_cf_botmanagement_default: false,
  cf_botmanagement_default: true,
  urlsearchparams_delete_has_value_arg: false,
  no_urlsearchparams_delete_has_value_arg: true,
  strict_compression_checks: false,
  no_strict_compression_checks: true,
  brotli_content_encoding: false,
  no_brotli_content_encoding: true,
  strict_crypto_checks: false,
  no_strict_crypto_checks: true,
  rtti_api: false,
  webgpu: false,
  crypto_preserve_public_exponent: false,
  no_crypto_preserve_public_exponent: true,
  vectorize_query_metadata_optional: false,
  vectorize_query_original: true,
  unsafe_module: false,
  js_rpc: false,
  no_global_importscripts: false,
  global_importscripts: true,
  nodejs_als: false,
  no_nodejs_als: true,
  queues_json_messages: false,
  no_queues_json_messages: true,
  python_workers: false,
  fetcher_no_get_put_delete: false,
  fetcher_has_get_put_delete: true,
  unwrap_custom_thenables: false,
  no_unwrap_custom_thenables: true,
  rpc: false,
  no_rpc: true,
  internal_stream_byob_return_view: false,
  internal_stream_byob_return_undefined: true,
  blob_standard_mime_type: false,
  blob_legacy_mime_type: true,
  fetch_standard_url: false,
  fetch_legacy_url: true,
  nodejs_compat_v2: true,
  no_nodejs_compat_v2: false
}
```

src/workerd/api/node/node.h
jasnell: I've reworked it to export constants instead. A flag that doesn't exist will return `undefined` and won't appear as a key in the object.

GregBrimble:
Yahoo! I have wanted this for _ages_. I thought it was going to be tucked away under `navigator` somewhere, but this is miles better and means we can continue to expand on introspective stuff like this more easily if the need arises.

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const compatFlagsTest = {
+  test() {
+    throws(() => compatFlags.no_nodejs_compat_v2 = true);
+    throws(() => compatFlags.not_a_real_compat_flag = true);
+    ok(compatFlags['nodejs_compat_v2']);
+    ok(!compatFlags['no_nodejs_compat_v2']);
+    ok(compatFlags['url_standard']);
+    ok(!compatFlags['url_original']);
+    ok(!compatFlags['not-a-real-compat-flag']);
```
dario-piotrowicz, line 16:
```suggestion
    ok(!compatFlags['not-a-real-compat-flag']);
    const entries = Object.entries(compatFlags);
    ok(entries.length > 0);
    ok(entries.find(([key,]) => key === 'nodejs_compat_v2')[1]);
    ok(!entries.find(([key,]) => key === 'no_nodejs_compat_v2')[1]);
```
or something along those lines, so that we can also test that the entries are enumerable? 🙂
(on the top of my head I can find of a great use case example for this, but I find the enumerability of the flags to be a nice to have feature 😄)

jasnell: Easiest would be just to call `Object.keys(compatFlags)` and check for specific keys in the resulting array.

dario-piotrowicz: yes, that would work too 🙂 (since you're already checking the values on the lines above anyways)

jasnell: Done

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,23 @@
+import {
+  ok,
+  throws,
+} from 'node:assert';
+
+import { compatFlags } from 'cloudflare:workers';
```
IgorMinar, line 6: for consistency with wrangler.toml, and capnp, I think we should call this symbol `compatibilityFlags`.

kentonv:
This was rolled back as I think the revised design that landed has a couple issues we need to resolve:

* Say a new flag comes along called `foo` and has a corresponding negative flag `no_foo`. Now I write `if (compatibilityFlags["no_foo"]) {` to guard some code that should only run with the old behavior. However, I then run my code on an old version of workerd where the `foo` flag isn't even defined yet. `compatibilityFlags["no_foo"]` will evaluate to `undefined`, which means my check will do the wrong thing -- it won't run the code, even though `foo` is not enabled (or even known about).

    One way to fix this would be to throw an exception on unknown flags, although this would unfortunately cause code which checks for new flags not to be able to run at all on old versions. Probably instead we need to change the API to use "positive" flags only. That is, you cannot check for "no_foo", you have to check for "foo" and negate the result. This way, the check evaluates correctly on old versions.

    It's too bad that this might lead to confusion given that negative flags can be used everywhere else where compat flags are used. Perhaps we can throw an exception if someone tries to check a known negative flag, to help them understand what went wrong.

* I have a few concerns with returning an object containing every known flag. It seems like it'll become less and less efficient as time goes on, though perhaps the cost is negligible. It also seems weird to me that it exposes the existence of experimental flags even to people who cannot use them in production. These flags are sometimes removed later -- will there be compatibility concerns with removing a flag that no one was allowed to use, just because it disappears from this object? Maybe a minor thing but the function call API felt better to me. OTOH maybe there is an argument that code which frequently checks flags will be faster using the object form, because V8 can more easily inline the check? (Such code could also store its own bool, though.)

IgorMinar:
@kentonv I'm in favor of thinking this one through a bit more.

I've thought about the version skew issue, I would think that by the time anyone writes code that checks for a new flag, we are already committed to the flag and it's fully rolled out in prod. There could be an issue with version skew in local development where someone could be pinned to an old version of workerd, but in that case we should just force them to update as their local dev is too far remote from the production version.

As for function access, I'd like to avoid it if we can, with constants we could actually tree-shake the unused branches at build time which could be interesting down the road, function design will not allow for that. A possible middle ground is for us to expose the flags as exported symbols as I suggested earlier: https://github.com/cloudflare/workerd/pull/2262#discussion_r1637184371 but this would still allow someone to iterate over all the flags using `*` imports.

In any case, I'd like to included on the review of future proposals of this feature as there are lots of small intricate details that matter to libraries and frameworks that we should consider.

PR #2271, opened 2024-06-14
Annotate exceptions arriving over JSRPC as "remote"

jclee:
...so that they get the expected ".remote" property when tunneled to JS, like exceptions from Durable Object fetch() calls do.

### DIFF
--- a/src/workerd/api/tests/js-rpc-test.js
+++ b/src/workerd/api/tests/js-rpc-test.js
@@ -1341,6 +1341,18 @@ export let testAsyncStackTrace = {
   }
 }
 
+// Test that exceptions thrown over RPC have the .remote property.
+export let testExceptionProperties = {
+  async test(controller, env, ctx) {
+    try {
+      await env.MyService.throwingMethod();
+    } catch (e) {
+      assert.strictEqual(e.remote, true);
+      assert.strictEqual(e.message, "METHOD THREW");
+    }
+  }
+}
+
 // Test that get(), put(), and delete() are valid RPC method names, not hijacked by Fetcher.
 export let canUseGetPutDelete = {
   async test(controller, env, ctx) {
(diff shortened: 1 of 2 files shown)


jasnell:
Any possibility of having tests with this?

jclee:
> Any possibility of having tests with this?

- [ ] I've got some on the internal PR... Will try adding workerd-specific coverage.

jasnell (approved):
Lgtm assuming internal ci is good

justin-mp (approved):

kentonv:
A couple inconsistencies with the way "remote" is handled for fetch():

1. Here you've placed the logic on the client side to add the remote annotation, but for fetch() it happens server-side (in WorkerEntrypoint). Maybe we should keep these the same? I think this means putting the logic in `JsRpcTargetBase::call()` instead.
2. WorkerEntrypoint only adds the annotation to JSG exceptions, not to internal errors, but your code here doesn't seem to distinguish.

jclee:
> A couple inconsistencies with the way "remote" is handled for fetch():
> 
> 1. Here you've placed the logic on the client side to add the remote annotation, but for fetch() it happens server-side (in WorkerEntrypoint). Maybe we should keep these the same? I think this means putting the logic in `JsRpcTargetBase::call()` instead.
> 2. WorkerEntrypoint only adds the annotation to JSG exceptions, not to internal errors, but your code here doesn't seem to distinguish.

- [ ] Ah, makes sense... taking a look.

jclee:
I moved the `.remote` annotation from the client side of the RPC boundary to the server side, and added a check to ensure only tunneled (`jsg.`-prefixed) exceptions get the annotation.  (I wasn't sure how to test non-remote exceptions from within the wd-test framework, but I managed to add some coverage for it on the internal PR.)

kentonv (approved):

PR #2275, opened 2024-06-14
JSRPC: Fix misc bugs

kentonv:
All of these were spotted by @jclee.

### DIFF
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -1390,14 +1390,15 @@ class TransientJsRpcTarget final: public JsRpcTargetBase {
 public:
   TransientJsRpcTarget(jsg::Lock& js, IoContext& ioCtx, jsg::JsObject object,
                        bool allowInstanceProperties = false)
-      : JsRpcTargetBase(ioCtx), object(js, object),
+      : JsRpcTargetBase(ioCtx),
+        handles(ioCtx.addObjectReverse(kj::heap<Handles>(js, object, kj::none))),
         allowInstanceProperties(allowInstanceProperties),
         pendingEvent(ioCtx.registerPendingEvent()) {
     // Check for the existence of a dispose function now so that the destructor doesn't have to
     // take an isolate lock if there isn't one.
     auto getResult = object.get(js, js.symbolDispose());
     if (getResult.isFunction()) {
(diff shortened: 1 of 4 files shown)


jasnell:
hmm... it would be helpful for reviewers if the specific bugs could be summarized a bit here. Someone more familiar with the code can pick them out but someone newer to the codebase likely wouldn't find them obvious

src/workerd/io/io-own.h
```diff
@@ -236,6 +258,35 @@ class IoPtr {
       : deleteQueue(kj::mv(deleteQueue)), ptr(ptr) {}
 };
 
+// Owned pointer held by a KJ I/O object living in the same thread as an IoContext. The underlying
+// object is destroyed when the ReverseIoOwn is dropped OR when the IoContext is destroyed,
+// whichever comes first. Accessing the ReverseIoOwn after the IoContext is destroyed will throw.
```
jasnell, line 263: This could use a bit more detail on when/why someone would use a `ReverseIoOwn<T>` as opposed to an `IoOwn<T>` or whatnot. The `IoOwn<T>` on its own has already been a source of confusion for folks so I'd err on the side of over-documenting than under.

src/workerd/api/worker-rpc.c++
```diff
@@ -1429,15 +1426,24 @@ public:
...
+    jsg::JsRef<jsg::JsObject> object;
+    kj::Maybe<jsg::V8Ref<v8::Function>> dispose;
+
+    Handles(jsg::Lock& js, jsg::JsObject object, kj::Maybe<v8::Local<v8::Function>> dispose)
+        : object(js, object),
+          dispose(dispose.map([&](v8::Local<v8::Function> func) {
+            return jsg::V8Ref<v8::Function>(js.v8Isolate, func);
+          })) {}
+  };
+  ReverseIoOwn<Handles> handles;
```
jasnell, line 1446: A comment here explaining why the `ReverseIoOwn<T>` is necessary here would be helpful for those coming into this code in the future.

jasnell (approved):
Would like to see the doc comments expanded a bit as discussed in comments but otherwise LGTM

jclee (approved):

kentonv:
Fixups:
* 89f7794c
* 3c631dac

PR #2281, opened 2024-06-17
Implement node.js buffer.isAscii|isUtf8

jasnell:
This change adds two new methods to the node.js Buffer class: isAscii and isUtf8. These methods are used to check if the buffer contains only ASCII or UTF-8 characters, respectively.

The implementation of these methods is based on the [simdutf](https://github.com/simdutf/simdutf) library, just like in Node.js itself.


src/workerd/api/node/buffer.h
```diff
@@ -77,6 +77,8 @@ class BufferUtil final: public jsg::Object {
                         kj::Array<kj::byte> bytes,
                         kj::Array<kj::byte> state);
   jsg::JsString flush(jsg::Lock& js, kj::Array<kj::byte> state);
+  bool isAscii(kj::Array<kj::byte> bytes);
```
IgorMinar, line 80: does this mean that the argument can be any of `<Buffer> | <ArrayBuffer> | <TypedArray> ` as in node? https://nodejs.org/api/buffer.html#buffer_buffer_isascii_input

IgorMinar (approved):
one q, but the rest of JS/public API surface part of this PR looks good to me. Thanks @jasnell!

src/workerd/api/node/buffer.h
jasnell: yes. the one difference is that I'm not checking to see if the underlying ArrayBuffer has been transfered

IgorMinar: Is that intentional?

jasnell: For now, yes, since we have significantly less use case for transfered buffers since we do not have worker threads. We do have `structuredClone(...)` and `ArrayBuffer`s can be transfered with streams but it's likely quite unlikely for it to be an issue.

IgorMinar: could you lave a note in the code as a comment? this kind of context is not obvious and will become handy down the road. thank you

src/workerd/api/node/buffer.c++
```diff
@@ -862,5 +863,15 @@ jsg::JsString BufferUtil::flush(jsg::Lock& js, kj::Array<kj::byte> state) {
   return ret;
 }
 
+bool BufferUtil::isAscii(kj::Array<kj::byte> buffer) {
+  if (buffer.size() == 0) return true;
```
jasnell, line 867:
```suggestion
  // The Node.js implementation has an additional check that will throw an error if
  // the input ArrayBuffer/ArrayBufferView has been detached/transferred. We're
  // not implementing that additional check here yet. Instead, we'll treat those the
  // same as zero-length. We can potentially add the check in later (with a compat
  // flag). 
  if (buffer.size() == 0) return true;
```

src/workerd/api/node/buffer.c++
```diff
@@ -862,5 +863,20 @@ jsg::JsString BufferUtil::flush(jsg::Lock& js, kj::Array<kj::byte> state) {
   return ret;
 }
 
+bool BufferUtil::isAscii(kj::Array<kj::byte> buffer) {
+  // The Node.js implementation has an additional check that will throw an error if
+  // the input ArrayBuffer/ArrayBufferView has been detached/transferred. We're
+  // not implementing that additional check here yet. Instead, we'll treat those the
+  // same as zero-length. We can potentially add the check in later (with a compat
+  // flag).
```
dom96, line 871: Why not do it now?

dom96 (approved):

PR #2285, opened 2024-06-18
make max timeout log more informative

Frederik-Baetens:


jasnell (approved):

kentonv:
This is a public-facing error message. If we're going to add stuff like this to it we need to make it really clear what these numbers mean or people will get confused. I don't think the current formatting is that clear -- it reports "max timeouts" and "finished timeouts" but doesn't explain that the maximum only applies to timeouts that haven't finished.

I'm also not sure that this will really tell us anything new about the specific case we're trying to debug...

kentonv:
(Of course, the old formatting of this error was not great either.)

Frederik-Baetens:
I added "active timeouts" to this, I think that'll make it clearer that those are the important timeouts when looking at this message.

Frederik-Baetens:
Further changed wording to "max active timeouts", "current active timeouts" and "You have exceeded the number of active timeouts you may set." to further clarify this error message.

Agree that it doesn't add much, but I'm hoping that subsequent occurrences of this error from within the same DO but with sleeps in between can maybe give us a clue about the rate at which timeouts are finishing.

PR #2287, opened 2024-06-18
Fix addExceptionDetail() when execution is being terminated.

kentonv:
In this case `js.tryCatch()` wouldn't actually catch the `JsExceptionThrown`. But we actually want to ignore this type of error too.


jasnell (approved):

kentonv:
Added another JsExceptionThrown fix.

Issue #2181, opened 2024-05-29
Improve Workers AI types


Cherry:
Today, the Workers AI types rely heavily on function overloads to specify arguments for different models. This unfortunately results in very difficult to debug types, and poor DX.

As an example with a more simplified non-AI class: https://tsplay.dev/Nan9ym

Throughout this during development, you get very little assistance with auto-complete, and the errors you get back, are extremely complex and hard to parse:
```
No overload matches this call.
  Overload 1 of 3, '(fruit: "apple" | "orange", color: "apple" | "orange", options: { foo: boolean; bar: boolean; }): Promise<void>', gave the following error.
    Argument of type '"lemon"' is not assignable to parameter of type '"apple" | "orange"'.
  Overload 2 of 3, '(fruit: "tomato" | "pear", color: "tomato" | "pear", options: { foo: string; bar?: boolean | undefined; }): Promise<void>', gave the following error.
    Argument of type '"lemon"' is not assignable to parameter of type '"tomato" | "pear"'.
  Overload 3 of 3, '(fruit: "lemon" | "lime", color: "lemon" | "lime", options: { foo?: string | undefined; bar: number; somethingElse: boolean; }): Promise<void>', gave the following error.
    Type 'null' is not assignable to type 'number'.
```

Let's use some real AI examples. These can all be found in https://tsplay.dev/wQB41N - scroll down to line 255 for examples.

## Invalid model names

```ts
AI.run('@hf/thebloke/neural-chat-7b-v3-awq', {
    messages: [
        {
            role: "system",
            content: `[very large system prompt]`
        },
        {
            role: "user",
            content: `How many pounds of food does a person eat in a year?`
        }
    ],
    stream: false
});
```
This errors with:
```
No overload matches this call.
  The last overload gave the following error.
    Argument of type '"@hf/thebloke/neural-chat-7b-v3-awq"' is not assignable to parameter of type '"@cf/unum/uform-gen2-qwen-500m" | "@cf/llava-hf/llava-1.5-7b-hf"'.
```

This is because I typo'd the model name and it should be `@hf/thebloke/neural-chat-7b-v3-1-awq`.

## Bad inputs

```ts
AI.run('@hf/thebloke/neural-chat-7b-v3-1-awq', {
    message: [
        {
            role: "system",
            content: `[very large system prompt]`
        },
        {
            role: "user",
            content: `How many pounds of food does a person eat in a year?`
        }
    ],
    stream: false
});
```
[...]

PR #2290, opened 2024-06-19
chore: added function calling types to workerd

Dhravya:


celso (approved):

Dhravya:

celso (approved):

jasnell (approved):

types/defines/ai.d.ts
```diff
@@ -149,108 +184,124 @@ export type AiOptions = {
...
+  | "@cf/fblgit/una-cybertron-7b-v2-awq"
+  | string;
+export type BaseAiTranslationModels = "@cf/meta/m2m100-1.2b" | string;
+export type BaseAiSummarizationModels = "@cf/facebook/bart-large-cnn" | string;
+export type BaseAiImageToTextModels =
+  | "@cf/unum/uform-gen2-qwen-500m"
+  | "@cf/llava-hf/llava-1.5-7b-hf"
+  | string;
 export declare abstract class Ai {
+  fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response>;
```
penalosa, line 257: Is this intended?

penalosa (changes requested):

types/defines/ai.d.ts
penalosa: Last I remember `fetch` had been explicitly excluded from the types?

celso: > Last I remember `fetch` had been explicitly excluded from the types?

I think you're right.

Dhravya: I removed the `fetch`

types/defines/ai.d.ts
```diff
@@ -149,108 +184,123 @@ export type AiOptions = {
   prefix?: string;
   extraHeaders?: object;
 };
+export type BaseAiTextClassificationModels =
+  | "@cf/huggingface/distilbert-sst-2-int8"
+  | string;
```
Cherry, line 189: Please correct me if I'm wrong, but as far as I know, adding `string` to every type here is going to fundamentally break the method overloading that occurs below, and therefore all of the `.run` types. All of these types are effectively now just `string` to TypeScript - the specific strings before them don't matter.

Even if you do `| string & {}` to enable smart auto-complete, I don't think that'll work well with method overriding either.

Example:
```typescript
type fruits = 'apple' | 'orange' | 'banana' | string;
type colors = 'green' | 'orange' | 'yellow' | string;

declare class Foo {
  run(item: fruits, other: number): void;
  run(item: colors, other: boolean): void;
}

const a = new Foo();

// this should type error, it doesn't
a.run('green', 1);

// this too
a.run('apple', false);
```

Because effectively, all I've done here with my function overloading is:
```typescript
declare class Foo {
  run(item: string, other: number): void;
  run(item: string, other: boolean): void;
}
```

(The same problem exists if you use `string & {}` here too.

The method overloading approach to these types is complex and flaky (see https://github.com/cloudflare/workerd/issues/2181 for details), but it _does_ mostly work once you get the model name situated, even if the specific errors aren't super helpful.

Dhravya: I got it, thanks a lot for spotting that. 
Is there a good solution for this? We essentially want to add an overload for string types, but type suggestions for models as well, as new models are added all the time

Dhravya: Something like this, maybe?
```
type Fruits = 'apple' | 'orange' | 'banana';
type Colors = 'green' | 'orange' | 'yellow';

declare class Foo {
  run(item: Fruits | (string & {}), other: number): void;
  run(item: Colors | (string & {}), other: boolean): void;
}

const a = new Foo();

// These should not cause a type error and provide type completions for fruits
a.run('banana', 1);

a.run('apple', 2);

a.run('green', true);
a.run('pineple', 1);
a.run('banana', 24);
```

celso: The goal here is not to enforce types on model names, that'd be hard to maintain, the catalog changes all the time (new models every week), but to facilitate auto-completion.

Cherry: With the method signature of the `run` function so heavily relying on overloads today, I don't think there's an ideal way to do this.

The following pseudo-example works:
```typescript
type fruits = 'apple' | 'orange' | 'banana';
type colors = 'green' | 'orange' | 'yellow';

declare class Foo {
  run(item: fruits, other: number): void;
  run(item: colors, other: boolean): void;
  run(item: string, other: string): void;
}

const a = new Foo();

// this should type error
a.run('green', 1);

// this should error too
a.run('apple', false);

// this shouldn't error
a.run('green', true);

// this also shouldn't error
a.run('banana', 2);

// and this shouldn't error either
a.run('foobar', 'foo');

// but this should error
a.run('foobar', 2);
```

But it does require all overloads to be disjunct and the last overload to be a union of all the option types - I think this is true of the `AI.run` function today so might work here?

A more complex solution might use `NoInfer`, but that's a very new TypeScript feature. 

It's too late to redesign this function signature without more thought and deprecation etc., but different functions for different models, or even just a singular `options` arg that includes the `model`, `inputs`, etc. would make this kind of thing much much easier and avoid the need for overloads at all.

Cherry: > The goal here is not to enforce types on model names, that'd be hard to maintain, the catalog changes all the time (new models every week), but to facilitate auto-completion.

@celso This would be a pretty significant change to the types and developer expected behaviour up until this point where that's _exactly_ what the types have done, and is generally the whole purpose of having a TypeScript library/SDK etc.

Why not simply just use `model: string, inputs: any` if that's the goal here and drop all the definitions entirely? That would obviously not be ideal, but that's essentially what you're going to do if your intent is now "not to enforce types on model names" and land this change.

I can't overstate how much this knee-capping of the types to be essentially useless and no longer provide any type safety for `inputs` would be a very bad idea. Please reconsider your goals and approach here. Why not perhaps a `AI.run<tasktype>(...)` signature that gives developers type safety for inputs?

Dhravya: I agree, for now, since the release is kinda urgent, I have removed the `| string` and this is something we'll consider later

penalosa (approved):

PR #2292, opened 2024-06-19
Add frameworks team to node CODEOWNERS

jasnell:


.github/CODEOWNERS
```diff
@@ -8,3 +8,4 @@ pnpm-lock.yaml @cloudflare/wrangler @cloudflare/workers-runtime-1
 types/ @cloudflare/wrangler @cloudflare/workers-runtime-1
 src/workerd/tools/ @cloudflare/wrangler @cloudflare/workers-runtime-1 @cloudflare/workers-durable-objects
 src/workerd/io/supported-compatibility-date.txt @cloudflare/wrangler @cloudflare/workers-runtime-1
+src/node/ @cloudflare/workers-frameworks
```
fhanau, line 11:
```suggestion
src/node/ @cloudflare/workers-runtime-1 @cloudflare/workers-durable-objects @cloudflare/workers-frameworks
```

Later lines overwrite permissions instead of adding to them – we need to still list the Runtime and DO teams here.

fhanau (approved):
LGTM otherwise. We should consider granting wider access since any change including files outside of src/node will still need a reviewer outside of the Frameworks team.

PR #2297, opened 2024-06-19
Actor exception autogate cleanup

jclee:
Cleans up the "updated-actor-exception-types" and "actor-exception-properties" autogates

### DIFF
--- a/src/workerd/util/autogate.h
+++ b/src/workerd/util/autogate.h
@@ -13,7 +13,6 @@ namespace workerd::util {
 // Workerd-specific list of autogate keys (can also be used in internal repo).
 enum class AutogateKey {
   TEST_WORKERD,
-  ACTOR_EXCEPTION_PROPERTIES, // adds .retryable and .overloaded properties to tunneled exceptions
   NumOfKeys // Reserved for iteration.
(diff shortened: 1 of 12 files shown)


src/workerd/util/autogate.h
```diff
@@ -42,9 +40,6 @@ class Autogate {
   static void initAutogate(
       capnp::List<capnp::Text>::Reader autogates);
 
-  // Convenience method for tests to use to invoke initAutogate()
-  static void initEmptyAutogateForTesting();
```
jasnell, line 46: Might be a good idea to keep this for future use, but not-blocking.

jclee: Yeah, I wasn't sure how much of the autogate-specific changes to leave in.

For this function in particular, though, it occurred to me after the fact that it might be more broadly useful if it were implemented as something taking a list of autogate name strings.  I can try adding that in a separate PR.

jasnell (approved):

dom96 (approved):
Nice, looks good. Thanks for cleaning these up!

justin-mp (approved):

PR #2301, opened 2024-06-20
Support case-insensitive encoding in TextDecoder API (Fixes #2300)

### LINKED ISSUE #2300: 🐛 Bug Report — Runtime APIs TextDecoder
TextDecoder does support `new TextDecoder("utf-8")` but does not support `TextDecoder("UTF-8")`

fhanau:
The Encoding [spec](https://encoding.spec.whatwg.org/#names-and-labels) requires that case-insensitive labels are supported.

### DIFF
--- a/src/workerd/api/encoding.c++
+++ b/src/workerd/api/encoding.c++
@@ -3,6 +3,7 @@
 //     https://opensource.org/licenses/Apache-2.0
 
 #include "encoding.h"
+#include "util.h"
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/buffersource.h>
 #include <unicode/ucnv.h>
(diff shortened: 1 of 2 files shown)


src/workerd/api/encoding.c++
```diff
@@ -256,6 +257,7 @@ kj::StringPtr getEncodingId(Encoding encoding) {
 }
 
 Encoding getEncodingForLabel(kj::StringPtr label) {
+  kj::String labelInsensitive = toLower(label);
```
fhanau, line 260: This calls `kj::String toLower(kj::String&& str)` in util.h. Calling that function with a kj::StringPtr is safe and just results in the string being implicitly copied, right?

jasnell (approved):

src/workerd/api/encoding.c++
fhanau: Turns out that this is using the `jsg::ByteString toLower(kj::StringPtr str);` constructor in the same file instead. This feels wonky, but works correctly for this use case.

PR #2304, opened 2024-06-20
Specify pnpm version in release action

garrettgu10:
The Miniflare testing was broken and complaining about this. Not sure how to test without merging it in first, but CI is already broken on main so 🤷

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,8 +175,8 @@ jobs:
 
       - name: Install pnpm
         uses: pnpm/action-setup@v3
-        # version is no longer being defined here, since we can automatically use the version
-        # defined in workers-sdk/package.json instead.
+        with:
+          version: 9.1.3
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:


.github/workflows/release.yml
```diff
@@ -175,8 +175,8 @@ jobs:
 
       - name: Install pnpm
         uses: pnpm/action-setup@v3
-        # version is no longer being defined here, since we can automatically use the version
-        # defined in workers-sdk/package.json instead.
+        with:
+          version: 9.1.3
```
fhanau, line 179: Using `~9.1.3` or `^9.1.3` might delay the time when this will be broken next – or that syntax might be unsupported and just won't work, we can go ahead with this for now.

fhanau (approved):

fhanau:
@petebacondarwin @CarmenPopoviciu #2256 was supposed to provide a durable fix here but didn't work. Based on https://github.com/pnpm/action-setup?tab=readme-ov-file#version the version field should being optional here, but it looks like the `packageManager` field would have to be set in Wrangler, not sure if that has any side effects.

PR #2308, opened 2024-06-21
[BUILD] Split up bazel io target

fhanau:
This makes it easier to avoid superfluous dependencies, especially in tests.
Making io a smaller target will help us rebuild less code following trivial
changes – any change to a header included in the io target requires all io and
dependent source files to be rebuilt.
Crucially, this also isolates several tests from io/JSG (and by extension V8). Since
we currently make V8 alwayslink, this makes these test binaries much
smaller (~140MB => ~1MB on macOS), improving CI performance as less data
needs to be fetched from cache.

### DIFF
--- a/src/workerd/io/BUILD.bazel
+++ b/src/workerd/io/BUILD.bazel
@@ -20,36 +20,33 @@ wd_cc_library(
     # HACK: Currently, the `io` and `api` packages are interdependent. We fold all the sources
     #   from `api` into `io`. In principle, it should be possible to pull them apart so `api`
     #   depends on `io` but not vice versa.
-    # TODO(cleaunp): Fix this.
+    # TODO(cleanup): Fix this.
     srcs = glob(
         ["*.c++"],
         exclude = [
             "worker-interface.c++",
             "*-test.c++",
             "trace.c++",
+            "io-gate.c++",
+            "observer.c++",
+            "actor-cache.c++",
+            "actor-sqlite.c++",
+            "actor-storage.c++",
+            "worker-entrypoint.c++",
         ],
     ) + ["//src/workerd/api:srcs"],
     hdrs = glob(
         ["*.h"],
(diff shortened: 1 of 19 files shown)


src/workerd/io/BUILD.bazel
```diff
@@ -20,36 +20,34 @@ wd_cc_library(
...
-        #
-        # ```
-        # ERROR: C:/.../src/workerd/io/BUILD.bazel: Compiling src/workerd/io/compatibility-date.c++ failed: undeclared inclusion(s) in rule '//src/workerd/io:io':
-        # this rule is missing dependency declarations for the following files included by 'src/workerd/io/compatibility-date.c++':
-        #   'bazel-out/x64_windows-opt/bin/src/workerd/io/_virtual_includes/io/workerd/io/trace.h'
-        # In file included from <built-in>:429:
-        # ```
-        #
-        # This feels like a weird interaction between symlinks on Windows and Bazel.
-        # `trace.h` is `#pragma once`d though, so including it in both `io` and `trace` targets
```
dom96, line 45: Isn't this context useful to keep?

src/workerd/io/io-gate-test.c++
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
+//#include <workerd/io/io-gate.h>
```
dom96, line 5: Should this be removed?

dom96 (approved):

src/workerd/io/BUILD.bazel
fhanau: This context would be helpful, but the underlying issue appears to be fixed – I was able to add trace.h to the exclusions list here and the Windows build is still working. My guess is that this got fixed in Bazel at some point.

PR #2314, opened 2024-06-22
Cleanup missing streams exports

jasnell:


src/node/internal/streams_writable.js
```diff
@@ -841,3 +841,11 @@ Writable.fromWeb = function (writableStream, options) {
 Writable.toWeb = function (streamWritable) {
   return newWritableStreamFromStreamWritable(streamWritable)
 }
+
+export function fromWeb(writableStream, options) {
+  return newStreamWritableFromWritableStream(writableStream, options)
+}
```
dario-piotrowicz, line 847: For more dry and easy to maintain code, would't it be better to define the function only once instead of duplicating it here and in the field above?

What I mean is, instead of
```js
Writable.fromWeb = function (writableStream, options) {
  return newStreamWritableFromWritableStream(writableStream, options)
}

export function fromWeb(writableStream, options) {
  return newStreamWritableFromWritableStream(writableStream, options)
}
```
would't it be better something like:
```js
export function fromWeb(writableStream, options) {
  return newStreamWritableFromWritableStream(writableStream, options)
}

Writable.fromWeb = fromWeb;
```
?


> [!NOTE]
> This comment applies to the other functions in this PR

dario-piotrowicz (approved):

PR #2317, opened 2024-06-23
Couple of minor cleanups/simplifications in the Node.js buffer impl

jasnell:


src/workerd/api/node/buffer.c++
```diff
@@ -319,22 +319,32 @@ kj::Array<kj::byte> BufferUtil::concat(
...
+  auto view = dest.asPtr();
+
+  uint32_t seen = 0;
   for (auto& src : list) {
     if (src.size() == 0) continue;
-    auto amountToCopy = kj::min(src.size(), remaining);
-    std::copy(src.begin(), src.begin() + amountToCopy, ptr + offset);
-    offset += amountToCopy;
-    remaining -= amountToCopy;
+    KJ_ASSERT(seen + src.size() <= length);
```
fhanau, line 336: This assert is unsafe: If there is more input provided over all chunks than the length parameter (i.e. if truncation is supposed to happen), the assert will be triggered.

jasnell: oh, this was a left over from a quick test that I forgot to pull out

fhanau (approved):

PR #2321, opened 2024-06-24
COMPILED_MAGIC_SUFFIX check from memcmp equals 0 -> arrayPtr equality

AdityaAtulTewari:
This change follows the pattern of other safety changes to use `ArrayPtr` equality rather than `memcmp == 0`, which has bounds checking.

This PR also adds a test that compiled binaries run correctly using `sh_test`.

### DIFF
--- a/samples/BUILD.bazel
+++ b/samples/BUILD.bazel
@@ -1,0 +1,4 @@
+exports_files([
+    "helloworld/config.capnp",
+    "helloworld/worker.js",
+])
--- a/src/workerd/server/BUILD.bazel
+++ b/src/workerd/server/BUILD.bazel
@@ -110,8 +110,8 @@ wd_cc_library(
     }),
     visibility = ["//visibility:public"],
     deps = [
-        ":alarm-scheduler",
         ":actor-id-impl",
+        ":alarm-scheduler",
         ":workerd_capnp",
         "//src/workerd/api:html-rewriter",
         "//src/workerd/api:pyodide",
@@ -136,6 +136,24 @@ wd_cc_capnp_library(
     visibility = ["//visibility:public"],
     deps = [
         ":workerd_capnp",
+    ],
+)
+
+sh_test(
+    name = "helloworld_compile_test",
+    size = "small",
+    srcs = ["tests/compile-tests/compile-test.sh"],
+    args = [
+        "$(location :workerd)",
+        "$(location //samples:helloworld/config.capnp)",
+        "8080",
+        "$(location tests/compile-tests/compile-helloworld-test.ok)",
+    ],
+    data = [
+        "tests/compile-tests/compile-helloworld-test.ok",
+        ":workerd",
+        "//samples:helloworld/config.capnp",
+        "//samples:helloworld/worker.js",
     ],
 )
 
--- a/src/workerd/server/tests/compile-tests/compile-test.sh
+++ b/src/workerd/server/tests/compile-tests/compile-test.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# $1 -> workerd binary path
+# $2 -> path to desired file to compile
+# $3 -> port to curl
+# $4 -> desired output
+
+# Help Function
+function show_help() {
+  echo "
+The Compile Test script is designed to aid in testing compiled workerd binaries.
+It works by verifying that the result of a curl command on a running compiled binary matches an expected output.
+usage: compile-test.sh [-d] [-h] <workerd command> <file-to-compile> <port-to-curl> <expected-output-file>
+  options:
+    -d print out where tmp files are created and do not delete them
+    -h this help message
+
+  Note: all flags must occur before arguments
+"
+}
+
+while getopts "h" option; do
+  case ${option} in
+    h)
+      show_help
+      exit
+      ;;
+  esac
+done
+
+shift $(expr $OPTIND - 1)
+WORKERD_BINARY=$1
+CAPNP_SOURCE=$2
+PORT=$3
+EXPECTED=$4
+
+CAPNP_BINARY=$(mktemp)
+
+# Compile the app
+$WORKERD_BINARY compile $CAPNP_SOURCE > $CAPNP_BINARY
+
+# Run the app
+$CAPNP_BINARY &
+KILL=$!
+OUTPUT=$(mktemp)
+FIXED_OUTPUT=$(mktemp)
+FIXED_EXPECTED=$(mktemp)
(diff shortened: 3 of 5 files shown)


dom96 (approved):
What's the rationale for this change? Is it better memory safety?

Also how did you test this? Do we have any tests for this code path in CI?

jasnell (approved):

AdityaAtulTewari:
> What's the rationale for this change? Is it better memory safety?

The rationale is that ArrayPtr does bounds checking on the pointers that are passed into it.

> Also how did you test this? Do we have any tests for this code path in CI?

This is tested by anything that calls into the cli main, I basically tried combinations that I thought would be wrong and most cases caused about 70 tests to fail. Happy to send any info you would like to see before merging this. I'm also happy to find out how to write an explicit test to make sure this behavior is correct.

kentonv:
I suspect there is not an explicit test for this functionality. The thing you need to test is using the `workerd compile` command to create a self-contained binary, then run that binary.

(Yes, it's possible to make other tests fail by making the comparison succeed spuriously, since that will engage the self-contained binary mode when it's not intended.)

AdityaAtulTewari:
Does this mean that we want to create a new test for this?

kentonv:
It would be nice to have a test, yes. You could test this using a shell test pretty easily, I think.

samples/BUILD.bazel
```diff
@@ -0,0 +1,2 @@
+exports_files(["helloworld/config.capnp"])
```
mikea, line 1: nit: `(["f1", "f2"])`

src/workerd/server/BUILD.bazel
```diff
@@ -139,6 +139,14 @@ wd_cc_capnp_library(
     ],
 )
 
+sh_test(
```
mikea, line 142: file needs to be formatted with buildifier

src/workerd/server/tests/compile-tests/compile-test.sh
```diff
@@ -0,0 +1,72 @@
...
+set -o nounset
+set -o pipefail
+
+# $1 -> workerd binary path
+# $2 -> path to desired file to compile
+# $3 -> port to curl
+# $4 -> desired output
+
+# Help Function
+function show_help() {
```
mikea, line 13: I would convert show_help to an elaborate file-based comment where you document besides the command line the purpose of the test.

src/workerd/server/tests/compile-tests/compile-test.sh
```diff
@@ -0,0 +1,72 @@
...
+usage: compile-test.sh [-d] [-h] <workerd command> <file-to-compile> <port-to-curl> <expected-output-file>
+  options:
+    -d print out where tmp files are created and do not delete them
+    -h this help message
+
+  Note: all flags must occur before arguments
+"
+}
+
+DEBUG=false
```
mikea, line 24: let's skip debug. It is always trivial to add set -x if you are debugging the script.

src/workerd/server/tests/compile-tests/compile-test.sh
```diff
@@ -0,0 +1,72 @@
...
+  set -x;
+fi
+
+shift $(expr $OPTIND - 1)
+WORKERD_BINARY=$1
+CAPNP_SOURCE=$2
+PORT=$3
+OK_FILE=$4
+
+COMPILE=$(mktemp)
```
mikea, line 48: bad name

src/workerd/server/tests/compile-tests/compile-test.sh
```diff
@@ -0,0 +1,72 @@
...
+
+if $DEBUG; then
+  set -x;
+fi
+
+shift $(expr $OPTIND - 1)
+WORKERD_BINARY=$1
+CAPNP_SOURCE=$2
+PORT=$3
+OK_FILE=$4
```
mikea, line 46: EXPECTED_OUTPUT

src/workerd/server/tests/compile-tests/compile-test.sh
```diff
@@ -0,0 +1,72 @@
...
+
+COMPILE=$(mktemp)
+$WORKERD_BINARY compile $CAPNP_SOURCE > $COMPILE
+$COMPILE &
+KILL=$!
+OUT=$(mktemp)
+FIXED_OUT=$(mktemp)
```
mikea, line 53: this block would benefit from some comments like

```
# compile the app

# run the app

# verify test request

# etc
```

src/workerd/server/tests/compile-tests/compile-test.sh
```diff
@@ -0,0 +1,72 @@
...
+FIXED_OK=$(mktemp)
+sleep 1
+curl localhost:$PORT -o $OUT
+sed -e's/[[:space:]]*$//' $OUT > $FIXED_OUT
```
mikea, line 57: what's the purpose of this sed?

src/workerd/server/tests/compile-tests/compile-test.sh
```diff
@@ -0,0 +1,72 @@
...
+OK_FILE=$4
+
+COMPILE=$(mktemp)
+$WORKERD_BINARY compile $CAPNP_SOURCE > $COMPILE
+$COMPILE &
+KILL=$!
+OUT=$(mktemp)
+FIXED_OUT=$(mktemp)
+FIXED_OK=$(mktemp)
+sleep 1
```
mikea, line 55: sleep 1 means test will spin for 1 second without doing anything. could we do better here?

src/workerd/server/tests/compile-tests/compile-test.sh
AdityaAtulTewari: On windows the output can include an extra `\r` character, the idea is to use `-Z` however it is not possible for that to work on macOS.

src/workerd/server/tests/compile-tests/compile-test.sh
AdityaAtulTewari: I added more to it, however I think the ability to use shell scripts like shell scripts is helpful for development.

src/workerd/server/tests/compile-tests/compile-test.sh
AdityaAtulTewari: I think I found a way to do this, please check it.

src/workerd/server/tests/compile-tests/compile-test.sh
AdityaAtulTewari: I think I found a way to do this, please check it.

mikea (approved):

PR #2324, opened 2024-06-24
Run test imports at top level, fix pydantic and numpy imports

garrettgu10:
Importing stuff within a `test()` function does not invoke the top-level entropy patches we implemented, so this change should make our testing much more complete. 

Should reproduce https://github.com/cloudflare/python-workers-examples/issues/9.


dom96 (approved):
At the risk of doubling our CI time... I wonder if we should test both scenarios.

PR #2325, opened 2024-06-24
BIGNUM utility cleanups

jasnell:
Eliminates some duplication and condenses some of the handling of BIGNUM types... just one more of a number of PRs that'll be reworking and cleaning up the crypto impl a bit before coalescing with the node.js crypto impl details.

### DIFF
--- a/src/workerd/api/crypto/asymmetric.c++
+++ b/src/workerd/api/crypto/asymmetric.c++
@@ -13,7 +13,7 @@
 #include <map>
 #include <kj/function.h>
 #include <type_traits>
-#include "util.h"
+#include <workerd/api/util.h>
 #include <workerd/io/features.h>
 
 namespace workerd::api {
@@ -548,13 +548,6 @@ protected:
   CryptoKey::RsaKeyAlgorithm keyAlgorithm;
 
 private:
-  static kj::Array<kj::byte> bigNumToArray(const BIGNUM& n) {
-    kj::Vector<kj::byte> result(BN_num_bytes(&n));
-    result.resize(result.capacity());
-    BN_bn2bin(&n, result.begin());
-    return result.releaseAsArray();
-  }
-
   SubtleCrypto::JsonWebKey exportJwk() const override final {
     const auto& rsa = JSG_REQUIRE_NONNULL(EVP_PKEY_get0_RSA(getEvpPkey()), DOMOperationError,
         "No RSA data backing key", tryDescribeOpensslErrors());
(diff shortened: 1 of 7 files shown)


src/workerd/api/crypto/impl.h
```diff
@@ -293,6 +294,14 @@ const SslDisposer<T, sslFree> SslDisposer<T, sslFree>::INSTANCE;
 // Using BN_clear_free here ensures that any potentially sensitive information in the
 // BIGNUM is also cleansed when it is freed.
 
+using UniqueBignum = std::unique_ptr<BIGNUM, void(*)(BIGNUM*)>;
+kj::Maybe<kj::Own<BIGNUM>> toBignum(kj::ArrayPtr<const kj::byte> data);
```
mikea, line 298: stylistical suggestion: ArrayPtr/Array have `.as()` method that is designed to help with things like `ptr.as<Bignum>()`. I found it very helpful for Rust conversions and makes code much easier to read.

src/workerd/api/crypto/impl.h
```diff
@@ -10,6 +10,7 @@
 #include "crypto.h"
 #include <workerd/api/util.h>
 #include <kj/encoding.h>
+#include <openssl/bn.h>
```
mikea, line 13: looks like a simple forward declaration of BIGNUM would do here?

src/workerd/api/crypto/impl.c++
```diff
@@ -3,9 +3,10 @@
 //     https://opensource.org/licenses/Apache-2.0
 
 #include "impl.h"
-#include "util.h"
+#include "../util.h"
```
mikea, line 6: I am surprised this works and is not banned in our source code :) Let's spell it out like we usually do?

src/workerd/api/crypto/impl.c++
```diff
@@ -220,6 +221,42 @@ void checkPbkdfLimits(jsg::Lock& js, size_t iterations) {
...
+
+kj::Maybe<kj::Array<kj::byte>> bignumToArray(const BIGNUM& n) {
+  auto result = kj::heapArray<kj::byte>(BN_num_bytes(&n));
+  if (BN_bn2bin(&n, result.begin()) != result.size()) return kj::none;
+  return kj::mv(result);
+}
+
+kj::Maybe<kj::Array<kj::byte>> bignumToArrayPadded(const BIGNUM& n) {
+  auto result = kj::heapArray<kj::byte>(BN_num_bytes(&n));
+  if (BN_bn2binpad(&n, static_cast<unsigned char*>(result.begin()),
```
mikea, line 242: I wonder whwhy cast is needed here? kj::byte is already unsigned char, isn't it?

src/workerd/api/crypto/asymmetric.c++
```diff
@@ -1065,8 +1055,8 @@ kj::Own<EVP_PKEY> rsaJwkReader(SubtleCrypto::JsonWebKey&& keyDataJwk) {
   // RSA_set0_*() transfers BIGNUM ownership to the RSA key, so we don't need to worry about
   // calling BN_free().
   OSSLCALL(RSA_set0_key(rsaKey.get(),
-      BN_bin2bn(modulus.begin(), modulus.size(), nullptr),
-      BN_bin2bn(publicExponent.begin(), publicExponent.size(), nullptr),
+      toBignumUnowned(modulus),
```
mikea, line 1058: I wonder if we should add `T* kj::Own<T>::disown()` rather than create separate functions for such use-cases?

src/workerd/api/crypto/asymmetric.c++
```diff
@@ -563,16 +556,16 @@ private:
...
+    jwk.e = kj::encodeBase64Url(KJ_REQUIRE_NONNULL(bignumToArray(KJ_REQUIRE_NONNULL(rsa.e))));
 
     if (getType() == "private"_kj) {
-      jwk.d = kj::encodeBase64Url(bigNumToArray(KJ_REQUIRE_NONNULL(rsa.d)));
-      jwk.p = kj::encodeBase64Url(bigNumToArray(KJ_REQUIRE_NONNULL(rsa.p)));
-      jwk.q = kj::encodeBase64Url(bigNumToArray(KJ_REQUIRE_NONNULL(rsa.q)));
-      jwk.dp = kj::encodeBase64Url(bigNumToArray(KJ_REQUIRE_NONNULL(rsa.dmp1)));
-      jwk.dq = kj::encodeBase64Url(bigNumToArray(KJ_REQUIRE_NONNULL(rsa.dmq1)));
-      jwk.qi = kj::encodeBase64Url(bigNumToArray(KJ_REQUIRE_NONNULL(rsa.iqmp)));
+      jwk.d = kj::encodeBase64Url(KJ_REQUIRE_NONNULL(bignumToArray(KJ_REQUIRE_NONNULL(rsa.d))));
```
mikea, line 563: it didn't throw on invalid parse before, right? Is this important that it now does?

jasnell: I don't think it is going to matter one way or the other. It's exceedingly unlikely this would every throw in this case.

src/workerd/api/crypto/asymmetric.c++
jasnell: Yeah, I was considering the same thing. The ability to release ownership is the only reason why I didn't eliminate the `std::unique_ptr` use here entirely.

src/workerd/api/crypto/impl.h
jasnell: Perhaps if we introduce a `Bignum` helper that wraps `BIGNUM` at some point. Given that this is mostly just shifting around the way the existing code already worked, I'd prefer to keep it as is for now but I'm going to be continuing to iterate on this area.

PR #2327, opened 2024-06-25
Additional crypto cleanups, add scrypt impl

jasnell:
More reorganization of crypto impl... start consolidating key parts of the Node.js crypto and web crypto impls starting with HKDF and PBKDF2. Add scrypt impl


src/workerd/api/crypto/scrypt.c++
```diff
@@ -0,0 +1,31 @@
...
+namespace workerd::api {
+
+kj::Maybe<kj::Array<kj::byte>> scrypt(size_t length,
+                                      uint32_t N,
+                                      uint32_t r,
+                                      uint32_t p,
+                                      uint32_t maxmem,
+                                      kj::ArrayPtr<const kj::byte> pass,
+                                      kj::ArrayPtr<const kj::byte> salt) {
+  auto buf = kj::heapArray<kj::byte>(length);
```
fhanau, line 18: Node also checks the size of pass and salt here (probably not needed) and applies some error handling by examining the last error if present after calling `EVP_PBE_scrypt()` (which we should adopt too).

src/node/internal/crypto_scrypt.ts
```diff
@@ -0,0 +1,194 @@
...
+  N: 16384,
+  r: 8,
+  p: 1,
+  maxmem: 32 << 20,
+};
+
+function validateParameters(password: ArrayLike, salt: ArrayLike,
+                            keylen: number, options: ScryptOptions) : ValidatedScryptOptions {
+  // TODO(soon): Add support for KeyObject input.
+  password = validateByteSource(password, 'password');
```
fhanau, line 87: Node uses getArrayBufferOrView for password and salt instead (not sure if it makes a difference).

fhanau (commented):
LGTM otherwise, let's just make sure we have some error handling here.

src/workerd/api/crypto/scrypt.c++
jasnell: Length check will happen on the ts side.

fhanau: LGTM, I must have missed the length checks before. Error handling works too.

fhanau (approved):

PR #2329, opened 2024-06-25
Fix Pydantic import entropy

garrettgu10:


hoodmane:
Do we have test coverage for this? We should make sure there's a test that fails without the change and passes with it. Perhaps this is the point of #2324?

garrettgu10:
@hoodmane  yeah, the top level import tests including pydantic are examples of tests that fail without this pr

PR #2330, opened 2024-06-25
Implement Node.js spkac APIs

jasnell:


src/node/internal/crypto.d.ts
```diff
@@ -44,6 +44,11 @@ export function createSecretKey(key: ArrayBuffer | ArrayBufferView): CryptoKey;
 export function createPrivateKey(key: InnerCreateAsymmetricKeyOptions): CryptoKey;
 export function createPublicKey(key: InnerCreateAsymmetricKeyOptions): CryptoKey;
 
+// Spkac
+export function verifySpkac(input: ArrayBufferView|ArrayBuffer): boolean;
```
fhanau, line 48: The Node.js docs indicate that exportPublicKey and exportChallenge return a Buffer – is that equivalent to `null | ArrayBuffer` here? All three functions also need to support string and Buffer inputs.

src/node/internal/crypto_spkac.ts
```diff
@@ -0,0 +1,103 @@
...
+  isAnyArrayBuffer,
+} from 'node-internal:internal_types';
+
+import {
+  ERR_INVALID_ARG_TYPE,
+} from 'node-internal:internal_errors'
+
+type ArrayLike = cryptoImpl.ArrayLike;
+
+function getArrayBufferOrView(buffer: unknown, name: string, encoding?: string) {
```
fhanau, line 46: `getArrayBufferOrView` is already available in `crypto_util.ts`

src/node/internal/crypto_spkac.ts
```diff
@@ -0,0 +1,103 @@
...
+// rely on any object state.
+
+// For backwards compatibility reasons, this cannot be converted into a
+// ES6 Class.
+export function Certificate(this: any) {
+  if (!(this instanceof Certificate))
+    return new (Certificate as any)();
+}
+
+Certificate.prototype.verifySpkac = verifySpkac;
```
fhanau, line 97: Since the non-static methods of Certificate are deprecated, do we need to define them?

src/workerd/api/crypto/spkac.c++
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+  return kj::none;
+}
+
+kj::Maybe<kj::Array<kj::byte>> exportChallenge(kj::ArrayPtr<const kj::byte> input) {
+  KJ_IF_SOME(spki, tryGetSpki(input)) {
+    kj::byte* buf = nullptr;
+    int buf_size = ASN1_STRING_to_UTF8(&buf, spki->spkac->challenge);
+    if (buf_size < 0 || buf == nullptr) return kj::Array<kj::byte>();
```
fhanau, line 87: nit: Could we also just return kj::none here? It should lead to the same result on the JS level.

src/workerd/api/crypto/spkac.c++
```diff
@@ -0,0 +1,95 @@
...
+
+kj::Maybe<kj::Own<BIO>> tryNewBio() {
+  auto bioptr = BIO_new(BIO_s_mem());
+  if (!bioptr) return kj::none;
+  return kj::disposeWith<BIO_free_all>(bioptr);
+}
+}  // namespace
+
+bool verifySpkac(kj::ArrayPtr<const kj::byte> input) {
+  // So, this is fun. SPKAC uses MD5 as the digest algorithm. This is a problem because
```
fhanau, line 51: If we can't verify spkac, should we even ship the method? It would be confusing to developers if they can call a method but it only produces errors. We also can't test it reliably.

fhanau:
- [ ] Check the Certificate checkbox in src/node/crypto.ts, at least with a [.] if we're not shipping verifySpkac().

src/workerd/api/crypto/spkac.c++
jasnell: Well, we do have the option of manually implementing it rather than relying on boringssl's implementation. It's a fair question tho and the answer is: I'm not sure.

src/node/internal/crypto_spkac.ts
jasnell: I've still seen the deprecated/legacy methods used in the wild. Likely a good idea to retain them here.

src/node/internal/crypto.d.ts
jasnell: These are the types for the internal call. The `Buffer` and `string` handling should be taken care of in the ts file that defines the public API.

src/node/internal/crypto_spkac.ts
fhanau: Ok, in this case it makes sense since the deprecated methods look sane and there's little overhead in having them.

src/workerd/api/crypto/spkac.c++
jasnell: For now I've added a `logWarningOnce(...)` that will explain the situation if the method is called from within an active IoContext. I think this is good enough and we can revisit this later if folks give any pushback.

fhanau (approved):
LGTM now.

PR #2332, opened 2024-06-26
Add new impliedByAfterDate annotation to compatibility-date

jasnell:
The new annotation allows us to specify that a compatibility flag is implied if the named other flag is enabled, but only at the specified compatibility date or later. If the flag is marked experimental and experimental flags are not allowed, or if the flag is explicitly turned off, the annotation will not apply.

This will be used, for instance, to specify that `nodejs_compat_v2` is implied on by `nodejs_compat` after a given date.


irvinebroque:
To confirm, what this lets us then provide is:

- If you are using `nodejs_compat` today, and have a compat date of 06-25-24 — nothing changes
- If you are then change your compat date to `<future date>` — we will (behind the scenes) use `nodejs_compat_v2` instead

...allowing us to maintain existing behavior, and provide an explicit opt-in (via a compatibility date)

src/workerd/io/compatibility-date.c++
```diff
@@ -160,6 +169,28 @@ void compileCompatibilityFlags(kj::StringPtr compatDate, capnp::List<capnp::Text
...
+            });
+          }
+        }
+      }
+    }
+    KJ_IF_SOME(impliedBy, maybeImpliedBy) {
+      // We only want to add the implied by flag if it is not explicitly disabled.
+      // We also only want to add it if the flag is not experimental or experimental
+      // flags are allowed...
+      if (!disableByFlag && (!isExperimental || allowExperimentalFeatures)) {
```
kentonv, line 192: I think we should remove the `&& (!isExperimental || allowExperimentalFeatures)` part. Otherwise, interpretation of flags depends on whether or not the user is permitted to use experimental flags, which means Cloudflare employees will get differing behavior without realizing it.

If an experimental flag is turned on because it was implied, but experimental flags aren't allowed, I think it's fine that we just end up erroring.

In practice this probably won't ever matter since I don't think we'll ever define an experimental flag to be implied by a non-experimental flag anyway.

kentonv (commented):
Code looks good but can we add a test to `compatibility-date-test.c++`? Can base the test around the nodejs_compat_v2 flag specifically.

jasnell:
Test added. experimental check removed.

src/workerd/io/compatibility-date-test.c++
```diff
@@ -180,6 +180,118 @@ KJ_TEST("compatibility flag parsing") {
...
+      " brotliContentEncoding = true,"
+      " strictCrypto = true,"
+      " rttiApi = false,"
+      " webgpu = false,"
+      " cryptoPreservePublicExponent = true,"
+      " vectorizeQueryMetadataOptional = true,"
+      " unsafeModule = false,"
+      " jsRpc = false,"
+      " noImportScripts = true,"
+      " nodeJsAls = false,"
```
IgorMinar, line 227: shouldn't `nodeJsAls` be an implied flag that is activated automatically when `nodeJsCompat` is on?

src/workerd/io/compatibility-date.capnp
```diff
@@ -444,7 +453,8 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
 
   nodeJsCompatV2 @50 :Bool
       $compatEnableFlag("nodejs_compat_v2")
-      $compatDisableFlag("no_nodejs_compat_v2");
+      $compatDisableFlag("no_nodejs_compat_v2")
+      $impliedByAfterDate(name = "nodeJsCompat", date = "2024-08-05");
```
IgorMinar, line 457: is this the date we want to draw in the sand for now? 

@irvinebroque how do you feel about this date? Could we push it out to September 1? I'm concerned that early august for on by default is a bit too aggressive. We should announce this and give it at least a month of opt-in usage before it's on by default.

We can announce in mid/late july, and make it the default in early September (way before birthday week to avoid collisions with other launches)

IgorMinar (commented):
I can't review the c++ code, but the tests look exactly like what I would expect. I just have some minor product comments.

src/workerd/io/compatibility-date.capnp
irvinebroque: See internal thread

src/workerd/io/compatibility-date-test.c++
jasnell: No, it's a separate flag intended to make `AsyncLocalStorage` available regardless of the rest of `nodejs_compat`. The `nodeJsCompat`/v2 flags already enable it. There's no need for it to be implied.

src/workerd/io/compatibility-date.capnp
jasnell, line 457:
```suggestion
      $impliedByAfterDate(name = "nodeJsCompat", date = "2024-09-02");
```

src/workerd/io/compatibility-date-test.c++
```diff
@@ -180,6 +180,118 @@ KJ_TEST("compatibility flag parsing") {
   // Can explicitly disable flag that's enabled for all dates.s
   expectCompileCompatibilityFlags("2021-05-17", {"r2_internal_beta_bindings"}, "()", {},
       CompatibilityDateValidation::FUTURE_FOR_TEST, true, false);
+
+  // nodejs_compat implies nodejs_compat_v2 on or after 2024-08-05
+  expectCompileCompatibilityFlags("2024-08-05", {"nodejs_compat"},
```
jasnell, line 185:
```suggestion
  expectCompileCompatibilityFlags("2024-09-02", {"nodejs_compat"},
```

src/workerd/io/compatibility-date-test.c++
```diff
@@ -180,6 +180,118 @@ KJ_TEST("compatibility flag parsing") {
...
+      " unwrapCustomThenables = true,"
+      " fetcherRpc = true,"
+      " internalStreamByobReturn = true,"
+      " blobStandardMimeType = true,"
+      " fetchStandardUrl = true,"
+      " nodeJsCompatV2 = true,"
+      " globalFetchStrictlyPublic = false,"
+      " newModuleRegistry = false)", {},
+      CompatibilityDateValidation::FUTURE_FOR_TEST, false, false);
+  expectCompileCompatibilityFlags("2024-08-04", {"nodejs_compat"},
```
jasnell, line 240:
```suggestion
  expectCompileCompatibilityFlags("2024-09-01", {"nodejs_compat"},
```

src/workerd/io/compatibility-date.capnp
jasnell: Date agreed on internally is 2024-09-02 ... PR is updated to reflect that.

kentonv (approved):

PR #2334, opened 2024-06-26
Complete typescript migration for Pyodide glue code

garrettgu10:
Last batch of types being added to the Pyodide glue code. There's a bit of jankiness around the filesystem (specifically the `info` object which can technically be any type). There was also a spot where the node type for directories was `5` vs `"5"`, but that didn't end up getting used in a critical path. I made them all `"5"` for consistency.

### DIFF
--- a/src/pyodide/internal/tar.ts
+++ b/src/pyodide/internal/tar.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import { default as TarReader } from "pyodide-internal:packages_tar_reader";
 
 // This is based on the info about the tar file format on wikipedia
@@ -6,21 +5,21 @@ import { default as TarReader } from "pyodide-internal:packages_tar_reader";
 // https://en.wikipedia.org/wiki/Tar_(computing)#File_format
 
 const decoder = new TextDecoder();
-function decodeString(buf) {
+function decodeString(buf: Uint8Array): string {
   const nullIdx = buf.indexOf(0);
   if (nullIdx >= 0) {
     buf = buf.subarray(0, nullIdx);
   }
   return decoder.decode(buf);
 }
-function decodeField(buf, offset, size) {
+function decodeField(buf: Uint8Array, offset: number, size: number): string {
   return decodeString(buf.subarray(offset, offset + size));
 }
-function decodeNumber(buf, offset, size) {
+function decodeNumber(buf: Uint8Array, offset: number, size: number): number {
   return parseInt(decodeField(buf, offset, size), 8);
 }
 
-function decodeHeader(buf, reader) {
+function decodeHeader(buf: Uint8Array, reader: Reader): FSInfo {
   const nameBase = decodeField(buf, 0, 100);
   const namePrefix = decodeField(buf, 345, 155);
   let path = namePrefix + nameBase;
@@ -46,12 +45,12 @@ function decodeHeader(buf, reader) {
 }
 
 export function parseTarInfo(reader = TarReader): [FSInfo, string[]] {
-  const directories = [];
+  const directories: FSInfo[] = [];
   const soFiles = [];
-  const root = {
+  const root: FSInfo = {
     children: new Map(),
     mode: 0o777,
-    type: 5,
+    type: "5",
     modtime: 0,
     size: 0,
     path: "",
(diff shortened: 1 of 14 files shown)


src/pyodide/internal/tar.ts
```diff
@@ -108,22 +107,22 @@ export function parseTarInfo(reader = TarReader): [FSInfo, string[]] {
...
-      directory = directories.pop();
+      directory = directories.pop()!;
     }
     // go down to target (in many tar files this second loop body is evaluated 0
     // times)
     const parts = info.path.slice(0, -1).split("/");
     for (let i = directories.length; i < parts.length - 1; i++) {
       directories.push(directory);
-      directory = directory.children.get(parts[i]);
+      directory = directory.children!.get(parts[i])!;
```
dom96, line 117: Why can `children` be `undefined`? Should the type be changed?

src/pyodide/internal/tarfs.ts
```diff
@@ -18,19 +17,22 @@ const FSOps = {
...
   },
   lookup(parent, name) {
-    return parent.info.children.get(name);
+    return parent.info.children!.get(name)!;
   },
   read(stream, position, buffer) {
-    return stream.node.info.reader.read(
+    if (stream.node.contentsOffset == undefined) {
+      throw new Error("contentsOffset is undefined");
+    }
```
dom96, line 28: Can we change the types so that these fields are guaranteed to not be undefined?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -24,16 +23,16 @@ import { default as Limiter } from "pyodide-internal:limiter";
...
   if (!MAIN_MODULE_NAME.endsWith(".py")) {
     throw new Error("Main module needs to end with a .py file extension");
   }
   const mainModuleName = MAIN_MODULE_NAME.slice(0, -3);
   return pyodide.pyimport(mainModuleName);
 }
 
-let pyodidePromise;
-function getPyodide() {
+let pyodidePromise: Promise<Pyodide> | undefined;
```
dom96, line 34: `null` may be better here:

```suggestion
let pyodidePromise: Promise<Pyodide> | null;
```

src/pyodide/types/FS.d.ts
```diff
@@ -1,18 +1,96 @@
...
+  isFile: (mode: number) => boolean;
+  genericErrors: Error[];
+}
+
+interface FSOps {
+  getNodeMode: (parent: FSNode | null, name: string, info: any) => {
+    permissions: number;
+    isDir: boolean;
+  };
+  setNodeAttributes: (node: FSNode, info: any, isDir: boolean) => void;
```
dom96, line 36: Not sure if this is right, but would be good to change 

```suggestion
  getNodeMode: (parent: FSNode | null, name: string, info: MetadataFSInfo) => {
    permissions: number;
    isDir: boolean;
  };
  setNodeAttributes: (node: FSNode, info: MetadataFSInfo, isDir: boolean) => void;
```

src/pyodide/types/Pyodide.d.ts
```diff
@@ -0,0 +1,16 @@
...
+
+interface PyModule {
+  [handlerName: string]: {
+    callRelaxed: (...args: any[]) => any
+  };
+}
+
+interface Pyodide {
+  _module: Module;
+  registerJsModule: (handle: string, mod: object) => void;
```
dom96, line 11: Maybe add a TODO so we remember to make `object` stricter

dom96 (approved):
Nice!

src/pyodide/internal/tarfs.ts
garrettgu10: contentsOffset is undefined when the node is a directory 😶

src/pyodide/types/FS.d.ts
garrettgu10: Technically this object can be any type, Emscripten just passes them around. We use Map, FSInfo, and number

src/pyodide/types/Pyodide.d.ts
garrettgu10: I think a js module can be any object though?

src/pyodide/python-entrypoint-helper.ts
hoodmane: I think `undefined` is more accurate. If you say `let x;` it is the same as `let x = undefined;`.

src/pyodide/types/Pyodide.d.ts
hoodmane: Yeah I'm not sure what you could put here that would be more helpful.

PR #2339, opened 2024-06-27
Simple REPL

garrettgu10:
Simple REPL servers for both JavaScript and Python. I've found these helpful when debugging the runtime and it would be nice to have these merged in and available.


samples/repl-server/client.js
```diff
@@ -0,0 +1,19 @@
+const repl = require("repl");
+
+const myeval = async (cmd, ctx, fname, callback) => {
+  const res = await fetch('http://localhost:8080', {
```
jasnell, line 4: Might be more efficient to set this up as a single long-running fetch where each command is streamed in rather than a new fetch for every command. But for now, this works.

jasnell (approved):

samples/repl-server/client.js
garrettgu10: Hmm, I wonder if that could allow us to stream stdout instead of having to batch it per-command. Will experiment with this.

PR #2340, opened 2024-06-28
More node:crypto cleanup and consolidation

jasnell:


src/workerd/api/crypto/impl.h
```diff
@@ -362,6 +362,13 @@ class ZeroOnFree {
 // be thrown. Otherwise the method will return normally.
 void checkPbkdfLimits(jsg::Lock& js, size_t iterations);
 
+// Either succeeds with exactly |length| bytes of cryptographically
+// strong pseudo-random data, or fails. This function may block.
+// Don't assume anything about the contents of |buffer| on error.
+// As a special case, |length == 0| can be used to check if the CSPRNG
```
fhanau, line 368: Comment can be simplified as we only support the simplified check now. Also bufer => buffer

src/workerd/api/crypto/impl.h
```diff
@@ -362,6 +362,13 @@ class ZeroOnFree {
...
+// is properly seeded without consuming entropy.
+bool CSPRNG(kj::ArrayPtr<kj::byte> bufer);
```
jasnell, line 370:
```suggestion
bool CSPRNG(kj::ArrayPtr<kj::byte> buffer);
```

src/workerd/api/crypto/impl.h
jasnell: can you suggest a specific simplification

fhanau (approved):
LGTM overall. I assume the oneshot methods being added will be used on the TS side in the future?

jasnell:
> I assume the oneshot methods being added will be used on the TS side in the future?

yep.. will handle those in a separate PR

PR #2342, opened 2024-06-28
Update V8: 12.6 -> 12.7

harrishancock:
This PR updates V8 from 12.6 to 12.7. My only concern here is the TracedReference dereferencing API removal. For now, I've just reverted the removal, and things seem to work -- but we need to find a way to move off of it soon.


jasnell (approved):
looks like a possible windows issue to investigate but over all LGTM once CI comes back green

fhanau:
Based on https://chromium-review.googlesource.com/c/v8/v8/+/5600069 and https://chromium-review.googlesource.com/c/v8/v8/+/5596550 we should also update FP16 and abseil here to keep their versions in sync with V8.

harrishancock:
Thanks @fhanau, I added an Abseil and FP16 update. I also updated to V8 12.7.224.11, which is now the latest patch release.

fhanau (approved):

fhanau:
The Windows build was stalled due to a transient bazel issue, I kicked it and it passed. PR LGTM.

PR #2344, opened 2024-06-28
Consolidate dh code into workerd/api/crypto, impl DH keygen

jasnell:
Further consolidation of the node:crypto impl into workerd/api/crypto and implementation of DH key generation.


src/workerd/api/crypto/dh.c++
```diff
@@ -40,80 +36,88 @@ BIGNUM* (*FindDiffieHellmanGroup(const char* name))(BIGNUM*) {
...
-CryptoImpl::DiffieHellmanHandle::DiffieHellmanHandle(kj::String& name) : verifyError(0) {
-  JSG_REQUIRE(InitGroup(name), Error, "DiffieHellman init failed");
-}
-
-bool CryptoImpl::DiffieHellmanHandle::InitGroup(kj::String& name) {
-  auto group = FindDiffieHellmanGroup(name.begin());
+kj::Own<DH> initDhGroup(kj::StringPtr name) {
+  auto group = findDiffieHellmanGroup(name.begin());
   JSG_REQUIRE(group != nullptr, Error, "Failed to init DiffieHellmanGroup: invalid group. Only "
               "groups {modp14, modp15, modp16, modp17, modp18} are supported.");
```
garrettgu10, line 43: Do we want to consolidate this list with the list in findDiffieHellman? May not be necessary if the list is unlikely to change

src/workerd/api/crypto/dh.c++
```diff
@@ -147,67 +151,80 @@ bool CryptoImpl::DiffieHellmanHandle::Init(
...
+                                unsigned char* data,
+                                size_t prime_size) {
+  // DH_size returns number of bytes in a prime number.
+  // DH_compute_key returns number of bytes in a remainder of exponent, which
+  // may have less bytes than a prime number. Therefore add 0-padding to the
+  // allocated buffer.
+  if (remainder_size != prime_size) {
+    KJ_ASSERT(remainder_size < prime_size);
+    const size_t padding = prime_size - remainder_size;
+    memmove(data + padding, data, remainder_size);
```
garrettgu10, line 170: I don't think this is constant-time. Does it matter in this context?

src/workerd/api/crypto/dh.c++
jasnell: not critical right now

src/workerd/api/crypto/dh.c++
jasnell: I wouldn't consider it critical right now. If we didn't have spectre mitigations in place then it might be an issue but as it is I think this should be fine.

garrettgu10: Hmm. Makes me a bit queasy but I guess it's fine.

garrettgu10 (approved):

fhanau (approved):

PR #2350, opened 2024-07-03
Fixes Socket::startTls assert failure when request IoContext ends.

dom96:
The `tlsStarter` is an `IoOwn` which means that it gets deallocated when the request's IoContext ends. In the `startTls` implementation we wait for a `flush` to complete before we use the `tlsStarter`, the `flush` may complete after the lifetime of the IoContext at which point the `tlsStarter` will be null and the `KJ_ASSERT_NONNULL` will fail.

This PR throws a JSG error instead which should (hopefully) fail more gracefully.

### DIFF
--- a/src/workerd/api/sockets.c++
+++ b/src/workerd/api/sockets.c++
@@ -339,7 +339,12 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
       }
     }
 
-    // All non-secure sockets should have a tlsStarter.
+    // All non-secure sockets should have a tlsStarter. Though since tlsStarter is an IoOwn, if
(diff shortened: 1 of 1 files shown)


jasnell (approved):

jclee:
FWIW, seems OK to me.  I think the main effect of this change is that the raised exception in JS would manifest as a TypeError with a more useful message, rather than an Error with an "internal error" message.

Error message seems OK. Not sure if there's any preexisting error text that would be useful to be consistent with...  I notice we have errors for similar situations documented as "attempting to access an inactive Request context", but it looks like these would mostly be "Cannot perform I/O on behalf of a different request." errors, which is probably less useful than the current PR text.

https://developers.cloudflare.com/workers/runtime-apis/request/#errors-when-attempting-to-access-an-inactive-request-context

jclee (approved):

PR #2352, opened 2024-07-03
Implement JSG_DETACHED_METHOD and JSG_DETACHED_METHOD_NAMED

jasnell:
These new macros allow defining instance methods on JSG objects that can be detached from the object using destructuring assignment, etc. They fill an existing gap between JSG_METHOD and JSG_STATIC_METHOD.

```cpp
struct Foo : public jsg::Object {
  static void foo() {};
  void bar() {}
  static void baz() {}

  JSG_RESOURCE_TYPE(Foo) {
    JSG_DETACHED_METHOD(foo);
    JSG_METHOD(bar);
    JSG_STATIC_METHOD(baz);
  }
};
```

```js
const obj = new Foo();

// The bar method can only be called attached or bound to the Foo
obj.bar();  // Ok!

const { bar } = obj;

bar.call(obj); // Ok!
bar.apply(obj); // Ok!
bar.bind(obj)(); // Ok!
bar(); // Error: illegal invocation!

// The baz method is only defined statically on the Foo class
Foo.baz();  // Ok!
obj.baz(); // Error! baz is not a function

// Since foo is defined as a detached method, it can be called
// with or without the object as the receiver

obj.foo(); // Ok!
const { foo } = obj;
foo(); // Ok!

foo.call(obj); // Ok!
foo.call(null); // Ok!
foo.apply(obj); // Ok!
foo.apply(null); // Ok!
```

~~There is a bit of complexity if the implementation of `foo()` does in fact need to know if there is a `this` bound to it. In that case the current only solution is to define the method signature of `foo()` as accepting a `const v8::FunctionCallBackInfo<v8::Value>&` as the first argument, which allows the implementation to check and unwrap the `this` object if it is bound. We can potentially simplify that a bit. See the included jsg-test for an example of how to do this and some discussion around how to simplify.~~ added a commit that simplifies this bit

### DIFF
--- a/src/workerd/jsg/jsg-test.c++
+++ b/src/workerd/jsg/jsg-test.c++
@@ -422,6 +422,53 @@ KJ_TEST("jsg::Lock getUuid") {
   KJ_ASSERT(called);
 }
 
+// ========================================================================================
+struct DetachedMethodContext: public ContextGlobalObject {
+
+  struct FooObject: public jsg::Object {
+    static bool foo() { return true; }
+
+    // TODO(simplify): The baz() method here shows what is currently the only way to
+    // extract the `this` object that a detached or static method may have been bound
+    // to (using either bind(...) in JS or the .call(...) or .apply(...) methods).
+    // We could potentially simplify this by introducing a new `jsg::Self<T>` type of
+    // special placeholder argument that would be automatically filled in with the
+    // bound `this`, if any.
+    static bool baz(const v8::FunctionCallbackInfo<v8::Value>& info,
+                    const TypeHandler<Ref<FooObject>>& handler) {
+      auto& js = jsg::Lock::from(info.GetIsolate());
+      KJ_IF_SOME(self, handler.tryUnwrap(js, info.This())) {
+        return self->abc();
+      }
+      return false;
+    }
+
+    bool abc() { return true; }
+
+    static jsg::Ref<FooObject> constructor() { return jsg::alloc<FooObject>(); }
+    JSG_RESOURCE_TYPE(FooObject) {
+      JSG_DETACHED_METHOD(foo);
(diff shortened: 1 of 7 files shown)


mikea (approved):
This LG, but I wonder why isn't it possible to fix for all methods? If all it takes is a slightly different `this` calculation, can't the code behind `JSG_METHOD` do the same?

irvinebroque:
🙌 Excited about this one. Very unintuitive that this doesn't work today!

kentonv (changes requested):
It appears that this design requires the C++ method to be static, so it does not get bound to the object it arrived on at all. I think this isn't what people are looking for here. For example, in the specific case of `ctx`, this won't work -- the methods are not detached functions, they are actually class methods. (Some of them don't currently actually use `this` because they use `IoContext::current()` directly, but this is actually a longstanding bug.)

In order to support that properly we would instead need to bind each function to each `this`, which implies each function must have separate instances allocated for each parent object, rather than using the shared prototype. This would have negative performance implications.

But I think if that's what you really want, the right way to accomplish it is to define a property whose type is a `jsg::Function`. No need for a new macro to support that?

jasnell:
> I think this isn't what people are looking for here

this is doing exactly what it was meant to. This is not specific to the internal discussion for `waitUntil` but also applies to module namespace objects that are implemented as JSG resource types. This pr does not make any changes to the definition of `waitUntil` and should not be blocked solely on whether it works for that specific case or not.

kentonv:
If the goal is to support "namespace objects", especially modules, would it make sense to define a `JSG_NAMESPACE` macro (or even `JSG_MODULE`) that substitutes for `JSG_RESOURCE_TYPE`? I think within such a declaration you'd be expected to define strictly static methods.

PR #2353, opened 2024-07-03
Multiple api/crypto key related cleanups (part 1)

jasnell:
This is to start separating and consolidating parts of the key handling code. There are much larger changes to come. This is separated out so that changes are easiler to review. This makes no functional changes, just restructures.

The intent here is to make this code easier to maintain and expand to include the Node.js key handling mechanisms without duplicating code, etc. The idea is to start separating things out to build some better abstractions. There is a lot more to do here so a lot of the questions of Why will likely be better answered in future Prs.


fhanau (approved):
LGTM.

PR #2359, opened 2024-07-05
adding loadShed outcome

mikea:
loadShed means the server didn't start processing the request. It could happen when server is overloaded or in some other not-very-healthy state.

Not used anywhere yet, the plan is to use for load-shedding experiments. There's downstream change open updating the master copy.


jasnell (approved):

mikea:
renamed to loadShed per @kentonv feedback.

PR #2362, opened 2024-07-08
Only ask for nodejs_als in the async-context sample

isker:
I think it's best to encourage people to use this narrower flag instead of making the entirety of the emulated nodejs API available, as the documentation explains.

I don't think there's any build interactions with samples/ at all, so I manually ran this sample with my changes and confirmed it works.


isker:

jasnell (approved):

PR #2376, opened 2024-07-09
Remove websocket locality

jp4a50:
Refactor workerd::api::WebSocket such that it no longer needs callers to indicate "locality" to decide whether to register pending events.

Whether to register pending events (and whether to tie the lifetime of the websocket connection to the IoContext) can be determined purely based on existing internal state.

I've reasoned through each case and this should be a non-functional change. I'm somewhat relying on test coverage to back that up, though.

This change was prompted by me trying to figure out when the server side of a websocket terminated in a worker or not (so I could add some appropriate logic) and this was obfuscated by the `setRemoteOnPair` hack.

### DIFF
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -58,8 +58,7 @@ WebSocket::WebSocket(jsg::Lock& js,
               ws,
               kj::mv(KJ_REQUIRE_NONNULL(package.maybeTags)),
               package.closedOutgoingConnection)),
-      outgoingMessages(IoContext::current().addObject(kj::heap<OutgoingMessagesMap>())),
-      locality(LOCAL) {}
+      outgoingMessages(IoContext::current().addObject(kj::heap<OutgoingMessagesMap>())) {}
   // This constructor is used when reinstantiating a websocket that had been hibernating, which is
   // why we can go straight to the Accepted state. However, note that we are actually in the
   // `Hibernatable` "sub-state"!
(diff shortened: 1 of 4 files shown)


src/workerd/api/web-socket.c++
```diff
@@ -344,23 +353,31 @@ kj::Promise<DeferredProxy<void>> WebSocket::couple(kj::Own<kj::WebSocket> other)
...
+  }
 
-    // To begin deferred proxying, we can use this magic `KJ_CO_MAGIC` expression, which fulfills
-    // our outer promise for a DeferredProxy<void>, which wraps a promise for the rest of this
-    // coroutine.
-    KJ_CO_MAGIC BEGIN_DEFERRED_PROXYING;
+  // Either:
+  // 1. This websocket is just proxying through, in which case we can allow the IoContext to go
+  // away while still being able to successfully pump the websocket connection.
+  // 2. This is a hibernatable websocket and we are falling through to deferred proxying to
```
jp4a50, line 372: FWIW this approach of treating hibernatable websockets as "deferred-proxyable" feels a little bit hacky and I'm inclined to suggest that the problem it solves should be solved outside this class. However, this is intended to be an NFC so I'm not looking to change this behaviour here.

src/workerd/api/web-socket.c++
```diff
@@ -332,7 +329,19 @@ kj::Promise<DeferredProxy<void>> WebSocket::couple(kj::Own<kj::WebSocket> other)
...
+    }
+    return kj::none;
+  };
+  auto isHibernatable = [&](workerd::api::WebSocket& ws) {
+    KJ_IF_SOME(state, ws.farNative->state.tryGet<Accepted>()) {
+      return state.isHibernatable();
+    }
+    return false;
+  };
+  if (tryGetPeer() != kj::none) {
```
MellowYarker, line 344: Hmm, if I'm understanding correctly, prior to this commit one end of a `WebSocketPair` would be passed to `state.acceptWebSocket()`, which would set the `peer`s locality to `REMOTE`, and set its own state to `Accepted`.

Then, later on we would call `ws.couple(clientWs)`, where `ws` is the other websocket from the `WebSocketPair` (which wasn't accepted and was instead added to the `Response()`). This would mean `locality` would be `REMOTE` and we would _not_ enter this branch where we `registerPendingEvent()`.

After this change, since both ends of the `WebSocketPair()` would have a valid `peer` set (since the ctor for `WebSocketPair()` calls `setPeer()` on both) I suspect we _would_ `registerPendingEvent()` when previously we wouldn't.

I think what you really want here is to add `if (auto p = tryGetPeer(); peer != kj::none && !isHibernatable(peer)) {...}`

jp4a50: Yep, you are absolutely right, and I initially did check `isHibernatable()` before calling `registerPendingEvent()`.

However, `registerPendingEvent()` has no effect for actors anyway, and I thought that this reflected the original intent of`setRemoteOnPair` better (specifically to cause deferred proxying, not necessarily to stop registering the pending event), so I decided to not put a condition in. Also, in theory I suppose we *could* implement something akin to hibernatable websockets for non-actors as well, in which case I think we *would* want to register a pending event.

Having said all that, I'm not averse to putting the `isHibernatable()` check back in if you feel strongly.

MellowYarker: > registerPendingEvent() has no effect for actors

Huh! I had no idea, thanks for pointing this out.

MellowYarker:
Mind opening an internal PR prior to merging this? I'd like to see builds pass there too since lots of hibernation tests are internal only.

MellowYarker (approved):

PR #2377, opened 2024-07-10
Type generating worker

penalosa:
https://github.com/cloudflare/workerd/pull/1959, but just the transition to using a Worker for type generation.


penalosa:
We're doing this differently

PR #2379, opened 2024-07-11
Refactor type generation scripts

penalosa:
This is the first part of landing [Types as a Service](https://github.com/cloudflare/workerd/pull/1959).

This is a relatively large PR, but it's pretty low risk. It mostly contains some refactoring of the type generation code (e.g. files like [types/src/transforms/overrides/compiler.ts](https://github.com/cloudflare/workerd/pull/2379/files#diff-b3672e23392be040c988f7f7a2dac703816fd138336e12d98ab1614cbe169354) which have not changed in functionality). In large part this refactoring is to make it easier for a future PR (another part of #1959) to run this code in a Worker (for instance, changing `import ... from "assert"` to `import ... from "node:assert"` is in support of this goal, as well as the enhancing of the [memory-based compiler host we use for running TypeScript](https://github.com/cloudflare/workerd/pull/2379/files#diff-526655eb36a1321787a081b4d485e11a22d99ded40afdd565a8bc06d8543255a)).

This PR changes the output of the generated `index.d.ts` files to no longer `declare` things for which that's not necessary. This is semantically the same thing, and should have no impact on the usage of these types. We probably should've done it like this all along:

```diff
- declare interface SomeInterface {}
+ interface SomeInterface {}

- declare type SomeType =  {}
+ type SomeType = {}
```
However, this has resulted in some test changes—the behaviour is otherwise unchanged, and the output after this PR should be the same as the output before.

This PR _doesn't_ cover:
- Generating types from internal modules. This means that e.g. `cloudflare:sockets` is still defined by hand, although this PR _does_ add some additional mechanics to make this possible in future.
- The changes made in #1959 to [collation of comments from standards](https://github.com/cloudflare/workerd/pull/1959/files#diff-e4608382ec0590ba5b6b14735401d1fa68bc5c6f30ac856b64f6257af8356b23). This will come in a standalone followup.
- Typechecking internal modules with generated types
- Implementing a worker to generate types
- _Using_ that worker when generating the `@cloudflare/workers-types` package in CI

To aid reviewing, [this link shows the diff](https://github.com/penalosa/workers-types-diff/pull/1/files) between `@cloudflare/workers-types@4.20240701.0` and the output of the type generation scripts run from this branch.

### DIFF
--- a/types/src/generator/index.ts
+++ b/types/src/generator/index.ts
@@ -2,7 +2,7 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
-import assert from "assert";
+import assert from "node:assert";
 import {
   FunctionType,
   Member,
@@ -13,13 +13,13 @@ import {
   Type,
   Type_Which,
 } from "@workerd/jsg/rtti.capnp.js";
-import ts from "typescript";
+import ts, { factory as f } from "typescript";
 import { createStructureNode } from "./structure";
+import { getTypeName } from "./type";
 
 export { getTypeName } from "./type";
-export { parseApiAstDump } from "./parameter-names";
 
-type StructureMap = Map<string, Structure>;
+export type StructureMap = Map<string, Structure>;
 // Builds a lookup table mapping type names to structures
 function collectStructureMap(root: StructureGroups): StructureMap {
   const map = new Map<string, Structure>();
@@ -40,7 +40,7 @@ function collectStructureMap(root: StructureGroups): StructureMap {
 // when certain compatibility flags are enabled (e.g. `Navigator`,
 // standards-compliant `URL`). However, these types are always included in
 // the `*_TYPES` macros.
-function collectIncluded(map: StructureMap): Set<string> {
+function collectIncluded(map: StructureMap, root?: string): Set<string> {
   const included = new Set<string>();
 
   function visitType(type: Type): void {
@@ -101,9 +101,17 @@ function collectIncluded(map: StructureMap): Set<string> {
     }
   }
 
-  // Visit all structures with `JSG_(STRUCT_)TS_ROOT` macros
-  for (const structure of map.values()) {
-    if (structure.getTsRoot()) visitStructure(structure);
+  if (root === undefined) {
+    // If no root was specified, visit all structures with
+    // `JSG_(STRUCT_)TS_ROOT` macros
+    for (const structure of map.values()) {
+      if (structure.getTsRoot()) visitStructure(structure);
+    }
+  } else {
+    // Otherwise, visit just that root
+    const structure = map.get(root);
+    assert(structure !== undefined, `Unknown root: ${root}`);
+    visitStructure(structure);
   }
 
   return included;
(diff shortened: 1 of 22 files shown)


types/src/generator/index.ts
```diff
@@ -144,39 +152,167 @@ function collectClasses(map: StructureMap): Set<string> {
...
+// "node-internal:async_hooks"). We use this to make sure we don't include
+// duplicate definitions if an internal module references a type from another
+// internal module. In this case, we'll include the definition in the one that
+// exported it.
+function collectModuleTypeExports(
+  root: StructureGroups,
+  map: StructureMap
+): Map</* structureName */ string, /* moduleSpecifier */ string> {
+  const typeExports = new Map<string, string>();
+  root.getModules().forEach((module) => {
```
jasnell, line 167: nit: forEach is quite a bit slower in practice than a normal for loop.

types/src/generator/index.ts
```diff
@@ -144,39 +152,167 @@ function collectClasses(map: StructureMap): Set<string> {
...
+}
+
+export function collectTypeScriptModules(root: StructureGroups): string {
+  let result = "";
+
+  root.getModules().forEach((module) => {
+    if (!module.isTsDeclarations()) return;
+    const declarations = module
+      .getTsDeclarations()
+      .replaceAll(/^\/\/\/.+$/gm, (match) => {
```
jasnell, line 305: nit: a comment explaining what this regex is doing would be helpful for future generations

types/src/generator/type.ts
```diff
@@ -103,11 +103,22 @@ export function isUnsatisfiable(typeNode: ts.TypeNode) {
 
 // Strings to replace in fully-qualified structure names with nothing
 const replaceEmpty =
-  /^workerd::api::public_beta::|^workerd::api::|^workerd::jsg::|::|[ >]/g;
+  /^workerd::api::public_beta::|^workerd::api::node::|^workerd::api::gpu::|^workerd::api::|^workerd::jsg::|::|[ >]/g;
```
jasnell, line 106: adding some comments here and when to update this would be helpful. My fear is that this is a bit buried and obscure

types/src/index.ts
```diff
@@ -1,25 +1,30 @@
 #!/usr/bin/env node
-import assert from "assert";
+import { readFileSync, readdirSync } from "fs";
 import { mkdir, readFile, readdir, writeFile } from "fs/promises";
+import assert from "node:assert";
```
jasnell, line 4: nit: given that this is an ESM, likely best to make all of the node built in specifiers use the `node:` prefix for consistency

jasnell (approved):

types/src/generator/index.ts
penalosa: In this case `root.getModules()` is a `capnp-ts` API

types/src/index.ts
penalosa: https://github.com/cloudflare/workerd/pull/2379/commits/409f10dca728a04a7800f20173a96aa75a4ae1b8

types/src/generator/type.ts
penalosa: https://github.com/cloudflare/workerd/pull/2379/commits/409f10dca728a04a7800f20173a96aa75a4ae1b8

types/src/generator/index.ts
penalosa: https://github.com/cloudflare/workerd/pull/2379/commits/409f10dca728a04a7800f20173a96aa75a4ae1b8

types/src/transforms/internal-namespace.ts
```diff
@@ -0,0 +1,210 @@
...
+  root: StructureGroups,
+  structureMap: StructureMap
+) {
+  const moduleRoots = new Map</* specifier */ string, Structure>(); // TODO: add members here as well?
+  root.getModules().forEach((module) => {
+    if (!module.isStructureName()) return;
+    const structure = structureMap.get(module.getStructureName());
+    assert(structure !== undefined);
+    const specifier = module.getSpecifier();
+    // const rootName = getTypeName(module.getStructureName());
```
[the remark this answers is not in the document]
penalosa: https://github.com/cloudflare/workerd/pull/2379/commits/d2b3f91c604455911b93222c8b7a8b0d5eb6fb0f

PR #2380, opened 2024-07-11
Better SQLite error message generation.

smerritt:
The error messages now have the error code in them, either something like SQLITE_TOOBIG or SQLITE_FORMAT if it's a known code, or SQLITE_UNKNOWN_ERROR_CODE(%d) if it's not known to us.

SQLite doesn't provide any function to do this for us. It has an internal function `sqlite3ErrName` that looks a lot like the giant switch statement in this commit, but (a) it's not exposed to library consumers, and (b) it's only present if compiled with certain flags like SQLITE_DEBUG that also turn on a bunch of other stuff.

I considered patching SQLite to expose `sqlite3ErrName`, but I think it's easier for workerd maintainers to deal with adding the occasional new case to a switch statement we own than it is to maintain yet another SQLite patch.

### DIFF
--- a/src/workerd/util/sqlite.c++
+++ b/src/workerd/util/sqlite.c++
@@ -5,6 +5,7 @@
 #include "sqlite.h"
 #include <kj/debug.h>
 #include <kj/refcount.h>
+#include <kj/string-tree.h>
 #include <workerd/util/sentry.h>
 
 #if _WIN32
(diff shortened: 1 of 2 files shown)


src/workerd/util/sqlite.c++
```diff
@@ -30,6 +31,102 @@
...
+  case SQLITE_ERROR:
+    return "SQLITE_ERROR"_kjc;
+  case SQLITE_INTERNAL:
+    return "SQLITE_INTERNAL"_kjc;
+  case SQLITE_PERM:
+    return "SQLITE_PERM"_kjc;
+  case SQLITE_ABORT:
+    return "SQLITE_ABORT"_kjc;
+  case SQLITE_BUSY:
+    return "SQLITE_BUSY"_kjc;
```
jasnell, line 52: it would likely be a bit cleaner to use a macro expansion for this instead...

something like...

```cpp
  #define SQLITE_ERRORS(V) \
    V(SQLITE_OK) \
    V(SQLITE_ERROR) \
    ...

#define V(name) case name: return #name##_kj;
  switch (1) {
    SQLITE_ERRORS(V)
  }
#undef V
```

This would reduce the duplication and make things a bit more compact.

non-blocking tho

src/workerd/util/sqlite.c++
```diff
@@ -30,6 +31,102 @@
...
+
+kj::String namedErrorCode(int errorCode) {
+  // Only include the numeric value if we don't know the name.
+  //
+  // This avoids error messages containing things like "SQLITE_CONSTRAINT(19)" which might leave
+  // users wondering what constraint 19 is; 19 is just the numeric value of SQLITE_CONSTRAINT.
+  KJ_IF_SOME(name, errorName(errorCode)) {
+    return kj::str(name);
+  }
+  return kj::str("SQLITE_UNKNOWN_ERROR_CODE(", errorCode, ")");
```
jasnell, line 116: why not just have `errorName(...)` always return this instead of `kj::none`? Should simplify things a bit.

src/workerd/util/sqlite.c++
```diff
@@ -30,6 +31,102 @@
...
+  case SQLITE_ROW:
+    return "SQLITE_ROW"_kjc;
+  case SQLITE_DONE:
+    return "SQLITE_DONE"_kjc;
+  default:
+    return kj::none;
+  }
+}
+
+kj::String namedErrorCode(int errorCode) {
```
jasnell, line 108: this can likely just return a kj::StringPtr instead

src/workerd/util/sqlite.c++
smerritt: Fair enough. I'm not a big fan of how macro-heavy our code is sometimes, but as long as it's scoped to just this function, it should be fine. Done in commit b15e6940.

src/workerd/util/sqlite.c++
smerritt: It needs to dynamically build the result in the unknown-code case.

src/workerd/util/sqlite.c++
smerritt: Done in fixup commit b15e6940

jasnell (approved):

PR #2385, opened 2024-07-12
Add WebSocketObserver

jp4a50:
This allows observation of traffic on websocket connections terminating within workers.

I'm not 100% satisfied with either the abstraction or the naming (especially `coupledToLocalWebSocket` as a name for a method which creates a `WebSocketObserver` object), but I couldn't think of a clearly better alternative.

Note that this currently won't observe receives on hibernatable websockets. I'll add a TODO for that.

### DIFF
--- a/src/workerd/io/observer.h
+++ b/src/workerd/io/observer.h
@@ -21,11 +21,27 @@ class WorkerInterface;
 class LimitEnforcer;
 class TimerChannel;
 
+class WebSocketObserver: public kj::Refcounted {
+public:
+  // Called when a worker sends a message on this WebSocket (includes close messages).
+  virtual void sentMessage(size_t bytes) { };
+  // Called when a worker receives a message on this WebSocket (includes close messages).
+  virtual void receivedMessage(size_t bytes) { };
+};
+
 // Observes a specific request to a specific worker. Also observes outgoing subrequests.
 //
 // Observing anything is optional. Default implementations of all methods observe nothing.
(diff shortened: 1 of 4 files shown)


jasnell (approved):
LGTM. We need to figure out a way to test these types of changes in workerd

src/workerd/io/observer.h
```diff
@@ -21,11 +21,24 @@ class WorkerInterface;
...
+
 // Observes a specific request to a specific worker. Also observes outgoing subrequests.
 //
 // Observing anything is optional. Default implementations of all methods observe nothing.
 class RequestObserver: public kj::Refcounted {
 public:
+  // Called when a WebSocket terminating at a worker is coupled to the response. A WebSocketObserver
+  // may optionally be returned to observe messages being sent and received on the resulting
+  // websocket connection by the worker.
+  virtual kj::Maybe<kj::Own<WebSocketObserver>> coupledToLocalWebSocket() { return kj::none; };
```
jasnell, line 40: nit: why not simply `tryGetWebSocketObserver()` for the name?

jp4a50: The idea behind the naming was that it is specifically called when the request is converted to an incoming websocket connection which terminates in a "local" websocket i.e. in a worker. `tryGetWebSocketObserver()` doesn't indicate any of those circumstances, but I guess I could just document the above rather than trying to encode it into the name.

jasnell (approved):

PR #2389, opened 2024-07-12
Implement rsa toPem, toDer, and fromJwk utilities

jasnell:
Separating out more of the node.js crypto key handling support. There's a lot to do with all this so I'm piecing it out into separate PRs that individually are easier to review. This adds methods but does not add the stuff that uses it yet. Tests for these individual pieces will also be added separately. A fair amount of this is adapted from https://github.com/nodejs/node/blob/main/src/crypto/crypto_keys.cc

### DIFF
--- a/src/workerd/api/crypto/impl.c++
+++ b/src/workerd/api/crypto/impl.c++
@@ -298,4 +298,28 @@ bool CSPRNG(kj::ArrayPtr<kj::byte> buffer) {
 
   return false;
 }
+
+
+kj::Maybe<kj::ArrayPtr<const kj::byte>> tryGetAsn1Sequence(kj::ArrayPtr<const kj::byte> data) {
+  if (data.size() < 2 || data[0] != 0x30)
+    return kj::none;
+
+  if (data[1] & 0x80) {
+    // Long form.
+    size_t n_bytes = data[1] & ~0x80;
+    if (n_bytes + 2 > data.size() || n_bytes > sizeof(size_t))
+      return kj::none;
+    size_t length = 0;
+    for (size_t i = 0; i < n_bytes; i++)
+      length = (length << 8) | data[i + 2];
+    auto start = 2 + n_bytes;
+    auto end = start + kj::min(data.size() - 2 - n_bytes, length);
+    return data.slice(start, end);
+  }
+
+  // Short form.
+  auto start = 2;
+  auto end = start + kj::min(data.size() - 2, data[1]);
+  return data.slice(start, end);
+}
 }  // namespace workerd::api
--- a/src/workerd/api/crypto/rsa.c++
+++ b/src/workerd/api/crypto/rsa.c++
@@ -269,7 +269,6 @@ kj::Maybe<AsymmetricKeyData> Rsa::fromJwk(KeyType keyType,
 
   static constexpr auto kInvalidBase64Error = "Invalid RSA key in JSON Web Key; invalid base64."_kj;
 
-  // TODO(now): Use a decodeBase64Url variant
   auto nDecoded = toBignumUnowned(simdutfBase64UrlDecodeChecked(n, kInvalidBase64Error));
   auto eDecoded = toBignumUnowned(simdutfBase64UrlDecodeChecked(e, kInvalidBase64Error));
   JSG_REQUIRE(RSA_set0_key(rsa.get(), nDecoded, eDecoded, nullptr) == 1, Error,
@@ -392,14 +391,14 @@ kj::Array<const kj::byte> Rsa::toDer(KeyEncoding encoding,
       switch (encoding) {
         case KeyEncoding::PKCS1: {
           JSG_REQUIRE(i2d_RSAPublicKey_bio(bio.get(), rsa) == 1, Error,
-              "Failed to write RSA public key to PEM", tryDescribeOpensslErrors());
+              "Failed to write RSA public key to DER", tryDescribeOpensslErrors());
           break;
         }
         case workerd::api::KeyEncoding::SPKI: {
           auto evpPkey = OSSL_NEW(EVP_PKEY);
           EVP_PKEY_set1_RSA(evpPkey.get(), rsa);
           JSG_REQUIRE(i2d_PUBKEY_bio(bio.get(), evpPkey.get()) == 1, Error,
-              "Failed to write RSA public key to PEM", tryDescribeOpensslErrors());
+              "Failed to write RSA public key to SPKI", tryDescribeOpensslErrors());
           break;
         }
         default: {
--- a/src/workerd/api/crypto/rsa.c++
+++ b/src/workerd/api/crypto/rsa.c++
@@ -6,7 +6,9 @@
 #include <openssl/bn.h>
 #include <openssl/crypto.h>
 #include <openssl/evp.h>
+#include <openssl/pem.h>
 #include <map>
+#include "simdutf.h"
 
 namespace workerd::api {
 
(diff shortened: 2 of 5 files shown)


src/workerd/api/crypto/rsa.h
```diff
@@ -22,19 +22,38 @@ class Rsa final {
...
+  kj::Array<kj::byte> sign(const kj::ArrayPtr<const kj::byte> data) const KJ_WARN_UNUSED_RESULT;
+
+  static kj::Maybe<AsymmetricKeyData> fromJwk(
+      KeyType keyType,
+      const SubtleCrypto::JsonWebKey& jwk) KJ_WARN_UNUSED_RESULT;
 
   SubtleCrypto::JsonWebKey toJwk(KeyType keytype,
-                                 kj::Maybe<kj::String> maybeHashAlgorithm) const;
+                                 kj::Maybe<kj::String> maybeHashAlgorithm) const
+                                KJ_WARN_UNUSED_RESULT;
```
jasnell, line 35:
```suggestion
                                 KJ_WARN_UNUSED_RESULT;
```

src/workerd/api/crypto/rsa.c++
```diff
@@ -28,6 +30,32 @@ kj::Maybe<T> fromBignum(kj::ArrayPtr<kj::byte> value) {
...
+  auto size = simdutf::maximal_binary_length_from_base64(input.begin(), input.size());
+  auto buf = kj::heapArray<kj::byte>(size);
+  auto result = simdutf::base64_to_binary(input.begin(),
+                                          input.size(),
+                                          buf.asChars().begin(),
+                                          simdutf::base64_url);
+  if (result.error != simdutf::SUCCESS) return kj::none;
+  KJ_ASSERT(result.count <= size);
+  return buf.first(result.count).attach(kj::mv(buf));
+}
```
jasnell, line 53: Reviewer note: This is a more efficient alternative to the existing base64 url decoding method we have based on kj::decodeBase64. This one follows the standard forgiving base64 decode algorithm defined by whatwg specs. Soon I expect this to be moved out of this rsa.c++ file and into a shared location.

src/workerd/api/crypto/impl.c++
```diff
@@ -298,4 +298,28 @@ bool CSPRNG(kj::ArrayPtr<kj::byte> buffer) {
...
+  if (data[1] & 0x80) {
+    // Long form.
+    size_t n_bytes = data[1] & ~0x80;
+    if (n_bytes + 2 > data.size() || n_bytes > sizeof(size_t))
+      return kj::none;
+    size_t length = 0;
+    for (size_t i = 0; i < n_bytes; i++)
+      length = (length << 8) | data[i + 2];
+    auto start = 2 + n_bytes;
+    auto end = start + kj::min(data.size() - 2 - n_bytes, length);
```
fhanau, line 316: By doing kj::min, this ends up returning truncated data if the input data is shorter than it should be as indicated by the length field instead of rejecting it. Is that what we want to do here?

src/workerd/api/crypto/rsa.c++
```diff
@@ -260,6 +478,13 @@ void Rsa::validateRsaParams(jsg::Lock& js,
   }
 }
 
+bool Rsa::isRSAPrivateKey(kj::ArrayPtr<const kj::byte> keyData) {
```
fhanau, line 481: Please add a comment pointing to the specification for this

src/workerd/api/crypto/rsa.c++
```diff
@@ -225,6 +253,196 @@ SubtleCrypto::JsonWebKey Rsa::toJwk(KeyType keyType,
...
+    auto d = JSG_REQUIRE_NONNULL(jwk.d.map([](auto& str) { return str.asPtr(); }), Error,
+                                 "Invalid RSA key in JSON Web Key; missing or invalid "
+                                 "Private Exponent parameter (\"d\").");
+    auto p = JSG_REQUIRE_NONNULL(jwk.p.map([](auto& str) { return str.asPtr(); }), Error,
+                                 "Invalid RSA key in JSON Web Key; missing or invalid "
+                                 "First Prime Factor parameter (\"p\").");
+    auto q = JSG_REQUIRE_NONNULL(jwk.q.map([](auto& str) { return str.asPtr(); }), Error,
+                                 "Invalid RSA key in JSON Web Key; missing or invalid "
+                                 "Second Prime Factor parameter (\"q\").");
+    auto dp = JSG_REQUIRE_NONNULL(jwk.dp.map([](auto& str) { return str.asPtr(); }), Error,
```
fhanau, line 288: Are the CRT parameters required for a JWK RSA key to be valid? I assumed that "plain" RSA keys are also supported.

src/workerd/api/crypto/rsa.c++
```diff
@@ -225,6 +253,196 @@ SubtleCrypto::JsonWebKey Rsa::toJwk(KeyType keyType,
...
+  switch (keyType) {
+    case KeyType::PUBLIC: {
+      switch (encoding) {
+        case KeyEncoding::PKCS1: {
+          JSG_REQUIRE(PEM_write_bio_RSAPublicKey(bio.get(), rsa) == 1, Error,
+              "Failed to write RSA public key to PEM", tryDescribeOpensslErrors());
+          break;
+        }
+        case workerd::api::KeyEncoding::SPKI: {
+          JSG_REQUIRE(PEM_write_bio_RSA_PUBKEY(bio.get(), rsa) == 1, Error,
```
fhanau, line 339: Off-topic: It's really confusing that `PEM_write_bio_RSAPublicKey`/`PEM_write_bio_RSA_PUBKEY` have nearly the same name but are used for different data formats...

src/workerd/api/crypto/rsa.c++
```diff
@@ -225,6 +253,196 @@ SubtleCrypto::JsonWebKey Rsa::toJwk(KeyType keyType,
...
+                                     KeyType keyType,
+                                     kj::Maybe<CipherOptions> options) const {
+  ClearErrorOnReturn clearErrorOnReturn;
+  auto bio = OSSL_BIO_MEM();
+  switch (keyType) {
+    case KeyType::PUBLIC: {
+      switch (encoding) {
+        case KeyEncoding::PKCS1: {
+          JSG_REQUIRE(i2d_RSAPublicKey_bio(bio.get(), rsa) == 1, Error,
+              "Failed to write RSA public key to PEM", tryDescribeOpensslErrors());
```
fhanau, line 395: Nit: `to PEM` needs to be `to DER` here.

src/workerd/api/crypto/rsa.c++
```diff
@@ -225,6 +253,196 @@ SubtleCrypto::JsonWebKey Rsa::toJwk(KeyType keyType,
...
+                               "Modulus parameter (\"n\").");
+  auto e = JSG_REQUIRE_NONNULL(jwk.e.map([](auto& str) { return str.asPtr(); }), Error,
+                               "Invalid RSA key in JSON Web Key; missing or invalid "
+                               "Exponent parameter (\"e\").");
+
+  auto rsa = OSSL_NEW(RSA);
+
+  static constexpr auto kInvalidBase64Error = "Invalid RSA key in JSON Web Key; invalid base64."_kj;
+
+  // TODO(now): Use a decodeBase64Url variant
```
fhanau, line 272: Nit: I assume this refers to wanting to move code into a new decodeBase64Url() function, which is worth doing but not essential in this PR. TODO(now)s within the code base feel a bit scary to me as they indicate high urgency, perhaps this can be TODO(soon)?

LGTM otherwise.

fhanau (approved):

src/workerd/api/crypto/rsa.c++
jasnell: Yep... have had this same conversation before in node ;-)

src/workerd/api/crypto/rsa.c++
jasnell: I'm not sure where the spec actually is. Pulled this from Node.js and implemented it to match node's behavior.

src/workerd/api/crypto/rsa.c++
jasnell, line 395:
```suggestion
              "Failed to write RSA public key to DER", tryDescribeOpensslErrors());
```

src/workerd/api/crypto/rsa.c++
```diff
@@ -225,6 +253,196 @@ SubtleCrypto::JsonWebKey Rsa::toJwk(KeyType keyType,
...
+        case KeyEncoding::PKCS1: {
+          JSG_REQUIRE(i2d_RSAPublicKey_bio(bio.get(), rsa) == 1, Error,
+              "Failed to write RSA public key to PEM", tryDescribeOpensslErrors());
+          break;
+        }
+        case workerd::api::KeyEncoding::SPKI: {
+          auto evpPkey = OSSL_NEW(EVP_PKEY);
+          EVP_PKEY_set1_RSA(evpPkey.get(), rsa);
+          JSG_REQUIRE(i2d_PUBKEY_bio(bio.get(), evpPkey.get()) == 1, Error,
+              "Failed to write RSA public key to PEM", tryDescribeOpensslErrors());
```
jasnell, line 402:
```suggestion
              "Failed to write RSA public key to SPKI", tryDescribeOpensslErrors());
```

src/workerd/api/crypto/rsa.c++
jasnell: May not be strictly required for some cases. Will adjust this as necessary when I start working on the tests and verifying that we're matching node.js' behavior correctly.

src/workerd/api/crypto/rsa.c++
jasnell: Comment is obsolete

jasnell, line 272:
```suggestion
```

src/workerd/api/crypto/impl.c++
jasnell: This would match the behavior in node.js I believe, but I will verify for sure when I start implementing the tests. So if this needs to change it can be changed then.

PR #2402, opened 2024-07-16
Add truncated flag for tail worker events

jasnell:
When set, indicates that the tail worker event was truncated due to the event buffer being full.

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -272,6 +272,7 @@ void Trace::copyTo(rpc::Trace::Builder builder) {
     }
   }
 
+  builder.setTruncated(truncated);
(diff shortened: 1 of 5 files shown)


src/workerd/io/trace.c++
```diff
@@ -621,6 +624,7 @@ void WorkerTracer::addDiagnosticChannelEvent(kj::Date timestamp,
                    message.size();
   if (newSize > MAX_TRACE_BYTES) {
```
jclee, line 625: I think there's also a less-likely `if (newSize > MAX_TRACE_BYTES)` case in the fetch event, to cover things like large header or cfJson values.

jclee (approved):

PR #2403, opened 2024-07-17
Add missed setTruncated call

jasnell:
Missed this in the original commit that added the truncated flag


fhanau (approved):

mikea (approved):

jclee (approved):
Ah, I forgot about that one, too.

Issue #2291, opened 2024-06-19
🐛 Bug: WorkerEntrypoint wrapped with Proxy throws error when invoked


rtbenfield:
A `WorkerEntrypoint` class implementation that is wrapped in a `Proxy` throws the following runtime error when the service binding is invoked. This was created in a Worker with a service binding to its own entrypoint for testing.

```✘ [ERROR] workerd/io/worker.c++:1819: error: worker is not an actor but class name was requested; n = RpcEntrypoint```

```ts
class RpcEntrypointImpl extends WorkerEntrypoint {
	myRpcCall() {
		return { ok: true }
	}
}
export const RpcEntrypoint = new Proxy(RpcEntrypointImpl, {})

// logs RpcEntrypoint__proto__ [Function: WorkerEntrypoint]
console.log('RpcEntrypoint__proto__', RpcEntrypoint.__proto__)
```

I found the error at the line below and attempted to trace how `WorkerEntrypoint` exports are determined. From what I understand, it should be detecting from `__proto__`, though in the JS side this seems to be retained by the proxy wrapper. This is my first time digging through the workerd code though so I could be off track.
https://github.com/cloudflare/workerd/blob/92c30c97870551bb2954c77ba69c419da0c7a058/src/workerd/io/worker.c%2B%2B#L1819

#### Why proxy a WorkerEntrypoint?
I'm attempting the `Proxy` wrapper as a way to apply OpenTelemetry instrumentation to a `WorkerEntrypoint` class. The proxy would extract the env form the constructor for configuration, wrap all RPC method invocations in spans, and export those spans automatically when the RPC method completes.

irvinebroque:
seems like what you're trying to do is kind of equivalent to applying a decorator to all the different RPC methods?

rtbenfield:
Yeah, applying a decorator is essentially the goal. The `Proxy` approach is what [`@microlabs/otel-cf-workers`](https://www.npmjs.com/package/@microlabs/otel-cf-workers) is currently using for all its instrumentation. It makes it simple to apply the decorator to all the methods and automatically instrument the env bindings.

It may be possible to accomplish this with JS decorators on the method level, though I would expect a class-level decorator to have the same problem and need a `Proxy` to work. So one would have to apply the decorator annotation to every method on the class. My understanding is that JS decorators are not in the runtime yet, so we'd also have to do some additional transpiling.

[...]

PR #2406, opened 2024-07-17
Handle proxy object prototypes more correctly

jasnell:
Refs: https://github.com/cloudflare/workerd/issues/2291

### DIFF
--- a/src/workerd/jsg/jsvalue.c++
+++ b/src/workerd/jsg/jsvalue.c++
@@ -137,6 +137,44 @@ JsObject JsObject::jsonClone(Lock& js) {
   return JsObject(obj);
 }
 
+JsValue JsObject::getPrototype(Lock& js) {
+  if (inner->IsProxy()) {
+    // Here we emulate the behavior of v8's GetPrototypeV2() function for proxies.
+    // If the proxy has a getPrototypeOf trap, we call it and return the result.
+    // Otherwise we return the prototype of the target object.
+    // Note that we do not check if the target object is extensible or not, or
+    // if the returned prototype is consistent with the target's prototype if
+    // the target is not extensible. See the comment below for more details.
+    auto proxy = inner.As<v8::Proxy>();
+    JSG_REQUIRE(!proxy->IsRevoked(), TypeError, "Proxy is revoked");
+    auto handler = proxy->GetHandler();
+    JSG_REQUIRE(handler->IsObject(), TypeError, "Proxy handler is not an object");
+    auto jsHandler = JsObject(handler.As<v8::Object>());
+    auto trap = jsHandler.get(js, "getPrototypeOf"_kj);
+    auto target = proxy->GetTarget();
+    if (trap.isUndefined()) {
+      JSG_REQUIRE(target->IsObject(), TypeError, "Proxy target is not an object");
+      // Run this through getPrototype to handle the case where the target is also a proxy.
+      return JsObject(target.As<v8::Object>()).getPrototype(js);
+    }
+    JSG_REQUIRE(trap.isFunction(), TypeError, "Proxy getPrototypeOf trap is not a function");
+    v8::Local<v8::Function> fn = ((v8::Local<v8::Value>)trap).As<v8::Function>();
+    v8::Local<v8::Value> args[] = { target };
+    auto ret = JsValue(check(fn->Call(js.v8Context(), jsHandler.inner, 1, args)));
+    JSG_REQUIRE(ret.isObject() || ret.isNull(), TypeError,
+                "Proxy getPrototypeOf trap did not return an object or null");
+    // TODO(maybe): V8 performs additional checks on the returned value to
+    // see if the proxy and the target are extensible or not, and if the
+    // returned prototype is consistent with the target's prototype if they
+    // are not extensible. To strictly match v8's behavior we should do the
+    // same but (a) v8 does not expose the necessary APIs to do so, and (b)
+    // it is not clear if we actually need to perform the additional check
+    // given how we are currently using this function.
+    return ret;
+  }
+  return JsValue(inner->GetPrototypeV2());
+}
+
 bool JsValue::isTruthy(Lock& js) const {
   KJ_ASSERT(!inner.IsEmpty());
   return inner->BooleanValue(js.v8Isolate);
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -1164,7 +1164,7 @@ private:
       jsg::JsObject object,
       rpc::JsRpcTarget::CallParams::Reader callParams,
       bool allowInstanceProperties) {
-    auto prototypeOfObject = KJ_ASSERT_NONNULL(js.obj().getPrototype().tryCast<jsg::JsObject>());
+    auto prototypeOfObject = KJ_ASSERT_NONNULL(js.obj().getPrototype(js).tryCast<jsg::JsObject>());
 
     // Get the named property of `object`.
     auto getProperty = [&](kj::StringPtr kjName) {
@@ -1231,7 +1231,7 @@ private:
             }
 
             // Decide whether the new object is a suitable RPC target.
-            if (object.getPrototype() == prototypeOfObject) {
+            if (object.getPrototype(js) == prototypeOfObject) {
               // Yes. It's a simple object.
               allowInstanceProperties = true;
             } else if (object.isInstanceOf<JsRpcTarget>(js)) {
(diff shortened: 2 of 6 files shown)


src/workerd/jsg/jsvalue.c++
```diff
@@ -137,6 +137,43 @@ JsObject JsObject::jsonClone(Lock& js) {
...
+      JSG_REQUIRE(target->IsObject(), TypeError, "Proxy target is not an object");
+      return JsValue(target.As<v8::Object>()->GetPrototype());
+    }
+    JSG_REQUIRE(trap.isFunction(), TypeError, "Proxy getPrototypeOf trap is not a function");
+    v8::Local<v8::Function> fn = ((v8::Local<v8::Value>)trap).As<v8::Function>();
+    v8::Local<v8::Value> args[] = { target };
+    auto ret = JsValue(check(fn->Call(js.v8Context(), jsHandler.inner, 1, args)));
+    JSG_REQUIRE(ret.isObject() || ret.isNull(), TypeError,
+                "Proxy getPrototypeOf trap did not return an object or null");
+    // TODO(maybe): V8 performs additional checks on the returned value to
```
kentonv, line 165: Can you give an example (JavaScript code) where this makes a difference? I don't really understand from the description.

kentonv:
Can you link to V8's code for this, for comparison?

jasnell:
https://github.com/v8/v8/blob/main/src/objects/objects.cc#L1379-L1432

src/workerd/jsg/jsvalue.c++
jasnell: The motivation here, as far as I understand it, is that if the target is sealed or frozen (aka non-extensible) then the Proxy trap is not really supposed to be able to change the prototype that it returns as that could change the behavior of the proxied object in a way that violates the seal. For our purposes, however, it's not entirely certain we *need* to worry about that check, and implementing the check at the c++ layer would require patching v8, which I'm trying to avoid here. If we need it, then so be it... In that case I'll just patch v8 to expose v8's JSProxy::GetPrototypeOf implementation as a new method on `v8::Proxy`

src/workerd/jsg/jsvalue.c++
```diff
@@ -137,6 +137,44 @@ JsObject JsObject::jsonClone(Lock& js) {
...
+    JSG_REQUIRE(!proxy->IsRevoked(), TypeError, "Proxy is revoked");
+    auto handler = proxy->GetHandler();
+    JSG_REQUIRE(handler->IsObject(), TypeError, "Proxy handler is not an object");
+    auto jsHandler = JsObject(handler.As<v8::Object>());
+    auto trap = jsHandler.get(js, "getPrototypeOf"_kj);
+    auto target = proxy->GetTarget();
+    if (trap.isUndefined()) {
+      JSG_REQUIRE(target->IsObject(), TypeError, "Proxy target is not an object");
+      // Run this through getPrototype to handle the case where the target is also a proxy.
+      return JsObject(target.As<v8::Object>()).getPrototype(js);
```
harrishancock, line 158: Is it possible to construct a cyclic object which could cause this to infinitely recurse?

jasnell: It *might* be but I would think that our existing cpu limit mitigations would handle that case. That said, I can't think of a way off hand to create a cyclic proxy on the target.

src/workerd/jsg/jsvalue.c++
kentonv: I dunno, we might need it. Keep in mind we follow prototype chains manually in JSRPC, and then invoke the method we find. Will we potentially be violating the seal when we do that?

I don't really understand this explanation though. Isn't your code strictly doing things that someone could do in pure JS? It shouldn't be possible to violate a seal or freeze from pure JS, should it? Again, example code would really help in understanding this.

If we can't be really sure we understand this then I would say we'd better go the route of patching V8 with a `GetPrototypeJS()` that actually runs V8's logic.

jasnell: > Isn't your code strictly doing things that someone could do in pure JS? It shouldn't be possible to violate a seal or freeze from pure JS, should it?

That's the point. Reporting a different prototype when the object is sealed violates the seal and is why v8 has this additional check. The example I showed is the error as it exists in browsers and node.js. What I don't have a good sense of is just how critical this check is. It's certainly an acceptable argument that If We Don't Know, err on the side of caution. It'll likely have to wait until I'm back from PTO tho unless you'd like to take a stab at the v8 patch

kentonv: Wait sorry, when I looked before I missed your first comment somehow and only saw the second.

kentonv: OK, looking at your code example helps. But it confuses me because in my opinion the V8 behavior seems wrong to me.

In my mind a Proxy is just a convenient way to define a wrapper around some other object that intercepts and redirects some or all functionality. The wrapper is a different object from the target. The reason a Proxy has a `target` at all is mostly convenience -- so that functionality you don't explicitly override can fall back to operating on the target.

In that model, it makes no sense to me that because the target happens to be frozen, the proxy would be restricted in what it's allowed to override.

But I am probably not understanding something about proxies.

Still not totally sure what to do here though. I worry that if we implement this "wrong" it could come back to bite us later and will be hard to fix.

jasnell: > Still not totally sure what to do here though. I worry that if we implement this "wrong" it could come back to bite us later and will be hard to fix.

yep, I share this concern. I'm just not sure how likely the issue is to come up in a practical sense. I think I will just go ahead and do the v8 patch but it will have to be when I'm back.

kentonv: Put another way, I don't understand why creating a Proxy that overrides the prototype is being treated as equivalent to re-assigning the prototype on the underlying object. The Proxy is not the same object, and it should be allowed to have whatever prototype it wants IMO.

Maybe we just go ahead with this.

kentonv (approved):
I'm going to go ahead and approve this.

I'm still a little worried about the difference in behavior when the target is frozen, but I actually question whether the V8 behavior is even correct. In any case it doesn't seem like it creates any kind of problem that we can't solve with a compat flag later.

jasnell:
How about this... for now, we can emit a periodic warning log if the proxy is returning a different prototype. This will at least give us a sense of whether anyone is potentially hitting this. I agree we can fix this up with a compat flag later if absolutely necessary

src/workerd/jsg/jsvalue.c++
```diff
@@ -137,6 +137,52 @@ JsObject JsObject::jsonClone(Lock& js) {
...
+    JSG_REQUIRE(ret.isObject() || ret.isNull(), TypeError,
+                "Proxy getPrototypeOf trap did not return an object or null");
+    // TODO(maybe): V8 performs additional checks on the returned value to
+    // see if the proxy and the target are extensible or not, and if the
+    // returned prototype is consistent with the target's prototype if they
+    // are not extensible. To strictly match v8's behavior we should do the
+    // same but (a) v8 does not expose the necessary APIs to do so, and (b)
+    // it is not clear if we actually need to perform the additional check
+    // given how we are currently using this function.
+    if (target->IsObject() && jsg::JsObject(target.As<v8::Object>()) != ret) {
```
kentonv, line 173: As I understand it, this check will almost always fail, because getPrototypeOf() isn't expected to return `target`, it's expected to return `target`'s prototype.

Either way I don't agree with this warning unless the object has actually had `preventExtensions()` applied. It's perfectly valid for a Proxy to return some other prototype. Let's just remove this warning.

kentonv (changes requested):

src/workerd/jsg/jsvalue.c++
jasnell: doh, heh, yeah I'm comparing against the wrong object lol. Sigh. The intent of the warning is to give us some idea if we'd ever actually see the non-extensible case that we're intentionally skipping. But also fine with not having the warning for now.

kentonv: Again, I think if we see the warning, it doesn't necessarily tell us that anything is wrong, because Proxies are actually supposed to be able to override the prototype (if the object is not frozen/sealed).

I'd be OK with the warning if it were limited to frozen/sealed objects, but of course if we could detect that then we could just as well throw the exception that V8 throws.

kentonv (approved):

kentonv:
(Clicked merge since James is on vacation.)

PR #2409, opened 2024-07-17
Redoing "#2073: Update Request to support cache option" with compat flags

AdityaAtulTewari:
RM-18292

Updates the Request object to support specifying a cache mode per the standard. Only undefined, 'no-store' and 'no-cache' may be specified, and the fetch implementation specifically does not implement the semantics for either 'no-store' or 'no-cache' yet. 
Also adds the compatibility flag for `cacheHeaderEnabled`. This just lays the ground work for the additional changes.

The next step is to start adding the support for adding the appropriate header fields in the fetch implementation behind an autogate.

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1076,6 +1076,10 @@ jsg::Ref<Request> Request::constructor(
         }
 
         KJ_IF_SOME(c, initDict.cache) {
+          bool cacheHeaderEnabled = FeatureFlags::get(js).getCacheHeaderEnabled();
+          if(!cacheHeaderEnabled) {
+            JSG_FAIL_REQUIRE(TypeError, kj::str("Unsupported cache mode: ", c, "\nYou may need to enable a compatability flag."));
+          }
           cacheMode = getCacheModeFromName(c);
         }
 
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1076,10 +1076,9 @@ jsg::Ref<Request> Request::constructor(
         }
 
         KJ_IF_SOME(c, initDict.cache) {
-          bool cacheHeaderEnabled = FeatureFlags::get(js).getCacheHeaderEnabled();
-          if(!cacheHeaderEnabled) {
-            JSG_FAIL_REQUIRE(TypeError, kj::str("Unsupported cache mode: ", c, "\nYou may need to enable a compatability flag."));
-          }
+          JSG_REQUIRE(FeatureFlags::get(js).getCacheOptionEnabled(), TypeError, kj::str(
+            "Unsupported cache mode: ", c,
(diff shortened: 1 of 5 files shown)


src/workerd/api/http.h
```diff
@@ -944,6 +944,7 @@ class Request: public Body {
         readonly cf?: Cf;
       });
     }
+    JSG_READONLY_PROTOTYPE_PROPERTY(cache, getCache);
```
jasnell, line 947: nit: let's keep this grouped with the others above

src/workerd/io/compatibility-date.capnp
```diff
@@ -504,4 +504,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("legacy_module_registry")
       $experimental;
   # Enables of the new module registry implementation.
+
+  cacheHeaderEnabled @53 :Bool
+    $compatEnableFlag("cache_header_enabled")
+    $compatDisableFlag("cache_header_disabled")
```
jasnell, line 510: `cache_option_enabled` and `cache_option_disabled` would likely be better. We're not directly impacting the use of the `Cache-Control` header but the option passed in on the `Request` constructor.

src/workerd/api/http.c++
```diff
@@ -1051,6 +1075,14 @@ jsg::Ref<Request> Request::constructor(
               "response status code).");
         }
 
+        KJ_IF_SOME(c, initDict.cache) {
+          bool cacheHeaderEnabled = FeatureFlags::get(js).getCacheHeaderEnabled();
+          if(!cacheHeaderEnabled) {
+            JSG_FAIL_REQUIRE(TypeError, kj::str("Unsupported cache mode: ", c, "\nYou may need to enable a compatability flag."));
```
jasnell, line 1081: nit: be sure to line-wrap at <= 100 cols

jasnell: You can also simplify this slightly using...

```cpp
JSG_REQUIRE(FeatureFlags::get(js).getCacheHeaderEnabled(), TypeError, kj::str(...))`
cacheMode = getCacheModeFromName(c);
```

src/workerd/api/http.h
AdityaAtulTewari: Should I also be splitting this into an instance vs prototype? I'm not exactly sure what line is an appropriate place to put that without splitting it.

jasnell: instance is only here for legacy reasons, anyone who has their compat date set that far back or has that flag turned off likely isn't going to update to support this so I wouldn't bother. Let's keep this as a prototype property.

AdityaAtulTewari: Let me know if this is fixed now

jasnell (approved):

PR #2414, opened 2024-07-19
[types] Add missing `reply` fn to `ForwardableEmailMessage`

### LINKED ISSUE #2283: 🐛 BUG: `ForwardableEmailMessage` is missing `reply` from its type definition
The `ForwardableEmailMessage` interface is [defined](https://github.com/cloudflare/workerd/blob/6ac4ba0296a9d83a96019e2f3ebc44c562b7e660/types/defines/email.d.ts#L18-L44) as follows:

```
interface ForwardableEmailMessage extends EmailMessage {
  readonly raw: ReadableStream<Uint8Array>;
  readonly headers: Headers;
  readonly rawSize: number;

  setReject(reason: string): void;
  forward(rcptTo: string, headers?: Headers): Promise<void>;
}
```

As per https://developers.cloudflare.com/email-routing/email-workers/runtime-api/#emailmessage-definition this type definition should also include a `reply` method: `reply(message: EmailMessage): Promise<void>;`, but it doesn't.

If the docs are the correct source of truth, we should update the `ForwardableEmailMessage` type definition to reflect that.

CarmenPopoviciu:
The `ForwardableEmailMessage` interface should expose a `reply` function, but it currently doesn't.This commit adds the missing function to the interface definition.

Please refer to https://developers.cloudflare.com/email-routing/email-workers/runtime-api/#emailmessage-definition

Fixes #2283

### DIFF
--- a/types/defines/email.d.ts
+++ b/types/defines/email.d.ts
@@ -41,6 +41,12 @@ interface ForwardableEmailMessage extends EmailMessage {
    * @returns A promise that resolves when the email message is forwarded.
    */
   forward(rcptTo: string, headers?: Headers): Promise<void>;
+  /**
+   * Reply to the sender of this email message with a new EmailMessage object.
+   * @param message The reply message.
+   * @returns A promise that resolves when the email message is replied.
+   */
+  reply(message: EmailMessage): Promise<void>;
 }
 
 /**


CarmenPopoviciu:

Issue #483, opened 2023-03-29
HostMetadata calculated to be of `unknown` type as of latest release


jmaroeder:
Hello, with the latest release of workers-types `4.20230321.0`, `HostMetadata` appears to be of `unknown` type, even when specifying a type as part of the parameters to `Request`.

Here is some example code which compiles fine with `4.20230228.0`:

```typescript
interface MyHostMetadata {
  foo: string;
}

let request: Request<MyHostMetadata> = new Request('');

console.log(request.cf?.hostMetadata.foo);
```

However, the latest version leads to the following error from `tsc`:
```
src/index.ts:7:13 - error TS18046: 'request.cf.hostMetadata' is of type 'unknown'.

7 console.log(request.cf?.hostMetadata.foo);
              ~~~~~~~~~~~~~~~~~~~~~~~~
```

mrbbot:
Hey! 👋 Apologies for the very delayed response on this. Looks like this got a little buried. 🙁 To explain what's going on here, this is a simplified snippet of the generated types:

```ts
declare class Request<CfHostMetadata = unknown, Cf = CfProperties<CfHostMetadata>> {
  get cf(): Cf | undefined;
}

declare type CfProperties<HostMetadata = unknown> =
  | IncomingRequestCfProperties<HostMetadata>
  | RequestInitCfProperties;

declare type IncomingRequestCfProperties<HostMetadata = unknown> =
    IncomingRequestCfPropertiesBase
  & IncomingRequestCfPropertiesCloudflareForSaaSEnterprise<HostMetadata>;
declare interface IncomingRequestCfPropertiesBase extends Record<string, unknown> {
  // ...
}
declare interface IncomingRequestCfPropertiesCloudflareForSaaSEnterprise<HostMetadata> {
  hostMetadata: HostMetadata;
}

declare interface RequestInitCfProperties extends Record<string, unknown> {
  // ...
}
```

[...]

PR #2421, opened 2024-07-22
Update Vectorize types for GA release

ndisidore:
### Overview

This updates the types for Vectorize ahead of the GA release.
We have intentionally moved the "current" Vectorize Index to a legacy type, promoting the "new" index to the default `VectorizeIndex` such that it becomes the standard going forward.

This is strictly a types change and sister to
https://github.com/cloudflare/workers-sdk/pull/6252

### DIFF
--- a/src/cloudflare/internal/test/vectorize/vectorize-mock.js
+++ b/src/cloudflare/internal/test/vectorize/vectorize-mock.js
@@ -2,7 +2,7 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
-/** @type {Array<VectorizeMatch>} */
+/** @type {Array<VectorizeQueryMatch>} */
 const exampleVectorMatches = [
   {
     id: "b0daca4a-ffd8-4865-926b-e24800af2a2d",
@@ -130,7 +130,7 @@ export default {
       } else if (request.method === "POST" && pathname.endsWith("/insert")) {
         /** @type {{vectors: Array<VectorizeVector>}} */
         const data = await request.json();
-        if (data.vectors.find((v) => v.id == "fail-with-test-error")) {
+        if (data.vectors.find((v) => v.id === "fail-with-test-error")) {
           return Response.json(
             {
               code: 9999,
@@ -142,34 +142,34 @@ export default {
           );
         }
 
-        return Response.json({
-          ids: [
-            ...data.vectors.map(({ id }) => id),
-            ...exampleVectors.map(({ id }) => id),
-          ],
-          count: data.vectors.length + exampleVectors.length,
-        });
+        /** @type {VectorizeAsyncMutation} */
+        const res = {
+          // fudge a bit and set the mutation id to some internals so our asserts can check more
+          mutationId: `total vectors: ${data.vectors.length + exampleVectors.length}`,
+        };
+        return Response.json(res);
       } else if (request.method === "POST" && pathname.endsWith("/upsert")) {
         /** @type {{vectors: Array<VectorizeVector>}} */
         let data = await request.json();
         if (data.vectors.length > 1) data.vectors.splice(-1);
-        return Response.json({
-          ids: [
-            ...data.vectors.map(({ id }) => id),
-            ...exampleVectors.map(({ id }) => id),
-          ],
-          count: data.vectors.length + exampleVectors.length,
-        });
+        /** @type {VectorizeAsyncMutation} */
+        const res = {
+          // fudge a bit and set the mutation id to some internals so our asserts can check more
+          mutationId: `total vectors: ${data.vectors.length + exampleVectors.length}`,
+        };
+        return Response.json(res);
       } else if (
         request.method === "POST" &&
         pathname.endsWith("/deleteByIds")
       ) {
         /** @type {{ids: Array<string>}} */
         const body = await request.json();
-        return Response.json({
-          ids: body.ids,
-          count: body.ids.length,
-        });
+        /** @type {VectorizeAsyncMutation} */
+        const res = {
+          // fudge a bit and set the mutation id to some internals so our asserts can check more
+          mutationId: `deleted vectors: ${body.ids.length}`,
+        };
+        return Response.json(res);
       } else if (request.method === "POST" && pathname.endsWith("/getByIds")) {
         /** @type {{ids: Array<string>}} */
         const body = await request.json();
--- a/src/cloudflare/internal/vectorize.d.ts
+++ b/src/cloudflare/internal/vectorize.d.ts
@@ -56,11 +56,24 @@ type VectorizeVectorMetadataFilter = {
  */
 type VectorizeDistanceMetric = "euclidean" | "cosine" | "dot-product";
 
-interface VectorizeQueryOptions {
+/**
+ * Metadata return levels for a Vectorize query.
+ *
+ * Default to "none".
+ *
+ * @property all      Full metadata for the vector return set, including all fields (including those un-indexed) without truncation. This is a more expensive retrieval, as it requires additional fetching & reading of un-indexed data.
+ * @property indexed  Return all metadata fields configured for indexing in the vector return set. This level of retrieval is "free" in that no additional overhead is incurred returning this data. However, note that indexed metadata is subject to truncation (especially for larger strings).
+ * @property none     No indexed metadata will be returned.
+ */
+type VectorizeMetadataRetrievalLevel = "all" | "indexed" | "none";
+
+interface VectorizeQueryOptions<
+  MetadataReturn extends boolean | VectorizeMetadataRetrievalLevel = boolean,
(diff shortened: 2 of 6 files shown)


src/cloudflare/internal/vectorize.d.ts
```diff
@@ -126,25 +126,81 @@ interface VectorizeMatches {
...
   count: number;
 }
 
 /**
-* Results of an operation that performed a mutation on a set of vectors
-* with the v2 version of Vectorize.
-* Here, `mutationId` is the identifier for the last mutation processed by Vectorize.
-*/
-interface VectorizeVectorMutationV2 {
-  /* The identifier for the last mutation processed by Vectorize. */
```
ndisidore, line 143: If I understand correctly, this previous comment was incorrect - each mutation gets a unique mutation id specific to the operation

DaniFoldi:
Hi @ndisidore,

I'm really excited for Vectorize GA, it has great potential. However, I have a few concerns about this PR, which I hope you don't mind me raising, in order of most to least important:
- The type of `VectorizeVectorMutation` would suddenly change for all users in a non-backwards-compatible way. If my guess is correct, the behavior of existing v1 datasets wouldn't suddenly change, and therefore the types of my current worker would become out of sync with the actual runtime. See my idea below.
- This removes the `VectorizeMutationV2` interface, which was available in previous workers-types releases, even outside of experimental (if it were only in experimental, this wouldn't be an issue at all). Here's the type in the currently latest release with `2023-07-01` type compat date: https://workers-types.pages.dev/4.20240718.0/2023-07-01/#VectorizeVectorMutationV2. It's very unlikely that anyone external would be using it today, however removing it in a minor is a semver violation, probably better to add the `@deprecated` jsdoc flag so tsc warns users about it, and remove it in the next major.
- I'd also add `@deprecated` to all `...legacy` types, so users are warned if they are using them.

My suggestion for keeping it one type is to use a generic with a default to distinguish between the types. Here's a link to a quick example: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAsgrsAhsAlgewHYDUCMUC8UA3lHAM4QBOAkgCYBcUZwlKGA5lAL4BQoksBMnTYATAWKkKlAHKIAthEbNWHANxQUDJizadefcNHhJUmADxYqZEVAgAPYBAy0yUAOS53UAD4esot6EnjjuAHwSVpQ2mHaOzq7+oVAA-IKmIrhQjCbCmAE8PLQQAMYANoiU0CWYzFDyODlCZhg8APRtUF3dPV0AeimFxeWV1bXA9aJNGRYh4e2dvUsDQ6UVVVA1GHXyAMzTeRjmnoFhC0vLg0A

The key points are: if nothing changes, the old one is used, to avoid changing existing v1 indexes. Users could be told to use either the type with a generic or the V2 override. Wrangler will probably have a way to detect which type a binding should be and it can emit the correct type.

ndisidore:
Hey @DaniFoldi :wave: 
Thanks for you interest and also your feedback!

The generics approach you linked to is a fun one and is certainly something we can consider. We'd looked at something vaguely similar but opted against it primarily because with that, the "naked" type, while it wouldn't be breaking, would default users to the depreciated legacy binding format. This would likely be the case for the forseeable future lest we push the breaking change problem to a later date.
For better or worse, workers types doesn't follow semver format so its tough to do something like this properly.

Given the above, that Vectorize is in beta, and that this change is strictly for types, we've decided that we're okay with a breaking change here, but we still welcome (and are gathering) feedback

DaniFoldi:
Hey @ndisidore :wave:,

Thanks for the response, I appreciate your time!

While I understand your reasoning, I firmly believe that the workers-types package does follow semver (and I believe the original intention was to expose unsafe types only in the `experimental` entrypoint, see DO storage sqlite methods for example, so to me this counts as a stable part of a stable package) and that is also why `EmailExportedHandler` still has a type name that is inconsistent with others (even though it is still in beta), and why `Request` has weird generics (see Brendan's description why: https://github.com/cloudflare/workerd/issues/483#issuecomment-1714140584).

I hope that this api type change will be communicated to users via docs (and I would also add an inline jsdoc mentioning the legacy types for anyone still using v1 and seeing errors after they update) and changelog so users are aware of the changes they need to make. Going forward it could also help if you marked beta types (or all types of beta products that are available in "stable" workers-types entrypoints) with `@alpha`, `@beta` or `@experimental` tags (I've been using jsdoc and tsdoc interchangeably, even though these would be tsdoc in this case).

src/cloudflare/internal/vectorize.d.ts
garvit-gupta: The new comment is more accurate description of a mutation id.

src/cloudflare/internal/vectorize.d.ts
```diff
@@ -126,25 +126,81 @@ interface VectorizeMatches {
...
+ * Results of an operation that performed a mutation on a set of vectors with the post-beta version of Vectorize.
+ * Actual mutations are processed async where the `mutationId` is the unique identifier for the operation.
+ */
+interface VectorizeVectorMutation {
```
garvit-gupta, line 143: We will need some more changes to support types for the new V2 operations. Furthermore, I think there would be some breaking changes we need to account for the query V2, get_by_ids V2 and delete_by_ids V2 responses. For reference: https://github.com/cloudflare/workers-sdk/blob/719cfad52a0e2ca8080eed99f0cbf29a5f8f42cd/packages/wrangler/src/vectorize/types.ts

src/cloudflare/internal/test/vectorize/vectorize-mock.js
garvit-gupta, line 1: We may need some changes in vectorize-api to support the new V2 operations: index_info, create_metadata_index, list_metadata_index, delete_metadata_index.

sejoker (approved):

src/cloudflare/internal/vectorize.d.ts
```diff
@@ -117,15 +119,38 @@ type VectorizeMatch = Pick<Partial<VectorizeVector>, "values"> &
...
 
+/**
+ * Represents a matched vector for a query along with its score and (if specified) the matching vector information.
+ */
+type VectorizeQueryMatch = Pick<Partial<VectorizeVector>, "values"> &
+  Omit<VectorizeVector, "values"> & {
+    /** The score or rank for similarity, when returned as a result */
+    score: number;
+    /** Indexed metadata associated with the vector. */
+    indexedMetadata?: Record<string, VectorizeVectorMetadata>;
```
garvit-gupta, line 138: Are we still maintaining this field? I may have thought wrong, but I understood that we would use a unified `metadata` field in the query match response.

ndisidore: Missed this: lot of in flight changes :octopus: 
This should be fixed and "final" now!

types/defines/vectorize.d.ts
```diff
@@ -177,3 +196,50 @@ declare abstract class VectorizeIndex {
    */
   public getByIds(ids: string[]): Promise<VectorizeVector[]>;
 }
+
+/**
+ * A Vectorize Vector Search Index for querying vectors/embeddings.
+ *
+ * Mutations in this version are async, returning a mutation id.
+ */
+declare abstract class Vectorize {
```
penalosa, line 205: There's a lot of duplication here—did the generics and aliasing approach not work out for some reason?

types/defines/vectorize.d.ts
```diff
@@ -127,15 +142,19 @@ interface VectorizeVectorMutation {
...
+ */
+interface VectorizeAsyncMutation {
+  /** The unique identifier for the async mutation operation containing the changeset. */
   mutationId: string;
 }
 
+/**
+ * A Vectorize Vector Search Index for querying vectors/embeddings.
+ *
+ * @deprecated This is a legacy version from beta version Vectorize.
```
penalosa, line 156: This should not be deprecated yet, since V2 is not yet released and stable

types/defines/vectorize.d.ts
ndisidore: The generics approach would work fine - however, we would lose out on the specificity (and correctness) of some of the tsdoc comments if we took that approach. Compare https://github.com/cloudflare/workerd/pull/2421/files#diff-d770090d5d2a9db9ac5822eb39c8f82411484c0dab8a12cbfd618e924eda58a5R232 & https://github.com/cloudflare/workerd/pull/2421/files#diff-4386ee8e0b9580e2b88770e50a5e19d669bd8d8957df5e215b633330a9f2cd23R224

types/defines/vectorize.d.ts
ndisidore: Replaced with a warning that it will be soon
https://github.com/cloudflare/workerd/pull/2421/files#diff-4386ee8e0b9580e2b88770e50a5e19d669bd8d8957df5e215b633330a9f2cd23R135

Frederik-Baetens (approved):

PR #2425, opened 2024-07-22
Initialize cache control headers, repeat of #2074 for cache: no-store.

AdityaAtulTewari:


src/workerd/api/http.c++
```diff
@@ -1195,6 +1196,22 @@ void Request::shallowCopyHeadersTo(kj::HttpHeaders& out) {
 }
 
 kj::Maybe<kj::String> Request::serializeCfBlobJson(jsg::Lock& js) {
+  if(cacheMode != CacheMode::NONE &&
+    util::Autogate::isEnabled(util::AutogateKey::HTTP_REQUEST_CACHE)) {
+    // Works because of the fact there are only two non-none cache modes
+    auto ttl = (CacheMode::NOSTORE == cacheMode) ? -1 : 0;
```
hoodmane, line 1202: I think it would be good to have a switch here so that if we add a new `CacheMode` it will become a compiler error? Or are we not expecting to add any `CacheMode`s ever?

src/workerd/api/http.c++
```diff
@@ -1195,6 +1196,22 @@ void Request::shallowCopyHeadersTo(kj::HttpHeaders& out) {
...
+    CfProperty clone;
+    KJ_IF_SOME(_, cf.get(js)) {
+      clone = cf.deepClone(js);
+    } else {
+      clone = CfProperty(js, js.obj());
+    }
+    auto obj = KJ_ASSERT_NONNULL(clone.get(js));
+
+    obj.set(js, "cacheTtl", js.num(ttl));
```
hoodmane, line 1211: What does it mean to set `cacheTtl` to `-1`? Documentation for `cacheTtl` says:

> Force response to be cached for a given number of seconds. (e.g. 300)

So I'd expect it to be nonnegative? What is the expected behavior difference between `NOSTORE` and `NOCACHE`?

src/workerd/api/http.c++
```diff
@@ -1816,7 +1836,26 @@ jsg::Promise<jsg::Ref<Response>> fetchImplNoOutputLock(
   // If the jsRequest has a CacheMode, we need to handle that here.
   // Currently, the only cache mode we support is undefined, but we will soon support
   // no-cache and no-store. These additional modes will be hidden behind an autogate.
```
hoodmane, line 1838: Update this comment?

src/workerd/api/http.c++
```diff
@@ -1816,7 +1836,26 @@ jsg::Promise<jsg::Ref<Response>> fetchImplNoOutputLock(
...
-  KJ_ASSERT(jsRequest->getCacheMode() == Request::CacheMode::NONE);
+  auto headerIds = ioContext.getHeaderIds();
+  const auto cacheMode = jsRequest->getCacheMode();
+  if(util::Autogate::isEnabled(util::AutogateKey::HTTP_REQUEST_CACHE)) {
+    switch(cacheMode) {
+      case Request::CacheMode::NOSTORE:
+      case Request::CacheMode::NOCACHE:
+      if(headers.get(headerIds.cacheControl) == kj::none) {
```
hoodmane, line 1845: Why would the `cacheControl`/`pragma` header already be set?

src/workerd/api/http-test.js
```diff
@@ -269,9 +269,14 @@ export const cacheMode = {
...
-        'TypeError',
-        "Unsupported cache mode: no-store");
+      {
+        const req = new Request('https://example.org', {cache : "no-store"});
+        assert.strictEqual(req.cache, "no-store");
+      }
+      {
+        const req = await fetch('https://example.org', { cache: "no-store"});
+        console.log(req);
+      }
```
hoodmane, line 279: Remove this?

src/workerd/api/http-test.js
```diff
@@ -269,9 +269,14 @@ export const cacheMode = {
       await assertFetchCacheRejectsError('no-transform');
       await assertFetchCacheRejectsError('unsupported');
     } else {
-      await assertRequestCacheThrowsError('no-store',
-        'TypeError',
-        "Unsupported cache mode: no-store");
+      {
+        const req = new Request('https://example.org', {cache : "no-store"});
```
hoodmane, line 273: Should there be a test for `no-cache` too? What about a more end-to-end test that covers the changes to `fetchImplNoOutputLock`?

src/workerd/api/http.c++
AdityaAtulTewari: iirc the user can set them.

src/workerd/api/http-test.js
AdityaAtulTewari: yes.

src/workerd/api/http.c++
AdityaAtulTewari: cacheTtl: -1 internally means to bypass cache.

hoodmane: Could you update the documentation for `cacheTtl` to include that?

hoodmane (approved):
Looks good to me.

src/workerd/api/http.c++
```diff
@@ -1176,6 +1178,41 @@ void Request::shallowCopyHeadersTo(kj::HttpHeaders& out) {
 }
 
 kj::Maybe<kj::String> Request::serializeCfBlobJson(jsg::Lock& js) {
+  if (cacheMode != CacheMode::NONE &&
+      util::Autogate::isEnabled(util::AutogateKey::HTTP_REQUEST_CACHE)) {
```
kentonv, line 1182: Given that setting of `cacheMode` is already behind an experimental compatibility flag, it seems unnecessary to also use an autogate here.

kentonv:
Since everything is gated behind `cacheMode` which isn't settable without the experimental compat flag, this looks fine from a "won't break anyone" perspective. I think the autogate can be removed as the compat flag already gates the new logic just fine. Rolling out an autogate would be needless additional work.

AdityaAtulTewari:
That makes sense and the change has been update to reflect not needing an autogate.

src/workerd/api/http.c++
```diff
@@ -1176,7 +1178,41 @@ void Request::shallowCopyHeadersTo(kj::HttpHeaders& out) {
...
+
+  int ttl = 2;
+  switch (cacheMode) {
+    case CacheMode::NOSTORE:
+      ttl = -1;
+      break;
+    case CacheMode::NOCACHE:
+      ttl = 0;
+    case CacheMode::NONE:
+      KJ_ASSERT(false);
```
jasnell, line 1193: `KJ_UNREACHABLE`?

src/workerd/api/http.c++
```diff
@@ -1176,7 +1178,41 @@ void Request::shallowCopyHeadersTo(kj::HttpHeaders& out) {
...
+  if (obj.has(js, "cacheTtl")) {
+    jsg::JsValue oldTtl = obj.get(js, "cacheTtl");
+    JSG_REQUIRE(oldTtl == js.num(ttl), TypeError,
+        kj::str("CacheTtl: ", oldTtl, ", is not compatible with cache: ",
+            getCacheModeName(cacheMode).orDefault("none"_kj), " header."));
+  } else {
+    obj.set(js, "cacheTtl", js.num(ttl));
+  }
+
+  obj.set(js, "cf-cache-level", js.str("byc"_kjc));
```
jasnell, line 1214: This should not be set if `CacheMode::NOCACHE` should it?

AdityaAtulTewari: We aren"t supporting no-cache yet. That's why it there is no break in the switch in that case.

jasnell (approved):

PR #2428, opened 2024-07-23
Check exception type to set retryCountsAgainstLimit for alarms

jqmmes:
If an exception is of type OVERLOADED we set `retryCountsAgainstLimites = true` for alarms.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -467,6 +467,10 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
         }
       }
 
+      // We only want to retry against limits if it's a user error. By default let's check if the
+      // output gate is broken.
+      auto shouldRetryCountsAgainstLimits = !context.isOutputGateBroken();
+
       // We want to alert if we aren't going to count this alarm retry against limits
       if (auto desc = e.getDescription();
           !jsg::isTunneledException(desc) && !jsg::isDoNotLogException(desc)
@@ -476,10 +480,15 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
         // We don't usually log these messages, but it's useful to know the real reason we failed
         // to correctly investigate stuck alarms.
         LOG_NOSENTRY(ERROR, "output lock broke during alarm execution without an interesting error description", actorId, e);
+        if (e.getDetail(jsg::EXCEPTION_IS_USER_ERROR) != kj::none) {
+          // The handler failed because the user overloaded the object. It's their fault, we'll not
+          // retry forever.
+          shouldRetryCountsAgainstLimits = true;
+        }
       }
       return WorkerInterface::AlarmResult {
         .retry = true,
-        .retryCountsAgainstLimit = !context.isOutputGateBroken(),
+        .retryCountsAgainstLimit = shouldRetryCountsAgainstLimits,
         .outcome = outcome
       };
     })
@@ -497,6 +506,9 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
             actorId = kj::str(s);
           }
         }
+        // We only want to retry against limits if it's a user error. By default let's assume it's our
+        // fault.
+        auto shouldRetryCountsAgainstLimits = false;
         if (auto desc = e.getDescription();
             !jsg::isTunneledException(desc) && !jsg::isDoNotLogException(desc)) {
           if (isInterestingException(e)) {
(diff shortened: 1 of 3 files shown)


src/workerd/api/global-scope.c++
```diff
@@ -467,6 +467,10 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
         }
       }
 
+      // We only want to retry against limits if it's an user error. By default let's assume it's our
+      // fault.
+      auto shouldRetryAgainstLimits = false;
```
MellowYarker, line 472: Can we change this to `shouldRetryCountAgainstLimits`?

src/workerd/api/global-scope.c++
```diff
@@ -467,6 +467,10 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
         }
       }
 
+      // We only want to retry against limits if it's an user error. By default let's assume it's our
```
MellowYarker, line 470: nit: if it's _a_ user error

src/workerd/api/global-scope.c++
```diff
@@ -497,6 +506,9 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
             actorId = kj::str(s);
           }
         }
+        // We only want to retry against limits if it's an user error. By default let's assume it's our
```
MellowYarker, line 509: nit: if it's _a_ user error

src/workerd/api/global-scope.c++
```diff
@@ -467,6 +467,10 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
         }
       }
 
+      // We only want to retry against limits if it's an user error. By default let's assume it's our
+      // fault.
+      auto shouldRetryAgainstLimits = false;
+
       // We want to alert if we aren't going to count this alarm retry against limits
       if (auto desc = e.getDescription();
```
justin-mp, line 475: I find the way this is structured confusing now, especially since both legs have `context.isOutputGateBroken` and we now OR together two things say that the retryCountsAgainstLimit.

I would much prefer to see the `!context.isOutputGateBroken()` be part of the calculation of the `shouldRetryAgainstLimits` variable so that the return value can look like:
```
return WorkerInterface::AlarmResult {
        .retry = true,
        .retryCountsAgainstLimit = shouldRetryAgainstLimits,
        .outcome = outcome
      };
```

(at which point you should also rename the variable `retryCountsAgainstLimit`.

justin-mp (changes requested):

src/workerd/api/global-scope.c++
```diff
@@ -476,10 +480,15 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(
         // We don't usually log these messages, but it's useful to know the real reason we failed
         // to correctly investigate stuck alarms.
         LOG_NOSENTRY(ERROR, "output lock broke during alarm execution without an interesting error description", actorId, e);
+        if (e.getType() == kj::Exception::Type::OVERLOADED) {
+          // The handler failed because the user overloaded the object. It's their fault, we'll not
```
a-robinson, line 484: Overloaded errors are not currently always the user's fault, so I don't think we can use this broad of a check without tightening up where OVERLOADED errors can happen.

To be specific, one case that I have in mind is that if storage operations slow way down (e.g. if the underlying storage layer is unhealthy), then this particular OVERLOADED error gets thrown and the user really has no ability to do anything about it: https://github.com/cloudflare/workerd/blob/main/src/workerd/io/worker.c%2B%2B#L3016-L3018

a-robinson (changes requested):

src/workerd/api/global-scope.c++
justin-mp: Given that we saw exceededCpu errors, is there a way to check for that specific kind of overload?

jqmmes: > Overloaded errors are not currently always the user's fault, so I don't think we can use this broad of a check without tightening up where OVERLOADED errors can happen.
> 
> To be specific, one case that I have in mind is that if storage operations slow way down (e.g. if the underlying storage layer is unhealthy), then this particular OVERLOADED error gets thrown and the user really has no ability to do anything about it: [`main`/src/workerd/io/worker.c++#L3016-L3018](https://github.com/cloudflare/workerd/blob/main/src/workerd/io/worker.c%2B%2B?rgh-link-date=2024-07-23T19%3A07%3A16Z#L3016-L3018)

For this specific error case, we can match either the description (which I don't think is a good practice, but would catch this specific case for now) or check for the presence of `broken.exceededMemory`...

jqmmes:
I've pushed an update to the PR.
@a-robinson this update now checks if we didn't fail due to `broken.outputGateBroken`. Not sure if this is enough for us to feel comfortable marking the retry as not internal-error.
@justin-mp I've removed the OR logic for the outputGateBroken as that's now part of the `shouldRetryCountsAgainstLimit` variable.

jqmmes:
Following our discussion today, I've updated the PR adding a new custom exception detail that allow us to know if a specific exception is, or is not, a user error.

CC @justin-mp / @a-robinson

src/workerd/io/actor-cache.c++
```diff
@@ -243,6 +243,9 @@ void ActorCache::evictOrOomIfNeeded(Lock& lock) {
     // Add trace info sufficient to tell us which operation caused the failure.
     exception.addTraceHere();
     exception.addTrace(__builtin_return_address(0));
+    // We know this exeption happens due to user error. Let's add an exception detail so we can
+    // parse it later.
+    exception.setDetail(jsg::EXCEPTION_IS_USER_ERROR, kj::heapArray<byte>(0));
```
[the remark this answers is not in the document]
jqmmes: Yup, I just used the same style we use elsewhere

justin-mp (approved):

PR #2429, opened 2024-07-24
Gate prototypes behind compatability flag for cache: no-store

AdityaAtulTewari:
Small bug fix, where prototype for `cache` was not gated behind the compatibility flag in #2409.

### DIFF
--- a/src/workerd/api/http-test.js
+++ b/src/workerd/api/http-test.js
@@ -85,6 +85,7 @@ export default {
 
 export const inspect = {
   async test(ctrl, env, ctx) {
+    const cacheEnabled = env.cacheEnabled === "true"
     // Check URL with duplicate search param keys
     const url = new URL("http://user:pass@placeholder:8787/path?a=1&a=2&b=3");
     assert.strictEqual(util.inspect(url),
@@ -124,7 +125,8 @@ export const inspect = {
       body: "message",
       headers: { "Content-Type": "text/plain" }
     });
-    assert.strictEqual(util.inspect(request),
+    if(cacheEnabled) {
+      assert.strictEqual(util.inspect(request),
 `Request {
   cache: undefined,
   keepalive: false,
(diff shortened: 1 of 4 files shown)


src/workerd/api/http-test.js
```diff
@@ -209,17 +233,36 @@ export const inspect = {
...
       assert.strictEqual(req.cache, undefined);
     }
     {
-      const req = new Request('https://example.org', { cache: 'no-store' });
-      assert.strictEqual(req.cache, 'no-store');
+      if(cacheEnabled) {
+        const req = new Request('https://example.org', { cache: 'no-store' });
+        assert.strictEqual(req.cache, 'no-store');
+      } else {
+        assert.throws(() => {
```
jp4a50, line 246: Shouldn't we also be putting this new behaviour (throwing when `cache` is present) behind the compatibility flag too?

kentonv:
Can we please completely revert the previous change, and then re-land? I want to be able to see a diff from the known-good state to the final state, so I can verify that everything new has been properly gated.

src/workerd/api/http-test.js
kentonv: When the compat flag is off, we need to be using the old `jsg::Unimplemented` definition, which would make this code irrelevant.

AdityaAtulTewari: we can't exactly be using that dynamically it's part of the type right? Should we just be throwing the same error?

kentonv: Sure you can make it dynamic. Rename `jsg::Unimplemented getCache();` to `jsg::Unimplemented getUnimplementedCache();`, then do something like this in the `JSG_RESOURCE_TYPE` block:

```
if (flags.getCacheOptionEnabled()) {
  JSG_READONLY_PROTOTYPE_PROPERTY(cache, getCache);
} else {
  JSG_READONLY_PROTOTYPE_PROPERTY(cache, getUnimplementedCache);
}
```

So the exported `cache` property in JS hooks up to different implementations depending on the flag.

PR #2430, opened 2024-07-24
Load Pyodide runtime from external capnproto file

hoodmane:
These changes are needed to load the Pyodide runtime from an external capnproto file. In workerd, we'll keep loading Pyodide in the same way, but we moved this to WorkerdAPI. In our internal codebase, we now download the bundle and call `setPyodideBundleData` when we load the first Python worker. Then when setting up the module registry, we add the bundle from `pyodideBundleGlobal`.

TODO in followups:
1. Make workerd load the bundle in the same way
2. Add support for multiple versions of the Pyodide bundle

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -7,6 +7,16 @@
 
 namespace workerd::api::pyodide {
 
+kj::Maybe<kj::Array<unsigned char>> pyodideBundleDataGlobal = kj::none;
+kj::Maybe<kj::Own<capnp::FlatArrayMessageReader>> pyodideBundleReaderGlobal = kj::none;
+kj::Maybe<jsg::Bundle::Reader> pyodideBundleGlobal = kj::none;
+
+void setPyodideBundleData(kj::Array<unsigned char> data) {
+  pyodideBundleReaderGlobal = kj::heap<capnp::FlatArrayMessageReader>(kj::arrayPtr(
+        reinterpret_cast<const capnp::word*>(data.begin()), data.size() / sizeof(capnp::word))).attach(kj::mv(data));
+  pyodideBundleGlobal = KJ_REQUIRE_NONNULL(pyodideBundleReaderGlobal)->getRoot<jsg::Bundle>();
+}
+
 static int readToTarget(kj::ArrayPtr<const kj::byte> source, int offset, kj::ArrayPtr<kj::byte> buf) {
   int size = source.size();
   if (offset >= size || offset < 0) {
--- a/src/workerd/api/modules.h
+++ b/src/workerd/api/modules.h
@@ -20,7 +20,9 @@ namespace workerd::api {
 template <class Registry>
 void registerModules(Registry& registry, auto featureFlags) {
   node::registerNodeJsCompatModules(registry, featureFlags);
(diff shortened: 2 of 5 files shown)


src/workerd/api/pyodide/pyodide.c++
```diff
@@ -7,6 +7,17 @@
...
+kj::Maybe<kj::Array<unsigned char>> pyodideBundleDataGlobal = kj::none;
+kj::Maybe<kj::Own<capnp::FlatArrayMessageReader>> pyodideBundleReaderGlobal = kj::none;
+kj::Maybe<jsg::Bundle::Reader> pyodideBundleGlobal = kj::none;
+
+void setPyodideBundleData(kj::Array<unsigned char> data) {
+  pyodideBundleReaderGlobal = kj::heap<capnp::FlatArrayMessageReader>(kj::arrayPtr(
+        reinterpret_cast<const capnp::word*>(data.begin()), data.size() / sizeof(capnp::word)));
+  pyodideBundleGlobal = KJ_REQUIRE_NONNULL(pyodideBundleReaderGlobal)->getRoot<jsg::Bundle>();
+  pyodideBundleDataGlobal = kj::mv(data);
+}
```
dom96, line 19: I don't feel too strongly about this, but I think it might be worth to reduce the number of globals here. Creating a reader should be fast, so maybe you could instead only keep the `pyodideBundleDataGlobal` global and create the `jsg::Bundle::Reader` on-demand?

src/workerd/server/workerd-api.c++
```diff
@@ -446,7 +446,9 @@ void WorkerdApi::compileModules(
     if (hasPythonModules(confModules)) {
       KJ_REQUIRE(featureFlags.getPythonWorkers(),
           "The python_workers compatibility flag is required to use Python.");
-      // Inject pyodide bootstrap module.
+      // Inject Pyodide bundle
+      modules->addBuiltinBundle(PYODIDE_BUNDLE, kj::none);
```
dom96, line 450: The `PYODIDE_BUNDLE` will get replaced with `pyodideBundleGlobal` here, right? Maybe add a todo.

dom96 (approved):

src/workerd/server/workerd-api.c++
hoodmane: Yeah that's a todo, I guess @garrettgu10 is working on it =)

src/workerd/api/pyodide/pyodide.c++
hoodmane: I don't understand the ownership semantics here, I'm concerned that the `jsg::Bundle::Reader` doesn't own its data and instead the data is owned by one of `pyodideBundleReaderGlobal` or `pyodideBundleDataGlobal`, so if I let either of those go out of scope while the `ModuleRegistry` object is still around, then when a file is imported, we'll read from freed data.

dom96: Looking at other examples of `capnp::FlatArrayMessageReader` it seems it is important to keep it all alive. But it looks like you can reduce the number of globals at least a little bit by using `attach`: https://github.com/cloudflare/workerd/blob/main/src/workerd/server/workerd.c%2B%2B#L991

Issue #698, opened 2023-04-07
🐛 BUG: The cache field on RequestInitializerDict is not implemented in fetch


dgraham:
### Which Cloudflare product(s) does this pertain to?

Workers for Platforms

### What version of `Wrangler` are you using?

Latest

### What operating system are you using?

All

### Describe the Bug

Making a fetch request with the `cache` field throws the following exception.

```
The cache field on RequestInitializerDict is not implemented
```

```js
const response = await fetch(url, {
  cache: 'no-store',
  method: 'POST',
  body: JSON.stringify(body),
  headers: {'Content-Type': 'application/json'}
})
```

This was reported in our database driver in https://github.com/planetscale/database-js/pull/102. We're using the `cache` attribute to advise the runtime to avoid caching any responses since the response body is an SQL result set that varies on every request.

A similar issue was reported in https://github.com/cloudflare/workers-sdk/issues/192, but I'm not sure what the resolution was  there.

Is support for the standard [Request.cache](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache) attribute planned? Ideally, we'd like to treat all JS runtimes identically and rely on the fetch interfaces to hide implementation details from the driver. Even "supporting" the attribute without throwing, but with no actual implementation, would be helpful here.

ignoramous:
Still seeing this:

```
✘ [ERROR]   Error: The 'cache' field on 'RequestInitializerDict' is not implemented.
```

Any workaround till this attr is implemented?

MaazDev:
@ignoramous if ur using planetscale https://github.com/planetscale/database-js/pull/102#issuecomment-1508219636

satpalsr:
I have same error with upstash redis db `The 'cache' field on 'RequestInitializerDict' is not implemented.`
```
import { Redis } from "@upstash/redis"

const redis = new Redis({
  url: process.env.UPSTASH_REDIS_REST_URL!,
  token: process.env.UPSTASH_REDIS_REST_TOKEN!,
})


try {
      data = await redis.get(userEmail);
} catch (err) {
      return new Response(`Could not access DB ${err}`, { status: 500 })
}
```

LarchLiu:
> I have same error with upstash redis db `The 'cache' field on 'RequestInitializerDict' is not implemented.` I deployed Next.js app with pages.
> 
> ```
> import { Redis } from "@upstash/redis"
> 
> const redis = new Redis({
[...]

PR #2434, opened 2024-07-24
Http Request Cache Part 1

AdityaAtulTewari:
RM-18292

refs https://github.com/cloudflare/workerd/issues/698 and https://github.com/cloudflare/workers-sdk/issues/2514

Follow-up to reintroduce the behavior that was reverted in #2432 

Updates the Request object to support specifying a cache mode per the standard. Only undefined, 'no-store' and 'no-cache' may be specified, and the fetch implementation specifically does not implement the semantics for either 'no-store' or 'no-cache' yet.
Also adds the compatibility flag for cacheHeaderEnabled. This just lays the ground work for the additional changes.

The next step is to start adding the support for adding the appropriate header fields in the fetch implementation behind an autogate


src/workerd/api/http.h
```diff
@@ -694,9 +694,10 @@ struct RequestInitializerDict {
   jsg::WontImplement credentials;
 
   // In browsers this controls the local browser cache. For Cloudflare Workers it could control the
-  // Cloudflare edge cache. Note that this setting is different from using the `Cache-Control`
-  // header since `Cache-Control` would be forwarded to the origin.
-  jsg::Unimplemented cache;
+  // Cloudflare edge cache. While the standard defines a number of values for this property, our
+  // implementation supports only three: undefined (identifying the default caching behavior that
+  // has been implemented by the runtime), "no-store", and "no-cache".
+  jsg::Optional<kj::String> cache;
```
kentonv, line 700: I think we have to guard this type change with the compat flag.

But I see now the issue you have with this, which is that JSG_STRUCT doesn't currently support making properties conditional on compat flag. It may be necessary to extend the JSG_STRUCT macro somehow, or allow multiple JSG_STRUCT declarations where one is chosen based on flags...

AdityaAtulTewari: Is it not sufficient to error on construction?

AdityaAtulTewari: Perhaps also we could have another type that is WasUnimplemented, that checks the compat flag upon construction?

kentonv: > Is it not sufficient to error on construction?

Hmm. I guess it could be, as long as:
* Every function which accepts RequestInit performs the check immediately.
* We throw _exactly_ the same error that we have in the past. This may sound a little extreme, but given we've already broken people twice by doing things that we thought wouldn't break anyone, I think we have to be extreme in our caution at this point.

Here's an idea: What if we extend JSG_STRUCT so that you can give the struct a `validate()` method, which is automatically called after construction, and has the opportunity to throw an exception. That can be used to confidently solve the first point above.

AdityaAtulTewari: > > Is it not sufficient to error on construction?
> 
> Hmm. I guess it could be, as long as:
> 
> * Every function which accepts RequestInit performs the check immediately.
> * We throw _exactly_ the same error that we have in the past. This may sound a little extreme, but given we've already broken people twice by doing things that we thought wouldn't break anyone, I think we have to be extreme in our caution at this point.

So after looking through this I somewhat disagree with my earlier statement because in the future if we decide to implement some behavior that was previously `jsg::Unimplemented` we want it to be intuitive to gate just that field/function.


> Here's an idea: What if we extend JSG_STRUCT so that you can give the struct a validate() method, which is > automatically called after construction, and has the opportunity to throw an exception. That can be used to confidently solve the first point above.

While this method seems good it doesn't allow for the normal type inferences that `jsg::Unimplemented` does to generate it's error messages

Here is an idea I had for an alternative:
We have a templated type called `jsg::WasUnimplemented<T>`, then this type is also castable to `jsg::Unimplemented`.
In the `WasUnimplemented` constructor, the desired function call to check the feature flag is passed in, and immediately checked, this then triggers the same `jsg::Unimplemented` error.
I'm not 100% sure of all the details, but the essence is we need a type that represents something that we have now implemented and at one point hadn't to preserve our error messages.

kentonv: I don't understand. Why do you need the type inference? You can just copy exactly the error string that was generated before, for this specific field, and have your code in `validate()` throw that message -- it should then be identical behavior, right?

AdityaAtulTewari: My idea seems incoherent and doesn't makes sense, the validate has been added in the last commit.

src/workerd/api/http-test.wd-test
```diff
@@ -8,11 +8,24 @@ const unitTests :Workerd.Config = (
           ( name = "worker", esModule = embed "http-test.js" )
         ],
         bindings = [
-          ( name = "SERVICE", service = "http-test" )
+          ( name = "SERVICE", service = "http-test" ),
+          ( name = "cacheEnabled", text = "false" ),
```
kentonv, line 12: The name you chose here uses an inconsistent style from the immediate preceding line.

Please make an effort to be consistent with surrounding code, even in the absence of specific rules.

```suggestion
          ( name = "CACHE_ENABLED", text = "false" ),
```

src/workerd/jsg/struct.h
```diff
@@ -118,10 +118,15 @@ class StructWrapper<Self, T, TypeTuple<FieldWrappers...>, kj::_::Indexes<indices
     //   it prescribes lexicographically-ordered member initialization, with base members ordered
     //   before derived members. Objects with mutating getters might be broken by this, but it
     //   doesn't seem worth fixing absent a compelling use case.
-
-    return T {
+    auto t = T {
       kj::get<indices>(fields).unwrap(static_cast<Self&>(*this), isolate, context, in)...
     };
+
+    if constexpr (requires { t.validate(); }) {
```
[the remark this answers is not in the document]
AdityaAtulTewari: Should I add a similar comment to JSG_STRUCT.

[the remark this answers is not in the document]
kentonv: Yeah that sounds like a good idea.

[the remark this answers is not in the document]
jasnell: This comment was resolved but it doesn't look like the suggested comment was added here? I see it added to the JSG_STRUCT comment tho.

src/workerd/jsg/jsg.h
```diff
@@ -637,6 +637,9 @@ namespace {
 // all in JavaScript when the optional is null in C++ (as opposed to the field being present but
 // assigned the value `undefined`).
 //
+// Note that if a `validate` function is provided, then it will be called after the struct is
+// unwrapped from v8. This would be an appropriate time to throw an error.
```
jasnell, line 641: nit: might be good to include the example from the `README.md` here also but up to you.  At the very least let's document the expected signature and indicate that `validate` is expected to throw its own errors and return void.

src/workerd/api/http.h
```diff
@@ -751,6 +752,8 @@ struct RequestInitializerDict {
   // jsg::Optional<kj::String> priority;
   // TODO(conform): Might support later?
 
+  void validate();
```
jasnell, line 755: nit: let's move this to after the `JSG_STRUCT` and `JSG_STRUCT_TS_OVERRIDE` macros and add a brief comment pointing folks to the new validate functionality in the `JSG_STRUCT` definition. Otherwise this might be a bit confusion for someone coming into this code fresh thinking they might need to call `validate()` themselves.

src/workerd/api/http.c++
```diff
@@ -1771,6 +1815,17 @@ jsg::Promise<jsg::Ref<Response>> fetchImplNoOutputLock(
   kj::HttpHeaders headers(ioContext.getHeaderTable());
   jsRequest->shallowCopyHeadersTo(headers);
 
+  // If the jsRequest has a CacheMode, we need to handle that here.
+  // Currently, the only cache mode we support is undefined, but we will soon support
+  // no-cache and no-store. These additional modes will be hidden behind an autogate.
+  if (jsRequest->getCacheMode() != Request::CacheMode::NONE) {
```
jasnell, line 1821: nit: Given the introduction of the `validate()` method, can this be simplified to a `KJ_ASSERT`?  I'm fine with this as is but the `validate()` check should gate the acceptable value(s) before we get to this point correct?

src/workerd/api/http.c++
```diff
@@ -1169,6 +1198,17 @@ kj::Maybe<kj::String> Request::serializeCfBlobJson(jsg::Lock& js) {
...
 }
 
+void RequestInitializerDict::validate() {
+  KJ_IF_SOME(c, cache){
+    // Check compatability
+    JSG_REQUIRE(Worker::Api::current().getFeatureFlags().getCacheOptionEnabled(), Error, kj::str(
+      "The 'cache' field on 'RequestInitializerDict' is not implemented."));
+
+    // Check that the string is a supported type
+    auto cacheStruct = getCacheModeFromName(c);
```
jasnell, line 1208: We ought to be able to enforce the `!= Request::CacheMode::NONE` check here while the `no-cache` and `no-store` support is still pending, correct?

src/workerd/jsg/README.md
```diff
@@ -640,6 +640,23 @@ the necessary mapping between the C++ struct and the JavaScript object. Only the
...
 
+If the struct has a validate() method, it is called when the struct is unwrapped from v8.
+This is an opportunity for it to throw a TypeError based on some custom logic.
+
+```cpp
+struct ValdiatingFoo {
+  kj::String abc;
+
+  void validate() {
+    JSG_REQUIRE(abc.size() != 0, TypeError, "Field 'abc' had no length in 'ValidatingFoo'.");
```
jasnell, line 651: suggestion: might be worthwhile expanding the example a bit to show how to incorporate a compat flag check into the validation... but given that JSG doesn't directly know about compat flags, I wouldn't say it is critical

src/workerd/jsg/struct.h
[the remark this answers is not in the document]
AdityaAtulTewari: Sorry about that, I did not push that.

src/workerd/api/http.c++
AdityaAtulTewari: In this PR yes. In general changing this seems to cause problems with the returning promise in `fetch` tests.

AdityaAtulTewari: So yes, but then how do we "test" these changes?

src/workerd/jsg/README.md
AdityaAtulTewari: I added a comment explaing that you have to use the direct API. It might make sense to force passing a reference to a jsg lock, but I wasn't sure how to make one in `tryUnwrap`.

jasnell: I beliefve you can get the current `jsg::Lock&` from the `info.GetIsolate()` like `jsg::Lock::from(info.GetIsolate())` etc

AdityaAtulTewari: do we want `validate` to have to take in a `jsg::Lock`?

jasnell: I think it would be a good idea.

src/workerd/jsg/struct.h
```diff
@@ -118,10 +118,19 @@ class StructWrapper<Self, T, TypeTuple<FieldWrappers...>, kj::_::Indexes<indices
...
-
-    return T {
+    auto t = T {
       kj::get<indices>(fields).unwrap(static_cast<Self&>(*this), isolate, context, in)...
     };
+
+    // Note that if a `validate` function is provided, then it will be called after the struct is
+    // unwrapped from v8. This would be an appropriate time to throw an error.
+    // Signature: void validate(jsg::Lock& js);
+    if constexpr (requires { t.validate(); }) {
```
jasnell, line 128: should the ` requires { t.validate(); }` here include the `jsg::Lock&` argument?

src/workerd/api/http.h
```diff
@@ -751,13 +752,18 @@ struct RequestInitializerDict {
   // jsg::Optional<kj::String> priority;
   // TODO(conform): Might support later?
 
+
   JSG_STRUCT(method, headers, body, redirect, fetcher, cf, mode, credentials, cache,
              referrer, referrerPolicy, integrity, signal);
   JSG_STRUCT_TS_OVERRIDE(RequestInit<Cf = CfProperties> {
     headers?: HeadersInit;
     body?: BodyInit | null;
     cf?: Cf;
```
jasnell, line 761: hmm, does the TS override here need to account for the difference in the `cache` property definition? I can't recall off the top of my head how the `jsg::Unimplemented cache;` is currently reflected in the generated types.

AdityaAtulTewari: we might want to introduce a test case for that. Is there another TS test case that I can look at that has similar behavior.

src/workerd/jsg/struct.h
AdityaAtulTewari: I think it should. That is a good point.

kentonv: Hmm... why didn't this lead to a test failure?

AdityaAtulTewari: When you don't have an argument it just assumes you will take `void validate(any...);

AdityaAtulTewari:
Based on the results of this PR: https://github.com/cloudflare/workerd/pull/2480 the typescript results are no different. Do we want to add tests for every frontend?

jasnell (approved):
Need @kentonv to review.

Also, @workers-frameworks folks.. can you look at this from a type generation persoective also. /cc @jculvey @dario-piotrowicz

kentonv (approved):
I've reviewed the interface changes and am convinced no change is observable in production if the compat flag isn't enabled.

I haven't carefully reviewed implementation nor TS types -- leaving that to James and others.

kentonv:
Note that when this rolls out we need to pay extra-close attention to the workers that were broken in the past. Please make sure to inform whoever is going to do the next release after this is merged.

src/workerd/jsg/jsg.h
```diff
@@ -641,6 +641,20 @@ namespace {
...
+// Signature: void validate(jsg::Lock& js);
+// Example:
+// struct ValdiatingFoo {
```
dario-piotrowicz, line 648:
```suggestion
// struct ValidatingFoo {
```

src/workerd/jsg/README.md
```diff
@@ -640,6 +640,24 @@ the necessary mapping between the C++ struct and the JavaScript object. Only the
 properties listed in the macro are mapped. (In this case, `onlyInternal` is not
 included in the JavaScript object.)
 
+If the struct has a validate() method, it is called when the struct is unwrapped from v8.
+This is an opportunity for it to throw a TypeError based on some custom logic.
+The signature for this method is `void validate(jsg::Lock&);`
+
+```cpp
+struct ValdiatingFoo {
```
dario-piotrowicz, line 648:
```suggestion
struct ValidatingFoo {
```

dario-piotrowicz (approved):
Looks good to me (from the types generation point of view) 🙂

dario-piotrowicz:
In the non-experimental types files we get this addition:
Screenshot 2024-08-19 at 09 33 11

So, the comment plus the `?: undefined` declaration

I think this quite ugly... but maybe acceptable? 🤷 

Ideally I think we should not add anything in those files, but I have no idea about the technical limitations and whatnot, if this is the best we can do I think it's fine

dario-piotrowicz (approved):
Looks perfect! thanks for the ts improvement @AdityaAtulTewari! 🫶

PR #2442, opened 2024-07-25
fix compile_flags.txt to include correct paths

anonrig:
Updates compile_flags.txt to include correct paths for ada-url and simdutf.

### DIFF
--- a/compile_flags.txt
+++ b/compile_flags.txt
@@ -3,8 +3,6 @@
 -xc++
 -nostdinc
 -Ibazel-bin/external/dawn/include
--Ibazel-bin/external/ada-url/_virtual_includes/ada-url/
--Ibazek-bin/external/simdutf/virtual_includes/simdutf/
 -Ibazel-bin/external/com_cloudflare_lol_html/_virtual_includes/lolhtml
 -Ibazel-bin/external/perfetto/
 -Iexternal/perfetto-sdk/sdk/
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,0 +1,23 @@
+# Development
+
+## Clangd
+
+To get language support in VSCode and other IDEs we recommend using `clangd`-based language server.
+The server reads supplied `compile_flags.txt` file for correct options and include paths to resolve symbols
+in opened files.
+
+To support `clangd` project-level operations like `Find References` `compile_commands.json` file needs to
+be generated that will list all files in the project. If you do this, then it needs to be re-generated periodically
+when new files appear.
+
+1.  Install `gen-compile-commands`
+
+```sh
+just prepare
+```
+
+2.  Generate `compile_commands.json`
+
+```sh
+just compile-commands
+```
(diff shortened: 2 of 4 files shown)


mikea (approved):

compile_flags.txt
```diff
@@ -3,8 +3,8 @@
 -xc++
 -nostdinc
 -Ibazel-bin/external/dawn/include
--Ibazel-bin/external/ada-url/_virtual_includes/ada-url/
--Ibazek-bin/external/simdutf/virtual_includes/simdutf/
+-Ibazel-bin/external/ada-url/
```
fhanau, line 6: I don't have headers listed in either of these paths – using just `external/simdutf/` and `external/ada-url/` instead seems to work instead. (In some cases we need the bazel-bin/ prefix because there are generated headers but here this is not the case.)

anonrig: Since, we already have `Iexternal/ada-url` and simdutf, I'll just remove these lines then. Thanks for the review @fhanau

docs/development.md
```diff
@@ -0,0 +1,17 @@
+# Development
+
+## Compile flags
+
+In order to fix Intellisense issues in VSCode and several other IDEs related to missing header files in C++ dependencies such as `ada-url` and `simdutf`, you need to generate a `compile_commands.json` document. Follow the steps below to fix the issue:
```
mikea, line 5: Strictly speaking this is not true: compile_flags.txt is enough for clangd to resolve includes. What doesn't work is find references, since clangd can't enumerate all files in the project.
You need to generate compile_commands.json only if you want these additional features to work.

anonrig: Would you mind recommending a replacement for this line?

docs/development.md
```diff
@@ -0,0 +1,17 @@
...
+
+1.  Install `gen-compile-commands`
+
+```sh
+cargo install --git https://github.com/mikea/gen-compile-commands`
```
mikea, line 10: let's replace with `cargo install gen-compile-commands`

docs/development.md
mikea:
```
## Clangd

To get language support in VSCode and other IDEs we recommend using `clangd`-based language server.
The server reads supplied `compile_flags.txt` file for correct options and include paths to resolve symbols
in opened files.

To support `clangd` project-level operations like `Find References` `compile_commands.json` file needs to
be generated that will list all files in the project. If you do this, then it needs to be re-generated periodically
when new files appear.

....
```

docs/development.md
```diff
@@ -0,0 +1,17 @@
...
+```
+
+2.  Generate `compile_commands.json`
+
+```sh
+gen-compile-commands --root . --compile-flags compile_flags.txt --out compile_commands.json --src-dir ./src
```
mikea, line 16: let's add Justfile and give `just compile-commands` instructions?

mikea (approved):

docs/development.md
anonrig: sounds good!

fhanau (approved):

PR #2443, opened 2024-07-25
Implement Vectorize GA binding changes

ndisidore:
Sister to https://github.com/cloudflare/workerd/pull/2421

This creates a new shared implementation of the Vectorize type that is common to both v1 and v2, conditionally switching on the version to serve up the correct api. This allows us to keep the module name common as `cloudflare-internal:vectorize-api`

### DIFF
--- a/src/cloudflare/internal/vectorize-api.ts
+++ b/src/cloudflare/internal/vectorize-api.ts
@@ -16,144 +16,166 @@ enum Operation {
   VECTOR_DELETE = 5,
 }
 
-class VectorizeIndexImpl implements VectorizeIndex {
+type VectorizeVersion = "v1" | "v2";
+
+/*
+ * The Vectorize beta VectorizeIndex shares the same methods, so to keep things simple, they share one implementation.
+ * The types here are specific to Vectorize GA, but the types here don't actually matter as they are stripped away
+ * and not visible to end users.
+ */
+class VectorizeIndexImpl implements Vectorize {
   public constructor(
     private readonly fetcher: Fetcher,
-    private readonly indexId: string
+    private readonly indexId: string,
+    private readonly indexVersion: VectorizeVersion
   ) {}
 
-  public async describe(): Promise<VectorizeIndexDetails> {
-    const res = await this._send(
-      Operation.INDEX_GET,
-      `indexes/${this.indexId}`,
-      {
-        method: "GET",
-      }
-    );
+  public async describe(): Promise<VectorizeIndexInfo> {
+    const endpoint =
+      this.indexVersion === "v2" ? `info` : `binding/indexes/${this.indexId}`;
+    const res = await this._send(Operation.INDEX_GET, endpoint, {
+      method: "GET",
+    });
 
-    return await toJson<VectorizeIndexDetails>(res);
+    return await toJson<VectorizeIndexInfo>(res);
   }
 
   public async query(
     vector: VectorFloatArray | number[],
-    options: VectorizeQueryOptions
+    options: VectorizeQueryOptions<VectorizeMetadataRetrievalLevel>
   ): Promise<VectorizeMatches> {
-    const compat = {
-      queryMetadataOptional: flags.vectorizeQueryMetadataOptional,
-    };
-    const res = await this._send(
-      Operation.VECTOR_QUERY,
-      `indexes/${this.indexId}/query`,
-      {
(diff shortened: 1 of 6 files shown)


src/cloudflare/internal/vectorize-v2-api.ts
ndisidore, line 1: Not super excited about this having a `-v2-` in this path.
With wrapped bindings, I'm not 100% sure how dynamic they are..
If we were to move the existing `vectorize-api` to `vectorize-beta-api` and promote this to `vectorize-api`, I'm not sure if it would break existing users or if they would be okay since things are build at compile time.

ndisidore: Confirmed with the runtime team that we can't rename existing modules and have them continue to work, so we may have to live with this (and not really that big of a deal)

mikea: > Confirmed with the runtime team that we can't rename existing modules and have them continue to work, so we may have to live with this (and not really that big of a deal)

It is true that if you change your module name, then existing pipeline will break. There are two ways out of this:

- create `-v2-` binding as you did with its own consequences
- make existing binding confrom to both standards by using version switch.

Since I'd prefer us not to have a copy of vectorize-api.ts, let me expand on it a little:

You change your api to look like:

```
export function makeBinding(env: {
  fetcher: Fetcher;
  indexId: string;
  version?: "v2";
}) {
}
```

and make your implementation respect version value. Now old pipelines won't have version at all, so you will fall back to the old implementation.

New pipelines will have "v2" version inner binding specified and will trigger new behavior.

You will have to change the way new pipelines are generated, so I'd rather you added version in there rather
that copies of api, unless changes are so drastic, which they are not.

As for types, I think only current path needs to be properly typed. Old path can use "any" all the way it wants.

mikea (changes requested):

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -16,144 +16,164 @@ enum Operation {
   VECTOR_DELETE = 5,
 }
 
-class VectorizeIndexImpl implements VectorizeIndex {
+/*
+ * The Vectorize beta VectorizeIndex shares the same methods, so to keep things simple, they share one implementation.
```
ndisidore, line 20: This is what we discussed @mikea and I this holds true, I think we're happy with this approach!

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -217,8 +237,13 @@ async function toJson<T = unknown>(response: Response): Promise<T> {
 export function makeBinding(env: {
   fetcher: Fetcher;
   indexId: string;
-}): VectorizeIndex {
-  return new VectorizeIndexImpl(env.fetcher, env.indexId);
+  indexVersion?: "v2";
```
ndisidore, line 240: IIRC capn'p has a `number` type so this must be a string

mikea (approved):

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -16,144 +16,164 @@ enum Operation {
...
-    const res = await this._send(
-      Operation.INDEX_GET,
-      `indexes/${this.indexId}`,
-      {
-        method: "GET",
-      }
-    );
+  public async describe(): Promise<VectorizeIndexInfo> {
+    const endpoint =
+      this.indexVersion === "v2" ? `info` : `binding/indexes/${this.indexId}`;
```
ndisidore, line 33: For v2, account info and index id are included in the base url of the fetcher so these endpoints just become the final path part.

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -217,8 +239,13 @@ async function toJson<T = unknown>(response: Response): Promise<T> {
...
   fetcher: Fetcher;
   indexId: string;
-}): VectorizeIndex {
-  return new VectorizeIndexImpl(env.fetcher, env.indexId);
+  indexVersion?: VectorizeVersion;
+}): Vectorize {
+  return new VectorizeIndexImpl(
+    env.fetcher,
+    env.indexId,
+    env.indexVersion ?? "v1"
```
ndisidore, line 247: This ensures backwards compatibility, since previously built pipelines won't have these inner globals.
We default to v1 if is it not present.

PR #2444, opened 2024-07-25
EW-8589 add a compat date to support custom ports for worker subrequests

danlapid:


src/workerd/io/compatibility-date.capnp
```diff
@@ -517,4 +517,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # Enables bypassing FL by translating pipeline tunnel configuration to subpipeline.
   # This flag is used only by the internal repo and not directly by workerd.
 
+  allowCustomPorts @55 :Bool
+    $compatEnableFlag("allow_custom_ports")
+    $compatEnableDate("2024-08-05");
```
kentonv, line 522: Not sure if you intend to reopen this later but FWIW I would suggest introducing this as `$experimental` first and only opening it up and giving it a date once the implementation has rolled out and is known to work.

PR #2445, opened 2024-07-25
Generate package memory snapshot for validator.

dom96:
Modifies memory snapshot generation to generate a _package_ snapshot- one that only includes imports of modules that are not present in the user's worker bundle.

Reuses the existing AST walker to collect the import names and pushes them out to the validator.

Tested upstream.

### DIFF
--- a/src/pyodide/internal/process_script_imports.py
+++ b/src/pyodide/internal/process_script_imports.py
@@ -1,6 +1,12 @@
+# This script is used to prepare a worker prior to a _package_ memory snapshot being taken.
+# All it does is walk through the imports in each of the worker's modules and attempts to import
+# them. Local imports are not possible because the worker file path is explicitly removed from the
+# module search path.
+CF_LOADED_MODULES=[]
 def _do_it():
     import ast
     from pathlib import Path
+    import sys
 
     def find_imports(source: str) -> list[str]:
         try:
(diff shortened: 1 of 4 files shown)


src/pyodide/internal/process_script_imports.py
```diff
@@ -23,12 +29,20 @@ def process_script(script):
...
+                CF_LOADED_MODULES.append(mod)
             except ImportError:
                 pass
 
     def process_scripts():
-        for script in Path("/session/metadata").glob("**/*.py"):
+        # Currently this script assumes that it is generating a _package_ snapshot- one that
+        # only includes non-vendored packages. Because of this we do not wish to import local
+        # modules, the easiest way to ensure they cannot be imported is to remove
+        # `/session/metadata` from the sys path.
```
hoodmane, line 40: Sure, that makes sense. I wonder should we instead delay putting `"/session/metadata"` on the path until after this code? But I guess it doesn't make too much of a difference.

hoodmane (approved):
Looks good to me. Thanks!

hoodmane:
Changes like this are going to become annoying after we start making compat dates correspond to a fixed pyodide capnp bundle...

src/workerd/api/pyodide/pyodide.h
```diff
@@ -178,10 +186,11 @@ class ArtifactBundler : public jsg::Object {
     auto& cb = KJ_REQUIRE_NONNULL(uploadMemorySnapshotCb);
     hasUploaded = true;
     auto& context = IoContext::current();
-    return context.awaitIo(js, cb(kj::mv(snapshot)));
+    return context.awaitIo(js, cb({
+        .snapshot = kj::mv(snapshot), .importedModulesList = {}}));
   };
 
-  void storeMemorySnapshot(jsg::Lock& js, kj::Array<kj::byte> snapshot) {
+  void storeMemorySnapshot(jsg::Lock& js, MemorySnapshotResult snapshot) {
```
hoodmane, line 193: Instead of a `MemorySnapshotResult` I'd recommend that this take two separate arguments, the snapshot and the list of strings. That way it will be backwards compatible to the previous api and cause less trouble with #2430.

src/pyodide/internal/process_script_imports.py
dom96: The advantage of doing it here is that it ensures the `/session/metadata` path is not there right before the imports are done, so it's less likely we'll accidentally include it in the path.

src/workerd/api/pyodide/pyodide.h
dom96: I'll see how painful rebasing is.

PR #2447, opened 2024-07-26
EW-8589 add a compat date to support custom ports for worker subrequests

danlapid:
This PR enables worker subrequests to fetch hosts in any port on sites that are grey-clouded or cloudflare proxy standard ports on orange-clouded sites.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -517,4 +517,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # Enables bypassing FL by translating pipeline tunnel configuration to subpipeline.
   # This flag is used only by the internal repo and not directly by workerd.
 
+  allowCustomPorts @55 :Bool
+      $compatEnableFlag("allow_custom_ports")
+      $experimental;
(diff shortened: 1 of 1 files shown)


danlapid:

danlapid:
recheck

src/workerd/io/compatibility-date.capnp
```diff
@@ -517,4 +517,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # Enables bypassing FL by translating pipeline tunnel configuration to subpipeline.
   # This flag is used only by the internal repo and not directly by workerd.
 
+  allowCustomPorts @55 :Bool
+    $compatEnableFlag("allow_custom_ports")
+    $compatEnableDate("2024-08-05");
```
kentonv, line 522: We should define this as experimental first, implement it and verify that it works as expected, and only then go back and set a default-on date. Otherwise there's a risk that it doesn't actually work yet on the default-on date and then we're in big trouble.

```suggestion
    $experimental;
```

danlapid: Makes sense, I amended it in.

kentonv (approved):

PR #2448, opened 2024-07-26
webgpu: move GPU access to external process

edevil:
Support having a separate process that talks to the GPU, and to which workerd can talk to using the dawn wire procotol. There are now two implementations of `DawnContainer`, one that relies on a remote connection via the wire protocol, and another that attempts to access the GPU directly via Dawn native, which is the default.

To use the remote protocol workerd needs to be built with the enable_experimental_webgpu_remote flag and the `webgpuBackend` config needs to be set with the correct path for the unix socket on which the server is listening on.

A server can be launched via the //src/workerd/server:voodoo module passing as a parameter the unix socket path where it will be listening on.

### DIFF
--- a/src/workerd/api/BUILD.bazel
+++ b/src/workerd/api/BUILD.bazel
@@ -15,6 +15,8 @@ filegroup(
                 "**/*test*.c++",
                 "pyodide.c++",
                 "pyodide/pyodide.c++",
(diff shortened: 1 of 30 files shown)


jasnell:
The WebGPU API has been removed.

PR #2451, opened 2024-07-26
Load Pyodide runtime through HTTP

garrettgu10:
This loads the Pyodide bundle through HTTP. It does this in a synchronous way, by creating a new kj::Thread and immediately disowning it, causing the current thread to be blocked on the completion of the new thread.

When a disk cache directory is provided at the command line using pyodide-bundle-disk-cache, the directory is used to cache pyodide bundles.

Existing test cases have been edited to use the `pyodide.capnp.bin` file generated through bazel, so changes in `src/pyodide` will be reflected instantly if tested through `bazel test ...`. Otherwise, the correct version according to compat date will be fetched from the internet. 

Here's a plan for how to test the disk cache writing (reading is covered by bazel tests) functionality:
```
mkdir /tmp/pyodide
./bazel-bin/src/workerd/server/workerd serve samples/pyodide/config.capnp --experimental --pyodide-bundle-disk-cache-dir=/tmp/pyodide --verbose
# should print loading pyodide package from internet
./bazel-bin/src/workerd/server/workerd serve samples/pyodide/config.capnp --experimental --pyodide-bundle-disk-cache-dir=/tmp/pyodide --verbose
# should no longer print loading pyodide package from internet & just work
```


src/workerd/server/workerd-api.c++
```diff
@@ -446,7 +447,9 @@ void WorkerdApi::compileModules(
     if (hasPythonModules(confModules)) {
       KJ_REQUIRE(featureFlags.getPythonWorkers(),
           "The python_workers compatibility flag is required to use Python.");
-      // Inject pyodide bootstrap module.
+      // Inject Pyodide bundle
+      modules->addBuiltinBundle(KJ_ASSERT_NONNULL(pyodideBundleGlobal), kj::none);
```
hoodmane, line 451:
```suggestion
      if(Autogate::isEnabled(util::AutogateKey::PYODIDE_LOAD_EXTERNAL)) {
          modules->addBuiltinBundle(KJ_ASSERT_NONNULL(pyodideBundleGlobal), kj::none);
      }
```

garrettgu10:
One point of jank is that the tip-of-tree pyodide bundle is still named `pyodide-2.capnp.bin` since the version number is still hardcoded in workerd. Ideally we want this to be named `pyodide-latest.capnp.bin` or something similar but this would require us to finish the story on compat-date -> pyodide version mapping.

garrettgu10:
Another thing we need to figure out before merging -- how can we determine if a worker is a Python worker or not? We need to know this so that we don't fetch the Pyodide bundle for non-Python test cases.

hoodmane:
Since it's blocking, why not call `fetchPyodideBundle` in workerd-api just before it's needed?

src/workerd/server/server.c++
```diff
@@ -2565,6 +2566,83 @@ void Server::abortAllActors() {
   }
 }
 
+kj::Path getPyodideBundlePath(kj::String &version) {
```
hoodmane, line 2569: Maybe `getPyodideBundleFileName`? I guess it's returning a path, but it's a single component path.

src/pyodide/BUILD.bazel
```diff
@@ -1,3 +1,5 @@
+package(default_visibility = ["//visibility:public"])
```
hoodmane, line 1: Why not just add this to the rules you use?

src/workerd/server/server.c++
hoodmane: Also I think the idomatic kj thing is to accept a `StringPtr` rather than a `String&`.
```suggestion
kj::Path getPyodideBundlePath(kj::StringPtr version) {
```

src/workerd/server/server.c++
```diff
@@ -2565,6 +2566,83 @@ void Server::abortAllActors() {
...
+  }
+}
+
+void Server::fetchPyodideBundle(api::pyodide::PythonConfig& pyConfig) {
+  auto version = kj::str("2"); //TODO: hardcoded version number, should probably get this from pyConfig eventually
+
+  if(api::pyodide::hasPyodideBundle(kj::str(version))) {
+    KJ_LOG(WARNING, "Pyodide version ", kj::str(version), " already exists in pyodide bundle table");
+    return;
+  }
```
hoodmane, line 2600: Nit: I don't think you need all these `kj::str`'s:
```suggestion
  auto version = "2"_kj; //TODO: hardcoded version number, should probably get this from pyConfig eventually

  if(api::pyodide::hasPyodideBundle(version)) {
    KJ_LOG(WARNING, "Pyodide version ", version, " already exists in pyodide bundle table");
    return;
  }
```

src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -1,42 +1,50 @@
 load("//:build/wd_test.bzl", "wd_test")
 
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+
+copy_file(
+    name = "pyodide-2.capnp.bin@rule",
+    src = "//src/pyodide:pyodide.capnp.bin",
```
hoodmane, line 7: I think this should be version `"dev"` or something like that, since it uses the bundle from workerd source. To get version 2 we should download it with bazel. If we're not ready to do that, maybe just call this `pyodide-dev.capnp.bin` and set the version to `"dev"` in the code. Don't want to confuse ourselves more than necessary.

dom96:
> finish the story on compat-date -> pyodide version mapping.

This story is complete with https://github.com/cloudflare/workerd/pull/2464. You should be able to use it to determine the bundle filename to download.

> how can we determine if a worker is a Python worker or not?

By checking the WorkerBundle. We've got the `hasPythonModule` for this in EW which you should be able to recreate here.

garrettgu10:
@dom96 ah, makes sense. We should probably copy pyodide-2.capnp.bin to pyodide-0.26.0a2.capnp.bin on our bucket so we can use that.

src/workerd/server/tests/python/BUILD.bazel
dom96: A comment in the bazel file here to explain this also wouldn't go amiss :)

src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -1,59 +1,67 @@
...
+copy_file(
+    name = "pyodide-dev.capnp.bin@rule",
+    src = "//src/pyodide:pyodide.capnp.bin",
+    out = "pyodide-bundle-cache/pyodide-dev.capnp.bin"
+)
+
 wd_test(
     src = "hello/hello.wd-test",
-    args = ["--experimental"],
+    args = ["--experimental", "--pyodide-bundle-disk-cache-dir", "$(location pyodide-dev.capnp.bin@rule)/.."],
```
dom96, line 13: Might be time to define a `py_wd_test` and avoid repeating this

src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -1,59 +1,67 @@
...
-#     args = ["--experimental", "--disk-cache-dir", "../all_pyodide_wheels"],
+#     args = ["--experimental", "--pyodide-package-disk-cache-dir", "../all_pyodide_wheels", "--pyodide-bundle-disk-cache-dir", "$(location pyodide-dev.capnp.bin@rule)/.."],
 #     data = glob(
 #         [
 #             "langchain/*",
 #         ],
 #         exclude = ["**/*.wd-test"],
-#     ) + ["@all_pyodide_wheels//:whls"],
+#     ) + ["@all_pyodide_wheels//:whls", "pyodide-dev.capnp.bin@rule"],
 # )
```
dom96, line 54: Should we remove this?

src/workerd/server/workerd-api.c++
```diff
@@ -446,7 +533,17 @@ void WorkerdApi::compileModules(
...
+      if(util::Autogate::isEnabled(util::AutogateKey::PYODIDE_LOAD_EXTERNAL)) {
+          if (fetchPyodideBundle(impl->pythonConfig, "dev"_kj)) {
+            modules->addBuiltinBundle(getPyodideBundle("dev"_kj), kj::none);
+          } else {
+            // TODO: hardcoded version number
+            KJ_REQUIRE(fetchPyodideBundle(impl->pythonConfig, "2"_kj), "Failed to get both dev and hardcoded Pyodide version");
+            modules->addBuiltinBundle(getPyodideBundle("2"_kj), kj::none);
+          }
+      }
+      // Inject pyodide bootstrap module (TODO: load this from the capnproto bundle?)
```
dom96, line 546: Do we do so already in EW? If so we should do it here too.

dom96 (commented):
LGTM but I think you should integrate `PythonSnapshotRelease` into here before merging. You probably want to move `getPythonSnapshotRelease` out of EW and into workerd, then use it here to get the right Python version to load.

src/workerd/server/workerd-api.c++
garrettgu10: I don't think we do. This is the three-line module that does basically nothing

garrettgu10:
Nice!

src/workerd/server/tests/python/BUILD.bazel
garrettgu10: Good idea. I was debating adding this to wd_test but your suggestion is way cleaner

hoodmane:
Could you split "refactor bazel test setup" out into a separate PR?

src/workerd/io/compatibility-date.c++
```diff
@@ -296,4 +296,64 @@ kj::Maybe<kj::String> normalizeCompatDate(kj::StringPtr date) {
   return CompatDate::parse(date).map([](auto v) { return v.toString(); });
 }
 
+struct PythonSnapshotParsedField {
```
garrettgu10, line 299: For reviewers: this is a copy of the relevant code from downstream -- there will be a small refactor to make downstream use this version instead.

src/workerd/server/tests/python/BUILD.bazel
dom96: seems to still be using wd_test here

src/workerd/server/workerd-api.c++
```diff
@@ -432,6 +436,89 @@ kj::Maybe<jsg::ModuleRegistry::ModuleInfo> WorkerdApi::tryCompileModule(
...
+    kj::Path filename = getPyodideBundleFileName(version);
+    auto replacer = dir->replaceFile(filename, kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+    replacer->get().writeAll(bytes);
+    replacer->commit();
+  }
+}
+
+bool fetchPyodideBundle(const api::pyodide::PythonConfig& pyConfig, kj::StringPtr version) {
+  if(api::pyodide::hasPyodideBundle(version)) {
```
dom96, line 465: Nit
```suggestion
  if (api::pyodide::hasPyodideBundle(version)) {
```

src/workerd/server/workerd-api.c++
```diff
@@ -432,6 +436,89 @@ kj::Maybe<jsg::ModuleRegistry::ModuleInfo> WorkerdApi::tryCompileModule(
   KJ_UNREACHABLE;
 }
 
+kj::Path getPyodideBundleFileName(kj::StringPtr version) {
+  return kj::Path(kj::str("pyodide-", version, ".capnp.bin"));
+}
+
+kj::Maybe<kj::Own<const kj::ReadableFile>> getPyodideBundleFile(const kj::Maybe<kj::Own<const kj::Directory>> &maybeDir, kj::StringPtr version) {
```
dom96, line 443: nit: wrap the line

src/workerd/server/workerd-api.c++
```diff
@@ -432,6 +436,89 @@ kj::Maybe<jsg::ModuleRegistry::ModuleInfo> WorkerdApi::tryCompileModule(
...
+    kj::Path filename = getPyodideBundleFileName(version);
+    auto file = dir->tryOpenFile(filename);
+
+    return file;
+  }
+
+  return kj::none;
+}
+
+void writePyodideBundleFileToDisk(const kj::Maybe<kj::Own<const kj::Directory>> &maybeDir, kj::StringPtr version, kj::ArrayPtr<byte> bytes) {
```
dom96, line 454: nit: wrap the line

src/workerd/server/workerd-api.c++
```diff
@@ -446,7 +533,18 @@ void WorkerdApi::compileModules(
     if (hasPythonModules(confModules)) {
       KJ_REQUIRE(featureFlags.getPythonWorkers(),
           "The python_workers compatibility flag is required to use Python.");
-      // Inject pyodide bootstrap module.
+      // Inject Pyodide bundle
+      if(util::Autogate::isEnabled(util::AutogateKey::PYODIDE_LOAD_EXTERNAL)) {
+          if (fetchPyodideBundle(impl->pythonConfig, "dev"_kj)) {
+            modules->addBuiltinBundle(getPyodideBundle("dev"_kj), kj::none);
+          } else {
+            auto version = KJ_ASSERT_NONNULL(getPythonSnapshotRelease(featureFlags)).getPyodide();
```
dom96, line 541: I wonder if we actually want to use more than just the Pyodide version in the bundle filename. Presumably we want `pyodide` + `pyodideRevision` + `backport`. @hoodmane what are your thoughts on this?

hoodmane: Agreed, `pyodide` + `pyodideRevision` + `backport` sounds right. But also, we need to upload a bundle with this name before this will work...

src/workerd/server/tests/python/BUILD.bazel
garrettgu10: It's removed in the separate "refactor python bazel tests" PR that hood requested

src/workerd/server/workerd-api.c++
garrettgu10: I got it!

src/workerd/server/workerd-api.c++
```diff
@@ -446,7 +537,23 @@ void WorkerdApi::compileModules(
     if (hasPythonModules(confModules)) {
       KJ_REQUIRE(featureFlags.getPythonWorkers(),
           "The python_workers compatibility flag is required to use Python.");
-      // Inject pyodide bootstrap module.
+      // Inject Pyodide bundle
+      if(util::Autogate::isEnabled(util::AutogateKey::PYODIDE_LOAD_EXTERNAL)) {
+          if (fetchPyodideBundle(impl->pythonConfig, "dev"_kj)) {
+            modules->addBuiltinBundle(getPyodideBundle("dev"_kj), kj::none);
+          } else {
```
hoodmane, line 544: We'll want to select between `"dev"` and the appropriate version by setting an appropriate compat flag so that tests can be a bit more declarative. I think it would be confusing if you try to test a specific version but the test uses latest because it exists. But we can deal with that in a followup.

src/workerd/server/workerd-api.c++
```diff
@@ -432,6 +436,93 @@ kj::Maybe<jsg::ModuleRegistry::ModuleInfo> WorkerdApi::tryCompileModule(
...
+  KJ_IF_SOME(dir, maybeDir) {
+    kj::Path filename = getPyodideBundleFileName(version);
+    auto replacer = dir->replaceFile(filename, kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+    replacer->get().writeAll(bytes);
+    replacer->commit();
+  }
+}
+
+bool fetchPyodideBundle(const api::pyodide::PythonConfig& pyConfig, kj::StringPtr version) {
```
hoodmane, line 468: I think it would be tidier to have `fetchPyodideBundle` return a `kj::Maybe<>` directly, but we can tidy up that sort of thing in followup PRs.

hoodmane (approved):
Thanks @garrettgu10! Let's merge this.

PR #2455, opened 2024-07-29
Python: Adjust API for Pyodide bundle to support multiple versions

hoodmane:
A relatively minor change that cleans up the bundle code in pyodide.c++/pyodide.h and adds a TreeMap to handle multiple bundle versions.


garrettgu10:
Can we test this downstream?

PR #2462, opened 2024-07-30
add buffer.transcode for nodejs_compat

anonrig:
> work in progress. opening pull-request to receive some initial feedback


src/workerd/api/node/buffer.c++
```diff
@@ -876,5 +852,17 @@ bool BufferUtil::isUtf8(kj::Array<kj::byte> buffer) {
...
 }
 
+kj::Array<kj::byte> BufferUtil::transcode(kj::Array<kj::byte> source, kj::String rawFromEncoding, kj::String rawToEncoding) {
+  auto fromEncoding = getEncoding(rawFromEncoding.asPtr());
+  auto toEncoding = getEncoding(rawToEncoding.asPtr());
+
+  // TODO: Throw appropriate error here.
+  if (!i18n::canBeTranscoded(fromEncoding) || !i18n::canBeTranscoded(toEncoding)) {
+    JSG_FAIL_REQUIRE(Error, "Unable to transcode Buffer");
+  }
```
jasnell, line 862:
```suggestion
  JSG_REQUIRE(i18n::canBeTranscoded(fromEncoding) &&
              i18n::canBeTranscoded(toEncoding), Error,
              "Unable to transcode Buffer");
```

src/workerd/api/node/buffer.c++
```diff
@@ -876,5 +852,17 @@ bool BufferUtil::isUtf8(kj::Array<kj::byte> buffer) {
...
+
+  return i18n::transcode(source.asPtr(), fromEncoding, toEncoding);
```
jasnell, line 864: If the `i18n::transcode` method here accepts a `kj::ArrayPtr<...>` as the first argument, you should be able to just pass `source` here... the `asPtr()` should be unnecessary... at least I think so :-)

src/workerd/api/node/i18n.h
```diff
@@ -0,0 +1,127 @@
...
+    case Encoding::ASCII:
+    case Encoding::LATIN1:
+    case Encoding::UCS2:
+    case Encoding::UTF8: return true;
+    default: return false;
+  }
+}
+
+
+
```
jasnell, line 81:
```suggestion

```

src/workerd/api/node/buffer.c++
anonrig: nice! definitely works.

src/workerd/api/node/i18n.h
```diff
@@ -0,0 +1,127 @@
...
+class Converter final {
+ public:
+  explicit Converter(Encoding encoding, std::string_view substitude = "");
+  explicit Converter(UConverter* converter, std::string_view substitude = "");
```
jasnell, line 85:
```suggestion
  explicit Converter(Encoding encoding, std::string_view sub = "");
  explicit Converter(UConverter* converter, std::string_view sub = "");
```

We don't typically use `std::string_view`... I'm fine with it but using `kj::StringPtr` would be more consistent with the rest of our codebase

src/workerd/api/node/i18n.h
```diff
@@ -0,0 +1,121 @@
...
+      KJ_ASSERT(U_SUCCESS(status));
+    }
+  }
+
+ private:
+  struct ConverterDisposer {
+    void operator()(UConverter* pointer) const { ucnv_close(pointer); }
+  };
+
+  std::unique_ptr<UConverter, ConverterDisposer> conv_;
```
jasnell, line 112: While I don't have any problems with the use of `std::unique_ptr` here, using `kj::Own` is more consistent with the rest of the codebase

src/workerd/api/node/i18n.h
anonrig: Makes sense. Updating now.

src/workerd/api/node/i18n.h
```diff
@@ -0,0 +1,121 @@
...
+};
+
+using TranscodeResult = kj::OneOf<kj::Array<kj::byte>, UErrorCode>;
```
jasnell, line 115: nit: not being used?

jasnell: Oh, I see it's used in the i18n.c++ file... I would move the definition into that file if it is only used there

src/workerd/api/node/i18n.h
```diff
@@ -0,0 +1,121 @@
...
+    KJ_ASSERT_NONNULL(conv_.get());
+    return ucnv_getMinCharSize(conv_.get());
+  }
+
+  void reset() {
+    KJ_ASSERT_NONNULL(conv_.get());
+    ucnv_reset(conv_.get());
+  }
+
+  void setSubstitudeChars(std::string_view sub) {
```
jasnell, line 98: nit: can we move these into the i18n.c++ file. I'd like to minimize the ICU imports necessary to bring into this header file.

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,158 @@
...
+  auto limit = source.size() + to.maxSize();
+  KJ_STACK_ARRAY(kj::byte, out, limit, 0, limit);
+  char* target = out.asChars().begin();
+  const char* source_ = source.asChars().begin();
+  UErrorCode status{};
+  ucnv_convertEx(to.conv(), from.conv(), &target, target + limit, &source_, source_ + source.size(),
+                 nullptr, nullptr, nullptr, nullptr, true, true, &status);
+
+  if (U_SUCCESS(status)) {
+    return out.slice(0, target - out.asChars().begin()).attach();
```
jasnell, line 52: This is going to lead to a UAF, I believe. The `out` here is stack allocated and you're returning an `ArrayPtr` from it. I don't believe the `attach()` call here causes it to be copied at all.

I think what you want here is `auto out = kj::heapArray<kj::byte>(limit)` instead of `KJ_STACK_ARRAY` and then `return out.slice(0, target - out.begin()).attach(kj::mv(out))`

src/workerd/api/node/buffer.h
```diff
@@ -94,6 +97,7 @@ class BufferUtil final: public jsg::Object {
     JSG_METHOD(write);
     JSG_METHOD(isAscii);
     JSG_METHOD(isUtf8);
+    JSG_METHOD(trancode);
```
jasnell, line 100:
```suggestion
    JSG_METHOD(transcode);
```

src/workerd/api/node/buffer.c++
```diff
@@ -85,30 +88,20 @@ void SwapBytes(kj::ArrayPtr<kj::byte> bytes) {
...
-  } else if (encoding == "base64") {
+  } else if (input == "base64"_kj) {
     return Encoding::BASE64;
-  } else if (encoding == "base64url") {
+  } else if (input == "base64url"_kj) {
     return Encoding::BASE64URL;
-  } else if (encoding == "hex") {
+  } else if (input == "hex"_kj) {
     return Encoding::HEX;
   }
```
jasnell, line 107: This isn't new in this PR but I've been thinking that we should change this `KJ_UNREACHABLE` into a proper `JSG_FAIL_REQUIRE` error instead. This value should be checked on the JS side before we get here but if it isn't, then it will lead to a crash.

So instead of `KJ_UNREACHABLE` at the end of this let's make it ...

```
JSG_FAIL_REQUIRE(Error, kj::str("Invalid encoding: ", input));
```

Or something similar.

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,205 @@
...
+  }
+}
+
+typedef kj::Maybe<kj::Array<kj::byte>> (*TranscodeImpl)(kj::ArrayPtr<kj::byte> source,
+                                                        Encoding fromEncoding, Encoding toEncoding);
+
+kj::Maybe<kj::Array<kj::byte>> TranscodeDefault(kj::ArrayPtr<kj::byte> source,
+                                                Encoding fromEncoding, Encoding toEncoding) {
+  Converter to(toEncoding);
+  std::string substitude(to.minSize(), '?');
```
jasnell, line 62: nit: While I don't mind the use of `std::string` here, for consistency here we're typically use `kj::String` here.

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,205 @@
...
+  to.setSubstitudeChars(substitude);
+  Converter from(fromEncoding);
+
+  auto limit = source.size() + to.maxSize();
+  auto out = kj::heapArray<kj::byte>(limit);
```
jasnell, line 67: I wonder if we shouldn't cap this at a maximum size just to be safe. I know that Node.js does not but....

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,205 @@
...
+  UErrorCode status{};
+  const size_t length_in_chars = source.size() * sizeof(UChar);
+  Converter from(fromEncoding);
+  auto out = kj::heapArray<UChar>(source.size());
+  const auto source_ = source.asChars().begin();
+  ucnv_toUChars(from.conv(), out.begin(), length_in_chars, source_, source.size(), &status);
+  if (U_SUCCESS(status)) {
+    return out.asBytes().attach(kj::mv(out));
+  }
+  return kj::none;
```
jasnell, line 91:
```suggestion

  return kj::none;
```

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,205 @@
...
+kj::Maybe<kj::Array<kj::byte>> TranscodeFromUCS2(kj::ArrayPtr<kj::byte> source,
+                                                 Encoding fromEncoding, Encoding toEncoding) {
+  UErrorCode status{};
+  KJ_STACK_ARRAY(kj::byte, sourcebuf, 1024, 0, 1024);
+  Converter to(toEncoding);
+  std::string substitude(to.minSize(), '?');
+  to.setSubstitudeChars(substitude);
+
+  const size_t length_in_chars = source.size() * sizeof(UChar);
+  sourcebuf.copyFrom(source.slice(0, length_in_chars));
```
jasnell, line 103: Is this additional copy really necessary here?

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,205 @@
...
+}
+
+} // namespace
+
+Converter::Converter(Encoding encoding, kj::StringPtr substitude) {
+  UErrorCode status = U_ZERO_ERROR;
+  auto name = getEncodingName(encoding);
+  auto conv = ucnv_open(name, &status);
+  KJ_ASSERT(U_SUCCESS(status));
+  conv_ = kj::Own<UConverter>(conv, ConverterDisposer::INSTANCE);
```
jasnell, line 134: kj has a `kj::disposeWith(...)` API that makes this a bit easier. Do a search for it in the code to see examples.

src/workerd/api/node/i18n.h
```diff
@@ -0,0 +1,65 @@
...
+  ASCII,
+  LATIN1,
+  UTF8,
+  UTF16LE,
+  BASE64,
+  BASE64URL,
+  HEX,
+
+  // Required by transcode support.
+  UCS2,
```
jasnell, line 25: nit: might expand to comment a little to indicate this is equivalent to UTF16LE

src/workerd/api/node/i18n.h
```diff
@@ -0,0 +1,65 @@
...
+    return true;
+  default:
+    return false;
+  }
+}
+
+class Converter final {
+public:
+  explicit Converter(Encoding encoding, kj::StringPtr substitude = ""_kj);
+  explicit Converter(UConverter* converter, kj::StringPtr substitude = ""_kj);
```
jasnell, line 46:
```suggestion
  explicit Converter(UConverter* converter, kj::StringPtr substitude = ""_kj);
  KJ_DISALLOW_COPY_AND_MOVE(Converter);
```

src/workerd/api/node/util.c++
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 #include "util.h"
+#include <kj/string.h>
```
jasnell, line 5: nit: unnecessary to add here?

src/node/internal/internal_buffer.ts
```diff
@@ -2294,6 +2294,18 @@ export function isUtf8(value: ArrayBufferView) {
...
+export function transcode(source: ArrayBufferView, fromEncoding: string, toEncoding: string) {
+  const normalizedFromEncoding = normalizeEncoding(fromEncoding);
+  if (!Buffer.isEncoding(normalizedFromEncoding)) {
+    throw new ERR_UNKNOWN_ENCODING(fromEncoding);
+  }
+  const normalizedToEncoding = normalizeEncoding(toEncoding);
+  if (!Buffer.isEncoding(normalizedToEncoding)) {
+    throw new ERR_UNKNOWN_ENCODING(toEncoding);
+  }
+  return bufferUtil.transcode(source, fromEncoding, toEncoding);
```
jasnell, line 2306:
```suggestion
  return bufferUtil.transcode(source, normalizedFromEncoding, normalizedToEncoding);
```

?

jasnell: Although, I do wonder if we should normalize these to int values matching the internal enum rather than passing these through as strings. Passing them down as strings means we will always be additionally copying the values and performing an unnecessary allocation and utf8 write in order to get those as kj::String internally. If we know they are valid encodings at this point, using the enum would be more efficient.

src/workerd/api/node/buffer.c++
anonrig: I agree that an error is better than crashing!

I'm fine with updating this in here. And additional PR for this small change will not benefit anyone. I'm updating it now. Thanks for the review.

src/workerd/api/node/i18n.c++
anonrig: kj::String does not have a constructor of `(size_t, char)`

src/node/internal/internal_buffer.ts
anonrig: Yup reducing the string serialization cost would be beneficial. Is there an easy way to share ENUM values with JS without duplication?

jasnell: Look for `JSG_STATIC_CONSTANT` or `JSG_STATIC_CONSTANT_NAMED` ... I think that's what you might want here.

src/workerd/api/node/i18n.c++
jasnell: Does `kj::repeat(...)` get you what you want (bit more verbose)

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,253 @@
...
+  size_t expected_utf16_length =
+      simdutf::utf16_length_from_utf8(source.asChars().begin(), source.size());
+  // Workers are limited to 128MB so this isn't actually a realistic concern, but sanity check.
+  JSG_REQUIRE(expected_utf16_length <= 134217728, Error,
+              "Expected UTF-16le length is too large to transcode");
+  auto destbuf = kj::heapArray<UChar>(expected_utf16_length);
+
+  size_t actual_length =
+      simdutf::convert_utf8_to_utf16le(source.asChars().begin(), source.size(), destbuf.begin());
+  KJ_ASSERT(actual_length == expected_utf16_length, "Expected UTF16 length mismatch");
```
jasnell, line 139: To be safest, convert these `KJ_ASSERT` uses into `JSG_REQUIRE` so that we get an actual JS error thrown and not the dreaded "Internal error"

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,253 @@
...
+  return ucnv_getMinCharSize(conv_.get());
+}
+
+void Converter::reset() {
+  KJ_ASSERT_NONNULL(conv_.get());
+  ucnv_reset(conv_.get());
+}
+
+void Converter::setSubstitudeChars(kj::StringPtr sub) {
+  KJ_ASSERT_NONNULL(conv_.get());
```
jasnell, line 211: Per the ICU API, setting the subst chars to zero-length input will fail. I think with Node.js we always use `?` as the default if I am remembering correctly, so maybe the default should be `?` here too instead of zero-length? Otherwise, check for zero-length here and skip the call to setSubstChars if there's no input.

src/node/internal/internal_buffer.ts
```diff
@@ -2294,6 +2294,21 @@ export function isUtf8(value: ArrayBufferView) {
...
+  }
+  const normalizedFromEncoding = normalizeEncoding(fromEncoding);
+  if (!Buffer.isEncoding(normalizedFromEncoding)) {
+    throw new ERR_UNKNOWN_ENCODING(fromEncoding);
+  }
+  const normalizedToEncoding = normalizeEncoding(toEncoding);
+  if (!Buffer.isEncoding(normalizedToEncoding)) {
+    throw new ERR_UNKNOWN_ENCODING(toEncoding);
+  }
+  return bufferUtil.transcode(source, normalizedFromEncoding, normalizedToEncoding);
```
jasnell, line 2309: The return value of `bufferUtil.transcode(...)` here is going to be an `ArrayBuffer`. This needs to be wrapped in a `Buffer`. So you'll want something like `return Buffer.from(bufferUtil.transcode(...` instead

src/workerd/api/node/i18n.c++
jasnell: Actually, I think it's any input `< ucnv_getMinCharSize(conv_.get())` that will fail

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,253 @@
...
+
+  return kj::none;
+}
+
+kj::Maybe<kj::Array<kj::byte>> TranscodeUcs2FromUtf8(kj::ArrayPtr<kj::byte> source,
+                                                     Encoding fromEncoding, Encoding toEncoding) {
+  size_t expected_utf16_length =
+      simdutf::utf16_length_from_utf8(source.asChars().begin(), source.size());
+  // Workers are limited to 128MB so this isn't actually a realistic concern, but sanity check.
+  JSG_REQUIRE(expected_utf16_length <= 134217728, Error,
```
jasnell, line 133: nit: let's move the `134217728` to a constant instead of duplicating it multiple times in different places

src/node/internal/internal_buffer.ts
anonrig: I added a comment for this. I'll follow-up after this pull-request lands.

justfile
```diff
@@ -1,3 +1,9 @@
+default:
```
jasnell, line 1: nit: this change should probably be in a separate commit

src/node/internal/internal_buffer.ts
```diff
@@ -2294,6 +2294,22 @@ export function isUtf8(value: ArrayBufferView) {
...
+  }
+  const normalizedFromEncoding = normalizeEncoding(fromEncoding);
+  if (!Buffer.isEncoding(normalizedFromEncoding)) {
+    throw new ERR_UNKNOWN_ENCODING(fromEncoding);
+  }
+  const normalizedToEncoding = normalizeEncoding(toEncoding);
+  if (!Buffer.isEncoding(normalizedToEncoding)) {
+    throw new ERR_UNKNOWN_ENCODING(toEncoding);
+  }
+  // Optimization opportunity: Pass int encoding values instead of strings.
```
jasnell, line 2309:
```suggestion
  // TODO(soon): Optimization opportunity: Pass int encoding values instead of strings.
```

src/workerd/api/node/tests/buffer-nodejs-test.js
```diff
@@ -5705,3 +5706,78 @@ export const isUtf8Test = {
...
+    // Invalid arguments should fail
+    throws(() => transcode(null, 'utf8', 'ascii'));
+    throws(() => transcode(Buffer.from('a'), 'b', 'utf8'));
+    throws(() => transcode(Buffer.from('a'), 'uf8', 'b'));
+
+    // Throws error for buffer bigger than 128mb.
+    throws(
+      () => {
+        const MAX_LENGTH = 134217728;
+        transcode(Buffer.from('a'.repeat(MAX_LENGTH), 'utf8', 'utf16le'));
```
jasnell, line 5779: nit: since the check is duplicated in several places in i18n, let's include variants here that are sure to test each case

src/workerd/api/node/util.h
```diff
@@ -245,4 +245,7 @@ class UtilModule final: public jsg::Object {
     api::node::MIMEParams::ValueIterator::Next, \
     api::node::MIMEParams::KeyIterator::Next
 
+// An isolate has a 128mb memory limit.
+const int ISOLATE_LIMIT = 134217728;
```
jasnell, line 249: just put this in i18n.c++ where it is used

jasnell (approved):
Few nits but otherwise LGTM

jasnell (approved):

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+                                                        Encoding fromEncoding, Encoding toEncoding);
+
+kj::Maybe<kj::Array<kj::byte>> TranscodeDefault(kj::ArrayPtr<kj::byte> source,
+                                                Encoding fromEncoding, Encoding toEncoding) {
+  Converter to(toEncoding);
+  auto substitude = kj::str(kj::repeat('?', to.minCharSize()));
+  to.setSubstitudeChars(substitude);
+  Converter from(fromEncoding);
+
+  size_t limit = source.size() + to.maxCharSize();
```
danlapid, line 71: Isn't this supposed to be multiplication?

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+  }
+}
+
+typedef kj::Maybe<kj::Array<kj::byte>> (*TranscodeImpl)(kj::ArrayPtr<kj::byte> source,
+                                                        Encoding fromEncoding, Encoding toEncoding);
+
+kj::Maybe<kj::Array<kj::byte>> TranscodeDefault(kj::ArrayPtr<kj::byte> source,
+                                                Encoding fromEncoding, Encoding toEncoding) {
+  Converter to(toEncoding);
+  auto substitude = kj::str(kj::repeat('?', to.minCharSize()));
```
danlapid, line 67: nit: typo substitude in place of substitute throughout the file

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  return kj::none;
+}
+
+kj::Maybe<kj::Array<kj::byte>> TranscodeToUCS2(kj::ArrayPtr<kj::byte> source, Encoding fromEncoding,
+                                               Encoding toEncoding) {
+  Converter from(fromEncoding);
+  // Workers are limited to 128MB so this isn't actually a realistic concern, but sanity check.
+  JSG_REQUIRE(source.size() <= ISOLATE_LIMIT, Error, "Source buffer is too large to transcode");
```
danlapid, line 91: Wouldn't the destination buffer be twice the size of the source buffer?
Latin 1 is 8bit encoding while utf16 is 16bit encoding

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+  ucnv_convertEx(to.conv(), from.conv(), &target, target + limit, &source_, source_ + source.size(),
+                 nullptr, nullptr, nullptr, nullptr, true, true, &status);
+  if (U_SUCCESS(status)) {
+    return out.slice(0, target - out.asChars().begin()).attach(kj::mv(out));
+  }
+
+  return kj::none;
+}
+
+kj::Maybe<kj::Array<kj::byte>> TranscodeToUCS2(kj::ArrayPtr<kj::byte> source, Encoding fromEncoding,
```
danlapid, line 87: If this function only transcodes from latin1 I think it's better to rename it and remove the fromencoding arg

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+    break;
+  case Encoding::UTF8:
+    if (toEncoding == Encoding::UTF16LE) {
+      transcode_function = &TranscodeUcs2FromUtf8;
+    }
+    break;
+  case Encoding::UTF16LE:
+    switch (toEncoding) {
+    case Encoding::UTF16LE:
+      transcode_function = &TranscodeDefault;
```
danlapid, line 239: Can't we optimize the case of fromEncoding == toEncoding to just be return copy of src?

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+  auto substitude = kj::str(kj::repeat('?', to.minCharSize()));
+  to.setSubstitudeChars(substitude);
+
+  const size_t length_in_chars = source.size() * sizeof(UChar);
+  // Workers are limited to 128MB so this isn't actually a realistic concern, but sanity check.
+  JSG_REQUIRE(length_in_chars <= ISOLATE_LIMIT, Error, "Buffer is too large to transcode");
+
+  // TODO: Optimization opportunity: Remove sourcebuf allocation by calculating length beforehand.
+  KJ_STACK_ARRAY(UChar, sourcebuf, source.size(), 1024, 536870888);
+  memcpy(sourcebuf.begin(), sourcebuf.begin(), length_in_chars);
```
danlapid, line 116: Something doesn't quite make sense here.
1. You're copying from sourcebuf.begin() to sourcebuf.begin() which is a nullop?
2. You have allocated sourcebuf to be source.size() but because of the error in calculation of length_in_chars you are copying twice that amount from sourcebuf onto itself thus both reading out of bounds and writing out of bounds

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+
+  auto destbuf = kj::heapArray<kj::byte>(length_in_chars);
```
danlapid, line 118: dstSize here should be `length_in_chars * to.maxCharSize()`

src/workerd/api/node/i18n.c++
```diff
@@ -0,0 +1,258 @@
...
+Converter::Converter(Encoding encoding, kj::StringPtr substitude) {
+  UErrorCode status = U_ZERO_ERROR;
+  auto name = getEncodingName(encoding);
+  auto conv = ucnv_open(name, &status);
+  JSG_REQUIRE(U_SUCCESS(status), Error, "Failed to initialize converter");
+  conv_ = kj::Own<UConverter>(conv, ConverterDisposer::INSTANCE);
+  setSubstitudeChars(substitude);
+}
+
+Converter::Converter(UConverter* converter, kj::StringPtr substitude)
```
danlapid, line 188: Is this constructor required? 
If so, can we not assume converter would need to be NONNULL and move the `KJ_ASSERT_NONNULL(conv_.get());` check from all the methods to this constructor so that we check that at creation rather than on every method call?

src/workerd/api/node/i18n.c++
jasnell: ha, good catch. My eyes completely overlooked that one

src/workerd/api/node/i18n.c++
anonrig: Yeah, you're correct. Updated it now.

src/workerd/api/node/i18n.c++
anonrig: Yeah, nice catch. Updating the implementation.

src/workerd/api/node/i18n.c++
anonrig: Yeah, this is indeed true, and matches the current implementation of Node.js. Updating it now.

src/workerd/api/node/i18n.c++
anonrig: Actually it is not needed.

src/workerd/api/node/i18n.c++
anonrig: yeah, we don't need the sourcebuf at all. I've updated the implementation.

danlapid (approved):

jasnell:
Adding the "needs-internal-pr" label on this to ensure that we make certain that the internal CI run is green before this lands.

jasnell (approved):

anonrig:
Internal PR is now resolved. We can merge this now.

PR #2464, opened 2024-07-30
Adds pythonRelease annotation to compat flags.

dom96:
This new annotation will be used to figure out the Python release that a Python Worker belongs to. The main use case for this is for figuring out memory snapshot compatibility between workers.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -15,6 +15,30 @@ struct ImpliedByAfterDate @0x8f8c1b68151b6cff {
   date @1 :Text;
 }
 
+struct PythonSnapshotRelease @0x89c66fb883cb6975 {
+  # Used to indicate a specific Python release that introduces a change which likely breaks
+  # existing memory snapshots.
+  #
+  # The versions/dates specified here are used to generate a filename for the package memory
+  # snapshots created by the validator. They are also used to generate a filename of the Pyodide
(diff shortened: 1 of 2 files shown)


garrettgu10 (approved):

src/workerd/io/compatibility-date.capnp
```diff
@@ -69,6 +69,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
 
   annotation impliedByAfterDate @0xe3e5a63e76284d89 (field) :ImpliedByAfterDate;
 
+  annotation pythonRelease @0xef74c0cc5d18cc0c (field) :Text;
+  # A tag to tell us which compat flags represent a Python Workers release and thus need to be
+  # taken into account for memory snapshot compatibility.
```
hoodmane, line 74: For memory snapshot compatibility we need much more than this. I think we should hash `pyodide.asm.wasm` and `pyodide.asm.js` and key on that.

dom96: Won't these files only change across compat flags?

hoodmane: Yeah. The Python version will also only change across compat flags though.

dom96: Renamed the annotation to `pythonSnapshotRelease` and reworded the description to make things clearer.

hoodmane (approved):
Thanks @dom96!

PR #2466, opened 2024-07-31
Refactor http-test to make subsequent edits cleaner

AdityaAtulTewari:


src/workerd/api/http-test.js
```diff
@@ -206,44 +206,56 @@ export const inspect = {
   }
 };
 
+async function requestCacheThrowsError(cacheHeader, errorName, errorMessage){
```
mikea, line 209: I suggest naming them all `assert*` since these are, essentially assertions.

Or `verify*` if you think they will be complicated.

mikea, line 209: need to format new code: `){`

src/workerd/api/http-test.js
```diff
@@ -206,44 +206,56 @@ export const inspect = {
...
+      name: errorName,
+      message: errorMessage,
+    }
+  );
+}
+
+async function requestCacheThrowsDefaultError(cacheHeader) {
+  requestCacheThrowsError(
+    cacheHeader,
+    'Error',
```
mikea, line 222: could these be default argument values instead?

mikea (changes requested):

src/workerd/api/http-test.js
AdityaAtulTewari: Yes.

mikea (approved):

jasnell (approved):

PR #2467, opened 2024-07-31
Clear WritableStreamSink queue on abort with compat flag

jasnell:
The motivation here is to address an issue with `WritableStream` instances wrapping a `WritableStreamSink` implementation not processing `abort()` calls proactively. Without the flag the `abort()` will be handled lazily, meaning that a consumer must be actively consuming the data written to the stream in order to trigger the abort and close the stream. This is problematic if the consumer has dropped or otherwise stopped reading from the writable's internal buffer ... the pending abort will never be processed and resolved. This means that pending writes and the pending abort itself will never be resolved in those cases. With the flag, the queue of pending writes is drained immediately when `abort()` is called and the stream is transitioned to an error state immediately. This should cause the underlying `WritableStreamSink` to be aborted and dropped immediately and the pending write and abort promises should all get settled.

internal ticket: EW-8607

### DIFF
--- a/src/workerd/api/streams/internal.c++
+++ b/src/workerd/api/streams/internal.c++
@@ -878,6 +878,11 @@ void ReadableStreamInternalController::releaseReader(
   }
 }
 
+void WritableStreamInternalController::Writable::abort(kj::Exception&& ex) {
+  canceler.cancel(kj::cp(ex));
+  sink->abort(kj::mv(ex));
+}
+
 WritableStreamInternalController::~WritableStreamInternalController() noexcept(false) {
   if (writeState.is<WriterLocked>()) {
     writeState.init<Unlocked>();
@@ -912,7 +917,7 @@ jsg::Promise<void> WritableStreamInternalController::write(
     KJ_CASE_ONEOF(errored, StreamStates::Errored) {
       return js.rejectedPromise<void>(errored.addRef(js));
     }
-    KJ_CASE_ONEOF(writable, Writable) {
+    KJ_CASE_ONEOF(writable, IoOwn<Writable>) {
       if (value == kj::none) {
         return js.resolvedPromise();
       }
@@ -1052,7 +1057,7 @@ jsg::Promise<void> WritableStreamInternalController::closeImpl(jsg::Lock& js, bo
       auto reason = errored.getHandle(js);
       return rejectedMaybeHandledPromise<void>(js, reason, markAsHandled);
     }
-    KJ_CASE_ONEOF(writable, Writable) {
+    KJ_CASE_ONEOF(writable, IoOwn<Writable>) {
       auto prp = js.newPromiseAndResolver<void>();
       if (markAsHandled) {
         prp.promise.markAsHandled(js);
@@ -1115,7 +1120,7 @@ jsg::Promise<void> WritableStreamInternalController::flush(
       auto reason = errored.getHandle(js);
       return rejectedMaybeHandledPromise<void>(js, reason, markAsHandled);
     }
-    KJ_CASE_ONEOF(writable, Writable) {
+    KJ_CASE_ONEOF(writable, IoOwn<Writable>) {
       auto prp = js.newPromiseAndResolver<void>();
       if (markAsHandled) {
         prp.promise.markAsHandled(js);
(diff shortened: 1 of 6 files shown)


jasnell:
@anonrig ... just fyi on this PR. This is a good simple example of the introduction of a compat flag to preserve backwards compat behavior... just for illustration.

src/workerd/api/streams/common.h
```diff
@@ -200,7 +200,14 @@ class WritableStreamSink {
   virtual kj::Maybe<kj::Promise<DeferredProxy<void>>> tryPumpFrom(
       ReadableStreamSource& input, bool end);
 
-  virtual void abort(kj::Exception reason) = 0;
+  enum class AbortOption {
+    NONE,
+    // Instructs the sink to reject pending writes when aborting. This is used when
+    // the InternalWritableStreamAbortClearsQueue flag is set.
+    DRAIN,
```
kentonv, line 207: Hmm, I usually associate the word "drain" with "gracefully complete all pending requests before shutting down", but here it seems to have the opposite meaning?

Maybe `FINISH_QUEUED_WRITES` and `CANCEL_QUEUED_WRITES` would be clearer names?

src/workerd/api/streams/internal.c++
```diff
@@ -1162,6 +1162,20 @@ jsg::Promise<void> WritableStreamInternalController::doAbort(
...
   KJ_IF_SOME(writable, state.tryGet<Writable>()) {
     auto exception = js.exceptionToKj(js.v8Ref(reason));
+
+    if (FeatureFlags::get(js).getInternalWritableStreamAbortClearsQueue()) {
+      // If this flag is set, we will clear the queue proactively and immediately
+      // error the stream rather than handling the abort lazily. In this case, the
+      // stream will be put into an errored state immediately after draining the
+      // queue. All pending writes and other operations in the queue will be rejected
+      // immediately and an immediately resolved or rejected promise will be returned.
+      writable->abort(kj::cp(exception), WritableStreamSink::AbortOption::DRAIN);
```
kentonv, line 1172: If there is currently a `write()` call in progress (that is, the `Promise` returned by a previous `write()` has not resolved yet), then this is UB, at least in the case of `EncodedAsyncOutputStream::abort()` which doesn't make any attempt to cancel that promise before deleting the underlying KJ stream.

One option is to change `WritableStreamSink`'s contract to say it is responsible for canceling writes on abort(). Another option would be to have `WritableStreamInternalController` itself be responsible for wrapping each write promise in a `kj::Canceler` which would be canceled here, before calling `abort()` on the sink. The latter seems better to me since it's handled in one place rather than every sink implementation.

Note that similarly, any `pumpTo()` needs to be canceled before aborting.

jasnell: Yep, was thinking along the same lines... I agree a kj::Canceler here would be a good approach.

jasnell: Ok, updated but needs a careful review to make sure the new behavior is correct. It does not appear that we have any tests that verify the edge cases on this. I'll be looking as seeing if we can reasonably expand the test coverage here. But, in the meantime please take a look

anonrig (approved):

jasnell:
Thank you for the review @anonrig ... going to wait for @kentonv to give another review before merging this.

src/workerd/api/streams/internal.c++
```diff
@@ -1565,9 +1585,10 @@ jsg::Promise<void> WritableStreamInternalController::writeLoopAfterFrontOutputLo
   };
 
   const auto maybeAbort = [this](jsg::Lock& js, auto& request) -> bool {
-    auto& writable = state.get<Writable>();
+    auto& writable = state.get<IoOwn<Writable>>();
```
kentonv, line 1588: Not new in this PR, but should we use `KJ_ASSERT_NONNULL(state.tryGet` here? `get` is UB if for whatever reason the state is a different type. IMO it should only ever be used if the type was checked literally right before.

src/workerd/api/streams/internal.c++
```diff
@@ -2333,14 +2363,24 @@ void IdentityTransformStreamImpl::abort(kj::Exception reason) {
       request.fulfiller->reject(kj::cp(reason));
     }
     KJ_CASE_ONEOF(request, WriteRequest) {
-      KJ_FAIL_ASSERT("abort() is supposed to wait for any pending write() to finish");
+      switch (option) {
+        case AbortOption::NONE: {
+          KJ_FAIL_ASSERT("abort() is supposed to wait for any pending write() to finish");
+        }
+        case AbortOption::ABORT_QUEUED_WRITES: {
+          request.fulfiller->reject(kj::cp(reason));
```
kentonv, line 2371: Now that all the write requests are canceled first, is this needed?

Maybe we no longer need to pass `AbortOption` down to the `WritableStreamSink`?

src/workerd/api/system-streams.c++
```diff
@@ -299,7 +299,7 @@ kj::Promise<void> EncodedAsyncOutputStream::end() {
   return promise.attach(ioContext.registerPendingEvent());
 }
 
-void EncodedAsyncOutputStream::abort(kj::Exception reason) {
+void EncodedAsyncOutputStream::abort(kj::Exception reason, AbortOption ignored) {
   inner.init<Ended>();
```
kentonv, line 303: Interesting to note that for compressed streams, this implementation has always failed to flush the buffer before aborting, so the compat flag doesn't make a whole lot of difference (either way, previously-written data may be lost).

Makes me wonder if the compat flag is even needed.

kentonv:
Sorry for delay, I had started reviewing this a few days ago and then got distracted away.

src/workerd/api/system-streams.c++
jasnell: Yeah that's a good point... hmmm

jasnell: I think the compat flag is still likely necessary. There's an observable difference in behavior around the rejected queued write promises here that someone could be currently working around ( at least in theory )...likely safest to have the flag

jasnell: but I think you're right about not needing the `AbortOption` enum class now.

kentonv (approved):

PR #2473, opened 2024-08-02
VS-257: Fix Vectorize query options

garvit-gupta:
This change fixes 2 aspects of Vectorize query options.

1. Options itself should be an optional field.

2. Options.returnMetadata should be of type boolean for Vectorize V1 and of type `VectorizeMetadataRetrievalLevel` for Vectorize V2. The existing type relied on a generic to assign type appropriately for the operation, but did not work as expected for Vectorize V2. It continued to expect a boolean field for V2 as well instead of `VectorizeMetadataRetrievalLevel`.

Follows: https://github.com/cloudflare/workerd/pull/2443

### DIFF
--- a/src/cloudflare/internal/test/vectorize/vectorize-mock.js
+++ b/src/cloudflare/internal/test/vectorize/vectorize-mock.js
@@ -97,7 +97,7 @@ export default {
       ) {
         return Response.json({});
       } else if (request.method === "POST" && pathname.endsWith("/query")) {
-        /** @type {VectorizeQueryOptions<VectorizeMetadataRetrievalLevel> & {vector: number[]}} */
+        /** @type {VectorizeQueryOptions & {vector: number[]}} */
         const body = await request.json();
         let returnSet = structuredClone(exampleVectorMatches);
         if (
@@ -114,7 +114,7 @@ export default {
           returnSet.forEach((v) => {
             delete v.values;
           });
-        if (!body?.returnMetadata)
+        if (!body?.returnMetadata || body?.returnMetadata === "none")
           returnSet.forEach((v) => {
             delete v.metadata;
           });
--- a/src/cloudflare/internal/vectorize-api.ts
+++ b/src/cloudflare/internal/vectorize-api.ts
@@ -42,9 +42,14 @@ class VectorizeIndexImpl implements Vectorize {
 
   public async query(
     vector: VectorFloatArray | number[],
-    options: VectorizeQueryOptions<VectorizeMetadataRetrievalLevel>
+    options?: VectorizeQueryOptions
   ): Promise<VectorizeMatches> {
     if (this.indexVersion === "v2") {
+      if (options && options.returnMetadata && !isVectorizeMetadataRetrievalLevel(options.returnMetadata) ) {
+        throw new Error(
+          `Invalid returnMetadata option. Expected: "none", "indexed" or "all"; got: ${options.returnMetadata}`
+        );
+      }
       const res = await this._send(Operation.VECTOR_QUERY, `query`, {
         method: "POST",
         body: JSON.stringify({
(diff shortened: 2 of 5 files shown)


types/defines/vectorize.d.ts
```diff
@@ -59,13 +59,11 @@ type VectorizeDistanceMetric = "euclidean" | "cosine" | "dot-product";
  */
 type VectorizeMetadataRetrievalLevel = "all" | "indexed" | "none";
 
-interface VectorizeQueryOptions<
-  MetadataReturn extends boolean | VectorizeMetadataRetrievalLevel = boolean,
-> {
```
penalosa, line 64: This is a breaking change to the types—is it absolutely needed?

RamIdeas: Agree with @penalosa this is a breaking change. If the problem is the `returnMetadata` field "continued to expect a boolean field for V2 as well instead of VectorizeMetadataRetrievalLevel" can we help you make that work rather than remove the generic entirely?

garvit-gupta: This change should have no effect on the way users would interact with Vectorize V1 using worker bindings. So I am not sure if I would term this as a breaking change.

That being said, I would be happy to understand if there could be any alternative options that could be suitable for our purpose.

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -68,7 +78,7 @@ class VectorizeIndexImpl implements Vectorize {
         {
           method: "POST",
           body: JSON.stringify({
-            ...options,
+            ...(options ?? {}),
```
threepointone, line 81: this shouldn't be necessary, undefined and null can spread into an object

types/defines/vectorize.d.ts
garvit-gupta: @RamIdeas @penalosa Do you have any suggestions on what I can try to make generics work, or to have a smaller impact on the type definitions?

RamIdeas: Garvit and I discussed this on a call and decided this isn't a breaking change to v1 users and the union type (over the generic type) is more ergonomic for both v1 and v2 users (+ incorrect usage is easily caught in js validation at runtime)

src/cloudflare/internal/vectorize-api.ts
garvit-gupta: Removed the check

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -42,9 +42,14 @@ class VectorizeIndexImpl implements Vectorize {
 
   public async query(
     vector: VectorFloatArray | number[],
-    options: VectorizeQueryOptions<VectorizeMetadataRetrievalLevel>
+    options?: VectorizeQueryOptions
   ): Promise<VectorizeMatches> {
     if (this.indexVersion === "v2") {
+      if (options && options.returnMetadata && !isVectorizeMetadataRetrievalLevel(options.returnMetadata) ) {
```
RamIdeas, line 48: nit

```suggestion
      if (!isVectorizeMetadataRetrievalLevel(options?.returnMetadata) ) {
```

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -59,6 +64,11 @@ class VectorizeIndexImpl implements Vectorize {
 
       return await toJson<VectorizeMatches>(res);
     } else {
+      if (options && options.returnMetadata && typeof options.returnMetadata !== 'boolean') {
```
RamIdeas, line 67: Nit

```suggestion
      if (typeof options?.returnMetadata !== 'boolean') {
```

RamIdeas (approved):

src/cloudflare/internal/test/vectorize/vectorize-mock.js
```diff
@@ -114,7 +114,7 @@ export default {
           returnSet.forEach((v) => {
             delete v.values;
           });
-        if (!body?.returnMetadata)
+        if (!body?.returnMetadata || body?.returnMetadata === "none")
```
jasnell, line 117:
```suggestion
         if ((body?.returnMetadata || 'none') === "none")
```

nit: Might be a bit cleaner

jasnell (approved):

anonrig (approved):

anonrig:
@jasnell is internal pr needed for this change?

jasnell:
To be safe I think we should have an internal CI run yes.

jasnell:
Internal CI run was goodl

PR #2478, opened 2024-08-05
reduce encoding serialization cost

anonrig:
Reduces encoding variable serialization cost between javascript and c++. by converting it to uint8_t we can reduce the "initial" cost we pay by calling almost all buffer functions.

### DIFF
--- a/src/node/internal/internal_stringdecoder.ts
+++ b/src/node/internal/internal_stringdecoder.ts
@@ -26,7 +26,7 @@
 /* todo: the following is adopted code, enabling linting one day */
 /* eslint-disable */
 
-import { Buffer, isEncoding } from 'node-internal:internal_buffer';
+import { Buffer } from 'node-internal:internal_buffer';
 import { normalizeEncoding } from 'node-internal:internal_utils';
 import {
   ERR_INVALID_ARG_TYPE,
@@ -43,15 +43,18 @@ const kBufferedBytes = 5;
 const kEncoding = 6;
 const kSize = 7;
 
-const encodings : Record<string,number> = {
-  ascii: 0,
-  latin1: 1,
-  utf8: 2,
-  utf16le: 3,
-  base64: 4,
-  base64url: 5,
-  hex: 6,
-};
+// The order of this array should be in sync with i18n.h
+// Encoding enum. Index of this array defines the uint8_t
+// value of an encoding.
+const encodings = [
+  'ascii',
+  'latin1',
+  'utf8',
+  'utf16le',
+  'base64',
+  'base64url',
+  'hex',
+];
 
 const kNativeDecoder = Symbol('kNativeDecoder');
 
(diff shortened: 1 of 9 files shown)


src/workerd/api/node/buffer.c++
```diff
@@ -88,24 +88,8 @@ void SwapBytes(kj::ArrayPtr<kj::byte> bytes) {
...
-    return Encoding::BASE64;
-  } else if (input == "base64url"_kj) {
-    return Encoding::BASE64URL;
-  } else if (input == "hex"_kj) {
-    return Encoding::HEX;
-  }
-
-  JSG_FAIL_REQUIRE(Error, kj::str("Invalid encoding: ", input));
+constexpr Encoding getEncoding(uint8_t input) {
+  return static_cast<Encoding>(input);
```
jasnell, line 92: Likely better just to skip this helper. Change the existing call sites to do the static_cast directly

src/workerd/api/node/buffer.h
```diff
@@ -35,20 +36,20 @@ class BufferUtil final: public jsg::Object {
 
   kj::Array<kj::byte> decodeString(jsg::Lock& js,
                                    jsg::JsString string,
-                                   kj::String encoding);
```
jasnell, line 38: nit: for this, probably a good idea to include a doc comment referencing the Encoding enum

anonrig: or I can just add something similar to `using EncodingValue = uint8_t;`. wdyt?

jasnell: Yep, that works too

jasnell (approved):

src/workerd/api/node/buffer.h
```diff
@@ -102,6 +103,14 @@ class BufferUtil final: public jsg::Object {
     // For StringDecoder
     JSG_METHOD(decode);
     JSG_METHOD(flush);
+
+    JSG_STATIC_CONSTANT_NAMED(ASCII, static_cast<EncodingValue>(Encoding::ASCII));
```
npaun, line 107: If we find ourselves adding more enums regularly, it would be cool to add some support to JSG to just export an entire enum without all this boilerplate.

src/node/internal/internal_stringdecoder.ts
```diff
@@ -43,15 +43,15 @@ const kBufferedBytes = 5;
...
-};
+const encodings = [
+  'ascii',
+  'latin1',
+  'utf8',
+  'utf16le',
+  'base64',
+  'base64url',
+  'hex',
+];
```
npaun, line 54: This array needs to be in the exact same order as the C++ enum or it breaks. I would add a comment about this, if there isn't a better way to do it.

src/workerd/api/node/buffer.h
anonrig: Agreed, but it is highly unlikely. The other option is to use a JSG Object which will create a heavy JS object.

src/node/internal/internal_stringdecoder.ts
anonrig: Added a comment.

npaun (approved):

Issue #2471, opened 2024-08-01
🤔  Discussion — SQLite: disabling foreign keys for migration/import?


geelen:
We have a couple of related issues:

## 1. [D1 export produces unimportable files](https://github.com/cloudflare/workers-sdk/issues/5683)

D1 Export follows SQL dump, which outputs SQL in the following order for a DB with two tables, A and B:
  1. `pragma defer_foreign_keys=TRUE`
  2. Table definition A
  3. Inserts for row A
  4. Table definition B
  5. Inserts for row B

If table A has a column that references B, the INSERT on step 3 will fail with `Parse error: no such table: main.B`. If we reorder 3 and 4, then `defer_foreign_keys` does what you'd expect and allows the rows in A to be added with no corresponding rows in B to reference (without `defer_foreign_keys`, step 3 would fail with `FOREIGN KEY constraint failed`).

This appears to be an unavoidable difference between SQLite's default behaviour of setting `PRAGMA foreign_keys=OFF;` at the beginning of DB exports, and D1/workerd using `pragma defer_foreign_keys=TRUE`. From the [SQLite docs](https://www.sqlite.org/foreignkeys.html), these are "DML errors" and are separate to the section on immediate/deferred constraints:

> If the database schema contains foreign key errors that require looking at more than one table definition to identify, then those errors are not detected when the tables are created. Instead, such errors prevent the application from preparing SQL statements that modify the content of the child or parent tables in ways that use the foreign keys. Errors reported when content is changed are "DML errors" and errors reported when the schema is changed are "DDL errors". So, in other words, misconfigured foreign key constraints that require looking at both the child and parent are DML errors. The English language error message for foreign key DML errors is usually "foreign key mismatch" but can also be "no such table" if the parent table does not exist. Foreign key DML errors are reported if:
>
>   • The parent table does not exist, or
>   • The parent key columns named in the foreign key constraint do not exist, or
>   • The parent key columns named in the foreign key constraint are not the primary key of the parent table and are not subject to a unique constraint using collating sequence specified in the CREATE TABLE, or
>   • The child table references the primary key of the parent without specifying the primary key columns and the number of primary key columns in the parent do not match the number of child key columns.

The easy solution would be to reorder the output of `d1 export` so that all tables are generated first, then all `INSERT` statements. That should make all foreign key violations deferrable, as there's no way for a dump to contain a DML error. But that doesn't solve the related issues.

[...]

PR #2479, opened 2024-08-05
SQLite: allow `foreign_key_check` for whole DB, not just a single table

geelen:
In response to https://github.com/cloudflare/workerd/issues/2471, it would be useful to allow some migrations to completely disable foreign key constraints. However, at the end of the migration, it might be important to know whether any foreign key violations remain, before setting `PRAGMA foreign_keys=ON;`.

This is possible by using `PRAGMA foreign_key_check;`, but right now we only allow that if it contains a single argument, indicating a table name. This PR allows both no-arg and single-arg forms of that pragma.

### DIFF
--- a/src/workerd/util/sqlite.c++
+++ b/src/workerd/util/sqlite.c++
@@ -326,6 +326,7 @@ enum class PragmaSignature {
   NO_ARG,
   BOOLEAN,
   OBJECT_NAME,
+  OPTIONAL_OBJECT_NAME,
   NULL_NUMBER_OR_OBJECT_NAME
 };
 struct PragmaInfo {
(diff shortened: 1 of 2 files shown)


anonrig (approved):

kossnocorp:
I'm so glad to see it happening. I'm struggling with this at this very moment, and from what I see, it should help me. If it's not too hard, can you please post a comment when this reaches production? 🙏

justin-mp:
Out of curiosity, what was the rationale for blocking the whole-DB check?

I imagine it can be pretty expensive.  Can it result in a full scan of the entire database?

geelen:
I'm not sure. From looking at the code, we didn't have a `NO_ARG_OR_OBJECT_NAME` construct, so maybe we just defaulted to allowing `OBJECT_NAME` because that seems more useful?

There's nothing stopping users from enumerating all the tables and calling `foreign_key_check` on each, so I don't think it's worth preventing users from accessing this, but maybe @kentonv has another take?

src/workerd/util/sqlite.c++
```diff
@@ -326,6 +326,7 @@ enum class PragmaSignature {
   NO_ARG,
   BOOLEAN,
   OBJECT_NAME,
+  NO_ARG_OR_OBJECT_NAME,
```
kentonv, line 329: My first reading of this, I interpreted it as "no (arg or object name)", i.e. "no arg and also no object name".

Maybe `OPTIONAL_OBJECT_NAME` would be clearer?

kentonv (approved):
I think it was unintentional that this was disallowed. Allowing it seems fine.

introduced here: 1a3ac57abbe0951c84ba4823a64f5bb27b882a3d
refactored here: f4d1e5478f80cbf5e7f4032f0c65100bb8dd70c6

PR #2480, opened 2024-08-06
Create typescript tests for apis

AdityaAtulTewari:


build/wd_test.bzl
```diff
@@ -11,12 +14,35 @@ def wd_test(
         specified.) The extension `.wd-test` is also permitted instead of `.capnp`, in order to
         avoid confusing other build systems that may assume a `.capnp` file should be compiled. As
         an extension, `.gpu-wd-test` is supported to enable special handling for GPU tests.
-     data: Files which the .capnp config file may embed. Typically JavaScript files.
+     srcs: Additional source files that can be compiled if they are typescript to be consumed by the test.
```
hoodmane, line 17:
```suggestion
     srcs: Additional Javascript or Typescript source files to be used by the test.
```

build/wd_test.bzl
```diff
@@ -11,12 +14,35 @@ def wd_test(
...
+     data: Additional Files which the .capnp config file may embed.
```
hoodmane, line 18:
```suggestion
     data: Additional files which the .capnp config file may embed.
```

build/wd_test.bzl
```diff
@@ -11,12 +14,35 @@ def wd_test(
...
 
+    for d in data:
+        if d.endswith(".js"):
+            fail("javascript files should be passed as srcs")
+
     # Add workerd binary to "data" dependencies.
-    data = data + [src, "//src/workerd/server:workerd"]
+    data = data + [src, "//src/workerd/server:workerd"] + srcs
+
+    ts_srcs = [src for src in srcs if src.endswith(".ts")]
```
hoodmane, line 29: Is there a reason you need the new `srcs` argument at all? Why not just filter `data` for arguments that end in `*.ts`? It seems simpler to me.

src/workerd/api/http-test-ts.ts
hoodmane, line 1: What does the failure look like if there is a type error? Would it be possible to add a `test_ts_testing` that forces a type error and asserts that it looks correct? Or did you manually verify this? 

It looks to me like type errors generate build errors rather than test failures. Ideally we should probably use `tsc --noCheck` at build time and run `tsc --noEmit` as part of the test. But annoyingly, it looks like the `--noCheck` CLI option is only going to be released as part of typescript v5.6:
https://github.com/microsoft/TypeScript/pull/58839

So maybe we can add this as a TODO for when typescript 5.6 comes out.

hoodmane (approved):
Thanks @AdityaAtulTewari.

build/wd_test.bzl
AdityaAtulTewari: [ ] Honestly it was cleaner to help me debug at first, I can collapse them.

src/workerd/api/http-test-ts.ts
AdityaAtulTewari: - [ ] That makes sense I can add that todo

PR #2482, opened 2024-08-06
fix edgeworker clashing with zlib function

anonrig:
Fixes internal build to fail due to clash in function naming


dom96 (approved):
Thanks for the quick fix!

jasnell (approved):

PR #2483, opened 2024-08-06
add more recipes to justfile

anonrig:
Update clang to 18 and add several alias

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,8 +1,10 @@
+alias b := build
+alias t := test
+
 default:
   @just --list
 
-# Override this value by calling `just --set clang_version 18`
-clang_version := "15"
+clang_version := "18"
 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -24,3 +28,7 @@ test *args="//...":
 
 test-asan *args="//...":
   just test {{args}} --config=asan
+
+build-test *args="//...":
+  just build {{args}}
```
fhanau, line 33: Nit: I'd go with a combined build-test step (i.e. deleting this line) which is slightly faster and will often show failing test results (if any) before having to build everything.

fhanau (approved):

npaun (approved):

fhanau (approved):

PR #2484, opened 2024-08-06
add file class to node:buffer

anonrig:
- Adds `File` to `node:buffer`
- Fixes Blob and File typescript definitions
- Fixes passing `NaN` as a `lastModified` to File options.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -26,8 +26,8 @@ devDependencies:
     specifier: ^4.1.10
     version: 4.1.10
   '@types/node':
-    specifier: ^20.11.6
-    version: 20.11.6
+    specifier: ^20.14.8
+    version: 20.14.14
   '@types/prettier':
     specifier: ^2.7.1
     version: 2.7.1
@@ -190,8 +190,8 @@ packages:
     resolution: {integrity: sha512-iiUgKzV9AuaEkZqkOLDIvlQiL6ltuZd9tGcW3gwpnX8JbuiuhFlEGmmFXEXkN50Cvq7Os88IY2v0dkDqXYWVgA==}
     dev: true
 
-  /@types/node@20.11.6:
-    resolution: {integrity: sha512-+EOokTnksGVgip2PbYbr3xnR7kZigh4LbybAfBAw5BpnQ+FqBYUsvCEjYd70IXKlbohQ64mzEYmMtlWUY8q//Q==}
+  /@types/node@20.14.14:
+    resolution: {integrity: sha512-d64f00982fS9YoOgJkAMolK7MN8Iq3TDdVjchbYHdEmjth/DHowx82GnoA+tVUAN+7vxfYUgAzi+JXbKNd2SDQ==}
     dependencies:
       undici-types: 5.26.5
     dev: true
--- a/src/node/buffer.ts
+++ b/src/node/buffer.ts
@@ -2,6 +2,8 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+
 import {
   constants,
   kMaxLength,
@@ -13,12 +15,10 @@ import {
   transcode,
 } from 'node-internal:internal_buffer';
 
-// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const atob = globalThis.atob;
-// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const btoa = globalThis.btoa;
-// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const Blob = globalThis.Blob;
+const File = globalThis.File;
 
 export {
   atob,
@@ -28,6 +28,7 @@ export {
   kStringMaxLength,
   Blob,
   Buffer,
+  File,
   SlowBuffer,
   isAscii,
   isUtf8,
@@ -35,16 +36,14 @@ export {
 };
 
 export default {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   atob,
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   btoa,
   constants,
   kMaxLength,
   kStringMaxLength,
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   Blob,
   Buffer,
+  File,
   SlowBuffer,
   isAscii,
   isUtf8,
--- a/src/node/internal/crypto_random.ts
+++ b/src/node/internal/crypto_random.ts
@@ -70,7 +70,7 @@ export function randomBytes(size: number, callback?: RandomBytesCallback) : Uint
 }
 
 export function randomFillSync(
-    buffer: ArrayBufferView|ArrayBuffer,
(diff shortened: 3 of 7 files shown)


src/node/internal/web_crypto.d.ts
```diff
@@ -1,12 +1,4 @@
 // Just enough of web crypto api to write the code in node-compat
-
-declare namespace crypto {
-  function getRandomValues<T extends ArrayBufferView | null>(array: T): T;
-  function randomUUID(): string;
-
-  const subtle: SubtleCrypto;
-}
```
npaun, line 8: This change doesn't seem to relate to the rest?

npaun (approved):

src/node/internal/web_crypto.d.ts
anonrig: updating `@types/node` makes this code redundant, and actually fails the linter, because of duplication. (latest `@types/node` already has the same exact namespace and function definitions.)

npaun: Aha!

pnpm-lock.yaml
```diff
@@ -26,8 +26,8 @@ devDependencies:
     specifier: ^4.1.10
     version: 4.1.10
   '@types/node':
-    specifier: ^20.11.6
-    version: 20.11.6
```
jasnell, line 30: @workers-frameworks @jculvey @dario-piotrowicz ... can you take a look at the types changes here to see if they make sense.

jasnell (approved):
LGTM but let's have the @workers-frameworks  folks take a look.

src/node/buffer.ts
```diff
@@ -19,6 +19,8 @@ const atob = globalThis.atob;
 const btoa = globalThis.btoa;
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const Blob = globalThis.Blob;
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
vicb, line 22: nit: disable eslint for the [whole section](https://eslint.org/docs/latest/use/configure/rules#disabling-rules) instead

src/node/internal/crypto_random.ts
```diff
@@ -70,7 +72,7 @@ export function randomBytes(size: number, callback?: RandomBytesCallback) : Uint
 }
 
 export function randomFillSync(
-    buffer: ArrayBufferView|ArrayBuffer,
+    buffer: BufferLike,
```
vicb, line 75: Is dropping `ArrayBufferView` and `ArrayBuffer` expected?

src/node/internal/crypto_random.ts
```diff
@@ -55,6 +55,8 @@ import {
   arrayBufferToUnsignedBigInt,
 } from 'node-internal:crypto_util';
 
+type BufferLike = Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | BigInt64Array | BigUint64Array;
```
vicb, line 58: nit: add a comment about the meaning of "BufferLike"?

src/workerd/api/blob.c++
```diff
@@ -305,6 +305,10 @@ jsg::Ref<File> File::constructor(jsg::Lock& js, jsg::Optional<Bits> bits,
   double lastModified;
   KJ_IF_SOME(m, maybeLastModified) {
     lastModified = m;
+
+    if (kj::isNaN(lastModified)) {
+      lastModified = 0;
+    }
```
vicb, line 311: nit: rewrite as:

```
    if (kj::isNaN(lastModified)) {
      lastModified = 0;
    } else {
      lastModified = n;
}
```
or using a ternary

src/workerd/api/node/tests/buffer-nodejs-test.js
```diff
@@ -5792,3 +5793,104 @@ export const transcodeTest = {
     }
   }
 };
+
+// Tests are taken from Node.js
+// https://github.com/nodejs/node/blob/main/test/parallel/test-file.js
```
dario-piotrowicz, line 5798: it's pretty minor, but when linking to a gh repo I always prefer permalinks so that if there are changes the link is still valid:
```suggestion
// https://github.com/nodejs/node/blob/a4f609fa/test/parallel/test-file.js
```

dario-piotrowicz (approved):
Looks good to me 🙂

src/workerd/api/node/tests/buffer-nodejs-test.js
```diff
@@ -5792,3 +5793,104 @@ export const transcodeTest = {
...
+export const fileTest = {
+  test() {
+    throws(() => new File(), TypeError);
+    throws(() => new File([]), TypeError);
+    throws(() => File.prototype.name, TypeError);
+    throws(() => File.prototype.lastModified, TypeError);
+
+    {
```
vicb, line 5806: Not sure if possible with the test framework here, but it would be nice to different tests, each with a description - there are already blocks scoping the test cases, it would be nice to add description and different fn (i.e. à la Jest `describe` / `test`)

src/workerd/api/node/tests/buffer-nodejs-test.js
vicb: Tip: hit 'y' on github to get the SHA based URL

vicb (commented):
I added a few comments on this file.

This PR also touches `crypto`, what is the reason, could it be a distinct commit?

As I was explaining to Dario earlier, I usually add a lot of comments when I review code. Feel free to ignore them if not applicable.

src/node/internal/crypto_random.ts
jasnell: Ah, yeah that's a good catch. I don't think the definition of `BufferLike` is complete enough here. This should accept `DataView`, `ArrayBuffer`, and `SharedArrayBuffer`

src/workerd/api/node/tests/buffer-nodejs-test.js
jasnell: They can be split into individual `export const fooTest = { test() { ... } }` blocks

src/node/internal/crypto_random.ts
anonrig: ok, i found a better solution: `NodeJS.ArrayBufferView` defined in `@types/node`

anonrig: Hot take: We should have a linter rule to avoid calling `ArrayBufferView` and keep our internal functions types in sync with `@nodejs/types`

src/node/internal/crypto_random.ts
```diff
@@ -142,7 +142,7 @@ export function randomFill(buffer: ArrayBufferView|ArrayBuffer,
   // We're currently not actually implementing the fill itself asynchronously,
   // so we defer to randomFillSync here, but we invoke the callback asynchronously.
   new Promise<void>((res) => {
-    randomFillSync(buffer, offset, size);
+    randomFillSync((buffer as Uint8Array), offset, size);
```
dario-piotrowicz, line 145: is this casting necessary?

Above you're updating the `randomFillSync` function's `buffer` parameter type to `NodeJS.ArrayBufferView` and that is the type of `buffer` isn't it?

anonrig: Yeah you're right. I missed this.

anonrig:
Internal PR needs to land first, since we need to update `@types/node` in there as well.

PR #2491, opened 2024-08-07
R2-2317 add support for SSE-C via `workerd` bindings

helloimalastair:
Give R2 Bindings support for creating/interacting with *SSE-C* encrypted objects.

Tests run on branch `mikkel/R2-2317`. Please merge once feature lands

### DIFF
--- a/src/workerd/api/r2-api.capnp
+++ b/src/workerd/api/r2-api.capnp
@@ -60,6 +60,10 @@ struct R2Conditional {
   # timestamp.
 }
 
+struct R2SSECOptions {
+  key @0 :Text;
+}
+
 struct R2Checksums {
   # The JSON name of these fields must comform to the representation of the ChecksumAlgorithm in
   # the R2 gateway worker.
@@ -93,6 +97,7 @@ struct R2GetRequest {
   range @1 :R2Range;
   rangeHeader @3 :Text;
   onlyIf @2 :R2Conditional;
+  ssec @4 :R2SSECOptions;
 }
 
 struct R2PutRequest {
@@ -106,6 +111,7 @@ struct R2PutRequest {
   sha384 @7 :Data $Json.hex;
   sha512 @8 :Data $Json.hex;
   storageClass @9 :Text;
+  ssec @10 :R2SSECOptions;
 }
 
 struct R2CreateMultipartUploadRequest {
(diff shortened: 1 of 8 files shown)


helloimalastair:

harshal317:
- [ ] R2 API Support

src/workerd/api/r2-api.capnp
```diff
@@ -106,19 +111,22 @@ struct R2PutRequest {
...
 }
 
 struct R2UploadPartRequest {
   object @0 :Text;
   uploadId @1 :Text;
   partNumber @2 :UInt32;
+  ssec @3 :R2SSECOptions;
 }
 
 struct R2CompleteMultipartUploadRequest {
```
Frederik-Baetens, line 132: The R2 api side has an ssec field. Aren't you missing that here?

src/workerd/api/r2-bucket.c++
```diff
@@ -559,6 +582,22 @@ jsg::Promise<kj::Maybe<jsg::Ref<R2Bucket::HeadResult>>> R2Bucket::put(jsg::Lock&
       KJ_IF_SOME(s, o.storageClass) {
         putBuilder.setStorageClass(s);
       }
+      KJ_IF_SOME(ssecKey, o.ssecKey) {
```
Frederik-Baetens, line 585: Can you de-duplicate this code? Seems like an exact copy of the code used in initGetOptions?

src/workerd/api/r2-api.capnp
helloimalastair: I have a PR up to remove that, since it isn't actually necessary. Should be merged soon-ish

src/workerd/api/r2-bucket.c++
helloimalastair: Split into separate function.

anonrig (approved):

src/workerd/api/r2-bucket.c++
```diff
@@ -296,6 +320,13 @@ void initGetOptions(jsg::Lock& js, Builder& builder, Options& o) {
       }
     }
   }
+  KJ_IF_SOME(ssecKey, o.ssecKey) {
```
Frederik-Baetens, line 323: Small nit, but you can move this if some into the buildSsceKey function, since the buildSsceKey already returns a kj::Maybe. Then if the argument is none, the result will just be none too, and the KJ_IF_SOME below will handle that case already.

Frederik-Baetens (approved):

PR #2492, opened 2024-08-07
wip: Add workflow entrypoint

sidharthachatterjee:
- Adds a new entrypoint class for Workflows with a constructor that sets up `ctx` and `env`
- Adds RPC types for Workflows

### DIFF
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -1831,4 +1831,18 @@ jsg::Ref<DurableObjectBase> DurableObjectBase::constructor(
   return jsg::alloc<DurableObjectBase>();
 }
 
+jsg::Ref<Workflow> Workflow::constructor(
+    const v8::FunctionCallbackInfo<v8::Value>& args,
+    jsg::Ref<ExecutionContext> ctx, jsg::JsObject env) {
+  // HACK: We take `FunctionCallbackInfo` mostly so that we can set properties directly on
+  //   `This()`. There ought to be a better way to get access to `this` in a constructor.
+  //   We *also* delcare `ctx` and `env` params more explicitly just for the sake of type checking.
+  jsg::Lock& js = jsg::Lock::from(args.GetIsolate());
+
+  jsg::JsObject self(args.This());
+  self.set(js, "ctx", jsg::JsValue(args[0]));
+  self.set(js, "env", jsg::JsValue(args[1]));
+  return jsg::alloc<Workflow>();
+}
+
 }; // namespace workerd::api
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -1546,6 +1546,9 @@ Worker::Worker(kj::Own<const Script> scriptParam,
                             } else if (handle == entrypointClasses.workerEntrypoint) {
                               impl->statelessClasses.insert(kj::mv(handler.name), kj::mv(cls));
                               return;
+                            } else if (handle == entrypointClasses.workflow) {
+                              impl->statelessClasses.insert(kj::mv(handler.name), kj::mv(cls));
+                              return;
                             }
 
                             handle = KJ_UNWRAP_OR(handle.getPrototype(js).tryCast<jsg::JsObject>(), {
--- a/src/workerd/server/workerd-api.c++
+++ b/src/workerd/server/workerd-api.c++
@@ -3,6 +3,7 @@
 //     https://opensource.org/licenses/Apache-2.0
 
 #include "workerd-api.h"
+#include "workerd/api/worker-rpc.h"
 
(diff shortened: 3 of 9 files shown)


sidharthachatterjee:

src/workerd/io/worker.c++
```diff
@@ -1546,6 +1546,9 @@ Worker::Worker(kj::Own<const Script> scriptParam,
                             } else if (handle == entrypointClasses.workerEntrypoint) {
                               impl->statelessClasses.insert(kj::mv(handler.name), kj::mv(cls));
                               return;
+                            } else if (handle == entrypointClasses.workflow) {
```
jasnell, line 1549: for the time being, should we gate access to this behind the `experimental` flag?

src/workerd/api/worker-rpc.c++
```diff
@@ -1831,4 +1831,18 @@ jsg::Ref<DurableObjectBase> DurableObjectBase::constructor(
...
+    jsg::Ref<ExecutionContext> ctx, jsg::JsObject env) {
+  // HACK: We take `FunctionCallbackInfo` mostly so that we can set properties directly on
+  //   `This()`. There ought to be a better way to get access to `this` in a constructor.
+  //   We *also* delcare `ctx` and `env` params more explicitly just for the sake of type checking.
+  jsg::Lock& js = jsg::Lock::from(args.GetIsolate());
+
+  jsg::JsObject self(args.This());
+  self.set(js, "ctx", jsg::JsValue(args[0]));
+  self.set(js, "env", jsg::JsValue(args[1]));
+  return jsg::alloc<Workflow>();
```
jasnell, line 1845: nit: would be nice to dedupe this by moving into a utility method

src/workerd/io/worker.c++
elithrar: Yes.

sidharthachatterjee: Very good call out

src/workerd/server/workerd-api.c++
```diff
@@ -234,6 +235,7 @@ WorkerdApi::EntrypointClasses WorkerdApi::getEntrypointClasses(jsg::Lock& lock)
   return {
     .workerEntrypoint = typedLock.getConstructor<api::WorkerEntrypoint>(lock.v8Context()),
     .durableObject = typedLock.getConstructor<api::DurableObjectBase>(lock.v8Context()),
+    .workflow = typedLock.getConstructor<api::Workflow>(lock.v8Context()),
```
kentonv, line 238: This will need to land together with an internal PR making the same change to the internal version of this file, otherwise I think the edge runtime will crash due to this being left null.

src/workerd/io/worker.c++
```diff
@@ -1546,6 +1546,9 @@ Worker::Worker(kj::Own<const Script> scriptParam,
...
+                              impl->statelessClasses.insert(kj::mv(handler.name), kj::mv(cls));
```
kentonv, line 1550: We probably want to extend `validateHandlers()` such that it reports `Workflow` classes differently from regular `WorkerEntrypoint`s. This way, the config service will be able to distinguish between workflow entrypoints and RPC entrypoints, so that it can detect when someone has defined a Workflow and prompt them to configure it as such, and also block people from creating service bindings to workflows (which don't make sense).

src/workerd/api/worker-rpc.c++
```diff
@@ -1831,4 +1831,18 @@ jsg::Ref<DurableObjectBase> DurableObjectBase::constructor(
   return jsg::alloc<DurableObjectBase>();
 }
 
+jsg::Ref<Workflow> Workflow::constructor(
+    const v8::FunctionCallbackInfo<v8::Value>& args,
+    jsg::Ref<ExecutionContext> ctx, jsg::JsObject env) {
+  // HACK: We take `FunctionCallbackInfo` mostly so that we can set properties directly on
+  //   `This()`. There ought to be a better way to get access to `this` in a constructor.
+  //   We *also* delcare `ctx` and `env` params more explicitly just for the sake of type checking.
```
anonrig, line 1839:
```suggestion
  //   We *also* declare `ctx` and `env` params more explicitly just for the sake of type checking.
```

kentonv (approved):

jasnell (approved):
LGTM but ideally this would come with at least one test. Is there a plan to add testing soon?

sidharthachatterjee:
> LGTM but ideally this would come with at least one test. Is there a plan to add testing soon?

Yeah, we've been iterating quickly on API shape. I'm going to open another PR with some comprehensive testing shortly.

jasnell:
Internal CI run appears to be successful and green so removing the `needs-internal-pr` label.

PR #2494, opened 2024-08-07
Adds APIs to jsg::Lock to perform manual external memory accounting.

jasnell:
While ultimately we want to explore more automatic memory accounting for jsg objects, there's still a need for manual adjustments. This PR adds two new APIs to `jsg::Lock` for manual external memory adjustments.

Why is this needed? Let's say you have a `jsg::Object` that holds onto an internal `kj::Array<kj::byte>` that is never wrapped in a `v8::ArrayBuffer` ... v8 won't never know about that memory allocation for that `kj::Array` unless we manually inform it.

```
public:
  Foo(jsg::Lock& js) : foo(kj::heapArray<int>(1024).attach(js.getExternalMemoryAdjuster(1024))) {}
private:
  kj::Array<kj::byte> foo;
}
```

Manual adjustment is imperfect and should only be used when the need is clear. We want to work on making the adjustments largely automatic when possible to do so, but for now this gives us a path forward to having better external memory tracking.

Note that when a `kj::Array<kj::byte>` is wrapped by a `v8::ArrayBuffer`, the `v8::ArrayBuffer` itself will handle reporting for us, so we don't need to manually adjust in those cases.

### DIFF
--- a/src/workerd/jsg/jsg.c++
+++ b/src/workerd/jsg/jsg.c++
@@ -278,6 +278,29 @@ JsSymbol Lock::symbolAsyncDispose() {
   return IsolateBase::from(v8Isolate).getSymbolAsyncDispose();
 }
 
+void Lock::adjustExternalMemory(ssize_t amount) {
+  v8Isolate->AdjustAmountOfExternalAllocatedMemory(static_cast<int64_t>(amount));
+}
+
+namespace {
+struct ExternalMemoryAdjuster {
+  int64_t size;
+  v8::Isolate* isolate;
+  ExternalMemoryAdjuster(v8::Isolate* isolate, size_t size)
+      : size(static_cast<int64_t>(size)), isolate(isolate) {
+    isolate->AdjustAmountOfExternalAllocatedMemory(size);
+  }
+
+  ~ExternalMemoryAdjuster() noexcept(false) {
+    isolate->AdjustAmountOfExternalAllocatedMemory(-size);
+  }
+};
+}  // namespace
+
+kj::Own<void> Lock::getExternalMemoryAdjuster(size_t amount) {
+  return kj::heap<ExternalMemoryAdjuster>(v8Isolate, amount);
+}
+
 Name::Name(kj::String string)
     : hash(kj::hashCode(string)),
       inner(kj::mv(string)) {}
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -2085,6 +2085,20 @@ public:
     return *reinterpret_cast<Lock*>(v8Isolate->GetData(2));
   }
 
+  // Manually make adjustments to the amount of external memory reported to V8.
+  // This is useful when we have a large amount of external memory allocated that
+  // typically would not be visible to v8's memory tracking. In the future we hope
+  // to make most memory accounting automatic, making most direct uses of this
+  // API unnecessary but there will always be times when manual adjustments are
+  // necessary.
+  void adjustExternalMemory(ssize_t amount);
+
+  // Reports amount of external memory to be manually attributed to the isolate.
+  // When the returned kj::Own<void> is dropped, the amount will be subtracted
+  // from the isolate's external memory accounting. It is important that these
+  // be held onto only during the lifetime of the isolate.
+  kj::Own<void> getExternalMemoryAdjuster(size_t amount);
+
   Value parseJson(kj::ArrayPtr<const char> data);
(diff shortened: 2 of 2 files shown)


src/workerd/jsg/jsg.c++
```diff
@@ -278,6 +278,29 @@ JsSymbol Lock::symbolAsyncDispose() {
...
+  int64_t size;
+  v8::Isolate* isolate;
+  ExternalMemoryAdjuster(v8::Isolate* isolate, size_t size)
+      : size(static_cast<int64_t>(size)), isolate(isolate) {
+    isolate->AdjustAmountOfExternalAllocatedMemory(size);
+  }
+
+  ~ExternalMemoryAdjuster() noexcept(false) {
+    isolate->AdjustAmountOfExternalAllocatedMemory(-size);
+  }
```
jasnell, line 296: Review note: Obviously this can be a problem if the `kj::Own<void>` for this is held beyond the lifespan of the `isolate`. Should we include more protections against that by making the `isolate` effectively a weak ref rather than a bare pointer? Should we forgo this entirely in favor of just having the manual `adjustExternalMemory(...)` call?

The key use case for this would be to allow for something like..

```
auto ary = kj::heapArray<int>(10).attach(js.getExternalMemoryAdjuster(10 * sizeof(int)));
```

Such that the external memory will be automatically adjusted in the isolate when the array is freed. This, of course, is only really appropriate if the array does not get bound to an `ArrayBuffer` at any point, which would cause the data to be double accounted.

npaun (approved):
Makes sense to me as an interface

src/workerd/jsg/jsg.c++
```diff
@@ -278,6 +278,29 @@ JsSymbol Lock::symbolAsyncDispose() {
   return IsolateBase::from(v8Isolate).getSymbolAsyncDispose();
 }
 
+void Lock::adjustExternalMemory(ssize_t amount) {
```
anonrig, line 281: nit: since we static_cast to int64_t, why not have a int64_t parameter? and also why `ssize_t` here but `size_t` in ExternalMemoryAdjuster constructor?

src/workerd/jsg/jsg.h
```diff
@@ -2085,6 +2085,20 @@ class Lock {
...
+  // typically would not be visible to v8's memory tracking. In the future we hope
+  // to make most memory accounting automatic, making most direct uses of this
+  // API unnecessary but there will always be times when manual adjustments are
+  // necessary.
+  void adjustExternalMemory(ssize_t amount);
+
+  // Reports amount of external memory to be manually attributed to the isolate.
+  // When the returned kj::Own<void> is dropped, the amount will be subtracted
+  // from the isolate's external memory accounting. It is important that these
+  // be held onto only during the lifetime of the isolate.
```
anonrig, line 2099: Would you mind providing an example use-case in this comment? It's hard to come up with one, as a reviewer.

anonrig (approved):

src/workerd/jsg/jsg.h
jasnell: There's an example in the original PR post. This would be for allocations that are held but not ever associated with a `v8::ArrayBuffer`

src/workerd/jsg/jsg.c++
jasnell: `ssize_t` because the `adjustExternalMemory` will take positive or negative values.

`size_t` in the constructor because that only needs to accept positive numbers and we'll invert it internally.

As for why `size_t` and not `int64_t`... really just so it is explicit in the API what the values are intended to be.

jasnell:
Going to go ahead and merge. If we want to tweak the APIs further for any reason we can, but these are good enough to get us started.

src/workerd/jsg/jsg.c++
kentonv, line 281: I think this API is error-prone: it's too easy to adjust the memory up and forget to adjust it back down later.

Could we create an RAII-based API instead? Have this method return some sort of an object whose destructor adjusts the memory back down. The object could also have a method to modify the amount of memory it's representing, for cases where we're tracking memory for a data structure that changes over time.

I think this object can just be a trivial wrapper around an integer; no allocation required. It can have a move constructor which zeros out the old value.

kentonv: Oh, I missed that this PR actually introduces an RAII API as well.

I think we should _only_ offer the RAII API. We can extend it to work well even when the memory allocation changes over time, using the technique I suggested. Also can avoid the need for a heap allocation.

PR #2496, opened 2024-08-07
node:zlib implem: Expose zlib.constants

npaun:


npaun:
recheck

src/node/internal/zlib.d.ts
```diff
@@ -1 +1,115 @@
...
+export const BROTLI_DECODER_PARAM_LARGE_WINDOW: number;
+export const BROTLI_DECODER_NO_ERROR: number;
+export const BROTLI_DECODER_SUCCESS: number;
+export const BROTLI_DECODER_NEEDS_MORE_INPUT: number;
+export const BROTLI_DECODER_NEEDS_MORE_OUTPUT: number;
+export const BROTLI_DECODER_ERROR_FORMAT_EXUBERANT_NIBBLE: number;
+export const BROTLI_DECODER_ERROR_FORMAT_RESERVED: number;
+export const BROTLI_DECODER_ERROR_FORMAT_EXUBERANT_META_NIBBLE: number;
+export const BROTLI_DECODER_ERROR_FORMAT_SIMPLE_HUFFMAN_ALPHABET: number;
+export const BROTLI_DECODER_ERROR_FORMAT_SIMPLE_HUFFMAN_SAME: number;
```
fhanau, line 95: Do we really need all of these defines? I can't find a reference to this define in the node:zlib spec.

src/node/internal/internal_zlib.ts
```diff
@@ -13,7 +13,122 @@ function crc32(data: ArrayBufferView | string, value: number = 0): number {
   validateUint32(value, 'value');
   return zlibUtil.crc32(data, value);
 }
+const constants = {
```
anonrig, line 16: If you have explicit testing you can make JS implementation a lot simpler

```suggestion
const constants = Object.keys().filter(k => k.startsWith())
```

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -213,3 +213,10 @@ export const crc32Test = {
     });
   }
 }
+
+export const constantsTest = {
+  test() {
+    strictEqual(constants.Z_NEED_DICT, 2);
+    strictEqual(constants.BROTLI_DECODER_ERROR_FORMAT_EXUBERANT_META_NIBBLE, -3);
```
anonrig, line 220: We need better testing here. Particularly, snapshot testing in here would be the safest.

src/workerd/api/node/zlib-util.h
```diff
@@ -1,9 +1,15 @@
 // Copyright (c) 2017-2022 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
+#include "src/workerd/jsg/jsg.h"
+#include "zlib.h"
 #include <kj/array.h>
```
anonrig, line 6: Can we keep the internal/external/system wise includes in an alphabetical order and within different groups?

src/node/internal/zlib.d.ts
npaun: I'm not sure - they're definitely all present when you import `node:zlib` but it's true that not all are documented.

src/workerd/api/node/zlib-util.h
```diff
@@ -12,11 +18,155 @@ class ZlibUtil final : public jsg::Object {
   ZlibUtil() = default;
   ZlibUtil(jsg::Lock&, const jsg::Url&) {}
 
+    // Certain zlib constants are defined by NodeJS itself
+    static constexpr auto Z_MIN_CHUNK = 64;
+    static constexpr auto Z_MAX_CHUNK = std::numeric_limits<double>::infinity();
```
anonrig, line 23: We should have 128mb limit here instead of this. fyi @jasnell

src/workerd/api/node/zlib-util.h
npaun: I thought clang-format would take care of this. I'll look into it.

src/workerd/api/node/tests/zlib-nodejs-test.js
npaun: Do you have an example of some existing snapshot testing in workerd? I'll snapshot all the constants then?

anonrig: Something like this should work deepStrictEqual(JSON.stringify(constants), "")

src/node/internal/zlib.d.ts
fhanau: Maybe we can limit it to the documented ones https://nodejs.org/api/zlib.html#brotli-constants – it's probably not overly consequential, but we won't be able to remove constants later on based on the compatibility limit and everything we add to the JSG API increases the compile time of the main translation unit that includes all JSG definitions (workerd-api.c++ for workerd), which is a bottleneck for us.

npaun: They're in Node's type definitions for zlib FWIW. 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ec8ec925cf490ce8825692cc2ea67c6cefd6d996/types/node/zlib.d.ts#L346

src/workerd/api/node/zlib-util.h
npaun: I fixed it manually this time, but we could definitely look into configuring clang-format to do this,.

src/node/internal/internal_zlib.ts
npaun: I don't see how that code would produce an object, but I think you're suggesting I select all the constants beginning with `Z_`, `BROTLI_`, etc. at runtime?

anonrig: I was implying something like this Object.keys().filter().reduce(fn, {})

src/node/internal/internal_zlib.ts
```diff
@@ -14,6 +14,122 @@ function crc32(data: ArrayBufferView | string, value: number = 0): number {
...
+BROTLI_DECODER_ERROR_FORMAT_DISTANCE: zlibUtil.BROTLI_DECODER_ERROR_FORMAT_DISTANCE,
+BROTLI_DECODER_ERROR_DICTIONARY_NOT_SET: zlibUtil.BROTLI_DECODER_ERROR_DICTIONARY_NOT_SET,
+BROTLI_DECODER_ERROR_INVALID_ARGUMENTS: zlibUtil.BROTLI_DECODER_ERROR_INVALID_ARGUMENTS,
+BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MODES: zlibUtil.BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MODES,
+BROTLI_DECODER_ERROR_ALLOC_TREE_GROUPS: zlibUtil.BROTLI_DECODER_ERROR_ALLOC_TREE_GROUPS,
+BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MAP: zlibUtil.BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MAP,
+BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_1: zlibUtil.BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_1,
+BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_2: zlibUtil.BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_2,
+BROTLI_DECODER_ERROR_ALLOC_BLOCK_TYPE_TREES: zlibUtil.BROTLI_DECODER_ERROR_ALLOC_BLOCK_TYPE_TREES,
+BROTLI_DECODER_ERROR_UNREACHABLE: zlibUtil.BROTLI_DECODER_ERROR_UNREACHABLE,
```
jasnell, line 129: nit: these aren't actually going to be constant defined this way.  These would likely need to be added use `Object.defineProperties(...)` with `writable: false` and `configurable: false`.

src/node/internal/zlib.d.ts
```diff
@@ -1 +1,115 @@
...
+export const BROTLI_DECODER_ERROR_FORMAT_DISTANCE: number;
+export const BROTLI_DECODER_ERROR_DICTIONARY_NOT_SET: number;
+export const BROTLI_DECODER_ERROR_INVALID_ARGUMENTS: number;
+export const BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MODES: number;
+export const BROTLI_DECODER_ERROR_ALLOC_TREE_GROUPS: number;
+export const BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MAP: number;
+export const BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_1: number;
+export const BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_2: number;
+export const BROTLI_DECODER_ERROR_ALLOC_BLOCK_TYPE_TREES: number;
+export const BROTLI_DECODER_ERROR_UNREACHABLE: number;
```
anonrig, line 115: EOF new line is missing

src/node/internal/internal_zlib.ts
```diff
@@ -14,6 +14,20 @@ function crc32(data: ArrayBufferView | string, value: number = 0): number {
...
+const constants = {};
+// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+Object.entries(Object.getPrototypeOf(zlibUtil))
+  .filter(([k,]) => k.startsWith(constPrefix))
+  .forEach(([k, v]) => Object.defineProperty(constants, k.slice(constPrefix.length), {
+    value: v,
+    writable: false,
+    configurable: false,
+    enumerable: true
+  }));
```
npaun, line 28: @anonrig, @jasnell
Here's my implementation of dynamically setting up the zlib constants. I find it kind of ugly though - let me know if you have any suggestions.

src/node/internal/zlib.d.ts
jasnell: I would keep them. In Node.js the actual exported values are the source of truth and not the documentation. It is possible that someone just forgot to document these or any number of things.

src/node/internal/internal_zlib.ts
jasnell: This is ok. I personally would have just listed them out individually but this works fine also

jasnell: Fwiw, however, using `Object.defineProperties(...)` and setting multiple properties in one go there is a lot more performant than repeatedly calling `Object.defineProperty` one at a time.

src/workerd/api/node/zlib-util.h
```diff
@@ -12,11 +19,180 @@ class ZlibUtil final : public jsg::Object {
   ZlibUtil() = default;
   ZlibUtil(jsg::Lock&, const jsg::Url&) {}
 
+  // Certain zlib constants are defined by NodeJS itself
+  static constexpr auto Z_MIN_CHUNK = 64;
+  static constexpr auto Z_MAX_CHUNK = std::numeric_limits<double>::infinity();
```
jasnell, line 24: This should likely be capped at `128 * 1024`

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -213,3 +214,117 @@ export const crc32Test = {
...
+  "BROTLI_DECODER_ERROR_FORMAT_DISTANCE": -16,
+  "BROTLI_DECODER_ERROR_DICTIONARY_NOT_SET": -19,
+  "BROTLI_DECODER_ERROR_INVALID_ARGUMENTS": -20,
+  "BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MODES": -21,
+  "BROTLI_DECODER_ERROR_ALLOC_TREE_GROUPS": -22,
+  "BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MAP": -25,
+  "BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_1": -26,
+  "BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_2": -27,
+  "BROTLI_DECODER_ERROR_ALLOC_BLOCK_TYPE_TREES": -30,
+  "BROTLI_DECODER_ERROR_UNREACHABLE": -31
```
jasnell, line 327: nit: if constants are added later it would be easy for this list to get out of date. I wonder if it would make sense to have the test fail if there are any unexpected constants in the `constants` object. You can check for that fairly easily by grabbing the `Object.keys()` and removing each item that is explicitly tested here then throwing if there are any left over at the end.

src/workerd/api/BUILD.bazel
```diff
@@ -74,6 +74,7 @@ wd_cc_library(
         "//src/workerd/io",
         "//src/workerd/jsg:rtti",
         "//src/workerd/server:workerd_capnp",
+        "@capnp-cpp//src/kj/compat:kj-brotli",
```
fhanau, line 77: Nit: kj-brotli is not directly used by rtti.c++, no need to add this here.

src/workerd/api/node/tests/zlib-nodejs-test.js
npaun: Are you suggesting to test for this list of keys, and not check the values, or am I misunderstanding the suggestion?

npaun:
Thanks for the suggestions, guys. It should be ready to go now.

anonrig (approved):

src/workerd/api/node/tests/zlib-nodejs-test.js
jasnell: For for the list of keys and the values, but throw if there are additional keys in `constants` not tested for explicitly.

jasnell (approved):

src/workerd/api/node/zlib-util.h
```diff
@@ -12,11 +19,184 @@ class ZlibUtil final : public jsg::Object {
...
+
+  static constexpr auto Z_MAX_MEMLEVEL = 9;
+  static constexpr auto Z_DEFAULT_MEMLEVEL = 8;
+  static constexpr auto Z_MIN_LEVEL = -1;
+  static constexpr auto Z_MAX_LEVEL = 9;
+  static constexpr auto Z_DEFAULT_LEVEL = Z_DEFAULT_COMPRESSION;
+  static constexpr auto Z_MIN_WINDOWBITS = 8;
+  static constexpr auto Z_MAX_WINDOWBITS = 15;
+  static constexpr auto Z_DEFAULT_WINDOWBITS = 15;
+  enum class ZlibMode {
```
anonrig, line 36:
```suggestion
  using ZlibModeValue = uint8_t;
  enum ZlibMode : uint8_t {
```

src/workerd/api/node/zlib-util.h
```diff
@@ -12,11 +19,184 @@ class ZlibUtil final : public jsg::Object {
...
+
+    JSG_STATIC_CONSTANT_NAMED(CONST_DEFLATE, static_cast<uint8_t>(ZlibMode::DEFLATE));
+    JSG_STATIC_CONSTANT_NAMED(CONST_INFLATE, static_cast<uint8_t>(ZlibMode::INFLATE));
+    JSG_STATIC_CONSTANT_NAMED(CONST_GZIP, static_cast<uint8_t>(ZlibMode::GZIP));
+    JSG_STATIC_CONSTANT_NAMED(CONST_GUNZIP, static_cast<uint8_t>(ZlibMode::GUNZIP));
+    JSG_STATIC_CONSTANT_NAMED(CONST_DEFLATERAW, static_cast<uint8_t>(ZlibMode::DEFLATERAW));
+    JSG_STATIC_CONSTANT_NAMED(CONST_INFLATERAW, static_cast<uint8_t>(ZlibMode::INFLATERAW));
+    JSG_STATIC_CONSTANT_NAMED(CONST_UNZIP, static_cast<uint8_t>(ZlibMode::UNZIP));
+    JSG_STATIC_CONSTANT_NAMED(CONST_BROTLI_DECODE, static_cast<uint8_t>(ZlibMode::BROTLI_DECODE));
+    JSG_STATIC_CONSTANT_NAMED(CONST_BROTLI_ENCODE, static_cast<uint8_t>(ZlibMode::BROTLI_ENCODE));
```
anonrig, line 91:
```suggestion
    JSG_STATIC_CONSTANT_NAMED(CONST_DEFLATE, static_cast<ZlibModeValue>(ZlibMode::DEFLATE));
    JSG_STATIC_CONSTANT_NAMED(CONST_INFLATE, static_cast<ZlibModeValue>(ZlibMode::INFLATE));
    JSG_STATIC_CONSTANT_NAMED(CONST_GZIP, static_cast<ZlibModeValue>(ZlibMode::GZIP));
    JSG_STATIC_CONSTANT_NAMED(CONST_GUNZIP, static_cast<ZlibModeValue>(ZlibMode::GUNZIP));
    JSG_STATIC_CONSTANT_NAMED(CONST_DEFLATERAW, static_cast<ZlibModeValue>(ZlibMode::DEFLATERAW));
    JSG_STATIC_CONSTANT_NAMED(CONST_INFLATERAW, static_cast<ZlibModeValue>(ZlibMode::INFLATERAW));
    JSG_STATIC_CONSTANT_NAMED(CONST_UNZIP, static_cast<ZlibModeValue>(ZlibMode::UNZIP));
    JSG_STATIC_CONSTANT_NAMED(CONST_BROTLI_DECODE, static_cast<ZlibModeValue>(ZlibMode::BROTLI_DECODE));
    JSG_STATIC_CONSTANT_NAMED(CONST_BROTLI_ENCODE, static_cast<ZlibModeValue>(ZlibMode::BROTLI_ENCODE));
```

jasnell:
If you would, rebase this PR on `main` again and let's give the internal PR CI run another go.

npaun:
@jasnell Done. However I still didn't sort out the zlib version inconsistency so I think it'll fail again.

jasnell:
That's fine, I just noticed that the internal streams.ew-test failed and it's likely because of the commit that got reverted this morning. The rebase at least helps to rule that out

src/workerd/api/node/zlib-util.h
```diff
@@ -1,9 +1,16 @@
 // Copyright (c) 2017-2022 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
-#include <kj/array.h>
+
 #include <workerd/jsg/jsg.h>
 
+#include <kj/array.h>
+#include <kj/compat/brotli.h>
+#include <zlib.h>
```
anonrig, line 9: This is probably why your internal builds are failing as well:

```suggestion
#include "zlib.h"
```

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -213,3 +214,117 @@ export const crc32Test = {
...
+  "Z_NO_COMPRESSION": 0,
+  "Z_BEST_SPEED": 1,
+  "Z_BEST_COMPRESSION": 9,
+  "Z_DEFAULT_COMPRESSION": -1,
+  "Z_FILTERED": 1,
+  "Z_HUFFMAN_ONLY": 2,
+  "Z_RLE": 3,
+  "Z_FIXED": 4,
+  "Z_DEFAULT_STRATEGY": 0,
+  "ZLIB_VERNUM": 4880,
```
jasnell, line 245: Since we know there's a version mismatch between workerd and the internal repo, I would likely just remove the check for `ZLIB_VERNUM` in the test for now.

anonrig: We can just check for Object.keys().sort() strict equality?

npaun: I'll do this for now, but we should fix our dependency mess when we can.

src/workerd/api/node/zlib-util.h
npaun: Unfortunately does not work.

PR #2497, opened 2024-08-08
EW-8447 Fix CPU profiling

harrishancock:
When breakpoint debugging was introduced, it added a new thread, the InspectorService thread. This thread is necessary in order to continue performing WebSocket I/O while the Isolate is paused at a breakpoint. The feature _also_ moved inspector message dispatching onto this same InspectorService thread. For most inspector messages, this turns out to be benign, but not for CPU profiling start/stop messages: those must be dispatched on the thread which we desire to profile. The end result was that breakpoint debugging caused us to profile JavaScript on the InspectorService thread (e.g., anything executed on the Chrome Devtools console) instead of the Isolate thread, as originally intended.

This turns out to be relatively easy to fix, because breakpoint debugging would already dispatch inspector messages on the Isolate thread in a specific case: when the Isolate was paused at a breakpoint, meaning there are already exists some machinery to pass inspector messages across threads. This commit simply makes it so that all messages are passed to the Isolate thread for dispatching, instead of only when paused at a breakpoint.

We (obviously) do not have any tests for CPU profiling, or else we would have caught this issue much earlier. This commit does not add any such tests. Since we'd like to fix this issue sooner than later, my intent is to land this commit first and add tests later -- we will not close related tickets until such tests are implemented.

I have smoke tested both profiling and breakpoint debugging; both work with the change in this commit.

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -2307,8 +2307,11 @@ struct MessageQueue {
 
 class Worker::Isolate::InspectorChannelImpl final: public v8_inspector::V8Inspector::Channel {
 public:
-  InspectorChannelImpl(kj::Own<const Worker::Isolate> isolateParam, kj::WebSocket& webSocket)
-      : ioHandler(webSocket), state(kj::heap<State>(this, kj::mv(isolateParam))) {
+  InspectorChannelImpl(kj::Own<const Worker::Isolate> isolateParam,
+      ExecutorNotifierPair isolateThreadExecutorNotifierPair,
+      kj::WebSocket& webSocket)
+      : ioHandler(kj::mv(isolateThreadExecutorNotifierPair), webSocket),
+        state(kj::heap<State>(this, kj::mv(isolateParam))) {
     ioHandler.connect(*this);
   }
 
@@ -2558,11 +2561,12 @@ private:
   // the InspectorChannelImpl and the InspectorClient.
   class WebSocketIoHandler final {
   public:
-    WebSocketIoHandler(kj::WebSocket& webSocket)
-        : webSocket(webSocket) {
+    WebSocketIoHandler(ExecutorNotifierPair isolateThreadExecutorNotifierPair, kj::WebSocket& webSocket)
+        : isolateThreadExecutor(kj::mv(isolateThreadExecutorNotifierPair.executor)),
+          incomingQueueNotifier(kj::mv(isolateThreadExecutorNotifierPair.notifier)),
+          webSocket(webSocket) {
       // Assume we are being instantiated on the InspectorService thread, the thread that will do
       // I/O for CDP messages. Messages are delivered to the InspectorChannelImpl on the Isolate thread.
-      incomingQueueNotifier = XThreadNotifier::create();
       outgoingQueueNotifier = XThreadNotifier::create();
     }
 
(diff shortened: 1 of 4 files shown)


src/workerd/io/worker.c++
```diff
@@ -2595,7 +2599,15 @@ private:
     // Message pumping promise that should be evaluated on the InspectorService
     // thread.
     kj::Promise<void> messagePump() {
-      return receiveLoop().exclusiveJoin(dispatchLoop()).exclusiveJoin(transmitLoop());
+      // Although inspector I/O must happen on the InspectorService thread (to make sure breakpoints
+      // don't block inspector I/O), inspector messages must be actually dispatched on the Isolate
+      // thread. So, we run the dispatch loop on the Isolate thread.
+      auto dispatchLoopPromise = isolateThreadExecutor->executeAsync([this]() {
```
danlapid, line 2605: nit: I would move the executeAsync into the dispatchLoop method, it's a bit clearer for future readers imho. 
otherwise lgtm

danlapid (approved):

mar-cf (approved):

src/workerd/io/worker.c++
harrishancock: I felt the same intuition, but wasn't confident the result was actually more readable, given the infamous footgunnery of lambda coroutines. The main choices that I see are:

1. Rename `dispatchLoop()` to `dispatchLoopImpl()` and make a new `dispatchLoop()` whose body is a call to `dispatchLoopImpl()` inside a lambda passed to `executeAsync()`.
2. Or, move `dispatchLoop()`'s body into a constexpr lambda which accepts `incomingQueueNotifier` and `channel` as parameters. (Alternatively, it could accept a `WebSocketIoHandler& self parameter`, and `self.` qualify the member accesses.)
3. Or, move `dispatchLoop()`'s body into a non-constexpr lambda which captures `this`, and wrap the lambda in `kj::coCapture()`.
4. Or, convert `dispatchLoop()` to a recursive `.then()` function, which I think would also require a separate `dispatchLoopImpl()`, since we only want to call `executeAsync()` a single time.
5. Or, the PR as written.

danlapid: Not sure what you mean, seeing as how `this` (which is just a ptr) should outlive the promise returned by executeAsync there should be no difference between the current implementation and 
```c++
// Must be called on the Isolate thread.
kj::Promise<void> dispatchLoop() {
  return isolateThreadExecutor->executeAsync([this]() {
    for (;;) {
      co_await incomingQueueNotifier->awaitNotification();
      KJ_IF_SOME(c, channel) {
        co_await c.dispatchProtocolMessages(this->incomingQueue);
      }
    }
  });
}
```

irvinebroque:
👏

src/workerd/io/worker.c++
harrishancock: Ah, you are right that that alternative actually is safe, because `executeAsync()` saves the lambda for the duration of the promise it returns, which is a guarantee I forgot it provides: https://github.com/capnproto/capnproto/blob/6cb55ac7d523c24047e6bbef0679027561f3c0d4/c%2B%2B/src/kj/async.h#L1040-L1041

To make sure we're on the same page, though: my understanding is that `this` being a pointer does not make this inherently safe. That is, my understanding is that it's the lifetime of the captured pointers/references (i.e., the lifetime of the lambda), not the lifetime of the pointed-to objects (e.g. `*this`), that matters.

My fear is that someone might see a `this`-capturing lambda coroutine and assume it is a generally safe pattern to use, and pass such a thing to, say, `Promise<T>::then()`, which does not preserve the lambda's lifetime. Because of this, I'd much prefer a blanket ban against non-constexpr, non-`coCapture()`ed coroutine lambdas, even if some usages like this one can be assumed to be safe.

harrishancock:
Added `needs-internal-pr` label until I figure out how to reconcile this with our internal codebase, which has no concept of an "Isolate thread". We might need multiple `attachInspector()` implementations. :/

harrishancock:
> We might need multiple attachInspector() implementations. :/

Obviously in retrospect, we already do. I modified the implementation which our internal codebase uses so that no internal change is required anymore (this change is effectively a no-op from its perspective), smoke tested profiling and breakpoint debugging once more on both codebases, and removed the `needs-internal-pr` label.

danlapid:
> > We might need multiple attachInspector() implementations. :/
> 
> Obviously in retrospect, we already do. I modified the implementation which our internal codebase uses so that no internal change is required anymore (this change is effectively a no-op from its perspective), smoke tested profiling and breakpoint debugging once more on both codebases, and removed the `needs-internal-pr` label.
> 
> Once @danlapid / @mar-cf give me a thumbs up on the latest diff, let's merge.

LGTM

irvinebroque:
If we ship this and cut a workerd release, then can get the ball rolling on testing out in Wrangler/Miniflare, both with the current fork of Chrome Devtools and an updated one.

mhart:
Is this in the latest release of workerd? I tried CPU profiling with whichever workerd is distributed with `wrangler 3.72.0` and it froze on me when I tried to take a profile. Both aspects of the chrome tools UI, as well as the workerd server itself stopped responding to requests. I couldn't find a way of successfully taking a profile.

irvinebroque:
I don't think it made it to `wrangler 3.72.0`

https://github.com/cloudflare/workers-sdk/pull/6490

Screenshot 2024-08-16 at 5 34 07 PM

mhart:
Ah, thanks @irvinebroque – I wonder if we should add the workerd version to `wrangler --version` output?

PR #2499, opened 2024-08-08
improve bazel builds for node

anonrig:
Moves Node declarations to a separate file to improve caching.


src/workerd/api/BUILD.bazel
```diff
@@ -71,9 +71,9 @@ wd_cc_library(
         ":html-rewriter",
         "//src/pyodide",
         "//src/pyodide:pyodide_extra_capnp",
-        "//src/workerd/io",
```
fhanau, line 74: We should not remove this line here – the target will still compile without it, but we should aim to follow an Include What You Use approach, even if we're currently far away from that.

src/workerd/io/BUILD.bazel
```diff
@@ -58,6 +58,7 @@ wd_cc_library(
         "//src/workerd/util:perfetto",
         "@capnp-cpp//src/kj/compat:kj-brotli",
         "@capnp-cpp//src/kj/compat:kj-gzip",
+        # TODO: Remove simdutf once api/crypto has it's own BUILD.bazel
```
fhanau, line 61: Nit: I feel hesitant about even adding a TODO here – I did try to modularizing the io target before, but achieving this will be lots of work based on the large amount of interdependencies we have currently. When we're doing this, it will enable much more cleanup than dropping simdutf here.

src/workerd/api/BUILD.bazel
anonrig: `src/workerd/api/node:node` also has a dependency of workerd/io. do you think I should still re-add it?

src/workerd/server/BUILD.bazel
```diff
@@ -123,7 +123,6 @@ wd_cc_library(
         "//src/workerd/io",
         "//src/workerd/io:worker-entrypoint",
         "//src/workerd/jsg",
-        "//src/workerd/util:perfetto",
```
fhanau, line 126: Nit: Same as above – we should follow IWYU here so that we don't depend on transitively pulling in util:perfetto through some other target here.

src/workerd/api/BUILD.bazel
```diff
@@ -71,9 +71,9 @@ wd_cc_library(
...
         "//src/workerd/jsg:rtti",
         "//src/workerd/server:workerd_capnp",
+        "//src/workerd/api/node:node",
```
fhanau, line 76: nit: Bazel supports `//src/workerd/api/node` as a short-form for this, the [buildifier](https://github.com/bazelbuild/buildtools/blob/main/buildifier/README.md) tool can be used to automatically detect such cases.

src/workerd/server/BUILD.bazel
fhanau: Similarly, we should add an explicit dependency on `//src/workerd/api/node` here.

fhanau (approved):
Thank you for doing this, I wasn't aware that separating node from io was possible! Left a number of suggestions here but these are largely stylistic concerns and this will be a win either way. Make sure to open a branch on the downstream repo before merging, it is possible that it will require a few changes now that the io target is smaller.

src/workerd/api/BUILD.bazel
fhanau: I think so – IWYU results in a longer list of dependencies but makes the code easier to maintain in the long run.
Imagine if (hypothetically) we ended up removing the node target here, recompiled and ended up with surprising error messages about headers not being found and then having to add one or more additional targets here that we used to transitively depend on, if IWYU is being followed we can avoid this by directly depending on all targets we need.
The motivation for IWYU within a C++ headers context is described [here](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md), but much of the same applies for bazel targets.

mikea (approved):
thanks for doing this, @fhanau has left excellent suggestions.

jasnell (approved):
LGTM with Felix's suggestions

PR #2500, opened 2024-08-08
startuptime is now reported when it isn't exceeded.

danlapid:
limitenforcer's enterStartup* functions now receive a reference to a OneOf an exception or a duration. 
Which can be populated by the elapsed duration if the limit was not exceeded.


dom96 (commented):
Should the base of this PR be different? It looks like it contains @anonrig's commit which is `main` so shouldn't show up here

danlapid:
> Should the base of this PR be different? It looks like it contains @anonrig's commit which is `main` so shouldn't show up here

Hmm, you're right, that wasn't here yesterday though.
I'll try to rebase

src/workerd/io/worker.c++
```diff
@@ -216,69 +218,74 @@ void reportStartupError(
...
-          KJ_FAIL_REQUIRE("script startup threw exception");
+          KJ_LOG(ERROR, "script startup exceeded resource limits", id, ex);
+          kj::throwFatalException(kj::cp(ex));
         }
-      });
-    } else {
-      kj::throwFatalException(kj::cp(permanentException.emplace(
-          KJ_EXCEPTION(FAILED, "returned empty handle but didn't throw exception?", id))));
+      }
+      KJ_CASE_ONEOF_DEFAULT {
```
dom96, line 248: Might be better to be specific here and use `KJ_CASE_ONEOF` for the other possible type (same for other KJ_CASE_ONEOF_DEFAULT uses below). This way if the OneOf type gets another variant we'll get a compile-time error rather than this code path just being used implicitly.

src/workerd/io/worker.c++
```diff
@@ -1572,6 +1589,14 @@ Worker::Worker(kj::Own<const Script> scriptParam,
...
             }
 
+            KJ_IF_SOME(s, startupTime) {
+              KJ_SWITCH_ONEOF(limitErrorOrTime) {
+                KJ_CASE_ONEOF(startupTimeElapsed, kj::Duration) {
+                  s = startupTimeElapsed;
+                }
+                KJ_CASE_ONEOF_DEFAULT { }
+              }
+            }
```
dom96, line 1599: It seems this is the only place where the duration part of the `limitErrorOrTime` is used. It makes me wonder whether a separate argument for this wouldn't be better, unless I'm missing something it would avoid a lot of the refactoring that you needed to do here (which largely is just changing from `KJ_IF_SOME` to `KJ_CASE_ONEOF`).

danlapid: I thought about it in the beginning but I decided against it.
Having a OneOf is clearer in tems of what to expect from enterStartup*.
The result will be EITHER an time exceeded error or the time elapsed, never both, never neither.
I realize it required a bit of refactoring to do this change but it makes a lot more sense semantically.

src/workerd/io/worker.c++
danlapid: I changed some of them now.
Some of them make sense as a ONEOF instead of default but the ones who specifically check for an exception shouldn't care if another value is added.
I imagine this as more of a Result type where we changed from `Result<(), Error>` to `Result<Duration, Error>`, we might in the future change to `Result<OneOf<Duration, ??>, Error>`. 
So in some cases it's just more interesting to check whether or not we failed rather than what's the result's success value.

src/workerd/io/worker.c++
dom96: Ahh, didn't realise it was an argument used as an output. That makes sense.

dom96 (approved):

jasnell:
Marking with `needs-internal-pr` to ensure that this is tested internally before merging. If the internal CI has already been done and passed, the label can be removed.

PR #2503, opened 2024-08-09
Use new "response stream disconnected" status for disconnections during proxying

jclee:
Add new responseStreamDisconnected request status, to be used when an exception is thrown due to disconnection during deferred proxying, which is common when the client or server hangs up early, particularly when proxying websockets.  Should allow us to differentiate these cases from those where JS is still running, such that a thrown exception could still be caught and handled by the worker.

Also, add instrumentation to report failure during deferred proxying.

### DIFF
--- a/src/workerd/io/observer.h
+++ b/src/workerd/io/observer.h
@@ -56,11 +56,20 @@ public:
   // the prewarm metric will be incremented.
   virtual void setIsPrewarm() {}
 
+  // Describes the source of a failure
+  enum class FailureSource: uint8_t {
+    // Failure occurred during deferred proxying
+    DEFERRED_PROXY,
+
+    // Failure occurred elsewhere
+    OTHER,
+  };
+
   // Report that the request failed with the given exception. This only needs to be called in
   // cases where the wrapper created with wrapWorkerInterface() wouldn't otherwise see the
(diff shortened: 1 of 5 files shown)


justin-mp (approved):
The code changes look good to me (in conjunction to the internal PRs to the schema as needed by outcome.capnp), but I'm surprised that there's no test change for workerd.  Is this because RequestObserver::reportFailure in workerd is an empty implementation, and, as such, doesn't change with this code?

MellowYarker (approved):

jclee:
> The code changes look good to me (in conjunction to the internal PRs to the schema as needed by outcome.capnp), but I'm surprised that there's no test change for workerd. Is this because RequestObserver::reportFailure in workerd is an empty implementation, and, as such, doesn't change with this code?

Yeah, as far as I can tell, we don't really have test coverage in workerd for `RequestObserver` method invocation, yet (and similarly for a few other interfaces that only have internal implementations).

Most workerd tests seem to be `.wd-test` integration tests that don't yet have a way to introspect `RequestObserver` calls.  It might be possible to add some sort of test-only API to allow `.wd-test`s to detect the calls -- or to add unit or integration tests in `*-test.c++` files?  But for this task, I ended up just augmenting the existing internal integration tests, which were already instrumented detect the behavior changes.

PR #2504, opened 2024-08-09
Tidy up pyodide bundle API

hoodmane:
This adds a Mutex around the bundleReaders map and makes `getPyodideBundle` return a Maybe<bundle> rather than asserting that the bundle is defined. This means we can get rid of `hasPyodideBundle` in favor of checking whether `getPyodideBundle` returns `none`.

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -8,20 +8,21 @@
 namespace workerd::api::pyodide {
 
 // singleton that owns bundle
-kj::TreeMap<kj::String, kj::Own<capnp::FlatArrayMessageReader>> bundleReaders;
 
-bool hasPyodideBundle(kj::StringPtr version) {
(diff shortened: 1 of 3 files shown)


dom96 (approved):

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -8,20 +8,20 @@
 namespace workerd::api::pyodide {
 
 // singleton that owns bundle
-kj::TreeMap<kj::String, kj::Own<capnp::FlatArrayMessageReader>> bundleReaders;
+static kj::MutexGuarded<kj::TreeMap<kj::String, kj::Tuple<kj::Own<capnp::FlatArrayMessageReader>, jsg::Bundle::Reader>>> bundleReaders;
```
kentonv, line 11: `kj::Tuple` should only be used in template metaprogramming. Here, you should instead declare a two-element struct and use that. It results in code that's much easier to read, since the members are named.

PR #2505, opened 2024-08-09
Add formatting infrastructure to the project

### LINKED ISSUE #2507: Code formatting: C++
Format C++ code
Definition of Done:
1. documentation
2. optional commit hooks
3. proper blame handling
4. CI checks

danlapid:
Fixes #2507

Add infrastructure to format c++ files with clang-format.
Added:
1. clang-format file matching internal one
2. formatting tool based on python for cross-platform
3. pre-commit and pre-push hooks for formatting validation
4. Lint CI action.

In terms of CR workflow here's what I was thinking:
Reviewing the entire CR is going to be difficult which is why it's split into 3 distinct commits:
1. Establish the framework for reformatting.
3. Plain format.py run
4. add .git-blame-ignore-revs file with format commit's hash.

Seeing as how this CR will probably take a while I will not try to resolve conflicts during the CR. Instead whenever we finish the CR I will revert to 
1. commit 1
2. rebase from main
3. apply format
4. add .git-blame-ignore-revs
5. merge without further CR

At the very least should have @jp4a50 approval before merge.


danlapid:
In terms of CR workflow here's what I was thinking:
Reviewing the entire CR is going to be difficult which is why it's split into 4 distinct commits:
1. Establish the framework for reformatting.
2. Actual code changes required for the reformatting to succeed.
3. Plain format.py run
4. add .git-blame-ignore-revs file with 3's commit hash.

Seeing as how this CR will probably take a while I will not try to resolve conflicts during the CR. Instead whenever we finish the CR I will revert to 
1. commit 2
2. rebase from main
3. apply format
4. add .git-blame-ignore-revs
5. merge without further CR

githooks/pre-commit
```diff
@@ -10,3 +10,43 @@ then
   echo -e "To commit anyway, use --no-verify\n"
   exit 1
 fi
+
+clang_format_check() {
```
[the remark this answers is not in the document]
danlapid: So generally speaking githooks are optional imho, they're nice that they're there but you don't have to enable them.
Every good git hook is enforced by a CI step anyway (In the PR I added a lint CI stage).

tools/cross/format.py
```diff
@@ -0,0 +1,135 @@
+import argparse
```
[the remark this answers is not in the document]
danlapid: I would like that eventually but I'd rather keep this PR minimal to C++.

tools/cross/format.py
```diff
@@ -0,0 +1,135 @@
...
+        if result.returncode != 0:
+            logging.error("Code has lint. Fix with: make format")
+            exit(1)
+    else:
+        subprocess.run(["clang-format", "-i"] + files, check=False)
+
+
+def git_get_modified_files(
+    target: str, source: typing.Optional[str], staged: bool
+) -> typing.List[str]:
```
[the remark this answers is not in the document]
danlapid: I think many people still use python3.8 and I'd rather not drop support for that just for prettier type hints.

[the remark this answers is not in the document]
anonrig: I see. I think we need a .python-version file in the root, and be explicit about what version we support

githooks/pre-commit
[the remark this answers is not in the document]
hoodmane: I like the githooks personally.

tools/cross/format.py
```diff
@@ -0,0 +1,135 @@
+import argparse
+import os
+import re
+import subprocess
+import logging
+import typing
```
hoodmane, line 6: Nit: you should use `from` imports for type hints and such:
```suggestion
from argparse import ArgumentParser, Namespace
from typing import List, Optional
```

fhanau:
Thank you for this!

On updating LLVM: Leaving it at 15 for now has been a conscious choice: My philosophy has been that it should be possible to build workerd using the LLVM version available in the default APT repo in the latest stable release of Ubuntu and Debian, so that developers can build workerd without having to build LLVM themselves or install it from an alternative repository like https://apt.llvm.org/ which is used in the CI build. For Debian (bookworm) that's currently 16 (also available on the earlier Debian Bullseye), for Ubuntu 22.04 it is LLVM 15, the recently introduced Ubuntu 24.04 supports LLVM 18. If we upgrade the version we use to build on CI to 18, we can easily introduce changes that break the build with lower versions.
With the V8 12.8 update (#2441, should land pretty soon), we will have to update this to LLVM 16 anyway, which may introduce some headaches for internal developers that are on Ubuntu 22.04-based systems.

What we can do is to require LLVM 18 and document that installing it may require additional developer actions, or we can update the version we use on CI to 18, but leave the documented minimum version as is and just state that LLVM18 is recommended and add a CI job that uses LLVM16 so that it doesn't break. I would prefer to address compiler upgrades separately from this PR TBH, for now we can just say that clang-format-18 is required for linting.

danlapid:
> Thank you for this!
> 
> On updating LLVM: Leaving it at 15 for now has been a conscious choice: My philosophy has been that it should be possible to build workerd using the LLVM version available in the default APT repo in the latest stable release of Ubuntu and Debian, so that developers can build workerd without having to build LLVM themselves or install it from an alternative repository like [apt.llvm.org](https://apt.llvm.org/) which is used in the CI build. For Debian (bookworm) that's currently 16 (also available on the earlier Debian Bullseye), for Ubuntu 22.04 it is LLVM 15, the recently introduced Ubuntu 24.04 supports LLVM 18. If we upgrade the version we use to build on CI to 18, we can easily introduce changes that break the build with lower versions. With the V8 12.8 update (#2441, should land pretty soon), we will have to update this to LLVM 16 anyway, which may introduce some headaches for internal developers that are on Ubuntu 22.04-based systems.
> 
> What we can do is to require LLVM 18 and document that installing it may require additional developer actions, or we can update the version we use on CI to 18, but leave the documented minimum version as is and just state that LLVM18 is recommended and add a CI job that uses LLVM16 so that it doesn't break. I would prefer to address compiler upgrades separately from this PR TBH, for now we can just say that clang-format-18 is required for linting.

Yeah sure, I thought it'd be nice to upgrade our llvm version but I don't mind reverting this change and just requiring clang-format-18 for format

.github/workflows/release.yml
```diff
@@ -88,14 +88,15 @@ jobs:
...
+          sudo ./llvm.sh 18
+          sudo apt-get install -y --no-install-recommends clang-18 lld-18 libunwind-18 libc++abi1-18 libc++1-18 libc++-18-dev
+          echo "build:linux --action_env=CC=/usr/lib/llvm-18/bin/clang --action_env=CXX=/usr/lib/llvm-18/bin/clang++" >> .bazelrc
+          echo "build:linux --host_action_env=CC=/usr/lib/llvm-18/bin/clang --host_action_env=CXX=/usr/lib/llvm-18/bin/clang++" >> .bazelrc
       - name: Setup macOS
         if: runner.os == 'macOS'
         run: |
-          sudo xcode-select -s "/Applications/Xcode_15.1.app"
+          brew install llvm@18
+          export PATH="$(brew --prefix llvm@18)/bin/bin:$PATH"
```
fhanau, line 99: We have always built workerd with the Xcode-provided clang/LLVM toolchain. Apple LLVM generally has some differences from mainline LLVM, and with the way toolchain resolution works in https://github.com/bazelbuild/apple_support I think that bazel will always build using Xcode's LLVM even if this is added to PATH. Building with a different LLVM may come with unexpected consequences just based on the differences Apple Clang has, currently Apple Clang 15 is based on LLVM 16 which feels recent enough.

danlapid: I have been building workerd locally with brew llvm@18 with no issues but as with the previous comment, I'll just revert this entire change.

src/workerd/jsg/function.h
```diff
@@ -519,4 +519,9 @@ template <> constexpr bool isGcVisitor<GcVisitor&>() { return true; }
     } \
   })
 
+// Clang format encounters a segfault when JSG_VISITABLE_LAMBDA is called from inside of a
```
fhanau: Nit: JSG_VISITABLE_LAMBDA is not a function and can't be called per se, I think something like "JSG_VISITABLE_LAMBDA  is used[...]" or "a JSG_VISITABLE_LAMBDA is declared" work better here

.github/workflows/lint.yml
```diff
@@ -0,0 +1,23 @@
+name: Tests
+
+on:
+  pull_request:
+    paths:
+      - src/**
```
fhanau, line 6: Nit: let's also add `.clang-format` itself here, if we end up changing the clang-format definition and CI doesn't check if there's lint for that commit it could easily lead to confusion.

tools/cross/format.py
```diff
@@ -0,0 +1,135 @@
+import logging
```
fhanau, line 1: Nit: add something like `#!/usr/bin/env python3` here and add execute permissions to the file to allow running it as just `./tools/cross/format.py` (i.e. without explicitly listing python in the command)

tools/cross/format.py
```diff
@@ -0,0 +1,135 @@
...
+def clang_format(files: List[str], check=False):
+    if not files:
+        logging.info("No changes to format")
+        exit(0)
+    if check:
+        result = subprocess.run(
+            ["clang-format", "--dry-run", "--Werror"] + files, check=False
+        )
+        if result.returncode != 0:
+            logging.error("Code has lint. Fix with: make format")
```
fhanau, line 84: Replace `make format` => `python3 ./tools/cross/format.py`

fhanau:
Unfortunately format checking fails for me with `Configuration file(s) do(es) not support Objective-C: [...]/workerd/.clang-format` (macOS, clang-format 18.1.8 installed via homebrew). Looks like clang-format is mistaking some files to be Objective-C instead of C++?

danlapid:
> Unfortunately format checking fails for me with `Configuration file(s) do(es) not support Objective-C: [...]/workerd/.clang-format` (macOS, clang-format 18.1.8 installed via homebrew). Looks like clang-format is mistaking some files to be Objective-C instead of C++?

Yea that warning is printed for about 5 files where we have embedded typescript in macros but macro expansion happens before syntax parsing in clang-format so it gets confused.
It's not an error but a warning.
We can make it go away by marking objective c files as "don't format" but there isn't any easy fix to make clang format be able to format these files.
It would require quite a massive change in the way it works.
We can revisit it later but I don't think it should block formatting the other hundreds of files we have.

tools/cross/format.py
danlapid: I added the `#!usr/bin/env python3` but I don't have permission to add execute permissions to files.

fhanau: > I added the `#!usr/bin/env python3` but I don't have permission to add execute permissions to files.

You don't have permission as in it violates development guidelines for workerd or as in the system doesn't let you? Does `chmod +x tools/cross/format.py` not work?

.clang-format
```diff
@@ -1,10 +1,100 @@
...
+# because existing users happened to include a dependency for it), one could use the following
+# order:
+#   local folder files
+#   project files
+#   3p dependency includes
+#   standard language headers (put kj/ here?)
+#   system headers
+# While this would be ideal it's also important to note that this isn't the (non-documented) style
+# that KJ uses, so it may be worth documenting the style & making it consistent.
+SortIncludes: false
```
npaun, line 26: This might be a good time to define IncludeCategories and turn this flag on.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories

tools/cross/format.py
danlapid: Chmod works, when I try to push I get the error.

fhanau:
> > Unfortunately format checking fails for me with `Configuration file(s) do(es) not support Objective-C: [...]/workerd/.clang-format` (macOS, clang-format 18.1.8 installed via homebrew). Looks like clang-format is mistaking some files to be Objective-C instead of C++?
> 
> Yea that warning is printed for about 5 files where we have embedded typescript in macros but macro expansion happens after syntax parsing and language recognition in clang-format so it gets confused. It's not an error but a warning. We can make it go away by marking objective c files as "don't format" but there isn't any easy fix to make clang format be able to format these files. It would require quite a massive change in the way it works. We can revisit it later but I don't think it should block formatting the other hundreds of files we have.

This doesn't look like warning to me, or at least I get a non-zero return code which will cause pushes to fail. Perhaps this is caused by `--Werror` being enabled, does running `python3 tools/cross/format.py --check` result in a different return code for you?

danlapid:
> > > Unfortunately format checking fails for me with `Configuration file(s) do(es) not support Objective-C: [...]/workerd/.clang-format` (macOS, clang-format 18.1.8 installed via homebrew). Looks like clang-format is mistaking some files to be Objective-C instead of C++?
> > 
> > 
> > Yea that warning is printed for about 5 files where we have embedded typescript in macros but macro expansion happens after syntax parsing and language recognition in clang-format so it gets confused. It's not an error but a warning. We can make it go away by marking objective c files as "don't format" but there isn't any easy fix to make clang format be able to format these files. It would require quite a massive change in the way it works. We can revisit it later but I don't think it should block formatting the other hundreds of files we have.
> 
> This doesn't look like warning to me, or at least I get a non-zero return code which will cause pushes to fail. Perhaps this is caused by `--Werror` being enabled, does running `python3 tools/cross/format.py --check` result in a different return code for you?

I've changed quite a lot since morning, try pulling and retrying 😄

tools/cross/format.py
fhanau: Ok that's weird, we certainly have scripts with execute permissions in the repo, e.g. `tools/unix/workspace-status.sh`. I think this is non-blocking overall but what error message do you get?

.clang-format
fhanau: I disagree – we don't have this set up internally either and as suggested by the comments getting include sorting right is non-trivial. IMHO formatting code correctly is more important than formatting includes correctly and getting includes right is likely to slow landing this down

fhanau:
> > > > Unfortunately format checking fails for me with `Configuration file(s) do(es) not support Objective-C: [...]/workerd/.clang-format` (macOS, clang-format 18.1.8 installed via homebrew). Looks like clang-format is mistaking some files to be Objective-C instead of C++?
> > > 
> > > 
> > > Yea that warning is printed for about 5 files where we have embedded typescript in macros but macro expansion happens after syntax parsing and language recognition in clang-format so it gets confused. It's not an error but a warning. We can make it go away by marking objective c files as "don't format" but there isn't any easy fix to make clang format be able to format these files. It would require quite a massive change in the way it works. We can revisit it later but I don't think it should block formatting the other hundreds of files we have.
> > 
> > 
> > This doesn't look like warning to me, or at least I get a non-zero return code which will cause pushes to fail. Perhaps this is caused by `--Werror` being enabled, does running `python3 tools/cross/format.py --check` result in a different return code for you?
> 
> I've changed quite a lot since morning, try pulling and retrying 😄

`ERROR:root:clang-format version must be at least 19.0.0`
LLVM 19 is scheduled to be published on September 3, but historically the new version hasn't always been available on macOS Homebrew until a few weeks afterwards. I think this will worsen DX on macOS, is supporting clang-format 18 not salvageable?

.clang-format
danlapid: Why is getting include sorting right non-trivial though?
What's wrong with something like:
```
SortIncludes: true
IncludeBlocks: Regroup
IncludeCategories:
  # c++ system headers
  - Regex: <[a-zA-Z0-9_]+>
    Priority: 5
  # kj/capnp headers
  - Regex: <(kj|capnp)/.+>
    Priority: 4
  # 3rd party headers
  - Regex: <.+>
    Priority: 3
  # workerd headers
  - Regex: <workerd/.+>
    Priority: 2
  # local headers
  - Regex: '".*"'
    Priority: 1
```

danlapid:
> worsen DX on macOS, is supporting



> > > > > Unfortunately format checking fails for me with `Configuration file(s) do(es) not support Objective-C: [...]/workerd/.clang-format` (macOS, clang-format 18.1.8 installed via homebrew). Looks like clang-format is mistaking some files to be Objective-C instead of C++?
> > > > 
> > > > 
> > > > Yea that warning is printed for about 5 files where we have embedded typescript in macros but macro expansion happens after syntax parsing and language recognition in clang-format so it gets confused. It's not an error but a warning. We can make it go away by marking objective c files as "don't format" but there isn't any easy fix to make clang format be able to format these files. It would require quite a massive change in the way it works. We can revisit it later but I don't think it should block formatting the other hundreds of files we have.
> > > 
> > > 
> > > This doesn't look like warning to me, or at least I get a non-zero return code which will cause pushes to fail. Perhaps this is caused by `--Werror` being enabled, does running `python3 tools/cross/format.py --check` result in a different return code for you?
> > 
> > 
> > I've changed quite a lot since morning, try pulling and retrying 😄
> 
> `ERROR:root:clang-format version must be at least 19.0.0` LLVM 19 is scheduled to be published on September 3, but historically the new version hasn't always been available on macOS Homebrew until a few weeks afterwards. I think this will worsen DX on macOS, is supporting clang-format 18 not salvageable?

Sadly it seems that clang-format-18 spews out different results in my updated linux environment on macOS and on github actions.

fhanau:
> Sadly it seems that clang-format-18 spews out different results in my updated linux environment on macOS and on github actions.

That's unfortunate – does that only affect a few files or is it a broad issue? Perhaps we can manage this by manually excluding a few files until 19 comes out, or add
```
// clang-format off
// clang-format on
```
sections where appropriate.

danlapid:
> > Sadly it seems that clang-format-18 spews out different results in my updated linux environment on macOS and on github actions.
> 
> That's unfortunate – does that only affect a few files or is it a broad issue? Perhaps we can manage this by manually excluding a few files until 19 comes out, or add
> 
> ```
> // clang-format off
> // clang-format on
> ```
> 
> sections where appropriate.

Got it sorted with clang-format-18, I edited the comment above.

.clang-format
fhanau: The categories here look reasonable, but if b.c++ includes a.h and b.h, does this ensure that b.h is listed first to check that b.h can be included without depending on other headers? We can try checking includes, my concern is that this will delay landing this PR if there's any complications with include changes, it's already pretty expansive.

danlapid: So for any <filename>.c++ the first include will always be "<filename>.h", with the above config.
And any header file should always be include-able without first including other headers, if that's not true then said head file should be fixed.

npaun: Here's what clang's manual says about this:

> The main header for a source file automatically gets category 0. so that it is generally kept at the beginning of the #includes (https://llvm.org/docs/CodingStandards.html#include-style).

fhanau:
> Got it sorted with clang-format-18, I edited the comment above.

Running the script works on macOS now. Unfortunately, it takes 16s in total on M1 which would slow down the commit process after freezing on `src/workerd/jsg/iterator.h` – when excluding that file this goes down to 4s which is still quite some overhead but perhaps acceptable. Maybe we can manually exclude that file?

npaun (approved):

danlapid:
> overhead but perhaps acceptable. Maybe

The pre-commit hook only runs on staged files, it shouldn't slow down commits at all unless you editted that specific files.
If that were a problem I would probably multithread the format.py script but on a regular commit with just a few c++ files the format is almost instant.

fhanau (approved):

.clang-format
danlapid: To summarise here, I think adding header include sorting is a huge +.
But I do understand Felix's worries.
I will defer to @jp4a50 whether we should add it to this PR or a later one. Either way the above configuration should be fitting and can be used at a later date.

tools/cross/format.py
```diff
@@ -0,0 +1,139 @@
...
+    git_parser.add_argument(
+        "--staged",
+        help="consider files with staged modifications only",
+        action="store_true",
+        default=False,
+    )
+    options = parser.parse_args()
+    if options.subcommand == "git":
+        if options.staged:
+            if options.source is not None or options.target != "HEAD":
```
hoodmane, line 48: Nit:
```suggestion
    if (
        options.subcommand == "git"
        and options.staged
        and (options.source is not None or options.target != "HEAD")
    ):
```

tools/cross/format.py
```diff
@@ -0,0 +1,139 @@
...
+                    "--staged cannot be used with --source or --target; use --staged with --source=HEAD"
+                )
+                exit(1)
+    return options
+
+
+def check_clang_format() -> bool:
+    try:
+        # Run clang-format with --version to check its version
+        output = subprocess.check_output([CLANG_FORMAT, "--version"]).decode("utf-8")
```
hoodmane, line 59: Pass an encoding argument rather than decoding the output:
```suggestion
        output = subprocess.check_output([CLANG_FORMAT, "--version"], encoding="utf-8")
```

tools/cross/format.py
```diff
@@ -0,0 +1,139 @@
...
+def find_cpp_files(dir_path) -> List[str]:
+    files = []
+    for root, _, filenames in os.walk(dir_path):
+        for filename in filenames:
+            if filename.endswith((".c++", ".h")):
+                files.append(os.path.join(root, filename))
+    return files
+
+
+def clang_format(files: List[str], check=False):
```
hoodmane, line 79:
```suggestion
def clang_format(files: List[str], check: bool = False):
```

tools/unix/apply-big-move.sh
hoodmane, line 1: Fancy!

tools/unix/find-python3.sh
```diff
@@ -0,0 +1,21 @@
...
+
+  # If not found, search for python
+  if [ -z "$python" ]; then
+    python=$(which python 2>/dev/null)
+    if [ -z "$python" ]; then
+      local python_version=$($python -V 2>&1 | head -n 1 | cut -d ' ' -f 2- | cut -d '.' -f1)
+      if [ "$python_version" != "3" ]; then
+        unset python
+      fi
+    fi
```
hoodmane, line 14: Presumably this inner check should be testing for the opposite?

hoodmane (approved):
Thanks @danlapid, I made some minor comments.

.clang-format
danlapid: To enable sortincludes with the above config and have workerd compile and pass all tests the following changes need to be made:
src/workerd/jsg/jsg.h: Add clang format off around the includes in the end.

// clang-format off
// These two includes are needed for the JSG type glue macros to work.
#include "buffersource.h"
#include "modules.h"
#include "resource.h"
#include "dom-exception.h"
#include "struct.h"
#include "promise.h"
#include "function.h"
#include "iterator.h"
#include "jsvalue.h"
// clang-format on


src/workerd/api/memory-cache.h: remove the include to function.h
delete `#include <workerd/jsg/function.h>`

jasnell (approved):

tools/unix/apply-big-move.sh
```diff
@@ -0,0 +1,181 @@
...
+    fail "Cannot find python3. Run install python3 first."
+  fi
+}
+
+apply_big_move() {
+  echo "Applying format..."
+
+  source "$(dirname -- $BASH_SOURCE)/find-python3.sh"
+  PYTHON_PATH=$(get_python3)
+  $PYTHON_PATH "$(dirname -- $BASH_SOURCE)/../cross/format.py"
```
jp4a50, line 44: FWIW this is a slightly different approach taken than the internal big move script in that you are running against only those files changed wrt. the HEAD commit rather than all files which should be formatted.

The intention of the original approach was that you can use this script to actually apply the big move in the first place by running `apply-big-move.sh --apply` so you can be sure you are doing the same (idempotent) thing when applying the big move in the first place and then applying it to each branch's commit.

That being said, assuming you run clang-format on (at least) the set of files that the `format.py` script might format, this should still work fine.

tools/unix/apply-big-move.sh
```diff
@@ -0,0 +1,181 @@
...
+
+# Ensure that if the file changes while bash is reading it, bash won't go off the rails. Normally,
+# bash reads one line from the file at a time and then executes it before reading the next.
+# Wrapping the whole script body in a function ensures that bash buffers it in memory. Placing
+# `exit 0` on the same line ensures that bash won't attempt to read anything more after the
+# function returns.
+#
+# This ensures that if further changes are made to this file before the Big Move actually happens,
+# then the rebase command that rebases to the commit before the Big Move won't confuse bash.
+main "$@"; exit 0
```
jp4a50, line 181: I skimmed over this, assuming that everything below `apply_big_move()` is unchanged. LMK if not.

jp4a50:
One thing that I did when I did the internal formatting was to go through the diff and find cases where clang-format made the code much less readable so I could selectively turn off clang-format for certain lines. I won't block on this because it was incredibly painstaking but if you fancy killing about half a day with some monotonous work then it would be an added bonus :sweat_smile:

jp4a50 (approved):
GG

tools/unix/apply-big-move.sh
danlapid: Well, the format.py script is the only way I used to clang-format so it should match :)

jp4a50: Ohh, I just realized that you actually changed the semantics of the original script when porting it - you added the `--git` option to the script when porting it. This makes a lot more sense now.

hoodmane:
I'm having some trouble with `apply-big-move`:
1. it calls `exec ./apply-big-move.sh` so if you called it via eg `./tools/unix/apply-big-move.sh` then it won't work. I tried cding into `tools/unix` which got me a bit further.
2. have to install clang-format-18 and ensure that there is a symlink from `clang-format` to `clang-format-18` on your path. It checks whether `clang-format --version` is 18 after starting a rebase and doesn't clean it up when it quits, so you have to `git reset --hard my-branch` first.
3. Then I see:
    ```
    Fetching tags...
    Rewriting commit 63b43725: Add formatting infrastructure to the project.
    Updated 1 path from 45fe9bd9
    Applying format...
    Changes applied successfully.
    [detached HEAD ee5ebcdf5] Add formatting infrastructure to the project.
     Author: Dan Lapid <dan.lapid@gmail.com>
     Date: Fri Aug 9 15:27:49 2024 +0100
     1 file changed, 0 insertions(+), 0 deletions(-)
     mode change 100644 => 100755 tools/unix/apply-big-move.sh
    Rewriting commit 37bed02d: Add Python development compatibility flag for tests
    Updated 0 paths from e34e2818
    Applying format...
    Changes applied successfully.
    fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
    ```
    And the latest commit is `d916b2af Add formatting infrastructure to the project.` Maybe invoking from within the `tools/unix` 
    directory isn't good?
4. I switched to `bash tools/unix/apply-big-move.sh` which gets further and actually formats a bunch of stuff. This fails on `FileNotFoundError: [Errno 2] No such file or directory: 'node_modules/.bin/prettier'`
5. Run `npm i prettier` and try again.
6. Now it does a lot more stuff but still fails with 
    ```
    fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
    ```
    But now HEAD is detached on a commit that makes my changes on top of `d916b2af`! So all I have to do is manually update my original branch with e.g., `git switch -C orig-branch` and I'm good to go.

PR #2511, opened 2024-08-09
Refine the ExternalMemoryAdjuster API more

jasnell:
Reflecting additional feedback from Kenton

### DIFF
--- a/src/workerd/jsg/jsg.c++
+++ b/src/workerd/jsg/jsg.c++
@@ -278,27 +278,60 @@ JsSymbol Lock::symbolAsyncDispose() {
   return IsolateBase::from(v8Isolate).getSymbolAsyncDispose();
 }
 
-void Lock::adjustExternalMemory(ssize_t amount) {
-  v8Isolate->AdjustAmountOfExternalAllocatedMemory(static_cast<int64_t>(amount));
+void ExternalMemoryAdjustment::maybeDeferAdjustment(v8::Isolate* isolate, size_t amount) {
+  if (isolate == nullptr) return;
+  if (v8::Locker::IsLocked(isolate)) {
+    isolate->AdjustAmountOfExternalAllocatedMemory(static_cast<int64_t>(-amount));
+  } else {
+    // Otherwise, if we don't have the isolate locked, defer the adjustment to the next
+    // time that we do.
+    auto& jsgIsolate = *reinterpret_cast<IsolateBase*>(isolate->GetData(0));
+    jsgIsolate.deferExternalMemoryDecrement(static_cast<int64_t>(amount));
+  }
 }
 
-namespace {
-struct ExternalMemoryAdjuster {
-  int64_t size;
-  v8::Isolate* isolate;
-  ExternalMemoryAdjuster(v8::Isolate* isolate, size_t size)
-      : size(static_cast<int64_t>(size)), isolate(isolate) {
-    isolate->AdjustAmountOfExternalAllocatedMemory(size);
-  }
+ExternalMemoryAdjustment::ExternalMemoryAdjustment(v8::Isolate* isolate, size_t amount)
+    : amount(amount), isolate(isolate) {
+  KJ_DASSERT(isolate != nullptr);
+  isolate->AdjustAmountOfExternalAllocatedMemory(amount);
+}
 
-  ~ExternalMemoryAdjuster() noexcept(false) {
-    isolate->AdjustAmountOfExternalAllocatedMemory(-size);
-  }
-};
-}  // namespace
+ExternalMemoryAdjustment::ExternalMemoryAdjustment(ExternalMemoryAdjustment&& other) :
+    amount(other.amount), isolate(other.isolate) {
+  other.amount = 0;
+  other.isolate = nullptr;
+}
 
-kj::Own<void> Lock::getExternalMemoryAdjuster(size_t amount) {
-  return kj::heap<ExternalMemoryAdjuster>(v8Isolate, amount);
+ExternalMemoryAdjustment& ExternalMemoryAdjustment::operator=(ExternalMemoryAdjustment&& other) {
+  // If we currently have an amount, adjust it back to zero.
+  // In the case we don't have the isolate lock here, the adjustment
+  // will be deferred until the next time we do.
+  if (amount > 0) maybeDeferAdjustment(isolate, amount);
+  amount = other.amount;
+  isolate = other.isolate;
+  other.amount = 0;
+  other.isolate = nullptr;
+  return *this;
+}
+
+ExternalMemoryAdjustment::~ExternalMemoryAdjustment() noexcept(false) {
+  if (amount != 0) {
+    maybeDeferAdjustment(isolate, amount);
+  }
+}
+
+void ExternalMemoryAdjustment::adjust(Lock& js, ssize_t amount) {
+  amount = kj::max(amount, -static_cast<ssize_t>(this->amount));
+  this->amount += amount;
+  isolate->AdjustAmountOfExternalAllocatedMemory(amount);
+}
+
+void ExternalMemoryAdjustment::set(Lock& js, size_t amount) {
+  adjust(js, amount - this->amount);
+}
+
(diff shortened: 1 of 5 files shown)


src/workerd/jsg/jsg.c++
```diff
@@ -278,27 +278,57 @@ JsSymbol Lock::symbolAsyncDispose() {
   return IsolateBase::from(v8Isolate).getSymbolAsyncDispose();
 }
 
-void Lock::adjustExternalMemory(ssize_t amount) {
-  v8Isolate->AdjustAmountOfExternalAllocatedMemory(static_cast<int64_t>(amount));
+void ExternalMemoryAdjuster::maybeDeferAdjustment(v8::Isolate* isolate, int64_t amount) {
+  if (isolate == nullptr || v8::Locker::IsLocked(isolate)) {
```
anonrig, line 282:
```suggestion
  if (isolate != nullptr || v8::Locker::IsLocked(isolate)) {
```

kentonv: I think it actually needs to be:

```suggestion
  if (isolate == nullptr) return;  // moved away
  if (v8::Locker::IsLocked(isolate)) {
```

src/workerd/jsg/jsg.h
```diff
@@ -2042,6 +2042,38 @@ class JsMessage;
...
+  void decrement(Lock&, int64_t amount);
+
+  ~ExternalMemoryAdjuster();
+
+  inline v8::Isolate* getIsolate() const { return isolate; }
+
+  inline int64_t getAmount() const { return amount; }
+
+private:
+  int64_t amount = 0;
```
kentonv, line 2071: This would never actually be negative, would it?

src/workerd/jsg/jsg.h
```diff
@@ -2042,6 +2042,38 @@ class JsMessage;
...
+public:
+  ExternalMemoryAdjuster() = default;
+  ExternalMemoryAdjuster(v8::Isolate* isolate, int64_t amount);
+  ExternalMemoryAdjuster(ExternalMemoryAdjuster&& other);
+  ExternalMemoryAdjuster& operator=(ExternalMemoryAdjuster&& other);
+  KJ_DISALLOW_COPY(ExternalMemoryAdjuster);
+
+  // Increment the amount of external memory reported. We pass in the
+  // Lock& only to prove that we are holding the isolate lock.
+  void increment(Lock&, int64_t amount);
```
kentonv, line 2058: I think there should be a simple `set()` method too, which calculates the adjustment based on the previous value.

Maybe `increment` and `decrement` can be combined into a single `adjust` and you pass a negative value for decrement?

src/workerd/jsg/jsg.c++
```diff
@@ -278,27 +278,57 @@ JsSymbol Lock::symbolAsyncDispose() {
...
+}
+
+ExternalMemoryAdjuster::ExternalMemoryAdjuster(ExternalMemoryAdjuster&& other) :
+    amount(other.amount), isolate(other.isolate) {
+  other.amount = 0;
+  other.isolate = nullptr;
+}
+
+ExternalMemoryAdjuster& ExternalMemoryAdjuster::operator=(ExternalMemoryAdjuster&& other) {
+  amount = other.amount;
```
kentonv, line 304: If `amount` is non-zero we need to apply an adjustment of `-amount` before overwriting it.

src/workerd/jsg/jsg.c++
```diff
@@ -278,27 +278,57 @@ JsSymbol Lock::symbolAsyncDispose() {
...
+  isolate = other.isolate;
+  other.amount = 0;
+  other.isolate = nullptr;
+  return *this;
+}
 
-  ~ExternalMemoryAdjuster() noexcept(false) {
-    isolate->AdjustAmountOfExternalAllocatedMemory(-size);
+ExternalMemoryAdjuster::~ExternalMemoryAdjuster() {
+  if (amount != 0 && isolate != nullptr) {
```
kentonv, line 312: Seems safe to check `amount != 0` only, right? It should be illegal to have a non-zero amount with a null isolate.

```suggestion
  if (amount != 0) {
```

src/workerd/jsg/jsg.h
```diff
@@ -2042,6 +2042,38 @@ class JsMessage;
...
+
+  // Decrement the amount of external memory reported. We pass in the
+  // Lock& only to prove that we are holding the isolate lock.
+  void decrement(Lock&, int64_t amount);
+
+  ~ExternalMemoryAdjuster();
```
kentonv, line 2064:
```suggestion
  ~ExternalMemoryAdjuster() noexcept(false);
```

src/workerd/jsg/jsg.h
kentonv: BTW, minor nitpick but IMO `size_t`/`ssize_t` are actually the right types to use here, they are meant to represent a quantity of memory.

src/workerd/jsg/jsg.c++
jasnell: aha, doh, good catch

src/workerd/jsg/jsg.c++
```diff
@@ -278,27 +278,68 @@ JsSymbol Lock::symbolAsyncDispose() {
...
+  int64_t adjustment = 0;
+  if (amount > 0) {
+    this->amount += amount;
+    adjustment = static_cast<int64_t>(amount);
+  } else if (amount < 0) {
+    size_t previousAmount = this->amount;
+    size_t amountToDecrement = kj::min(this->amount, -amount);
+    this->amount -= amountToDecrement;
+    KJ_ASSERT(this->amount < previousAmount);  // Be overly cautious about arithmetic overflow.
+    adjustment = -static_cast<int64_t>(amountToDecrement);
```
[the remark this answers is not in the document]
kentonv: My code above throws an exception if an adjustment would make the adjuster's size negative, since that's probably a bug. But if you really want to clamp the size to zero, then it could be...

```cpp
  amount = kj::max(amount, -this->amount);
  this->amount += amount;
  isolate->AdjustAmountOfExternalAllocatedMemory(amount);
```

src/workerd/jsg/jsg.h
```diff
@@ -2042,6 +2042,36 @@ class JsMessage;
 
 class DOMException;
 
+// RAII class to adjust the amount of external memory attributed to an isolate.
+// The adjustment will be automatically decremented when the object is destroyed.
+// The allocation amount can be adjusted up or down during the lifetime of an object.
+class ExternalMemoryAdjuster final {
```
kentonv, line 2048: Bikeshedding: Maybe `ExternalMemoryAdjustment` is a more natural name? The object itself represents an adjustment.

src/workerd/jsg/jsg.c++
```diff
@@ -278,27 +278,68 @@ JsSymbol Lock::symbolAsyncDispose() {
   return IsolateBase::from(v8Isolate).getSymbolAsyncDispose();
 }
 
-void Lock::adjustExternalMemory(ssize_t amount) {
-  v8Isolate->AdjustAmountOfExternalAllocatedMemory(static_cast<int64_t>(amount));
+void ExternalMemoryAdjuster::maybeDeferAdjustment(v8::Isolate* isolate, size_t amount) {
```
kentonv, line 281:
```suggestion
void ExternalMemoryAdjuster::maybeDeferAdjustment(v8::Isolate* isolate, ssize_t amount) {
```

kentonv, line 281: Minor, ignore if you want, but I found it slightly confusing that this represents an inverted adjustment, and we talk about deferring a "decrement"... if we just made this `ssize_t` and called it a "deferred adjustment" (which always happens to be negative in practice since it's used by the destructor) it seems clearer.

jasnell: Will leave it as is for now, can revisit this later if it proves helpful to do so

kentonv (approved):

PR #2512, opened 2024-08-10
[nfc] 08/10/24 Build/Docs cleanup

fhanau:
Cleaning out a number of smaller changes I had stashed for a while.

- Emphasize that building on macOS requires Xcode.
- Clean up Windows dependency script, the workaround is no longer needed.
- Fix cargo-culted comment in encoding-test.js
- Drop unused perfetto dependency for node target
- Mark a few classes as final, expand anonymous namespace usage.

### DIFF
--- a/src/workerd/server/workerd-api.c++
+++ b/src/workerd/server/workerd-api.c++
@@ -56,6 +56,7 @@ namespace workerd::server {
 
 using api::pyodide::PythonConfig;
 
+namespace {
 JSG_DECLARE_ISOLATE_TYPE(JsgWorkerdIsolate,
   // Declares the listing of host object types and structs that the jsg
   // automatic type mapping will understand. Each of the various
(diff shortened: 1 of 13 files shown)


src/workerd/server/workerd-api.c++
```diff
@@ -56,6 +56,7 @@ namespace workerd::server {
 
 using api::pyodide::PythonConfig;
 
+namespace {
```
fhanau, line 59: Adding this closure is high-impact because JsgWorkerdIsolate instantiates the entire workerd JS API – on macOS compiling without optimization sees little impact to compile time or object code size (41.1MB => 40.6MB), but under -O3 there is a large impact on both (38.2s => 51.8s, 15.1MB => 10.3MB). Compiling workerd-api.c++ is a bottleneck for the build, but I think this is worth it since such tradeoffs are what we are looking for when optimization is enabled, we might also see faster LTO link times as there is less code to be optimized at that stage.

anonrig (approved):

jasnell (approved):

PR #2515, opened 2024-08-11
Lazily initialize _cf_KV sqlite table.

kentonv:
This way, if a Durable Object never actually uses the KV storage interface -- only uses SQL -- we'll never create the table and can skip preparing statements for it.

This will also make it easier for the edge runtime to clean up actors that never had data.

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -1862,6 +1862,10 @@ public:
                 auto db = kj::heap<SqliteDatabase>(*as,
                     kj::Path({d.uniqueKey, kj::str(idPtr, ".sqlite")}),
                     kj::WriteMode::CREATE | kj::WriteMode::MODIFY | kj::WriteMode::CREATE_PARENT);
+
+                // Before we do anything, make sure the database is in WAL mode.
+                db->run("PRAGMA journal_mode=WAL;");
+
                 return kj::heap<ActorSqlite>(kj::mv(db), outputGate,
                     []() -> kj::Promise<void> { return kj::READY_NOW; },
                     *sqliteHooks).attach(kj::mv(sqliteHooks));
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -264,6 +264,7 @@ async function test(state) {
     () => sql.exec('CREATE TABLE _cf_invalid (name TEXT)'),
     /not authorized/
   )
+  storage.put("blah", 123);  // force creation of _cf_KV table
   assert.throws(
     () => sql.exec('SELECT * FROM _cf_KV'),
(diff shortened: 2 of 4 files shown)


src/workerd/server/server.c++
```diff
@@ -1857,6 +1857,10 @@ public:
                 auto db = kj::heap<SqliteDatabase>(*as,
                     kj::Path({d.uniqueKey, kj::str(idPtr, ".sqlite")}),
                     kj::WriteMode::CREATE | kj::WriteMode::MODIFY | kj::WriteMode::CREATE_PARENT);
+
+                // Before we do anything, make sure the database is in WAL mode.
+                db->run("PRAGMA journal_mode=WAL;");
```
anonrig, line 1862: Is there a test to ensure that the db is in WAL mode?

kentonv: I guess not at present.

Technically workerd works fine in any mode, we only enable WAL because it's better.

Storage Relay Service (what we use on the edge) absolutely requires WAL mode, so all its tests would certainly blow up with out it. But it also doesn't depend on workerd setting it; it sets WAL mode itself in internal code.

justin-mp (approved):

justin-mp:
Hmm, the `//src/workerd/api:sql-test` the test is unhappy:
```
workerd/server/server.c++:3791: debug: [ TEST ] main
(18:00:15) [2,869 / 2,870] Testing @@workerd//src/workerd/api:sql-test; 1s linux-sandbox
workerd/io/worker.c++:1925: info: uncaught exception; source = Uncaught (in promise); stack = AssertionError [ERR_ASSERTION]: The input did not match the regular expression /access to _cf_KV.key is prohibited/. Input:

'Error: no such table: _cf_KV: SQLITE_ERROR'

    at new AssertionError (node-internal:internal_assertionerror:419:15)
    at createAssertionError (node-internal:internal_assert:35:19)
    at validateThrownError (node-internal:internal_assert:552:15)
    at Module.throws (node-internal:internal_assert:97:13)
    at test (worker:267:10)
    at async DurableObjectExample.fetch (worker:1131:7)
workerd/io/io-context.c++:355: info: uncaught exception; exception = workerd/jsg/_virtual_includes/jsg/workerd/jsg/value.h:1355: failed: jsg.Error: AssertionError: The input did not match the regular expression /access to _cf_KV.key is prohibited/. Input:

'Error: no such table: _cf_KV: SQLITE_ERROR'
```

Did we miss a bypass or initialization somewhere?

kentonv:
> Did we miss a bypass or initialization somewhere?

No, the test was checking for precise error text that depended on the `_cf_KV` table existing, but now it's lazily initialized. Fixed by adding a storage.put() before the check.

PR #2519, opened 2024-08-12
add several `node:zlib` classes

anonrig:
Lacks tests but before adding more, I think this is getting ready to have an additional pair of eyes.

Remaining todo:

- [ ] Add tests
- [x] Investigate why eslint is causing so much trouble (hence eslint-ignore)

### DIFF
--- a/src/node/internal/internal_zlib_base.ts
+++ b/src/node/internal/internal_zlib_base.ts
@@ -1,0 +1,681 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+// Copyright Joyent and Node contributors. All rights reserved. MIT license.
+
+import { default as zlibUtil, type ZlibOptions } from 'node-internal:zlib';
+import { Buffer, kMaxLength } from 'node-internal:internal_buffer';
+import {
+  checkRangesOrGetDefault,
+  checkFiniteNumber,
+} from 'node-internal:validators';
+import {
+  ERR_OUT_OF_RANGE,
+  ERR_BUFFER_TOO_LARGE,
+  ERR_INVALID_ARG_TYPE,
+  NodeError,
+} from 'node-internal:internal_errors';
+import { Transform, type DuplexOptions } from 'node-internal:streams_transform';
+import { eos as finished } from 'node-internal:streams_util';
+import {
+  isArrayBufferView,
+  isAnyArrayBuffer,
+} from 'node-internal:internal_types';
+
+// Explicitly import `ok()` to avoid typescript error requiring every name in the call target to
+// be annotated with an explicit type annotation.
+import assert, { ok } from 'node-internal:internal_assert';
+
+const {
+  CONST_INFLATE,
+  CONST_GUNZIP,
+  CONST_GZIP,
+  CONST_UNZIP,
+  CONST_Z_DEFAULT_CHUNK,
+  CONST_Z_DEFAULT_STRATEGY,
+  CONST_Z_DEFAULT_MEMLEVEL,
+  CONST_Z_DEFAULT_WINDOWBITS,
+  CONST_Z_DEFAULT_COMPRESSION,
+  CONST_Z_FIXED,
+  CONST_Z_MAX_LEVEL,
+  CONST_Z_MAX_MEMLEVEL,
+  CONST_Z_MAX_WINDOWBITS,
+  CONST_Z_MIN_LEVEL,
+  CONST_Z_MIN_MEMLEVEL,
+  CONST_Z_SYNC_FLUSH,
+  CONST_Z_NO_FLUSH,
+  CONST_Z_BLOCK,
+  CONST_Z_MIN_CHUNK,
+  CONST_Z_PARTIAL_FLUSH,
+  CONST_Z_FULL_FLUSH,
+  CONST_Z_FINISH,
+  CONST_BROTLI_ENCODE,
+  CONST_BROTLI_DECODE,
+  CONST_BROTLI_OPERATION_PROCESS,
+  CONST_BROTLI_OPERATION_EMIT_METADATA,
+} = zlibUtil;
+
+export const owner_symbol = Symbol('owner');
+
+const FLUSH_BOUND_IDX_NORMAL: number = 0;
+const FLUSH_BOUND_IDX_BROTLI: number = 1;
+const FLUSH_BOUND: [[number, number], [number, number]] = [
+  [CONST_Z_NO_FLUSH, CONST_Z_BLOCK],
+  [CONST_BROTLI_OPERATION_PROCESS, CONST_BROTLI_OPERATION_EMIT_METADATA],
+];
+
+const kFlushFlag = Symbol('kFlushFlag');
+const kError = Symbol('kError');
+
+function processCallback(this: zlibUtil.ZlibStream): void {
+  // This callback's context (`this`) is the `_handle` (ZCtx) object. It is
+  // important to null out the values once they are no longer needed since
+  // `_handle` can stay in memory long after the buffer is needed.
+  // eslint-disable-next-line @typescript-eslint/no-this-alias
+  const handle = this;
+  const self = this[owner_symbol];
+  const state = self._writeState;
+
+  if (self.destroyed) {
+    this.buffer = null;
+    this.cb();
+    return;
+  }
+
+  const [availOutAfter, availInAfter] = state as unknown as [number, number];
+
+  const inDelta = handle.availInBefore - availInAfter;
+  self.bytesWritten += inDelta;
+
+  const have = handle.availOutBefore - availOutAfter;
+  let streamBufferIsFull = false;
+  if (have > 0) {
+    const out = self._outBuffer.slice(self._outOffset, self._outOffset + have);
+    self._outOffset += have;
+    streamBufferIsFull = !self.push(out);
+  } else {
+    assert.strictEqual(have, 0, 'have should not go down');
+  }
+
+  if (self.destroyed) {
+    this.cb();
+    return;
+  }
+
+  // Exhausted the output buffer, or used all the input create a new one.
+  if (availOutAfter === 0 || self._outOffset >= self._chunkSize) {
+    handle.availOutBefore = self._chunkSize;
+    self._outOffset = 0;
+    self._outBuffer = Buffer.allocUnsafe(self._chunkSize);
+  }
+
+  if (availOutAfter === 0) {
+    // Not actually done. Need to reprocess.
+    // Also, update the availInBefore to the availInAfter value,
+    // so that if we have to hit it a third (fourth, etc.) time,
+    // it'll have the correct byte counts.
+    handle.inOff += inDelta;
+    handle.availInBefore = availInAfter;
+
+    if (!streamBufferIsFull) {
+      this.write(
+        handle.flushFlag,
+        this.buffer as NodeJS.TypedArray, // in
+        handle.inOff, // in_off
+        handle.availInBefore, // in_len
+        self._outBuffer, // out
+        self._outOffset, // out_off
+        self._chunkSize
+      ); // out_len
+    } else {
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      const oldRead = self._read;
+      self._read = (n): void => {
+        self._read = oldRead;
+        this.write(
+          handle.flushFlag,
+          this.buffer as NodeJS.TypedArray, // in
+          handle.inOff, // in_off
+          handle.availInBefore, // in_len
+          self._outBuffer, // out
+          self._outOffset, // out_off
+          self._chunkSize
+        ); // out_len
+        self._read(n);
+      };
+    }
+    return;
+  }
+
+  if (availInAfter > 0) {
+    // If we have more input that should be written, but we also have output
+    // space available, that means that the compression library was not
+    // interested in receiving more data, and in particular that the input
+    // stream has ended early.
+    // This applies to streams where we don't check data past the end of
+    // what was consumed; that is, everything except Gunzip/Unzip.
+    self.push(null);
+  }
+
+  // Finished with the chunk.
+  this.buffer = null;
+  this.cb();
+}
+
+// If a flush is scheduled while another flush is still pending, a way to figure
+// out which one is the "stronger" flush is needed.
+// This is currently only used to figure out which flush flag to use for the
+// last chunk.
+// Roughly, the following holds:
+// Z_NO_FLUSH (< Z_TREES) < Z_BLOCK < Z_PARTIAL_FLUSH <
+//     Z_SYNC_FLUSH < Z_FULL_FLUSH < Z_FINISH
+const flushiness: number[] = [];
+const kFlushFlagList = [
+  CONST_Z_NO_FLUSH,
+  CONST_Z_BLOCK,
+  CONST_Z_PARTIAL_FLUSH,
+  CONST_Z_SYNC_FLUSH,
+  CONST_Z_FULL_FLUSH,
+  CONST_Z_FINISH,
+];
+for (let i = 0; i < kFlushFlagList.length; i++) {
+  flushiness[kFlushFlagList[i] as number] = i;
+}
+
+type BufferWithFlushFlag = Buffer & { [kFlushFlag]: number };
+
+// Set up a list of 'special' buffers that can be written using .write()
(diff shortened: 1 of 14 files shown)


src/node/.eslintrc.json
```diff
@@ -10,32 +10,32 @@
   ],
   "overrides": [],
   "parserOptions": {
-      "ecmaVersion": "latest",
-      "sourceType": "module",
-      "project": "src/node/tsconfig.json"
-    },
+    "ecmaVersion": "latest",
+    "sourceType": "module",
+    "project": "./tsconfig.json"
```
npaun, line 15: Should be moved to another PR

src/node/tsconfig.json
```diff
@@ -21,9 +21,7 @@
     "moduleResolution": "node",
     "declaration": true,
     "types": [
-      // todo: consume generated workerd types
-      // "@cloudflare/workers-types"
-      "@types/node",
+      "@types/node"
```
npaun, line 24: Also not related

src/node/internal/validators.ts
```diff
@@ -218,4 +258,8 @@ export default {
   validateOneOf,
   validateString,
   validateUint32,
+
+  // Zlib specific
```
npaun, line 262: I think I'd either remove this comment, or move the functions to the zlib file.

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,261 @@
...
+#define V(code) if (err == code) return #code;
+  ZLIB_ERROR_CODES(V)
+#undef V
+  return "Z_UNKNOWN_ERROR";
+}
+#endif  // ZLIB_ERROR_CODES
+
+// Certain zlib constants are defined by Node.js itself
+static constexpr auto Z_MIN_CHUNK = 64;
+static constexpr auto Z_MAX_CHUNK = 128 * 1024;
```
npaun, line 40:
```suggestion
static constexpr auto Z_MAX_CHUNK = 128 * 1024 * 1024;
```

This is fixed in my PR too but just pointing it out here

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,261 @@
...
+  GUNZIP,
+  DEFLATERAW,
+  INFLATERAW,
+  UNZIP,
+  BROTLI_DECODE,
+  BROTLI_ENCODE
+};
+
+struct CompressionError {
+  CompressionError(kj::StringPtr _message, kj::StringPtr _code, int _err)
```
npaun, line 71: I don't personally like variable names starting with `_` but if it's a convention elsewhere in our codebase, then ignore me.

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,261 @@
...
+    stream.avail_in = inputLength;
+    stream.next_in = input.begin();
+    stream.avail_out = outputLength;
+    stream.next_out = output.begin();
+  };
+  int getFlush() const { return flush; };
+  void setFlush(int value) {
+    flush = value;
+  };
+  void getAfterWriteOffsets(kj::ArrayPtr<kj::byte> writeResult) const {
```
npaun, line 100: Is this function designed this way to conform to a NodeJS interface? I'm not a big fan of out-parameters, nor of storing struct fields as an array.

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,261 @@
...
+};
+
+class ZlibContext final {
+public:
+  ZlibContext() = default;
+
+  KJ_DISALLOW_COPY(ZlibContext);
+
+  void close();
+  void setBuffers(kj::ArrayPtr<kj::byte> input, uint32_t inputLength, kj::ArrayPtr<kj::byte> output, uint32_t outputLength) {
```
npaun, line 90: What is our convention on when it's ok to implement methods inline? My instinct would be to avoid it, because it slows down compilation time, but this is a project-dependent thing.

src/workerd/api/node/zlib-util.c++
```diff
@@ -1,11 +1,373 @@
...
 #include "zlib-util.h"
 
 namespace workerd::api::node {
+
 uint32_t ZlibUtil::crc32Sync(kj::Array<kj::byte> data, uint32_t value) {
   // Note: Bytef is defined in zlib.h
   return crc32(value, reinterpret_cast<const Bytef*>(data.begin()), data.size());
 }
+
+void ZlibContext::initialize(int _level, int _windowBits, int _memLevel, int _strategy, jsg::Optional<kj::Array<kj::byte>> _dictionary) {
```
npaun, line 15: Just commenting to note that this function is the same for both one-shot and async and we should see if we can reuse it.

src/workerd/api/node/zlib-util.c++
```diff
@@ -1,11 +1,373 @@
...
+          // Make it possible for After() to tell a bad dictionary from bad
+          // input.
+          err = Z_NEED_DICT;
+        }
+      }
+
+      while (stream.avail_in > 0 &&
+             mode == ZlibMode::GUNZIP &&
+             err == Z_STREAM_END &&
+             stream.next_in[0] != 0x00) {
```
npaun, line 243:
```suggestion
             stream.next_in[0] != '\0') {
```

nit: I think this is more common when we working with bytes

src/workerd/api/node/zlib-util.c++
```diff
@@ -1,11 +1,373 @@
...
+  // of room.  If there was avail_out left over, then it means
+  // that all the input was consumed.
+  switch (mode) {
+    case ZlibMode::DEFLATE:
+    case ZlibMode::GZIP:
+    case ZlibMode::DEFLATERAW:
+      err = deflate(&stream, flush);
+      break;
+    case ZlibMode::UNZIP:
+      if (stream.avail_in > 0) {
```
npaun, line 178: This case is very complex. I would consider to extract it into its own function and invoke it from here.

src/node/internal/zlib.d.ts
```diff
@@ -113,3 +115,41 @@ export const CONST_BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_1: number;
...
+  dictionary?: ArrayBufferView | undefined; // deflate/inflate only, empty dictionary by default
+  info?: boolean | undefined;
+  maxOutputLength?: number | undefined;
+}
+
+type ErrorHandler = (errno: number, code: string, message: string) => void;
+type ProcessHandler = () => void;
+
+export class ZlibStream {
+  public [owner_symbol]: Zlib;
```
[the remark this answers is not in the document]
anonrig: Before private properties in javascript (`this.#private_property`) were a thing, people used Symbol's to store private properties. Each symbol is unique and therefore is different from each other.

```
const valSymbol = Symbol('hello');

class Hello {
  setAttr(val) {
    this[valSymbol] = val;
  }

  getAttr() { return this[valSymbol]; }
}

Since zlib was implemented before private properties, this symbol is used to store the variable reference to the parent class instance.

File: src/workerd/api/node/zlib-util.h (hunk 5 above)
Comment (anonrig) [reply src/workerd/api/node/zlib-util.h:71]: This is a pretty common Node.js convention, but I'm pretty sure @jasnell uses it a lot as well. How would you initialize CompressionError where there is a `message` attribute? How would you name this variable int that scenario?

File: src/workerd/api/node/zlib-util.h (hunk 6 above)
Comment (anonrig) [reply src/workerd/api/node/zlib-util.h:100]: I tried to not diverge from Node.js implementation as much as I can to make `maintenance` of this module a lot easier, unless there is a specific reason to do so.

File: src/workerd/api/node/zlib-util.h (hunk 7 above)
Comment (anonrig) [reply src/workerd/api/node/zlib-util.h:90]: i'm not sure. i have this way in my mind that if the function is simple to implement in header, i put it in header. but technically we should have `zlib-util-inl.h` class if we want to conform to Google code style. @jasnell?

File: src/workerd/api/node/zlib-util.h (hunk 5 above)
Comment (npaun) [reply src/workerd/api/node/zlib-util.h:71]: Same name for both is what I'm used to, but if we're doing underscores I'll change.

File: src/node/.eslintrc.json (hunk 1 above)
Comment (anonrig) [reply src/node/.eslintrc.json:15]: Agree: https://github.com/cloudflare/workerd/pull/2554

File: src/node/internal/internal_zlib_base.ts [hunk 12]
```diff
@@ -0,0 +1,558 @@
...
+  let offset = self._outOffset;
+  const chunkSize = self._chunkSize;
+
+  let error;
+  self.on('error', function onError(er) {
+    error = er;
+  });
+
+  while (true) {
+    // TODO: This was `writeSync` before, but it's not anymore.
```
jasnell, line 211: Is this our todo or an existing todo from the node.js source?

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,204 @@
 // Copyright (c) 2017-2022 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
-
+// Copyright Joyent and Node contributors. All rights reserved. MIT license.
 #include <workerd/jsg/jsg.h>
```
jasnell, line 5:
```suggestion
#pragma once
#include <workerd/jsg/jsg.h>
```

src/workerd/api/node/zlib-util.h
jasnell: For this code that is largely adopted from Node.js I prefer that we maintain the Node.js conventions in order to make porting (and more importantly updating later) easier

src/workerd/api/node/zlib-util.h
jasnell: There may be some difference of opinion on it in the team but for me, it really comes down to the complexity of the function. Generally try to avoid inlining in the headers unless the body is very trivial (e.g. a trivial constructor, or a trivial getter/setter). When in doubt, if it's going to be more than a few lines, don't inline.

jasnell:
Overall looking good so far. Definitely want to see more tests land with this tho so will hold off signing off until we get those.

src/node/internal/internal_zlib_base.ts
anonrig: This is our TODO. I implemented the `sync` path, not didn't implement the async path on C++.

src/workerd/api/node/zlib-util.h
anonrig: Header files should only be included only once. Why do we need to have a pragma once here? Do we not trust the compiler to include a header file only once, or is it something else?

npaun: The default behaviour is to include every time. `#pragma once` tells the compiler not to include `zlib-util.h` more than once.

src/workerd/api/node/zlib-util.h
npaun: @jasnell That's my usual approach as well. However wouldn't we get faster builds if we totally crack down on inline implementations?

src/node/internal/internal_zlib_base.ts
jasnell: Ok, to differentiate, make the todo like... This will help since the `(soon)` is our convention vs. node.js convention of `TODO(username): `

```suggestion
    // TODO(soon): This was `writeSync` before, but it's not anymore.
```

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,206 @@
...
+  z_stream stream{};
+};
+
+using CompressionStreamErrorHandler = jsg::Function<void(int, kj::StringPtr, kj::StringPtr)>;
+
+template <typename CompressionContext>
+class CompressionStream {
+public:
+  CompressionStream() = default;
+  ~CompressionStream();
```
jasnell, line 146:
```suggestion
  ~CompressionStream() noexcept(false);
```

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,206 @@
...
-    BROTLI_ENCODE
+  class ZlibStream final : public jsg::Object, public CompressionStream<ZlibContext> {
+  public:
+    ZlibStream(ZlibMode mode);
+    KJ_DISALLOW_COPY_AND_MOVE(ZlibStream);
+    static jsg::Ref<ZlibStream> constructor(ZlibModeValue mode);
+
+    // Instance methods
+    void initialize(int windowBits, int level, int memLevel, int strategy, kj::Array<kj::byte> writeState, jsg::Function<void()> writeCallback, jsg::Optional<kj::Array<kj::byte>> dictionary);
+    void write(jsg::Lock& js,int flush, kj::Array<kj::byte> input, int inputOffset, int inputLength, kj::Array<kj::byte> output, int outputOffset, int outputLength);
```
jasnell, line 185:
```suggestion
    void write(jsg::Lock& js, int flush, kj::Array<kj::byte> input, int inputOffset, int inputLength, kj::Array<kj::byte> output, int outputOffset, int outputLength);
```

src/node/zlib.ts
```diff
@@ -1,8 +1,74 @@
+import * as zlib from 'node-internal:internal_zlib';
 import { crc32, constants } from 'node-internal:internal_zlib';
+import { nodeJsZlib } from 'workerd:compatibility-flags';
 
-export { crc32, constants };
+function protectMethod(method: unknown): unknown {
+  if (nodeJsZlib) {
+    return function notImplemented() {
+      throw new Error('Compatibility flag "nodejs_compat_zlib" is not enabled');
```
jasnell, line 8:
```suggestion
      throw new Error('Compatibility flag "nodejs_zlib" is not enabled');
```

src/workerd/io/compatibility-date.capnp
```diff
@@ -580,4 +580,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
           packages = "2024-03-01", backport = 0)
     $experimental;
   # Enables Python Workers and uses the bundle from the Pyodide source directory directly. For testing only.
+
+  nodeJsZlib @59 :Bool
+      $compatEnableFlag("nodejs_zlib")
+      $compatDisableFlag("no_nodejs_zlib")
+      $experimental;
+  # Enables node:zlib implementation while it is in-development.
```
jasnell, line 588:
```suggestion
  # Enables node:zlib implementation while it is in-development.
  # Once the node:zlib implementation is complete, this will be automatically enabled when
  # nodejs_compat is enabled.
```

jasnell (approved):
Looks good once we're sure the internal CI is green.

src/workerd/api/node/zlib-util.h
```diff
@@ -1,57 +1,245 @@
...
+  bool initializeZlib();
+  kj::Maybe<CompressionError> setDictionary();
+
+  CompressionError constructError(kj::StringPtr message) const {
+    if (stream.msg != nullptr) message = kj::StringPtr(stream.msg);
+
+    return {kj::str(message), kj::str(ZlibStrerror(err)), err};
+  };
+
+  kj::MutexGuarded<bool> initialized{false};
```
mikea, line 145: this is rather unusual: please leave a comment which threads access it and why this synchronization is necessary.

mikea (approved):
appreciate the effort to make all .ts code typed.

src/workerd/api/node/zlib-util.h
jasnell: @anonrig and I discussed this a bit yesterday. I don't think this is actually needed and it should be removed in the next iteration as this moves forward. A mutex is used here in Node.js because there are cases that need to be protected from multiple threads accessing this but for workerd it shouldn't be necessary.

npaun (changes requested):
Tests are failing under ASan. I'm worried there might be some kind of corruption.

src/workerd/api/node/zlib-util.h
anonrig: Yeah, it is unnecessary. While investigating the ASAN build, I removed it.

npaun:
Do we know why the unnecessary mutex guard caused the test failure?

It doesn't immediately jump out at me what this has to do with a mutex
`Fatal uncaught kj::Exception: workerd/api/node/zlib-util.c++:311: failed: expected status == 0 || status == (-3); jsg.Error: Uncaught error on closing zlib stream`

anonrig:
> Do we know why the unnecessary mutex guard caused the test failure?

Yes, the class destructor was checking the reference not the value, and causing the sanitizer to fail.

npaun (approved):

npaun:
Aside: That implicit conversion to `operator T*` in libkj is kind of wack.

PR #2522, opened 2024-08-12
Use V8's zlib

npaun:
This PR changes the source of zlib to match what V8 is using, so that we don't have version conflicts between workerd and the internal edgeworker repo.

zlib is built from source. The build config is a translation from the BUILD.gn file into bazel syntax.

Note that this is a bit of a different build process from edgeworker, which uses precompiled binaries from v8 for zlib.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -282,34 +282,12 @@ http_archive(
 # the build process. To update the dependency, update the reference commit in
 # rust-deps/BUILD.bazel and run `bazel run //rust-deps:crates_vendor -- --repin`
 
-# Based on https://github.com/bazelbuild/bazel/blob/master/third_party/zlib/BUILD.
-_zlib_build = """
-cc_library(
+git_repository(
     name = "zlib",
-    srcs = glob(["*.c"]),
-    hdrs = glob(["*.h"]),
-    includes = ["."],
-    # Workaround for zlib warnings and mac compilation. Some issues were resolved in v1.3, but there are still implicit function declarations.
-    copts = [
-        "-w",
-        "-Dverbose=-1",
-    ] + select({
-        "@platforms//os:linux": [ "-Wno-implicit-function-declaration" ],
-        "@platforms//os:macos": [ "-Wno-implicit-function-declaration" ],
-        "//conditions:default": [],
-    }),
-    visibility = ["//visibility:public"],
-)
-"""
-
-http_archive(
-    name = "zlib",
-    build_file_content = _zlib_build,
-    sha256 = "38ef96b8dfe510d42707d9c781877914792541133e1870841463bfa73f883e32",
-    strip_prefix = "zlib-1.3.1",
-    # Using the .tar.xz artifact from the release page – for many other dependencies we use a
-    # snapshot based on the tag of a release instead.
-    urls = ["https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.xz"],
+    build_file = "//:build/BUILD.zlib",
+    # Must match the version used by v8
+    commit = "c2469fdd73f192383d2d94288da0ff5b9a3869f5",
+    remote = "https://chromium.googlesource.com/chromium/src/third_party/zlib.git",
 )
 
 http_file(
--- a/build/BUILD.zlib
+++ b/build/BUILD.zlib
@@ -1,0 +1,194 @@
+# This config closely follows the original GN build file
+# Ref: https://chromium.googlesource.com/chromium/src/third_party/zlib.git/+/refs/heads/main/BUILD.gn
+# The main difference is that we don't set ZLIB_DLL because it messes things up on Windows
+
+package(default_visibility = ["//visibility:public"])
+
+zlib_warnings = [
+    "-Wno-incompatible-pointer-types",
+    "-Wunused-variable",
+]
+
+cc_library(
+    name = "zlib_common_headers",
+    hdrs = [
+        "chromeconf.h",
+        "deflate.h",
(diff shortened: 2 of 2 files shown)


fhanau:
Couple thoughts on this – 
- we could use our own high-performance zlib fork ([cloudflare/zlib](https://github.com/cloudflare/zlib/)) here, but I think using Chromium zlib is better for consistency since we use V8 a lot. Chromium zlib also has a number of performance optimizations (in fact some of the decompression optimizations in Cloudflare zlib originate from Chromium zlib) and should work well.
- I looked into switching a while back but got bogged down by configuring the build to configure SIMD support properly. I think we should do this eventually but it's not required for now – the [GN build file](https://chromium.googlesource.com/chromium/src/third_party/zlib/+/71660e196bcfc74b2c657a46b512037cb49079a2/BUILD.gn) is pretty complex and we'd also have to properly document what level of hardware extensions workerd requires. This is best done together with compiling with e.g. x86-64-v3 by default.
- I don't think we should test for `ZLIB_VERNUM` at all – in my experience it is not relevant to developers, except when checking if some C-level API that got added in a certain version will be available, or perhaps if a bug fix added in a certain version is added. Either way, we don't depend on zlib being a specific fixed version and shouldn't have to update a test every time the zlib version is incremented.

Change LGTM but it looks like there's SIMD-related compile errors on Windows so building this using Bazel might be more complex

jasnell:
@anonrig ... your experience with simd may be helpful in the longer term effort here

npaun:
Thanks for the notes @fhanau 

I don't care about ZLIB_VERNUM, itself just that there is consistency between what is being used in workerd and edgeworker, so as to eliminate a possible source of trouble. 

I'll have a look tomorrow into what is going wrong on Windows. Fingers crossed that it's not too hard to fix.

WORKSPACE
```diff
@@ -300,14 +300,11 @@ cc_library(
...
+git_repository(
     name = "zlib",
-    build_file_content = _zlib_build,
-    sha256 = "38ef96b8dfe510d42707d9c781877914792541133e1870841463bfa73f883e32",
-    strip_prefix = "zlib-1.3.1",
-    # Using the .tar.xz artifact from the release page – for many other dependencies we use a
-    # snapshot based on the tag of a release instead.
-    urls = ["https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.xz"],
+    build_file_content=_zlib_build,
+    commit = "71660e196bcfc74b2c657a46b512037cb49079a2",
```
anonrig, line 306: Is this the commit that we use on internal repo? Keeping this value in sync with the V8 version might be a challenge.

anonrig:
Node.js ZLIB configuration might give you a better understanding for the SIMD related issues. If you couldn't solve this in a reasonable amount of time, I have a Windows machine at home. 

https://github.com/anonrig/node/blob/main/deps/zlib/zlib.gyp

WORKSPACE
npaun: In edgeworker we link against the object files that v8 already built for zlib rather than the source, but we have several files worth of scaffolding for this. I got the commit hash from the v8 repo and indeed it's going to get annoying to keep it in sync. But it's the same problem with ICU isn't it?

npaun:
May resolve #2540 if we decide to go with this approach.

npaun:
Internal PR passes

anonrig (approved):

fhanau:
Thank you for doing this! Should certainly improve performance in use cases where developers use compression APIs within workerd a lot.

build/BUILD.zlib
```diff
@@ -0,0 +1,194 @@
...
+    name = "zlib_adler32_simd",
+    srcs = [
+        "adler32_simd.c",
+        "adler32_simd.h",
+    ],
+    copts = select({
+        "@platforms//cpu:x86_64": ["-mssse3"],
+        "//conditions:default": [],
+    }),
+    defines = select({
```
fhanau, line 48: These should really be `local_defines` – defines are added to every target that depends on zlib otherwise, which is really polluting the global defines namespace and unnecessarily making the command line of workerd longer. https://bazel.build/reference/be/c-cpp#cc_library.local_defines

build/BUILD.zlib
```diff
@@ -0,0 +1,194 @@
...
+)
+
+cc_library(
+    name = "zlib_crc32_simd",
+    srcs = [
+        "crc32_simd.c",
+        "crc32_simd.h",
+        "crc_folding.c",
+    ],
+    copts = [
```
fhanau, line 104: This adds a requirement on ISA extensions that we didn't have before – IMHO this is reasonable since these extensions have been around for a long time but we need to document it.
If we're doing that it's reasonable to define the same set of extensions for the entire build and remove the individual flags from here (I agree that we should follow the structure of BUILD.gn for the individual targets but to make the file less complex we should not have to define ISA extension flags for each target). I will prepare a PR for this.

build/BUILD.zlib
npaun: I agree; the main reason I left it all as define is to match BUILD.gn. I can take a crack at limiting the scope in another PR.

npaun: There's also another nuisance: I'm using `defines` to avoid nested selects which don't work in bazel. I am actually not entirely sure how to get around this.

fhanau: Great question! We use `selects.config_setting_group` (https://bazel.build/docs/configurable-attributes#and-chaining) for this. Started with a PR on converting those.

PR #2526, opened 2024-08-13
Fix some incorrect include statements.

danlapid:
Replacing without reordering quote includes to angle bracket includes where appropriate.
The following logic is used to find the incorrect includes:
  find #include "kj/
  find #include "capnp/
  find #include "workerd/
  find #include "v8- // AND the header is in the v8 include directory
  find #include "openssl/
  find #include ".*?/.* // Not all of these are wrong
  find #include c system headers // replace <stdint.h> with <cstdint>

If we decide to enable sortincludes in clang-format this should fix some incorrect grouping.

### DIFF
--- a/src/workerd/api/crypto/x509.c++
+++ b/src/workerd/api/crypto/x509.c++
@@ -1,8 +1,8 @@
 #include "impl.h"
 #include "x509.h"
-#include "openssl/bio.h"
-#include "openssl/pem.h"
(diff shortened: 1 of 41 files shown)


anonrig (approved):

npaun (approved):
👏 👏 👏

jasnell (approved):

PR #2527, opened 2024-08-13
Implements globalThis.Cloudflare.compatibilityFlags API

jasnell:
A simple built-in API for determining if a given compat flag is set.

```
const { compatibilityFlags } = globalThis.Cloudflare || {};

console.log(compatibilityFlags['url_standard']); 
```

Internal discussion wiki: (sorry external folks, you won't be able to see this) https://wiki.cfdata.org/display/~jsnell/Exposing+active+compat+flags+to+workers

### DIFF
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -110,6 +110,22 @@ public:
   }
 };
 
+// Exposed as a global to provide access to certain Cloudflare-specific
+// configuration details. This is not a standard API and great care should
+// be taken when deciding to expose new properties or methods here.
+class Cloudflare: public jsg::Object {
+public:
+  // Return an object containing the state of all compatibility flags known to the runtime.
+  jsg::JsObject getCompatibilityFlags(jsg::Lock& js);
+
+  JSG_RESOURCE_TYPE(Cloudflare) {
+    JSG_LAZY_READONLY_INSTANCE_PROPERTY(compatibilityFlags, getCompatibilityFlags);
+
+    JSG_TS_OVERRIDE({ readonly compatibilityFlags: Record<string, boolean>;
+    });
+  }
+};
+
 class PromiseRejectionEvent: public Event {
 public:
   PromiseRejectionEvent(
@@ -507,6 +523,10 @@ public:
     return jsg::alloc<Performance>();
   }
 
+  jsg::Ref<Cloudflare> getCloudflare() {
+    return jsg::alloc<Cloudflare>();
+  }
+
   // The origin is unknown, return "null" as described in
   // https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque.
   kj::StringPtr getOrigin() {
@@ -570,6 +590,7 @@ public:
     JSG_LAZY_INSTANCE_PROPERTY(caches, getCaches);
     JSG_LAZY_INSTANCE_PROPERTY(scheduler, getScheduler);
     JSG_LAZY_INSTANCE_PROPERTY(performance, getPerformance);
+    JSG_LAZY_INSTANCE_PROPERTY(Cloudflare, getCloudflare);
     JSG_READONLY_INSTANCE_PROPERTY(origin, getOrigin);
 
     JSG_NESTED_TYPE(Event);
@@ -824,6 +845,6 @@ private:
   api::WorkerGlobalScope, api::ServiceWorkerGlobalScope, api::TestController,                      \
       api::ExecutionContext, api::ExportedHandler,                                                 \
       api::ServiceWorkerGlobalScope::StructuredCloneOptions, api::PromiseRejectionEvent,           \
-      api::Navigator, api::Performance, api::AlarmInvocationInfo, api::Immediate
+      api::Navigator, api::Performance, api::AlarmInvocationInfo, api::Immediate, api::Cloudflare
 // The list of global-scope.h types that are added to worker.c++'s JSG_DECLARE_ISOLATE_TYPE
 }  // namespace workerd::api
--- a/src/workerd/api/node/node.h
+++ b/src/workerd/api/node/node.h
@@ -12,6 +12,7 @@
 #include <workerd/jsg/modules-new.h>
 #include <capnp/dynamic.h>
 #include <node/node.capnp.h>
+#include <workerd/io/compatibility-date.h>
 
 namespace workerd::api::node {
 
--- a/src/workerd/api/tests/compat-flags-test.js
+++ b/src/workerd/api/tests/compat-flags-test.js
@@ -1,0 +1,58 @@
+import { ok, strictEqual, throws } from 'node:assert';
+
+const { compatibilityFlags } = globalThis.Cloudflare;
+ok(Object.isSealed(compatibilityFlags));
+ok(Object.isFrozen(compatibilityFlags));
+ok(!Object.isExtensible(compatibilityFlags));
+
+// It shuld be possible to shadow the Cloudflare global
+const Cloudflare = 1;
+strictEqual(Cloudflare, 1);
+
+export const compatFlagsTest = {
+  test() {
+    // The compatibility flags object should be sealed, frozen, and not extensible.
+    throws(() => (compatibilityFlags.no_nodejs_compat_v2 = '...'));
+    throws(() => (compatibilityFlags.not_a_real_compat_flag = '...'));
+    throws(() => {
+      delete compatibilityFlags['nodejs_compat_v2'];
+    });
+
+    // The compatibility flags object should have no prototype.
+    strictEqual(Object.getPrototypeOf(compatibilityFlags), null);
+
+    // The compatibility flags object should have the expected properties...
+    // That is... the only keys that should appear on the compatibilityFlags
+    // object are the enable flags.
+    //
+    // If a key does not appear, it can mean one of three things:
+    // 1. It is a disable flag.
+    // 2. It is an experimental flag and the experimental option is not set.
+    // 3. The flag does not exist.
+    //
+    // At this level we make no attempt to differentiate between these cases
+    ok(compatibilityFlags['nodejs_compat_v2']);
+    ok(compatibilityFlags['url_standard']);
+
+    ok(!compatibilityFlags['no_nodejs_compat_v2']);
+    ok(!compatibilityFlags['url_original']);
+    strictEqual(compatibilityFlags['no_nodejs_compat_v2'], undefined);
+    strictEqual(compatibilityFlags['url_original'], undefined);
+
+    // Since we are not specifying the experimental flag, experimental flags should
+    // not be included in the output.
+    strictEqual(compatibilityFlags['durable_object_rename'], undefined);
+    strictEqual('durable_object_rename' in compatibilityFlags, false);
+
+    // If a flag does not exist, the value will be undefined.
+    strictEqual(compatibilityFlags['not-a-real-compat-flag'], undefined);
+    strictEqual('not-a-real-compat-flag' in compatibilityFlags, false);
+
+    // The compatibility flags object should have the expected keys.
+    const keys = Object.keys(compatibilityFlags);
+    ok(keys.includes('nodejs_compat_v2'));
+    ok(keys.includes('url_standard'));
+    ok(!keys.includes('url_original'));
+    ok(!keys.includes('not-a-real-compat-flag'));
+  },
+};
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -918,4 +918,40 @@ void ServiceWorkerGlobalScope::clearImmediate(kj::Maybe<jsg::Ref<Immediate>> may
   }
 }
 
+jsg::JsObject Cloudflare::getCompatibilityFlags(jsg::Lock& js) {
+  auto flags = FeatureFlags::get(js);
+  auto obj = js.objNoProto();
+  auto dynamic = capnp::toDynamic(flags);
+  auto schema = dynamic.getSchema();
+
+  bool skipExperimental = !flags.getWorkerdExperimental();
+
+  for (auto field: schema.getFields()) {
+    // If this is an experimental flag, we expose it only if the experimental mode
+    // is enabled.
+    auto annotations = field.getProto().getAnnotations();
+    bool skip = false;
+    if (skipExperimental) {
+      for (auto annotation: annotations) {
+        if (annotation.getId() == EXPERIMENTAl_ANNOTATION_ID) {
+          skip = true;
+          break;
+        }
+      }
+    }
+    if (skip) continue;
+
+    // Note that disable flags are not exposed.
+    for (auto annotation: annotations) {
+      if (annotation.getId() == COMPAT_ENABLE_FLAG_ANNOTATION_ID) {
+        obj.setReadOnly(
+            js, annotation.getValue().getText(), js.boolean(dynamic.get(field).as<bool>()));
+      }
+    }
+  }
+
+  obj.seal(js);
+  return obj;
+}
+
 }  // namespace workerd::api
--- a/src/workerd/api/tests/compat-flags-test.wd-test
+++ b/src/workerd/api/tests/compat-flags-test.wd-test
@@ -1,0 +1,15 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "compat-flags-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "compat-flags-test.js")
+        ],
+        compatibilityDate = "2023-01-15",
+        compatibilityFlags = ["nodejs_compat_v2"],
+      )
+    ),
+  ],
+);
--- a/src/workerd/io/features.h
+++ b/src/workerd/io/features.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <workerd/io/compatibility-date.capnp.h>
+#include <workerd/io/compatibility-date.h>
 #include <workerd/jsg/jsg.h>
 
 namespace workerd {
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -2541,6 +2541,7 @@ public:
   JsSymbol symbolShared(kj::StringPtr) KJ_WARN_UNUSED_RESULT;
   JsSymbol symbolInternal(kj::StringPtr) KJ_WARN_UNUSED_RESULT;
   JsObject obj() KJ_WARN_UNUSED_RESULT;
+  JsObject objNoProto() KJ_WARN_UNUSED_RESULT;
   JsMap map() KJ_WARN_UNUSED_RESULT;
   JsValue external(void*) KJ_WARN_UNUSED_RESULT;
   JsValue error(kj::StringPtr message) KJ_WARN_UNUSED_RESULT;
--- a/src/workerd/jsg/jsvalue.c++
+++ b/src/workerd/jsg/jsvalue.c++
@@ -60,6 +60,12 @@ void JsObject::set(Lock& js, kj::StringPtr name, const JsValue& value) {
   set(js, js.strIntern(name), value);
 }
 
+void JsObject::setReadOnly(Lock& js, kj::StringPtr name, const JsValue& value) {
+  v8::Local<v8::String> nameStr = js.strIntern(name);
+  check(inner->DefineOwnProperty(js.v8Context(), nameStr, value,
+      static_cast<v8::PropertyAttribute>(v8::ReadOnly | v8::DontDelete)));
+}
+
 JsValue JsObject::get(Lock& js, const JsValue& name) {
   return JsValue(check(inner->Get(js.v8Context(), name.inner)));
 }
@@ -133,6 +139,10 @@ JsArray JsObject::previewEntries(bool* isKeyValue) {
 
 void JsObject::recursivelyFreeze(Lock& js) {
   jsg::recursivelyFreeze(js.v8Context(), inner);
+}
+
+void JsObject::seal(Lock& js) {
+  check(inner->SetIntegrityLevel(js.v8Context(), v8::IntegrityLevel::kSealed));
 }
 
 JsObject JsObject::jsonClone(Lock& js) {
@@ -419,6 +429,10 @@ JsRegExp Lock::regexp(kj::StringPtr str, RegExpFlags flags, kj::Maybe<uint32_t> 
 
 JsObject Lock::obj() {
   return JsObject(v8::Object::New(v8Isolate));
+}
+
+JsObject Lock::objNoProto() {
+  return JsObject(v8::Object::New(v8Isolate, v8::Null(v8Isolate), nullptr, nullptr, 0));
 }
 
 JsMap Lock::map() {
--- a/src/workerd/jsg/jsvalue.h
+++ b/src/workerd/jsg/jsvalue.h
@@ -341,6 +341,7 @@ public:
 
   void set(Lock& js, const JsValue& name, const JsValue& value);
   void set(Lock& js, kj::StringPtr name, const JsValue& value);
+  void setReadOnly(Lock& js, kj::StringPtr name, const JsValue& value);
   JsValue get(Lock& js, const JsValue& name) KJ_WARN_UNUSED_RESULT;
   JsValue get(Lock& js, kj::StringPtr name) KJ_WARN_UNUSED_RESULT;
 
@@ -377,6 +378,7 @@ public:
   using JsBase<v8::Object, JsObject>::JsBase;
 
   void recursivelyFreeze(Lock&);
+  void seal(Lock&);
   JsObject jsonClone(Lock&);
 };
 


src/cloudflare/workers.ts
```diff
@@ -11,3 +11,8 @@ export const WorkerEntrypoint = entrypoints.WorkerEntrypoint;
 export const DurableObject = entrypoints.DurableObject;
 export const RpcStub = entrypoints.RpcStub;
 export const RpcTarget = entrypoints.RpcTarget;
+
+/* eslint-disable */
+import { default as flags } from 'workerd:compatibility-flags';
+export const compatibilityFlags = (flags as any).compatibilityFlags;
```
anonrig, line 17: I think we can remove eslint-disable here. Doesn't this work? 

```
export {
  compatibilityFlags: flags.compatibilityFlags,
}
```

src/cloudflare/workers.ts
```diff
@@ -11,3 +11,8 @@ export const WorkerEntrypoint = entrypoints.WorkerEntrypoint;
 export const DurableObject = entrypoints.DurableObject;
 export const RpcStub = entrypoints.RpcStub;
 export const RpcTarget = entrypoints.RpcTarget;
+
+/* eslint-disable */
+import { default as flags } from 'workerd:compatibility-flags';
```
anonrig, line 16:
```suggestion
import { compatibilityFlags } from 'workerd:compatibility-flags';
```

src/workerd/api/node/node.h
```diff
@@ -23,7 +24,51 @@ class CompatibilityFlags : public jsg::Object {
...
+        for (auto annotation : annotations) {
+          if (annotation.getId() == EXPERIMENTAl_ANNOTATION_ID) {
+            skip = true;
+            break;
+          }
+        }
+      }
+      if (skip) continue;
+
+      bool value = dynamic.get(field).as<bool>();
```
anonrig, line 53: Since type is defined, does this work?

```suggestion
      bool value = dynamic.get(field).as();
```

src/workerd/api/node/node.h
```diff
@@ -23,7 +24,51 @@ class CompatibilityFlags : public jsg::Object {
...
+      // is enabled.
+      auto annotations = field.getProto().getAnnotations();
+      bool skip = false;
+      if (skipExperimental) {
+        for (auto annotation : annotations) {
+          if (annotation.getId() == EXPERIMENTAl_ANNOTATION_ID) {
+            skip = true;
+            break;
+          }
+        }
```
anonrig, line 49: This can be simplified using something like `std::find_if` depending on type type of `annotations`. Does KJ support something like this?

src/workerd/api/node/node.h
```diff
@@ -23,7 +24,51 @@ class CompatibilityFlags : public jsg::Object {
...
+      }
+    }
+
+    obj.seal(js);
+    return obj;
+  }
+
   JSG_RESOURCE_TYPE(CompatibilityFlags, workerd::CompatibilityFlags::Reader flags) {
+    JSG_LAZY_READONLY_INSTANCE_PROPERTY(compatibilityFlags, getCompatibilityFlags);
```
anonrig, line 70: Does this mean the result will always be the same, hence caching?

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,38 @@
...
+import { compatibilityFlags } from 'cloudflare:workers';
+
+const kOn = 'on';
+const kOff = 'off';
+
+export const compatFlagsTest = {
+  test() {
+    throws(() => compatibilityFlags.no_nodejs_compat_v2 = "...");
+    throws(() => compatibilityFlags.not_a_real_compat_flag = "...");
+//    throws(() => { delete compatibilityFlags['nodejs_compat_v2']; });
```
anonrig, line 16:
```suggestion
```

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,38 @@
...
+
+    strictEqual(compatibilityFlags['nodejs_compat_v2'], kOn);
+    strictEqual(compatibilityFlags['no_nodejs_compat_v2'], kOff);
+    strictEqual(compatibilityFlags['url_standard'], kOn);
```
anonrig, line 20: the `wd-test` does not have this compatibility flag, how does this return true?

src/cloudflare/workers.ts
jasnell: No, this doesn't work and the eslint-disable complains

src/workerd/api/node/node.h
jasnell: Probably but I prefer the more verbose option here.

src/workerd/api/node/node.h
jasnell: essentially yes. If this were `JSG_READONLY_INSTANCE_PROPERTY`, then the getter would be called every time the property is accessed, which is wasteful. By using the lazy option, v8 will call the getter only once and then will replace the getter in the internal property definition with the returned value so the getter is never called again.

src/workerd/api/tests/compat-flags-test.js
jasnell: Actually, this should go the other direction... the line should not be commented out.

src/workerd/api/tests/compat-flags-test.js
jasnell: Because it is on by default based on the compat date configured for the worker.

src/workerd/api/node/node.h
jasnell: Probably.. but let's get agreement on the basic approach before applying optional optimizations/simplifications.

0001-Implements-cloudflare-compatibility-flags-API.patch
```diff
@@ -0,0 +1,161 @@
+From fdac25bc6ad66aed94bf7e89e0f2afc1158c32d8 Mon Sep 17 00:00:00 2001
```
npaun, line 1: File accidentally added?

src/workerd/api/node/node.h
anonrig: Ah nice. Thanks for the explanation!

0001-Implements-cloudflare-compatibility-flags-API.patch
jasnell: doh

src/workerd/api/node/node.h
```diff
@@ -12,6 +12,7 @@
 #include <workerd/jsg/modules-new.h>
 #include <capnp/dynamic.h>
 #include <node/node.capnp.h>
+#include <workerd/io/compatibility-date.h>
```
jasnell, line 15:
```suggestion
```

jasnell:
This PR should be ready for review.

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,41 @@
...
+const kOff = 'off';
+
+const { compatibilityFlags } = globalThis.Cloudflare;
+ok(Object.isSealed(compatibilityFlags));
+ok(Object.isFrozen(compatibilityFlags));
+ok(!Object.isExtensible(compatibilityFlags));
+
+export const compatFlagsTest = {
+  test() {
+    throws(() => compatibilityFlags.no_nodejs_compat_v2 = "...");
```
anonrig, line 17: Can you add a test for no proto as well?

anonrig (approved):

dario-piotrowicz:
Could/should the `Cloudflare` type be improved?

This is what this branch currently generates:
Screenshot 2024-08-15 at 16 59 16

jasnell:
> Could/should the Cloudflare type be improved?

Try now.

src/workerd/api/global-scope.h
```diff
@@ -106,6 +106,24 @@ class Performance: public jsg::Object {
...
+public:
+
+  // Return an object containing the state of all compatibility flags known to the runtime.
+  jsg::JsObject getCompatibilityFlags(jsg::Lock& js);
+
+  JSG_RESOURCE_TYPE(Cloudflare) {
+    JSG_LAZY_READONLY_INSTANCE_PROPERTY(compatibilityFlags, getCompatibilityFlags);
+
+    JSG_TS_OVERRIDE({
+      readonly compatibilityFlags: Record<string, string>;
```
dario-piotrowicz, line 122: what about a union?

```suggestion
      readonly compatibilityFlags: Record<string, 'on'|'off'>;
```

dario-piotrowicz:
> > Could/should the Cloudflare type be improved?
> 
> Try now. Updated the value to be `Record<string, string>`

Yes, looks better now 😄 
(but as I mentioned in my other comment maybe we'd want to have the type more restrictive? 😃)

Screenshot 2024-08-15 at 18 19 03

dario-piotrowicz (approved):
Looks good to me 🙂 (from just a js/ts point of view)

jasnell:
This is otherwise ready to land but going to hold off until @kentonv signs off.

mhart:
This looks great!

Given there's some complexity around `no_*` and `on` vs `off`, I think we'll need to provide clear guidance on how this should be used. ie, how we recommend people use it – and make sure it covers all the use cases it needs to.

My understanding is that 99% of the time, you're just going to want to check if a particular flag `=== "on"`, correct?

And all the flags listed here will be present: https://developers.cloudflare.com/workers/configuration/compatibility-dates/, correct?

And by default all the `Flag to enable` ones will be `on`, correct?

jasnell:
Not sure I completely follow the question so hopefully this answers it by way of example:

The `url_standard` and `url_legacy` flags are mutually exclusive. If `url_standard` is `'on'`, `url_legacy` will be `'off'`

If we have a `foo` and `no_foo`, if `foo` is set it will be `'on'` and `no_foo` will be `'off'`. If `no_foo` is set, `no_foo` will be `'on'` and `foo` will be `off`.

And yes, tor the most part, most users will generally only be checking for `=== 'on'`. The `'off'` is largely just there to reliably differentiate between a flag that exists but is off vs. a flag that does not exist.

dario-piotrowicz:
@jasnell sorry if it's a silly question... I was just wondering, would there be any benefit in also exposing the compat date?

or is it pretty unnecessary since the compat date basically just dictates the compat flags? (so those are ultimately all the ones you need to know)

jasnell:
I was debating the same thing. Short answer is I don't know. I'm not sure it's going to be that useful, certainly possible tho. I wouldn't do it in this PR tho.

dario-piotrowicz:
> I was debating the same thing. Short answer is I don't know. I'm not sure it's going to be that useful, certainly possible tho. I wouldn't do it in this PR tho.

ok thanks, I was just curious 🙂👍

mhart:
> And all the flags listed here will be present: https://developers.cloudflare.com/workers/configuration/compatibility-dates/, correct?

@jasnell and is this correct? ie, all the flags listed there will be available on this object, right? Regardless of whether a user has specified them in their wrangler.toml.

jasnell:
Yep, all of those should be available. All flags known to the system will be listed.

mhart:
> And by default all the `Flag to enable` ones will be `on`, correct?

Great! And this ^ is also true, right?

ie, by default, all of the flags listed on that page will have `on` for the property listed next to `Flag to enable` (ie, `export_commonjs_default` in the screenshot below)

Screenshot 2024-08-16 at 9 25 19 AM

Again, I'm just being super careful we're making this as clear as possible to users – and that there's not some weird edge cases or anything.

mhart:
Sorry, with "by default" – I mean "by default if you use the latest compatibility date"

jasnell:
For any flag that has a "Default as of" date, if the configured compat date for the worker is on or after the default on date, then yes, the related "Flag to enable" will be `'on'` and the "Flag to disable" will be `'off`'.

mhart (approved):

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,50 @@
+import {
+  ok,
+  strictEqual,
+  throws,
+} from 'node:assert';
+
+const kOn = 'on';
+const kOff = 'off';
```
IgorMinar, line 8: why not boolean? this makes the API incompatible with current tree-shaking tools. could we use booleans please?

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,50 @@
...
+    throws(() => compatibilityFlags.no_nodejs_compat_v2 = "...");
+    throws(() => compatibilityFlags.not_a_real_compat_flag = "...");
+    throws(() => { delete compatibilityFlags['nodejs_compat_v2']; });
+
+    // The compatibility flags object should have no prototype.
+    strictEqual(Object.getPrototypeOf(compatibilityFlags), null);
+
+    // The compatibility flags object should have the expected properties.
+    strictEqual(compatibilityFlags['nodejs_compat_v2'], kOn);
+    strictEqual(compatibilityFlags['no_nodejs_compat_v2'], kOff);
```
IgorMinar, line 30: I like that you added support `'no_'` variant as well - this will improve code readability in some cases.

IgorMinar (changes requested):
I'm very supportive of this feature, but let's please not rush merging this in.

Could we please settle the following first:

- [ ] why do we care about CJS support? (in other words, why not `import.meta.cloudflare`) https://x.com/jasnell/status/1824114538272411825 
- [x] why capital `C` in "globalThis.Cloudflare"? As far as I can think of, it is customary to use lower-case letters in JS for variables like this one, for example `globalThis.navigator` or `globalThis.process`.
- [ ] are we intentionally not exposing the `compatibility_date`? I'm unsure if we should, so I'm mainly wondering if the omission is intentional.
- [ ] can you please document the plan for typescript support? without a design doc, it's not clear to me if this will just work or if we need to do anything special. I expect that we'll need to change RTT to support dynamic type generation that is currently in works by @andyjessop 
- [ ] can we ensure that these flags use an API that enable tree-shaking of the code, I don't believe that's the case now due to `'on'` and `'off'` values instead of using `boolean` type

mhart:
> * why capital `C` in "globalThis.Cloudflare"? As far as I can think of, it is customary to use lower-case letters in JS for variables like this one, for example `globalThis.navigator` or `globalThis.process`.

1. There's prior art here with `globalThis.Deno` and `globalThis.Bun` – [lots of code checks for compatibility this way](https://github.com/ije/md4w/blob/483ff363e2e767b440884b750ec973e185ce27f8/js/index.js#L5-L25) – so we have consistency there and won't stick out as the odd one out.
2. There's [some (not much) existing code that uses lowercase `globalThis.cloudflare`](https://github.com/search?q=globalThis.Cloudflare&type=code)

src/workerd/api/tests/compat-flags-test.js
mhart: Can you explain why I boolean would make a difference for tree-shaking?

Are you saying this would be tree-shaken:

```js
if (globalThis.Cloudflare?.compatibilityFlags.nodejs_compat_v2)
```

But this wouldn't?

```js
if (globalThis.Cloudflare?.compatibilityFlags.nodejs_compat_v2 === "on")
```

IgorMinar:
> > * why capital `C` in "globalThis.Cloudflare"? As far as I can think of, it is customary to use lower-case letters in JS for variables like this one, for example `globalThis.navigator` or `globalThis.process`.
> 
> 1. There's prior art here with `globalThis.Deno` and `globalThis.Bun` – [lots of code checks for compatibility this way](https://github.com/ije/md4w/blob/483ff363e2e767b440884b750ec973e185ce27f8/js/index.js#L5-L25) – so we have consistency there and won't stick out as the odd one out.
> 2. There's [some (not much) existing code that uses lowercase `globalThis.cloudflare`](https://github.com/search?q=globalThis.Cloudflare&type=code)
> 3. Unlike `navigator` and `process`, this is going to introduce a _non-standard_ property, so it's better that it diverges from standard norms

These are good argument for capital C. Thanks @mhart

src/workerd/api/tests/compat-flags-test.js
jasnell: > ... why not boolean?

Largely due to @kentonv's argument about being able to tell the difference between a disabled flag vs. a non-existent one. Specifically, using a boolean encourages code like `if (!compatFlags['foo']) { ... }` which will be indistinguishable between `undefined` and explicitly `false`. By ensuring that flags that are recognized-but-explicitly-off still use a truthy value we make these checks more reliable and explicit. Only non-existent flags will be `undefined`.

The only other way to address this would be to explicitly throw an error for non-existent flags, which requires more boilerplate and complexity to get right (module developers would need to worry about an additional error being thrown)

IgorMinar: I should clarify that string values are supported as well by rollup and esbuild: https://esbuild.github.io/api/#define

However they are more prone to coding patterns that are not (reliably) tree-shakable for example:

```
if (globalThis.Cloudflare?.compatibilityFlags.nodejs_compat_v2 === "o" + "n")
```

or 

```
if (globalThis.Cloudflare?.compatibilityFlags.nodejs_compat_v2 === onFlag())
```

Additionally, the string values literally translate to `true` and `false`, and the flag system is designed  with `boolean`s at heart (hence the `no_` support). So using string `'on'` and `'off'` seems just to add no benefits, and only makes for a more verbose and clunky API.

jasnell:
> ... but let's please not rush merging this in.

There is no rush.

> why do we care about CJS support? (in other words, why not import.meta.cloudflare)

The vast majority of modules on npm are still commonjs and will remain so for an indefinite amount of time. `globalThis.Cloudflare` means we can support any module, CJS or ESM. `import.meta.cloudflare` would mean we are intentionally choosing to ignore the majority of the very ecosystem we are working to become more compatible with.

`globalThis.Cloudflare` is also consistent with what other runtimes have done (e.g. `globalThis.Deno`, `globalThis.Bun`, etc)

> are we intentionally not exposing the compatibility_date? I'm unsure if we should, so I'm mainly wondering if the omission is intentional.

https://github.com/cloudflare/workerd/pull/2527#issuecomment-2292438196 ....

short answer is: I'm definitely open to adding it if there's a useful use case for it.

> can you please document the plan for typescript support? without a design doc, it's not clear to me if this will just work or if we need to do anything special. I expect that we'll need to change RTT to support dynamic type generation that is currently in works

The type of `globalThis.Cloudflare.compatibilityFlags` is `Record<string, 'on'|'off'>` ... is there anything more than that needed? If our types need to include all of the actual flag names then the type generation logic could fairly easily extract those, but I'm not super convinced it needs to be any more complicated than `Record<string, 'on'|'off'>`

> can we ensure that these flags use an API that enable tree-shaking of the code, I don't believe that's the case now due to 'on' and 'off' values instead of using boolean type

Can you explain a bit more why the current proposed API doesn't work for tree shaking?

src/workerd/api/tests/compat-flags-test.js
jasnell: https://esbuild.github.io/try/#dAAwLjIzLjAALS1kZWZpbmU6Z2xvYmFsVGhpcy5DbG91ZGZsYXJlLmNvbXBhdGliaWxpdHlGbGFncy5mb289IidvbiciIC0tdHJlZS1zaGFraW5nPXRydWUAZnVuY3Rpb24gZm9vKCkge30KZnVuY3Rpb24gYmFyKCkge30KCmlmIChnbG9iYWxUaGlzLkNsb3VkZmxhcmUuY29tcGF0aWJpbGl0eUZsYWdzLmZvbyA9PT0gJ29uJykgewogIGZvbygpOwp9IGVsc2UgewogIGJhcigpOwp9 illustrates that tree shaking is certainly possible with esbuild with the string values. Yes, there are ways to write the code that causing tree shaking to fail that's also possible with a `boolean` value also.

kentonv:
I think we should go back to boolean values.

With the string on/off values, since it is unintuitive, people are likely to forget and write truthy checks instead, and then spend hours trying to figure out why workerd is ignoring them when they try to configure the flag to be off in their tests.

I don't think the string values solve the problem with negative flags. People will still write `if (Cloudflare.compatibilityFlags.no_foo === 'on') { ... }`, which will still do the wrong thing when running on an old version of workerd where the flag doesn't exist yet.

I think there are really two ways forward on that:
1. Don't include negative flags, only positive ones. Though that comes with its own confusion. Perhaps we want the negative flags to exist but throw an exception when accessed saying to please use the positive version instead?
2. Just live with it. If people are running old versions of workerd they may run into libraries that incorrectly check for newer features, and will have to file bugs to get those libraries to reverse the check.

I could be convinced either way.

> are we intentionally not exposing the compatibility_date?

I would argue against exposing it, otherwise people might check against the date instead of against the specific flag, which would work most of the time but would be incorrect. Also, at present we don't actually have the compat date available on the edge -- it is consumed by the config service and converted to a boolean list before deployment.

dario-piotrowicz:
What about something like this?
```ts
type CompatibilityFlags = {
  enabled: Set<string>;
  disabled: Set<string>;
};

interface Cloudflare {
  compatibilityFlag: CompatibilityFlags;
}
```

Seems very clear and not easy to get wrong to me, plus with [`union`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/union)s it would also be very straightforward to get all possible flags

kentonv:
@dario-piotrowicz That still has the same backwards-compatibility problem. If I write `if (Cloudflare.compatibilityFlags.enabled.get("someNegativeFlag") {` and run it on an old version of workerd it will incorrectly treat the flag as disabled. (Or if I write `if (Cloudflare.compatibilityFlags.disabled.get("somePositiveFlag")`.)

jasnell:
PR updated with the changes:

1. Values switch back to boolean
2. The `compatibilityFlags` keys only include the enable flags whose values are explicitly `true`. See the comments in the updated test for more explanation.

src/workerd/api/global-scope.h
```diff
@@ -106,6 +106,24 @@ class Performance: public jsg::Object {
...
+public:
+
+  // Return an object containing the state of all compatibility flags known to the runtime.
+  jsg::JsObject getCompatibilityFlags(jsg::Lock& js);
+
+  JSG_RESOURCE_TYPE(Cloudflare) {
+    JSG_LAZY_READONLY_INSTANCE_PROPERTY(compatibilityFlags, getCompatibilityFlags);
+
+    JSG_TS_OVERRIDE({
+      readonly compatibilityFlags: Record<string, 'on'|'off'>;
```
anonrig, line 122:
```suggestion
      readonly compatibilityFlags: Record<string, boolean>;
```

anonrig (approved):

anonrig (approved):

kentonv:
> The compatibilityFlags keys only include the enable flags whose values are explicitly true. See the comments in the updated test for more explanation.

Why? That means you can't check `in` to see if the workerd version knows about the flag.

Not necessarily saying I disagree with the change but was wondering what the motivation is. Sorry if I missed it somewhere in the long thread.

jasnell:
@kentonv:
> Why? That means you can't chec `in` to see if the workerd version knows about the flag.

https://github.com/cloudflare/workerd/pull/2527#issuecomment-2293687482 Bullet point 1

I'd rather not have an API that throws on a known property getter as that carries its own DX hurdles.

kentonv:
> > Why? That means you can't chec in to see if the workerd version knows about the flag.
>
> https://github.com/cloudflare/workerd/pull/2527#issuecomment-2293687482 Bullet point 1
>
> I'd rather not have an API that throws on a known property getter as that carries its own DX hurdles.

Oh, but this isn't what I was asking for there.

By "negative flags" I didn't mean flags whose value is false, I meant flags declared with `$compatDisableFlag` (as opposed to `$compatEnableFlag`). Maybe I should have said "disable flags" instead of "negative flags". Disable flags are the ones that have issues with backwards-compatibility, because on old versions they will be undefined, which is falsy, even though the flag should logically be treated as true for old versions.

My suggestion is that we do not include disable-flags in the API at all, and instead tell people to test for the enable-flag, so that treating undefined as false is correct.

jasnell:
Updated to include all enable flags.

jasnell:
Rebased on master to resolve merge conflicts after the formatting change.

kentonv (approved):
The logic regarding experimental flags seems odd to me but not enough to hold this up further.

jasnell:
Ok, just waiting on a final review from @IgorMinar

IgorMinar:
> My suggestion is that we do not include disable-flags in the API at all, and instead tell people to test for the enable-flag, so that treating undefined as false is correct.

Exactly!!! I'm reviewing the lastest, but I'm so happy that this is where we ended up. Full review soon.

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+ok(!Object.isExtensible(compatibilityFlags));
+
+// It shuld be possible to shadow the Cloudflare global
+const Cloudflare = 1;
+strictEqual(Cloudflare, 1);
+
+export const compatFlagsTest = {
+  test() {
+    // The compatibility flags object should be sealed, frozen, and not extensible.
+    throws(() => (compatibilityFlags.no_nodejs_compat_v2 = '...'));
```
IgorMinar, line 15: now that we don't support disable flags, this test throws for a reason that is different than when the test was originally written, and is the same assertion as `not_a_real_compat_flag`. I suggest you either add a comment to say that you are explicitly testing for disable flags which should work as non-existent flags, or remove this line altogether.

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+    // If a key does not appear, it can mean one of three things:
+    // 1. It is a disable flag.
+    // 2. It is an experimental flag and the experimental option is not set.
+    // 3. The flag does not exist.
+    //
+    // At this level we make no attempt to differentiate between these cases
+    ok(compatibilityFlags['nodejs_compat_v2']);
+    ok(compatibilityFlags['url_standard']);
+
+    ok(!compatibilityFlags['no_nodejs_compat_v2']);
```
IgorMinar, line 37: this assertion doesn't prove that the value is undefined. I suggest that you test using `in` rather than by potentially coercing the return value via `!`

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+    ok(!compatibilityFlags['url_original']);
+    strictEqual(compatibilityFlags['no_nodejs_compat_v2'], undefined);
```
IgorMinar, line 39: oh, given this test, I'd remove all the `ok(!` assertions above as they only test that javascript works as expected.

Alternatively, it might be valuable to test for a **lack** of presence of the property descriptor via `in` or `getOwnPropertyDescriptor`.

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+    strictEqual(compatibilityFlags['url_original'], undefined);
```
IgorMinar, line 40: the same as above

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+
+    // Since we are not specifying the experimental flag, experimental flags should
+    // not be included in the output.
+    strictEqual(compatibilityFlags['durable_object_rename'], undefined);
```
IgorMinar, line 44: could you instead assert that the property descriptor is missing? it's a more solid assertion.

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+    strictEqual('durable_object_rename' in compatibilityFlags, false);
```
IgorMinar, line 45: ah.. you do, so I'd remove the first one as that only tests javascript's behavior and not our use of it.

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+
+    // If a flag does not exist, the value will be undefined.
+    strictEqual(compatibilityFlags['not-a-real-compat-flag'], undefined);
```
IgorMinar, line 48: same as above, I'd remove

src/workerd/api/tests/compat-flags-test.js
```diff
@@ -0,0 +1,58 @@
...
+    strictEqual('not-a-real-compat-flag' in compatibilityFlags, false);
+
+    // The compatibility flags object should have the expected keys.
+    const keys = Object.keys(compatibilityFlags);
+    ok(keys.includes('nodejs_compat_v2'));
+    ok(keys.includes('url_standard'));
+    ok(!keys.includes('url_original'));
+    ok(!keys.includes('not-a-real-compat-flag'));
+  },
```
IgorMinar, line 57: yeah. this is great!

IgorMinar (approved):
I like this version of flags api a lot! thank you @jasnell and all the reviewers for patiently iterating on this. I no longer have any concerns and I'm happy with this being merged.

I would still love to get an ok from the @andyjessop or someone (@penalosa, @petebacondarwin,
 @lrapoport-cf ) from the wrangler team on the typescript support as I requested a while back.

I do, however, think the API is simple enough now and TS support should be non-controversial and unless they review quickly we don't need to wait for them and instead to a post-merge review.

PR #2532, opened 2024-08-14
more user-friendly jsg exception message

mikea:
Strictly speaking this is not backward compatible, but I doubt anyone cares and it has saved me some time while debugging an issue.

### DIFF
--- a/src/workerd/jsg/struct.h
+++ b/src/workerd/jsg/struct.h
@@ -103,8 +103,9 @@ public:
                      kj::isSameType<typename FieldWrappers::Type, Unimplemented>()) && ...)) {
         return T{};
       }
-      jsg::throwTypeError(isolate, "Cannot initialize a dictionary with required members from an "
-                                    "undefined or null value.");
+      jsg::throwTypeError(isolate, kj::str("Cannot initialize ", typeid(T).name(),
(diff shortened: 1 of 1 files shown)


anonrig (approved):

src/workerd/jsg/struct.h
```diff
@@ -103,8 +103,9 @@ class StructWrapper<Self, T, TypeTuple<FieldWrappers...>, kj::_::Indexes<indices
                      kj::isSameType<typename FieldWrappers::Type, Unimplemented>()) && ...)) {
         return T{};
       }
-      jsg::throwTypeError(isolate, "Cannot initialize a dictionary with required members from an "
-                                    "undefined or null value.");
+      jsg::throwTypeError(isolate, kj::str("Cannot initialize ", typeid(T).name(),
```
jasnell, line 106: minor nit but is this going to give a user friendly name that is useful for end users or something only we can appreciate?

jasnell (approved):

npaun (approved):

src/workerd/jsg/struct.h
mikea: Friendliness is debatable, but this is definitely something that will have a name similar to TS type.

PR #2533, opened 2024-08-14
zlib one-shot methods

npaun:
- Reuses ZlibContext from @anonrig's implementation of the stream-based API.
- Sends the entire input buffer at once, and uses only the finishFlush flag, like Node's does.
- Uses GrowableBuffer to collect the output as it's generated. This is basically kj::Vector, but it can be limited to a maximum size and adds some methods to make usage as a buffer more ergonomic.

### DIFF
--- a/src/node/internal/internal_zlib.ts
+++ b/src/node/internal/internal_zlib.ts
@@ -3,11 +3,14 @@
 //     https://opensource.org/licenses/Apache-2.0
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 
-import { default as zlibUtil, type ZlibOptions } from 'node-internal:zlib';
+import {
+  default as zlibUtil,
+  type ZlibOptions,
+  type CompressCallback,
+} from 'node-internal:zlib';
 import { Buffer } from 'node-internal:internal_buffer';
 import { validateUint32 } from 'node-internal:validators';
 import { ERR_INVALID_ARG_TYPE } from 'node-internal:internal_errors';
-import { isArrayBufferView } from 'node-internal:internal_types';
 import { Zlib } from 'node-internal:internal_zlib_base';
 
 const {
@@ -19,6 +22,253 @@ const {
   CONST_GZIP,
   CONST_UNZIP,
 } = zlibUtil;
+
+export function crc32(
+  data: ArrayBufferView | string,
+  value: number = 0
+): number {
+  validateUint32(value, 'value');
+  return zlibUtil.crc32(data, value);
+}
+
+export function inflateSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {}
+): Buffer {
+  return Buffer.from(zlibUtil.zlibSync(data, options, zlibUtil.CONST_INFLATE));
+}
+
+export function deflateSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {}
+): Buffer {
+  return Buffer.from(zlibUtil.zlibSync(data, options, zlibUtil.CONST_DEFLATE));
+}
+
+export function gunzipSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {}
+): Buffer {
+  return Buffer.from(zlibUtil.zlibSync(data, options, zlibUtil.CONST_GUNZIP));
+}
+
+export function gzipSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {}
+): Buffer {
+  return Buffer.from(zlibUtil.zlibSync(data, options, zlibUtil.CONST_GZIP));
+}
+
+export function inflateRawSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {}
+): Buffer {
+  return Buffer.from(
+    zlibUtil.zlibSync(data, options, zlibUtil.CONST_INFLATERAW)
+  );
+}
+
+export function deflateRawSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {}
+): Buffer {
+  return Buffer.from(
+    zlibUtil.zlibSync(data, options, zlibUtil.CONST_DEFLATERAW)
+  );
+}
+
+export function unzipSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {}
+): Buffer {
+  return Buffer.from(
+    zlibUtil.zlibSync(data, options, zlibUtil.CONST_DEFLATERAW)
+  );
+}
+
+function normalizeArgs(
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): [ZlibOptions, CompressCallback<Error, Buffer>] {
+  if (typeof optionsOrCallback === 'function') {
+    return [{}, optionsOrCallback];
+  } else if (typeof callbackOrUndefined === 'function') {
+    return [optionsOrCallback, callbackOrUndefined];
+  }
+
+  throw new ERR_INVALID_ARG_TYPE('callback', 'Function', callbackOrUndefined);
+}
+
+function wrapCallback(
+  callback: CompressCallback<Error, Buffer>
+): CompressCallback<string, ArrayBuffer> {
+  return (error: string | undefined, result: ArrayBuffer | undefined) => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    queueMicrotask(() => {
+      callback(
+        error ? new Error(error) : undefined,
+        result ? Buffer.from(result) : undefined
+      );
+    });
+  };
+}
+
+export function inflate(
+  data: ArrayBufferView | string,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function inflate(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function inflate(
+  data: ArrayBufferView | string,
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): void {
+  const [options, callback] = normalizeArgs(
+    optionsOrCallback,
+    callbackOrUndefined
+  );
+  zlibUtil.zlib(data, options, zlibUtil.CONST_INFLATE, wrapCallback(callback));
+}
+
+export function unzip(
+  data: ArrayBufferView | string,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function unzip(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function unzip(
+  data: ArrayBufferView | string,
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): void {
+  const [options, callback] = normalizeArgs(
+    optionsOrCallback,
+    callbackOrUndefined
+  );
+  zlibUtil.zlib(data, options, zlibUtil.CONST_UNZIP, wrapCallback(callback));
+}
+
+export function inflateRaw(
+  data: ArrayBufferView | string,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function inflateRaw(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function inflateRaw(
+  data: ArrayBufferView | string,
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): void {
+  const [options, callback] = normalizeArgs(
+    optionsOrCallback,
+    callbackOrUndefined
+  );
+  zlibUtil.zlib(
+    data,
+    options,
+    zlibUtil.CONST_INFLATERAW,
+    wrapCallback(callback)
+  );
+}
+
+export function gunzip(
+  data: ArrayBufferView | string,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function gunzip(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function gunzip(
+  data: ArrayBufferView | string,
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): void {
+  const [options, callback] = normalizeArgs(
+    optionsOrCallback,
+    callbackOrUndefined
+  );
+  zlibUtil.zlib(data, options, zlibUtil.CONST_GUNZIP, wrapCallback(callback));
+}
+
+export function deflate(
+  data: ArrayBufferView | string,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function deflate(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function deflate(
+  data: ArrayBufferView | string,
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): void {
+  const [options, callback] = normalizeArgs(
+    optionsOrCallback,
+    callbackOrUndefined
+  );
+  zlibUtil.zlib(data, options, zlibUtil.CONST_DEFLATE, wrapCallback(callback));
+}
+
+export function deflateRaw(
+  data: ArrayBufferView | string,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function deflateRaw(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function deflateRaw(
+  data: ArrayBufferView | string,
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): void {
+  const [options, callback] = normalizeArgs(
+    optionsOrCallback,
+    callbackOrUndefined
+  );
+  zlibUtil.zlib(
+    data,
+    options,
+    zlibUtil.CONST_DEFLATERAW,
+    wrapCallback(callback)
+  );
+}
+
+export function gzip(
+  data: ArrayBufferView | string,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function gzip(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  callback: CompressCallback<Error, Buffer>
+): void;
+export function gzip(
+  data: ArrayBufferView | string,
+  optionsOrCallback: ZlibOptions | CompressCallback<Error, Buffer>,
+  callbackOrUndefined?: CompressCallback<Error, Buffer>
+): void {
+  const [options, callback] = normalizeArgs(
+    optionsOrCallback,
+    callbackOrUndefined
+  );
+  zlibUtil.zlib(data, options, zlibUtil.CONST_GZIP, wrapCallback(callback));
+}
 
 const constPrefix = 'CONST_';
 export const constants: Record<string, number> = {};
@@ -40,19 +290,6 @@ Object.defineProperties(
       ])
   )
 );
-
-export function crc32(
-  data: ArrayBufferView | string,
-  value: number = 0
-): number {
-  if (typeof data === 'string') {
-    data = Buffer.from(data);
-  } else if (!isArrayBufferView(data)) {
-    throw new ERR_INVALID_ARG_TYPE('data', 'ArrayBufferView', typeof data);
-  }
-  validateUint32(value, 'value');
-  return zlibUtil.crc32(data, value);
-}
 
 export class Gzip extends Zlib {
   public constructor(options: ZlibOptions) {
--- a/src/node/internal/zlib.d.ts
+++ b/src/node/internal/zlib.d.ts
@@ -2,6 +2,24 @@ import { owner_symbol, type Zlib } from 'node-internal:internal_zlib_base';
 
 export function crc32(data: ArrayBufferView, value: number): number;
 
+export type CompressCallback<ErrT, BufT> = (
+  error?: ErrT,
+  result?: BufT
+) => void;
+
+export function crc32(data: ArrayBufferView | string, value: number): number;
+export function zlibSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  mode: number
+): ArrayBuffer;
+export function zlib(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  mode: number,
+  cb: CompressCallback<string, ArrayBuffer>
+): ArrayBuffer;
+
 // zlib.constants (part of the API contract for node:zlib)
 export const CONST_Z_NO_FLUSH: number;
 export const CONST_Z_PARTIAL_FLUSH: number;
--- a/src/node/zlib.ts
+++ b/src/node/zlib.ts
@@ -29,6 +29,24 @@ const createInflate = protectMethod(zlib.createInflate);
 const createInflateRaw = protectMethod(zlib.createInflateRaw);
 const createUnzip = protectMethod(zlib.createUnzip);
 
+const inflate = protectMethod(zlib.inflate);
+const inflateSync = protectMethod(zlib.inflateSync);
+const deflate = protectMethod(zlib.deflate);
+const deflateSync = protectMethod(zlib.deflateSync);
+
+const inflateRaw = protectMethod(zlib.inflateRaw);
+const inflateRawSync = protectMethod(zlib.inflateRawSync);
+const deflateRaw = protectMethod(zlib.deflateRaw);
+const deflateRawSync = protectMethod(zlib.deflateRawSync);
+
+const gzip = protectMethod(zlib.gzip);
+const gzipSync = protectMethod(zlib.gzipSync);
+const gunzip = protectMethod(zlib.gunzip);
+const gunzipSync = protectMethod(zlib.gunzipSync);
+
+const unzip = protectMethod(zlib.unzip);
+const unzipSync = protectMethod(zlib.unzipSync);
+
 export {
   crc32,
   constants,
@@ -50,6 +68,22 @@ export {
   createInflate,
   createInflateRaw,
   createUnzip,
+
+  // One-shot methods
+  inflate,
+  inflateSync,
+  deflate,
+  deflateSync,
+  inflateRaw,
+  inflateRawSync,
+  deflateRaw,
+  deflateRawSync,
+  gzip,
+  gzipSync,
+  gunzip,
+  gunzipSync,
+  unzip,
+  unzipSync,
 };
 
 export default {
@@ -73,4 +107,20 @@ export default {
   createInflate,
   createInflateRaw,
   createUnzip,
+
+  // One-shot methods
+  inflate,
+  inflateSync,
+  deflate,
+  deflateSync,
+  inflateRaw,
+  inflateRawSync,
+  deflateRaw,
+  deflateRawSync,
+  gzip,
+  gzipSync,
+  gunzip,
+  gunzipSync,
+  unzip,
+  unzipSync,
 };
--- a/src/workerd/api/node/tests/zlib-nodejs-test.js
+++ b/src/workerd/api/node/tests/zlib-nodejs-test.js
@@ -239,12 +239,9 @@ export const crc32Test = {
     }
 
     [undefined, null, true, 1, () => {}, {}].forEach((invalid) => {
-      throws(
-        () => {
-          zlib.crc32(invalid);
-        },
-        { code: 'ERR_INVALID_ARG_TYPE' }
-      );
+      throws(() => {
+        zlib.crc32(invalid);
+      }, new TypeError("Failed to execute 'crc32' on 'ZlibUtil': parameter 1 is not of type 'string or ArrayBuffer or ArrayBufferView'."));
     });
 
     [null, true, () => {}, {}].forEach((invalid) => {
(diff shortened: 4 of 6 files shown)


src/workerd/api/node/zlib-util.c++
```diff
@@ -3,9 +3,159 @@
 //     https://opensource.org/licenses/Apache-2.0
 #include "zlib-util.h"
 
+#include <workerd/jsg/exception.h>
+
+#include <zlib.h>
+
+#include <iostream>
```
jasnell, line 10:
```suggestion
#include <workerd/jsg/exception.h>
#include <zlib.h>
#include <iostream>
```

src/workerd/api/node/zlib-util.c++
```diff
@@ -3,9 +3,159 @@
...
+    stream.avail_in = data.size();
+    auto status = impl().init();
+    JSG_REQUIRE(status == Z_OK, Error, "Failed to initialize zlib");
+  }
+
+  kj::Array<kj::byte> processBuffer() {
+    int status = Z_OK;
+
+    do {
+      result.growBy(ZLIB_OUT_BUF_SIZE);
```
jasnell, line 55: This should incorporate a check to make sure the `result` does not grow beyond a reasonable limit (I'd say max `128 * 1024`).

src/workerd/api/node/zlib-util.c++
```diff
@@ -3,9 +3,159 @@
...
+      stream.next_out = result.writePos();
+      stream.avail_out = result.available();
+
+      status = impl().processOnce();
+      switch (status) {
```
jasnell, line 60:
```suggestion
      switch (impl().processOnce()) {
```

src/workerd/api/node/zlib-util.h
```diff
@@ -201,6 +216,6 @@ class ZlibUtil final : public jsg::Object {
   };
 };
 
-#define EW_NODE_ZLIB_ISOLATE_TYPES api::node::ZlibUtil
+#define EW_NODE_ZLIB_ISOLATE_TYPES api::node::ZlibUtil, api::node::ZlibUtil::Options
```
jasnell, line 219:
```suggestion
#define EW_NODE_ZLIB_ISOLATE_TYPES \
  api::node::ZlibUtil, \
  api::node::ZlibUtil::Options
```

nit: just to match the formatting style we use throughout the rest of the code

src/workerd/api/node/zlib-util.c++
npaun: ughhh.... can't wait for clang-format to be merged

src/workerd/api/node/zlib-util.h
npaun: Our clang format config will squash this back down to one line

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +21,149 @@ const {
...
+  // FIXME: not actually implemented
+}
+
+export function inflate(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {},
+  callback: CompressCallback
+): void {
+  zlibUtil.zlibWithCb(data, options, zlibUtil.CONST_INFLATE, (error, result) =>
+    callback(error ? Error(error) : null, result ? Buffer.from(result) : null)
```
jasnell, line 94: In these cases, in order to best match node.js the closest we can, these callbacks should likely be deferred using `queueMicrotask(...)`

```suggestion
    queueMicrotask(() => callback(error ? Error(error) : null, result ? Buffer.from(result) : null));
```

src/node/internal/zlib.d.ts
```diff
@@ -2,6 +2,24 @@ import { owner_symbol, type Zlib } from 'node-internal:internal_zlib_base';
...
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  mode: number
+): ArrayBuffer;
+export function zlibWithCb(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  mode: number,
+  cb: CompressCallback
+): ArrayBuffer;
```
jasnell, line 21: Is this separate `zlibWithCb` actually needed? Couldn't the `internal_zlib.ts` just call `zlibSync` directly?

src/workerd/api/node/zlib-util.c++
```diff
@@ -4,14 +4,105 @@
...
+  // size
+public:
+  inline explicit GrowableBuffer(size_t _chunkSize, size_t _maxCapacity) {
+    auto maxChunkSize = kj::min(_chunkSize, _maxCapacity);
+    builder = kj::heapArrayBuilder<kj::byte>(maxChunkSize);
+    chunkSize = maxChunkSize;
+    maxCapacity = _maxCapacity;
+  }
+
+  inline size_t size() const {
```
jasnell, line 39: nit: explicitly marking these `inline` is probably not very effective here.

src/workerd/api/node/zlib-util.h
```diff
@@ -69,6 +71,15 @@ enum class ZlibMode : ZlibModeValue {
...
 };
 
+// Zlib OS_CODE. Needed only for test fixtures
+#if defined(__APPLE__)
+static constexpr auto ZLIB_OS_CODE = 0x13;
+#elif defined(__WIN32)
+static constexpr auto ZLIB_OS_CODE = 0xa;
+#else  // non-Apple UNIX
+static constexpr auto ZLIB_OS_CODE = 0x3;
+#endif
```
jasnell, line 81: Do we really need to export this? The test could simply mask out this field and ignore it.

src/workerd/api/node/zlib-util.h
```diff
@@ -86,7 +97,7 @@ class ZlibContext final {
 public:
   ZlibContext() = default;
 
-  KJ_DISALLOW_COPY(ZlibContext);
+  //KJ_DISALLOW_COPY(ZlibContext);
```
jasnell, line 100: nit... why comment this out? The `ZlibContext` likely shouldn't be copyable.

src/workerd/api/node/zlib-util.h
```diff
@@ -131,7 +142,7 @@ class ZlibContext final {
       jsg::Optional<kj::Array<kj::byte>> _dictionary);
   kj::Maybe<CompressionError> setParams(int level, int strategy);
 
-private:
+  //private:
```
jasnell, line 145: nit... why comment this out?

src/workerd/api/node/zlib-util.h
```diff
@@ -234,11 +245,44 @@ class ZlibUtil final: public jsg::Object {
...
+    JSG_STRUCT(flush,
+        finishFlush,
+        chunkSize,
+        windowBits,
+        level,
+        memLevel,
+        strategy,
+        dictionary,
+        maxOutputLength);
+    JSG_MEMORY_INFO(Options) {}
```
jasnell, line 269:
```suggestion
```

src/node/internal/zlib.d.ts
npaun: @anonrig suggested doing it this way for performance reasons

src/workerd/api/node/zlib-util.c++
npaun: Yeah I copied and pasted most of this from libkj. I wouldn't personally write it like that

src/workerd/api/node/zlib-util.h
npaun: Oh that's better. I don't know why I didn't think of that

src/workerd/api/node/zlib-util.h
npaun: I'm not finished yet. This was some crap for testing.

src/workerd/api/node/zlib-util.h
npaun: What does that do btw? It came in when I copied and pasted some stuff

src/node/internal/zlib.d.ts
anonrig: I think you misunderstood me here. We can have a single function on C++ land, and trigger the callback if the 4th argument is callback. If not, it can just return a value and act like sync.

src/node/zlib.ts
```diff
@@ -73,4 +107,23 @@ export default {
...
+  deflateRawSync,
+  gzip,
+  gzipSync,
+  gunzip,
+  gunzipSync,
+  unzip,
+  unzipSync,
+
+  // For unit tests only
+  ZLIB_OS_CODE,
```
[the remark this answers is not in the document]
npaun: James suggested a good way (just mask that field in the tests)

src/workerd/api/node/zlib-util.h
jasnell: it has to do with collecting memory details for a heap snapshot. Most of the time you won't need it.

src/node/internal/zlib.d.ts
npaun: I'm not sure I see the advantage of having a single function on the C++ side. It would have a more confusing return type (Some array if no callback, None if there is a callback). There's very little code to reuse either.

src/workerd/api/node/zlib-util.c++
npaun: Do you want me to get rid of `inline`? I was thinking of just keeping it to match kj::Vector but it's no big deal to remove if it looks wonky to you.

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -667,3 +664,176 @@ export const testFailedInit = {
...
+    zlib.deflate(
+      'bird bird bird',
+      { windowBits: 11, level: 4 },
+      (_, result) => {
+        strictEqual(result.toString('base64'), 'OE9LyixKUUiCEQAmfgUk');
+      }
+    );
+
+    zlib.deflate('garbage data', { level: -9000 }, (error, _) => {
+      strictEqual(error.message, 'Error: Invalid compression level');
```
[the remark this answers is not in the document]
npaun: I don't think jsg can do this

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +21,149 @@ const {
...
+  );
+}
+
+export function gunzip(
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {},
+  callback: CompressCallback
+): void {
+  zlibUtil.zlibWithCb(data, options, zlibUtil.CONST_GUNZIP, (error, result) =>
+    callback(error ? Error(error) : null, result ? Buffer.from(result) : null)
```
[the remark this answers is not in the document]
npaun: What is the difference between `new Error` and `Error`? The ES standard says they do the same thing if I'm reading it right.

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -667,3 +664,176 @@ export const testFailedInit = {
 // - [x] test-zlib-failed-init.js
 // - [ ] test-zlib-invalid-input.js
 // - [ ] test-zlib-reset-before-write.js
+
+const BIG_DATA = 'horse'.repeat(50_000) + 'cow'.repeat(49_000);
```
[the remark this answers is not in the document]
npaun: This is used by several tests

npaun:
@jasnell Do we want to respect `chunkSize` for output even though these are one-shot methods? I think we could theoretically ignore it and pick something larger for performance, but that's not how Node does it.

src/node/internal/internal_zlib.ts
```diff
@@ -3,13 +3,13 @@
 //     https://opensource.org/licenses/Apache-2.0
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 
-import { default as zlibUtil, type ZlibOptions } from 'node-internal:zlib';
+import { default as zlibUtil, ZlibOptions } from 'node-internal:zlib';
```
anonrig, line 6: This needs to be reverted 

```suggestion
import { default as zlibUtil, type ZlibOptions } from 'node-internal:zlib';
```

src/workerd/api/node/zlib-util.h
```diff
@@ -234,11 +240,43 @@ class ZlibUtil final: public jsg::Object {
...
+  uint32_t crc32Sync(InputSource data, uint32_t value);
+  void zlibWithCb(
+      jsg::Lock& js, InputSource data, Options options, ZlibModeValue mode, CompressCallback cb);
+  kj::Array<kj::byte> zlibSync(InputSource data, Options options, ZlibModeValue mode);
 
   JSG_RESOURCE_TYPE(ZlibUtil) {
     JSG_METHOD_NAMED(crc32, crc32Sync);
     JSG_NESTED_TYPE(ZlibStream);
+    JSG_METHOD(zlibSync);
+    JSG_METHOD(zlibWithCb);
```
anonrig, line 279: Let's follow Node.js function naming here to avoid confusion:

sync functions have Sync suffix (zlibSync)
callback functions doesn't have anything (zlib)

```suggestion
    JSG_METHOD(zlib);
```

npaun:
@jasnell As for the `info: true` thing we talked about, I didn't implement it because `@types/node` does not support it, so I don't think it'll be useful. 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6cd7a0270f8a301a8d01f5e0634da3e429e8e60e/types/node/zlib.d.ts#L237

LMK if I need to do it anyway

src/workerd/api/node/zlib-util.h
npaun: I was a bit scared when something went wrong in edgeworker when you had a function named `crc32`. Maybe I misunderstood the problem?

anonrig: You can use jsg method named and expose different function to JS just like we do for crc32Sync

src/node/internal/internal_zlib.ts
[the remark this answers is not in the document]
anonrig: Error() calls new Error()

src/workerd/api/node/tests/zlib-nodejs-test.js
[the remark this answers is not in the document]
npaun: These were some junky tests to help me figure out how node's zlib implementation was supposed to behave. In another PR, I'll replace them with Node's equivalent tests where possible.

src/node/internal/internal_zlib.ts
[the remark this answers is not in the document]
npaun: So if I switch to new it gets rid of one level of indirection?

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -667,3 +664,176 @@ export const testFailedInit = {
...
+    );
+    strictEqual(
+      zlib
+        .deflateRawSync('as idle as a painted ship upon a painted ocean')
+        .toString('base64'),
+      'SyxWyEzJSVVILFZIVChIzMwrSU1RKM7ILFAoLcjPQxLLT05NzAMA'
+    );
+  },
+};
+export const inflateTest = {
```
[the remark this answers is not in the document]
npaun: These were some junky tests to help me figure out how node's zlib implementation was supposed to behave. In another PR, I'll replace them with Node's equivalent tests where possible.

src/node/internal/internal_zlib.ts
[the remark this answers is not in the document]
anonrig: Yes. It's mostly implemented like this

```js
function Error(...args) {
  if (!(this instanceof Error)) return new Error(...args)
  // rest of implementation
}
```

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +20,147 @@ const {
...
+  data: ArrayBufferView | string,
+  options: ZlibOptions = {},
+  callback: CompressCallback
+): void {
+  zlibUtil.zlibWithCb(
+    data,
+    options,
+    zlibUtil.CONST_INFLATERAW,
+    (error, result) =>
+      callback(error ? Error(error) : null, result ? Buffer.from(result) : null)
```
anonrig, line 116: Just an idea:

```suggestion
      const args = error ? [error] : [null, Buffer.from(result)];
      callback(...args)
```

npaun: It looks nicer but TS  is getting mad at me. I think I'll need your help tomorrow morning.

src/node/internal/internal_zlib.ts
jasnell: Not certain why this was marked resolved? From what I can see there still appears to be a need to wrap the `callback(...)` call here in `queueMicrotask(...)`

src/workerd/api/node/tests/zlib-nodejs-test.js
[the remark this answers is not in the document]
jasnell: If the error is being emitted from jsg then yeah, there won't be an `error.code` on the error, only the message.

src/workerd/api/node/zlib-util.c++
jasnell: Up to you.

src/node/internal/internal_zlib.ts
npaun: I tend to mark stuff as resolved as soon as I make the change locally, even if I haven't pushed yet. I haven't forgotten.

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +20,250 @@ const {
...
+  } else {
+    return [optionsOrCallback, callbackOrUndefined as CompressCallback];
+  }
+}
+
+function wrapCallback(
+  callback: CompressCallback
+): (error: string | null, result: ArrayBuffer | null) => void {
+  return (error, result) => {
+    queueMicrotask(() => {
```
npaun, line 101: eslint sez `  101:5  error  Unsafe call of an `any` typed value  @typescript-eslint/no-unsafe-call`

Can anyone spot what I've done wrong here?

jasnell: might just need to eslint-ignore these.

npaun: I'm surprised that other callers of queueMicrotask aren't getting slapped with that

src/workerd/api/node/zlib-util.c++
```diff
@@ -450,4 +547,67 @@ void ZlibUtil::ZlibStream::reset(jsg::Lock& js) {
   }
 }
 
+kj::Array<kj::byte> syncProcessBuffer(ZlibContext& ctx, GrowableBuffer& result) {
+  KJ_DEFER(ctx.close());
+
+  do {
+    // TODO: We respect the chunk size to match node, but it's a bit silly.. should we just
+    // override it for perf?
+    result.addChunk();
```
jasnell, line 556: Probably safe to ignore chunk size for the one-shot methods

jasnell: Well... probably should still validate that it is a correct value just to be safe, but likely ok to override it for perf.

src/workerd/api/node/zlib-util.c++
```diff
@@ -450,4 +547,67 @@ void ZlibUtil::ZlibStream::reset(jsg::Lock& js) {
...
+    KJ_CASE_ONEOF(dataBuf, kj::Array<kj::byte>) {
+      ctx.setInputBuffer(dataBuf);
+    }
+
+    KJ_CASE_ONEOF(dataStr, jsg::NonCoercible<kj::String>) {
+      ctx.setInputBuffer(dataStr.value.asBytes());
+    }
+  }
+
+  return syncProcessBuffer(ctx, result);
```
jasnell, line 601: nit: may be useful here to add a comment reminding folks that `ctx` will be closed when `syncProcessBuffer` returns.

jasnell: alternatively, would it make sense to have `~ZlibContext` call `close()` if necessary so that we don't have to worry about the `KJ_DEFER(ctx.close())` call?

npaun: I wish ZlibContext was RAII so I didn't have to do this manually but maybe it has to be that way to match NodeJS? What do you think @anonrig?

npaun: Heh James and I commented the same thing at the same time

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -667,3 +664,122 @@ export const testFailedInit = {
...
+    );
+
+    throws(
+      () => zlib.inflateSync('garbage data'),
+      Error('incorrect header check')
+    );
+
+    strictEqual(
+      zlib
+        .inflateSync(Buffer.from('OE9LyixKUUiCEQAmfgUk', 'base64'), {
```
jasnell, line 684: let's include test variations that test a zero-length input and an overly large input (zip bomb style)?

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +20,250 @@ const {
   CONST_UNZIP,
 } = zlibUtil;
 
+export function crc32(
+  data: ArrayBufferView | string,
+  value: number = 0
+): number {
+  validateUint32(value, 'value');
```
jasnell, line 27: let's throw on overly large inputs. Anything larger than 128 * 1024 should just automatically be rejected... even that is being very generous.

src/workerd/api/node/zlib-util.c++
```diff
@@ -4,14 +4,101 @@
...
+    }
+  }
 
-uint32_t ZlibUtil::crc32Sync(kj::Array<kj::byte> data, uint32_t value) {
-  // Note: Bytef is defined in zlib.h
-  return crc32(value, reinterpret_cast<const Bytef*>(data.begin()), data.size());
+  KJ_UNREACHABLE;
 }
 
+class GrowableBuffer {
```
fhanau, line 24: This can be put into an anonymous namespace, and be marked final while we're at it

src/node/internal/internal_zlib.ts
```diff
@@ -41,19 +285,6 @@ Object.defineProperties(
   )
 );
 
-export function crc32(
```
anonrig, line 44: I think we should move this function change to a different pull-request.

src/node/internal/internal_zlib.ts
```diff
@@ -6,10 +6,10 @@
 import { default as zlibUtil, type ZlibOptions } from 'node-internal:zlib';
 import { Buffer } from 'node-internal:internal_buffer';
 import { validateUint32 } from 'node-internal:validators';
-import { ERR_INVALID_ARG_TYPE } from 'node-internal:internal_errors';
-import { isArrayBufferView } from 'node-internal:internal_types';
 import { Zlib } from 'node-internal:internal_zlib_base';
 
+type CompressCallback = (error: Error | null, result: Buffer | null) => void;
```
anonrig, line 11: This type is wrong. Technically this supports (null, null), but it is not possible.

```suggestion
type CompressCallback = ((error: Error, result: null) => void) | ((null, result: Buffer) => void)
```

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +20,250 @@ const {
...
+): Buffer {
+  return Buffer.from(
+    zlibUtil.zlibSync(data, options, zlibUtil.CONST_DEFLATERAW)
+  );
+}
+
+function normalizeArgs(
+  optionsOrCallback: ZlibOptions | CompressCallback,
+  callbackOrUndefined?: CompressCallback
+): [ZlibOptions, CompressCallback] {
```
anonrig, line 89: You need to have typescript overloads if you want to support all possible cases, like:

```
function normalizeArgs(options: ZlibOptions, callback: CompressCallback);
function normalizeArgs(options: CompressCallback);
..
```

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +20,250 @@ const {
...
+  if (typeof optionsOrCallback === 'function') {
+    return [{}, optionsOrCallback];
+  } else {
+    return [optionsOrCallback, callbackOrUndefined as CompressCallback];
```
anonrig, line 93: This `as CompressCallback` hints us that type safety is not properly configured for this path. Using proper function overload types, would fix this need.

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +20,250 @@ const {
...
+  }
+}
+
+function wrapCallback(
+  callback: CompressCallback
+): (error: string | null, result: ArrayBuffer | null) => void {
```
anonrig, line 99: If you make `CompressCallback` take a generic parameter, you wouldn't need to have this type here. `CompressCallback<ArrayBuffer>` for example would result in ArrayBuffer as result.

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +20,250 @@ const {
...
+  return (error, result) => {
+    queueMicrotask(() => {
+      const args: [Error | null, Buffer | null] = error
+        ? [new Error(error), null]
+        : [null, Buffer.from(result as ArrayBuffer)];
```
anonrig, line 104: Again this shouldn't be needed:

```suggestion
        : [null, Buffer.from(result)];
```

src/node/internal/zlib.d.ts
```diff
@@ -2,6 +2,24 @@ import { owner_symbol, type Zlib } from 'node-internal:internal_zlib_base';
 
 export function crc32(data: ArrayBufferView, value: number): number;
 
+type CompressCallback = (
+  error: string | null,
+  result: ArrayBuffer | null
+) => void;
```
anonrig, line 8: This pull-request includes 3 types of CompressCallback type. We should have 1.

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -239,12 +239,9 @@ export const crc32Test = {
...
     [undefined, null, true, 1, () => {}, {}].forEach((invalid) => {
-      throws(
-        () => {
-          zlib.crc32(invalid);
-        },
-        { code: 'ERR_INVALID_ARG_TYPE' }
-      );
+      throws(() => {
+        zlib.crc32(invalid);
+      }, TypeError("Failed to execute 'crc32' on 'ZlibUtil': parameter 1 is not of type 'string or ArrayBuffer or ArrayBufferView'."));
```
anonrig, line 244:
```suggestion
      }, new TypeError("Failed to execute 'crc32' on 'ZlibUtil': parameter 1 is not of type 'string or ArrayBuffer or ArrayBufferView'."));
```

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -667,3 +664,122 @@ export const testFailedInit = {
...
+export const inflateSyncTest = {
+  test() {
+    strictEqual(
+      zlib.inflateSync(zlib.deflateSync(BIG_DATA)).toString(),
+      BIG_DATA
+    );
+
+    throws(
+      () => zlib.inflateSync('garbage data'),
+      Error('incorrect header check')
```
anonrig, line 679:
```suggestion
      new Error('incorrect header check')
```

src/workerd/api/node/tests/zlib-nodejs-test.js
anonrig: Previous code `{ code: 'ERR_INVALID_ARG_TYPE' }` was checking for `validateUint32()` function output, and it was more correct.

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -667,3 +664,122 @@ export const testFailedInit = {
...
+export const deflateSyncTest = {
+  test() {
+    function maskOsId(buf) {
+      // Clear the OS ID byte in gzip, which varies based on the platform used to run the tests
+      return buf.fill(0x0, 9, 10);
+    }
+
+    throws(
+      () => zlib.deflateSync('hello world', { windowBits: 9000 }),
+      Error('Invalid windowBits')
```
anonrig, line 703: Can you add `new` to every error class initialization?

src/workerd/api/node/zlib-util.c++
```diff
@@ -450,4 +547,67 @@ void ZlibUtil::ZlibStream::reset(jsg::Lock& js) {
...
+    KJ_IF_SOME(error, ctx.getError()) {
+      JSG_FAIL_REQUIRE(Error, error.message);
+    }
+
+    result.adjustUnused(ctx.getAvailOut());
+  } while (ctx.getAvailOut() == 0);
+
+  KJ_IF_SOME(error, ctx.getError()) {
+    JSG_FAIL_REQUIRE(Error, error.message);
+  }
```
anonrig, line 570: This seems unnecessary since there is one on line 561?

src/workerd/api/node/zlib-util.c++
```diff
@@ -450,4 +547,67 @@ void ZlibUtil::ZlibStream::reset(jsg::Lock& js) {
...
+
+  return result.releaseAsArray();
+}
+
+kj::Array<kj::byte> ZlibUtil::zlibSync(InputSource data, Options opts, ZlibModeValue mode) {
+  ZlibContext ctx;
```
anonrig, line 576: You don't seem to call the constructor?

```suggestion
  ZlibContext ctx{};
```

src/workerd/api/node/zlib-util.c++
anonrig: The context shouldn't live without the CompressionStream class. Since this change diverges the implementation of one-shot classes from Node.js, I don't think we should diverge more. It is complicated as it is right now.

src/workerd/api/node/zlib-util.c++
```diff
@@ -450,4 +547,67 @@ void ZlibUtil::ZlibStream::reset(jsg::Lock& js) {
...
+
+    KJ_CASE_ONEOF(dataStr, jsg::NonCoercible<kj::String>) {
+      ctx.setInputBuffer(dataStr.value.asBytes());
+    }
+  }
+
+  return syncProcessBuffer(ctx, result);
+}
+
+void ZlibUtil::zlibWithCb(
```
anonrig, line 604: Abbreviations only increase the cognitive load of understanding. Let's keep the code as readable as possible (for non-native speakers)

```suggestion
void ZlibUtil::zlibWithCallback(
```
```suggestion
void ZlibUtil::zlibWithCb(
```

src/workerd/api/node/zlib-util.h
```diff
@@ -4,14 +4,16 @@
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 #pragma once
 
-#include <workerd/jsg/jsg.h>
 #include <workerd/jsg/exception.h>
+#include <workerd/jsg/jsg.h>
+#include <zlib.h>
```
anonrig, line 9:
```suggestion
#include "zlib.h"
```

anonrig (changes requested):

src/workerd/api/node/zlib-util.c++
npaun: In C++, the code I've written invokes the constructor with no arguments.

src/workerd/api/node/tests/zlib-nodejs-test.js
npaun: This test is for the error relating to an invalid data argument. This comment is referring to the error for an invalid value argument, which I have not changed.

src/workerd/api/node/zlib-util.c++
npaun: I added a wrapper called ZlibContextRAII. It's a bit more verbose but I find this the more intuitive approach for me.

src/node/internal/internal_zlib.ts
npaun: Thanks, that's much better.

npaun: Hmm, I'm not sure the new type is right either. To me it says "a function that handles errors" OR "a function that handles results". You can't pass in a function that handles both with that type. FWIW, DefinitelyTyped uses a type like the one I had originally.

src/node/internal/internal_zlib.ts
npaun: Normally I'm in favour of splitting PRs into small chunks, however, this time I think the changes are OK to go together, because I'm reusing the input source stuff between zlibsync and crc32sync

npaun:
Addressed the latest round of suggestions; thanks everyone. Please have another look.

src/node/internal/internal_zlib.ts
```diff
@@ -20,6 +23,253 @@ const {
...
+
+function wrapCallback(
+  callback: CompressCallback<Error, Buffer>
+): CompressCallback<string, ArrayBuffer> {
+  return (error: string | null, result: ArrayBuffer | null) => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    queueMicrotask(() => {
+      callback(
+        error ? new Error(error) : null,
+        result ? Buffer.from(result) : Buffer.alloc(0)
```
anonrig, line 110: Shouldn't second parameter be null, if there is an error?

src/node/internal/zlib.d.ts
```diff
@@ -2,6 +2,21 @@ import { owner_symbol, type Zlib } from 'node-internal:internal_zlib_base';
 
 export function crc32(data: ArrayBufferView, value: number): number;
 
+type CompressCallback<ErrT, BufT> = (error: ErrT | null, result: BufT) => void;
```
anonrig, line 5: I don't understand typescript. How does this work without an export?

```suggestion
export type CompressCallback<ErrT, BufT> = (error: ErrT | null, result: BufT) => void;
```

src/workerd/api/node/zlib-util.c++
```diff
@@ -450,4 +556,61 @@ void ZlibUtil::ZlibStream::reset(jsg::Lock& js) {
   }
 }
 
+kj::Array<kj::byte> syncProcessBuffer(ZlibContext& ctx, GrowableBuffer& result) {
+  do {
+    // TODO: We respect the chunk size to match node, but it's a bit silly.. should we just
+    // override it for perf?
```
anonrig, line 562: It feels like after @jasnell's comment, this is no longer needed?

```suggestion
```

anonrig (approved):

src/node/internal/internal_zlib.ts
npaun: It should be undefined in Node. However, DefinitelyTyped type says it's always Buffer. Not sure what to do with this mess.

jasnell: I would make it `undefined` then.

src/node/internal/zlib.d.ts
npaun: Apparently TS allows this: https://github.com/microsoft/TypeScript/issues/38592

npaun: (but I will fix)

anonrig:
@npaun just fyi: merged commit has `WIP: ` prefix on it.

PR #2534, opened 2024-08-15
Removes Python memory snapshot upload code.

dom96:
We now upload via the validator so this code is no longer needed.

### DIFF
--- a/src/pyodide/internal/snapshot.ts
+++ b/src/pyodide/internal/snapshot.ts
@@ -26,8 +26,7 @@ let LOADED_BASELINE_SNAPSHOT: number;
 
 const TOP_LEVEL_SNAPSHOT =
   ArtifactBundler.isEwValidating() || SHOULD_SNAPSHOT_TO_DISK;
-const SHOULD_UPLOAD_SNAPSHOT =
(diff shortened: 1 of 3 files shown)


garrettgu10 (approved):

danlapid (approved):

src/pyodide/internal/snapshot.ts
```diff
@@ -26,8 +26,7 @@ let LOADED_BASELINE_SNAPSHOT: number;
 
 const TOP_LEVEL_SNAPSHOT =
   ArtifactBundler.isEwValidating() || SHOULD_SNAPSHOT_TO_DISK;
-const SHOULD_UPLOAD_SNAPSHOT =
-  ArtifactBundler.isEnabled() || TOP_LEVEL_SNAPSHOT;
+const SHOULD_UPLOAD_SNAPSHOT = TOP_LEVEL_SNAPSHOT;
```
hoodmane, line 29: Rather than having two redundant names for this, why not combine them? I think `SHOULD_UPLOAD_SNAPSHOT` is probably the better name.

hoodmane (approved):
Always nice to see PRs that remove code.

PR #2539, opened 2024-08-16
IMAGES-1258: Don't pollute global scope with generally named types

### LINKED ISSUE #2537: Type collisions with common names
The latest versions of `workers-types` is now including image options:

`export type InfoResponse`
`export type OutputOptions`
`export type Transform`

https://github.com/Cloudflare-Mining/Cloudflare-Datamining/blob/38e659727fe7679e28731e5f2b7adf9ca8121dff/data/workers-types/2021-11-03/index.ts#L4748

etc. which are now available as globals when people load `workers-types`.

There have been some in the past like `Params` from `Pages`, but I wonder if this is something that will want to be thought about at some point with better handling of prefixes for non-standard things? There are a lot of things here that are prefixing with `Ai` or `Pages`. I'm not sure what the solution here is, but would love to open a discussion around it.

ns476:
Fixes #2537

### DIFF
--- a/src/cloudflare/internal/images-api.ts
+++ b/src/cloudflare/internal/images-api.ts
@@ -15,7 +15,7 @@ type RawInfoResponse =
       height: number;
     };
 
-class TransformationResultImpl implements TransformationResult {
+class TransformationResultImpl implements ImageTransformationResult {
   public constructor(private readonly bindingsResponse: Response) {}
 
   public contentType(): string {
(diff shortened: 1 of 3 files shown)


RamIdeas (approved):

Cherry:
Would be good to see this merged soon if possible 👀

penalosa (approved):

penalosa (approved):

jasnell (approved):

PR #2542, opened 2024-08-16
improve nodejs buffer base64 and base64url performance

anonrig:
Improves the performance of base64 and base64url conversion from binary. Applies to `buffer.toString('base64')` and `buffer.toString('base64url')`

### DIFF
--- a/src/workerd/api/node/buffer.c++
+++ b/src/workerd/api/node/buffer.c++
@@ -458,10 +458,17 @@ jsg::JsString toStringImpl(
       return js.str(data);
     }
     case Encoding::BASE64: {
-      return js.str(kj::encodeBase64(slice));
+      size_t length = simdutf::base64_length_from_binary(slice.size());
+      auto out = kj::heapArray<kj::byte>(length);
+      simdutf::binary_to_base64(reinterpret_cast<const char*>(slice.begin()), slice.size(), reinterpret_cast<char*>(out.begin()));
+      return js.str(out);
     }
     case Encoding::BASE64URL: {
-      return js.str(kj::encodeBase64Url(slice));
+      auto options = simdutf::base64_url;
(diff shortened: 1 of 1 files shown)


src/workerd/api/node/buffer.c++
```diff
@@ -458,9 +458,16 @@ jsg::JsString toStringImpl(
...
+      auto out = kj::heapArray<kj::byte>(length);
+      simdutf::binary_to_base64(reinterpret_cast<const char*>(slice.begin()), slice.size(), reinterpret_cast<char*>(out.begin()));
+      return js.str(out);
     }
     case Encoding::BASE64URL: {
+      auto options = simdutf::base64_url;
+      size_t length = simdutf::base64_length_from_binary(slice.size(), options);
+      auto out = kj::heapArray<kj::byte>(length);
+      simdutf::binary_to_base64(reinterpret_cast<const char*>(slice.begin()), slice.size(), reinterpret_cast<char*>(out.begin()), options);
       return js.str(kj::encodeBase64Url(slice));
```
jasnell, line 471: Is this returning the right thing? Shouldn't it return `js.str(out)` here also?

npaun, line 471: simdutf can produce is base64 url by setting the last parameter to the enum constant `base64_url` or `base64_url_with_padding`. Is there a reason we don't want to use it?

anonrig: 🥲

npaun (approved):

jasnell (approved):

fhanau:
LGTM – I was looking at kj::encodeBase64 a few days ago not even based on performance but since it performs excessive memory allocations. It should probably be optimized still but this change should fix any performance issues for the JS APIs.

mikea (approved):

Issue #2509, opened 2024-08-09
Code formatting: Python


danlapid:
Format Python code
Definition of Done:
1. documentation
2. optional commit hooks
3. proper blame handling
4. CI checks

npaun:
We're probably going to go with `ruff`: https://github.com/astral-sh/ruff

hoodmane:
I have for my ruff config
```
lint.select = [
  "B0",     # bugbear (all B0* checks enabled by default)
  "B904",   # bugbear (Within an except clause, raise exceptions with raise ... from err)
  "B905",   # bugbear (zip() without an explicit strict= parameter set.)
  "C9",     # mccabe complexity
  "E",      # pycodestyles
  "W",      # pycodestyles
  "F",      # pyflakes
  "I",      # isort
  "PGH",    # pygrep-hooks
  "PLC",    # pylint conventions
  "PLE",    # pylint errors
  "UP",     # pyupgrade
]
lint.ignore = [
	"E402",  # module import not at top of file
	"E731",  # do not assign a lambda use a def
	"E741", # Ambiguous variable name
	"UP031", # Use format specifiers instead of percent format
	"UP038" # Use X | Y in isinstance check instead of (X, Y)
]
```

PR #2543, opened 2024-08-16
Add TS/JS/JSON formatting to the project

npaun:
This PR extends @danlapid's work in #2505 to format JS/TS/JSON code too.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Apply prettier to the project
+0523bf8b36a937348f1bb79eceda2463a5c220b5
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,34 @@
+name: Lint
+
+on:
+  pull_request:
+  push:
+    branches:
+    - main
+    
+jobs:
+  lint:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          show-progress: false
+      - name: Setup Linux
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          wget https://apt.llvm.org/llvm.sh
+          sed -i '/apt-get install/d' llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh 18
+          sudo apt-get install -y --no-install-recommends clang-format-18
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        # The pnpm version will be determined by the `packageManager` field in `.npmrc`
+      - name: Install project deps with pnpm
+        run: |
+          pnpm i
+      - name: Lint
+        run: |
+          python3 ./tools/cross/format.py --check
+        env:
+          CLANG_FORMAT: clang-format-18
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,6 @@
+{
+  "printWidth": 80,
+  "tabWidth": 2,
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
--- a/samples/nodejs-compat-diagnosticschannel/worker.js
+++ b/samples/nodejs-compat-diagnosticschannel/worker.js
@@ -32,6 +32,7 @@ export default {
   async fetch(request) {
     doSomething();
 
+    // prettier-ignore
     console.log(c.runStores(1, (...args) => {
       console.log(this, ...args, als1.getStore(), als2.getStore());
       return 1;
--- a/src/cloudflare/internal/test/images/images-api-test.js
+++ b/src/cloudflare/internal/test/images/images-api-test.js
@@ -2,7 +2,7 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 // @ts-ignore
-import * as assert from "node:assert";
+import * as assert from 'node:assert';
 
 /**
  * @typedef {{'images': ImagesBinding}} Env
@@ -15,10 +15,10 @@ export const test_images_info_bitmap = {
    * @param {Env} env
    */
   async test(_, env) {
-    const blob = new Blob(["png"]);
+    const blob = new Blob(['png']);
     const info = await env.images.info(blob.stream());
     assert.deepStrictEqual(info, {
-      format: "image/png",
+      format: 'image/png',
       fileSize: 123,
       width: 123,
       height: 123,
@@ -32,10 +32,10 @@ export const test_images_info_svg = {
    * @param {Env} env
    */
   async test(_, env) {
-    const blob = new Blob(["<svg></svg>"]);
+    const blob = new Blob(['<svg></svg>']);
     const info = await env.images.info(blob.stream());
     assert.deepStrictEqual(info, {
-      format: "image/svg+xml",
+      format: 'image/svg+xml',
     });
   },
 };
@@ -46,7 +46,7 @@ export const test_images_info_error = {
    * @param {Env} env
    */
   async test(_, env) {
-    const blob = new Blob(["BAD"]);
+    const blob = new Blob(['BAD']);
 
     /**
      * @type {any} e;
@@ -61,7 +61,7 @@ export const test_images_info_error = {
 
     assert.equal(true, !!e);
     assert.equal(e.code, 123);
-    assert.equal(e.message, "IMAGES_INFO_ERROR 123: Bad request");
+    assert.equal(e.message, 'IMAGES_INFO_ERROR 123: Bad request');
   },
 };
 
@@ -72,20 +72,20 @@ export const test_images_transform = {
    */
 
   async test(_, env) {
-    const blob = new Blob(["png"]);
+    const blob = new Blob(['png']);
 
     const result = await env.images
       .input(blob.stream())
       .transform({ rotate: 90 })
-      .output({ format: "image/avif" });
+      .output({ format: 'image/avif' });
 
     // Would be image/avif in real life, but mock always returns JSON
-    assert.equal(result.contentType(), "application/json");
+    assert.equal(result.contentType(), 'application/json');
     const body = await result.response().json();
 
     assert.deepStrictEqual(body, {
-      image: "png",
-      output_format: "image/avif",
+      image: 'png',
+      output_format: 'image/avif',
       transforms: [{ rotate: 90 }],
     });
   },
@@ -98,7 +98,7 @@ export const test_images_transform_bad = {
    */
 
   async test(_, env) {
-    const blob = new Blob(["BAD"]);
+    const blob = new Blob(['BAD']);
 
     /**
      * @type {any} e;
@@ -109,14 +109,14 @@ export const test_images_transform_bad = {
       await env.images
         .input(blob.stream())
         .transform({ rotate: 90 })
-        .output({ format: "image/avif" });
+        .output({ format: 'image/avif' });
     } catch (e2) {
       e = e2;
     }
 
     assert.equal(true, !!e);
     assert.equal(e.code, 123);
-    assert.equal(e.message, "IMAGES_TRANSFORM_ERROR 123: Bad request");
+    assert.equal(e.message, 'IMAGES_TRANSFORM_ERROR 123: Bad request');
   },
 };
 
@@ -127,7 +127,7 @@ export const test_images_transform_consumed = {
    */
 
   async test(_, env) {
-    const blob = new Blob(["png"]);
+    const blob = new Blob(['png']);
 
     /**
      * @type {any} e;
@@ -139,8 +139,8 @@ export const test_images_transform_consumed = {
         .input(blob.stream())
         .transform({ rotate: 90 });
 
-      await transformer.output({ format: "image/avif" });
-      await transformer.output({ format: "image/avif" });
+      await transformer.output({ format: 'image/avif' });
+      await transformer.output({ format: 'image/avif' });
     } catch (e2) {
       e = e2;
     }
@@ -149,7 +149,7 @@ export const test_images_transform_consumed = {
     assert.equal(e.code, 9525);
     assert.equal(
       e.message,
-      "IMAGES_TRANSFORM_ERROR 9525: ImageTransformer consumed; you may only call .output() once"
+      'IMAGES_TRANSFORM_ERROR 9525: ImageTransformer consumed; you may only call .output() once'
     );
   },
 };
--- a/src/node/_stream_duplex.js
+++ b/src/node/_stream_duplex.js
@@ -3,12 +3,16 @@
 //     https://opensource.org/licenses/Apache-2.0
 //
 /* eslint-disable */
+
+// prettier-ignore
 import {
   Duplex,
   from,
   fromWeb,
   toWeb,
 } from 'node-internal:streams_duplex';
+
+// prettier-ignore
 export {
   Duplex,
   from,
--- a/src/node/crypto.ts
+++ b/src/node/crypto.ts
@@ -4,9 +4,7 @@
 //
 /* eslint-disable */
 
-import {
-  ERR_METHOD_NOT_IMPLEMENTED
-} from 'node-internal:internal_errors';
+import { ERR_METHOD_NOT_IMPLEMENTED } from 'node-internal:internal_errors';
 
 export const getRandomValues = crypto.getRandomValues;
 export const subtle = crypto.subtle;
@@ -47,21 +45,11 @@ import {
   Hmac,
 } from 'node-internal:crypto_hash';
 
-import {
-  hkdf,
-  hkdfSync,
-} from 'node-internal:crypto_hkdf';
+import { hkdf, hkdfSync } from 'node-internal:crypto_hkdf';
 
-import {
-  pbkdf2,
-  pbkdf2Sync,
-  ArrayLike,
-} from 'node-internal:crypto_pbkdf2';
+import { pbkdf2, pbkdf2Sync, ArrayLike } from 'node-internal:crypto_pbkdf2';
 
-import {
-  scrypt,
-  scryptSync,
-} from 'node-internal:crypto_scrypt';
+import { scrypt, scryptSync } from 'node-internal:crypto_scrypt';
 
 import {
   KeyObject,
@@ -79,9 +67,7 @@ import {
 
 import { Certificate } from 'node-internal:crypto_spkac';
 
-import {
-  X509Certificate,
-} from 'node-internal:crypto_x509';
+import { X509Certificate } from 'node-internal:crypto_x509';
 
 export {
   // DH
@@ -136,9 +122,10 @@ export {
   Certificate,
   // X509
   X509Certificate,
-}
+};
 
 export function getCiphers() {
+  // prettier-ignore
   return ["aes-128-cbc", "aes-192-cbc", "aes-256-cbc", "aes-128-ctr", "aes-192-ctr", "aes-256-ctr",
   "aes-128-ecb", "aes-192-ecb", "aes-256-ecb", "aes-128-gcm", "aes-192-gcm", "aes-256-gcm",
   "aes-128-ofb", "aes-192-ofb", "aes-256-ofb", "des-ecb", "des-ede", "des-ede-cbc", "rc2-cbc"];
@@ -147,12 +134,16 @@ export function getCiphers() {
 export function getCurves() {
   // Hardcoded list of supported curves. Note that prime256v1 is equivalent to secp256r1, we follow
   // OpenSSL's and bssl's nomenclature here.
+
+  // prettier-ignore
   return ['secp224r1', 'prime256v1', 'secp384r1', 'secp521r1'];
 }
 
 export function getHashes() {
   // Hardcoded list of hashes supported in boringssl, node's approach looks pretty clunky. This is
   // expected to change infrequently based of bssl's stability-focused approach.
+
+  // prettier-ignore
   return ['md4', 'md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512', 'md5-sha1', 'RSA-MD5',
           'RSA-SHA1', 'RSA-SHA224', 'RSA-SHA256', 'RSA-SHA384', 'RSA-SHA512', 'DSA-SHA',
           'DSA-SHA1', 'ecdsa-with-SHA1'];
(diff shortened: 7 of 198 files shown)


samples/async-context/worker.js
```diff
@@ -2,25 +2,26 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
-import { AsyncLocalStorage, AsyncResource } from 'node:async_hooks';
+import { AsyncLocalStorage, AsyncResource } from "node:async_hooks";
```
jasnell, line 5: Nit: I know it is a minor thing, but I generally have a style preference for single-quotes on JS string literals rather than double-quotes.

npaun: We can switch that over and re-run prettier

samples/async-context/worker.js
```diff
@@ -2,25 +2,26 @@
...
 
 const als = new AsyncLocalStorage();
 
 export default {
   async fetch(request) {
-    const differentScope = als.run(123, () => AsyncResource.bind(() => {
-      console.log(als.getStore());
-    }));
+    const differentScope = als.run(123, () =>
+      AsyncResource.bind(() => {
```
jasnell, line 12: This change is not correct. The original uses the implied "return" on the fat-arrow function, causing the call to `als.run(..)` to return the return value of the `AsyncResource.bind(...)`. In the new version, `als.run(...)` will return undefined.

Compare:

```
function foo() { return 1 }

const e = () => foo();
const f = () => {
  foo();
}

console.log(e(), f());
```

src/cloudflare/internal/test/images/images-api-test.js
```diff
@@ -135,9 +135,7 @@ export const test_images_transform_consumed = {
     let e;
 
     try {
-      let transformer = env.images
-        .input(blob.stream())
-        .transform({ rotate: 90 });
+      let transformer = env.images.input(blob.stream()).transform({ rotate: 90 });
```
jasnell, line 138: For me, the way it was before formatting is clearer.

src/node/_stream_duplex.js
```diff
@@ -3,16 +3,6 @@
...
-  toWeb,
-} from 'node-internal:streams_duplex';
-export {
-  Duplex,
-  from,
-  fromWeb,
-  toWeb,
-};
+import { Duplex, from, fromWeb, toWeb } from "node-internal:streams_duplex";
+export { Duplex, from, fromWeb, toWeb };
```
jasnell, line 7: Here too, I much prefer the original formatting for these.

jasnell (changes requested):

samples/async-context/worker.js
anonrig: I'm also in favor of single quote

src/cloudflare/internal/test/images/images-api-test.js
npaun: Agreed. Maybe we should reduce the maximum line length a bit?

samples/async-context/worker.js
npaun: Screenshot 2024-08-16 at 10 26 42 AM
 22 AM" src="https://github.com/user-attachments/assets/c3fec609-5476-4115-b802-8b75ec1fbb2d">

I might be blind but I don't think it made any semantic changes. It just slapped in a comma on the end of the function. Does that do any harm?

tools/cross/format.py
```diff
@@ -69,16 +76,16 @@ def check_clang_format() -> bool:
...
-            if filename.endswith((".c++", ".h")):
+            if filename.endswith(exts):
                 files.append(os.path.join(root, filename))
     return files
 
 
-def clang_format(files: List[str], check: bool = False):
+def clang_format(files: List[str], check: bool = False) -> None:
     if not files:
         logging.info("No changes to format")
```
danlapid, line 90: This should just change to `return True`

tools/cross/format.py
```diff
@@ -93,6 +100,19 @@ def clang_format(files: List[str], check: bool = False):
         subprocess.run([CLANG_FORMAT, "--verbose", "-i"] + files, check=False)
 
 
+def prettier(files: List[str], check: bool = False) -> None:
+    if not files:
+        logging.info("No changes to format")
+        exit(0)
+    if check:
+        result = subprocess.run([*PRETTIER, "--check", *files])
```
danlapid, line 108: nit: `PRETTIER + ["--check", *files]`

tools/cross/format.py
```diff
@@ -93,6 +100,19 @@ def clang_format(files: List[str], check: bool = False):
...
+        if result.returncode != 0 :
+            logging.error("Code has lint. Fix with python ./tools/cross/format.py")
+            exit(1)
+    else:
+        subprocess.run([*PRETTIER, "--write", *files])
```
danlapid, line 113: nit: `PRETTIER + ["--write"] + files`

tools/cross/format.py
```diff
@@ -114,19 +134,39 @@ def git_get_modified_files(
...
+    cpp_files = set(find_files_by_exts("src/workerd", (".c++", ".h")))
+
+    if files_in_diff:
+        cpp_files &= files_in_diff
+
+    clang_format(list(cpp_files), check)
+
+
+def format_js(files_in_diff: Optional[Set[str]], check: bool) -> None:
+    js_files = set(find_files_by_exts("src", (".js", ".ts", ".json")))
```
danlapid, line 147: nit: I'd separate json to another function, we can still use prettier but does it really fit with js and ts in this list?

danlapid:
I didn't feel like doing it so who am I to throw it on you but if you feel like adding bazel take a look at https://github.com/bazelbuild/buildtools/blob/main/buildifier/README.md
basically
`buildifier --lint=fix path/to/file`
`buildifier --mode=check path/to/file`

.github/workflows/lint.yml
```diff
@@ -5,6 +5,7 @@ on:
     paths:
       - src/**
       - .clang-format
+      - .prettierc.json
```
danlapid, line 8: This list feels incomplete when looking at this PR's diff

npm/scripts/build-shim-package.mjs
danlapid, line 1: How did .mjs files get formatted if they're not in format.py's file extensions? (I think the extension should be added but it's also weird to me that these got formatted.

npaun: U caught me. I did the first run of prettier by hand on the command line.

tools/cross/format.py
```diff
@@ -114,19 +134,39 @@ def git_get_modified_files(
         return files_in_diff
 
 
+def format_cpp(files_in_diff: Optional[Set[str]], check: bool) -> None:
+    cpp_files = set(find_files_by_exts("src/workerd", (".c++", ".h")))
+
+    if files_in_diff:
```
danlapid, line 140: nit: `if files_in_diff is not None:`

npm/scripts/build-shim-package.mjs
danlapid: 😃

danlapid:
Do we want to use prettier for markdown as well?
I've never tried it before so this is a really open ended question

samples/async-context/worker.js
danlapid: I believe @npaun is right, the linter didn't end a trailing semicolon, the return value AsyncResource.bind is still the return value of the arrow function
```js
const differentScope = als.run(123, () => AsyncResource.bind(() => {
  console.log(als.getStore());
}));
```
is functionally equivalent to
```js
const differentScope = als.run(123, () =>
  AsyncResource.bind(() => {
    console.log(als.getStore());
  })
);
```

.github/workflows/lint.yml
```diff
@@ -24,8 +25,12 @@ jobs:
           chmod +x llvm.sh
           sudo ./llvm.sh 18
           sudo apt-get install -y --no-install-recommends clang-format-18
+          curl -fsSL https://deb.nodesource.com/setup_22.x | sudo bash
+          sudo apt-get install -y nodejs
+          npm i -g prettier
```
fhanau, line 30: We're not defining a version of prettier here – I assume that means we'll be using the latest stable version and CI will break without warning whenever there's a prettier update that adds new formatting rules/fixes bugs with existing ones that affect our code base. Can we define a fixed version here?
A related issue is that locally we will run whatever prettier version is available locally – if it disagrees with the prettier version on CI developers won't be able to contribute code until they upload their version. Perhaps we can use the existing bazel facilities to run prettier? I noticed that prettier is already listed in package.json, although I'm not sure how it is used so far. Perhaps it could work kind of like the `//src/cloudflare:cloudflare.capnp@eslint` and `//src/node:node.capnp@eslint` bazel targets we have for eslint?

.github/workflows/lint.yml
```diff
@@ -24,8 +25,12 @@ jobs:
           chmod +x llvm.sh
           sudo ./llvm.sh 18
           sudo apt-get install -y --no-install-recommends clang-format-18
+          curl -fsSL https://deb.nodesource.com/setup_22.x | sudo bash
```
fhanau, line 28: Non-blocking: We use Node 20 everywhere else so far and that should be installed by default on the runner image already (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime), although I'm not sure if using it involves any additional steps.

anonrig: There is a `setup-node` github workflow maintained by Github itself. We should use the official distributions.

.github/workflows/lint.yml
anonrig: I think we should just run `pnpm install` and later execute the prettier using `pnpm lint` etc. To reduce the friction of maintaining these kinds of things.

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "printWidth": 100
```
anonrig, line 4:
```suggestion
  "printWidth": 100,
  "singleQuote": true
```

samples/nodejs-compat-diagnosticschannel/worker.js
```diff
@@ -1,46 +1,55 @@
...
+        (...args) => {
+          console.log(this, ...args, als1.getStore(), als2.getStore());
+          return 1;
+        },
+        {},
+        1,
+        2,
+        3
+      )
+    );
```
anonrig, line 44: These kinds of changes needs to be ignored using `prettier-ignore`. I think the previous one was more readable.

src/node/crypto.ts
```diff
@@ -136,26 +116,60 @@ export {
...
   // X509
   X509Certificate,
-}
+};
 
 export function getCiphers() {
-  return ["aes-128-cbc", "aes-192-cbc", "aes-256-cbc", "aes-128-ctr", "aes-192-ctr", "aes-256-ctr",
-  "aes-128-ecb", "aes-192-ecb", "aes-256-ecb", "aes-128-gcm", "aes-192-gcm", "aes-256-gcm",
-  "aes-128-ofb", "aes-192-ofb", "aes-256-ofb", "des-ecb", "des-ede", "des-ede-cbc", "rc2-cbc"];
+  return [
```
anonrig, line 122: Ditto: I think this needs to be ignored as well.

anonrig (commented):
I think this is extremely hard to review and land without making sure we don't break anything. What we should do is, to enable prettier new rules on a specific folder, and divide the large chunk to smaller chunks while opening specific pull-requests according to that.

Also, I think we should use a specific prettier config that reduces the changes, and later iterate on finding the "sweet" formatting config that fits to our needs.

samples/async-context/worker.js
jasnell: aha, my eyes were seeing a `{` after the `=>` that wasn't there

src/cloudflare/internal/test/images/images-api-test.js
npaun: Line length adjustment fixed this one

.github/workflows/lint.yml
npaun: I chose the pnpm install approach in order to keep the `lint` Github action fast and simple.

src/node/crypto.ts
npaun: I personally feel that benefits of saving time and having consistent formatting outweigh choosing a good style for each block of code. I would only ignore extremely egregious formatting changes.

.github/workflows/lint.yml
npaun: In the end I decided to remove the path constraints, and allow `tools/cross/format.py` to format JS files in the entire codebase. If we're worried about that, I could go the opposite way and scope everything to `src/`

tools/cross/format.py
```diff
@@ -113,21 +126,61 @@ def git_get_modified_files(
...
+FORMATTERS = [
+    FormatConfig(
+        directory='src/workerd',
+        extensions=('.c++', '.h'),
+        formatter=clang_format
+    ),
+    FormatConfig(
+        directory='.',
+        extensions=('.js', '.ts', '.cjs', '.ejs', '.mjs'),
+        formatter= prettier
```
danlapid, line 145:
```suggestion
        formatter=prettier
```

danlapid: We need a python formatter as well 😆

src/node/crypto.ts
danlapid: I agree with @npaun but do we want to change the config so that arrays are more concise in general?
That way we don't need to ignore this.

.github/workflows/lint.yml
npaun: After all, I'd rather not do excludes like node_modules, etc. so I'm going to switch it to src/

danlapid (approved):
LGTM but I'm not the one to decide what our js lint config should be. I didn't inspect any of the js/ts file changes and so this approve does not cover those parts.

.github/workflows/lint.yml
danlapid: Just make sure this list fits whatever is being formatted :)

tools/cross/format.py
npaun: Which one do we like? I used `autopep8` at my old job but not sure if there's a cooler one we should be using instead.

danlapid: I think black is currently the best one out there but a lot of people are swearing by ruff because it's written in rust.

npaun: I found black produced some really ugly output back when I tested it. I  guess I'll go with ruff

danlapid: I think you should probably ask whoever is in charge of most of our python code, is that @hoodmane maybe?

.github/workflows/lint.yml
```diff
@@ -3,12 +3,17 @@ name: Lint
 on:
   pull_request:
     paths:
-      - src/**
-      - .clang-format
+    - .clang-format
+    - .prettierrc.json
+    - src/**
```
penalosa, line 8: This should also include `types/**`, `npm/**`, and `samples/**` at a minimum

npm/lib/node-install.ts
```diff
@@ -9,23 +9,23 @@
 import {
   downloadedBinPath,
   pkgAndSubpathForCurrentPlatform,
-} from "./node-platform";
+} from './node-platform';
```
penalosa, line 12: These files have already been formatted with prettier, mostly—it seems like this change is just to change the `'` to `"`?

.git-blame-ignore-revs
```diff
@@ -1,2 +1,5 @@
 # Apply clang-format to the project.
 16d8647946a4f2b99f0d6b8cc05a83833f85749e
+
+# Apply prettier to the project
+3ef1cecbb1411c3ce4f2e79bca8053e9bcb603c2
```
penalosa, line 5: We need to make sure this PR isn't squash merged for this to work, which might need a branch rules change. Alternatively this could be a fast-follow with the commit hash of the squashed commit?

npm/lib/node-install.ts
npaun: In this discussion, several colleagues said they preferred single quotes, so I decided to apply it throughout the project.

https://github.com/cloudflare/workerd/pull/2543#discussion_r1720089448

.git-blame-ignore-revs
npaun: It looks like GH will allow a merge commit if needed. In any case, I'll do what it takes to make sure the right commit hash is in place when this gets merged.

tools/cross/format.py
```diff
@@ -113,21 +126,61 @@ def git_get_modified_files(
...
+        ))
+    else:
+        files_in_diff = None
 
-    # TODO: lint js, ts, bazel files
+    all_ok = all(format(config, files_in_diff, options.check)
+                 for config in FORMATTERS)
 
+    if not all_ok:
+        logging.error("Code has lint. Fix with python ./tools/cross/format.py")
```
penalosa, line 182: `Code has linting issue`, maybe?

tools/cross/format.py
```diff
@@ -69,28 +78,32 @@ def check_clang_format() -> bool:
...
-            exit(1)
+            [CLANG_FORMAT, "--verbose", "--dry-run", "--Werror"] + files)
+        return result.returncode == 0
     else:
-        subprocess.run([CLANG_FORMAT, "--verbose", "-i"] + files, check=False)
+        subprocess.run([CLANG_FORMAT, "--verbose", "-i"] + files)
+        return True
+
+
+def prettier(files: List[str], check: bool = False) -> bool:
```
penalosa, line 100: I'm not sure why this function is needed, or why we're running prettier from a py file? Running `pnpm exec prettier` from the repo root _should_ do everything we need—it shouldn't be necessary to explicitly list all the files?

npm/lib/node-install.ts
penalosa: Fair enough—as long as we're consistent. Worth noting that we use `"` in [workers-sdk](https://github.com/cloudflare/workers-sdk/blob/main/.prettierrc#L3)

tools/cross/format.py
npaun: Ts script also handles clang-format, and will eventually take care of Python and Skylark (Bazel) formatting. @danlapid created this script to allow developers to reformat only files changed by their current commit. It's also set up as a git pre-commit hook.

npm/lib/node-install.ts
npaun: @jasnell, @anonrig

.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 80,
```
anonrig, line 2: I think printWidth 80 is too small. Most people use large screens and laptops >= 14 inch. Can we make this 100 or 120?

npaun: prettier's printWidth is not a maximum line length limit. See their explanation on the subject: https://prettier.io/docs/en/options.html#print-width

.github/workflows/lint.yml
```diff
@@ -24,6 +29,8 @@ jobs:
           chmod +x llvm.sh
           sudo ./llvm.sh 18
           sudo apt-get install -y --no-install-recommends clang-format-18
+          npm install -g pnpm@latest-7
```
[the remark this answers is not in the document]
npaun: Yep, I'll copy what you did in #2555

samples/nodejs-compat-streams-split2/split2.js
```diff
@@ -16,128 +16,133 @@ ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
 IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 */
 
-'use strict'
+'use strict';
```
jasnell, line 19: This file should likely be left as is and ignored. It was copied from the module and isn't our code.

src/node/_stream_duplex.js
jasnell: For now does prettier have an option to ignore lines/blocks using comments? If so, let's use that with these

samples/nodejs-compat-streams-split2/split2.js
npaun: Added to .prettierignore

.github/workflows/lint.yml
npaun: I decided to just ask git to list all tracked files + all untracked files that are not git ignored. This list should stay up to date.

tools/cross/format.py
npaun: Discussion can be continued in #2509

.github/workflows/lint.yml
```diff
@@ -3,12 +3,17 @@ name: Lint
...
-      - src/**
-      - .clang-format
+    - .clang-format
+    - .prettierrc.json
+    - src/**
   push:
     branches:
     - main
-
+    paths:
```
danlapid, line 12: I think we want to format all on any push to main, regardless of paths

.github/workflows/lint.yml
danlapid: Not sure what you mean, on my screen it's still not showing those directories

src/node/_stream_duplex.js
npaun: `// prettier-ignore`

Added

samples/nodejs-compat-diagnosticschannel/worker.js
npaun: ignored

.github/workflows/lint.yml
npaun: You're right, thanks.

.github/workflows/lint.yml
npaun: I changed only format.py and forgot GH actions. Will fix.

npm/lib/node-install.ts
garrettgu10: Obvs not a big deal but I prefer double quotes since JSON requires them, and being consistent with workers-sdk would be a nice benefit

anonrig (approved):

garrettgu10 (approved):
I only looked at the `pyodide` folder but it looked nice, thanks!

npm/lib/node-install.ts
npaun: Quote fight!

npaun: I'm on team single quotes, but it's not a big deal for me either.

anonrig: We can use different configuration on different folders. We just need to add `.prettierrc` to the folder where we want to have double quotes.

npaun: I would prefer to have a unified style in all of workerd, but I'm not opposed to per-folder if there is a strong reason

jasnell (approved):

PR #2544, opened 2024-08-16
Do not allow code with warnings to pass CI

npaun:
This PR enables `-Werror`, when building in the CI. As a result, warnings are treated as a fatal error. This ensures that warnings will be addressed, and won't pile up over time in our codebase. Here are some specific details about the configuration.

* Warnings aren't fatal locally. This is for convenience while working on code.
* Warnings aren't fatal in `external/`. This is because these warnings come from our dependencies, and need to be fixed there.
* Deprecation warnings and `#warning` are not fatal. They're just here to give us a heads up that something will need to be fixed eventually.
* Warnings are only fatal on Linux builds. I didn't want to make this a maddening quest of fixing macOS, Windows AND Linux compiler warnings every time we do a PR.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -43,6 +43,7 @@ build --@dawn//src/tint:tint_build_wgsl_writer=True
 # Our dependencies (ICU, zlib, etc.) produce a lot of these warnings, so we disable them. Depending
 # on the clang version, zlib either produces warnings for -Wdeprecated-non-prototype or does not
 # have that option, so disable -Wunknown-warning-option there too.
+build --per_file_copt='external@-Wno-error'
 build --per_file_copt='external/v8@-Wno-deprecated-declarations'
 build --per_file_copt='external/com_googlesource_chromium_icu@-Wno-ambiguous-reversed-operator,-Wno-deprecated-declarations'
 build --host_per_file_copt='external/com_googlesource_chromium_icu@-Wno-ambiguous-reversed-operator,-Wno-deprecated-declarations'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,6 +84,9 @@ jobs:
           sudo apt-get install -y --no-install-recommends clang-16 lld-16 libunwind-16 libc++abi1-16 libc++1-16 libc++-16-dev libclang-rt-16-dev
           echo "build:linux --action_env=CC=/usr/lib/llvm-16/bin/clang --action_env=CXX=/usr/lib/llvm-16/bin/clang++" >> .bazelrc
           echo "build:linux --host_action_env=CC=/usr/lib/llvm-16/bin/clang --host_action_env=CXX=/usr/lib/llvm-16/bin/clang++" >> .bazelrc
+          echo "build:linux --copt='-Werror'" >> .bazelrc
+          echo "build:linux --copt='-Wno-error=#warnings'" >> .bazelrc
+          echo "build:linux --copt='-Wno-error=deprecated-declarations'" >> .bazelrc
           sed -i -e "s%llvm-symbolizer%/usr/lib/llvm-16/bin/llvm-symbolizer%" .bazelrc
       - name: Setup macOS
         if: matrix.os.name == 'macOS'
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -1602,7 +1602,8 @@ Worker::Worker(kj::Own<const Script> scriptParam,
                 }
                 KJ_CASE_ONEOF(limitError, kj::Exception) { }
               }
-            }
+            } else {} // Added to suppress a compiler warning
+
             startupMetrics->done();
           } catch (const kj::Exception& e) {
             lock.throwException(kj::cp(e));


.github/workflows/test.yml
```diff
@@ -84,6 +84,9 @@ jobs:
           sudo apt-get install -y --no-install-recommends clang-16 lld-16 libunwind-16 libc++abi1-16 libc++1-16 libc++-16-dev libclang-rt-16-dev
           echo "build:linux --action_env=CC=/usr/lib/llvm-16/bin/clang --action_env=CXX=/usr/lib/llvm-16/bin/clang++" >> .bazelrc
           echo "build:linux --host_action_env=CC=/usr/lib/llvm-16/bin/clang --host_action_env=CXX=/usr/lib/llvm-16/bin/clang++" >> .bazelrc
+          echo "build:linux --copt='-Werror'" >> .bazelrc
+          echo "build:linux --copt='-Wno-error=#warnings'" >> .bazelrc
+          echo "build:linux --copt='-Wno-error=deprecated-declarations'" >> .bazelrc
```
danlapid, line 89: Do we want need this if we have .bazelrc:47 `build --per_file_copt='external/v8@-Wno-deprecated-declarations'`
It's possible both are required, just not sure.

npaun: This would handle a project other than v8 having deprecated declarations.

danlapid: Yeah that's why I asked if we want this in any project other than v8.
I guess it's a more fitting question for @jasnell  maybe?

src/workerd/io/worker.c++
```diff
@@ -1602,7 +1602,8 @@ Worker::Worker(kj::Own<const Script> scriptParam,
                 }
                 KJ_CASE_ONEOF(limitError, kj::Exception) { }
               }
-            }
+            } else {}
```
anonrig, line 1605: These change seems unrelated. Can you remove it?

anonrig (commented):
Can you also update `compile_flags.txt`?

danlapid:
> Can you also update `compile_flags.txt`?

I think that would break "Warnings aren't fatal locally. This is for convenience while working on code."

src/workerd/io/worker.c++
npaun: This fixes the one fatal warning we have currently

.github/workflows/test.yml
fhanau: I think we regularly have deprecation warnings for JSG calling deprecated V8 functions (including right now?), it will be difficult to always have those under control.

.bazelrc
```diff
@@ -43,6 +43,7 @@ build --@dawn//src/tint:tint_build_wgsl_writer=True
 # Our dependencies (ICU, zlib, etc.) produce a lot of these warnings, so we disable them. Depending
 # on the clang version, zlib either produces warnings for -Wdeprecated-non-prototype or does not
 # have that option, so disable -Wunknown-warning-option there too.
+build --per_file_copt='external/@-Wno-error'
```
fhanau, line 46: nit: `external@-Wno-error` looks much cleaner to me

.github/workflows/test.yml
danlapid: But these would fall under .bazelrc:47 no?

fhanau:
> - Warnings aren't fatal locally. This is for convenience while working on code.
> - Warnings are only fatal on Linux builds. I didn't want to make this a maddening quest of fixing macOS, Windows AND Linux compiler warnings every time we do a PR.
> 

I think these two aspects are key, thank you for covering them. 1) is important for development velocity, 2) means we won't have to fix the many unhelpful Windows warnings we have. Including macOS via `build:unix` could be a good option as both systems share the vast majority of warnings when they happen but it conceivably could make things harder for non-mac developers and is not needed at this stage.

src/workerd/io/worker.c++
fhanau: What is the warning? The statement looks really superfluous to me, if there's no other way to handle this let's add a comment here

.github/workflows/test.yml
fhanau: No, that directive only adds the `-Wno-deprecated-declarations` flag when compiling source files whose paths 
match the `external/v8` pattern, i.e. V8 source files only.

src/workerd/io/worker.c++
npaun: -Wdangling-else on KJ_IF_SOME. I see that James fixed a bunch of these in the past.  The macro tries to block that warning but it's still coming through. Maybe we should look into that.


Thinking out loud a bit: 

The macro's basic structure is 

```c++
if (has value)
   if (init; false) { 
      // no op
   } else 
```

So a usage expands to something like

```c++
if (has value) 
    if (init; false) { 
       // no op
    } else {
       // here is my actual block when there is a value 
    }
```

Clang is not happy about that outer if which has no braces.

npaun: I can't trigger it on its own so I'm thinking either a clang bug, or something glitching with the pragmas

npaun: I spent way too much time on this - the problem is triggered if KJ_IF_SOME is passed as an argument to any other macro.

e.g.
```c++
#define USELESS(x) x
USELESS(
    KJ_IF_SOME(foo, bar) {
       // blah
    }
)
```

npaun: For now I added a comment to explain why that else is there.

fhanau (approved):
✅

danlapid (approved):

danlapid (approved):

PR #2552, opened 2024-08-19
Add Pyodide development compat flag

hoodmane:
Garrett already added the logic to consume this flag in workerd. We will want to test with the flag both on and off. Downstream we also need some additional code to consume the flag.


dom96 (approved):

hoodmane:
I seem to have accidentally made two PRs that did almost the same thing. This was superseded by #2560.

PR #2553, opened 2024-08-19
Adding PipelineTransform class

hhoughgg:
This PR adds a class called PipelineTransform which allows us to test PipelineTransform implementations before sending them real data from "Cloudflare Pipelines".

It also means that you don't have to deal with streams directly and can just work with JSON/Javascript Objects.

### DIFF
--- a/src/cloudflare/internal/pipeline-transform.ts
+++ b/src/cloudflare/internal/pipeline-transform.ts
@@ -37,7 +37,7 @@ async function* readLines(
 type Batch = {
   id: string; // unique identifier for the batch
   shard: string; // assigned shard
-  ts: number; // timestamp of the event
+  ts: number; // creation timestamp of the batch
 
   format: Format;
   size: {
@@ -118,6 +118,10 @@ export class PipelineTransformImpl extends entrypoints.WorkerEntrypoint {
   }
 
   #sendJson(data: object[]): JsonStream {
+    if (!(data instanceof Array)) {
+      throw new Error('transformJson must return an array of objects');
+    }
+
     let written = 0;
     const encoder = new TextEncoder();
     const readable = new ReadableStream<Uint8Array>({
--- a/src/cloudflare/internal/pipeline-transform.ts
+++ b/src/cloudflare/internal/pipeline-transform.ts
@@ -1,0 +1,147 @@
+// Copyright (c) 2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import entrypoints from 'cloudflare-internal:workers';
+
+async function* readLines(
+  stream: ReadableStream<string>
+): AsyncGenerator<string> {
+  let start = 0;
+  let end = 0;
+  let partial = '';
+
+  // @ts-expect-error must have a '[Symbol.asyncIterator]()' method
+  for await (const chunk of stream) {
+    const full = partial + chunk;
+    for (const char of full) {
+      if (char === '\n') {
+        yield full.substring(start, end);
+        end++;
+        start = end;
+      } else {
+        end++;
+      }
+    }
+
+    partial = full.substring(start, end);
+    start = 0;
+    end = 0;
+  }
+
+  if (partial.length > 0) {
+    yield partial;
+  }
+}
+
+type Batch = {
+  id: string; // unique identifier for the batch
+  shard: string; // assigned shard
+  ts: number; // timestamp of the event
+
+  format: Format;
+  size: {
+    bytes: number;
+    rows: number;
+  };
+  data: unknown;
+};
+
+type JsonStream = Batch & {
+  format: Format.JSON_STREAM;
+  data: ReadableStream<Uint8Array>;
+};
+
+enum Format {
+  JSON_STREAM = 'json_stream', // jsonl
+}
+
+export class PipelineTransformImpl extends entrypoints.WorkerEntrypoint {
+  #batch?: Batch;
+  #initalized: boolean = false;
+
+  // stub overriden on the sub class
+  // eslint-disable-next-line @typescript-eslint/require-await
+  public async transformJson(_data: object[]): Promise<object[]> {
+    throw new Error('should be implemented by parent');
+  }
+
+  // called by the dispatcher which then calls the subclass methods
+  // @ts-expect-error thinks ping is never used
+  private async _ping(): Promise<void> {
+    // making sure the function was overriden by an implementing subclass
+    if (this.transformJson !== PipelineTransformImpl.prototype.transformJson) {
+      return await Promise.resolve(); // eslint
+    } else {
+      throw new Error(
+        'the transformJson method must be overridden by the PipelineTransform subclass'
+      );
+    }
+  }
+
+  // called by the dispatcher which then calls the subclass methods
+  // the reason this is typescript private and not javascript private is that this must be
+  // able to be called by the dispatcher but should not be called by the class implementer
+  // @ts-expect-error _transform is called by rpc
+  private async _transform(batch: Batch): Promise<JsonStream> {
+    if (this.#initalized) {
+      throw new Error('pipeline entrypoint has already been initialized');
+    }
+
+    this.#batch = batch;
+    this.#initalized = true;
+
+    switch (this.#batch.format) {
+      case Format.JSON_STREAM:
+        const data = await this.#readJsonStream();
+        const transformed = await this.transformJson(data);
+        return this.#sendJson(transformed);
+      default:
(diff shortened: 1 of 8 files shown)


hhoughgg:

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,131 @@
...
+
+import entrypoints from 'cloudflare-internal:workers'
+
+async function* readLines(stream: ReadableStream<string>): AsyncGenerator<string> {
+  let partial = ''
+
+  // @ts-ignore
+  for await (const chunk of stream) {
+    const full = partial + chunk
+    const lines = full.split('\n')
```
oliy, line 13: Probably a later optimization, but could probably implement a streaming split to avoid extra memory allocations.  Might not be worth the effort, if the chunks are relatively small.

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  // stub overriden on the sub class
+  // @ts-ignore
+  public async transformJson(data: object[]): Promise<object[]> {
+    throw new Error('should be implemented by parent')
+  }
+
+  // called by the dispatcher which then calls the subclass methods
+  // @ts-ignore
+  private async _transform(batch: Batch, ping?: boolean): Promise<JsonStream> {
```
oliy, line 63: If null batch is the signal for a "ping", no explicit ping parameter is necessary, which clutters up the interface.

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (this.#initalized) {
+      throw new Error('pipeline entrypoint has already been initialized')
```
oliy, line 65: Would we ever need to reset/restart a stream for reliability purposes?  Seems like this should handle either case of multiple streams or single infinite stream.

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,132 @@
...
+          const data = await this.readJsonStream()
+          const transformed = await this.transformJson(data)
+          return this.sendJson(transformed)
+        default:
+          throw new Error('unsupported batch format')
+      }
+    }
+  }
+
+  private async readJsonStream(): Promise<object[]> {
```
anonrig, line 96: Why don't we use private properties supported by javascript language?
```suggestion
  async #readJsonStream(): Promise<object[]> {
```

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const data: object[] = []
+    for await (const line of readLines(decoder)) {
+      data.push(JSON.parse(line))
+    }
+
+    return data
+  }
+
+  private async sendJson(data: object[]): Promise<JsonStream> {
```
anonrig, line 112:
```suggestion
  async #sendJson(data: unknown[]): Promise<JsonStream> {
```

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,132 @@
...
+      }
+    })
+
+    return {
+      id: this.#batch!.id,
+      shard: this.#batch!.shard,
+      ts: this.#batch!.ts,
+      format: Format.JSON_STREAM,
+      data: readable.pipeThrough(new TextEncoderStream())
+    } as JsonStream
```
anonrig, line 130: Why is this `as JsonStream` required?

src/cloudflare/internal/test/pipeline-transform/transform-test.js
```diff
@@ -0,0 +1,112 @@
+// @ts-nocheck
+// Copyright (c) 2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+// @ts-ignore
+import * as assert from 'node:assert'
```
anonrig, line 7: This should also work since `assert` has a default export

```suggestion
import assert from 'node:assert'
```

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,132 @@
...
+  private async _transform(batch: Batch): Promise<JsonStream> {
+    if (this.#initalized) {
+      throw new Error('pipeline entrypoint has already been initialized')
+    }
+
+    // test call to make sure the pipeline entrypoint is setup correctly
+    if (batch === null) {
+      // making sure the function was overriden by an implementing subclass
+      if (this.transformJson !== PipelineTransformImpl.prototype.transformJson) {
+        return null as unknown as JsonStream
```
anonrig, line 77: I think we should add `Promise<JsonStream | null>` into the return value, since this can clearly cause a bug on runtime. Or, we should just throw in here.

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // stub overriden on the sub class
+  // @ts-ignore
+  public async transformJson(data: object[]): Promise<object[]> {
+    throw new Error('should be implemented by parent')
+  }
+
+  // called by the dispatcher which then calls the subclass methods
+  // @ts-ignore
+  private async _transform(batch: Batch): Promise<JsonStream> {
```
anonrig, line 68: Since we check for `batch == null`, we should update the type of the variable.

```suggestion
  private async _transform(batch?: Batch): Promise<JsonStream> {
```

src/cloudflare/internal/pipeline-transform.ts
hhoughgg: https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes
I think I would have to declare it as a new variable without shard and then set it after

src/cloudflare/internal/pipeline-transform.ts
hhoughgg: Went ahead and made this a separate method instead

a-robinson (commented):
Mostly looks good! Just a few small things.

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,147 @@
...
+  // called by the dispatcher which then calls the subclass methods
+  // the reason this is typescript private and not javascript private is that this must be
+  // able to be called by the dispatcher but should not be called by the class implementer
+  // @ts-expect-error _transform is called by rpc
+  private async _transform(batch: Batch): Promise<JsonStream> {
+    if (this.#initalized) {
+      throw new Error('pipeline entrypoint has already been initialized');
+    }
+
+    this.#batch = batch;
```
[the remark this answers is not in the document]
hhoughgg: My assumption is that each dynamic dispatch call would create a new worker. I think in the future we may add the ability to have the customer directly process the stream which would allow us to send multiple batches to a single worker.

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    return data;
+  }
+
+  #sendJson(data: object[]): JsonStream {
+    let written = 0;
+    const encoder = new TextEncoder();
+    const readable = new ReadableStream<Uint8Array>({
+      start(controller): void {
+        for (const obj of data) {
```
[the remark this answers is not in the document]
hhoughgg: Yeah - should at least give a nice error message. There is additional validation on the pipeline side.

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (!(data instanceof Array)) {
+      throw new Error('transformJson must return an array of objects');
+    }
+
+    let written = 0;
+    const encoder = new TextEncoder();
+    const readable = new ReadableStream<Uint8Array>({
+      start(controller): void {
+        for (const obj of data) {
+          const encoded = encoder.encode(`${JSON.stringify(obj)}\n`);
```
a-robinson, line 130: In terms of validating the response from the customer's code, it's [also possible for them to return values that will throw when passed to `JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions), and we'd *probably* benefit from handling them more gracefully than just an uncaught exception?

Although then again in the `instanceof Array` check above all we do is throw an exception, so actually maybe this is already fine and we just need the calling code to detect this case correctly?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions

PR #2554, opened 2024-08-19
fix node eslintrc.json project path

anonrig:
This was locally causing issues for me. Using a relative path in `project` causes clion errors to go away.


npaun (approved):
Thanks! I've been having the same issue on my machine.

PR #2555, opened 2024-08-19
use a specific pnpm version

anonrig:
Currently, we don't have any way of understanding which version of `pnpm` we are using. 

This pull-request:

- Adds `packageManager` field that is used both by `corepack` and `pnpm` which automatically downloads the correct pnpm version.
- Updates `pnpm` to latest version
- Updates 2 bazel plugins in order to update the pnpm version. Unfortunately, the existing versions doesn't support pnpm 9.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -174,9 +174,11 @@ jobs:
           repository: cloudflare/workers-sdk
 
       - name: Install pnpm
-        uses: pnpm/action-setup@v3
+        uses: pnpm/action-setup@v4
         with:
-          version: ^9.1.3
+          # This version will be different depending on the value of `packageManager` field
+          # in package.json file with the setting in `.npmrc` file.
+          version: latest
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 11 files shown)


npaun (approved):
Just this morning I thought to myself "huh this project isn't using corepack for some reason". And now it's solved.

anonrig:
> Just this morning I thought to myself "huh this project isn't using corepack for some reason". And now it's solved.

My goal wasn't to add `corepack` since it is experimental and might be removed. Thankfully, pnpm has recently added a support for auto-installing the correct version depending on the `packageManager` value.

npaun:
So the preferred workflow is to install any old version of pnpm, then get pnpm to install the right one for us?

anonrig:
> So the preferred workflow is to install any old version of pnpm, then get pnpm to install the right one for us?

Preferably, pnpm install `pnpm@latest` which will handle the correct version for us, so there won't be any unnecessary diff in `pnpm-lock.yml`

fhanau (approved):

jasnell (approved):

anonrig:
the syntax for `.npmrc` was wrong. i force pushed. @fhanau can you update the internal pull-request if you don't mind?

fhanau:
Internal CI has passed. Note that the internal build will not be accessing the workerd .npmrc file, it's just the TS and tsconfig changes that are relevant to the internal build, so the last CI run is sufficient. This is ready to merge.

PR #2557, opened 2024-08-20
Mark io target static variables as const to avoid race conditions

fhanau:
Also cleans up remaining instances of uninitialized variables in that target.

- In some cases using constexpr might be possible with little additional effort (in particular by making StringPtr constructors constexpr), but that was not the main objective here.


src/workerd/api/crypto/keys.c++
```diff
@@ -90,7 +90,7 @@ kj::Array<kj::byte> AsymmetricKeyCryptoKeyImpl::exportKeyExt(
   auto bio = OSSL_BIO_MEM();
 
   struct EncDetail {
-    char* pass = &EMPTY_PASSPHRASE[0];
+    char* pass = const_cast<char*>(&EMPTY_PASSPHRASE[0]);
```
fhanau, line 93: I haven't used const_cast before – let me know if this usage pattern is incorrect.

src/workerd/util/sqlite.h
```diff
@@ -51,13 +50,57 @@ class SqliteDatabase {
   // expected.
   operator sqlite3*() { return db; }
 
+// Class which regulates a SQL query, especially to control how queries created in JavaScript
+// application code are handled.
+class Regulator {
```
fhanau, line 55: The compiler insisted on Regulator being defined before the static constexpr Regulator variable, so I moved the subclass inside.

anonrig (approved):

jasnell (approved):

src/workerd/api/crypto/keys.c++
mikea: I think this case is ok since we don't expect anything to be written.

src/workerd/io/io-gate.h
```diff
@@ -51,10 +51,10 @@ class InputGate {
     virtual void inputGateWaiterAdded() {}
     virtual void inputGateWaiterRemoved() {}
 
-    static Hooks DEFAULT;
+    static const Hooks DEFAULT;
   };
 
-  InputGate(Hooks& hooks = Hooks::DEFAULT);
+  InputGate(Hooks& hooks = const_cast<Hooks&>(Hooks::DEFAULT));
```
kentonv, line 57: Generally any `const_cast` should come with a comment saying why it's OK. In this case: "const_cast OK because the object has no members, it's just a vtable."

PR #2558, opened 2024-08-20
EW-8447 Add regression test for CPU profiling

### LINKED ISSUE #1754: 🐛 Bug Report — CPU profiling reports entire execution as `(program)`

harrishancock:
This PR adds a regression test for the CPU profiling bug reported in #1754 and fixed in #2497. It is based on @mrbbot's original reproduction.

I backported the test to the commit just prior to #2497, and confirmed that the test caught the original breakage.

I wrote the test in JavaScript, because it seemed to have the richest ecosystem of tools for working with the Chrome Devtool Protocol. I originally intended to use Playwright to initiate a CDP connection to workerd, but it seemed to make too many assumptions that it was connecting to a browser. I tried the [chrome-remote-interface](https://github.com/cyrus-and/chrome-remote-interface) library next, and it seemed to work well.
    
Fixes #1754.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -42,6 +42,9 @@ importers:
       capnpc-ts:
         specifier: ^0.7.0
         version: 0.7.0
+      chrome-remote-interface:
+        specifier: ^0.33.2
+        version: 0.33.2
       esbuild:
         specifier: ^0.15.18
         version: 0.15.18
@@ -304,12 +307,19 @@ packages:
     resolution: {integrity: sha512-oKnbhFyRIXpUuez8iBMmyEa4nbj4IOQyuhc/wy9kY7/WVPcwIO9VA668Pu8RkO7+0G76SLROeyw9CpQ061i4mA==}
     engines: {node: '>=10'}
 
+  chrome-remote-interface@0.33.2:
+    resolution: {integrity: sha512-wvm9cOeBTrb218EC+6DteGt92iXr2iY0+XJP30f15JVDhqvWvJEVACh9GvUm8b9Yd8bxQivaLSb8k7mgrbyomQ==}
+    hasBin: true
+
   color-convert@2.0.1:
     resolution: {integrity: sha512-RRECPsj7iu/xb5oKYcsFHSppFNnsj/52OVTRKb4zP5onXwVF3zVmmToNcOfGC+CRDpfK/U584fMg38ZHCaElKQ==}
     engines: {node: '>=7.0.0'}
 
   color-name@1.1.4:
     resolution: {integrity: sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA==}
+
+  commander@2.11.0:
+    resolution: {integrity: sha512-b0553uYA5YAEGgyYIGYROzKQ7X5RAqedkfjiZxwi0kL1g3bOaBNNZfYkzt/CL0umgD5wc9Jec2FbB98CjkMRvQ==}
 
   concat-map@0.0.1:
     resolution: {integrity: sha512-/Srv4dswyQNBfohGpz9o6Yb3Gz3SrUDqBH5rTuhGR7ahtlbYKnVxw2bCFMRljaA7EXHaXZ8wsHdodFvbkhKmqg==}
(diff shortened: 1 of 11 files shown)


pnpm-lock.yaml
danlapid, line 1: Can you try
```sh
npm install -g pnpm@latest-7
pnpm install
```
I believe it should revert most changes to this file :)

src/workerd/server/tests/server-harness.mjs
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  // Return a promise for the inspector port.
+  async getListenInspectorPort() {
+    return await this.#listenInspectorPort;
+  }
+
+  // Return a promise for the named listen port.
+  async getListenPort(name) {
+    assert.notEqual(this.#listenPorts.get(name), undefined);
```
danlapid, line 108:
```suggestion
    assert.notEqual(this.#listenPorts, null);
    assert.notEqual(this.#listenPorts.get(name), undefined);
```

pnpm-lock.yaml
harrishancock: Thanks, that helped.

danlapid (approved):
Love it! very concise and clear

pnpm-lock.yaml
harrishancock: Amusingly I had to redo this because #2555 just updated us to pnpm 9.

PR #2559, opened 2024-08-20
Add StepConfig workflow types

LuisDuarte1:
`StepConfig` was missing from previous type iterations, making the `cloudflare:workers` `Workflow` types kind of unusable :sweat_smile:. 

All utility types are prefixed with Workflow to avoid polluting the `cloudflare:workers` scope

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -195,7 +195,7 @@ declare module "cloudflare:workers" {
     webSocketError?(ws: WebSocket, error: unknown): void | Promise<void>;
   }
 
-  export type DurationLabel =
+  export type WorkflowDurationLabel =
     | "second"
     | "minute"
     | "hour"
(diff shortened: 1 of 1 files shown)


LuisDuarte1:

jasnell (approved):

types/defines/rpc.d.ts
```diff
@@ -195,22 +195,35 @@ declare module "cloudflare:workers" {
...
+
+  export type WorkflowStepConfig = {
+    retries?: {
+      limit: number;
+      delay: string | number;
+      backoff?: WorkflowBackoff;
+    };
+    timeout?: string | number;
+  };
+
```
penalosa, line 219: I realise there are already Workflow types in this file, but I think they should probably be in a separate `workflows.d.ts` so we can more easily separate them.

types/defines/rpc.d.ts
```diff
@@ -195,22 +195,35 @@ declare module "cloudflare:workers" {
     webSocketError?(ws: WebSocket, error: unknown): void | Promise<void>;
   }
 
-  export type DurationLabel =
+  export type WorkflowDurationLabel =
```
penalosa, line 198: This is a breaking change—I assume this is being considered okay because Workflows is experimental?

LuisDuarte1: Yhep this is okay as we are still iterating on the types, we just thought that would be better to prefix everything with Workflow to not pollute the scope

PR #2560, opened 2024-08-20
Add Python development compatibility flag for tests

hoodmane:
This also moves toward deriving the bundle name from the compat date.


src/workerd/io/compatibility-date.capnp
```diff
@@ -424,12 +424,20 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   pythonWorkers @43 :Bool
       $compatEnableFlag("python_workers")
       $pythonSnapshotRelease(pyodide = "0.26.0a2", pyodideRevision = "2024-03-01",
-          packages = "2024-03-01", backport = 0);
+          packages = "2024-03-01", backport = 0)
+      $impliedByAfterDate(name = "pythonWorkersDevPyodide", date = "2000-01-01");
```
garrettgu10, line 428: Not sure if I'm understanding this bit. If I set my compatibility date to 2024-02-29, does this mean the `python_workers_development` flag will be set automatically?

src/workerd/server/tests/python/import_tests.bzl
```diff
@@ -23,7 +23,7 @@ const unitTests :Workerd.Config = (
           (name = "{}", pythonRequirement = ""),
         ],
         compatibilityDate = "2024-05-02",
-        compatibilityFlags = ["python_workers"],
+        compatibilityFlags = ["python_workers_development"],
```
garrettgu10, line 26: I wonder if we can do a find-and-replace on this string to generate the two variants of tests we talked about. I'll fiddle around with the bazel and make a PR.

garrettgu10 (approved):
🚀🚀🚀🚀🚀

src/workerd/server/tests/python/env-param/env.wd-test
```diff
@@ -14,7 +14,7 @@ const unitTests :Workerd.Config = (
           ),
         ],
         compatibilityDate = "2024-01-15",
-        compatibilityFlags = ["python_workers"],
```
danlapid, line 17: Would it make more sense to duplicate all of those tests to run against the old flag as well?

src/workerd/io/compatibility-date.capnp
```diff
@@ -424,12 +424,20 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
   # Enables Python Workers. Access to this flag is not restricted, instead bundles containing
   # Python modules are restricted in EWC.
   #
   # WARNING: Python Workers are still an experimental feature and thus subject to change.
 
+  pythonWorkersDevPyodide @58 :Bool
```
danlapid, line 434: I know it fits nicely under pythonWorkers but I still think it's better to have it be at the end of the file so that it's sorted by the flag id, next person to add a flag will be very confused as to why his receiving an error of 58 already in use.

src/workerd/io/compatibility-date.c++
```diff
@@ -356,4 +356,11 @@ kj::Maybe<PythonSnapshotRelease::Reader> getPythonSnapshotRelease(
   return result;
 }
 
+kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
```
danlapid, line 359: Not sure if this code fits logically within compatibility-date.c++ does it? I guess we have a lot of other pyodideRelease code here already but still

src/workerd/io/compatibility-date.capnp
hoodmane: I believe that it says if `pythonWorkersDevPyodide` is on and the compat date is after `2000-01-01` (aka always) then `pythonWorkers` should be on too.

src/workerd/server/tests/python/env-param/env.wd-test
hoodmane: Yeah it would. I think @garrettgu10 has offered to do this in a followup.

danlapid (approved):

Issue #2551, opened 2024-08-19
Migration to Eslint 9


anonrig:
We are using 8.57.0 as of August 19. The latest version is v9. We eventually need to update to latest version of Eslint to update our dependencies as well.

- [ ] Update eslint to v9
- [ ] Update eslint-config-prettier
- [ ] Update eslint-plugin-import
- [ ] Update eslint-plugin-prettier
- [ ] Update typescript-eslint*

We also don't properly configure typescript-eslint in `src/node/.eslintrc.json` and fully leverage the typescript-eslint package type checking capabilities.

PR #2562, opened 2024-08-20
update eslint to v9

anonrig:
Updates all eslint dependencies to latest versions, and update inconsistencies with prettier, typescript-eslint and eslint recommended configurations.

Fixes https://github.com/cloudflare/workerd/issues/2551


build/wd_ts_bundle.bzl
```diff
@@ -1,7 +1,7 @@
 load("@aspect_rules_ts//ts:defs.bzl", "ts_config", "ts_project")
+load("@capnp-cpp//src/capnp:cc_capnp_library.bzl", "cc_capnp_library")
```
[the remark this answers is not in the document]
garrettgu10: I think it's just alphabetized

src/pyodide/BUILD.bazel
```diff
@@ -1,8 +1,9 @@
 load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
-load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@capnp-cpp//src/capnp:cc_capnp_library.bzl", "cc_capnp_library")
 load("//:build/capnp_embed.bzl", "capnp_embed")
+load("//:build/pyodide_bucket.bzl", "PYODIDE_PACKAGE_BUCKET_URL")
```
npaun, line 6: Lots of unrelated changes here but they do make the file nicer.

npaun (approved):

build/wd_ts_bundle.bzl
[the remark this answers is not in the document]
anonrig: This change is caused by `buildifier` the default formatter of bazel files. Unfortunately, we don't run the formatter by default, but clion like IDE's properly fix the formatting issues.

src/pyodide/BUILD.bazel
anonrig: This is buildifier fixing the formatting errors.

fhanau (approved):

fhanau:
LGTM outside of the CI issues, make sure to run the upstream CI on this too to ensure it won't break.

mikea (approved):

build/wd_ts_bundle.bzl
```diff
@@ -80,6 +80,7 @@ def wd_ts_bundle_capnp(
     if lint:
         # todo: lint js_srcs too, not just ts_srcs
         eslint_bin.eslint_test(
+            size = "enormous",
```
fhanau, line 83: Nit: I haven't encountered the @eslint job being overly slow before. Not to bikeshed too much but large should be sufficient right?

fhanau (approved):

build/wd_ts_bundle.bzl
anonrig: previously we weren't using typescript-eslint correctly. type-aware linting wasn't running, hence, we didn't encounter this before. this will increase with the amount of js+ts code we write.

fhanau: That makes sense to me. My point is that large (60s) should be sufficient and we won't need enormous (240s) – at present the two lint jobs appear to take 4.7s and 8.8s, respectively based on CI. I think the intention behind these categories is in part to avoid having enormous tests where possible to avoid a CI slowdown, this fits well within a 60s window.

jasnell (approved):

build/wd_ts_bundle.bzl
anonrig: Ah I see. Sounds good. I'll change it to `large`. Thanks for the review & explanation.

PR #2563, opened 2024-08-20
docs: clarify ExternalServer DNS resolution behavior

isker:
From reading the source code of workerd and KJ, I can see that ExternalServer does DNS resolution only once ever. This was certainly surprising to me, and I suspect it would be a surprise to many, as ExternalServer is the seemingly most natural way to get a binding to a particular networked service, but it is in fact only suitable to look up domain names that are guaranteed to not change over the lifetime of the workerd service on which it is configured, like container sidecars.

So, be more explicit about the behavior in the documentation.

### DIFF
--- a/src/workerd/server/workerd.capnp
+++ b/src/workerd/server/workerd.capnp
@@ -652,7 +652,9 @@ struct ExternalServer {
   # - "[1234:5678::abcd]:80": Connect to the given IPv6 address and port.
   # - "unix:/path/to/socket": Connect to the given Unix Domain socket by path.
   # - "unix-abstract:name": On Linux, connect to the given "abstract" Unix socket name.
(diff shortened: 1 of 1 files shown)


kentonv:
I think this is a bug. If we document it, we should say that it's a bug that may be fixed later. But maybe we should just fix it?

isker:
You added the TODO two years ago in https://github.com/cloudflare/workerd/commit/dc27ddcb80f29a2da23ad370f138a2b5b21ae585. So it doesn’t seem to be a bug as much as consciously acknowledged missing functionality. And I don’t think that commit regressed DNS resolution in any way, I think ExternalServer has always been this way.

Either way, I have got what it takes to clarify the documentation but not, I think, to add this functionality. So if you’d prefer the latter to the exclusion of the former, go ahead and close this PR.

PR #2566, opened 2024-08-20
small perf improvements on workerd/api

anonrig:
Small performance improvements on workerd/api

### DIFF
--- a/src/workerd/api/data-url.c++
+++ b/src/workerd/api/data-url.c++
@@ -10,6 +10,18 @@ kj::Maybe<DataUrl> DataUrl::tryParse(kj::StringPtr url) {
   return kj::none;
 }
 
+static constexpr kj::FixedArray<uint8_t, 256> ascii_whitespace_table = []() consteval {
+  kj::FixedArray<uint8_t, 256> result{};
+  for (uint8_t c: {0x09, 0x0a, 0x0c, 0x0d, 0x20}) {
+    result[c] = true;
+  }
+  return result;
+}();
+
+constexpr bool isAsciiWhitespace(uint8_t c) noexcept {
+  return ascii_whitespace_table[c];
+}
+
 kj::Maybe<DataUrl> DataUrl::from(const jsg::Url& url) {
   if (url.getProtocol() != "data:"_kj) return kj::none;
   auto clone = url.clone(jsg::Url::EquivalenceOption::IGNORE_FRAGMENTS);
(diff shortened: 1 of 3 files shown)


src/workerd/api/data-url.c++
```diff
@@ -21,12 +22,16 @@ kj::Maybe<DataUrl> DataUrl::from(const jsg::Url& url) {
...
+  static constexpr std::array<uint8_t, 256> ascii_whitespace_table = []() consteval {
+    std::array<uint8_t, 256> result{};
+    for (uint8_t c : { 0x09, 0x0a, 0x0c, 0x0d, 0x20 }) {
+      result[c] = true;
+    }
+    return result;
+  }();
+
+  static constexpr auto isAsciiWhitespace = [](uint8_t c) {
+    return ascii_whitespace_table[c];
```
npaun, line 34: LGTM, but out of curiosity couldn't you just do something like:

```suggestion
  static constexpr auto isAsciiWhitespace = [](uint8_t c) {
   if constexpr (c == 0x09 || c== 0x0a blah blah)
```

npaun (approved):

src/workerd/api/data-url.c++
npaun: Ah no, that wouldn't be as good as the table for runtime arguments

jasnell (approved):

src/workerd/api/data-url.c++
anonrig: Hash map approach is faster because it uses less instructions: https://godbolt.org/z/zavxo36PM

src/workerd/api/data-url.c++
```diff
@@ -21,12 +22,16 @@ kj::Maybe<DataUrl> DataUrl::from(const jsg::Url& url) {
   // Per the fetch spec, it doesn't matter if the comma is within a quoted
   // string value in the MIME type... which is fun.
 
-  static const auto isAsciiWhitespace = [](auto c) {
-    return c == 0x09 /* tab */ ||
-          c == 0x0a /* lf  */ ||
-          c == 0x0c /* ff  */ ||
-          c == 0x0d /* cr  */ ||
-          c == 0x20 /* sp  */;
+  static constexpr std::array<uint8_t, 256> ascii_whitespace_table = []() consteval {
```
fhanau, line 25: Have you considered using kj::FixedArray here? Looks like at least some of its methods are constexpr and we generally try to lightly use the C++ STL. LGTM otherwise.

anonrig: Nice. I didn't know kj::FixedArray existed. Just updated now. I also updated the previous optimization to use fixed array as well. Appreciate the review @fhanau

src/workerd/api/data-url.c++
```diff
@@ -10,6 +10,18 @@ kj::Maybe<DataUrl> DataUrl::tryParse(kj::StringPtr url) {
   return kj::none;
 }
 
+constexpr kj::FixedArray<uint8_t, 256> ascii_whitespace_table = []() consteval {
```
fhanau, line 13: Can this be static? Not sure if it makes much of a difference with constexpr. Same goes for mimetype.c++ `token_table` and `quoted_string_token_table`.

anonrig: Yeah, no worries!

fhanau (approved):

PR #2573, opened 2024-08-21
Add Python formatting with Ruff

hoodmane:


npaun:
I don't see a ruff config file in the PR. Shouldn't we add your config?

danlapid:
Make sure to run the formatter and commit the changed files

hoodmane:
Well I want to see the CI fail first.

hoodmane:
> Shouldn't we add your config?

I think the configuration I suggested before is all configuring the lints for `ruff check`, for `ruff format` I think we can just use the defaults and it will work fine. I looked through [the list of `ruff format` configuration options](https://docs.astral.sh/ruff/settings/#format) and none of them look that interesting to me.

tools/cross/format.py
```diff
@@ -102,6 +104,26 @@ def prettier(files: List[str], check: bool = False) -> bool:
...
+        if check:
+            # In ci, fail.
+            logging.error(msg)
+            return False
+        else:
+            # In a local checkout, let it go. If the user wants Python
+            # formatting they can install ruff and run again.
+            logging.warning(msg)
+            return True
+        return not check
```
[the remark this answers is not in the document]
hoodmane: Yeah, thanks

hoodmane:
I think I'll add ruff linting in a separate PR.

npaun (approved):
Don't forget to add the relevant commit to .git-blame-ignore-revs

.github/workflows/lint.yml
```diff
@@ -27,6 +30,9 @@ jobs:
       - name: Install project deps with pnpm
         run: |
           pnpm i
+      - name: Install Ruff
+        run: |
+          pip install ruff
```
anonrig, line 35: Can we use ruff-action? https://github.com/ChartBoost/ruff-action

tools/cross/format.py
```diff
@@ -102,6 +104,25 @@ def prettier(files: List[str], check: bool = False) -> bool:
...
+        if check:
+            # In ci, fail.
+            logging.error(msg)
+            return False
+        else:
+            # In a local checkout, let it go. If the user wants Python
+            # formatting they can install ruff and run again.
+            logging.warning(msg)
+            return True
+    cmd = [RUFF, "format"]
```
danlapid, line 119: Do we have any kind of "verbose" flag for ruff so we see what files it formats or checks?

anonrig: Actually, I was just about to open a pull-request to remove it, since it reduces formatting performance. Any particular reason to enable it?

npaun: I find it convenient for local use to know which files were checked, and if they needed their formatting corrected or not.

.github/workflows/lint.yml
hoodmane: Well with the current approach it works locally for people to do `python ./tools/cross/format.py` if they have ruff installed, whereas if we used the action there wouldn't be a good local approach. I think @danlapid said he'd prefer to modify `format.py` over using ci-specific automation.

tools/cross/format.py
danlapid: Agree with @npaun also, formatting shouldn't take too long for regular runs where a user just has a few files staged for commits.
The CI run might take a while but it's gonna be faster than build&test anyway.

danlapid: @anonrig  What kind of time difference did you see between with verbose vs without?

anonrig: I've opened a pull-request about 2 changes that I think will fulfill everybody's needs. https://github.com/cloudflare/workerd/pull/2576

> @anonrig What kind of time difference did you see between with verbose vs without?

I'm not sure. I have to benchmark it using a tool like `hyperfine` to came up with a scientific result.

danlapid: Yea sounds right.

.github/workflows/lint.yml
```diff
@@ -13,6 +13,9 @@ jobs:
       - uses: actions/checkout@v4
         with:
           show-progress: false
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
```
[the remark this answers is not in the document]
hoodmane: To be honest, I only see `3.12` repeated in two places, so I'm not sure this would be helpful.

tools/cross/format.py
hoodmane: > I find it convenient for local use to know which files were checked, and if they needed their formatting corrected or not.

Unfortunately I don't think that ruff has a mode that does this without printing way too much output.

npaun:
PEP 8 recommends 4 spaces for Python, so that's my personal preference

npaun:
Commenting just in case: Make sure to merge this PR via "Merge" and not "Squash". Otherwise, .git-blame-ignore-revs will not work properly.

hoodmane:
I wish it were possible to disable squash merge ahead of time on a particular PR, I have a tendency to accidentally squash merge patch release PRs and things like that.

npaun:
Yeah, same here.

.github/workflows/lint.yml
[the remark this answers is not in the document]
fhanau: FWIW the Ubuntu 24.04 dev image defaults to python 3.12, so this is currently a no-op (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime). There's a case for keeping it though to use a fixed python version even if we update the build image, or if we want to update to python 3.13 early (comes out in 6 weeks).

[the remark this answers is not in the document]
anonrig: This is also an indicator for IDEs to automatically use a specific version of Python whenever the interpreter is not configured. I wish this change was included in this pull-request.

.github/workflows/lint.yml
anonrig: `ruff-action` is used to install ruff. it is not used to execute ruff. right now, we will always use the latest ruff version.

PR #2576, opened 2024-08-21
update prettier and clang-format cli options in format.py

anonrig:
- Replaced `--check` with `--list-different` to reduce noise (hides the files that does not need formatting) 
- Added `--log-level=warn` on formatting where now it prints only the files that actually changed.

### DIFF
--- a/tools/cross/format.py
+++ b/tools/cross/format.py
@@ -85,7 +85,7 @@ def filter_files_by_exts(
 
 
 def clang_format(files: List[str], check: bool = False) -> bool:
-    cmd = [CLANG_FORMAT, "--verbose"]
+    cmd = [CLANG_FORMAT]
     if check:
         cmd += ["--dry-run", "--Werror"]
     else:
@@ -95,7 +95,7 @@ def clang_format(files: List[str], check: bool = False) -> bool:
 
 
 def prettier(files: List[str], check: bool = False) -> bool:
-    cmd = [PRETTIER]
+    cmd = [PRETTIER, "--log-level=warn"]
     if check:
         cmd.append("--check")
     else:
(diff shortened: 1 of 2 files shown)


npaun:
If it's just to cut down on noise I think --check + --log-level=warn will look nicer than --list-different?

anonrig:
> Is this for taste or performance? If it's for the latter, what kind of difference are you seeing in formatting time?

Mostly for taste. I don't have `githooks` enabled, and I run `python3 tools/cross/format.py` manually, which just forces me to clear my terminal everytime I run it.

My local benchmarks using `hyperfine -i "node_modules/.bin/prettier . --check" "node_modules/.bin/prettier . --list-different"` shows that `--list-different` is 3-5% faster but I don't think that's really my reasoning for this pull-request.

hoodmane:
I agree with @anonrig that we should get rid of `clang-format --verbose`. I don't want the formatting command to generate multiple screens worth of output.

jasnell:
+1 to going with the less verbose output

anonrig:
Would you mind reviewing this pull-request?

danlapid (approved):

tools/cross/format.py
```diff
@@ -97,9 +97,13 @@ def clang_format(files: List[str], check: bool = False) -> bool:
 def prettier(files: List[str], check: bool = False) -> bool:
     cmd = [PRETTIER]
     if check:
-        cmd.append("--check")
+        # This flag will only print files that require formatting
+        # and change exit code of the process depending on state.
+        cmd.append("--list-different")
```
npaun, line 102:
```suggestion
        cmd.append("--check")
```

I'd prefer the output of check because it gives you some context as to what it's printing.

tools/cross/format.py
```diff
@@ -97,9 +97,13 @@ def clang_format(files: List[str], check: bool = False) -> bool:
...
     else:
         cmd.append("--write")
+        # This will only print files that requires formatting.
+        cmd.append("--log-level=warn")
```
npaun, line 106:
```suggestion
    cmd.append("--log-level=warn")
```

But this is great so it doesn't print when there's nothing wrong

jasnell (approved):

npaun (approved):

PR #2577, opened 2024-08-21
Add autoformatting for Bazel files

npaun:


tools/cross/format.py
```diff
@@ -120,53 +144,56 @@ def git_get_modified_files(
...
     ),
     FormatConfig(
         directory="src",
-        extensions=(".js", ".ts", ".cjs", ".ejs", ".mjs"),
+        globs=("*.js", "*.ts", "*.cjs", "*.ejs", "*.mjs"),
         formatter=prettier,
     ),
-    FormatConfig(directory="src", extensions=(".json",), formatter=prettier),
-    # TODO: lint bazel files
+    FormatConfig(directory="src", globs=(".json",), formatter=prettier),
```
danlapid, line 174: nit:
```suggestion
    FormatConfig(directory="src", globs=("*.json",), formatter=prettier),
```

fhanau:
See internal PR 8342 for some prior work on enabling buildifier automatically on the downstream repo – I was planning to patch up that PR today but looks like you got here first. It uses a different approach to downloading buildifier that I think is preferable, but I'll have to incorporate Dan's suggestions on how to invoke it – will provide some detailed notes on that later today and review this.

danlapid:
> See internal PR 8342 for some prior work on enabling buildifier automatically on the downstream repo – I was planning to patch up that PR today but looks like you got here first. It uses a different approach to downloading buildifier that I think is preferable, but I'll have to incorporate Dan's suggestions on how to invoke it – will provide some detailed notes on that later today and review this.

Seems to me like we mind as well land this change as is and you can convert the installation method in a separate PR later.

danlapid (approved):
Dependant on @fhanau approval of bazel file changes.

tools/cross/format.py
```diff
@@ -74,17 +83,30 @@ def check_clang_format() -> bool:
...
     ]
 
 
-def clang_format(files: List[str], check: bool = False) -> bool:
+def matches_any_glob(globs: Tuple[str, ...], file: Path) -> bool:
+    return any(file.match(glob) for glob in globs)
+
+
+def is_relative_to(path: Path, prefix: Path) -> bool:
+    # Polyfill for method introduced in Python 3.9
```
fhanau, line 101: Is anyone on a system still using python 3.9?

npaun: @danlapid suggested we keep 3.8 compatibility as it isn't too hard to do

danlapid: No strong feelings on my end, I don't hate the polyfill but I don't care if we want to update the minimum required to any version.

fhanau:
This also applies for the other linters we added be we should document that buildifier needs to be installed as running the formatter is now a key part of development, even if invoked indirectly.

tools/cross/format.py
npaun: OK let's go with 3.9. I'm just itching to remove those imports from `typing`

npaun:
I added a new issue to document the linters: https://github.com/cloudflare/workerd/issues/2581

PR #2578, opened 2024-08-21
blank but executable pre-push

npaun:


npaun:
It seems to block some specific content

PR #2585, opened 2024-08-22
POC for basic web platform test integration

jasnell:
This is just a proof of concept to start working through what is necessary to integrate web platform tests into wd-tests.

There are some limitations to this approach...

1. All module imports are evaluated outside of an IoContext, even if they are dynamic imports. This means that no test can schedule any actual i/o. Any test that depends on setTimeout, for instance, won't work with this pattern. This also means that none of the fetch tests would work under this pattern (EventSource, timers, Web Sockets, etc, all wouldn't be possible with this approach)
2. The test harness messes with the global scope. Ideally we'd be running these things in something like a `ShadowRealm` ... otherwise there's always a possibility of the harness conflicting with the global scope, tho that is unlikely in practice


build/BUILD.wpt
```diff
@@ -0,0 +1,9 @@
+filegroup(
+  name = "url",
+  srcs = glob(
+    include = ["url/**/*"],
+    allow_empty = True,
+  ),
+  visibility = ["//visibility:public"],
+)
```
jasnell, line 8: We would need to export filegroups for each of the web platform test groups we want... we don't need to export filegroups for everything since most of the wpts will never be relevant

src/workerd/api/wpt/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
...
+    args = [
+      "--experimental",
+    ],
+    data = [
+      f.removesuffix(".wd-test") + ".js",
+      "@wpt//:url",
+    ],
+) for f in glob(
+    ["**/*.wd-test"],
+)]
```
jasnell, line 16: Most likely should list each individually rather than using the glob here so that we can selectively import the appropriate wpt filegroups.

src/workerd/api/wpt/url-test.js
```diff
@@ -0,0 +1,42 @@
...
+globalThis.test = (callback, message) => {
+  try {
+    callback();
+  } catch (err) {
+    const aerr = new AggregateError([err], message);
+    globalThis.errors.push(aerr);
+  }
+}
+
+globalThis.errors = [];
```
jasnell, line 29: In order to be able to run the tests we need to define the test harness methods along with a way of collecting the results. This is a bit wonky here.

src/workerd/api/wpt/url-test.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const test = {
+  async test(_, env) {
+    const foo = await import('url-origin.any.js');
```
jasnell, line 33: The tests end up running when the module is evaluated. We have to set the global test harness methods before the dynamic import, then check the `globalThis.errors` when we're done. The `globalThis.errors` is only specific to this test file, it is not actually part of the WPT test harness. We have to do this because the tests themselves may not actually propagate the error (in the case of url-origin.any.js we end up with a swallowed rejected promise) if we don't have this.

src/workerd/api/wpt/url-test.wd-test
```diff
@@ -0,0 +1,28 @@
...
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "url-test.js"),
+          (name = "url-origin.any.js",
+           esModule = embed "../../../../../wpt/url/url-origin.any.js"),
+          (name = "resources/urltestdata.json",
+           json = embed "../../../../../wpt/url/resources/urltestdata.json"),
+          (name = "resources/urltestdata-javascript-only.json",
+           json = embed "../../../../../wpt/url/resources/urltestdata-javascript-only.json"),
+        ],
```
jasnell, line 15: Each of the resources needed for the test end up needing to be embedded directly since we do not allow dynamic evaluation.

anonrig: Can we add *.wd-test inside this folder to gitignore and dynamically generate the config files?

src/workerd/api/wpt/url-test.js
anonrig: We also need a way to ignore certain errors. Possibly from a `module_name.json` file with a JSON schema that conforms to our needs

src/workerd/api/wpt/url-test.wd-test
jasnell: probably? I would need to buff up on my Bazel Kung Fu for that as my Bazel Kung Fu is weak.

src/workerd/api/wpt/url-test.js
jasnell: Yeah, I'm thinking that can be a separate json module that is imported and used by the harness.

src/workerd/api/wpt/url-test.js
```diff
@@ -0,0 +1,23 @@
...
+export const test = {
+  async test() {
+    // The tests will be run when the module is imported. Note that
+    // there are limitations to this in that the module is not allowed
+    // to perform any I/O... so any test that requires setTimeout,
+    // for instance, will fail... this also means we cannot run
+    // fetch tests from WPT... sad face.
+    const foo = await import('url-origin.any.js');
+
+    if (globalThis.errors.length > 0) {
```
anonrig, line 16: Maybe we can make this if statement a function in harness so we don't have to copy paste in everywhere?

src/workerd/api/wpt/url-test.wd-test
mikea: here is a good example: https://github.com/cloudflare/workerd/blob/main/src/workerd/server/tests/python/import_tests.bzl

src/workerd/api/wpt/BUILD.bazel
```diff
@@ -0,0 +1,4 @@
+load("//:build/wd_test.bzl", "wd_test")
+load("//src/workerd/api/wpt:generate-tests.bzl", "gen_wpt_tests")
+
+gen_wpt_tests(glob(["**/*.js"]))
```
jasnell, line 4: Some documentation comments around this would be helpful.

src/workerd/api/wpt/generate-tests.bzl
```diff
@@ -0,0 +1,86 @@
...
+        modules = [
+          (name = "worker", esModule = embed "{}.js"),
+          (name = "harness", esModule = embed "../../../../../workerd/src/wpt/harness.js"),
+          {}
+        ],
+        bindings = [
+          (name = "wpt", service = "wpt"),
+        ],
+        compatibilityDate = "2024-07-01",
+        compatibilityFlags = ["nodejs_compat_v2"],
```
jasnell, line 23:
```suggestion
        compatibilityFlags = ["nodejs_compat_v2", "experimental"],
```

Should include experimental just to be safe.

jasnell: Also, for compatibilityDate... could that be set dynamically to the current day whenever these are run? I don't want us to have to go back and manually change that repeatedly over time.

src/workerd/api/wpt/url-test.js
```diff
@@ -0,0 +1,17 @@
+import * as harness from 'harness';
+
+export const urlConstructor = {
+  async test() {
+    harness.prepare();
+    await import('url-constructor.any.js');
+    harness.validate();
```
jasnell, line 7: Side note: Our workerd tests could benefit from having a `before() { ... }` and `after() { ... }` mechanism.

src/wpt/harness.js
```diff
@@ -0,0 +1,41 @@
+import { strictEqual } from 'node:assert';
```
jasnell, line 1: Side note... let's try not to forget adding the copyright headers to these (speaking to myself here mostly)

src/workerd/api/wpt/generate-tests.bzl
anonrig: I've updated the implementation to use dynamic dates, and added experimental flag.

src/workerd/api/wpt/url-test.js
jasnell: I suppose we could have a pattern like...

```
import * as harness from 'harness';

export const urlConstructor = {
  test: harness.test('url-constructor.any.js');
};
```

Where the `harness.test(...)` handles any of the necessary boilerplate for the most typical cases?

anonrig: Yeah sounds like a good idea

npaun:
@jasnell @anonrig I think this might be almost ready to merge. I think I found a pretty elegant bazel solution.

build/BUILD.wpt
```diff
@@ -1,8 +1,17 @@
...
-    name = "url",
-    srcs = glob(
-        include = ["url/**/*"],
-        allow_empty = True,
+directories = glob(
+    ["*"],
+    exclude = glob(
+        [
+            "*",
+            ".*",
```
anonrig, line 6: Does this ignore `.github` folder? Can you validate?

build/wpt_test.bzl
```diff
@@ -0,0 +1,110 @@
...
+
+def _wpt_test_gen_impl(ctx):
+    src = ctx.actions.declare_file("{}.wd-test".format(ctx.attr.test_name))
+    ctx.actions.write(
+        output = src,
+        content = WPT_TEST_TEMPLATE.format(
+            test_name = ctx.attr.test_name,
+            test_js = wd_relative_path(ctx.file.test_js),
+            modules = generate_external_modules(ctx.attr.wpt_directory.files),
+            date = "2024-10-10",  # FIXME: we don't wanna be hardcoding that type of shizz
```
anonrig, line 36: This needs to be `2024-10-11`

npaun: Can we hardcode this value or will it need to be updated regularly?

anonrig: It cannot be hardcoded since we might introduce a change that is only available after a certain date.

build/BUILD.wpt
npaun: I tweaked the globs a little to get rid of it.

npaun:
Hmm the checksum of the Github Release for WPT keeps changing...

So far for https://github.com/web-platform-tests/wpt/archive/refs/tags/merge_pr_48695.tar.gz I've seen:

sha256-Hxn/D6x6lI9ISlCQFq620sb8x9iXplVzXPV6zumX84A=
sha256-2qTT6fguxmM+8ND0AgIlysNkqRHzELEiaCqEo6Y3PrY=
sha256-W9YnJn4yn9LpOYB3oOx2aSaG88t6qqwiWDL7zsogGdc=

build/wpt_test.bzl
npaun: Mike helped me find a fix for this

build/wpt_test.bzl
```diff
@@ -0,0 +1,110 @@
...
+        modules = [
+          (name = "worker", esModule = embed "{test_js}"),
+          (name = "harness", esModule = embed "../../../../../workerd/src/wpt/harness.js"),
+          {modules}
+        ],
+        bindings = [
+          (name = "wpt", service = "wpt"),
+        ],
+        compatibilityDate = embed "../../../../../workerd/src/workerd/io/trimmed-supported-compatibility-date.txt",
+        compatibilityFlags = ["nodejs_compat_v2", "experimental"],
```
jasnell, line 59: Nit: if the compat date is after 2024-09-23 then you can just use `nodejs_compat` instead of `nodejs_compat_v2`

src/wpt/harness.js
jasnell: @npaun ... can you go through and add the necessary copyright headers to the new files?

npaun: WPT's copyright?

jasnell:
LGTM but I'm unable to approve the PR officially since I'm the original author

src/wpt/harness.js
anonrig: Workerd.

anonrig (approved):

PR #2587, opened 2024-08-22
Add ruff.toml with Python lint configuration

hoodmane:


tools/cross/format.py
```diff
@@ -129,10 +134,18 @@ def ruff(files: list[Path], check: bool = False) -> bool:
             # formatting they can install ruff and run again.
             logging.warning(msg)
             return True
+    # Format
     cmd = [RUFF, "format"]
     if check:
         cmd.append("--diff")
     result = subprocess.run(cmd + files)
+    if result.returncode != 0:
+        return False
```
danlapid, line 143: Do we want an early return here? I'd assume we want to run both and return False if either failed.

danlapid:
Love it!

tools/cross/format.py
hoodmane: Well, the logic here will return false if either failed.

hoodmane: Actually though we should do the lint check before the format because the linter fixes can break the formatting.

danlapid: Well what I was thinking was:
Someone doesn't lint or format, pushes his commit, gets a lint CI formatting fail, fixes the errors then gets a lint CI lint fails - rather annoying.
Let's check both so that we can see all errors printed out.
We don't fail-fast between languages either.

hoodmane: Okay.

danlapid (approved):
LGTM

hoodmane:
Thanks @danlapid!

PR #2594, opened 2024-08-22
[nfc] explicit test targets

mikea:
Also marking sql-test as large since it is slow sometimes.

### DIFF
--- a/src/workerd/api/BUILD.bazel
+++ b/src/workerd/api/BUILD.bazel
@@ -208,16 +208,258 @@ kj_test(
     deps = ["//src/workerd/tests:test-fixture"],
 )
 
-[wd_test(
-    src = f,
+wd_test(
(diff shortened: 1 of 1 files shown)


src/workerd/api/BUILD.bazel
```diff
@@ -228,26 +470,66 @@ wd_test(
...
-)]
+    data = ["http-test-ts.ts"],
+)
 
 # Enable GPU tests if experimental GPU support is enabled. Unfortunately, this depends on the right
 # drivers being available on Linux and macOS, so turn it off by default. Run GPU tests with
 # `bazel test //src/workerd/api:gpu/<test name>`.
-[wd_test(
-    src = f,
+wd_test(
```
anonrig, line 482: Nit: Can we make this a function so we don't repeat everything?

anonrig (approved):

PR #2598, opened 2024-08-23
Fix llvm-symbolizer in CI

npaun:
Make sure we actually install llvm-symbolizer, so that traces in CI test failures make sense.


danlapid:
Should probably make a test fail in CI to make sure this works

npaun:
Check this one out from when I tried it on one of my PRs: https://github.com/cloudflare/workerd/actions/runs/10530661974/job/29181058646

danlapid (approved):

fhanau (approved):

PR #2599, opened 2024-08-23
improve zlib for nodejs_compat

anonrig:
Improves nodejs compat and adds a test that fails without the changes in the pull-request.

### DIFF
--- a/src/workerd/api/node/zlib-util.h
+++ b/src/workerd/api/node/zlib-util.h
@@ -6,8 +6,9 @@
 
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/exception.h>
+#include <workerd/jsg/buffersource.h>
+#include <zlib.h>
 
-#include "zlib.h"
 #include <kj/array.h>
 #include <kj/compat/brotli.h>
 #include <kj/vector.h>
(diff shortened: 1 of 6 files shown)


jasnell (approved):
looks like some lint issues and I'm not sure why the scrypt test is suddenly timing out but that appears unrelated. Will be investigating that separately in a bit.

src/workerd/api/node/zlib-util.h
```diff
@@ -210,24 +219,44 @@ class ZlibUtil final: public jsg::Object {
...
+    void write_(jsg::Lock& js,
+        int flush,
+        jsg::Optional<kj::Array<kj::byte>> input,
+        int inputOffset,
+        int inputLength,
+        kj::ArrayPtr<kj::byte> output,
+        int outputOffset,
+        int outputLength);
+
+    // TODO(soon): Find a way to expose functions with templates using JSG_METHOD.
```
npaun, line 235: Give this a try:

JSG_METHOD_NAMED(write, template write_<true>)
JSG_METHOD_NAMED(writeSync, write_<false>)

npaun (approved):

src/workerd/api/node/zlib-util.h
anonrig: unfortunately, it also doesn't work

npaun: Ah you also need an explicit instantiation of the template, to tell C++ to actually generate the code. Looks kind of clunky but it would be something like:

```c++
template void ZlibUtil::ZlibStream::write_<true>(all your params here);
template void ZlibUtil::ZlibStream::write_<false>(all your params here);
```

anonrig: yeah, i'll leave it to a different pull-request for now.

jasnell:
Since this touches `buffersource.h`, please run an internal CI before merging.

anonrig:
> Since this touches `buffersource.h`, please run an internal CI before merging.

Internal PR passes.

PR #2601, opened 2024-08-24
[nfc] 08-23-24 build cleanup

fhanau:
[build] workerd/io target cleanup, fold up IoContext::reportPromiseRejectEvent()
\- Dissolve glob in io target
\- Isolate more source files into smaller target
\- reportPromiseRejectEvent() does not involve IoContext internals and is only used within Worker, fold it into worker.c++.
This makes io easier to maintain and makes the target a bit smaller.

[nfc] Extended api include cleanup aided by include-what-you-use
\- Add _LIBCPP_REMOVE_TRANSITIVE_INCLUDES to compile_flags.txt so errors for missing headers are available within the editor too

I wanted to divide the io target further and see what impact IWYU has in practice. IWYU made it easier to clean up some headers, but the real-world impact is kinda disappointing (1080MB => 1071MB size of preprocessed workerd source). Still, just having frequently used headers like http.h be smaller should also help with managing include sizes in the downstream project. In some cases I also fixed headers being unable to compile on their own or moved functions from a header to the corresponding cpp file if that allowed removing includes from the header.

### DIFF
--- a/src/workerd/api/crypto/crypto.h
+++ b/src/workerd/api/crypto/crypto.h
@@ -10,8 +10,8 @@
 #include <workerd/jsg/buffersource.h>
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/buffersource.h>
-#include <openssl/err.h>
-#include "../streams.h"
+#include <workerd/api/streams/writable.h>
+#include <openssl/base.h>  // for EVP_MD_CTX, X509
 
 namespace workerd::api {
 namespace node {
--- a/src/workerd/api/crypto/impl.h
+++ b/src/workerd/api/crypto/impl.h
@@ -10,8 +10,9 @@
 #include "crypto.h"
 #include <workerd/api/util.h>
 #include <kj/encoding.h>
-#include <openssl/evp.h>
-#include <openssl/bio.h>
+#include <openssl/base.h>
+#include <openssl/bn.h>
+#include <openssl/err.h>
 
 typedef struct bignum_st BIGNUM;
 
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -6,24 +6,20 @@
 
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/async-context.h>
-#include <workerd/util/abortable.h>
 #include <kj/compat/http.h>
 #include <map>
 #include "basics.h"
 #include "cf-property.h"
-#include "streams.h"
+#include <workerd/api/streams/readable.h>
 #include "form-data.h"
 #include "web-socket.h"
 #include "url.h"
-#include "url-standard.h"
 #include "blob.h"
 #include <workerd/io/compatibility-date.capnp.h>
 #include "worker-rpc.h"
 #include "queue.h"
 
 namespace workerd::api {
-
-struct QueueResponse;
 
 class Headers final: public jsg::Object {
 private:
--- a/src/workerd/api/kv.h
+++ b/src/workerd/api/kv.h
@@ -4,10 +4,14 @@
 
 #pragma once
 
-#include <workerd/jsg/jsg.h>
-#include "streams.h"
+#include <workerd/api/streams/readable.h>
 #include <workerd/io/limit-enforcer.h>
+#include <workerd/jsg/jsg.h>
 
+namespace kj {
+class HttpClient;
+class HttpHeaders;
+}  // namespace kj
 namespace workerd {
 class IoContext;
 }
--- a/src/workerd/api/sockets.h
+++ b/src/workerd/api/sockets.h
@@ -8,7 +8,8 @@
 #include <workerd/jsg/modules.h>
 #include <workerd/jsg/modules-new.h>
 #include <workerd/jsg/url.h>
-#include "streams.h"
+#include <workerd/api/streams/readable.h>
+#include <workerd/api/streams/writable.h>
 
 namespace workerd::api {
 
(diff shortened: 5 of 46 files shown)


fhanau:
- [ ] create downstream PR to make any needed changes there before merging this.

anonrig (approved):
amazing work!

fhanau:
Added a commit here that marks `crypto_scrypt-test` as large – it appears to time out in debug and asan configurations on CI sometimes including on main.

anonrig (approved):

anonrig:
Can we merge this to fix the timeout issues in CI?

src/workerd/api/crypto/crypto.h
```diff
@@ -10,8 +10,8 @@
 #include <workerd/jsg/buffersource.h>
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/buffersource.h>
-#include <openssl/err.h>
-#include "../streams.h"
+#include "openssl/base.h"  // for EVP_MD_CTX, X509
+#include "workerd/api/streams/writable.h"
```
danlapid, line 14: nit:
```suggestion
#include <openssl/base.h>  // for EVP_MD_CTX, X509
#include <workerd/api/streams/writable.h>
```

src/workerd/api/crypto/crypto.h
```diff
@@ -10,8 +10,8 @@
 #include <workerd/jsg/buffersource.h>
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/buffersource.h>
-#include <openssl/err.h>
-#include "../streams.h"
+#include "openssl/base.h"  // for EVP_MD_CTX, X509
```
danlapid, line 13: nit:
```suggestion
#include <openssl/base.h>  // for EVP_MD_CTX, X509
```

src/workerd/api/crypto/impl.h
```diff
@@ -10,8 +10,9 @@
 #include "crypto.h"
 #include <workerd/api/util.h>
 #include <kj/encoding.h>
-#include <openssl/evp.h>
-#include <openssl/bio.h>
+#include <openssl/base.h>
+#include "openssl/bn.h"
```
danlapid, line 14: nit
```suggestion
<include "openssl/bn.h>
```

src/workerd/api/http.h
```diff
@@ -6,25 +6,21 @@
 
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/async-context.h>
-#include <workerd/util/abortable.h>
 #include <kj/compat/http.h>
 #include <map>
 #include "basics.h"
 #include "cf-property.h"
-#include "streams.h"
+#include "workerd/api/streams/readable.h"
```
danlapid, line 13:
```suggestion
#include <workerd/api/streams/readable.h>
```

src/workerd/api/kv.h
```diff
@@ -4,10 +4,14 @@
 
 #pragma once
 
-#include <workerd/jsg/jsg.h>
-#include "streams.h"
+#include "workerd/api/streams/readable.h"
```
danlapid, line 7:
```suggestion
#include <workerd/api/streams/readable.h>
```

src/workerd/api/sockets.h
```diff
@@ -8,7 +8,8 @@
 #include <workerd/jsg/modules.h>
 #include <workerd/jsg/modules-new.h>
 #include <workerd/jsg/url.h>
-#include "streams.h"
+#include "workerd/api/streams/readable.h"
```
danlapid, line 11:
```suggestion
#include <workerd/api/streams/readable.h>
```

src/workerd/api/sockets.h
```diff
@@ -8,7 +8,8 @@
...
+#include "workerd/api/streams/writable.h"
```
danlapid, line 12:
```suggestion
#include <workerd/api/streams/writable.h>
```

src/workerd/api/system-streams.h
```diff
@@ -6,7 +6,7 @@
 // Implementations of ReadableStreamSource / WritableStreamSink which wrap system streams (sockets),
 // handle encoding/decoding, and optimize pumping between them when possible.
 
-#include "streams.h"
+#include "workerd/api/streams/common.h"  // for StreamEncoding, ...
```
danlapid, line 9:
```suggestion
#include <workerd/api/streams/common.h>  // for StreamEncoding, ...
```

src/workerd/io/io-context.h
```diff
@@ -8,9 +8,8 @@
 #include <workerd/io/io-channels.h>
 #include <workerd/io/io-gate.h>
 #include "io-own.h"
-#include "io-timers.h"
-#include "io-thread-context.h"
-#include "limit-enforcer.h"
+#include "workerd/io/io-timers.h"
```
danlapid, line 11:
```suggestion
#include <workerd/io/io-timers.h>
```

src/workerd/io/io-context.h
```diff
@@ -8,9 +8,8 @@
...
+#include "workerd/io/io-thread-context.h"
```
danlapid, line 12:
```suggestion
#include <workerd/io/io-thread-context.h>
```

danlapid (approved):
I'd like to keep include brackets correct but LGTM otherwise.

src/workerd/api/crypto/crypto.h
fhanau: I blame Google's style guide for these 😄 – they eschew the use of angled brackets for some reason, and since IWYU was developed with Google3 in mind that's what it is producing.

danlapid: It's actually our fault, I tried to fix it but I don't understand bazel well enough.
IWYU will include with "" if file is in `-I` include path and with <> if file is in `-isystem` include path.
Ideally I'd like our bazel configuration to include all these libraries as ONLY `-isystem` instead of the current behavior which is both

PR #2605, opened 2024-08-25
Use v8::LocalVector in multiple places

jasnell:
v8 Local handles aren't really supposed to be heap allocated at all. A short while back v8 introduced a new `v8::LocalVector<T>` collection type that more correctly handles the allocation and handling of a collection of locals. Update workerd to use it.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -690,14 +690,14 @@ void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, v8::Local<v8::Funct
           (jsg::Lock & js, const v8::FunctionCallbackInfo<v8::Value>& args)->v8::Local<v8::Value> {
             return js.tryCatch([&]() -> v8::Local<v8::Value> {
               auto function = fn.getHandle(js);
-              auto context = js.v8Context();
 
-              kj::Vector<v8::Local<v8::Value>> argv(args.Length());
+              v8::LocalVector<v8::Value> argv(js.v8Isolate, args.Length());
               for (int n = 0; n < args.Length(); n++) {
-              argv.add(args[n]);
+              argv[n] = args[n];
               }
 
-              return jsg::check(function->Call(context, js.v8Null(), args.Length(), argv.begin()));
+              return jsg::check(
+                  function->Call(js.v8Context(), js.v8Null(), argv.size(), argv.data()));
             }, [&](jsg::Value exception) {
               reportError(js, jsg::JsValue(exception.getHandle(js)));
               return js.v8Undefined();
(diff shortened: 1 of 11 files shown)


src/workerd/api/global-scope.c++
```diff
@@ -690,14 +690,14 @@ void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, v8::Local<v8::Funct
           (jsg::Lock & js, const v8::FunctionCallbackInfo<v8::Value>& args)->v8::Local<v8::Value> {
             return js.tryCatch([&]() -> v8::Local<v8::Value> {
               auto function = fn.getHandle(js);
-              auto context = js.v8Context();
 
-              kj::Vector<v8::Local<v8::Value>> argv(args.Length());
+              v8::LocalVector<v8::Value> argv(js.v8Isolate, args.Length());
```
anonrig, line 694: Nit: Can we have a helper function to convert `FunctionCallbackInfo<T>` to `LocalVector<T>`? We can add a todo to simplify it once `FunctionCallbackInfo` has iterator support?

src/workerd/api/worker-rpc.c++
```diff
@@ -501,17 +501,18 @@ JsRpcPromiseAndPipleine callImpl(jsg::Lock& js,
...
         // JS.
-        if (argv.size() > 0) {
-          serializeJsValue(js, js.arr(argv.asPtr()), [&](capnp::MessageSize hint) {
+        if (args.Length() > 0) {
+          // This is a function call with arguments.
+          v8::LocalVector<v8::Value> argv(js.v8Isolate, args.Length());
+          for (int n = 0; n < args.Length(); n++) {
+            argv[n] = args[n];
+          }
+          auto arr = v8::Array::New(js.v8Isolate, argv.data(), argv.size());
```
anonrig, line 513: Can we just eliminate LocalVector in here, and just use v8::array constructor with length, and update it? Doesn't this make an additional copy?

src/workerd/api/worker-rpc.c++
```diff
@@ -1361,7 +1363,7 @@ private:
         "arguments, the server must use class-based syntax (extending WorkerEntrypoint) "
         "instead.");
 
-    KJ_STACK_ARRAY(v8::Local<v8::Value>, arguments, kj::max(argCountFromClient + 2, arity), 8, 8);
+    v8::LocalVector<v8::Value> arguments(js.v8Isolate, kj::max(argCountFromClient + 2, arity));
```
anonrig, line 1366: Should we just change `KJ_STACK_ARRAY` implementation?

jasnell: KJ_STACK_ARRAY is a kj thing that is not specific to v8 at all. It wouldn't make sense to change it's implementation here.

src/workerd/api/worker-rpc.c++
jasnell: No, using individual calls to `->Set(...)` would be a lot slower. The `v8::LocalVector<T>` is not copying anything. This form is still more performant

anonrig (approved):

PR #2610, opened 2024-08-27
Move the default implied on date for nodejs-compat-v2

jasnell:
Push back the date where nodejs-compat implies nodejs-compat-v2

@dom96 @vickykont ... this would need to go out into a release this week.

### DIFF
--- a/src/workerd/io/compatibility-date-test.c++
+++ b/src/workerd/io/compatibility-date-test.c++
@@ -178,8 +178,8 @@ KJ_TEST("compatibility flag parsing") {
   expectCompileCompatibilityFlags("2021-05-17", {"r2_internal_beta_bindings"}, "()", {},
       CompatibilityDateValidation::FUTURE_FOR_TEST, true, false);
 
-  // nodejs_compat implies nodejs_compat_v2 on or after 2024-08-05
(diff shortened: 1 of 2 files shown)


src/workerd/io/compatibility-date-test.c++
```diff
@@ -179,7 +179,7 @@ KJ_TEST("compatibility flag parsing") {
       CompatibilityDateValidation::FUTURE_FOR_TEST, true, false);
 
   // nodejs_compat implies nodejs_compat_v2 on or after 2024-08-05
```
vicb, line 181: Should the comment be updated.
I try to never repeat values from the code in a comment because those always get out of sync at some point.
Using a constant + comment can help.

jasnell, line 181:
```suggestion
  // nodejs_compat implies nodejs_compat_v2 on or after 2024-09-23
```

dom96 (approved):
LGTM but I wonder if the date shouldn't be moved further forward just in case. It can always be moved back relatively easily if necessary, right?

jasnell:
The date is intended to align with the start of birthday week.

Issue #2572, opened 2024-08-21
🐛 Bug Report — Runtime APIs - R2Conditional parses wildcard etag as strong etag


harshal317:
Was looking into this issue here: https://github.com/cloudflare/workers-sdk/issues/6411

Specifically when using the R2Conditional type when passing in an `onlyIf` option for R2_BINDING.put() we parse this as a strong etag regardless of the content

https://github.com/cloudflare/workerd/blob/8cf46289709352b8c8e5e16718e3231cab48689b/src/workerd/api/r2-bucket.c%2B%2B#L965-L983

As a result when the conditional check is done in the R2 layer the condition which is meant to always be true when an object exists ends up being false. 

Mostly creating this for tracking, I'll take a stab at updating the runtime to build a wildcard etag when we are passed a `*`

PR #2611, opened 2024-08-27
Fix R2 wildcard etag parsing

harshal317:
This commit addresses issue #2572

### DIFF
--- a/src/workerd/api/r2-bucket.c++
+++ b/src/workerd/api/r2-bucket.c++
@@ -935,10 +935,17 @@ kj::Array<R2Bucket::Etag> parseConditionalEtagHeader(kj::StringPtr condHeader,
   }
 }
 
-kj::Array<R2Bucket::Etag> buildSingleStrongEtagArray(kj::StringPtr etagValue) {
-  struct R2Bucket::StrongEtag etag = {.value = kj::str(etagValue)};
+kj::Array<R2Bucket::Etag> buildSingleEtagArray(kj::StringPtr etagValue) {
   kj::ArrayBuilder<R2Bucket::Etag> etagArrayBuilder = kj::heapArrayBuilder<R2Bucket::Etag>(1);
-  etagArrayBuilder.add(kj::mv(etag));
+
+  if (etagValue == "*") {
+    struct R2Bucket::WildcardEtag etag = {};
+    etagArrayBuilder.add(kj::mv(etag));
+  } else {
+    struct R2Bucket::StrongEtag etag = {.value = kj::str(etagValue)};
+    etagArrayBuilder.add(kj::mv(etag));
+  }
+
   return etagArrayBuilder.finish();
 }
 
@@ -967,12 +974,12 @@ R2Bucket::UnwrappedConditional::UnwrappedConditional(const Conditional& c)
   KJ_IF_SOME(e, c.etagMatches) {
     JSG_REQUIRE(!isQuotedEtag(e.value), TypeError,
         "Conditional ETag should not be wrapped in quotes (", e.value, ").");
-    etagMatches = buildSingleStrongEtagArray(e.value);
+    etagMatches = buildSingleEtagArray(e.value);
   }
   KJ_IF_SOME(e, c.etagDoesNotMatch) {
     JSG_REQUIRE(!isQuotedEtag(e.value), TypeError,
         "Conditional ETag should not be wrapped in quotes (", e.value, ").");
-    etagDoesNotMatch = buildSingleStrongEtagArray(e.value);
+    etagDoesNotMatch = buildSingleEtagArray(e.value);
   }
   KJ_IF_SOME(d, c.uploadedAfter) {
     uploadedAfter = d;
--- a/src/workerd/api/r2-test.js
+++ b/src/workerd/api/r2-test.js
@@ -1,0 +1,98 @@
+// Copyright (c) 2023 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import assert from 'node:assert';
+
+export default {
+  // Handler for HTTP request binding makes to R2
+  async fetch(request, env, ctx) {
+    // We only expect PUT/Get
+    assert(['GET', 'PUT'].includes(request.method));
+
+    // Each request should have a metadata size header indicating how much
+    // we should read to understand what type of request this is
+    const metadataSizeString = request.headers.get('cf-r2-metadata-size');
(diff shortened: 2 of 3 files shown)


src/workerd/api/r2-bucket.c++
```diff
@@ -935,10 +935,19 @@ kj::Array<R2Bucket::Etag> parseConditionalEtagHeader(kj::StringPtr condHeader,
   }
 }
 
-kj::Array<R2Bucket::Etag> buildSingleStrongEtagArray(kj::StringPtr etagValue) {
-  struct R2Bucket::StrongEtag etag = {.value = kj::str(etagValue)};
+kj::Array<R2Bucket::Etag> buildSingleEtagArray(kj::StringPtr etagValue) {
+  auto value = kj::str(etagValue);
```
jasnell, line 939: Creating the copy here shouldn't be necessary.

src/workerd/api/r2-test.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    // Assert it's the correct version
+    assert(jsonRequest.version = 1)
+
+    if (request.method === "PUT") {
+      assert(jsonRequest.method === "put")
+
+      if (jsonRequest.object === "onlyIfStrongEtag") {
+        assert(jsonRequest.onlyIf.etagMatches.length === 1)
```
jasnell, line 42: Since you're using `node:assert`, it is better to use `assert.strictEqual(...)` here instead of `assert(...)`

jasnell: You could also probably greatly simplify things using `assert.deepStrictEqual(...)` but that's only a guess.

src/workerd/api/r2-test.js
```diff
@@ -0,0 +1,89 @@
...
+    const metadataSize = parseInt(metadataSizeString)
+    assert(!Number.isNaN(metadataSize))
+
+    const reader = request.body.getReader({ mode: 'byob' })
+    const jsonArray = new Uint8Array(metadataSize)
+    const { value } = await reader.readAtLeast(metadataSize, jsonArray)
+    const jsonRequest = JSON.parse(new TextDecoder().decode(value))
+
+    // Currently not using the body in these test so I'm going to just discard
+    while (true) {
```
jasnell, line 26: You can simplify this loop to just...

```js
for await (const _ of request.body) {}
```

And eliminate creating the reader

src/workerd/api/r2-test.js
```diff
@@ -0,0 +1,89 @@
...
+export default {
+  // Handler for HTTP request binding makes to R2
+  async fetch(request, env, ctx) {
+    // We only expect PUT/Get
+    assert(request.method === "PUT" || request.method === "GET")
+
+    // Each request should have a metadata size header indicating how much
+    // we should read to understand what type of request this is
+    const metadataSizeString = request.headers.get("cf-r2-metadata-size")
+    assert(metadataSizeString !== null)
```
jasnell, line 16:
```js
asssert.notStrictEqual(metadataSizeString, null);
```

src/workerd/api/r2-test.js
```diff
@@ -0,0 +1,89 @@
...
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import assert from 'node:assert';
+
+export default {
+  // Handler for HTTP request binding makes to R2
+  async fetch(request, env, ctx) {
+    // We only expect PUT/Get
+    assert(request.method === "PUT" || request.method === "GET")
```
jasnell, line 11:
```js
assert(['GET', 'PUT'].includes(request.method));
```

src/workerd/api/r2-test.js
```diff
@@ -0,0 +1,89 @@
...
+
+      if (jsonRequest.object === "onlyIfWildcard") {
+        assert(jsonRequest.onlyIf.etagMatches.length === 1)
+        assert(jsonRequest.onlyIf.etagDoesNotMatch.length === 1)
+
+        assert(jsonRequest.onlyIf.etagMatches[0].type === "wildcard")
+        assert(jsonRequest.onlyIf.etagDoesNotMatch[0].type === "wildcard")
+      }
+
+      return new Response(new TextEncoder().encode(JSON.stringify({
```
jasnell, line 60: You can simplify this using...

```js
return Response.json({
  name: 'objectKey',
  // ...
});
```

src/workerd/api/r2-test.js
```diff
@@ -0,0 +1,89 @@
...
+        etagMatches: "strongEtag",
+        etagDoesNotMatch: "strongEtag"
+      }
+    })
+    await env.BUCKET.put("onlyIfWildcard", "content", {
+      onlyIf: {
+        etagMatches: "*",
+        etagDoesNotMatch: "*"
+      }
+    })
```
jasnell, line 87: What is the expectation when someone uses a weak etag? e.g. `W/"foo"`

A test verifying the expectation would help round this out.

harshal317: I don't think we need to support weak eTags here, and we hadn't been. We do support them through the HTTP headers to follow the spec but a weak eTag on etagMatches is never going to be true and we always use weak comparison on etagDoesNotMatch so regardless of strong/weak we treat it the same. 

Maybe worth updating the docs to clarify this.

src/workerd/api/r2-test.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // Assert it's the correct version
+    assert((jsonRequest.version = 1));
+
+    if (request.method === "PUT") {
+      assert(jsonRequest.method === "put");
+
+      if (jsonRequest.object === "onlyIfStrongEtag") {
+        assert(jsonRequest.onlyIf.etagMatches.length === 1);
```
anonrig, line 42: If you use `assert.strictEqual`, we will have better debugging if someone breaks this test.

src/workerd/api/r2-test.js
harshal317: That's a lot cleaner, thanks for the tip!

src/workerd/api/r2-test.js
harshal317: Changed to deepStrictEqual

jasnell (approved):

anonrig (approved):

harshal317:
Oops thought maybe clicking close would give me an option to merge.

harshal317:
@jasnell @anonrig should I get a review from anyone else? If not, could one of you help me with merging this

jasnell:
Have you run an internal CI job with this update to verify everything works internally? If so, this should be ready to land. If not, let's do that first.

PR #2614, opened 2024-08-27
Cancel previous builds within the same PR on push

npaun:
What this does:
- Makes the test workflow responsible for triggering fixup, labels and lint. 
- Cancels the entire workflow if another commit is pushed to the same PR
- Allows all other builds to run through.

This should cut down on the number of runners in use, and lead to less wait time for builds.

### DIFF
--- a/.github/workflows/fixup.yml
+++ b/.github/workflows/fixup.yml
@@ -1,7 +1,6 @@
 name: "Just say no to fixup commits"
 on:
-  pull_request:
-    types: [synchronize, opened, reopened]
+  workflow_call:
 
 jobs:
   block-fixup:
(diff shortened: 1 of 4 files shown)


jasnell (approved):

danlapid:
I like the idea but we need to make sure we aren't reducing our concurrency at all
I don't want a full CI run to be synchronous now. 
I think we just want it to be auto cancelled
You don't need the workflow for that, simply set the group to include ${{ github.workflow }} and they'll be fine.
Other than that we still need to make sure matrix jobs have different groups from each other
That might need to happen at the <job>.concurrency.group level though.

npaun:
This won't reduce our concurrency; the jobs still run in parallel. The advantage of the reusable workflow, where I was able to do to that, is just that I don't need to copy paste the concurrency group name into each workflow.

danlapid (approved):

npaun:
Screenshot 2024-08-27 at 12 25 35 PM
Just to be sure I grabbed this screenshot of a bunch of jobs running at once.

PR #2616, opened 2024-08-27
zlib one-shot methods

npaun:
This is the same as previously approved PR #2533, except that some compressed test strings are modified to match the output generated by V8's zlib. This way we won't have any more inconsistencies between workerd and the internal edgeworker repo

### DIFF
--- a/src/node/internal/zlib.d.ts
+++ b/src/node/internal/zlib.d.ts
@@ -2,6 +2,24 @@ import { owner_symbol, type Zlib } from 'node-internal:internal_zlib_base';
 
 export function crc32(data: ArrayBufferView, value: number): number;
 
+export type CompressCallback<ErrT, BufT> = (
+  error?: ErrT,
+  result?: BufT
+) => void;
+
+export function crc32(data: ArrayBufferView | string, value: number): number;
+export function zlibSync(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  mode: number
+): ArrayBuffer;
+export function zlib(
+  data: ArrayBufferView | string,
+  options: ZlibOptions,
+  mode: number,
+  cb: CompressCallback<string, ArrayBuffer>
+): ArrayBuffer;
+
 // zlib.constants (part of the API contract for node:zlib)
 export const CONST_Z_NO_FLUSH: number;
 export const CONST_Z_PARTIAL_FLUSH: number;
--- a/src/node/internal/internal_zlib.ts
+++ b/src/node/internal/internal_zlib.ts
@@ -3,11 +3,14 @@
 //     https://opensource.org/licenses/Apache-2.0
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 
-import { default as zlibUtil, type ZlibOptions } from 'node-internal:zlib';
+import {
+  default as zlibUtil,
+  type ZlibOptions,
+  type CompressCallback,
+} from 'node-internal:zlib';
 import { Buffer } from 'node-internal:internal_buffer';
 import { validateUint32 } from 'node-internal:validators';
 import { ERR_INVALID_ARG_TYPE } from 'node-internal:internal_errors';
-import { isArrayBufferView } from 'node-internal:internal_types';
 import { Zlib } from 'node-internal:internal_zlib_base';
 
 const {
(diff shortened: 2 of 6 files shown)


src/node/internal/zlib.d.ts
```diff
@@ -2,6 +2,24 @@ import { owner_symbol, type Zlib } from 'node-internal:internal_zlib_base';
 
 export function crc32(data: ArrayBufferView, value: number): number;
 
+export type CompressCallback<ErrT, BufT> = (
```
anonrig, line 5: This needs to be (err, undefined) | (null, val)

npaun: Why is `| undefined` better than `?` in this context?

anonrig: So, this type allows the following callback output: `(new Error(), Buffer.from('aa'))` but in reality, we can't have an error and a buffer at the same time. Therefore you need to fix the type to something like this

```
export type CompressCallback<T = ArrayBufferView> = ((err: Error, val: undefined) => void | ((err: null, val: T) => void)
```

npaun: I don't think that type is right either. It would allow the callback signature to be `callback(err: undefined, val: Buffer)`, i.e. a callback that can only handle success.

anonrig: so, this is for our use case. this type is not exposed to outside, so null vs. undefined does not really make a difference for us.

npaun: Sorry let's forget about null or undefined. 

My definition:
type X = (T | null, U | null) -> void

Type X wrongly allows (null, null) as arguments. That's bad. 

Your definition 
type Y = (T, nil) -> void | (nil, U) -> void 

Type Y allows me to implement this callback. 
f(a: nil, b: U) -> void 

This callback will never be able to handle errors properly. Yet it's permitted by type Y

anonrig: > This callback will never be able to handle errors properly. Yet it's permitted by type Y

I don't follow?

npaun: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAwg9gWzAJwgZzTAhgGxwIywGMBrKAXigAoqJlk5kAuKAUXsYBopU0BXHMBZ8AdgBMIAMwCWIiGICUFAHxQAbnGmKoAH2q0OzKKIky5Y7rwFCoaYMlkBzJeVUatCgNwAob5NFEwNJwIlCyGiQQVES4BMQkLPBIvJixhKRKAN7eUFAA9Hm2iND4cGIgUAAWdNDeAL6+-iKBwaESCHDYeOkkBgxGJlKy8pbo1ix2DiLOUNm5BVAAKpXSaFBNLSFha3CSUKCQsIgo6Knd8VD4fMBhNzEiAOQ3lVjiONB0-WtYNwCCeHAAO5oHJQIghNBwd4AOhwcEcVCsgi89V8C2WqygKHhyCwCCqWDWIjg+3A0FW-HQ3nCcEiVHanTS8S8QA

npaun: Or on the flip side, if I try to work with that type, Typescript tells me both args resolve to "never":
https://www.typescriptlang.org/play/?#code/C4TwDgpgBAwg9gWzAJwgZzTAhgGxwIywGMBrKAXigAoqJlk5kAuKAUXsYBopU0BXHMBZ8AdgBMIAMwCWIiGICUFAHxQAbnGmKoAH2q0OzKKIky5Y7rwFCoaYMlkBzJeVUatCgNwAob5NFEwNJwIlCyGiQQVES4BMQkLPBIvJixhKRKAN7eUFAA9HlQACrIIE5QwHDGaNDAABbSaBXg0AhYkU2gkGhEDmDAFQ0iZBDEdVBYyI58CBAiA7xwOGroFVVyK8g5YZLUALJY9QB0yFjiiFRKqgAMRwCsWdu5MXjpJFRyAO5shlQARHBxvg4CA-gpuCYpLJ5F5tgBfKAQHA1KDZXLPNLxKiQszybh-ACSA0+jBIaAAhGCfLk4d5aX4AkEQlAJAg4NhXli6AwjDjoRYeOhrCw7A4RM5UdsCsUGk1-CJAsFQo0oHBdl1oEkUOhUpzSFB8HwBtIBjERAByAZ1M5iHDQbmMJqHCZ4bZEEJoJYQI44OCOKhWQSw+nhOCRKis9mYjKeIA

anonrig: Then I guess you need something like:

```
type CompressCallback = ((error: Error, result: undefined) => void) & ((error: undefined, result: string) => void);

function invoke(callback: CompressCallback) {
  // some body here 
}

function demoCallback(error: undefined, result: string | undefined) {
  // This function is of type CompressCallback but it can't handle errors at Allows
  console.log(result);
}

// This program has no type issues
invoke(demoCallback);
```

But I personally prefer something like this:


```
type CompressCallback = ((error: Error, result: undefined) => void) & ((error: undefined, result: string) => void);

function invoke(callback: CompressCallback) {
  // some body here 
}

function demoCallback(error: Error, result: undefined) : void;
function demoCallback(error: undefined, result: string) : void;
function demoCallback(error: Error | undefined, result: string | undefined) {
  // This function is of type CompressCallback but it can't handle errors at Allows
  console.log(result);
}

// This program has no type issues
invoke(demoCallback);
```

npaun: Here's the fun bit... using the `&` type allows both an error and a result, just like my original type:
https://www.typescriptlang.org/play/?#code/C4TwDgpgBAwg9gWzAJwgZzTAhgGxwIywGMBrKAXigAoqJlk5kAuKAUXsYBopU0BXHMBZ8AdgBMIAMwCWIiGICUFAHxQAbnGmKoAMmq0OzKKIky5Y7rwFCoaYMlkBzJeVUatCgNwAob5NFEwNJwIlCyGiQQVES4BMQkLPBIvJixhKRKAN7eUFAA9Hm2iND4cGIgUAAWdNDeAL6+-iKBwaESCHDYeOkkBgxG7P1QAD7G4lKy8pbo1ix2DiKOI2Omk9rZuQVQACqV0mhQTS0hYQdwklCgkLCIKOip3fFQ+HzAYW8xIgDkb5VY4jhoHR+gcsG8AIJ4OAAdzQOSgRBCaDggIAdDg4I4qFZBF56r4trt9lAUJjkFgEFUsAcRHBLuBoPt+OhvOE4JEqO1Oml4l4gA

anonrig (approved):

PR #2618, opened 2024-08-27
Add packages to ArtifactBundler

garrettgu10:
Related to downstream PR 8540

### DIFF
--- a/src/pyodide/internal/loadPackage.ts
+++ b/src/pyodide/internal/loadPackage.ts
@@ -70,11 +70,13 @@ async function loadBundleFromArtifactBundler(
   const packagesVersion = PACKAGES_VERSION;
   const filename = LOCKFILE['packages'][requirement]['file_name'];
   const fullPath = 'python-package-bucket/' + packagesVersion + '/' + filename;
-  return new Promise((resolve) => {
-    ArtifactBundler.onPackageReceived(fullPath, (r: Reader) =>
-      resolve([requirement, r])
+  const reader = ArtifactBundler.getPackage(fullPath);
+  if (!reader)
+    throw new Error(
+      'Failed to get package ' + fullPath + ' from ArtifactBundler'
     );
-  });
+  return Promise.resolve([requirement, reader]);
+  // ^ this is okay to do during startup since it resolves immediately
 }
 
 /**
--- a/src/pyodide/internal/loadPackage.ts
+++ b/src/pyodide/internal/loadPackage.ts
@@ -9,16 +9,21 @@
  * that contains all the packages ready to go.
  */
 
-import { default as LOCKFILE } from 'pyodide-internal:generated/pyodide-lock.json';
-import { WORKERD_INDEX_URL } from 'pyodide-internal:metadata';
+import {
+  WORKERD_INDEX_URL,
+  LOCKFILE,
+  LOAD_WHEELS_FROM_R2,
+  LOAD_WHEELS_FROM_ARTIFACT_BUNDLER,
+  PACKAGES_VERSION,
+} from 'pyodide-internal:metadata';
 import {
   SITE_PACKAGES,
-  LOAD_WHEELS_FROM_R2,
   getSitePackagesPath,
 } from 'pyodide-internal:setupPackages';
 import { parseTarInfo } from 'pyodide-internal:tar';
 import { default as DiskCache } from 'pyodide-internal:disk_cache';
 import { createTarFS } from 'pyodide-internal:tarfs';
+import { default as ArtifactBundler } from 'pyodide-internal:artifacts';
 
 async function decompressArrayBuffer(
   arrBuf: ArrayBuffer
@@ -33,13 +38,18 @@ async function decompressArrayBuffer(
   }
 }
 
-async function loadBundle(requirement: string): Promise<[string, ArrayBuffer]> {
+// loadBundleFromR2 loads the package from the internet (through fetch) and uses the DiskCache as
+// a backing store. This is only used in local dev.
+async function loadBundleFromR2(
+  requirement: string
+): Promise<[string, Reader]> {
   // first check if the disk cache has what we want
   const filename = LOCKFILE['packages'][requirement]['file_name'];
   const cached = DiskCache.get(filename);
   if (cached) {
     const decompressed = await decompressArrayBuffer(cached);
-    return [requirement, decompressed];
+    const reader = new ArrayBufferReader(decompressed);
+    return [requirement, reader];
   }
 
   // we didn't find it in the disk cache, continue with original fetch
@@ -50,7 +60,21 @@ async function loadBundle(requirement: string): Promise<[string, ArrayBuffer]> {
   const decompressed = await decompressArrayBuffer(compressed);
 
   DiskCache.put(filename, compressed);
-  return [requirement, decompressed];
+  const reader = new ArrayBufferReader(decompressed);
+  return [requirement, reader];
+}
+
+async function loadBundleFromArtifactBundler(
+  requirement: string
+): Promise<[string, Reader]> {
+  const packagesVersion = PACKAGES_VERSION;
+  const filename = LOCKFILE['packages'][requirement]['file_name'];
+  const fullPath = 'python-package-bucket/' + packagesVersion + '/' + filename;
+  return new Promise((resolve) => {
+    ArtifactBundler.onPackageReceived(fullPath, (r: Reader) =>
+      resolve([requirement, r])
+    );
+  });
 }
 
 /**
@@ -73,9 +97,11 @@ class ArrayBufferReader {
   }
 }
 
-export async function loadPackages(Module: Module, requirements: Set<string>) {
-  if (!LOAD_WHEELS_FROM_R2) return;
-
+async function loadPackagesImpl(
+  Module: Module,
+  requirements: Set<string>,
+  loadBundle: (req: string) => Promise<[string, Reader]>
+) {
   let loadPromises = [];
   let loading = [];
   for (const req of requirements) {
@@ -87,8 +113,7 @@ export async function loadPackages(Module: Module, requirements: Set<string>) {
   console.log('Loading ' + loading.join(', '));
 
   const buffers = await Promise.all(loadPromises);
-  for (const [requirement, buffer] of buffers) {
-    const reader = new ArrayBufferReader(buffer);
+  for (const [requirement, reader] of buffers) {
     const [tarInfo, soFiles] = parseTarInfo(reader);
     SITE_PACKAGES.addSmallBundle(tarInfo, soFiles, requirement);
   }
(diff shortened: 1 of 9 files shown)


src/workerd/api/pyodide/pyodide.c++
```diff
@@ -27,6 +27,19 @@ void PyodideBundleManager::setPyodideBundleData(
       kj::mv(version), {.messageReader = kj::mv(messageReader), .bundle = bundle});
 }
 
+const kj::Maybe<kj::ArrayPtr<const unsigned char>> PyodidePackageManager::getPyodidePackage(
+    kj::StringPtr id) const {
+  KJ_IF_SOME(t, packages.lockShared()->find(id)) {
```
anonrig, line 32: Nit: This function can be simplified with just returning and calling kj::maybe's .map() to convert it into asPtr().

anonrig (approved):

src/workerd/api/pyodide/pyodide.h
```diff
@@ -166,25 +167,61 @@ class PyodideMetadataReader: public jsg::Object {
...
+  // stored on the V8 heap. Ideally, we'd like to simply store a map from package paths to Promises.
+  // Instead, we store a CrossThreadWaitList and get a new Promise from the waitlist as needed.
+  kj::HashMap<kj::String, kj::Own<SmallPackagesTarReader>> fetchedPackages;
+  kj::HashMap<kj::String, CrossThreadWaitList> waitlists;
+
+public:
+  void insert(kj::String path, kj::Promise<kj::Own<SmallPackagesTarReader>> promise);
+
+  // Note it is explicitly not allowed to call getPromise more than once on the same package
+  // This is because allowing this causes ownership of the SmallPackagesTarReader to be muddied.
```
dom96, line 204: What happens when we do this accidentally? Will we get a nice error? If not can you add some code to ensure we do?

dom96 (approved):

src/pyodide/internal/loadPackage.ts
```diff
@@ -73,9 +99,11 @@ class ArrayBufferReader {
   }
 }
 
-export async function loadPackages(Module: Module, requirements: Set<string>) {
-  if (!LOAD_WHEELS_FROM_R2) return;
-
+async function loadPackagesImpl(
+  Module: Module,
+  requirements: Set<string>,
+  loadBundle: (req: string) => Promise<[string, Reader]>
```
hoodmane, line 105: Nit: Making it `loadBundle`'s responsibility to include the bundle name in the result isn't the best for readability since we have to go read those implementations to see they never change it. Maybe just add a helper here?
```ts
async function load(req) {
   const bundle = await loadBundle(req);
   return [req, bundle];
}
```

src/pyodide/internal/loadPackage.ts
```diff
@@ -50,7 +60,23 @@ async function loadBundle(requirement: string): Promise<[string, ArrayBuffer]> {
...
+  requirement: string
+): Promise<[string, Reader]> {
+  const packagesVersion = PACKAGES_VERSION;
+  const filename = LOCKFILE['packages'][requirement]['file_name'];
+  const fullPath = 'python-package-bucket/' + packagesVersion + '/' + filename;
+  const reader = ArtifactBundler.getPackage(fullPath);
+  if (!reader)
+    throw new Error(
+      'Failed to get package ' + fullPath + ' from ArtifactBundler'
+    );
```
hoodmane, line 77: I think it's better for readability to include braces.

src/pyodide/internal/loadPackage.ts
```diff
@@ -50,7 +60,23 @@ async function loadBundle(requirement: string): Promise<[string, ArrayBuffer]> {
...
+  return Promise.resolve([requirement, reader]);
+  // ^ this is okay to do during startup since it resolves immediately
```
hoodmane, line 79: This is weird, presumably it would work to directly return the value right?
```suggestion
  return [requirement, reader];
```
(We don't even necessarily need to make the function async, if you do `await notThenable` it just returns `notThenable`. But this is perhaps not ideal for readability.)

src/pyodide/internal/setupPackages.ts
```diff
@@ -122,22 +126,18 @@ class SitePackagesDir {
...
-  requirements: Set<string>
-): [SitePackagesDir, boolean] {
+export function buildSitePackages(requirements: Set<string>): SitePackagesDir {
   const [bigTarInfo, bigTarSoFiles] = parseTarInfo();
 
-  let LOAD_WHEELS_FROM_R2 = true;
   let requirementsInBigBundle = new Set([...STDLIB_PACKAGES]);
-  if (bigTarInfo.children!.size > 10) {
-    LOAD_WHEELS_FROM_R2 = false;
+  if (!LOAD_WHEELS_FROM_R2 && !LOAD_WHEELS_FROM_ARTIFACT_BUNDLER) {
```
hoodmane, line 133: Can you add a comment explaining this condition? It's not immediately clear what you are up to here. If we are using the packages built into the binary, then `requirementsInBigBundle` is the set of stdlib packages? Otherwise, it includes all requirements? So I guess the point is that the big bundle either has:
1. everything (currently) or
2. just standard library packages (when your autogate is on)?

If this isn't already explained in a comment somewhere, please explain it.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -188,6 +188,7 @@ try {
       if (typeof mainModule[pyHandlerName] === 'function') {
         handlers[handlerName] = makeHandler(pyHandlerName);
       }
+      handlers[handlerName] = makeHandler(pyHandlerName);
```
hoodmane, line 191: This doesn't look right to me.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -11,10 +11,8 @@ namespace workerd::api::pyodide {
 
 const kj::Maybe<jsg::Bundle::Reader> PyodideBundleManager::getPyodideBundle(
     kj::StringPtr version) const {
-  KJ_IF_SOME(t, bundles.lockShared()->find(version)) {
-    return t.bundle;
-  }
-  return kj::none;
+  return bundles.lockShared()->find(version).map(
+      [](const MessageBundlePair& t) { return t.bundle; });
```
hoodmane, line 15: It's unfortunate that C++ can't infer the type of the lambda argument...

src/workerd/api/pyodide/pyodide.h
```diff
@@ -176,15 +203,20 @@ struct MemorySnapshotResult {
 // CPU architecture-specific artifacts. The logic for loading these is in getArtifacts.
 class ArtifactBundler: public jsg::Object {
 public:
+  kj::Maybe<const PyodidePackageManager&> packageManager;
+  // ^ lifetime should be static since there is normally one worker set for the whole process. see worker-set.h
```
hoodmane, line 207: Or in more detail the ownership structure goes:
`WorkerSet` lifetime = `PackageManager` lifetime and `Worker` lifetime = `ArtifactBundler` lifetime and `WorkerSet` owns and will outlive `Worker`, so `PackageManager` outlives `ArtifactBundler`

src/workerd/api/pyodide/pyodide.h
```diff
@@ -229,6 +262,16 @@ class ArtifactBundler: public jsg::Object {
...
   }
 
+  kj::Maybe<jsg::Ref<PackagesTarReader>> getPackage(kj::String path) {
+    KJ_IF_SOME(pacman, packageManager) {
+      KJ_IF_SOME(ptr, pacman.getPyodidePackage(path)) {
+        return jsg::alloc<PackagesTarReader>(ptr);
+      }
+    }
+
+    return kj::none;
```
hoodmane, line 272: Is it a logic error if we reach this point?

garrettgu10: Yes, it is a logic error, but it ends up getting surfaced by the JavaScript caller instead of the C++ code. I think it's more flexible for the API to return a nullable value here and make it the caller's responsibility to check. There's also not a huge risk for missing a nullability check since the TypeScript type matches.

src/pyodide/python-entrypoint-helper.ts
garrettgu10: oops, merge error

src/pyodide/internal/loadPackage.ts
garrettgu10: Yeah, you're right, this API is a bit clunky. I ended up just doing this in a then() lower down

src/pyodide/internal/setupPackages.ts
garrettgu10: This is meant as a transitionary thing, to support the current model as we're rolling it out. Eventually we are going to make the big bundle completely empty.

src/pyodide/internal/loadPackage.ts
```diff
@@ -50,7 +58,23 @@ async function loadBundle(requirement: string): Promise<[string, ArrayBuffer]> {
...
+  const reader = new ArrayBufferReader(decompressed);
+  return reader;
+}
+
+async function loadBundleFromArtifactBundler(
+  requirement: string
+): Promise<Reader> {
+  const packagesVersion = PACKAGES_VERSION;
+  const filename = LOCKFILE['packages'][requirement]['file_name'];
+  const fullPath = 'python-package-bucket/' + packagesVersion + '/' + filename;
```
dom96, line 70: Nit

```suggestion
  const fullPath = 'python-package-bucket/' + PACKAGES_VERSION + '/' + filename;
```

src/pyodide/internal/loadPackage.ts
```diff
@@ -50,7 +58,23 @@ async function loadBundle(requirement: string): Promise<[string, ArrayBuffer]> {
...
-  return [requirement, decompressed];
+  const reader = new ArrayBufferReader(decompressed);
+  return reader;
+}
+
+async function loadBundleFromArtifactBundler(
+  requirement: string
+): Promise<Reader> {
+  const packagesVersion = PACKAGES_VERSION;
+  const filename = LOCKFILE['packages'][requirement]['file_name'];
```
dom96, line 69: Any chance that `requirement` might not be present in the lockfile?

src/workerd/api/pyodide/pyodide.h
```diff
@@ -176,15 +203,22 @@ struct MemorySnapshotResult {
 // CPU architecture-specific artifacts. The logic for loading these is in getArtifacts.
 class ArtifactBundler: public jsg::Object {
 public:
+  kj::Maybe<const PyodidePackageManager&> packageManager;
```
dom96, line 206: Why not create this in ArtifactBundler or pass in a `kj::Own<PyodidePackageManager>`?

dom96 (approved):
LGTM with a few questions/nits.

src/workerd/api/pyodide/pyodide.h
garrettgu10: This is shared between all workers of a WorkerSet (we want it to be shared since it has ownership over the in-memory cache for packages), so owning an instance doesn't make sense here IMO. I believe there's a comment immediately below explaining the ownership and lifetime of this reference.

src/pyodide/internal/setupPackages.ts
hoodmane: Well it does kind of make sense to revendor the unvendored stdlib modules.

PR #2619, opened 2024-08-27
implement brotli classes

anonrig:
Implements:
- `new BrotliCompress()`
- `new BrotliDecompress()`
- `createBrotliDecompress()`
- `createBrotliCompress()`
functions and classes to `node:zlib`

### DIFF
--- a/src/workerd/api/node/zlib-util.c++
+++ b/src/workerd/api/node/zlib-util.c++
@@ -4,18 +4,24 @@
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 
 #include "zlib-util.h"
-#include "workerd/jsg/exception.h"
 
+#include "nbytes.h"
+
+// The following implementation is adapted from Node.js
+// and therefore follows Node.js style as opposed to kj style.
+// Latest implementation of Node.js zlib can be found at:
+// https://github.com/nodejs/node/blob/main/src/node_zlib.cc
 namespace workerd::api::node {
+
 kj::ArrayPtr<kj::byte> ZlibUtil::getInputFromSource(InputSource& data) {
   KJ_SWITCH_ONEOF(data) {
     KJ_CASE_ONEOF(dataBuf, kj::Array<kj::byte>) {
-      JSG_REQUIRE(dataBuf.size() < Z_MAX_CHUNK, RangeError, "Memory limit exceeded");
+      JSG_REQUIRE(dataBuf.size() < Z_MAX_CHUNK, RangeError, "Memory limit exceeded"_kj);
       return dataBuf.asPtr();
     }
 
     KJ_CASE_ONEOF(dataStr, jsg::NonCoercible<kj::String>) {
-      JSG_REQUIRE(dataStr.value.size() < Z_MAX_CHUNK, RangeError, "Memory limit exceeded");
+      JSG_REQUIRE(dataStr.value.size() < Z_MAX_CHUNK, RangeError, "Memory limit exceeded"_kj);
       return dataStr.value.asBytes();
     }
   }
@@ -40,27 +46,27 @@ public:
     maxCapacity = _maxCapacity;
   }
 
-  inline size_t size() const {
+  size_t size() const {
     return builder.size();
   }
-  inline bool empty() const {
+  bool empty() const {
     return size() == 0;
   }
-  inline size_t capacity() const {
+  size_t capacity() const {
     return builder.capacity();
   }
-  inline size_t available() const {
+  size_t available() const {
     return capacity() - size();
   }
 
-  inline kj::byte* begin() KJ_LIFETIMEBOUND {
+  kj::byte* begin() KJ_LIFETIMEBOUND {
     return builder.begin();
   }
-  inline kj::byte* end() KJ_LIFETIMEBOUND {
+  kj::byte* end() KJ_LIFETIMEBOUND {
     return builder.end();
   }
 
-  inline kj::Array<kj::byte> releaseAsArray() {
+  kj::Array<kj::byte> releaseAsArray() {
     // TODO(perf):  Avoid a copy/move by allowing Array<T> to point to incomplete space?
     if (!builder.isFull()) {
       setCapacity(size());
@@ -68,20 +74,20 @@ public:
     return builder.finish();
   }
 
-  inline void adjustUnused(size_t unused) {
+  void adjustUnused(size_t unused) {
     resize(capacity() - unused);
   }
 
-  inline void resize(size_t size) {
+  void resize(size_t size) {
     if (size > builder.capacity()) grow(size);
     builder.resize(size);
   }
 
-  inline void addChunk() {
+  void addChunk() {
     reserve(size() + chunkSize);
   }
 
-  inline void reserve(size_t size) {
+  void reserve(size_t size) {
     if (size > builder.capacity()) {
       grow(size);
     }
@@ -438,18 +444,20 @@ void ZlibContext::setOutputBuffer(kj::ArrayPtr<kj::byte> output) {
   stream.avail_out = output.size();
 }
 
-jsg::Ref<ZlibUtil::ZlibStream> ZlibUtil::ZlibStream::constructor(ZlibModeValue mode) {
-  return jsg::alloc<ZlibStream>(static_cast<ZlibMode>(mode));
+template <typename CompressionContext>
+jsg::Ref<ZlibUtil::CompressionStream<CompressionContext>> ZlibUtil::CompressionStream<
+    CompressionContext>::constructor(ZlibModeValue mode) {
+  return jsg::alloc<CompressionStream>(static_cast<ZlibMode>(mode));
 }
 
 template <typename CompressionContext>
-CompressionStream<CompressionContext>::~CompressionStream() noexcept(false) {
+ZlibUtil::CompressionStream<CompressionContext>::~CompressionStream() {
   JSG_ASSERT(!writing, Error, "Writing to compression stream"_kj);
   close();
 }
 
 template <typename CompressionContext>
-void CompressionStream<CompressionContext>::emitError(
+void ZlibUtil::CompressionStream<CompressionContext>::emitError(
     jsg::Lock& js, const CompressionError& error) {
   KJ_IF_SOME(onError, errorHandler) {
     onError(js, error.err, kj::mv(error.code), kj::mv(error.message));
@@ -463,7 +471,7 @@ void CompressionStream<CompressionContext>::emitError(
 
 template <typename CompressionContext>
 template <bool async>
-void CompressionStream<CompressionContext>::writeStream(jsg::Lock& js,
+void ZlibUtil::CompressionStream<CompressionContext>::writeStream(jsg::Lock& js,
     int flush,
     kj::ArrayPtr<kj::byte> input,
     uint32_t inputLength,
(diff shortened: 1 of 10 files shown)


npaun:
GH won't let me comment on the line, but turn on the brotli tests in `zlibEmptyBuffer`

src/workerd/api/node/zlib-util.h
```diff
@@ -91,6 +90,9 @@ struct CompressionError {
 
 class ZlibContext {
 public:
+  explicit ZlibContext(ZlibMode _mode) {
+    mode = _mode;
+  }
```
npaun, line 95:
```suggestion
  explicit ZlibContext(ZlibMode _mode): mode(_mode) {}
```

src/workerd/api/node/zlib-util.h
```diff
@@ -173,42 +175,74 @@ class ZlibContext {
...
-  void close();
-  bool checkError(jsg::Lock& js);
-  void emitError(jsg::Lock& js, const CompressionError& error);
-  template <bool async>
-  void writeStream(jsg::Lock& js,
-      int flush,
-      kj::ArrayPtr<kj::byte> input,
+  explicit BrotliContext(ZlibMode _mode) {
+    mode = _mode;
+  }
```
npaun, line 182:
```suggestion
  explicit BrotliContext(ZlibMode _mode): mode(_mode) {}
```

src/workerd/api/node/zlib-util.c++
```diff
@@ -438,18 +440,25 @@ void ZlibContext::setOutputBuffer(kj::ArrayPtr<kj::byte> output) {
   stream.avail_out = output.size();
 }
 
-jsg::Ref<ZlibUtil::ZlibStream> ZlibUtil::ZlibStream::constructor(ZlibModeValue mode) {
-  return jsg::alloc<ZlibStream>(static_cast<ZlibMode>(mode));
+template <class CompressionContext>
```
npaun, line 443:
```suggestion
template <typename CompressionContext>
```

There's no difference between `typename` and `class` in this context but we should be consistent throughout.

src/workerd/api/node/zlib-util.c++
```diff
@@ -438,18 +440,25 @@ void ZlibContext::setOutputBuffer(kj::ArrayPtr<kj::byte> output) {
...
+jsg::Ref<ZlibUtil::CompressionStream<CompressionContext>> ZlibUtil::CompressionStream<
+    CompressionContext>::constructor(ZlibModeValue mode) {
+  return jsg::alloc<CompressionStream>(static_cast<ZlibMode>(mode));
 }
 
-template <typename CompressionContext>
-CompressionStream<CompressionContext>::~CompressionStream() noexcept(false) {
+template <class CompressionContext>
```
npaun, line 449:
```suggestion
template <typename CompressionContext>
```

anonrig:
> GH won't let me comment on the line, but turn on the brotli tests in `zlibEmptyBuffer`

@npaun Those are the one-shot methods that this pull-request doesn't cover.

src/workerd/api/node/zlib-util.h
```diff
@@ -250,38 +286,111 @@ class ZlibUtil final: public jsg::Object {
...
+    JSG_RESOURCE_TYPE(CompressionStream) {
+      JSG_METHOD(close);
+      JSG_METHOD_NAMED(write, template write<true>);
+      JSG_METHOD_NAMED(writeSync, template write<false>);
+      JSG_METHOD(reset);
+      JSG_METHOD(setErrorHandler);
+    }
+
+  protected:
+    CompressionContext* context() {
```
npaun, line 300: Is this a pointer because node? I'd rather use a reference

npaun:
@anonrig  Right, that's my job then hehe

src/workerd/api/node/zlib-util.h
anonrig: This follows Node.js implementation but this is actually needed in `BrotliCompressionStream<CompressionContext>::initialize` function.

The use case is `this->context()->initialize` where if we define this context as a protected variable and expose it, the compiler will complain, since both BrotliCompressionStream and CompressionStream has context variables. (PS: BrotliCompressionStream extends CompressionStream)

src/workerd/api/node/zlib-util.h
```diff
@@ -250,38 +282,111 @@ class ZlibUtil final: public jsg::Object {
...
-      JSG_METHOD(close);
-      JSG_METHOD(write);
-      JSG_METHOD(writeSync);
       JSG_METHOD(params);
-      JSG_METHOD(setErrorHandler);
-      JSG_METHOD(reset);
+    }
+
+    CompressionContext* context() {
+      return this->CompressionStream<CompressionContext>::context();
```
npaun, line 377: Not sure what this method does because it should be inherited.

anonrig: it returns the context of the inherited class. This won't work because the usage of excessive templates makes the compilers work extremely hard when used with `this->context.myFunction()`

src/workerd/api/node/zlib-util.h
npaun: I don't quite get why there would be two of them. I see only one variable named context_

anonrig: I'll open a PR after this pull-request lands, to show the build error. For now, if this is not blocking, let's proceed with the PR. wdyt @npaun?

npaun: Not blocking. I'll start tagging my comments with _nit:_ for clarity.

npaun:
Just flagging that there's a memory issue in the linux-asan build

src/workerd/api/node/zlib-util.h
```diff
@@ -217,31 +247,33 @@ class ZlibUtil final: public jsg::Object {
...
-  class ZlibStream final: public jsg::Object, public CompressionStream<ZlibContext> {
+  template <class CompressionContext>
+  class CompressionStream: public jsg::Object {
   public:
-    ZlibStream(ZlibMode mode);
-    KJ_DISALLOW_COPY_AND_MOVE(ZlibStream);
-    static jsg::Ref<ZlibStream> constructor(ZlibModeValue mode);
+    explicit CompressionStream(ZlibMode _mode): context_(_mode) {}
+    CompressionStream() = default;
+    ~CompressionStream();
```
jasnell, line 255:
```suggestion
    ~CompressionStream() noexcept(false);
```

anonrig: Not possible:

```
src/workerd/api/node/zlib-util.h:260:5: error: exception specification is not available until end of class definition
  260 |     ~CompressionStream() noexcept(false);
```

npaun (approved):
LGTM but I'd wait for James || Felix to approve as well before merging because it's a bigger change

src/workerd/api/node/zlib-util.h
```diff
@@ -173,42 +178,72 @@ class ZlibContext {
...
 protected:
-  CompressionContext context;
+  ZlibMode mode;
+  const uint8_t* nextIn = nullptr;
+  uint8_t* nextOut = nullptr;
+  size_t availIn = 0;
+  size_t availOut = 0;
+  BrotliEncoderOperation flush = BROTLI_OPERATION_PROCESS;
+
+  // TODO(addaleax): These should not need to be stored here.
```
jasnell, line 203: Probably a good idea to add some extra comments at the top of this file pointing out that a lot of this was adapted from Node.js and therefore follows a lot of Node.js style as opposed to kj style. Some folks coming into this code later on might not have any idea who or what an "addaleax" is ;-)

anonrig: Agreed. Just added.

src/workerd/api/node/zlib-util.c++
```diff
@@ -438,18 +444,20 @@ void ZlibContext::setOutputBuffer(kj::ArrayPtr<kj::byte> output) {
...
-  return jsg::alloc<ZlibStream>(static_cast<ZlibMode>(mode));
+template <typename CompressionContext>
+jsg::Ref<ZlibUtil::CompressionStream<CompressionContext>> ZlibUtil::CompressionStream<
+    CompressionContext>::constructor(ZlibModeValue mode) {
+  return jsg::alloc<CompressionStream>(static_cast<ZlibMode>(mode));
 }
 
 template <typename CompressionContext>
-CompressionStream<CompressionContext>::~CompressionStream() noexcept(false) {
+ZlibUtil::CompressionStream<CompressionContext>::~CompressionStream() {
```
jasnell, line 454: Why did you take the `noexcept(false)` off this signature? It should be retained.

src/workerd/api/node/zlib-util.c++
```diff
@@ -476,11 +484,11 @@ void CompressionStream<CompressionContext>::writeStream(jsg::Lock& js,
 
   writing = true;
 
-  context.setBuffers(input, inputLength, output, outputLength);
-  context.setFlush(flush);
+  context()->setBuffers(input, inputLength, output, outputLength);
+  context()->setFlush(flush);
```
jasnell, line 488: Nit: Can `context()` return the reference instead of a pointer? ...`context().setBuffer(...)` just looks cleaner

jasnell (approved):

src/workerd/api/node/zlib-util.c++
anonrig: It's producing the following error that requires some hacky solution to get around it:

```
src/workerd/api/node/zlib-util.h:260:5: error: exception specification is not available until end of class definition
  260 |     ~CompressionStream() noexcept(false);
```

src/workerd/api/node/zlib-util.c++
anonrig: I'll follow up with a new pull-request for this.

src/workerd/api/node/zlib-util.h
```diff
@@ -250,38 +291,113 @@ class ZlibUtil final: public jsg::Object {
...
+    JSG_RESOURCE_TYPE(CompressionStream) {
+      JSG_METHOD(close);
+      JSG_METHOD_NAMED(write, template write<true>);
+      JSG_METHOD_NAMED(writeSync, template write<false>);
+      JSG_METHOD(reset);
+      JSG_METHOD(setErrorHandler);
+    }
+
+  protected:
+    CompressionContext* context() {
```
jasnell, line 305: Our typical style would be to return a `CompressionContext&` here instead... and if you do want/need to provide the pointer, provide a separate `CompressionContext* get()` or something.

src/workerd/api/node/zlib-util.h
```diff
@@ -250,38 +291,113 @@ class ZlibUtil final: public jsg::Object {
...
+    // the allocated memory chunk just before the "payload" memory we return
+    // to zlib.
+    static void* AllocForZlib(void* data, uInt items, uInt size);
+    static void* AllocForBrotli(void* data, size_t size);
+    static void FreeForZlib(void* data, void* pointer);
+
+  private:
+    // Used to store allocations in Brotli* operations.
+    // This declaration should be physically positioned before
+    // context to avoid `heap-use-after-free` ASan error.
```
jasnell, line 321: non-blocking nit: I would expand the comment slight to connect this more concretely with the `AllocFor_` and `FreeFor_` APIs above.

PR #2621, opened 2024-08-27
Add `replicaRouting` compat flag

MellowYarker:
This will remain `experimental` for some time.

Note that turning this on in local dev would effectively be a no-op, but rather than allow the flag to be set and continue regular operation, we'll throw an exception.

### DIFF
--- a/src/workerd/io/io-channels.h
+++ b/src/workerd/io/io-channels.h
@@ -151,6 +151,7 @@ public:
       const ActorIdFactory::ActorId& id,
       kj::Maybe<kj::String> locationHint,
       ActorGetMode mode,
+      bool enableReplicaRouting,
       SpanParent parentSpan) = 0;
 
   // Get an actor stub from the given namespace for the actor with the given name.
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -721,9 +721,10 @@ public:
       const ActorIdFactory::ActorId& id,
       kj::Maybe<kj::String> locationHint,
       ActorGetMode mode,
+      bool enableReplicaRouting,
       SpanParent parentSpan) {
     return getIoChannelFactory().getGlobalActor(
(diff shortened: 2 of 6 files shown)


src/workerd/io/io-channels.h
```diff
@@ -151,7 +151,8 @@ class IoChannelFactory {
       const ActorIdFactory::ActorId& id,
       kj::Maybe<kj::String> locationHint,
       ActorGetMode mode,
-      SpanParent parentSpan) = 0;
+      SpanParent parentSpan,
```
justin-mp, line 154: nit: can the SpanParent remain as the last parameter?  Our style seems to be that observability stuff goes at the end after all functionality-related parameters.

src/workerd/io/io-context.h
```diff
@@ -721,9 +721,10 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
       const ActorIdFactory::ActorId& id,
       kj::Maybe<kj::String> locationHint,
       ActorGetMode mode,
-      SpanParent parentSpan) {
+      SpanParent parentSpan,
+      bool enableReplicaRouting) {
```
justin-mp, line 725: nit: again, can the SpanParent be the last parameter?

src/workerd/io/io-channels.h
MellowYarker: Funnily enough I actually did this at first, then swapped the parameters right before I pushed.

justin-mp: Should have trusted your gut!

jasnell (approved):

MellowYarker:
https://github.com/cloudflare/workerd/compare/1560c005eceb996c33fc2a0a70c4e42d7e84f3ec..619b393cbc722c98489f056021b24eab3b414247 is a rebase on `main`.

PR #2623, opened 2024-08-28
call synctime when receiving new websocket messages

Frederik-Baetens:
This seems safe to me since I don't really see a fundamental difference between websocket messages and fetch requests, and I think the hibernatable path already ends up calling synctime because I think it calls into IoContext::IncomingRequest::delivered() through HibernatableWebSocketCustomEventImpl::run(). (I'm not sure whether that run method gets ran for every new ws message though)

I think you may be able to get more accurate timestamps by sending messages to a websocket pair that you just hold in a worker internally, but I think you may already be able to do that with hibernatable websockets?

### DIFF
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -962,6 +962,9 @@ kj::Promise<kj::Maybe<kj::Exception>> WebSocket::readLoop(
         a.getMetrics().receivedWebSocketMessage(size);
       }
 
+      // This should count as a new IO event, so we should call syncTime
+      context.getIoChannelFactory().getTimer().syncTime();
+
       // Re-enter the context with context.run(). This is arguably a bit unusual compared to other
       // I/O which is delivered by return from context.awaitIo(), but the difference here is that we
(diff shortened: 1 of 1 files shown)


Frederik-Baetens:
I'm not sure if this should be wrapped in:

```
KJ_IF_SOME(a, contexts.getActor()) {
  ioChannelFactory->getTimer().syncTime();
}
```

`IoContext::IncomingRequest::delivered()` does that, but I'm not sure what the benefit of doing that is.

kentonv:
This is almost certainly already happening somewhere else in the stack. This doesn't belong in code specific to one event type.

kentonv (changes requested):
Yeah, syncTime() already happens as part of `context.run()` which is called a few lines later.

kentonv:
https://github.com/cloudflare/workerd/blob/main/src/workerd/io/io-context.c%2B%2B#L1003

Frederik-Baetens:
Thanks kenton, missed that.

kentonv:
I suppose we should close this, then.

PR #2624, opened 2024-08-28
Refactor the D1 binding to use D1DatabaseSession for all actions

lambrospetrou:
## Overview

This change prepares the D1 binding for the upcoming Sessions API (relevant https://blog.cloudflare.com/building-d1-a-global-database). Most of the changes is moving things around, but the highlights are as below:
1. A new `D1DatabaseSession` class is introduced that maintains a `commitTokenOrConstraint` property, and also takes ownership of the `fetch`+`_send/_sendOrThrow` methods.
    - This class proxies all calls to the D1 eyeball worker, and automatically adds the commit token as an HTTP header to the requests and extracts it from the response headers.
    - Our eyeball worker and the D1 DO will extract the commit token and accordingly decide which replica should route the requests to.
2. There is a new `D1DatabaseSessionAlwaysPrimary` that extends `D1DatabaseSession` and essentially always sends the commit token constraint `first-primary`.
    - This session subclass also provides the existing `dump, exec` functions that the new Sessions API does not have.
4. As specified by the SPEC, the default if no commit token is provided to the `withSession()` function is `"first-unconstrained"` (to run the first query (which may be a read or a write) with no constraints, and then use causal consistency for subsequent queries).

## Notes

I decided to design it such that `D1DatabaseSession` proxies all the calls and maintains the commit token so that the other classes don't really care about which session they use, or having to know about it.

This way, all the complex logic about the commit token and how the interaction with the D1 DOs is going to be inside our eyeball worker, which we control fully, it's easier to deploy and fix, and is the first shared entrypoint with the REST API too, so once the REST API can pass the commit token to use, the rest of the logic is exactly the same.

I am not sure if using HTTP headers is common in Cloudflare to pass this kind of information to our APIs, so if that's not allowed/suggested, then we will need to pass it as query parameter instead.

Suggestions for naming of the header or the query parameter welcome.

## Compatibility flag

There is a new `experimental` compatibility flag `enable_d1_with_sessions_api` to expose this new `withSessions()` method. The default binding still uses the `D1DatabaseSession` object though, so this PR's changes will actually be running after merge, always sending the `first-primary` constraint. 

The `d1-api-test.js` is used by the D1 eyeball worker E2E tests so I added the new tests in a new test suite with the compatibility flag enabled only for that.

## Tests

```sh
bazel run //src/cloudflare/internal/test/d1:d1-api-test
...
Executing tests from //src/cloudflare/internal/test/d1:d1-api-test
-----------------------------------------------------------------------------
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test:testWithoutSessions
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test:testWithoutSessions


bazel run //src/cloudflare/internal/test/d1:d1-api-test-with-sessions
...
Executing tests from //src/cloudflare/internal/test/d1:d1-api-test-with-sessions
-----------------------------------------------------------------------------
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test-with-sessions:test_d1_api_happy_path
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test-with-sessions:test_d1_api_happy_path
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_default
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_default
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_first_primary
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_first_primary
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_first_unconstrained
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_first_unconstrained
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_some_ranomd_token
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test-with-sessions:test_d1_api_happy_path_withsessions_some_ranomd_token
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test-with-sessions:test_d1_api_withsessions_old_token_skipped
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test-with-sessions:test_d1_api_withsessions_old_token_skipped
workerd/server/server.c++:3847: debug: [ TEST ] d1-api-test-with-sessions:test_d1_api_withsessions_token_handling
workerd/server/server.c++:3855: debug: [ PASS ] d1-api-test-with-sessions:test_d1_api_withsessions_token_handling
```

The first target runs the default database tests, without using the sessions API, and without the compatibility flag enabled.

The second target runs all the test suites with the compatibility flag enabled.


lambrospetrou:

src/cloudflare/internal/d1-api.ts
```diff
@@ -52,38 +54,102 @@ type SQLError = {
 
 type ResultsFormat = 'ARRAY_OF_OBJECTS' | 'ROWS_AND_COLUMNS' | 'NONE';
 
+type D1SessionCommitTokenOrConstraint = string;
+type D1SessionCommitToken = string;
+
+// TODO Figure out what header to use.
+const D1_SESSION_COMMIT_TOKEN_HTTP_HEADER = 'x-cf-d1-session-commit-token';
```
rozenmd, line 61: I guess query parameters are probably better to avoid losing them

lambrospetrou: What do you mean avoid losing them?

The reason I chose a header is to make it easy to use in the REST API queries we offer as well in the same way. Since there we pass params in the body instead of query params.

We could support both query params and body params (in that order) though, and that would be fine.

I don't have strong preference, either way works for me.

src/cloudflare/internal/d1-api.ts
```diff
@@ -52,38 +52,96 @@ type SQLError = {
...
+  protected commitTokenOrConstraint: D1SessionCommitTokenOrConstraint;
+
+  public constructor(
+    envFetcher: Fetcher,
+    commitTokenOrConstraint: D1SessionCommitTokenOrConstraint
+  ) {
+    this.fetcher = envFetcher;
+    this.commitTokenOrConstraint = commitTokenOrConstraint;
+
+    // TODO Do format validation for the commit token?
```
justin-mp, line 125: nit: the workerd repository follows the KJ style guide for TODOs.  See https://github.com/capnproto/capnproto/blob/v2/style-guide.md#comments.

src/cloudflare/internal/d1-api.ts
```diff
@@ -52,38 +52,96 @@ type SQLError = {
...
+    if (!this.commitTokenOrConstraint) {
+      throw new Error('D1_SESSION_ERROR: invalid commit token or constraint');
     }
-    return await response.arrayBuffer();
+  }
+
+  // Update the commit token IFF the given newCommitToken is more recent.
```
justin-mp, line 131: I take it that this is the part of the code that protects against the case where there are multiple requests in flight from the same session.  It might be worth dropping a comment saying so!

src/cloudflare/internal/d1-api.ts
justin-mp: I think @rozenmd means that it's really easy to accidentally drop request headers when chaining fetch requests (since you usually have to copy them into the new sub-request that you're making).  Query parameters are often more explicit since they're part of the URL and we generally proxy the URL.

I don't have a preference either way!

src/cloudflare/internal/d1-api.ts
```diff
@@ -98,34 +156,50 @@ class D1Database {
...
-          return p + (c.meta['duration'] as number);
-        }, 0),
-      };
+      init.headers = h;
     }
+    return this.fetcher.fetch(input, init).then((resp) => {
+      const newCommitToken = resp.headers.get(
+        D1_SESSION_COMMIT_TOKEN_HTTP_HEADER
+      );
+      if (newCommitToken) {
```
justin-mp, line 206: It feels like we should require that if we send a commit token, we should expect a commit token in return and must raise an error if that's not the case.  Maybe drop a TODO?

justin-mp (commented):
I won't comment on the details of the code, but the class structure seems reasonable, especially how the statements use the session class.

src/cloudflare/internal/d1-api.ts
lambrospetrou: OK, I will switch to query params since there is preference there.

src/cloudflare/internal/d1-api.ts
lambrospetrou: I was actually going to ask you about this. Is there even any validation we can do to the token here? Mabe just the length?

src/cloudflare/internal/d1-api.ts
lambrospetrou: Yeap, that's its purpose. I will add a comment further explaining it.

src/cloudflare/internal/d1-api.ts
lambrospetrou: Hmm, good point. I haven't about it being required, but it makes sense to expect a token back if you are using a session, since otherwise the session semantics might break. I guess in the current state, the same token will be sent again next time, which will potentially "go back in time" in case there was any write afterwards.

It's a bit tricky during the rollout though of this change, since once this is merged there is always going to be a token sent (i.e. `first-primary`), but the backend API won't send anything back yet... 

I will put a TODO for now, and once the feature is fully rolled out we can add this check.

geelen (approved):
Looks good! A bit of feedback, but nothing major, so consider this approved.

src/cloudflare/internal/test/d1/d1-api-test.wd-test
```diff
@@ -9,7 +9,7 @@ const unitTests :Workerd.Config = (
           (name = "worker", esModule = embed "d1-api-test.js")
         ],
         compatibilityDate = "2023-01-15",
-        compatibilityFlags = ["nodejs_compat"],
+        compatibilityFlags = ["nodejs_compat", "enable_d1_with_sessions_api"],
```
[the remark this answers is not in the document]
lambrospetrou: It took me some time to figure out the Workerd configuration, but it seems simpler now actually too. I added the new tests into a new test suite and extracted the common tests into a shared file.

I left the existing file without the sessions API tests because it's used by the Eyeball E2E tests by its name. I will adjust those too later to include Sessions tests.

src/cloudflare/internal/test/d1/d1-api-test-common.js
```diff
@@ -0,0 +1,515 @@
...
+  changes: anything,
+  last_row_id: anything,
+  changed_db: anything,
+  size_after: anything,
+  rows_read: anything,
+  rows_written: anything,
+  ...values,
+});
+
+export async function testD1ApiQueriesHappyPath(DB) {
```
lambrospetrou, line 57: This was cut-pasted as-is, without any changes apart from adding the last section of DELETE statements, so that the database is cleaned up.

lambrospetrou, line 57: This was cut-pasted as-is, without any changes apart from adding the last section of DELETE statements, so that the database is cleaned up.

src/cloudflare/internal/d1-api.ts
lambrospetrou: FYI, I am going to stick with headers now at least for the communication between the D1 Binding and the eyeball worker, since it's easier to not requiring to parse full body responses to extract the token (even if they are already parsed before returned to users).

For the public API we can do query parameters if needed, and the eyeball can handle both.

src/cloudflare/internal/d1-api.ts
lambrospetrou: I am inclined to add the validation later, since I don't know yet exactly what a token could look like or how we would validate it other than "is it not an empty string".

Also, this PR already adds a bit of logic changes, so don't want to introduce also a breaking change that requires eyeball changes to be merged and released first.

dom96 (approved):

danlapid (approved):

PR #2625, opened 2024-08-28
[build] Refactor zlib configuration

fhanau:
- Drop visibility attribute, only zlib library itself needs to be visible
- Drop explicit includes directive, this is generally bad style in bazel
- Drop support for unsupported architectures, add target_compatible_with for arch-specific targets

Will make another PR that switches to using `local_defines` in the afternoon.


anonrig (approved):
Would this reduce build time?

npaun:
Seems like something got bungled with visibility?

fhanau:
It's linking with system zlib instead of this version of zlib when dropping the `includes = ["."]` – unexpected, but easy to work around.

npaun (approved):

PR #2627, opened 2024-08-29
[build] Further zlib build refactor, document new hardware requirements

fhanau:
Also fixes a potential build failure for arm64 – the `zlib_arm_crc32` target did not have the required crc32 extension enabled. This did not surface on Apple Silicon builds as the extension is supported by all Apple Silicon CPUs and enabled unconditionally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,7 @@ Prebuilt binaries are distributed via `npm`. Run `npx workerd ...` to use these.
 * On macOS:
   * macOS 11.5 or higher
   * The Xcode command line tools, which can be installed with `xcode-select --install`
+* x86_64 CPU with at least SSE4.2 and CLMUL ISA extensions, or arm64 CPU with CRC extension (enabled by default under armv8.1-a). These extensions are supported by all recent x86 and arm64 CPUs.
 
 ### Local Worker development with `wrangler`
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -182,6 +182,7 @@ Prebuilt binaries are distributed via `npm`. Run `npx workerd ...` to use these.
 * On macOS:
   * macOS 11.5 or higher
   * The Xcode command line tools, which can be installed with `xcode-select --install`
+* x86_64 CPU with at least SSE4.2 and CLMUL ISA extensions, or arm64 CPU with CRC extension (enabled by default under). These extensions are supported by all recent x86 and arm64 CPUs.
```
fhanau, line 185: Since the question will inevitably come up: This level of ISA extensions is supported by Intel Westmere (2010), AMD Bulldozer (2011), armv8.1-a (introduced 2014, although CRC can also be added as an optional extension to plain armv8-a), and higher. I want to eventually require something like -march=x86-64-v3/-march=armv8.2-a+crypto, but that will require prior discussion.

npaun: > enabled by default under 

I think you accidentally a word

jasnell: >> enabled by default under

> I think you accidentally a word

Heh, I think you both accidentally forgot a word ;-)

jasnell (approved):

npaun (approved):

PR #2629, opened 2024-08-29
implement domainToASCII and domainToUnicode

anonrig:
Implements domainToASCII and domainToUnicode and exports them under `node:url` where `URL` and `URLSearchParams` also exist.


src/node/url.ts
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
npaun, line 1: Where does the unsafe assignment error come from? I would suggest moving the disable to that line.

anonrig: There are 2 lines, specifically `const XX = globalThis.XX`

src/workerd/api/node/url.c++
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+#include "url.h"
+
+#include "ada.h"
```
npaun, line 6:
```suggestion
#include <ada.h>
```

src/workerd/api/node/url.c++
```diff
@@ -0,0 +1,39 @@
...
+    return {};
+  }
+
+  auto out = ada::parse<ada::url_aggregator>("ws://x");
+  JSG_REQUIRE(out.has_value(), Error, "URL parsing failed"_kj);
+  if (!out->set_hostname({domain.begin(), domain.size()})) {
+    return {};
+  }
+
+  auto result = ada::idna::to_unicode(out->get_hostname());
```
npaun, line 35: It's a bit odd that the toASCII function is invoking to_unicode. Were they swapped by accident? If not, please add a comment to explain what to_unicode does.

src/workerd/api/node/url.c++
```diff
@@ -0,0 +1,39 @@
...
+  auto result = out->get_hostname();
+  return kj::ArrayPtr<const char>(result.data(), result.size());
+}
+
+kj::ArrayPtr<const char> UrlUtil::domainToASCII(kj::String domain) {
+  if (domain.size() == 0) {
+    return {};
+  }
+
+  auto out = ada::parse<ada::url_aggregator>("ws://x");
```
npaun, line 29: I'm guessing `ws://x` is a fake URL to initialize with? If so, please make a constant for it.

src/workerd/api/node/url.c++
anonrig: src/workerd/api/node/url.c++:6:10: error: 'ada.h' file not found with <angled> include; use "quotes" instead
    6 | #include <ada.h>
      |          ^~~~~~~
      |          "ada.h"

src/node/url.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return urlUtil.domainToASCII(`${domain}`);
+}
+
+function domainToUnicode(domain: string): string {
+  if (arguments.length < 1) {
+    throw new ERR_MISSING_ARGS('domain');
+  }
+
+  return urlUtil.domainToASCII(`${domain}`);
```
jasnell, line 21: shouldn't this to `urlUtil.domainToUnicode`?

src/workerd/api/node/url.c++
anonrig: Making a constant means that the URL is shared between multiple calls. Technically that would be faster, but for the sake of being 100% compliant with Node.js, I prefer to keep it as it is.

src/node/url.ts
anonrig: Yup

src/workerd/api/node/url.c++
npaun: Bazel is going to make me tear my hair out. I'll try to figure out what it's actually doing with the includes.

src/node/url.ts
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import { default as urlUtil } from 'node-internal:url';
+import { ERR_MISSING_ARGS } from 'node-internal:internal_errors';
+
+const URL = globalThis.URL;
+const URLSearchParams = globalThis.URLSearchParams;
```
jasnell, line 6:
```suggestion
const { URL, URLSearchParams } = globalThis;
export { URL, URLSearchParams };
```

npaun (approved):

src/node/url.ts
```diff
@@ -0,0 +1,31 @@
...
+
+function domainToASCII(domain: string): string {
```
jasnell, line 8:
```suggestion
export function domainToASCII(domain: string): string {
```

src/node/url.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (arguments.length < 1) {
+    throw new ERR_MISSING_ARGS('domain');
+  }
+
+  return urlUtil.domainToASCII(`${domain}`);
+}
+
+function domainToUnicode(domain: string): string {
```
jasnell, line 16:
```suggestion
export function domainToUnicode(domain: string): string {
```

src/node/url.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (arguments.length < 1) {
+    throw new ERR_MISSING_ARGS('domain');
+  }
+
+  return urlUtil.domainToUnicode(`${domain}`);
+}
+
+export { domainToASCII, domainToUnicode, URL, URLSearchParams };
```
jasnell, line 24:
```suggestion
export { URL, URLSearchParams };
```

src/workerd/api/node/url.c++
```diff
@@ -0,0 +1,44 @@
...
+  // Since it will change the implementation of `set_hostname` according to URL
+  // spec.
+  auto out = ada::parse<ada::url_aggregator>("ws://x");
+  JSG_REQUIRE(out.has_value(), Error, "URL parsing failed"_kj);
+  if (!out->set_hostname({domain.begin(), domain.size()})) {
+    return {};
+  }
+  auto result = out->get_hostname();
+  return kj::ArrayPtr<const char>(result.data(), result.size());
+}
```
jasnell, line 24: nit: there is a fair amount of duplication in these two methods. Completely optional but it might make sense to move the bulk of the impl is a separate utility method that is used by both of these, just to eliminate the duplication.

src/workerd/api/node/url.c++
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+#include "url.h"
+
+#include "ada.h"
+
+namespace workerd::api::node {
+kj::ArrayPtr<const char> UrlUtil::domainToASCII(kj::String domain) {
+  if (domain.size() == 0) {
+    return {};
```
jasnell, line 11: I wonder if these should return `kj::Maybe<kj::ArrayPtr<const char>>` instead so we can differentiate between a response that produces a zero-length result vs. a response that fails (like in the `!out->set_hostname(...` block below. Not blocking tho

src/workerd/api/node/url.c++
```diff
@@ -0,0 +1,44 @@
...
+  // It is important to have an initial value that contains a special scheme.
+  // Since it will change the implementation of `set_hostname` according to URL
+  // spec.
+  auto out = ada::parse<ada::url_aggregator>("ws://x");
+  JSG_REQUIRE(out.has_value(), Error, "URL parsing failed"_kj);
+  if (!out->set_hostname({domain.begin(), domain.size()})) {
+    return {};
+  }
+  auto result = out->get_hostname();
+  return kj::ArrayPtr<const char>(result.data(), result.size());
```
jasnell, line 23: there is a `kj::arrayPtr(...)` method that slightly reduces the boilerplate here, just fyi

src/workerd/api/node/url.c++
anonrig: Technically yes, but Node.js implementation returns empty string for invalid value. On top of that to_unicode() function in Ada also returns empty string if it's invalid.

jasnell (approved):

npaun (approved):

jasnell:
@IgorMinar ... please take a look

IgorMinar (approved):
LGTM

For those wondering how this PR relates to https://github.com/unjs/unenv/pull/299, @anonrig clarified with me that these two methods are nearly impossible to implement in userland in a performant way. So we are implementing them natively, while other parts of `node:url` could come in via unenv.

// fyi: @pi0

jasnell:
@pi0 ... so you know what to expect... these will be exported by the runtime as the `node:url` module. The bundler will essentially need to alias `node:url`, use the `process.getBuiltinModule('node:url')` API to get the built-in version to extend, the re-export the fully patched module.

IgorMinar:
> @pi0 ... so you know what to expect... these will be exported by the runtime as the `node:url` module. The bundler will essentially need to alias `node:url`, use the `process.getBuiltinModule('node:url')` API to get the built-in version to extend (that has these methods), then re-export the fully patched module.

Yes! except unenv will do exactly this in a new file we'll create under https://github.com/unjs/unenv/tree/main/src/runtime/node/url called `$cloudflare.ts`, following the existing precedence for hybrid modules: https://github.com/unjs/unenv/blob/main/src/runtime/node/buffer/%24cloudflare.ts

Bundlers then pick up `url/$cloudflare.ts` as instructed by the [cloudflare preset](https://github.com/unjs/unenv/blob/main/src/presets/cloudflare.ts#L24-L32) (which also needs to be updated to include `url`) and combine it all into the final bundle.

jasnell:
@anonrig ... as discussed in chat, let's make sure this is added to the test:

```patch
diff --git a/src/workerd/api/node/tests/url-nodejs-test.js b/src/workerd/api/node/tests/url-nodejs-test.js
index ab683c9f..c75fbfad 100644
--- a/src/workerd/api/node/tests/url-nodejs-test.js
+++ b/src/workerd/api/node/tests/url-nodejs-test.js
@@ -77,3 +77,11 @@ export const urlAndSearchParams = {
     );
   },
 };
+
+export const getBuiltinModule = {
+  async test() {
+    const bim = process.getBuiltinModule('node:url');
+    const url = await import('node:url');
+    strictEqual(bim, url.default);
+  }
+};
diff --git a/src/workerd/api/node/tests/url-nodejs-test.wd-test b/src/workerd/api/node/tests/url-nodejs-test.wd-test
index 1be17b89..92589f61 100644
--- a/src/workerd/api/node/tests/url-nodejs-test.wd-test
+++ b/src/workerd/api/node/tests/url-nodejs-test.wd-test
@@ -8,7 +8,7 @@ const unitTests :Workerd.Config = (
           (name = "worker", esModule = embed "url-nodejs-test.js")
         ],
         compatibilityDate = "2023-10-01",
-        compatibilityFlags = ["nodejs_compat"],
+        compatibilityFlags = ["nodejs_compat_v2"],
       )
     ),
   ],
```

jasnell (approved):

PR #2630, opened 2024-08-29
Log when deleteAll is called with an alarm still set on the object

a-robinson:
To get a better sense of the prevalence of this usage pattern.


jclee (approved):

justin-mp (approved):

kentonv:
Should this use LOG_PERIODICALLY in case this is very common?

a-robinson:
> Should this use LOG_PERIODICALLY in case this is very common?

I was hoping to avoid that so that we could get more complete data, operating under the assumption that deleteAll itself is only used very infrequently, and that it should be even less frequent in combination with alarms. But now I feel obligated to since it'll look silly if that assumption ends up being wrong 🙃

Switched it over.

PR #2633, opened 2024-08-30
Add PyodidePackageManager implementation

garrettgu10:
This is a scaled-back version of #2618. 

I'm pushing ahead with this change while we figure out what we're doing with the ArtifactBundler & package loading promises.

### DIFF
--- a/src/workerd/api/pyodide/pyodide.h
+++ b/src/workerd/api/pyodide/pyodide.h
@@ -29,6 +29,15 @@ private:
   const kj::MutexGuarded<kj::HashMap<kj::String, MessageBundlePair>> bundles;
 };
 
+class PyodidePackageManager {
+public:
+  void setPyodidePackageData(kj::String id, kj::Array<unsigned char> data) const;
+  const kj::Maybe<kj::ArrayPtr<const unsigned char>> getPyodidePackage(kj::StringPtr id) const;
+
+private:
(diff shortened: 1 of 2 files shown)


src/workerd/api/pyodide/pyodide.h
```diff
@@ -29,6 +29,15 @@ class PyodideBundleManager {
   const kj::MutexGuarded<kj::HashMap<kj::String, MessageBundlePair>> bundles;
 };
 
+class PyodidePackageManager {
+public:
+  void setPyodidePackageData(kj::String id, kj::Array<unsigned char> data) const;
+  const kj::Maybe<kj::ArrayPtr<const unsigned char>> getPyodidePackage(kj::StringPtr id) const;
```
anonrig, line 35: If I understood your intentions, ArrayPtr should be cosnt, not kj::Maybe right?

```suggestion
  kj::Maybe<const kj::ArrayPtr<const unsigned char>> getPyodidePackage(kj::StringPtr id) const;
```

garrettgu10: I originally tried making the ArrayPtr const instead, but that caused weird template instantiation errors. From reading kj/common.h, my understanding is that making the kj::Maybe const signals that the inner thing should be const as well.

anonrig (approved):

PR #2634, opened 2024-08-30
Add NonRetryableError

LuisDuarte1:


src/cloudflare/internal/workflows-api.ts
```diff
@@ -0,0 +1,10 @@
+export class NonRetryableError extends Error {
+  // `__brand` is needed for Workflows' engine to validate if the user returned a NonRetryableError
+  // this provides better DX because they can just extend NonRetryableError for their own Errors
+  public readonly __brand: string = 'NonRetryableError';
```
jasnell, line 4: Non-blocking but...This is a bit unfortunate. I'd almost prefer something like `NonRetryableError.isInstance(err)` that can check an internal private... e.g.

```js
const kTag = Symbol('kTag');

class NonRetryableError extends Error {
  [kTag] = kTag;  

  static isInstance(value) { return value[kTag] === kTag; }
}
```

types/defines/workflows.d.ts
```diff
@@ -0,0 +1,8 @@
+
+/**
+ * NonRetryableError allows for a Workflow to throw a "fatal" error as in,
+ * an error that makes the instance fail immediately without triggering a retry.
+ */
+export class NonRetryableError extends Error {
+  public constructor(message: string, name?: string);
```
jasnell, line 7: Are you intentionally omitting the `__brand` string from the types? If so, would be good to document the reason so someone doesn't come along later and add it.

src/cloudflare/internal/workflows-api.ts
LuisDuarte1: Is your suggested implementation serialized by RPC? If yes, I'm fine with replacing it 😅

jasnell: ah, no, it would not be. Would you mind adding a comment then to indicate that the field is a public own property because it needs to be serialized for RPC?

anonrig: Why don't we use name property as a way if distinguishing this error type?

types/defines/workflows.d.ts
LuisDuarte1: done

anonrig (approved):

PR #2636, opened 2024-08-31
Implement node:module createRequire API

jasnell:
The `createRequire()` function is used in node.js to create a `require()` function that can be used from inside an ESM module. This is useful largely for porting existing CommonJS code to ESM as it allows the ESM module to keep using `require()` to load other modules.

This is being implemented in the runtime rather than as a polyfill because it needs to integrate tightly with the module registry implementation.

The implementation is as close to Node.js' as we can currently make it.

Note that the return value of `createRequire()` is a function. In Node.js, this function has additional properties like `require.resolve(...)`, `require.main`, `require.extensions`, and `require.cache`. We are not currently implementing these additional properties but we could in the future if there is a need for them. (`require.cache` is unlikely to ever be implemented since workers does not have a module cache the way Node.js does).

Because it was extremely trivial to do so, is closely related, and we already have to maintain a list of known node builtins,
this also implements the `require('module').isBuiltin(...)` API.

This was implemented because the frameworks team asked for it.

Refs: https://nodejs.org/docs/latest/api/module.html#modulecreaterequirefilename

### DIFF
--- a/src/node/internal/module.d.ts
+++ b/src/node/internal/module.d.ts
@@ -1,0 +1,6 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+export function createRequire(path: string): (specifier: string) => unknown;
+export function isBuiltin(specifier: string): boolean;
--- a/src/node/module.ts
+++ b/src/node/module.ts
@@ -1,0 +1,122 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+/* eslint-disable */
+
+import { default as moduleUtil } from 'node-internal:module';
+import { ERR_INVALID_ARG_VALUE } from 'node-internal:internal_errors';
+
+export function createRequire(
+  path: string | URL
+): (specifier: string) => unknown {
+  // Note that per Node.js' requirements, path must be one of either
+  // an absolute file path or a file URL. We do not currently handle
+  // module specifiers as URLs yet but we'll try to get close.
+
+  path = `${path}`;
+  if (
+    !(path as string).startsWith('/') &&
+    !(path as string).startsWith('file:')
+  ) {
+    throw new ERR_INVALID_ARG_VALUE(
+      'path',
+      path,
+      'The argument must be a file URL object, ' +
+        'a file URL string, or an absolute path string.'
+    );
+  }
+
+  return moduleUtil.createRequire(path as string);
+}
+
+// Indicates only that the given specifier is known to be a
+// Node.js built-in module specifier with or with the the
+// 'node:' prefix. A true return value does not guarantee that
+// the module is actually implemented in the runtime.
+export function isBuiltin(specifier: string): boolean {
+  return moduleUtil.isBuiltin(specifier);
+}
+
+// Intentionally does not include modules with mandatory 'node:'
+// prefix like `node:test`.
+// See: See https://nodejs.org/docs/latest/api/modules.html#built-in-modules-with-mandatory-node-prefix
+// TODO(later): This list duplicates the list that is in
+// workerd/jsg/modules.c++. Later we should source these
+// from the same place so we don't have to maintain two lists.
+export const builtinModules = [
+  '_http_agent',
+  '_http_client',
+  '_http_common',
+  '_http_incoming',
+  '_http_outgoing',
+  '_http_server',
+  '_stream_duplex',
+  '_stream_passthrough',
+  '_stream_readable',
+  '_stream_transform',
+  '_stream_wrap',
+  '_stream_writable',
+  '_tls_common',
+  '_tls_wrap',
+  'assert',
+  'assert/strict',
+  'async_hooks',
+  'buffer',
+  'child_process',
+  'cluster',
+  'console',
+  'constants',
+  'crypto',
+  'dgram',
+  'diagnostics_channel',
+  'dns',
+  'dns/promises',
+  'domain',
+  'events',
+  'fs',
+  'fs/promises',
+  'http',
+  'http2',
+  'https',
+  'inspector',
+  'inspector/promises',
+  'module',
+  'net',
+  'os',
+  'path',
+  'path/posix',
+  'path/win32',
+  'perf_hooks',
+  'process',
+  'punycode',
+  'querystring',
+  'readline',
+  'readline/promises',
+  'repl',
+  'stream',
+  'stream/consumers',
+  'stream/promises',
+  'stream/web',
+  'string_decoder',
+  'sys',
+  'timers',
+  'timers/promises',
+  'tls',
+  'trace_events',
+  'tty',
+  'url',
+  'util',
+  'util/types',
+  'v8',
+  'vm',
+  'wasi',
+  'worker_threads',
+  'zlib',
+];
+Object.freeze(builtinModules);
+
+export default {
+  createRequire,
+  isBuiltin,
+  builtinModules,
+};
--- a/src/workerd/api/node/module.c++
+++ b/src/workerd/api/node/module.c++
@@ -1,0 +1,112 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+#include "module.h"
+#include <workerd/jsg/url.h>
+
+namespace workerd::api::node {
+
+bool ModuleUtil::isBuiltin(kj::String specifier) {
+  return jsg::checkNodeSpecifier(specifier) != kj::none;
+}
+
+jsg::JsValue ModuleUtil::createRequire(jsg::Lock& js, kj::String path) {
+  // Node.js requires that the specifier path is a File URL or an absolute
+  // file path string. To be compliant, we will convert whatever specifier
+  // is into a File URL if possible, then take the path as the actual
+  // specifier to use.
+  auto parsed = JSG_REQUIRE_NONNULL(jsg::Url::tryParse(path.asPtr(), "file:///"_kj), TypeError,
+      "The argument must be a file URL object, "
+      "a file URL string, or an absolute path string.");
+
+  // We do not currently handle specifiers as URLs, so let's treat any
+  // input that has query string params or hash fragments as errors.
+  if (parsed.getSearch().size() > 0 || parsed.getHash().size() > 0) {
+    JSG_FAIL_REQUIRE(
+        Error, "The specifier must not have query string parameters or hash fragments.");
+  }
+
+  // The specifier must be a file: URL
+  JSG_REQUIRE(parsed.getProtocol() == "file:"_kj, TypeError, "The specifier must be a file: URL.");
+
+  return jsg::JsValue(js.wrapReturningFunction(js.v8Context(),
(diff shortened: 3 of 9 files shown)


src/workerd/api/node/module.c++
```diff
@@ -0,0 +1,109 @@
+#include "module.h"
```
jasnell, line 1:
```suggestion
// Copyright (c) 2017-2022 Cloudflare, Inc.
// Licensed under the Apache 2.0 license found in the LICENSE file or at:
//     https://opensource.org/licenses/Apache-2.0
#include "module.h"
```

src/workerd/api/node/module.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
jasnell, line 1:
```suggestion
// Copyright (c) 2017-2022 Cloudflare, Inc.
// Licensed under the Apache 2.0 license found in the LICENSE file or at:
//     https://opensource.org/licenses/Apache-2.0
#pragma once
```

src/workerd/api/node/tests/module-create-require-test.js
```diff
@@ -0,0 +1,68 @@
+import { createRequire, isBuiltin } from 'node:module';
```
jasnell, line 1:
```suggestion
// Copyright (c) 2017-2022 Cloudflare, Inc.
// Licensed under the Apache 2.0 license found in the LICENSE file or at:
//     https://opensource.org/licenses/Apache-2.0
import { createRequire, isBuiltin } from 'node:module';
```

src/workerd/api/node/module.c++
```diff
@@ -0,0 +1,112 @@
...
+  JSG_REQUIRE(parsed.getProtocol() == "file:"_kj, TypeError, "The specifier must be a file: URL.");
+
+  return jsg::JsValue(js.wrapReturningFunction(js.v8Context(),
+      [referrer = kj::str(parsed.getPathname())](
+          jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>& args) -> v8::Local<v8::Value> {
+    auto registry = jsg::ModuleRegistry::from(js);
+
+    // TODO(soon): This will need to be updated to support the new module registry
+    // when that is fully implemented.
+    JSG_REQUIRE(registry != nullptr, Error, "Module registry not available.");
```
anonrig, line 39:
```suggestion
    JSG_REQUIRE_NONNULL(registry, Error, "Module registry not available.");
```

jasnell: While that works it's largely by accident. `JSG_REQUIRE_NONNULL` was intended to be used with `kj::Maybe<T>` not `T*`. I'd rather keep the check as is as it makes it a bit clearer that `registry` is a raw pointer.

src/node/module.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (!path.startsWith('/') && !path.startsWith('file:')) {
+    throw new ERR_INVALID_ARG_VALUE(
+      'path',
+      path,
+      'The argument must be a file URL object, ' +
+        'a file URL string, or an absolute path string.'
+    );
+  }
+
+  return moduleUtil.createRequire(`${path}`);
```
vicb, line 23: nit: use `path`

src/node/module.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import { default as moduleUtil } from 'node-internal:module';
+import { ERR_INVALID_ARG_VALUE } from 'node-internal:internal_errors';
+
+export function createRequire(path: string | URL): any {
```
vicb, line 8: Could the returned type be refined?

It would help with:

> Note that the return value of createRequire() is a function. In Node.js, this function has additional properties like require.resolve(...), require.main, require.extensions, and require.cache. We are not currently implementing these additional properties but we could in the future if there is a need for them. (require.cache is unlikely to ever be implemented since workers does not have a module cache the way Node.js does).

src/node/module.ts
```diff
@@ -0,0 +1,33 @@
...
+  // Note that per Node.js' requirements, path must be one of either
+  // an absolute file path or a file URL. We do not currently handle
+  // module specifiers as URLs yet but we'll try to get close.
+
+  path = `${path}`;
+  if (!path.startsWith('/') && !path.startsWith('file:')) {
```
vicb, line 14: Would be be possible to extend the support to builtin modules?

src/node/module.ts
pi0: I'm not sure how much compatibility you are looking for but one main purpose of cache is to handle circular dependencies. Nevertheless, we can also implement the missing APIs (or at least a stub of them) in unenv layer if needed, keeping workerd exports for native part.

pi0:
Re unenv: I think unenv should still return all known built-ins and also return `true` for known built-ins. Libraries depend on both to determine if an import is Node.js built-in or not and if we only return true for the native-subset of workerd, it can break their behavior or at least make inconsistencies. I understand this is something useful for workerd itself to support but probably not for mixed compatibility.

vicb:
> Re unenv: I think unenv should still return all known built-ins for `builtinModules` and also return `true` for `isBuiltin`

Good point.

src/node/module.ts
jasnell: What you mean?

src/node/module.ts
jasnell: Ah yes, forgot to update that.

src/node/module.ts
anonrig: path variable was already converted into string couple of lines above this line

jasnell: oh right! ha, yeah forgot to change this one lol.

jasnell, line 23:
```suggestion
  return moduleUtil.createRequire(path);
```

jasnell:
> Re unenv: I think unenv should still return all known built-ins for builtinModules and also return true for isBuiltin if it is a known built in (same behavior as Node.js). Libraries depend on both to determine if an import is Node.js built-in or not and if we only return true for the native-subset of workerd, it can break their behavior or at least make inconsistencies. I understand this is something useful for workerd itself to support but probably not for mixed compatibility.

Not quite sure I'm following. Are you saying the runtime should *not* expose `isBuiltin()`?

src/node/internal/module.d.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+export function createRequire(path: string): any;
```
vicb, line 5:
```suggestion
export function createRequire(path: string): (specifier: string) => any;
```

vicb:
> Not quite sure I'm following. Are you saying the runtime should _not_ expose `isBuiltin()`?

I think `workerd` should expose it but `unenv` will patch it to return `true` for the module it polyfills.
With `unenv` the `workerd` impl is available at `process.getBuiltinModule('module').isBuiltin`

Is that correct @pi0 ?

jasnell:
@vicb ... I'm just a bit confused because `module.isBuiltin(...)` should return `true` for all of the known Node.js built-in specifiers, not just the ones that are built into the runtime. For instance, in this implementation `isBuiltin('fs')` would return true even tho it is not implemented in the runtime.

vicb:
> in this implementation `isBuiltin('fs')` would return true even tho it is not implemented in the runtime.

Now I'm confused too 🤔

vicb:
```js
import module from "node:module";

for (const m of modules()) {
  if (!module.isBuiltin(m)) {
    console.error(m);
  }
}

function modules() {
  // return the output of module.builtInModules as of Node.js v22.6.0 minus "internal/..." packages
}
```

```
-> % node test.mjs
test
-> % deno run test.mjs 
test
-> % bun test.mjs
test
```

Noe that `test` is (kind of) expected per the [Node.js doc](https://nodejs.org/docs/latest/api/module.html#modulebuiltinmodules)

vicb:
Would this comment

> If we have .isBuiltin exposed, it would be nice to also have .builtinModules

be easy to add to the PR?

jasnell:
> ... Noe that test is (kind of) expected per the [Node.js doc](https://nodejs.org/docs/latest/api/module.html#modulebuiltinmodules)

It's not clear what you're saying in this comment.

image

> If we have .isBuiltin exposed, it would be nice to also have .builtinModules

It would be fairly trivial yes.

vicb:
> > ... Noe that test is (kind of) expected per the [Node.js doc](https://nodejs.org/docs/latest/api/module.html#modulebuiltinmodules)
> 
> It's not clear what you're saying in this comment.

The docs says "Note: the list doesn't contain [prefix-only modules](https://nodejs.org/docs/latest/api/modules.html#built-in-modules-with-mandatory-node-prefix) like node:test."

I interpreted that as `test` being "special", implying that `isBuiltin(test) === false` is expected - "kind of" because I'm unsure if my interpretation is correct.

jasnell:
`require('module').builtinModules` added.

jasnell:
Marking this ready for review

vicb:
A few comments:

Node 22.6 had 2 extras builtin modules:  `test/mock_loader` and `wasi`

I got my [previous cross runtime](https://github.com/cloudflare/workerd/pull/2636#issuecomment-2326812194) wrong as the reference was `deno` instead of node. By correctly using the node ref, there are missing modules in runtimes:

```
-> % bun test.mjs 
test/mock_loader

-> % deno run test.mjs 
_http_client
_http_common
_http_incoming
_http_server
_stream_wrap
_tls_common
_tls_wrap
inspector/promises
test/mock_loader
trace_events
```

Which means that `.isBuiltin()` behavior is not consistent across runtimes. At least `deno` returns what it actually implements.

`test/mock_loader` might not have made it to the list (see this [bug](https://github.com/nodejs/node/issues/54071)).

I think adding `wasi` to current PR and clearly document the behavior of `workerd` which is isBuiltin() === implemented is Node is good to me

src/node/module.ts
```diff
@@ -0,0 +1,112 @@
...
+  'timers/promises',
+  'tls',
+  'trace_events',
+  'tty',
+  'url',
+  'util',
+  'util/types',
+  'v8',
+  'vm',
+  'worker_threads',
```
vicb, line 103: add `wasi`?

src/node/module.ts
```diff
@@ -0,0 +1,112 @@
...
+      path,
+      'The argument must be a file URL object, ' +
+        'a file URL string, or an absolute path string.'
+    );
+  }
+
+  return moduleUtil.createRequire(path as string);
+}
+
+export function isBuiltin(specifier: string): boolean {
```
vicb, line 32: Clearly document what the return value means

src/node/module.ts
```diff
@@ -0,0 +1,112 @@
...
+  return moduleUtil.isBuiltin(specifier);
+}
+
+// Intentionally does not include 'node:' prefixed-modules like `node:test`
```
vicb, line 36:
```suggestion
// Intentionally does not include modules with mandatory `node:` prefix like `node:test`.
// See https://nodejs.org/docs/latest/api/modules.html#built-in-modules-with-mandatory-node-prefix
```

vicb (changes requested):
LGTM with a few minor comments

vicb (approved):

vicb:
Thanks for the updates James

anonrig (approved):

jasnell:
Running an internal CI run before landing.

PR #2638, opened 2024-09-01
Remove response stream disconnected status gate

joshthoward:


jqmmes (approved):

jasnell (approved):

jclee:
Thanks!  This looks good to me, but probably worth keeping in mind that the internal `web-socket-proxy` test in edgeworker currently covers the behavior difference between the two autogate settings, so it would also need to be updated at the same time as the workerd dependency that removes the conditional behavior.  I believe it is simply a matter of merging the versions of the tests with `all-autogates` and `not-all-autogates` test results (I think the relevant tests can be found by grepping for `RESPONSE_STREAM_DISCONNECTED_STATUS`).  Let me know if you'd like me to help.

jclee (approved):

joshthoward:
The internal PR is passing checks and approved. I am going to go ahead and merge, update the internal PR to point to the merge commit, and then merge the internal PR. Please LMK if that isn't the right process.

PR #2648, opened 2024-09-04
Store alarm value in sqlite database, for sqlite-backed DOs

jclee:
Implements the setting of the alarm value in the sqlite database, when setAlarm()/getAlarm() methods are called.

Also implements `armAlarmHandler()` and `cancelDeferredDelete()` internally, in a manner similar to ActorCache, rather than calling out to the hook functions.  Currently, only the `setAlarm()` hook function is used, and just to request notification from the AlarmScheduler.

Still need to implement proper sequencing of alarm update vs. commitCallback, but wanted to see if the general approach is OK.

### DIFF
--- a/src/workerd/api/actor-alarms-test.js
+++ b/src/workerd/api/actor-alarms-test.js
@@ -11,12 +11,12 @@ export class DurableObjectExample {
 
   async waitForAlarm(scheduledTime) {
     let self = this;
-    let prom = new Promise((resolve) => {
-      self.resolve = resolve;
-    });
+    const { promise, resolve, reject } = Promise.withResolvers();
+    self.resolve = resolve;
+    self.reject = reject;
 
     try {
-      await prom;
+      await promise;
       if (Date.now() < scheduledTime.valueOf()) {
         throw new Error(
           `Date.now() is before scheduledTime! ${Date.now()} vs ${scheduledTime.valueOf()}`
@@ -45,11 +45,15 @@ export class DurableObjectExample {
   }
 
   async alarm() {
-    let time = await this.state.storage.getAlarm();
-    if (time) {
-      throw new Error(`time not null inside alarm handler ${time}`);
+    try {
+      let time = await this.state.storage.getAlarm();
+      if (time !== null) {
+        throw new Error(`time not null inside alarm handler ${time}`);
+      }
+      this.resolve();
+    } catch (e) {
+      this.reject(e);
     }
-    this.resolve();
   }
 }
 
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -385,112 +385,73 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
   auto& context = IoContext::current();
   auto& actor = KJ_ASSERT_NONNULL(context.getActor());
   auto& persistent = KJ_ASSERT_NONNULL(actor.getPersistent());
-  auto maybeDeferredDelete = persistent.armAlarmHandler(scheduledTime);
 
-  KJ_IF_SOME(deferredDelete, maybeDeferredDelete) {
-    auto& handler = KJ_REQUIRE_NONNULL(exportedHandler);
-    if (handler.alarm == kj::none) {
+  KJ_SWITCH_ONEOF(persistent.armAlarmHandler(scheduledTime)) {
+    KJ_CASE_ONEOF(armResult, ActorCacheInterface::RunAlarmHandler) {
+      auto& handler = KJ_REQUIRE_NONNULL(exportedHandler);
+      if (handler.alarm == kj::none) {
 
-      lock.logWarningOnce("Attempted to run a scheduled alarm without a handler, "
-                          "did you remember to export an alarm() function?");
-      return WorkerInterface::AlarmResult{
-        .retry = false, .outcome = EventOutcome::SCRIPT_NOT_FOUND};
-    }
+        lock.logWarningOnce("Attempted to run a scheduled alarm without a handler, "
+                            "did you remember to export an alarm() function?");
+        return WorkerInterface::AlarmResult{
+          .retry = false, .outcome = EventOutcome::SCRIPT_NOT_FOUND};
+      }
 
-    auto& alarm = KJ_ASSERT_NONNULL(handler.alarm);
+      auto& alarm = KJ_ASSERT_NONNULL(handler.alarm);
 
-    return context
-        .run([exportedHandler, &context, timeout, retryCount, &alarm,
-                 maybeAsyncContext = jsg::AsyncContextFrame::currentRef(lock)](
-                 Worker::Lock& lock) mutable -> kj::Promise<WorkerInterface::AlarmResult> {
-      jsg::AsyncContextFrame::Scope asyncScope(lock, maybeAsyncContext);
-      // We want to limit alarm handler walltime to 15 minutes at most. If the timeout promise
-      // completes we want to cancel the alarm handler. If the alarm handler promise completes first
-      // timeout will be canceled.
-      auto timeoutPromise = context.afterLimitTimeout(timeout).then(
-          [&context]() -> kj::Promise<WorkerInterface::AlarmResult> {
-        // We don't want to delete the alarm since we have not successfully completed the alarm
-        // execution.
+      return context
+          .run([exportedHandler, &context, timeout, retryCount, &alarm,
+                   maybeAsyncContext = jsg::AsyncContextFrame::currentRef(lock)](
+                   Worker::Lock& lock) mutable -> kj::Promise<WorkerInterface::AlarmResult> {
+        jsg::AsyncContextFrame::Scope asyncScope(lock, maybeAsyncContext);
+        // We want to limit alarm handler walltime to 15 minutes at most. If the timeout promise
+        // completes we want to cancel the alarm handler. If the alarm handler promise completes first
+        // timeout will be canceled.
+        auto timeoutPromise = context.afterLimitTimeout(timeout).then(
+            [&context]() -> kj::Promise<WorkerInterface::AlarmResult> {
+          // We don't want to delete the alarm since we have not successfully completed the alarm
+          // execution.
+          auto& actor = KJ_ASSERT_NONNULL(context.getActor());
+          auto& persistent = KJ_ASSERT_NONNULL(actor.getPersistent());
+          persistent.cancelDeferredAlarmDeletion();
+
+          LOG_NOSENTRY(WARNING, "Alarm exceeded its allowed execution time");
+          // Report alarm handler failure and log it.
+          auto e = KJ_EXCEPTION(OVERLOADED,
+              "broken.dropped; worker_do_not_log; jsg.Error: Alarm exceeded its allowed execution time");
+          context.getMetrics().reportFailure(e);
+
+          // We don't want the handler to keep running after timeout.
+          context.abort(kj::mv(e));
+          // We want timed out alarms to be treated as user errors. As such, we'll mark them as
+          // retriable, and we'll count the retries against the alarm retries limit. This will ensure
+          // that the handler will attempt to run for a number of times before giving up and deleting
+          // the alarm.
+          return WorkerInterface::AlarmResult{
+            .retry = true, .retryCountsAgainstLimit = true, .outcome = EventOutcome::EXCEEDED_CPU};
+        });
+
+        return alarm(lock, jsg::alloc<AlarmInvocationInfo>(retryCount))
+            .then([]() -> kj::Promise<WorkerInterface::AlarmResult> {
+          return WorkerInterface::AlarmResult{.retry = false, .outcome = EventOutcome::OK};
+        }).exclusiveJoin(kj::mv(timeoutPromise));
+      })
+          .catch_([&context, deferredDelete = kj::mv(armResult.deferredDelete)](
+                      kj::Exception&& e) mutable {
         auto& actor = KJ_ASSERT_NONNULL(context.getActor());
         auto& persistent = KJ_ASSERT_NONNULL(actor.getPersistent());
         persistent.cancelDeferredAlarmDeletion();
 
-        LOG_NOSENTRY(WARNING, "Alarm exceeded its allowed execution time");
-        // Report alarm handler failure and log it.
-        auto e = KJ_EXCEPTION(OVERLOADED,
-            "broken.dropped; worker_do_not_log; jsg.Error: Alarm exceeded its allowed execution time");
         context.getMetrics().reportFailure(e);
 
-        // We don't want the handler to keep running after timeout.
-        context.abort(kj::mv(e));
-        // We want timed out alarms to be treated as user errors. As such, we'll mark them as
-        // retriable, and we'll count the retries against the alarm retries limit. This will ensure
-        // that the handler will attempt to run for a number of times before giving up and deleting
-        // the alarm.
-        return WorkerInterface::AlarmResult{
-          .retry = true, .retryCountsAgainstLimit = true, .outcome = EventOutcome::EXCEEDED_CPU};
-      });
+        // This will include the error in inspector/tracers and log to syslog if internal.
+        context.logUncaughtExceptionAsync(UncaughtExceptionSource::ALARM_HANDLER, kj::mv(e));
 
-      return alarm(lock, jsg::alloc<AlarmInvocationInfo>(retryCount))
-          .then([]() -> kj::Promise<WorkerInterface::AlarmResult> {
-        return WorkerInterface::AlarmResult{.retry = false, .outcome = EventOutcome::OK};
-      }).exclusiveJoin(kj::mv(timeoutPromise));
-    })
-        .catch_([&context, deferredDelete = kj::mv(deferredDelete)](kj::Exception&& e) mutable {
-      auto& actor = KJ_ASSERT_NONNULL(context.getActor());
-      auto& persistent = KJ_ASSERT_NONNULL(actor.getPersistent());
-      persistent.cancelDeferredAlarmDeletion();
-
-      context.getMetrics().reportFailure(e);
-
-      // This will include the error in inspector/tracers and log to syslog if internal.
-      context.logUncaughtExceptionAsync(UncaughtExceptionSource::ALARM_HANDLER, kj::mv(e));
-
-      EventOutcome outcome = EventOutcome::EXCEPTION;
-      KJ_IF_SOME(status, context.getLimitEnforcer().getLimitsExceeded()) {
-        outcome = status;
-      }
-
-      kj::String actorId;
-      KJ_SWITCH_ONEOF(actor.getId()) {
-        KJ_CASE_ONEOF(f, kj::Own<ActorIdFactory::ActorId>) {
-          actorId = f->toString();
+        EventOutcome outcome = EventOutcome::EXCEPTION;
+        KJ_IF_SOME(status, context.getLimitEnforcer().getLimitsExceeded()) {
+          outcome = status;
         }
-        KJ_CASE_ONEOF(s, kj::String) {
-          actorId = kj::str(s);
-        }
-      }
 
-      // We only want to retry against limits if it's a user error. By default let's check if the
-      // output gate is broken.
-      auto shouldRetryCountsAgainstLimits = !context.isOutputGateBroken();
-
-      // We want to alert if we aren't going to count this alarm retry against limits
-      if (auto desc = e.getDescription(); !jsg::isTunneledException(desc) &&
-          !jsg::isDoNotLogException(desc) && context.isOutputGateBroken()) {
-        LOG_NOSENTRY(ERROR, "output lock broke during alarm execution", actorId, e);
-      } else if (context.isOutputGateBroken()) {
-        // We don't usually log these messages, but it's useful to know the real reason we failed
-        // to correctly investigate stuck alarms.
-        LOG_NOSENTRY(ERROR,
-            "output lock broke during alarm execution without an interesting error description",
-            actorId, e);
-        if (e.getDetail(jsg::EXCEPTION_IS_USER_ERROR) != kj::none) {
-          // The handler failed because the user overloaded the object. It's their fault, we'll not
-          // retry forever.
-          shouldRetryCountsAgainstLimits = true;
-        }
-      }
-      return WorkerInterface::AlarmResult{.retry = true,
-        .retryCountsAgainstLimit = shouldRetryCountsAgainstLimits,
-        .outcome = outcome};
-    })
-        .then(
-            [&context](
-                WorkerInterface::AlarmResult result) -> kj::Promise<WorkerInterface::AlarmResult> {
-      return context.waitForOutputLocks().then(
-          [result]() { return kj::mv(result); }, [&context](kj::Exception&& e) {
-        auto& actor = KJ_ASSERT_NONNULL(context.getActor());
         kj::String actorId;
         KJ_SWITCH_ONEOF(actor.getId()) {
           KJ_CASE_ONEOF(f, kj::Own<ActorIdFactory::ActorId>) {
@@ -500,21 +461,20 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
             actorId = kj::str(s);
           }
         }
-        // We only want to retry against limits if it's a user error. By default let's assume it's our
-        // fault.
-        auto shouldRetryCountsAgainstLimits = false;
-        if (auto desc = e.getDescription();
-            !jsg::isTunneledException(desc) && !jsg::isDoNotLogException(desc)) {
-          if (isInterestingException(e)) {
-            LOG_EXCEPTION("alarmOutputLock"_kj, e);
-          } else {
-            LOG_NOSENTRY(ERROR, "output lock broke after executing alarm", actorId, e);
-          }
-        } else {
+
+        // We only want to retry against limits if it's a user error. By default let's check if the
+        // output gate is broken.
+        auto shouldRetryCountsAgainstLimits = !context.isOutputGateBroken();
+
+        // We want to alert if we aren't going to count this alarm retry against limits
+        if (auto desc = e.getDescription(); !jsg::isTunneledException(desc) &&
+            !jsg::isDoNotLogException(desc) && context.isOutputGateBroken()) {
+          LOG_NOSENTRY(ERROR, "output lock broke during alarm execution", actorId, e);
+        } else if (context.isOutputGateBroken()) {
           // We don't usually log these messages, but it's useful to know the real reason we failed
           // to correctly investigate stuck alarms.
           LOG_NOSENTRY(ERROR,
-              "output lock broke after executing alarm without an interesting error description",
+              "output lock broke during alarm execution without an interesting error description",
               actorId, e);
           if (e.getDetail(jsg::EXCEPTION_IS_USER_ERROR) != kj::none) {
             // The handler failed because the user overloaded the object. It's their fault, we'll not
@@ -524,12 +484,57 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
         }
         return WorkerInterface::AlarmResult{.retry = true,
           .retryCountsAgainstLimit = shouldRetryCountsAgainstLimits,
-          .outcome = EventOutcome::EXCEPTION};
+          .outcome = outcome};
+      })
+          .then([&context](WorkerInterface::AlarmResult result)
+                    -> kj::Promise<WorkerInterface::AlarmResult> {
+        return context.waitForOutputLocks().then(
+            [result]() { return kj::mv(result); }, [&context](kj::Exception&& e) {
+          auto& actor = KJ_ASSERT_NONNULL(context.getActor());
+          kj::String actorId;
+          KJ_SWITCH_ONEOF(actor.getId()) {
+            KJ_CASE_ONEOF(f, kj::Own<ActorIdFactory::ActorId>) {
+              actorId = f->toString();
+            }
+            KJ_CASE_ONEOF(s, kj::String) {
+              actorId = kj::str(s);
+            }
+          }
+          // We only want to retry against limits if it's a user error. By default let's assume it's our
+          // fault.
+          auto shouldRetryCountsAgainstLimits = false;
+          if (auto desc = e.getDescription();
+              !jsg::isTunneledException(desc) && !jsg::isDoNotLogException(desc)) {
+            if (isInterestingException(e)) {
+              LOG_EXCEPTION("alarmOutputLock"_kj, e);
+            } else {
+              LOG_NOSENTRY(ERROR, "output lock broke after executing alarm", actorId, e);
+            }
+          } else {
+            // We don't usually log these messages, but it's useful to know the real reason we failed
+            // to correctly investigate stuck alarms.
+            LOG_NOSENTRY(ERROR,
+                "output lock broke after executing alarm without an interesting error description",
+                actorId, e);
+            if (e.getDetail(jsg::EXCEPTION_IS_USER_ERROR) != kj::none) {
+              // The handler failed because the user overloaded the object. It's their fault, we'll not
+              // retry forever.
+              shouldRetryCountsAgainstLimits = true;
+            }
+          }
+          return WorkerInterface::AlarmResult{.retry = true,
+            .retryCountsAgainstLimit = shouldRetryCountsAgainstLimits,
+            .outcome = EventOutcome::EXCEPTION};
+        });
       });
-    });
-  } else {
-    return WorkerInterface::AlarmResult{.retry = false, .outcome = EventOutcome::CANCELED};
+    }
+    KJ_CASE_ONEOF(armResult, ActorCacheInterface::CancelAlarmHandler) {
+      return armResult.waitBeforeCancel.then([]() {
+        return WorkerInterface::AlarmResult{.retry = false, .outcome = EventOutcome::CANCELED};
+      });
+    }
   }
+  KJ_UNREACHABLE;
 }
 
 jsg::Promise<void> ServiceWorkerGlobalScope::test(
--- a/src/workerd/io/actor-sqlite-test.c++
+++ b/src/workerd/io/actor-sqlite-test.c++
@@ -1,0 +1,1140 @@
+// Copyright (c) 2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "actor-sqlite.h"
+#include "io-gate.h"
+
+#include <workerd/util/capnp-mock.h>
+#include <workerd/util/test.h>
+
+#include <kj/debug.h>
+#include <kj/test.h>
+
+namespace workerd {
+namespace {
+
+static constexpr kj::Date oneMs = 1 * kj::MILLISECONDS + kj::UNIX_EPOCH;
+static constexpr kj::Date twoMs = 2 * kj::MILLISECONDS + kj::UNIX_EPOCH;
+static constexpr kj::Date threeMs = 3 * kj::MILLISECONDS + kj::UNIX_EPOCH;
+static constexpr kj::Date fourMs = 4 * kj::MILLISECONDS + kj::UNIX_EPOCH;
+static constexpr kj::Date fiveMs = 5 * kj::MILLISECONDS + kj::UNIX_EPOCH;
+
+template <typename T>
+kj::Promise<T> eagerlyReportExceptions(kj::Promise<T> promise, kj::SourceLocation location = {}) {
+  return promise.eagerlyEvaluate([location](kj::Exception&& e) -> T {
+    KJ_LOG_AT(ERROR, location, e);
+    kj::throwFatalException(kj::mv(e));
+  });
+}
+
+// Expect that a synchronous result is returned.
+template <typename T>
+T expectSync(kj::OneOf<T, kj::Promise<T>> result, kj::SourceLocation location = {}) {
+  KJ_SWITCH_ONEOF(result) {
+    KJ_CASE_ONEOF(promise, kj::Promise<T>) {
+      KJ_FAIL_ASSERT_AT(location, "result was unexpectedly asynchronous");
+    }
+    KJ_CASE_ONEOF(value, T) {
+      return kj::mv(value);
+    }
+  }
+  KJ_UNREACHABLE;
+}
+
+struct ActorSqliteTestOptions final {
+  bool monitorOutputGate = true;
+};
+
+struct ActorSqliteTest final {
+  kj::EventLoop loop;
+  kj::WaitScope ws;
+
+  OutputGate gate;
+  kj::Own<const kj::Directory> vfsDir;
+  SqliteDatabase::Vfs vfs;
+  SqliteDatabase db;
+
+  struct Call final {
+    kj::String desc;
+    kj::Own<kj::PromiseFulfiller<void>> fulfiller;
+  };
+  kj::Vector<Call> calls;
+
+  struct ActorSqliteTestHooks final: public ActorSqlite::Hooks {
+  public:
+    explicit ActorSqliteTestHooks(ActorSqliteTest& parent): parent(parent) {}
+
+    kj::Promise<void> scheduleRun(kj::Maybe<kj::Date> newAlarmTime) override {
+      KJ_IF_SOME(h, parent.scheduleRunHandler) {
+        return h(newAlarmTime);
+      }
+      auto desc = newAlarmTime.map([](auto& t) {
+        return kj::str("scheduleRun(", t, ")");
+      }).orDefault(kj::str("scheduleRun(none)"));
+      auto [promise, fulfiller] = kj::newPromiseAndFulfiller<void>();
+      parent.calls.add(Call{kj::mv(desc), kj::mv(fulfiller)});
+      return kj::mv(promise);
+    }
+
+    ActorSqliteTest& parent;
+  };
+  kj::Maybe<kj::Function<kj::Promise<void>(kj::Maybe<kj::Date>)>> scheduleRunHandler;
+  ActorSqliteTestHooks hooks = ActorSqliteTestHooks(*this);
+
+  ActorSqlite actor;
+
+  kj::Promise<void> gateBrokenPromise;
+  kj::UnwindDetector unwindDetector;
+
+  explicit ActorSqliteTest(ActorSqliteTestOptions options = {})
+      : ws(loop),
+        vfsDir(kj::newInMemoryDirectory(kj::nullClock())),
+        vfs(*vfsDir),
+        db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY),
+        actor(kj::attachRef(db), gate, KJ_BIND_METHOD(*this, commitCallback), hooks),
+        gateBrokenPromise(options.monitorOutputGate ? eagerlyReportExceptions(gate.onBroken())
+                                                    : kj::Promise<void>(kj::READY_NOW)) {}
+
+  ~ActorSqliteTest() noexcept(false) {
+    if (!unwindDetector.isUnwinding()) {
+      // Make sure if the output gate has been broken, the exception was reported. This is
+      // important to report errors thrown inside flush(), since those won't otherwise propagate
+      // into the test body.
+      gateBrokenPromise.poll(ws);
+
+      // Make sure there's no outstanding async work we haven't considered:
+      pollAndExpectCalls({}, "unexpected calls at end of test");
+    }
+  }
+
+  kj::Promise<void> commitCallback() {
+    auto [promise, fulfiller] = kj::newPromiseAndFulfiller<void>();
+    calls.add(Call{kj::str("commit"), kj::mv(fulfiller)});
+    return kj::mv(promise);
+  }
+
+  // Polls the event loop, then asserts that the description of calls up to this point match the
+  // expectation and returns their fulfillers.  Also clears the call log.
+  //
+  // TODO(cleanup): Is there a better way to do mocks?  capnp-mock looks nice, but seems a bit
+  // heavyweight for this test.
+  kj::Vector<kj::Own<kj::PromiseFulfiller<void>>> pollAndExpectCalls(
+      std::initializer_list<kj::StringPtr> expCallDescs,
+      kj::StringPtr message = ""_kj,
+      kj::SourceLocation location = {}) {
+    ws.poll();
+    auto callDescs = KJ_MAP(c, calls) { return kj::str(c.desc); };
+    KJ_ASSERT_AT(callDescs == heapArray(expCallDescs), location, kj::str(message));
+    auto fulfillers = KJ_MAP(c, calls) { return kj::mv(c.fulfiller); };
+    calls.clear();
+    return kj::mv(fulfillers);
+  }
+
+  // A few driver methods for convenience.
+  auto get(kj::StringPtr key, ActorCache::ReadOptions options = {}) {
+    return actor.get(kj::str(key), options);
+  }
+  auto getAlarm(ActorCache::ReadOptions options = {}) {
+    return actor.getAlarm(options);
+  }
+  auto put(kj::StringPtr key, kj::StringPtr value, ActorCache::WriteOptions options = {}) {
+    return actor.put(kj::str(key), kj::heapArray(value.asBytes()), options);
+  }
+  auto setAlarm(kj::Maybe<kj::Date> newTime, ActorCache::WriteOptions options = {}) {
+    return actor.setAlarm(newTime, options);
+  }
+};
+
+KJ_TEST("initial alarm value is unset") {
+  ActorSqliteTest test;
+
+  KJ_ASSERT(expectSync(test.getAlarm()) == kj::none);
+}
+
+KJ_TEST("can set and get alarm") {
+  ActorSqliteTest test;
+
+  test.setAlarm(oneMs);
+  test.pollAndExpectCalls({"scheduleRun(1ms)"})[0]->fulfill();
+  test.pollAndExpectCalls({"commit"})[0]->fulfill();
+
+  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+}
+
+KJ_TEST("alarm write happens transactionally with storage ops") {
+  ActorSqliteTest test;
+
+  test.setAlarm(oneMs);
+  test.put("foo", "bar");
+  test.pollAndExpectCalls({"scheduleRun(1ms)"})[0]->fulfill();
+  test.pollAndExpectCalls({"commit"})[0]->fulfill();
+
+  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+  KJ_ASSERT(KJ_ASSERT_NONNULL(expectSync(test.get("foo"))) == kj::str("bar").asBytes());
+}
+
+KJ_TEST("storage op without alarm change does not wait on scheduler") {
+  ActorSqliteTest test;
+
+  test.put("foo", "bar");
+  test.pollAndExpectCalls({"commit"})[0]->fulfill();
+
+  KJ_ASSERT(KJ_ASSERT_NONNULL(expectSync(test.get("foo"))) == kj::str("bar").asBytes());
+  KJ_ASSERT(expectSync(test.getAlarm()) == kj::none);
+}
+
+KJ_TEST("alarm scheduling starts synchronously before implicit local db commit") {
+  ActorSqliteTest test;
+
+  // In workerd (unlike edgeworker), there is no remote storage, so there is no work done in
+  // commitCallback(); the local db is considered durably stored after the synchronous sqlite
+  // commit() call returns.  If a commit includes an alarm state change that requires scheduling
+  // before the commit call, it needs to happen synchronously.  Since workerd synchronously
+  // schedules alarms, we just need to ensure that the database is in a pre-commit state when
+  // scheduleRun() is called.
+
+  // Initialize alarm state to 2ms.
+  test.setAlarm(twoMs);
+  test.pollAndExpectCalls({"scheduleRun(2ms)"})[0]->fulfill();
+  test.pollAndExpectCalls({"commit"})[0]->fulfill();
+  test.pollAndExpectCalls({});
+
+  bool startedScheduleRun = false;
+  test.scheduleRunHandler = [&](kj::Maybe<kj::Date>) -> kj::Promise<void> {
+    startedScheduleRun = true;
+
+    KJ_EXPECT_THROW_MESSAGE(
+        "cannot start a transaction within a transaction", test.db.run("BEGIN TRANSACTION"));
+
+    return kj::READY_NOW;
+  };
+
+  test.setAlarm(oneMs);
+  KJ_ASSERT(!startedScheduleRun);
+  test.ws.poll();
+  KJ_ASSERT(startedScheduleRun);
+
+  test.pollAndExpectCalls({"commit"})[0]->fulfill();
+
+  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+}
+
+KJ_TEST("alarm scheduling starts synchronously before explicit local db commit") {
+  ActorSqliteTest test;
+
+  // Initialize alarm state to 2ms.
+  test.setAlarm(twoMs);
+  test.pollAndExpectCalls({"scheduleRun(2ms)"})[0]->fulfill();
+  test.pollAndExpectCalls({"commit"})[0]->fulfill();
+  test.pollAndExpectCalls({});
+
+  bool startedScheduleRun = false;
+  test.scheduleRunHandler = [&](kj::Maybe<kj::Date>) -> kj::Promise<void> {
+    startedScheduleRun = true;
+
+    // Not sure if there is a good way to detect savepoint presence without mutating the db state,
+    // but this is sufficient to verify the test properties:
+
+    // Verify that we are not within a nested savepoint.
+    KJ_EXPECT_THROW_MESSAGE(
+        "no such savepoint: _cf_savepoint_1", test.db.run("RELEASE _cf_savepoint_1"));
+
+    // Verify that we are within the root savepoint.
+    test.db.run("RELEASE _cf_savepoint_0");
+    KJ_EXPECT_THROW_MESSAGE(
+        "no such savepoint: _cf_savepoint_0", test.db.run("RELEASE _cf_savepoint_0"));
+
+    // We don't actually care what happens in the test after this point, but it's slightly simpler
+    // to readd the savepoint to allow the test to complete cleanly:
+    test.db.run("SAVEPOINT _cf_savepoint_0");
+
+    return kj::READY_NOW;
+  };
+
+  {
+    auto txn = test.actor.startTransaction();
+    txn->setAlarm(oneMs, {});
+
+    KJ_ASSERT(!startedScheduleRun);
+    txn->commit();
+    KJ_ASSERT(startedScheduleRun);
+
+    test.pollAndExpectCalls({"commit"})[0]->fulfill();
+  }
+
+  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+}
+
+KJ_TEST("alarm scheduling does not start synchronously before nested explicit local db commit") {
+  ActorSqliteTest test;
+
+  // Initialize alarm state to 2ms.
+  test.setAlarm(twoMs);
+  test.pollAndExpectCalls({"scheduleRun(2ms)"})[0]->fulfill();
+  test.pollAndExpectCalls({"commit"})[0]->fulfill();
+  test.pollAndExpectCalls({});
+
+  bool startedScheduleRun = false;
+  test.scheduleRunHandler = [&](kj::Maybe<kj::Date>) -> kj::Promise<void> {
+    startedScheduleRun = true;
+    return kj::READY_NOW;
+  };
+
+  {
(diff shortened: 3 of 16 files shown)


src/workerd/util/sqlite-metadata.c++
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+SqliteMetadata::Initialized& SqliteMetadata::ensureInitialized() {
+  KJ_SWITCH_ONEOF(state) {
+    KJ_CASE_ONEOF(uninitialized, Uninitialized) {
+      auto& db = uninitialized.db;
+
+      db.run(R"(
+        CREATE TABLE IF NOT EXISTS _cf_METADATA (
```
kentonv, line 49: I just merged #2680 which implements `deleteAll()` such that it deletes and recreates the entire database. That means after a `deleteAll()`, the _cf_METADATA table will be gone and needs to be recreated on-demand. You'll have to update this code to account for that, I guess.

Which raises the question: What should happen to the alarm on deleteAll()? With the old storage backend, the alarm is not actually deleted. To emulate that behavior, I suppose we have to recreate the alarm immediately, if it exists. We probably do need to emulate the behavior as we want to be able to migrate people transparently someday.

Relatedly, though, if an alarm is set on an object that has no other data in storage (whether because deleteAll() was called, or maybe nothing else was ever stored in the first place), and that alarm eventually runs and doesn't store anything nor set a new alarm... we need to make sure the object can be cleaned up. That won't happen unless we actually do another `deleteAll()` (or `db.reset()`) at that point.

We might need to add some code which runs whenever an alarm is unset, which checks if the database is completely empty, and resets it if it is...

That can probably be a separate PR though.

src/workerd/io/actor-sqlite.h
```diff
@@ -156,13 +161,57 @@ class ActorSqlite final: public ActorCacheInterface, private kj::TaskSet::ErrorH
...
+  struct UnknownAlarmTime {};
+  struct KnownAlarmTime {
+    enum class Status {
+      CLEAN,    // Alarm time has been committed
+      DIRTY,    // Alarm time has been changed and not yet committed
+      FLUSHING  // Alarm time is being committed in an ongoing transaction
+    } status;
+    kj::Maybe<kj::Date> time;
+  };
+  struct DeferredAlarmDelete {
```
kentonv, line 184: Could use some more commentary here on what this is for. It took me a while to realize this is when we're currently running the alarm handler, and we need to delete the alarm at the end of the handler unless it is changed during the handler...

src/workerd/io/actor-sqlite.c++
```diff
@@ -157,11 +158,68 @@ void ActorSqlite::onWrite() {
...
+    KJ_CASE_ONEOF(deferredDelete, DeferredAlarmDelete) {
+      if (deferredDelete.status == DeferredAlarmDelete::Status::READY) {
+        deferredDelete.status = DeferredAlarmDelete::Status::FLUSHING;
+        needsAlarmFlush = true;
+        newAlarmTime = kj::none;
+      }
+    }
+  }
+
+  // TODO(soon): ActorCache implements a 4x retry of failed flushes... Do we need anything similar
```
kentonv, line 189: Probably, yes. It's possible that the update will throw DISCONNECTED if it was sent to the server over a connection that was lost during the call. We need to retry in that case.

However, arguably, it could be the hook's responsibility to perform the retry, rather than do it here. In the workerd case, where the hook is implemented entirely locally, there'd be no reason to retry.

src/workerd/io/actor-sqlite.c++
```diff
@@ -157,11 +158,68 @@ void ActorSqlite::onWrite() {
...
+  // for alarm scheduling?
+
+  // We assume that exceptions thrown during commit will propagate to the caller, such that they
+  // will ensure cancelDeferredAlarmDeletion() is called, if necessary.
+
+  if (needsAlarmFlush) {
+    // TODO(soon): fix sequencing of alarm scheduling vs. commitCallback().
+    co_await hooks.setAlarm(newAlarmTime);
```
kentonv, line 197: A couple things missing here:

- We only want to update the alarm time _before_ the commit if the new time is sooner than what was already set. It looks like here you're performing the pre-commit update regardless of which direction the alarm is moving. Of course, this implies that we need to keep track of what the previous value was.
- If, during the call to `setAlarm()`, the alarm value is updated again, and is updated to be sooner than what had been passed to `setAlarm()`, then we need to repeat the `setAlarm()` with the newer value before committing.

src/workerd/io/actor-sqlite.c++
```diff
@@ -157,11 +158,68 @@ void ActorSqlite::onWrite() {
       // rather than after the callback.
       { auto drop = kj::mv(txn); }
 
-      return commitCallback();
+      return commitImpl();
```
kentonv, line 161: Looking at this I realize that the pre-commit alarm update actually needs to happen not just before calling the commit callback, but actually before committing the transaction. The alarm update should become part of the transaction.

(Technically the implicit transaction is only important for correctness in workerd -- in production, SRS provides another layer of safety via the commit callback. But we probably should make this correct for workerd. Also, lumping more stuff into fewer transactions performs better, as sqlite is then able to coalesce writes to the same pages.)

jclee: When you say the pre-commit alarm update needs to happen before committing the transaction, do you mean that we need to asynchronously block for any necessary alarm notification change requests to complete just prior to the above `txn->commit()` call?

If so, does that put us in a situation where we're holding a sqlite transaction open while waiting for async work to complete, like Justin mentioned on the spec?  Or is there a way to avoid that?

kentonv: Yes, that's what I meant.

It should be OK to hold it open. It's already being held open until the `kj::evalLater` above resolves -- arbitrary async stuff could have happened in the meantime. It's possible that more JavaScript code will run, even, in which case changes made in that code will be merged into the same implicit transaction. Again, that'll all fine: the point of the implicit transaction is only to make sure that two queries executed in sequence without an await are applied atomically, but there's no reason that it _can't_ extend that atomicity further, such that it actually crosses an await.

There is some complexity if the application requests an explicit asynchronous transaction in the meantime. When that happens we normally commit any implicit transaction immediately -- see the constructor of ExplicitTransaction. Perhaps the way we handle this is: Before committing the implicit transaction to make way for an explicit one, if an alarm update is needed, we at least _start_ the `setAlarm()` call, but we commit the transaction without waiting for it. In workerd, this works because `setAlarm()` updates the alarm synchronously, so by the time we commit the transaction, the alarm has in fact been updated. Whereas in production, we rely on SRS commits as the true guardian of atomicity, and the SRS commit won't be sent until after the alarm has successfully udpated.

src/workerd/io/actor-sqlite.c++
a-robinson: Seconded. We need retries somewhere.

src/workerd/io/actor-sqlite.c++
```diff
@@ -177,6 +235,29 @@ void ActorSqlite::requireNotBroken() {
   }
 }
 
+void ActorSqlite::maybeDeleteDeferredAlarm() {
+  KJ_IF_SOME(d, currentAlarmTime.tryGet<DeferredAlarmDelete>()) {
+    // TODO(now): Do we need to do anything special, if we've already started flushing the alarm?
```
a-robinson, line 240: I don't think so, no. I'm not sure how we'd get into that state and then end up back here. In any case, the ActorCache version of this code doesn't do anything special for that, so I wouldn't here either.

src/workerd/io/actor-sqlite.c++
```diff
@@ -177,6 +235,29 @@ void ActorSqlite::requireNotBroken() {
...
+    d.status = DeferredAlarmDelete::Status::READY;
+    // TODO(now): Do we need to do anything special to ensure that the commit code for this sqlite
```
a-robinson, line 242: @kentonv is more familiar with this than me, but it sure looks to me like we'd be better off going through an `ActorSqlite` code path that will ensure this gets flushed properly in accordance with output gates. I would have expected this to go through `ActorSqlite::setAlarm` and/or the code path in `commitImpl`, since as is this looks like it won't do anything more than update the sqlite DB without ever calling out to the `ActorSqliteHooks`.

src/workerd/io/actor-sqlite.c++
```diff
@@ -331,11 +452,52 @@ void ActorSqlite::shutdown(kj::Maybe<const kj::Exception&> maybeException) {
...
+        // the pending write.
+        alarmDeleteNeeded = false;
+      }
+    }
+  }
+
+  // TODO(now): Here, similar to the ActorCache code, we're assuming that if the alarm time
+  // matches but is not in a clean state (i.e. is dirty or flushing), that it's OK to schedule the
+  // delete, and that it's OK to use the current alarm time to initialize the alarm state back to
+  // "clean" if the alarm is cancelled.  Is that OK?
```
a-robinson, line 478: It's apparently not disastrous given that we've been relying on it for quite some time now. I'd allow it, since it seems pretty likely that if we're getting invoked at the given time that the time was probably successfully flushed.

I suppose that the sqlite storage does add a second source of truth that makes this a little more confusing though. But even so, if the alarm time matched, and we successfully ran an alarm invocation at that time, then it seems correct to delete it?

So I think this is fine, but let me know if you see an issue that I'm missing.

src/workerd/io/actor-sqlite.c++
```diff
@@ -331,11 +452,52 @@ void ActorSqlite::shutdown(kj::Maybe<const kj::Exception&> maybeException) {
...
+  static const DeferredAlarmDeleter disposer;
+  return kj::Own<void>(this, disposer);
 }
 
 void ActorSqlite::cancelDeferredAlarmDeletion() {
-  hooks.cancelDeferredAlarmDeletion();
+  if (currentAlarmTime.is<DeferredAlarmDelete>()) {
+    // TODO(now): Should we do anything special if client tries to cancel alarm deletion after
+    // flush has already been issued?
+    // TODO(now): Is it correct to reset alarm to CLEAN state here?
```
a-robinson, line 495: As discussed above, yeah I think it's correct.

src/workerd/io/actor-sqlite.c++
```diff
@@ -331,11 +452,52 @@ void ActorSqlite::shutdown(kj::Maybe<const kj::Exception&> maybeException) {
...
+  }
+
+  static const DeferredAlarmDeleter disposer;
+  return kj::Own<void>(this, disposer);
 }
 
 void ActorSqlite::cancelDeferredAlarmDeletion() {
-  hooks.cancelDeferredAlarmDeletion();
+  if (currentAlarmTime.is<DeferredAlarmDelete>()) {
+    // TODO(now): Should we do anything special if client tries to cancel alarm deletion after
```
a-robinson, line 493: That'd be an invalid use of the API, so hopefully we don't have to worry about it. But reporting an error to sentry if it happens might not be a bad idea.

src/workerd/io/actor-sqlite.h
```diff
@@ -156,13 +161,57 @@ class ActorSqlite final: public ActorCacheInterface, private kj::TaskSet::ErrorH
...
   void requireNotBroken();
+
+  // If alarm is in an uninitialized state, sets alarm time from db state.
+  //
+  // TODO(cleanup): Alternately, we could eliminate the UnknownAlarmTime state and just init the
+  // alarm to a known time at startup, but that requires a db query in the constructor.  Are the
+  // perf savings worth the extra complexity?
+  void ensureAlarmTimeInitialized();
+
+  // Called when alarm handler token object is destroyed, to delete alarm if not reset or
```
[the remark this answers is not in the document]
jclee: Sorry, I meant the `kj::Own<void>` thing that `armAlarmHandler()` returns... not sure if there's a proper name for it.

src/workerd/io/actor-sqlite.c++
jclee: Re: the sqlite storage adding a second source of truth... My original intent was for the sqlite storage to be the source of truth, at least for the local, possibly-not-yet-persisted alarm state, but I wasn't sure if it would be expensive to query sqlite each time we need to check the alarm value, so I'm essentially caching it in the `currentAlarmTime` member.  If it's actually not that expensive to run the query, I think I could make the code a bit simpler.

src/workerd/io/actor-sqlite.c++
jclee: Oh... I had been assuming that the `commitCallback()` had to return for the change to be considered durably persisted, but I guess that's only true in production, and that the workerd local database state is effectively also the workerd durable state.

In which case, do we theoretically need similar logic in the explicit transaction's `commit()` -- i.e. starting any necessary alarm notification request prior to the root transaction's `run("RELEASE _cf_savepoint_"...)`?

src/workerd/io/actor-sqlite.c++
jclee: My understanding is that calling `metadata.setAlarm(kj::none);` will run the sqlite query to update the alarm state in the sqlite database, and that this will cause sqlite to call `ActorSqlite::onWrite()`, which should set up an implicit transaction to be committed on a later run of the event loop, which would then invoke the necessary hooks functions?

But looking at the current implementation of `onWrite()`, it looks like it would only set up an implicit transaction if there is not an ongoing explicit transaction... which I guess means that someone might be able to start an explicit transaction, and any alarm handler invocations that finish during the transaction would not delete their alarms until the explicit transaction was committed.

It seems kind of weird for an object to be holding open an explicit transaction across alarm handler boundaries, but I'm not sure we disallow it, so maybe this case needs to be handled, too?

src/workerd/io/actor-sqlite.c++
a-robinson: I'd think it's fine to just hit sqlite every time. I don't know what @kentonv thinks.

It could have a very minor affect on billing since it would mean getAlarm() counts as one row read, but (A) I think we're going technically going to start causing a row to be read by the first getAlarm even with this caching, and (B) rows read are so preposterously dirt cheap that it's hard to imagine it becoming a big issue. That'll also depend on how #2707 and the related work plays out. I believe the goal was to not start charging for getAlarm, which may require some extra care to achieve :/

src/workerd/io/actor-sqlite.h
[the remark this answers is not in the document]
a-robinson: I'd just call it the `DeferredAlarmDeleter`

src/workerd/io/actor-sqlite.c++
a-robinson: Oh, nice, thanks for calling out how that `onWrite()` callback is hooked up.

I'm not 100% certain on this without digging back in, but it looks like an `ExplicitTxn` is only ever created by the `startTransaction()` function, which is only called from `DurableObjectStorage::transaction`, which wraps the `startTransaction()` call in a `blockConcurrencyWhile()`, which should eliminate any concerns about an alarm handler finishing up while some other request is doing a txn. So I think we're ok here.

And I was going to say I was worried about raw `BEGIN` statements being sent by the user, but then saw that we have the `Regulator` class configured to disallow those from JS. So I think we're ok here.

And on the plus side, if something goes wrong and we end up not deleting an alarm time, we'll just harmlessly re-run the alarm, so it's not the highest of stakes.

src/workerd/api/actor-alarms-test.js
```diff
@@ -11,8 +11,9 @@ export class DurableObjectExample {
 
   async waitForAlarm(scheduledTime) {
     let self = this;
-    let prom = new Promise((resolve) => {
+    let prom = new Promise((resolve, reject) => {
       self.resolve = resolve;
+      self.reject = reject;
     });
```
jasnell, line 17: nit: `const { promise, resolve, reject } = Promise.withResolvers();` is probably a bit cleaner here.

```js
self.resolve = resolve;
self.reject = reject;

// ...

try {
  await promise;
  /// ...
```

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -0,0 +1,440 @@
...
+    ActorSqliteTestHooks(ActorSqliteTest& parent): parent(parent) {}
+
+    kj::Promise<kj::Maybe<kj::Date>> getAlarm() override {
+      parent.calls.add(kj::str("getAlarm"));
+      return kj::Maybe<kj::Date>(kj::none);
+    }
+
+    kj::Promise<void> setAlarm(kj::Maybe<kj::Date> newAlarmTime) override {
+      auto time = newAlarmTime.map([](auto& t) { return kj::str(t); }).orDefault(kj::str("none"));
+      parent.calls.add(kj::str("setAlarm(", time, ")"));
```
jasnell, line 69: minor nit...

```suggestion
      auto time = newAlarmTime.map([](auto& t) {
        return kj::str("setAlarm(", t, ")");
      }).orDefault(kj::str("setAlarm(\"none\")"));
      parent.calls.add(kj::mv(time));
```

Should save an extraneous copy and allocation. Obviously super minor tho.

src/workerd/api/actor-alarms-test.js
jclee: Oh, nice... I'd forgotten that this had been added.

src/workerd/io/actor-sqlite.c++
shrima-cf: With the current implementation of counting rowsRead/rowsWritten, alarms will also be counted. Could you open a ticket under https://jira.cfdata.org/browse/STOR-3551 as a bug so we don't loose track?

shrima-cf: https://jira.cfdata.org/browse/STOR-3744 - will need to be fixed before we enable usage based billing

src/workerd/io/actor-sqlite.c++
jclee: Hmm... that's my intuition of how sqlite uses `onWrite()`, anyway, but thinking more about it, I should probably verify the behavior.  One thing I'm not quite sure of is how it interacts with transactions... I assume that:

- if someone calls `setAlarm()` within a transaction, they don't expect the alarm handler to actually be called until after the transaction is committed (and any transactions enclosing it).  I'm pretty sure the current PR does this, but should probably add a test for it.
- if someone calls `setAlarm()` within a transaction, then rolls back the transaction, they expect the alarm state to revert to the pre-transaction value.  I think the sqlite alarm state will revert correctly in the current PR, but not sure the cached value in `currentAlarmTime` would.  Which I guess is an argument for dropping the caching and just querying the sqlite value directly.  (and of course would also benefit from a test case)

Are these assumptions correct?

> which wraps the startTransaction() call in a blockConcurrencyWhile(), which should eliminate any concerns about an alarm handler finishing up while some other request is doing a txn. So I think we're ok here.

Ah, cool... I was thinking that maybe `blockConcurrencyWhile()` only prevented the start of new JS events, but still allowed the completion of ongoing async C++ promises, and wasn't sure if maybe that would allow things like the alarm handler cleanup to sneak in.

a-robinson: > if someone calls setAlarm() within a transaction, they don't expect the alarm handler to actually be called until after the transaction is committed (and any transactions enclosing it). I'm pretty sure the current PR does this, but should probably add a test for it.

Agreed.

> if someone calls setAlarm() within a transaction, then rolls back the transaction, they expect the alarm state to revert to the pre-transaction value. I think the sqlite alarm state will revert correctly in the current PR, but not sure the cached value in currentAlarmTime would. Which I guess is an argument for dropping the caching and just querying the sqlite value directly. (and of course would also benefit from a test case)

Agreed, and good catch!

> Ah, cool... I was thinking that maybe blockConcurrencyWhile() only prevented the start of new JS events, but still allowed the completion of ongoing async C++ promises, and wasn't sure if maybe that would allow things like the alarm handler cleanup to sneak in.

`blockConcurrencyWhile()` prevents any events from being delivered to JS, including both new incoming requests but also responses to previously issued outgoing I/O. We don't necessarily stop ongoing async C++ work from *happening*, but we do prevent it from waking JS code back up.

src/workerd/io/actor-sqlite.c++
kentonv: I think it's fine to query sqlite again every time and I think it's fine that this is billed as a row read.

DaniFoldi: Hi :wave:,

About half a year ago on Discord it was confirmed by @a-robinson that getAlarm would not be billed: https://discord.com/channels/595317990191398933/773219443911819284/1232360354210648136 - so (without access to the tickets and without knowing your plans) to make sure that people are aware of the change, please include this as a note somewhere in the docs or changelog.

Not knowing pricing (just a guess based on D1) right now, I'm not worried about sudden increases in bills, I'm just asking for this to be documented, please 😁

jclee:
I've reworked the state tracking to rely on alarm db queries for the local state, which is probably less bug-prone.  I've also implemented the sequencing of alarm scheduling updates vs. database commits.

The things that I know still need to be implemented:

* need to fix/test deleteAll interaction.
* can remove unused hook methods, simplify workerd alarm manager
* possibly need tighter check in armAlarmHandler()
* probably should populate lastConfirmedScheduledAlarm from alarm manager on startup
* need to support alarm rescheduling when alarm is early
* eventually, need to handle alarm cleanup when deleting alarm that is the only thing in the db?
* on the edgeworker side, hook up alarm scheduling, implement retries, add test coverage.

a-robinson:
Sorry for needing to ask this, but do you know what I can click on to only see what's changed since the first round of review?

In the options I see, the top "Compare" link appears to be a rebase on master, the bottom "Compare" link appears empty, and [8b607ac](https://github.com/cloudflare/workerd/pull/2648/commits/8b607ac5de00c75f933a951128c97cae35c8d4fd) appears to be a mixture of old changes with new changes.

Screenshot 2024-09-17 at 11 56 14 AM

kentonv:
Hmm... these two fixups couldn't be all the changes, though.

kentonv:
Ahh, there were more fixup-ish commits before the rebase-on-master... I think this is the whole set:

https://github.com/cloudflare/workerd/compare/17c964ac6340f6d877c5aebc79a723c2794195db..e7d6809862baa17f3a6b36ac99593da62d5fa19b

kentonv:
TBH it seems like things changed enough in actor-sqlite.c++ to re-review it from scratch. sqlite-metadata didn't change AFAICT.

jclee:
Ugh...  Sorry, I'd pushed a version of my branch with WIP changes and thought it was OK because the github UI was still showing the commits post-rebase, but I guess it removes them after a little while.  Here's a diff between the state the branch was in when last reviewed (branch jlee/srs-alarms-wip1) and the state after adding my WIP changes but before rebasing (branch jlee/srs-alarms-wip2):

https://github.com/cloudflare/workerd/compare/jlee/srs-alarms-wip1...jlee/srs-alarms-wip2

The current state of the PR is the rebased result plus a couple minor fixes.

(I'm not sure how helpful the interim commits are, as quite a lot changed, particularly in the last commit, but I tried to isolate the more mechanical changes.)

src/workerd/io/actor-sqlite.c++
```diff
@@ -18,8 +36,11 @@ ActorSqlite::ActorSqlite(kj::Own<SqliteDatabase> dbParam,
       commitCallback(kj::mv(commitCallback)),
       hooks(hooks),
       kv(*db),
+      metadata(*db),
       commitTasks(*this) {
   db->onWrite(KJ_BIND_METHOD(*this, onWrite));
+  // TODO(now): populate lastConfirmedScheduledAlarm asynchronously from alarm manager.
```
kentonv, line 42: TODO(now)?

But do we actually need this? We can assume the time in the alarm manager is _no later than_ the time in sqlite. I think if we set `lastConfirmedScheduledAlarm` to the value sqlite, the worst that will happen is we might update the alarm when we didn't need to.

src/workerd/io/actor-sqlite.c++
```diff
@@ -152,6 +180,9 @@ void ActorSqlite::onWrite() {
       // Don't commit if shutdown() has been called.
       requireNotBroken();
 
+      // Start the schedule request before commit(), for correctness in workerd.
+      auto precommitAlarmPromise = startPrecommitAlarmScheduling();
```
kentonv, line 184:
```suggestion
      auto precommitAlarmState = startPrecommitAlarmScheduling();
```

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +200,67 @@ void ActorSqlite::onWrite() {
...
       { auto drop = kj::mv(txn); }
 
-      return commitCallback();
+      return commitImpl(kj::mv(precommitAlarmPromise));
     })));
   }
 }
 
+kj::Promise<void> ActorSqlite::requestScheduledAlarm(kj::Maybe<kj::Date> requestedTime) {
+  co_await hooks.scheduleRun(requestedTime);
```
kentonv, line 209: Hmm I don't remember... if `scheduleRun()` throws synchronously, will `requestScheduleAlarm()` throw synchronously, or does the co_await catch the exception and turn it into a rejected promise?

We need the exception to flow through synchronously in the workerd case (so that the transaction is never committed). It might be best not to use coroutines here, to be safe.

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +200,67 @@ void ActorSqlite::onWrite() {
...
+  return kj::mv(state);
+}
+
+kj::Promise<void> ActorSqlite::commitImpl(ActorSqlite::PrecommitAlarmState precommitAlarmState) {
+  // We assume that exceptions thrown during commit will propagate to the caller, such that they
+  // will ensure cancelDeferredAlarmDeletion() is called, if necessary.
+
+  KJ_IF_SOME(pending, pendingCommit) {
+    co_await pending.addBranch();
+    co_return;
```
kentonv, line 233: I guess we're potentially canceling `precommitAlarmState.schedulingPromise` here -- and worse, it may or may not have actually completed.

But I think this is OK, because the commit we're merging with will have executed the `while` loop below and accounted for any alarm change required by our new commit?

Probably worth commenting to point this out.

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +200,67 @@ void ActorSqlite::onWrite() {
...
+
+  while (willFireEarlier(localAlarmState, lastConfirmedScheduledAlarm)) {
+    co_await requestScheduledAlarm(localAlarmState);
+    localAlarmState = metadata.getAlarm();
+  }
+
+  auto commitCallbackPromise = commitCallback();
+
+  // Fulfill and clear the pending commit.
+  fulfiller->fulfill();
```
kentonv, line 253: I think we actually don't want to fulfill this until the commit callback completes, otherwise we're potentially prematurely unblocking the output gate.

We do still want to set `pendingCommit = none` here so that no _new_ commits merge with this one, but any that have already merged should wait until the commit is actually confirmed.

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +200,67 @@ void ActorSqlite::onWrite() {
...
+  pendingCommit = kj::none;
+  co_await commitCallbackPromise;
+
+  lastConfirmedAlarmDbState = localAlarmState;
+
+  if (willFireEarlier(lastConfirmedScheduledAlarm, localAlarmState)) {
+    co_await requestScheduledAlarm(localAlarmState);
```
kentonv, line 260: Technically we don't have to make the app wait for this.

src/workerd/io/actor-sqlite.c++
```diff
@@ -383,11 +511,37 @@ void ActorSqlite::shutdown(kj::Maybe<const kj::Exception&> maybeException) {
...
-  return hooks.armAlarmHandler(scheduledTime, noCache);
+  KJ_ASSERT(!haveDeferredDelete);
+  KJ_ASSERT(!inAlarmHandler);
+
+  auto localAlarmState = metadata.getAlarm();
+  if (localAlarmState != scheduledTime) {
+    if (localAlarmState == lastConfirmedAlarmDbState) {
+      // If there's a clean scheduledTime that is different from ours, this run should be
+      // canceled.
+      // TODO(now): should probably also check that no other db requests are in-flight?
```
kentonv, line 522: TODO(now)

src/workerd/io/actor-sqlite.c++
```diff
@@ -402,6 +556,10 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::onNoPendingFlush() {
 
 const ActorSqlite::Hooks ActorSqlite::Hooks::DEFAULT = ActorSqlite::Hooks{};
 
+kj::Promise<void> ActorSqlite::Hooks::scheduleRun(kj::Maybe<kj::Date> newAlarmTime) {
+  return kj::READY_NOW;
```
kentonv, line 560: Default implementation should probably throw an exception saying alarms are not implemented?

src/workerd/io/actor-sqlite.c++
```diff
@@ -383,11 +511,37 @@ void ActorSqlite::shutdown(kj::Maybe<const kj::Exception&> maybeException) {
 }
 
 kj::Maybe<kj::Own<void>> ActorSqlite::armAlarmHandler(kj::Date scheduledTime, bool noCache) {
-  return hooks.armAlarmHandler(scheduledTime, noCache);
+  KJ_ASSERT(!haveDeferredDelete);
+  KJ_ASSERT(!inAlarmHandler);
+
+  auto localAlarmState = metadata.getAlarm();
+  if (localAlarmState != scheduledTime) {
```
kentonv, line 518: If we receive a premature alarm (`scheduledTime < localAlarmState`) then we need to update the alarm manager to the correct time, right?

kentonv: To be clear, the scenario would be:
* Application updates alarm from X to Y, where Y is later than X.
* The database commit goes through and is persisted.
* The alarm update (which happens after the commit, since it's moving later) fails.
* The machine dies.
* Later, the alarm fires at time X. At this point, no writes/commits/etc. are in progress or anything.
* We need to respond by updating the alarm time to Y (and skip running the alarm for now).

src/workerd/util/sqlite-metadata.c++
```diff
@@ -0,0 +1,64 @@
...
+    state = Uninitialized{db};
+  } else {
+    // The metadata table was initialized in the past. We can go ahead and prepare our statements.
+    // (We don't call ensureInitialized() here because the `CREATE TABLE IF NOT EXISTS` query it
+    // executes would be redundant.)
+    state = Initialized(db);
+  }
+}
+
+kj::Maybe<kj::Date> SqliteMetadata::getAlarm() {
```
kentonv, line 22: After seeing how much `getAlarm()` is called in actor-sqlite, I'm thinking maybe it's worth caching the value in `SqliteMetadata` after all. I'd do the caching here and not in actor-sqlite as I think it'll be easier here.

a-robinson:
Thanks for the context on what's changed! I don't know how github hasn't made iterative reviews just work by now, it's clearly not impossible given that third party services like reviewable.io are able to manage it.

kentonv:
Speaking of GitHub code review being bad, it appears the UI has decided to hide three comments from my recent review, because I mean, you don't really need to see _all_ the comments, right? /s

Screenshot from 2024-09-17 14-43-31

I doubt you'd miss it @jclee but I've definitely had cases in the past where people have totally skimmed over this box and never saw the comments within, so heads up!

src/workerd/util/sqlite-metadata.c++
jclee: Sure, makes sense.  Yeah, caching at the SqliteMetadata layer would be a lot easier than the way I was trying to do it in ActorSqlite.

One thing I wasn't quite sure of was whether a cached value would be correct when a transaction gets rolled back.  But I suppose it would be sufficient to notify SqliteMetadata of a rollback after calling "ROLLBACK TRANSACTION" or "ROLLBACK TO..." and refresh the cached value at that point.

src/workerd/io/actor-sqlite.c++
```diff
@@ -293,6 +420,7 @@ kj::Own<ActorCacheInterface::Transaction> ActorSqlite::startTransaction() {
 }
 
 ActorCacheInterface::DeleteAllResults ActorSqlite::deleteAll(WriteOptions options) {
+  // TODO(soon): handle deleteAll and alarm state interaction.
```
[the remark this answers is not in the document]
jclee: I added a commit (615deee2) to restore the metadata table's alarm state if necessary, as part of deleteAll().

src/workerd/io/actor-sqlite.c++
```diff
@@ -5,10 +5,28 @@
...
+#include <workerd/util/sentry.h>
 #include "io-gate.h"
 
 namespace workerd {
 
+namespace {
+
+// Returns true if a given (set or unset) alarm will fire earlier than another.
+static bool willFireEarlier(kj::Maybe<kj::Date> alarm1, kj::Maybe<kj::Date> alarm2) {
+  KJ_IF_SOME(t2, alarm2) {
```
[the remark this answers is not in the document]
jclee: I was thinking that, too -- it's certainly a lot simpler -- but I wasn't sure if there would be weird interaction if someone actually tried to set a `kj::maxValue` alarm... i.e. missing an update such that `getAlarm()` would return a non-null value when null was expected or vice versa.

But there's otherwise little behaviorial difference between an unset alarm and an alarm at infinity, and I guess these comparisons currently only affect the scheduled alarm value, which users can't currently read.  So maybe the simpler version is fine.

src/workerd/util/sqlite-metadata.c++
kentonv: Yeah, you could even just invalidate the cache so that the next time it is requested, it has to be read from sqlite again...

src/workerd/io/actor-sqlite.c++
jclee: Ah, yeah, that makes sense; I can take it out.

I figured there were cases where we might do extra alarm updates when recovering from failures, but I wasn't sure if we maybe also wanted to try to detect "shouldn't happen" cases like where the scheduled alarm time is later than the sqlite state (or unset?).  But maybe those can be detected other ways; we certainly could log the appearances we notice at handler run time.

src/workerd/io/actor-sqlite.c++
```diff
@@ -464,6 +469,21 @@ ActorCacheInterface::DeleteAllResults ActorSqlite::deleteAll(WriteOptions option
...
 
   uint count = kv.deleteAll();
+
+  // TODO(correctness): Since workerd doesn't have a separate durability step, in the unlikely
+  // event of a failure here, between deleteAll() and setAlarm(), we could theoretically lose the
+  // current alarm state when running under workerd.  Not sure if there's a practical way to avoid
+  // this.
+
+  // Reinitialize metadata wrapper, to allow on-demand recreation of metadata table.
+  metadata = SqliteMetadata(*db);
```
kentonv, line 479: Hmm, overwriting using an assignment like this is sort of surprising -- I wouldn't have expected `SqliteMetadata` to be assignable. It's also a little inefficient in that it's going to perform a database query to check for the existence of the metadata table, which in this case we know is definitely not there.

I think instead SqliteMetadata should probably register itself as a reset listener on the SqliteDatabase, and in its `beforeReset()` callback it should de-initialize itself.

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +200,67 @@ void ActorSqlite::onWrite() {
...
+  }
+
+  while (willFireEarlier(localAlarmState, lastConfirmedScheduledAlarm)) {
+    co_await requestScheduledAlarm(localAlarmState);
+    localAlarmState = metadata.getAlarm();
+  }
+
+  auto commitCallbackPromise = commitCallback();
+
+  // Fulfill and clear the pending commit.
```
kentonv, line 252: What happens if code above throws? I think we need to reject the fulfiller on the way out so that other callers can't end up hung? (We can keep `pendingCommit` set, though... so all future commits will also throw, which seems fine, because any commit throwing is going to break the output gate anyway.)

src/workerd/util/sqlite-metadata.c++
jclee: Added caching/invalidation in 935e9ecd

src/workerd/io/actor-sqlite.c++
jclee: I think it shouldn't actually cancel schedulingPromise, since `startPrecommitAlarmScheduling()` currently does not issue a scheduling request and populate the promise, if it sees that pendingCommit is already populated -- I assume that's OK?  But I can add more comments to clarify the reasoning.

src/workerd/io/actor-sqlite.c++
jclee: Hmm... what you're proposing makes sense, but I'm a bit puzzled because I have a test case where several invocations of `commitImpl()` are in-flight, and the output gate for all of them does not unblock until the commit promise completes.

https://github.com/cloudflare/workerd/blob/935e9ecdd8ddda26547e49da6b3dc951cfa6db32/src/workerd/io/actor-sqlite-test.c%2B%2B#L640

I suppose this must be because earlier output gate promises block all the later ones?  If so, I'm having trouble thinking of a sequence of events where moving the fulfill() call later results in a difference in behavior, such that I can add a test against a bug.  But I can update the code, so that we're not relying strictly on the output gate behavior for correctness.

src/workerd/io/actor-sqlite.c++
jclee: Am I mistaken in thinking that if the code throws, the fulfiller would be destroyed, which should reject the promise and unblock any caller waiting on the forked promise?

But I suppose in that case it will have the PromiseFulfiller exception message instead of the thrown exception, which would be more useful.  I can try adding that.

jclee: Oh, interesting... Apparently we have logic in PromiseFulfiller to detect when it is being destroyed during unwind for a thrown exception and use that instead of "PromiseFulfiller destroyed", so we're even currently recording a copy of the thrown exception correctly.

https://github.com/capnproto/capnproto/blob/7ad12031/c%2B%2B/src/kj/async-test.c%2B%2B#L417-L429

I can add a test case.

src/workerd/io/actor-sqlite.c++
jclee: So, I'd been thinking through how to implement this, and I think it might require a slight change in API to do it correctly...  The issue is that `armAlarmHandler()` can only return either `kj::none`, which cancels the alarm without retry, or the `kj::Own<void>` deleter value.  If we issue the alarm rescheduling request asynchronously and return `kj::none`, I think it potentially leaves a window where the AlarmManager will no longer try the old alarm time but the new alarm time has not yet been confirmed.

I'm thinking of adding a `kj::Promise<void>` to the set of types that `armAlarmManager()` can return, such that the caller can wait for the rescheduling to be successful before returning a canceled result.

src/workerd/io/actor-sqlite.c++
jclee: Wow, good catch... the coroutine here was indeed turning synchronously thrown exceptions into rejected promises.

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +201,83 @@ void ActorSqlite::onWrite() {
...
+  co_await commitCallbackPromise;
+  lastConfirmedAlarmDbState = localAlarmState;
+
+  // Notify any merged commitImpl() requests that the db update completed.
+  fulfiller->fulfill();
+
+  // If the db state is now later than the known-scheduled alarm, issue a request to update it to
+  // match the db state.  We don't need to hold open the output gate, so we add the scheduling
+  // request to commitTasks.
+  if (willFireEarlier(lastConfirmedScheduledAlarm, localAlarmState)) {
```
a-robinson, line 276: So IIUC:

1. We'll update the scheduled alarm time synchronouslybefore committing if we're moving it earlier
2. We'll update the scheduled alarm time asynchronously after committing if we're moving it later

But what if we do one and then the other, before we get back the result from the first request? Imagine something like:

1. Alarm scheduled/confirmed at time T
2. Set alarm to time T+100, which commits and kicks off a background request to move the alarm time back
3. Set alarm time to T+5 *before* the background request resolves, such that `lastConfirmedScheduledAlarm` is still set to T.
4. Since `lastConfirmedScheduledAlarm` was T, and T is earlier than T+5, we decide to update the alarm *after* committing.
5. Wait on `commitCallbackPromise`
6. The background request to move the alarm time back succeeds, updating `lastConfirmedScheduledAlarm` to T+100.
7. `commitCallbackPromise` resolves, but now T+5 *is* earlier than `lastConfirmedScheduledAlarm`, so this check evaluates to false and we end up never scheduling the alarm for T+5.

Am I missing something or could this leave us with an alarm that won't fire until T+100 even though it should fire at T+5?

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +201,83 @@ void ActorSqlite::onWrite() {
...
+    co_await pending.addBranch();
+    co_return;
+  }
+  auto [promise, fulfiller] = kj::newPromiseAndFulfiller<void>();
+  pendingCommit = promise.fork();
+
+  // Wait for the first precommit alarm scheduling request to complete, if any.
+  auto localAlarmState = precommitAlarmState.localAlarmState;
+  KJ_IF_SOME(p, precommitAlarmState.schedulingPromise) {
+    co_await p;
```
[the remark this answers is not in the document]
jclee: If it throws, I believe it should cause the coroutine context to be destroyed, which should destroy `fulfiller`, which should cause pendingCommit and all other forked promises to reject.  Kenton had the same concern, so maybe I'm misunderstanding something, but I did add a test case that I think would fail if a thrown exception caused hangs:

https://github.com/cloudflare/workerd/blob/bee6326be6a8f378cee2f5086526fd178e2e3f49/src/workerd/io/actor-sqlite-test.c%2B%2B#L826-L859

src/workerd/io/actor-sqlite.c++
jclee: Wow... good catch.  I was worried there might be cases like this, but hadn't been able to find one.

I'm not sure of the best way to resolve the issue... I think we may need to have `startPrecommitAlarmScheduling()/commitImpl()` block on the completion of all outstanding scheduling requests prior to the initial `willFireEarlier()` check (while also, of course, ensuring that the first scheduling request is still issued synchronously in `startPrecommitAlarmScheduling()` for correctness in workerd).  I think that will limit the throughput of commits in flight when some contain alarm changes, but should otherwise work correctly.

That will also mostly limit scheduling requests to having one in-flight at a time, so it would probably make sense to explicitly enforce that as well, to eliminate the possibility of other races between the final alarm scheduling in `commitImpl()`.  Per Kenton's other comment, though, I think we can allow the output gate to remain open for that request.

@kentonv Does that make sense?

src/workerd/io/actor-sqlite.c++
```diff
@@ -283,7 +406,28 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::setAlarm(
...
+    // After the first write in the handler occurs, the logic here is correct again as the current
+    // alarm time would match what we are reporting to the user from getAlarm().
+    //
+    // TODO(now): optimization no longer useful in sqlite?
+    if (metadata.getAlarm() == newAlarmTime) {
+      return kj::none;
+    }
+  }
+  metadata.setAlarm(newAlarmTime);
+  haveDeferredDelete = false;
```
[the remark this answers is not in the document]
jclee: Oh, good point... this does not work with transaction rollbacks.  I think to fix this we'd need more complex tracking of the deferred delete state -- e.g. only setting `haveDeferredDelete` to false when a transaction with an alarm change is committed.

It looks like `ActorCache` avoids this problem by explicitly tracking alarm changes in `Transaction` objects, and maybe the simplest solution might be to do something similar here.  It looks like the root `setAlarm()` method might have a similar problem though, or might interact weirdly with `txn.setAlarm()`?  I'd need to read more code to see if we do something to explicitly disallow use of the root `setAlarm()` method during transactions, or other unusual interactions between root and transaction-level methods.  As currently implemented in ActorSqlite, it looks like the transaction-level methods are all aliases for the root-level methods.

src/workerd/io/actor-sqlite.c++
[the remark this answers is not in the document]
a-robinson: You're right, I forgot about the auto-reject-on-destruct behavior of fulfillers. Thanks for explaining!

src/workerd/io/actor-sqlite.c++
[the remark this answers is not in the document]
a-robinson: > It looks like ActorCache avoids this problem by explicitly tracking alarm changes in Transaction objects, and maybe the simplest solution might be to do something similar here

Yeah, I think so.

> It looks like the root setAlarm() method might have a similar problem though

No, not as far as I can tell. It applies the change immediately to its in-memory state and immediately schedules a flush to storage along with an output gate, the same as any write in ActorCache.

> I'd need to read more code to see if we do something to explicitly disallow use of the root setAlarm() method during transactions, or other unusual interactions between root and transaction-level methods.

In the ActorCache model, there really weren't any interactions. Root methods apply directly and immediately even if there is an ongoing transaction, while changes in transaction get buffered up until commit time and only flushed then.

src/workerd/io/actor-sqlite.c++
a-robinson: Although if we're too tight on time before when this needs to ship, it wouldn't be disastrous to ship what you have and occasionally have delayed alarms in that kind of edge case for a week or two while we get a better fix in.

src/workerd/io/actor-sqlite.c++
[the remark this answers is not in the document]
jclee: > In the ActorCache model, there really weren't any interactions. Root methods apply directly and immediately even if there is an ongoing transaction, while changes in transaction get buffered up until commit time and only flushed then.

Ah, OK, then ActorCache doesn't have the problem I was thinking of.  I was confused by the fact ActorSqlite seems to have different semantics...  Its transaction methods are direct passthroughs to the root methods, such that any operation becomes part of the innermost nested open transaction.  So if someone did:
```
async alarm() {
  await this.storage.startTransaction((txn) => {
    this.storage.setAlarm(Date.now());
    txn.rollback();
  });
}
```
...for ActorSqlite, the `setAlarm()` would be part of the transaction, so the alarm state change would be rolled back, and the user might have some expectation that the alarm would be deleted at the end of the handler.  But for ActorCache, it sounds like the change would not be rolled back and the alarm would not be deleted at the end of the handler.

@kentonv I guess this difference in behavior might be intentional, to support other features, and would also happen for other methods like `this.storage.put()`?  If so, I assume it would be most useful for ActorSqlite to delete the alarm at the end of the handler in the above example, rather than maybe trying to match the ActorCache behavior.

src/workerd/io/actor-sqlite.h
```diff
@@ -22,10 +23,9 @@ class ActorSqlite final: public ActorCacheInterface, private kj::TaskSet::ErrorH
   // for alarm operations.
   class Hooks {
   public:
-    virtual kj::Promise<kj::Maybe<kj::Date>> getAlarm();
-    virtual kj::Promise<void> setAlarm(kj::Maybe<kj::Date> newAlarmTime);
-    virtual kj::Maybe<kj::Own<void>> armAlarmHandler(kj::Date scheduledTime, bool noCache);
-    virtual void cancelDeferredAlarmDeletion();
+    // Makes a request to the alarm manager to run the alarm handler at the given time, returning
+    // a promise that resolves when the scheduling has succeeded.
+    virtual kj::Promise<void> scheduleRun(kj::Maybe<kj::Date> newAlarmTime);
```
jclee, line 28: So, one additional issue I've found while implementing the edgeworker side of scheduling is that ActorCache explicitly uses separate `setAlarm()` and `deleteAlarm()` RPC calls.  I thought the single `scheduleRun()` method might be sufficient, but I see that the `deleteAlarm()` method has an optional parameter to allow specifying an explicit alarm time, which ActorCache passes in when doing the deferred deletion at the end of a handler.

I'm not sure of the exact situation this might be protecting against, but it seems likely that constraining the deferred deletion is important for the at-least-once delivery guarantee, so I think I should revise the Hooks API to match.

a-robinson: I don't have a specific example scenario in mind -- it looks like the motivation for adding that was due to the internal AlarmManager -- but it is a good idea to be paranoid about deleting alarms, since an alarm failing to run is potentially a major bug, while an alarm being invoked when it shouldn't is just a minor annoyance.

src/workerd/io/actor-sqlite.c++
kentonv: I think that when making the post-commit request to move the alarm time _later_, you can _immediately_ update `lastConfirmedScheduledAlarm` to the new value when you kick off the request, without actually waiting for confirmation. This works because it's always OK if the alarm time accidentally gets left being too early -- you'll just reschedule it when it fires prematurely.

src/workerd/io/actor-sqlite.c++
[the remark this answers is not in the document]
kentonv: Yeah, under sqlite, when a transaction is open, there's no way for us to support doing queries that skip the transaction. And we also wanted transactions to include SQL queries with the new SQL interface. So we just said: ok, transactions now cover everything done within the transaction callback whether you issue the queries against the `txn` object or the underlying `storage` object. `txn` is essentially deprecated and people should stop declaring it as an argument.

src/workerd/io/actor-sqlite.c++
a-robinson: Yeah, I think that works.

Although I still worry about whether we need to serialize the `hooks.scheduleRun` calls. Either that or `scheduleRun` will need to be implemented such that any internal retries can't possibly cause updates to be applied out-of-order.

jclee: I implemented the check for whether to do the immediate update in `requestScheduledAlarm()`, since `commitImpl()` isn't the only place we're using it to set alarms later...  We're also doing it in `armAlarmHandler()`, to reschedule alarms that are too early, and I assume that those requests need the same treatment, to avoid similar problems.

src/workerd/io/actor-sqlite.c++
kentonv: Oh... I forgot about that. Probably worth a comment.

I knew the fulfiller would be automatically rejected but thought it would be with an unhelpful "PromiseFulfiller was destroyed without fulfilling the promise" message.

src/workerd/api/global-scope.c++
```diff
@@ -524,12 +484,63 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
...
+            .retryCountsAgainstLimit = shouldRetryCountsAgainstLimits,
+            .outcome = EventOutcome::EXCEPTION};
+        });
       });
-    });
-  } else {
-    return WorkerInterface::AlarmResult{.retry = false, .outcome = EventOutcome::CANCELED};
+    }
+    KJ_CASE_ONEOF(armResult, ActorCacheInterface::CancelAlarmHandler) {
+      // TODO(cleanup): might be possible to use kj::READY_NOW promises instead of kj::none to
```
kentonv, line 532: Yes I think that would be fine.

src/workerd/io/actor-sqlite.c++
```diff
@@ -169,11 +223,97 @@ void ActorSqlite::onWrite() {
...
+
+  bool movingAlarmLater = willFireEarlier(lastConfirmedScheduledAlarm, requestedTime);
+  if (movingAlarmLater) {
+    // Since we are setting the alarm to be later, we can update lastConfirmedScheduledAlarm
+    // immediately and still preserve the invariant that the scheduled alarm time is equal to or
+    // earlier than the persisted db alarm value.  Doing the immediate update ensures that
+    // subsequent invocations of commitImpl() will compare against the correct value in their
+    // precommit alarm checks, even if other later-setting requests are still in-flight, without
+    // needing to wait for them to complete.
+    lastConfirmedScheduledAlarm = requestedTime;
```
kentonv, line 243: Perhaps `lastConfirmedScheduledAlarm` should actually be called something like... `alarmScheduledNoLaterThan`? Makes clearer why it's safe to update before the call when scheduling later, but only after the call when scheduling earlier.

jclee: Thanks! It was bugging me that it no longer represented "confirmed" alarms, but I wasn't sure what a good name was.

src/workerd/io/actor-sqlite.h
jclee: FWIW, I took an attempt at passing the specific alarm time on alarm deletion, but plumbing the deletion time from the end of the handler to the point of scheduling adds a fair bit of complexity to what is already pretty complex code...  Maybe there's a better way to do it, than I'm doing, but I'm thinking it may be better to omit for now if it isn't strictly needed for correctness.

https://github.com/cloudflare/workerd/commit/62936f69a8da23926c787c1e7f29606fbcfa77e6

src/workerd/io/actor-sqlite.c++
```diff
@@ -18,8 +29,16 @@ ActorSqlite::ActorSqlite(kj::Own<SqliteDatabase> dbParam,
       commitCallback(kj::mv(commitCallback)),
       hooks(hooks),
       kv(*db),
+      metadata(*db),
       commitTasks(*this) {
   db->onWrite(KJ_BIND_METHOD(*this, onWrite));
+  lastConfirmedAlarmDbState = metadata.getAlarm();
```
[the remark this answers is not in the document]
jclee: I'm not sure myself how we're accounting for sqlite reads that are required as part of DO operation, but I guess one additional thing to note is that both SqliteMetadata and SqliteKv perform queries on startup to read whether their tables have already been created... Not sure if those queries increment the same counters and need to be similarly accounted for.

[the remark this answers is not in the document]
a-robinson: Fair point. cc @ramyabygari , who (based on the internal ticket) I believe is going to be the person working on ensuring alarm reads aren't billed

src/workerd/io/actor-sqlite.c++
jclee: Fixed in d1983124b .

src/workerd/util/sqlite-metadata.c++
jclee: * Added code reestablish the alarm state after `deleteAll()`, in 615deee2 .
* Added TODO to address database cleanup after final alarm is removed, in 2d833464 .

src/workerd/io/actor-sqlite.c++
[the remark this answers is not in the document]
jclee: Changed to simpler version in 784a5536 .

src/workerd/io/actor-sqlite.c++
jclee: Removed TODO and populated value in 784a55365 .

src/workerd/io/actor-sqlite.c++
jclee: Addressed in f1c401766 .

src/workerd/io/actor-sqlite.c++
jclee: Added more documentation to commitImpl() in 784a55365 and d1983124b

src/workerd/io/actor-sqlite.c++
jclee: Implemented in 3a10dcfd .

src/workerd/io/actor-sqlite.c++
jclee: Implemented in bee6326be .

src/workerd/io/actor-sqlite.c++
jclee: Added comment in d1983124b .

src/workerd/io/actor-sqlite.c++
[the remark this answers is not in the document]
jclee: Added comment in d1983124b .

src/workerd/io/actor-sqlite.c++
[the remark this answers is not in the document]
jclee: Added transaction alarm dirtiness tracking and test cases in 2d833464 .

src/workerd/io/actor-sqlite.c++
jclee: Added test case and implementation in ffefa438 .

src/workerd/util/sqlite-metadata-test.c++
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "sqlite-metadata.h"
+#include <kj/test.h>
+
+namespace workerd {
+namespace {
+
+KJ_TEST("SQLite-METADATA") {
```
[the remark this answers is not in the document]
jclee: Added invalidate() test in 93b152796 .

jclee:
Thanks for the feedback so far! I think I've addressed the known issues on the workerd side -- please let me know if further changes would help.

I'm still working on adding alarm scheduling retries and some more testing on the edgeworker side.

I still need to rebase this PR to resolve a conflict in `#include`s, but can wait if it's going to complicate review on github.  I'm also not sure if the current granularity of commits will be useful post-merge, so I was thinking of mostly merging the ActorSqlite alarms implementation together in one commit.  Is that OK?

a-robinson:
> I still need to rebase this PR to resolve a conflict in #includes, but can wait if it's going to complicate review on github

Mind waiting until I get through the latest batch of commits? I'm reading through them now. (Thanks for separating them out, by the way)

> I was thinking of mostly merging the ActorSqlite alarms implementation together in one commit. Is that OK?

SGTM

src/workerd/io/actor-sqlite.c++
a-robinson: I still think there's something to be said for serializing calls `hooks.scheduleRun` and coalescing queued-up updates together -- we not only have to worry about changes being applied to the scheduler out-of-order, but also about responses from the scheduler coming back out of order and causing us to run the callback that updates `alarmScheduledNoLaterThan` out-of-order.

But this looks good enough to ship 👍

src/workerd/io/actor-sqlite.c++
```diff
@@ -406,27 +433,17 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::setAlarm(
...
-    //
-    // If we're in the alarm handler and haven't set the time yet, we can't perform this
-    // optimization as the current alarm time may be equal to the requested time but we indicate
-    // to the actor in getAlarm() that there is no alarm set, therefore we need to act like that
-    // in setAlarm().
-    //
-    // After the first write in the handler occurs, the logic here is correct again as the current
-    // alarm time would match what we are reporting to the user from getAlarm().
-    //
-    // TODO(now): optimization no longer useful in sqlite?
```
a-robinson: Is the idea that this is no longer useful because sqlite won't actually rewrite the page if it's unchanged? Or if it's something else, why?

src/workerd/io/actor-sqlite.c++
```diff
@@ -546,7 +563,6 @@ void ActorSqlite::shutdown(kj::Maybe<const kj::Exception&> maybeException) {
 
 kj::OneOf<ActorSqlite::CancelAlarmHandler, ActorSqlite::RunAlarmHandler> ActorSqlite::
     armAlarmHandler(kj::Date scheduledTime, bool noCache) {
-  KJ_ASSERT(!haveDeferredDelete);
```
a-robinson: Why did this need to be removed? I'd still think it should always be false when starting a new alarm handler?

a-robinson (approved):
Nice work on this @jclee , it's very tricky stuff.

kentonv (approved):

src/workerd/io/actor-sqlite.c++
jclee: I wasn't sure that's the case...  I think we only currently set it to false when:

- setAlarm() is called on the root level transaction
- committing a transaction with a dirty alarm value
- starting a new alarm handler when the handler time is later than the local database state

We're not currently setting it to false when a DeferredAlarmDeleter is destroyed and starts the deferred alarm deletion, and maybe we should be?  But I'm also not totally sure if it's possible for the DeferredAlarmDeleter to be attached to something that takes a few turns of the event loop to fully resolve, such that it's possible for other work in the event queue to start a new handler before DeferredAlarmDeleter's destructor runs.

But thinking about it more, yeah, this seems unlikely, and I think should already be covered by the log-once warnings, so I'll try reintroducing the assert and fix the DeferredAlarmDeleter to clear the flag.

src/workerd/io/actor-sqlite.c++
jclee: I'd added an optimization in the SqliteMetadata caching layer to skip writes if the write already matched the cached value, so yeah, writing the same value wouldn't currently even reach sqlite, here.

Also, I think the motivation for the analogous comment in ActorCache is because we don't want to skip the alarm dirtiness logic when the current alarm time matches the requested alarm time but we still have a deferred delete.  But in the case of the below code, we're running the dirtiness logic regardless (even though I guess we'd similarly only need to run it when we have a deferred delete, assuming handlers and transactions nest sanely).

src/workerd/io/actor-sqlite.c++
jclee: Readded the check as a logged-once warning and added code to clear the flag, in cb67b25 .

jclee:
While iterating on the PR, I noticed that the comparison I was doing for the "if scheduled alarm is earlier, reschedule later" logic was backward in both the implementation and tests :( ... fixed in 6111915a .

I'm now going to rebase to resolve the `#include` conflict and squash the ActorSqlite changes together.  Pre-rebase diff: https://github.com/cloudflare/workerd/compare/jlee/srs-alarms-wip3...jlee/srs-alarms-wip4?expand=1

PR #2649, opened 2024-09-04
Update Github Actions dependencies

fhanau:
- This finally fixes warnings about the deprecated node16 being used for CLAssistant.
- Use short-form for dependencies in BUILD.dawn


anonrig (approved):
Should we also enable dependabot?

fhanau:
> Should we also enable dependabot?

IMHO dependabot can just be spamming PRs sometimes... We don't use GitHub Actions dependencies that much and managing dependencies is quite sane for them in the general case (e.g. we use `actions/checkout@v4` which maps to the latest available 4.x release, so we get non-major updates automatically). These dependencies have been mostly low-maintenance and the current approach of updating them as needed/based on deprecation warnings works well enough.

PR #2650, opened 2024-09-04
Fix feature detection for stable Pyodide 0.26

hoodmane:
We want Pyodide to decide we're a browser. It checks a few extra things in 0.26.0 before being convinced of this. I think this is the only change to the loader code required to get Pyodide 0.26.2 working.


dom96 (approved):
LGTM, be sure to create an internal PR for this too

hoodmane:
Okay. Also, internally we don't use the development bundle for most Python tests except in the validator. Perhaps I should fix that too...

PR #2653, opened 2024-09-04
enable linter for module.ts

anonrig:
We no longer need to disable linters for files that use URL or queueMicrotask()

### DIFF
--- a/src/node/internal/web_crypto.d.ts
+++ b/src/node/internal/web_crypto.d.ts
@@ -1,9 +1,0 @@
-// Just enough of web crypto api to write the code in node-compat
-type SubtleCrypto = unknown;
-
-type CryptoKey = unknown;
-
-declare const CryptoKey: {
-  prototype: CryptoKey;
-  new (): CryptoKey;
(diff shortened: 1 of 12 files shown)


fhanau (approved):

anonrig:
I had to force push. I added `dom` as the library to tsconfig to enable types for URL. This come with a caveat (or a benefit depending on your pov.), where webcrypto types are also implemented.

src/node/internal/web_crypto.d.ts
```diff
@@ -1,9 +1,9 @@
-// Just enough of web crypto api to write the code in node-compat
-type SubtleCrypto = unknown;
-
-type CryptoKey = unknown;
-
-declare const CryptoKey: {
-  prototype: CryptoKey;
-  new (): CryptoKey;
-};
+// // Just enough of web crypto api to write the code in node-compat
```
anonrig, line 1: I forgot to remove this file.

npaun (approved):
Cool

PR #2654, opened 2024-09-04
Added interface type for Cloudflare pipelines binding

hhoughgg:


types/defines/pipeline-transform.d.ts
```diff
@@ -11,3 +11,18 @@ export abstract class PipelineTransform {
    */
   public transformJson(data: object[]): Promise<object[]>;
 }
+
+export type PipelineResponse = {
```
a-robinson, line 15: Sorry for commenting on API design here given that the code is probably already written, but what is this return type communicating that wouldn't be equally-well communicated by a rejected promise / thrown Error?

In what cases will `success` be false? In what cases will `error` be filled in vs the `Promise` being rejected?

I'd have expected `send` to just return a `Promise<void>` unless there was something more than success/error to be returned.

a-robinson: For example that's what the Queues `send()` method returns: https://github.com/cloudflare/workerd/blob/main/src/workerd/api/queue.h#L68

types/defines/pipeline-transform.d.ts
```diff
@@ -11,3 +11,18 @@ export abstract class PipelineTransform {
...
+  success: boolean
+  error?: string
+}
+
+export interface Pipeline {
+  /**
+   * send recieves an array of javascript objects which are sent
+   * into the pipeline for processing via binding
+   * @param data The data to be send
```
a-robinson, line 24: Either "The data to send" or "The data to be sent", but not "The data to be send"

types/defines/pipeline-transform.d.ts
```diff
@@ -11,3 +11,18 @@ export abstract class PipelineTransform {
...
+
+export type PipelineResponse = {
+  success: boolean
+  error?: string
+}
+
+export interface Pipeline {
+  /**
+   * send recieves an array of javascript objects which are sent
+   * into the pipeline for processing via binding
```
a-robinson, line 23:
```suggestion
   * into the pipeline for processing
```

types/defines/pipeline-transform.d.ts
```diff
@@ -11,3 +11,18 @@ export abstract class PipelineTransform {
...
+   * @param data The data to be send
+   * @returns A promise containing the outcome of the send
+   */
+  send(data: object[]): Promise<PipelineResponse>
```
elithrar, line 27: Why an `object` - not every `object` is serializable? Does this cause issues with writing objects out to storage / consuming from a non-JS system?

How does this fit with Queues? https://github.com/cloudflare/workerd/blob/c9eca148301f683bf7b4bef2c642ef6380ea531e/src/workerd/api/queue.h#L51-L87

hhoughgg: This will act in the same way as a queue send in that the object must be structured cloneable since this is an RPC call. The only difference is the limit is 1mb instead of 128kb. https://developers.cloudflare.com/queues/configuration/javascript-apis/#queue

I think the idea behind using an object is that we want to make sure all the data in Pipelines is a row style layout. Rows can also be represented in other languages no problem and conversions of rows/columns to other formats is a well worn path.

oliy:
> I have read the CLA Document and I hereby sign the CLA

oliy:
recheck

fhanau:
@oliy Not sure why the CLA approval didn't work – maybe try commenting without the leading quote?

jasnell (approved):

a-robinson (approved):

oliy:

oliy:
recheck

PR #2655, opened 2024-09-04
Add just test-stream to run a test with live output

npaun:
I find it a lot more convenient than rooting around in the test log

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,6 +1,7 @@
 alias b := build
 alias t := test
 alias f := format
+alias st := stream-test
 
 default:
   @just --list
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -28,6 +28,9 @@ test *args="//...":
 test-asan *args="//...":
   just test {{args}} --config=asan
 
+test-stream args:
```
anonrig, line 31: Can you add an example argument as a comment to here? Should we have `//...` as the default value?

anonrig, line 31: Maybe lets call it `stream-tests` and add a shortcut for `st` so we can execute it using `just st` 
```suggestion
stream-tests args:
```

npaun: I think that would be too slow and too much output - I think it's best when running a single test.

npaun: Nice!

anonrig: Would you mind adding a comment for running an example test?

anonrig (approved):

PR #2656, opened 2024-09-04
enable linters on more node.js implementations

anonrig:
Removes eslint-disable on more files and fixes the types along the way.

### DIFF
--- a/src/node/tsconfig.json
+++ b/src/node/tsconfig.json
@@ -12,7 +12,7 @@
     "noFallthroughCasesInSwitch": true,
     "noImplicitOverride": true,
     "noImplicitReturns": true,
-    "noPropertyAccessFromIndexSignature": true,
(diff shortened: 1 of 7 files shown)


src/node/tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "noFallthroughCasesInSwitch": true,
     "noImplicitOverride": true,
     "noImplicitReturns": true,
-    "noPropertyAccessFromIndexSignature": true,
+    "noPropertyAccessFromIndexSignature": false,
```
anonrig, line 15: Note to reviewer:

I'm not quite sure why we enabled this in the first place. A object with a type of `Record<unknown, unknown>` has to access properties using `obj['my_field']` rather than `obj.my_field`. Without this change, we create a lot of unnecessary strings.

jasnell (approved):

PR #2658, opened 2024-09-04
Fix initialization of brotli parameters

npaun:
Brotli parameters were being read from the wrong argument.


src/workerd/api/node/zlib-util.c++
```diff
@@ -837,6 +836,8 @@ bool ZlibUtil::BrotliCompressionStream<CompressionContext>::initialize(jsg::Lock
     return false;
   }
 
+  auto results = params.template asArrayPtr<uint32_t>();
```
anonrig, line 839: Is there a test that actually fails without this change?

anonrig (approved):

src/workerd/api/node/zlib-util.c++
npaun: Not yet; but I'll make sure that it's covered either from the NodeJS test suite or from a custom test in another PR.

PR #2660, opened 2024-09-05
Implement brotli one shot methods

npaun:


src/workerd/api/node/zlib-util.c++
```diff
@@ -878,6 +836,109 @@ void ZlibUtil::CompressionStream<CompressionContext>::FreeForZlib(void* data, vo
...
+      opts.windowBits.orDefault(Z_DEFAULT_WINDOWBITS), opts.memLevel.orDefault(Z_DEFAULT_MEMLEVEL),
+      opts.strategy.orDefault(Z_DEFAULT_STRATEGY), kj::mv(opts.dictionary));
+  ctx.setFlush(opts.finishFlush.orDefault(Z_FINISH));
+  ctx.setInputBuffer(getInputFromSource(data));
+  return syncProcessBuffer(ctx, result);
+}
+
+void ZlibUtil::zlibWithCallback(
+    jsg::Lock& js, InputSource data, Options options, ZlibModeValue mode, CompressCallback cb) {
+  try {
```
jasnell, line 892: These should use `js.tryCatch(...)` instead. Specifically, the `CompressCallback` could end up triggering a `jsg::JsExceptionThrown` which would not be handled by this, and there are fatal exception cases we need to make sure are handled. It would be something like...

Also, there's a risk here that `cb(...)` is invoked twice. It should only ever be invoked exactly once.

src/workerd/api/node/zlib-util.c++
```diff
@@ -878,6 +837,126 @@ void ZlibUtil::CompressionStream<CompressionContext>::FreeForZlib(void* data, vo
...
+
+template <typename Context>
+kj::Array<kj::byte> ZlibUtil::brotliSync(InputSource data, BrotliOptions opts) {
+  ContextRAII<Context> ctx(Context::Mode);
+
+  auto chunkSize = opts.chunkSize.orDefault(ZLIB_PERFORMANT_CHUNK_SIZE);
+  auto maxOutputLength = opts.maxOutputLength.orDefault(Z_MAX_CHUNK);
+
+  JSG_REQUIRE(Z_MIN_CHUNK <= chunkSize && chunkSize <= Z_MAX_CHUNK, Error, "Invalid chunkSize");
+  JSG_REQUIRE(maxOutputLength <= Z_MAX_CHUNK, Error, "Invalid maxOutputLength");
```
jasnell, line 917: Nit: it would be helpful to expand the error messages here with a bit more detail about why they are invalid. But if we're matching node.js' error messages, then this is fine as is for now.

src/workerd/api/node/zlib-util.c++
npaun: I think your example got truncated

src/workerd/api/node/zlib-util.c++
npaun: This doesn't match them very well because the NodeJS messages are more structured. I'd like to take a look to see what we can do to improve this later on.

src/workerd/api/node/zlib-util.c++
```diff
@@ -878,6 +837,126 @@ void ZlibUtil::CompressionStream<CompressionContext>::FreeForZlib(void* data, vo
...
+}
+
+template <typename Context>
+void ZlibUtil::brotliWithCallback(
+    jsg::Lock& js, InputSource data, BrotliOptions options, CompressCallback cb) {
+  try {
+    cb(js, kj::none, brotliSync<Context>(kj::mv(data), kj::mv(options)));
+  } catch (kj::Exception& ex) {
+    auto tunneledError = jsg::tunneledErrorType(ex.getDescription());
+    cb(js, tunneledError.message, kj::none);
```
jasnell, line 957: Same issue as above... The `cb(...)` should only every be called once and this should likely be using `js.tryCatch(...)` instead.

src/workerd/api/node/zlib-util.c++
jasnell: no, but my thought process did lol. I originally intended to write an example but got distracted and forgot to.

src/workerd/api/node/zlib-util.c++
jasnell: Ok, please add a TODO(soon) comment then that will help us keep track and remember

src/workerd/api/node/zlib-util.c++
npaun: I think I figured it out - please have a look.

src/node/zlib.ts
```diff
@@ -49,6 +49,10 @@ const gunzip = protectMethod(zlib.gunzip);
 const gunzipSync = protectMethod(zlib.gunzipSync);
 const unzip = protectMethod(zlib.unzip);
 const unzipSync = protectMethod(zlib.unzipSync);
+const brotliCompress = protectMethod(zlib.brotliCompress);
+const brotliDecompressSync = protectMethod(zlib.brotliDecompressSync);
+const brotliDecompress = protectMethod(zlib.brotliDecompress);
+const brotliCompressSync = protectMethod(zlib.brotliCompressSync);
```
anonrig, line 55:
```suggestion
const brotliCompress = protectMethod(zlib.brotliCompress);
const brotliCompressSync = protectMethod(zlib.brotliCompressSync);
const brotliDecompressSync = protectMethod(zlib.brotliDecompressSync);
const brotliDecompress = protectMethod(zlib.brotliDecompress);
```

src/workerd/api/node/zlib-util.c++
```diff
@@ -3,18 +3,20 @@
...
+#include "workerd/jsg/jsg.h"
 
 // The following implementation is adapted from Node.js
 // and therefore follows Node.js style as opposed to kj style.
 // Latest implementation of Node.js zlib can be found at:
 // https://github.com/nodejs/node/blob/main/src/node_zlib.cc
 namespace workerd::api::node {
-
-kj::ArrayPtr<kj::byte> ZlibUtil::getInputFromSource(InputSource& data) {
+kj::ArrayPtr<kj::byte> getInputFromSource(ZlibUtil::InputSource& data) {
```
anonrig, line 19: You can keep it as it is, and just move it while passing to an argument:

```suggestion
kj::ArrayPtr<kj::byte> getInputFromSource(ZlibUtil::InputSource data) {
```

src/workerd/api/node/zlib-util.c++
```diff
@@ -624,6 +626,20 @@ void BrotliContext::setBuffers(kj::ArrayPtr<kj::byte> input,
...
+
+void BrotliContext::setOutputBuffer(kj::ArrayPtr<kj::byte> output) {
+  nextOut = output.begin();
+  availOut = output.size();
+}
+
+kj::uint BrotliContext::getAvailOut() const {
+  return availOut;
+}
+
```
anonrig, line 642: These can be added to header file

src/workerd/api/node/zlib-util.c++
```diff
@@ -878,6 +838,129 @@ void ZlibUtil::CompressionStream<CompressionContext>::FreeForZlib(void* data, vo
...
+    ctx.work();
+
+    KJ_IF_SOME(error, ctx.getError()) {
+      JSG_FAIL_REQUIRE(Error, error.message);
+    }
+
+    result.adjustUnused(ctx.getAvailOut());
+  } while (ctx.getAvailOut() == 0);
+
+  return result.releaseAsArray();
```
anonrig, line 868: shouldn't we need attach here, in order to prevent a memory leak? @jasnell

```suggestion
  return result.releaseAsArray().attach(kj::mv(result));
```

src/workerd/api/node/zlib-util.c++
```diff
@@ -3,18 +3,20 @@
 //     https://opensource.org/licenses/Apache-2.0
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 
-#include "util.h"
 #include "zlib-util.h"
 
+#include "util.h"
+#include <workerd/jsg/exception.h>
```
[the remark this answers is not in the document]
npaun: Aren't we using a function from it?

src/workerd/api/node/zlib-util.c++
```diff
@@ -878,6 +838,129 @@ void ZlibUtil::CompressionStream<CompressionContext>::FreeForZlib(void* data, vo
...
+
+template <typename Context>
+void ZlibUtil::brotliWithCallback(
+    jsg::Lock& js, InputSource data, BrotliOptions options, CompressCallback cb) {
+
+  auto res = js.tryCatch([&]() {
+    return CompressCallbackArg(brotliSync<Context>(kj::mv(data), kj::mv(options)));
+  }, [&](jsg::Value&& exception) {
+    return CompressCallbackArg(jsg::JsValue(exception.getHandle(js)));
+  });
```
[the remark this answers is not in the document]
npaun: James requested it for correctness reasons. I'll add a comment

src/workerd/api/node/zlib-util.h
```diff
@@ -212,8 +215,9 @@ class BrotliContext {
   void* alloc_opaque_brotli = nullptr;
 };
 
-class BrotliEncoderContext final: public BrotliContext {
+class BrotliEncoderContext: public BrotliContext {
```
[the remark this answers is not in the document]
npaun: Cause I need to inherit from it to provide an optional RAII wrapper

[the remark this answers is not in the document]
npaun: Cause I need to inherit from it to provide an optional RAII wrapper

src/workerd/api/node/zlib-util.c++
npaun: This is to keep it out of the header since it's private. Speeds compiles a little.

src/workerd/api/node/zlib-util.c++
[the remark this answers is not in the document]
anonrig: jsg/jsg.h already includes exception

src/workerd/api/node/zlib-util.h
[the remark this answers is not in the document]
anonrig: Can you add a TODO here? Once we make sure that the tests succeed, we can get rid of RAII wrapper.

src/workerd/api/node/zlib-util.c++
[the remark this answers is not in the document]
npaun: This is actually an interesting point. These big umbrella includes don't interact so well with IDE auto imports. There is a tool to identify redundant includes called include-what-you-use. Maybe I should try to see how it fares on our codebase

src/workerd/api/node/zlib-util.h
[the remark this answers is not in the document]
npaun: I prefer it to manually calling .close(). We're in C++ after all 😛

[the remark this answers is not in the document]
anonrig: Yes, by removing RAII, I mean moving RAII behavior to the Context.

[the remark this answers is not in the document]
npaun: Ah! Yes let's do that.

src/workerd/api/node/zlib-util.c++
anonrig: Then I recommend making it const

src/workerd/api/node/zlib-util.h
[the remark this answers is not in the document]
anonrig: Let's do that once we know that our implementation is correct. Later we can improve the performance as well.

src/workerd/api/node/zlib-util.c++
npaun: I'm not sure that I follow. It's not a member function so it can't be const. It's already static but I can add the keyword.

src/workerd/api/node/zlib-util.c++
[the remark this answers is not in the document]
anonrig: I am also interested in include-what-you-use but that's for another pull-request I suppose.

src/workerd/api/node/zlib-util.c++
anonrig: I mean:

```suggestion
kj::ArrayPtr<kj::byte> getInputFromSource(const ZlibUtil::InputSource& data) {
```

src/workerd/api/node/zlib-util.c++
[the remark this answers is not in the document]
npaun: By the time we're through, every single mistake will be blocked by the CI. (almost)

src/workerd/api/node/zlib-util.c++
npaun: Ah and I guess that means the return type should be kj::ArrayPtr<const kj::byte> though I need to double check with all these funky templates

src/node/internal/internal_zlib.ts
```diff
@@ -100,25 +115,22 @@ function normalizeArgs(
...
-function wrapCallback(
-  callback: CompressCallback<Error, Buffer>
-): CompressCallback<string, ArrayBuffer> {
-  return (error: string | null, result: ArrayBuffer | undefined) => {
+function wrapCallback(callback: CompressCallback): InternalCompressCallback {
+  return (res: Error | ArrayBuffer) => {
     queueMicrotask(() => {
-      if (error) {
-        callback(new Error(error));
+      if (res instanceof Error) {
```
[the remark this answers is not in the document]
npaun: I'll look into whether some template voodoo is possible

src/workerd/api/node/zlib-util.c++
npaun: I didn't do this because the full setBuffers is defined just above, and I thought it was easier to read with them all in one place.

anonrig (approved):

src/workerd/api/node/zlib-util.c++
jasnell: No, that's not needed in this case. `releaseAsArray()` is going to return a `kj::Array<T>` that owns the data, basically moving the underlying storage away from `result`. You'd be uselessly attaching an empty object here.

jasnell (approved):

PR #2664, opened 2024-09-05
Throw error if the python package disk cache dir is missing.

dom96:
I ran into this when regenerating snapshots, should be loud here so we aren't accidentally not using the disk cache when we're supposed to. Unless there are cases where we rely on this silently failing?


jasnell (approved):
LGTM but would it be reasonable for workerd to try to create the directory if it doesn't exist?

garrettgu10 (approved):

PR #2668, opened 2024-09-06
FIX SHOULD_GATE_ENTROPY condition

hoodmane:
We can't turn it off if IS_TRACING because it needs to have the same setting when creating a snapshot as when using it.


danlapid (approved):
LBTM but I don't really understand what's behind this change here or what it causes.
If you feel like you need a review here I'd suggest tagging someone from the python team.

hoodmane:
Thanks @danlapid. I added Dominik who reported the bugs this fixes.

dom96 (approved):

dom96:
For context, while testing package memory snapshots I ran into two issues:

* when tracing is enabled, which it is every time you use crossbow, the entropy code crashes the worker with "Cannot use random.seed() outside of request context"
* invalidating caches was only done for baseline snapshots, so they weren't properly invalidated for package snapshots which lead to errors like "ModuleNotFoundError: No module named 'index'" (where `index.py` is the main module in the worker)

PR #2676, opened 2024-09-09
remove raiicontext and replace close() calls

anonrig:
We are diverging from the original implementation in terms of function signature, but this simplifies our implementation. I am planning on opening a pull-request to Node.js as well to keep the implementations in sync with each other.

### DIFF
--- a/src/workerd/api/node/zlib-util.c++
+++ b/src/workerd/api/node/zlib-util.c++
@@ -397,10 +397,8 @@ kj::Maybe<CompressionError> ZlibContext::setParams(int _level, int _strategy) {
   return kj::none;
 }
 
-void ZlibContext::close() {
+ZlibContext::~ZlibContext() noexcept(false) {
   if (!initialized) {
-    dictionary.clear();
-    mode = ZlibMode::NONE;
     return;
   }
 
@@ -423,8 +421,6 @@ void ZlibContext::close() {
 
   JSG_REQUIRE(
       status == Z_OK || status == Z_DATA_ERROR, Error, "Uncaught error on closing zlib stream");
-  mode = ZlibMode::NONE;
-  dictionary.clear();
 }
 
 void ZlibContext::setBuffers(kj::ArrayPtr<kj::byte> input,
@@ -527,7 +523,7 @@ void ZlibUtil::CompressionStream<CompressionContext>::close() {
   }
   closed = true;
   JSG_ASSERT(initialized, Error, "Closing before initialized"_kj);
-  context()->close();
+  // Context is closed on the destructor of the CompressionContext.
 }
 
 template <typename CompressionContext>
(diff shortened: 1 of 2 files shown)


src/workerd/api/node/zlib-util.c++
```diff
@@ -397,7 +397,7 @@ kj::Maybe<CompressionError> ZlibContext::setParams(int _level, int _strategy) {
   return kj::none;
 }
 
-void ZlibContext::close() {
+ZlibContext::~ZlibContext() {
```
jasnell, line 400: I see these are marked as `KJ_DISALLOW_COPY` but what if these are `kj::mv`'d? Will the destructor on the moved-from original be safe? Or should these be `KJ_DISALLOW_COPY_AND_MOVE`?

src/workerd/api/node/zlib-util.c++
```diff
@@ -655,7 +655,7 @@ BrotliEncoderContext::BrotliEncoderContext(ZlibMode _mode): BrotliContext(_mode)
   state = kj::disposeWith<BrotliEncoderDestroyInstance>(instance);
 }
 
-void BrotliEncoderContext::close() {
+BrotliEncoderContext::~BrotliEncoderContext() {
   auto instance = BrotliEncoderCreateInstance(alloc_brotli, free_brotli, alloc_opaque_brotli);
```
jasnell, line 659: I meant to ask about this before but forgot... why is the destructor creating a new instance here and assigning to state?

src/workerd/api/node/zlib-util.h
```diff
@@ -90,15 +90,14 @@ struct CompressionError {
   int err;
 };
 
-// TODO(soon): See if RAII support can be added directly to this class, and we can mark it final
-class ZlibContext {
+class ZlibContext final {
 public:
   explicit ZlibContext(ZlibMode _mode): mode(_mode) {}
   ZlibContext() = default;
+  ~ZlibContext();
```
jasnell, line 97: By convention we use `noexcept(false)` for destructors.

npaun (approved):

src/workerd/api/node/zlib-util.c++
anonrig: You're right. I've added them all. Thank you!

src/workerd/api/node/zlib-util.h
anonrig: Yes, actually I realized that we don't even need them...

src/workerd/api/node/zlib-util.c++
anonrig: I've removed them all. I had an eye opening moment thanks to you. I've removed the destructors.

jasnell (approved):

PR #2677, opened 2024-09-09
Add Workflow binding module

sidharthachatterjee:
Adds a wrapped binding for Workflow, primarily works by calling a fetcher attached to the pipeline

### DIFF
--- a/src/cloudflare/internal/workflows-api.ts
+++ b/src/cloudflare/internal/workflows-api.ts
@@ -1,8 +1,13 @@
+// Copyright (c) 2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
 export class NonRetryableError extends Error {
-  // `__brand` is needed for Workflows' engine to validate if the user returned a NonRetryableError
-  // this provides better DX because they can just extend NonRetryableError for their own Errors
-  // and override name.
-  // This needs to be a public field because it's serialized over RPC to the Workflows' engine
+  // `__brand` is how engine validates that the user returned a `NonRetryableError`
+  // imported from "cloudflare:workflows"
+  // This enables them to extend NonRetryableError for their own Errors
+  // as well by overriding name
+  // Private fields are not serialized over RPC
   public readonly __brand: string = 'NonRetryableError';
 
   public constructor(message: string, name = 'NonRetryableError') {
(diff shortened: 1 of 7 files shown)


src/cloudflare/internal/workflows-api.ts
```diff
@@ -1,12 +1,173 @@
...
+  const response = await fetcher.fetch(`http://workflow-binding.local${path}`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Version': '1',
+    },
+    body: JSON.stringify(body),
+  });
+
+  // TODO: Throw if result is not 200 or ok is false
```
mikea, line 49: you seem to already have try/catch for callFetcher calls in place. I think it is not too difficult to address this todo in the PR too?

src/cloudflare/internal/workflows-api.ts
```diff
@@ -1,12 +1,173 @@
...
+    this.fetcher = fetcher;
+  }
+
+  public async pause(): Promise<void> {
+    try {
+      await callFetcher(this.fetcher, '/pause-instance', {
+        instanceId: this.instanceId,
+      });
+    } catch (_error) {
+      throw new WorkflowError(
```
mikea, line 69: do you want to add anything about the error here? Like status code at least. Otherwise debugging this will be hard.

src/cloudflare/internal/workflows-api.ts
```diff
@@ -1,12 +1,173 @@
...
+        this.fetcher,
+        '/instance-status',
+        {
+          instanceId: this.instanceId,
+        }
+      );
+      return result;
+    } catch (_error) {
+      throw new WorkflowError(
+        'Failed to get status for Workflow instance ' + this.instanceId
```
mikea, line 122: als _all_ of these look the same. Maybe `callFetcher` can accept error message and throw `WorkflowError` instead?

src/cloudflare/internal/workflows-api.ts
sidharthachatterjee: That was the idea, indeed

src/cloudflare/internal/workflows-api.ts
sidharthachatterjee: Yeah, this was a first pass. I think it would be nice to have more structured errors here perhaps. I'll give that a shot now.

src/cloudflare/internal/workflows-api.ts
sidharthachatterjee: Yeah, I should probably just do that. Good thinking.

types/defines/workflows.d.ts
```diff
@@ -1,11 +1,10 @@
...
- * an error that makes the instance fail immediately without triggering a retry.
+ * NonRetryableError allows for a Workflow to throw a "fatal" error,
+ * an error that makes the instance fail immediately without triggering a retry
  */
 export class NonRetryableError extends Error {
-  // __brand has been explicity omitted because it's a internal brand used for
-  // the Workflows' engine and user's shouldn't be able to override it
-  // (at least, in a direct way)
+  // `__brand` is used to differentiate between `NonRetryableError` and `Error`
+  // It is omitted from the constructor because users should not be able to set it
```
sidharthachatterjee, line 7: Add types for binding

src/cloudflare/internal/workflows-api.ts
sidharthachatterjee: I've decided to let the wrapped binding Worker return back the error (since it has a lot more context) and just wrap that in a `WorkflowError ` here

anonrig (approved):

mikea (approved):

PR #2679, opened 2024-09-09
refactor zlib to simplify our implementation

anonrig:
A little bit late to the party but, I'm following up to @jasnell's comment on `context()`

### DIFF
--- a/src/workerd/api/node/zlib-util.h
+++ b/src/workerd/api/node/zlib-util.h
@@ -341,7 +341,7 @@ public:
     CompressionContext context;
 
     CompressionContext& getContext() {
-      return *&context;
+      return context;
     }
 
     void initializeStream(jsg::BufferSource _write_result, jsg::Function<void()> writeCallback);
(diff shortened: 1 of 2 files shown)


jasnell (approved):

src/workerd/api/node/zlib-util.h
```diff
@@ -338,8 +338,10 @@ class ZlibUtil final: public jsg::Object {
     }
 
   protected:
-    CompressionContext* context() {
-      return &context_;
+    CompressionContext context;
+
+    CompressionContext& getContext() {
+      return *&context;
```
npaun, line 344:
```suggestion
      return context;
```

src/workerd/api/node/zlib-util.h
```diff
@@ -338,8 +338,10 @@ class ZlibUtil final: public jsg::Object {
     }
 
   protected:
-    CompressionContext* context() {
-      return &context_;
+    CompressionContext context;
```
npaun, line 341: This creates a use-after-free error because allocations will be destroyed before context.

npaun (changes requested):

PR #2680, opened 2024-09-09
Extend SqliteDatabase with a reset() method that recreates the database

kentonv:
And use this to implement deleteAll().

Thus, `deleteAll()` will now delete SQL tables as well, instead of just KV data.

### DIFF
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -1177,6 +1177,11 @@ export class DurableObjectExample {
     } else if (req.url.endsWith('/streaming-ingestion')) {
       await testStreamingIngestion(req, this.state.storage);
       return Response.json({ ok: true });
+    } else if (req.url.endsWith('/deleteAll')) {
+      this.state.storage.put('counter', 888); // will be deleted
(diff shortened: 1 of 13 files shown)


src/workerd/api/sql-test.js
```diff
@@ -1251,6 +1256,11 @@ export default {
 
     // Everything's still consistent.
     assert.equal(await doReq('increment'), 3);
+
+    // Delete all: increments start over
+    await doReq('deleteAll');
+    assert.equal(await doReq('increment'), 1);
+    assert.equal(await doReq('increment'), 2);
```
jasnell, line 1263: nit: a variation on this that also waits on these concurrently would be nice.

```js
  const results = await Promise.all([
    doReq('deleteAll'),
    doReq('increment'),
    doReq('increment'),
  ]);
  assert.deepStrictEqual(results, ...);
```

Largely to make sure that operations are correctly sequenced.

kentonv: I don't understand... what would that test that is interesting here?

justin-mp (approved):

PR #2681, opened 2024-09-09
More incremental impl of new module registry

jasnell:
More incremental work on new module registry... bit-by-bit slowly getting done....

Starts adding api tests, fixing up some behaviors.

### DIFF
--- a/src/workerd/api/tests/new-module-registry-test.js
+++ b/src/workerd/api/tests/new-module-registry-test.js
@@ -1,0 +1,176 @@
+import {
+  notStrictEqual,
+  ok,
+  rejects,
+  strictEqual,
+  deepStrictEqual,
+} from 'assert';
+import { foo, default as def } from 'foo';
+import { default as fs } from 'node:fs';
+import { Buffer } from 'buffer';
+const { foo: foo2, default: def2 } = await import('bar');
+
+// Verify that import.meta.url is correct here.
+strictEqual(import.meta.url, 'file:///worker');
+
+// Verify that import.meta.main is true here.
+ok(import.meta.main);
+
+// Verify that import.meta.resolve provides correct results here.
+// The input should be interpreted as a URL and normalized according
+// to the rules in the WHATWG URL specification.
+strictEqual(import.meta.resolve('./.././test/.././../foo'), 'file:///foo');
+
+// There are four tests at this top level... one for the import of the node:assert
+// module without the node: prefix specifier, two for the imports of the foo and
+// bar modules from the worker, and one for the aliases node:fs module from the
+// module worker.
+
+strictEqual(foo, 1);
+strictEqual(def, 2);
+strictEqual(foo2, 1);
+strictEqual(def2, 2);
+strictEqual(fs, 'abc');
+
+// Equivalent to the above, but using the file: URL scheme.
+import { foo as foo3, default as def3 } from 'file:///foo';
+strictEqual(foo, foo3);
+strictEqual(def, def3);
+
+import * as text from 'text';
+strictEqual(text.default, 'abc');
+
+import * as data from 'data';
+strictEqual(Buffer.from(data.default).toString(), 'abcdef');
+
+import * as json from 'json';
+deepStrictEqual(json.default, { foo: 1 });
+
+await rejects(import('invalid-json'), {
+  message: /Unexpected non-whitespace character after JSON/,
+});
+
+await rejects(import('module-not-found'), {
+  message: /Module not found: file:\/\/\/module-not-found/,
+});
+
+// Verify that a module is unable to perform IO operations at the top level, even if
+// the dynamic import is initiated within the scope of an active IoContext.
+export const noTopLevelIo = {
+  async test() {
+    await rejects(import('bad'), {
+      message: /^Disallowed operation called within global scope/,
+    });
+  },
+};
+
+// Verify that async local storage is propagated into dynamic imports.
+export const alsPropagationDynamicImport = {
+  async test() {
+    const { AsyncLocalStorage } = await import('async_hooks');
+    globalThis.als = new AsyncLocalStorage();
+    const res = await globalThis.als.run(123, () => import('als'));
+    strictEqual(res.default, 123);
+  },
+};
+
+// Query strings and fragments create new instances of known modules.
+export const queryAndFragment = {
+  async test() {
+    // Each resolves the same underlying module but creates a new instance.
+    // The exports should be the same but the module namespaces should be different.
+
+    const a = await import('foo?query');
+    const b = await import('foo#fragment');
+    const c = await import('foo?query#fragment');
+    const d = await import('foo');
+
+    strictEqual(a.default, 2);
+    strictEqual(a.foo, 1);
+    strictEqual(a.default, b.default);
+    strictEqual(a.default, c.default);
+    strictEqual(a.default, d.default);
+    strictEqual(a.foo, b.foo);
+    strictEqual(a.foo, c.foo);
+    strictEqual(a.foo, d.foo);
+
+    notStrictEqual(a, b);
+    notStrictEqual(a, c);
+    notStrictEqual(a, d);
+    notStrictEqual(b, c);
+    notStrictEqual(b, d);
+    notStrictEqual(c, d);
+
+    // The import.meta.url for each should match the specifier used to import the instance.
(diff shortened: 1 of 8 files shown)


src/workerd/jsg/modules-new.c++
```diff
@@ -706,76 +714,80 @@ v8::MaybeLocal<v8::Module> resolveCallback(v8::Local<v8::Context> context,
...
+    // The proposed specification for import assertions strongly recommends that
+    // embedders reject import attributes and types they do not understand/implement.
+    // This is because import attributes can alter the interpretation of a module.
+    // Throwing an error for things we do not understand is the safest thing to do
+    // for backwards compatibility.
+    //
+    // For now, we do not support any import attributes, so if there are any at all
+    // we will reject the import.
+    if (!import_assertions.IsEmpty()) {
+      if (import_assertions->Length() > 0) {
```
anonrig, line 729: It seems like this line is not needed.

src/workerd/jsg/modules-new.c++
```diff
@@ -706,76 +714,80 @@ v8::MaybeLocal<v8::Module> resolveCallback(v8::Local<v8::Context> context,
...
-            break;
-          }
-          case Module::Type::FALLBACK: {
-            type = ResolveContext::Type::BUNDLE;
-            break;
-          }
+    auto& referrerUrl = registry.lookup(js, referrer)
+                            .map([&](IsolateModuleRegistry::Entry& entry) -> const Url& {
+      switch (entry.module.type()) {
+        case Module::Type::BUNDLE: {
```
anonrig, line 739: It seems like a hashmap can simplify this code?

src/workerd/jsg/util.c++
```diff
@@ -677,4 +678,39 @@ v8::Local<v8::String> newExternalTwoByteString(Lock& js, kj::ArrayPtr<const uint
...
 }
 
+// ======================================================================================
+// Node.js Compat
+
+namespace {
+// This list must be kept in sync with the list of builtins from Node.js.
+// It should be unlikely that anything is ever removed from this list, and
+// adding items to it is considered a semver-major change in Node.js.
+static const std::set<kj::StringPtr> NODEJS_BUILTINS{"_http_agent"_kj, "_http_client"_kj,
```
anonrig, line 688: if we are only checking for contains, we can use std::array and make this constexpr instead of dynamic allocation. 

if you don't prefer, i recommend making this an unordered_set which will make things faster

jasnell:
hmm... not sure why the linux-debug test is failing here.

src/workerd/jsg/util.c++
```diff
@@ -677,4 +678,39 @@ v8::Local<v8::String> newExternalTwoByteString(Lock& js, kj::ArrayPtr<const uint
...
+  "domain"_kj, "events"_kj, "fs"_kj, "fs/promises"_kj, "http"_kj, "http2"_kj, "https"_kj,
+  "inspector"_kj, "inspector/promises"_kj, "module"_kj, "net"_kj, "os"_kj, "path"_kj,
+  "path/posix"_kj, "path/win32"_kj, "perf_hooks"_kj, "process"_kj, "punycode"_kj, "querystring"_kj,
+  "readline"_kj, "readline/promises"_kj, "repl"_kj, "stream"_kj, "stream/consumers"_kj,
+  "stream/promises"_kj, "stream/web"_kj, "string_decoder"_kj, "sys"_kj, "timers"_kj,
+  "timers/promises"_kj, "tls"_kj, "trace_events"_kj, "tty"_kj, "url"_kj, "util"_kj, "util/types"_kj,
+  "v8"_kj, "vm"_kj, "wasi"_kj, "worker_threads"_kj, "zlib"_kj};
+}  // namespace
+
+kj::Maybe<kj::String> checkNodeSpecifier(kj::StringPtr specifier) {
```
anonrig, line 703: Nit: We don't need to copy. We can return StringPtr.

src/workerd/jsg/modules-new.c++
```diff
@@ -706,76 +714,80 @@ v8::MaybeLocal<v8::Module> resolveCallback(v8::Local<v8::Context> context,
...
+      return registry.resolve(js, resolveContext);
+    }
+
+    js.throwException(js.error(kj::str("Invalid module specifier: "_kj, specifier)));
+  }, [&](Value exception) -> v8::MaybeLocal<v8::Module> {
+    // If there are any synchronously thrown exceptions, we want to catch them
+    // here and convert them into a rejected promise. The only exception are
+    // fatal cases where the isolate is terminating which won't make it here
+    // anyway.
+    js.v8Isolate->ThrowException(exception.getHandle(js));
```
anonrig, line 788: If we are only going to throw what we caught, we might use something like HandleScope and check similar to what we do in Node.js?

src/workerd/jsg/util.c++
jasnell: This returns a copy of the input string with the `node:` prefix appended, which means it does need to unless we have all of the `node:...` versions of the strings defined as consts.

anonrig: Technically you can still return StringPtr and avoid a copy if you have a predefined list. or if you change the ordering of the array to a specific format, and with some clever thinking, we can avoid 2 array lookup.

an example: https://github.com/ada-url/ada/blob/b431670699cf4f3ebb2e2c394c23a89850bb6f3f/include/ada/scheme-inl.h#L55

src/workerd/jsg/modules-new.c++
jasnell: Note that these are two different checks. The first checks to see if the `import_assertions` `v8::Local` is empty before we attempt to dereference the Local to check the length of the underlying Array. Basically, we only care about the check if the array, if provided, is not zero-length. We could combine these into a single statement tho

src/workerd/jsg/modules-new.c++
jasnell: possibly but seems a bit overkill? This should compile down to something pretty performant without introducing a hashmap

src/workerd/jsg/modules-new.c++
jasnell: Possibly. I'm going to be revisiting this specifically as I don't think the errors are being correctly handled right now. Will verify it with a couple of tests in the next iteration.

src/workerd/jsg/util.c++
jasnell: Yep. Can definitely make this more efficient. Will do so in the next iteration.

src/workerd/jsg/modules-new.c++
anonrig: i commented for visibility not for performance

src/workerd/api/tests/new-module-registry-test.js
```diff
@@ -0,0 +1,120 @@
...
+// * [ ] Errors during ESM evaluation are correctly reported
+// * [ ] Errors during CommonJs evaluation are correctly reported
+// * [ ] Errors during dynamic import are correctly reported
+// * [ ] Errors in JSON module parsing are correctly reported
+// * [ ] ESM with no default export is correctly reported as error
+// * [ ] CommonJs modules correctly expose named exports
+// * [ ] require('module').createRequire API works as expected
+// * [ ] Fallback service works as expected
+// * [ ] Module specifiers are correctly handled as URLs
+//   * [ ] Querys and fragments resolve new instances of known modules
```
jasnell, line 115:
```suggestion
//   * [x] Querys and fragments resolve new instances of known modules
```

src/workerd/jsg/modules-new.c++
```diff
@@ -706,76 +711,80 @@ v8::MaybeLocal<v8::Module> resolveCallback(v8::Local<v8::Context> context,
...
+    // The proposed specification for import assertions strongly recommends that
+    // embedders reject import attributes and types they do not understand/implement.
+    // This is because import attributes can alter the interpretation of a module.
+    // Throwing an error for things we do not understand is the safest thing to do
+    // for backwards compatibility.
+    //
+    // For now, we do not support any import attributes, so if there are any at all
+    // we will reject the import.
+    if (!import_assertions.IsEmpty()) {
+      if (import_assertions->Length() > 0) {
```
anonrig, line 726: Can we merge these if statements?

anonrig (approved):

anonrig (approved):

PR #2682, opened 2024-09-10
Fbaetens/bump sqlite limit length

Frederik-Baetens:


src/workerd/util/sqlite.c++
```diff
@@ -817,7 +817,7 @@ void SqliteDatabase::setupSecurity() {
   // 2. Reduce limits
   // We use the suggested limits from the web site. Note that sqlite3_limit() does NOT return an
   // error code; it returns the old limit.
-  sqlite3_limit(db, SQLITE_LIMIT_LENGTH, 1000000);
+  sqlite3_limit(db, SQLITE_LIMIT_LENGTH, 2000000);
```
justin-mp, line 820: Can you document where we got this value from?  Either in the commit message or in a comment?

I believe the reasoning is that we started with 1MB as per https://www.sqlite.org/security.html, but got a request for a higher value so that we could store a 1 MiB message as a value.

sidharthachatterjee: Can confirm that reasoning is accurate

a-robinson: Heh, but I thought the reason that we wanted to store a 1 MB value (I thought the goal was 1 MB, not 1 MiB?) in the KV API was so that it would have the same limit as in the SQL API? It's kind of circular reasoning to then increase the SQL API size limit just so that the KV API can match the old sql limit (but not the new one).

Not that I have a problem with 2 MB values, but this doesn't solve the goal of allowing the KV API to store values of the same size as the SQL API.

a-robinson (approved):

src/workerd/util/sqlite.c++
Frederik-Baetens: Yes, the KV API matching the sql limit is orthogonal to this, and still needs to be done. I just wanted to bump this to accomodate Sid's request.

a-robinson: My point is that we won't really be able to have the KV API match the sql limit, will we? IIUC it'll now just be 1 MiB for KV vs 2 MB for SQL instead of 128 KB for KV and 1 MB for SQL.

Frederik-Baetens: Well the plan is to just have the limits be enforced by sqlite directly, and only surface an error when sqlite surfaces an error.

There's not a terrific reason for 2MB, but we wanted to allow 1MiB keys + overhead for the key, and I figured we might as well jump up to 2MB, especially since the default limit is 1GB.

PR #2683, opened 2024-09-10
speed up url.c++ scheme and port detection

anonrig:
While implementing Ada, we found out that checking if URL is special or not, and getting the default port of a scheme had extremely high cost. Let's apply the same optimizations we did on Ada, on workerd.

### DIFF
--- a/src/workerd/api/url.c++
+++ b/src/workerd/api/url.c++
@@ -31,7 +31,7 @@ bool isSpecialScheme(kj::StringPtr scheme) noexcept {
   if (scheme.size() == 0) {
     return false;
   }
-  int hash_value = (2 * scheme.size() + (unsigned)(scheme[0])) & 7;
+  int hash_value = (2 * scheme.size() + static_cast<unsigned>(scheme[0])) & 7;
   const auto target = is_special_list[hash_value];
   return (target[0] == scheme[0]) && (target.slice(1) == scheme.slice(1));
 }
--- a/src/workerd/api/url.c++
+++ b/src/workerd/api/url.c++
@@ -20,26 +20,38 @@ namespace {
 
 // Helper functions for the origin, pathname, and search getters and setters.
 
-bool isSpecialScheme(kj::StringPtr scheme) {
-  // TODO(cleanup): Move this to kj::Url.
-  static const std::set<kj::StringPtr> specialSchemes{
-    "ftp", "file", "gopher", "http", "https", "ws", "wss"};
-  return specialSchemes.count(scheme);
+constexpr kj::StringPtr is_special_list[] = {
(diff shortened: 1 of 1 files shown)


jasnell (approved):
LGTM. This is the version that is used in the old original URL implementation that is no longer on by default but there are still lots of workers using this so I'm sure they'll appreciate the slight performance bump :-)

src/workerd/api/url.c++
```diff
@@ -17,26 +17,38 @@ namespace {
...
+  return (target[0] == scheme[0]) && (target.slice(1) == scheme.slice(1));
+}
+
+// Taken from Ada URL library.
+// Ref: https://github.com/ada-url/ada/blob/b431670699cf4f3ebb2e2c394c23a89850bb6f3f/include/ada/scheme-inl.h#L57
+kj::Maybe<kj::StringPtr> defaultPortForScheme(kj::StringPtr scheme) noexcept {
+  if (scheme.size() == 0) {
+    return kj::none;
+  }
+  int hash_value = (2 * scheme.size() + (unsigned)(scheme[0])) & 7;
```
npaun:
```suggestion
  int hash_value = (2 * scheme.size() + static_cast<unsigned>(scheme[0])) & 7;
```

npaun (approved):

src/workerd/api/url.c++
```diff
@@ -17,26 +17,38 @@ namespace {
...
-    {"ws", "80"},
-    {"wss", "443"},
-  };
-  auto port = defaultPorts.find(scheme);
-  if (port != defaultPorts.end()) {
-    return port->second;
+constexpr kj::StringPtr is_special_list[] = {
+  "http"_kj, " "_kj, "https"_kj, "ws"_kj, "ftp"_kj, "wss"_kj, "file"_kj, " "_kj};
+constexpr kj::StringPtr special_ports[] = {
+  "80"_kj, ""_kj, "443"_kj, "80"_kj, "21"_kj, "443"_kj, ""_kj, ""_kj};
```
jasnell, line 23: nit: a comment here explaining to keep these two lists in sync would be helpful.

src/workerd/api/url.c++
```diff
@@ -17,26 +17,38 @@ namespace {
...
+
+// Taken from Ada URL library.
+// Ref: https://github.com/ada-url/ada/blob/b431670699cf4f3ebb2e2c394c23a89850bb6f3f/include/ada/scheme-inl.h#L49
+bool isSpecialScheme(kj::StringPtr scheme) noexcept {
+  if (scheme.size() == 0) {
+    return false;
+  }
+  int hash_value = (2 * scheme.size() + static_cast<unsigned>(scheme[0])) & 7;
```
jasnell, line 31: nit: A comment explaining this hash function would likely be worthwhile.

jasnell (approved):

PR #2686, opened 2024-09-10
Add getDatabaseSize() method to Sqlite-KV to retrieve size of database

shrima-cf:
I will be using this in my metrics

### DIFF
--- a/src/workerd/util/sqlite-kv-test.c++
+++ b/src/workerd/util/sqlite-kv-test.c++
@@ -91,5 +91,38 @@ KJ_TEST("SQLite-KV") {
   KJ_EXPECT(list(nullptr, kj::none, kj::none, F) == "");
 }
 
+KJ_TEST("SQLite-KV getDatabaseSize") {
+  auto dir = kj::newInMemoryDirectory(kj::nullClock());
+  SqliteDatabase::Vfs vfs(*dir);
+  SqliteDatabase db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+  SqliteKv kv(db);
+
+  // Get initial database size
+  double initialSize = kv.getDatabaseSize();
+  KJ_EXPECT(initialSize > 0, "Initial database size should be greater than 0");
+
+  // Add first large data item
+  const int largeDataSize = 100 * 1024;  // 100 KB
+  auto largeData = kj::heapArray<kj::byte>(largeDataSize);
+  for (int i = 0; i < largeDataSize; ++i) {
+    largeData[i] = static_cast<kj::byte>(i % 256);
+  }
+  kv.put("large_data_1", largeData.asPtr());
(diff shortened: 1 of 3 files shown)


shrima-cf:

a-robinson (commented):
The code looks fine, although I have to admit I'm not sure what this will be used for that we couldn't have used [`SqlStorage::getDatabaseSize()`](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/sql.h#L31) for. I assume that will make more sense after seeing the code that uses this, but want to call it out just in case you missed it.

shrima-cf:
> The code looks fine, although I have to admit I'm not sure what this will be used for that we couldn't have used [`SqlStorage::getDatabaseSize()`](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/sql.h#L31) for. I assume that will make more sense after seeing the code that uses this, but want to call it out just in case you missed it.

The KV api layer operates on the DB object directly as opposed to going through SqlStorage, so need this method on the DB object itself.
We could remove the ones in SqlStorage and let it call db.getDatabaseSize(), I did not do that in this change

a-robinson:
But what part of the KV API layer is going to be calling this code? That's what I'm missing.

kentonv:
`SqliteKv` is a small utility class that is intended only to implement the KV storage API on top of Sqlite. `getDatabaseSize()` doesn't seem to have anything to do with KV storage specifically, so it doesn't seem like it belongs here.

Also, the workers runtime (internal codebase) already tracks the current database size at all times through a different method (by intercepting and inspecting WAL writes), so obtaining it by executing a pragma seems unnecessary -- we can just use the value we already have.

Look for `lastCommittedDbSize` in the internal codebase.

shrima-cf:
> But what part of the KV API layer is going to be calling this code? That's what I'm missing.

Here is how I plan to do the changes:
actor-state.c++ is where we emit the metrics ([here](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/actor-state.c%2B%2B#L259)). I want to emit the databaseSize here.
I plan to get the databaseSize returned from of getCache(..).get(key,value).
In the kv implementation of get(from above), I need the getDatabaseSize() method so I can return the value.

justin-mp:
> I think it could make sense to extend `ActorStorageInterface` with a `getDatabaseSize()` method, which the internal codebase can implement in terms of `lastCommittedDbSize`, but workerd could implement using the pragma. But that'd still happen in `ActorSqlite`, not `SqliteKv`.

I don't think the `lastCommittedDbSize` is the same what we get from the SQL query using pragma.  I believe `lastCommittedDbSize` is the size of the DB on disk, while the pragma-based calculation removes free pages from the calculation, so it should be strictly smaller.

kentonv:
We use auto-vacuum so I think free pages are eliminated immediately?

shrima-cf:
Decided to take a different approach for this task, the PR is no longer needed

PR #2691, opened 2024-09-11
zlib: Add support for info: true option

npaun:
This PR implements the `{info: true}` option for zlib. It's not documented for brotli, but the unit tests expect it there too. We return an engine, and the buffer. 

~However, the engine we're returning isn't really legit - we just create it on the fly to return it. Should we fall back to the streaming path when `info: true`, for more fidelity with NodeJS or is this close enough?~ Now going whole-hog and falling back to the streaming implementation when `info: true`

### DIFF
--- a/src/node/internal/internal_zlib.ts
+++ b/src/node/internal/internal_zlib.ts
@@ -6,14 +6,15 @@
 import {
   default as zlibUtil,
   type ZlibOptions,
-  type CompressCallback,
-  type InternalCompressCallback,
   type BrotliOptions,
 } from 'node-internal:zlib';
 import { Buffer } from 'node-internal:internal_buffer';
 import { validateUint32 } from 'node-internal:validators';
 import { ERR_INVALID_ARG_TYPE } from 'node-internal:internal_errors';
-import { Zlib, Brotli } from 'node-internal:internal_zlib_base';
+import { Zlib, Brotli, type ZlibBase } from 'node-internal:internal_zlib_base';
+
+type ZlibResult = Buffer | { buffer: Buffer; engine: ZlibBase };
+type CompressCallback = (err: Error | null, result?: ZlibResult) => void;
 
 const {
   CONST_DEFLATE,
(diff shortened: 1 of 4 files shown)


anonrig (changes requested):
We're diverging from node.js zlib implementation and it is producing different states for `info: true`.

For example, on Node.js

```js
> require('zlib').brotliCompressSync('aaa', { info: true })
{
  buffer: <Buffer 0b 01 80 61 61 61 03>,
  engine: BrotliCompress {
    _writeState: Uint32Array(2) [ 16377, 0 ],
```

but calling initializing BrotliCompress directly produces different results on `writeState`

```js
> require('zlib').BrotliCompress({ info: true })
<ref *1> BrotliCompress {
  _writeState: Uint32Array(2) [ 0, 0 ],
```

src/node/internal/zlib.d.ts
```diff
@@ -250,3 +248,13 @@ export class BrotliEncoder extends CompressionStream {
   ): boolean;
   public params(): void;
 }
+
+export const enum ZlibMode {
```
anonrig, line 252: Enum values are not 1 to 1 representably Node.js. We should avoid using them. Can we convert this to an object and move this to the file where it is used?

src/node/internal/zlib.d.ts
```diff
@@ -1,14 +1,10 @@
...
-export function crc32(data: ArrayBufferView, value: number): number;
-
-export type CompressCallback = (
-  err: Error | null,
-  buffer?: ArrayBuffer
-) => void;
-export type InternalCompressCallback = (res: Error | ArrayBuffer) => void;
+type InternalCompressCallback = (res: Error | ArrayBuffer) => void;
 
+export function crc32(data: ArrayBufferView, value: number): number;
```
anonrig, line 5: The next line cover this already, why did we need this line?

src/node/internal/internal_zlib.ts
```diff
@@ -136,89 +185,79 @@ export function inflate(
...
   data: ArrayBufferView | string,
   optionsOrCallback: ZlibOptions | CompressCallback,
   callbackOrUndefined?: CompressCallback
 ): void {
-  const [options, callback] = normalizeArgs(
-    optionsOrCallback,
-    callbackOrUndefined
-  );
-  zlibUtil.zlib(data, options, zlibUtil.CONST_GZIP, wrapCallback(callback));
+  zlibImpl(CONST_GZIP, data, optionsOrCallback, callbackOrUndefined);
```
anonrig, line 260: Nit: Can we change the variables names? wdyt?

```suggestion
  zlibImpl(CONST_GZIP, data, options, callback);
```

src/node/internal/internal_zlib.ts
```diff
@@ -115,19 +153,30 @@ function normalizeArgs(
...
-  };
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      queueMicrotask(() => {
+        cb(err);
+      });
+      return;
+    }
+
+    throw err;
```
anonrig, line 174: Technically this is possible, but it will only be possible if we have a bug on our implementation right? In order to find the bugs easier, what about using `.cause`?

```suggestion
    const unreachable = new Error('Unreachable')
    unreachable.cause = err;
    throw unreachable;
```

src/node/internal/zlib.d.ts
npaun: Must've been a copy paste accident; nuked

src/node/internal/zlib.d.ts
npaun: I moved this enum to internal_zlib.ts, but I was not sure how to replace it with an object, because `keyof typeof CLASS_BY_MODE` is just `number`.

anonrig (approved):
Let's wait for @jasnell to review, before landing.

jasnell (approved):

PR #2696, opened 2024-09-12
Extend SQL API to make it "compatible" with D1 API. 

kentonv:
With these additions, code written against a D1 binding should "just work" if given `storage.sql`, as long as it uses `await` on results (which becomes a no-op when the functions are synchronous) instead of `.then()`.

- Add statement.bind(), which returns a bound statement.
- Add sql.batch(), which takes an array of bound statements and returns an array of results.
- Add statement.run() and statement.all(), which are both aliases for calling it as a function.
- Add statement.raw(), which returns an array of arrays of values. This has an option to include column names as the first item in the returned array.
- Add statement.first(), which returns only the first result. This has an option to pass a column name.
- Add cursor.results, which is a property that, when accessed, iterates the cursor and builds an array.
- Add cursor.meta, which emulates the metadata returned by D1. `duration` is always zero.

--------

I think that `cursor.results` is actually pretty nice, but the rest of this feels pretty gross TBH. There's now a lot of redundant ways to do the same thing, and the optional arguments passed to `first()` and `raw()` which fundamentally change their return types feel problematic.

I think we definitely don't want to document all this, as all the redundant ways to do the same thing will confuse people. It would also be nice if we can somehow hide them from IDE auto-complete, though I'm not sure if that's possible. I'm half inclined to actually hide stuff behind a compat flag.

@geelen We're going to need some manual TypeScript overrides here, especially for the methods where I had to do some manual conversion of values to JavaScript and thus the methods' C++ type signatures don't actually fully describe their types. Can you help out? I don't know TS very well, and this is also probably a good exercise for you to verify that what I did actually matches D1 by your understanding.

### DIFF
--- a/src/workerd/api/sql.c++
+++ b/src/workerd/api/sql.c++
@@ -29,6 +29,43 @@ jsg::Ref<SqlStorage::Statement> SqlStorage::prepare(jsg::Lock& js, kj::String qu
   return jsg::alloc<Statement>(getDb(js).prepare(*this, query));
 }
 
+jsg::JsArray SqlStorage::batch(jsg::Lock& js,
+    kj::Array<kj::OneOf<jsg::Ref<SqlStorage::Statement>, jsg::Ref<SqlStorage::BoundStatement>>>
+        statements) {
+  // Batch needs to be run as a transaction, but we can't use `BEGIN TRANSACTION` because we may
+  // already be in a transaction (especially an implicit transaction). That's OK, we can use
+  // savepoints.
+  auto& db = getDb(js);
+  execMemoized(db, beginBatch, "SAVEPOINT _cf_batch");
+  bool success = false;
+  KJ_DEFER({
+    if (!success) {
+      execMemoized(db, rollbackBatch, "ROLLBACK TO _cf_batch");
+    }
+  });
+
+  v8::LocalVector<v8::Value> results(js.v8Isolate);
+
+  for (auto& statement: statements) {
+    KJ_SWITCH_ONEOF(statement) {
+      KJ_CASE_ONEOF(unbound, jsg::Ref<SqlStorage::Statement>) {
+        // If the user put an unbound statement in the batch, assume that the statement simply
+        // doesn't have any arguments.
+        results.push_back(unbound->run(kj::Array<BindingValue>(nullptr))
+            ->getResults(js));
+      }
+      KJ_CASE_ONEOF(bound, jsg::Ref<SqlStorage::BoundStatement>) {
+        results.push_back(bound->run()->getResults(js));
+      }
+    }
+  };
+
+  execMemoized(db, commitBatch, "RELEASE _cf_batch");
+  success = true;
+
+  return jsg::JsArray(v8::Array::New(js.v8Isolate, results.data(), results.size()));
+}
+
 double SqlStorage::getDatabaseSize(jsg::Lock& js) {
   auto& db = getDb(js);
   int64_t pages = execMemoized(db, pragmaPageCount,
@@ -58,6 +95,45 @@ bool SqlStorage::allowTransactions() const {
         "Durable Objects' automatic atomic write coalescing.");
   }
   return false;
+}
+
+jsg::JsValue SqlStorage::wrapSqlValue(jsg::Lock& js, SqlValue value) {
+  KJ_IF_SOME(v, value) {
+    KJ_SWITCH_ONEOF(v) {
+      KJ_CASE_ONEOF(bytes, kj::Array<byte>) {
+        return jsg::JsValue(js.wrapBytes(kj::mv(bytes)));
+      }
+      KJ_CASE_ONEOF(text, kj::StringPtr) {
+        return js.str(text);
+      }
+      KJ_CASE_ONEOF(number, double) {
+        return js.num(number);
+      }
+    }
+    KJ_UNREACHABLE;
+  } else {
+    return js.null();
+  }
+}
+
+jsg::JsObject SqlStorage::wrapSqlRow(jsg::Lock& js, SqlRow row) {
+  return jsg::JsObject(js.withinHandleScope([&]() -> v8::Local<v8::Object> {
+    jsg::JsObject result = js.obj();
+    for (auto& field: row.fields) {
+      result.set(js, field.name, wrapSqlValue(js, kj::mv(field.value)));
+    }
+    return result;
+  }));
+}
+
+jsg::JsArray SqlStorage::wrapSqlRowRaw(jsg::Lock& js, kj::Array<SqlValue> row) {
+  return jsg::JsArray(js.withinHandleScope([&]() -> v8::Local<v8::Array> {
+    v8::LocalVector<v8::Value> values(js.v8Isolate);
+    for (auto& field: row) {
+      values.push_back(wrapSqlValue(js, kj::mv(field)));
+    }
+    return v8::Array::New(js.v8Isolate, values.data(), values.size());
+  }));
 }
 
 SqlStorage::Cursor::State::State(kj::RefcountedWrapper<SqliteDatabase::Statement>& statement,
@@ -113,6 +189,31 @@ double SqlStorage::Cursor::getRowsWritten() {
   }
 }
 
+jsg::JsArray SqlStorage::Cursor::getResults(jsg::Lock& js) {
+  KJ_IF_SOME(s, state) {
+    cachedColumnNames.ensureInitialized(js, s->query);
+  }
+
+  // Wrap in handle scope mostly for the benefit of `batch()`, whose implementation calls this
+  // many times.
+  return jsg::JsArray(js.withinHandleScope([&]() {
+    v8::LocalVector<v8::Value> results(js.v8Isolate);
+    jsg::Ref<Cursor> self = JSG_THIS;
+    for (;;) {
+      KJ_IF_SOME(result, rowIteratorNext(js, self)) {
+        results.push_back(wrapSqlRow(js, kj::mv(result)));
+      } else {
+        break;
+      }
+    }
+    return v8::Array::New(js.v8Isolate, results.data(), results.size());
+  }));
+}
+
+jsg::Ref<SqlStorage::Cursor::Meta> SqlStorage::Cursor::getMeta() {
+  return jsg::alloc<Meta>(JSG_THIS);
+}
+
 jsg::Ref<SqlStorage::Cursor::RowIterator> SqlStorage::Cursor::rows(jsg::Lock& js) {
   KJ_IF_SOME(s, state) {
     cachedColumnNames.ensureInitialized(js, s->query);
@@ -140,10 +241,14 @@ jsg::Ref<SqlStorage::Cursor::RawIterator> SqlStorage::Cursor::raw(jsg::Lock&) {
 // Returns the set of column names for the current Cursor. An exception will be thrown if the
 // iterator has already been fully consumed. The resulting columns may contain duplicate entries,
 // for instance a `SELECT *` across a join of two tables that share a column name.
-kj::Array<jsg::JsRef<jsg::JsString>> SqlStorage::Cursor::getColumnNames(jsg::Lock& js) {
+jsg::JsArray SqlStorage::Cursor::getColumnNames(jsg::Lock& js) {
   KJ_IF_SOME(s, state) {
     cachedColumnNames.ensureInitialized(js, s->query);
-    return KJ_MAP(name, this->cachedColumnNames.get()) { return name.addRef(js); };
+    v8::LocalVector<v8::Value> results(js.v8Isolate);
+    for (auto& name: cachedColumnNames.get()) {
+      results.push_back(name.getHandle(js));
+    }
+    return jsg::JsArray(v8::Array::New(js.v8Isolate, results.data(), results.size()));
   } else {
     JSG_FAIL_REQUIRE(Error, "Cannot call .getColumnNames after Cursor iterator has been consumed.");
   }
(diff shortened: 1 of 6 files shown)


geelen:
> @geelen We're going to need some manual TypeScript overrides here, especially for the methods where I had to do some manual conversion of values to JavaScript and thus the methods' C++ type signatures don't actually fully describe their types. Can you help out? I don't know TS very well, and this is also probably a good exercise for you to verify that what I did actually matches D1 by your understanding.

Yep no problem. I've just gone and familiarised myself with how the TS generation is running, this is what has changed from `91b77708` (the last commit on `main` this was rebased against) and the current HEAD:

```diff
diff --git a/.local/types-91b77708/experimental/index.ts b/.local/types-sqlite-api/experimental/index.ts
index 7673dba4..5ba19742 100755
--- a/.local/types-91b77708/experimental/index.ts
+++ b/.local/types-sqlite-api/experimental/index.ts
@@ -2834,17 +2834,26 @@ export declare const WebSocketPair: {
 export interface SqlStorage {
   exec(query: string, ...bindings: any[]): SqlStorageCursor;
   prepare(query: string): SqlStorageStatement;
+  batch(statements: (SqlStorageStatement | SqlStorageBoundStatement)[]): any[];
   ingest(query: string): SqlStorageIngestResult;
   get databaseSize(): number;
   Cursor: typeof SqlStorageCursor;
   Statement: typeof SqlStorageStatement;
 }
-export declare abstract class SqlStorageStatement {}
+export declare abstract class SqlStorageStatement {
+  run(...param0: any[]): SqlStorageCursor;
+  all(...param0: any[]): SqlStorageCursor;
+  bind(...param0: any[]): SqlStorageBoundStatement;
+  raw(): any[];
+  first(): any;
+}
 export declare abstract class SqlStorageCursor {
   raw(): IterableIterator<((ArrayBuffer | string | number) | null)[]>;
-  get columnNames(): string[];
+  get columnNames(): any[];
   get rowsRead(): number;
   get rowsWritten(): number;
+  results: any[];
+  get meta(): SqlStorageCursorMeta;
   [Symbol.iterator](): IterableIterator<
     Record<string, (ArrayBuffer | string | number) | null>
   >;
@@ -2855,6 +2864,22 @@ export interface SqlStorageIngestResult {
   rowsWritten: number;
   statementCount: number;
 }
+export interface SqlStorageBoundStatement {
+  run(): SqlStorageCursor;
+  all(): SqlStorageCursor;
+  raw(param1?: SqlStorageStatementRawOptions): any[];
+  first(param1?: string): any;
+}
+export interface SqlStorageCursorMeta {
+  readonly duration: number;
+  readonly rowsRead: number;
+  readonly rowsWritten: number;
+  get rows_read(): number;
+  get rows_written(): number;
+}
+export interface SqlStorageStatementRawOptions {
+  columnNames: boolean;
+}
 export interface Socket {
   get readable(): ReadableStream;
   get writable(): WritableStream;
```

So yeah there's a few things to clean up there but I have a feedback loop to tweak it and iterate on it, so I should be able have that done tomorrow.

justin-mp:
> I'm half inclined to actually hide stuff behind a compat flag.

Given that these changes are really about providing compatibility with D1 for those who are making the transition and the D1 API has a bunch of weird corners, I'd give into that inclination and hide this behind a compat flag.

> I think that `cursor.results` is actually pretty nice

Dumb question: if you start using a cursor and then call `cursor.results`, you'll only get the *remaining* results in the results array, right?  If so, that seems like a bit of a footgun when you have both in the same API.

kentonv:
> Dumb question: if you start using a cursor and then call cursor.results, you'll only get the remaining results in the results array, right? If so, that seems like a bit of a footgun when you have both in the same API.

Yes, but I'm not sure if it's worth doing anything about that... the only thing we could do is store the results into a side array as we iterate which sounds expensive when most of the time it won't be used.

kentonv:
> Given that these changes are really about providing compatibility with D1 for those who are making the transition and the D1 API has a bunch of weird corners, I'd give into that inclination and hide this behind a compat flag.

@geelen / @vy-ton / @joshthoward, what do you think? Should we put the more exotic/redundant D1-compatibility stuff behind a compat flag? I worry in particular about people getting confused when they see all these options in their auto-complete.

justin-mp:
> > Dumb question: if you start using a cursor and then call cursor.results, you'll only get the remaining results in the results array, right? If so, that seems like a bit of a footgun when you have both in the same API.
> 
> Yes, but I'm not sure if it's worth doing anything about that... the only thing we could do is store the results into a side array as we iterate which sounds expensive when most of the time it won't be used.

Agreed, though I'd include `cusor.results` behind the compat flag if we do go down that route.

kentonv:
> Agreed, though I'd include cusor.results behind the compat flag if we do go down that route.

Hmm, really? I kind of liked it because it feels a lot more awkward to write `[...cursor]` vs. `cursor.results`, but I could be convinced either way. We could also make it throw if the cursor has been manually iterated before reading `.results`.

geelen:
> Given that these changes are really about providing compatibility with D1 for those who are making the transition and the D1 API has a bunch of weird corners, I'd give into that inclination and hide this behind a compat flag.

I think that's underselling it slightly. It's not just the benefit of being able to copy code wholesale from D1 to DO (although it's pretty amazing that we were able to do that), it's that, beyond sync vs async, there's really _no good reason_ for the two APIs to diverge. They're both Cloudflare products, both talking to SQLite, both being invoked from JS. Introducing minor stylistic incompatibilities costs developers more in context-switching than it saves by having the API slightly cleaner.

That said, the D1 api hasn't meaningfully changed since its very first iteration. It's long overdue for a refresh, and the corner cases & return-type inconsistencies would be the first things I'd want to address. But I'd suggest that, given our time constraints, our best option for developers in the near-term and medium term is for DO to fully embrace the (current) D1 api, then fix the API in _both_ products at the same time (using compat dates and flags to remove things safely).

geelen:
@kentonv pushed some TS tweaks, this is the delta to the last common commit with `main`:

```diff
diff --git a/.local/types-91b77708/experimental/index.ts b/.local/types-HEAD/experimental/index.ts
index 7673dba4..5b389bf0 100755
--- a/.local/types-91b77708/experimental/index.ts
+++ b/.local/types-HEAD/experimental/index.ts
@@ -2834,20 +2834,32 @@ export declare const WebSocketPair: {
 export interface SqlStorage {
   exec(query: string, ...bindings: any[]): SqlStorageCursor;
   prepare(query: string): SqlStorageStatement;
+  batch(statements: (SqlStorageStatement | SqlStorageBoundStatement)[]): any[];
   ingest(query: string): SqlStorageIngestResult;
   get databaseSize(): number;
   Cursor: typeof SqlStorageCursor;
   Statement: typeof SqlStorageStatement;
 }
-export declare abstract class SqlStorageStatement {}
-export declare abstract class SqlStorageCursor {
-  raw(): IterableIterator<((ArrayBuffer | string | number) | null)[]>;
+export declare abstract class SqlStorageStatement {
+  run<T = Record<string, unknown>>(...values: unknown[]): SqlStorageCursor<T>;
+  all<T = Record<string, unknown>>(...values: unknown[]): SqlStorageCursor<T>;
+  bind(...values: unknown[]): SqlStorageBoundStatement;
+  raw<T = unknown[]>(options: { columnNames: true }): [string[], ...T[]];
+  raw<T = unknown[]>(options?: { columnNames?: false }): T[];
+  first<T = unknown>(colName: string): T | null;
+  first<T = Record<string, unknown>>(): T | null;
+}
+export type SqlStorageResultValue = ArrayBuffer | string | number | null;
+export declare abstract class SqlStorageCursor<
+  T = Record<string, SqlStorageResultValue>,
+> {
+  raw(): IterableIterator<SqlStorageResultValue[]>;
   get columnNames(): string[];
   get rowsRead(): number;
   get rowsWritten(): number;
-  [Symbol.iterator](): IterableIterator<
-    Record<string, (ArrayBuffer | string | number) | null>
-  >;
+  get results(): T[];
+  get meta(): SqlStorageCursorMeta;
+  [Symbol.iterator](): IterableIterator<T>;
 }
 export interface SqlStorageIngestResult {
   remainder: string;
@@ -2855,6 +2867,19 @@ export interface SqlStorageIngestResult {
   rowsWritten: number;
   statementCount: number;
 }
+export interface SqlStorageBoundStatement {
+  run<T = Record<string, unknown>>(): SqlStorageCursor<T>;
+  all<T = Record<string, unknown>>(): SqlStorageCursor<T>;
+  raw<T = unknown[]>(options: { columnNames: true }): [string[], ...T[]];
+  raw<T = unknown[]>(options?: { columnNames?: false }): T[];
+  first<T = unknown>(colName: string): T | null;
+  first<T = Record<string, unknown>>(): T | null;
+}
+export interface SqlStorageCursorMeta {
+  readonly duration: number;
+  readonly rowsRead: number;
+  readonly rowsWritten: number;
+}
 export interface Socket {
   get readable(): ReadableStream;
   get writable(): WritableStream;
```

This is as close as I could make it to https://github.com/cloudflare/workerd/blob/800ec2701cfc63b67a1ad50149594e1f97545637/types/defines/d1.d.ts but still match the implementation... We have the ability to test these TS definitions I think but I've run out of time today so I can pick it up Monday

justin-mp:
> > Agreed, though I'd include cusor.results behind the compat flag if we do go down that route.
> 
> Hmm, really? I kind of liked it because it feels a lot more awkward to write `[...cursor]` vs. `cursor.results`, but I could be convinced either way. We could also make it throw if the cursor has been manually iterated before reading `.results`.

I'm worried about documenting the behavior if it's not behind the compat flag.  Throwing an exception if the cursor has been manually iterated before calling results makes that easy to document, so I think that's a reasonable solution.

src/workerd/api/sql.c++
```diff
@@ -60,6 +97,45 @@ bool SqlStorage::allowTransactions() const {
...
+}
+
+jsg::JsObject SqlStorage::wrapSqlRow(jsg::Lock& js, SqlRow row) {
+  return jsg::JsObject(js.withinHandleScope([&]() -> v8::Local<v8::Object> {
+    jsg::JsObject result = js.obj();
+    for (auto& field: row.fields) {
+      result.set(js, field.name, wrapSqlValue(js, kj::mv(field.value)));
+    }
+    return result;
+  }));
```
jasnell, line 126:
```suggestion
  return js.withinHandleScope([&]() -> jsg::JsObject {
    jsg::JsObject result = js.obj();
    for (auto& field: row.fields) {
      result.set(js, field.name, wrapSqlValue(js, kj::mv(field.value)));
    }
    return result;
  });
```

src/workerd/api/sql.c++
```diff
@@ -113,24 +189,49 @@ double SqlStorage::Cursor::getRowsWritten() {
...
 }
 
+jsg::JsArray SqlStorage::Cursor::getResults(jsg::Lock& js) {
+  KJ_IF_SOME(s, state) {
+    cachedColumnNames.ensureInitialized(js, s->query);
+  }
+
+  // Wrap in handle scope mostly for the benefit of `batch()`, whose implementation calls this
+  // many times.
+  return jsg::JsArray(js.withinHandleScope([&]() {
```
jasnell, line 199:
```suggestion
  return jsg::JsArray(js.withinHandleScope([&]() -> v8::Local<v8::Array> {
```

For consistency with the `wrapSqlRowRaw(...)` call above.

src/workerd/api/sql.c++
```diff
@@ -273,6 +379,116 @@ jsg::Ref<SqlStorage::Cursor> SqlStorage::Statement::run(jsg::Arguments<BindingVa
...
+jsg::Ref<SqlStorage::BoundStatement> SqlStorage::Statement::bind(
+    jsg::Arguments<BindingValue> bindings) {
+  return jsg::alloc<BoundStatement>(JSG_THIS, kj::mv(bindings));
+}
+
+jsg::JsArray SqlStorage::Statement::raw(
+    const v8::FunctionCallbackInfo<v8::Value>& args,
+    const jsg::TypeHandler<BindingValue>& bindingTypeHandler,
+    const jsg::TypeHandler<RawOptions>& optionsTypeHandler) {
+  jsg::Lock& js = jsg::Lock::from(args.GetIsolate());
```
jasnell, line 391: Nit: does this need to use `args.GetIsolate()` or could the `jsg::Lock&` be passed in?

src/workerd/api/sql.c++
```diff
@@ -273,6 +379,116 @@ jsg::Ref<SqlStorage::Cursor> SqlStorage::Statement::run(jsg::Arguments<BindingVa
...
+
+  auto result = KJ_UNWRAP_OR(Cursor::rowIteratorNext(js, cursor), {
+    // No results.
+    return js.null();
+  });
+
+  KJ_IF_SOME(c, column) {
+    for (auto& field: result.fields) {
+      if (field.name == c) {
+        return jsg::JsValue(wrapSqlValue(js, kj::mv(field.value)));
```
jasnell, line 482: `wrapSqlValue(...)` already returns a `jsg::JsValue` doesn't it?

src/workerd/api/sql.c++
```diff
@@ -286,4 +502,44 @@ void SqlStorage::visitForMemoryInfo(jsg::MemoryTracker& tracker) const {
...
+}
+
+jsg::Arguments<SqlStorage::BindingValue> SqlStorage::BoundStatement::cloneBindings() {
+  // Annoyingly the bound statement could theoretically be executed multiple times and so we need
+  // to clone the bindings each time in order to pass an owned copy to the query while also keeping
+  // a copy for future calls. In theory with some refactoring we could avoid this but this is only
+  // a compatibility shim anyway so we're not optimizing for it.
+
+  return KJ_MAP(binding, bindings) -> BindingValue {
+    KJ_IF_SOME(b, binding) {
```
jasnell, line 526: Why not binding.map(...) here since that would return `kj::none` ? Just curious.

src/workerd/api/sql.c++
kentonv: This doesn't work. `withinHandleScope` is not specialized for `jsg::JsObject` so it won't properly allow the handle to escape the scope.

src/workerd/api/sql.c++
kentonv: At present, the JSG stuff for wrapping methods does not support taking `jsg::Lock&` and `const v8::FunctionCallbackInfo<v8::Value>&` at the same time. It should probably be added but I didn't really want to get into JSG improvements in this PR...

src/workerd/api/sql.c++
jasnell: Ah right. Ok. Mind adding a comment?

kentonv:
> I think that's underselling it slightly. It's not just the benefit of being able to copy code wholesale from D1 to DO (although it's pretty amazing that we were able to do that), it's that, beyond sync vs async, there's really no good reason for the two APIs to diverge.

I'd argue the sync vs. async part actually leads to reasons for many differences:
* The batch API is only needed when the database is remote. There's no need for it in DO.
* The ability to bind statements separately from running them is only really needed for the batch API.
* `first()` isn't needed when you have cursors, you just read the first result from the cursor.
* `raw()` cannot return a cursor without breaking compatibility, but that means `run()` and `raw()` work very differently, which is weird.

I don't think the differences are entirely "stylistic" here.

> That said, the D1 api hasn't meaningfully changed since its very first iteration. It's long overdue for a refresh, and the corner cases & return-type inconsistencies would be the first things I'd want to address. But I'd suggest that, given our time constraints, our best option for developers in the near-term and medium term is for DO to fully embrace the (current) D1 api, then fix the API in both products at the same time (using compat dates and flags to remove things safely).

Well, I don't think it's that bad to launch without compatibility and then reconcile once we're ready to update D1, if it's not going to be that long...

kentonv:
After more discussion the consensus is that we don't want to ship this even under a compat flag. In fact, we're removing prepared statements from the API (#2722), since we intend to make `exec()` memoize them anyway. So the API is just going to be `exec("SQL CODE", param, param, ...)`, which is nice and simple. We can offer D1 API compatibility as a library if people want it.

PR #2697, opened 2024-09-12
Draft changes

shrima-cf:


src/workerd/util/sqlite.h
```diff
@@ -19,6 +20,20 @@ namespace workerd {
 using kj::byte;
 using kj::uint;
 
+class SqliteObserver {
+public:
+  virtual void addRowsRead(uint64_t rowsRead) {}
```
kentonv, line 25: I would combine `addRowsRead()` and `addRowsWritten()` into a single `queryExecuted()` call with two parameters. These are always both called together so it avoids a redundant virtual call, and may make it easier to track other metrics (like counting the number of total queries).

src/workerd/util/sqlite.h
```diff
@@ -19,6 +20,20 @@ namespace workerd {
...
+  virtual void addRowsWritten(uint64_t rowsWritten) {}
+  virtual void setSrsStoredBytesCallback(kj::Function<uint64_t()> callback) {}
```
kentonv, line 27: I think `setSrsStoredBytesCallback()` could be part of `ActorObserver` or maybe even `InternalActorObserver` (in the internal codebase), since it is not actually called by `SqliteDatabase` or anywhere in workerd.

src/workerd/util/sqlite.h
```diff
@@ -19,6 +20,20 @@ namespace workerd {
...
+
+  virtual ~SqliteObserver() noexcept(false) {}
```
kentonv, line 29: You don't need to declare this desturctor. In KJ code, `kj::Own` separately keeps track of how to dispose the object, so the destructor is never called directly on a virtual interface. Declaring the destructor virtual will actually result in a slight performance hit without any benefit for KJ code.

Note if you get a compiler warning about non-virtual destructors, the solution is to mark the derived class as `final`.

src/workerd/util/sqlite.h
```diff
@@ -19,6 +20,20 @@ namespace workerd {
...
+
+  static SqliteObserver& getDefaultObserver() {
+    static SqliteObserver defaultObserver;
+    return defaultObserver;
+  }
```
kentonv, line 34: We normally just declare a constant for these default instances, like:

`static const SqliteObserver DEFAULT;`

src/workerd/server/server.c++
```diff
@@ -1982,11 +1982,11 @@ public:
...
 
-            actorContainer->actor.emplace(
-                kj::refcounted<Worker::Actor>(*service.worker, actorContainer->getTracker(),
-                    kj::str(idPtr), true, kj::mv(makeActorCache), className, kj::mv(makeStorage),
-                    lock, kj::mv(loopback), timerChannel, kj::refcounted<ActorObserver>(),
-                    actorContainer->tryGetManagerRef(), hibernationEventTypeId));
+            actorContainer->actor.emplace(kj::refcounted<Worker::Actor>(*service.worker,
+                actorContainer->getTracker(), kj::str(idPtr), true, kj::mv(makeActorCache),
+                className, kj::mv(makeStorage), lock, kj::mv(loopback), timerChannel,
+                kj::refcounted<ActorObserver>(), kj::Own<SqliteObserver>(),
```
kentonv, line 1988: `kj::Own<SqliteObserver>()` here constructs a null pointer, which will presumably lead to a segfault when accessed later.

Given that the implementation ultimately has a single class implementing both `ActorObserver` and `SqliteObserver`, perhaps `ActorObserver` itself should be declared to be a subclass of `SqliteObserver` so that it doesn't need to be passed separately here?

src/workerd/util/sqlite.h
```diff
@@ -42,7 +57,13 @@ class SqliteDatabase {
     uint64_t statementCount;
   };
 
-  SqliteDatabase(const Vfs& vfs, kj::PathPtr path, kj::Maybe<kj::WriteMode> maybeMode = kj::none);
+  SqliteDatabase(const Vfs& vfs, kj::PathPtr path, kj::Maybe<kj::WriteMode> maybeMode = kj::none)
+      : SqliteDatabase(vfs, path, SqliteObserver::getDefaultObserver(), maybeMode) {}
+
+  SqliteDatabase(const Vfs& vfs,
+      kj::PathPtr path,
+      SqliteObserver& sqliteObserver,
```
kentonv, line 65: Style nitpick: I would suggest adding `sqliteObserver` as the fourth argument of the existing constructor, and declare it with a default value, like:

`SqliteObesrver& sqliteObserver = SqliteObserver::DEFAULT`

src/workerd/server/server.c++
shrima-cf: That sounds fair

Issue #2694, opened 2024-09-11
zlib.createBrotliCompress emits unexpected error


npaun:
The following function call emits two errors:

```js
zlib.createBrotliCompress({
      params: {
        // This is a boolean flag
        [zlib.constants.BROTLI_PARAM_DISABLE_LITERAL_CONTEXT_MODELING]:
          42,
      },
    });
```

The first is `info: uncaught exception; desc = Error: Setting parameter failed`
The second is `Error [ERR_ZLIB_INITIALIZATION_FAILED]` (which matches NodeJS)

PR #2701, opened 2024-09-12
fix the execution order of brotli class

anonrig:
Found the bug. It seems the execution order was wrong. I couldn't find a proper way to test this, but manually running the tests show no `uncaught exception`.

Fixes https://github.com/cloudflare/workerd/issues/2694


anonrig:
It seems this error is only visible on the logs, since error is thrown on the queueMicrotask call.

> workerd/io/worker.c++:1054: info: uncaught exception; desc = Error: Setting parameter failed

npaun:
@anonrig Yep you found what I was talking about

anonrig:
ok, i think i found it. we should have called the parent class constructor after we initialize the handle.

npaun (approved):

PR #2702, opened 2024-09-12
remove the need for `nodejs_zlib` flag

anonrig:
Let's merge this once the existing bugs are fixed.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -584,7 +584,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   nodeJsZlib @59 :Bool
       $compatEnableFlag("nodejs_zlib")
       $compatDisableFlag("no_nodejs_zlib")
-      $experimental;
+      $impliedByAfterDate(name = "nodeJsCompat", date = "2024-09-23");
   # Enables node:zlib implementation while it is in-development.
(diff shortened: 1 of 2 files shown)


jasnell (approved):
This isn't typically how we would do this. The compat flag should remain and these checks should remain.

Instead, for the nodejs_zlib compat flag definition, remove the `$experimental` annotation and add an `$impliedByAfterDate(name = "nodeJsCompat", date = "2024-09-23");`

So that it becomes on by default when `nodeJsCompat` is turned on.

anonrig:
Thanks @jasnell. I've updated the pull-request.

src/workerd/io/compatibility-date.capnp
```diff
@@ -584,7 +584,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   nodeJsZlib @59 :Bool
       $compatEnableFlag("nodejs_zlib")
       $compatDisableFlag("no_nodejs_zlib")
-      $experimental;
```
jasnell, line 587: There is a compatibility-date-test.c++ that can/should be updated to verify this. Look at line 182 in the current test. That verifies that the `nodejs_compat_v2` flag is appropriately set on the compat date, the `nodejs_zlib` flag can be checked there also.

npaun (approved):

jasnell (approved):

PR #2705, opened 2024-09-12
Extract Zlib/Brotli custom allocator

### LINKED ISSUE #2678: Track brotli allocations
We need to complete this TODO before marking them as stable.

https://github.com/cloudflare/workerd/blob/main/src/workerd/api/node/zlib-util.c%2B%2B#L909

npaun:
* Zlib and Brotli use malloc and free by default to manage memory. If they leak, it's an upstream bug.
* However, they also allow for custom allocators to be passed.
* This PR extracts our custom allocator functions AllocForZlib, AllocForBrotli, FreeForZlib into a new class called Allocator.
* Allocator is now used for both the streaming and one-shot implementations.
* I dropped the stuffing of size into the allocated regions because the backing kj::heapArrays are already sized objects.
* At present, the information collected by allocator is not used. However, we could block excessively large allocations, and report allocations to V8 in the future.
* Our current implementation is a hash map of pointers, but what we'd really like is a kind of arena allocator, in my opinion at least.

Closes #2678

### DIFF
--- a/src/workerd/api/node/zlib-util.c++
+++ b/src/workerd/api/node/zlib-util.c++
@@ -602,7 +602,7 @@ void ZlibUtil::ZlibStream::initialize(int windowBits,
     jsg::Function<void()> writeCallback,
     jsg::Optional<kj::Array<kj::byte>> dictionary) {
   initializeStream(kj::mv(writeState), kj::mv(writeCallback));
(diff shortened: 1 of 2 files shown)


jasnell (approved):
Let's have @anonrig review also

anonrig (approved):

PR #2706, opened 2024-09-12
Add initial lime tracing

fhanau:
See downstream PR for discussion.

### DIFF
--- a/src/workerd/api/actor.c++
+++ b/src/workerd/api/actor.c++
@@ -27,18 +27,18 @@ public:
     auto& context = IoContext::current();
 
     return context.getMetrics().wrapActorSubrequestClient(context.getSubrequest(
-        [&](SpanBuilder& span, IoChannelFactory& ioChannelFactory) {
-      if (span.isObserved()) {
-        span.setTag("actor_id"_kjc, kj::str(actorId));
+        [&](TraceContext& tracing, IoChannelFactory& ioChannelFactory) {
+      if (tracing.span.isObserved()) {
+        tracing.span.setTag("actor_id"_kjc, kj::str(actorId));
       }
 
       // Lazily initialize actorChannel
       if (actorChannel == kj::none) {
-        actorChannel = context.getColoLocalActorChannel(channelId, actorId, span);
+        actorChannel = context.getColoLocalActorChannel(channelId, actorId, tracing.span);
       }
 
       return KJ_REQUIRE_NONNULL(actorChannel)
-          ->startRequest({.cfBlobJson = kj::mv(cfStr), .parentSpan = span});
+          ->startRequest({.cfBlobJson = kj::mv(cfStr), .tracing = tracing});
     },
         {.inHouse = true,
           .wrapMetrics = true,
@@ -68,19 +68,19 @@ public:
     auto& context = IoContext::current();
 
     return context.getMetrics().wrapActorSubrequestClient(context.getSubrequest(
-        [&](SpanBuilder& span, IoChannelFactory& ioChannelFactory) {
-      if (span.isObserved()) {
-        span.setTag("actor_id"_kjc, id->toString());
+        [&](TraceContext& tracing, IoChannelFactory& ioChannelFactory) {
+      if (tracing.span.isObserved()) {
+        tracing.span.setTag("actor_id"_kjc, id->toString());
       }
 
       // Lazily initialize actorChannel
       if (actorChannel == kj::none) {
-        actorChannel = context.getGlobalActorChannel(
-            channelId, id->getInner(), kj::mv(locationHint), mode, enableReplicaRouting, span);
+        actorChannel = context.getGlobalActorChannel(channelId, id->getInner(),
+            kj::mv(locationHint), mode, enableReplicaRouting, tracing.span);
       }
 
       return KJ_REQUIRE_NONNULL(actorChannel)
-          ->startRequest({.cfBlobJson = kj::mv(cfStr), .parentSpan = span});
+          ->startRequest({.cfBlobJson = kj::mv(cfStr), .tracing = tracing});
     },
         {.inHouse = true,
           .wrapMetrics = true,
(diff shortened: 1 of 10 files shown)


jasnell:
We're soon going to be implementing streaming traces *alongside* the current in-memory traces. We'd need to figure out how these spans fit into that model.

src/workerd/io/trace.c++
```diff
@@ -596,9 +627,57 @@ void WorkerTracer::log(kj::Date timestamp, LogLevel logLevel, kj::String message
...
+        }
+        KJ_CASE_ONEOF(val, double) {
+          return kj::str(val);
+        }
+        KJ_CASE_ONEOF(val, bool) {
+          return kj::str(val);
+        }
+      }
+      KJ_UNREACHABLE;
+    }();
```
jasnell, line 675: Question: should this count towards the max byte limit for the trace?

fhanau: It already does – log() is called to actually write the data and keeps track of `trace->bytesUsed `.

mikea:
> We're soon going to be implementing streaming traces _alongside_ the current in-memory traces. We'd need to figure out how these spans fit into that model.

I suggest we introduce a single `TraceContext` that will carry around both of current span builders and all the future tracing infrastucture around. Its interface might be disjoint (separate create child span calls for each branch), but we won't have to change function signatures again.

fhanau:
> > We're soon going to be implementing streaming traces _alongside_ the current in-memory traces. We'd need to figure out how these spans fit into that model.
> 
> I suggest we introduce a single `TraceContext` that will carry around both of current span builders and all the future tracing infrastucture around. Its interface might be disjoint (separate create child span calls for each branch), but we won't have to change function signatures again.

Created two structs to track both span builders/span parents. I deliberately did not make this a fully-fledged class since places where the spans will be used are expected to remain somewhat disparate, but it does make passing through parameters in `getSubrequest()` and `SubrequestMetadata` easier. We can refactor/extend it later when this seems helpful.

src/workerd/io/trace.h
```diff
@@ -373,10 +378,14 @@ class WorkerTracer final: public kj::Refcounted {
...
       PipelineLogLevel pipelineLogLevel);
   explicit WorkerTracer(PipelineLogLevel pipelineLogLevel);
+  ~WorkerTracer() {
+    weakRef->invalidate();
+  }
   KJ_DISALLOW_COPY_AND_MOVE(WorkerTracer);
 
   // Adds log line to trace.  For Spectre, timestamp should only be as accurate as JS Date.now().
-  void log(kj::Date timestamp, LogLevel logLevel, kj::String message);
+  void log(kj::Date timestamp, LogLevel logLevel, kj::String message, bool isSpan = false);
```
mikea, line 387: need to documetn isSpan parameter

src/workerd/io/trace.h
```diff
@@ -373,10 +378,14 @@ class WorkerTracer final: public kj::Refcounted {
...
+  void addSpan(const Span& span, kj::String spanContext);
```
mikea, line 388: need documentation

src/workerd/io/trace.h
```diff
@@ -408,13 +417,18 @@ class WorkerTracer final: public kj::Refcounted {
...
+  }
+
 private:
   PipelineLogLevel pipelineLogLevel;
   kj::Own<Trace> trace;
 
   // own an instance of the pipeline to make sure it doesn't get destroyed
   // before we're finished tracing
   kj::Maybe<kj::Own<PipelineTracer>> parentPipeline;
+  kj::Own<WeakRef<WorkerTracer>> weakRef;
```
mikea, line 431: maybe `self` is a better name?

src/workerd/api/actor.c++
```diff
@@ -24,18 +24,18 @@ public:
     auto& context = IoContext::current();
 
     return context.getMetrics().wrapActorSubrequestClient(context.getSubrequest(
-        [&](SpanBuilder& span, IoChannelFactory& ioChannelFactory) {
-      if (span.isObserved()) {
-        span.setTag("actor_id"_kjc, kj::str(actorId));
+        [&](TraceContext& spans, IoChannelFactory& ioChannelFactory) {
```
mikea, line 27: sorry for bikeshedding, but I think we can come up with better class/parameter name pair than this.

If we're going to keep `TraceContext`, then maybe `TraceContext& tracing`? (my vote)
If you like `spans`, then maybe `TraceSpans`?

mikea (commented):
Let's get naming sorted out while I review the rest of the code.

I also worry that `Lime` as in `limeSpans` wouldn't age well. How about `user tracing` framework: `UserSpanBuilder` and `userSpans`?

src/workerd/io/trace.h
```diff
@@ -413,13 +417,18 @@ class WorkerTracer final: public kj::Refcounted {
...
```
mikea, line 431: also let's document this since you're saying yourself it is tricky.

src/workerd/io/trace.h
```diff
@@ -641,4 +683,89 @@ inline SpanBuilder SpanBuilder::newChild(kj::ConstString operationName, kj::Date
       kj::mv(operationName), startTime);
 }
 
+// Lime tracing
```
mikea, line 686: I'm not super convinced we need `LimeSpanBuilder` at all. It looks to me that `SpanBuilder` is totally sufficient for both branches of tracking. Having single builder class will simplify maintenance a lot.

src/workerd/io/trace.h
fhanau: Added a comment describing my current understanding of the issue.

fhanau:
Updated this based on the review comments, documentation should look quite a bit better now. Changing the name to user tracing will be done in a follow-up PR. workerd side is ready for review.

mikea (approved):
lg. Lime rename can happen in a follow-up

PR #2707, opened 2024-09-12
Introduce SqliteObserver and use it to collect stats related to queries and db size

shrima-cf:
https://jira.cfdata.org/browse/STOR-3710

### DIFF
--- a/src/workerd/util/sqlite-test.c++
+++ b/src/workerd/util/sqlite-test.c++
@@ -782,5 +782,75 @@ KJ_TEST("reset database") {
   }
 }
 
+KJ_TEST("SQLite observer addQueryStats") {
+  class TestSqliteObserver: public SqliteObserver {
+  public:
+    void addQueryStats(uint64_t read, uint64_t written) override {
+      rowsRead += read;
+      rowsWritten += written;
+    }
+
+    uint64_t rowsRead = 0;
+    uint64_t rowsWritten = 0;
+  };
+
+  TempDirOnDisk dir;
+  SqliteDatabase::Vfs vfs(*dir);
+  TestSqliteObserver sqliteObserver = TestSqliteObserver();
+  SqliteDatabase db(
+      vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY, sqliteObserver);
+
+  db.run(R"(
+    CREATE TABLE things (
+      id INTEGER PRIMARY KEY
+    );
+  )");
+
+  // There are some rows read and written when we create the db, we offset this in the test
+  int rowsReadBefore = sqliteObserver.rowsRead;
+  int rowsWrittenBefore = sqliteObserver.rowsWritten;
+  constexpr int dbRowCount = 3;
+  {
+    db.run("INSERT INTO things (id) VALUES (10)");
+    db.run("INSERT INTO things (id) VALUES (11)");
+    db.run("INSERT INTO things (id) VALUES (12)");
+  }
+  KJ_EXPECT(sqliteObserver.rowsRead - rowsReadBefore == dbRowCount);
+  KJ_EXPECT(sqliteObserver.rowsWritten - rowsWrittenBefore == dbRowCount);
+
+  rowsReadBefore = sqliteObserver.rowsRead;
+  rowsWrittenBefore = sqliteObserver.rowsWritten;
+  {
+    auto getCount = db.prepare("SELECT COUNT(*) FROM things");
+    KJ_EXPECT(getCount.run().getInt(0) == dbRowCount);
+  }
+  KJ_EXPECT(sqliteObserver.rowsRead - rowsReadBefore == dbRowCount);
+  KJ_EXPECT(sqliteObserver.rowsWritten - rowsWrittenBefore == 0);
+
+  // Verify if addQueryStats works correctly when we call query.nextRow()
+  rowsReadBefore = sqliteObserver.rowsRead;
+  rowsWrittenBefore = sqliteObserver.rowsWritten;
+  {
+    auto stmt = db.prepare("SELECT * FROM things");
+    auto query = stmt.run();
+    KJ_ASSERT(!query.isDone());
+    while (!query.isDone()) {
+      query.nextRow();
+    }
+  }
+  KJ_EXPECT(sqliteObserver.rowsRead - rowsReadBefore == dbRowCount);
+  KJ_EXPECT(sqliteObserver.rowsWritten - rowsWrittenBefore == 0);
+
+  // Verify addQueryStats works correctly when db is reset
+  rowsReadBefore = sqliteObserver.rowsRead;
+  rowsWrittenBefore = sqliteObserver.rowsWritten;
+  {
+    auto query = db.run("INSERT INTO things (id) VALUES (100)");
+    db.reset();
+  }
+  KJ_EXPECT(sqliteObserver.rowsRead - rowsReadBefore == 1);
+  KJ_EXPECT(sqliteObserver.rowsWritten - rowsWrittenBefore == 1);
+}
+
 }  // namespace
 }  // namespace workerd
--- a/src/workerd/util/sqlite.c++
+++ b/src/workerd/util/sqlite.c++
@@ -367,12 +367,18 @@ static constexpr PragmaInfo ALLOWED_PRAGMAS[] = {{"data_version"_kj, PragmaSigna
 
 // =======================================================================================
 
+SqliteObserver SqliteObserver::DEFAULT = SqliteObserver{};
+
 constexpr SqliteDatabase::Regulator SqliteDatabase::TRUSTED;
 
-SqliteDatabase::SqliteDatabase(const Vfs& vfs, kj::Path path, kj::Maybe<kj::WriteMode> maybeMode)
+SqliteDatabase::SqliteDatabase(const Vfs& vfs,
+    kj::Path path,
+    kj::Maybe<kj::WriteMode> maybeMode,
+    SqliteObserver& sqliteObserver)
     : vfs(vfs),
       path(kj::mv(path)),
-      readOnly(maybeMode == kj::none) {
+      readOnly(maybeMode == kj::none),
+      sqliteObserver(sqliteObserver) {
   init(maybeMode);
 }
 
@@ -979,6 +985,9 @@ SqliteDatabase::Query::Query(SqliteDatabase& db,
 }
 
 SqliteDatabase::Query::~Query() noexcept(false) {
+  //Update the db stats that we have collected for the query
+  db.sqliteObserver.addQueryStats(rowsRead, rowsWritten);
+
   // We only need to reset the statement if we don't own it. If we own it, it's about to be
   // destroyed anyway.
   if (ownStatement.get() == nullptr) {
@@ -1096,6 +1105,11 @@ void SqliteDatabase::Query::nextRow() {
   db.currentRegulator = regulator;
 
   int err = sqlite3_step(statement);
+  // TODO(perf): This is slightly inefficient to call for every row read, but not bad enough to
+  // fix it immediately. The alternate way would be to getRowsRead/Written once when we emit it
+  // in the Dtor, and handle the case where the statement could be null when the Query gets destructed
+  rowsRead = getRowsRead();
+  rowsWritten = getRowsWritten();
   if (err == SQLITE_DONE) {
     done = true;
   } else if (err != SQLITE_ROW) {
--- a/src/workerd/util/sqlite.h
+++ b/src/workerd/util/sqlite.h
@@ -20,6 +20,16 @@ namespace workerd {
 using kj::byte;
 using kj::uint;
 
+// Used to collect periodic metrics about queries and size of sqlite db
+class SqliteObserver {
+public:
+  virtual void addQueryStats(uint64_t rowsRead, uint64_t rowsWritten) {}
+  // The method is not used by the SqliteDatabase, it is added here for convenience
+  virtual void setSqliteStoredBytes(uint64_t sqliteStoredBytes) {}
+
+  static SqliteObserver DEFAULT;
+};
+
 // C++/KJ API for SQLite.
 //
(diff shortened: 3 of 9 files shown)


src/workerd/util/sqlite.h
```diff
@@ -205,10 +217,15 @@ class SqliteDatabase {
     afterResetCallback = kj::mv(callback);
   }
 
+  SqliteObserver& getSqliteObserver() {
```
shrima-cf, line 220: I added this here so I can use this in srs client code to get the observer from the db object

src/workerd/util/sqlite.h
```diff
@@ -20,6 +20,15 @@ namespace workerd {
 using kj::byte;
 using kj::uint;
 
+// Used to collect periodic metrics about queries and size of sqlite db
+class SqliteObserver {
+public:
+  virtual void updateQueryStats(uint64_t rowsRead, uint64_t rowsWritten) {}
+  virtual void setSrsStoredBytes(uint64_t srsStoredBytes) {}
+
+  static SqliteObserver DEFAULT;
```
justin-mp, line 29: Could this be `constexpr` like `SqliteDatabase::TRUSTED`?

src/workerd/util/sqlite.c++
```diff
@@ -992,6 +998,10 @@ SqliteDatabase::Query::~Query() noexcept(false) {
       sqlite3_clear_bindings(&statement);
     }
   }
+  // Update Query stats only if db is not reset
+  if (maybeStatement != kj::none) {
+    db.sqliteObserver.updateQueryStats(getRowsRead(), getRowsWritten());
```
justin-mp, line 1003: I would have thought that the `sqlite3_reset()` call on line 994 would affect the statistics here.  I guess it doesn't?

src/workerd/util/sqlite.c++
```diff
@@ -992,6 +998,10 @@ SqliteDatabase::Query::~Query() noexcept(false) {
       sqlite3_clear_bindings(&statement);
     }
   }
+  // Update Query stats only if db is not reset
```
justin-mp, line 1001: This comment implies that if you reset the database, the rows you've already read or written will not be accounted for.  I think you may need to add something to `SqliteDatabase::Query::beforeSqliteReset()` to record this information.

src/workerd/util/sqlite-test.c++
```diff
@@ -782,5 +782,51 @@ KJ_TEST("reset database") {
...
+  // There are some rows read and written when we create the db, we offset this in the test
+  int rowsReadBefore = sqliteObserver.rowsRead;
+  int rowsWrittenBefore = sqliteObserver.rowsWritten;
+  constexpr int dbRowCount = 3;
+  {
+    db.run("INSERT INTO things (id) VALUES (10)");
+    db.run("INSERT INTO things (id) VALUES (11)");
+    db.run("INSERT INTO things (id) VALUES (12)");
+  }
+  KJ_EXPECT(sqliteObserver.rowsRead - rowsReadBefore == dbRowCount);
```
justin-mp, line 818: Wait, an insert counts as a row read and a row written?  That's unexpected.

src/workerd/util/sqlite.h
```diff
@@ -20,6 +20,15 @@ namespace workerd {
 using kj::byte;
 using kj::uint;
 
+// Used to collect periodic metrics about queries and size of sqlite db
+class SqliteObserver {
+public:
+  virtual void updateQueryStats(uint64_t rowsRead, uint64_t rowsWritten) {}
+  virtual void setSrsStoredBytes(uint64_t srsStoredBytes) {}
```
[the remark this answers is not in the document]
shrima-cf: Good catch! Will fix it

src/workerd/util/sqlite.h
shrima-cf: This can't be a constexpr because of the virtual methods in it. 
It can't be a const either because the methods will eventually update the state in the implementation of the derived class

src/workerd/util/sqlite.c++
shrima-cf: Hmm, we do need it in there. I just have to make sure we don't run into race condition where the rows gets updated from the beforeSqliteReset and from the destructor

src/workerd/util/sqlite-test.c++
shrima-cf: id is the primary key, so an insert here does a read to make sure the primary key is not already present and then inserts, so 1 read for the lookup and 1 read for the write

src/workerd/util/sqlite.c++
```diff
@@ -1177,6 +1188,9 @@ sqlite3_stmt* SqliteDatabase::Query::getStatement() {
 }
 
 void SqliteDatabase::Query::beforeSqliteReset() {
+  // update the db stats before resetting the database
+  db.sqliteObserver.updateQueryStats(getRowsRead(), getRowsWritten());
```
shrima-cf, line 1192: The asan error is happening because the statement could be dangling by the time we reach here. I am trying to see how I can fix this, but so far it does not look like there is a good way for us to figure this out.
We could store rowsRead and rowsWritten by the query and update it everytime we do any operation on the query and use it here, I am not 100% convinced I want to go that route yet.

src/workerd/util/sqlite.h
```diff
@@ -205,10 +217,15 @@ class SqliteDatabase {
...
 
+  SqliteObserver& getSqliteObserver() {
+    return sqliteObserver;
+  }
+
 private:
   const Vfs& vfs;
   kj::Path path;
   bool readOnly;
+  SqliteObserver& sqliteObserver;
```
jasnell, line 228: A comment here identifying who owns the `SqliteObserver` would be helpful given that it'll be necessary to make sure whatever it is outlives this.

src/workerd/util/sqlite.c++
justin-mp: No, I meant that the calls to getRowsRead()/getRowsWritten() would have been affected by the reset, since those call into `sqlite3_stmt_status(statement, LIBSQL_STMTSTATUS_ROWS_WRITTEN, 0);`.  It seems `sqlite3_reset()` doesn't touch those values.

src/workerd/util/sqlite.h
shrima-cf: Will do!

src/workerd/util/sqlite.c++
```diff
@@ -992,6 +998,8 @@ SqliteDatabase::Query::~Query() noexcept(false) {
       sqlite3_clear_bindings(&statement);
     }
   }
+  //Update the db stats that we have collected for the query
```
justin-mp, line 1001: Personally, I would do this before all of the reset stuff.  That way you know the `sqlite3_reset` doesn't affect your counts even without checking the implementation.

src/workerd/util/sqlite.c++
```diff
@@ -1096,6 +1104,8 @@ void SqliteDatabase::Query::nextRow() {
   db.currentRegulator = regulator;
 
   int err = sqlite3_step(statement);
+  rowsRead += getRowsRead();
```
justin-mp, line 1107: `getRowsRead()` is culmulative, so I think you want to use `=`, not `+=`.  Same for the write case below.

You should be able to detect this in sqlite-test.c++ by having a single statement that returns multiple rows.

src/workerd/util/sqlite.h
```diff
@@ -43,7 +52,12 @@ class SqliteDatabase {
     uint64_t statementCount;
   };
 
-  SqliteDatabase(const Vfs& vfs, kj::Path path, kj::Maybe<kj::WriteMode> maybeMode = kj::none);
+  // Whoever creates the database, is required to make sure the SqliteObserver is alive atleast
+  // as long as the DB is alive.
```
justin-mp, line 56:
```suggestion
  // Whoever creates the database is required to make sure the SqliteObserver is alive at least
  // as long as the SqliteDatabase is alive.
```

src/workerd/util/sqlite.h
```diff
@@ -390,6 +409,12 @@ class SqliteDatabase::Query final: private ResetListener {
   kj::Maybe<sqlite3_stmt&> maybeStatement;  // null if database was reset
   bool done = false;
 
+  // Storing the rowsRead and rowsWritten here to use in cases where a DB is reset
+  // If the DB is reset, getRowdRead and getRowsWritten will fail as the statement it
+  // refers to gets destroyed in the even of DB reset
```
justin-mp, line 414:
```suggestion
  // Storing the rowsRead and rowsWritten here to use in cases where a DB is reset.
  // When the DB is reset, getRowdRead and getRowsWritten will fail as the statement they
  // refer to gets destroyed as part of the reset process.
```

src/workerd/util/sqlite.c++
shrima-cf: You are right, will fix and add a test

src/workerd/util/sqlite.c++
shrima-cf: I chose to do it here so we update the stats only once. The way the update is done here uses the internal variables of the query, so reset doesn't have an effect to any of the entities involved here

shrima-cf: I realize the confusion is arising from the updateQueryStats, the implementation is open to interpretation. I will rename it to addQueryStats, so it is clear that the stats will have to accumulated.

justin-mp: I'm confused, I meant just move this block of code to the start of the dtor!

shrima-cf: Ah, I thought you wanted me to move this out of dtor and into the nextRow itself. 
Will do that! I still think the name was confusing, so I changed it

justin-mp (approved):

shrima-cf:
@justin-mp The new changes are the ones from [here](https://github.com/cloudflare/workerd/pull/2711/files)

justin-mp (approved):
The `makeActorCache` changes seem reasonable to me.

src/workerd/util/sqlite.h
```diff
@@ -20,6 +20,15 @@ namespace workerd {
 using kj::byte;
 using kj::uint;
 
+// Used to collect periodic metrics about queries and size of sqlite db
+class SqliteObserver {
+public:
+  virtual void addQueryStats(uint64_t rowsRead, uint64_t rowsWritten) {}
+  virtual void setSqliteStoredBytes(uint64_t sqliteStoredBytes) {}
```
kentonv, line 27: It's a little weird that this method is declared here but not actually called by `SqliteDatabase`. In theory a cleaner approach would be to have a separate interface like `SrsDataSizeObserver` declared in the internal codebase.

But I'm fine with leaving it like this as long as we have a comment explaining that this method is here for convenience but not actually called by `SqliteDatabase`.

src/workerd/util/sqlite.h
```diff
@@ -43,7 +52,12 @@ class SqliteDatabase {
     uint64_t statementCount;
   };
 
-  SqliteDatabase(const Vfs& vfs, kj::Path path, kj::Maybe<kj::WriteMode> maybeMode = kj::none);
+  // Whoever creates the database, is required to make sure the SqliteObserver is alive at least
```
kentonv, line 55: This comment is stating something that [KJ style already requires by default](https://github.com/capnproto/capnproto/blob/v2/style-guide.md#ownership):

> A pointer or reference passed as a constructor parameter must remain valid for the lifetime of the constructed object.

So I think the comment can be omitted.

In any case, the same requirement applies to the existing `vfs` argument, so if the comment is going to be explicit about lifetimes it should mention that too.

src/workerd/util/sqlite.c++
```diff
@@ -1096,6 +1105,8 @@ void SqliteDatabase::Query::nextRow() {
   db.currentRegulator = regulator;
 
   int err = sqlite3_step(statement);
+  rowsRead = getRowsRead();
```
kentonv, line 1108: This seems inefficient as we're querying SQLite for rows read/written counts on every single row just in case the database is later resent before the query is destroyed -- which is very rare.

Instead, how about putting code into `Query::beforeSqliteReset()` that posts the metrics early (calls `addQueryStats` directly), and then have the destructor only post metrics if it hasn't been reset (i.e. `maybeStatement` is non-null).

src/workerd/util/sqlite.h
kentonv: I actually disagree this needs a comment. This reference was passed into the constructor, so it's implied that it is the caller's responsibility to make sure it stays alive. The code here doesn't actually know anything about the ownership of this object beyond that.

src/workerd/util/sqlite.h
justin-mp: I don't think you actually need this for the internal PR.

src/workerd/util/sqlite.h
shrima-cf: I was wondering the same thing, I added the comment in response to James' question regarding the same.
If it is part of the style guide, then it makes sense to remove the comment

src/workerd/util/sqlite.c++
shrima-cf: I did try that approach first. The problem with that comes in the way we do getRowsRead() and the way the Statement and Query are setup. There are a couple of corner cases that have to be handled - when the Query has its own statement vs passed in a statement, handling the beforeSqliteReset in both Query and Statement.
In the end, I decided to go this route as it is easy to reason about.

As far as inefficiency in querying Sqlite for rows read/written, looking at the implementation of getRowsRead/Written, it does a `sqlite3_stmt_status(...)` which does a pointer read, so its not as inefficient as we think it is

src/workerd/util/sqlite.h
shrima-cf: Added a comment

src/workerd/util/sqlite.c++
shrima-cf: I added a comment with more details

kentonv (approved):

PR #2709, opened 2024-09-13
use `ExternalMemoryAdjuster` on zlib allocations

anonrig:
Let's also inform v8 about the allocations that we've made.

### DIFF
--- a/src/workerd/api/node/BUILD.bazel
+++ b/src/workerd/api/node/BUILD.bazel
@@ -14,7 +14,6 @@ wd_cc_library(
         "@capnp-cpp//src/kj/compat:kj-gzip",
         "@nbytes",
         "@simdutf",
-        "@zlib",
     ],
     visibility = ["//visibility:public"],
     deps = [
--- a/src/workerd/api/node/zlib-util.h
+++ b/src/workerd/api/node/zlib-util.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <workerd/jsg/jsg.h>
+#include <workerd/api/streams/compression.h>
 
 #include <zlib.h>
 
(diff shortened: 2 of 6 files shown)


src/workerd/api/node/zlib-util.h
```diff
@@ -291,17 +291,23 @@ class ZlibUtil final: public jsg::Object {
...
+  // isolate pointer and use that to get the external memory adjustment.
   class Allocator final {
   public:
     static void* AllocForZlib(void* data, uInt items, uInt size);
     static void* AllocForBrotli(void* data, size_t size);
     static void FreeForZlib(void* data, void* pointer);
 
+    struct Allocation {
+      kj::Array<kj::byte> data;
+      kj::Maybe<jsg::ExternalMemoryAdjustment> memory;
```
npaun, line 306:
```suggestion
      kj::Maybe<jsg::ExternalMemoryAdjustment> memoryAdjustment;
```

nit: Makes it clearer which field is the allocation and what is the metadata.

npaun (approved):
LGTM but definitely needs a review from @jasnell too

jasnell (approved):

jasnell:
Let's make sure to run an internal CI run on this to confirm. Should be just fine but let's make sure.

anonrig:
I've reimplemented the allocator and moved it to api/streams, since the compression there also needs to be tracked using v8. appreciate a re-review @jasnell @npaun

src/workerd/api/node/BUILD.bazel
```diff
@@ -14,7 +14,6 @@ wd_cc_library(
         "@capnp-cpp//src/kj/compat:kj-gzip",
         "@nbytes",
         "@simdutf",
-        "@zlib",
```
npaun, line 17: What's the reason for removing this? Does kj-gzip cover it?

src/workerd/api/streams/compression.c++
```diff
@@ -3,14 +3,51 @@
...
+#include "nbytes.h"
 #include <workerd/io/features.h>
-#include <zlib.h>
 #include <deque>
 #include <vector>
 #include <iterator>
 
 namespace workerd::api {
 
+void CompressionAllocator::configure(z_stream* stream) {
```
npaun, line 15: Maybe call it configureForZlib, as that's the only use?

src/workerd/api/node/BUILD.bazel
anonrig: Yes, this also adds lots of warnings to edgeworker.

npaun (approved):
Re-approved

anonrig:
Had to revert the v8 tracking changes due to an internal bug. I'll investigate differently. Meanwhile, the PR includes a little bit of refactoring to simplify the function signatures.

PR #2718, opened 2024-09-16
Add Sqlite test for read/write row counters for large rows

shrima-cf:


src/workerd/util/sqlite-test.c++
```diff
@@ -663,6 +663,36 @@ KJ_TEST("SQLite write row counters (basic)") {
...
 }
 
+KJ_TEST("SQLite read/write row counters (large row insert)") {
+  auto dir = kj::newInMemoryDirectory(kj::nullClock());
+  SqliteDatabase::Vfs vfs(*dir);
+  SqliteDatabase db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+  db.run("CREATE TABLE large_things (id INTEGER PRIMARY KEY, large_value TEXT)");
+
+  // SQLite's default page size is 4096 bytes
```
anonrig, line 673: Can you add an assertion for the default page size? It seems  `pragma_page_size() ` is available.

src/workerd/util/sqlite-test.c++
```diff
@@ -663,6 +663,36 @@ KJ_TEST("SQLite write row counters (basic)") {
   }
 }
 
+KJ_TEST("SQLite read/write row counters (large row insert)") {
```
justin-mp, line 666: It would be good to comment why we have this test (either in the commit message or with a comment here).

I believe it's to make sure that extra page reads aren't included in the row count.

Does it also make sense to have more than one row for this test?  I don't know how big rows are actually stored in SQLite.

src/workerd/util/sqlite-test.c++
shrima-cf: Added!

src/workerd/util/sqlite-test.c++
shrima-cf: Big rows are stored  in what is called overflow pages - https://www.sqlite.org/fileformat2.html#ovflpgs

I don't think having multiple rows in the test is going to give us any new information.

Will add the comment

justin-mp (approved):

PR #2719, opened 2024-09-17
Patch v8 to support promise cross-context resolution

jasnell:
Draft for now until I add tests tests and more tests

```
export default {
  async fetch(req, env, ctx) {
    if (globalThis.resolve === undefined) {

      {
        setTimeout(resolve, 1000);
        ctx.waitUntil(promise);
      }

      // This is our first request. We will create a new promise and resolver
      // pair and store it in a global. This request will then wait for the
      // promise to be resolved before continuing.
      globalThis.resolve = resolve;
      globalThis.promise = promise;
      const ab = AbortSignal.abort();
      console.log(ab.aborted);
      promise.then(() => {
        // This will be run within the correct IoContext now...
        try {
          console.log('test1', ab.aborted);
        } catch (err) {
          // We would get here if the IoContext was incorrect because
          // of the call to ab.aborted
          console.log(err.message);
        }
      });
    } else {
      // This is our second request. We will resolve the promise created in the
      // first request.
      console.log('....');
      globalThis.resolve();
      globalThis.resolve = undefined;
      console.log('test2');
    }

  }
};

```

Some details on how this works:

* Per the original fixup, every IoContext has an associated opaque JS object called the "promise tag"
* Originally, this promise tag was just a simple, ordinary JS object. With this PR it becomes an opaque wrapper around an `kj::Own<IoContext::WeakRef>`
* When an `IoContext` becomes current/locked, that `IoContext`s promise tag is associated with the `v8::Isolate` as the "current promise tag"
* Per the original fixup, whenever a `Promise` is created, the Isolate's "current promise tag" is associated with the Promise in an internal field.
* In the original fixup, when a `Promise` is followed (.then(...), .catch(...), etc), we check to see if the current promise tag matches the promises context tag. If they are not the same, then the promise is a "cross-context promise" and we arrange for proper cross-context signaling.
* This PR *adds* that when the `resolve(...)` or `reject(...)` methods are called, and we determine that the promise is a cross-context promise, scheduling of the promise continuations will be deferred until after we re-enter the promise's original `IoContext`.
* If the promise's original `IoContext` has been destroyed, we drop the continuations on the floor and emit a warning.

### DIFF
--- a/patches/v8/0019-Implement-cross-request-context-promise-resolve-hand.patch
+++ b/patches/v8/0019-Implement-cross-request-context-promise-resolve-hand.patch
@@ -1,0 +1,471 @@
+From 8cf4fe0f72fbaf25028ebd152559bd363da6599e Mon Sep 17 00:00:00 2001
+From: James M Snell <jasnell@gmail.com>
+Date: Mon, 16 Sep 2024 09:56:04 -0700
+Subject: [PATCH] Implement cross-request context promise resolve handling
+
+---
+ include/v8-callbacks.h                      | 21 ++++++++++
+ include/v8-isolate.h                        |  2 +
+ src/api/api.cc                              |  7 +++-
+ src/builtins/promise-abstract-operations.tq |  6 ++-
+ src/builtins/promise-resolve.tq             |  4 +-
+ src/execution/isolate-inl.h                 | 21 +++++++---
+ src/execution/isolate.cc                    | 24 +++++++++++-
+ src/execution/isolate.h                     | 19 +++++++--
+ src/heap/factory.cc                         | 11 +++---
+ src/objects/js-promise.h                    |  5 +++
+ src/objects/objects.cc                      | 43 +++++++++++++++++++++
+ src/roots/roots.h                           |  3 +-
+ src/runtime/runtime-promise.cc              | 30 +++++++++++---
+ src/runtime/runtime.h                       |  3 +-
+ 14 files changed, 171 insertions(+), 28 deletions(-)
+
+diff --git a/include/v8-callbacks.h b/include/v8-callbacks.h
+index 25f0c8189e5..f7fbae861a5 100644
+--- a/include/v8-callbacks.h
++++ b/include/v8-callbacks.h
+@@ -471,6 +471,27 @@ using PromiseCrossContextCallback = MaybeLocal<Promise> (*)(Local<Context> conte
+                                                             Local<Promise> promise,
+                                                             Local<Object> tag);
+ 
++/**
++ * PromiseCrossContextResolveCallback is called when resolving or rejecting a
++ * pending promise whose context tag is not strictly equal to the isolate's
++ * current promise context tag. The callback is called with the promise to be
++ * resolved, it's context tag, and a function that when called, causes the
++ * reactions to the resolved promise to be enqueued. The idea is that the
++ * embedder sets this callback in the case it needs to defer the actual
++ * scheduling of the reactions to the given promise to a later time.
++ * Importantly, when this callback is invoked, the state of the promise
++ * should have already been updated. We're simply possibly deferring the
++ * enqueue of the reactions to the promise.
++ */
++using PromiseCrossContextResolveCallback = Maybe<void> (*)(
++    v8::Isolate* isolate,
++    Local<Value> tag,
++    Local<Data> reactions,
++    Local<Value> argument,
++    std::function<void(v8::Isolate* isolate,
++                       Local<Data> reactions,
++                       Local<Value> argument)> callback);
++
+ }  // namespace v8
+ 
+ #endif  // INCLUDE_V8_ISOLATE_CALLBACKS_H_
+diff --git a/include/v8-isolate.h b/include/v8-isolate.h
+index 8112231319b..9fe57f7bde4 100644
+--- a/include/v8-isolate.h
++++ b/include/v8-isolate.h
+@@ -1732,6 +1732,8 @@ class V8_EXPORT Isolate {
+ 
+   class PromiseContextScope;
+   void SetPromiseCrossContextCallback(PromiseCrossContextCallback callback);
++  void SetPromiseCrossContextResolveCallback(
++      PromiseCrossContextResolveCallback callback);
+ 
+   Isolate() = delete;
+   ~Isolate() = delete;
+diff --git a/src/api/api.cc b/src/api/api.cc
+index d3fc6d77ea3..248a78ad24c 100644
+--- a/src/api/api.cc
++++ b/src/api/api.cc
+@@ -12246,12 +12246,17 @@ void Isolate::SetPromiseCrossContextCallback(PromiseCrossContextCallback callbac
+   isolate->set_promise_cross_context_callback(callback);
+ }
+ 
++void Isolate::SetPromiseCrossContextResolveCallback(PromiseCrossContextResolveCallback callback) {
++  i::Isolate* isolate = reinterpret_cast<i::Isolate*>(this);
++  isolate->set_promise_cross_context_resolve_callback(callback);
++}
++
+ Isolate::PromiseContextScope::PromiseContextScope(Isolate* isolate, v8::Local<v8::Object> tag)
+     : isolate_(reinterpret_cast<i::Isolate*>(isolate)) {
+   DCHECK(!isolate_->has_promise_context_tag());
+   DCHECK(!tag.IsEmpty());
+   i::Handle<i::Object> handle = Utils::OpenHandle(*tag);
+-  isolate_->set_promise_context_tag(*handle);
++  isolate_->set_promise_context_tag(handle);
+ }
+ 
+ Isolate::PromiseContextScope::~PromiseContextScope() {
+diff --git a/src/builtins/promise-abstract-operations.tq b/src/builtins/promise-abstract-operations.tq
+index fdec3d268e4..456d051d3c0 100644
+--- a/src/builtins/promise-abstract-operations.tq
++++ b/src/builtins/promise-abstract-operations.tq
+@@ -23,6 +23,9 @@ extern transitioning runtime PromiseRejectEventFromStack(
+ 
+ extern transitioning runtime PromiseContextCheck(
+     implicit context: Context)(JSPromise): JSPromise;
++
++extern transitioning runtime PromiseResolveContextCheck(
++    implicit context: Context)(JSPromise): JSAny;
+ }
+ 
+ // https://tc39.es/ecma262/#sec-promise-abstract-operations
+@@ -239,7 +242,8 @@ transitioning builtin RejectPromise(
+   // the runtime handle this operation, which greatly reduces
+   // the complexity here and also avoids a couple of back and
+   // forth between JavaScript and C++ land.
+-  if (IsIsolatePromiseHookEnabledOrDebugIsActiveOrHasAsyncEventDelegate(
++  if (ToBoolean(runtime::PromiseResolveContextCheck(promise)) ||
++      IsIsolatePromiseHookEnabledOrDebugIsActiveOrHasAsyncEventDelegate(
+           promiseHookFlags) ||
+       !promise.HasHandler()) {
+     // 7. If promise.[[PromiseIsHandled]] is false, perform
+diff --git a/src/builtins/promise-resolve.tq b/src/builtins/promise-resolve.tq
+index 202180adbba..c93ac5905d7 100644
+--- a/src/builtins/promise-resolve.tq
++++ b/src/builtins/promise-resolve.tq
+@@ -96,7 +96,9 @@ transitioning builtin ResolvePromise(
+   // We also let the runtime handle it if promise == resolution.
+   // We can use pointer comparison here, since the {promise} is guaranteed
+   // to be a JSPromise inside this function and thus is reference comparable.
+-  if (IsIsolatePromiseHookEnabledOrDebugIsActiveOrHasAsyncEventDelegate() ||
++
++  if (ToBoolean(runtime::PromiseResolveContextCheck(promise)) ||
++      IsIsolatePromiseHookEnabledOrDebugIsActiveOrHasAsyncEventDelegate() ||
+       TaggedEqual(promise, resolution))
+     deferred {
+       return runtime::ResolvePromise(promise, resolution);
+diff --git a/src/execution/isolate-inl.h b/src/execution/isolate-inl.h
+index 9e0969460fd..aa2b1cea6b7 100644
+--- a/src/execution/isolate-inl.h
++++ b/src/execution/isolate-inl.h
+@@ -134,26 +134,35 @@ bool Isolate::is_execution_terminating() {
+          i::ReadOnlyRoots(this).termination_exception();
+ }
+ 
+-Tagged<Object> Isolate::promise_context_tag() {
+-  return promise_context_tag_;
++Handle<Object> Isolate::promise_context_tag() {
++  return root_handle(RootIndex::kPromiseContextTag);
+ }
+ 
+ bool Isolate::has_promise_context_tag() {
+-  return promise_context_tag_ != ReadOnlyRoots(this).the_hole_value();
++  return heap()->promise_context_tag() != ReadOnlyRoots(this).the_hole_value();
+ }
+ 
+ void Isolate::clear_promise_context_tag() {
+-  set_promise_context_tag(ReadOnlyRoots(this).the_hole_value());
++  heap()->set_promise_context_tag(ReadOnlyRoots(this).the_hole_value());
+ }
+ 
+-void Isolate::set_promise_context_tag(Tagged<Object> tag) {
+-  promise_context_tag_ = tag;
++void Isolate::set_promise_context_tag(Handle<Object> tag) {
++  heap()->set_promise_context_tag(*tag);
+ }
+ 
+ void Isolate::set_promise_cross_context_callback(PromiseCrossContextCallback callback) {
+   promise_cross_context_callback_ = callback;
+ }
+ 
++void Isolate::set_promise_cross_context_resolve_callback(
++    PromiseCrossContextResolveCallback callback) {
++  promise_cross_context_resolve_callback_ = callback;
++}
++
++bool Isolate::has_promise_context_resolve_callback() {
++  return promise_cross_context_resolve_callback_ != nullptr;
++}
++
+ #ifdef DEBUG
+ Tagged<Object> Isolate::VerifyBuiltinsResult(Tagged<Object> result) {
+   if (is_execution_terminating() && !v8_flags.strict_termination_checks) {
+diff --git a/src/execution/isolate.cc b/src/execution/isolate.cc
+index 94194d59bae..19e01f57240 100644
+--- a/src/execution/isolate.cc
++++ b/src/execution/isolate.cc
+@@ -593,8 +593,6 @@ void Isolate::Iterate(RootVisitor* v, ThreadLocalTop* thread) {
+                       FullObjectSlot(&thread->pending_message_));
+   v->VisitRootPointer(Root::kStackRoots, nullptr,
+                       FullObjectSlot(&thread->context_));
+-  v->VisitRootPointer(Root::kStackRoots, nullptr,
+-                      FullObjectSlot(&promise_context_tag_));
+ 
+   for (v8::TryCatch* block = thread->try_catch_handler_; block != nullptr;
+        block = block->next_) {
+@@ -7459,5 +7457,27 @@ MaybeHandle<JSPromise> Isolate::RunPromiseCrossContextCallback(Handle<NativeCont
+   return v8::Utils::OpenHandle(*result);
+ }
+ 
++Maybe<void> Isolate::RunPromiseCrossContextResolveCallback(v8::Isolate* isolate,
++                                                           Handle<JSObject> tag,
++                                                           Handle<Object> reactions,
++                                                           Handle<Object> argument,
++                                                           PromiseReaction::Type type) {
++  CHECK(promise_cross_context_resolve_callback_ != nullptr);
++  return promise_cross_context_resolve_callback_(
++      isolate,
++      v8::Utils::ToLocal(tag),
++      v8::Utils::ToLocal(reactions),
++      v8::Utils::ToLocal(argument),
++      [type](v8::Isolate* isolate,
++             v8::Local<v8::Data> reactions,
++             v8::Local<v8::Value> argument) {
++    JSPromise::ContinueTriggerPromiseReactions(
++        reinterpret_cast<Isolate*>(isolate),
++        Utils::OpenHandle(*reactions),
++        Utils::OpenHandle(*argument),
++        type);
++  });
++}
++
+ }  // namespace internal
+ }  // namespace v8
+diff --git a/src/execution/isolate.h b/src/execution/isolate.h
+index 59926e68e97..1bd5193b054 100644
+--- a/src/execution/isolate.h
++++ b/src/execution/isolate.h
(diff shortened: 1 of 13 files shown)


jasnell:
TODOs

* Handle reject() ... Currently this only deals with the resolve case. Needs to handle the reject case also but that will follow the identical pattern. 
* Tests! Make sure all the various ways promises are resolved are handled. Async iterators, custom thenables, etc
* Documentation ... We should add documentation to the website

patches/v8/0019-Implement-cross-request-context-promise-resolve-hand.patch
```diff
@@ -0,0 +1,302 @@
...
++ * embedder sets this callback in the case it needs to defer the actual
++ * scheduling of the reactions to the given promise to a later time.
++ * Importantly, when this callback is invoked, the state of the promise
++ * should have already been updated. We're simply possibly deferring the
++ * enqueue of the reactions to the promise.
++ */
++using PromiseCrossContextResolveCallback = void (*)(
++    v8::Isolate* isolate,
++    Local<Value> tag,
++    Local<Value> reactions,
```
kentonv, line 43: I was hoping this callback could take a `Local<Promise::Resolver>` and just say "please resolve me in the right context". That seems cleaner as it doesn't expose any of the internals of promises and doesn't need an `std::function` since the embedder ends up calling an existing public API.

Maybe this callback should hook in a little bit earlier than it does? I was basically thinking that when you call `resolve()`, the _first thing_ that happens should be to check the tag. If we're in the wrong context, the callback gets invoked without altering the state of the promise at all...

jasnell: That approach would require quite a more invasive approach when it comes to handling things like custom thenables and resolving a promise with a promise (e.g. `resolve({ then(res, rej) { ... } })` and `resolve(promise)` ... since the v8 currently updates the state of the promise before evaluating those. The actual fulfillment or rejection of the promise is deferred until completion of those.

Also, per https://tc39.es/ecma262/#sec-promise-resolve-functions the language spec requires that when `resolve()` or `reject()` are called, the state of the promise is updated *immediately* and synchronously. Deferring the state change would break this spec-defined invariant and may have other side effects.

While we may be able to hide more of the details of the `reactions` and `argument` from the callback, I think the approach taken in this PR is the best both in terms of minimizing the changes to v8 and preserving spec-defined invariants.

kentonv: > That approach would require quite a more invasive approach when it comes to handling things like custom thenables and resolving a promise with a promise (e.g. resolve({ then(res, rej) { ... } }) and resolve(promise) ... since the v8 currently updates the state of the promise before evaluating those. The actual fulfillment or rejection of the promise is deferred until completion of those.

OK I guess that makes sense. Or to put it another way:
* When `resolve(promise2)` is called, we need `promise2` to run in the calling context, not in the resolver's context. So we can't call the callback yet.
* But we do need to mark the original promise resolved at this time. It would be problematic to wait until `promise2` finishes before doing that, otherwise someone else could call `resolve()` during that time with a non-promise value and it would actually preempt `promise2`, incorrectly.
* So, by the time `promise2` resolves, and we need to call the callback, we can't just pass the resolver itself, because it's already marked resolved and so won't work.

jasnell:
Patch updated to address the linux-debug CI failure, and added an additional test case.

patches/v8/0019-Implement-cross-request-context-promise-resolve-hand.patch
```diff
@@ -0,0 +1,339 @@
...
++extern transitioning runtime PromiseResolveContextCheck(
++    implicit context: Context)(JSPromise): JSAny;
+ }
+ 
+ // https://tc39.es/ecma262/#sec-promise-abstract-operations
+@@ -239,7 +242,8 @@ transitioning builtin RejectPromise(
+   // the runtime handle this operation, which greatly reduces
+   // the complexity here and also avoids a couple of back and
+   // forth between JavaScript and C++ land.
+-  if (IsIsolatePromiseHookEnabledOrDebugIsActiveOrHasAsyncEventDelegate(
```
jasnell, line 100: There are generally two paths for a promise to be resolved/rejected... one via runtime (logic in objects.cc) and one via these torque built-ins. There is existing logic on the torque side to always defer to the runtime logic when either promise hooks are being used or the inspector is enabled. In this patch we add a third case to this check to defer to the runtime logic if the promise tag does not match the current isolate promise tag.

patches/v8/0019-Implement-cross-request-context-promise-resolve-hand.patch
```diff
@@ -0,0 +1,339 @@
...
++++ b/src/execution/isolate.cc
+@@ -7459,5 +7459,27 @@ MaybeHandle<JSPromise> Isolate::RunPromiseCrossContextCallback(Handle<NativeCont
+   return v8::Utils::OpenHandle(*result);
+ }
+ 
++void Isolate::RunPromiseCrossContextResolveCallback(v8::Isolate* isolate,
++                                                    Handle<JSObject> tag,
++                                                    Handle<Object> reactions,
++                                                    Handle<Object> argument,
++                                                    PromiseReaction::Type type) {
```
jasnell, line 153: Here we call out to our embedder-configured callback, giving it an std::function and a couple of arguments to call when we are ready to actually schedule the continuations to run.

The `reactions` is an opaque value that is holding the reactions to be scheduled.
The `argument` is the resolve value or exception that the promise was fulfilled with.

patches/v8/0019-Implement-cross-request-context-promise-resolve-hand.patch
```diff
@@ -0,0 +1,339 @@
...
++        !isolate->has_promise_context_resolve_callback());
++  if (needs_promise_context_switch) {
++    isolate->RunPromiseCrossContextResolveCallback(
++        reinterpret_cast<v8::Isolate*>(isolate),
++        Handle<JSObject>(Cast<JSObject>(obj), isolate),
++        reactions,
++        value,
++        PromiseReaction::kFulfill);
++    return isolate->factory()->undefined_value();
++  }
```
jasnell, line 252: This is the actual meat of the patch. If the promise has no context tag, or the context tag matches, or there is no configured callback, then we'll schedule the continuations synchronously per the default behavior. If the promise tag does not match, then we defer scheduling the continuations by calling out to our embedder-configured callback. This needs to happen in both the resolve and reject paths.

jasnell:
CI is failing due to a transient issue downloading stuff....

src/workerd/io/worker.c++
```diff
@@ -1142,6 +1142,59 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
...
+    // promise.
+    lock->v8Isolate->SetPromiseCrossContextResolveCallback(
+        [](v8::Isolate* isolate, v8::Local<v8::Value> tag, v8::Local<v8::Data> reactions,
+            v8::Local<v8::Value> argument,
+            std::function<void(v8::Isolate * isolate, v8::Local<v8::Data> reactions,
+                v8::Local<v8::Value> argument)> callback) {
+      auto& js = jsg::Lock::from(isolate);
+
+      try {
+        auto& ref = jsg::unwrapOpaqueRef<kj::Own<IoContext::WeakRef>>(isolate, tag);
```
kentonv, line 1157: Ugh I remembered that this doesn't work.

Recapping what we already discussed in person:
* We might be in a different thread.
* The WeakRef is a thread-specific object, we cannot touch it from here. (In fact, storing an `IoContext::WeakRef` in a GC'd object at all can lead to problems.)
* We can't call `context->run()` directly from here either, because we're on the wrong thread.
* Instead, we should extend the `DeleteQueue`, since it is already designed for this sort of cross-thread signalling.
* We could append an item to the `DeleteQueue` whose destructor invokes the code we want. That gets us half-way there.
* But we also have to cause the destination IoContext to wake up promptly to process the queue.
* To that end, each IoContext should create a `kj::newPromiseAndCrossThreadResolver()` when first constructed.
* Put the fulfiller end into the `DeleteQueue`, so that it can be called by the `PromiseCrossContextResolveCallback` from another thread.
* Wait on the promise end. When it resolves, call `run()` to force the `DeleteQueue` to be processed. Then inside the run callback (while the isolate is still locked), replace the promise and fulfiller with a new pair, so that it can be signaled again.

patches/v8/0019-Implement-cross-request-context-promise-resolve-hand.patch
```diff
@@ -0,0 +1,339 @@
...
+   static Handle<Object> TriggerPromiseReactions(Isolate* isolate,
+diff --git a/src/objects/objects.cc b/src/objects/objects.cc
+index 76e284e19a2..32ccc84d8f8 100644
+--- a/src/objects/objects.cc
++++ b/src/objects/objects.cc
+@@ -4878,6 +4878,21 @@ Handle<Object> JSPromise::Fulfill(DirectHandle<JSPromise> promise,
+   // 6. Set promise.[[PromiseState]] to "fulfilled".
+   promise->set_status(Promise::kFulfilled);
+ 
++  Tagged<Object> obj = promise->context_tag();
```
erikcorry, line 239: I think it's preferable to move this to a GC-safe Handle as soon as possible.

```
  Handle<Object> obj(promise->context_tag(), isolate);
  Handle<Object> tag(isolate->promise_context_tag(), isolate);
  bool needs_promise_context_switch =
      !(*obj == Smi::zero() ||
        obj.is_identical_to(tag) ||
        !isolate->has_promise_context_resolve_callback());
  if (needs_promise_context_switch) {
    isolate->RunPromiseCrossContextResolveCallback(
        reinterpret_cast<v8::Isolate*>(isolate),
        Cast<JSObject>(obj),
```

erikcorry:
Here's a diff to the diff file, which uses Handles more, and raw Tagged pointers less.  Feel free to use it, or I'll just do a separate change tomorrow (where I'll probably also unify the 0007 and 0019 patches to one file).
[handles.txt](https://github.com/user-attachments/files/17035159/handles.txt)

src/workerd/io/io-own.h
```diff
@@ -120,9 +121,20 @@ class DeleteQueue: public kj::AtomicRefcounted {
   DeleteQueue(): crossThreadDeleteQueue(State{kj::Vector<OwnedObject*>()}) {}
 
   void scheduleDeletion(OwnedObject* object) const;
+  void scheduleAction(jsg::Lock& js, kj::Function<void(jsg::Lock&)>&& action) const;
 
   struct State {
     kj::Vector<OwnedObject*> queue;
+    // Collection of actions to be run immediately after pending deletions are processed.
```
kentonv, line 128: This comment seems like it's mostly addressed to me, telling me why my idea didn't work. :) I agree, you're right, my idea doesn't work, but probably future readers won't start by thinking of it and so would be better served by a different comment, like:

> Actions that some other IoContext has requested be executed in this IoContext. When adding an action to this list, `crossThreadFulfiller` should be fulfilled, signaling the target IoContext to wake up and run actions. After draining the actions queue, the target IoContext should replace `crossThreadFulfiller` with a new one which will wake it up again.
>
> In particular, these actions are used to implement cross-context promise resolution.
>
> Keep in mind the IoContext could be destroyed before the cross-thread signal runs, in which case the actions will never run.

src/workerd/io/io-own.c++
```diff
@@ -33,6 +33,38 @@ void DeleteQueue::scheduleDeletion(OwnedObject* object) const {
...
+      if (IoContext::hasCurrent()) {
+        auto err = v8::Exception::Error(
+            js.str("A promise was resolved or rejected from a different request context than "
+                   "the one it was created in. However, the creating context has already been "
+                   "destroyed and continuations for that request likely will not run correctly"_kj))
+                       .As<v8::Object>();
+        jsg::check(err->Set(js.v8Context(), js.str("name"_kj), js.str("Warning"_kj)));
+        auto stack = jsg::check(err->Get(js.v8Context(), js.str("stack"_kj)));
+
+        // Safe to mutate here since we have the exclusive lock on the queue above.
```
kentonv, line 59: I'm concerned that we're calling a lot of JavaScript here while holding the DeleteQueue's lock. It seems entirely possible that we'll end up calling something else that tries to take the lock -- e.g. if we trigger GC. So I think we actually need to drop the lock before calling any V8 APIs.

This comment here implies that the lock is providing some sort of safety but I don't see what that is -- the lock only protects the DeleteQueue itself, so if we're done manipulating the DeleteQueue, we can drop it.

src/workerd/io/io-own.c++
```diff
@@ -33,6 +33,38 @@ void DeleteQueue::scheduleDeletion(OwnedObject* object) const {
...
+        IoContext::current().logWarning(kj::str(stack));
+        // TODO(soon): Add documentation link to this warning.
+      }
+      action(js);
```
kentonv, line 63: Why call the action in the wrong context here? Shouldn't we simply drop it without calling it?

src/workerd/io/io-own.h
```diff
@@ -145,6 +157,30 @@ class DeleteQueue: public kj::AtomicRefcounted {
...
+
+// Object which can push actions into a specific DeleteQueue then signal it's
+// owning IoContext to wake up to process the queue. This is a bit of a hack of
+// the DeleteQueue concept that allows us to use the same queue for more than
+// just deletions.
+class IoDeferral {
+public:
+  IoDeferral(kj::Own<const DeleteQueue> deleteQueue): deleteQueue(kj::mv(deleteQueue)) {}
+
+  // Defers the specified action to the owning IoContext. The next time the
```
kentonv, line 174: This comment implies that we won't proactively wake up the target context, but instead just wait until it happens to be entered for some other reason. But actually, we do proactively wake it up.

I think maybe this class should be called `IoCrossContextExecutor` and the `defer()` method should be called `execute()`.

src/workerd/io/io-own.c++
```diff
@@ -33,6 +33,38 @@ void DeleteQueue::scheduleDeletion(OwnedObject* object) const {
   }
 }
 
+void DeleteQueue::scheduleAction(jsg::Lock& js, kj::Function<void(jsg::Lock&)>&& action) const {
+  if (IoContext::hasCurrent() && IoContext::current().deleteQueue.get() == this) {
+    // Same thread! No need to enqueue.
+    action(js);
```
kentonv, line 39: IMO even if we're already in the right context, it would be better to enqueue the action for asynchronous execution the same way as if we weren't. Otherwise, in some cases it can be tricky to write the callback in a way that's safe to run both synchronously and asynchronously. The code below that schedules cross-context should work just fine even if we happen to be in the same context.

Our target use case won't ever schedule on the same context anyway, of course, so it can't hurt there.

src/workerd/io/io-own.h
```diff
@@ -195,6 +231,9 @@ class DeleteQueuePtr: public kj::Own<DeleteQueue> {
   ~DeleteQueuePtr() noexcept(false) {
     auto ptr = get();
     if (ptr != nullptr) {
+      // The delete queue state may include a kj::CrossThreadPromiseFulfiller that
```
kentonv, line 234: Please only delete the CrossThreadPromiseFulfiller itself in AllowAsyncDestructorScope. We might also be deleting scheduled action functions here, and we should keep the restriction that those should not contain KJ objects.

kentonv (commented):
Not quite done reviewing but need to go to meetings, so here's my comments so far.

src/workerd/io/io-own.c++
jasnell: Discussed in chat... summary of outcome: We'll introduce a compat flag for this entire change. With the flag off, the current behavior of running these actions in the wrong IoContext will be preserved, just in case. With the compat flag set the actions will be dropped on the floor if the original IoContext has been destroyed.

src/workerd/io/io-own.h
```diff
@@ -167,14 +168,15 @@ class DeleteQueue: public kj::AtomicRefcounted {
...
 // just deletions.
-class IoDeferral {
+class IoCrossContextExecutor {
 public:
-  IoDeferral(kj::Own<const DeleteQueue> deleteQueue): deleteQueue(kj::mv(deleteQueue)) {}
+  IoCrossContextExecutor(kj::Own<const DeleteQueue> deleteQueue)
+      : deleteQueue(kj::mv(deleteQueue)) {}
 
-  // Defers the specified action to the owning IoContext. The next time the
+  // Tries to execute the specified action to the owning IoContext. The next time the
```
kentonv, line 176: I think this comment is still misleading. Instead of saying "The next time the IoContext is entered", it should say that the target IoContext will be signaled to run the action as soon as it is able.

src/workerd/io/worker.c++
```diff
@@ -1145,6 +1145,96 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
...
+              jsg::throwInternalError(js.v8Isolate, kj::mv(ex));
+            }
+          });
+        } catch (jsg::JsExceptionThrown&) {
+          // Exceptions here are generally unexpected but possible because the jsg::Promise
+          // then can fail if the isolate is in the process of being torn down. Let's just
+          // return control back to V8 which should handle the case.
+        } catch (...) {
+          auto ex = kj::getCaughtExceptionAsKj();
+          KJ_LOG(ERROR, "Scheduling deferred promise continuations failed unexpectedly", ex);
```
kentonv, line 1233: I think the exception will get logged twice since `throwInternalError()` will also log it?

(Same issue above on line 1223.)

src/workerd/io/io-context.c++
```diff
@@ -1304,11 +1310,33 @@ void IoContext::throwNotCurrentJsError(kj::Maybe<const std::type_info&> maybeTyp
...
+  // The promise that is returned is held by the IoContext itself, so when the
+  // IoContext is destroyed, the promise will be canceled and the loop will
+  // end. On each iteration of the loop we want to reset the cross thread
+  // signal in the delete queue, then wait on the promise. Once the promise
+  // is fulfilled, we will run an empty task to prompt the IoContext to drain
+  // the DeleteQueue.
+  try {
+    for (;;) {
+      co_await context->deleteQueue->resetCrossThreadSignal();
+      co_await context->run([](auto&) {});
```
kentonv, line 1329: I think we should drain the action queue in the lambda here, rather than every time the IoContext is entered. That seems like it will lead to more predictable behavior. We know this will always be executed when an action is queued.

src/workerd/io/io-context.c++
```diff
@@ -1304,11 +1310,33 @@ void IoContext::throwNotCurrentJsError(kj::Maybe<const std::type_info&> maybeTyp
...
+    }
+  } catch (...) {
+    auto exception = kj::getCaughtExceptionAsKj();
+    KJ_IF_SOME(a, context->actor) {
+      a.shutdownActorCache(exception);
```
kentonv, line 1334: Probably better to call `abort(kj::mv(exception))` here? (As in, `IoContext::abort`.) It will shut down the actor cache and also do a bunch of other things to cancel the request.

src/workerd/io/io-context.c++
```diff
@@ -1304,11 +1310,33 @@ void IoContext::throwNotCurrentJsError(kj::Maybe<const std::type_info&> maybeTyp
...
+    }
+    kj::throwFatalException(kj::mv(exception));
```
kentonv, line 1336: I guess this exception is never handled anywhere since the caller just stores the promise without ever waiting on it.

We probably need to log the exception instead of throwing it here? Or maybe `abort()`ing with it will be sufficient to make it get logged somewhere...

src/workerd/io/worker.c++
```diff
@@ -1145,6 +1145,96 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
...
+          auto& ref = jsg::unwrapOpaqueRef<kj::Own<IoCrossContextExecutor>>(isolate, tag);
+          ref->execute(js,
+              [reactions = v8::Global<v8::Data>(isolate, reactions),
+                  argument = jsg::JsRef(js, jsg::JsValue(argument)),
+                  callback = kj::mv(callback)](jsg::Lock& js) mutable {
+            try {
+              callback(js.v8Isolate, reactions.Get(js.v8Isolate), argument.getHandle(js));
+            } catch (jsg::JsExceptionThrown&) {
+              // Exceptions here are generally unexpected but possible. Let's just
+              // return control back to V8 which should handle the case.
```
kentonv, line 1220: Technically we are not returning control back to V8, we're returning it to the place that calls action callbacks.

If we move that to be inside a run() callback then probably we should actually let exceptions flow through here?

src/workerd/io/worker.c++
```diff
@@ -1145,6 +1145,96 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
...
+            } catch (...) {
+              auto ex = kj::getCaughtExceptionAsKj();
+              KJ_LOG(ERROR, "Scheduling deferred promise continuations failed unexpectedly", ex);
+              jsg::throwInternalError(js.v8Isolate, kj::mv(ex));
+            }
+          });
+        } catch (jsg::JsExceptionThrown&) {
+          // Exceptions here are generally unexpected but possible because the jsg::Promise
+          // then can fail if the isolate is in the process of being torn down. Let's just
+          // return control back to V8 which should handle the case.
```
kentonv, line 1230: In this case we are in fact returning to V8. But will V8 actually do the right thing if an exception is scheduled during the callback? Normally in V8 code we would return an empty handle or empty maybe to signal that an exception is scheduled, so that the caller can return early. Maybe the callback should return a boolean so that the call site can do the right thing in the exception case?

kentonv (commented):
OK I actually read through everything this time (except the V8 patch).

src/workerd/io/worker.c++
jasnell: Got it figured out I think. V8 was not properly checking if an exception was being scheduled in the reject path causing it to not properly handle errors here. I *think* the patch now covers the relevant case but it's possible there are other places? I need to keep digging but let's see if we can make progress on landing this now

src/workerd/io/worker.c++
```diff
@@ -1213,25 +1213,14 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
...
-              callback(js.v8Isolate, reactions.Get(js.v8Isolate), argument.getHandle(js));
-            } catch (jsg::JsExceptionThrown&) {
-              // Exceptions here are generally unexpected but possible. Let's just
-              // return control back to V8 which should handle the case.
-            } catch (...) {
-              auto ex = kj::getCaughtExceptionAsKj();
-              KJ_LOG(ERROR, "Scheduling deferred promise continuations failed unexpectedly", ex);
-              jsg::throwInternalError(js.v8Isolate, kj::mv(ex));
-            }
+            callback(js.v8Isolate, reactions.Get(js.v8Isolate), argument.getHandle(js));
```
kentonv, line 1216: Oh it occurs to me, this is a callback _to_ v8, so it can't possibly throw anyway, right? So more reason why it was right to remove the try/catch.

Though it raises a different question: What if the callback failed, e.g. due to termination? That doesn't seem to be signaled at all. Does the callback actually need to return a bool or something?

jasnell: I looked at this but the only thing the callback does is call the function to schedule the microtasks and that always returns `undefined`. It does not appear to signal any failure conditions and does not appear to return an empty handle to signal any kind of failure.

kentonv: Fair enough.

kentonv (approved):
Approving workerd code... trusting @erikcorry to review V8 code.

jasnell:
Rebased and squashed

erikcorry (approved):
LGTM

PR #2724, opened 2024-09-17
Fbaetens/bump sqlite kv limit & improve error reporting

Frederik-Baetens:


src/workerd/io/actor-cache.c++
```diff
@@ -577,6 +582,7 @@ kj::OneOf<ActorCache::GetResultList, kj::Promise<ActorCache::GetResultList>> Act
   {
     auto lock = lru.cleanList.lockExclusive();
     for (auto& key: keys) {
+      ActorStorageLimits::checkMaxKeySize(key); // new, can't find where this check is done for getMultiple, should I leave this out?
```
Frederik-Baetens, line 585: Will delete this comment, but yeah, I think this is new, so should I leave this check out since it wasn't being done previously as far as I can tell?

Frederik-Baetens:
This also has an internal PR

src/workerd/api/actor-state.c++
```diff
@@ -554,16 +544,12 @@ jsg::Promise<void> DurableObjectStorageOperations::putMultiple(
     // deleting an undefined field is confusing, throwing could break otherwise working code, and
     // a stray undefined here or there is probably closer to what the user desires.
 
-    ActorStorageLimits::checkMaxKeySize(field.name);
-
     kj::Array<byte> buffer = serializeV8Value(js, field.value);
-    ActorStorageLimits::checkMaxValueSize(field.name, buffer);
```
justin-mp, line 560: Hmm, by doing this means that if we get a giant value we're going to keep it around until the check kicks in later.  I guess this is probably safe, if only because if the value were too huge, the isolate would have run out of memory before we got here.

src/workerd/api/actor-state.c++
```diff
@@ -554,16 +544,12 @@ jsg::Promise<void> DurableObjectStorageOperations::putMultiple(
...
 
     units += billingUnits(field.name.size() + buffer.size());
 
     kvs.add(ActorCacheOps::KeyValuePair{kj::mv(field.name), kj::mv(buffer)});
   }
-
```
justin-mp, line 566: nit: spurious whitespace change?

src/workerd/io/actor-cache.c++
justin-mp: You should leave this check out because it's not backwards compatible.

That being said, we probably want to add this check.  In a follow-up PR, we can add it as a compatibility date update.

It's weird that we don't call checkMaxPairsCount() for putMultiple and deleteMultiple.  According to our [docs](https://developers.cloudflare.com/durable-objects/api/transactional-storage-api/), get, put, and delete are all supposed to have a 128 entry limit.

src/workerd/util/sqlite-kv.h
```diff
@@ -5,9 +5,23 @@
 #pragma once
 
 #include "sqlite.h"
+#include <workerd/jsg/exception.h>
 
 namespace workerd {
 
+namespace {
+
+// Small class which is used to customeze certain aspects of the underlying sql operations
```
justin-mp, line 14: spelling: customeze

src/workerd/util/sqlite-kv.h
```diff
@@ -5,9 +5,23 @@
...
+// In this case we just customise the error reporting to emit JSG user visible errors instead
+// of KJ exceptions which become internal errors.
+class SqliteKvRegulator: public SqliteDatabase::Regulator {
+  void onError(kj::StringPtr message) const override {
+    JSG_ASSERT(false, Error, message);
```
justin-mp, line 19: This implementation will throw JSG errors for all possible SQLite errors.  Almost everything about this class is under our control so every error EXCEPT the TOOBIG error should be an internal error.  Is it possible to report only TOOBIG errors this way?

Frederik-Baetens: Yes, but not without either doing matching on the string message (bad), or changing the onError API (out of scope before birthday week).

What I'll do now is only pass the regulator to the put query, so only sql errors for put get passed through. Or would you prefer I do string matching as a temporary kludge here?

src/workerd/io/actor-cache.c++
Frederik-Baetens: :+1: removed this check for now.

src/workerd/util/sqlite-kv.h
Frederik-Baetens: Changed it to only report errors on the put query. Let me know your thoughts on that. I acknowledge it's a bit of a hack, but I think its a low consequence change here. Reporting an SQLite error that's actually our fault to the user doesn't seem like a massive issue for sqlite here.

Frederik-Baetens: Hmm, including the error code in the onError doesn't seem that difficult actually. I'll try to get something out based on that.

src/workerd/api/pyodide/pyodide-test.c++
```diff
@@ -3,6 +3,7 @@
 //     https://opensource.org/licenses/Apache-2.0
 
 #include "pyodide.h"
+
```
Frederik-Baetens, line 6: This got in from running the formatter.

src/workerd/util/sqlite-kv.h
Frederik-Baetens: Pushed a fixup commit passing the error code to onError. Let me know what you think of that, and I'll squash it if you think it's a good approach here.

src/workerd/util/sqlite-kv.h
```diff
@@ -15,8 +17,22 @@ namespace {
...
+        case SQLITE_TOOBIG:
+          // We want to return SQLITE_TOOBIG to the user since it's usually because of user error.
+          JSG_ASSERT(false, Error, message);
+          return;
+        // We don't want to return other errors to the user since they're usually our fault.
+        // In that case we do nothing because the contract of onError is not to handle the error in
+        // its entirity, but instead to optionally handle it, and do nothing otherwise.
+        // When onError does nothing, the code calling into onError is still responsible for
+        // handling the error by other means, usually by throwing a KJ exception itself.
+        default: {
```
justin-mp, line 32: rather than doing this dance with the braces, just do

```
default:
    return;
```

You may also want to put a KJ_UNREACHABLE at the after the switch statement to be clear you're never supposed to get there.

Frederik-Baetens: Changed it to your suggestion, but I don't mind the fall through behavior as much because other codepaths (when the error code is kj::none) already just fall through to the end of the method. It seems nice to share that fallthrough path even if it doesn't do anything right now.

src/workerd/util/sqlite-kv.h
```diff
@@ -15,8 +17,22 @@ namespace {
 // In this case we just customize the error reporting to emit JSG user visible errors instead
 // of KJ exceptions which become internal errors.
 class SqliteKvRegulator: public SqliteDatabase::Regulator {
-  void onError(kj::StringPtr message) const override {
-    JSG_ASSERT(false, Error, message);
+  void onError(kj::Maybe<int> sqliteErrorCode, kj::StringPtr message) const override {
```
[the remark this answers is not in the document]
Frederik-Baetens: Did that, I had put it in the header file to be able to put it in an anonymous namespace, but I suppose that doesn't matter much.

justin-mp (commented):
I think the code itself is in good shape, but we're missing a test in sqlite-kv-test.c++.

Or is there a reason that we shouldn't include a test?

src/workerd/util/sqlite-kv.h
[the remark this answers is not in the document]
justin-mp: FYI for future reference, don't put anonymous namespaces in headers.  This causes everything that includes the header to compile its own copy of the stuff in the anonymous namespace.  This bloats the binary and could lead to ODR violations.

Frederik-Baetens:
I tested this in the internal PR, but I'll add a test there too then.

justin-mp (approved):

PR #2728, opened 2024-09-17
Add lock wait time metrics for Memory Cache.

mar-cf:
Record lock wait time for Memory cache and record as a tag in the trace.


mar-cf:
After https://github.com/cloudflare/workerd/pull/2717 is merged. I'll send this out for review. @mikea

a-robinson (approved):

Issue #2716, opened 2024-09-16
🐛 Bug Report — Runtime APIs - nodejs compat - crypto - `getRandomBytes()`


petebacondarwin:
The `getRandomBytes()` function will fail when called if it is not bound to the `webcrypto` object.
See https://github.com/cloudflare/workers-sdk/pull/6721 for a test case.
Currently the calls fail with: `TypeError: Illegal invocation: function called with incorrect `this` reference.`.

This has been fixed in unenv via https://github.com/unjs/unenv/pull/309 but I think that the `getRandomBytes()` function should not require any particular `this` to work.

kentonv:
What's `getRuntimeBytes()`? Did you mean `crypto.getRandomValues()`?

Assuming so, it has the same behavior in Chrome:

![Screenshot from 2024-09-16 15-47-53](https://github.com/user-attachments/assets/88c832d9-eca3-4371-ba43-5d70590071a3)

petebacondarwin:
To be more concrete. Using `nodejs_compat` (not v2, so no polyfills) running this Worker:

```ts
import { getBuiltinModule } from "node:process";

export default {
	async fetch(
		request: Request,
		env: Env,
		ctx: ExecutionContext
	): Promise<Response> {
		const results: string[] = [];
		const nodeCrypto = getBuiltinModule("crypto");
		const webcrypto = nodeCrypto.webcrypto;
		const getRandomValues = webcrypto.getRandomValues;
		const a = crypto.getRandomValues(new Uint8Array(6)).toString(); // global
		const c = webcrypto.getRandomValues(new Uint8Array(6)).toString(); // webcrypto
		const b = nodeCrypto.getRandomValues(new Uint8Array(6)).toString(); // namespace import
		const d = getRandomValues(new Uint8Array(6)).toString(); // named import
		return Response.json(results);
	},
};
```

Results in exceptions on these lines:

```ts
const b = nodeCrypto.getRandomValues(new Uint8Array(6)).toString(); // namespace import
const d = getRandomValues(new Uint8Array(6)).toString(); // named import
```

I could accept that the bare `getRandomValues()` call might not work. But I would expect the `nodeCrypto.getRandomValues()` call to work, which is the point of this issue.

petebacondarwin:
Also given that this is for Node.js compatibility, I tested it on vanilla Node.js (18):

```
> const crypto = require("crypto")
undefined
> crypto.getRandomValues(new Uint8Array(6))
[...]

PR #2730, opened 2024-09-17
fix invalid invocation on getRandomValues()

anonrig:
Fixes https://github.com/cloudflare/workerd/issues/2716

### DIFF
--- a/src/workerd/api/node/tests/crypto_random-test.js
+++ b/src/workerd/api/node/tests/crypto_random-test.js
@@ -413,3 +413,27 @@ export const timingSafeEqualTest = {
     timingSafeEqual(new Uint8Array(1), new Uint8Array(1));
   },
 };
+
+// Ref: https://github.com/cloudflare/workerd/issues/2716
+export const getRandomValuesIllegalInvocation = {
+  async test() {
+    const crypto = await import('node:crypto');
+    {
+      // The following assertion doesn't fail as of Node.js v20.17.0
+      const getRandomValues = crypto.getRandomValues;
+      strictEqual(getRandomValues(new Uint8Array(6)).length, 6);
+    }
+    throws(
+      () => {
+        // This ensures that we are replicating Node.js behavior as of v20.17.0
+        const getRandomValues = crypto.webcrypto.getRandomValues;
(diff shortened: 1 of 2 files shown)


src/workerd/api/node/tests/crypto_random-test.js
```diff
@@ -413,3 +413,12 @@ export const timingSafeEqualTest = {
     timingSafeEqual(new Uint8Array(1), new Uint8Array(1));
   },
 };
+
+// Ref: https://github.com/cloudflare/workerd/issues/2716
+export const getRandomValuesIllegalInvocation = {
+  async test() {
+    const crypto = await import('node:crypto');
+    const getRandomValues = crypto.getRandomValues;
```
petebacondarwin, line 421: Can you also check `const getRandomValues = crypto.webcrypto.getRandomValues`?

I think that this still is not able to be used unbound.

petebacondarwin (commented):
Thanks for fixing this. Can you check that other scenario too?

src/workerd/api/node/tests/crypto_random-test.js
petebacondarwin: Gosh! I was sure I tested that previously but I just tried now and I can confirm that this case does indeed throw.
Thanks!

src/workerd/api/node/tests/crypto_random-test.js
```diff
@@ -413,3 +413,27 @@ export const timingSafeEqualTest = {
     timingSafeEqual(new Uint8Array(1), new Uint8Array(1));
   },
 };
+
+// Ref: https://github.com/cloudflare/workerd/issues/2716
+export const getRandomValuesIllegalInvocation = {
+  async test() {
+    const crypto = await import('node:crypto');
+    {
+      // The following assertion doesn't fail as of Node.js v20.17.0
```
petebacondarwin, line 422: This doesn't fail for me in node 18 either...

petebacondarwin (approved):

IgorMinar:
thank you @anonrig!!

anonrig:
@IgorMinar 🫡

dom96 (approved):

PR #2734, opened 2024-09-18
Tweaked TS generation of DO SQlite types

geelen:
After #2722, we now have a final, very minimal SQLite API for durable objects. So I've just tweaked the TS generation to give a few generics to make things easier to work with. These are the changes:

```diff
diff --git a/.local/types-main/2023-07-01/index.ts b/bazel-bin/types/definitions/2023-07-01/index.ts
index 45460a4d..556a55d3 100755
--- a/.local/types-main/2023-07-01/index.ts
+++ b/bazel-bin/types/definitions/2023-07-01/index.ts
@@ -2778,20 +2778,24 @@ export declare const WebSocketPair: {
   };
 };
 export interface SqlStorage {
-  exec(query: string, ...bindings: any[]): SqlStorageCursor;
+  exec<T extends Record<string, SqlStorageValue>>(
+    query: string,
+    ...bindings: any[]
+  ): SqlStorageCursor<T>;
   get databaseSize(): number;
   Cursor: typeof SqlStorageCursor;
   Statement: typeof SqlStorageStatement;
 }
 export declare abstract class SqlStorageStatement {}
-export declare abstract class SqlStorageCursor {
-  raw(): IterableIterator<((ArrayBuffer | string | number) | null)[]>;
+export type SqlStorageValue = ArrayBuffer | string | number | null;
+export declare abstract class SqlStorageCursor<
+  T extends Record<string, SqlStorageValue>,
+> {
+  raw<U extends SqlStorageValue[]>(): IterableIterator<U>;
   get columnNames(): string[];
   get rowsRead(): number;
   get rowsWritten(): number;
-  [Symbol.iterator](): IterableIterator<
-    Record<string, (ArrayBuffer | string | number) | null>
-  >;
+  [Symbol.iterator](): IterableIterator<T>;
 }
 export interface Socket {
   get readable(): ReadableStream;
```

And the relevant diff between `2023-07-01` and `experimental`:

```diff
 export declare abstract class ScheduledEvent extends ExtendableEvent {
   readonly scheduledTime: number;
   readonly cron: string;
@@ -2782,6 +2836,8 @@ export interface SqlStorage {
     query: string,
     ...bindings: any[]
   ): SqlStorageCursor<T>;
+  prepare(query: string): SqlStorageStatement;
+  ingest(query: string): SqlStorageIngestResult;
   get databaseSize(): number;
   Cursor: typeof SqlStorageCursor;
   Statement: typeof SqlStorageStatement;
@@ -2797,6 +2853,12 @@ export declare abstract class SqlStorageCursor<
   get rowsWritten(): number;
   [Symbol.iterator](): IterableIterator<T>;
 }
+export interface SqlStorageIngestResult {
+  remainder: string;
+  rowsRead: number;
+  rowsWritten: number;
+  statementCount: number;
+}
 export interface Socket {
   get readable(): ReadableStream;
   get writable(): WritableStream;
```

### DIFF
--- a/src/workerd/api/sql.h
+++ b/src/workerd/api/sql.h
@@ -33,15 +33,12 @@ public:
   JSG_RESOURCE_TYPE(SqlStorage, CompatibilityFlags::Reader flags) {
     JSG_METHOD(exec);
 
-    // Prepared statement API is experimental-only and deprecated. exec() will automatically
-    // handle caching prepared statements, so apps don't need to worry about it.
     if (flags.getWorkerdExperimental()) {
+      // Prepared statement API is experimental-only and deprecated. exec() will automatically
+      // handle caching prepared statements, so apps don't need to worry about it.
       JSG_METHOD(prepare);
-    }
 
-    // Make sure that the 'ingest' function is still experimental-only if and when
-    // the SQL API becomes publicly available.
-    if (flags.getWorkerdExperimental()) {
+      // 'ingest' functionality is still experimental-only
       JSG_METHOD(ingest);
     }
 
@@ -49,6 +46,10 @@ public:
 
     JSG_NESTED_TYPE(Cursor);
     JSG_NESTED_TYPE(Statement);
+
+    JSG_TS_OVERRIDE({
+      exec<T extends Record<string, SqlStorageValue>>(query: string, ...bindings: any[]): SqlStorageCursor<T>
+    });
   }
 
   void visitForMemoryInfo(jsg::MemoryTracker& tracker) const;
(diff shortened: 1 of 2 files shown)


Frederik-Baetens:
> We don't have a way of conditionally adding them in, do we?

We do, `wrangler types --x-include-runtime` generates types based on your compat flags and bindings. I'd expect that to work out of the box if you're just using an established experimental flag, and if it doesn't, we can definitely extend that mechanism to conditionally generate types.

src/workerd/api/sql.h
```diff
@@ -33,22 +33,25 @@ class SqlStorage final: public jsg::Object, private SqliteDatabase::Regulator {
...
       JSG_METHOD(ingest);
     }
 
     JSG_READONLY_PROTOTYPE_PROPERTY(databaseSize, getDatabaseSize);
 
     JSG_NESTED_TYPE(Cursor);
     JSG_NESTED_TYPE(Statement);
+
+    JSG_TS_OVERRIDE({
+      prepare: never;
```
kentonv, line 51: I don't think we have to explicitly override these to hide them... they will only show up in then first place if someone is generating types with "experimental" enabled, in which case they should actually get those declarations, right?

geelen: Oh my mistake, I was looking at the wrong output directory.

kentonv (approved):

kentonv:
LGTM but CI seems unhappy...

PR #2736, opened 2024-09-18
test clang-format changes

anonrig:
Ref: https://github.com/cloudflare/workerd/pull/2725#issuecomment-2358783982


.clang-format
```diff
@@ -100,7 +100,7 @@ SpaceBeforeCtorInitializerColon: false
 SpaceBeforeInheritanceColon: false
 SpaceBeforeParens: ControlStatementsExceptControlMacros
 SpaceBeforeRangeBasedForLoopColon: false
-SpaceBeforeCpp11BracedList: false
+SpaceBeforeCpp11BracedList: true
```
npaun, line 103: Changing this affects a bunch of other things. I think it would be better to switch to the apt.llvm.org version on your laptop.

PR #2741, opened 2024-09-18
Enable clang-format AllowShortBlocksOnASingleLine

fhanau:
This looks much cleaner and readable IMHO. There is also an "Always" option for this, but its output feels less readable to me.

=============

See https://clang.llvm.org/docs/ClangFormatStyleOptions.html#allowshortblocksonasingleline. I don't remember how I found this but it might have been improved/fixed in LLVM 18/19.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -32,6 +32,7 @@ IncludeCategories:
 
 AllowShortIfStatementsOnASingleLine: true
 AllowShortLoopsOnASingleLine: true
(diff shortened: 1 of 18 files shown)


npaun (approved):
This is quite nice.

Don't forget:
- .git-blame-ignore-revs
-  Merge via merge commit

jasnell (approved):

PR #2742, opened 2024-09-18
introduce swar routines for capitalization

anonrig:
Improved the performance of `toUpper` and `toLower` functionality by introducing SWAR routines.

### DIFF
--- a/src/workerd/api/util.c++
+++ b/src/workerd/api/util.c++
@@ -40,19 +40,61 @@ kj::ArrayPtr<const char> split(kj::ArrayPtr<const char>& text, char c) {
   return result;
 }
 
+constexpr uint64_t broadcast(uint8_t v) noexcept {
+  return 0x101010101010101ull * v;
+}
+
+// SWAR routine designed to convert ASCII uppercase letters to lowercase.
+// Let's process 8 bytes (64 bits) at a time using a single 64-bit int,
+// treating as 8 parallel 8-bit values.
+// PS: This will enable the use of auto-vectorization.
+constexpr void toLowerAscii(char* input, size_t length) noexcept {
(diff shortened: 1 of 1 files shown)


jasnell (approved):
Might be good to migrate these into kj

src/workerd/api/util.c++
```diff
@@ -40,19 +40,61 @@ kj::ArrayPtr<const char> split(kj::ArrayPtr<const char>& text, char c) {
...
+}
+
+// SWAR routine designed to convert ASCII uppercase letters to lowercase.
+// Let's process 8 bytes (64 bits) at a time using a single 64-bit int,
+// treating as 8 parallel 8-bit values.
+// PS: This will enable the use of auto-vectorization.
+constexpr void toLowerAscii(char* input, size_t length) noexcept {
+  uint64_t broadcast_80 = broadcast(0x80);
+  uint64_t broadcast_Ap = broadcast(128 - 'A');
+  uint64_t broadcast_Zp = broadcast(128 - 'Z' - 1);
```
npaun, line 54: I think the compiler can figure this out, but in principle, shouldn't we mark these as `static constexpr`?

fhanau: Note that static or a storage class in general is not needed where there is constexpr; in fact clang-19 warns about `static constexpr` so just constexpr is sufficient.

npaun: @fhanau TIL. Thanks!

anonrig: I changed them to `constexpr const`

vicb:
I guess the reason to repeat the same code twice (excluding the 2 boundaries) is perf?

vicb:
Also the one thing I don't get is how overflow is not a problem?

PR #2743, opened 2024-09-18
Clear DeleteQueue actions under isolate lock

jasnell:


src/workerd/io/io-context.c++
```diff
@@ -1184,6 +1184,17 @@ void IoContext::runFinalizers(Worker::AsyncLock& asyncLock) {
           : context(context),
             warnings(kj::mv(warnings)) {}
       void run(Worker::Lock& lock) override {
+        {
+          // If there are actions in the delete queue, clear them out while we
+          // have the isolate lock. It is critical to do this here since the
+          // actions capture a v8::Global that can only be destroyed while the
```
kentonv, line 1190: I think we'd better change the `v8::Global` to `jsg::V8Ref`, which can handle being destroyed outside of lock. Otherwise I'm not confident that there aren't other ways these might get destroyed without the lock...

jasnell: I originally had a V8Ref here but the reactions is not a v8::Value and that was throwing an assert because the Data wasn't a v8::Value.

kentonv: Use `jsg::Data` for `v8::Data`.

jasnell: heh... you see, this is how you know it's been a long week already. I completely forgot that jsg::Data was a thing lol. Sigh.

PR #2745, opened 2024-09-19
Partially update boringssl, add warnings for invalid parameters

fhanau:
~~Boringssl now has built-in bazel support, so we no longer have to use a commit from main-with-bazel instead of the primary branch. A few code changes are needed based on the RSA struct now being opaque, which is why we deferred updating this so far.~~

Roll boringssl, fix RSA struct usage and refactor DH code
- This is only a partial update for now due to incompatible changes to DH introduced in https://boringssl-review.googlesource.com/c/boringssl/+/62226. Includes some code changes based on the RSA struct now being opaque as well as slight code simplifications for DH.

Warn when creating DH handle with invalid parameters
- https://boringssl-review.googlesource.com/c/boringssl/+/62226 added some additional DH checks for "egregiously large or invalid" parameters. Add a warning when creating a DH handle that would cause boringssl to throw. Some of the additional checks are already covered by our implementation, so we only need this for some checks on the p parameter.

============

This will allows us to see if we can safely update boringssl or if there is code depending on using DH with impractical parameters.


WORKSPACE
```diff
@@ -52,13 +52,22 @@ http_archive(
     urls = ["https://github.com/capnproto/capnproto/tarball/6446b721a9860eebccf9d3c73b27610491359b5a"],
 )
 
+# Needed for boringssl
```
fhanau, line 55: This really shouldn't be necessary, or at least be handled by a `boringssl_build_deps()` function, but boringssl does not support workspace-based builds all that well.

fhanau:
This is failing `//src/workerd/api/node:tests/crypto_dh-test` because of https://boringssl-review.googlesource.com/c/boringssl/+/62226 or https://boringssl-review.googlesource.com/c/boringssl/+/62228, which reject some input parameters. We could still update to c08ccc9ed166a82b92edd70ab215ae1f2501e838, but that's a much smaller update. Moving this back to draft stage.

src/workerd/api/crypto/dh.c++
```diff
@@ -100,6 +103,10 @@ kj::Own<DH> initDh(kj::OneOf<kj::Array<kj::byte>, int>& sizeOrKey,
             }
             return 1;
           };
+          // Operations on an "egregiously large" prime will throw with recent boringssl.
```
jasnell, line 106: I assume later we'd be (hopefully) updating these to throws? If so, a `TODO` comment indicating next steps would be good on these.

src/workerd/api/crypto/dh.c++
```diff
@@ -100,6 +103,10 @@ kj::Own<DH> initDh(kj::OneOf<kj::Array<kj::byte>, int>& sizeOrKey,
...
+          if (size > OPENSSL_DH_MAX_MODULUS_BITS) {
+            KJ_LOG(WARNING, "DiffieHellman init: requested prime size too large");
```
jasnell, line 108: Should these be `NOSENTRY` or do we want them appearing there?

jasnell (approved):

src/workerd/api/crypto/dh.c++
fhanau: Yup, I had a TODO message in a previous version of this. Added one here again.

src/workerd/api/crypto/dh.c++
fhanau: I do prefer having warnings appear on Sentry as it makes them easier to find. These should be temporary so hopefully this is fine

PR #2749, opened 2024-09-19
[build] Roll pnpm dependencies

fhanau:
- Roll pnpm dependencies
  Some updates are not included to avoid breakage (kicking the can down the road), but we make progress on a number of packages.
- Update workerd NPM package deployment target to node18
  Node 16 has been EOL since August 2023.
- Roll esbuild 0.15 : 0.21

=======================

This is a general update PR, I've updated dependencies here before but that was usually focused on specific packages. This does not attempt to update Typescript to 5.6 or update typescript-eslint beyond 8.4.0 (causes lint failures that go over my head).

@penalosa: Any objections to updating the NPM deployment target to node18? Would we need to document this anywhere? I'm also apprehensive about updating esbuild since it's quite outdated by now, it could break the NPM job and we'd only find out when trying to release
@anonrig Let me know if you have concerns too. This is intended to be mostly safe. Since none of the files being updated is used by the downstream build I'll make a similar downstream PR afterwards.

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -117,17 +117,17 @@ jobs:
       - run: mkdir -p npm/workerd/lib
       - run: mkdir -p npm/workerd/bin
       - name: Build node-install
-        run: npx esbuild npm/lib/node-install.ts --outfile=npm/workerd/install.js --bundle --target=node16 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
+        run: npx esbuild npm/lib/node-install.ts --outfile=npm/workerd/install.js --bundle --target=node18 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
         env:
           WORKERD_VERSION: ${{ needs.version.outputs.version }}
           LATEST_COMPATIBILITY_DATE: ${{ needs.version.outputs.date }}
       - name: Build node-shim
-        run: npx esbuild npm/lib/node-shim.ts --outfile=npm/workerd/bin/workerd --bundle --target=node16 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
+        run: npx esbuild npm/lib/node-shim.ts --outfile=npm/workerd/bin/workerd --bundle --target=node18 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
         env:
           WORKERD_VERSION: ${{ needs.version.outputs.version }}
           LATEST_COMPATIBILITY_DATE: ${{ needs.version.outputs.date }}
       - name: Build node-path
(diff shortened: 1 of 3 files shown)


.github/workflows/npm.yml
```diff
@@ -117,17 +117,17 @@ jobs:
       - run: mkdir -p npm/workerd/lib
       - run: mkdir -p npm/workerd/bin
       - name: Build node-install
-        run: npx esbuild npm/lib/node-install.ts --outfile=npm/workerd/install.js --bundle --target=node16 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
+        run: npx esbuild npm/lib/node-install.ts --outfile=npm/workerd/install.js --bundle --target=node18 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
```
anonrig, line 120: Before this step, we should call `pnpm@setup` github workflow and install the dependency.

`npx esbuild` will install the latest version, if it doesn't exist.

 ```suggestion
        run: node_modules/.bin/esbuild npm/lib/node-install.ts --outfile=npm/workerd/install.js --bundle --target=node18 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
```

File: .github/workflows/npm.yml [hunk 2]
```diff
@@ -117,17 +117,17 @@ jobs:
...
         env:
           WORKERD_VERSION: ${{ needs.version.outputs.version }}
           LATEST_COMPATIBILITY_DATE: ${{ needs.version.outputs.date }}
       - name: Build node-shim
-        run: npx esbuild npm/lib/node-shim.ts --outfile=npm/workerd/bin/workerd --bundle --target=node16 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
+        run: npx esbuild npm/lib/node-shim.ts --outfile=npm/workerd/bin/workerd --bundle --target=node18 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
```
anonrig, line 125: What's the reason for not targeting LTS? node18 active support ended, and even node 20(LTS) will end in 1 month. 

Ref: https://endoflife.date/nodejs

.github/workflows/npm.yml
anonrig, line 120: PS: We are also using `pnpm` but running eslint using `npx`.

.github/workflows/npm.yml
fhanau: My thinking was that we should target the oldest version developers would reasonably be using, I'm not sure how quickly folks update to the latest version in the Node.js ecosystem but as long as 18 is being supported we should target since it (presumably) costs us nothing for this small application

.github/workflows/npm.yml
fhanau: Maybe this is best addressed in another PR aimed at improving the NPM CI job – bumping the node version here makes sense within the scope of updating npm packages, but I've never looked at how we build these files in detail

.github/workflows/npm.yml
anonrig: It's hard to distinguish when a technology is not used. Are you thinking about EOL change to drop node18?

.github/workflows/npm.yml
anonrig: Sounds good to me. Thank you.

anonrig (approved):

.github/workflows/npm.yml
fhanau: Actually looked at the generated output – it's identical across node versions which is unsurprising since there's not a whole lot of code to begin with. The esbuild version only has cosmetic differences between 0.15 and 0.23, so it is inconsequential too. I think this can go ahead as-is, the node18 label is least likely to cause confusion IMHO.

PR #2752, opened 2024-09-19
enable v8_enable_local_off_stack_check compile time flag

anonrig:
Referencing the original commit introducing the flag:

> This CL introduces a compile-time flag v8_enable_local_off_stack_check
> which enforces a run-time DCHECK, that all v8::Local<T> objects are
> indeed stack-allocated. The check is disabled by default. It will
> fail for all heap data structures containing local handles.

Local handles (object of the v8::Local) should never be allocated on the heap, and we should use `v8::LocalVector` instead of heap allocation. This flag ensures that we follow v8 recommendation.


jasnell (approved):

.bazelrc
```diff
@@ -61,6 +61,9 @@ build --host_per_file_copt='external/zlib@-Wno-unknown-warning-option,-Wnodeprec
 build --per_file_copt=external/com_google_tcmalloc@-DANNOTATE_MEMORY_IS_INITIALIZED=ABSL_ANNOTATE_MEMORY_IS_INITIALIZED
 build --per_file_copt=external/com_google_protobuf@-Wno-deprecated-declarations,-Wno-deprecated-pragma
 build --host_per_file_copt=external/com_google_protobuf@-Wno-deprecated-declarations,-Wno-deprecated-pragma
+# Enables heap allocation debug checks for local handles.
+# Ref: https://chromium-review.googlesource.com/c/v8/v8/+/4905902
+build --per_file_copt='external/v8@-DV8_ENABLE_LOCAL_OFF_STACK_CHECK'
```
fhanau, line 66: Actually I was wrong about this only being needed when building V8: Since the define present in include/v8-local-handle.h and include/v8-fast-api-calls.h we also need to apply it when building our own code. Let's change this to a regular `copt`.

fhanau:
Nit: We reproduce V8 build flags in `compile_flags.txt` so that they are available to clangd when editing code – let's add the define there too.

anonrig:
I've submitted a patch to v8 to fix this error. https://chromium-review.googlesource.com/c/v8/v8/+/5876953

jasnell:
Should be noted that the v8 patch is temporary. A future v8 update will include it and the patch will be dropped.

jasnell (approved):

fhanau (approved):
Code LGTM, this can probably go in after the update to 13.0. I wonder if there's any performance impact with this. It looks like Local<> stops being trivially copyable and the Local constructor calls StackAllocated<true>::VerifyOnStack() which is mostly a DCHECK – I'd imagine that's an acceptable impact.
I could also see libkj's fibers performing arcane magic on the stack that cause the check to fail when it shouldn't, but let's try this out.

kentonv:
> I could also see libkj's fibers performing arcane magic on the stack that cause the check to fail when it shouldn't, but let's try this out.

We don't call any V8 APIs on fibers so this shouldn't be an issue.

jasnell:
Optional Side Quest: we should see about a clang-tidy rule to catch uses of things like `std::vector<v8::Local<T>>`, I know this compile time flag is supposed to help check that but a clang-tidy rule might be a bit faster to help flag issues in development.

PR #2753, opened 2024-09-19
SQL: Add next(), toArray(), and one() methods to cursor.

kentonv:
(I don't actually know how to run the type test so I hope it just works.)

### DIFF
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -61,6 +61,64 @@ async function test(state) {
     assert.equal(result[2]['value'], 3);
   }
 
+  {
+    // Test multiple rows, manual iteration with next().
+    const cursor = sql.exec(
+      'SELECT 1 AS col\n' +
+        'UNION ALL\n' +
+        'SELECT "foo" AS col\n' +
+        'UNION ALL\n' +
+        'SELECT 3 AS col;'
+    );
+    assert.deepEqual(cursor.next(), { done: false, value: { col: 1 } });
+    assert.deepEqual(cursor.next(), { done: false, value: { col: 'foo' } });
+    assert.deepEqual(cursor.next(), { done: false, value: { col: 3 } });
+    assert.deepEqual(cursor.next(), { done: true });
+  }
+
+  {
+    // Test multiple rows using .toArray()
+    const cursor = sql.exec(
+      'SELECT 1 AS value\n' +
+        'UNION ALL\n' +
+        'SELECT "foo" AS value\n' +
+        'UNION ALL\n' +
+        'SELECT 3 AS value;'
+    );
+    assert.deepEqual(cursor.toArray(), [
+      { value: 1 },
+      { value: 'foo' },
+      { value: 3 },
+    ]);
+  }
+
+  {
+    // Test one row with .one()
+    let cursor = sql.exec('SELECT 123 AS foo, "abc" AS bar');
+    assert.deepEqual(cursor.one(), { foo: 123, bar: 'abc' });
+    // Cursor has been consumed.
+    assert.deepEqual([...cursor], []);
+
+    // Multiple results throws.
+    assert.throws(
+      () =>
+        sql
+          .exec(
+            'SELECT 1 AS value\n' + 'UNION ALL\n' + 'SELECT "foo" AS value;'
+          )
+          .one(),
+      /Expected exactly one result from SQL query, but got multiple results/
+    );
+
+    // No results throws.
+    sql.exec('CREATE TABLE IF NOT EXISTS empty (x INTEGER)');
+    assert.throws(
+      () => sql.exec('SELECT * from empty;').one(),
+      /Expected exactly one result from SQL query, but got no results/
+    );
+    sql.exec('DROP TABLE empty');
+  }
+
   // Test partial query ingestion
   assert.deepEqual(sql.ingest(`SELECT 123; SELECT 456;    `).remainder, '    ');
   assert.deepEqual(sql.ingest(`SELECT 123; SELECT 456;`).remainder, '');
--- a/src/workerd/api/sql.c++
+++ b/src/workerd/api/sql.c++
@@ -125,7 +125,7 @@ jsg::Ref<SqlStorage::Cursor::RowIterator> SqlStorage::Cursor::rows(jsg::Lock& js
 kj::Maybe<SqlStorage::Cursor::RowDict> SqlStorage::Cursor::rowIteratorNext(
     jsg::Lock& js, jsg::Ref<Cursor>& obj) {
   auto names = obj->cachedColumnNames.get();
-  return iteratorImpl(js, obj, [&](State& state, uint i, Value&& value) {
+  return iteratorImpl(js, obj, [&](State& state, uint i, SqlValue&& value) {
     return RowDict::Field{
       // A little trick here: We know there are no HandleScopes on the stack between JSG and here,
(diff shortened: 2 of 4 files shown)


src/workerd/api/sql.h
```diff
@@ -107,6 +107,19 @@ class SqlStorage final: public jsg::Object, private SqliteDatabase::Regulator {
...
+  // JSG do it. We can't use jsg::TypeHandler because SqlValue contains StringPtr, which doesn't
+  // support unwrapping. We don't actually ever use unwrapping, but requesting a TypeHandler forces
+  // JSG to try to generate the code for unwrapping, leading to compiler errors.
+  //
+  // TODO(cleanup): Think hard about how to make JSG support this better. Part of the problem is
+  //   that we're being too clever with optimizations to avoid copying strings when we don't need
+  //   to.
+  static jsg::JsValue wrapSqlValue(jsg::Lock& js, SqlValue value);
+  static jsg::JsObject wrapSqlRow(jsg::Lock& js, SqlRow row);
+  static jsg::JsArray wrapSqlRowRaw(jsg::Lock& js, kj::Array<SqlValue> row);
```
justin-mp, line 122: I don't think this function is used anywhere.

src/workerd/api/sql-test.js
```diff
@@ -61,6 +61,64 @@ async function test(state) {
...
+
+  {
+    // Test one row with .one()
+    let cursor = sql.exec('SELECT 123 AS foo, "abc" AS bar');
+    assert.deepEqual(cursor.one(), { foo: 123, bar: 'abc' });
+    // Cursor has been consumed.
+    assert.deepEqual([...cursor], []);
+
+    // Multiple results throws.
+    assert.throws(
```
justin-mp, line 103: Given that exec.one() does special things with rowsRead/rowsWritten, should we test the values here?

justin-mp (commented):
nit: I think you want the commit description destination of the rename of the first commit to be SqlStorage::SqlValue (storage and colons).

src/workerd/api/sql.h
kentonv: It isn't now, but likely to be in the future when we figure out `raw().toArray()` vs. `toRawArray()` or whatever.

justin-mp: I'd probably remove it, but I know we're on a schedule here.

justin-mp (approved):

src/workerd/api/sql-test.js
kentonv: I'll need to significantly revise the test since currently it doesn't actually read from tables so the counters are always zero... will do that in a later PR if you don't mind.

justin-mp: Fine with me.

src/workerd/api/sql.c++
```diff
@@ -115,24 +154,85 @@ double SqlStorage::Cursor::getRowsWritten() {
...
+  KJ_IF_SOME(s, state) {
+    cachedColumnNames.ensureInitialized(js, s->query);
+  }
+
+  auto self = JSG_THIS;
+  auto maybeRow = rowIteratorNext(js, self);
+  bool done = maybeRow == kj::none;
+  return {
+    .done = done,
+    .value = kj::mv(maybeRow),
```
geelen, line 167: I know why you've done this (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol) but it doesn't actually make this object iterable unless we make `[Symbol.iterator]()` return `this` (which we could do?)

Maybe that's exactly what we want, and then `.raw()` returns a copy of the Cursor but with a different implementation of `next()` that returns arrays. That makes things as JS-standardy as I think we can get?

geelen: I actually came here to write that we should make `.next()` return `T | null` and then also add a `.done` property:

```js
const row = sql.exec('...').next()
// instead of
const { value: row } = sql.exec('...').next()
```

And `.done` means could check for 0 results without actually calling `.next()`:

```js
const cursor = sql.exec('...')
if (!cursor.done) throw new Error("Wasn't expecting results!!")
```

But both of those seem like marginal benefits, at the cost of diverging from the Iterator standard. Anyway, thought I'd note them here for completeness.

src/workerd/api/sql.c++
```diff
@@ -115,24 +154,85 @@ double SqlStorage::Cursor::getRowsWritten() {
...
+  auto& query = KJ_ASSERT_NONNULL(state)->query;
+  query.nextRow();
+  bool hadOneResult = query.isDone();
+
+  // Save off row counts before the query goes away, just like iteratorImpl() would do when done.
+  rowsRead = query.getRowsRead();
+  rowsWritten = query.getRowsWritten();
+
+  // End the query even if it wasn't done.
+  state = kj::none;
```
geelen, line 210: I really like that the query gets definitely closed as part of calling `.one()`. Very tidy.

src/workerd/api/sql.c++
geelen: @justin-mp you mentioned being a bit confused re: Iterators and Iterables, and I admit I needed to refresh my memory, so I just wrote out a couple of examples to confirm the way things work re: `Symbol.iterator` and `next()`:

```ts
class CustomNext {
  private iterator: Iterator<any>
  public done: boolean = false

  constructor(cursor: SqlStorageCursor) {
    this.iterator = cursor[Symbol.iterator]()
  }

  // Can return anything we want, will never be called
  // by a for-of loop or an array splat because we don't
  // implement [Symbol.iterator]()
  next() {
    const { value, done } = this.iterator.next()
    this.done = Boolean(done)
    return value
  }
}

class CustomIterator {
  private iterator: Iterator<any>

  constructor(cursor: SqlStorageCursor) {
    this.iterator = cursor[Symbol.iterator]()
  }

  // This is what is called when one of these objects is
  // passed to a for-of loop or [...splat] syntax. We must
  // return an object that implements next() => { value, done }
  [Symbol.iterator]() {
    return this
  }

  // Must follow the Iterator protocol
  next() {
    const { value, done } = this.iterator.next()
    return { value, done }
  }
}
```

Anyway, I am now fully convinced we should follow the `CustomIterator` example above (follow the spec) vs the `CustomNext` example (yolo)

geelen (commented):
Code looks fine, I'll fix the TS errors and push

src/workerd/api/sql.c++
```diff
@@ -62,6 +62,45 @@ bool SqlStorage::allowTransactions() const {
...
+    for (auto& field: row.fields) {
+      result.set(js, field.name, wrapSqlValue(js, kj::mv(field.value)));
+    }
+    return result;
+  }));
+}
+
+jsg::JsArray SqlStorage::wrapSqlRowRaw(jsg::Lock& js, kj::Array<SqlValue> row) {
+  return jsg::JsArray(js.withinHandleScope([&]() -> v8::Local<v8::Array> {
+    v8::LocalVector<v8::Value> values(js.v8Isolate);
```
anonrig, line 96:
```suggestion
    v8::LocalVector<v8::Value> values(js.v8Isolate);
    values.reserve(row.size());
```

kentonv: Good optimization but I have to merge this tonight and I need to sign off shortly so as soon as CI is green I'm going to merge, will do this in some future PR.

anonrig (approved):

Issue #2652, opened 2024-09-04
Minor regression in types params names with `4.20240903.0`


Cherry:
Diff: https://npmdiff.dev/@cloudflare%2fworkers-types/4.20240821.1/4.20240903.0/package/experimental/index.ts
Raw diff: https://github.com/Cloudflare-Mining/Cloudflare-Datamining/commit/66e958488df276e1d052cdb93e13b365e20fbd22#diff-498f7d22754cd92e5654b3be0adcae3df0c297abe440099d42dbc8cf43af66fd

`FixedLengthStream` previously had named paramaters:

```ts
declare class FixedLengthStream extends IdentityTransformStream {
  constructor(
    expectedLength: number | bigint,
    queuingStrategy?: IdentityTransformStreamQueuingStrategy,
  );
}
```

Now these are more generic with just `param1` and `param2`:
```ts
declare class FixedLengthStream extends IdentityTransformStream {
  constructor(
    param1: number | bigint,
    param2?: IdentityTransformStreamQueuingStrategy,
  );
}
```

A few more methods/classes had the same thing occur including:

- `URL.canParse`
- `URL.parse`
- `IdentityTransformStreamQueuingStrategy`
- `URLSearchParams`
- `URLSearchParams.append`
- `URLSearchParams.delete`
- `URLSearchParams.get`
- `URLSearchParams.getAll`
- `URLSearchParams.has`
- `URLSearchParams.set`

penalosa:
@jasnell @anonrig have the implementations for streams or URLs moved? We generate parameter names by parsing the c++ AST so it's relatively sensitive to refactoring

anonrig:
The last change seems to be `clang-format` https://github.com/cloudflare/workerd/commits/main/src/workerd/jsg/url.c%2B%2B

jasnell:
No changes that I'm aware of that would cause this. Really odd. Could the formatting changes have broken the ast transform?

jasnell:
@penalosa ... I think we should first rule out a regression in the type generation logic. If that points to a change on the runtime side, then we'll investigate further but there really hasn't been much here that could have changed as far as I know.

Cherry:
It looks like this was partially resolved in the latest update:

https://npmdiff.dev/@cloudflare%2fworkers-types/4.20240909.0/4.20240919.0/package/experimental/index.ts

`FixedLengthStream` has the correct param names again, as well as `IdentityTransformStream`, but not any of the others yet.

PR #2754, opened 2024-09-20
Fix type params

penalosa:
Fixes https://github.com/cloudflare/workerd/issues/2652

A regression introduced in https://github.com/cloudflare/workerd/commit/fc7f91d59e114f697784e9a2caf9b0e3e93f46c2 meant that parameter names for some APIs were no longer being included in the generated types, since `param-names-ast.c++` didn't import them. This PR updates the includes within `param-names-ast.c++` to match `rtti.c++`.

### DIFF
--- a/src/workerd/tools/param-names-ast.c++
+++ b/src/workerd/tools/param-names-ast.c++
@@ -4,6 +4,7 @@
 
 // This file includes all Worker APIs, and is used to generate a Clang AST dump for later
 // lookup of parameter names for inclusion in the TS types (since RTTI doesn't include this information)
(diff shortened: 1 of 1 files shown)


mikea:
If they need to be in sync, then maybe we should create a separate `.h` file and include it from both places so that there are no discrepancies?

jasnell (approved):

RamIdeas:
> If they need to be in sync, then maybe we should create a separate `.h` file and include it from both places so that there are no discrepancies?

This sounds like a good idea to prevent this happening again

Issue #2470, opened 2024-08-01
`AiTextToImageOutput` should be a ReadableStream instead of `Uint8Array`


Kynson:
Hello,

I am using `env.AI.run('@cf/stabilityai/stable-diffusion-xl-base-1.0', input)` to generate image. The return type of this function is `AiTextToImageOutput` which is a type alias of Uint8Array. Yet, `env.AI.run()` seems to return a `ReadableStream` instead of `Uint8Array`

## Minimal reproducible example
```javascript
const result = await env.AI.run(
    "@cf/stabilityai/stable-diffusion-xl-base-1.0",
   {
      prompt:
        "An artistic cartoon of a cat playing with a new technology gadget in a sunny computer lab, surrounded by colorful flowers and Server guarantees serene background",
      num_steps: 20,
      guidance: 7,
   }
  );
// logs:
// ReadableStream {
//  locked: false,
//  [state]: 'readable',
//  [supportsBYOB]: true,
//  [length]: undefined
// }
console.log(result);
```

## Tested with 
wrangler version: 3.67.1
@cloudflare/workers-types version: 4.20240725.0
node version: 22.5.0

PR #2757, opened 2024-09-20
Fix AiTextToImageOutput type from UInt8Array -> ReadableStream

dochne:
As noted in https://github.com/cloudflare/workerd/issues/2470 - the response type from `env.AI.run` with (in my case) `@cf/bytedance/stable-diffusion-xl-lightning`  claims it should return a UInt8Array, but actually returns a ReadableStream.

This is fine when feeding it into a `new Response()` and returning, but is problematic when trying to use the data in other ways.

You can replicate this with the example model usage [here](https://developers.cloudflare.com/workers-ai/models/stable-diffusion-xl-lightning/)

```
[ai]
binding = "AI"
```

```ts
export interface Env {
  AI: Ai;
}

export default {
  const inputs = {prompt: "cyberpunk cat"};

  const response = await env.AI.run(
    "@cf/bytedance/stable-diffusion-xl-lightning",
    inputs
  );

  // Expected: Uint8Array. Actual: ReadableStream
  console.log(response);
  return new Response("Don't error");
}
```

### DIFF
--- a/types/defines/ai.d.ts
+++ b/types/defines/ai.d.ts
@@ -156,7 +156,7 @@ export type AiTextToImageInput = {
   strength?: number;
   guidance?: number;
 };
-export type AiTextToImageOutput = Uint8Array;
+export type AiTextToImageOutput = ReadableStream<Uint8Array>;
 export declare abstract class BaseAiTextToImage {
   inputs: AiTextToImageInput;
   postProcessedOutputs: AiTextToImageOutput;


dochne:

types/defines/ai.d.ts
```diff
@@ -156,7 +156,7 @@ export type AiTextToImageInput = {
   strength?: number;
   guidance?: number;
 };
-export type AiTextToImageOutput = Uint8Array;
+export type AiTextToImageOutput = ReadableStream;
```
dochne, line 159: Let me know if this is actually defined somewhere else and I'm just modifying an artifact.

dario-piotrowicz: I wonder if this should be `ReadableStream<Uint8Array>` 🤔 

That's what we have for images for example: https://github.com/cloudflare/workerd/blob/93953aedc17e13deda0212a86aad71711140343b/types/defines/images.d.ts#L120
So I am guessing that this one should probably be the same too 🤔 

(the AI team should know best anyways)

dochne: Updated to add in the UInt8Array!

dario-piotrowicz: @dochne Awesome! thank you very much! 🫶

G4brym (approved):

penalosa (approved):

dario-piotrowicz:
Thanks a bunch again for the fix @dochne 🫶

PR #2759, opened 2024-09-20
moving rust dependencies to deps/rust and structuring like downstream

mikea:


deps/rust/BUILD.bazel
```diff
@@ -70,31 +73,44 @@ crates_vendor(
     ],
 )
 
-# TODO(cleanup): Switching to rust_library() would reduce the overhead of this by not creating a
-# crate-of-crates but this previously broke the Windows build.
+# define the path to the lolhtml headers, this is clunky but allows us to avoid cloning the
+# repository twice.
+alias(
```
fhanau, line 78: This alias is no longer being used based on `deps/rust/BUILD.lolhtml` – I think the approach we had so far made it easy to find and update the lolhtml version number if needed, but I'm not attached to it so leaving BUILD.lolhtml and deleting this alias is an option too

deps/rust/BUILD.bazel
```diff
@@ -70,31 +73,44 @@ crates_vendor(
...
+    name = "lol_html_api",
+    actual = "@crates_vendor__lol_html-1.2.1//:c-api/include/lol_html.h",
+    tags = ["manual"],
+    visibility = ["//visibility:public"],
+)
+
+# windows build fixes due to rust libraries missing symbols:
+# tps://github.com/bazelbuild/rules_rust/blob/144d34fd/ffi/cc/global_allocator_library/BUILD.bazel#L3
```
fhanau, line 86: Fix the url, and maybe add a comment saying that we use empty_lib to have a Rust "staticlib" target as this allows us to the missing symbols on Windows (otherwise having emptylib could be confusing)

src/workerd/server/BUILD.bazel
```diff
@@ -117,6 +117,7 @@ wd_cc_library(
         ":actor-id-impl",
         ":alarm-scheduler",
         ":workerd_capnp",
+        "//deps/rust:runtime",
```
fhanau, line 120: One wrinkle is that every C++ binary that depends on Rust will have to depend on this target. Currently we could just depend on `@workerd//deps/rust:runtime` in `deps/rust/BUILD.lolhtml`, but eventually we'd have to do this for every Rust target which is not clean either. I guess we can do this as-is for now, in the long term we'll have to think something up to avoid confusion

fhanau (approved):

fhanau:
Nit: The WORKSPACE comment "To update the dependency, update the reference commit in rust-deps/BUILD.bazel and run \`bazel run //rust-deps:crates_vendor -- --repin\`" needs to be updated now that we need repin full.

src/workerd/server/BUILD.bazel
mikea: I don't think this is a big issue: downstream this library is a no-op, so we have to sprinkle it only in workerd to make things work on win. I'm fine with it.

mikea:
> Nit: The WORKSPACE comment "To update the dependency, update the reference commit in rust-deps/BUILD.bazel and run `bazel run //rust-deps:crates_vendor -- --repin`" needs to be updated now that we need repin full.

done.

PR #2761, opened 2024-09-20
Add some debug logging to track down "called null capability" exception.

kentonv:


kentonv:
nm I think I found it

PR #2776, opened 2024-09-23
Revert "improve encodeBase64Url performance (#2774)"

anonrig:
This reverts commit b3d98e0049a364818ef8428153a8fb2436f99cc3.

I'll follow up on a different PR to fix edgeworker related tests.

### DIFF
--- a/src/workerd/api/util.c++
+++ b/src/workerd/api/util.c++
@@ -4,8 +4,6 @@
 
 #include "util.h"
 
-#include "simdutf.h"
-
 #include <workerd/io/io-context.h>
 #include <workerd/util/mimetype.h>
 #include <workerd/util/thread-scopes.h>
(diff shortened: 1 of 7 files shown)


fhanau (approved):

jasnell (approved):

src/workerd/api/util.c++
```diff
@@ -290,14 +288,4 @@ void maybeWarnIfNotText(jsg::Lock& js, kj::StringPtr str) {
...
           "checking the Content-Type header before interpreting entities as text."));
 }
-
-kj::String fastEncodeBase64Url(kj::ArrayPtr<const byte> bytes) {
-  auto expected_length = simdutf::base64_length_from_binary(bytes.size(), simdutf::base64_url);
-  auto output = kj::heapArray<char>(expected_length);
-  auto actual_length = simdutf::binary_to_base64(
-      bytes.asChars().begin(), bytes.size(), output.asChars().begin(), simdutf::base64_url);
-  KJ_ASSERT(expected_length == actual_length);
-  return kj::String(kj::mv(output));
```
jclee, line 300: Seems likely this might be the issue?  I think `kj::String` has an internal assert to require that any string pointer content be null-terminated.  Might be able to fix it by allocating a slightly bigger array and adding the terminator before passing to `kj::String`.

jclee (approved):

src/workerd/api/util.c++
anonrig: You're absolutely right. Yes, I'll follow up in a new pull-request to fix this. Let's fix edgeworker first!

jclee: Thanks!

PR #2777, opened 2024-09-23
Install buildifier via Bazel

npaun:


devtools/BUILD.bazel
```diff
@@ -0,0 +1,43 @@
+load("@bazel_skylib//lib:selects.bzl", "selects")
+
+selects.config_setting_group(
```
fhanau, line 3: We have similar config setting groups defined in `rust-deps/BUILD.bazel` for the cargo_bazel binary. We should define these in the build folder so that we only need one group of them. For completion, we could also add darwin-amd64 since we have it there (or not, if we assume that no one working on this code base is on Intel Mac).

WORKSPACE
```diff
@@ -597,3 +597,32 @@ new_local_repository(
         visibility = ["//visibility:public"],)""",
     path = "empty",
 )
+
+# Dev tools
+http_file(
```
fhanau, line 602: This will not be fun to maintain manually (if you look at `cargo_bazel_linux_x64` and other variants I've been lazy with upgrading ones other than my platform). I think this can go ahead as-is though, #2775 will eventually make this easier.

tools/cross/format.py
```diff
@@ -117,7 +117,12 @@ def prettier(files: list[Path], check: bool = False) -> bool:
 
 
 def buildifier(files: list[Path], check: bool = False) -> bool:
-    cmd = [BUILDIFIER, "--mode=check" if check else "--mode=fix"]
```
danlapid, line 120: Bazel is annoyingly slow for simple tasks
I would personally prefer to use bazel to vendor the binary into a const path and then just invoke it directly without bazel.

npaun:
Based on @danlapid's suggestion, I added some code to let `format.py` run the tool directly and avoid invoking bazel each time.

danlapid (approved):
Love it!

fhanau:
What I meant with my comment re. config_setting_group was to still have them to define an alias for the binary with the right architecture, but to define the groups in a new file in the build folder and use that to fold up the groups that we have so far in rust-deps/BUILD.bazel. That feels cleaner than doing platform selection in python, but I assume that would cause issues with running the binary directly instead of through bazel, so not opposed to this approach.

npaun:
@fhanau Yep, your suggestion makes sense, but I wanted direct running of the binary to speed things up where possible.

tools/cross/format.py
```diff
@@ -189,18 +217,18 @@ class FormatConfig:
     # FormatConfig(
     #     directory="src/workerd", globs=("*.c++", "*.h"), formatter=clang_format
     # ),
-    # FormatConfig(
-    #     directory="src",
-    #     globs=("*.js", "*.ts", "*.cjs", "*.ejs", "*.mjs"),
-    #     formatter=prettier,
-    # ),
-    # FormatConfig(directory="src", globs=("*.json",), formatter=prettier),
+    FormatConfig(
```
npaun, line 220: Restoring prettier because it is already installed by bazel

tools/cross/format.py
```diff
@@ -189,18 +217,18 @@ class FormatConfig:
...
+        formatter=prettier,
+    ),
+    FormatConfig(directory="src", globs=("*.json",), formatter=prettier),
     # FormatConfig(directory=".", globs=("*.py",), formatter=ruff),
-    # FormatConfig(
-    #     directory=".",
-    #     globs=("*.bzl", "WORKSPACE", "BUILD", "BUILD.*"),
-    #     formatter=buildifier,
-    # ),
+    FormatConfig(
```
npaun, line 227: Restoring buildifier because this PR installs it via Bazel

tools/cross/format.py
mikea, line 220: was prettier intended to be included? Imo it requires running `npm install` or something and is not fetched automatically?

mikea (approved):
Thanks for doing this. I have #2775 in the works that will hopefully make managing these easier.

tools/cross/format.py
npaun: It's listed as a runtime dependency of workerd, so it's somewhere... I'm currently doing a clean build to find out where it puts it.

mikea:
> What I meant with my comment re. config_setting_group was to still have them to define an alias for the binary with the right architecture, but to define the groups in a new file in the build folder and use that to fold up the groups that we have so far in rust-deps/BUILD.bazel. That feels cleaner than doing platform selection in python, but I assume that would cause issues with running the binary directly instead of through bazel, so not opposed to this approach.

I thought about this too, and while it can be done I don't see a particular value of doing it that way. You'll have to do `bazel run` then, which has its own drawbacks. This change is simple enough so lg.

tools/cross/format.py
npaun: Found it (bazel-bin/node_modules/prettier/bin/prettier.cjs)

PR #2778, opened 2024-09-23
Handle errors thrown by reportError in queueMicrotask

jasnell:


src/workerd/api/global-scope.c++
```diff
@@ -706,7 +706,20 @@ void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, v8::Local<v8::Funct
...
             }, [&](jsg::Value exception) {
-              reportError(js, jsg::JsValue(exception.getHandle(js)));
+              // The reportError call itself can potentially throw errors. Let's catch
+              // and report them as well.
+              js.tryCatch([&] { reportError(js, jsg::JsValue(exception.getHandle(js))); },
+                  [&](jsg::Value exception) {
+                if (IoContext::hasCurrent()) {
+                // If we're running in a request context then we want to abort the
+                // request with the exception here.
+                IoContext::current().abort(js.exceptionToKj(kj::mv(exception)));
```
kentonv, line 716: Is there some background on why this is considered the right thing to do?

anonrig (commented):
Can you add a test case for this?

src/workerd/api/global-scope.c++
jasnell: If an error is thrown in the error event handler it was getting printed but otherwise would continue as if nothing happened. For Node.js compat, however, which uses queueMicrotask as the underlying impl of process.nextTick we need roughly the equivalent of a process crash, the closest we have and would want to get is erroring the request. Otherwise it just becomes a foot gun. @anonrig discovered the inconsistency while trying to port node.js tests for zlib over.

src/workerd/api/tests/error-in-error-event-test.js
```diff
@@ -0,0 +1,30 @@
+import { strictEqual } from 'assert';
+
+export const test = {
+  async test(_, env) {
+    try {
+      await env.subrequest.fetch('http://example.org');
```
anonrig, line 6:
```suggestion
      await env.subrequest.fetch('http://example.org');
      throw new Error('should have thrown');
```

anonrig (approved):

anonrig (approved):

src/workerd/api/global-scope.c++
kentonv: I forget, is reportError() a node compat API or does it exist in browsers? In browsers, what happens if it throws?

I do agree that node APIs should emulate node behavior and the best way to emulate a process crash would be to abort the I/O context. But I don't agree that "crash on error" is _inherently_ the right thing to do.

jasnell: `reportError(...)` is a standard api.

image

kentonv: So in the browser, the event handler throwing doesn't "abort" anything. It feels wrong, then, for this event handler to work differently on Workers.

Can we limit this aborting behavior to happen specifically when the microtask was queued using `process.nextTick`?

Or maybe apps should just register a callback themselves that calls `ctx.abort()` if that's what they want?

src/workerd/api/global-scope.c++
```diff
@@ -706,7 +706,20 @@ void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, v8::Local<v8::Funct
               return jsg::check(
                   function->Call(js.v8Context(), js.v8Null(), argv.size(), argv.data()));
             }, [&](jsg::Value exception) {
-              reportError(js, jsg::JsValue(exception.getHandle(js)));
+              // The reportError call itself can potentially throw errors. Let's catch
+              // and report them as well.
+              js.tryCatch([&] { reportError(js, jsg::JsValue(exception.getHandle(js))); },
+                  [&](jsg::Value exception) {
+                if (IoContext::hasCurrent()) {
+                // If we're running in a request context then we want to abort the
```
kentonv, line 714: BTW what's with the indentation here? Seems all wrong but the formatter isn't complaining?

src/workerd/api/global-scope.c++
kentonv: The thing that worries me here is that Node in general has a policy of "abort on uncaught errors", but browsers do not, and Workers have always behaved more like browsers. It's weird to have one single place where we actually do abort on uncaught exceptions suddenly. This doesn't feel consistent.

I'm all for apps being able to opt into it if they want it but the defaults should be consistent I think?

jasnell: I generally disagree but I can update this to only about if the nodejs-compat flag is enabled. That way it is explicitly opt in.

src/workerd/api/global-scope.c++
anonrig: cc @danlapid @npaun any ideas?

danlapid: Clang format was disabled by https://github.com/cloudflare/workerd/pull/2785
Until https://github.com/cloudflare/workerd/pull/2791 gets merged.

kentonv: Doesn't CI still enforce?

src/workerd/api/global-scope.c++
kentonv: But that's still inconsistent? Doesn't node also abort on unhandled rejections and such? Also I feel like node-compat should just make node APIs available, but not change the behavior of non-node APIs.

I could imagine a _separate_ compat flag which opts into "eagerly abort on errors", which would apply both here and to unhandled rejections.

src/workerd/api/global-scope.c++
danlapid: Nope, sadly the above mentioned PR just disabled it completely both for local formatting and for CI

kentonv: OK, but still, the formatter was disabled _after_ this PR was uploaded... so this PR did pass the formatter before it was disabled, seemingly.

jasnell: Yeah, the formatting here was picked by the formatter, not by anything I did. I updated and checked in with `--no-verify` to ensure good formatting on the updated commit.

kentonv:
> he PR is updated to simply log the error rather than abort anything.

That sounds good to me. But I don't see the update yet.

jasnell:
> ... But I don't see the update yet.

Grrr... I hadn't noticed that the precommit hook had blocked it. There now.

jasnell:
Linter is complaining about the better formatting so will go with the weird formatting for now.

src/workerd/api/global-scope.c++
```diff
@@ -706,7 +706,28 @@ void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, v8::Local<v8::Funct
...
+                KJ_IF_SOME(obj, val.tryCast<jsg::JsObject>()) {
+                auto stack = obj.get(js, "stack"_kj);
+                if (!stack.isUndefined()) {
+                js.reportError(stack);
+                return;
+                }
+                } else {
+                }  // Here to avoid a compile warning
+                // Otherwise just log the stringified value generically.
+                js.reportError(val);
```
kentonv, line 729: Nitpick: Would be nicne to add some doc comments to jsg::Lock::reportError(), it's currently quite hard to figure out what exactly it does.

src/workerd/api/global-scope.c++
```diff
@@ -706,7 +706,28 @@ void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, v8::Local<v8::Funct
...
+                // Let's log the error and just continue. It won't be possible to actually
+                // catch or handle this error so logging is really the only way to notify
+                // folks about it.
+                auto val = jsg::JsValue(exception.getHandle(js));
+                // If the value is an object that has a stack propery, log that so we get
+                // the stack trace if it is an exception.
+                KJ_IF_SOME(obj, val.tryCast<jsg::JsObject>()) {
+                auto stack = obj.get(js, "stack"_kj);
+                if (!stack.isUndefined()) {
+                js.reportError(stack);
```
kentonv, line 723: ugh this indentation is impossible to read...

kentonv (approved):

src/workerd/api/global-scope.c++
jasnell: Yep. I tried to avoid it but the linter was complaining. Once that is fixed we can fix this up.

PR #2779, opened 2024-09-24
Add lock wait time metrics for Memory Cache.

mar-cf:


src/workerd/api/memory-cache.c++
```diff
@@ -239,14 +239,26 @@ SharedMemoryCache::Use::~Use() noexcept(false) {
 }
 
 kj::Maybe<kj::Own<CacheValue>> SharedMemoryCache::Use::getWithoutFallback(
-    const kj::String& key) const {
-  auto data = cache->data.lockExclusive();
+    const kj::String& key, SpanBuilder& span) const {
+  kj::Locked<ThreadUnsafeData> data;
+  {
```
mikea, line 244: I am not super fond of using lambdas for variable initialization, but this case will benefit from it imo:

```
data = [&span]() { ... }();
```

src/workerd/api/memory-cache.c++
```diff
@@ -239,14 +239,26 @@ SharedMemoryCache::Use::~Use() noexcept(false) {
...
+    auto memoryCacheLockTiming =
```
mikea, line 245: why do we need this timing virtual object at all? Why can't this MemoryCacheLockRecord be defined here in this file, and do real work?

mar-cf: I think we must have been trying to avoid including timing metrics in workerd.

I removed it and make the real measurement here.

src/workerd/io/observer.h
```diff
@@ -30,6 +32,30 @@ class WebSocketObserver: public kj::Refcounted {
   virtual void receivedMessage(size_t bytes) {};
 };
 
+class MemoryCacheLockRecord {
+public:
+  explicit MemoryCacheLockRecord(SpanBuilder& span, const kj::MonotonicClock& timer)
+      : span(span),
+        timer(timer),
+        startTime(timer.now()) {}
```
mar-cf, line 40: We could just use the kj::systemPreciseMonotonicClock().now() if we don't mind making a timing a measurement here.

src/workerd/io/observer.h
```diff
@@ -30,6 +32,30 @@ class WebSocketObserver: public kj::Refcounted {
   virtual void receivedMessage(size_t bytes) {};
 };
 
+class MemoryCacheLockRecord {
```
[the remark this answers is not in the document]
mar-cf: Done.

src/workerd/io/observer.h
```diff
@@ -30,6 +32,30 @@ class WebSocketObserver: public kj::Refcounted {
...
+public:
+  explicit MemoryCacheLockRecord(SpanBuilder& span, const kj::MonotonicClock& timer)
+      : span(span),
+        timer(timer),
+        startTime(timer.now()) {}
+
+  ~MemoryCacheLockRecord() noexcept(false) {
+    //auto memoryCacheLockWaitTime = kj::systemPreciseMonotonicClock().now() - startTime;
```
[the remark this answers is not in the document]
mar-cf: Removed.

src/workerd/io/observer.h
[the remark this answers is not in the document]
mar-cf: Moved to tracing header and generalized.

src/workerd/api/memory-cache.c++
```diff
@@ -239,14 +242,22 @@ SharedMemoryCache::Use::~Use() noexcept(false) {
 }
 
 kj::Maybe<kj::Own<CacheValue>> SharedMemoryCache::Use::getWithoutFallback(
-    const kj::String& key) const {
-  auto data = cache->data.lockExclusive();
+    const kj::String& key, SpanBuilder& span) const {
+  kj::Locked<ThreadUnsafeData> data = [&] {
+    auto memoryCacheLockRecord =
+        ScopedDurationTagger(span, "memory_cache_lock_wait_time_ns"_kjc, cache->timer);
```
mikea, line 248: since this is used twice lets extract single `constexpr memoryCachekLockWaitTimeTag = ...`

mikea (approved):

PR #2780, opened 2024-09-24
Update v8 to 13.0

dom96:
Thanks to @erikcorry for the help with this one!


fhanau:
Pushed the changes to the shared linkage patch here, hope that's ok. The updated patch is cleaner and should be easier to upstream.

fhanau (approved):

PR #2782, opened 2024-09-24
Revert the revert of `fastEncodeBase64Url` method

anonrig:
Applying the suggestion of adding a null terminator before passing the input to `kj::String`

Reverts https://github.com/cloudflare/workerd/pull/2776

### DIFF
--- a/src/workerd/api/util.c++
+++ b/src/workerd/api/util.c++
@@ -292,11 +292,14 @@ void maybeWarnIfNotText(jsg::Lock& js, kj::StringPtr str) {
 }
 
 kj::String fastEncodeBase64Url(kj::ArrayPtr<const byte> bytes) {
+  if (KJ_UNLIKELY(bytes.size() == 0)) {
+    return {};
+  }
   auto expected_length = simdutf::base64_length_from_binary(bytes.size(), simdutf::base64_url);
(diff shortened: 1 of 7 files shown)


anonrig:
Internal PR passes

src/workerd/api/util.c++
```diff
@@ -288,4 +290,17 @@ void maybeWarnIfNotText(jsg::Lock& js, kj::StringPtr str) {
...
+
+kj::String fastEncodeBase64Url(kj::ArrayPtr<const byte> bytes) {
+  if (KJ_UNLIKELY(bytes.size() == 0)) {
+    return {};
+  }
+  auto expected_length = simdutf::base64_length_from_binary(bytes.size(), simdutf::base64_url);
+  auto output = kj::heapArray<char>(expected_length + 1);
+  auto actual_length = simdutf::binary_to_base64(
+      bytes.asChars().begin(), bytes.size(), output.asChars().begin(), simdutf::base64_url);
+  output[actual_length] = '\0';
```
jclee, line 302: FWIW, I noticed the old code had an assert that `expected` was close to `actual`.  But maybe this isn't crucial to recheck, if we trust simdutf to make the same guarantee with regard to `base64_length_from_binary()` and `binary_to_base64()`.

PR #2784, opened 2024-09-24
improve utf8 to utf16 conversion performance

anonrig:
Introduces `fastEncodeUtf16` function that uses simdutf under the hood to convert utf8 inputs to utf16. The changeset also includes the usage of `simdutf::validate_ascii` to check if an input is a valid ascii.


anonrig:
Passes all internal tests.

jasnell (approved):

PR #2785, opened 2024-09-24
disabling all precommit hooks that depend on external binaries

mikea:
This got out of hand and many developers report problems. All third-party dependencies need to be fetched by bazel if they are a requirement for workerd development.

Feel free to uncomment particular formatters once dependency management is fixed.

This is not a reflection on linters/formatters at all: they are super useful and we should continue increasing the coverage.

### DIFF
--- a/tools/cross/format.py
+++ b/tools/cross/format.py
@@ -186,21 +186,21 @@ class FormatConfig:
 
 
 FORMATTERS = [
-    FormatConfig(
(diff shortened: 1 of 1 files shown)


mar-cf (approved):

anonrig (approved):

anonrig:
I think we can just disable the git hooks as well.

mikea:
> I think we can just disable the git hooks as well.

let them be? I'd like to see these linters reborn.

jasnell (approved):

PR #2789, opened 2024-09-24
Add type definition for R2 event notification message body

bthwaites:
Adding a type that can be used as a parameter for MessageBatch when consuming events from a Queue that were created by an R2 Event Notification.

### DIFF
--- a/types/defines/r2.d.ts
+++ b/types/defines/r2.d.ts
@@ -1,0 +1,43 @@
+/**
+ * The body of an R2 Event Notification sent to a Queue
+ */
+interface R2EventNotification {
+  /**
+   * The owner of the R2 Bucket.
+   */
+	account: string
+  /**
+   * The name of the R2 Bucket.
+   */
+	bucket: string
+	object?: {
+    /**
+     * The name of the object that triggered the Event Notification rule.
+     */
+	  key: string
+    /**
+     * The object size in bytes. Not present for deletions.
+     */
+	  size?: number
+    /**
+     * The object eTag. Not present for deletions.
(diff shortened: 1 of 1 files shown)


types/defines/r2.d.ts
```diff
@@ -0,0 +1,43 @@
...
+	  size?: number
+    /**
+     * The object eTag. Not present for deletions.
+     */
+	  eTag?: string
+	}
+  /**
+   * The R2 action that triggered the Event Notification rule.
+   */
+	action: string
```
jrf0110, line 30: Is action a set of known strings? If so, you could make a discriminated union which discriminates on `action` that way you don't have to put all of the possible keys on a single interface (like `size` and `eTag` do not exist on deletions)

bthwaites: Sounds like a good idea. I do want to keep the type consistent with what we use internally, so I'm working on an internal refactor to use a discriminated union. I'll update this PR as soon as it's ready.

bthwaites: @jrf0110 I finally got back around to updating this - what do you think of the new type?

types/defines/r2.d.ts
```diff
@@ -0,0 +1,79 @@
...
+  | {
+      action: 'CompleteMultipartUpload'
+      object: {
+        key: string
+        size: number
+        eTag: string
+      }
+    }
+  | {
+      action: 'AbortMultipartUpload'
```
petebacondarwin, line 53: This action is not mentioned in the docs.

PR #2793, opened 2024-09-24
[build] Split up more bazel targets, compile_flags.txt fixes

fhanau:
[build] Factor out api:hyperdrive target

[nfc] Assorted build cleanup
- Add back/expand comments after moving dependencies to update-deps system
- Fix compile_flags.txt for Xcode 16 by adding missing path
- Remove Windows include paths from compile_flags.txt – the team does not
  regularly use Windows and the spaces in the file path make using the file to
  build code (e.g. clang++ @compile_flags.txt <file>) fail.
- Add -fbracket-depth=512 to compile_flags.txt so that it can be used to
  compile workerd-api.c++

[build] Split files off of node and server targets
- Fix missing pragma once in dh-primes.h

[nfc] Clean up Rust-related comments, gitignore

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -138,7 +138,7 @@ build --cxxopt="-fbracket-depth=512" --host_cxxopt="-fbracket-depth=512"
 # Additional Rust flags (see https://doc.rust-lang.org/rustc/codegen-options/index.html)
 # Need to disable debug-assertions for fastbuild, should be off automatically for opt. As long as
 # rust-level LTO is not enabled, LLVM bitcode is not needed so -C embed-bitcode=n provides a free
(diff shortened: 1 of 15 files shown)


mikea:
I tried to do the same here, but it broke windows build too: https://github.com/cloudflare/workerd/pull/2759

would be great if you figured it out :)

fhanau:
I figured

> I tried to do the same here, but it broke windows build too: #2759
> 
> would be great if you figured it out :)

And I figured out why it broke too 😢  – see the comments I just pushed. Will eventually transform this into a cleanup PR instead.

npaun (approved):

mikea (approved):

PR #2803, opened 2024-09-25
merge isAsciiWhitespace implementations

anonrig:
Merges 2 implementations and introduces 2 new cc_library targets on bazel.

### DIFF
--- a/src/workerd/api/BUILD.bazel
+++ b/src/workerd/api/BUILD.bazel
@@ -12,6 +12,8 @@ filegroup(
             exclude = [
                 "html-rewriter.c++",
                 "data-url.c++",
+                "encoding.c++",
+                "util.c++",
                 "rtti.c++",
                 "**/*test*.c++",
                 "pyodide.c++",
@@ -23,6 +25,8 @@ filegroup(
             exclude = [
                 "html-rewriter.c++",
                 "data-url.c++",
+                "encoding.c++",
+                "util.c++",
                 "rtti.c++",
                 "**/*test*.c++",
                 "pyodide.c++",
@@ -38,6 +42,8 @@ api_header_exclusions = [
     "html-rewriter.h",
     "deferred-proxy.h",
     "data-url.h",
+    "encoding.h",
+    "util.h",
     "modules.h",
     "rtti.h",
     "**/*test*.h",
@@ -122,9 +128,39 @@ kj_test(
 )
 
 wd_cc_library(
+    name = "util",
+    srcs = ["util.c++"],
+    hdrs = ["util.h"],
+    implementation_deps = ["@simdutf"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//src/workerd/jsg",
+        "//src/workerd/util",
+        "@capnp-cpp//src/kj",
+    ],
+)
+
+wd_cc_library(
+    name = "encoding",
+    srcs = ["encoding.c++"],
+    hdrs = ["encoding.h"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":util",
+        "//src/workerd/io:compatibility_date_capnp",
+        "//src/workerd/jsg",
+        "//src/workerd/util",
+        "@capnp-cpp//src/kj",
+    ],
+)
+
+wd_cc_library(
     name = "data-url",
     srcs = ["data-url.c++"],
     hdrs = ["data-url.h"],
+    implementation_deps = [
+        ":encoding",
+    ],
     visibility = ["//visibility:public"],
     deps = [
         "//src/workerd/jsg:url",
@@ -206,6 +242,7 @@ kj_test(
 kj_test(
     src = "api-rtti-test.c++",
     deps = [
+        ":encoding",
         ":html-rewriter",
         "//src/workerd/io",
         "//src/workerd/jsg:rtti",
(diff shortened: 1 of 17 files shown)


src/workerd/api/BUILD.bazel
```diff
@@ -38,6 +44,8 @@ api_header_exclusions = [
     "html-rewriter.h",
     "deferred-proxy.h",
     "data-url.h",
+    "encoding.h",
+    "util.h",
     "modules.h",
```
fhanau, line 49: If a source file is being excluded somewhere, the corresponding header should also be excluded from the target, otherwise it's easy to get linker errors due to missing symbols (e.g. when depending on io but not on util-iocontext). I think the goal of splitting util.c++ to move it out of the io target is worth pursing, but either util-iocontext.c++ must be a part of the io target; or the one function it has should be moved into io-context.c++/.h itself (second option is cleaner IMHO).

src/workerd/api/BUILD.bazel
```diff
@@ -159,6 +205,8 @@ wd_cc_capnp_library(
     kj_test(
         src = f,
         deps = [
+            ":encoding",
+            ":util-iocontext",
```
fhanau, line 209: These tests need to depend on util-iocontext even though none of them directly use it, that could be avoided based on what I commented above

src/workerd/api/BUILD.bazel
```diff
@@ -121,10 +129,48 @@ kj_test(
     ],
 )
 
+wd_cc_library(
+    name = "util",
+    srcs = ["util.c++"],
+    hdrs = ["util.h"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//src/workerd/jsg",
```
fhanau, line 138: Hope this isn't too much bikeshedding, just want to get this target right: @simdutf should be implementation_deps here since it's only used in the .c++ file

anonrig: The comments and the reviews are extremely helpful. Thank you! (And I don't consider this as bike-shedding)

src/workerd/io/io-timers.h
```diff
@@ -86,4 +86,7 @@ class TimeoutManager {
   virtual kj::Maybe<kj::Date> getNextTimeout() const = 0;
 };
 
+// Returns exactly what Date.now() would return.
```
fhanau, line 89: This is not needed here right?

src/workerd/io/BUILD.bazel
```diff
@@ -332,13 +335,17 @@ kj_test(
 
 kj_test(
     src = "promise-wrapper-test.c++",
-    deps = [":io"],
+    deps = [
+        ":io",
+        "//src/workerd/api:encoding",
```
fhanau, line 340: Neither of these two tests depends directly on encoding right? Let's IWYU here and leave these as they were before

src/workerd/api/util.c++
```diff
@@ -259,15 +257,6 @@ kj::String redactUrl(kj::StringPtr url) {
   return kj::String(redacted.releaseAsArray());
 }
 
-double dateNow() {
```
jasnell, line 262: Unrelated change?

src/workerd/api/BUILD.bazel
```diff
@@ -177,6 +213,7 @@ kj_test(
     src = "data-url-test.c++",
     deps = [
         ":data-url",
+        ":encoding",
```
fhanau, line 216: The test doesn't depend on encoding either. The same goes for api-rtti-test – encoding.h is included in there but not actually needed I think.

src/workerd/api/util.c++
jasnell: ah, nevermind, I see that things were shifted around a bit.

src/workerd/api/data-url.c++
```diff
@@ -11,18 +13,6 @@ kj::Maybe<DataUrl> DataUrl::tryParse(kj::StringPtr url) {
...
 
-static constexpr kj::FixedArray<uint8_t, 256> ascii_whitespace_table = []() consteval {
-  kj::FixedArray<uint8_t, 256> result{};
-  for (uint8_t c: {0x09, 0x0a, 0x0c, 0x0d, 0x20}) {
-    result[c] = true;
-  }
-  return result;
-}();
-
-constexpr bool isAsciiWhitespace(uint8_t c) noexcept {
```
fhanau, line 22: Conceptually, this could affect performance – `isAsciiWhitespace()` does a single table lookup so having it defined in a different source file will make using it much slower based on needing a function call now. The compiler may or may not be able to inline it properly if LTO is enabled (such as with the downstream release build), so this is non-blocking.

src/workerd/api/util.c++
anonrig: yes, this change is required in order to avoid depending on io on this file, which eventually creates a circular dependency. the best way to avoid this by moving it to io-util.h

src/workerd/api/data-url.c++
anonrig: I moved the implementation to encoding.h and made it constexpr again. Since it is an extremely small fixed array, I don't think it would have any negative impact, but let me know if you disagree. I can change it back without any hassle.

jasnell (approved):

anonrig:
...just in case...

fhanau (approved):

mikea (approved):

PR #2804, opened 2024-09-25
Have nodejs_zlib implied by nodejs_compat_v2

jasnell:
Ensures that if either nodejs_compat or nodejs_compat_v2 are enabled, then nodejs_zlib will be enabled also

There was a limitation in the original definition of the implied by annotation in that it would only allow specifying a single field. The effect was that `nodejs_zlib` would be implied by `nodejs_compat` but not `noejs_compat_v2`. If I updated it to be implied by `nodejs_compat_v2` then it would transitively work for both `nodejs_compat` and `nodejs_compat_v2` but *wouldn't* do the right thing if the worker had `nodejs_compat, no_nodejs_compat_v2`... Confusing.

This PR modifies the implied by annotation to allow multiple flags to be specified so that it works correct in any combination.

### DIFF
--- a/src/workerd/api/node/tests/zlib-nodejs-test.js
+++ b/src/workerd/api/node/tests/zlib-nodejs-test.js
@@ -2603,3 +2603,22 @@ const PSS_VECTORS_JSON = `{
     ]
   }
 }`;
+
+export const compatFlagTest = {
+  async test(_, env) {
+    const waitForIt = async (test) => {
+      const res = await env[test].fetch('http://example.org');
+      return await res.text();
+    };
+    const results = await Promise.allSettled([
+      waitForIt('compat'),
+      waitForIt('compatv2'),
+      waitForIt('compatNoV2'),
+    ]);
+    assert.deepStrictEqual(results, [
+      { status: 'fulfilled', value: 'true' },
+      { status: 'fulfilled', value: 'true' },
+      { status: 'fulfilled', value: 'true' },
+    ]);
+  },
+};
--- a/src/workerd/io/compatibility-date.c++
+++ b/src/workerd/io/compatibility-date.c++
@@ -153,7 +153,7 @@ void compileCompatibilityFlags(kj::StringPtr compatDate,
     kj::Maybe<CompatDate> enableDate;
     kj::StringPtr enableFlagName;
     kj::StringPtr disableFlagName;
-    kj::Maybe<ImpliedBy> maybeImpliedBy;
+    kj::Vector<ImpliedBy> impliedByVector;
 
     for (auto annotation: field.getProto().getAnnotations()) {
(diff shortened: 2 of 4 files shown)


src/workerd/io/compatibility-date.capnp
```diff
@@ -584,7 +587,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   nodeJsZlib @59 :Bool
       $compatEnableFlag("nodejs_zlib")
       $compatDisableFlag("no_nodejs_zlib")
-      $impliedByAfterDate(name = "nodeJsCompat", date = "2024-09-23");
+      $impliedByAfterDate(names = ["nodeJsCompat", "nodeJsCompatV2"], date = "2024-09-23");
   # Enables node:zlib implementation while it is in-development.
   # Once the node:zlib implementation is complete, this will be automatically enabled when
   # nodejs_compat is enabled.
```
vicb, line 593: Nit: or _2

src/workerd/io/compatibility-date.capnp
```diff
@@ -11,7 +11,10 @@ $Cxx.allowCancellation;
 struct ImpliedByAfterDate @0x8f8c1b68151b6cff {
   # Annotates a compatibility flag to indicate that it is implied by the enablement
   # of the named flag after the specified date.
```
vicb, line 13: Nit: flag(s)

src/workerd/api/node/tests/zlib-nodejs-test.js
```diff
@@ -2603,3 +2603,22 @@ const PSS_VECTORS_JSON = `{
...
 }`;
+
+export const compatFlagTest = {
+  async test(_, env) {
+    const waitForIt = async (test) => {
+      const res = await env[test].fetch('http://example.org');
+      return await res.text();
+    };
+    const results = await Promise.allSettled([
+      waitForIt('test1'),
```
vicb, line 2614: Nit: rename test1 -> compat, 2 -> compat_v2, 3 -> compat-no_compat_v2

vicb (approved):
Awesome, thanks! 
I left a few minor comments

anonrig (approved):

src/workerd/io/compatibility-date.c++
```diff
@@ -153,7 +153,7 @@ void compileCompatibilityFlags(kj::StringPtr compatDate,
     kj::Maybe<CompatDate> enableDate;
     kj::StringPtr enableFlagName;
     kj::StringPtr disableFlagName;
-    kj::Maybe<ImpliedBy> maybeImpliedBy;
+    kj::Vector<ImpliedBy> maybeImpliedBy;
```
anonrig, line 156: Nit:

```suggestion
    kj::Vector<ImpliedBy> impliedBy;
```

anonrig (approved):

PR #2806, opened 2024-09-26
Mount python worker files after taking memory snapshot

hoodmane:
This ensures that the contents of worker files cannot be accessed prior to taking the snapshot and so won't appear in the linear memory.

### DIFF
--- a/src/pyodide/internal/setupPackages.ts
+++ b/src/pyodide/internal/setupPackages.ts
@@ -189,19 +189,26 @@ export function getSitePackagesPath(Module: Module): string {
  * details, so even though we want these directories to be on sys.path, we
  * handle that separately in adjustSysPath.
  */
-export function mountLib(Module: Module, info: TarFSInfo): void {
+export function mountSitePackages(Module: Module, info: TarFSInfo): void {
   const tarFS = createTarFS(Module);
-  const mdFS = createMetadataFS(Module);
   const site_packages = getSitePackagesPath(Module);
   Module.FS.mkdirTree(site_packages);
-  Module.FS.mkdirTree('/session/metadata');
   if (!LOAD_WHEELS_FROM_R2 && !LOAD_WHEELS_FROM_ARTIFACT_BUNDLER) {
     // if we are not loading additional wheels, then we're done
     // with site-packages and we can mount it here. Otherwise, we must mount it in
     // loadPackages().
     Module.FS.mount(tarFS, { info }, site_packages);
   }
+}
+
+export function mountWorkerFiles(Module: Module) {
+  Module.FS.mkdirTree('/session/metadata');
+  const mdFS = createMetadataFS(Module);
   Module.FS.mount(mdFS, {}, '/session/metadata');
+  simpleRunPython(
+    Module,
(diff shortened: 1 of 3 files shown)


danlapid (approved):

src/pyodide/internal/setupPackages.ts
```diff
@@ -189,19 +189,26 @@ export function getSitePackagesPath(Module: Module): string {
...
   }
+}
+
+export function mountWorkerFiles(Module: Module) {
+  Module.FS.mkdirTree('/session/metadata');
+  const mdFS = createMetadataFS(Module);
   Module.FS.mount(mdFS, {}, '/session/metadata');
+  simpleRunPython(
+    Module,
+    `import sys; sys.path.append("/session/metadata"); del sys`
```
danlapid, line 210: I might be totally off base here but wouldn't we want `sys.path.insert(0, ...)` here?
If a user has http.py file with his main.py file we'd want his http.py file to take precedence over the builtin http module when he uses `import http` no?

danlapid: Edit: I can see this was moved so obviously it isn't specific to this PR but just a general question

danlapid: Actually, and sorry for being annoying but the move actually does change behaviour where now site-packages are before session/metadata in path in contrast to previous behaviour.

hoodmane: Good point I put that back. Good idea about `sys.path.insert(0, ...)` I think that is right. I'll make that change in a separate PR though.

src/workerd/io/compatibility-date.capnp
```diff
@@ -424,7 +424,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   pythonWorkers @43 :Bool
       $compatEnableFlag("python_workers")
       $pythonSnapshotRelease(pyodide = "0.26.0a2", pyodideRevision = "2024-03-01",
-          packages = "2024-03-01", backport = 0)
+          packages = "2024-03-01", backport = 1)
```
dom96, line 427: Curious why you're bumping this.

dom96 (approved):
Thanks for fixing this!

src/workerd/io/compatibility-date.capnp
hoodmane: Well I made a new bundle with the updated code with backport number 1. I want the new bundle to be used, so I have to bump the backport number.

dom96: Ahh, because the .ts file changes here need a new bundle in order to be in effect, right?

It's going to take a bit of a mind shift to keep that in mind.

hoodmane: Yeah, it's annoying.

dom96: yeah.. this is the thing that I'm concerned is going to end up causing us too much pain. Like even my small PR will need a regenerated bundle, right? https://github.com/cloudflare/workerd/pull/2830

hoodmane: Thanks!

PR #2807, opened 2024-09-26
Switch from external pyodide autogate to feature flag

hoodmane:
We want to be able to synchronize this behavior change between upload time and runtime. Autogates don't allow us to do this as far as I understand but feature flags do.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -615,4 +615,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # compatibility flag we arrange to have such promise continuations scheduled to run
   # in the correct IoContext if it is still alive, or dropped on the floor with a warning
   # if the correct IoContext is not still alive.
+  pythonExternalBundle @63 :Bool
+      $compatEnableFlag("python_external_bundle")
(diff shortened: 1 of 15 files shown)


danlapid:
Why did you not delete the autogate?

danlapid:
Just for my context, is this autogate currently deployed at all?
If any existing worker depends on it being ON then that worker would presumably fail to run after this change

hoodmane:
We should see some internal test failures, but this autogate shouldn't be deployed outside of VET. Nothing should rely on it being on yet.

src/workerd/io/compatibility-date.capnp
```diff
@@ -612,4 +612,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # compatibility flag we arrange to have such promise continuations scheduled to run
   # in the correct IoContext if it is still alive, or dropped on the floor with a warning
   # if the correct IoContext is not still alive.
+  pythonExternalBundle @63 :Bool
+      $compatEnableFlag("python_external_bundle")
+      $experimental;
```
dom96, line 617: Please add some text to describe what this compat flag does.

dom96 (approved):

PR #2810, opened 2024-09-26
Fixes parsing of imports in multi-line string literals.

dom96:


src/workerd/api/pyodide/pyodide-test.c++
```diff
@@ -115,5 +115,50 @@ KJ_TEST("supports backslash") {
   KJ_REQUIRE(result[5] == "c");
 }
 
+KJ_TEST("multiline-strings ignored") {
+  auto files = kj::heapArrayBuilder<kj::String>(4);
+  files.add(kj::str("FOO=\"\"\"\nimport x\nfrom y import z\n\"\"\""));
```
hoodmane, line 120: Doesn't C++ have a way to add multiline/raw strings? It's a bit hard to read the test cases.

hoodmane (approved):
Thanks!

src/workerd/api/pyodide/pyodide-test.c++
jclee: Yes... R-strings, like: https://github.com/cloudflare/workerd/blob/b1656ad/src/workerd/api/node/buffer-test.c%2B%2B#L16-L24
or: https://github.com/cloudflare/workerd/blob/b1656ad/src/workerd/api/util-test.c%2B%2B#L100

PR #2811, opened 2024-09-26
remove workerd/io as a dependency from pyodide

anonrig:
Removes workerd/io as a dependency from pyodide header file and creates a BUILD.bazel inside pyodide folder to simplify things.


jasnell (approved):

mikea (approved):

src/workerd/api/BUILD.bazel
```diff
@@ -79,6 +79,7 @@ wd_cc_library(
         "//src/pyodide",
         "//src/pyodide:pyodide_extra_capnp",
         "//src/workerd/api/node",
+        "//src/workerd/api/pyodide",
```
fhanau, line 82: This wasn't listed here so far. I was wondering why the build was working so far despite that – turns out that we were failing to exclude pyodide/pyodide.h from hdrs so far. With the pyodide folder now being a subpackage based on adding a BUILD.bazel file there, the file won't be captured by the glob so this is no longer an issue.

fhanau (approved):

danlapid:
Any reason not to progress this?

PR #2812, opened 2024-09-26
simplify workerd/api/node bazel

anonrig:
Moves `workerd/io` to implementation_deps and removes unused includes.

### DIFF
--- a/src/workerd/api/node/BUILD.bazel
+++ b/src/workerd/api/node/BUILD.bazel
@@ -10,6 +10,7 @@ wd_cc_library(
     ),
     hdrs = glob(["**/*.h"]),
     implementation_deps = [
+        "//src/workerd/io",
         "@ada-url",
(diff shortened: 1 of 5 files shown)


src/workerd/api/node/util.c++
```diff
@@ -3,8 +3,6 @@
 //     https://opensource.org/licenses/Apache-2.0
 #include "util.h"
 
-#include <workerd/jsg/jsg.h>
```
fhanau, line 6: Some of these includes should be kept based on IWYU – jsg.h is being used here. https://include-what-you-use.org/

anonrig: `#include "util.h"` already includes them. should we still include them here?

fhanau: Yes https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md

mikea (approved):

PR #2823, opened 2024-09-27
Expose zlib constants directly under zlib module

### LINKED ISSUE #2805: 🐛 Bug Report — Runtime APIs — node:zlib not exporting all necessary APIs via the default export
The latest update of https://workers-nodejs-compat-matrix.pages.dev/ surfaced that we missed exposing some of the newly added `node:zlib` symbols via the default export:

Screenshot 2024-09-25 at 5 02 51 PM

There is 44 symbols missing, they are listed in this range: https://github.com/IgorMinar/workers-nodejs-compat-matrix/blob/325bbf17a9cf432f185a37aa915b6f0457e4cc9b/data/workerd.json#L11300-L11514 and marked as `"missing"`.

Can we correctly export these? Thank you

// @jasnell @anonrig

npaun:
This is supported in NodeJS but is deprecated.
Fixes #2805


src/node/zlib.ts
```diff
@@ -100,12 +213,122 @@ export {
   brotliDecompressSync,
   brotliCompress,
   brotliCompressSync,
+
+  // NodeJS also exports all constants directly under zlib, but this is deprecated
```
anonrig, line 217: @jasnell afaik, there isn't any other alternative way to expose this, but any idea?

anonrig (approved):

src/node/zlib.ts
jasnell: Well... you could move all of the constants here into a separate `internal/zlib-constants` import as named exports only... then do

```js
import * as zlibConstants from 'node-internal:zlib-constants';

export const constants = { ...zlibConstants };
export * from 'node-internal:zlib-constants';
```

But might not be worth the extra hassle. I'd just go with this for now.

jasnell (approved):

src/node/zlib.ts
anonrig: JavaScript is a weird language.

PR #2829, opened 2024-09-29
[o11y] Fix incorrect R2 span names, add initial cache, memory-cache user spans

fhanau:
Splitting this off of #2799, which will take some more work.

### DIFF
--- a/src/workerd/api/cache.c++
+++ b/src/workerd/api/cache.c++
@@ -517,8 +517,9 @@ jsg::Promise<bool> Cache::delete_(
 }
 
 kj::Own<kj::HttpClient> Cache::getHttpClient(
-    IoContext& context, kj::Maybe<kj::String> cfBlobJson, kj::ConstString operationName) {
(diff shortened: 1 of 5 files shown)


src/workerd/api/cache.c++
```diff
@@ -518,7 +518,8 @@ jsg::Promise<bool> Cache::delete_(
 
 kj::Own<kj::HttpClient> Cache::getHttpClient(
     IoContext& context, kj::Maybe<kj::String> cfBlobJson, kj::ConstString operationName) {
```
mikea, line 520: I think we should make `operationName` a LiteralConstString. Would this also simplify the next line?

mikea (approved):

src/workerd/api/cache.c++
fhanau: Thank you for the suggestion, that does make things easier! I tried to get the ConstString stuff right and the approach I had was memory safe at least but this is certainly more elegant/saves a memory allocation.

PR #2830, opened 2024-09-30
Improve errors thrown by simpleRunPython.

dom96:
We run a lot of important code through this function, this is a simple change that should make it easier to track down which code is broken.


hoodmane (approved):

hoodmane:
Thanks!

PR #2833, opened 2024-09-30
convert jsg::dict to use hashmap instead of array

anonrig:
Replaces jsg::dict internals to use kj::HashMap instead of kj::Array.

### DIFF
--- a/src/workerd/api/r2-bucket.h
+++ b/src/workerd/api/r2-bucket.h
@@ -257,13 +257,13 @@ public:
     }
 
     const jsg::Optional<jsg::Dict<kj::String>> getCustomMetadata() const {
-      return customMetadata.map([](const jsg::Dict<kj::String>& m) {
-        return jsg::Dict<kj::String>{
-          .fields =
-              KJ_MAP(f, m.fields) {
-          return jsg::Dict<kj::String>::Field{.name = kj::str(f.name), .value = kj::str(f.value)};
-        },
-        };
+      return customMetadata.map([](const jsg::Dict<kj::String>& cm) {
+        auto res = jsg::Dict<kj::String>{};
+        res.fields.reserve(cm.fields.size());
+        for (auto& m: cm.fields) {
+          res.fields.insert(kj::str(m.key), kj::str(m.value));
+        }
+        return kj::mv(res);
       });
     }
 
--- a/src/workerd/api/urlpattern.c++
+++ b/src/workerd/api/urlpattern.c++
@@ -44,27 +44,23 @@ kj::Maybe<URLPattern::URLPatternComponentResult> execRegex(jsg::Lock& js,
     jsg::JsRef<jsg::JsRegExp>& regex,
     kj::ArrayPtr<const kj::String> nameList,
     kj::StringPtr input) {
-  using Groups = jsg::Dict<kj::String, kj::String>;
-
   KJ_IF_SOME(array, regex.getHandle(js)(js, input)) {
     // Starting at 1 here looks a bit odd but it is intentional. The result of the regex
     // is an array and we're skipping the first element.
     uint32_t index = 1;
     uint32_t length = array.size();
-    kj::Vector<Groups::Field> fields(length - 1);
-
+    jsg::Dict<kj::String> dict{};
+    dict.fields.reserve(length - 1);
     while (index < length) {
       auto value = array.get(js, index);
-      fields.add(Groups::Field{
-        .name = kj::str(nameList[index - 1]),
-        .value = value.isUndefined() ? kj::str() : kj::str(value),
-      });
+      dict.fields.insert(
+          kj::str(nameList[index - 1]), value.isUndefined() ? kj::String() : kj::str(value));
       index++;
     }
 
     return URLPattern::URLPatternComponentResult{
       .input = kj::str(input),
-      .groups = Groups{.fields = fields.releaseAsArray()},
+      .groups = kj::mv(dict),
     };
   }
 
--- a/src/workerd/api/actor-state.c++
+++ b/src/workerd/api/actor-state.c++
@@ -549,9 +549,9 @@ jsg::Promise<void> DurableObjectStorageOperations::putMultiple(
 
     kj::Array<byte> buffer = serializeV8Value(js, field.value);
 
-    units += billingUnits(field.name.size() + buffer.size());
+    units += billingUnits(field.key.size() + buffer.size());
 
-    kvs.add(ActorCacheOps::KeyValuePair{kj::mv(field.name), kj::mv(buffer)});
+    kvs.add(ActorCacheOps::KeyValuePair{kj::mv(field.key), kj::mv(buffer)});
   }
 
   jsg::Promise<void> maybeBackpressure =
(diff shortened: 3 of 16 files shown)


kentonv (commented):
Did you find any use cases where this is actually a win? I looked at several (not all) of the cases covered in this PR and none of them seem to benefit from O(1) lookups. In all these cases, building a HashMap is actually a loss, since it sets up a hash table that is never used.

I think this might suggest we should introduce a new type, like jsg::HashDict, rather than update the existing type.

kentonv:
I don't understand your comment. How does changing from `Array` to `HashMap` avoid copying? AFAICT it's exactly the same data structure, except now there's a hash table on the side. That's only helpful if we ever use it for a lookup, which in many cases we don't.

> For example, in an ideal world, this code should not exist:

What do you think this code should look like instead? How does using a HashMap improve this code?

src/workerd/api/r2-bucket.h
```diff
@@ -256,15 +256,8 @@ class R2Bucket: public jsg::Object {
...
-      return customMetadata.map([](const jsg::Dict<kj::String>& m) {
-        return jsg::Dict<kj::String>{
-          .fields =
-              KJ_MAP(f, m.fields) {
-          return jsg::Dict<kj::String>::Field{.name = kj::str(f.name), .value = kj::str(f.value)};
-        },
-        };
-      });
+    jsg::Optional<jsg::Dict<kj::String>> getCustomMetadata() {
+      return customMetadata.map([](jsg::Dict<kj::String>& m) { return kj::mv(m); });
```
kentonv, line 260: You seem to have changed this to move away the custom metadata, which means a second call to `getCustomMetadata()` would return an empty object. But since this is used to initialize a lazy property, it'll only be called once. If nothing else in the class uses `customMetadata` then maybe it's OK. If so, this needs some comments explaining.

But IMO that kind of change should not be hidden inside a commit which is supposed to be about changing the internal format of `jsg::Dict`. This change should probably be a separate PR, or at least a separate commit, to call it out, in case it breaks something.

jasnell: Definitely agree here. This should be called out or done separately. If this method is meant only to be ever called once to populate the lazy property, I would also consider adding a `KJ_ASSERT` before calling the `return customMetadata...`

anonrig: Reverted the change. I'll follow up in a different pull-request as suggested. Thank you for the reviews.

anonrig:
> I don't understand your comment. How does changing from Array to HashMap avoid copying?

@kentonv kj::Array allows copy whereas kj::HashMap has `KJ_DISALLOW_COPY`, hence moving into kj::HashMap implementation makes copy operations more visible.

src/workerd/api/r2-bucket.h
```diff
@@ -256,14 +256,14 @@ class R2Bucket: public jsg::Object {
...
-              KJ_MAP(f, m.fields) {
-          return jsg::Dict<kj::String>::Field{.name = kj::str(f.name), .value = kj::str(f.value)};
-        },
-        };
+    jsg::Optional<jsg::Dict<kj::String>> getCustomMetadata() {
+      return customMetadata.map([](const jsg::Dict<kj::String>& cm) {
+        auto res = jsg::Dict<kj::String>{};
+        res.fields.reserve(cm.fields.size());
+        for (auto& m: cm.fields) {
+          res.fields.insert(kj::str(kj::mv(m.key)), kj::str(kj::mv(m.value)));
```
jasnell, line 264: You're moving the `m.key` and `m.value` here, where the original was copying. If the `customMetadata` is not being consumed here then these should not move?

src/workerd/api/urlpattern.c++
```diff
@@ -44,27 +44,23 @@ kj::Maybe<URLPattern::URLPatternComponentResult> execRegex(jsg::Lock& js,
...
+    jsg::Dict<kj::String> dict{};
+    dict.fields.reserve(length - 1);
     while (index < length) {
       auto value = array.get(js, index);
-      fields.add(Groups::Field{
-        .name = kj::str(nameList[index - 1]),
-        .value = value.isUndefined() ? kj::str() : kj::str(value),
-      });
+      dict.fields.insert(
+          kj::str(nameList[index - 1]), value.isUndefined() ? kj::str() : kj::str(value));
```
jasnell, line 57: Suggest using `kj::String()` instead of `kj::str()` ... same effect but I think it's slightly more efficient.

jasnell (approved):

kentonv:
> @kentonv kj::Array allows copy

No it doesn't.

kentonv (changes requested):
While it makes intuitive sense for `jsg::Dict` to be backed by `kj::HashMap` (hence the ancient TODO comment), in practice it doesn't appear that any of the use cases actually benefit from it -- and in fact this PR will hurt performance by causing us to construct a lot of hash tables that are never referenced.

If we do actually find use cases where maintaining the hash table is helpful (i.e. we actually use it to do lookups at some point), then I think the right approach would be to introduce a new type like `jsg::HashDict` which provides this functionality, so as not to slow down all the use cases that don't need it.

anonrig:
@kentonv This already caught a bug. Having a vector means, the uniqueness of the key is not guaranteed, and it seems it is required on some cases:

If you apply the following patch to this pull-request:

```diff
diff --git a/src/workerd/api/sql.c++ b/src/workerd/api/sql.c++
index 1157b416..94cde329 100644
--- a/src/workerd/api/sql.c++
+++ b/src/workerd/api/sql.c++
@@ -242,6 +242,9 @@ kj::Maybe<SqlStorage::SqlRow> SqlStorage::Cursor::rowIteratorNext(
     jsg::Dict<SqlValue, jsg::JsString> result{};
     result.fields.reserve(fields.size());
     for (auto& field: fields) {
+      KJ_IF_SOME(entry, result.fields.find(field.key)) {
+        JSG_FAIL_REQUIRE(Error, kj::str("Key already exists ", field.key.toString(js)));
+      }
       result.fields.insert(kj::mv(field.key), kj::mv(field.value));
     }
     return kj::mv(result);
```

And run tests...

```
workerd/server/server.c++:3851: debug: [ TEST ] main
To execute a transaction, please use the state.storage.transaction() API instead of the SQL BEGIN TRANSACTION or SAVEPOINT statements. The JavaScript API is safer because it will automatically roll back on exceptions, and because it interacts correctly with Durable Objects' automatic atomic write coalescing.
workerd/io/worker.c++:2037: info: uncaught exception; source = Uncaught (in promise); stack = Error: Key already exists c
    at Function.from (<anonymous>)
    at test (worker:806:30)
    at async DurableObjectExample.fetch (worker:1213:7)
workerd/io/io-context.c++:352: info: uncaught exception; exception = workerd/jsg/_virtual_includes/jsg/workerd/jsg/value.h:1368: failed: jsg.Error: Key already exists c
stack: src/workerd/server/workerd@58f5480 src/workerd/server/workerd@58f73b6 src/workerd/server/workerd@58f729f src/workerd/server/workerd@58f6f28 src/workerd/server/workerd@bb8f440 src/workerd/server/workerd@bb7936c src/workerd/server/workerd@5588b20 src/workerd/server/workerd@55897e0
workerd::api::ServiceWorkerGlobalScope::request(kj::HttpMethod, kj::StringPtr, kj::HttpHeaders const&, kj::AsyncInputStream&, kj::HttpService::Response&, kj::Maybe<kj::StringPtr>, workerd::Worker::Lock&, kj::Maybe<workerd::api::ExportedHandler&>)::$_6::operator()(workerd::api::DeferredProxy<void>) at global-scope.c++:0:0
```

PR #2835, opened 2024-10-01
improve stringwrapper bytestring unwrap

anonrig:
Improves StringWrapper `kj::Maybe<ByteString> tryUnwrap` performance

### DIFF
--- a/src/workerd/jsg/value.h
+++ b/src/workerd/jsg/value.h
@@ -8,6 +8,7 @@
 // Handling of various basic value types: numbers, booleans, strings, optionals, maybes, variants,
 // arrays, buffers, dicts.
 
+#include "simdutf.h"
 #include "util.h"
 #include "web-idl.h"
 #include "wrappable.h"
@@ -468,20 +469,23 @@ public:
       kj::Maybe<v8::Local<v8::Object>> parentObject) {
     // TODO(cleanup): Move to a HeaderStringWrapper in the api directory.
     v8::Local<v8::String> str = check(handle->ToString(context));
-    auto result =
-        ByteString(KJ_ASSERT_NONNULL(tryUnwrap(context, str, (kj::String*)nullptr, parentObject)));
-    if (str->ContainsOnlyOneByte()) {
-      auto buf = kj::heapArray<kj::byte>(str->Length() + 1);
-      str->WriteOneByte(context->GetIsolate(), buf.begin());
-      for (auto b: buf) {
-        if (b >= 128) {
-          result.warning = ByteString::Warning::CONTAINS_EXTENDED_ASCII;
-          break;
-        }
+    auto result = ByteString(KJ_ASSERT_NONNULL(
+        tryUnwrap(context, str, static_cast<kj::String*>(nullptr), parentObject)));
+
+    if (!simdutf::validate_ascii(result.begin(), result.size())) {
+      // If storage is one-byte or the string contains only one-byte
+      // characters, we know that it contains extended ASCII characters.
+      //
+      // The order of execution matters, since ContainsOnlyOneByte()
+      // will scan the whole string for two-byte storage.
+      if (str->ContainsOnlyOneByte()) {
+        result.warning = ByteString::Warning::CONTAINS_EXTENDED_ASCII;
+      } else {
+        // Storage is two-bytes and it contains two-byte characters.
+        result.warning = ByteString::Warning::CONTAINS_UNICODE;
       }
-    } else {
-      result.warning = ByteString::Warning::CONTAINS_UNICODE;
     }
+
     return kj::mv(result);
   }
 };
--- a/src/workerd/jsg/BUILD.bazel
+++ b/src/workerd/jsg/BUILD.bazel
@@ -52,7 +52,6 @@ wd_cc_library(
         "web-idl.h",
         "wrappable.h",
     ],
-    implementation_deps = ["@simdutf"],
     visibility = ["//visibility:public"],
     deps = [
         ":exception",
@@ -65,6 +64,7 @@ wd_cc_library(
         "//src/workerd/util:thread-scopes",
         "//src/workerd/util:uuid",
         "@capnp-cpp//src/kj",
+        "@simdutf",
         "@workerd-v8//:v8",
     ],
 )


src/workerd/jsg/value.h
```diff
@@ -470,14 +471,11 @@ class StringWrapper {
     v8::Local<v8::String> str = check(handle->ToString(context));
     auto result =
         ByteString(KJ_ASSERT_NONNULL(tryUnwrap(context, str, (kj::String*)nullptr, parentObject)));
-    if (str->ContainsOnlyOneByte()) {
+    if (str->ContainsOnlyOneByte()) [[likely]] {
```
kentonv, line 474: Wait, this whole block looks suspect. We already decoded the string content to a kj::String... shouldn't we scan that, instead of allocating a whole second buffer and using `WriteOneByte()` to copy the content again?

Also I note the doc comment for `ContainsOnlyOneByte()` says that it "Will read the entire string in some cases." -- in what cases? Can we avoid that scan as well?

anonrig: Yes, you're right. I've changed the code a little bit to avoid any unnecessary performance impact.

src/workerd/jsg/value.h
```diff
@@ -468,21 +469,22 @@ class StringWrapper {
...
+
+    // This will avoid reading the whole underlying string.
+    if (str_view.is_one_byte()) [[likely]] {
+      if (!simdutf::validate_ascii(
+              reinterpret_cast<const char*>(str_view.data8()), str_view.length())) [[unlikely]] {
+        result.warning = ByteString::Warning::CONTAINS_EXTENDED_ASCII;
       }
-    } else {
-      result.warning = ByteString::Warning::CONTAINS_UNICODE;
+      return kj::mv(result);
```
kentonv, line 483: I wonder if having two return statements -- particularly one with `kj::mv` and one without -- will confuse the compiler out of return value optimization? I would change this back to using an `else` block and have a single `return result;` at the end.

src/workerd/jsg/value.h
```diff
@@ -468,21 +469,22 @@ class StringWrapper {
...
     }
-    return kj::mv(result);
+
+    result.warning = ByteString::Warning::CONTAINS_UNICODE;
```
kentonv, line 486: Hmm, I suspect it's possible for the view to be two-byte but for all the code points to have the upper byte be zero, hence it _could_ be represented as 1-byte, so this warning might be incorrect. (I suppose that's why `ContainsOnlyOneByte()` potentially has to scan the whole string?)

src/workerd/jsg/value.h
anonrig: Agreed. :+1:

src/workerd/jsg/value.h
anonrig: I've taken another try at this. I think we don't need to call simdutf and scan the whole string.

src/workerd/jsg/value.h
```diff
@@ -468,18 +469,26 @@ class StringWrapper {
...
+
+    if (str_view.is_one_byte()) [[likely]] {
+      // Storage is one-byte.
+      if (!simdutf::validate_ascii(
+              reinterpret_cast<const char*>(str_view.data8()), str_view.length())) [[unlikely]] {
+        result.warning = ByteString::Warning::CONTAINS_EXTENDED_ASCII;
+      }
+    } else if (str->ContainsOnlyOneByte()) {
+      // Storage is two-byte, but it contains only one byte.
+      KJ_ASSERT(!str_view.is_one_byte());
```
kentonv, line 485: This assert seems superfluous since we just checked this condition on line 477.

src/workerd/jsg/value.h
```diff
@@ -468,18 +469,26 @@ class StringWrapper {
...
+      auto buf = reinterpret_cast<const char*>(str_view.data16());
```
kentonv, line 486: Err... my first impression was that this cast obviously doesn't work, you can't just reinterpret a pointer from a char16 array to a char8 array.

But I guess it _does_ work in that we already know that all the actual characters are 8-bit, so basically this is an array where every other byte is NUL, which is ASCII, and it's only the non-NUL chars that you intend to validate?

But then you need to double `str_view.length()` on the next line, right? And all this deserves a massive comment.

How about instead, just use the result buffer?

`validate_ascii(result.begin(), result.size())`

The result is UTF-8 encoded, but validate_ascii should handle that just fine, presumably?

In fact, you could do the above `validate_ascii` check first, and only call `ContainsOnlyOneByte()` in the case that it fails (to decide which warning to return), to optimize for the success case...

src/workerd/jsg/value.h
```diff
@@ -468,18 +469,26 @@ class StringWrapper {
...
+      if (!simdutf::validate_ascii(buf, str_view.length())) [[unlikely]] {
```
fhanau, line 487: Not that familiar with text encodings beyond ASCII – I assume that if ContainsOnlyOneByte() applies the most significant byte of a two-byte value is expected to be zero?

src/workerd/jsg/value.h
anonrig: Actually, yes. We're over-engineering it. Directly calling simdutf and eliminating the happy path in the beginning makes things a lot simpler to understand and comprehend in the first look.

src/workerd/jsg/value.h
anonrig: Yes, you are correct. ASCII uses only one byte per character in the range of 0 to 127. In two-byte systems like utf-16, ascii characters are often represented with a leading zero byte followed by the ASCII value in the second byte.If ContainsOnlyOneByte returns true, it means that the value can be represented in a single byte, even though the underlying storage is two-bytes.

fhanau (approved):
Updated version is much cleaner, LGTM

src/workerd/jsg/value.h
```diff
@@ -468,20 +469,25 @@ class StringWrapper {
...
+
+    if (!simdutf::validate_ascii(result.begin(), result.size())) {
+      v8::String::ValueView str_view(context->GetIsolate(), str);
+
+      // If storage is one-byte or the string contains only one-byte
+      // characters, we know that it contains extended ASCII characters.
+      //
+      // The order of execution matters, since ContainsOnlyOneByte()
+      // will scan the whole string for two-byte storage.
+      if (str_view.is_one_byte() || str->ContainsOnlyOneByte()) {
```
fhanau, line 483: v8::String has IsOneByte() and ContainsOnlyOneByte() methods – can we elide the ValueView here?

src/workerd/jsg/value.h
```diff
@@ -468,20 +469,25 @@ class StringWrapper {
...
-    if (str->ContainsOnlyOneByte()) {
-      auto buf = kj::heapArray<kj::byte>(str->Length() + 1);
-      str->WriteOneByte(context->GetIsolate(), buf.begin());
-      for (auto b: buf) {
-        if (b >= 128) {
-          result.warning = ByteString::Warning::CONTAINS_EXTENDED_ASCII;
-          break;
-        }
+    auto result = ByteString(KJ_ASSERT_NONNULL(
+        tryUnwrap(context, str, static_cast<kj::String*>(nullptr), parentObject)));
```
mikea, line 473: nit: I wonder if you can simply `tryUnwrap<kj::String*>` rather than cast.

mikea (approved):

src/workerd/jsg/value.h
anonrig: According to the documentation: v8::String IsOneByte() can return false negatives.

src/workerd/jsg/value.h
anonrig: Wdym? I couldn't follow

anonrig: Ok, I understand it right now. The following code:

```c++
auto result = ByteString(KJ_ASSERT_NONNULL(
        tryUnwrap<kj::String*>(context, str, nullptr, parentObject)));
```

Throws build errors:

```
bazel-out/k8-fastbuild/bin/src/workerd/jsg/_virtual_includes/jsg/workerd/jsg/value.h:473:9: error: member 'tryUnwrap' used before its declaration
  473 |         tryUnwrap<kj::String*>(context, str, nullptr, parentObject)));
      |         ^
      |         this->
bazel-out/k8-fastbuild/bin/external/capnp-cpp/src/kj/_virtual_includes/kj/kj/debug.h:361:27: note: expanded from macro 'KJ_ASSERT_NONNULL'
  361 | #define KJ_ASSERT_NONNULL KJ_REQUIRE_NONNULL
      |                           ^
bazel-out/k8-fastbuild/bin/external/capnp-cpp/src/kj/_virtual_includes/kj/kj/debug.h:301:42: note: expanded from macro 'KJ_REQUIRE_NONNULL'
  301 |     auto _kj_result = ::kj::_::readMaybe(value); \
      |                                          ^
bazel-out/k8-fastbuild/bin/src/workerd/jsg/_virtual_includes/jsg/workerd/jsg/value.h:454:25: note: member is declared here
  454 |   kj::Maybe<kj::String> tryUnwrap(v8::Local<v8::Context> context,
      |                         ^
bazel-out/k8-fastbuild/bin/src/workerd/jsg/_virtual_includes/jsg/workerd/jsg/value.h:466:25: note: member is declared here
  466 |   kj::Maybe<ByteString> tryUnwrap(v8::Local<v8::Context> context,
      |                         ^
bazel-out/k8-fastbuild/bin/src/workerd/jsg/_virtual_includes/jsg/workerd/jsg/value.h:473:9: error: no matching member function for call to 'tryUnwrap'
  473 |         tryUnwrap<kj::String*>(context, str, nullptr, parentObject)));
      |         ^~~~~~~~~~~~~~~~~~~~~~
bazel-out/k8-fastbuild/bin/external/capnp-cpp/src/kj/_virtual_includes/kj/kj/debug.h:361:27: note: expanded from macro 'KJ_ASSERT_NONNULL'
  361 | #define KJ_ASSERT_NONNULL KJ_REQUIRE_NONNULL
      |                           ^
bazel-out/k8-fastbuild/bin/external/capnp-cpp/src/kj/_virtual_includes/kj/kj/debug.h:301:42: note: expanded from macro 'KJ_REQUIRE_NONNULL'
  301 |     auto _kj_result = ::kj::_::readMaybe(value); \
```

anonrig: Mostly because tryUnwrap is an overload, and it doesn't take a template argument.

src/workerd/jsg/value.h
fhanau: Does the same not apply for ValueView too? There is no documentation for the is_one_byte() method, looking at the String::ValueView::ValueView definition the is_one_byte_ variable is set to `flat_content.IsOneByte();` so it should both be the same value unless the flattening eliminates false negatives here.

anonrig: I'm not sure.

fhanau: Another thing: ContainsOnlyOneByte will not always scan the whole thing – it performs the equivalent of the IsOneByte() check and returns early if it is in one-byte representation. That means we can get rid of the first condition entirely assuming that ValueView has the same false positives, ContainsOnlyOneByte will return without checking the string itself in every case where IsOneByte() applies.
Do we have reason to assume that ValueView avoids these false positives? I don't think so, and if it does I don't see how it would do so more cheaply than ContainsOnlyOneByte does without needing the same amount of computation.
TLDR: My understanding is the following: IsOneByte() returns if the string is known to be one-byte, if it is not known if it is or not it returns false (hence "false negatives"). ContainsOnlyOneByte() first checks if it is known to be such, if not it scans the string to find out. ValueView::is_one_byte() is functionally equivalent to IsOneByte(), it does not perform such scanning.

anonrig: Agreed. Removed the ValueView initialization. Thank you for the review @fhanau

jasnell (approved):

PR #2839, opened 2024-10-01
Skip Python parsing of package imports when requirements are empty.

dom96:
This will avoid any potential parser bugs if the deployment isn't using any packages.

### DIFF
--- a/src/pyodide/internal/snapshot.ts
+++ b/src/pyodide/internal/snapshot.ts
@@ -10,6 +10,7 @@ import {
   SHOULD_SNAPSHOT_TO_DISK,
   IS_CREATING_BASELINE_SNAPSHOT,
   MEMORY_SNAPSHOT_READER,
+  REQUIREMENTS,
 } from 'pyodide-internal:metadata';
 import { reportError, simpleRunPython } from 'pyodide-internal:util';
 import { default as MetadataReader } from 'pyodide-internal:runtime-generated/metadata';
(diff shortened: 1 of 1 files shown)


anonrig (approved):

src/pyodide/internal/snapshot.ts
```diff
@@ -261,6 +262,13 @@ function memorySnapshotDoImports(Module: Module): Array<string> {
     return [];
   }
 
+  if (REQUIREMENTS.length == 0) {
+    // Don't attempt to scan for package imports if the Worker has specified no package
+    // requirements, as this means their code isn't going to be importing any modules that we need
+    // to include in a snapshot.
+    return [];
+  }
```
hoodmane, line 270: What's the point of this change? Is it a simple optimization? What about imports to standard library modules?

hoodmane: We should probably also filter out imports to modules in `SNAPSHOT_IMPORTS` in this method.

dom96: The inspiration for this change is the bug I ran into with the import parser. It meant that it was failing even workers that didn't have requirements specified. So this change is just a simple change to ensure this code doesn't even run if requirements aren't specified.

hoodmane: Wasn't it a good thing that we hit the bug and fixed it?

dom96: True, but it meant we were breaking deployments of workers that don't use requirements (something that's out to the world). I'd rather have isolated the breakage to only those workers which use requirements (something that's only usable internally).

hoodmane (approved):

PR #2842, opened 2024-10-02
Fix URL bugs found by fuzzing

fhanau:
Fix minor buffer overflow in url.c++
When the protocol string is empty, we should not try to drop a trailing colon. This was found by using a new fuzzer for URL-related APIs.

Fix URL assertion failure found by fuzzing
This aligns canonicalizeOpaquePathname() with the other canonicalize functions.
Also elides a superfluous memory allocation.

Also see the upstream draft PR.


anonrig (approved):
I prefer if we have some tests for it, but LGTM

PR #2851, opened 2024-10-07
add memory tracking to workerd::mimetype

anonrig:
Adds memory tracking to mimetype and separates several cc libraries to improve build (which is required)

### DIFF
--- a/src/workerd/util/BUILD.bazel
+++ b/src/workerd/util/BUILD.bazel
@@ -29,7 +29,6 @@ wd_cc_library(
 wd_cc_library(
     name = "util",
     srcs = [
-        "mimetype.c++",
         "stream-utils.c++",
         "wait-list.c++",
     ],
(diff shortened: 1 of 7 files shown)


jasnell (approved):

src/workerd/util/BUILD.bazel
```diff
@@ -59,6 +55,36 @@ wd_cc_library(
...
+    deps = [
+        "//src/workerd/jsg:memory-tracker",
+        "@capnp-cpp//src/kj",
+        "@capnp-cpp//src/kj:kj-async",
+    ],
+)
+
+wd_cc_library(
+    name = "strings",
+    srcs = [],
```
fhanau, line 75: Late to this, but two minor potential improvements:
- Empty arguments like srcs and deps here are not needed since cc_library and by extension wd_cc_library supports defaut arguments
- If we apply the principles of IWYU to bazel, we'd want all bazel targets to be "self-sufficient", that is to depend directly on all targets they use. Then the strings target should depend on libkj, otherwise parsing the header would fail. This is a pretty small issue though, almost every source file will be including KJ in some capacity anyway.

PR #2854, opened 2024-10-07
Support ndjson-format for insert & upsert ops in vectorize api

garrettgu10:
In order to support local bindings in miniflare, we have to send requests through the core Vectorize service. This service requires input for `/insert` and `/upsert` operations to be in ndjson format. This PR adds an initialization option that optionally tells the binding to put the input to these operations in ndjson. Existing behavior should be identical.

### DIFF
--- a/src/cloudflare/internal/vectorize-api.ts
+++ b/src/cloudflare/internal/vectorize-api.ts
@@ -18,7 +18,7 @@ enum Operation {
 
 type VectorizeVersion = 'v1' | 'v2';
 
-function toNdJson(arr: object[]) {
+function toNdJson(arr: object[]): string {
   return arr.reduce((acc, o) => acc + JSON.stringify(o) + '\n', '').trim();
 }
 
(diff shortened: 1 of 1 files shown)


src/cloudflare/internal/vectorize-api.ts
```diff
@@ -18,6 +18,10 @@ enum Operation {
 
 type VectorizeVersion = 'v1' | 'v2';
 
+function toNdJson(arr: object[]) {
+  return arr.map((o) => JSON.stringify(o)).join('\n');
```
jasnell, line 22: how big are these arrays expected to be? Asking because using `map` and `join` aren't the fastest ways to do this if the inputs are going to be large (as they end up requiring multiple iterations).

jasnell (approved):

src/cloudflare/internal/vectorize-api.ts
garrettgu10: The array in this case is as large as the number of vectors being inserted. I'll rewrite it to use a single iteration.

garrettgu10:
This change is no longer needed since we decided to block all write requests in local dev mode, which includes both `/insert` and `/upsert`. Will reopen if it turns out we need it.

garrettgu10:
Reopening this as it appears we may want to enable write operations behind a wrangler flag

PR #2860, opened 2024-10-08
Refactor Python startup code to prepare for Python pools

hoodmane:
For Python pools, we have to run `instantiateEmscriptenModule` before we bind the feature flags and set up the global scope and module registry. We will need to bundle all of the relevant code into a single script with no imports. This PR makes most of the typescript changes that will be needed to enable this.

I split part of `python.ts` into a new file called `emscriptenSetup.ts` which intended for execution as part of the pool. To do this, `emscriptenSetup.ts` cannot import anything that uses C++ APIs. After this PR, it nearly satisfies this constraint except for the use of UnsafeEval by `newWasmModule`. I will fix that in a followup.

Aside from moving code from `python.ts` to `emscriptenSetup.ts`, the code changes relate to the fact that `emscriptenSettings` cannot receive information from snapshot.ts. We have to cause the relevant effects dynamically. Instead of setting `INITIAL_MEMORY`, we use `Module.growMemory`. We patch Emscripten to read the value of `Module.noInitialRun` directly before calling `main()` (instead of loading it early and ignoring subsequent changes). And instead of calling `preloadDynamicLibs()` from a prerun hook, we use `addRunDependency()` to stall the initialization of the runtime until we have a chance to load the dynamic libraries.


src/pyodide/internal/python.ts
```diff
@@ -33,7 +30,6 @@ import {
  * _createPyodideModule and pyodideWasmModule together are produced by the
```
danlapid, line 30: fix comment?

src/pyodide/internal/emscriptenSetup.ts
```diff
@@ -0,0 +1,211 @@
...
+ * This file is intended to be executed in the Python pool (once it exists). As such, it cannot
+ * import anything that transitively uses C++ extension modules. It has to work in a vanilla v8
+ * isolate. Also, we will have to bundle this file and all of its transitive imports into a single
+ * js file.
+ */
+
+import { reportError } from 'pyodide-internal:util';
+
+/**
+ * _createPyodideModule and pyodideWasmModule together are produced by the
```
danlapid, line 11: fix comment?

src/pyodide/internal/python.ts
```diff
@@ -222,9 +75,13 @@ export async function loadPyodide(
   lockfile: PackageLock,
   indexURL: string
 ): Promise<Pyodide> {
-  const emscriptenSettings = getEmscriptenSettings(lockfile, indexURL);
   const Module = await enterJaegerSpan('instantiate_emscripten', () =>
-    instantiateEmscriptenModule(emscriptenSettings)
+    instantiateEmscriptenModule(
```
danlapid, line 79: Do you intend to export this call to c++ in a separate PR?

src/pyodide/internal/emscriptenSetup.ts
hoodmane: Well what it says is still true and probably even more helpful now that they are imported in separate files.

src/pyodide/internal/python.ts
hoodmane: Yeah

src/pyodide/internal/emscriptenSetup.ts
```diff
@@ -0,0 +1,214 @@
...
+import { _createPyodideModule } from 'pyodide-internal:generated/pyodide.asm';
+
+/**
+ * A preRun hook. Make sure environment variables are visible at runtime.
+ */
+function setEnv(Module: Module): void {
+  Object.assign(Module.ENV, Module.API.config.env);
+}
+
+function getWaitForDynlibs(resolveReadyPromise: PreRunHook): PreRunHook {
```
dom96, line 23: Nit: this doesn't appear to wait for anything, rather it returns a function that resolves a promise. Should the name be changed to better reflect this?

dom96 (approved):
Nice!

danlapid (approved):
LGTM but I really am not quite the expert on this code so I would appreciate @dom96's review over my own, seeing as he approved it already I think this is g2g

src/pyodide/internal/emscriptenSetup.ts
hoodmane: By calling `addRunDependency("dynlibs")` it causes Emscripten to wait until we call `removeRunDependency("dynlibs")`. I will add a comment explaining this since it is indeed a bit opaque.

hoodmane: Oh there already is a comment for it.

PR #2861, opened 2024-10-08
Don't direct people to a site that has an 'install bazel' button

erikcorry:


danlapid (approved):
Seems right to me, @fhanau any opposition?

anonrig (approved):

fhanau (approved):

PR #2866, opened 2024-10-08
Fix a bunch of SQLite-related bugs

kentonv:
* Fix bug where if executing a prepared statement threw (e.g. due to violating a uniqueness constraint), the prepared statement would be permanently broken. (Note we have decided not to make the prepared statement API public, but this bug affected an experimental user.)
* Fix bugs where if the _cf_KV or _cf_METADATA tables were created during a transaction which ended up being rolled back, then the in-memory state could be wrong, resulting in further access doing the wrong thing.
* Throw a more descriptive error message when the user tries to execute `BEGIN TRANSACTION`. The "not authorized" error confused people into thinking that transactions weren't supported!


src/workerd/util/sqlite.c++
```diff
@@ -466,21 +466,123 @@ kj::StringPtr SqliteDatabase::getCurrentQueryForDebug() {
...
-    SQLITE_CALL(sqlite3_prepare_v3(db, sqlCode.begin(), sqlCode.size(), prepFlags, &result, &tail));
+    auto prepareResult =
+        sqlite3_prepare_v3(db, sqlCode.begin(), sqlCode.size(), prepFlags, &result, &tail);
+    if (prepareResult == SQLITE_AUTH) {
+      KJ_IF_SOME(error, parseContext.authError) {
+        // Throw the tailored auth error.
+        kj::throwFatalException(kj::mv(error));
+      }
+    }
+    if (prepareResult != SQLITE_OK) {
```
anonrig, line 582:
```suggestion
    } else if (prepareResult != SQLITE_OK) {
```

jasnell (approved):

src/workerd/util/sqlite-test.c++
```diff
@@ -926,5 +926,152 @@ KJ_TEST("SQLite failed statement reset") {
...
+    // The savepoint still exists until we release it...
+    db.run("RELEASE SAVEPOINT foo");
+  }
+
+  // Prepared statements work.
+  {
+    auto begin = db.prepare("BEGIN TRANSACTION");
+    auto commit = db.prepare("COMMIT TRANSACTION");
+
+    MockRollbackCallback cb1;
```
justin-mp, line 1025: nit: might be worth leaving a "// No transactions are open yet so the callback is dropped immediately." comment so folks aren't confused by `cb1.wasCommitted()` is true.

justin-mp (approved):

src/workerd/util/sqlite.c++
kentonv: No, that would introduce a bug in the case that `parseContext.authError` is not set.

jasnell: A quick comment here might be useful but good either way.

kentonv:
Updates (just to comments): https://github.com/cloudflare/workerd/compare/b3fe27f28e15e4d4c6c929b5fad5333603116bd8..e1766f90b6f398d8347550b2ab3019cc16be4885

kentonv:
windows build failure unrelated; fixed in #2891

PR #2867, opened 2024-10-09
Implement process.exit(...) in the runtime

jasnell:
An implementation of `process.exit(...)` as part of Node.js compat.

Unlike in Node.js, calling `process.exit(...)` does not actually exit the workerd process, for obvious reasons. Instead, when called within the scope of a request, calling `process.exit(...)` will cause the current IoContext to be aborted and will call `isolate->TerminateExecution()` to prevent further execution of JavaScript while leaving the isolate in a good state to run other requests.

This needs to be evaluated very carefully. The approach is a bit of a hack and there might be a better way of doing this that I'm not seeing. So let's not rush this in.

### DIFF
--- a/src/node/internal/util.d.ts
+++ b/src/node/internal/util.d.ts
@@ -122,3 +122,4 @@ export function isBoxedPrimitive(
 
 export function getBuiltinModule(id: string): any;
 export function getCallSite(frames: number): Record<string, string>[];
+export function processExitImpl(code: number): void;
--- a/src/workerd/api/node/tests/process-exit-test.js
+++ b/src/workerd/api/node/tests/process-exit-test.js
@@ -1,0 +1,63 @@
+import { fail, ok, rejects, strictEqual } from 'assert';
+
+let called = false;
+
+process.exit(9999);
+
+export const test = {
+  async test(_, env) {
+    // We don't really have a way to verifying that the correct log messages
+    // we emitted. For now, we need to manually check the log out, and here
+    // we check only that we received an internal error and that no other
+    // lines of code ran after the process.exit call.
+    await rejects(env.subrequest.fetch('http://example.org'), {
+      message: /^The Node.js process.exit/,
+    });
+    ok(!called);
+  },
+};
+
+let fooCreateCount = 0;
+
+export const test2 = {
+  async test(_, env) {
+    // We don't really have a way to verifying that the correct log messages
+    // we emitted. For now, we need to manually check the log out, and here
+    // we check only that we received an internal error and that no other
+    // lines of code ran after the process.exit call.
+    {
+      const obj = env.foo.get(
+        env.foo.idFromName('210bd0cbd803ef7883a1ee9d86cce06f')
+      );
+      await rejects(obj.fetch('http://example.org'), {
+        message: /^The Node.js process.exit/,
+      });
+      await rejects(obj.fetch('http://example.org'), {
+        message: /^The Node.js process.exit/,
+      });
+      // The durable object should have been created twice.
+      strictEqual(fooCreateCount, 2);
+    }
+  },
+};
+
+export default {
+  async fetch() {
+    try {
+      process.exit(123);
+    } finally {
+      called = true;
+    }
+  },
+};
+
+export class Foo {
+  constructor() {
+    fooCreateCount++;
+  }
+
+  fetch() {
+    process.exit(123);
+    fail('Should never get here.');
+  }
+}
--- a/src/workerd/api/node/util.c++
+++ b/src/workerd/api/node/util.c++
@@ -3,6 +3,7 @@
 //     https://opensource.org/licenses/Apache-2.0
 #include "util.h"
 
+#include <workerd/io/io-context.h>
 #include <workerd/jsg/jsg.h>
 
 #include <kj/vector.h>
@@ -238,6 +239,54 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
   }
 
   return js.undefined();
+}
+
+namespace {
+[[noreturn]] void handleProcessExit(jsg::Lock& js, int code) {
+  // There are a few things happening here. First, we abort the current IoContext
+  // in order to shut down this specific request....
+  auto message =
+      kj::str("The Node.js process.exit(", code, ") API was called. Canceling the request.");
+  auto& ioContext = IoContext::current();
+  // If we have a tail worker, let's report the error.
+  KJ_IF_SOME(tracer, ioContext.getWorkerTracer()) {
+    // Why create the error like this in tracing? Because we're adding the exception
+    // to the trace and ideally we'd have the JS stack attached to it. Just using
+    // JSG_KJ_EXCEPTION would not give us that, and we only want to incur the cost
+    // of creating and capturing the stack when we actually need it.
+    auto ex = KJ_ASSERT_NONNULL(js.error(message).tryCast<jsg::JsObject>());
+    tracer.addException(ioContext.now(), ex.get(js, "name"_kj).toString(js),
+        ex.get(js, "message"_kj).toString(js), ex.get(js, "stack"_kj).toString(js));
+    ioContext.abort(js.exceptionToKj(ex));
+  } else {
+    ioContext.abort(JSG_KJ_EXCEPTION(FAILED, Error, kj::mv(message)));
+  }
+  // ...then we tell the isolate to terminate the current JavaScript execution.
+  // Oddly however, this does not appear to *actually* terminate the thread of
+  // execution unless we trigger the Isolate to handle the intercepts, which
+  // calling v8::JSON::Stringify does. Weird... but ok? As long as it works
+  // TODO(soon): Investigate if there is a better approach to triggering the
+  // interrupt handling.
+  js.v8Isolate->TerminateExecution();
+  jsg::check(v8::JSON::Stringify(js.v8Context(), js.str()));
+  // This should be unreachable here as we expect the isolate to terminate and
+  // an exception to have been thrown.
+  KJ_UNREACHABLE;
+}
+}  // namespace
+
+void UtilModule::processExitImpl(jsg::Lock& js, int code) {
+  if (IoContext::hasCurrent()) {
+    handleProcessExit(js, code);
+  }
+
+  // Create an error object so we can easily capture the stack where the
+  // process.exit call was made.
+  auto err = KJ_ASSERT_NONNULL(
+      js.error("process.exit(...) called without a current request context. Ignoring.")
+          .tryCast<jsg::JsObject>());
+  err.set(js, "name"_kj, js.str());
+  js.logWarning(kj::str(err.get(js, "stack"_kj)));
 }
 
 }  // namespace workerd::api::node
--- a/src/workerd/io/io-context.c++
+++ b/src/workerd/io/io-context.c++
@@ -1089,6 +1089,16 @@ void IoContext::runImpl(Runnable& runnable,
         // Check if we hit a limit.
         limitEnforcer->requireLimitsNotExceeded();
 
+        // If we were terminated because abort() was called, then it's not an unknown
+        // reason...
+        if (!abortFulfiller->isWaiting()) {
+          // Nothing to do here. Do not log anything. The assumption is that we've
+          // terminated because the IoContext was aborted and isolate->TerminateExection()
+          // was called (likely because of someone using process.exit(...) in Node.js
+          // compat mode).
+          return;
+        }
+
         // That should have thrown, so we shouldn't get here.
         KJ_FAIL_ASSERT("script terminated for unknown reasons");
       } else {
--- a/src/node/internal/process.ts
+++ b/src/node/internal/process.ts
@@ -88,8 +88,13 @@ export function getBuiltinModule(id: string): any {
   return utilImpl.getBuiltinModule(id);
 }
 
+export function exit(code: number) {
+  utilImpl.processExitImpl(code);
+}
+
 export default {
   nextTick,
   env,
+  exit,
   getBuiltinModule,
 };
--- a/src/workerd/api/node/BUILD.bazel
+++ b/src/workerd/api/node/BUILD.bazel
@@ -54,6 +54,7 @@ wd_cc_library(
     ],
     visibility = ["//visibility:public"],
     deps = [
+        "//src/workerd/io",
         "//src/workerd/io:compatibility-date_capnp",
         "//src/workerd/jsg",
         "//src/workerd/util:mimetype",
@@ -206,4 +207,10 @@ wd_test(
     src = "tests/module-create-require-test.wd-test",
     args = ["--experimental"],
     data = ["tests/module-create-require-test.js"],
+)
+
+wd_test(
+    src = "tests/process-exit-test.wd-test",
+    args = ["--experimental"],
+    data = ["tests/process-exit-test.js"],
 )
--- a/src/workerd/api/node/tests/process-exit-test.wd-test
+++ b/src/workerd/api/node/tests/process-exit-test.wd-test
@@ -1,0 +1,23 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "process-exit-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "process-exit-test.js")
+        ],
+        compatibilityDate = "2024-10-01",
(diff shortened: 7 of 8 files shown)


src/workerd/api/node/util.c++
```diff
@@ -240,4 +241,30 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
   return js.undefined();
 }
 
+void UtilModule::processExitImpl(jsg::Lock& js, int code) {
```
anonrig, line 244: node.js stringifies every code. for example running `process.exit('00')` or `process.exit('11')` works. let's make sure we accept any input.

src/node/internal/util.d.ts
```diff
@@ -122,3 +122,4 @@ export function isBoxedPrimitive(
 
 export function getBuiltinModule(id: string): any;
 export function getCallSite(frames: number): Record<string, string>[];
+export function processExitImpl(code: number): void;
```
anonrig, line 125: why not?

```suggestion
export function exitProcess(code: number): void;
```

src/workerd/api/node/tests/process-exit-test.js
```diff
@@ -0,0 +1,26 @@
...
+      message: /^The Node.js process.exit/,
+    });
+    ok(!called);
+  },
+};
+
+export default {
+  async fetch() {
+    try {
+      process.exit(123);
```
anonrig, line 21: I think `process.exit(1)` and `process.exit(0)` should make a difference.

jasnell: A difference in what way?

src/workerd/api/node/util.c++
jasnell: I believe there are already tests elsewhere that verify that the value coercion works with the jsg type mappings. Manually checking `process.exit('00')` with this and it works as expected.

src/node/internal/util.d.ts
jasnell: I actually prefer `processExitImpl` here since that's pretty much exactly what it is ;-) ... the `process.exit Impl`

src/workerd/api/node/tests/process-exit-test.js
anonrig: Maybe we should behave differently depending on the exit codes, such as if 0, we succeed and keep the behavior as is, and for 1 (failure) do something else?

jasnell: Yeah, but behave differently *how*?  In either case the call to `process.exit(...)` is expected to be terminal in a way we do not support but need to approximate. There's not expected to be any way of handling the exit outside of `process.on('exit')` which doesn't really change anything about the behavior. Either way we need to stop JS from executing further on this request and need to shut down the IoContext. Really not sure what else we can do here so open to more specific suggestions ;-)

danlapid:
Maybe trigger a specific trace event on process.exit?
I think it can be very confusing to debug a worker that had some random library decide to process.exit().
Possibly even attaching a stacktrace?

jasnell:
> Maybe trigger a specific trace event on process.exit?

The error that is reported via `ioContext.abort(...)` where there is a current IoContext should include the JS stack trace identifying the location of the error, AFAICT. When there isn't an active IoContext the warning that is emitted will also include the stacktrace.

danlapid:
> > Maybe trigger a specific trace event on process.exit?
> 
> The error that is reported via `ioContext.abort(...)` where there is a current IoContext should include the JS stack trace identifying the location of the error, AFAICT. When there isn't an active IoContext the warning that is emitted will also include the stacktrace.

But will that be reported to a trace worker?

jasnell:
> But will that be reported to a trace worker?

Ah, good question. No, because we're aborting the IoContext it would not be reported. I've updated to report to the tracer (if it exists) before calling abort.

src/workerd/api/node/tests/process-exit-test.js
anonrig: I don't have any suggestions on this, and once I thought about this, I realized that the behavior shouldn't change.

anonrig (approved):

jasnell:
Sidenote: When documenting this, we need to remember to emphasize that folks really *should not* be making use of `process.exit(...)` in workers, and that any code that does depend on it should be rewritten to avoid it. While we're making it act reasonably here, the pattern of abruptly aborting the request context and terminating javascript execution can be problematic, especially if that happens in the middle of updating some global state.

vicb:
> aborting the request context and terminating javascript execution can be problematic

It's way above my understanding of workerd. But could the workerd impl be more problematic than a `process.exit()` on i.e. Node?

jasnell:
I'm not sure I understand the question. The key difference in workerd is that we're not actually going to destroy the isolate so additional requests can be processed. Depending on when the `process.exit(...)` is called and what else is happening, the globalThis state could potentially be left in a bad state but there's not a lot we can do about that. I suppose we could take the additional step of condemning the isolate but I'm not sure that's the right thing. Interested in @kentonv's thoughts on that

vicb:
> I'm not sure I understand the question.

That's probably because I don't know enough about workerd.

> globalThis state could potentially be left in a bad state

That sounds scary to me - do we really want to allow this?
(This might be a stupid question again, but I think all questions are worth asking)

jasnell:
> That sounds scary to me - do we really want to allow this?

Well, the bad state would essentially be bad application state, not runtime state... and it's entirely dependent on what the user code is doing.... which gets us back to warning developers away from depending on / using `process.exit` at all. Also, in the future if/once we get Lightweight Isolate Clones this problem goes away entirely. I think for right now, as long as the runtime state is good and we've got proper warning markers in place telling folks to avoid using `process.exit(...)` we should be fine.

danlapid:
> > That sounds scary to me - do we really want to allow this?
> 
> Well, the bad state would essentially be bad application state, not runtime state... and it's entirely dependent on what the user code is doing.... which gets us back to warning developers away from depending on / using `process.exit` at all. Also, in the future if/once we get Lightweight Isolate Clones this problem goes away entirely. I think for right now, as long as the runtime state is good and we've got proper warning markers in place telling folks to avoid using `process.exit(...)` we should be fine.

Perhaps a compat flag that would switch between aborting and throwing a JS exception would be nice to allow users to “gracefully” shutdown in case an external library process.exit’d

vicb:
> Well, the bad state would essentially be bad application state, not runtime state...

Ok so that was my initial understanding and the reason why I asked the initial question: how different would it be in workerd vs Node? Would this impl of `process.exit()` in workerd potentially cause more damage than calling `process.exit()` in Node?

jasnell:
Just throwing a catchable error is likely to be *more* dangerous -- we really shouldn't allow the user to ignore that `process.exit(...)` was called as that would be more likely to lead to conditions of application code getting into a bad state. I think we want to be noisy and disruptive when `process.exit(...)` is called -- if for no other reason than to make it explicitly clear that code running in workers should NOT be depending on it.

vicb:
I think we should start charging customers $10 each time they call that `process.exit` ;)

src/workerd/api/node/util.c++
```diff
@@ -240,4 +241,39 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
...
+    // in order to shut down this specific request....
+    auto message =
+        kj::str("The Node.js process.exit(", code, ") API was called. Canceling the request.");
+    auto& ioContext = IoContext::current();
+    // If we have a tail worker, let's report the error.
+    KJ_IF_SOME(tracer, ioContext.getWorkerTracer()) {
+      auto ex = KJ_ASSERT_NONNULL(js.error(message).tryCast<jsg::JsObject>());
+      tracer.addException(ioContext.now(), ex.get(js, "name"_kj).toString(js),
+          ex.get(js, "message"_kj).toString(js), ex.get(js, "stack"_kj).toString(js));
+      ioContext.abort(js.exceptionToKj(ex));
```
kentonv, line 256: Why use a different exception here vs. the non-traced case? Seems weird for tracing to change behavior?

src/workerd/api/node/util.c++
```diff
@@ -240,4 +241,39 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
...
+    } else {
+      ioContext.abort(JSG_KJ_EXCEPTION(FAILED, Error, kj::mv(message)));
+    }
+    // ...then we tell the isolate to terminate the current JavaScript execution.
+    // Oddly however, this does not appear to *actually* terminate the thread of
+    // execution unless we trigger the Isolate to handle the intercepts, which
+    // calling v8::JSON::Stringify does. Weird... but ok? As long as it works
+    // TODO(soon): Investigate if there is a better approach to triggering the
+    // interrupt handling.
+    js.v8Isolate->TerminateExecution();
```
kentonv, line 266: Does `IoContext::runImpl()` do the right thing after this? It looks like it'll throw a KJ exception `script terminated for unknown reasons` which would end up in our Sentry.

src/workerd/api/node/util.c++
```diff
@@ -240,4 +241,39 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
...
+    v8::JSON::Stringify(js.v8Context(), js.str()).IsEmpty();
```
kentonv, line 267: How did you come up with JSON stringification as the best solution here?

I mean, I guess it's fine, just curious what else you tried.

IgorMinar (approved):
I don't have sufficient knowledge of workerd code to review the implementation, but the goal of:

>  calling process.exit(...) will cause the current IoContext to be aborted and will call isolate->TerminateExecution() to prevent further execution of JavaScript while leaving the isolate in a good state to run other requests.

seems like a reasonable behavior that closely mimics what node.js does, so I'm supportive of this goal as specified.

src/workerd/api/node/util.c++
jasnell: Picking through the v8 code to see what paths would trigger handling the scheduled interrupts with the least amount of overhead. The JSON stringify method handles the interrupts very early in the function and is super easy to call so it presented the easiest option here. Tho I admit it's rather funky.

src/workerd/api/node/util.c++
jasnell: Yeah this felt a bit funky but the reason (admittedly not very good) is that `tracer.addException(...)` is really set up for us to extract the `name`, `message`, and `stack` (most importantly the stack)... the stack being the part that is most important to send to the tail worker. I also wanted to avoid the additional overhead of creating the JS error in the case tracing is not being used. I don't feel strongly about how any of this bit is implemented tho.

src/workerd/api/node/util.c++
jasnell: What do you suggest as an alternative?

jasnell: The one idea I've come up with is checking immediately before the `script terminated for unknown reasons` assert to see if the `abortFulfiller` is still waiting. If it is not, then we assume that the script terminated because the IoContext was aborted and we emit a `NOSENTRY` error instead.

pkiv:
We're really excited about this one! Biggest issue for us is that there are dependencies that have process.exit, that will not be called in the production code path. But because of build time checks, we can't actually use those dependencies. So this will be a huge unblock. Thanks everyone!

anonrig:
> We're really excited about this one! Biggest issue for us is that there are dependencies that have process.exit, that will not be called in the production code path. But because of build time checks, we can't actually use those dependencies. So this will be a huge unblock. Thanks everyone!

@pkiv if there is any other blockers for you, please open an issue at the workerd repository. we'll like to take a look!

src/workerd/api/node/util.c++
jasnell: Actually, would it actually? The `process.exit(...)` impl does not actually throw a `jsg::JsExceptionThrown` in this case that I can see. I'm not getting that log in my testing on this.

src/workerd/api/node/util.c++
jasnell: Added a comment.

src/workerd/api/node/util.c++
jasnell: In any case, let's go with this and if it ends up being a problem in sentry we can revisit. Use of `process.exit(...)` *should* be limited and I kind of do want to know if folks are using it.

jasnell:
>  think we should start charging customers $10 each time they call that process.exit

Feel free to direct all payments into my venmo account.

src/workerd/api/node/util.c++
kentonv: > The one idea I've come up with is checking immediately before the script terminated for unknown reasons assert to see if the abortFulfiller is still waiting. If it is not, then we assume that the script terminated because the IoContext was aborted and we emit a NOSENTRY error instead.

That sounds fine. We can also rethrow the abort exception from runImpl() in this case.

Don't log NOSETRY, though. Just don't log anything, because there was no error. (If you separately want to log NOSENTRY to find out if people are using process.exit(), then do that in process.exit()'s own implementation... but I don't like that kind of logging, I worry it'll never get cleaned up and will just waste log space.)

> The process.exit(...) impl does not actually throw a jsg::JsExceptionThrown in this case that I can see. I'm not getting that log in my testing on this.

Err, is it actually terminating execution or not? Hard to see how it wouldn't end up throwing if it's correctly terminating execution...

kentonv:
We should add a test for behavior in Durable Objects. The behavior should be that it restarts the object -- same as `DurableObjectState::abort()` would.

src/workerd/api/node/util.c++
```diff
@@ -240,4 +241,43 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
...
```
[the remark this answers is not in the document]
jasnell: `process.exit(...)` is not expected to throw normally and it should not be possible to catch and handle an error here.

jasnell:
DO test added

src/workerd/api/node/util.c++
jasnell: Added the jsg::check and logic to verify that that execution is terminating. Still not actually seeing the "script terminated for unknown reasons" log in the tests.

kentonv: What if you KJ_DBG() next to the line where that message is thrown?

I am guessing that the error is being thrown but the abort error is winning the race out of WorkerEntrypoint so the "script terminated for unknown reasons" error is discarded. But that might not be reliably true...

jasnell: Added the additional check here. Haven't been able to trigger this assert, however, so unsure how necessary it is.

src/workerd/io/io-context.c++
```diff
@@ -1089,6 +1089,16 @@ void IoContext::runImpl(Runnable& runnable,
...
         limitEnforcer->requireLimitsNotExceeded();
 
+        // If we were terminated because abort() was called, then it's not an unknown
+        // reason...
+        if (!abortFulfiller->isWaiting()) {
+          // Nothing to do here. Do not log anything. The assumption is that we've
+          // terminated because the IoContext was aborted and isolate->TerminateExection()
+          // was called (likely because of someone using process.exit(...) in Node.js
+          // compat mode).
+          return;
```
kentonv, line 1099: You can't just return here. The caller may assume that this means the call actually completed successfully. This could, for example, cause an alarm handler to think that the alarm finished successfully and therefore doesn't need to be retried. This could lead to data loss.

You can either:
a. Return `abortPromise.addBranch()`, to propagate the abort exception.
b. Return `kj::NEVER_DONE` and count on the caller joining with the abort exception.

I think (a) is better.

irvinebroque:
This landed in https://github.com/cloudflare/workerd/releases/tag/v1.20241018.0 but isn't yet available to use on latest Wrangler (3.83.0) — unenv throws an unimplemented error.

I think we have added to the runtime but until unenv is updated, this is not actually usable in practice. Verified by going to the Quick Editor in the dash — because there is no polyfilling there, you can see this working.

vicb:
> This landed in https://github.com/cloudflare/workerd/releases/tag/v1.20241018.0 but isn't yet available to use on latest Wrangler (3.83.0) — unenv throws an unimplemented error.
> 
> I think we have added to the runtime but until unenv is updated, this is not actually usable in practice. Verified by going to the Quick Editor in the dash — because there is no polyfilling there, you can see this working.

The [PR](https://github.com/cloudflare/workers-sdk/pull/7080) should be in the next release on Tuesday

PR #2869, opened 2024-10-09
More refactoring to enable Python pools

hoodmane:
We won't be able to import `UnsafeEval` directly from `builtin_wrappers.ts` so this adds a `setUnsafeEval` function to inject it from the outside.


AdityaAtulTewari (approved):
LGTM

PR #2872, opened 2024-10-09
Rename GetPrototype() uses to GetPrototypeV2()

jasnell:
This is expected to be relatively temporary. V8 is in the process of deprecating and removing the original `GetPrototype()` implementation. Once that original impl is removed, I believe the plan is to rename `GetPrototypeV2()` back to `GetPrototype()`, at which point we'll likely need to change these back. That's ok, we want to make sure there's no issue with updating to the new behavior well in advance of that.


erikcorry:
Apparently there is some subtle difference between the two calls if we are calling it on the global object.

I didn't understand that, but assuming that we understand that, LGTM.

anonrig (approved):
looks good to me. i tried to do a similar research to address @erikcorry's comments, but if we are in par, let's go!

PR #2873, opened 2024-10-09
move thenables to coroutines on worker-interface

anonrig:
Saw an internal initiative that was abandoned, which moves several `.then` and `.catch` statements to coroutines. Let's make a small change on `worker-interface.c++`


npaun (approved):

jasnell (approved):

anonrig:
Breaking downstream.

PR #2875, opened 2024-10-09
enable linter and simplify stream/consumers.js

anonrig:
Enablers linter and simplifies implementation by using `Array.fromAsync()`

### DIFF
--- a/src/node/stream/consumers.js
+++ b/src/node/stream/consumers.js
@@ -23,13 +23,10 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-/* todo: the following is adopted code, enabling linting one day */
 import { Buffer } from 'node-internal:internal_buffer';
 
 export async function blob(stream) {
-  const chunks = [];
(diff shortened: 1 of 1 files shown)


npaun (approved):

src/node/stream/consumers.js
```diff
@@ -23,13 +23,10 @@
...
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-/* todo: the following is adopted code, enabling linting one day */
 import { Buffer } from 'node-internal:internal_buffer';
 
 export async function blob(stream) {
-  const chunks = [];
-  for await (const chunk of stream) chunks.push(chunk);
-  return new Blob(chunks);
+  return new Blob(await Array.fromAsync(stream));
```
jasnell, line 29: hmm... The `Array.fromAsync(...)` is still only Stage 3 in tc39 so there's a slight risk here. However, changing it back later if necessary would be easy so... ok :-)

jasnell (approved):

src/node/stream/consumers.js
anonrig: PS: Node.js has also started using it. https://github.com/search?q=repo%3Anodejs%2Fnode%20ArrayFromAsync&type=code

PR #2878, opened 2024-10-09
Update Workflow class type definition to accept a single event

sidharthachatterjee:
We previously passed in an Array of events but have since decided to switch to a single event per Workflow instance. This updates the types.


a-robinson (approved):
🎉

PR #2881, opened 2024-10-09
Use name instead of id for instances in Workflows binding

sidharthachatterjee:
This is a breaking chance but the binding is currently experimental so that is okay 

The Workflow binding currently accepts an `id` for its create and get functions. We now allow users to pass in a user defined "instance name". This can be used to lookup instances that were previously created.

### DIFF
--- a/types/defines/workflows.d.ts
+++ b/types/defines/workflows.d.ts
@@ -15,18 +15,18 @@ declare module "cloudflare:workflows" {
 declare abstract class Workflow {
   /**
    * Get a handle to an existing instance of the Workflow.
-   * @param id Id for the instance of this Workflow
+   * @param name Name of the instance of this Workflow
    * @returns A promise that resolves with a handle for the Instance
    */
-  public get(id: string): Promise<Instance>;
+  public get(name: string): Promise<Instance>;
 
   /**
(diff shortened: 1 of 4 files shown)


jasnell (approved):

types/defines/workflows.d.ts
```diff
@@ -15,18 +15,18 @@ declare module "cloudflare:workflows" {
...
   /**
-   * Create a new instance and return a handle to it. If a provided id exists, an error will be thrown.
-   * @param id Id to create the instance of this Workflow with
+   * Create a new instance and return a handle to it. If a provided instance name exists, an error will be thrown.
+   * @param name Name to create the instance of this Workflow with
    * @param params The payload to send over to this instance
    * @returns A promise that resolves with a handle for the Instance
    */
-  public create(id: string, params: object): Promise<Instance>;
+  public create(name: string, params: object): Promise<Instance>;
```
LuisDuarte1, line 29: I think it would be better to create a `CreateInstanceParams` interface, to allow for better extensibility (we are making breaking changes anyway)

LuisDuarte1, line 29: `params` should be unknown and not an objext

LuisDuarte1 (changes requested):

LuisDuarte1 (approved):

PR #2883, opened 2024-10-09
Restore: Change WorkerInterface so prewarm returns a promise.

danlapid:
Change WorkerInterface so prewarm returns a promise.

### DIFF
--- a/src/workerd/io/worker-interface.c++
+++ b/src/workerd/io/worker-interface.c++
@@ -16,10 +16,8 @@ namespace {
 // interface the promise resolved to.
 class PromisedWorkerInterface final: public kj::Refcounted, public WorkerInterface {
 public:
-  PromisedWorkerInterface(
-      kj::TaskSet& waitUntilTasks, kj::Promise<kj::Own<WorkerInterface>> promise)
-      : waitUntilTasks(waitUntilTasks),
-        promise(promise.then([this](kj::Own<WorkerInterface> result) { worker = kj::mv(result); })
+  PromisedWorkerInterface(kj::Promise<kj::Own<WorkerInterface>> promise)
+      : promise(promise.then([this](kj::Own<WorkerInterface> result) { worker = kj::mv(result); })
                     .fork()) {}
 
   kj::Promise<void> request(kj::HttpMethod method,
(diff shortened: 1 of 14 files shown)


danlapid:
Note: I have put all changes from the original PR in a separate commit so you don't have to review the first commit again.

harrishancock (approved):

src/workerd/io/worker-interface.c++
```diff
@@ -76,21 +76,21 @@ public:
...
+      co_return co_await w->customEvent(kj::mv(event));
     } else {
       co_await promise;
       co_return co_await KJ_ASSERT_NONNULL(worker)->customEvent(kj::mv(event));
     }
   }
 
 private:
-  kj::ForkedPromise<void> promise;
+  kj::Promise<void> promise;
```
kentonv, line 87: This seems unsafe if more than one method is called?

It's often the case that a `WorkerInterface` object is used to deliver only one event, but not always.

danlapid: Well even if it's called multiple times it shouldn't be called simultaneously from multiple threads which means that the first execution will still resolve the promise and set worker after which all subsequent executions will not try to await the promise.
This isn't really all too different to the original solution as we didn't addBranch on every infocation.

danlapid: The only usage for this ForkedPromise was that we had branched in the case of prewarm because that handling was done in a waitUntilTask and by the time it ran the promise might have already been resolved.
I removed that part of the code when switching prewarm to return a promise so it's no longer required.

kentonv: You could get two overlapping calls in a single thread. These returns promises after all -- the caller could call another method without first waiting for the first promise to complete.

> This isn't really all too different to the original solution as we didn't addBranch on every invocation.

`co_await forkedPromise` implicitly waits on a branch, hence doesn't consume the promise. `co_await promise`, however, consumes the promise.

danlapid: I see, makes sense now, I'll revert this change, thanks for the explanation 😄

PR #2885, opened 2024-10-10
SQLite-backed DOs: check actor brokenness before starting implicit transactions

jclee:


src/workerd/io/actor-sqlite.c++
```diff
@@ -198,6 +198,7 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
 
 void ActorSqlite::onWrite() {
   if (currentTxn.is<NoTxn>()) {
+    requireNotBroken();
```
kentonv, line 201: Let's move this above the `if()`, so that all writes will fail when broken.

jclee:
Tried using the `fixup.sh` script, but it pushed a rebase I didn't expect...  At any rate, here were the changes in its output:
* 87bd15f2
* 64ff1c5b

src/workerd/io/actor-sqlite.c++
jclee: I moved the line, and also added a rudimentary test.

kentonv (approved):

jclee:
Also adding an extra check for brokenness to the `getSqliteDatabase()` function, to reduce the possibility of read-only queries succeeding against a broken actor.

jclee:
Also fixing lint.

jclee:
Dropped the check for brokenness in the getSqliteDatabase() function... it causes an internal test to hang in a way I don't fully understand yet.  I probably just need to update the test to match, but want to get this in first.

PR #2886, opened 2024-10-10
Esbuild bundle the Python pool setup code

hoodmane:
We want to execute a single standalone JavaScript file in the pool scope. It should work in a normal v8 isolate with no modifications. This adds the esbuild bundle step, but keeps everything else working the same.


danlapid:
Tagging @fhanau for Bazel changes

danlapid (approved):

WORKSPACE
```diff
@@ -274,6 +274,13 @@ http_archive(
     url = "https://github.com/aspect-build/rules_ts/releases/download/v3.0.0/rules_ts-v3.0.0.tar.gz",
 )
 
+http_archive(
```
fhanau, line 277: - [ ] Since we recently added update-deps.py for workerd, we should be using that for all new dependencies

hoodmane: Great to see a reduction in the number of differences between developing with workerd vs internally. How's it look now?

fhanau: Thank you, LGTM! Didn't look closely at js_file but this should be good to go based on Dan's review

hoodmane: Well Dan wrote `js_file.bzl` =) I should probably add him as a coauthor on the commit.

PR #2887, opened 2024-10-10
Rename Workflow to WorkflowEntrypoint

sidharthachatterjee:
`Workflow` conflicts with the binding class which is also called `Workflow`

Since we have `WorkerEntrypoint`, `WorkflowEntrypoint` feels like a much more appropriate and descriptive name of the base class

This is of course a breaking change, but these are all scoped under `experimental` for now so that is okay

### DIFF
--- a/src/cloudflare/internal/workers.d.ts
+++ b/src/cloudflare/internal/workers.d.ts
@@ -12,7 +12,7 @@ export class WorkerEntrypoint {
   public env: unknown;
 }
 
-export class Workflow {
+export class WorkflowEntrypoint {
   public constructor(ctx: unknown, env: unknown);
 
   public ctx: unknown;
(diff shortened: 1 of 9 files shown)


sidharthachatterjee:
I think this doesn't require an internal PR anymore because I'm only seeing references to workerd submodule but I'd like one of us to confirm

LuisDuarte1 (approved):

fhanau:
> I think this doesn't require an internal PR anymore because I'm only seeing references to workerd submodule but I'd like one of us to confirm

Can you elaborate what you mean here? For the downstream project, workerd is being used the same way as before (as a submodule) so there should be an internal PR for this.

sidharthachatterjee:
> > I think this doesn't require an internal PR anymore because I'm only seeing references to workerd submodule but I'd like one of us to confirm
> 
> Can you elaborate what you mean here? For the downstream project, workerd is being used the same way as before (as a submodule) so there should be an internal PR for this.

The file that I saw references to `entrypointClasses` doesn't seem to exist anymore in the downstream project

sidharthachatterjee:
Never mind, I was looking at the wrong base branch @fhanau Internal PR is now ready as well and waiting on CI

a-robinson (approved):

a-robinson (approved):
LGTM but will wait on internal CI before merging.

PR #2892, opened 2024-10-10
protobuf upgrade

mikea:


build/deps/build_deps.jsonc
```diff
@@ -78,6 +78,22 @@
...
+      "name": "aspect_rules_esbuild",
+      "type": "github_release",
+      "owner": "aspect-build",
+      "repo": "rules_esbuild",
+      "file_regex": "\\.tar\\.gz$",
+      // All releases are marked as prerelease, so github_last_release() fails.
+      "freeze_version": "v0.21.0"
+    },
+    {
+      "name": "rules_pkg",
```
fhanau, line 92: Is rules_pkg needed by protobuf now?

build/deps/gen/dep_com_google_protobuf.bzl
```diff
@@ -0,0 +1,18 @@
+# WARNING: THIS FILE IS AUTOGENERATED BY update-deps.py DO NOT EDIT
+
+load("@//:build/http.bzl", "http_archive")
+
+TAG_NAME = "v28.2"
```
fhanau, line 5: This puts us far ahead of the protobuf version used in perfetto, V8 and the downstream repo, but the build seems to work so maybe that's just fine?

fhanau (approved):

build/deps/build_deps.jsonc
mikea: yes

build/deps/gen/dep_com_google_protobuf.bzl
mikea: I guess so. Looks like it is version-compatible?

anonrig (approved):

PR #2894, opened 2024-10-10
[build] Update package versions to drop bazel-lib patch, convert more deps to update-deps

fhanau:
- Drop unused rules_license


mikea (approved):
thanks

PR #2896, opened 2024-10-10
Make callback the last arg for Workflow do

sidharthachatterjee:
The callback is now always the last argument with an optional `config` as second argument. This allows users to do: 

```
step.do('does not truncate', async () => {
    return event.payload.char.repeat(900);
}),

step.do(
    'does truncate',
    {
        retries: {
            limit: 5,
            backoff: 'constant',
            delay: '1 second',
        },
    },
    async () => {
        return event.payload.char.repeat(2000);
    }
),
```

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -225,11 +225,22 @@ declare module "cloudflare:workers" {
   };
 
   export type WorkflowStep = {
-    do: <T extends Rpc.Serializable>(
-      name: string,
-      callback: () => Promise<T>,
-      config?: WorkflowStepConfig
-    ) => Promise<T>;
+    do:
+      | (<T extends Rpc.Serializable>(
+          name: string,
+          callback: () => Promise<T>
+        ) => Promise<T>)
+      | (<T extends Rpc.Serializable>(
+          name: string,
+          config: WorkflowStepConfig,
+          callback: () => Promise<T>
+        ) => Promise<T>)
+      | (<T extends Rpc.Serializable>(
+          name: string,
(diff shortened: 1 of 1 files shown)


types/defines/rpc.d.ts
```diff
@@ -225,11 +225,22 @@ declare module "cloudflare:workers" {
...
+          callback: () => Promise<T>
+        ) => Promise<T>)
+      | (<T extends Rpc.Serializable>(
+          name: string,
+          config: WorkflowStepConfig,
+          callback: () => Promise<T>
+        ) => Promise<T>)
+      | (<T extends Rpc.Serializable>(
+          name: string,
+          config: WorkflowStepConfig | (() => Promise<T>),
```
a-robinson, line 240: I'm not understanding -- this third option here allows you to pass a function for `config` and for `callback`? It also makes `calback` optional, meaning that you could pass just a `string` and a `WorkflowStepConfig` with no third argument?

I'd have thought that only the first two type options were needed. What is this one doing?

sidharthachatterjee: I too was confused. See https://www.typescriptlang.org/docs/handbook/2/functions.html#overload-signatures-and-the-implementation-signature

a-robinson: My notes from our chat to summarize why this is needed:

> so the first two signatures are overload signatures and the latter is the implementation signature? I guess I'm not sure how those docs map to the fact that these are type overrides defined separately from the implementation

exactly

> and so even though the implementation is somewhere else, typescript gets mad without the third one?

indeed

a-robinson (approved):

PR #2898, opened 2024-10-10
[build] fix lld linker path

fhanau:
This appears to be needed for the clang driver to find LLD when using older clang versions.

=============

This (hopefully) fixes macOS CI release builds.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -320,7 +320,7 @@ build:release_macos --linkopt="-Wl,-no_function_starts"
 # We could use Xcode 16's -Wl,-deduplicate option instead, but LLD's ICF appears to be superior.
 # We also want to enable ICF for Linux, but there it causes warnings when dynamically linking with
 # libc++.
-build:macos_lld --linkopt="-fuse-ld=/opt/homebrew/bin/ld64.lld"
+build:macos_lld -fuse-ld=lld --ld-path=ld64.lld
 build:macos_lld_icf --config=macos_lld
 build:macos_lld_icf --linkopt="-Wl,--icf=safe"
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -320,7 +320,7 @@ build:release_macos --linkopt="-Wl,-no_function_starts"
 # We could use Xcode 16's -Wl,-deduplicate option instead, but LLD's ICF appears to be superior.
 # We also want to enable ICF for Linux, but there it causes warnings when dynamically linking with
 # libc++.
-build:macos_lld --linkopt="-fuse-ld=/opt/homebrew/bin/ld64.lld"
+build:macos_lld --linkopt=-fuse-ld=lld --linkopt=--ld-path=/opt/homebrew/bin/ld64.lld
```
anonrig, line 323: Q: Is there an assertion in anywhere in the build pipeline so that homebrew is installed and these expectations (such as `/homebrew/bin/ld64.lld` are available?

fhanau: With #2884, `macos_lld` is a build configuration we opt in to in the release job, and we run `brew install lld` before that. Homebrew itself is always installed in the image.
Now that I think about it the `/opt/homebrew/bin/ld64.lld` requirement is precisely why this is actually failing – the homebrew binary paths are different on x86 and Apple Silicon, and since the release job is on x86 we'll need different paths here.

mikea (approved):

.bazelrc
fhanau: Figured out a way to not use an absolute path – glad there's a platform-independent option!

Issue #2516, opened 2024-08-11
🐛 Bug Report — Request Throws 'Invalid HTTP method' Error when Using Lowercase Method "patch"


Kynson:
### The Issue
`new Request('', {method: 'patch'})` unexpectedly throws error 'Invalid HTTP method string: patch' but `new Request('', {method: 'PATCH'})` works fine.

Minimal Reproducible Example: [Workers Playground](https://workers.cloudflare.com/playground#LYVwNgLglgDghgJwgegGYHsHALQBM4RwDcABAEbogB2+CAngLzbPYDqApmQNJQQBimYACFKNRHWABzdABYo6AFoQAygCYAXghEB2AAoAuFiw7deArCOq0J0uYpUat6PQFgAUAGF0VCOx-YAESgAZxh0YN55Kn0SDCw8AmISKjhgdgYAIigadgAPADoAK2CM0lQoMD9U9KycguKM9y8fPwhsABU6GHYYuBgYMCgAYwIo5EK4ADc4YKGEWAgAamB0XHB2d3dkACpttxJtkgABCC72XHZUEgBvAHkyQvYhiABfEgBRKkn97eQ3NzyYSQJAuqDg4AgNzcAEgdntoQdjvAEKkbgAldgARxA7GCrxICCxOLxMMRR2RqOun0mbz83wRh3JiEp71yTxA0G8zV8uXxz1ypMZhIgIAQVGCN10CHQwBC7AAPBjQt5guwAHwvQV-aEzOhUIaxdgQIYACwAFITsbiIAAaEh0u38gCUUOh0Ko7AA7iQMVa8WaAOQBu3XNIQE2rGIB+DGk0Bl5OmHQ4WiqjJL0+3FhcXsQO+PEBp1EGEvG1uF7F9xGZgmHj8QSWMT0KSyeRKNSaHS6JreXz+ILKiKc6KxQQJQikFJpTJhuD4QilWIVKrTjJkMDoMiNTy91odM69fqDEbD8bBbybNzXDJDGUxqBkCq8OgAfXn7Ay+gyqgADKoZNgP4AMyAdoGQ2jed6jI+gynC+qBgHAkglPoADaGRUKs7DFC+t7ADGGQALoQcAcDZC+KxrJUn61Bc9QlC8VbVrWZgNqI1gtnY7aOF2zDuEAA)

### Expected Behaviour
Both lowercase and uppercase method string should be supported (other methods support both lowercase and uppercase). 

[...]

PR #2913, opened 2024-10-11
Upper case all http methods in fetch API with compat flag

jasnell:
This introduces a non-standard behavior that would become the default. Please review carefully. See the original bug report for context (linked below).

Fixes: https://github.com/cloudflare/workerd/issues/2516

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1004,9 +1004,14 @@ jsg::Ref<Request> Request::constructor(
         KJ_IF_SOME(m, initDict.method) {
           KJ_IF_SOME(code, tryParseHttpMethod(m)) {
             method = code;
-          } else {
-            KJ_IF_SOME(code, kj::tryParseHttpMethod(toUpper(m))) {
-              method = code;
+          } else KJ_IF_SOME(code, kj::tryParseHttpMethod(toUpper(m))) {
+            method = code;
+            if (!FeatureFlags::get(js).getUpperCaseAllHttpMethods()) {
+              // This is actually the spec defined behavior. We're expected to only
+              // upper case get, post, put, delete, head, and options per the spec.
+              // Other methods, even if they would be recognized if they were uppercased,
+              // are supposed to be rejected.
+              // Refs: https://fetch.spec.whatwg.org/#methods
               switch (method) {
                 case kj::HttpMethod::GET:
                 case kj::HttpMethod::POST:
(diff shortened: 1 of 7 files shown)


src/workerd/api/http.c++
```diff
@@ -1002,12 +1002,15 @@ jsg::Ref<Request> Request::constructor(
...
-          auto originalMethod = kj::str(m);
           KJ_IF_SOME(code, tryParseHttpMethod(m)) {
             method = code;
-          } else {
-            KJ_IF_SOME(code, kj::tryParseHttpMethod(toUpper(kj::mv(m)))) {
-              method = code;
+          } else KJ_IF_SOME(code, kj::tryParseHttpMethod(toUpper(m))) {
+            method = code;
+            if (!FeatureFlags::get(js).getUpperCaseAllHttpMethods()) {
+              // This is actually the spec defined behavior. We're expected to only
```
anonrig, line 1010: Would you mind adding a link to the spec in here?

anonrig (approved):

src/workerd/api/http.c++
```diff
@@ -1002,12 +1002,15 @@ jsg::Ref<Request> Request::constructor(
...
+              // upper case get, post, put, delete, head, and options per the spec.
+              // Other methods, even if they would be recognized if they were uppercased,
+              // are supposed to be rejected.
```
jasnell, line 1013:
```suggestion
              // are supposed to be rejected.
              // Refs: https://fetch.spec.whatwg.org/#methods
```

src/workerd/api/http.c++
jasnell: Done

kentonv:
Up to you but personally I'd be in favor of not bothering with a compat flag here and just making the change for everyone. It only turns an error case into a non-error case and I can't imagine why anyone would be probing for this error.

src/workerd/api/http.c++
```diff
@@ -1017,12 +1021,11 @@ jsg::Ref<Request> Request::constructor(
                 case kj::HttpMethod::OPTIONS:
                   break;
                 default:
-                  JSG_FAIL_REQUIRE(
-                      TypeError, kj::str("Invalid HTTP method string: ", originalMethod));
+                  JSG_FAIL_REQUIRE(TypeError, kj::str("Invalid HTTP method string: ", m));
```
kentonv, line 1024: Isn't this now logging `m` after it has been moved away?

Below, too.

kentonv: Wait never mind, it's not moved away anymore.

jasnell:
I'd prefer the compat flag if only to provide a means of reverting back to the more strict spec-compliance mode. I'm not a fan of intentionally moving away from the spec requirements without at least providing a safety hatch.

PR #2914, opened 2024-10-11
Provide ExecutionModel enum to tail worker

fhanau:


fhanau:
- [ ] isActor => isDurableObject

src/workerd/io/trace.c++
```diff
@@ -260,13 +260,15 @@ Trace::Trace(kj::Maybe<kj::String> stableId,
     kj::Maybe<kj::String> scriptName,
     kj::Maybe<kj::Own<ScriptVersion::Reader>> scriptVersion,
     kj::Maybe<kj::String> dispatchNamespace,
+    bool isActor,
```
jasnell, line 263: Really not a fan of boolean arguments like this. Would prefer an enum.

Also, would prefer adding this to the end of the arg list.

kentonv: Hmm, in fact, perhaps the API we expose should be an enum instead of a boolean... like:

```
executionModel: "stateless" | "durableObject" | "workflow";
```

fhanau: Changing to an enum is something I can do. What is the meaning of stateless and workflow in this context?

kentonv: "stateless" would be a regular worker (i.e. not a DO). "workflow" would be the new Workflows product (not launched yet). Workflows are a different execution model from both regular workers and DOs. They are more like "Durable Procedures". cc @sidharthachatterjee

src/workerd/api/trace.c++
```diff
@@ -284,6 +285,10 @@ kj::StringPtr TraceItem::getOutcome() {
   return outcome;
 }
 
+bool TraceItem::getIsActor() {
```
anonrig, line 288:
```suggestion
bool TraceItem::getIsActor() const {
```

src/workerd/api/trace.h
```diff
@@ -105,6 +105,7 @@ class TraceItem final: public jsg::Object {
 
   uint getCpuTime();
   uint getWallTime();
+  bool getIsActor();
```
anonrig, line 108:
```suggestion
  bool getIsActor() const;
```

src/workerd/api/trace.c++
fhanau: I agree that we should be doing this, but this is best done as part of a separate cleanup that applies this for all functions in Trace. Let's keep this PR self-contained if possible.

src/workerd/api/trace.c++
```diff
@@ -116,10 +116,11 @@ jsg::Optional<kj::Array<kj::String>> getTraceScriptTags(const Trace& trace) {
...
-kj::String getTraceOutcome(const Trace& trace) {
-  // TODO(cleanup): Add to enumToStr() to capnp?
-  auto enums = capnp::Schema::from<EventOutcome>().getEnumerants();
-  uint i = static_cast<uint>(trace.outcome);
+template <typename Enum>
+kj::String enumToStr(const Enum& var) {
+  // TODO(cleanup): Port this to capnproto.
+  auto enums = capnp::Schema::from<Enum>().getEnumerants();
+  uint i = static_cast<uint>(var);
   KJ_ASSERT(i < enums.size(), "invalid outcome");
```
jasnell, line 124: Should update this error message since it's no longer specific to `outcome`

src/workerd/io/worker-interface.capnp
```diff
@@ -136,6 +136,14 @@ struct Trace @0x8e8d911203762d34 {
...
   }
 
+  enum ExecutionModel {
+    stateless @0;
+    durableObject @1;
+    workflow @2;
+  }
+  executionModel @25 :ExecutionModel;
+  # the execution model of the worker being traced. Can be stateless for a regular worker,
+  # durableObject for a DO worker or workflow for the upcoming Workflows.
```
jasnell, line 146: Nit: can this be moved after `truncated` since it is using a higher ordinal value? It's already difficult enough to make sure we're not duplicating ordinals, having them out of order doesn't help ;-)

jasnell (approved):
This'll cause a gnarly merge conflict with my initial streaming trace PR but I guess thats ok ;-) .... LGTM

fhanau:
> This'll cause a gnarly merge conflict with my initial streaming trace PR but I guess thats ok ;-) .... LGTM

Thank you!

PR #2916, opened 2024-10-14
Fix: change WorkflowStep into an abstract class

LuisDuarte1:
For some reason, Typescript resolves types in a different away compared to abstract classes. As a consequence, it would not apply the overloading properly and always treats `WorkflowStep.do` as a 3 argument function

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -224,26 +224,12 @@ declare module "cloudflare:workers" {
     timestamp: Date;
   };
 
-  export type WorkflowStep = {
-    do:
-      | (<T extends Rpc.Serializable>(
-          name: string,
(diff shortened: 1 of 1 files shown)


a-robinson (approved):
LGTM assuming this works. Have you tried generating the types with this and examined them or is it more of a guess-and-check situation?

sidharthachatterjee:
> LGTM assuming this works. Have you tried generating the types with this and examined them or is it more of a guess-and-check situation?

Good question. To avoid thrashing, let's generate these and test locally on smoke tests @LuisDuarte1

a-robinson:
Then no worries here either, but it looks like it can't be merged yet because "This branch has conflicts that must be resolved"

jasnell (approved):

PR #2917, opened 2024-10-14
add node:path win32 implementations

anonrig:
Implement win32 implementations of `node:path` and enables linter.


anonrig:
> My only concern here is why... do we have evidence that folks are actually trying to use this and want the `path.win32` semantics in the runtime? Not blocking but I'm not convinced this is worthwhile having in the runtime vs. using the polyfill.

@jasnell Polyfill doesn't add these functions. We can add them to polyfill, but then `path` module will become a hybrid module. That's what I was trying to avoid. Ref: https://github.com/unjs/unenv/blob/main/src/presets/cloudflare.ts#L16

jasnell (approved):
Code LGTM. Not convinced it's super necessary but no reason to block

PR #2923, opened 2024-10-14
Add waitForBookmark to DurableObjectStorage

justin-mp:
This exposes the internal `waitForBookmark` method to JavaScript. We’ll need this in order to implement decent consistency models with SQLite replication.

### DIFF
--- a/src/workerd/api/actor-state.c++
+++ b/src/workerd/api/actor-state.c++
@@ -735,6 +735,10 @@ kj::Promise<kj::String> DurableObjectStorage::onNextSessionRestoreBookmark(kj::S
   return cache->onNextSessionRestoreBookmark(bookmark);
 }
 
+kj::Promise<void> DurableObjectStorage::waitForBookmark(kj::String bookmark) {
+  return cache->waitForBookmark(bookmark);
+}
+
 ActorCacheOps& DurableObjectTransaction::getCache(OpName op) {
   JSG_REQUIRE(!rolledBack, Error, kj::str("Cannot ", op, " on rolled back transaction"));
   auto& result = *JSG_REQUIRE_NONNULL(cacheTxn, Error,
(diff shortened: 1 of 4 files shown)


justin-mp:
https://github.com/cloudflare/workerd/pull/2923/commits/59ce3bdf603f069454a8cca75729dfe28261531a is a rebase on master.

src/workerd/api/actor-state.c++
```diff
@@ -735,6 +735,10 @@ kj::Promise<kj::String> DurableObjectStorage::onNextSessionRestoreBookmark(kj::S
   return cache->onNextSessionRestoreBookmark(bookmark);
 }
 
+kj::Promise<void> DurableObjectStorage::waitForBookmark(kj::String bookmark) {
+  return cache->waitForBookmark(bookmark);
```
jasnell, line 739: Can this safely be called multiple times from javascript? As it is here, every call on this will result in a new JavaScript promise backed by a kj::Promise... specifically, if I do..

```js
const p1 = whatever.waitForBookmark('...');
const p2 = whatever.waitForBookmark('...');
```

Then `p1 !== p2`

Should this always return the same JS promise when the bookmark id is the same and is already being waited on?

justin-mp: The internal PR waits on the same forked promise, so in production this will wait on the same thing.  `p1` will still be `!== p2`, is that a problem?

jasnell: Not a problem just something to be aware of. Creating multiple promises that wait on the same thing can add performance cost.

jasnell (approved):

src/workerd/api/actor-state.c++
kentonv: I think this is fine as-is. Don't think people are likely to call it multiple times in a row like that.

kentonv (approved):

justin-mp:
https://github.com/cloudflare/workerd/compare/59ce3bdf603f069454a8cca75729dfe28261531a..4d5c89c41bfa35525b89f1ebb996cf1c344ce556 is a rebase on main.

PR #2924, opened 2024-10-15
add process.platform

anonrig:
Adds process.platform. This can help us test OS specific code. For example: `node:url`

### DIFF
--- a/src/workerd/api/node/tests/process-nodejs-test.js
+++ b/src/workerd/api/node/tests/process-nodejs-test.js
@@ -1,0 +1,8 @@
+import assert from 'node:assert';
+
+export const processPlatform = {
+  test() {
+    assert.strictEqual(typeof process.platform, 'string');
+    assert.ok(['darwin', 'win32', 'linux'].includes(process.platform));
+  },
+};
--- a/src/workerd/api/node/util.h
+++ b/src/workerd/api/node/util.h
@@ -161,6 +161,16 @@ public:
 
   jsg::Name getResourceTypeInspect(jsg::Lock& js);
 
+#ifdef _WIN32
+  static constexpr kj::StringPtr processPlatform = "win32"_kj;
+#elif defined(__linux__)
+  static constexpr kj::StringPtr processPlatform = "linux"_kj;
+#elif defined(__APPLE__)
+  static constexpr kj::StringPtr processPlatform = "darwin"_kj;
+#else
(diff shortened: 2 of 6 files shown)


src/workerd/api/node/util.c++
```diff
@@ -289,4 +289,16 @@ void UtilModule::processExitImpl(jsg::Lock& js, int code) {
   js.logWarning(kj::str(err.get(js, "stack"_kj)));
 }
 
+kj::String UtilModule::getProcessPlatform() const {
+#ifdef _WIN32
+  return kj::str("win32");
+#elif defined(__linux__)
```
jasnell, line 295: It probably ought to be noted that in the production environment this will always return `linux` ... my fear is that by providing this we might accidentally send the message to users that the value in production might be `win32` or `darwin` or that they will be able to write code in workerd that only ever runs in workerd. Not blocking but I think it's worth calling out in documentation.

src/workerd/api/node/util.c++
```diff
@@ -289,4 +289,16 @@ void UtilModule::processExitImpl(jsg::Lock& js, int code) {
   js.logWarning(kj::str(err.get(js, "stack"_kj)));
 }
 
+kj::String UtilModule::getProcessPlatform() const {
```
jasnell, line 292: Let's have this return `kj::StringPtr` and have these values be literal constants to avoid the extraneous allocation on every call.

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -0,0 +1,7 @@
+import assert from 'node:assert';
+
+export const processPlatform = {
+  test() {
+    assert.strictEqual(typeof process.platform, 'string');
```
jasnell, line 5: nit: probably should test for one of the specific known values.

src/workerd/api/node/util.h
```diff
@@ -251,6 +267,7 @@ class UtilModule final: public jsg::Object {
 
     JSG_METHOD(getBuiltinModule);
     JSG_METHOD(processExitImpl);
+    JSG_READONLY_INSTANCE_PROPERTY(processPlatform, getProcessPlatform);
```
jasnell, line 270: One last nit.. since the value wouldn't be expected to change, let's make this `JSG_LAZY_READONLY_INSTANCE_PROPERTY` so the getter is only ever called once.

PR #2929, opened 2024-10-15
update AI models types

akazwz:
update AI models types


Cherry:
Could we get this merged please? Workers AI types are extremely out of date now, with how fast the product moves.

G4brym:
This pr appears to be out of date, as the these types are already present in the main branch:
https://github.com/cloudflare/workerd/blob/main/types/defines/ai.d.ts#L676

PR #2931, opened 2024-10-15
Remove direct eval in pyodide.asm.js and minify setupEmscripten.

danlapid:
This replaces direct eval in pyodide.asm.js with indirect eval, see comment in the commit. This also minifies setupEmscripten.
This should fix build warnings.


hoodmane:
These direct eval code paths are never actually executed (they'll fail since eval is off) so if possible we could just disable the warning.

danlapid:
> These direct eval code paths are never actually executed (they'll fail since eval is off) so if possible we could just disable the warning.

They actually do get executed and and the tests fail without them

PR #2932, opened 2024-10-15
Make NonRetryableError non-abstract in types

LuisDuarte1:
Basically, we won't get NonRetryableError serialization done properly because we would have to implement our own Error Serializer for this which might not be worth it, at least for launch. Even if we fix this later, it's okay to let users throw a raw NonRetryableError anyway.

### DIFF
--- a/types/defines/workflows.d.ts
+++ b/types/defines/workflows.d.ts
@@ -1,13 +1,9 @@
-/**
- * NonRetryableError allows for a user to throw a fatal error
- * that makes a Workflow instance fail immediately without triggering a retry
- */
 declare module "cloudflare:workflows" {
-  export abstract class NonRetryableError extends Error {
-    /**
-     * `__brand` is used to differentiate between `NonRetryableError` and `Error`
-     * and is omitted from the constructor because users should not set it
-     */
+  /**
+   * NonRetryableError allows for a user to throw a fatal error
+   * that makes a Workflow instance fail immediately without triggering a retry
+   */
+  export class NonRetryableError extends Error {
     public constructor(message: string, name?: string);
   }
 }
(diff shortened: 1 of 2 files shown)


types/defines/workflows.d.ts
```diff
@@ -3,7 +3,7 @@
  * that makes a Workflow instance fail immediately without triggering a retry
  */
 declare module "cloudflare:workflows" {
-  export abstract class NonRetryableError extends Error {
+  export class NonRetryableError extends Error {
     /**
      * `__brand` is used to differentiate between `NonRetryableError` and `Error`
```
Skye-31, line 8: While this is getting updated, why is this a jsdoc comment? Users don't need to see it

LuisDuarte1: good point, i'll change it into a normal one

types/defines/workflows.d.ts
```diff
@@ -1,13 +1,11 @@
...
-    /**
-     * `__brand` is used to differentiate between `NonRetryableError` and `Error`
-     * and is omitted from the constructor because users should not set it
-     */
+  /**
+   * NonRetryableError allows for a user to throw a fatal error
+   * that makes a Workflow instance fail immediately without triggering a retry
+   */
+  export class NonRetryableError extends Error {
+    // `__brand` is used to differentiate between `NonRetryableError` and `Error`
```
a-robinson, line 7: What is `__brand` and why does it merit mentioning here?

a-robinson (approved):

types/defines/workflows.d.ts
jasnell: Things like js rpc that are based on v8 serialization do not currently handle custom class-based Error objects very well. With `Error` objects only the core properties like `name`, `message`, and `stack` are handled out of the box. I'm not sure if `__brand` should be mentioned here but I believe the original intent was for brand to be used to signal that this is not a regular error. Unfortunately, v8 won't serialize it either way.

LuisDuarte1: done

PR #2934, opened 2024-10-16
Remove direct eval in pyodide.asm.js

danlapid:
This replaces direct eval in pyodide.asm.js with indirect eval, see comment in the commit.


danlapid:
Problem with previous PR was actually the minify sadly.

hoodmane:
These `eval` calls should never be executed anyways.

hoodmane (approved):

PR #2936, opened 2024-10-16
Workerd side of python isolate pool

danlapid:


src/pyodide/BUILD.bazel
```diff
@@ -207,25 +209,25 @@ data = wd_ts_bundle_capnp(
...
     internal_modules = [
-        "generated/emscriptenSetup.js",
+        # "generated/emscriptenSetup.js",
     ] + glob(
         [
             "internal/*.ts",
             "internal/topLevelEntropy/*.ts",
             # The pool directory is only needed by typescript, it shouldn't be used at runtime.
-            "internal/pool/*.ts",
+            # "internal/pool/*.ts",
```
hoodmane, line 218: We should run `tsc --noEmit` on this directory to make sure that our types are correct. Probably alongside the esbuild command, since esbuild by itself ignores types.

src/pyodide/emscripten_setup.capnp
```diff
@@ -0,0 +1,13 @@
...
+  code @0 :Text;
+  pyodideAsmWasm @1 :Data;
+  pythonStdlibZip @2 :Data;
+}
+
+const emscriptenSetup :EmscriptenSetup = (
+  code = embed "emscriptenSetup.js",
+  pyodideAsmWasm = embed "pyodide.asm.wasm",
+  pythonStdlibZip = embed "python_stdlib.zip",
+);
```
hoodmane, line 13: It would be nice to put this and `pyodide.capnp` together into the same capnproto.bin file so we don't have to upload/download two separate things.

danlapid: The benefit of uploading two separate things is they are cached separately.
If we suspect this to change less often than the latter we'll get a big performance improvement.
This is a huge package (>10MB) while the other package is tiny.

src/pyodide/internal/pool/builtin_wrappers.ts
```diff
@@ -1,6 +1,7 @@
 import type { getRandomValues as getRandomValuesType } from 'pyodide-internal:topLevelEntropy/lib';
 import type { default as UnsafeEvalType } from 'internal:unsafe-eval';
 
+// TODO: export the following two functions to EmscriptenApi
```
hoodmane, line 4: Easiest way would be to import them in `emscriptenSetup.ts` and assign them to `emscriptenModule` at the end of `instantiateEmscriptenModule`.

src/pyodide/types/emscripten.d.ts
```diff
@@ -68,4 +68,8 @@ interface Module {
   addRunDependency(x: string): void;
   removeRunDependency(x: string): void;
   noInitialRun: boolean;
+  setUnsafeEval(mod: typeof import('internal:unsafe-eval').default): void;
```
danlapid, line 71: Type imports must be here instead of an import statement because we use this file as an "ambient" type declaration.
See https://stackoverflow.com/questions/39040108/import-class-in-definition-file-d-ts

src/workerd/api/BUILD.bazel
```diff
@@ -93,15 +96,19 @@ wd_cc_library(
     name = "pyodide",
     srcs = [
         "pyodide/pyodide.c++",
+        "pyodide/setup-emscripten.c++",
```
danlapid, line 99: I think we'll want to merge this file into pyodide.c++

src/pyodide/emscripten_setup.capnp
hoodmane: But then in order to get the performance improvement we'll need to make extra tooling to keep track of when the big package hasn't actually changed. If we try to do it manually we'll probably mess up.

danlapid: Not sure how that's different tooling to what we have now for the regular package

hoodmane: But I suppose if we upload separately, we can get that tooling working later.

src/workerd/api/pyodide/setup-emscripten.c++
```diff
@@ -0,0 +1,80 @@
...
+}
+
+jsg::JsValue resolvePromise(jsg::Lock& js, jsg::JsValue prom) {
+  auto promise = KJ_ASSERT_NONNULL(prom.tryCast<jsg::JsPromise>());
+  if (promise.state() == jsg::PromiseState::PENDING) {
+    js.runMicrotasks();
+  }
+  KJ_ASSERT(promise.state() == jsg::PromiseState::FULFILLED);
+  return promise.result();
+}
```
hoodmane, line 30: Interesting. So this is asserting that each promise will be resolved in one trip around the event loop or... something?

danlapid:
```c++
  /**
   * Runs the default MicrotaskQueue until it gets empty and perform other
   * microtask checkpoint steps, such as calling ClearKeptObjects. Asserts that
   * the MicrotasksPolicy is not kScoped. Any exceptions thrown by microtask
   * callbacks are swallowed.
   */
```
So it's not one trip around the event loop but rather, until we're completely done processing any events.

hoodmane: But only events that are microtasks not the main event loop?

danlapid: What do you believe we are running here that does not count as a microtask?

danlapid: I don't expect network IO to work if that's the question

src/pyodide/emscripten_setup.capnp
hoodmane: What I mean is either we always update both at the same time and then there is no caching benefit to separating them, or we need to somehow know when the bigger bundle hasn't changed.

src/workerd/api/pyodide/setup-emscripten.c++
hoodmane: What about instantiating webassembly?

danlapid: Well it works so I guess it is 😁

hoodmane: Good enough for me =)

src/workerd/io/worker.h
```diff
@@ -294,6 +297,10 @@ class Worker::Isolate: public kj::AtomicRefcounted {
     return id;
   }
 
+  inline void setId(kj::String newId) {
```
danlapid, line 300: Unused

src/workerd/server/workerd-api.c++
```diff
@@ -128,6 +129,92 @@ static const PythonConfig defaultConfig{
 };
 }  // namespace
 
+namespace {
```
danlapid, line 132: Note for code review:
No changes to anything in this namespace scope, this is only a move from below because I need these functions earlier in the file

danlapid:
Removing from draft stage as this is pretty much the end state of this PR.
Everything under `src/workerd/jsg` has been pulled into separate PRs so you can ignore that (it'll disappear when I rebase after those PRs are merged).
@hoodmane feel free to start reviewing Internal PR # 9006 is nearing a ready state and will be removed from draft soon as well so you can safely look at it to understand what the internal code is going to look like.

danlapid:
@hoodmane this is a breaking change to the bundle, I'll need an explanation of how we can do that safely  in a way that won't break backwards compatibility

hoodmane:
> this is a breaking change to the bundle, I'll need an explanation of how we can do that safely in a way that won't break backwards compatibility

You should be able to:
1. release the new bundle as a patch version bump
2. Update the patch version in compatibility-date.capnp in this PR

Then when we roll it out, the bundle patch version changes at the same time as the C++ code.

src/pyodide/BUILD.bazel
```diff
@@ -197,8 +197,6 @@ INTERNAL_MODULES = glob(
     [
         "internal/*.ts",
         "internal/topLevelEntropy/*.ts",
-        # The pool directory is only needed by typescript, it shouldn't be used at runtime.
-        "internal/pool/*.ts",
```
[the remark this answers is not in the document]
danlapid: Does wd_ts_bundle do any more checks than esbuild?

src/pyodide/BUILD.bazel
danlapid: Is there an easy way to configure esbuild to type check perhaps? I’d prefer not writing a whole new babel target for this

src/workerd/api/pyodide/setup-emscripten.c++
```diff
@@ -0,0 +1,105 @@
...
+      jsg::v8StrIntern(js.v8Isolate, "pyodide-internal:generated/emscriptenSetup"), 0, 0, false, -1,
+      {}, false, false, true);
+  v8::ScriptCompiler::Source source(contentStr, origin);
+  return jsg::check(v8::ScriptCompiler::CompileModule(js.v8Isolate, &source));
+}
+
+jsg::JsValue resolvePromise(jsg::Lock& js, jsg::JsValue prom) {
+  auto promise = KJ_ASSERT_NONNULL(prom.tryCast<jsg::JsPromise>());
+  if (promise.state() == jsg::PromiseState::PENDING) {
+    js.runMicrotasks();
```
dom96, line 24: Can doing this block the thread for too long and prevent other isolates from doing work?

I guess we will run this in its own thread, so this won't be an issue?

src/workerd/server/workerd-api.c++
```diff
@@ -174,8 +262,25 @@ struct WorkerdApi::Impl final {
...
+        if (features->getPythonExternalBundle() ||
+            util::Autogate::isEnabled(util::AutogateKey::PYTHON_EXTERNAL_BUNDLE)) {
+          auto pythonRelease = KJ_ASSERT_NONNULL(getPythonSnapshotRelease(*features));
+          auto version = getPythonBundleName(pythonRelease);
+          bundle = KJ_ASSERT_NONNULL(
+              fetchPyodideBundle(pythonConfig, version), "Failed to get Pyodide bundle");
+        }
+        auto context = lock.newContext<api::ServiceWorkerGlobalScope>({}, lock.v8Isolate);
+        v8::Context::Scope scope(context.getHandle(lock));
+        // Init emscripten syncronously, the python script will import setup-emscripten and
```
dom96, line 278:
```suggestion
        // Init emscripten synchronously, the python script will import setup-emscripten and
```

dom96 (approved):
Nice, some small nits/questions, but otherwise LGTM.

src/pyodide/BUILD.bazel
hoodmane: There are a couple of esbuild plugins that could possibly be used or we could make our own. It isn't that common for people to do it this way though, usually the build script for an esbuild project looks like `tsc --noEmit && node esbuild.config.mjs`

src/pyodide/internal/pool/emscriptenSetup.ts
```diff
@@ -210,6 +210,8 @@ export async function instantiateEmscriptenModule(
 
     // Wait until we've executed all the preRun hooks before proceeding
     const emscriptenModule = await emscriptenSettings.readyPromise;
+    emscriptenModule.setUnsafeEval = setUnsafeEval;
+    emscriptenModule.setGetRandomValues = setGetRandomValues;
```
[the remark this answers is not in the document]
hoodmane: No, unfortunately nothing in the pool can get imported directly or else we get two copies that are unrelated to each other.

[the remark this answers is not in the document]
danlapid: We're storing callbacks that will set a global variable in the emscripten context.
This emscripten context is ran as pure JS without any import functionality before we even have a module registry.
Later on when we actually have a request incoming and we have a module registry loaded and everything that comes with it we can use these callbacks to set the UnsafeEval and GetRandomValues functions in the emscripten context.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -484,6 +486,24 @@ void DiskCache::put(jsg::Lock& js, kj::String key, kj::Array<kj::byte> data) {
   }
 }
 
+jsg::JsValue SetupEmscripten::getModule(jsg::Lock& js) {
+  KJ_IF_SOME(module, emscriptenModule) {
+    return module.getHandle(js);
+  } else {
+    auto& runtime = KJ_ASSERT_NONNULL(workerd::Worker::Api::current().getEmscriptenRuntime());
```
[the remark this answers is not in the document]
danlapid: Well it can be guaranteed that if we have a python worker this HAS to be instantiated before we call getModule.
If for some reason someone called SetupEmscripten.getModule without an emscripten runtime this should definitely fail and there's now way to recover from it.
That being said, for any isolate that isn't a python isolate emscriptenRuntime will always be a kj::none and SetupEmscripten.getModule will simply never be called.

src/workerd/api/pyodide/setup-emscripten.c++
danlapid: Previously this logic would run in the global await that the loading of the pyodide bundle would block on. It's essentially the exact same logic just moved.
This means that we aren't blocking for any more time than we previously did but simply blocking at stage that's a bit earlier.

For workerd that doesn't matter at all.
For our internal use case it will maintain the original behaviour for any isolate that isn't part of the pool (yeah those still exist).
If the isolate is part of the pool then this will run in a separate thread and actually improve the current status.

src/workerd/server/workerd-api.c++
danlapid: oopsie

src/pyodide/BUILD.bazel
danlapid: Well this is a lot more involved than you make it seem, I'm just restoring the pool directory here so it'll be type checked in the same way it was before this PR.

hoodmane:
Changes to typescript look good to me.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -490,7 +490,7 @@ jsg::JsValue SetupEmscripten::getModule(jsg::Lock& js) {
 }
 
 void SetupEmscripten::visitForGc(jsg::GcVisitor& visitor) {
-  visitor.visit(emscriptenRuntime.emscriptenRuntime);
+  visitor.visit(const_cast<EmscriptenRuntime&>(emscriptenRuntime).emscriptenRuntime);
```
dom96, line 493: Nit: might be worth writing a comment why this cast is necessary and why it's safe

dom96 (approved):

PR #2937, opened 2024-10-16
Revert "Erikcorry roll v8 13.0 (#2918)"

AdityaAtulTewari:
This reverts commit 2160e96fdaa7a37c5a1ca56423fc56f36f60688a.


erikcorry (approved):

src/workerd/jsg/v8-platform-wrapper.h
```diff
@@ -22,9 +22,8 @@ class V8PlatformWrapper: public v8::Platform {
     return inner.NumberOfWorkerThreads();
   }
 
-  std::shared_ptr<v8::TaskRunner> GetForegroundTaskRunner(
```
fhanau, line 25: Instead of repeatedly rolling this change back and forth, can't we just keep the update and use `V8_MAJOR_VERSION` to select between both versions so that the internal build just works with either V8 version? Otherwise I have to recompile almost everything for workerd whenever we roll this back/reland it. Having a temporary version mismatch should be acceptable as long as it won't be for extended time frames.

PR #2939, opened 2024-10-16
add more functions for node:url

anonrig:
Had to update `node:path` posix implementation to make sure new functions added to `node:url` is implemented correctly. The last remaining part of `node:url` is the legacy_url which I'll open a new pull-request to add it.

Adds:
- `fileURLToPath`
- `pathToFileURL`
- `toPathIfFileURL`

### DIFF
--- a/src/node/internal/internal_url.ts
+++ b/src/node/internal/internal_url.ts
@@ -1,0 +1,226 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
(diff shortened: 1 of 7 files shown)


src/node/internal/internal_url.ts
```diff
@@ -0,0 +1,210 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+import {
```
jasnell, line 4: nit: This likely needs the node.js attribution header also

jasnell (approved):

PR #2941, opened 2024-10-17
VS-284: Add Bindings support for Vectorize queryById operation

garvit-gupta:
Add a new Vectorize operation for Bindings: `queryById()` that allows users to query their index using a vector already present in the index.

### DIFF
--- a/src/cloudflare/internal/vectorize-api.ts
+++ b/src/cloudflare/internal/vectorize-api.ts
@@ -18,6 +18,10 @@ enum Operation {
 
 type VectorizeVersion = 'v1' | 'v2';
 
+type QueryImplV2Params =
+  | { vector: VectorFloatArray | number[]; vectorId?: undefined }
+  | { vector?: undefined; vectorId: string };
+
 function toNdJson(arr: object[]): string {
   return arr.reduce((acc, o) => acc + JSON.stringify(o) + '\n', '').trim();
 }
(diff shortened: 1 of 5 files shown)


src/cloudflare/internal/vectorize-api.ts
```diff
@@ -114,6 +114,45 @@ class VectorizeIndexImpl implements Vectorize {
...
   }
 
+  public async queryById(
+    vectorId: string,
+    options?: VectorizeQueryOptions
+  ): Promise<VectorizeMatches> {
+    if (this.indexVersion === 'v1') {
+      throw new Error(`QueryById operation is not supported for v1 indexes.`);
+    } else {
+      if (options?.returnMetadata) {
```
garrettgu10, line 124: Can we factor out this duplicate code between query() and queryById()? Something like a queryImpl() that accepts an object that contains either a `vector` or `vectorId` field, then spreads the object into the body.

garrettgu10 (approved):

src/cloudflare/internal/vectorize-api.ts
garvit-gupta: Great suggestion! Removed the duplication.

fhanau:
Linux CI appears thoroughly broken on this PR for some reason – likely unrelated to the changes and just a cache connectivity issue, but we can't merge until that's resolved

PR #2943, opened 2024-10-17
Implements the start of an SDK for Python Workers.

dom96:
Implements the beginnings of a Python SDK, an API that makes writing Python Workers easier as it avoids dealing with awkward details of the Pyodide JS FFI. This PR implements the following:

* a new `cloudflare.workers` module which contains the API
* a `fetch` function defined in that module which accepts native Python types like HTTPMethod, it calls into Pyodide's `pyfetch` which offers many other niceties
* a `Response` function which creates a new `Response` instance and accepts native Python types like HTTPStatus
* support for returning a Pyodide FetchResponse from the `on_fetch` handler
* a custom `FetchResponse` which makes it possible to access the JS ReadableStream via `body` (this is often used in examples which rewrite a response). This is probably something we want to upstream to Pyodide.

This is just the start and the plan is to add more APIs into this module. For example, I want to make an idiomatic version of `FormData` and I'm sure we'll run into other cases where an idiomatic wrapper around the JS API makes sense.

### Test Plan

```
$ bazel run @workerd//src/workerd/server/tests/python:sdk/sdk
```

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -1,0 +1,78 @@
+# This module defines a Workers API for Python. It is similar to the API provided by
+# JS Workers, but with changes and additions to be more idiomatic to the Python
+# programming language.
+from http import HTTPMethod, HTTPStatus
+from typing import TypedDict, Unpack
+
+import js
+
+import pyodide.http
+from pyodide.http import pyfetch
+
+JSBody = (
+    "js.Blob | js.ArrayBuffer | js.TypedArray | js.DataView | js.FormData |"
+    "js.ReadableStream | js.URLSearchParams"
+)
+Body = "str | JSBody"
+Headers = dict[str, str] | list[tuple[str, str]]
+
+
+class FetchKwargs(TypedDict, total=False):
+    headers: Headers | None
+    body: "Body | None"
+    method: HTTPMethod = HTTPMethod.GET
+
+
+# TODO: This is just here to make the `body` field available as many Workers
+# examples which rewrite responses make use of it. It's something we should
+# likely upstream to Pyodide.
+class FetchResponse(pyodide.http.FetchResponse):
+    @property
+    def body(self) -> Body:
+        """
+        Returns the body from the JavaScript Response instance.
+        """
+        return self.js_response.body
+
+
+async def fetch(
+    resource: str,
+    **other_options: Unpack[FetchKwargs],
+) -> FetchResponse:
+    if "method" in other_options and isinstance(other_options["method"], HTTPMethod):
+        other_options["method"] = other_options["method"].value
+    resp = await pyfetch(resource, **other_options)
+    return FetchResponse(resp.url, resp.js_response)
+
+
+class Response:
+    def __init__(
+        self,
+        body: Body,
+        status: HTTPStatus | int = HTTPStatus.OK,
+        statusText="",
+        headers: Headers = None,
+    ):
+        """
+        Represents the response to a request.
+
+        Based on the JS API of the same name:
+        https://developer.mozilla.org/en-US/docs/Web/API/Response/Response.
+        """
+        options = {
+            "status": status.value if isinstance(status, HTTPStatus) else status,
+        }
+        if len(statusText) > 0:
+            options["statusText"] = statusText
+        if headers:
+            if isinstance(headers, list):
+                # We should have a list[tuple[str, str]]
+                options["headers"] = js.Headers.new(headers)
+            elif isinstance(headers, dict):
+                options["headers"] = js.Headers.new(headers.items())
+            else:
+                raise TypeError(
+                    "Response() received unexpected type for headers argument"
+                )
+
+        self.js_response = js.Response.new(body, **options)
--- a/src/pyodide/python-entrypoint-helper.ts
+++ b/src/pyodide/python-entrypoint-helper.ts
@@ -93,6 +93,11 @@ export async function setupPackages(pyodide: Pyodide): Promise<void> {
     const pymajor = pyodide._module._py_version_major();
     const pyminor = pyodide._module._py_version_minor();
     pyodide.site_packages = `/lib/python${pymajor}.${pyminor}/site-packages`;
+
+    // Inject modules that enable JS features to be used idiomatically from Python.
+    pyodide.FS.mkdir(`${pyodide.site_packages}/cloudflare`);
+    await injectSitePackagesModule(pyodide, 'workers', 'cloudflare/workers');
+
     // Install patches as needed
     if (TRANSITIVE_REQUIREMENTS.has('aiohttp')) {
       await applyPatch(pyodide, 'aiohttp');
@@ -142,9 +147,17 @@ function makeHandler(pyHandlerName: string): Handler {
         'prep_python',
         async () => await preparePython()
       );
-      return await enterJaegerSpan('python_code', () => {
+
+      const result = await enterJaegerSpan('python_code', () => {
         return mainModule[pyHandlerName].callRelaxed(...args);
       });
+
+      // Support returning a pyodide.ffi.FetchResponse.
+      if (result && result.js_response !== undefined) {
+        return result.js_response;
+      } else {
+        return result;
+      }
     } catch (e) {
       console.warn('Error in makeHandler');
       reportError(e);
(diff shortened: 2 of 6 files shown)


src/pyodide/python-entrypoint-helper.ts
```diff
@@ -142,9 +147,17 @@ function makeHandler(pyHandlerName: string): Handler {
...
         async () => await preparePython()
       );
-      return await enterJaegerSpan('python_code', () => {
+
+      const result = await enterJaegerSpan('python_code', () => {
         return mainModule[pyHandlerName].callRelaxed(...args);
       });
+
+      // Support returning a pyodide.ffi.FetchResponse.
+      if (typeof result.js_response !== 'undefined') {
```
anonrig, line 156:
```suggestion
      if (result.js_response !== undefined) {
```

src/pyodide/internal/workers.py
```diff
@@ -0,0 +1,54 @@
+# This module defines a Workers API for Python. It is similar to the API provided by
```
danlapid, line 1: Can this be just a package we publish similar to workers-rs that people can choose to depend on and use?

src/pyodide/internal/workers.py
```diff
@@ -0,0 +1,54 @@
...
+    headers: Headers = None,
+    body: Body | None = None,
+    method: HTTPMethod = HTTPMethod.GET,
+    **other_options,
+) -> FetchResponse:
+    if headers:
+        other_options["headers"] = headers
+
+    if body:
+        other_options["body"] = body
```
hoodmane, line 30: To give a type to a kwargs argument, you can make a typed dict:
```py
class FetchKwargs(TypedDict, total=False):
    headers: Headers
    body : Body
```
and then you can do:
```py
async def fetch(
    resource: str,
    **kwargs: Unpack[FetchKwargs]
)
```
 and then you can get rid of this stuff.

src/pyodide/internal/workers.py
```diff
@@ -0,0 +1,54 @@
...
+def Response(
+    body: Body, status=HTTPStatus.OK, statusText="", headers: Headers = None
+) -> js.Response:
+    """
+    Represents the response to a request.
+
+    Based on the JS API of the same name:
+    https://developer.mozilla.org/en-US/docs/Web/API/Response/Response.
+    """
+    options = to_js(
```
hoodmane, line 47: We probably want to allow the user to pass a dictionary for headers? Maybe
```py
if not headers:
   headers = {}
if isinstance(headers, dict):
   headers = Headers.new(headers.items())
```

src/pyodide/python-entrypoint-helper.ts
hoodmane: Probably we want to use `?.` so that if the handler incorrectly returns `None` we don't raise an error here.

src/pyodide/internal/workers.py
dom96: It could be, but I don't see a reason why it shouldn't be included implicitly. All of our examples should be using this module, so getting everyone to also add dependencies on this module wouldn't be a good experience.

danlapid: Well the biggest benefit would be to have this be it's own repo with it's own issues and PRs and have everyone be able to easily contribute and have releases be automated in CI so that nobody needs to wait for a workerd release.
I think that approach turned out rather well for us with Workers-rs which is getting a lot of outside contributions.
If we limited contributions to only people who are able to build and debug workerd it would be a real shame.

dom96: Having this in its own repo is probably doable, but I don't think it should be a separate package that developers have to opt into. It should be the default way to write Python Workers.

That said, I have my reservations about creating more repos for Python related code (we already have at least 4).

src/pyodide/internal/workers.py
dom96: Cool trick, thanks!

src/pyodide/internal/workers.py
```diff
@@ -0,0 +1,76 @@
...
+    resp = await pyfetch(resource, **other_options)
+    return FetchResponse(resp.url, resp.js_response)
+
+
+def Response(
+    body: Body,
+    status: HTTPStatus | int = HTTPStatus.OK,
+    statusText="",
+    headers: Headers = None,
+) -> js.Response:
```
hoodmane, line 53: If the name is `Capitalized` then it should probably be a class, for instance with support for `isinstance`. If you want the result to be actually a `js.Response` you could add a `__instancecheck__` and a `__new__`. Or with the change to `python-entrypoint-helper` you could make a Python class with a `js_response` field. That may be the most reasonable approach. Maybe we should think about upstreamng changes to make it convenient to construct a `pyodide.ffi.FetchResponse` directly.

src/workerd/server/tests/python/sdk/mock.py
```diff
@@ -0,0 +1,17 @@
...
+async def on_fetch(request):
+    if request.url.endswith("/sub"):
+        return Response("Hi there!")
+    elif request.url.endswith("/sub_headers"):
+        assert request.headers.get("Custom-Req-Header") == "123"
+        return Response(
+            "Hi there!", headers={"Custom-Header-That-Should-Passthrough": True}
+        )
+    else:
+        raise AssertionError("Unexpected path " + request.url)
```
hoodmane, line 13: It's a bit odd to raise an `AssertionError` on something that came across the application boundary like this. A `ValueError` might be most appropriate here.

src/workerd/server/tests/python/sdk/worker.py
```diff
@@ -0,0 +1,84 @@
...
+            headers={"Custom-Header-That-Should-Passthrough": "modified"},
+        )
+    elif request.url.endswith("/modify_tuple"):
+        resp = await fetch("https://example.com/sub")
+        return Response(
+            resp.body,
+            headers=[
+                ("Custom-Header-That-Should-Passthrough", "modified"),
+                ("Custom-Header-That-Should-Passthrough", "another"),
+            ],
```
hoodmane, line 21: What does the code look like if we want to add extra headers? Is that convenient?

src/pyodide/internal/workers.py
```diff
@@ -0,0 +1,76 @@
...
+# TODO: This is just here to make the `body` field available as many Workers
+# examples which rewrite responses make use of it. It's something we should
+# likely upstream to Pyodide.
+class FetchResponse(pyodide.http.FetchResponse):
+    @property
+    def body(self) -> Body:
+        """
+        Returns the body from the JavaScript Response instance.
+        """
+        return self.js_response.body
```
hoodmane, line 35: Yeah this should be an easy thing to upstream.

src/pyodide/internal/workers.py
```diff
@@ -0,0 +1,76 @@
+# This module defines a Workers API for Python. It is similar to the API provided by
+# JS Workers, but with changes and additions to be more idiomatic to the Python
+# programming language.
+from http import HTTPMethod, HTTPStatus
+from typing import TypedDict, Unpack
+
+import js
+
+import pyodide.http
+from pyodide.http import pyfetch
```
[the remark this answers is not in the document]
dom96: I didn't do this because I want the code to be explicit when referring to Pyodide's `FetchResponse`, since I am defining a new class of the same name. Did I miss something?

src/workerd/server/tests/python/sdk/mock.py
dom96: fwiw this was a `assert False` but Ruff complained and suggested using `raise AssertionError` so I followed its advice. I don't think the exact error matters much here since this is a test, but I'll change it.

src/workerd/server/tests/python/sdk/worker.py
dom96: How do you mean? You can just add extra tuples into the list or use a dictionary (like above).

This is testing that duplicate headers can be specified and most shouldn't use this form when specifying headers.

src/pyodide/internal/workers.py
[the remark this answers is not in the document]
hoodmane: Ah. Then maybe it's nice to do:
```py
from pyodide.http import pyfetch, FetchResponse as OrigFetchResponse
```
but either way.

hoodmane (approved):

PR #2946, opened 2024-10-17
WC-2707 Only submit non-empty traces to trace workers

matthewdavidrodgers:
Currently, any invocation for a stage generates a trace, even for PipelineLogLevel::NONE. The only difference is that NONE does not attach logs, exceptions, or eventInfo to the trace.

This leads us to see traces with a null event for stages you opted to not get traces from, which we often see when there's a FPW in a pipeline that we did not intend to see.

There are quite a few places throughout this and edgeworker that note that we do this to "keep the request tree contiguous", but from my investigation and chats with other people, these traces are not used for anything other than delivery to trace/tail handlers, i.e. this isn't used to report metrics or billing. So there's no reason to filter these out before delivering traces.

This is of course not as efficient as it could be, as we are allocating for empty traces and then discarding them before submitting. A more comprehensive approach would be to not generate traces at all when the stage / subpipeline's log level is NONE, but that's probably beyond me currently (although I would welcome the help).

I do have a tests that are tweaked to reflect these changes in edgeworker, but there seem to be some quirks / nuances I don't understand there (for example, the tests pass normally but do not pass for process sandboxing) so I was recommended to get a PR up to workerd to begin the discussion here.

### DIFF
--- a/src/workerd/api/trace.c++
+++ b/src/workerd/api/trace.c++
@@ -607,17 +607,24 @@ kj::Promise<void> sendTracesToExportedHandler(kj::Own<IoContext::IncomingRequest
     tracer->setEventInfo(context.now(), Trace::TraceEventInfo(traces));
   }
 
+  auto nonEmptyTraces = kj::Vector<kj::Own<Trace>>(kj::size(traces));
+  for (auto& trace: traces) {
+    if (trace->eventInfo != kj::none) {
+      nonEmptyTraces.add(kj::mv(trace));
+    }
+  }
+
   // Add the actual JS as a wait until because the handler may be an event listener which can't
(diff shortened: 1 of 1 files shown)


anonrig (commented):
Can we have a test for this, so we don't regress?

matthewdavidrodgers:
@anonrig i have ew-tests written in edgeworker that cover the changes; is there a suitable place to do something similar here?

anonrig:
> @anonrig i have ew-tests written in edgeworker that cover the changes; is there a suitable place to do something similar here?

If there is a test in the downstream repo, I don't think we need an additional one here. Thanks!

anonrig (approved):

fhanau:
Rebased the branch and created a downstream PR so we can ship this.

PR #2949, opened 2024-10-18
Erikcorry roll v8 13.0 (#2918)

erikcorry:
This is a tagged version for use by edgeworker 2024.10.930.
It should stick around even after it is merged to main.


erikcorry:
This PR should not be committed to main until edgeworker's origin/main is also on V8 13.0

AdityaAtulTewari (approved):

PR #2950, opened 2024-10-18
feat: make create instance api take a single options parameter

LuisDuarte1:
Not only this allows for a better DX, it allows us to customize and add new fields later without breaking changes.

Supersedes https://github.com/cloudflare/workerd/pull/2881

### DIFF
--- a/src/cloudflare/internal/workflows-api.ts
+++ b/src/cloudflare/internal/workflows-api.ts
@@ -96,11 +96,13 @@ class WorkflowImpl {
     return new InstanceImpl(result.instanceId, this.fetcher);
   }
 
-  public async create(id: string, params?: unknown): Promise<Instance> {
+  public async create(
+    options?: WorkflowInstanceCreateOptions
+  ): Promise<Instance> {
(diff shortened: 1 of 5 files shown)


src/cloudflare/internal/workflows-api.ts
```diff
@@ -96,11 +96,13 @@ class WorkflowImpl {
...
 
-  public async create(id: string, params?: unknown): Promise<Instance> {
+  public async create(
+    options?: WorkflowInstanceCreateOptions
+  ): Promise<Instance> {
     const result = await callFetcher<{ instanceId: string }>(
       this.fetcher,
       '/create',
-      { id, params }
+      options === undefined ? {} : options
```
anonrig, line 105: Nit:

```suggestion
      options ?? {}
```

anonrig (approved):

src/cloudflare/internal/workflows-api.ts
LuisDuarte1: done :)

anonrig (approved):

PR #2952, opened 2024-10-18
enable compile cache for faster boots

anonrig:
Enables compile cache in workerd.

### DIFF
--- a/src/workerd/jsg/modules.c++
+++ b/src/workerd/jsg/modules.c++
@@ -23,13 +23,40 @@ namespace {
 // we'd likely need to have find return an atomic refcount or something similar.
 class CompileCache {
 public:
-  void add(const void* key, std::unique_ptr<v8::ScriptCompiler::CachedData> cached) const {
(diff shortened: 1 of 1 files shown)


jasnell (approved):

irvinebroque:
What's the TLDR of what this does and how much it improves performance?

anonrig:
> What's the TLDR of what this does and how much it improves performance?

@irvinebroque This particular change enables on-demand compile cache for built-in modules (such as `node:buffer`, `node:url` etc.). Simplify, the second import of a built-in module becomes extremely cheap, and hence improves the boot time (due to the less time required to parse and analyze the module). This is particularly different than Node.js, since Node compiles all built-in modules on compile time, which makes the first import extremely cheap. I haven't done any benchmarks yet, but Node.js has seen 56% improvement in boot time by implementing the whole build time caching. (Ref: https://github.com/nodejs/node/pull/24950). We can invest on this and go through a similar path depending on the `nodejs_compat` flags for example.

PS: V8 has seen 20-40% improvements - https://v8.dev/blog/improved-code-caching.

PR #2953, opened 2024-10-18
Make sure kCharLookupTable is indexed by unsigned char instead of sig…

jasnell:
…ned char

### DIFF
--- a/src/workerd/api/util.c++
+++ b/src/workerd/api/util.c++
@@ -146,7 +146,7 @@ kj::String redactUrl(kj::StringPtr url) {
   };
 
   for (const char& c: url) {
-    uint8_t lookup = kCharLookupTable[c];
+    uint8_t lookup = kCharLookupTable[static_cast<const kj::byte>(c)];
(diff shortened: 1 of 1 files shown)


src/workerd/api/util.c++
```diff
@@ -146,7 +146,7 @@ kj::String redactUrl(kj::StringPtr url) {
   };
 
   for (const char& c: url) {
-    uint8_t lookup = kCharLookupTable[c];
+    uint8_t lookup = kCharLookupTable[static_cast<const kj::byte>(c)];
```
anonrig, line 149: Why do we need this? `const char` is an alias to `const kj::byte` isn't it?

anonrig: Ah I get it, char is not `uint8_t`

anonrig (approved):

src/workerd/api/util.c++
jasnell: `kj::byte` is an alias for `unsigned char`

PR #2956, opened 2024-10-19
Roll boringssl

fhanau:
Follow-up to #2745. After confirming that the few cases where DH inputs that are now rejected by boringssl are not being actively used in production, we can return a proper error for them and update the boringssl version for workerd. With the new version, we're moving 13 months forward. Even if the version used internally is out-of-sync for now, the added checks here ensure that the behavior is the same.
Test cases are updated based on the new behavior.

### DIFF
--- a/build/deps/deps.jsonc
+++ b/build/deps/deps.jsonc
@@ -14,11 +14,12 @@
       "type": "github_tarball",
       "owner": "google",
       "repo": "boringssl",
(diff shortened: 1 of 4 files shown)


fhanau:
Downstream build (same branch name) passes.

AdityaAtulTewari (approved):
LGTM

PR #2961, opened 2024-10-20
Simplify PipelineTracer

jasnell:
Eliminate internal nesting of PipelineTracers. Use the resolved on complete promise to propagate traces up to the parent.

This was originally done as part of the larger PR to introduce streaming traces but makes sense to pull out separately here.

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -591,7 +591,7 @@ kj::Rc<WorkerTracer> PipelineTracer::makeWorkerTracer(PipelineLogLevel pipelineL
       kj::mv(dispatchNamespace), kj::mv(scriptId), kj::mv(scriptTags), kj::mv(entrypoint),
       executionModel);
   traces.add(kj::addRef(*trace));
-  return kj::rc<WorkerTracer>(kj::addRef(*this), kj::mv(trace), pipelineLogLevel);
+  return kj::rc<WorkerTracer>(addRefToThis(), kj::mv(trace), pipelineLogLevel);
 }
 
 void PipelineTracer::addTrace(rpc::Trace::Reader reader) {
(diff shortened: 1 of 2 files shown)


harrishancock (approved):

src/workerd/io/trace.h
```diff
@@ -368,12 +361,12 @@ class PipelineTracer final: public kj::Refcounted {
   // to the host where tracing was initiated.
   void addTrace(rpc::Trace::Reader reader);
 
+  void addTracesFromChild(kj::ArrayPtr<kj::Own<Trace>> traces);
```
anonrig, line 364: a documentation to this function would be nice

src/workerd/io/trace.c++
```diff
@@ -590,15 +591,15 @@ kj::Rc<WorkerTracer> PipelineTracer::makeWorkerTracer(PipelineLogLevel pipelineL
...
+  return kj::rc<WorkerTracer>(addRefToThis(), kj::mv(trace), pipelineLogLevel);
 }
 
 void PipelineTracer::addTrace(rpc::Trace::Reader reader) {
   traces.add(kj::refcounted<Trace>(reader));
 }
 
 WorkerTracer::WorkerTracer(
-    kj::Own<PipelineTracer> parentPipeline, kj::Own<Trace> trace, PipelineLogLevel pipelineLogLevel)
+    kj::Rc<PipelineTracer> parentPipeline, kj::Own<Trace> trace, PipelineLogLevel pipelineLogLevel)
```
anonrig, line 602: What's the reason for this change?

anonrig (approved):

src/workerd/io/trace.c++
jasnell: `kj::Rc<T>` is a newer alternative to `kj::Own<T>` that makes the refcounted semantics clearer (and safer). We haven't adopted it much yet but it definitely beats passing around bare references. Since I'm in here modifying this code anyway it makes sense to update to this.

src/workerd/io/trace.h
jasnell, line 364:
```suggestion
  // When collecting traces from multiple stages in a pipeline, this is called by the
  // tracer for a subordinate stage to add its collected traces to the parent pipeline.
  void addTracesFromChild(kj::ArrayPtr<kj::Own<Trace>> traces);
```

PR #2965, opened 2024-10-21
[nfc] Update to python 3.13, minor python and TS cleanup

fhanau:
- Update rules_python, facilitating python 3.13 update
- Remove actions/setup-python from lint.yml – python 3.12 already ships on the Ubuntu 24.04 runner.
- Update python deps (markupsafe)
- Refactor wd_test macro and tsconfig so that we can define TypeScript-based JS tests without having to manually define a ts_project


.github/workflows/lint.yml
```diff
@@ -19,9 +19,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           show-progress: false
-      - uses: actions/setup-python@v5
```
fhanau, line 22: @hoodmane I know this was added intentionally, but the Ubuntu 24.04 image already comes with python 3.12. I'm open to keeping this if you see a need for it despite that, or if the lint script would benefit from python 3.13 (developers won't have access to 3.13 for now so I guess that's a reason to not depend on it for now).

src/workerd/api/tsconfig.json
```diff
@@ -11,7 +11,8 @@
     },
     "composite": true,
     "declaration": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "allowJs": true
```
fhanau, line 15: allowJs added to match ts_project in wd_test – this is not needed now but we might have wd_tests using both JS + TS in the future.

.github/workflows/lint.yml
hoodmane: Removing seems fine to me, no strong feelings.

PR #2967, opened 2024-10-21
add complete node:url implementation

anonrig:
Adds all remaining functionality for removing the requirement of the polyfill from wrangler for `node:url` module. (We can remove the hybrid polyfill). It is important to add the remaining modules because `node:url` is one of the most used modules in the ecosystem.

Here's the functionality implemented:

- Legacy `new Url()` class
- `url.urlToHttpOptions`
- `url.format()`
- `url.parse()`
- `url.resolve()`
- `url.resolveObject()`

### DIFF
--- a/src/node/internal/legacy_url.ts
+++ b/src/node/internal/legacy_url.ts
@@ -1,0 +1,1018 @@
+import { validateString, validateObject } from 'node-internal:validators';
+import {
+  CHAR_SPACE,
+  CHAR_TAB,
+  CHAR_CARRIAGE_RETURN,
+  CHAR_LINE_FEED,
+  CHAR_NO_BREAK_SPACE,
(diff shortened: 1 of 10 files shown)


jasnell (approved):
Makes me sad to have to support this but LGTM

src/node/internal/legacy_url.ts
```diff
@@ -0,0 +1,1018 @@
...
+  this.host = null;
+  this.port = null;
+  this.hostname = null;
+  this.hash = null;
+  this.search = null;
+  this.query = null;
+  this.pathname = null;
+  this.path = null;
+  this.href = null;
+} as unknown as {
```
anonrig, line 159: @mikea any suggestions on making this more maintainable?

PR #2970, opened 2024-10-22
SQLite: Cache prepared statements behind `sql.exec()`.

kentonv:
`sql.exec()` currently parses the query every time it is invoked. This change makes it so that we retain some number of prepared statements in a cache and reuse them, keyed by the query text. Since query strings are likely commonly internalized strings, cache lookups should be O(1) most of the time, as internalized strings compare equal by identity without the need to compare content.

### DIFF
--- a/src/workerd/api/sql.h
+++ b/src/workerd/api/sql.h
@@ -29,9 +29,6 @@ public:
   // JSG, which does not need to make a copy.
   using SqlValue = kj::Maybe<kj::OneOf<kj::Array<byte>, kj::StringPtr, double>>;
 
-  // One row of a SQL query result. This is an Object whose properties correspond to columns.
-  using SqlRow = jsg::Dict<SqlValue, jsg::JsString>;
-
   jsg::Ref<Cursor> exec(jsg::Lock& js, jsg::JsString query, jsg::Arguments<BindingValue> bindings);
   IngestResult ingest(jsg::Lock& js, kj::String query);
 
@@ -186,18 +183,17 @@ private:
     }
   }
 
-  // Utility functions to convert SqlValue, SqlRow, and Array<SqlValue> to JS values. In some
-  // cases we end up having to do this conversion before actually returning, so we can't have
-  // JSG do it. We can't use jsg::TypeHandler because SqlValue contains StringPtr, which doesn't
-  // support unwrapping. We don't actually ever use unwrapping, but requesting a TypeHandler forces
-  // JSG to try to generate the code for unwrapping, leading to compiler errors.
+  // Utility functions to convert SqlValue to a JS value. We can't just return the C++ values and
+  // let JSG do the work because we're trying to avoid having to make a copy of string contents out
+  // of SQLite's buffer when the conversion to JS is just going to make another copy. We can't use
+  // jsg::TypeHandler because SqlValue contains StringPtr, which doesn't support unwrapping. We
+  // don't actually ever use unwrapping, but requesting a TypeHandler forces JSG to try to generate
+  // the code for unwrapping, leading to compiler errors.
   //
   // TODO(cleanup): Think hard about how to make JSG support this better. Part of the problem is
   //   that we're being too clever with optimizations to avoid copying strings when we don't need
   //   to.
   static jsg::JsValue wrapSqlValue(jsg::Lock& js, SqlValue value);
-  static jsg::JsObject wrapSqlRow(jsg::Lock& js, SqlRow row);
-  static jsg::JsArray wrapSqlRowRaw(jsg::Lock& js, kj::Array<SqlValue> row);
 };
 
 class SqlStorage::Cursor final: public jsg::Object {
@@ -234,8 +230,8 @@ public:
     });
   }
 
-  JSG_ITERATOR(RowIterator, rows, SqlRow, jsg::Ref<Cursor>, rowIteratorNext);
-  JSG_ITERATOR(RawIterator, raw, kj::Array<SqlValue>, jsg::Ref<Cursor>, rawIteratorNext);
+  JSG_ITERATOR(RowIterator, rows, jsg::JsObject, jsg::Ref<Cursor>, rowIteratorNext);
+  JSG_ITERATOR(RawIterator, raw, jsg::JsArray, jsg::Ref<Cursor>, rawIteratorNext);
 
   RowIterator::Next next(jsg::Lock& js);
   jsg::JsArray toArray(jsg::Lock& js);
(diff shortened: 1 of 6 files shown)


kentonv:
Added a couple commits to fix a problem where queries were very commonly being left open until the cursor was GC'd, which made the statement cache ineffective.

justin-mp (approved):

src/workerd/api/sql.h
```diff
@@ -284,6 +288,11 @@ class SqlStorage::Cursor final: public jsg::Object {
   kj::Maybe<CachedColumnNames> ownCachedColumnNames;
   CachedColumnNames& cachedColumnNames;
 
+  // Invoke when `query.isDone()`, or when we want to prematurely cancel the query. This records
+  // row counters and then sets `stateRef` to `none` to drop the query and return the prepared
```
jclee, line 292: I guess technically it sets `state` to none, not `stateRef`, although they refer to the same thing in most cases.

src/workerd/api/sql.h
```diff
@@ -107,6 +107,58 @@ class SqlStorage final: public jsg::Object, private SqliteDatabase::Regulator {
...
+      return key.hashCode();
+    }
+  };
+
+  using StatementMap = kj::Table<kj::Rc<CachedStatement>, kj::HashIndex<StatementCacheCallbacks>>;
+
+  struct StatementCache {
+    StatementMap map;
+    kj::List<CachedStatement, &CachedStatement::lruLink> lru;
+    size_t totalSize = 0;
```
jclee, line 153: Should totalSize maybe feed back to jsg memory tracking metrics somehow?  Intuitively, I'd guess one could maybe add it to the `jsgGetMemorySelfSize()` return value, but I'm not familiar with exactly how that gets used.

src/workerd/util/sqlite-test.c++
```diff
@@ -1076,5 +1076,96 @@ KJ_TEST("SQLite onRollback") {
...
+  SqliteDatabase::Vfs vfs(*dir);
+  SqliteDatabase db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+  auto stmt = db.prepareMulti(SqliteDatabase::TRUSTED, kj::str(R"(
+    CREATE TABLE IF NOT EXISTS things (
+      id INTEGER PRIMARY KEY AUTOINCREMENT,
+      value INTEGER
+    );
+    INSERT INTO things(value) VALUES (123);
+    INSERT INTO things(value) VALUES (456);
```
jclee, line 1090: Would there be value in testing a multi statement that fails partway through?  I guess it wouldn't make a difference at construction time, since it's lazy.  And I guess it would just throw the appropriate exception at run time?

src/workerd/util/sqlite.c++
```diff
@@ -640,6 +643,13 @@ SqliteDatabase::StatementAndEffect SqliteDatabase::prepareSql(
 
           // Reduce `sqlCode` to include only what we haven't already executed.
           sqlCode = kj::StringPtr(tail, sqlCode.end());
+
+          KJ_IF_SOME(p, prelude) {
+            p.add(Statement(*this, regulator,
+                StatementAndEffect{.statement = kj::mv(ownResult),
+                  .stateChange = kj::mv(parseContext.stateChange)}));
```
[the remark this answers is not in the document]
kentonv: Eek, no, that is a bug. Good catch. I'd better add a test that does transactions.

(Though this bug wouldn't be possible to hit from JavaScript since we don't allow people to execute transaction statements from JS.)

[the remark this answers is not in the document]
kentonv: (But it's not actually an uninialized OneOf. The default value for `ParseContext.stateChange` is `NoChange()`.)

src/workerd/api/sql.h
```diff
@@ -191,7 +185,7 @@ class SqlStorage::Cursor final: public jsg::Object {
 
     void ensureInitialized(jsg::Lock& js, SqliteDatabase::Query& source);
 
-    JSG_MEMORY_INFO(cachedColumnNames) {
+    JSG_MEMORY_INFO(CachedColumnNames) {
```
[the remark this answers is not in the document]
kentonv: Yeah I was surprised by this, but I guess it just feeds into `sizeof()`. (I guess the size was even correct in this case, though if it wasn't, the result would presumably only be that memory was miscounted. No one would ever have noticed.)

src/workerd/api/sql.h
```diff
@@ -107,6 +107,58 @@ class SqlStorage final: public jsg::Object, private SqliteDatabase::Regulator {
...
+
+    inline auto hashCode(jsg::JsString key) const {
+      return key.hashCode();
+    }
+    inline auto hashCode(const jsg::HashableV8Ref<v8::String>& key) const {
+      return key.hashCode();
+    }
+  };
+
+  using StatementMap = kj::Table<kj::Rc<CachedStatement>, kj::HashIndex<StatementCacheCallbacks>>;
```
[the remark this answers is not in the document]
kentonv: It's because I wanted the key to be stored as a member of `CachedStatement` which is held by `kj::Rc`, not inlined into the table. `HashMap` doesn't quite have that structure. (I'll add a comment.)

src/workerd/api/sql.h
kentonv: Typo.

src/workerd/util/sqlite-test.c++
kentonv: Good call. I think the `prelude` vector would end up incorrect... the queries before the one that failed would get duplicated on the next attempt.

anonrig (approved):

src/workerd/api/sql.h
kentonv: Hmm I looked at `SqlStorage::visitForMemoryInfo` and I frankly have no idea what's going on there. It seems like it's double-counting pointer sizes but not counting the things they point to. Also I suppose it cannot dereference `IoOwn` because `visitForMemoryInfo()` is not necessarily called from within any particular `IoContext`? So is it even possible to count the statement cache?

I think I'm just not going to touch it.

kentonv:
Fixups:
* 16f2f4a2
* 99078377
* ac49e963
* f25004a7

However, the D1 API test is still failing. Going to fix that next.

kentonv:
OK, the D1 tests caught a real regression and the last commit fixes it.

src/workerd/util/sqlite-test.c++
```diff
@@ -1076,5 +1076,199 @@ KJ_TEST("SQLite onRollback") {
...
+    );
+    INSERT INTO things(value) VALUES (123);
+    INSERT INTO things(id, value) VALUES (1, 456);  -- fails, duplicate primary key
+  )"));
+
+  KJ_EXPECT_THROW_MESSAGE("SQLITE_CONSTRAINT", stmt.run());
+  KJ_EXPECT_THROW_MESSAGE("SQLITE_CONSTRAINT", stmt.run());
+  KJ_EXPECT_THROW_MESSAGE("SQLITE_CONSTRAINT", stmt.run());
+
+  // We ran the statement three times. Each time it should have inserted a new row containing
```
justin-mp, line 1190: We talked offline yesterday about `exec()`ing multiple statements and how if one fails in the middle, we should probably roll back all of the statements.  For my own education, in a world where we fix that bug, would `prepareMulti` implement the rollback or would something higher than `prepareMulti` handle the rollback?

If it's prepareMulti, should we drop a TODO(soon) here about how even this behavior is buggy?

justin-mp (approved):

src/workerd/util/sqlite-test.c++
kentonv: Technically `prepareMulti()` itself doesn't parse or execute anything. But I think your question is, would `SqliteDatabase` handle the rollback internally or would the caller have to do it. I think `SqliteDatabase` should handle it. I suspect logic will be needed in all of `prepareSql()`, `Statement::prepareForExecution()`, and `~Query()`, annoyingly.

> If it's prepareMulti, should we drop a TODO(soon) here about how even this behavior is buggy?

Added comment:

https://github.com/cloudflare/workerd/pull/2970/files#diff-2b573b1c6fa7749e40c7da2c29f37f0f4accf53f2406bdab6ca5378755260fecR1173-R1177

(No other changes, just rebased.)

PR #2971, opened 2024-10-22
Fixes Python import parser bug in multi-line string literals.

dom96:
We're switching to Ruff, but in the meantime the fix was simple so let's get it deployed.

### Test Plan

```
$ 
```

Also ran the parser on some real Python code (which I should have done in my previous changes), specifically:

* My cold start worker code
* The .py files in https://github.com/pyodide/pyodide/tree/main/src/py/pyodide 
* Some of the .py files in https://github.com/pyodide/pyodide/tree/main/src/tests

### DIFF
--- a/src/workerd/api/pyodide/pyodide-test.c++
+++ b/src/workerd/api/pyodide/pyodide-test.c++
@@ -187,5 +187,14 @@ import openai)SCRIPT"));
   KJ_REQUIRE(result[2] == "openai");
 }
 
+KJ_TEST("quote in multiline string") {
+  auto files = kj::heapArrayBuilder<kj::String>(1);
+  files.add(kj::str(R"SCRIPT(temp = """
+w["h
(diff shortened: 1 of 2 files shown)


src/workerd/api/pyodide/pyodide-test.c++
```diff
@@ -187,5 +187,14 @@ import openai)SCRIPT"));
   KJ_REQUIRE(result[2] == "openai");
 }
 
+KJ_TEST("quote in multiline string") {
+  auto files = kj::heapArrayBuilder<kj::String>(1);
+  files.add(kj::str(R"SCRIPT(temp = """
+w["h
```
hoodmane, line 193: Did this fail before the change?

dom96: yep

Issue #2969, opened 2024-10-21
🐛 Bug Report — Runtime APIs [URL.createObjectURL]


ImLunaHey:
`URL.createObjectURL` is missing.

looks like it was missed in this https://github.com/cloudflare/cloudflare-docs/issues/2382#issuecomment-1968783333

jasnell:
This is not a bug. We have intentionally not implemented `URL.createObjectIURL()` in workers. @anonrig has a PR that adds placeholders that will explicitly throw but implementing these is not currently in plan.

thorwebdev:
I'm running into this issue with `@11labs/react` https://github.com/thorwebdev/elevenlabs-voice-form-nextjs-cloudflare/blob/opennextjs-cloudflare/app/page.tsx#L12

AFAICS it should be running client-side, so not entirely sure why that would run in workers. Is there any way to make this work? The page is tagged as `use client` but that doesn't seem to make a difference.

PR #2973, opened 2024-10-22
add placeholders for createObjectURL()

anonrig:
Add placeholders for `URL.createObjectURL()` and `URL.revokeObjectURL()`

Ref: https://github.com/cloudflare/workerd/issues/2969

### DIFF
--- a/src/workerd/api/url-standard.c++
+++ b/src/workerd/api/url-standard.c++
@@ -53,6 +53,16 @@ bool URL::canParse(kj::String url, jsg::Optional<kj::String> maybeBase) {
   return jsg::Url::canParse(url, maybeBase.map([](kj::String& str) { return str.asPtr(); }));
 }
 
+jsg::JsString URL::createObjectURL(jsg::Lock& js, kj::OneOf<File, Blob> object) {
+  // TODO(soon): Implement this
+  JSG_FAIL_REQUIRE(Error, "URL.createObjectURL() is not implemented"_kj);
+}
+
+void URL::revokeObjectURL(jsg::Lock& js, kj::String object_url) {
+  // TODO(soon): Implement this
+  JSG_FAIL_REQUIRE(Error, "URL.revokeObjectURL() is not implemented"_kj);
+}
+
 jsg::Ref<URLSearchParams> URL::getSearchParams() {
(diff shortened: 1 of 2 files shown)


src/workerd/api/url-standard.h
```diff
@@ -205,6 +206,8 @@ class URL: public jsg::Object {
   // ].filter((test) => URL.canParse(test));
   //
   static bool canParse(kj::String url, jsg::Optional<kj::String> base = kj::none);
+  static jsg::JsString createObjectURL(jsg::Lock& js, kj::OneOf<workerd::api::File, workerd::api::Blob> object);
```
jasnell, line 209:
```suggestion
  static jsg::JsString createObjectURL(jsg::Lock& js, kj::OneOf<jsg::Ref<File>, jsg::Ref<Blob>> object);
```

jasnell (approved):
Needs `just f` but otherwise LGTM

src/workerd/api/url-standard.c++
```diff
@@ -53,6 +53,17 @@ bool URL::canParse(kj::String url, jsg::Optional<kj::String> maybeBase) {
   return jsg::Url::canParse(url, maybeBase.map([](kj::String& str) { return str.asPtr(); }));
 }
 
+jsg::JsString URL::createObjectURL(
+    jsg::Lock& js, kj::OneOf<jsg::Ref<File>, jsg::Ref<Blob>> object) {
+  // TODO(soon): Implement this
```
jasnell, line 58: Might be worth expanding a bit on what needs to be considered to implement this. It's not trivial by any stretch.

The `createObjectURL(...)` method is expected to store the `Blob` or `File` persistently in some scope such that the returned blob URL can be used to fetch it from that scope. In browsers and Node.js this is relatively straightforward. In workers, however... would that scope be expected to be single-metal? colo-wide? fleet-wide? If the URL is passed to a fetch on a different metal would it be expected to work? Where is the data actually stored? Is it volative in memory such that these URLs are also fully volatile? If they are fully volatile is there a risk of collisions? As so on... Implementing these is not trivial and we cannot simply copy what Node.js has done.

anonrig: Agreed. I'll be creating an internal proposal to see what our options are.

PR #2975, opened 2024-10-22
Switch back to id for Workflows binding

sidharthachatterjee:


danlapid (approved):

types/defines/workflows.d.ts
```diff
@@ -35,9 +28,9 @@ declare abstract class Workflow {
 
 interface WorkflowInstanceCreateOptions {
   /**
-   * A name for your Workflow instance. Must be unique within the Workflow.
+   * An id for your Workflow instance. Must be unique within the Workflow.
```
a-robinson, line 31: It'd be nice to mention that if an ID isn't specified, a unique ID will be generated for you. It looks like you already did that in the file above (src/cloudflare/internal/workflows.d.ts), just missed it here.

But this is fine to merge either way 👍

a-robinson (approved):

PR #2976, opened 2024-10-22
add built-in modules benchmark

anonrig:


src/workerd/tests/bench-builtin-modules.c++
```diff
@@ -0,0 +1,57 @@
...
+            for (let i = 0; i < 100; i++) {
+              buffer = await import("node:buffer");
+            }
+            assert.ok(buffer);
+            return new Response("OK");
+          },
+        };
+      )"_kj};
+    fixture = kj::heap<TestFixture>(kj::mv(params));
+  }
```
jasnell, line 40: My concern here is that if this is meant to test things like the efficacy of compile cache, it's not clear this will actually be a useful benchmark. The internal module will end up being compiled and cached only once per worker isolate, with the compile cache being populated and reused only when another isolate is loaded that imports the same module. We wouldn't see any performance impact on the first isolate and would only see it on the subsequent additional isolates as reduced cold start times.

src/workerd/tests/bench-builtin-modules.c++
```diff
@@ -0,0 +1,57 @@
...
+  void SetUp(benchmark::State& state) noexcept(true) override {
+    static auto flags = createCompatibilityFlags();
+    TestFixture::SetupParams params = {.featureFlags = flags.asReader(), .mainModuleSource = R"(
+        import assert from "node:assert";
+        export default {
+          async fetch(request) {
+            let buffer;
+            for (let i = 0; i < 100; i++) {
+              buffer = await import("node:buffer");
+            }
```
jasnell, line 33: Note that the `import(...)` will only be evaluated once, then cached, which means this is really only benchmarking the initial load once then the other 99 times we're just pulling the result from v8's cache.

anonrig: So, higher the count, higher the performance impact would be for compile cache. What about having this call only 2 times?

jasnell: I think what you actually need here is to have the test deploy two workers. One is a primer... you send a single request to it to perform the initial compile of `node:buffer` that generates the compile cache. Then, you send a single request to the second worker that also imports `node:buffer`. That one should see a reduced completion time on the very first `await import`.

PR #2983, opened 2024-10-23
Python external capnproto: fix "Exceeded message traversal limit" error

hoodmane:
If we reuse the bundle for too many different isolates we were getting an error. As suggested in the capnproto docs, this copies the data to a `MallocMessageBuilder` which then doesn't enforce the limit. Tested downstream.

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -25,7 +25,11 @@ void PyodideBundleManager::setPyodideBundleData(
     kj::String version, kj::Array<unsigned char> data) const {
   auto wordArray = kj::arrayPtr(
       reinterpret_cast<const capnp::word*>(data.begin()), data.size() / sizeof(capnp::word));
-  auto messageReader = kj::heap<capnp::FlatArrayMessageReader>(wordArray).attach(kj::mv(data));
+  // We're going to reuse this in the ModuleRegistry for every Python isolate, so set the traversal
+  // limit to infinity or else eventually a new Python isolate will fail.
+  auto messageReader = kj::heap<capnp::FlatArrayMessageReader>(
+      wordArray, capnp::ReaderOptions{.traversalLimitInWords = kj::maxValue})
+                           .attach(kj::mv(data));
   auto bundle = messageReader->getRoot<jsg::Bundle>();
   bundles.lockExclusive()->insert(
       kj::mv(version), {.messageReader = kj::mv(messageReader), .bundle = bundle});


anonrig (approved):

danlapid (approved):

kentonv:
You could also use `ReaderOptions` to set the traversal limit to UINT64_MAX.

hoodmane:
Okay I switched to setting `traversalLimitInWords = UINT64_MAX`. This is less code too. Thanks @kentonv!

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -25,7 +27,11 @@ void PyodideBundleManager::setPyodideBundleData(
     kj::String version, kj::Array<unsigned char> data) const {
   auto wordArray = kj::arrayPtr(
       reinterpret_cast<const capnp::word*>(data.begin()), data.size() / sizeof(capnp::word));
-  auto messageReader = kj::heap<capnp::FlatArrayMessageReader>(wordArray).attach(kj::mv(data));
+  // We're going to reuse this in the ModuleRegistry for every Python isolate, so set the traversal
+  // limit to infinity or else eventually a new Python isolate will fail.
+  auto messageReader = kj::heap<capnp::FlatArrayMessageReader>(
+      wordArray, capnp::ReaderOptions{.traversalLimitInWords = UINT64_MAX})
```
kentonv, line 33: You can use `kj::maxValue` here which automatically becomes whatever the maximum is for the type being assigned to. Avoids mistakes if the type isn't what you expect. Also don't need to include `stdint.h`.

```suggestion
      wordArray, capnp::ReaderOptions{.traversalLimitInWords = kj::maxValue})
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -25,7 +27,11 @@ void PyodideBundleManager::setPyodideBundleData(
     kj::String version, kj::Array<unsigned char> data) const {
   auto wordArray = kj::arrayPtr(
       reinterpret_cast<const capnp::word*>(data.begin()), data.size() / sizeof(capnp::word));
-  auto messageReader = kj::heap<capnp::FlatArrayMessageReader>(wordArray).attach(kj::mv(data));
+  // We're going to reuse this in the ModuleRegistry for every Python isolate, so set the traversal
```
kentonv, line 30: Note this is only safe if this data is trusted. I assume it is since it's shared python data, not from a user...

kentonv (approved):

dom96 (approved):
Thank you for the quick fix on this!

jasnell (approved):

PR #2986, opened 2024-10-23
cf-cache-level is the wrong header use cacheLevel instead

AdityaAtulTewari:


zaidoon1 (approved):

mikea (approved):

mikea:
would be nice to have a test failing though? (future PR)

AdityaAtulTewari:
The test has to be in edgeworker.

PR #2987, opened 2024-10-23
[rust] upstreaming rust/cxx integration

mikea:
TODO after this PR would be to reuse these parts of cxx support downstream. (will require small tweaks to init tokio),
but want to land this now to unblock.


mikea:
blocked by https://github.com/capnproto/capnproto/pull/2169

danlapid (approved):

mikea:
> This seems to break windows build 😢 

Yeah, windows rust build is  a bit tricky for some reason. Will require some fiddling.

fhanau:
> > This seems to break windows build 😢
> 
> Yeah, windows rust build is a bit tricky for some reason. Will require some fiddling.

Two things stick out: 
This happens when linking `external/capnp-cpp/src/kj/kj_c9e87b54.dll` – Bazel does not enable shared linkage by default on Windows and we have never enabled it, no idea why it is happening here.
The build is complaining about `undefined symbol: __declspec(dllimport) inet_pton` being missing. When we link Rust binaries within workerd, we have to add `ntdll.lib` on Windows to avoid missing symbol warnings. It appears that this might be needed here too (if we link statically, the `__declspec(dllimport)` part should not appear here).

Overall, rules_rust does not support Windows well unfortunately so I'd expect more such issues along the way (well, perhaps Windows is just not very suitable for software development in general).

danlapid:
> inet_pton

inet_pton is from Ws2_32.lib not ntdll.lib

src/rust/cxx-integration/BUILD
```diff
@@ -0,0 +1,21 @@
...
+
+wd_rust_crate(
+    name = "cxx-integration",
+    cxx_bridge_src = "lib.rs",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@capnp-cpp//src/kj",
+        # "//src/rust/process-tracing",
+        # "//src/rust/safe-libc",
+        # "//src/rust/version",
```
fhanau, line 11: Nit: It is doubtful if we'll need a version field as used internally anytime soon for workerd

src/rust/cxx-integration/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+        fn trigger_panic(msg: &str);
+    }
+
+    unsafe extern "C++" {
+        include!("rust/cxx-integration/cxx-bridge.h");
+    }
+}
+
+pub fn init() {
+    tokio::init();
```
fhanau, line 17: Do we plan to be using tokio soon within workerd? If it's only inited here and used in tests but not prod code it might be better to start depending on it later on

src/rust/cxx-integration-test/cxx-rust-integration-test.c++
```diff
@@ -0,0 +1,261 @@
...
+#include <rust/cxx.h>
+
+#include <signal.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include <kj/async.h>
+#include <kj/test.h>
+
+#include <exception>
```
fhanau, line 12: Nit: I know this code is just ported from downstream but these two headers are unused right?

fhanau (approved):

src/rust/cxx-integration/lib.rs
mikea: I want us to have the same set up so that there's less friction writing/moving code. E.g. if one decided upstream log fowarding or something :)

mikea:
> This happens when linking external/capnp-cpp/src/kj/kj_c9e87b54.dll – Bazel does not enable shared linkage by default on Windows and we have never enabled it, no idea why it is happening here.

maybe this is what rust_crate does on windows?

mikea:
I'll be CI-debugging windows build by commenting things out, sorry for the noise.

mikea:
Windows build issue is fixed by:
- upgrading to rust 1.82.0 (linker issue)
- switching to mainline `cxx` crate. I do not think it is possible to build workerd-cxx on windows using bazel, so I will try to publish to crates.io later. For now cxx will do to unlock everyone

PTAL @danlapid

danlapid (approved):

fhanau:
> Windows build issue is fixed by:
> 
> * upgrading to rust 1.82.0 (linker issue)
> * switching to mainline `cxx` crate. I do not think it is possible to build workerd-cxx on windows using bazel, so I will try to publish to crates.io later. For now cxx will do to unlock everyone
> 
> PTAL @danlapid

Glad you figured it out!

danlapid:
👑

danlapid:
@hoodmane we can integrate ruff now! 🤩

hoodmane:
Thanks @mikea !

mikea:
Np. I'll also follow up with wd_rust_binary tomorrow (me and @anonrig need it too)

PR #2981, opened 2024-10-22
[build] Roll bazel 7.3.1:7.4.1

fhanau:
- Enable BwoB on Windows again now that https://github.com/bazelbuild/bazel/issues/23462 has been fixed.

PR #2988, opened 2024-10-23
[build] Disable WebGPU for Windows non-release builds

fhanau:
The Windows build is currently much slower than the macOS/Linux builds on CI, especially on mostly cached builds.
Based on profiling on #2981, this is caused by the repo mapping stage for @down taking several minutes of Windows. Disable the experimental WebGPU feature on Windows to better meet compile time constraints. Release builds remain unaffected.


fhanau:
This does not seem to be as impactful as expected – will take some more work

PR #2989, opened 2024-10-23
[rust] update rust dependencies

mikea:
- moves all deps to single-digit like downstream
- added bunch of common crates needed in follow up PRs

just getting noisy change out of the way


danlapid (approved):

fhanau (approved):
We're adding a bunch of large dependencies here – using them all within workerd may cause headaches in terms of CI compile time overhead, but that's a concern for another day.

PR #2990, opened 2024-10-23
move compile_cache to jsg/compile_cache

anonrig:
Moves compile cache to it's own file to prepare for precompiling node internals.

### DIFF
--- a/src/workerd/jsg/BUILD.bazel
+++ b/src/workerd/jsg/BUILD.bazel
@@ -11,6 +11,7 @@ wd_cc_library(
     srcs = [
         "async-context.c++",
         "buffersource.c++",
+        "compile-cache.c++",
         "dom-exception.c++",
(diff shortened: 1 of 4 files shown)


jasnell (approved):

src/workerd/jsg/BUILD.bazel
```diff
@@ -11,6 +11,7 @@ wd_cc_library(
     srcs = [
         "async-context.c++",
         "buffersource.c++",
+        "compile-cache.c++",
```
danlapid, line 14: Is it a good opportunity to create to a separate compilation target maybe?

anonrig: I couldn't find a good way to seperate to a different target since compile cache depends on `jsg` and `jsg` depends on compile cache.

PR #2991, opened 2024-10-24
[build] Disentangle r2*.c++, memory-cache.c++ from io target

fhanau:
- Mark tools:dump_api_ast as no-downstream – this target is expensive and serves no purpose for the downstream build.
- Disable default dependencies for tracing to drop superfluous crate and match downstream


fhanau:
- [ ] add downstream PR to update bazel deps there as needed

jasnell (approved):

PR #2995, opened 2024-10-24
Implements more methods on Python Response API and adds FormData API.

dom96:
The FormData API does not have a native Python equivalent, but it is a very useful API in Workers. This PR implements it and ties it into the Response API implementation.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -27,6 +27,7 @@ ignore = [
   "PLR2004", # Magic value used in comparison
   "TRY003", # Avoid specifying long messages outside the exception class
   "UP038",  # Use X | Y in isinstance check instead of (X, Y)
+  "PLR0911", # too many return statements
 ]
 
 [lint.per-file-ignores]
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -1,19 +1,21 @@
 # This module defines a Workers API for Python. It is similar to the API provided by
 # JS Workers, but with changes and additions to be more idiomatic to the Python
 # programming language.
+from collections.abc import Generator, MutableMapping
 from http import HTTPMethod, HTTPStatus
 from typing import TypedDict, Unpack
 
 import js
 
 import pyodide.http
+from pyodide.ffi import JsException, to_js
 from pyodide.http import pyfetch
 
 JSBody = (
     "js.Blob | js.ArrayBuffer | js.TypedArray | js.DataView | js.FormData |"
     "js.ReadableStream | js.URLSearchParams"
 )
-Body = "str | JSBody"
+Body = "str | FormData | JSBody"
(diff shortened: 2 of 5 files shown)


.ruff.toml
```diff
@@ -27,6 +27,7 @@ ignore = [
   "PLR2004", # Magic value used in comparison
   "TRY003", # Avoid specifying long messages outside the exception class
   "UP038",  # Use X | Y in isinstance check instead of (X, Y)
+  "PLR0911", # too many return statements
```
danlapid, line 30: Can we inline this ignore to the specific statement where we want to ignore this error type?

dom96: we can, but I think we can ignore it everywhere in the same way we are ignoring "too many branches"

src/pyodide/internal/workers.py
```diff
@@ -71,8 +116,78 @@ def __init__(
...
+            )
+        except JsException as exc:
+            raise _to_python_exception(exc) from exc
+
+
+# TODO: Implement pythonic blob API
+FormDataValue = "str | js.Blob"
+
+
+class FormData:
```
hoodmane, line 156: Maybe this should subclass `abc.MutableMapping[str, FormDataValue]`?

hoodmane (approved):
Looks good to me. I left some minor comments.

src/workerd/server/tests/python/sdk/worker.py
```diff
@@ -1,8 +1,12 @@
...
-from cloudflare.workers import Response, fetch
+import js
 
+from cloudflare.workers import FormData, Response, fetch
 
+
+# Each path in this handler is its own test. The URLs that are being fetched
+# here are defined in server.py.
 async def on_fetch(request):
     if request.url.endswith("/modify"):
```
hoodmane, line 11: Would be nice to refactor this to use a dispatch table and a decorator.

hoodmane: Maybe I'll make a followup PR to rearrange this set of tests into a tidier form.

src/pyodide/internal/workers.py
```diff
@@ -71,8 +116,78 @@ def __init__(
...
+    def __init__(self, form_data: "js.FormData | None | dict[str, FormDataValue]"):
+        if form_data:
+            if isinstance(form_data, dict):
+                self.js_form_data = js.FormData.new()
+                for item in form_data.items():
+                    self.js_form_data.append(item[0], item[1])
```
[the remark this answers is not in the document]
dom96: it does not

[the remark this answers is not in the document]
hoodmane: Ah indeed looks like the `FormData` constructor just does not work that way.

PR #2996, opened 2024-10-24
Invalidate Python caches earlier so packages are picked up.

dom96:
Right now the validator fails to validate a Python Worker that imports packages when the validator is using the baseline snapshot. This should fix it.


danlapid (approved):

hoodmane (approved):
Thanks!

PR #2997, opened 2024-10-24
feat(build): add support for windows arm64

p-romain:
Hi, I will try to add the support for windows arm64 (binary build), since we are now few users running a Snapdragon X Elite computer.


p-romain:
First requirement is OK, the windows-2022 runner has `Microsoft.VisualStudio.Component.VC.14.29.16.11.ARM64` installed, probably via cross-compiler, so should work. Now I have to see if the `.bazelrc` need to be adapted.

p-romain:

fhanau:
Thank you for your PR! Unfortunately, this might be difficult for us to support in the near-term:
- Cross-compilation under bazel is a complex beast. For macOS this is well supported when using [bazelbuild/apple_support](https://github.com/bazelbuild/apple_support), but last time I tried I couldn't get cross-compilation to work on Linux despite significant changes (Rust support was one issue). Based on that I assume that this won't work as-is and that developing cross-compilation support would be lots of effort, especially since none .
- workerd Windows releases are built using free-tier GitHub actions runners for the time being, support for Linux and Windows arm64 runners is not available but expected by EOY. That should provide a much clearer path to supporting arm64 Windows.

p-romain:
Hi @fhanau thank you very much for all the explanations. Yes, this makes sense, I also noticed this on other projects hosted in Gitlab: we're waiting the GA of arm64 runners. I've also noticed that the windows 64 binary works on windows arm64 in emulation, so this will be my workaround in the meantime. At this time, because of workerd, we're not able to use wrangler 3 (out-of-the-box) on windows arm64.

mikea:
FWIW I would welcome a switch to windows cross-compilation if someone was able to do it. We've all spent enormous time banging our heads against windows. I remember that windows cross compilation was an explicit goal of clang team. Maybe it is finally possible with clang 20?

p-romain:
Finally I've experienced some issues with the x86 version emulated under arm64. The version 2 was not compatible with my needs (nodejs functions support). I was only able to do something with WSL. But after that, I've experienced lot of issues with Prisma, pg driver, Hono under Wrangler. Finally the 1 Mo bundle limitation has achieved me. I don't consider anymore cf worker for my backed, I've switched back to Symfony.

PR #2999, opened 2024-10-24
Add an option to newContext to disable memoization of global context …

danlapid:
…template.

For Python Isolate Pools we want to instantiate emscripten in a separate context on isolate creation (before we have compatibility flags) and later after receiving the compatibility flags create a new context. 
One blocker to this approach is all the templates that are instantiated in the first context are memoized. 
This PR allows us to create a new context without memoizing the global context templates. 
This feature is off by default and will be enabled in a future PR for python isolates only.

### DIFF
--- a/src/workerd/jsg/resource.h
+++ b/src/workerd/jsg/resource.h
@@ -458,14 +458,15 @@ template <typename TypeWrapper,
     const char* methodName,
     typename Method,
     Method method,
-    bool isContext>
+    bool isContext,
+    bool memoize>
 struct GetterCallback;
 
 #define JSG_DEFINE_GETTER_CALLBACK_STRUCTS(...)                                                    \
   template <typename TypeWrapper, const char* methodName, typename T, typename Ret,                \
-      typename... Args, Ret (T::*method)(Args...) __VA_ARGS__, bool isContext>                     \
+      typename... Args, Ret (T::*method)(Args...) __VA_ARGS__, bool isContext, bool memoize>       \
   struct GetterCallback<TypeWrapper, methodName, Ret (T::*)(Args...) __VA_ARGS__, method,          \
-      isContext> {                                                                                 \
+      isContext, memoize> {                                                                        \
     static constexpr bool enumerable = true;                                                       \
     static void callback(v8::Local<v8::Name>, const v8::PropertyCallbackInfo<v8::Value>& info) {   \
       liftKj(info, [&]() {                                                                         \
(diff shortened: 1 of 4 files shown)


src/workerd/jsg/resource.h
```diff
@@ -527,14 +533,15 @@ template <typename TypeWrapper,
     const char* methodName,
     typename Method,
     Method method,
-    bool isContext>
+    bool isContext,
+    bool memoize>
```
jasnell, line 537: Non-blocking nit... really not a big fan of bool args like this. Much prefer enums for this type of case. Won't block on it tho

src/workerd/jsg/resource.h
```diff
@@ -1492,12 +1549,14 @@ class ResourceWrapper {
...
 
-  template <bool isContext = false>
+  template <bool isContext = false, bool memoize = true>
   v8::Local<v8::FunctionTemplate> getTemplate(v8::Isolate* isolate, T*) {
     v8::Global<v8::FunctionTemplate>& slot = isContext ? contextConstructor : memoizedConstructor;
     if (slot.IsEmpty()) {
-      auto result = makeConstructor<isContext>(isolate);
-      slot.Reset(isolate, result);
+      auto result = makeConstructor<isContext, memoize>(isolate);
+      if (memoize) {
```
jasnell, line 1557: Can this be `if constexpr (memoize) {`?

danlapid: yep, no difference as compiler will optimize it away but I generally like the if constexpr feature so lets use it 😄

danlapid:
@jasnell is this good to go?

src/workerd/jsg/resource.h
```diff
@@ -1411,9 +1460,17 @@ class ResourceWrapper {
...
     // Fortunately, for types that are never used as the global object, we never have to
     // instantiate the `isContext = true` branch.
+    //
+    // When creating a new context the global context template will be memoized for future reuse in
+    // additional contexts. The global context can depend on `configuration`. For Python Isolate
+    // Pools we want to create a new context and initialize some code in it and later when a request
+    // arrives, edit the configuration and create a new context. To achieve this we create the first
+    // context with `memoize=false` to disable memoization of the global context created for the
+    // first context. Then we create a new context with `memoize=true` to enable memoization of the
+    // main context created after a request was received.
```
dom96, line 1470: Probably a stupid question, but I wonder why we cannot overwrite the memoized context after it has been memoized?

dom96 (commented):
Changes LGTM assuming this is the cleanest way to achieve what we need. I don't have a lot of context on this code so will leave to someone else to approve.

src/workerd/jsg/resource.h
danlapid: That would require the inverse of the same logic changing the code from that existed before this change rather than adding a different code flow for our exceptional case.

jasnell (approved):
LGTM but there's a merge conflict and let's be sure to run an internal CI job before merging to make sure there are no issues on the internal repo

PR #3004, opened 2024-10-25
create compile caches on compile step

anonrig:
> Work in progress. 

The goal of this pull-request is to create an intermediary step between TypeScript compiler and Workerd compilation step to create compile caches. Compile caches will be used by the main binary to improve the startup time of the built-in modules (and only the built-ins).

Technical details:

The newly introduced binary accepts a string to the path of a file list that contains every path to be compile cached. With the current change the build process will slightly change to the following:

1. Workerd runs typescript and compiles TS files.
2. Each node.js built-in will be passed to the newly introduced binary to create a compile cache for the given paths.
3. Workerd will be compiled and it will use the compile cache output from the creator binary if exists.

Limitations:
- Currently, we only enable it for Linux. We can enable it for Windows, but since we're only using Linux on production, I don't think it's beneficial to support Windows.
- We need to handle downstream repository and the isolate creation differences.

### DIFF
--- a/build/wd_compile_cache.bzl
+++ b/build/wd_compile_cache.bzl
@@ -5,41 +5,48 @@ Bazel rules for generating compilation caches from a list of input files.
 load("@bazel_skylib//rules:run_binary.bzl", "run_binary")
 
 def _gen_compile_cache_impl(ctx):
-    file_list_path = "{}-compile-cache-file-list".format(ctx.label)
-    file_list = ctx.actions.declare_file(file_list_path)
+    file_list = ctx.actions.declare_file("in")
 
     # Get the File objects from the labels
-    input_files = []
+    srcs = []
     for src in ctx.attr.srcs:
-        input_files.extend(src.files.to_list())
+        srcs.extend(src.files.to_list())
 
     ctx.actions.write(
         output = file_list,
-        content = "\n".join([f.short_path for f in input_files]) + "\n",
+        content = "\n".join([f.path for f in srcs]) + "\n",
     )
 
-    output = ctx.actions.declare_file(ctx.attr.output)
+    out = ctx.outputs.out
+
+    args = ctx.actions.args()
+    args.add(out)
+    args.add(file_list)
 
     ctx.actions.run(
-        outputs = [output],
-        inputs = input_files + [file_list],
-        arguments = [file_list.path],
+        outputs = [out],
+        inputs = [file_list] + srcs,
+        arguments = [args],
         executable = ctx.executable._tool,
         tools = [ctx.executable._tool],
     )
+
+    return [
+        DefaultInfo(files = depset([out])),
+    ]
 
 gen_compile_cache = rule(
     implementation = _gen_compile_cache_impl,
     attrs = {
         "srcs": attr.label_list(mandatory = True, allow_files = True),
-        "output": attr.string(mandatory = True),
+        "out": attr.output(mandatory = True),
         "_tool": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "//src/workerd/tools:create_compile_cache"),
     },
 )
 
 def wd_compile_cache(name, srcs):
     gen_compile_cache(
-        name = name,
+        name = name + "_gen",
         srcs = srcs,
-        output = "{}-compile-cache-output".format(name),
+        out = name,
     )
--- a/build/wd_compile_cache.bzl
+++ b/build/wd_compile_cache.bzl
@@ -2,7 +2,8 @@
 Bazel rules for generating compilation caches from a list of input files.
 """
 
-load("@bazel_skylib//rules:run_binary.bzl", "run_binary")
+def compile_cache_name(file_path):
+    return file_path + "_cache"
 
 def _gen_compile_cache_impl(ctx):
     file_list = ctx.actions.declare_file("in")
@@ -11,7 +12,9 @@ def _gen_compile_cache_impl(ctx):
     srcs = []
     for src in ctx.attr.srcs:
         srcs.extend(src.files.to_list())
-    outs = [ctx.actions.declare_file(src.path + "_cache") for src in srcs]
+
+    outs = [compile_cache_name(src.basename) for src in srcs]
+    outs = [ctx.actions.declare_file(compile_cache_name(src.basename)) for src in srcs]
 
     content = []
     for i in range(0, len(srcs)):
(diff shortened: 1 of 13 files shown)


src/workerd/tools/create-compile-cache.c++
```diff
@@ -0,0 +1,131 @@
+#include "fcntl.h"
+
+#include <workerd/io/limit-enforcer.h>
+#include <workerd/jsg/compile-cache.h>
+#include <workerd/jsg/setup.h>
+
+#include <kj/filesystem.h>
+#include <kj/main.h>
+
+#include <iostream>
```
jasnell, line 10: Nit: please add a comment at the top of this explaining its purpose and use.

src/workerd/tools/create-compile-cache.c++
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  // Key is the path of the file, and value is the content.
+  kj::Vector<kj::Tuple<kj::String, kj::Array<kj::byte>>> file_contents{};
+};
+
+}  // namespace
+}  // namespace workerd::api
+
+KJ_MAIN(workerd::api::CompileCacheCreator)
```
jasnell, line 131:
```suggestion
KJ_MAIN(workerd::api::CompileCacheCreator)

```

build/wd_compile_cache.bzl
```diff
@@ -0,0 +1,57 @@
...
+
+gen_compile_cache = rule(
+    implementation = _gen_compile_cache_impl,
+    attrs = {
+        "name": attr.string(mandatory = True, doc = "name"),
+        "data": attr.string_list(mandatory = True, doc = "files to create compile cache for"),
+    },
+    target_compatible_with = select({
+        "@platforms//os:linux": [],
+        "//conditions:default": ["@platforms//:incompatible"],
```
mikea, line 41: sorry, is there any real reason we avoid mac+windows here? I don't see why wouldn't this work?

build/wd_compile_cache.bzl
```diff
@@ -0,0 +1,57 @@
...
+    ctx.actions.write(
+        output = file_list_path,
+        content = "\n".join(ctx.attr.data) + "\n",
+    )
+    run_binary(
+        ctx.attr.name,
+        srcs = [],
+        outs = [compile_cache_output],
+        args = [file_list_path],
+        tool = "create_compile_cache_bin",
```
mikea, line 29: the tool should be passed as a label through private attribute. Example from gen_capnp.bzl:

```
        "_capnpc": attr.label(executable = True, allow_single_file = True, cfg = "exec", default = "@capnp-cpp//src/capnp:capnp_tool"),
```

this needs to go to attr and you need to use it here properly.

build/wd_compile_cache.bzl
```diff
@@ -0,0 +1,57 @@
...
+
+    file_list_path = "{}-compile-cache-file-list".format(ctx.attr.name)
+    compile_cache_output = "{}-compile-cache".format(ctx.attr.name)
+    ctx.actions.write(
+        output = file_list_path,
+        content = "\n".join(ctx.attr.data) + "\n",
+    )
+    run_binary(
+        ctx.attr.name,
+        srcs = [],
```
mikea, line 26: all input files must be passed here as `src`

build/wd_compile_cache.bzl
```diff
@@ -0,0 +1,57 @@
...
+
+def _gen_compile_cache_impl(ctx):
+    """Implementation of the gen_compile_cache rule.
+
+    This function:
+    1. Creates a file containing the list of input files
+    2. Runs a binary to generate the compilation cache
+    """
+
+    file_list_path = "{}-compile-cache-file-list".format(ctx.attr.name)
```
mikea, line 18: since this is a rule I suggest you have attr for each one of these, and have actual substitutions in the macro.

build/wd_compile_cache.bzl
anonrig: my initial implementation didn't use KJ_MAIN and I wanted to avoid Windows specific code because of that. since the implementation is now different, I guess there is no reason to avoid mac+windows.

build/wd_compile_cache.bzl
```diff
@@ -0,0 +1,55 @@
...
+        tool = ctx.attr._tool,
+    )
+    return compile_cache_output
+
+gen_compile_cache = rule(
+    implementation = _gen_compile_cache_impl,
+    attrs = {
+        "name": attr.string(mandatory = True, doc = "name"),
+        "srcs": attr.string_list(mandatory = True, doc = "files to create compile cache for"),
+        "file_list_path": attr.string(mandatory = True, doc = "temporary path for the file list"),
```
mikea, line 35: sorry for noise, but temp files don't have to be here. I missed the fact that this list is not used anywhere else. The output does need to be here.

src/workerd/tools/BUILD.bazel
```diff
@@ -71,3 +71,16 @@ run_binary(
     tool = "param_extractor_bin",
     visibility = ["//visibility:public"],
 )
+
+# This binary is used to generate compile cache for workerd.
+# This will later be used by workerd to access compile caches of JS internals.
+# NOTE: We only enable compile cache for linux at the moment.
+cc_binary(
```
mikea, line 78: wd_cc_binary please

src/workerd/tools/BUILD.bazel
```diff
@@ -71,3 +71,16 @@ run_binary(
...
+    name = "create_compile_cache_bin",
```
mikea, line 79: just `create_compile_cache` for name

src/workerd/tools/create-compile-cache.c++
```diff
@@ -0,0 +1,113 @@
+#include "fcntl.h"
```
mikea, line 1: <fctnl.h>

mikea (approved):

src/workerd/tools/create-compile-cache.c++
anonrig: Unfortunately, it doesn't work:

```
src/workerd/tools/create-compile-cache.c++:1:10: fatal error: 'fctnl.h' file not found
    1 | #include <fctnl.h>
      |          ^~~~~~~~~
```

src/workerd/jsg/compile-cache.c++
```diff
@@ -28,4 +34,21 @@ kj::Maybe<CompileCache::Data&> CompileCache::find(kj::StringPtr key) const {
   return kj::none;
 }
 
+void CompileCache::serialize(capnp::MessageBuilder& message) const {
+  capnp::DynamicStruct::Builder builder = message.initRoot<workerd::tools::CompileCache>();
```
kentonv, line 38: Why are you using the capnp dynamic API here? You've included the generated-code header so why not use it?

src/workerd/jsg/compile-cache.c++
```diff
@@ -28,4 +34,21 @@ kj::Maybe<CompileCache::Data&> CompileCache::find(kj::StringPtr key) const {
...
+  for (auto& current: *lock) {
+    auto entry = entries[i].as<capnp::DynamicStruct>();
+    capnp::Text::Reader key(current.key);
+    entry.set("path"_kj, key);
+    capnp::Data::Reader data(kj::ArrayPtr(current.value.data, current.value.length));
+    entry.set("data"_kj, data);
+    i++;
+  }
+}
+
```
kentonv, line 53:
```suggestion
  workerd::tools::CompileCache::Builder builder = message.initRoot<workerd::tools::CompileCache>();

  auto lock = cache.lockShared();
  auto entries = builder.initEntries(lock->size());

  size_t i = 0;
  for (auto& current: *lock) {
    auto entry = entries[i];
    entry.setPath(current.key);
    entry.setData(kj::ArrayPtr(current.value.data, current.value.length));
    i++;
  }
}

```

src/workerd/jsg/compile-cache.c++
anonrig: You're absolutely right. Thank you for the review.

anonrig:
Since multiple PRs have landed, we can close this pull-request.

PR #3005, opened 2024-10-25
Move TraceId implementation to workerd + some cleanups

jasnell:
The TraceId implementation will be reused for the new streaming tail worker implementation as well as the internal jaeger tracing impl. This moves the implementation out to workerd so it can be used there. A separate internal PR will update uses of TraceId to use this and remove the old internal implementation

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -5,14 +5,157 @@
 #include <workerd/io/trace.h>
 #include <workerd/util/thread-scopes.h>
 
+#include <openssl/rand.h>
+
 #include <capnp/message.h>
 #include <capnp/schema.h>
+#include <kj/compat/http.h>
 #include <kj/debug.h>
 #include <kj/time.h>
 
 #include <cstdlib>
 
 namespace workerd {
+
+namespace tracing {
+namespace {
+kj::Maybe<kj::uint> tryFromHexDigit(char c) {
+  if ('0' <= c && c <= '9') {
+    return c - '0';
+  } else if ('a' <= c && c <= 'f') {
+    return c - ('a' - 10);
+  } else if ('A' <= c && c <= 'F') {
+    return c - ('A' - 10);
+  } else {
+    return kj::none;
+  }
+}
+
+kj::Maybe<uint64_t> hexToUint64(kj::ArrayPtr<const char> s) {
+  KJ_ASSERT(s.size() <= 16);
+  uint64_t value = 0;
+  for (auto ch: s) {
+    KJ_IF_SOME(d, tryFromHexDigit(ch)) {
+      value = (value << 4) + d;
+    } else {
+      return kj::none;
+    }
+  }
+  return value;
+}
+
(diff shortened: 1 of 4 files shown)


src/workerd/io/trace.c++
```diff
@@ -5,15 +5,159 @@
...
+      value = (value << 4) + d;
+    } else {
+      return kj::none;
+    }
+  }
+  return value;
+}
+
+void addHex(kj::Vector<char>& out, uint64_t v) {
+  const char HEX_DIGITS[] = "0123456789abcdef";
```
anonrig, line 48:
```suggestion
  constexpr char HEX_DIGITS[] = "0123456789abcdef";
```

src/workerd/io/trace.c++
```diff
@@ -5,15 +5,159 @@
...
+  for (int i = 0; i < 16; ++i) {
+    out.add(HEX_DIGITS[v >> (64 - 4)]);
+    v = v << 4;
+  }
+};
+
+void addBigEndianBytes(kj::Vector<byte>& out, uint64_t v) {
+  for (int i = 0; i < 8; ++i) {
+    out.add(v >> (64 - 8));
```
anonrig, line 57: Nit: Before for loop:

out.reserve(out.size() + 8);

src/workerd/io/trace.c++
```diff
@@ -5,15 +5,159 @@
...
+}
+
+// Reference: https://github.com/jaegertracing/jaeger/blob/e46f8737/model/ids.go#L50
+kj::String TraceId::toGoString() const {
+  if (high == 0) {
+    kj::Vector<char> s(17);
+    addHex(s, low);
+    s.add('\0');
+    return kj::String(s.releaseAsArray());
+  } else {
```
anonrig, line 89: Nit: This pattern is repeated almost all of workerd, and I would like to raise my voice even though you're just moving the code. 

----

This else is not needed since previous if statement already returns. 

----

I can finally rest.

src/workerd/io/trace.c++
```diff
@@ -5,15 +5,159 @@
...
+// Reference https://www.w3.org/TR/trace-context/#trace-id
+kj::String TraceId::toW3C() const {
+  kj::Vector<char> s(32);
+  addHex(s, high);
+  addHex(s, low);
+  return kj::str(s.releaseAsArray());
+}
+
+TraceId TraceId::fromEntropy(kj::Maybe<kj::EntropySource&> entropySource) {
+  if (isPredictableModeForTest()) {
```
anonrig, line 131: For a future pull request:

We can make this function constexpr.

anonrig (approved):

src/workerd/io/trace.c++
jasnell: The vector passed in already has reserved size. Given that I'm moving this existing implementation out from the internal code base I'm going to leave it as is for now. We can take another pass on speeding it up later.

PR #3013, opened 2024-10-27
Bazel 8 support

fhanau:
- Remove incompatible flag overrides that are now on-by-default in Bazel
- Not using bzlmod so far, just adding placeholder MODULE.bazel file
- Apply workarounds for Windows symlink bugs


src/pyodide/helpers.bzl
```diff
@@ -131,12 +131,9 @@ def pyodide_static():
         [
             "internal/*.ts",
             "internal/topLevelEntropy/*.ts",
-            # The pool directory is only needed by typescript, it shouldn't be used at runtime.
```
fhanau, line 134: Changes in this file were added to work around Windows compile errors. I did not check if it is safe to drop them after adding the workaround, but I think they are a win either way, basically dropping build inputs where we don't need them.

anonrig (approved):

src/pyodide/helpers.bzl
danlapid: @hoodmane

src/pyodide/helpers.bzl
```diff
@@ -270,9 +267,11 @@ def _python_bundle(version, *, pyodide_asm_wasm = None, pyodide_asm_js = None, p
     else:
         esbuild(
             name = "emscriptenSetup@" + version,
+            # exclude emscriptenSetup from source set so that rules_ts won't also try to create a JS output
+            # for it. The file is provided in entry_point instead.
             srcs = native.glob([
                 "internal/pool/*.ts",
-            ]) + [
+            ], exclude = ["internal/pool/emscriptenSetup.ts"]) + [
```
danlapid, line 274: @hoodmane

Issue #2993, opened 2024-10-24
🐛 Bug Report — Runtime APIs - Durable Objects with SQLite storage - calling deleteAll() in alarm handler causes internal error


lucidplot:
I think I've found a bug in the SQLite storage backend for Durable Objects. When I call `deleteAll()` from an alarm handler, it causes an internal error. Which means the alarm handler fails and keeps getting retried.

However, if I do `await this.ctx.storage.deleteAlarm();` before calling `deleteAll()`, it works fine.

(I'm using this as a workaround with no problem. But it seems like a bug.)

Use case: I want to use alarms to delete all data, for eg a verify link that auto-expires.

FWIW this problem didn't seem to happen when I was using the KV storage backend.

joshthoward:
@lucidplot we've confirmed this and are working on a fix. Thank you for reporting this issue!

stayallive:
I think my issue is related or at least related to the fix. I see the PR was merged but I am assuming this is not live yet. I am wondering if this will also clear up the affected DO's since I have quite a substantial number of them stuck in a retry loop. The constructor is called but the alarm callback is never fired for these and delete storage or alarm for the DO IDs have no effect. It's starting to rack up to quite a number of requests / CPU time (I'm assuming).

![image](https://github.com/user-attachments/assets/f1b828f9-4e03-4498-b7d8-02e12eaf9651)

jclee:
> I think my issue is related or at least related to the fix. I see the PR was merged but I am assuming this is not live yet.

Correct... we are in the process of rolling out a version of the runtime containing the fix (as well as other updates), but at the time of your comment, it was not yet present in production.

If you wanted to address the error prior to the fix being fully deployed, one possible workaround (as lucidplot pointed out above) is for the Durable Object to explicitly delete the alarm before calling `deleteAll()`.  It turns out that the bug only manifests when `deleteAll()` is called while an alarm is in place -- which is always the case during alarm handlers, if the alarm is not explicitly deleted, since the handler-running code does not actually clear the alarm until the handler successfully completes.

stayallive:
I have been doing that, calling deleteAlarm and after deleteAll. I did that via a RPC call to a function that looks like:

```js
	public async destroy(): Promise<void> {
		await this.ctx.blockConcurrencyWhile(async () => {
			await this.ctx.storage.deleteAlarm();
			await this.ctx.storage.deleteAll();
		});
	}
```

[...]

PR #3018, opened 2024-10-28
Sqlite-backed DOs: Allow calling deleteAll() inside an alarm handler

jclee:
And also fix a potential issue with alarm exception propagation.

Should address https://github.com/cloudflare/workerd/issues/2993

---

* SRS actors: Ensure that deleteAll() always has a commit scheduled during reset

  ...particularly, in the case where an alarm is set.  Should fix errors seen when calling deleteAll() during an alarm handler.

* SRS actors: Don't attempt alarm deletion at end of handler if the output gate is already broken

  May fix cases where a broken output gate could cause an exception to propagate beyond the regular alarm output gate handling.

### DIFF
--- a/src/workerd/io/actor-sqlite-test.c++
+++ b/src/workerd/io/actor-sqlite-test.c++
@@ -94,7 +94,13 @@ struct ActorSqliteTest final {
         db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY),
         actor(kj::attachRef(db), gate, KJ_BIND_METHOD(*this, commitCallback), hooks),
         gateBrokenPromise(options.monitorOutputGate ? eagerlyReportExceptions(gate.onBroken())
-                                                    : kj::Promise<void>(kj::READY_NOW)) {}
+                                                    : kj::Promise<void>(kj::READY_NOW)) {
+    db.afterReset([this](SqliteDatabase& cbDb) {
+      KJ_ASSERT(&db == &cbDb);
+      KJ_ASSERT(actor.isCommitScheduled(),
+          "actor should have a commit scheduled during afterReset() callback");
+    });
+  }
 
   ~ActorSqliteTest() noexcept(false) {
     if (!unwindDetector.isUnwinding()) {
@@ -924,7 +930,9 @@ KJ_TEST("calling deleteAll() preserves alarm state if alarm is set") {
   test.pollAndExpectCalls({});
   KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
+  KJ_ASSERT(!test.actor.isCommitScheduled());
   ActorCache::DeleteAllResults results = test.actor.deleteAll({});
+  KJ_ASSERT(test.actor.isCommitScheduled());
   KJ_ASSERT(results.backpressure == kj::none);
   KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
@@ -954,7 +962,9 @@ KJ_TEST("calling deleteAll() preserves alarm state if alarm is not set") {
   test.pollAndExpectCalls({});
   KJ_ASSERT(expectSync(test.getAlarm()) == kj::none);
 
+  KJ_ASSERT(!test.actor.isCommitScheduled());
   ActorCache::DeleteAllResults results = test.actor.deleteAll({});
+  KJ_ASSERT(test.actor.isCommitScheduled());
   KJ_ASSERT(results.backpressure == kj::none);
   KJ_ASSERT(expectSync(test.getAlarm()) == kj::none);
 
@@ -973,10 +983,13 @@ KJ_TEST("calling deleteAll() preserves alarm state if alarm is not set") {
 KJ_TEST("calling deleteAll() during an implicit transaction preserves alarm state") {
   ActorSqliteTest test;
 
+  KJ_ASSERT(!test.actor.isCommitScheduled());
+
   // Initialize alarm state to 1ms.
   test.setAlarm(oneMs);
 
   ActorCache::DeleteAllResults results = test.actor.deleteAll({});
+  KJ_ASSERT(test.actor.isCommitScheduled());
   KJ_ASSERT(results.backpressure == kj::none);
   KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
--- a/src/workerd/io/actor-sqlite-test.c++
+++ b/src/workerd/io/actor-sqlite-test.c++
@@ -930,21 +930,48 @@ KJ_TEST("calling deleteAll() preserves alarm state if alarm is set") {
   test.pollAndExpectCalls({});
   KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
-  KJ_ASSERT(!test.actor.isCommitScheduled());
-  ActorCache::DeleteAllResults results = test.actor.deleteAll({});
-  KJ_ASSERT(test.actor.isCommitScheduled());
-  KJ_ASSERT(results.backpressure == kj::none);
-  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+  {
+    KJ_ASSERT(!test.actor.isCommitScheduled());
+    ActorCache::DeleteAllResults results = test.actor.deleteAll({});
+    KJ_ASSERT(test.actor.isCommitScheduled());
+    KJ_ASSERT(results.backpressure == kj::none);
+    KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
-  auto commitFulfiller = kj::mv(test.pollAndExpectCalls({"commit"})[0]);
-  KJ_ASSERT(results.count.wait(test.ws) == 0);
-  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+    auto commitFulfiller = kj::mv(test.pollAndExpectCalls({"commit"})[0]);
+    KJ_ASSERT(results.count.wait(test.ws) == 0);
+    KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
-  commitFulfiller->fulfill();
-  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+    commitFulfiller->fulfill();
+    KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
-  test.pollAndExpectCalls({});
-  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+    test.pollAndExpectCalls({});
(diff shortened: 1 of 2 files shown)


src/workerd/io/actor-sqlite.c++
```diff
@@ -513,10 +523,21 @@ ActorCacheInterface::DeleteAllResults ActorSqlite::deleteAll(WriteOptions option
...
   }
 
-  if (localAlarmState == kj::none && !deleteAllCommitScheduled) {
-    // If we're not going to perform a write to restore alarm state, we'll want to make sure the
-    // commit callback is called for the deleteAll().
-    commitTasks.add(outputGate.lockWhile(kj::evalLater([this]() mutable -> kj::Promise<void> {
+  // We want to ensure a commit is scheduled during the deleteAll() call, but if we end up doing a
+  // write to restore alarm state afterward, we can skip the original scheduled commit, since the
+  // write will queue its own commit.
+  kj::Maybe<kj::Own<bool>> maybeCanSkipDeleteAllCommit;
```
jclee, line 529: I'm not sure if cancelling the scheduled commit is strictly necessary, or if there isn't a simpler way to do it that preserves the desired invariants in all cases (exceptions, cancellation, etc.).  But without the cancellation, I noticed via the tests that there was otherwise a redundant commit call.

src/workerd/io/actor-sqlite.c++
```diff
@@ -513,10 +523,21 @@ ActorCacheInterface::DeleteAllResults ActorSqlite::deleteAll(WriteOptions option
...
+
+  if (!deleteAllCommitScheduled) {
```
jclee, line 531: In case it isn't apparent: the main fix here is removal of the `localAlarmState == kj::none` check, to ensure `deleteAllCommitScheduled` is always true by the time we reach `db.clear()`, which in turn ensures that `isCommitScheduled()` is true, since it's used to check an invariant in a storage relay callback.

The `localAlarmState == kj::none` check was originally put in place to avoid having a redundant commit, since the write to restore the alarmState also adds a commit.  But of course there's no way for the storage relay callback to know via the current `isCommitScheduled()` implementation that we're planning to add the commit after calling `db.clear()` (and maybe it isn't even correct to rely solely on committing after `db.clear()`, if it's important that the output lock is synchronously started before calling `db.clear()`?).

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -631,6 +631,62 @@ KJ_TEST("canceling deferred alarm deletion is idempotent") {
...
+
+  auto armResult = test.actor.armAlarmHandler(oneMs, false);
+  KJ_ASSERT(armResult.is<ActorSqlite::RunAlarmHandler>());
+
+  // Break gate
+  test.put("foo", "bar");
+  test.pollAndExpectCalls({"commit"})[0]->reject(KJ_EXCEPTION(FAILED, "a_rejected_commit"));
+  KJ_EXPECT_THROW_MESSAGE("a_rejected_commit", promise.wait(test.ws));
+  // Ensure taskFailed handler runs and notices brokenness:
+  test.ws.poll();
```
justin-mp, line 682: Up to and including this line, this test is the same as the test above it.  Why does it mean the handler succeeds in this case, but fails in the previous one?

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -930,21 +930,48 @@ KJ_TEST("calling deleteAll() preserves alarm state if alarm is set") {
...
+    KJ_ASSERT(!test.actor.isCommitScheduled());
+    ActorCache::DeleteAllResults results = test.actor.deleteAll({});
+    KJ_ASSERT(test.actor.isCommitScheduled());
+    KJ_ASSERT(results.backpressure == kj::none);
+    KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
 
-  auto commitFulfiller = kj::mv(test.pollAndExpectCalls({"commit"})[0]);
-  KJ_ASSERT(results.count.wait(test.ws) == 0);
-  KJ_ASSERT(expectSync(test.getAlarm()) == oneMs);
+    auto commitFulfiller = kj::mv(test.pollAndExpectCalls({"commit"})[0]);
```
justin-mp, line 996: For my own education, we only expect to see `scheduleRun(1ms)` when setAlarm is called, and not on subsequent resets?  Presumably, that's because the scheduled run gets scheduled outside of this deletion?

src/workerd/io/actor-sqlite.c++
kentonv: Hmm, I think you're overestimating the harm of a "redundant commit". `StorageRelayClient::commit()` already detects if there's nothing to commit (`if (!needCommit` at the beginning), so it's not that big a deal to schedule an extra one here.

And `deleteAll()` is not commonly called so I don't think it's worth optimizing away some promises.

Let's instead simplify this code to just always schedule (and apply) the deleteAll commit.

src/workerd/io/actor-sqlite-test.c++
kentonv: During normal alarm processing, if the alarm handler fails (i.e. the application throws an exception), `cancelDeferredAlarmDeletion()` is invoked.

src/workerd/io/actor-sqlite.c++
jclee: - [ ] Sure, makes sense... I'll try to simplify it.

src/workerd/io/actor-sqlite-test.c++
jclee: Right...  The intent behind the two tests was to exercise both the case where the caller calls `cancelDeferredAlarmDeletion()` prior to dropping the deleter and the case where it doesn't.  Strictly speaking, only the behavior of the non-cancelled test changes in this commit; the cancelled test was already succeeding because it didn't attempt to reset the alarm after brokenness, but I don't think the interaction between gate brokenness and cancellation was otherwise being covered by other tests.

- [ ] I can try to clarify the comments.

src/workerd/io/actor-sqlite-test.c++
jclee: Pretty much, yeah.  The `scheduleRun()` hook is called by ActorSqlite as part of the committing logic -- implemented in `startPrecommitAlarmScheduling()` and `commitImpl()` -- to request a change to the alarm manager's scheduled alarm notification time.  If ActorSqlite knows the currently scheduled alarm doesn't need to be updated, it doesn't call `scheduleRun()`.

But thinking about this a bit more, I guess the point of confusion might be why we don't see a `scheduleRun(kj::none)` in the test corresponding to the `db.reset()` and a subsequent `scheduleRun(1ms)` corresponding to the alarm state reset? That's because we don't run the committing logic for the `db.reset()`, only for the state reset.  And because we're restoring the original state, the committing logic sees no change, so `scheduleRun()` is not called.

I haven't yet implemented Kenton's suggestion of simplifying the deleteAll() logic by performing a redundant commit, but I'm pretty sure this won't change the situation, since the promise that deleteAll() queues calls `commitCallback()` directly, rather than through the alarm commit logic.  And I'm pretty sure that's OK?  But might need to think about it a bit more to be sure.

src/workerd/io/actor-sqlite.c++
kentonv: I guess what you're getting at is that the reset commit task doesn't actually do the same set of work as a regular write commit -- it doesn't actually commit a transaction and whatnot. Yeah, I agree it would be a problem if we committed the reset without actually committing the transaction to restore the alarm time.

So perhaps the reset commit task needs to actually skip calling `commitCallback` in the case that it knows that a further commit is coming for the metadata. Perhaps it can just be like:

```
    commitTasks.add(outputGate.lockWhile(kj::evalLater([this]() mutable -> kj::Promise<void> {
      // Don't commit if shutdown() has been called.
      requireNotBroken();

      deleteAllCommitScheduled = false;
      if (currentTxn.is<ImplicitTxn>()) {
        // An implicit transaction is already scheduled, so we'll count on it to perform a commit when it's
        // done. This is particularly important for the case where deleteAll() was called while an alarm
        // is outstanding; we'll immediately begin an implicit transaction that restores the alarm. We don't
        // want to commit the deletion without that transaction.
        return kj::READY_NOW;
      } else {
        return commitCallback();
      }
    })));
    deleteAllCommitScheduled = true;
```

kentonv: Another, maybe cleaner approach would be to unify the two commit tasks into a single one which checks both `currentTxn.is<ImplicitTxn>()` and `deleteAllCommitScheduled` to decide what actual work needs to be performed before committing?

jclee: I was also curious... I assume it's OK/moot that we call `requireNotBroken()` prior to clearing `deleteAllCommitScheduled`, since once the output gate is broken, we expect all subsequent user operations to fail and continue failing, such that the return value of `isCommitScheduled()` no longer matters?

kentonv: Yeah nothing matters once broken.

jclee: I tried combining the `deleteAll()` and `onWrite()` commit tasks, but it was turning out to be a more complicated change than your suggestion of just checking `if (currentTxn.is<ImplicitTxn*>()) {...}` in `deleteAll()`, so I went with that.

One thing I wasn't quite sure of, though, is the case involving explicit transactions:

- user calls `deleteAll()` while alarm state is present.
- `deleteAll()` adds its `kj::evalLater()` commit task to the event queue.
- `deleteAll()` calls `setAlarm()` to restore the post-delete alarm state, which fires `onWrite()`, which sets up the implicit transaction and adds its `kj::evalLater()` commit task to the event queue to run immediately after.
- before the event queue processes either commit task, it runs an event where user code starts an explicit transaction.
- the explicit transaction constructor runs the implicit transaction's commit(), then updates `currentTxn` to point to itself.
- the user code maybe makes a few additional changes within the explicit transaction.

When the `deleteAll()` commit task runs, it will see that `currentTxn` is no longer an implicit transaction, so it's going to just run `commitCallback()`.  Is that OK?

I assume the `commitCallback()` would end up persisting the interim state of the explicit transaction.  But presumably that's OK, because if the system goes down at that point, we recover from that by rolling back any explicit transaction savepoints we see in the persisted DB when reloading?  (If not, I think a similar sequence of events is already possible just using implicit and explicit transactions together, without calling `deleteAll()`.)

And I guess in that case, we'd be running a `commitCallback()` that effectively updates alarm state without also running the alarm rescheduling commit code, but maybe that's OK, since the very next event queue item should be the implicit transaction's commit task, which would run it?

kentonv (approved):

PR #1222, opened 2023-09-22
Fix various dropped promises

smerritt:
I found these by grabbing the changes from https://github.com/capnproto/capnproto/pull/1810, which makes the compiler warn about dropped promises.

PR #3019, opened 2024-10-28
Fix draining: Connections should stop being accepted after SIGTERM.

### LINKED ISSUE #3000: workerd exits during SIGTERM graceful drain while requests are still being handled
Workerd has [SIGTERM handling](https://github.com/cloudflare/workerd/blob/efb5ff9a9132b2d40489ddb5a705f7c3fb189b0a/src/workerd/server/workerd.c%2B%2B#L1308-L1309) to gracefully drain listeners before exiting.

In my testing, after receiving SIGTERM, workerd continues to accept new connections and handle new HTTP requests, and it does not set `connection: close` on responses.  It will, however, eventually exit while requests received after SIGTERM are still being handled.

Take `config.capnp`:
```capnp

const helloWorldExample :Workerd.Config = (
  services = [ (name = "main", worker = .helloWorld) ],
  sockets = [ ( name = "http", address = "*:8080", http = (), service = "main" ) ]
);

  modules = [
  ],
  compatibilityDate = "2024-10-22",
);
```

And `worker.js`:
```js
export default {
  async fetch() {
      await new Promise(resolve => setTimeout(resolve, 10_000));
      return new Response("hi\n");
  }
};
```

Sequence of events.
1. `workerd serve config.capnp`.
1. Start request 1 (`curl http://localhost:8080`).
1. `pkill -TERM -f workerd`.
1. Start request 2.
1. Request 1 completes. Start request 3.
1. Request 2 completes. Workerd exits while request 3 is still being handled. Curl says: `curl: (56) Recv failure: Connection reset by peer`.

The lack of connection-level draining (refusing new connections and/or setting `connection: close` on HTTP responses) is a bit surprising, but not necessarily unreasonable. (Though it does seem to contradict the stated intention [in the description of the PR](https://github.com/cloudflare/workerd/pull/244) that added the functionality.)

But requests being dropped on the floor is very surprising. 

I don't know if the behavior is actually that requests received after the completion of the last request that was received before draining began do not keep workerd alive, but that's what this example seems to indicate.

I've produced this on version 2024-10-22 on macOS and Linux.

kentonv:
This worked originally but #1222 accidentally broke it in an attempt to fix a compiler warning about dropped promises. And, unfortunately, the test didn't catch it.

This only affects workerd, not the edge runtime.

Fixes #3000

### DIFF
--- a/src/workerd/server/server-test.c++
+++ b/src/workerd/server/server-test.c++
@@ -369,6 +369,12 @@ public:
     return TestStream(ws, KJ_REQUIRE_NONNULL(sockets.find(addr), addr)->connect().wait(ws));
   }
 
+  // Try to connect to the address and return whether or not this connection attempt hangs,
+  // i.e. a listener exists but connections are not being accepted.
+  bool connectHangs(kj::StringPtr addr) {
+    return !KJ_REQUIRE_NONNULL(sockets.find(addr), addr)->connect().poll(ws);
+  }
+
   // Expect an incoming connection on the given address and from a network with the given
   // allowed / denied peer list.
   TestStream receiveSubrequest(kj::StringPtr addr,
@@ -2535,6 +2541,9 @@ KJ_TEST("Server: drain incoming HTTP connections") {
 
   // But conn2 is still open.
   KJ_EXPECT(!conn2.isEof());
+
+  // New connections shouldn't be accepted at this point.
+  KJ_EXPECT(test.connectHangs("test-addr"));
 
   // Finish the request on conn2.
   conn2.send(" / HTTP/1.1\nHost: foo\n\n");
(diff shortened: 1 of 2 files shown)


src/workerd/server/server-test.c++
```diff
@@ -369,6 +369,12 @@ public:
     return TestStream(ws, KJ_REQUIRE_NONNULL(sockets.find(addr), addr)->connect().wait(ws));
   }
 
+  // Try to connect to the address and return whether or notn this connection attempt hangs,
```
isker, line 372:
```suggestion
  // Try to connect to the address and return whether or not this connection attempt hangs,
```

isker (commented):
Thanks for the quick fix!

kentonv:
@ketanhwr @hoodmane ping

jasnell (approved):

PR #3021, opened 2024-10-28
Implement original tail worker support for local dev

jasnell:
Enables tail workers for local development in workerd

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -9,6 +9,7 @@
 #include <workerd/api/actor-state.h>
 #include <workerd/api/analytics-engine.capnp.h>
 #include <workerd/api/pyodide/pyodide.h>
+#include <workerd/api/trace.h>
 #include <workerd/api/worker-rpc.h>
 #include <workerd/io/actor-cache.h>
 #include <workerd/io/actor-id.h>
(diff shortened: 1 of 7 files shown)


src/workerd/server/server.c++
```diff
@@ -1535,14 +1660,52 @@ public:
...
+      // We can test this directly but it's more difficult to test indirect
+      // loops (dependency of dependency, etc). Here we're just going to keep
+      // it simple and just check the direct dependency.
+      tailWorkers.add(service.startRequest({}));
+    }
+
+    // When the tracer is complete, deliver the traces to both the parent
+    // and the tail workers.
+    waitUntilTasks.add(tracer->onComplete().then(
+        [tailWorkers = kj::mv(tailWorkers), parentTracer = mapAddRef(maybeTracer)](
```
danlapid, line 1689: Missing kj::coCapture

jasnell: Not sure it's actually needed in this case but doesn't hurt anything.

danlapid: Well you're creating a task that will run a lot later and capturing variables that should still be alive after your first co_await so I think it's required

jasnell:
Odd / non-obvious crashes in this PR on windows :-/ ... not getting any useful traces out of the CI run and unable to reproduce locally so I'll have to do a bit of trial-and-error debugging via fixup commits.

src/workerd/server/server.c++
```diff
@@ -1445,6 +1446,122 @@ void Server::InspectorServiceIsolateRegistrar::registerIsolate(
...
+        t->setFetchResponseInfo(Trace::FetchResponseInfo(fetchStatus));
+      }
+      t->setCPUTime(0 * kj::MILLISECONDS);
+      t->setWallTime(0 * kj::MILLISECONDS);
+      t->setOutcome(outcome);
+    }
+  }
+
+  WorkerInterface& wrapWorkerInterface(WorkerInterface& worker) override {
+    if (tracer != kj::none) {
```
danlapid, line 1465: assert inner is none?

src/workerd/server/server.c++
```diff
@@ -1445,6 +1446,122 @@ void Server::InspectorServiceIsolateRegistrar::registerIsolate(
...
+namespace {
+class RequestObserverWithTracer final: public RequestObserver, public WorkerInterface {
+public:
+  RequestObserverWithTracer(kj::Maybe<kj::Own<WorkerTracer>> tracer): tracer(kj::mv(tracer)) {}
+  ~RequestObserverWithTracer() noexcept(false) {
+    KJ_IF_SOME(t, tracer) {
+      if (fetchStatus != 0) {
+        t->setFetchResponseInfo(Trace::FetchResponseInfo(fetchStatus));
+      }
+      t->setCPUTime(0 * kj::MILLISECONDS);
```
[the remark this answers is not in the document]
danlapid: Well why would you have to manually set the cpu and wall time of the tracer to zero?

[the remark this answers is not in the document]
danlapid: If we decide to not set those in workerd for now I think a comment saying why we decided that is legit, otherwise if it's a TODO that's fine as well.
But I think it'll be easier for future readers to understand if we have a clear comment here on why this is set to 0 instead of some real value.
On the other hand if you had just not set it at all then people would just assume it's not yet implemented which is fine as well. I just think setting to 0 specifically sends a message that this has been intentional with that should come an explanation.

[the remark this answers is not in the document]
jasnell: I'm setting them here right now to rule out the possibility of potentially uninitialized fields causing the crashes that I'm seeing in the windows CI. If I end up keeping these I will add a comment but in general none of the mechanisms for tracking cpu/wall time are implemented for the workerd version of this, at least for now, these should be zero.

danlapid (approved):
LGTM, was the se translator useful? Do we want to add it in a separate PR?

jasnell:
> ... was the se translator useful? ...

Not really. Enabling the flag necessary seemed to make the build hang forever. I managed to spot the issue while I was waiting for it to finish building lol

kentonv:
Are there any tests for this?

src/workerd/server/workerd.capnp
```diff
@@ -630,6 +630,12 @@ struct Worker {
 
   moduleFallback @13 :Text;
 
+  # Tail worker configuration
+  logging :union {
```
kentonv, line 634: This could have just been:

```
tails @14 :List(ServiceDesignator);
```

The union adds unnecessarily complexity.

Is anyone using this yet? Can we change it?

Please remember to tag me on changes to capnp APIs.

PR #3024, opened 2024-10-30
Fix minor Use-After-Move in sockets API

fhanau:
As discovered by clang-tidy `clang-analyzer-cplusplus.Move`.

### DIFF
--- a/src/workerd/api/sockets.c++
+++ b/src/workerd/api/sockets.c++
@@ -417,7 +417,7 @@ void Socket::handleProxyStatus(jsg::Lock& js, kj::Promise<kj::Maybe<kj::Exceptio
 
 void Socket::handleProxyError(jsg::Lock& js, kj::Exception e) {
   resolveFulfiller(js, kj::cp(e));
-  openedResolver.reject(js, kj::mv(e));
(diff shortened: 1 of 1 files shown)


src/workerd/api/sockets.c++
```diff
@@ -417,7 +417,7 @@ void Socket::handleProxyStatus(jsg::Lock& js, kj::Promise<kj::Maybe<kj::Exceptio
 
 void Socket::handleProxyError(jsg::Lock& js, kj::Exception e) {
   resolveFulfiller(js, kj::cp(e));
-  openedResolver.reject(js, kj::mv(e));
```
fhanau, line 420: This might be causing an internal test failure – are KJ exceptions not safe to copy/is moving and then accessing e again in L422 the right approach here?

jasnell: The change looks correct. If a test is failing it's likely because the change is exposing another possible bug or the test was written to rely on the previously broken behavior.

dom96 (approved):

src/workerd/api/sockets.c++
fhanau: That ended up being the case, thank you!

PR #3027, opened 2024-10-30
[nfc] fixing *bundle rules dataflow

mikea:
data should flow from top to bottom and not sideways


anonrig (approved):

src/pyodide/BUILD.bazel
```diff
@@ -229,27 +235,41 @@ data = wd_ts_bundle_capnp(
...
-)
-
 genrule(
     name = "pyodide.capnp.bin@rule",
     srcs = [
-        "pyodide.capnp",
+        ":pyodide.capnp",
         "//src/workerd/jsg:modules.capnp",
-    ] + data,
+    ] + [
```
mikea, line 248: fyi @hoodmane this could probably be cleaned up, but I opted to keep it very explicit since I'm not entirely sure what's going on. If this is a common pattern, then maybe we need to add helper transformation function to the bundle .bzl, but returning values from macros is not a common pattern.

PR #3028, opened 2024-10-30
Implement InvocationSpanContext for user tracing / streaming tail workers

jasnell:
This implements the `InvocationSpanContext` class. This type holds a tuple of trace ID, invocation ID, and span ID and will be used in support for the updated streaming tail workers implementation. The basic idea is that every top-level request receives a trace ID which is shared across every invocation the occurs within that request. Every individual invocation has a unique invocation ID, and every span within that invocation receives a monotonically increasing span ID.

While this PR does not currently introduce uses of this type, the idea is for all invocations to always have a root `InvocationSpanContext` and for every child user span created to branch off this root. The current user span then would always have a current `InvocationSpanContext` that would be propagated outward for all traceable sub requests (generally any subrequest that is not an public internet fetch). A follow up PR will start to introduce the uses and propagation of the span context. This PR only introduces the class impl an the basic test for it. (I'm intentionally breaking this work into smaller incremental PRs to make it easier to review individual chunks).

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -155,6 +155,73 @@ kj::String KJ_STRINGIFY(const TraceId& id) {
   return id;
 }
 
+InvocationSpanContext::InvocationSpanContext(kj::Badge<InvocationSpanContext>,
+    kj::Maybe<kj::Rc<SpanIdCounter>> counter,
+    TraceId traceId,
+    TraceId invocationId,
+    kj::uint spanId,
+    kj::Maybe<kj::Rc<InvocationSpanContext>> parentSpanContext)
+    : counter(kj::mv(counter)),
+      traceId(kj::mv(traceId)),
+      invocationId(kj::mv(invocationId)),
+      spanId(spanId),
+      parentSpanContext(kj::mv(parentSpanContext)) {}
+
+kj::Rc<InvocationSpanContext> InvocationSpanContext::newChild() {
+  auto& c = KJ_ASSERT_NONNULL(counter, "unable to create child spans on this context");
+  return kj::rc<InvocationSpanContext>(kj::Badge<InvocationSpanContext>(), c.addRef(), traceId,
+      invocationId, c->next(), addRefToThis());
+}
+
+kj::Rc<InvocationSpanContext> InvocationSpanContext::newForInvocation(
+    kj::Maybe<kj::Rc<InvocationSpanContext>&> triggerContext,
+    kj::Maybe<kj::EntropySource&> entropySource) {
+  kj::Maybe<kj::Rc<InvocationSpanContext>> parent;
+  auto traceId = triggerContext
+                     .map([&](kj::Rc<InvocationSpanContext>& ctx) {
+    parent = ctx.addRef();
+    return ctx->traceId;
+  }).orDefault([&] { return TraceId::fromEntropy(entropySource); });
+  return kj::rc<InvocationSpanContext>(kj::Badge<InvocationSpanContext>(), kj::rc<SpanIdCounter>(),
+      kj::mv(traceId), TraceId::fromEntropy(entropySource), 0, kj::mv(parent));
+}
+
+TraceId TraceId::fromCapnp(rpc::InvocationSpanContext::TraceId::Reader reader) {
+  return TraceId(reader.getLow(), reader.getHigh());
+}
+
+void TraceId::toCapnp(rpc::InvocationSpanContext::TraceId::Builder writer) const {
+  writer.setLow(low);
+  writer.setHigh(high);
+}
+
+kj::Maybe<kj::Rc<InvocationSpanContext>> InvocationSpanContext::fromCapnp(
+    rpc::InvocationSpanContext::Reader reader) {
+  if (!reader.hasTraceId() || !reader.hasInvocationId()) {
+    // If the reader does not have a traceId or invocationId field then it is
+    // invalid and we will just ignore it.
+    return kj::none;
+  }
+
+  auto sc = kj::rc<InvocationSpanContext>(kj::Badge<InvocationSpanContext>(), kj::none,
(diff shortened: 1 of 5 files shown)


anonrig (approved):

fhanau:
Don't merge just yet – will review today

jasnell:
> Don't merge just yet – will review today

Yep, figured as much. Want to make sure this is going to be good with the work you're doing

src/workerd/io/trace.c++
```diff
@@ -155,6 +155,73 @@ kj::String KJ_STRINGIFY(const TraceId& id) {
...
+
+void TraceId::toCapnp(rpc::InvocationSpanContext::TraceId::Builder writer) const {
+  writer.setLow(low);
+  writer.setHigh(high);
+}
+
+kj::Maybe<kj::Rc<InvocationSpanContext>> InvocationSpanContext::fromCapnp(
+    rpc::InvocationSpanContext::Reader reader) {
+  if (!reader.hasTraceId() || !reader.hasInvocationId()) {
+    // If the reaer does not have a traceId or invocationId field then it is
```
anonrig, line 201:
```suggestion
    // If the reader does not have a traceId or invocationId field then it is
```

src/workerd/io/trace.c++
```diff
@@ -155,6 +155,73 @@ kj::String KJ_STRINGIFY(const TraceId& id) {
...
+    // invalid and we will just ignore it.
+    return kj::none;
+  }
+
+  auto sc = kj::rc<InvocationSpanContext>(kj::Badge<InvocationSpanContext>(), kj::none,
+      TraceId::fromCapnp(reader.getTraceId()), TraceId::fromCapnp(reader.getInvocationId()),
+      reader.getSpanId());
+  // If the traceId or invocationId are invalid, then we'll ignore them.
+  if (!sc->getTraceId() || !sc->getInvocationId()) return kj::none;
```
anonrig, line 210: At this point, maybe we should just have a helper method to check validness.
```suggestion
  // If the traceId or invocationId are invalid, then we'll ignore them.
  if (!sc->hasTraceId() || !sc->hasInvocationId()) return kj::none;
```

jasnell: Are you suggesting renaming these because there is no `hasTraceId()` or `hasInvocationId()`. These both return `const TraceId&` which implements an `operator bool()`, which means these work as is.

src/workerd/io/worker-interface.capnp
```diff
@@ -15,6 +15,16 @@ using import "/capnp/compat/byte-stream.capnp".ByteStream;
 using import "/workerd/io/outcome.capnp".EventOutcome;
 using import "/workerd/io/script-version.capnp".ScriptVersion;
 
+struct InvocationSpanContext {
+  struct TraceId {
+    high @0 :UInt64;
```
fhanau, line 20: This reminds me: high and low are an eyesore, we should really implement the TODO to use `__int128_t` some day.

src/workerd/io/trace.h
```diff
@@ -109,13 +109,114 @@ class TraceId final {
...
+  // whose ID is zero. Every child span context created within that context will
+  // have a span id that is one greater than the previously created one.
+  class SpanIdCounter final: public kj::Refcounted {
+  public:
+    SpanIdCounter() = default;
+    KJ_DISALLOW_COPY_AND_MOVE(SpanIdCounter);
+
+    inline kj::uint next() {
+      static constexpr kj::uint kMax = kj::maxValue;
+      KJ_ASSERT(id < kMax, "max number of spans exceeded");
```
fhanau, line 141: Not objecting to this assert, but if 2^32 spans are reached we have bigger problems 😄

src/workerd/io/trace.c++
```diff
@@ -155,6 +155,73 @@ kj::String KJ_STRINGIFY(const TraceId& id) {
...
+  return kj::mv(sc);
+}
+
+void InvocationSpanContext::toCapnp(rpc::InvocationSpanContext::Builder writer) const {
+  traceId.toCapnp(writer.initTraceId());
+  invocationId.toCapnp(writer.initInvocationId());
+  writer.setSpanId(spanId);
+  kj::mv(getParent());
```
fhanau, line 218: Where does the parent context move, or is this just used to invalidate it?

fhanau (approved):

src/workerd/io/worker-interface.capnp
jasnell: Agree. For the time being tho, let's leave that as a todo.

src/workerd/io/trace.h
```diff
@@ -109,13 +109,114 @@ class TraceId final {
...
+// invocation. Every invocation of every worker should have an InvocationSpanContext.
+// That may or may not have a trigger InvocationSpanContext.
+class InvocationSpanContext final: public kj::Refcounted,
+                                   public kj::EnableAddRefToThis<InvocationSpanContext> {
+public:
+  // Spans within a InvocationSpanContext are identified by a span id that is a
+  // monotically increasing number. Every InvocationSpanContext has a root span
+  // whose ID is zero. Every child span context created within that context will
+  // have a span id that is one greater than the previously created one.
+  class SpanIdCounter final: public kj::Refcounted {
```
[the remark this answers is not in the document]
jasnell: Yes, that's exactly it. I didn't put much effort into designing this bit and just Did Something That Works. If you have an alternative suggested approach here I'm happy to change this part.

src/workerd/io/trace.h
jasnell: Oh, I meant for this to be a `KJ_DASSERT` but this is just me being way over paranoid lol.

src/workerd/io/trace.c++
jasnell: Just invalidates.

PR #3029, opened 2024-10-30
Fix handling of termination exceptions.

jp4a50:
Exceptions caught in `v8::TryCatch` scopes need to be rethrown in order for the next try/catch handler down the stack to be notified of them. In some code paths, we rethrow the native `jsg::JsExceptionThrown` exception but don't rethrow the v8 exception using `v8::TryCatch::ReThrow()`.

This commit fixes one such code path which resulted in `IoContext::runImpl`'s `jsg::JsExceptionThrown`'s catch block finding that the corresponding `v8::TryCatch` object hadn't been notified of the termination (and didn't contain an exception).

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -2336,7 +2336,7 @@ public:
   //
   // Some kinds of exceptions explicitly will not be caught:
   // - Exceptions where JavaScript execution cannot continue, such as the "uncatchable exception"
-  //   produced by IsolateBase::terminateExecution().
+  //   produced by IsolateBase::TerminateExecution().
   // - C++ exceptions other than `kj::Exception`, e.g. `std::bad_alloc`. These exceptions are
   //   assumed to be serious enough that they cannot be caught as if they were JavaScript errors,
   //   and instead unwind must continue until C++ catches them.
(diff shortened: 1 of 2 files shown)


jp4a50:
Note that this change ensures that we now hit [this conditional branch](https://github.com/cloudflare/workerd/blob/main/src/workerd/io/io-context.c%2B%2B#L1082).

I wanted to generally prompt a discussion about how we want to handle various types of V8 exception, though. At the moment, it seems like our combination of `v8::TryCatch` and `jsg::JsExceptionThrown` is quite easy to misuse, leading to bugs like this.

Further context: we first started hitting this when we added [this catch clause](https://github.com/cloudflare/workerd/blob/main/src/workerd/io/io-context.h#L1204) which allowed termination exceptions to bubble up to the try/catch in `runImpl`. I managed to find an internal test which repro'd the behaviour and debugged/verified the fix using it.

jasnell (approved):

src/workerd/jsg/jsg.h
```diff
@@ -2354,13 +2354,14 @@ class Lock {
       } catch (JsExceptionThrown&) {
         // If tryCatch.HasCaught() is false, it typically means that JsExceptionThrown
         // was thrown without an exception actually being scheduled on the isolate.
-        // This may happen in particular when the JsExceptionThrows was the result of
-        // TerminateException() but V8 has since cleared the terminate flag because all
+        // This may happen in particular when the JsExceptionThrown was the result of
+        // TerminateExecution() but V8 has since cleared the terminate flag because all
```
jp4a50, line 2358: @jasnell what exactly is this comment referring to? Is it referencing [this V8 logic here](https://github.com/v8/v8/blob/main/src/api/api.cc#L2956)? I couldn't easily figure out under exactly what circumstances `CallDepthIsZero()` would be true since comments seemed to suggest that it pertains to "V8 API" call depth rather than JS call depth.

PR #3030, opened 2024-10-30
Add WPT urlpattern tests

npaun:
These tests were a bit clunky to add because the actual test code is in the `resources/` directory, and the typical `*.any.js` files are empty and contain only some comments.

New test cases:
- urlpattern-compare-tests.js
- urlpattern-hasregexpgroups-tests.js
- urlpatterntests.js

New harness features:
- assert_object_equals
- assert_implements


src/workerd/api/wpt/urlpattern-test.js
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import { run } from 'harness';
+
+export const urlpatternCompare = run('urlpattern-compare-tests.js', {
+  expectedFailures: [
```
jasnell, line 8:
```suggestion
  expectedFailures: [
    // Each of these *ought* to pass. They are included here because we
    // know they currently do not. Each needs to be investigated.
```

jasnell (approved):

src/workerd/api/wpt/urlpattern-test.js
jasnell: Yep

anonrig (approved):

PR #3034, opened 2024-10-30
Move ownership of metrics and limitEnforcer to the api type. 

danlapid:
Enable const variants of some const-safe jsg functions.
Move ownership of metrics and limitEnforcer to the api type.

This PR has no real function and should not change anything of itself.
Instead it's just another part of the work to enable python isolate pools.
This PR does some cleanup that will later enable us to create an isolate without the data that only becomes available once we have a request.

There is also a corresponding internal PR 9105


src/workerd/server/workerd-api.c++
```diff
@@ -157,17 +159,24 @@ struct WorkerdApi::Impl final {
...
+        observer(kj::atomicAddRef(*observerParam)),
+        jsgIsolate(v8System,
+            Configuration(*this),
+            kj::mv(observerParam),
+            limitEnforcer->getCreateParams()),
         memoryCacheProvider(memoryCacheProvider),
-        pythonConfig(pythonConfig) {}
+        pythonConfig(pythonConfig) {
+    jsgIsolate.runInLockScope(
+        [&](JsgWorkerdIsolate::Lock& lock) { limitEnforcer->customizeIsolate(lock.v8Isolate); });
```
dom96, line 178: In the previous version of the code this is run under `jsg::runInV8Stack`, I assume doing this here under `runInLockScope` is equivalent.

dom96 (approved):
Nice! Ownership changes look good. One tiny change seems to be to when we call `customizeIsolate`, but that likely isn't problematic. So LGTM!

src/workerd/server/workerd-api.c++
danlapid: Yeah, it's a utility function that translates to the same thing :)

danlapid:
> Nice! Ownership changes look good. One tiny change seems to be to when we call `customizeIsolate`, but that likely isn't problematic. So LGTM!

Yep you're right, that change is indeed intended, we're going to be using this isolate before we have an `Isolate` class wrapping it, we need it to have a limit enforcer tied to it. I would argue the previous location was just incorrect but since the constructors were always called one after the other in quick succession it never mattered.

PR #3037, opened 2024-10-31
SQC-333: Support Hyperdrive with SSL in dev mode

ReppCodes:
We have historically used sslmode=disable in when running `wrangler dev`, with the intention of not forcing folks to set up SSL on a local test DB.  Unfortunately this causes friction because many hosted databases will have SSL set up and required, and Hyperdrive in production actually requires this. To improve flexibilty, switch to `prefer` mode which will use SSL if available for a close-to-production experience, but will allow unencrypted connections if needed.


ReppCodes:

thomasgauvin (changes requested):
This means folks won't be able to connect without ssl

thomasgauvin (approved):

a-robinson (approved):

jasnell (approved):

anonrig (approved):

PR #3038, opened 2024-11-01
Add LazyWorkerInterface

danlapid:
This is a new WorkerInterface class that will be used in the internal PR, see comments to understand what it does.

### DIFF
--- a/src/workerd/io/worker-interface.h
+++ b/src/workerd/io/worker-interface.h
@@ -151,6 +151,105 @@ private:
 // for the promise, then invoke the destination object.
 kj::Own<WorkerInterface> newPromisedWorkerInterface(kj::Promise<kj::Own<WorkerInterface>> promise);
 
+template <typename Func>
+class LazyWorkerInterface final: public WorkerInterface {
+public:
+  LazyWorkerInterface(Func func): func(kj::mv(func)) {}
+
+  void ensureResolve() {
+    if (promise == kj::none) {
+      promise = KJ_ASSERT_NONNULL(func)()
+                    .then([this](kj::Own<WorkerInterface> result) { worker = kj::mv(result); })
+                    .eagerlyEvaluate(nullptr)
+                    .fork();
+      func = kj::none;
+    }
+  }
+
+  kj::Promise<void> request(kj::HttpMethod method,
+      kj::StringPtr url,
+      const kj::HttpHeaders& headers,
+      kj::AsyncInputStream& requestBody,
+      Response& response) override {
+    ensureResolve();
+    KJ_IF_SOME(w, worker) {
+      co_await w->request(method, url, headers, requestBody, response);
+    } else {
+      co_await KJ_ASSERT_NONNULL(promise);
+      co_await KJ_ASSERT_NONNULL(worker)->request(method, url, headers, requestBody, response);
+    }
+  }
+
+  kj::Promise<void> connect(kj::StringPtr host,
+      const kj::HttpHeaders& headers,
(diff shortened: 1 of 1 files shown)


src/workerd/io/worker-interface.h
```diff
@@ -151,6 +151,102 @@ class WorkerInterface: public kj::HttpService {
 // for the promise, then invoke the destination object.
 kj::Own<WorkerInterface> newPromisedWorkerInterface(kj::Promise<kj::Own<WorkerInterface>> promise);
 
+template <typename Func>
+class FuturePromisedWorkerInterface final: public WorkerInterface {
+public:
+  FuturePromisedWorkerInterface(Func func): func(kj::mv(func)) {}
+
+  void resolveFunc() {
+    promise = func()
```
kentonv, line 160: You probably only want to overwrite the promise if it isn't already set?

src/workerd/io/worker-interface.h
```diff
@@ -151,6 +151,102 @@ class WorkerInterface: public kj::HttpService {
 // for the promise, then invoke the destination object.
 kj::Own<WorkerInterface> newPromisedWorkerInterface(kj::Promise<kj::Own<WorkerInterface>> promise);
 
+template <typename Func>
+class FuturePromisedWorkerInterface final: public WorkerInterface {
+public:
+  FuturePromisedWorkerInterface(Func func): func(kj::mv(func)) {}
+
+  void resolveFunc() {
```
kentonv, line 159: Maybe `ensureResolve()`?

Should this be private?

src/workerd/io/worker-interface.h
```diff
@@ -151,6 +151,102 @@ class WorkerInterface: public kj::HttpService {
 // for the promise, then invoke the destination object.
 kj::Own<WorkerInterface> newPromisedWorkerInterface(kj::Promise<kj::Own<WorkerInterface>> promise);
 
+template <typename Func>
+class FuturePromisedWorkerInterface final: public WorkerInterface {
```
kentonv, line 155: Name suggestion: `LazyWorkerInterface`

danlapid:
@npaun @AdityaAtulTewari @harrishancock ping

src/workerd/io/worker-interface.h
```diff
@@ -151,6 +151,105 @@ class WorkerInterface: public kj::HttpService {
 // for the promise, then invoke the destination object.
 kj::Own<WorkerInterface> newPromisedWorkerInterface(kj::Promise<kj::Own<WorkerInterface>> promise);
 
+template <typename Func>
+class LazyWorkerInterface final: public WorkerInterface {
+public:
+  LazyWorkerInterface(Func func): func(kj::mv(func)) {}
```
jasnell, line 157: Non-blocking nit: might consider marking this explicitly `KJ_DISALLOW_COPY_AND_MOVE` but not critical with having the KJ_ASSERT_NONNULLs

jasnell (approved):

src/workerd/io/worker-interface.h
danlapid: Yea copies would be bad but they are implicitly disabled by virtue of having an Own member.
I don't see why moves should be a problem.

jasnell: Largely just an explicit hygiene thing for me... kind of a safeguard against someone changing any of the methods later and forgetting a `KJ_ASSERT_NONNULL`. But I'm being a bit extra paranoid :-)

PR #3040, opened 2024-11-01
gen-compile-cache binary

mikea:
Forked off #3004 and rewrote the driver in Rust


src/rust/gen-compile-cache/main.rs
```diff
@@ -0,0 +1,37 @@
...
+
+fn main() {
+    let args = Args::parse();
+
+    let file_list = fs::read_to_string(&args.file_list).expect("can't read file list");
+    for line in file_list.split("\n").filter(|l| l.len() > 0) {
+        let words: Vec<_> = line.split(" ").collect();
+        assert_eq!(2, words.len(), "incorrect input: `{line}`");
+
+        let input_path = words[0];
```
mikea, line 20: I wonder if there's a shorter way to collect into fixed sized arrays?

mikea:
gosh, windows delivers as usual:


```
the given toolchain supports creation of bazel-out/x64_windows-opt/bin/src/rust/gen-compile-cache/gen-compile-cache.exe.exe instead of bazel-out/x64_windows-opt/bin/src/rust/gen-compile-cache/gen-compile-cache.exe
```

anonrig (approved):

src/workerd/jsg/setup.h
```diff
@@ -617,10 +617,10 @@ class Isolate: public IsolateBase {
   };
 
   // The func must be a callback with the signature: void(jsg::Lock&)
```
danlapid, line 619: Don't forget to update comment now that this can return values.
Also should probably add a comment about being careful from returning v8 locals trying to escape the scope

danlapid: Is it even used in this PR?

mikea: yes, I use it to return bytes.

PR #3043, opened 2024-11-01
use ci.bazelrc to drive platform-dependent configuration

mikea:
I need to run bazel in lint job, but current setup makes it difficult to reuse settings


npaun (approved):

fhanau (approved):
Really glad we're doing this, will make maintenance easier

PR #3054, opened 2024-11-04
Extract code formatting config to its own file

npaun:
So we can use the same formatting tool in both workerd and our internal repo.


danlapid (approved):
❤️

PR #3055, opened 2024-11-04
[rust] publishing cxx bridge includes with workerd/ prefix

mikea:
This will allow us to have same crate names (cxx-integration) and not to confuse compiler regarding which one is used.

### DIFF
--- a/build/wd_rust_binary.bzl
+++ b/build/wd_rust_binary.bzl
@@ -32,7 +32,7 @@ def wd_rust_binary(
             name = name + "@cxx",
             src = cxx_bridge_src,
             hdrs = hdrs,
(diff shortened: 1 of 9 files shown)


fhanau (approved):
Change itself LGTM.
I guess if we start using cxx-integration from a source file that is shared between workerd and downstream, we'd need to pick the right cxx integration using #ifdef?

mikea:
> Change itself LGTM. I guess if we start using cxx-integration from a source file that is shared between workerd and downstream, we'd need to pick the right cxx integration using #ifdef?

I intend for ew cxx-integration header file to just include workerd + using statement so they will be the same.

PR #3057, opened 2024-11-04
[nfc] clippy lint job

mikea:


dom96 (approved):
Nice! Would it make sense to add a `lint` command to the justfile as well? or maybe add it to `format`?

fhanau:
With this we have two lint jobs: `Lint / lint ` and `Tests / lint / bazel`. To avoid confusion, can we indicate that the new lint job is for clippy or for Rust? (Ideally we'd have one unified lint job but that does not need to be handled here)

fhanau (approved):
LGTM otherwise

mikea:
> to avoid confusion, can we indicate that the new lint job is for clippy or for Rust?

clang-tidy will also come one day

PR #3060, opened 2024-11-05
[Vectorize] Fix field name for Vectorize describe operation

garvit-gupta:
Fix the typo in the field `vectorsCount`, and replace it with `vectorCount` in the Vectorize V2 Describe response type.


alekrumkamp:
Tracked in VS-338

anonrig (approved):

netgusto (approved):

PR #3067, opened 2024-11-06
Export setupContext publically to be used by additional contexts.

danlapid:
Currently Worker::Isolate::Impl does the setup of the context. Namely setting up the console and WebAssembly.Module@@HasInstance. This functionality is useful in general to any context we might create even before we have a Worker::Isolate.
This commit moves this functionality to a publicly facing api.

Another part split from #2936

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -467,9 +467,6 @@ private:
   kj::MutexGuarded<State> state;
 };
 
-// Defined later in this file.
-void setWebAssemblyModuleHasInstance(jsg::Lock& lock, v8::Local<v8::Context> context);
-
 static thread_local const Worker::Api* currentApi = nullptr;
 
 const Worker::Api& Worker::Api::current() {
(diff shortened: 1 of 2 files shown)


src/workerd/io/worker.c++
```diff
@@ -599,40 +596,12 @@ struct Worker::Isolate::Impl {
     KJ_DISALLOW_COPY_AND_MOVE(Lock);
 
     void setupContext(v8::Local<v8::Context> context) {
-      // Set WebAssembly.Module @@HasInstance
-      setWebAssemblyModuleHasInstance(*lock, context);
-
       // The V8Inspector implements the `console` object.
       KJ_IF_SOME(i, impl.inspector) {
```
danlapid, line 600: Only real change is that Set WebAssembly.Module @@HasInstance is now only set after alerting the inspector of a new context.
This shouldn't have any actual effect in practice.

dom96 (approved):
Thank you for splitting up these changes!

jasnell (approved):

PR #3069, opened 2024-11-06
Add ByteStreamObserver to enable monitoring of byte stream queue size and memory usage.

jp4a50:


src/workerd/io/observer.h
```diff
@@ -30,6 +30,18 @@ class WebSocketObserver: public kj::Refcounted {
...
 
+// Observes a byte stream. This interface should only be used by byte streams which internally queue
+// each write and each write of size N is accompanied by a corresponding read of the same size.
+// Thus, the current number of items in the internal queue can be calculated as the number of times
+// onWrite() has been called minus the number of times onRead() has been called.
+class ByteStreamObserver {
+public:
+  // Called when a write of size `bytes` is enqueued on the stream.
+  virtual void onWrite(size_t bytes) {};
+  // Called when a write of size `bytes` is dequeued from the stream.
```
danlapid, line 41:
```suggestion
  // Called when a read of size `bytes` is dequeued from the stream.
```

src/workerd/api/streams/transform.c++
```diff
@@ -64,7 +64,7 @@ jsg::Ref<TransformStream> TransformStream::constructor(jsg::Lock& js,
           .pull = maybeAddFunctor<UnderlyingSource::PullAlgorithm>(
               JSG_VISITABLE_LAMBDA((controller = controller.addRef()), (controller),
                   (jsg::Lock & js, auto c) mutable { return controller->pull(js); })),
-          .cancel = maybeAddFunctor<UnderlyingSource::CancelAlgorithm>(JSG_VISITABLE_LAMBDA(
```
danlapid, line 67: odd that the linter changed this line

danlapid (approved):
LGTM but I haven't checked if there are more streams we might want to attach this observer to.

src/workerd/io/observer.h
```diff
@@ -45,6 +57,12 @@ class RequestObserver: public kj::Refcounted {
     return kj::none;
   };
 
+  // This is called when a writable byte stream is created whilst processing this request. It will
+  // be destroyed when the corresponding byte stream is destroyed.
+  virtual kj::Maybe<kj::Own<ByteStreamObserver>> tryCreateWritableByteStreamObserver() {
+    return kj::none;
+  }
```
jasnell, line 64: Might be a bit easier to just have this return a `kj::Own<ByteStreamObserver>` that wraps a non-op singleton instance so that we can skip the `KJ_IF_SOME` block.

src/workerd/io/observer.h
```diff
@@ -30,6 +30,18 @@ class WebSocketObserver: public kj::Refcounted {
   virtual void receivedMessage(size_t bytes) {};
 };
 
+// Observes a byte stream. This interface should only be used by byte streams which internally queue
+// each write and each write of size N is accompanied by a corresponding read of the same size.
```
jasnell, line 34: This second sentence is a bit unclear as to what it means.

src/workerd/api/streams/internal.c++
```diff
@@ -933,6 +933,9 @@ jsg::Promise<void> WritableStreamInternalController::write(
 
       auto prp = js.newPromiseAndResolver<void>();
       increaseCurrentWriteBufferSize(js, byteLength);
+      KJ_IF_SOME(o, observer) {
+        o->onWrite(byteLength);
+      }
```
jasnell, line 938: An RAII approach the the API could work here also. Something like...

```
auto observedWrite = o->observeWrite(byteLength);
// .. then later when the chunk is consumed...
kj::mv(observedWrite); 
```

jasnell:
Nit: Some detail of what this is meant to be used for in the PR description would be helpful for context.

src/workerd/api/streams/internal-test.c++
```diff
@@ -230,7 +230,8 @@ KJ_TEST("WritableStreamInternalController queue size assertion") {
     // allowed to be queued.
 
     jsg::Ref<ReadableStream> source = ReadableStream::constructor(env.js, kj::none, kj::none);
-    jsg::Ref<WritableStream> sink = jsg::alloc<WritableStream>(env.context, kj::heap<MySink>());
```
jasnell, line 233: Would be helpful to include a test here with a mocked implementation of the observer to verify that the new instrumentation points work as expected would be helpful.

src/workerd/io/observer.h
jp4a50: I think I might be better off renaming `write` to `item` and `onWrite` to `enqueued` and `dequeued` to better reflect that we are observing operations on the internal queueing supported by streams rather than, for example, writing to the stream sink. WDYT?

danlapid: I don't know, I feel like that's less descriptive.
maybe `onWriteEvent` `onWriteEnqueued` `onWriteEventEnqueued` ?

src/workerd/io/observer.h
jp4a50: Good idea. It'd make sense to do this for both `ByteStreamObserver` and `WebSocketObserver` so I'll do both in a follow up PR since changing `tryCreateWebSocketObserver` will be a breaking API change.

src/workerd/io/observer.h
jp4a50: The term `chunk` seems to be used in the code and in the [writable stream JS API docs](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream) so I'll go for `onChunkEnqueued` and `onChunkDequeued`.

src/workerd/io/observer.h
jp4a50: Have tried to rephrase.

src/workerd/io/observer.h
danlapid: SGTM

PR #3070, opened 2024-11-06
[nfc] add rustfmt convenience target

mikea:
it is way faster than the rest of the format, why should I suffer? :)


danlapid (approved):
LGTM, btw, the rustfmt target fails on macos-rosetta for some reason it's trying to compile rustfmt instead of just pulling in the binary 😕 
Currently I got around it by running format.py directly instead of `just format`

PR #3077, opened 2024-11-07
[nfc] always download outputs in release builds

mikea:
after ci.bazelrc move minimal downloads started to apply to release builds too, which is not what we want.
Sorry for inconvenience @fhanau

### DIFF
--- a/build/ci.bazelrc
+++ b/build/ci.bazelrc
@@ -20,8 +20,11 @@ build:ci --color=yes
 build:ci --terminal_columns=100
 
 build:ci --config=v8-codegen-opt
-build:ci --remote_download_minimal --disk_cache=~/bazel-disk-cache
 test:ci --test_output=errors
+build:ci --disk_cache=~/bazel-disk-cache
+
+# test CI jobs don't need any top-level artifacts and just verify things
+build:ci-test --remote_download_outputs=minimal
 
 # limit storage usage on ci
 # Exclude large benchmarking binaries created in debug and asan configurations to avoid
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -21,7 +21,7 @@ build --experimental_remote_cache_eviction_retries=3
 # due to https://github.com/bazelbuild/bazel/issues/20576.
 # Windows build is failing on some runs due to FileAccessException errors, Bazel has had several
 # similar bugs in the past. Disable BwoB for now.
-build:windows --remote_download_all
+build:windows --remote_download_outputs=all
```
fhanau, line 24: Non-blocking: This overlaps with #2981, where we should be able to get rid of this again

build/ci.bazelrc
```diff
@@ -20,8 +20,13 @@ build:ci --color=yes
 build:ci --terminal_columns=100
 
 build:ci --config=v8-codegen-opt
-build:ci --remote_download_minimal --disk_cache=~/bazel-disk-cache
 test:ci --test_output=errors
+build:ci --disk_cache=~/bazel-disk-cache
+
+# normal CI build doesn't need outputs downloaded
+build:ci --remote_download_outputs=minimal
```
fhanau, line 27: I think it's safer to not set any default setting (so that we implicitly default to toplevel as of Bazel 7) and only opt in to minimal for the test job. I'm not sure if the npm publish/npm-types CI jobs rely on using toplevel

fhanau (approved):

PR #3078, opened 2024-11-07
[CI] Cross-compile Intel Mac release binary

fhanau:
Follow-up to #2958, where we added and tested cross-compilation support. This allows us to use Xcode 16 to compile the Intel Mac binary (macos-15 is Apple Silicon-based, macos-13 only has Xcode 15 available) and take advantage of the faster M1-based runners.

- Don't set minimum macOS version for host configuration, host tools are only needed during the build and do not need to be distributed beyond that. This fixes a misleading warning.

==============

Should save some precious macOS runner time. In a follow-up, we will also compile the macOS arm64 binary on GitHub CI, and eventually the Linux arm64 binary. Then all release binaries will be compiled here and we'll be able to automate the NPM publish and NPM types jobs so that releases are fully automated.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -211,8 +211,9 @@ build:linux --config=unix
 build:macos --config=unix
 
 # Support macOS 13 as the minimum version. There should be at least a warning when backward
-# compatibility is broken as -Wunguarded-availability-new is enabled by default.
-build:macos --macos_minimum_os=13.5 --host_macos_minimum_os=13.5
+# compatibility is broken as -Wunguarded-availability-new is enabled by default. Only enable for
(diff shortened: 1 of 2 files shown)


fhanau:
- [ ] After merging, manually confirm that the generated release artifact is actually an x86 binary (this worked in local testing for #2958, and when testing cross-compilation in that PR CI tests were slightly slower which is likely being caused by the Rosetta overhead).

As with all release-only changes, there is a small risk that this breaks the build, in which case I will promptly fix or revert.

mikea (approved):

npaun (approved):

PR #3080, opened 2024-11-07
[CI] Set right architecture for x64 macOS release artifact

fhanau:
This is needed now that we cross-compile from an Apple Silicon image.

=============

The problem here is that with #3078 the x64 macOS release artifact continues to be a valid x64 binary, but is uploaded as `macOS-ARM64-binary`, which prevents it from being added to tagged releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,12 +61,17 @@ jobs:
           - os-name: linux
             os: ubuntu-20.04
             bazel-config: release_linux
+            target-arch: X64
           - os-name: macOS
+            # This configuration is used for cross-compiling – macos-15 is Apple Silicon-based but
+            # we use it to compile the x64 release.
             os: macos-15
             bazel-config: release_macos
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -61,12 +61,17 @@ jobs:
           - os-name: linux
             os: ubuntu-20.04
             bazel-config: release_linux
+            target-arch: X64
```
fhanau, line 64: This is redundant at present (all releases built on GitHub Actions are for x86), but I plan to build the Apple Silicon release here too some time after #3079 has landed. Open to alternative suggestions.

npaun (approved):

.github/workflows/release.yml
npaun: It's pretty good as-is imo

fhanau: 👍
Merged and seems to work correctly now, thanks for approving!

PR #3081, opened 2024-11-08
Disable top-level await in require with a compat flag

jasnell:
~~Currently, the top level scope of a Worker script and modules that are imported may use top level await. However, because we restrict the use of i/o in the top level scope, the only thing that can be awaited at the top level scope is a dynamic import, which we require to be resolved synchronously. It is a bit nonsensical to support synchronously resolved dynamic imports at the top level when we can just use static imports more effectively. Additionally, the ecosystem and runtimes are moving to a state where top level await in modules is being strongly discouraged. This compat flag. This compatibility flag will make it impossible to use top level await.~~

(made the change with lots of distractions so keeping as a draft until I can do a second pass to make sure nothing was missed)

*Update*: Updated to apply the compat flag only to `require()` calls, along with a few other relevant cleanups. Still keeping as a draft because this will need some extensive testing.

### DIFF
--- a/src/workerd/api/node/module.c++
+++ b/src/workerd/api/node/module.c++
@@ -85,6 +85,18 @@ jsg::JsValue ModuleUtil::createRequire(jsg::Lock& js, kj::String path) {
 
     auto module = info.module.getHandle(js);
 
+    if (module->GetStatus() == v8::Module::Status::kEvaluating ||
+        module->GetStatus() == v8::Module::Status::kInstantiating) {
+      KJ_IF_SOME(synth, info.maybeSynthetic) {
+        KJ_IF_SOME(cjs, synth.tryGet<jsg::ModuleRegistry::CommonJsModuleInfo>()) {
+          return cjs.moduleContext->getExports(js);
+        }
+        KJ_IF_SOME(cjs, synth.tryGet<jsg::ModuleRegistry::NodeJsModuleInfo>()) {
+          return cjs.moduleContext->getExports(js);
+        }
+      }
+    }
+
     auto features = FeatureFlags::get(js);
     jsg::InstantiateModuleOptions options = jsg::InstantiateModuleOptions::DEFAULT;
     if (features.getNoTopLevelAwaitInRequire()) {
--- a/src/workerd/api/node/module.c++
+++ b/src/workerd/api/node/module.c++
@@ -3,6 +3,7 @@
 //     https://opensource.org/licenses/Apache-2.0
 #include "module.h"
 
+#include <workerd/io/features.h>
 #include <workerd/jsg/url.h>
 
 namespace workerd::api::node {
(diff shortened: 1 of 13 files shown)


jasnell:
After an initial review of impact on this, I think this PR needs to be updated to relax the restriction a bit more. TLA should be restricted only when modules are `require(...)`, but entry point and import would work fine.

jasnell:
Updated to further refine and also eliminates the limitation around circular dependencies of commonjs modules. Not marking this "ready for review" just yet because there are a few more additional cleanups I want to do to eliminate duplicated code.

src/workerd/api/node/module.c++
```diff
@@ -84,19 +85,34 @@ jsg::JsValue ModuleUtil::createRequire(jsg::Lock& js, kj::String path) {
 
     auto module = info.module.getHandle(js);
 
-    jsg::instantiateModule(js, module);
-    auto handle = jsg::check(module->Evaluate(js.v8Context()));
-    KJ_ASSERT(handle->IsPromise());
-    auto prom = handle.As<v8::Promise>();
-    if (prom->State() == v8::Promise::PromiseState::kPending) {
-      js.runMicrotasks();
+    if (module->GetStatus() == v8::Module::Status::kEvaluating ||
```
anonrig, line 88: Can you add a comment to here?

src/workerd/io/compatibility-date.capnp
```diff
@@ -657,4 +657,12 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   #
   # This is a compat flag so that we can opt in our test workers into it before rolling out to
   # everyone.
+
+  noTopLevelAwaitInRequire @67 :Bool
+      $compatEnableFlag("disable_top_level_await_in_require")
+      $compatDisableFlag("enable_top_level_await_in_require")
+      $compatEnableDate("2024-12-02");
```
anonrig, line 664: Why are we specifically targeting this date?

src/workerd/jsg/modules.c++
```diff
@@ -322,32 +334,43 @@ NonModuleScript NonModuleScript::compile(kj::StringPtr code, jsg::Lock& js, kj::
...
-      "dependency on itself, and that a synchronous require() is being called while the module "
-      "is being loaded.");
+  // If the previous instantiation failed, throw the exception.
+  if (status == v8::Module::Status::kErrored) {
+    isolate->ThrowException(module->GetException());
+    throw jsg::JsExceptionThrown();
+  }
+
+  // Nothing to do if the module is already evaluated.
+  if (status == v8::Module::Status::kEvaluated || status == v8::Module::Status::kEvaluating) return;
```
anonrig, line 352: Why are we returning here? Shouldn't we assert that it is not evaluated or evaluating before calling this function. Previous kEvaluating check on module loaders makes it hard to trace the lifecycle and evaluation timeline of a module.

src/workerd/jsg/modules.c++
```diff
@@ -322,32 +334,43 @@ NonModuleScript NonModuleScript::compile(kj::StringPtr code, jsg::Lock& js, kj::
...
+  }
 
-  jsg::check(module->InstantiateModule(context, &resolveCallback));
   auto prom = jsg::check(module->Evaluate(context)).As<v8::Promise>();
+
+  if (module->IsGraphAsync() && prom->State() == v8::Promise::kPending) {
+    // If top level await has been disable, error.
+    JSG_REQUIRE(options != InstantiateModuleOptions::NO_TOP_LEVEL_AWAIT, Error,
+        "Top-level await in module is not permitted at this time.");
+  }
```
anonrig, line 364: This can be simplified to a single line of JSG_REQUIRE but I guess it would be less readable...

anonrig: Would it be more readable and maintainable if we add these assumptions/assertions as a comment before the function? It seems there are lots of know-how required for this function that could be beneficial for a developer, if provided in a more structural way.

src/workerd/jsg/modules.h
```diff
@@ -191,7 +193,14 @@ class NonModuleScript {
   v8::Global<v8::UnboundScript> unboundScript;
 };
 
-void instantiateModule(jsg::Lock& js, v8::Local<v8::Module>& module);
+enum class InstantiateModuleOptions {
+  DEFAULT,
+  NO_TOP_LEVEL_AWAIT,
```
anonrig, line 198: Can we add a comment before these definitions?

anonrig (approved):

jasnell:
@kentonv 
> There are actually several async APIs that work at startup time, including WebCrypto and WebAssembly. I think there are real use cases for being able to use these at startup so I'm not excited about taking it away.

Those will not be impacted. Note that this PR has been updated to ONLY apply to use of `require()` and `createRequire()`. There is zero impact on the worker top-level evaluation and anything imported using static or dynamic import.

src/workerd/io/compatibility-date.capnp
jasnell: Picked an arbitrary date out of the air that isn't too far away. This will likely change. There's no significance to this specific date.

src/workerd/jsg/modules.c++
jasnell: No, because this is potentially re-entrant. If the module is already evaluated or actively evaluating then we don't want to *start* evaluating it again but we also don't want to assert here.

PR #3085, opened 2024-11-09
[nfc] 110824 assorted build cleanup

fhanau:
**Roll Rust deps, rules_rust**
- The rules_rust update includes checksums for Rust 1.82, fixing a warning.

**[build] Fix protobuf warning, compile_flags.txt** 

**[build] Port libc++ defines from downstream to reduce include overhead**


jasnell (approved):
Bit of a rubber-stamp LGTM

anonrig:
ASAN errors might be related

fhanau:
> ASAN errors might be related

Thank you for approving. The crash disappears after re-running the ASan job, so it is unlikely to be caused by this (there are two libc++ flags being added but they are unrelated to ASan and are working well with downstream). There is a downstream V8 patch dealing with ASan issues for html-rewriter tests, perhaps the html-rewriter crash here has the same root cause.

PR #3089, opened 2024-11-11
[o11y] Add rudimentary user tracing support for connect()

fhanau:
- Use more descriptive name for IOContext function

### DIFF
--- a/src/workerd/api/sockets.c++
+++ b/src/workerd/api/sockets.c++
@@ -180,6 +180,7 @@ jsg::Ref<Socket> connectImplNoOutputLock(jsg::Lock& js,
 
   auto& ioContext = IoContext::current();
   JSG_REQUIRE(!ioContext.isFiddle(), TypeError, "Socket API not supported in web preview mode.");
+  auto userSpan = ioContext.makeUserTraceSpan("connect"_kjc);
 
   // Extract the domain/ip we are connecting to from the address.
   kj::String domain;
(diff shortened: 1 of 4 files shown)


src/workerd/api/sockets.c++
```diff
@@ -180,6 +180,7 @@ jsg::Ref<Socket> connectImplNoOutputLock(jsg::Lock& js,
 
   auto& ioContext = IoContext::current();
   JSG_REQUIRE(!ioContext.isFiddle(), TypeError, "Socket API not supported in web preview mode.");
+  ioContext.makeUserTraceSpan("connect"_kjc);
```
a-robinson, line 183: I imagine it's intentional that we're dropping the return value immediately and not measuring anything about the duration of the connect?

It looks a little funny, but so long as it's by design the code LGTM.

a-robinson (approved):

src/workerd/api/sockets.c++
fhanau: I think you're right – we want the span to exist for the entire duration of the function, and I guess if no variable is declared for it, it goes out of scope immediately instead of at the end of the function. In some other cases we need to assign the returned span so that we can extend its lifetime for async I/O or to create subspans, but I guess this is still needed here too.

fhanau: @mikea would this make `makeUserTraceSpan()` and similarly `SpanBuilder::newChild()` good candidates for adding the `[[nodiscard]]` to avoid this mistake in the future?

mikea: looks like `[[nodiscard]]` is a must for them.

mikea (approved):

jasnell (approved):

PR #3092, opened 2024-11-12
[build] Move io-channels.h back into io target

fhanau:
io-channels.h has a circular dependency with io due to its dependency on io-util.h (parsing the header using bazel is expected to fail accordingly). This means any source file depending on it would need to be in a bazel target depending on io to avoid missing header errors, even if the source file doesn't use io itself. Move it back into io in accordance with IWYU.

\- Fix dependencies for io:actor-id, util:duration-exceeded-logger
\- Fix duration-exceeded-logger.h appearing in two targets

### DIFF
--- a/src/workerd/io/BUILD.bazel
+++ b/src/workerd/io/BUILD.bazel
@@ -59,6 +59,7 @@ wd_cc_library(
         "compatibility-date.h",
         "features.h",
         "hibernation-manager.h",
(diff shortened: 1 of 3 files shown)


fhanau:
Builds seem to be passing on the eponymous downstream branch.

dom96 (approved):

PR #3093, opened 2024-11-12
Upgrade V8 13.0.245.16 -> 13.1.201.8

ObsidianMinor:
Includes the `WebAssembly.compile` behavior patch.


anonrig (approved):

anonrig (approved):

fhanau:
- [ ] In WORKSPACE, absl should be updated to the commit used by V8 13.1 (dc257ad54f38739767a6cb26eb57fd51c37bfe3c)

fhanau:
LGTM – I assumed that we'd have to set `--@v8//:v8_wasm_random_fuzzers=false` here to benefit from https://chromium-review.googlesource.com/c/v8/v8/+/5873224, it seems to be off-by-default though. There are some new defines like CPPGC_ENABLE_API_CHECKS too, but it looks like they don't require manual handling either.

fhanau (approved):

danlapid:
@ObsidianMinor can we get this passing and merged please?

ObsidianMinor:
I ran around for 2 days until I eventually got it compiled locally and ran it under WinDbg just to find it it's broken and was only fixed here https://github.com/v8/v8/commit/f8bb2d4576cf61c68b2ac5263cf82c8ee4e0a01e

danlapid:
> I ran around for 2 days until I eventually got it compiled locally and ran it under WinDbg just to find it it's broken and was only fixed here [v8/v8@f8bb2d4](https://github.com/v8/v8/commit/f8bb2d4576cf61c68b2ac5263cf82c8ee4e0a01e)

Not even backporting the fix to 13.1, that's just mean

.bazelrc
```diff
@@ -194,9 +197,9 @@ build:unix --workspace_status_command=./tools/unix/workspace-status.sh
 build:unix --cxxopt='-std=c++20' --host_cxxopt='-std=c++20'
 build:unix --@capnp-cpp//src/kj:libdl=True
 
-build:unix --action_env=BAZEL_COMPILER=clang
-build:unix --action_env=CC=clang
-build:unix --action_env=CXX=clang++
+build:unix --action_env=BAZEL_COMPILER=clang-18
```
danlapid, line 200: I assume this is unintended?
We don't want to choose clang version here but rather use the system default.
We don't require a specific clang version but rather a minimum version so setting this here will break contributors build environments

patches/v8/0019-codegen-Don-t-pass-a-nullptr-in-InitUnwindingRecord-.patch
danlapid, line 1: If this ends up being the fix tag Mike to let him know he should remove this patch when updating to 13.2 🙏

.bazelrc
ObsidianMinor: Oh yeah woops that's supposed to be just for me

patches/v8/0019-codegen-Don-t-pass-a-nullptr-in-InitUnwindingRecord-.patch
ObsidianMinor: I believe git will automatically remove the commit since the patch already exists upstream? It should happen during the rebase. But I'll make sure to mention it.

danlapid:
@ObsidianMinor gg wp

PR #3096, opened 2024-11-12
Implements JS Blob API in Python SDK

dom96:
The big change here is the addition of the Blob API. But there are also some other improvements/fixes:

* Each class instance now exposes a `js_object` property which returns its underlying JS object.
* The FormData API is now compatible with the multidict package's API
* The FormData API can now accept Blobs and returns the Python Blobs too

### Test Plan

```
$ bazel run @workerd//src/workerd/server/tests/python:sdk/sdk
```

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -1,14 +1,16 @@
 # This module defines a Workers API for Python. It is similar to the API provided by
 # JS Workers, but with changes and additions to be more idiomatic to the Python
 # programming language.
-from collections.abc import Generator, MutableMapping
+from collections.abc import Generator, Iterable, MutableMapping
+from contextlib import ExitStack, contextmanager
+from enum import StrEnum
 from http import HTTPMethod, HTTPStatus
 from typing import TypedDict, Unpack
 
 import js
 
 import pyodide.http
-from pyodide.ffi import JsException, to_js
+from pyodide.ffi import JsException, create_proxy, destroy_proxies, to_js
 from pyodide.http import pyfetch
 
 JSBody = (
@@ -25,6 +27,8 @@ class FetchKwargs(TypedDict, total=False):
     method: HTTPMethod = HTTPMethod.GET
 
 
+# TODO: Pyodide's FetchResponse.headers returns a dict[str, str] which means
+#       duplicates are lost, we should fix that so it returns a http.client.HTTPMessage
 class FetchResponse(pyodide.http.FetchResponse):
     # TODO: Consider upstreaming the `body` attribute
     @property
@@ -37,6 +41,10 @@ class FetchResponse(pyodide.http.FetchResponse):
             return FormData(b)
         else:
             return b
+
+    @property
+    def js_object(self) -> "js.Response":
+        return self.js_response
 
     """
     Instance methods defined below.
@@ -77,6 +85,15 @@ def _to_python_exception(exc: JsException) -> Exception:
         return exc
 
 
+@contextmanager
+def _manage_pyproxies():
+    proxies = js.Array.new()
+    try:
+        yield proxies
+    finally:
+        destroy_proxies(proxies)
+
+
 class Response(FetchResponse):
     def __init__(
         self,
@@ -96,7 +113,7 @@ class Response(FetchResponse):
         # Initialise via the FetchResponse super-class which gives us access to
         # methods that we would ordinarily have to redeclare.
         js_resp = js.Response.new(
-            body.js_form_data if isinstance(body, FormData) else body, **options
+            body.js_object if isinstance(body, FormData) else body, **options
         )
         super().__init__(js_resp.url, js_resp)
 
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/workers.py
```diff
@@ -25,6 +27,8 @@ class FetchKwargs(TypedDict, total=False):
     method: HTTPMethod = HTTPMethod.GET
 
 
+# TODO: Pyodide's FetchResponse.headers returns a dict[str, str] which means
+#       duplicates are lost, we should fix that so it returns a http.client.HTTPMessage
```
hoodmane, line 31: Do you still think we should do this? I thought the duplicate values are combined with a comma. How does `http.client.HTTPMessage` handle values with a comma in them?

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,202 @@ def json(
...
+    def _supports_buffer_protocol(self, o):
+        try:
+            # memoryview used only for testing type; 'with' releases the view instantly
+            with memoryview(o):
+                return True
+        except TypeError:
+            return False
+
+    @contextmanager
+    def _make_blob_entry(self, e):
```
hoodmane, line 339: These should be top level functions, not methods. Or at least make them static.

src/pyodide/internal/workers.py
dom96: It's not super important, but it would be nice. Splitting values by comma is pretty bleh as far as API design.

But I guess on the other hand that is what JS' Headers expects so maybe it's fine.

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,202 @@ def json(
...
-            self.js_form_data = js.FormData.new()
+            return item
 
-    def __getitem__(self, key: str) -> list[FormDataValue]:
-        return list(self.js_form_data.getAll(key))
+    def _process_value_for_py(self, item: FormDataValue) -> "str | Blob":
+        if hasattr(item, "constructor") and (item.constructor.name in ("Blob", "File")):
+            return Blob(item)
+        else:
+            return item
```
hoodmane, line 193: nits: These also could be top level nonstatic functions. Maybe call them `_js_value_to_py` and `_py_value_to_js`?

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,202 @@ def json(
...
+    """
+
+    def __init__(
+        self, form_data: "js.FormData | None | dict[str, FormDataValue]" = None
+    ):
         if form_data:
             if isinstance(form_data, dict):
-                self.js_form_data = js.FormData.new()
+                self._js_form_data = js.FormData.new()
                 for item in form_data.items():
```
hoodmane, line 174:
```suggestion
                for k, v in form_data.items():
```

src/pyodide/internal/workers.py
dom96: ahh, I knew I was forgetting something :)

Done.

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,204 @@ def json(
...
 
 class FormData(MutableMapping[str, FormDataValue]):
-    def __init__(self, form_data: "js.FormData | None | dict[str, FormDataValue]"):
+    """
+    This API follows that of https://pypi.org/project/multidict/.
+    """
+
+    def __init__(
+        self, form_data: "js.FormData | None | dict[str, FormDataValue]" = None
+    ):
```
hoodmane, line 170: I think it's nicer style to do something like:
```
if not form_data:
   self._js_form_data = js.FormData.new()
   return 
if hasattr(form_data, "constructor") and form_data.constructor.name == "FormData":
   self._js_form_data = form_data
   return
if isinstance(form_data, dict):
   self._js_form_data = _dict_to_from_data(form_data)
   return
raise TypeError("Expected form_data to be a dict or an instance of FormData")
```
and then factor out the logic to convert a dictionary to a from data into a separate function.

dom96: Agreed

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,202 @@ def json(
...
+            raise TypeError("Expected single item in arguments to FormData.__setitem__")
+        self._js_form_data.set(key, _py_value_to_js(value))
 
-    def append(self, key: str, value: FormDataValue):
-        self.js_form_data.append(key, value)
+    def append(self, key: str, value: FormDataValue, filename: str | None = None):
+        if filename:
+            self._js_form_data.append(key, _py_value_to_js(value), filename)
+        else:
+            self._js_form_data.append(key, _py_value_to_js(value))
```
hoodmane, line 207: These can be unified:
```suggestion
        self._js_form_data.append(key, _py_value_to_js(value), filename)
```

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,202 @@ def json(
...
+        yield from self._js_form_data.keys()
 
     def __iter__(self):
         yield from self.keys()
 
     def items(self) -> Generator[tuple[str, FormDataValue], None, None]:
-        for item in self.js_form_data.entries():
-            yield (item[0], item[1])
+        for item in self._js_form_data.entries():
+            yield (item[0], _js_value_to_py(item[1]))
```
hoodmane, line 227:
```suggestion
        for k, v in self._js_form_data.entries():
            yield (k, _js_value_to_py(v))
```

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,202 @@ def json(
...
 
     def __len__(self):
         return len(self.keys())
+
+    def get_all(self, key: str) -> list[FormDataValue]:
+        return list(map(lambda x: _js_value_to_py(x), self._js_form_data.getAll(key)))
+
+    @property
+    def native(self) -> "js.FormData":
+        return self._js_form_data
```
hoodmane, line 240: Is this a better name than `js_form_data`? I guess using `native` everywhere is most likely advantageous so that people don't need to remember the name. Though I'd prefer something with `js` in it. What about `js_object`?

dom96: Yeah, I was going to ask if anyone has better name suggestions. `js_object` sounds good.

dom96: And yeah, I introduced it so that it's consistent across all the objects.

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,208 @@ def json(
...
+class Blob:
+    def __init__(
+        self,
+        blob_parts: "Iterable[BlobValue] | BlobValue",
+        **options: Unpack[BlobKwargs],
+    ):
+        if "content_type" in options:
+            options["type"] = options["content_type"]
+        if "endings" in options:
+            options["endings"] = str(options["endings"])
```
hoodmane, line 305: Seems like it might be better to be explicit here:
```suggestion
        content_type: str | None = None,
        endings: BlobEnding | str | None = None
    ):
        if endings:
           endings = str(endings)
```
and then:
```py
    js.Blob.new(args, type=content_type, endings=endings)
```
Also, I am pretty sure that you don't need `to_js` for `args`.

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,208 @@ def json(
...
+
+        is_single_item = not hasattr(blob_parts, "__iter__")
```
hoodmane, line 307: The correct way to test if something is iterable is:
```
def _is_iterable(obj):
    try:
        iter(obj)
        return True
    except TypeError:
        return False
```
Checking `hasattr(obj, "__iter__")` can give false negatives. I think we should add this as another top level helper function and then this should be:
```suggestion
        is_single_item = not _is_iterable(blob_parts)
```

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,208 @@ def json(
...
+        if is_single_item:
+            if "type" not in options:
+                # Inherit the `type` if we have a single item.
+                self._js_blob = (
+                    blob_parts.js_object if isinstance(blob_parts, Blob) else blob_parts
+                )
+                return
```
hoodmane, line 314: Won't this mess up if someone passed `endings`? I'd suggest:
```py
# Inherit the `type` if we have a single item.
if not content_type and isinstance(blob, Blob):
   content_type = blob.content_type
if not content_type and hasattr(blob, "constructor") and blob.constructor.name == "Blob":
   content_type = blob.type
```
and then always create a new blob in the same place.

src/pyodide/internal/workers.py
dom96: > Also, I am pretty sure that you don't need to_js for args.

You do, I get this otherwise: `pyodide.ffi.JsException: TypeError: Failed to construct 'Blob': constructor parameter 1 is not of type 'Array'.`

src/pyodide/internal/workers.py
dom96: If I do this then I also have to handle `filename`, but yeah, I'll do that

hoodmane: Well if you want to allow it, that sounds like a reasonable helpful behavior.

src/pyodide/internal/workers.py
hoodmane: That's unfortunate. There's been discussion before whether we should make `Array.isArray()` return `true` on a Python list. Not sure whether it would fix this error message. I also don't remember why I didn't do it -- it's possible I just never got to it, or that there was some problem it caused.

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,211 @@ def json(
...
+            yield (k, _js_value_to_py(v))
 
     def __delitem__(self, key: str):
         self.delete(key)
 
     def __len__(self):
         return len(self.keys())
+
+    def get_all(self, key: str) -> list[FormDataValue]:
+        return list(map(lambda x: _js_value_to_py(x), self._js_form_data.getAll(key)))
```
hoodmane, line 242: More idiomatic to use a list comprehension here:
```suggestion
        return [_js_value_to_py(x) for x in self._js_form_data.getAll(key)]
```

src/pyodide/internal/workers.py
```diff
@@ -149,54 +157,211 @@ def json(
...
+                    blob_parts.js_object if isinstance(blob_parts, Blob) else blob_parts
+                )
+                return
+
+            # Otherwise create a new Blob below.
+            blob_parts = [blob_parts]
+
+        with ExitStack() as stack:
+            args = [stack.enter_context(_make_blob_entry(e)) for e in blob_parts]
+            self._js_blob = js.Blob.new(to_js(args), type=content_type, endings=endings)
```
hoodmane, line 324: We need to be careful here or else this will be leaky. Probably the easiest thing to do is:
```suggestion
            pyproxies = js.Array()
            self._js_blob = js.Blob.new(to_js(args, pyproxies), type=content_type, endings=endings)
            destroy_proxies(pyproxies)
```
you need also a `from pyodide.ffi import to_js`.

hoodmane: Yeah we definitely don't need this here sorry. Everything returned from `_make_blob_entry` is either a string or a JsProxy so no PyProxies are created by this to_js call.

src/pyodide/internal/workers.py
```diff
@@ -141,62 +158,248 @@ def json(
...
+    @property
+    def content_type(self) -> str:
+        return self._js_blob.type
+
+    @property
+    def name(self) -> str:
+        if self._js_blob.constructor.name != "File":
+            raise TypeError(
+                "Underlying Blob type is not a File and so has no file name."
+            )
```
hoodmane, line 377: It's not obvious to me that it's a good idea to unify `Blob` and `File`, but if we do, it's not obvious to me we should raise `TypeError` here. Another option would be to return `None` here. In fact, if you just did:
```py
return self._js_blob.name
```
then it would return `None` for `Blob`s.

hoodmane (approved):
Looking pretty good now, thanks @dom96. I have one last remark.

PR #3097, opened 2024-11-12
Have Blob use BufferSource for arrayBuffer() also

jasnell:
Avoid use of the `kj::Array<kj::byte>` type wrapping entirely for reads (`arrayBuffer()` as well as `bytes()`)


anonrig (approved):
This should apply to all zlib functions as well, I suppose.

jasnell:
> This should apply to all zlib functions as well, I suppose.

There will be quite a few places to make this change. I'll be going through them incrementally with a number of small PRs

PR #3099, opened 2024-11-12
use llvm-symbolizer directly on justfile

anonrig:
Rather than accessing v18, let's directly use `llvm-symbolizer`. This also adds symbolizer to other commands so we don't receive `--test-env` changed warnings.

### DIFF
--- a/justfile
+++ b/justfile
@@ -7,7 +7,6 @@ default:
   @just --list
 
 pwd := `pwd`
-clang_version := "18"
 
 prepare:
   cargo install gen-compile-commands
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -19,20 +18,20 @@ clean:
   rm -f compile_commands.json
 
 build *args="//...":
-  bazel build {{args}}
+  bazel build {{args}} --test_env=LLVM_SYMBOLIZER=llvm-symbolizer
```
danlapid, line 21: Why not just edit bazel so that this is set as the default value if the environment variable isn't set?

fhanau:
On my system (devspace), I don't have a non-versioned llvm-symbolizer available, so that's a use case where the old approach would be better. (You could argue that I should create a symlink for plain llvm-symbolizer, but ideally that should be documented somewhere... not sure if there's a perfect solution here)

danlapid:
> On my system (devspace), I don't have a non-versioned llvm-symbolizer available, so that's a use case where the old approach would be better. (You could argue that I should create a symlink for plain llvm-symbolizer, but ideally that should be documented somewhere... not sure if there's a perfect solution here)

A near-perfect solution is iterate over all llvm-symbolizer from 30 to 15 and see if it exists in path and use it but seems like overkill to me.

justfile
mikea: I agree, let's add this to .bazelrc

anonrig: I see that we already have `build:unix --test_env=LLVM_SYMBOLIZER=llvm-symbolizer` inside `.bazelrc` file, and it doesn't work for me running on unix. I still need to pass `--test_env=LLVM_SYMBOLIZER=llvm-symbolizer`

mikea: wat? If it is indeed linux or macos, then there's

```
build:linux --config=unix
build:macos --config=unix
```

somethis is broken in your config?

PR #3102, opened 2024-11-13
Erikcorry/multi sandbox

erikcorry:


src/workerd/jsg/buffersource.h
```diff
@@ -69,19 +69,29 @@ using BufferSourceViewConstructor = v8::Local<v8::Value> (*)(Lock&, BackingStore
...
+    if (js.v8Isolate->GetGroup().SandboxContains(data.begin())) {
+      auto ptr = new kj::Array<byte>(kj::mv(data));
+      return BackingStore(
+          v8::ArrayBuffer::NewBackingStore(ptr->begin(), size,
+              [](void*, size_t, void* ptr) { delete reinterpret_cast<kj::Array<byte>*>(ptr); }, ptr),
+          size, 0, getBufferSourceElementSize<T>(), construct<T>, checkIsIntegerType<T>());
+    } else {
+      auto result = BackingStore(v8::ArrayBuffer::NewBackingStore(js.v8Isolate, size), size, 0,
+          getBufferSourceElementSize<T>(), construct<T>, checkIsIntegerType<T>());
+      memcpy(result.asArrayPtr().begin(), data.begin(), size);
```
jasnell, line 92: Use `result.AsArrayPtr().copyFrom(...)` here instead?

src/workerd/jsg/buffersource.h
```diff
@@ -69,19 +69,29 @@ using BufferSourceViewConstructor = v8::Local<v8::Value> (*)(Lock&, BackingStore
...
+// safely used outside of the isolate lock.  It cannot be passed to a different
+// isolate, unless they are in the same isolate group, which means they are in
+// the same sandbox.
 class BackingStore {
 public:
   template <BufferSourceType T = v8::Uint8Array>
-  static BackingStore from(kj::Array<kj::byte> data) {
+  static BackingStore from(Lock& js, kj::Array<kj::byte> data) {
     // Creates a new BackingStore that takes over ownership of the given kj::Array.
+    // The bytes may be moved if they are not inside the sandbox already.
```
jasnell, line 81:
```suggestion
    // The bytes may be copied if they are not inside the sandbox already.
```

src/workerd/jsg/buffersource.h
jasnell: Also, using `jsg::BackingStore::alloc<T>(js, N)` would be easier here?

src/workerd/jsg/value.h
```diff
@@ -916,28 +916,40 @@ class ArrayBufferWrapper {
...
-    //   unifying Array<T> and Vector<T> in the future (i.e. making all Array<T>s growable). So
-    //   it may be best to stick with allocating an Array<byte> on the heap after all...
     size_t size = value.size();
     if (size == 0) {
       // BackingStore doesn't call custom deleter if begin is null, which it often is for empty
       // arrays.
       return v8::ArrayBuffer::New(isolate, 0);
     }
     byte* begin = value.begin();
+    if (isolate->GetGroup().SandboxContains(begin)) {
```
jasnell, line 926: Perhaps this could be extracted into a separate utility since there are two places doing this same check and operation?

PR #3104, opened 2024-11-13
Update Node.js buffer.transcode to use BufferSource

jasnell:
Ensure more allocations are known to the Isolate and occur within the Isolate allocators

### DIFF
--- a/src/node/internal/internal_buffer.ts
+++ b/src/node/internal/internal_buffer.ts
@@ -2688,9 +2688,13 @@ export function transcode(
   if (normalizedToEncoding === undefined) {
     throw new ERR_UNKNOWN_ENCODING(toEncoding);
   }
-  return Buffer.from(
-    bufferUtil.transcode(source, normalizedFromEncoding, normalizedToEncoding)
+
+  const u8: Uint8Array = bufferUtil.transcode(
+    source,
+    normalizedFromEncoding,
+    normalizedToEncoding
   );
+  return Buffer.from(u8.buffer, u8.byteOffset, u8.byteLength);
 }
 
 export function resolveObjectURL(_id: string): unknown {
--- a/src/workerd/api/node/i18n.c++
+++ b/src/workerd/api/node/i18n.c++
@@ -23,9 +23,9 @@ namespace i18n {
 namespace {
 
 // An isolate has a 128mb memory limit.
-const int ISOLATE_LIMIT = 134217728;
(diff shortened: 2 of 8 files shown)


src/node/internal/internal_buffer.ts
```diff
@@ -2688,9 +2688,13 @@ export function transcode(
   if (normalizedToEncoding === undefined) {
     throw new ERR_UNKNOWN_ENCODING(toEncoding);
   }
-  return Buffer.from(
-    bufferUtil.transcode(source, normalizedFromEncoding, normalizedToEncoding)
+
+  const u8: Uint8Array = bufferUtil.transcode(
```
anonrig, line 2692: Why do we need this?

```suggestion
  const u8 = bufferUtil.transcode(
```

src/workerd/api/node/i18n.c++
```diff
@@ -40,100 +40,130 @@ const char* getEncodingName(Encoding input) {
...
-    kj::ArrayPtr<kj::byte> source, Encoding fromEncoding, Encoding toEncoding) {
+kj::Maybe<jsg::BufferSource> TranscodeFromUTF16(
+    jsg::Lock& js, kj::ArrayPtr<kj::byte> source, Encoding fromEncoding, Encoding toEncoding) {
   Converter to(toEncoding);
   auto substitute = kj::str(kj::repeat('?', to.minCharSize()));
   to.setSubstituteChars(substitute);
 
   auto utf16_input = kj::arrayPtr<char16_t>(
-      reinterpret_cast<char16_t*>(source.begin()), source.size() / sizeof(UChar));
+      reinterpret_cast<char16_t*>(source.begin()), source.size() / sizeof(char16_t));
```
anonrig, line 109: Nit: Technically this is correct, since UChar is equal to char16_t, but this function uses ICU which returns UChar. In the future, if things change, this would cause some issues (and many more).

anonrig (approved):

src/node/internal/internal_buffer.ts
jasnell: Mostly for clarity since `bufferUtil.transcode(...)` used to return an `ArrayBuffer`

src/workerd/api/node/i18n.c++
jasnell: `UChar` is *exceedingly* unlikely to change and doing so would cause other issues if it did. We'd see test failures all over I would expect.

PR #3107, opened 2024-11-13
enable built-in compile cache behind autogate

anonrig:
Enables compile cache behind an autogate flag.

### DIFF
--- a/src/workerd/jsg/modules.c++
+++ b/src/workerd/jsg/modules.c++
@@ -6,6 +6,8 @@
 #include "jsg.h"
 #include "setup.h"
 
+#include <workerd/util/autogate.h>
+
 #include <kj/mutex.h>
 
 #include <set>
@@ -365,6 +367,7 @@ static CompilationObserver::Option convertOption(ModuleInfoCompileOption option)
 v8::Local<v8::Module> compileEsmModule(jsg::Lock& js,
     kj::StringPtr name,
     kj::ArrayPtr<const char> content,
+    kj::ArrayPtr<const kj::byte> compileCache,
     ModuleInfoCompileOption option,
     const CompilationObserver& observer) {
   // destroy the observer after compilation finished to indicate the end of the process.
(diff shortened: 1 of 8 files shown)


src/workerd/jsg/modules.c++
```diff
@@ -365,53 +367,45 @@ static CompilationObserver::Option convertOption(ModuleInfoCompileOption option)
...
-  v8::ScriptCompiler::Source source(contentStr, origin, existingCacheData);
-  auto module =
-      jsg::check(v8::ScriptCompiler::CompileModule(js.v8Isolate, &source, compileOptions));
-
-  if (existingCacheData == nullptr) {
-    auto cachedData = std::shared_ptr<v8::ScriptCompiler::CachedData>(
-        v8::ScriptCompiler::CreateCodeCache(module->GetUnboundModuleScript()));
-    compileCache.add(name, kj::mv(cachedData));
+  if (util::Autogate::isEnabled(util::AutogateKey::COMPILE_CACHE_FOR_BUILTINS)) {
+    if (compileCache.size() == 0) {
```
jasnell, line 399: Is this meant to be `> 0` and not `== 0` here?

src/workerd/jsg/modules.c++
```diff
@@ -365,53 +367,45 @@ static CompilationObserver::Option convertOption(ModuleInfoCompileOption option)
 v8::Local<v8::Module> compileEsmModule(jsg::Lock& js,
     kj::StringPtr name,
     kj::ArrayPtr<const char> content,
+    kj::ArrayPtr<kj::byte> compileCache,
```
jasnell, line 370: Should this be `kj::ArrayPtr<const kj::byte>`?

src/workerd/server/workerd-api.c++
```diff
@@ -398,7 +398,8 @@ kj::Maybe<jsg::ModuleRegistry::ModuleInfo> WorkerdApi::tryCompileModule(jsg::Loc
               lock, Impl::compileJsonGlobal(lock, module.getJson())));
     }
     case config::Worker::Module::ES_MODULE: {
-      return jsg::ModuleRegistry::ModuleInfo(lock, module.getName(), module.getEsModule(),
+      // TODO(soon): Make sure passing `{}` to compile cache is desired.
+      return jsg::ModuleRegistry::ModuleInfo(lock, module.getName(), module.getEsModule(), {},
```
mikea, line 402: maybe nullptr looks better in this context?

Also please `nullptr /* compile cache */,`

src/workerd/jsg/modules.c++
anonrig: Yes!

mikea (approved):
@jasnell stole some of my comments. otherwise lg.

src/workerd/server/workerd-api.c++
anonrig: Definitely!

src/workerd/jsg/modules.c++
anonrig: Yes!

anonrig:
I believe this is ready to land @jasnell @mikea. Appreciate if you could take one more look before landing this. PS: @kentonv would you mind taking a look as well?

src/workerd/jsg/modules.c++
```diff
@@ -365,53 +367,46 @@ static CompilationObserver::Option convertOption(ModuleInfoCompileOption option)
...
 
   // Must pass true for `is_module`, but we can skip everything else.
-  const int resourceLineOffset = 0;
-  const int resourceColumnOffset = 0;
-  const bool resourceIsSharedCrossOrigin = false;
-  const int scriptId = -1;
-  const bool resourceIsOpaque = false;
-  const bool isWasm = false;
-  const bool isModule = true;
+  static constexpr int resourceLineOffset = 0;
```
danlapid, line 378: nit: static constexpr is superfluous, just constexpr would be the same

src/workerd/util/autogate.h
```diff
@@ -15,6 +15,7 @@ namespace workerd::util {
 enum class AutogateKey {
   TEST_WORKERD,
   PYTHON_EXTERNAL_BUNDLE,
+  COMPILE_CACHE_FOR_BUILTINS,
```
danlapid, line 18: Just to be sure, you intend for this to be an autogate and not a compat flag right?
Autogates change code paths per metal for all workers
Compat flags are per worker for all metals.

src/workerd/server/workerd-api.c++
```diff
@@ -398,8 +398,9 @@ kj::Maybe<jsg::ModuleRegistry::ModuleInfo> WorkerdApi::tryCompileModule(jsg::Loc
               lock, Impl::compileJsonGlobal(lock, module.getJson())));
     }
     case config::Worker::Module::ES_MODULE: {
+      // TODO(soon): Make sure passing nullptr to compile cache is desired.
```
danlapid, line 401: todo

danlapid (approved):
LGTM

src/workerd/jsg/modules.c++
anonrig: definitely

src/workerd/util/autogate.h
anonrig: Yes, it's intentional. I don't want to enable compile cache and break millions of people. Gradually enabling it is beneficial for safety.

src/workerd/server/workerd-api.c++
anonrig: @jasnell any thoughts on this? what's the code-path that triggers this particular case?

src/workerd/util/autogate.h
danlapid: If it is expected to be a breaking change for people it definitely needs to be a compat flag.
If it's not expected to be a breaking change but it might be in case of a bug we missed then it can be an autogate.

src/workerd/server/workerd-api.c++
danlapid: One such example is workerd-api.c++:557-558 where we compile PYTHON_ENTRYPOINT.
Would've been nice if that was cached.
But it's all relevant in workerd-api only which is only used for local dev so I don't think it matters all too much.

src/workerd/util/autogate.h
anonrig: It's not a breaking change, there might be a bug, hence: auto-gate.

jasnell: There's no need for a compat flag here as there is no breaking API or behavioral changes. Autogate is appropriate.

src/workerd/server/workerd-api.c++
jasnell: This is appropriate here. We are not yet enabling compile cache for anything but built-ins. The TODO could be reworded accordingly but this is fine also

PR #3109, opened 2024-11-13
v8 13.2 support

mikea:
This will let me upgrade it without juggling too many PRs in the process.

### DIFF
--- a/src/workerd/jsg/setup.c++
+++ b/src/workerd/jsg/setup.c++
@@ -252,11 +252,14 @@ HeapTracer& HeapTracer::getTracer(v8::Isolate* isolate) {
   return IsolateBase::from(isolate).heapTracer;
 }
 
+// todo: cleanup after 13.2 upgrade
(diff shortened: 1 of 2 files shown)


src/workerd/jsg/setup.c++
```diff
@@ -252,11 +252,14 @@ HeapTracer& HeapTracer::getTracer(v8::Isolate* isolate) {
   return IsolateBase::from(isolate).heapTracer;
 }
 
+// todo: cleanup after 13.2 upgrade
+#if V8_MINOR_VERSION < 2
```
danlapid, line 256: Mind adding a check for V8_MAJOR_VERSION == 13 as well please?

danlapid: We can add #error if v8 isn't 13 or minor_version > 2 just to make sure we can't forget to clean this up

mikea: sure, done. Though I wanted it to rot if we don't clean it up soon enough.

danlapid (approved):

PR #3112, opened 2024-11-14
Give DurableObjectStorage reference to primary

MellowYarker:


justin-mp:
> @justin-mp alright if I consolidate to 1 commit and mark you as a co-author?

Fine with me.

src/workerd/io/io-channels.h
```diff
@@ -4,6 +4,9 @@
 
 #pragma once
 
+#include "workerd/io/actor-cache.h"
+#include "workerd/io/io-own.h"
```
jasnell, line 8: Nit: are these includes actually needed in io-channels.h or is this just to make the decls available in the other files changed in this PR? If the latter, I'd prefer if the includes were added where they are actually used.

jasnell (approved):

jasnell (approved):

jasnell:
A PR so nice I approved it thrice. 

image

Really not sure what happened here, Github.

src/workerd/api/actor-state.h
```diff
@@ -177,6 +179,16 @@ class DurableObjectStorage: public jsg::Object, public DurableObjectStorageOpera
       : cache(kj::mv(cache)),
         enableSql(enableSql) {}
 
+  // This constructor is only used when we're setting up the `DurableObjectStorage` for a replica
+  // Durable Object instance. Replica's need to retain a reference to their primary so they can
```
justin-mp, line 183: nit: Replicas (plural, not possessive)

justin-mp (commented):
LGTM, but please squash the 2 commits and clean up the files in src/workerd/io since I don't think we need any of those changes.

justin-mp (approved):

src/workerd/io/io-channels.h
MellowYarker: Had initially needed these includes, but took the PR in another direction. Thanks for catching!

kentonv:
FYI, this build was merged with CI failing. It was an "optional" CI build, but the failure was in fact caused by this change and so should have been considered blocking.

PR #3117, opened 2024-11-14
Fix SQL statement caching of computed query strings.

kentonv:
I'd forgotten to internalize the string, but the hash map was comparing strings by identity, so computed strings would always miss cache. (Worse, their hash codes were computed by content, so they would cause excessive hash collisions.)


justin-mp (approved):

src/workerd/jsg/jsvalue.h
```diff
@@ -240,6 +240,12 @@ class JsString final: public JsBase<v8::String, JsString> {
 
   bool operator==(const JsString& other) const;
 
+  // "Internalize" the string. Returns a string with the same content but which is identity-equal
+  // to all other internalized strings with the same content. If the string is already
+  // internalized, this returns the same value. Note that strings originating from literals in the
+  // code are always internalized.
+  JsString internalize(Lock& js) const;
```
jclee, line 247: Huh... Looks like the V8 headers have a comment that says internalization is a "hint", which makes me wonder if there are cases where it is (or was, in some earlier V8?) not guaranteed to be applied.  Not that it particularly matters for this change, since it's just improving the performance of cache lookups, which seems consistent with V8's other uses of internalization.  And the tests should let us know if it breaks for some reason.

https://github.com/v8/v8/blob/13.1.201.8/include/v8-primitive.h#L113-L118

Issue #3090, opened 2024-11-11
Cannot use browser rendering with svelte kit


simonxabris:
I created a new sveltekit app with typescript using `create-cloudflare` and I wanted to use browser rendering, but I am unable to call `puppeteer.launch` with `.env.BROWSER`, because I get the following type error:

<details>
<summary>TS error output</summary>

<img width="650" alt="Screenshot 2024-11-11 at 10 25 51" src="https://github.com/user-attachments/assets/7ed75d4b-4bab-4507-93be-bd27479f994c">
</details>

I did a bit of digging and figured out where the problem comes from. This is how the `BrowserWorker` type is defined in `workers-types`:

```ts
export declare interface BrowserWorker {
    fetch: typeof fetch;
}
``` 

The problem with this is `fetch` will resolve to multiple different  type definitions depending on your TS config. In this case it resolves to three. The one in `workers-types`, the one in `@types/node` and the one in typescript's `DOM` lib definitions. The latter is incompatible with the former two and that is what's causing this error.

I tried experimenting with a solution that specifies the type of fetch required by `BrowserWorker` exactly, something like:

```
export declare interface BrowserWorker {
   fetch: import("@cloudflare/workers-types/2023-07-01").Fetcher["fetch"];
}
``` 

but this also didn't work fully, as then `RequestInfo` was resolved from the `DOM` lib and I couldn't fix that one.

Also, this is a problem for basically all full-stack frameworks as all of the use the `DOM` type definitions.

petebacondarwin:
The simplest solution here is just to cast the BROWSER binding to a BrowserWorker:

```ts
  await puppeteer.launch(platform!.env.BROWSER as BrowserWorker);
```

In my IDE this removes the complaint from the type checker.

petebacondarwin:
Alternatively, you can change the app.d.ts file so that `env.BROWSER` is of type `BrowserWorker`:

```ts
declare global {
  namespace App {
    interface Platform {
      env: {
        BROWSER: BrowserWorker;
      };
      cf: CfProperties;
      ctx: ExecutionContext;
    }
  }
}
```

petebacondarwin:
But perhaps the fix we should put into workers-types is to change the `Fetcher` interface. From:

```ts
type Fetcher<
  T extends Rpc.EntrypointBranded | undefined = undefined,
  Reserved extends string = never,
> = (T extends Rpc.EntrypointBranded
[...]

PR #3121, opened 2024-11-15
Move `URL` out of `RequestInfo` to match DOM types

penalosa:
Fixes https://github.com/cloudflare/workerd/issues/3090

To improve the typing situation when loading dom types in the same environment as workers-types, this PR moves the `URL` type out of `RequestInfo`, in favour of e.g. `fetch()` referencing `fetch(req: RequestInfo | URL)` directly. This matches how dom types are defined.

There should be no user-observable behaviour change for users using just workers-types, but this should improve the situation if e.g. the fetch definition comes from workers-types but the `RequestInfo` definition comes from dom types.

### DIFF
--- a/src/workerd/api/cache.h
+++ b/src/workerd/api/cache.h
@@ -72,11 +72,10 @@ public:
     JSG_METHOD(keys);
 
     JSG_TS_OVERRIDE({
-      delete(request: RequestInfo, options?: CacheQueryOptions): Promise<boolean>;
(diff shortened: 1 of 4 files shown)


petebacondarwin:
I manually applied this fix (or approximately this fix 😄) to the reproduction in the issue and it appears to resolve the typings problem.

andyjessop (approved):
Nice work 👍

petebacondarwin (approved):

jasnell (approved):

PR #3127, opened 2024-11-15
Ensure that crypto allocations are using BufferSource

jasnell:
... ensures that all ArrayBuffer instances are using BufferSource rather than external allocations.

### DIFF
--- a/src/workerd/api/crypto/crypto.c++
+++ b/src/workerd/api/crypto/crypto.c++
@@ -317,7 +317,7 @@ bool CryptoKey::verifyX509Private(const X509* cert) const {
   return impl->verifyX509Private(cert);
 }
 
-jsg::Promise<kj::Array<kj::byte>> SubtleCrypto::encrypt(jsg::Lock& js,
+jsg::Promise<jsg::BufferSource> SubtleCrypto::encrypt(jsg::Lock& js,
     kj::OneOf<kj::String, EncryptAlgorithm> algorithmParam,
     const CryptoKey& key,
     kj::Array<const kj::byte> plainText) {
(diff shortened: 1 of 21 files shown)


jasnell:
A handful of the remaining uses of `kj::heapArray` could actually be converted to `KJ_STACK_ARRAY` but will handle that cleanup in a separate PR.

anonrig (approved):

src/workerd/api/crypto/crypto.h
```diff
@@ -207,19 +207,28 @@ class CryptoKey: public jsg::Object {
       KJ_SWITCH_ONEOF(publicExponent) {
         KJ_CASE_ONEOF(array, BigInteger) {
           if (fixPublicExp) {
-            auto expCopy = kj::heapArray<kj::byte>(array.asPtr());
-            jsg::BackingStore expBack = jsg::BackingStore::from(kj::mv(expCopy));
+            // alloc will, by default create a Uint8Array
```
erikcorry, line 210: If the comment is intended to document what type is hiding behind `auto` then would it not be nicer to just write the type there.

If that's not what is intended, then the reason for the comment is unclear.

src/workerd/api/crypto/crypto.c++
```diff
@@ -538,18 +542,17 @@ jsg::Promise<jsg::Ref<CryptoKey>> SubtleCrypto::unwrapKey(jsg::Lock& js,
 
     validateOperation(unwrappingKey, normalizedAlgorithm.name, CryptoKeyUsageSet::unwrapKey());
 
-    kj::Array<kj::byte> bytes =
-        unwrappingKey.impl->unwrapKey(kj::mv(normalizedAlgorithm), wrappedKey);
+    auto bytes = unwrappingKey.impl->unwrapKey(js, kj::mv(normalizedAlgorithm), wrappedKey);
```
erikcorry, line 545: Not really a comment for this PR, but I wish we had less auto and more explicit types, and here you are going the other way.  For the reader it's just less clear.  Could be that the IDE tells you the type, but we don't always have that (eg here in the code review).

erikcorry (approved):
This is great, should help with the sandboxing of V8.

src/workerd/api/crypto/crypto.c++
jasnell: Noted ;-) .... I think we've found the c++ equivalent to spaces vs. tabs tho lol ... some folks seem to really like `auto` (like myself) and others really hate it

Pjrich1313 (commented):
Rerun

PR #3132, opened 2024-11-18
new clang warning fix

mikea:


fhanau (approved):
LGTM. IIRC there were more such warnings but I trust you checked that these are all that appear for workerd.

PR #3140, opened 2024-11-19
Implements support for external package loading in validator.

dom96:
Moves package loading to `prepareWasmLinearMemory` so that `preloadDynamicLibs` can use the loaded packages instead of the embedded ones.

Also fixes a stack-use-after-scope error in `getPyodidePackage`

### Test Plan

```
$ bazel test @workerd//src/cloudflare/internal/test/... --config=asan
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,6 +2,7 @@ common --enable_platform_specific_config
 build --verbose_failures
 build --build_tag_filters=-off-by-default
 test --test_tag_filters=-off-by-default
+test:asan --test_tag_filters=-off-by-default,-no-asan
 # exclude enormous tests by default
 build --test_size_filters=-enormous
 
--- a/src/pyodide/internal/python.ts
+++ b/src/pyodide/internal/python.ts
@@ -1,6 +1,7 @@
 import { enterJaegerSpan } from 'pyodide-internal:jaeger';
 import {
   SITE_PACKAGES,
+  TRANSITIVE_REQUIREMENTS,
   adjustSysPath,
   mountSitePackages,
   mountWorkerFiles,
@@ -50,6 +51,7 @@ import {
   setUnsafeEval,
   setGetRandomValues,
 } from 'pyodide-internal:generated/emscriptenSetup';
+import { loadPackages } from 'pyodide-internal:loadPackage';
 
 /**
  * After running `instantiateEmscriptenModule` but before calling into any C
@@ -59,9 +61,8 @@ import {
  */
 async function prepareWasmLinearMemory(Module: Module): Promise<void> {
   // Note: if we are restoring from a snapshot, runtime is not initialized yet.
-  mountSitePackages(Module, SITE_PACKAGES.rootInfo);
-  entropyMountFiles(Module);
   Module.noInitialRun = !SHOULD_RESTORE_SNAPSHOT;
+
   enterJaegerSpan('preload_dynamic_libs', () => preloadDynamicLibs(Module));
   enterJaegerSpan('remove_run_dependency', () =>
     Module.removeRunDependency('dynlibs')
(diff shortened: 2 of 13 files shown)


src/workerd/api/pyodide/pyodide.c++
```diff
@@ -35,10 +35,9 @@ void PyodideBundleManager::setPyodideBundleData(
       kj::mv(version), {.messageReader = kj::mv(messageReader), .bundle = bundle});
 }
 
-const kj::Maybe<kj::ArrayPtr<const unsigned char>> PyodidePackageManager::getPyodidePackage(
+const kj::Maybe<const kj::Array<unsigned char>&> PyodidePackageManager::getPyodidePackage(
```
hoodmane, line 38: What's the reason for this change?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -176,11 +163,12 @@ try {
   if (!IS_TRACING) {
     if (IS_WORKERD) {
       // If we're in workerd, we have to do setupPackages in the IoContext, so don't start it yet.
+      //
+      // Also, not doing this here seems to cause @workerd//src/cloudflare/internal/test/d1:python-d1-api-test
+      // to fail under --config=asan.
+      //
       // TODO: fix this.
       await getPyodide();
```
dom96, line 171: If I remove this call then some of our workerd tests fail under asan.

https://paste.cfdata.org/GdKtTZYLiz6LktnWP5gP

Hood mentioned this may be related to the isolate pool changes. @danlapid any ideas what might be going on here?

src/workerd/api/pyodide/pyodide.c++
dom96: I was getting stack-use-after-scope errors before this change. I'm guessing it was returning a pointer to the reference, rather than the array itself. This seems to fix it.

src/pyodide/internal/python.ts
```diff
@@ -62,6 +64,11 @@ async function prepareWasmLinearMemory(Module: Module): Promise<void> {
   mountSitePackages(Module, SITE_PACKAGES.rootInfo);
   entropyMountFiles(Module);
   Module.noInitialRun = !SHOULD_RESTORE_SNAPSHOT;
+
+  // NB. loadPackages adds the packages to the `SITE_PACKAGES` global which then gets used in
+  // preloadDynamicLibs.
+  await loadPackages(Module, TRANSITIVE_REQUIREMENTS);
```
hoodmane, line 70: I wonder line 70 and lines 64 and 65 should be moved into `loadPyodide`. Since they don't strictly speaking have anything to do with `prepareWasmLinearMemory`. And is it worth giving `loadPackages` its own jaeger span?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -119,7 +105,8 @@ function getMainModule(): Promise<PyModule> {
     }
     mainModulePromise = (async function () {
       const pyodide = await getPyodide();
-      await setupPackages(pyodide);
+      patchLoadPackage(pyodide);
```
hoodmane, line 108: Maybe move this back into `setupPatches`? It sounds related.

src/pyodide/internal/setupPackages.ts
```diff
@@ -44,6 +45,7 @@ class SitePackagesDir {
       path: '',
       name: '',
       parts: [],
+      reader: null,
```
hoodmane, line 48: What about setting `reader: EmbeddedPackagesTarReader` and removing the `node.reader ?? EmbeddedPackagesTarReader` logic in `snapshot.ts`? Would that work?

hoodmane (approved):
I made a few minor questions and suggestions.

src/pyodide/internal/setupPackages.ts
dom96: Probably, but doesn't matter much. I plan to get rid of `EmbeddedPackagesTarReader` in a follow up PR.

src/pyodide/python-entrypoint-helper.ts
dom96: Had to remove it again. No way around it. I'll probably disable these tests under asan for now but we should fix this asap, this asan issue is likely lurking there right now and just isn't being triggered.

.bazelrc
```diff
@@ -188,6 +188,7 @@ build:asan --config=sanitizer-common
 build:asan --copt="-fsanitize=address" --linkopt="-fsanitize=address"
 build:asan --test_env=ASAN_OPTIONS=abort_on_error=true
 build:asan --test_env=KJ_CLEAN_SHUTDOWN=1
+build:asan --test_tag_filters=-no-asan
```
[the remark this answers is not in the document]
dom96: good catch, thanks!

src/pyodide/python-entrypoint-helper.ts
danlapid: Pool code hasn't been merged yet so I don't see how it might be related

PR #3145, opened 2024-11-20
Create and assign an InvocationSpanContext for every invocation

jasnell:
@fhanau ... please consider this one carefully. What I'm not sure about is whether the `InvocationSpanContext` should be held by the `IncomingRequest` or something internal to the `RequestObserver`.

Eventually I'd like to abstract out the `WorkerTracer` here and actually hide that inside the `RequestObserver` and put it entirely behind `SpanBuilder` but that'll come later. Accordingly, however, I don't want to put `InvocationSpanContext` inside the `WorkerTracer`.

The other thing to keep in mind is that we'll need to be able to propagate the `InvocationSpanContext` across subrequests of various kinds so we'll need easy access to it to place it into things like the control header.

How is this expected to be used? Good question! When streaming tail workers is fully operational, the invocation span context will be recorded with each mark event. Every child span created during the invocation of a worker will become a child off this root and should have its own `kj::Rc<InvocationSpanContext>` and there should always be a "current invocation span context" at any given time during an invocation that is propagated via the async context just like the current SpanParent. The fact that the user SpanBuilder would need to have access to this suggests to me that *maybe* the root `InvocationSpanContext` should be owned by the `RequestObserver` but I'm not entirely sure.

### DIFF
--- a/src/workerd/io/io-context.c++
+++ b/src/workerd/io/io-context.c++
@@ -201,11 +201,13 @@ IoContext::IoContext(ThreadContext& thread,
 IoContext::IncomingRequest::IoContext_IncomingRequest(kj::Own<IoContext> contextParam,
     kj::Own<IoChannelFactory> ioChannelFactoryParam,
     kj::Own<RequestObserver> metricsParam,
-    kj::Maybe<kj::Own<WorkerTracer>> workerTracer)
+    kj::Maybe<kj::Own<WorkerTracer>> workerTracer,
+    kj::Rc<tracing::InvocationSpanContext> invocationSpanContext)
     : context(kj::mv(contextParam)),
       metrics(kj::mv(metricsParam)),
       workerTracer(kj::mv(workerTracer)),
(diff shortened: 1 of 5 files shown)


fhanau (commented):
Code changes LGTM, I'll have to think about IncomingRequest vs. RequestObserver a bit more.

fhanau:
> @fhanau ... please consider this one carefully. What I'm not sure about is whether the `InvocationSpanContext` should be held by the `IncomingRequest` or something internal to the `RequestObserver`.

The thoughts above seem reasonable, I don't have much to add. Intuitively, having the InvocationSpanContext be owned by RequestObserver in some capacity feels more sensible/easier to support, but I could also see it work with IncomingRequest. Having it in/around RequestObserver would be more in line with the current model, but I guess we're making big changes anyway.

fhanau (approved):

jasnell:
Ok, well, we can keep it here for now and can always move it later if we discover it makes more sense in `RequestObserver`

PR #3148, opened 2024-11-20
SQC-333: Revert previous changes

ReppCodes:
This commit reverts 1149737171addcb3705314ce56eaa2196f449710. The sslmode change in the previous commit does not work as expected. Connection attempts time out, presumably getting hung while negotiating SSL upgrade. Revert this change until we have the bandwidth to attempt it again.

### DIFF
--- a/src/workerd/api/hyperdrive.c++
+++ b/src/workerd/api/hyperdrive.c++
@@ -80,7 +80,7 @@ uint16_t Hyperdrive::getPort() {
 
 kj::String Hyperdrive::getConnectionString() {
   return kj::str(getScheme(), "://", getUser(), ":", getPassword(), "@", getHost(), ":", getPort(),
-      "/", getDatabase(), "?sslmode=prefer");
+      "/", getDatabase(), "?sslmode=disable");
 }
 
 kj::Promise<kj::Own<kj::AsyncIoStream>> Hyperdrive::connectToDb() {


jasnell (approved):

src/workerd/api/hyperdrive.c++
```diff
@@ -80,7 +80,7 @@ uint16_t Hyperdrive::getPort() {
 
 kj::String Hyperdrive::getConnectionString() {
   return kj::str(getScheme(), "://", getUser(), ":", getPassword(), "@", getHost(), ":", getPort(),
-      "/", getDatabase(), "?sslmode=prefer");
```
jasnell, line 83: you might consider putting the change behind an autogate so that the availability of the change can be rolled out with a salt gate instead of needing a new release.

ReppCodes: I'll have to explore that, I haven't spent a ton of time in the core edgeworker repos yet. Thanks for the tip!

I think this particular change should be okay going out with the next release, though.

thomasgauvin: Good to patch this now and have a proper fix in the next release, customers have to install specific version. All local development is broken right now

anonrig (approved):

thomasgauvin:
@jasnell @anonrig any chance we can merge this outside of the usual train? local development with Hyperdrive is completely broken right now

PR #3149, opened 2024-11-20
replace symbolizer with kj built-in

anonrig:
Since we have stringifyStackTraceWithLlvm on KJ, we can remove our own implementation. I'm not sure if this is a 1 to 1 replacement since previous implementation has sentry as a dependency...

Ref: https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/exception.c%2B%2B#L290


src/workerd/jsg/jsg.c++
```diff
@@ -9,6 +9,8 @@
 #include <workerd/jsg/util.h>
 #include <workerd/util/thread-scopes.h>
 
+#include <kj/exception.h>
```
jasnell, line 12: It's not clear why this additional include is needed?

mikea (approved):

src/workerd/jsg/jsg.c++
anonrig: Ooops.

jasnell (approved):

Issue #3066, opened 2024-11-06
🐛 Bug Report — URLSearchParams body sets incorrect fetch content-type


nickbabcock:
```js
export default {
    async fetch(request, env, ctx) {
        const r = new Request('https://example.com/', {
            method: 'POST',
            body: new URLSearchParams({ hello: 'world' })
        });
        return new Response(r.headers.get('content-type'))
    },
};
```

Expected ([ref](https://fetch.spec.whatwg.org/#bodyinit-unions)):
```
application/x-www-form-urlencoded;charset=UTF-8
```

Actual:
```
text/plain;charset=UTF-8
```

This behavior differs from browsers and node.js

Replicated in cloudflare workers playground and in a production workers deployment.

balusch:
It seems that `api::url::URLSearchParams` instead of `api::URLSearchParams` should be used in `Body::Initializer`:

https://github.com/cloudflare/workerd/blob/f32706b15ea3daa0e13a097d25d259d6ccf942be/src/workerd/api/http.h#L255

PR #3151, opened 2024-11-21
Add missing api::url::URLSearchParams in Body::Initializer.

balusch:
Fix https://github.com/cloudflare/workerd/issues/3066

### DIFF
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -14,6 +14,7 @@
 #include "form-data.h"
 #include "web-socket.h"
 #include <workerd/api/url.h>
+#include <workerd/api/url-standard.h>
 #include "blob.h"
 #include <workerd/io/compatibility-date.capnp.h>
 #include "worker-rpc.h"
@@ -252,7 +253,8 @@ public:
   // from any of the other source types, Body can create a new ReadableStream from the source, and
   // the POST will successfully retransmit.
   using Initializer = kj::OneOf<jsg::Ref<ReadableStream>, kj::String, kj::Array<byte>,
-                                jsg::Ref<Blob>, jsg::Ref<URLSearchParams>, jsg::Ref<FormData>>;
+                                jsg::Ref<Blob>, jsg::Ref<FormData>,
+                                jsg::Ref<URLSearchParams>, jsg::Ref<url::URLSearchParams>>;
 
   struct RefcountedBytes final: public kj::Refcounted {
     kj::Array<kj::byte> bytes;
--- a/src/workerd/api/url-standard.h
+++ b/src/workerd/api/url-standard.h
@@ -96,16 +96,14 @@ public:
     JSG_ITERABLE(entries);
 
     if (!flags.getSpecCompliantUrl()) {
-      // This is a hack. The spec-compliant URLSearchParams type is used in
-      // the Body constructor, see:
-      // https://github.com/balusch/workerd/tree/main/src/workerd/api/http.h#L255
-      // This means that when the TypeScript generation scripts are visiting root types for inclusion,
-      // we'll always visit the spec-compliant type even if we have the
-      // "url-standard" flag disabled. Rather than updating those usages based
-      // on which flags are enabled, we just delete the spec compliant
-      // declaration in an override if "url-standard" is disabled.
-      // We do the same for the non-spec-compliant URLSearchParams, see:
-      // https://github.com/balusch/workerd/tree/main/src/workerd/api/url.h#L219
+      // This is a hack. The spec-compliant URLSearchParams type is used in the Body constructor,
+      // see https://github.com/cloudflare/workerd/blob/v1.20241127.0/src/workerd/api/http.h#L255
+      // This means that when the TypeScript generation scripts are visiting root types for
+      // inclusion, we'll always visit the spec-compliant type even if we have the "url-standard"
+      // flag disabled. Rather than updating those usages based on which flags are enabled, we just
+      // delete the spec compliant declaration in an override if "url-standard" is disabled.
+      // We do the same for the non-spec-compliant URLSearchParams
+      // (https://github.com/cloudflare/workerd/blob/v1.20241127.0/src/workerd/api/url.h#L219).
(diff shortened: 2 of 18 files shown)


balusch:
Hi @jasnell! Could you please review this PR?

I'm excited and eager to contribute to this project. While browsing the issues, I noticed that #3066 was proposed without a corresponding PR. I decided to address it.

I’m still trying to fully understand the issue. I see that both `EW_URL_ISOLATE_TYPES` (from url.h) and `EW_URL_STANDARD_ISOLATE_TYPES` (from url-standard.h) are registered in the runtime. They seem similar, but url.h is not compliant with the WHATWG standard. Will they conflict, or does the latter override the former?

I suspect the problem arises because `EW_URL_STANDARD_ISOLATE_TYPES` overrides `EW_URL_ISOLATE_TYPES`, causing `api::URLSearchParams` to not match the JavaScript URLSearchParams object. As a result, the JavaScript URLSearchParams is coerced into a `kj::String`, leading to the content-type header error. I'm not very sure about that because the codebase is not easy to understand. If there is anything wrong, please correct it for me. :)

src/workerd/api/http.h
```diff
@@ -252,7 +253,7 @@ class Body: public jsg::Object {
   // from any of the other source types, Body can create a new ReadableStream from the source, and
   // the POST will successfully retransmit.
   using Initializer = kj::OneOf<jsg::Ref<ReadableStream>, kj::String, kj::Array<byte>,
-                                jsg::Ref<Blob>, jsg::Ref<URLSearchParams>, jsg::Ref<FormData>>;
+                                jsg::Ref<Blob>, jsg::Ref<url::URLSearchParams>, jsg::Ref<FormData>>;
```
jasnell, line 256: hmm, actually this should be both `jsg::Ref<URLSearchParams>` *and* `jsg::Ref<url::URLSearchParams>`, those are two different implementations (one the original and the second the actually spec-compliant one). They would be handled identically tho with two `KJ_CASE_ONEOF` blocks in the http.c++ above.

balusch: Thank you for your clarification! However, I’m still a bit confused about how they can co-exist. Since both `url.h` and `url-standard.h` register the same name (such as `URLSearchParams`) into the JavaScript runtime, won’t that lead to a conflict?

jasnell: They are controlled with a compatibility flag. If the compat flag is not set, the old original impl (`jsg::Ref<URLSearchParams>`) is used. With the flag set, the new impl (`jsg::Ref<url::URLSearchParams>`) is used. At the c++ level they both coexist but only one is ever exposed to JS in any given worker.

balusch: Thanks! That makes sense to me.

balusch:
Hi, @jasnell. I’ve force-pushed an update to add the missing `jsg::Ref<url::URLSearchParams>` in `Body::Initializer`. I’ve tested both versions with different `compatibilityDate`, and they both worked well. I also added you as a co-author in the commit message and hope you don’t mind.

jasnell (approved):

jasnell:
@balusch ... thank you for the fix! Really appreciated!

jasnell:
Looks like some linting issues with this. I'll fix those up and get this landed.

jasnell:
hmm, actually that passed linting locally. Maybe a CI issue. Let me rerun CI.

fhanau (approved):

fhanau:
CI seems unwell, but looks more like for unrelated network/cache issues... hope those resolve again

jasnell:
ah, ok, I see, there's an issue with our automatic type generation with two URLSearchParam definitions conflicting. We'll need to resolvre that before this can land.

src/workerd/api/url-standard.h
```diff
@@ -95,7 +95,19 @@ class URLSearchParams: public jsg::Object {
     JSG_METHOD(toString);
     JSG_ITERABLE(entries);
 
-    if (flags.getUrlSearchParamsDeleteHasValueArg()) {
+    if (!flags.getSpecCompliantUrl()) {
+      // This is a hack. The spec-compliant URLSearchParams type is used in
+      // the Body constructor, see:
+      // https://github.com/balusch/workerd/tree/main/src/workerd/api/http.h#L255
```
balusch, line 101: Maybe we could specify a specific commit of the official repository instead of using the `main` branch, so that the reference remains stable and is not affected by subsequent commits.

fhanau:
Not to bikeshed some more, but I updated this to fix the comment line length (workerd uses KJ style and thus 100) and use links to the latest tagged release which are shorter.

balusch:
Hi @fhanau Thanks for your update. However, the ci still failed and it seems that there is something wrong with the Bazel remote cache (maybe related to [this issue](https://github.com/bazelbuild/bazel/issues/15596) but not sure). Could you please re-trigger the ci?

fhanau:
> [ `2f58c06`](/cloudflare/workerd/commit/2f58c06fe6476268f3082e6d571a8abb718979a1)

The Windows build appears thoroughly broken unfortunately – I'll try to see on Monday if flushing the cache entirely helps here.

fhanau:
@penalosa any remaining concerns or can I merge this?

jasnell:
@fhanau ... let's do an internal CI run on this just to be extra safe.

fhanau:
Regenerated the types as required by CI now, which revealed the next problem... URLSearchParams is listed twice in the types definition? @penalosa

penalosa:
> Regenerated the types as required by CI now, which revealed the next problem... URLSearchParams is listed twice in the types definition? @penalosa

I _think_ that's fine. It's a bit messy but it shouldn't cause any type errors (since they point to the same definition). I can't really see any easy way around that unless we get a lot more invasive with the C++ definitions based on compat flag

penalosa:
> The Windows build appears thoroughly broken unfortunately – I'll try to see on Monday if flushing the cache entirely helps here. Edit: Now it suddenly seems to be working 🤷

This is a fork, so I don't think the cache is being used, which is probably why the builds are so slow

PR #3154, opened 2024-11-21
Add basic new streaming tail worker types

jasnell:
Largely preparation for the bigger implementation piece. Getting the basic types implemented.

~~I likely won't land this immediately even with sign offs. Want to progress the design a bit more first. But this lays groundwork.~~ This should be ready to land

Internal discussion: https://wiki.cfdata.org/display/~jsnell/Tail+Workers+2%3A+Streaming+Trace (sorry for those of you following along that don't have access to this)

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -154,6 +154,21 @@ TraceId TraceId::fromEntropy(kj::Maybe<kj::EntropySource&> entropySource) {
   return TraceId(getRandom64Bit(entropySource), getRandom64Bit(entropySource));
 }
 
+kj::String SpanId::toGoString() const {
+  kj::Vector<char> s(16);
+  addHex(s, id);
+  s.add('\0');
+  return kj::String(s.releaseAsArray());
+}
+
+SpanId SpanId::fromEntropy(kj::Maybe<kj::EntropySource&> entropySource) {
+  return SpanId(getRandom64Bit(entropySource));
+}
+
+kj::String KJ_STRINGIFY(const SpanId& id) {
+  return id;
+}
+
 kj::String KJ_STRINGIFY(const TraceId& id) {
   return id;
 }
@@ -162,18 +177,18 @@ InvocationSpanContext::InvocationSpanContext(kj::Badge<InvocationSpanContext>,
     kj::Maybe<kj::EntropySource&> entropySource,
     TraceId traceId,
     TraceId invocationId,
-    uint64_t spanId,
+    SpanId spanId,
     kj::Maybe<kj::Rc<InvocationSpanContext>> parentSpanContext)
     : entropySource(entropySource),
       traceId(kj::mv(traceId)),
       invocationId(kj::mv(invocationId)),
-      spanId(spanId),
+      spanId(kj::mv(spanId)),
       parentSpanContext(kj::mv(parentSpanContext)) {}
 
 kj::Rc<InvocationSpanContext> InvocationSpanContext::newChild() {
   KJ_ASSERT(!isTrigger(), "unable to create child spans on this context");
   return kj::rc<InvocationSpanContext>(kj::Badge<InvocationSpanContext>(), entropySource, traceId,
-      invocationId, getRandom64Bit(entropySource), addRefToThis());
+      invocationId, SpanId::fromEntropy(entropySource), addRefToThis());
 }
 
 kj::Rc<InvocationSpanContext> InvocationSpanContext::newForInvocation(
@@ -186,7 +201,8 @@ kj::Rc<InvocationSpanContext> InvocationSpanContext::newForInvocation(
     return ctx->traceId;
   }).orDefault([&] { return TraceId::fromEntropy(entropySource); });
   return kj::rc<InvocationSpanContext>(kj::Badge<InvocationSpanContext>(), entropySource,
-      kj::mv(traceId), TraceId::fromEntropy(entropySource), 0, kj::mv(parent));
+      kj::mv(traceId), TraceId::fromEntropy(entropySource), SpanId::fromEntropy(entropySource),
+      kj::mv(parent));
 }
 
 TraceId TraceId::fromCapnp(rpc::InvocationSpanContext::TraceId::Reader reader) {
@@ -1315,10 +1331,10 @@ kj::Maybe<tracing::TailEvent::Context> getParentContextFromSpan(
 }
 }  // namespace
 
-tracing::TailEvent::Context::Context(TraceId traceId, TraceId invocationId, kj::uint spanId)
-    : traceId(traceId),
-      invocationId(invocationId),
-      spanId(spanId) {}
+tracing::TailEvent::Context::Context(TraceId traceId, TraceId invocationId, SpanId spanId)
+    : traceId(kj::mv(traceId)),
+      invocationId(kj::mv(invocationId)),
+      spanId(kj::mv(spanId)) {}
 
 tracing::TailEvent::Context::Context(rpc::InvocationSpanContext::Reader reader)
     : traceId(TraceId::fromCapnp(reader.getTraceId())),
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -1098,6 +1098,59 @@ tracing::SpanClose tracing::SpanClose::clone() {
 }
 
 namespace {
+kj::Maybe<kj::String> readLabelFromReader(const rpc::Trace::Link::Reader& reader) {
+  if (!reader.hasLabel()) return kj::none;
+  return kj::str(reader.getLabel());
+}
+tracing::TraceId readTraceIdFromReader(const rpc::Trace::Link::Reader& reader) {
+  KJ_ASSERT(reader.hasContext());
+  auto context = reader.getContext();
+  return tracing::TraceId::fromCapnp(context.getTraceId());
+}
+tracing::TraceId readInvocationIdFromReader(const rpc::Trace::Link::Reader& reader) {
+  KJ_ASSERT(reader.hasContext());
+  auto context = reader.getContext();
+  return tracing::TraceId::fromCapnp(context.getInvocationId());
+}
+tracing::SpanId readSpanIdFromReader(const rpc::Trace::Link::Reader& reader) {
+  KJ_ASSERT(reader.hasContext());
+  auto context = reader.getContext();
+  return tracing::SpanId(context.getSpanId());
+}
+}  // namespace
+
+tracing::Link::Link(const InvocationSpanContext& other, kj::Maybe<kj::String> label)
+    : Link(kj::mv(label), other.getTraceId(), other.getInvocationId(), other.getSpanId()) {}
+
+tracing::Link::Link(
+    kj::Maybe<kj::String> label, TraceId traceId, TraceId invocationId, SpanId spanId)
+    : label(kj::mv(label)),
+      traceId(kj::mv(traceId)),
+      invocationId(kj::mv(invocationId)),
+      spanId(kj::mv(spanId)) {}
+
+tracing::Link::Link(rpc::Trace::Link::Reader reader)
+    : label(readLabelFromReader(reader)),
+      traceId(readTraceIdFromReader(reader)),
+      invocationId(readInvocationIdFromReader(reader)),
+      spanId(readSpanIdFromReader(reader)) {}
+
+void tracing::Link::copyTo(rpc::Trace::Link::Builder builder) {
+  KJ_IF_SOME(l, label) {
+    builder.setLabel(l);
+  }
+  auto ctx = builder.initContext();
+  traceId.toCapnp(ctx.initTraceId());
+  invocationId.toCapnp(ctx.initInvocationId());
+  ctx.setSpanId(spanId.getId());
+}
+
+tracing::Link tracing::Link::clone() {
+  return Link(
+      label.map([](kj::String& str) { return kj::str(str); }), traceId, invocationId, spanId);
+}
+
+namespace {
 tracing::Onset::Info getInfoFromReader(const rpc::Trace::Onset::Reader& reader) {
   auto info = reader.getInfo();
   switch (info.which()) {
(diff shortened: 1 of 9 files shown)


src/workerd/io/trace.c++
```diff
@@ -711,6 +783,549 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+        for (size_t n = 0; n < custom.size(); n++) {
+          custom[n].copyTo(customBuilder[n]);
+        }
+      }
+    }
+  }
+}
+
+tracing::SpanOpen tracing::SpanOpen::clone() {
+  static constexpr auto cloneInfo =
```
anonrig, line 1009: I'm not sure if `static constexpr` holds here.

```suggestion
  constexpr auto cloneInfo =
```

src/workerd/io/trace.c++
```diff
@@ -711,6 +783,549 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+            attrs.add(custom[n].clone());
+          }
+          return attrs.releaseAsArray();
+        }
+      }
+      KJ_UNREACHABLE;
+    });
+  };
+  return SpanOpen(
+      operationName.map([](auto& str) { return kj::ConstString(kj::str(str)); }), cloneInfo(info));
```
anonrig, line 1031: Is there a slightly more readable solution than `kj::ConstString(kj::str(str))` where we can add to KJ?

src/workerd/io/trace.c++
```diff
@@ -711,6 +783,549 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+}
+
+tracing::SpanClose::SpanClose(EventOutcome outcome): outcome(outcome) {}
```
anonrig, line 1034: I guess, this can be moved to the header file.

src/workerd/io/trace.c++
```diff
@@ -711,6 +783,549 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+    case rpc::Trace::TailEvent::Event::RETURN:
+      return tracing::Mark(tracing::Return(event.getReturn()));
+    case rpc::Trace::TailEvent::Event::DIAGNOSTIC_CHANNEL_EVENT:
+      return tracing::Mark(tracing::DiagnosticChannelEvent(event.getDiagnosticChannelEvent()));
+    case rpc::Trace::TailEvent::Event::EXCEPTION:
+      return tracing::Mark(tracing::Exception(event.getException()));
+    case rpc::Trace::TailEvent::Event::LOG:
+      return tracing::Mark(tracing::Log(event.getLog()));
+  }
+  KJ_UNREACHABLE;
```
anonrig, line 1231: Can you move this to the default case of the switch?

src/workerd/io/trace.c++
```diff
@@ -711,6 +783,549 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+          KJ_CASE_ONEOF(ret, Return) {
+            return Mark(ret.clone());
+          }
+          KJ_CASE_ONEOF(attr, Attribute) {
+            return Mark(attr.clone());
+          }
+        }
+      }
+    }
+    KJ_UNREACHABLE;
```
anonrig, line 1320: Ditto

anonrig (approved):

src/workerd/io/trace.c++
jasnell: honestly using `kj::ConstString` here at all is likely overkill. I may make this just `kj::String`

src/workerd/io/trace.c++
jasnell: Just Say No The `default:` Case.

src/workerd/io/trace.c++
jasnell: It could but I prefer to keep it with the rest of the impl. Just a style difference

jasnell:
Updated to ...

(a) remove the use of `ConstString` in a couple of places to simplify
(b) Add the `Resume` and `Hibernate` types
(c) Add tests
(d) Fill in the rest of the missing impl bits

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,126 @@ struct Trace @0x8e8d911203762d34 {
...
+  struct Attribute {
+    # An Attribute mark is used to add detail to a span over its lifetime.
+    # The Attribute struct can also be used to provide arbitrary additional
+    # properties for some other structs.
+    # Modeled after https://opentelemetry.io/docs/concepts/signals/traces/#attributes
+    struct Value {
+      inner :union {
+        text @0 :Text;
+        bool @1 :Bool;
+        int @2 :Int32;
```
fhanau, line 173: Getting the types right here is tricky – the tracing approach I had in mind was to support 64-bit integers but that doesn't fit into the double-based JS number type. If integers are limited to 32 bits we'll have to document it but it might just be the cleanest approach.
For float we should have Float64 though? That matches the JS number type.

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,126 @@ struct Trace @0x8e8d911203762d34 {
...
+        float @3 :Float32;
+      }
+    }
+    name @0 :Text;
+    value @1 :List(Value);
```
fhanau, line 178: Having a list here is consistent with OTel, but at odds with what I'm working on in #3180 / out of scope for Automated Tracing as envisioned it so far (there's nothing that would need to add more than one tag value). Maybe useful for custom spans in the future.

jasnell: I think for now let's stick with the otel consistency and allow this to be multi-valued

src/workerd/io/trace.c++
```diff
@@ -714,6 +806,711 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+
+  // There should always be a value and it always have at least one entry in the list.
+  KJ_ASSERT(reader.hasValue());
+  auto value = reader.getValue();
+  KJ_ASSERT(value.size() >= 1);
+  if (value.size() == 1) {
+    return readValue(value[0]);
+  }
+  kj::Vector<tracing::Attribute::Value> values(value.size());
+  for (auto v: value) {
```
anonrig, line 879:
```suggestion
  for (auto v&: value) {
```

src/workerd/io/trace.c++
```diff
@@ -714,6 +806,711 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+    case rpc::Trace::Onset::Info::RESUME: {
+      return tracing::Resume(info.getResume());
+    }
+    case rpc::Trace::Onset::Info::CUSTOM: {
+      return tracing::CustomEventInfo();
+    }
+  }
+  KJ_UNREACHABLE;
+}
+kj::Maybe<kj::String> getScriptNameFromReader(const rpc::Trace::Onset::Reader& reader) {
```
anonrig, line 1138: almost all getters have missing const qualifiers. can you add them before landing?

src/workerd/io/trace.c++
```diff
@@ -714,6 +806,711 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+}
+
+kj::Maybe<tracing::TailEvent::Context> readParentContextFromTailEvent(
+    rpc::Trace::TailEvent::Reader& reader) {
+  if (!reader.hasParentContext()) return kj::none;
+  return tracing::TailEvent::Context(reader.getParentContext());
+}
+
+tracing::TailEvent::Event readEventFromTailEvent(rpc::Trace::TailEvent::Reader& reader) {
+  auto event = reader.getEvent();
```
anonrig, line 1387:
```suggestion
  const auto event = reader.getEvent();
```

src/workerd/io/trace.h
```diff
@@ -397,6 +480,31 @@ class Exception final {
   kj::Maybe<kj::String> stack;
 
   void copyTo(rpc::Trace::Exception::Builder builder);
+  Exception clone();
+};
+
+struct Resume final {
```
anonrig, line 486: Can you add a documentation/comment to all newly added struct, explaining their purpose and what they represent?

anonrig (approved):

src/workerd/io/trace.c++
jasnell: `value` is a capnp Reader, which means the `auto v: value` is correct.

src/workerd/io/trace.c++
jasnell: These aren't getters. They are statics.

anonrig: It's hard to distinguish when the static function names doesn't start with capital letter (like node.js does)...

anonrig (approved):

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,125 @@ struct Trace @0x8e8d911203762d34 {
...
+        bool @1 :Bool;
+        int @2 :Int32;
+        float @3 :Float64;
+      }
+    }
+    name @0 :Text;
+    value @1 :List(Value);
+  }
+
+  struct Return {
```
fhanau, line 181: Are return events supposed to be issued for every span? A lot of the time there won't be any meaningful time between the return and the span close right?

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,125 @@ struct Trace @0x8e8d911203762d34 {
...
+      fetch @3 :FetchEventInfo;
+      jsrpc @4 :JsRpcEventInfo;
+    }
+  }
+
+  struct SpanClose {
+    # Marks the closing of a child span within the streaming tail session.
+    # Once emitted, no further mark events should occur within the closed
+    # span.
+    outcome @0 :EventOutcome;
```
fhanau, line 210: `EventOutcome` as defined in outcome.capnp is more for entire worker invocations right? Or will only a subset of the values for that be used here?

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,125 @@ struct Trace @0x8e8d911203762d34 {
...
+  }
+
+  struct Resume {
+    # A resume event indicates that we are resuming a previously hibernated
+    # tail session.
+    attachment @0 :Data;
```
fhanau, line 216: What is the attachment here for?

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,125 @@ struct Trace @0x8e8d911203762d34 {
...
+
+  struct TailEvent {
+    # A streaming tail worker receives a series of Tail Events. Tail events always
+    # occur within an InvocationSpanContext. The first TailEvent delivered to a
+    # streaming tail session is always an Onset. The final TailEvent delivered is
+    # always an Outcome. Between those can be any number of SpanOpen, SpanClose,
+    # and Mark events. Every SpanOpen *must* be associated with a SpanClose unless
+    # the stream was abruptly terminated.
+    context @0 :InvocationSpanContext;
+    parentContext @1 :InvocationSpanContext;
```
fhanau, line 263: Having two contexts feel redundant... I'm sure you've already considered that but is passing just the parent span ID not sufficent here?

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,125 @@ struct Trace @0x8e8d911203762d34 {
...
+    timestampNs @2 :Int64;
+    sequence @3 :UInt32;
```
fhanau, line 265: Sequence is a simple counter I assume?

src/workerd/io/worker-interface.capnp
```diff
@@ -157,6 +157,125 @@ struct Trace @0x8e8d911203762d34 {
...
+    event :union {
+      onset @4 :Onset;
+      outcome @5 :Outcome;
+      hibernate @6 :Hibernate;
+      spanOpen @7 :SpanOpen;
+      spanClose @8 :SpanClose;
+      attribute @9 :Attribute;
```
fhanau, line 272: Can a single attribute be an event? From the I assumed that attributes would always have an associated span and all attributes of a span would be reported together in the Return event.

src/workerd/io/worker-interface.capnp
jasnell: No, not for every span, only for cases where it makes sense. For instance, with a fetch subrequest like the following... the span open will correspond with the fetch request itself, the return marks the moment the fetch response was received, but the span close won't come until after the full response payload is read. Other types of spans may not have any kind of meaningful return and therefore wouldn't ever have a `Return` mark.

src/workerd/io/worker-interface.capnp
jasnell: Yep. I've been debating whether to define a more constrained enum for this. For a child span the only values that are meaningful here are: `UNKNOWN`, `OK`, `EXCEPTION`, and `CANCELED`

src/workerd/io/worker-interface.capnp
jasnell: Similar to hibernatable web sockets, when a tail span is put into hibernation, the tail worker will have the opportunity to store a small amount of v8 serialized data to be persisted with the hibernation state. That will be passed back into the tail worker when the tail stream is resumed.

src/workerd/io/worker-interface.capnp
jasnell: The triggering event may be coming from a completely different trace / invocation. I'm looking to see if we can remodel this bit as a `Link` tho.

src/workerd/io/worker-interface.capnp
jasnell: Yep, simple monotonic counter.

src/workerd/io/worker-interface.capnp
jasnell: Setting a span attribute will be an event, yes. They have nothing to do with the `Return` event. All `Attribute` events received within a span are attributes of that span.

fhanau: Having every tag be a separate event feels rather inefficient – if a span has say 3 - 6 small tags (or attributes in OTel terminology), which will apply for several user tracing span types, that means having many more events to be sent even though the tags are mostly being set all at the same time.
For custom spans it is conceivable that sometimes tags will be created at different points in time within the span, and that the tail worker relies on getting that tag information before the span is complete (this is kind of at odds with how I think span tags should be used but it's plausible).
To avoid the overhead, would adding an attributes list to the SpanEnd event make sense so that for user tracing tags are part of that event and not distinct events? When we support custom spans in the future we can add support for manually adding tags that will be their own events.

jasnell: Yep, for this case I've been considering making it so that the `Attribute` event, multiple attributes can actually be provided at once.

src/workerd/io/worker-interface.capnp
jasnell: Refactored this.

src/workerd/io/worker-interface.capnp
jasnell: For now, going to keep this using `EventOutcome`.

src/workerd/io/trace.c++
```diff
@@ -714,6 +816,763 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+    auto inner = reader.getInner();
+    switch (inner.which()) {
+      case rpc::Trace::Attribute::Value::Inner::TEXT: {
+        return kj::str(inner.getText());
+      }
+      case rpc::Trace::Attribute::Value::Inner::BOOL: {
+        return inner.getBool();
+      }
+      case rpc::Trace::Attribute::Value::Inner::FLOAT: {
+        return inner.getFloat();
```
fhanau, line 875: This should be getFloat64 after the type changed to float64.

src/workerd/io/trace.c++
```diff
@@ -714,6 +816,763 @@ tracing::Exception::Exception(rpc::Trace::Exception::Reader reader)
...
+  static auto writeValue = [](auto builder, const auto& value) mutable {
+    KJ_SWITCH_ONEOF(value) {
+      KJ_CASE_ONEOF(str, kj::String) {
+        builder.initInner().setText(str.asPtr());
+      }
+      KJ_CASE_ONEOF(b, bool) {
+        builder.initInner().setBool(b);
+      }
+      KJ_CASE_ONEOF(f, double) {
+        builder.initInner().setFloat(f);
```
fhanau, line 909: Same here, use setFloat64

src/workerd/io/trace.c++
jasnell: The `getFloat` is based on the name of the field, not the type. The definition is `float @n : Float64`, therefore `inner.getFloat()` is correct.

fhanau: Ok, my mistake, I saw getFloat64 being used internally.

fhanau (approved):

PR #3156, opened 2024-11-22
add PassthroughTraceWorker definitions

anonrig:
Adds the passthrough trace worker `.capnp` definitions


jasnell:
I'm going to hold off signing off on this until I can evaluate any overlap that may exist in the interfaces that still need to be defined for the streaming tail worker API.

PR #3177, opened 2024-11-26
[nfc] Refactor and add trace span RPC code

fhanau:
This will be used to better export user tracing spans in a follow-up PR.

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -1729,25 +1729,59 @@ void WorkerTracer::addSpan(const Span& span, kj::String spanContext) {
   // TODO(cleanup): Create a function in kj::OneOf to automatically convert to a given type (i.e
   // String) to avoid having to handle each type explicitly here.
   for (const Span::TagMap::Entry& tag: span.tags) {
-    auto value = [&]() {
-      KJ_SWITCH_ONEOF(tag.value) {
-        KJ_CASE_ONEOF(str, kj::String) {
-          return kj::str(str);
-        }
-        KJ_CASE_ONEOF(val, int64_t) {
-          return kj::str(val);
-        }
-        KJ_CASE_ONEOF(val, double) {
-          return kj::str(val);
-        }
-        KJ_CASE_ONEOF(val, bool) {
-          return kj::str(val);
-        }
-      }
-      KJ_UNREACHABLE;
-    }();
-    kj::String message = kj::str("[\"tag: "_kj, tag.key, " => "_kj, value, "\"]");
+    kj::String message = kj::str("[\"tag: "_kj, tag.key, " => "_kj, spanTagStr(tag.value), "\"]");
     addLog(span.endTime, LogLevel::LOG, kj::mv(message), true);
+  }
+}
+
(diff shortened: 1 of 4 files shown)


src/workerd/io/trace.capnp
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2017-2022 Cloudflare, Inc.
+# Licensed under the Apache 2.0 license found in the LICENSE file or at:
+#     https://opensource.org/licenses/Apache-2.0
+
+@0xc40f73be329a38d9;
+
+using Cxx = import "/capnp/c++.capnp";
+$Cxx.namespace("workerd::rpc");
+
+# This file only contains trace helper structures – due to C++ include size constraints we should
```
fhanau, line 10: This was a surprising find – I wanted to have `Trace` itself in here but it turns out that having any kind of capnp interface comes with a significant compile time overhead. This would have increased the compile time of source files using a commonly used downstream capnp file that includes this significantly, undoing some of the progress made through previous include cleanups.

src/workerd/io/trace.c++
```diff
@@ -867,28 +867,60 @@ void WorkerTracer::addSpan(const Span& span, kj::String spanContext) {
...
+    }
+    KJ_CASE_ONEOF(val, bool) {
+      return kj::str(val);
+    }
+  }
+  KJ_UNREACHABLE;
+}
+
+using RpcValue = rpc::TagValue;
+void serializeTagValue(RpcValue::Builder builder, const Span::TagValue& value) {
```
fhanau, line 894: The serialize/deserialize functions will be used in a follow-up, having them here instead of downstream avoids duplication once they are being used here.

src/workerd/io/trace.c++
```diff
@@ -867,28 +867,60 @@ void WorkerTracer::addSpan(const Span& span, kj::String spanContext) {
...
+  switch (value.which()) {
+    case RpcValue::BOOL:
+      return value.getBool();
+    case RpcValue::FLOAT64:
+      return value.getFloat64();
+    case RpcValue::INT64:
+      return value.getInt64();
+    case RpcValue::STRING:
+      return kj::heapString(value.getString());
+  }
```
anonrig, line 921: Can you add a default kj unreachable here?

src/workerd/io/trace.capnp
anonrig: Would you mind expanding the comment here to explain what those size constraints are?

fhanau: Absolutely, this is a quite subtle issue so it deserves some more detail.
`.capnp.h` includes are a major driver of slow builds for us – https://github.com/capnproto/capnproto/pull/2188 explores this in-depth and partially mitigates it. This is noticeable especially when compiling code that directly or indirectly depends on many or large capnp structures – `Trace` as defined in worker-interface.capnp is somewhat large and includes capnp interfaces, which cause a lot of code to be generated. This currently causes worker-interface.capnp to have a relatively large include overhead – it's worth paying attention to where Trace is defined to avoid having too many source files needing to transitively include it.
I expanded the comment a bit, hopefully this is more clear now.

fhanau:
@mikea PTAL

mikea (approved):

PR #3180, opened 2024-11-27
[o11y] Provide OTel-compatible user span representation in trace

fhanau:
#3177 should land before this (listed as the first comment here). Also see the corresponding upstream PR.

### DIFF
--- a/src/workerd/io/trace.h
+++ b/src/workerd/io/trace.h
@@ -778,6 +778,7 @@ enum class PipelineLogLevel {
 };
 
 struct Span;
+struct CompleteSpan;
 
 // TODO(someday): See if we can merge similar code concepts...  Trace fills a role similar to
 // MetricsCollector::Reporter::StageEvent, and Tracer fills a role similar to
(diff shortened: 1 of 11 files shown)


fhanau:
This should be ready for review now – I'm not 100% certain about the span representation but we can still change it while this is marked as experimental. Making trace spans available is covered in a follow-up PR (#3209), I fear that it will require some more review so not including it in this one.

src/workerd/io/trace.h
```diff
@@ -708,6 +706,33 @@ struct Span {
...
+// Utility functions for handling span tags.
+void serializeTagValue(rpc::TagValue::Builder builder, const Span::TagValue& value);
+Span::TagValue deserializeTagValue(rpc::TagValue::Reader value);
+
+// Stringify and clone for span tags, getting this to work with KJ_STRINGIFY() appears exceedingly
+// difficult.
+kj::String spanTagStr(const Span::TagValue& tag);
+Span::TagValue spanTagClone(const Span::TagValue& tag);
+
+struct CompleteSpan {
```
fhanau, line 718: Not sure if `CompleteSpan` is descriptive enough. Should this be `UserSubmitSpan` or something else? It's unfortunate that we'll need this on top of `Span`, but this needs to store more information than Span has.

fhanau:
- [ ] Update experimental types (`types/definitions/experimental/index.d.ts`)

fhanau:
Hoping to get reviews on this soon – this will be useful for doing another demo on user tracing using the new interchange format.

jasnell (approved):

PR #3183, opened 2024-11-27
implement node:dns module for nodejs_compat

anonrig:
Implements all functions in `node:dns` using Cloudflare DNS (1.1.1.1)


src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,94 @@
...
+export async function resolveTxt(hostname: string): Promise<string[][]> {
+  validateString(hostname, 'hostname');
+  const json = await sendDnsRequest(hostname, 'TXT');
+  assert(
+    json.Answer,
+    `DNS request should contain an "Answer" attribute but it didn't.`
+  );
+  // Each entry has quotation marks as a prefix and suffix.
+  // Node.js APIs doesn't have them.
+  return json.Answer.map((value) => [value.data.replaceAll('"', '')]);
```
vicb, line 75: would value.data.slice(1, -1) be faster ?

anonrig: Nice catch. Yes, definitely!

src/node/internal/internal_dns_client.ts
```diff
@@ -0,0 +1,65 @@
...
+export async function sendDnsRequest(
+  name: string,
+  type: string
+): Promise<SuccessResponse> {
+  // We are using cloudflare-dns.com and not 1.1.1.1 because of certificate issues.
+  // TODO(soon): Replace this when KJ certificate issues are resolved.
+  const server = new URL('https://cloudflare-dns.com/dns-query');
+  server.searchParams.set('name', name);
+  server.searchParams.set('type', type);
+  const response = await fetch(server, {
```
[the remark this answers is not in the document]
anonrig: Yes, also I need to find a way to distinguish found/not found DNS records.

```
❯ curl --header "accept: application/dns-json" "https://cloudflare-dns.com/dns-query?name=google.com2222&type=TXT" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   246  100   246    0     0   5104      0 --:--:-- --:--:-- --:--:--  5020
{
  "Status": 3,
  "TC": false,
  "RD": true,
  "RA": true,
  "AD": true,
  "CD": false,
  "Question": [
    {
      "name": "google.com2222",
      "type": 16
    }
  ],
  "Authority": [
    {
      "name": "",
      "type": 6,
      "TTL": 86400,
      "data": "a.root-servers.net. nstld.verisign-grs.com. 2024112701 1800 900 604800 86400"
    }
  ]
}
```

src/workerd/api/node/tests/dns-nodejs-test.js
```diff
@@ -37,3 +36,41 @@ export const errorCodesExist = {
...
+    // An idiomatic test that avoids checking for a specific value
+    function testResults(results) {
+      ok(Array.isArray(results));
+      ok(results.length > 0);
+      ok(
+        results.find((r) => r.find((v) => v.includes('spf1'))),
+        'One of the entries should have spf1'
+      );
+      ok(
+        results.at(0).at(0)[0] !== '"',
```
vicb, line 51:
```suggestion
        results[0][0][0] !== '"',
```

src/workerd/api/node/tests/dns-nodejs-test.js
```diff
@@ -37,3 +36,41 @@ export const errorCodesExist = {
...
+      const { promise, resolve, reject } = Promise.withResolvers();
+      dns.resolveTxt('cloudflare.com', (error, results) => {
+        if (error) {
+          reject(error);
+          return;
+        }
+        testResults(results);
+        resolve();
+      });
+      await promise;
```
vicb, line 73:
```suggestion
      // Callback API
      dns.resolveTxt('cloudflare.com', (error, results) => {
        if (error) {
          return fail(error);
        }
        testResults(results);
      });
```

vicb (commented):
Added a few inline comments

src/node/dns.ts
```diff
@@ -1,30 +1,70 @@
...
+  resolveNs,
+  resolvePtr,
+  resolveSoa,
+  resolveSrv,
+  setDefaultResultOrder,
+  getDefaultResultOrder,
+  setServers,
+} from 'node-internal:internal_dns';
+
+function toCallback<A = unknown, ReturnValue = unknown>(
```
jasnell, line 46: You know we implement `callbackify`? You can move its implementation into `node/internal` and reuse that here.

src/node/dns.ts
```diff
@@ -1,30 +1,70 @@
...
-export const resolveSrv = dnsUtil.resolveSrv.bind(dnsUtil);
-export const resolveTxt = dnsUtil.resolveTxt.bind(dnsUtil);
-export const reverse = dnsUtil.reverse.bind(dnsUtil);
-export const setDefaultResultOrder =
-  dnsUtil.setDefaultResultOrder.bind(dnsUtil);
-export const getDefaultResultOrder =
-  dnsUtil.getDefaultResultOrder.bind(dnsUtil);
-export const setServers = dnsUtil.setServers.bind(dnsUtil);
+export const reverse = toCallback(dns.reverse.bind(dns));
+export const resolveTxt = toCallback(dns.resolveTxt.bind(dns));
```
jasnell, line 67: Since you're going from promises to callbacks here, how are you handling synchronously throwing on argument validation? By default the JSG wrapper should cause synchronously thrown errors at the c++ level to be returned into promise rejections. Are you certain this will do the right thing for the errors the Node.js API expects to be thrown synchronously?

src/node/dns.ts
```diff
@@ -1,30 +1,70 @@
...
+
+function toCallback<A = unknown, ReturnValue = unknown>(
+  fn: (arg: A) => Promise<ReturnValue>
+): (
+  input: A,
+  callback: (error: Error | null, value?: ReturnValue) => void
+) => void {
+  return (
+    input: A,
+    callback: (error: Error | null, value?: ReturnValue) => void
```
jasnell, line 54: Node.js will throw an error on these if `callback` is not a function, correct?

src/node/internal/internal_dns_client.ts
```diff
@@ -0,0 +1,82 @@
...
+  Additional?: Answer;
+}
+
+export async function sendDnsRequest(
+  name: string,
+  type: string
+): Promise<SuccessResponse> {
+  // We are using cloudflare-dns.com and not 1.1.1.1 because of certificate issues.
+  // TODO(soon): Replace this when KJ certificate issues are resolved.
+  const server = new URL('https://cloudflare-dns.com/dns-query');
```
jasnell, line 52: Since this will route these requests as public internet requests I guess we should decide if using global fetch is really what we want vs. using a private fetch via a binding that would for sure route the request through internal infrastructure.

src/node/internal/internal_dns_client.ts
```diff
@@ -0,0 +1,82 @@
...
+    json = await response.json();
+  } catch {
+    throw new DnsError(name, errorCodes.BADQUERY, syscall);
+  }
+
+  if ('error' in json) {
+    throw new DnsError(name, errorCodes.BADRESP, syscall);
+  }
+
+  if (json.Answer?.at(0).name === '') {
```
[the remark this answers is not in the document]
anonrig: On typescript `[0]` doesn't return `undefined | ValueType` but `.at()` does. I think we should encourage it more. And we should also enable `noUncheckedIndexedAccess` on tsconfig as well (ref: https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess)

src/node/internal/internal_errors.ts
```diff
@@ -587,6 +587,29 @@ export class ERR_INVALID_URI extends NodeError {
...
+// Error: queryTxt ENOTFOUND google.com2
+//     at QueryReqWrap.onresolve [as oncomplete] (node:internal/dns/callback_resolver:45:19)
+//     at QueryReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
+//   errno: undefined,
+//   code: 'ENOTFOUND',
+//   syscall: 'queryTxt',
+//   hostname: 'google.com2'
+// }
+export class DnsError extends NodeError {
+  errno: undefined = undefined;
```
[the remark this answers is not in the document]
anonrig: Yes

src/node/dns.ts
anonrig: TIL. Thanks

src/node/dns.ts
anonrig: I'll remove `async` from functions and just return a promise from the function, which would throw validation errors synchronously but the function will return Promise

src/node/dns.ts
anonrig: Yes, definitely there is a `validateFunction` missing from here.

src/node/internal/internal_dns_client.ts
```diff
@@ -0,0 +1,82 @@
...
+  const server = new URL('https://cloudflare-dns.com/dns-query');
+  server.searchParams.set('name', name);
+  server.searchParams.set('type', type);
+
+  // syscall needs to be in format of `queryTxt`
+  const syscall = `query${type[0]!.toUpperCase()}${type.slice(1)}`;
+
+  let json: SuccessResponse | FailedResponse;
+  try {
+    const response = await fetch(server, {
```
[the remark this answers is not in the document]
anonrig: Is there an internal API to remove this sub request limit from this call?

src/node/internal/internal_dns_client.ts
anonrig: Yes, for sure, we should find a solution to this before landing this pull-request.

src/node/internal/internal_dns_client.ts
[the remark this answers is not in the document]
vicb: > On typescript [0] doesn't return undefined | ValueType but .at() does. 

Did you mean to say "does not"?

If that's the case TS is lying to you:

image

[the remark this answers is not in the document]
vicb: TS doesn't lie on my machine

image

src/workerd/api/node/tests/dns-nodejs-test.js
vicb: See [the previous comment](https://github.com/cloudflare/workerd/pull/3183/files/19f1629ff3537d5b115a4f51c7196a63fc9a90ad#r1864151200)

vicb: So I guess it should be `results[0]?.[0]?.[0]`

src/node/internal/internal_dns_client.ts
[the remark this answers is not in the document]
vicb: Both return `undefined | T`

src/workerd/api/node/tests/dns-nodejs-test.js
anonrig: @jasnell Does this hold, or do we need Promise.withResolvers() to wait for the callback response?

src/node/internal/internal_dns_client.ts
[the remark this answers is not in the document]
anonrig: TS playground disagrees: https://www.typescriptlang.org/play/?target=99#code/MYewdgzgLgBAllApgWwgLhtATnMBzAbQF0YBeGAgcgAtEAbOkSgGhkoHcQs6ATSogFAC4AMxgAKBCggEAzEQIAGEqXIAiamoCUMAN4CYMUJBB1EAOkZ5xaqNTgQYPEIghhKsEQEM4dbQG4BAF8hUQkpVHMvKHFZLQB+KJjFLR19Q2MIUwsrGzsHJxc3DxhvX21g0LFJJEjo2K0k8RS0gyNwLLNLEGtbe0dnV3dYACMAV18eCqCgA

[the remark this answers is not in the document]
vicb: Oh 🙀
TIL!
Still you should (even more) add the "?" here and below!

src/node/dns.ts
```diff
@@ -1,30 +1,54 @@
...
-export const reverse = dnsUtil.reverse.bind(dnsUtil);
-export const setDefaultResultOrder =
-  dnsUtil.setDefaultResultOrder.bind(dnsUtil);
-export const getDefaultResultOrder =
-  dnsUtil.getDefaultResultOrder.bind(dnsUtil);
-export const setServers = dnsUtil.setServers.bind(dnsUtil);
+export const reverse = callbackify(dns.reverse.bind(dns));
+export const resolveTxt = callbackify(dns.resolveTxt.bind(dns));
+export const resolveCaa = callbackify(dns.resolveCaa.bind(dns));
+export const resolveMx = callbackify(dns.resolveMx.bind(dns));
```
jasnell, line 51: Need to verify that this does the right thing for argument validation.

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,144 @@
...
+  throw new Error('Not implemented');
+}
+
+export function resolveCaa(
+  name: string
+): Promise<
+  { critical: number; issue?: string; iodef?: string; issuewild?: string }[]
+> {
+  validateString(name, 'name');
+  return sendDnsRequest(name, 'CAA').then((json) => {
```
jasnell, line 50: Why not async await syntax?

src/node/internal/internal_dns_client.ts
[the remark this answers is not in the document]
jasnell: Not using global fetch. If we were using a binding or exposing the fetcher a different way then yes. But honestly I think dns calls *should* be subject to the limit.

src/workerd/api/node/tests/dns-nodejs-test.js
jasnell: The use of `Promise.withResolvers()` looks correct here.

src/node/internal/internal_dns.ts
anonrig: Because `validateString()` should throw synchronously whereas the DNS request errors should be async errors.

jasnell: A comment along those lines would be helpful

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,264 @@
...
+  return sendDnsRequest(name, 'CNAME').then((json) => {
+    if (!('Answer' in json)) {
+      // DNS request should contain an "Answer" attribute, but it didn't.
+      throw new DnsError(name, errorCodes.NODATA, 'queryCname');
+    }
+
+    return json.Answer.map((a) => {
+      // Cloudflare DNS returns "nodejs.org." whereas
+      // Node.js returns "nodejs.org" as a CNAME data.
+      return a.data.slice(0, -1);
```
jasnell, line 54: To be extra safe here (in case the CF DNS behavior ever changed unexpectedly), it might be worth going ahead and checking to see if the answer endsWith "."

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,264 @@
...
+export function resolveCname(name: string): Promise<string[]> {
+  validateString(name, 'name');
+
+  // Validation errors needs to be sync.
+  // Return a promise rather than using async qualifier.
+  return sendDnsRequest(name, 'CNAME').then((json) => {
+    if (!('Answer' in json)) {
+      // DNS request should contain an "Answer" attribute, but it didn't.
+      throw new DnsError(name, errorCodes.NODATA, 'queryCname');
+    }
```
jasnell, line 49: To reduce code duplication, consider moving this check into a reusable utility function.

```js
function validateAnswer(json, name, op) {
  if (!('Answer' in json)) {
    throw new DnsError(name, errorCodes.NODATA, op);
  }
}
```

```
validateAnswer(json, name, 'queryCname');
validateAnswer(json, name, 'queryCaa');
// ... etc
```

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,264 @@
...
+      return a.data.slice(0, -1);
+    });
+  });
+}
+
+export function resolveCaa(
+  name: string
+): Promise<
+  { critical: number; issue?: string; iodef?: string; issuewild?: string }[]
+> {
```
jasnell, line 63: Nit: Might be a bit nicer to define a type for this.

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    // Cloudflare API returns "data": "10 smtp.google.com." hence
+    // we need to parse it.
+    return json.Answer.map((a) => {
+      const [priority, value]: string[] = a.data.split(' ');
+      if (priority == null || value == null) {
+        throw new DnsError(name, errorCodes.BADRESP, 'queryMx');
+      }
+      return {
+        exchange: value.slice(0, -1),
```
jasnell, line 108: I assume this is also because the value may end in a "."? If so, a comment here would be helpful along with a double check to make sure the value ends with the "." before slicing it.

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,253 @@
...
+  throw new Error('Not implemented');
+}
+
+export function resolve4(
+  name: string,
+  options?: { ttl?: boolean }
+): Promise<(string | TTLResponse)[]> {
+  validateString(name, 'name');
+  if (options != null) {
+    validateObject(options, 'options');
```
jasnell, line 55: I notice that Node.js does not actually enforce this. If `options` is anything other than an object Node.js appears to ignore it.

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (options.ttl != null) {
+      validateBoolean(options.ttl, 'options.ttl');
```
jasnell, line 57: Likewise here. Node.js appears to handle any truthy value for `options.ttl`

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,249 @@
...
+  name: string,
+  options?: { ttl?: boolean }
+): Promise<(string | TTLResponse)[]> {
+  validateString(name, 'name');
+
+  // The following change is done to comply with Node.js behavior
+  let ttl = false;
+  if (options != null && typeof options === 'object') {
+    ttl = Boolean(options.ttl);
+  }
```
jasnell, line 55: Should be able to simplify to just....

```suggestion
  const ttl = !!options?.ttl;
```

jasnell:
I think the only key remaining question here is whether this should be using a global fetch to the cloudflare dns endpoint or whether it should be using an internal route to DNS in production. 

@kentonv thoughts?

src/node/internal/internal_dns.ts
anonrig: I have how easy JS enables you to write safe JS code that looks unsafe.

src/node/internal/internal_errors.ts
```diff
@@ -587,6 +587,28 @@ export class ERR_INVALID_URI extends NodeError {
...
+  errno = undefined;
+  syscall: string;
+  hostname: string;
+
+  constructor(hostname: string, code: string, syscall: string) {
+    super(code, `${syscall} ${code} ${hostname}`);
+    this.syscall = syscall;
+    this.hostname = hostname;
+  }
+}
```
vicb, line 610:
```suggestion
  constructor(public hostname: string, code: string, public syscall: string) {
    super(code, `${syscall} ${code} ${hostname}`);
  }
}
```

src/node/internal/internal_utils.ts
```diff
@@ -49,11 +50,13 @@ export function slowCases(enc: string): Encoding | undefined {
     case 4:
       if (enc === 'UTF8') return UTF8;
       if (enc === 'ucs2' || enc === 'UCS2') return UTF16LE;
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-template-expression
       enc = `${enc}`.toLowerCase();
```
vicb, line 54: Why is this needed? enc is supposed to be a string.

Also `ucs2` is repeated l52 and l56

src/node/internal/internal_utils.ts
```diff
@@ -49,11 +50,13 @@ export function slowCases(enc: string): Encoding | undefined {
...
       if (enc === 'utf8') return UTF8;
       if (enc === 'ucs2') return UTF16LE;
       break;
     case 3:
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-template-expression
       if (enc === 'hex' || enc === 'HEX' || `${enc}`.toLowerCase() === 'hex') {
```
vicb, line 60: `${enc}.toLowerCase() === 'hex'` would only ever be usefull if enc is not a string. 
Maybe the type of enc needs to be fixed on that function?

src/node/internal/internal_utils.ts
```diff
@@ -200,3 +208,167 @@ export function getOwnNonIndexProperties(
...
   return result;
 }
+
+function callbackifyOnRejected(
+  reason: unknown,
+  cb: (error?: unknown) => void
+): void {
+  if (!reason) {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    cb(new ERR_FALSY_VALUE_REJECTION(`${reason}`));
```
vicb, line 218: would `String(reason)` allow removing the comment?

src/node/internal/internal_utils.ts
```diff
@@ -200,3 +208,167 @@ export function getOwnNonIndexProperties(
...
+    );
+  }
+
+  const descriptors = Object.getOwnPropertyDescriptors(original);
+  if (typeof descriptors['length']?.value === 'number') {
+    descriptors['length'].value++;
+  }
+  if (typeof descriptors['name']?.value === 'string') {
+    descriptors['name'].value += 'Callbackified';
+  }
```
vicb, line 365:
```suggestion
  if (typeof descriptors.length?.value === 'number') {
    descriptors.length.value++;
  }
  if (typeof descriptors.name?.value === 'string') {
    descriptors.name.value += 'Callbackified';
  }
```

src/rust/dns/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+}
+
+/// Parses an unknown RR format returned from Cloudflare DNS.
+/// Specification is available at
+/// `<https://datatracker.ietf.org/doc/html/rfc3597>`
+///
+/// The format of the record is as follows:
+/// \# 37 15 b3 08 ae 01 73 0a 6d 79 2d 73 65 72 76 69 63 65 06 72 65 67 65 78 70 0b 72 65 70 6c 61 63 65 6d 65 6e 74 00
+///       |--|  |--|  |  |  |  |--------------------------|  |  |--------------|  |  |--------------------------------|
```
vicb, line 89: Nice :)

Is `- Replacement length` wrongly positioned?

vicb (commented):
Great PR!
A few minor comments

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,243 @@
...
+        return {
+          ttl: a.TTL,
+          address: a.data,
+        };
+      }
+
+      return a.data;
+    });
+  });
+}
```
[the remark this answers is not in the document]
anonrig: Performance wise this doesn't make a difference, but looks nicer, I agree.

src/node/internal/internal_dns_client.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  return json;
+}
+
+export function validateAnswer(
+  answer: unknown,
+  name: string,
+  query: string
+): asserts answer is Answer[] {
+  if (answer == null) {
```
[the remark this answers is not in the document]
anonrig: This will be slower, and hence we know that if it exists, it will be an array, we can just check for `null` or `undefined` value.

src/node/internal/internal_dns_client.ts
```diff
@@ -0,0 +1,243 @@
...
+  };
+}
+
+export function normalizeCname({ data }: Answer): string {
+  // Cloudflare DNS returns "nodejs.org." whereas
+  // Node.js returns "nodejs.org" as a CNAME data.
+  if (data.endsWith('.')) {
+    return data.slice(0, -1);
+  }
+  return data;
```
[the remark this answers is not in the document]
anonrig: There is no behavioral/performance impact of the recommendation, and I personally think it reduces the readability.

src/node/internal/internal_dns_client.ts
```diff
@@ -0,0 +1,243 @@
...
+};
+export function normalizeCaa({ data }: Answer): CAA {
+  // CAA API returns "hex", so we need to convert it to UTF-8
+  const obj = { critical: 0 };
+  const record = dnsUtil.parseCaaRecord(data);
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  obj[record.field] = record.value;
+  obj.critical = record.critical;
+  return obj;
```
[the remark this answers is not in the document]
anonrig: This would be better:

```
const record = dnsUtil.parseCaaRecord(data);
  const obj: CAA = { critical: record.critical };
  obj[record.field] = record.value;
  return obj;
  ```

File: src/node/internal/internal_dns_client.ts [hunk 31]
```diff
@@ -0,0 +1,243 @@
...
+  order: number;
+  preference: number;
+};
+export function normalizeNaptr({ data }: Answer): NAPTR {
+  // Cloudflare DNS appends "." at the end whereas Node.js doesn't.
+  return dnsUtil.parseNaptrRecord(data);
+}
+
+export function normalizePtr({ data }: Answer): string {
+  if (data.endsWith('.')) {
```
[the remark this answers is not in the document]
anonrig: I'll be using these functions in resolveAny method, where the separation of these functions make it much more readable.

src/node/internal/internal_utils.ts
anonrig: Yeah, this is better, but we should definitely add an eslint rule that forbids this. I'll apply this change but there is nothing blocking others from reverting this.

src/node/internal/internal_utils.ts
anonrig: yes. i personally think we should just remove this linting rule in the future.

kentonv:
> I think the only key remaining question here is whether this should be using a global fetch to the cloudflare dns endpoint or whether it should be using an internal route to DNS in production.

It's important that these requests do not bypass egress workers, since that could be a security problem if the egress worker is intentionally blocking communications with the internet. If we had a DNS handler API, then we could let the egress worker implement that and proxy DNS requests explicitly -- but that would be a big project. So the only reasonable alternative right now is to go through global fetch() or global connect(). But we don't actually have a connect handler API yet either... so in that case requests would just fail in the presence of an egress worker. So global fetch() seems like the best option.

> Also, since this is a subrequest to a CF property, should the subrequest propagate tracing metadata, control header, etc.

I think that's a more general question -- should global fetch() to a CF property propagate tracing metadata etc.? I don't think that's something we need to answer in the context of this PR specifically.

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,243 @@
...
+  throw new Error('Not implemented');
+}
+
+export function resolve4(
+  name: string,
+  options?: { ttl?: boolean }
+): Promise<(string | TTLResponse)[]> {
+  validateString(name, 'name');
+
+  // The following change is done to comply with Node.js behavior
```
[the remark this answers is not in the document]
anonrig: `!!options?.ttl` because node.js omits any non-object value and doesn't validate if TTL is a boolean

src/rust/dns/lib.rs
anonrig: Yes! Nice catch.

src/rust/dns/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+    extern "Rust" {
+        fn parse_caa_record(record: &str) -> CaaRecord;
+        fn parse_naptr_record(record: &str) -> NaptrRecord;
+    }
+}
+
+/// Given a vector of strings, converts each slice to UTF-8 from HEX.
+///
+/// # Panics
```
mikea, line 26: We shouldn't panic. Especially during parsing. Let's replace this with errors. I suggest using `thiserror` - very easy to use.

src/rust/dns/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    struct NaptrRecord {
+        flags: String,
+        service: String,
+        regexp: String,
+        replacement: String,
+        order: u32,
+        preference: u32,
+    }
+    extern "Rust" {
+        fn parse_caa_record(record: &str) -> CaaRecord;
```
mikea, line 19: these should be Result. Parsing could fail by definition. Result is supported by cxx and is transformed into exception.

src/rust/dns/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+/// ```
+/// let record = parse_caa_record("\\# 15 00 05 69 73 73 75 65 70 6b 69 2e 67 6f 6f 67");
+/// assert_eq!(record.critical, false);
+/// assert_eq!(record.field, "issue")
+/// assert_eq!(record.value, "pki.goog")
+/// ```
+fn parse_caa_record(record: &str) -> ffi::CaaRecord {
+    // Let's remove "\\#" and the length of data from the beginning of the record
+    let data = record.split_ascii_whitespace().collect::<Vec<_>>()[2..].to_vec();
+    let critical = data[0].parse::<u8>().unwrap();
```
mikea, line 64: no panic please

src/rust/dns/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+fn parse_replacement(input: &[&str]) -> String {
+    if input.is_empty() {
+        return String::new();
+    }
+
+    let mut output: Vec<String> = vec![];
+    let mut length_index = 0;
+    let mut offset_index = 1;
+
+    while length_index < input.len() {
```
mikea, line 158: maybe a simple comment about what's going on here? Hard to decode intent.

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,229 @@
...
+  type SOA,
+  type SRV,
+  type TTLResponse,
+} from 'node-internal:internal_dns_client';
+import { DnsError } from 'node-internal:internal_errors';
+import { validateString } from 'node-internal:validators';
+import * as errorCodes from 'node-internal:internal_dns_constants';
+
+export function getServers(): string[] {
+  throw new Error('Not implemented');
```
jasnell, line 30: A comment on these "Not implemented" stubs describing them a bit more would be helpful for future contributors. Are we not implementing them for a particular reason? Is there a path to implementing them? Links to the Node.js docs describing them, etc. Basically just something that would give enough context for someone coming into the project later to know what, if anything, to do with these.

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,229 @@
...
+}
+
+export function resolve(): void {
+  throw new Error('Not implemented');
+}
+
+export function resolve4(
+  name: string,
+  options?: { ttl?: boolean }
+): Promise<(string | TTLResponse)[]> {
```
jasnell, line 48: Non-blocking: Ugh I've always hated the fact that these returns are polymorphic

src/node/internal/internal_dns.ts
```diff
@@ -0,0 +1,229 @@
...
+    validateAnswer(json.Answer, name, 'queryAaaa');
+
+    return json.Answer.map((a) =>
+      ttl ? { ttl: a.TTL, address: a.data } : a.data
+    );
+  });
+}
+
+export function resolveAny(): void {
+  throw new Error('Not implemented');
```
jasnell, line 86: A comment here especially would be helpful. Why are we not implementing the `resolveAny(...)` when we implement the rest of these.

jasnell (approved):
LGTM for everything except the Rust code, which I'll let someone else review. A couple nits regarding more code comments that I'd like to see addressed before this lands but otherwise the code good.

jasnell (approved):

src/rust/dns/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+    extern "Rust" {
+        fn parse_caa_record(record: &str) -> Result<CaaRecord>;
+        fn parse_naptr_record(record: &str) -> Result<NaptrRecord>;
+    }
+}
+
+/// Given a vector of strings, converts each slice to UTF-8 from HEX.
+///
+/// # Errors
```
mikea, line 40: I personally don't care about errors documentation much (panics are important). Is clippy requiring them? I thought we have it off. Feel free to drop if this is true.

src/rust/dns/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+/// `DnsParserError::InvalidHexString`
+/// `DnsParserError::ParseIntError`
+pub fn decode_hex(input: &[&str]) -> Result<Vec<String>, DnsParserError> {
+    let mut v = Vec::with_capacity(input.len());
+
+    for slice in input {
+        let num = u16::from_str_radix(slice, 16)?;
+        let ch = String::from_utf16(&[num])
+            .map_err(|_| DnsParserError::InvalidHexString("Invalid UTF-16 sequence".to_owned()))?;
```
mikea, line 49: you don't need map_err if you have thiserror with `from` setup correctly.

mikea (approved):
only one issue. otherwise rust side lg.

src/rust/dns/lib.rs
anonrig: Yes, if the function is pub then it requires it.

src/rust/dns/lib.rs
anonrig: I didn't want to add a specific error for utf16 encoding. invalid hex string with appropriate error message already points us towards this particular line.

irvinebroque:
dev docs?

PR #3185, opened 2024-11-28
Log outputGate broken shouldRetryCountsAgainstLimits for alarms

jqmmes:
We have some overloaded exceptions that are set as `EXCEPTION_IS_USER_ERROR`. These errors should be reported back to the alarm-manager as user errors, but some are being reported as internal errors.

This PR adds additional logging and does a hackish check to enforce that `broken.exceededMemory` errors during the alarm execution are treated as user errors.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -471,16 +471,17 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
             !jsg::isDoNotLogException(desc) && context.isOutputGateBroken()) {
           LOG_NOSENTRY(ERROR, "output lock broke during alarm execution", actorId, e);
         } else if (context.isOutputGateBroken()) {
-          // We don't usually log these messages, but it's useful to know the real reason we failed
-          // to correctly investigate stuck alarms.
-          LOG_NOSENTRY(ERROR,
-              "output lock broke during alarm execution without an interesting error description",
-              actorId, e);
           if (e.getDetail(jsg::EXCEPTION_IS_USER_ERROR) != kj::none) {
             // The handler failed because the user overloaded the object. It's their fault, we'll not
             // retry forever.
             shouldRetryCountsAgainstLimits = true;
           }
+
+          // We don't usually log these messages, but it's useful to know the real reason we failed
+          // to correctly investigate stuck alarms.
+          LOG_NOSENTRY(ERROR,
+              "output lock broke during alarm execution without an interesting error description",
+              actorId, e, shouldRetryCountsAgainstLimits);
         }
         return WorkerInterface::AlarmResult{.retry = true,
           .retryCountsAgainstLimit = shouldRetryCountsAgainstLimits,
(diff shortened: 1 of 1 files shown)


src/workerd/api/global-scope.c++
```diff
@@ -471,16 +471,28 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
             !jsg::isDoNotLogException(desc) && context.isOutputGateBroken()) {
           LOG_NOSENTRY(ERROR, "output lock broke during alarm execution", actorId, e);
         } else if (context.isOutputGateBroken()) {
-          // We don't usually log these messages, but it's useful to know the real reason we failed
-          // to correctly investigate stuck alarms.
-          LOG_NOSENTRY(ERROR,
-              "output lock broke during alarm execution without an interesting error description",
-              actorId, e);
+          if (e.getType() == kj::Exception::Type::OVERLOADED) {
+            // This is a bit hackish, but we know for a fact that this specific exception detail is
```
MellowYarker, line 475: Can we maybe use exception details here? https://github.com/capnproto/capnproto/blob/14132442b125d0383285e36809e467c6b6a759aa/c%2B%2B/src/kj/exception.h#L185

src/workerd/api/global-scope.c++
```diff
@@ -471,16 +471,28 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
...
+            // caused by user error, but not all overloaded exceptions are user fault. As such
+            // let's treat this special case.
+            // TODO(soon): Get a more robust way to process user errors comming from JSG exceptions.
+            if (e.getDescription() ==
```
MellowYarker, line 479: Why 2 `if`s when we really only care about both being true? Could the `if` above combine them with an `&&`?

Alternatively, on line 485 we could do an OR along with `overloaded && matchesString`?

MellowYarker (approved):

src/workerd/api/global-scope.c++
MellowYarker: Sorry, ignore me. I didn't realize that's what was happening below :sweat_smile:.

src/workerd/api/global-scope.c++
Frederik-Baetens: Commented the same thing but had forgotten to submit my review :sweat_smile:

src/workerd/api/global-scope.c++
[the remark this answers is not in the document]
MellowYarker: Heh, and then I [also commented this](https://github.com/cloudflare/workerd/pull/3185#discussion_r1862565943), but accidentally resolved _the wrong comment_ so it got hidden :laughing:

src/workerd/api/global-scope.c++
```diff
@@ -471,16 +471,28 @@ kj::Promise<WorkerInterface::AlarmResult> ServiceWorkerGlobalScope::runAlarm(kj:
...
+                "overloaded: jsg.Error: Durable Object's isolate exceeded its memory limit and was"
```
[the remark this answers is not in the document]
jqmmes: I want to simply use the exception details, but I'm not confident we're keeping those details all the time. I'm OK with dropping the description matching for now, and rely on being able to propagate the exception details everywhere.

PR #3190, opened 2024-11-29
Add API to set and get the V8 cage base address

erikcorry:
I want to use this to call madvise just before a crash so the V8 heap
for the crashing isolate is part of the core dump.

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -735,6 +735,15 @@ consteval size_t prefixLengthToStrip(const char (&s)[N]) {
   registry.template registerStructProperty<name##_JSG_NAME_DO_NOT_USE_DIRECTLY,                    \
       decltype(::kj::instance<Self>().name), &Self::name>()
 
+enum SetDataIndex {
+  SET_DATA_ISOLATE_BASE,
+  SET_DATA_TYPE_WRAPPER,
+  SET_DATA_ISOLATE,
+  SET_DATA_LOCK,
+  SET_DATA_CAGE_BASE,
+  SET_DATA_SLOTS_IN_USE,
+};
+
 // =======================================================================================
 // Special types
 //
@@ -2272,7 +2281,7 @@ class Lock {
   // This method is intended to be used in callbacks from V8 that pass an isolate pointer but
   // don't provide any further context. Most code should rely on the caller passing in a `Lock&`.
   static Lock& from(v8::Isolate* v8Isolate) {
-    return *reinterpret_cast<Lock*>(v8Isolate->GetData(2));
+    return *reinterpret_cast<Lock*>(v8Isolate->GetData(SET_DATA_ISOLATE));
   }
 
   // RAII construct that reports amount of external memory to be manually attributed to
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -735,12 +735,20 @@ consteval size_t prefixLengthToStrip(const char (&s)[N]) {
   registry.template registerStructProperty<name##_JSG_NAME_DO_NOT_USE_DIRECTLY,                    \
       decltype(::kj::instance<Self>().name), &Self::name>()
 
+// Indexes for adding API data to V8's Isolate object.
 enum SetDataIndex {
+  // The jsg::IsolateBase for a particular V8 isolate.
   SET_DATA_ISOLATE_BASE,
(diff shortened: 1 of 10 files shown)


src/workerd/jsg/setup.h
```diff
@@ -355,6 +355,14 @@ class IsolateBase {
 // intended to be invoked from a signal handler.
 kj::Maybe<kj::StringPtr> getJsStackTrace(void* ucontext, kj::ArrayPtr<char> scratch);
 
+// Set the location of the pointer cage base for the current isolate.  This is only
+// used by getJsCageBase().
+void setJsCageBase(void* cage_base);
```
harrishancock, line 360: Nit: prefer camelCase

src/workerd/jsg/setup.c++
```diff
@@ -562,6 +562,22 @@ void IsolateBase::jitCodeEvent(const v8::JitCodeEvent* event) noexcept {
...
+  if (isolate == nullptr) return nullptr;
+  // Returns null if we are not using pointer cages.
+  return isolate->GetData(4);
+}
+
+void setJsCageBase(void* base) {
+  if (!v8Initialized) return;
+  v8::Isolate* isolate = v8::Isolate::TryGetCurrent();
+  if (isolate == nullptr) return;
+  // Returns null if we are not using pointer cages.
```
harrishancock, line 577: I'm not sure this comment makes sense -- probably copy-pasta?

src/workerd/jsg/setup.c++
```diff
@@ -562,6 +562,22 @@ void IsolateBase::jitCodeEvent(const v8::JitCodeEvent* event) noexcept {
   }
 }
 
+void* getJsCageBase() {
+  if (!v8Initialized) return nullptr;
+  v8::Isolate* isolate = v8::Isolate::TryGetCurrent();
+  if (isolate == nullptr) return nullptr;
+  // Returns null if we are not using pointer cages.
```
harrishancock, line 569: Does this comment refer to the line above, where we return nullptr? Or the line below? If the line below, I'm not sure it makes sense, since someone could probably call `setJsCageBase()` with a non-null value even if we're not using pointer cages, right?

src/workerd/jsg/setup.c++
```diff
@@ -562,6 +562,22 @@ void IsolateBase::jitCodeEvent(const v8::JitCodeEvent* event) noexcept {
...
+  return isolate->GetData(4);
```
harrishancock, line 570: Where does the magic number 4 come from? Can it be hoisted out to a constexpr variable?

harrishancock (approved):

src/workerd/jsg/jsg.h
```diff
@@ -735,6 +735,15 @@ consteval size_t prefixLengthToStrip(const char (&s)[N]) {
   registry.template registerStructProperty<name##_JSG_NAME_DO_NOT_USE_DIRECTLY,                    \
       decltype(::kj::instance<Self>().name), &Self::name>()
 
+enum SetDataIndex {
```
anonrig, line 738: Would you mind adding a documentation/comment for each one of these values?

anonrig (approved):

src/workerd/jsg/setup.c++
```diff
@@ -562,6 +564,21 @@ void IsolateBase::jitCodeEvent(const v8::JitCodeEvent* event) noexcept {
...
 
+void* getJsCageBase() {
+  if (!v8Initialized) return nullptr;
+  v8::Isolate* isolate = v8::Isolate::TryGetCurrent();
+  if (isolate == nullptr) return nullptr;
+  // Returns null if setJsCageBase was never called.
+  return isolate->GetData(SET_DATA_CAGE_BASE);
+}
+
+void setJsCageBase(void* cageBase) {
```
jp4a50, line 575: [nit] Unless you plan to call this from lots of different call sites it seems like these methods mainly serve to silently hide subtle, presumably unexpected error cases (the early returns). Maybe it would be better to just not have these?

jp4a50 (approved):

src/workerd/jsg/setup.c++
erikcorry: It's a layering issue - Edgeworker doesn't have a lot of files that deal directly with the v8 Isolate, so I thought it made more sense to put it in jsg. The early returns are modelled on getJsStackTrace, and like that function the getJsCageBase is intended to be called during crash handling so it is written in a defensive style.

PR #3191, opened 2024-11-29
feat: better define step retry delay and step timeout types

LuisDuarte1:
Previously, both types were specified as a string, which is not intended behavior. We only allow for a specific formatted string - same as `WorkflowSleepDuration`.

I've added `WorkflowTimeoutDuration` and `WorkflowDelayDuration` for better DX in the types by reducing confusion (otherwise you would see a `WorkflowSleepDuration` in a timeout field) - also gives us flexibility if we need to separate them later.

While this could be a breaking change in the types, we already validate it at runtime, so there are no worries here.

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -217,15 +217,19 @@ declare module "cloudflare:workers" {
     | `${number} ${WorkflowDurationLabel}${"s" | ""}`
     | number;
 
+  export type WorkflowDelayDuration = WorkflowSleepDuration;
+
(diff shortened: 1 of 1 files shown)


vickykont (approved):
Looks good to me, but it might be best if someone that understands these types and what might introduce breaking changes approves.

sidharthachatterjee:
This is a breaking change. However, what makes it "breaking" is that the new types are actually accurate to what the implementation requires/anything that doesn't match these types (updated in this PR) actually fail. So we should merge this. I'll merge. Thanks folks!

PR #3192, opened 2024-11-29
Improve tail worker support

kentonv:
This fixes some issues from #3021, which I noticed while working on something unrelated.

The third commit is more of a suggestion. Do we think anyone is using `logging` yet? If so, maybe I should drop this commit. If not, maybe I should change the commit to actually remove `logging` in favor of the simpler flat array which I've named `tails`... let me know.

### DIFF
--- a/samples/tail-workers/config.capnp
+++ b/samples/tail-workers/config.capnp
@@ -13,7 +13,7 @@ const helloWorld :Workerd.Worker = (
     (name = "worker", esModule = embed "worker.js")
   ],
   compatibilityDate = "2024-10-14",
-  logging = ( toService = "log" ),
+  tails = ["log"],
 );
 
 const logWorker :Workerd.Worker = (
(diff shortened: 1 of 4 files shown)


jasnell (approved):

kentonv:
@jasnell Can you comment on whether `logging` is in use yet?

I know it theoretically could be, but have you explicitly documented it or told anyone about it or used it anywhere beyond what's in #3021?

jasnell:
Nope, intentionally hadn't documented it just yet but better to ask the wrangler folks. Not sure if anyone on that side has made use of it.

kentonv:
I don't see `toService` nor `toServices` mentioned in [workers-sdk](https://github.com/cloudflare/workers-sdk) so doesn't look like it's used in wrangler yet. OK, will update this to just replace it with `tails`.

danlapid (approved):

PR #3193, opened 2024-11-29
Implement connection props.

kentonv:
This adds `ctx.props` to the `ctx` object given to `WorkerEntrypoint`s. The property receives metadata about the particular service binding over which the entrypoint was invoked.

```
class MyEntrypoint extends WorkerEntrypoint {
  foo() {
    console.log("called by: " + this.ctx.props.caller);
  }
}
```

Service binding declarations in the workerd config may specify what metadata to pass:

```
bindings = [
  ( name = "FOO",
    service = (
      name = "my-service",
      entrypoint = "MyEntrypoint",
      props = (
        json = `{"caller": "my-calling-service"}
      )
    )
  )
]
```

Note that "caller" is just an example. The props can contain anything. Use cases include:
* Authentication of the caller's identity.
* Authorization / permissions (independent of caller identity).
* Specifying a particular resource. For example, if the `WorkerEntrypoint` represents a chat room, `props.roomId` could be the ID of the specific chat room to access.

This allows service bindings to implement a deeper capability-based security model, where bindings point to specific resources with specific permissions, instead of general APIs.

On Cloudflare, only users who have permission to modify your worker will have permission to create a binding containing arbitrary metadata. Meanwhile we will be creating a mechanism by which you can grant a service binding to your worker to someone, but where you specify the metadata. Thus, you can use the metadata to authenticate requests, without the need for any cryptography.

### DIFF
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -204,6 +204,9 @@ class TestController: public jsg::Object {
 
 class ExecutionContext: public jsg::Object {
  public:
+  ExecutionContext(jsg::Lock& js): props(js, js.obj()) {}
+  ExecutionContext(jsg::Lock& js, jsg::JsValue props): props(js, props) {}
+
   void waitUntil(kj::Promise<void> promise);
   void passThroughOnException();
 
@@ -211,9 +214,14 @@ class ExecutionContext: public jsg::Object {
   // and throwing an error at the client.
   void abort(jsg::Lock& js, jsg::Optional<jsg::Value> reason);
 
+  jsg::JsValue getProps(jsg::Lock& js) {
+    return props.getHandle(js);
+  }
+
   JSG_RESOURCE_TYPE(ExecutionContext, CompatibilityFlags::Reader flags) {
     JSG_METHOD(waitUntil);
     JSG_METHOD(passThroughOnException);
+    JSG_LAZY_INSTANCE_PROPERTY(props, getProps);
 
     if (flags.getWorkerdExperimental()) {
       // TODO(soon): Before making this generally available we need to:
@@ -228,6 +236,17 @@ class ExecutionContext: public jsg::Object {
       //   consistent with each other.
       JSG_METHOD(abort);
     }
+  }
+
+  void visitForMemoryInfo(jsg::MemoryTracker& tracker) const {
+    tracker.trackField("props", props);
+  }
+
+ private:
+  jsg::JsRef<jsg::JsValue> props;
+
+  void visitForGc(jsg::GcVisitor& visitor) {
+    visitor.visit(props);
   }
 };
 
--- a/src/workerd/io/frankenvalue.h
+++ b/src/workerd/io/frankenvalue.h
@@ -1,0 +1,73 @@
+// Copyright (c) 2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#pragma once
+
+#include <workerd/io/frankenvalue.capnp.h>
+#include <workerd/jsg/jsg.h>
+
+namespace workerd {
+
+// C++ class mirroring `Frankenvalue` as defined in `frankenvalue.capnp`.
+//
+// Represents a JavaScript value that has been stitched together from multiple sources outside of
+// a JavaScript evaluation context. The Frankevalue can be evaluated down to a JS value as soon
+// as it has a JS execution environment in which to be evaluated.
+//
+// This is used in particular to represent `ctx.props`.
+class Frankenvalue {
+ public:
+  Frankenvalue(): value(EmptyObject()) {}
+
+  bool empty() const {
+    return value.is<EmptyObject>() && properties.empty();
+  }
+
+  Frankenvalue clone() const;
+
+  // Convert to/from capnp format.
+  void toCapnp(rpc::Frankenvalue::Builder builder) const;
+  static Frankenvalue fromCapnp(rpc::Frankenvalue::Reader reader);
+
+  // Convert to/from JavaScript values. Note that round trips here don't produce the exact same
+  // Frankenvalue representation: toJs() puts all the contents together into a single value, and
+  // fromJs() always returns a Frakenvalue containing a single V8-serialized value.
+  jsg::JsValue toJs(jsg::Lock& js) const;
+  static Frankenvalue fromJs(jsg::Lock& js, jsg::JsValue value);
+
+  // Construct a Frakenvalue from JSON.
+  //
+  // (It's not possible to convert a Frakenvalue back to JSON, except by evaluating it in JS and
+  // then JSON-stringifying from there.)
+  static Frankenvalue fromJson(kj::String json);
+
+  // Add a property to the value, represented as another Frankenvalue. This is how you "stitch
+  // together" values!
+  //
+  // This is called `set` because the new property will override any existing property with the
+  // same name, but note that this strictly appends content. The replacement happens only when the
+  // Frankenvalue is finally converted to JS.
+  void setProperty(kj::String name, Frankenvalue value);
+
+ private:
+  struct EmptyObject {};
+  struct Json {
+    kj::String json;
+  };
+  struct V8Serialized {
+    kj::Array<byte> data;
+  };
+  kj::OneOf<EmptyObject, Json, V8Serialized> value;
+
+  struct Property;
+  kj::Vector<Property> properties;
+};
+
+// Can't be defined inline since `Frankenvalue` is still incomplete there.
+struct Frankenvalue::Property {
+  kj::String name;
+  Frankenvalue value;
+};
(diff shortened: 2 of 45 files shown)


kentonv:
I'm not entirely happy that every `CustomEvent` implementation had to change here. I might go back and try a different approach to reduce the noise a bit.

kentonv:
Last push was fixup: b9c66097

src/workerd/api/global-scope.h
```diff
@@ -229,6 +237,13 @@ class ExecutionContext: public jsg::Object {
       JSG_METHOD(abort);
     }
   }
+
+ private:
+  jsg::JsRef<jsg::JsValue> props;
+
+  void visitForGc(jsg::GcVisitor& visitor) {
+    visitor.visit(props);
+  }
```
jasnell, line 246: Non-blocking nit: this should likely implement `visitForMemoryInfo(...)` also

src/workerd/io/frankenvalue.h
```diff
@@ -0,0 +1,66 @@
...
+//     https://opensource.org/licenses/Apache-2.0
+
+#pragma once
+
+#include <workerd/io/frankenvalue.capnp.h>
+#include <workerd/jsg/jsg.h>
+
+namespace workerd {
+
+class Frankenvalue {
```
jasnell, line 12: A code comment description of what this is and how/why to use it would be helpful.

src/workerd/io/frankenvalue.h
```diff
@@ -0,0 +1,66 @@
...
+
+  // Convert to/from capnp format.
+  void toCapnp(rpc::Frankenvalue::Builder builder) const;
+  static Frankenvalue fromCapnp(rpc::Frankenvalue::Reader reader);
+
+  // Convert to/from JavaScript values. Note that round trips here don't produce the exact same
+  // Frankenvalue representation: toJs() puts all the contents together into a single value, and
+  // fromJs() always returns a Frakenvalue containing a single V8-serialized value.
+  jsg::JsValue toJs(jsg::Lock& js) const;
+  static Frankenvalue fromJs(jsg::Lock& js, jsg::JsValue value);
```
jasnell, line 30: non-blocking nit: If converting to Json is going to be a common use case, then maybe consider exposing a `jsg::JsString toJson(jsg::Lock& js) const;` also?

src/workerd/io/worker.c++
```diff
@@ -1666,7 +1667,12 @@ Worker::Worker(kj::Own<const Script> scriptParam,
                     KJ_SWITCH_ONEOF(handler.value) {
                       KJ_CASE_ONEOF(obj, api::ExportedHandler) {
                         obj.env = lock.v8Ref(bindingsScope.As<v8::Value>());
-                        obj.ctx = jsg::alloc<api::ExecutionContext>();
+                        // TODO(cleanup): Unfortunately, for non-class-based handlers, we have
+                        //   always created only a single `ctx` object and reused it for all
+                        //   requests. This is weird and obviously wrong but changing it probably
+                        //   requires a compat flag. Until then, connection properties will not be
+                        //   available for non-class handlers.
```
jasnell, line 1674: For later... we *could* add a getter to `ExecutionContext` that always returns the request-specific `props` accessed through, perhaps, the `IoContext`. That's obviously a bit more work and is fairly convoluted so I'm fine with this restriction for now.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -495,13 +502,13 @@ kj::Promise<WorkerInterface::ScheduledResult> WorkerEntrypoint::runScheduled(
...
-      context.run([scheduledTime, cron, entrypointName = entrypointName, &context,
+      context.run([this, scheduledTime, cron, entrypointName = entrypointName, &context,
                       &metrics = incomingRequest->getMetrics()](Worker::Lock& lock) mutable {
     TRACE_EVENT("workerd", "WorkerEntrypoint::runScheduled() run");
     jsg::AsyncContextFrame::StorageScope traceScope = context.makeAsyncTraceScope(lock);
 
-    lock.getGlobalScope().startScheduled(
-        scheduledTime, cron, lock, lock.getExportedHandler(entrypointName, context.getActor()));
+    lock.getGlobalScope().startScheduled(scheduledTime, cron, lock,
+        lock.getExportedHandler(entrypointName, kj::mv(props), context.getActor()));
```
jp4a50, line 511: I guess this must be guaranteed somehow since we also capture `entrypointName` by ptr but what ensures that `this` (and thus `props`) has not been destroyed by the time this code executes?

jp4a50 (approved):
LGTM

src/workerd/io/frankenvalue.h
kentonv: No, you generally never convert a Frakenvalue to JSON. Parts of the Frakenvalue might come _from_ JSON (e.g. the binding configuration), but other parts may come from other places and may or may not be JSON-serializable. The final destination of a Frakenvalue is always a JS value in the callee isolate.

src/workerd/io/worker.c++
kentonv: That would actually be easy, but I think it's actually weird and broken that an `ExecutioContext` instance can be used in a different context than the one where it originated, and it'll just operate on whatever context it's used it... I mean, you would _expect_ that each context gets a unique `ExecutionContext`, which can only operate on that context. I want to fix this via a compat flag later.

src/workerd/api/global-scope.h
```diff
@@ -204,16 +204,24 @@ class TestController: public jsg::Object {
...
   void abort(jsg::Lock& js, jsg::Optional<jsg::Value> reason);
 
+  jsg::JsValue getProps(jsg::Lock& js) {
+    return props.getHandle(js);
+  }
+
   JSG_RESOURCE_TYPE(ExecutionContext, CompatibilityFlags::Reader flags) {
     JSG_METHOD(waitUntil);
     JSG_METHOD(passThroughOnException);
+    JSG_LAZY_INSTANCE_PROPERTY(props, getProps);
```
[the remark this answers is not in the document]
kentonv: Performance: A lazy instance property becomes a regular JS property after the first access, which can then be optimized better if it is accessed repeatedly. A regular property would have to call back to C++ for every access.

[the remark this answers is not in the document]
jasnell: I think the question was more: Why `JSG_LAZY_INSTANCE_PROPERTY` and not `JSG_LAZY_READYONLY_INSTANCE_PROPERTY`... you'd get the same lazy replacement either way. Tho, marking the field read only does have an ever so slight performance impact.

src/workerd/io/worker-entrypoint.c++
kentonv: I think that for `WorkerEntrypoint` to be destroyed, first the promise returned by `runScheduled()` would have to have been canceled, which from what I could tell will actually drop the IoContext, canceling all waitUntil()s.

This is actually sort of inconsistent: Dropping the promise returned by request() (for HTTP) does NOT cancel waitUntil()s. But in that case, the code that uses `props` is not in a `waitUntil`, it's in a promise that _is_ still canceled when the request is canceled...

With all that said, I don't feel super-great about depending on this so I'm going to remove the `this` capture and instead capture `props` by-move.

Incidentally, the treatment of `entrypointName` doesn't seem to prove anything as it appears that `WorkerEntrypoint` as a whole does not own `entrypointName` but rather expects some higher power to keep it alive...

kentonv:
Fixups:
* 7f6d4cfa
* ec46051d
* 969a8e64
* 7b310a12

src/workerd/io/frankenvalue.h
```diff
@@ -0,0 +1,73 @@
...
+  // (It's not possible to convert a Frakenvalue back to JSON, except by evaluating it in JS and
+  // then JSON-stringifying from there.)
+  static Frankenvalue fromJson(kj::String json);
+
+  // Add a property to the value, represented as another Frankenvalue. This is how you "stitch
+  // together" values!
+  //
+  // This is called `set` because the new property will override any existing property with the
+  // same name, but note that this strictly appends content. The replacement happens only when the
+  // Frankenvalue is finally converted to JS.
```
jasnell, line 50: Non-blocking nit: It's likely worth documenting that `toJs(...)` will throw if there are properties added but the realized JS value is not actually an object.

src/workerd/io/frankenvalue.h
```diff
@@ -0,0 +1,73 @@
...
+  void setProperty(kj::String name, Frankenvalue value);
+
+ private:
+  struct EmptyObject {};
+  struct Json {
+    kj::String json;
+  };
+  struct V8Serialized {
+    kj::Array<byte> data;
+  };
```
jasnell, line 60: Non-blocking nit: Since these are just single valued, structs this might be a bit cleaner:

```
using Json = kj::String;
using V8Serialized = kj::Array<kj::byte>;
```

jasnell (approved):

PR #3194, opened 2024-11-29
Enable using context-specific typewrapper instances.

danlapid:
This PR has two commits:
1. Revert the previous commit to disable memoization as this is no longer required with the new approach
2. A commit to enable using context-specific typewrapper instances

Sometimes we'd like to have multiple contexts with different
typewrapper that were instantiated with different configuration objects.
To do that this commit extends the Isolate class, wrapper is now a
vector of wrappers, the 0 indexed wrapper is the default wrapper but
additional wrappers can be created with `instantiateWrapper` which
receives a new Configuration object. To associate a context with its
wrapper we use `context->SetAlignedPointerInEmbedderData(3, wrapper)`.


src/workerd/jsg/setup.h
```diff
@@ -392,20 +423,26 @@ kj::Maybe<kj::StringPtr> getJsStackTrace(void* ucontext, kj::ArrayPtr<char> scra
 template <typename TypeWrapper>
 class Isolate: public IsolateBase {
  public:
+  explicit Isolate(bool dud,
```
danlapid, line 426: replace dud with something better

src/workerd/jsg/setup.h
```diff
@@ -392,20 +423,26 @@ kj::Maybe<kj::StringPtr> getJsStackTrace(void* ucontext, kj::ArrayPtr<char> scra
...
+      v8::Isolate::CreateParams createParams = {})
+      : IsolateBase(system, kj::mv(createParams), kj::mv(observer)) {}
+
   // Construct an isolate that requires configuration. `configuration` is a value that all
   // individual wrappers' configurations must be able to be constructed from. For example, if all
   // wrappers use the same configuration type, then `MetaConfiguration` should just be that type.
   // If different wrappers use different types, then `MetaConfiguration` should be some value that
   // inherits or defines conversion operators to each required type -- or the individual
   // configuration types must declare constructors from `MetaConfiguration`.
+  // TODO(now): update this comment
```
danlapid, line 438: todo

src/workerd/jsg/setup.h
```diff
@@ -398,14 +423,21 @@ class Isolate: public IsolateBase {
...
   template <typename MetaConfiguration>
   explicit Isolate(const V8System& system,
       MetaConfiguration&& configuration,
       kj::Own<IsolateObserver> observer,
-      v8::Isolate::CreateParams createParams = {})
-      : IsolateBase(system, kj::mv(createParams), kj::mv(observer)),
-        wrapper(wrapperSpace.construct(ptr, kj::fwd<MetaConfiguration>(configuration))) {
-    wrapper->initTypeWrapper();
+      v8::Isolate::CreateParams createParams = {},
+      bool instantiateDefaultWrapper = true)
```
danlapid, line 434: rename `instantiateDefaultWrapper`

danlapid:
Hey @jasnell @kentonv 
This is an alternate implementation following Kenton's remarks on the previous PR.
I think this should address Kenton's concerns.

I did my best to ensure the case where we never needed a second typewrapper doesn't see any performance regressions.

The types check GitHub check is failing just because I need to rebase, I'll do that after the review since there's no significant changes to the types in this PR.

src/workerd/jsg/setup.h
```diff
@@ -414,23 +445,39 @@ class Isolate: public IsolateBase {
...
+
+  template <typename MetaConfiguration>
+  kj::Own<TypeWrapper>& instantiateWrapper(MetaConfiguration&& configuration) {
+    hasExtraWrappers = true;
+    return wrappers.add(kj::heap<TypeWrapper>(ptr, kj::fwd<MetaConfiguration>(configuration)));
+  }
+
   ~Isolate() noexcept(false) {
-    dropWrappers(kj::mv(wrapper));
+    dropWrappers(wrappers.releaseAsArray());
```
kentonv, line 463: You could avoid the need to templatize/inline `dropWrappers()` by instead doing something here like:

```
int dummy;
dropWrappers(kj::attachRef(dummy, kj::mv(wrappers)));
```

Or alternatively, `dropWrappers()` could take a `kj::FunctionParam<void()> drop` which is invoked at the point where the wrappers should be dropped, and then this could be:

```
dropWrappers([&]() { wrappers.clear(); });
```

(Using `FunctionParam` for this avoids any memory allocation, but still permits `dropWrappers()` to be declared in the c++ file.)

src/workerd/jsg/setup.h
```diff
@@ -414,23 +445,39 @@ class Isolate: public IsolateBase {
...
+  template <typename MetaConfiguration>
+  void instantiateDefaultWrapper(MetaConfiguration&& configuration) {
+    KJ_DASSERT(wrappers[0].get() == nullptr);
+    auto wrapper = wrapperSpace.construct(ptr, kj::fwd<MetaConfiguration>(configuration));
+    wrapper->initTypeWrapper();
+    wrappers[0] = kj::mv(wrapper);
+  }
+
+  template <typename MetaConfiguration>
+  kj::Own<TypeWrapper>& instantiateWrapper(MetaConfiguration&& configuration) {
```
kentonv, line 457: It's a little weird that this returns a reference to a `kj::Own`, especially as it is a reference to a vector slot which could move if the vector is resized. Could you just return `TypeWrapper&` instead? And then `newContextWithWrapper()` would also take `TypeWrapper&` as the param?

Alternatively, maybe the two methods should be combined into a single function `newContextWithSeparateWrapper()` that always allocates a new wrapper for the new context? That avoids leaking a reference to the `TypeWrapper` at all, which seems good since it's meant to be a private implementation type.

kentonv:
Approach seems reasonable, just some nitpick comments.

danlapid:
Last push was just a rebase.
Fixups after the rebase are in the fixup commit: [afa5410](https://github.com/cloudflare/workerd/pull/3194/commits/afa541012c9ea2653fc349647c66860b96b3af08)

src/workerd/jsg/setup.h
```diff
@@ -656,9 +698,37 @@ class Isolate: public IsolateBase {
...
 
+ protected:
+  inline TypeWrapper* getWrapperByContext(jsg::Lock& js) {
+    if (KJ_LIKELY(!hasExtraWrappers)) {
+      return wrappers[0].get();
+    } else {
+      return getWrapperByContext(js.v8Context());
+    }
+  }
+  inline TypeWrapper* getWrapperByContext(v8::Local<v8::Context>&& context) {
```
kentonv, line 709: You wouldn't need this overload if the overload below took a plain value instead of a reference. `v8::Local` just wraps a pointer anyway, so passing it by value is not expensive.

src/workerd/jsg/setup.h
```diff
@@ -541,59 +559,74 @@ class Isolate: public IsolateBase {
...
           v8Isolate->GetCurrentContext(), promise, jsg::TypeErrorContext::other());
     }
 
-    // Creates a new JavaScript "context", i.e. the global object. This is the first step to
-    // executing JavaScript code. T should be one of your API types which you want to use as the
-    // global object. `args...` are passed to the type's constructor.
     template <typename T, typename... Args>
-    JsContext<T> newContext(NewContextOptions options, Args&&... args) {
+    JsContext<T> newContextWithWrapper(
+        kj::Own<TypeWrapper>& wrapper, NewContextOptions options, Args&&... args) {
```
kentonv, line 612: Instead of passing `kj::Own<TypeWrapper>&`, why not just pass `TypeWrapper&`? It doesn't look like you actually do anything with the `Own` except deference it.

kentonv (approved):
Just some minor style nitpicks.

src/workerd/jsg/setup.h
danlapid: I can pass a `TypeWrapper*` instead but I dislike the idea of passing a `Typewrapper&` and then getting it's address with `&wrapper`

danlapid:
Added a fixup with pr fixups in [`9c25460` (#3194)](https://github.com/cloudflare/workerd/pull/3194/commits/9c25460dc9fad91ccee41b693a12685377cb4683) then rebased

danlapid:
Last push merged fixups

src/workerd/jsg/setup.h
kentonv: > I dislike the idea of passing a Typewrapper& and then getting it's address with &wrapper

Why? We do that all the time. I don't see the issue.

PR #3195, opened 2024-11-30
Enable cache: no-cache compat flag and enum

AdityaAtulTewari:
First part of `cache: no-cache` implementation:
- Work-shopping the test infrastructure in workerd.
- Allow no-cache construction
- Update typescript information

Note that the actual implementation is incompelete.

### DIFF
--- a/src/workerd/api/http-test-ts.ts
+++ b/src/workerd/api/http-test-ts.ts
@@ -50,40 +50,56 @@ async function assertFetchCacheRejectsError(
 
 export const cacheMode = {
   async test(ctrl: any, env: any, ctx: any) {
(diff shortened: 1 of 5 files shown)


fhanau:
I'll need some context here to review this well:
- Is the compat flag the right enabling mechanism here? Are there cases where supporting cache:no-store will be incompatible? Or is this just to support using it while it is not fully stable?
- Should this be landed while incomplete? Are there any internal PRs I should look out for?

jasnell (approved):
LGTM but let's be sure to test internally before merging

danlapid:
@AdityaAtulTewari this PR was merged when type checks didn't pass. Please fix types.

PR #3201, opened 2024-12-03
Make spanId a random 64-bit rather than using a counter

jasnell:
In the streaming tail worker design discussion it was decided that span IDs will follow the otel pattern and will be random 64-bit rather than monotonically increasing counter.

### DIFF
--- a/src/workerd/io/trace.c++
+++ b/src/workerd/io/trace.c++
@@ -126,29 +126,32 @@ kj::String TraceId::toW3C() const {
   return kj::str(s.releaseAsArray());
 }
 
-TraceId TraceId::fromEntropy(kj::Maybe<kj::EntropySource&> entropySource) {
-  if (isPredictableModeForTest()) {
-    return TraceId(0x2a2a2a2a2a2a2a2a, 0x2a2a2a2a2a2a2a2a);
-  }
-
-  uint64_t low = 0;
-  uint64_t high = 0;
+namespace {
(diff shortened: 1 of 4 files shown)


src/workerd/io/trace.c++
```diff
@@ -126,51 +126,54 @@ kj::String TraceId::toW3C() const {
...
 
-TraceId TraceId::fromEntropy(kj::Maybe<kj::EntropySource&> entropySource) {
-  if (isPredictableModeForTest()) {
-    return TraceId(0x2a2a2a2a2a2a2a2a, 0x2a2a2a2a2a2a2a2a);
-  }
-
-  uint64_t low = 0;
-  uint64_t high = 0;
+namespace {
+uint64_t getRandom64Bit(kj::Maybe<kj::EntropySource&>& entropySource) {
```
anonrig, line 130: Can you add a comment here describing the algorithm behind this generation?

anonrig (commented):
I assume it's not a breaking change, right?

src/workerd/io/trace.c++
jasnell, line 130:
```suggestion
// Generate a random 64-bit number. If an entropy source is given, use it,
// otherwise fall back to RAND_bytes. If, by some chance, the random
// generator returns 0, we'll try up to three times before just returning.
uint64_t getRandom64Bit(kj::Maybe<kj::EntropySource&>& entropySource) {
```

jasnell:
Not a breaking change because nothing is relying on this right now.

anonrig (approved):

PR #3202, opened 2024-12-03
remove async fs from build-types.ts

anonrig:
The PR aims to fix build errors related to SIGINT on types/definition task.

- Removes async fs from file
- Replace SIGINT with SIGTERM when we need to kill the spawned process.


fhanau (approved):
LGTM – not familiar with all of the JS I/O APIs but looks reasonable overall

anonrig:
Update: I'm waiting on the PR to update Node.js to v22 to land this.

penalosa (approved):

PR #3206, opened 2024-12-03
update nodejs to v22.11.0

anonrig:
Update Node.js to latest version to support features like `Promise.withResolvers()`


build/deps/build_deps.jsonc
```diff
@@ -93,6 +93,15 @@
...
       "file_regex": "\\.tar\\.gz$"
     },
+    {
+      // Needed for running the latest Node.js due to aspect_rules_js
+      "name": "rules_nodejs",
+      "type": "github_release",
+      "owner": "bazel-contrib",
+      "repo": "rules_nodejs",
+      "file_regex": "\\.tar\\.gz$",
+      "freeze_version": "v6.3.2"
```
fhanau, line 103: - [ ] No need to freeze this – later versions will work just as well

build/deps/build_deps.jsonc
```diff
@@ -93,6 +93,15 @@
       "repo": "rules_pkg",
       "file_regex": "\\.tar\\.gz$"
     },
+    {
+      // Needed for running the latest Node.js due to aspect_rules_js
```
fhanau, line 97: Nit: rules_js is not making the situation worse per se – we'd pull in an outdated version using `rules_js_dependencies()` without this change, but if we weren't using rules_js we'd still have to import it ourselves. This comment should say something like "due to aspect_rules_js using older version by default".

fhanau (approved):

build/deps/build_deps.jsonc
anonrig: Thanks! I updated the coment.

build/deps/build_deps.jsonc
anonrig: Thanks! Removed it.

PR #3209, opened 2024-12-04
[o11y] Add traceId to user trace

fhanau:
Builds on #3180. Also see the corresponding upstream PR.

### DIFF
--- a/src/workerd/io/trace.h
+++ b/src/workerd/io/trace.h
@@ -95,6 +95,9 @@ class TraceId final {
   // Replicates W3C Serialization
   kj::String toW3C() const;
 
+  // Return network order hex representation
+  kj::String toNetworkOrderHex() const;
+
   // Creates a random Trace Id, optionally using a given entropy source. If an
   // entropy source is not given, then we fallback to using BoringSSL's RAND_bytes.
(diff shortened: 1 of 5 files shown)


src/workerd/io/trace.h
```diff
@@ -92,6 +92,9 @@ class TraceId final {
   // Replicates W3C Serialization
   kj::String toW3C() const;
 
+  // Return network order hex representation
+  kj::String toNEHex() const;
```
jasnell, line 96: Nit: While it's more verbose, I prefer avoiding the abbreviation here. `s/toNEHex/toNetworkOrderHex`

src/workerd/io/trace.h
```diff
@@ -609,6 +616,7 @@ class WorkerTracer final: public kj::Refcounted, public BaseTracer {
 
   void addLog(kj::Date timestamp, LogLevel logLevel, kj::String message) override;
   void addSpan(CompleteSpan&& span) override;
+  void setTraceId(tracing::TraceId traceId);
```
jasnell, line 619: Nit: `const tracing::TraceId&` to prevent the extra copy? I know it's a small class but still.

fhanau:
Closing this PR – we'll not be adding more features here and support this in span tracing instead.

PR #3211, opened 2024-12-05
html-rewriter: Support streaming content replacements

npaun:
lol-html allows you to insert replacement values in various places in a HTML stream. Previously, these replacement values were always strings. However, recently lol-html added support for _streaming_ replacements in https://github.com/cloudflare/lol-html/pull/229. lol-html requires the streams to produce valid UTF-8 data, but it may be written in arbitrary length chunks (even partial UTF-8 characters).

This PR adds support for replacement values that are `ReadableStream`s or `Response`s.

### DIFF
--- a/deps/rust/cargo.bzl
+++ b/deps/rust/cargo.bzl
@@ -16,10 +16,9 @@ PACKAGES = {
     "cxx": crate.spec(version = "1"),
     "cxxbridge-cmd": crate.spec(version = "1"),
     "flate2": crate.spec(version = "1"),
-    "lol_html_c_api": crate.spec(
-        git = "https://github.com/cloudflare/lol-html.git",
-        rev = "cac9f2f59aea8ad803286b0aae0d667926f441c7",
-    ),
+    # Commit hash refers to lol-html v2.1.0. We then access the nested lol_html_c_api crate within.
+    # TODO(npaun): Seems like some changes are still ongoing on the lol-html side. Recheck this in a couple weeks.
+    "lol_html_c_api": crate.spec(git = "https://github.com/cloudflare/lol-html.git", rev = "cac9f2f59aea8ad803286b0aae0d667926f441c7"),
     "nix": crate.spec(version = "0"),
     "pico-args": crate.spec(version = "0"),
     "proc-macro2": crate.spec(version = "1"),
--- a/deps/rust/cargo.bzl
+++ b/deps/rust/cargo.bzl
@@ -17,7 +17,8 @@ PACKAGES = {
     "cxxbridge-cmd": crate.spec(version = "1"),
     "flate2": crate.spec(version = "1"),
     # Commit hash refers to lol-html v2.1.0. We then access the nested lol_html_c_api crate within.
-    # TODO(npaun): Seems like some changes are still ongoing on the lol-html side. Recheck this in a couple weeks.
+    # TODO(npaun): The next release of lol-html could change the way we access the nested crate.
+    # Check once https://github.com/cloudflare/lol-html/pull/247 is in a release.
     "lol_html_c_api": crate.spec(git = "https://github.com/cloudflare/lol-html.git", rev = "cac9f2f59aea8ad803286b0aae0d667926f441c7"),
     "nix": crate.spec(version = "0"),
     "pico-args": crate.spec(version = "0"),
--- a/src/workerd/api/html-rewriter.c++
+++ b/src/workerd/api/html-rewriter.c++
@@ -235,6 +235,12 @@ class Rewriter final: public WritableStreamSink {
   // Implementation for `Element::onEndTag` to avoid exposing private details of Rewriter.
   void onEndTag(lol_html_element_t* element, ElementCallbackFunction&& callback);
 
+  lol_html_streaming_handler_t registerReplacer(jsg::Ref<ReadableStream> content, bool isHtml);
+
+  ~Rewriter() {
+    KJ_ASSERT(registeredReplacers.size() == 0, "Some replacers were leaked by lol-html");
+  }
+
  private:
   // Wait for the write promise (if any) produced by our `output()` callback, then, if there is a
   // stored exception, abort the wrapped WritableStreamSink with it, then return the exception.
@@ -266,6 +272,15 @@ class Rewriter final: public WritableStreamSink {
 
   friend class ::workerd::api::HTMLRewriter;
 
+  // Keeps track of streams currently being used as replacement content for tokens.
+  // lol-html will invoke replacerThunk with a pointer to the RegisteredReplacer to be used.
+  class RegisteredReplacer {
+   public:
+    Rewriter& rewriter;
+    bool isHtml;
+    jsg::Ref<ReadableStream> stream;
+  };
+
   struct RegisteredHandler {
     // A back-reference to the rewriter which owns this particular registered handler.
     Rewriter& rewriter;
@@ -294,6 +309,15 @@ class Rewriter final: public WritableStreamSink {
   // Eagerly free this handler. Should only be called if we're confident the handler will never be
   // used again.
   void removeEndTagHandler(RegisteredHandler& registration);
+
+  // Field stores a list of readable streams that are being used by lol-html for replacements
+  kj::HashMap<void*, kj::Own<RegisteredReplacer>> registeredReplacers;
+
+  static int replacerThunk(lol_html_streaming_sink_t* sink, void* userData);
+  kj::Promise<void> replacerThunkPromise(
+      lol_html_streaming_sink_t* sink, RegisteredReplacer& registration);
+  int replacerThunkImpl(lol_html_streaming_sink_t* sink, RegisteredReplacer& registration);
+  static void removeRegisteredReplacer(void* userData);
 
   // Must be constructed AFTER the registered handler vector, since the function which constructs
   // this (buildRewriter()) modifies that vector.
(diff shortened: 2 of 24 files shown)


src/workerd/api/html-rewriter.h
```diff
@@ -241,6 +241,9 @@ class Element final: public HTMLRewriter::Token {
   kj::Maybe<Impl> impl;
 
   void htmlContentScopeEnd() override;
+
+  template <auto Func, auto StreamingFunc>
+  jsg::Ref<Element> contentRewriter(Content content, jsg::Optional<ContentOptions> options);
```
harrishancock, line 246: Nit: I'd find `rewriteContent()` to be a bit clearer of a name, or maybe even `rewriteContentGeneric()` with a doc comment explaining it's the shared implementation of all of the `before()`/`after()`/etc. functions.

src/workerd/api/html-rewriter.h
```diff
@@ -304,6 +307,8 @@ class EndTag final: public HTMLRewriter::Token {
 
  private:
   kj::Maybe<CType&> impl;
+  // TODO(npaun): is adding rewriter field like this legit?
```
harrishancock, line 310: I would say, not quite legit, because this could become dangling. The reason is that subclasses of Token are the C++-side implementations of JS wrapper objects, meaning JS can arbitrarily extend their lifetimes beyond `Rewriter`'s lifetime. This is the reason there's the TokenScope guard which nullifies `impl` above.

Of course, `checkToken()` is called in all of the places we use this `rewriter` reference, so the code as written _is_ safe, but it feels like leaving a footgun around.

It occurs to me that an alternative could be to make each Token subclass own their own HashMap of registered replacers. We would expect each Token's replacers to be dropped before its associated TokenScope ends, meaning we could have `~TokenScope()` assert if there are any registered replacers left in the Token.

deps/rust/BUILD.lolhtml
```diff
@@ -1,10 +1,10 @@
 cc_library(
     name = "lolhtml",
-    hdrs = ["@crates_vendor__lol_html-2.0.0//:c-api/include/lol_html.h"],
-    include_prefix = "c-api/include",
-    strip_include_prefix = "c-api/include",
+    hdrs = ["@crates_vendor__lol_html_c_api-1.1.2//:include/lol_html.h"],
```
harrishancock, line 3: Could you separate the dependency update into a second commit?

src/workerd/api/html-rewriter.c++
```diff
@@ -4,7 +4,7 @@
 
 #include "html-rewriter.h"
 
-#include "c-api/include/lol_html.h"
+#include "include/lol_html.h"
```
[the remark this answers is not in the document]
Nayawana: Thanks 🙏🏻👍🏻

src/workerd/api/html-rewriter.h
Nayawana: Thanks 🙏🏻👍🏻

src/workerd/api/html-rewriter.c++
```diff
@@ -578,6 +596,124 @@ kj::Promise<void> Rewriter::thunkPromise(CType* content, RegisteredHandler& regi
...
 
+lol_html_streaming_handler_t Rewriter::registerReplacer(
+    jsg::Ref<ReadableStream> content, bool isHtml) {
+  auto replacer = kj::heap<RegisteredReplacer>(*this, isHtml, kj::mv(content));
+  auto userData = replacer.get();
+  registeredReplacers.insert(userData, kj::mv(replacer));
+
+  return {.user_data = userData,
+    .write_all_callback = Rewriter::replacerThunk,
+    .drop_callback = Rewriter::removeRegisteredReplacer};
```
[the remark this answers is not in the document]
npaun: Nice trick.
I wonder if I could fix the clang-format style so we don't get this problem anymore.

src/workerd/api/html-rewriter.h
npaun: The dangling reference issue is fixed.
However, I didn't move the registered replacers to `Token` yet because it seems like the replacer callback can be invoked after their registering handler has exited. Happy to revisit this if you have any ideas to try later on.

src/workerd/api/html-rewriter.c++
```diff
@@ -294,6 +303,15 @@ class Rewriter final: public WritableStreamSink {
   // used again.
   void removeEndTagHandler(RegisteredHandler& registration);
 
+  // Field stores a list of readable streams that are being used by lol-html for replacements
+  kj::HashMap<void*, kj::Own<RegisteredReplacer>> registeredReplacers;
```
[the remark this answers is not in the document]
npaun: Implemented the first option for now.

deps/rust/cargo.bzl
```diff
@@ -16,7 +16,9 @@ PACKAGES = {
     "cxx": crate.spec(version = "1"),
     "cxxbridge-cmd": crate.spec(version = "1"),
     "flate2": crate.spec(version = "1"),
-    "lolhtml": crate.spec(git = "https://github.com/cloudflare/lol-html.git", rev = "a161bb319a61ddfb4c66e29153bbf8d6491f28cf"),
+    # Commit hash refers to lol-html v2.1.0. We then access the nested lol_html_c_api crate within.
+    # TODO(npaun): Seems like some changes are still ongoing on the lol-html side. Recheck this in a couple weeks.
```
anonrig, line 20: Can you be specific, as well as adding a link to any issue that references this? No-one can complete this comment at this current state.  "Couple of weeks" is extremely vague.

src/workerd/api/html-rewriter.c++
```diff
@@ -265,6 +271,13 @@ class Rewriter final: public WritableStreamSink {
 
   friend class ::workerd::api::HTMLRewriter;
 
+  class RegisteredReplacer {
```
anonrig, line 274: Can you add a comment to this class, explaining what it does etc.

src/workerd/api/html-rewriter.c++
```diff
@@ -578,6 +600,127 @@ kj::Promise<void> Rewriter::thunkPromise(CType* content, RegisteredHandler& regi
...
+    return kj::READY_NOW;
+  }
+
+  kj::Promise<void> end() override KJ_WARN_UNUSED_RESULT {
+    // Nothing specific needs to be done to tell lol_html we're done writing the stream
+    return kj::READY_NOW;
+  }
+
+  void abort(kj::Exception reason) override {
+    // TODO(npaun): Not sure what we can do... should we throw it again?
```
anonrig, line 651: This seems like a TODO that should be completed before merging this PR.

src/workerd/api/html-rewriter.c++
```diff
@@ -843,6 +1042,7 @@ jsg::Ref<Comment> Comment::after(Content content, jsg::Optional<ContentOptions>
 }
 
 jsg::Ref<Comment> Comment::replace(Content content, jsg::Optional<ContentOptions> options) {
+  // TODO(npaun): Seems like you can't streaming replace comments yet in lol-html
```
anonrig, line 1045: What is the actual TODO in here? Adding support for streaming replace comments?

src/workerd/api/html-rewriter.h
```diff
@@ -157,6 +157,21 @@ struct ContentOptions {
 
 class Rewriter;
 
+template <typename CType>
+struct BaseImpl {
```
anonrig, line 161: Can you add a comment to what this class does?

src/workerd/api/tests/htmlrewriter-test.js
```diff
@@ -610,6 +610,40 @@ export const manualWriting2 = {
...
+    // initiate the read after doing the writes.
+    const promise = response.text();
+
+    const rfc2616 = await (
+      await fetch('https://www.rfc-editor.org/rfc/rfc2616.txt')
+    ).text();
+    strictEqual(await promise, `<html>${rfc2616}bar</html>`);
+  },
+};
+
```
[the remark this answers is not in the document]
anonrig: Agree with @harrishancock. This needs more tests before landing it.

src/workerd/api/html-rewriter.c++
```diff
@@ -578,6 +596,124 @@ kj::Promise<void> Rewriter::thunkPromise(CType* content, RegisteredHandler& regi
...
+int Rewriter::replacerThunk(lol_html_streaming_sink_t* sink, void* userData) {
+  auto& registration = *reinterpret_cast<RegisteredReplacer*>(userData);
+  return registration.rewriter.replacerThunkImpl(sink, registration);
+}
+
+int Rewriter::replacerThunkImpl(
+    lol_html_streaming_sink_t* sink, RegisteredReplacer& registeredHandler) {
+  // TODO(npaun): This is basically the same as thunkImpl except we call replacerThunkPromise instead,
+  // and return an int (because that's what lol-html says to do)
+  // We could merge this with thunkImpl fairly easily if that's what we needed to do
```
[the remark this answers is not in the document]
npaun: I tried this locally but decided I didn't like it because there were too many template parameters.

src/workerd/api/html-rewriter.c++
```diff
@@ -578,6 +596,124 @@ kj::Promise<void> Rewriter::thunkPromise(CType* content, RegisteredHandler& regi
...
+    return kj::READY_NOW;
+  }
+
+  virtual kj::Promise<void> end() KJ_WARN_UNUSED_RESULT {
+    // Nothing specific needs to be done to tell lol_html we're done writing the stream
+    return kj::READY_NOW;
+  }
+
+  virtual void abort(kj::Exception reason) {
+    // TODO(npaun): Not sure what we can do... should we throw it again?
```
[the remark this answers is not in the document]
npaun: The test `streamingReplacementBadUTF8` covers this one. The control flow here gave me a huge headache but I think I get it now:

1. We try to write bad UTF-8 with lol_html_streaming_sink_write_utf8_chunk
2. It returns -1 which we convert to returning a KJ exception with message.
3. pumpTo will invoke abort(). I don't think we need to do anything specific in response to this.
4. replacerThunkImpl catches the exception and poisons the rewriter. This is just like thunkImpl for element handlers.
5. lol_html receives -1 from write_all_callback, and so it knows there was an error during the streaming replacement . 
6. This eventually bubbles up to the lol_html write failing

src/workerd/api/html-rewriter.h
```diff
@@ -157,6 +157,23 @@ struct ContentOptions {
 
 class Rewriter;
 
+// Each Token subclass has an inner BaseImpl subclass which holds a reference
+// to the rewriter, and the actual underlying lol-html C API handle for the token.
+template <typename CType>
+struct BaseImpl {
```
harrishancock, line 163: Nit: BaseImpl is quite a generic name to have in the workerd::api namespace. Can we nest this inside of HTMLRewriter::Token and call it ImplBase? (We typically name base classes FooBase, where Foo is our intended derived class name -- which in this case happens to be Impl, which is why I think that order makes more sense.)

src/workerd/api/html-rewriter.c++
```diff
@@ -579,6 +603,122 @@ kj::Promise<void> Rewriter::thunkPromise(CType* content, RegisteredHandler& regi
...
+// but the content must be valid UTF-8 or lol_html will reject it.
+class ReplacerStreamSink final: public WritableStreamSink {
+ public:
+  ReplacerStreamSink(lol_html_streaming_sink_t* sink, bool isHtml): sink(sink), isHtml(isHtml) {}
+
+  kj::Promise<void> write(kj::ArrayPtr<const byte> buffer) override KJ_WARN_UNUSED_RESULT {
+    auto err = lol_html_streaming_sink_write_utf8_chunk(
+        sink, buffer.asChars().begin(), buffer.size(), isHtml);
+    if (err != 0) {
+      return KJ_EXCEPTION(FAILED, "lol_html_streaming_sink_write_utf8_chunk failed");
```
harrishancock, line 630: We might want to just do `return getLastError()` here, and below.

src/workerd/api/tests/htmlrewriter-test.js
```diff
@@ -610,6 +610,157 @@ export const manualWriting2 = {
...
+    const encoder = new TextEncoder();
+
+    await writer.write(encoder.encode('<html>'));
+    await writer.write(encoder.encode('bar'));
+    await writer.write(encoder.encode('</html>'));
+    await writer.close();
+
+    // This variation uses the JavaScript TransformStream, so we can
+    // initiate the read after doing the writes.
+    // TODO(someday): Is it a good idea to pass on the specific message from lol_html? How can we do that?
```
harrishancock, line 759: I think we might get a friendly message for free by returning `getLastError()` from our `lol_html_streaming_sink_write_utf8_chunk()` call sites.

harrishancock (approved):

src/workerd/api/tests/htmlrewriter-test.js
npaun: This works. Neat!

PR #3212, opened 2024-12-05
JSRPC: Allow wrapping RPC targets in Proxy objects.

### LINKED ISSUE #3184: Feature request: RpcTarget wrapped in Proxy throws DataCloneError
An `RpcTarget` class instance that is wrapped in a `Proxy` throws the following runtime error when it is returned by an RPC call.

`DataCloneError: #<SimpleTarget> could not be cloned.`

```ts
export default class DefaultEntrypoint extends WorkerEntrypoint {
  override fetch(request: Request) {
    return Response.json({ hello: "world" });
  }

  // ✅ this is fine
  async workingRpcTarget() {
    return new SimpleTarget();
  }

  // ❌ this throws DataCloneError
  async brokenRpcTarget() {
    const target = new SimpleTarget();
    return new Proxy(target, {});
  }
}

class SimpleTarget extends RpcTarget {
  async targetCall() {
    return Promise.resolve({ hello: "world" });
  }
}
```

I found this while trying to extend our observability by capturing traces through `RpcTarget` instances. I have `WorkerEntrypoint` instrumented through `Proxy` thanks to the fix from #2291. The instrumentation is then checking if the return value of an RPC call is an instance of `RpcTarget` and attempts to wrap it in a `Proxy` to instrument those calls as well.

Could this be caused by the runtime failing to identify the `RpcTarget` inheritance of a `Proxy` and treating it as a plain object instead?

kentonv:
When serializing a value for RPC, if we encounter a `Proxy`, we will check if it claims to be an instance of `RpcTarget` (via the prototype chain) and, if so, treat it as one, wrapping it in a stub with the same semantics as an actual `RpcTarget`.

Additionally, if the `Proxy` claims to be a plain object (the prototype is `Object.prototype`), we will allow this too, wrapping it in a stub that allows access to the object's "own" properties. This is not exactly the same thing as we'd do if it were really a raw object: in that case, we would actually send the object content. However, it is consistent with our access control model, which says that for plain objects, "own" properties are exposed, but for `RpcTarget` instances, only prototype properties are exposed.

A `Proxy` can even be used to cause an object of class type that is _not_ derived from `RpcTarget` behave as if it were, by overriding `getPrototypeOf()` to return `RpcTarget.prototype`. This is a feature, not a bug! The reasons we don't normally support non-RpcTarget classes by default are (1) we want to reserve the right to add by-value serialization to existing API types in the future, without breaking compatibility, and (2) classes not explicitly written for RPC may not have been written with the understanding that their public API is a security boundary. However, if the developer uses a `Proxy` to fake that the object is an `RpcTarget`, then the developer is "opting in" to `RpcTarget` behavior, namely (1) the object is strictly pass-by-stub, and (2) they are comfortable with exposing the class's public interface.

Fixes #3184

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -303,6 +303,7 @@ http_archive(
         "//:patches/v8/0018-Return-rejected-promise-from-WebAssembly.compile-if-.patch",
         "//:patches/v8/0019-codegen-Don-t-pass-a-nullptr-in-InitUnwindingRecord-.patch",
         "//:patches/v8/0020-Add-another-slot-in-the-isolate-for-embedder.patch",
+        "//:patches/v8/0021-Add-ValueSerializer-SetTreatProxiesAsHostObjects.patch",
     ],
     strip_prefix = "v8-13.1.201.8",
     url = "https://github.com/v8/v8/archive/refs/tags/13.1.201.8.tar.gz",
--- a/patches/v8/0021-Add-ValueSerializer-SetTreatProxiesAsHostObjects.patch
+++ b/patches/v8/0021-Add-ValueSerializer-SetTreatProxiesAsHostObjects.patch
@@ -1,0 +1,119 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Kenton Varda <kenton@cloudflare.com>
+Date: Wed, 4 Dec 2024 22:36:05 -0600
+Subject: Add ValueSerializer::SetTreatProxiesAsHostObjects().
+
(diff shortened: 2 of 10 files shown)


danlapid:
This probably requires an internal-pr to add the patch right?

danlapid (approved):
Besides the comment on an internal pr LGTM

kentonv:
> This probably requires an internal-pr to add the patch right?

Yes, there is one.

jasnell (approved):

PR #3224, opened 2024-12-09
Add optional AlarmInvocationInfo parameter to Durable Object alarm() method

jclee:
The runtime passes in this parameter when invoking alarm(); the `retryCount` property gives a count of the number of previous times the runtime has tried to run this specific alarm event.

The runtime has been passing this parameter to `alarm()` calls since it was originally introduced earlier in the year, in PR #1673, but the parameter declaration has not yet been added to the worker-types TypeScript declarations.  This PR just updates the TypeScript declarations to include this parameter.

### DIFF
--- a/types/generated-snapshot/2021-11-03/index.d.ts
+++ b/types/generated-snapshot/2021-11-03/index.d.ts
@@ -467,7 +467,7 @@ interface Cloudflare {
 }
 interface DurableObject {
   fetch(request: Request): Response | Promise<Response>;
-  alarm?(): void | Promise<void>;
+  alarm?(alarmInfo?: AlarmInvocationInfo): void | Promise<void>;
   webSocketMessage?(
     ws: WebSocket,
     message: string | ArrayBuffer,
@@ -5312,7 +5312,7 @@ declare module "cloudflare:workers" {
     protected env: Env;
     constructor(ctx: DurableObjectState, env: Env);
     fetch?(request: Request): Response | Promise<Response>;
-    alarm?(): void | Promise<void>;
+    alarm?(alarmInfo?: AlarmInvocationInfo): void | Promise<void>;
     webSocketMessage?(
       ws: WebSocket,
       message: string | ArrayBuffer,
--- a/types/test/types/rpc.ts
+++ b/types/test/types/rpc.ts
@@ -306,6 +306,27 @@ class TestObject extends DurableObject {
   }
 }
 
+class TestAlarmObject extends DurableObject {
+  // Can declare alarm method consuming optional alarmInfo parameter
+  async alarm(alarmInfo?: AlarmInvocationInfo) {
+    if (alarmInfo !== undefined) {
+      const _isRetry: boolean = alarmInfo.isRetry;
+      const _retryCount: number = alarmInfo.retryCount;
+    }
+  }
+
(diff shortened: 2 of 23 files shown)


types/generated-snapshot/2021-11-03/index.d.ts
```diff
@@ -467,7 +467,7 @@ interface Cloudflare {
 }
 interface DurableObject {
   fetch(request: Request): Response | Promise<Response>;
-  alarm?(): void | Promise<void>;
+  alarm?(alarmInfo?: AlarmInvocationInfo): void | Promise<void>;
```
jclee, line 470: The changes to `types/generated-snapshot/*` were all generated by running the command recommended by a CI error message:
```
  The generated output of @cloudflare/workers-types has been changed by this
  PR. If this is intentional, run:
  bazel build //types && rm -rf types/generated-snapshot && cp -r bazel-bin/types/definitions types/generated-snapshot
```
I assume it's doing the right thing by regenerating these snapshot files, but I'm not familiar with the intent.

types/test/types/rpc.ts
```diff
@@ -306,6 +306,27 @@ class TestObject extends DurableObject {
   }
 }
 
+class TestAlarmObject extends DurableObject {
+  // Can declare alarm method consuming optional alarmInfo parameter
+  async alarm(alarmInfo?: AlarmInvocationInfo) {
+    if (alarmInfo !== undefined) {
```
jclee, line 312: FWIW, the current version of the runtime always passes this `alarmInfo` parameter, but early implementations did not, and it's likely that almost all Durable Object `alarm()` implementations in the wild do not currently declare the parameter.  I think that means that the parameter needs to be declared `alarmInfo?` for backwards compatibility, although it is a bit unfortunate that this requires TS DOs to do this check for `undefined` before using the parameter.

I'm not sure if there is a better way to declare the method in the TS `rpc.d.ts` file such that it allows class implementers to either consume a (non-undefined) `alarmInfo` parameter or to omit it entirely.

Probably not a big deal in the common case, though, since most implementers probably won't be writing an `alarm()` handler that needs the info in the parameter.

justin-mp: Dumb question: is there a way to set the `alarmInfo` as always passed as of a particular compatibility date?

jasnell (approved):

types/test/types/rpc.ts
jclee: Not dumb -- I was wondering myself, but am not familiar with how we are doing versioning and backwards compatibility of TypeScript declarations.

I assume the purpose of the various `types/generated-snapshot/*` files might be to allow for compatibility-breaking changes in types, but I'm not sure if there is a correspondence to compatibility dates.  I think marking the parameter as mandatory would not be backwards compatible, in the sense that it could cause TS to issue new errors or warnings for existing worker code -- either for `alarm()` handler definitions, or in cases where worker code tries to invoke `alarm()` handlers directly.

I suspect this feature will be infrequently used enough that it might not be worth breaking backwards compatibility for, by itself -- but if there comes a time when we're breaking backwards compatibility anyway, that might be a convenient time to make the parameter mandatory?

justin-mp (approved):

PR #3228, opened 2024-12-10
Implement the tail stream RPC interface

jasnell:
Implements the streaming tail worker RPC interface and `tailStream(...)` handler.

Note that this is just the implementation of the RPC interface. It does not yet include the various instrumentation points that would provide data to the interface.

Adds an autogate to enable the use of tail streams. 

On the tail worker side, simply export a `tailStream(...)` handler:

```js
export default {
  tailStream(onset) {
    return (event) => { };
  }
}
```

The `tailStream` function can return a function, an object, undefined or a promise for function, object, or undefined.

If undefined is returned, no more trace events will be delivered for this invocation. If a function is returned, it will receive all events. If an object is returned, it is expected to export functions that handle individual event types.

For instance,

```js
export default {
  tailStream(onset) {
    return {
      log(logEvent) {  },
      exception(exceptionEvent) {  },
    };
  }
}
```

On the producing side, a new `reportTailEvent(...)` is added to the `RequestObserver` API. 

```js
tracing::Onset onset(...);
ioContext.getMetrics().reportTailEvent(ioContext, kj::mv(onset));
```

This is intentionally a low-level API without many bells and whistles. Adding the actual instrumentation points will come separately, at which time we may make additional tweaks to the API.

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -1888,9 +1888,9 @@ class Server::WorkerService final: public Service,
         } else if (service->hasHandler("tail") || service->hasHandler("trace")) {
           legacyList.add(service->startRequest({}));
         }
-        legacyTailWorkers = legacyList.releaseAsArray();
-        streamingTailWorkers = streamingList.releaseAsArray();
       }
+      legacyTailWorkers = legacyList.releaseAsArray();
+      streamingTailWorkers = streamingList.releaseAsArray();
     } else {
       legacyTailWorkers = KJ_MAP(service, channels.tails) -> kj::Own<WorkerInterface> {
         KJ_ASSERT(service != this, "A worker currently cannot log to itself");


src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,937 @@
...
+          obj.set(js, EVENT_STR, ToJs(js, attrs, cache));
+        }
+      }
+    }
+  }
+
+  return obj;
+}
+
+class ServerTopLevelMembrane final: public capnp::MembranePolicy, public kj::Refcounted {
```
jasnell, line 574: This is adopted from the class of the same name over is worker-rpc.c++. It can, and probably should be deduplicated into a common utility but will do that separate.

src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,937 @@
...
+
+  kj::Own<MembranePolicy> addRef() override {
+    return kj::addRef(*this);
+  }
+
+ private:
+  kj::Maybe<kj::Own<kj::PromiseFulfiller<void>>> doneFulfiller;
+};
+
+class TailStreamTargetBase: public rpc::TailStreamTarget::Server {
```
jasnell, line 606: Review note: The TailStreamTargetBase forms the basis of the server-side of the tail stream RPC interface. It arranges to have the received tail events dispatched to the correct exported handler

src/workerd/server/server.c++
```diff
@@ -1653,50 +1653,80 @@ class Server::WorkerService final: public Service,
...
-        co_await worker->customEvent(kj::mv(event)).ignoreResult();
+    kj::Array<kj::Own<WorkerInterface>> legacyTailWorkers = nullptr;
+    kj::Array<kj::Own<WorkerInterface>> streamingTailWorkers = nullptr;
+    // If streaming tail workers is enabled, then we will initialize two lists:
+    // one with services that only export the tail or trace handler (legacy tail
+    // workers) and one that exports the tailStream handler. We'll check tailStreams
+    // first.
+    if (util::Autogate::isEnabled(util::AutogateKey::STREAMING_TAIL_WORKERS)) {
+      kj::Vector<kj::Own<WorkerInterface>> legacyList;
+      kj::Vector<kj::Own<WorkerInterface>> streamingList;
```
anonrig, line 1835: Depending on the size of `channels.tails` you can at least reserve the capacity for these vectors by `channels.tails.size() / 2`

src/workerd/io/observer.h
```diff
@@ -127,6 +128,20 @@ class RequestObserver: public kj::Refcounted {
     return nullptr;
   }
 
+  // If the worker is configured to support streaming tail workers, reportTailEvent
+  // will forward the given event on to the collection of streaming tail workers
+  // that are configured with this observer. Otherwise, this is a non-op.
+  virtual void reportTailEvent(IoContext& ioContext, tracing::TailEvent::Event&& event) {
```
anonrig, line 134: Q: How can a function be virtual and have an implementation at the same time?

src/workerd/io/observer.h
```diff
@@ -127,6 +128,20 @@ class RequestObserver: public kj::Refcounted {
...
+    reportTailEvent(ioContext, [event = kj::mv(event)]() mutable { return kj::mv(event); });
+  }
+
+  // If the worker is configured to support streaming tail workers, reportTailEvent
+  // will forward the event returned by the callback on to the collection of streaming
+  // fail workers that are configured with this observer. The callback will only be
+  // invoked if there are tail workers.
+  virtual void reportTailEvent(
+      IoContext& ioContext, kj::FunctionParam<tracing::TailEvent::Event()> fn) {}
```
anonrig, line 143: Should this be without `virtual` and the implementor just overrides it?

src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,821 @@
...
+      }
+      KJ_CASE_ONEOF(d, double) {
+        return js.num(d);
+      }
+      KJ_CASE_ONEOF(i, int32_t) {
+        return js.num(i);
+      }
+    }
+    KJ_UNREACHABLE;
+  };
```
anonrig, line 146: Since we are inside an anonymous namespace, we can just move this `ToJs` to a function in the anonymous namespace itself?

src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,821 @@
...
+  obj.set(js, "statusCode"_kj, js.num(info.statusCode));
+  return obj;
+}
+
+jsg::JsValue ToJs(jsg::Lock& js, const tracing::FetchEventInfo& info) {
+  auto obj = js.obj();
+  obj.set(js, "type"_kj, js.str("fetch"_kj));
+  obj.set(js, "method"_kj, js.str(kj::str(info.method)));
+  obj.set(js, "url"_kj, js.str(info.url));
+  obj.set(js, "cfJson"_kj, js.str(info.cfJson));
```
anonrig: Q: Is there a JSG API that allows creating an object with an array of object keys and values?

src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,821 @@
...
+jsg::JsValue ToJs(jsg::Lock& js, const EventOutcome& outcome) {
+  switch (outcome) {
+    case EventOutcome::OK:
+      return js.str("ok"_kj);
+    case EventOutcome::CANCELED:
+      return js.str("canceled"_kj);
+    case EventOutcome::EXCEPTION:
+      return js.str("exception"_kj);
+    case EventOutcome::KILL_SWITCH:
+      return js.str("kilSwitch"_kj);
```
anonrig: This should be `killSwitch`

src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,821 @@
...
+  return obj;
+}
+
+jsg::JsValue ToJs(jsg::Lock& js, const tracing::CustomEventInfo& info) {
+  auto obj = js.obj();
+  obj.set(js, "type"_kj, js.str("custom"_kj));
+  return obj;
+}
+
+jsg::JsValue ToJs(jsg::Lock& js, const EventOutcome& outcome) {
```
anonrig: All values can be defined in constexpr evaluation and be put in a kj::Map or something to remove the repetition (and copying)

src/workerd/server/server.c++
jasnell: eh, I'm not going to worry about it here. This is just the workerd local server at this point so it really doesn't need to be highly performance optimized. I think the internal server that uses this would benefit from a few more optimizations but here it's not that critical.

src/workerd/io/observer.h
jasnell: That how we allow for default implementations of virtual functions.

src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,821 @@
...
+
+jsg::JsValue ToJs(jsg::Lock& js, const tracing::TraceEventInfo& info) {
+  auto obj = js.obj();
+  obj.set(js, "type"_kj, js.str("trace"_kj));
+
+  auto names = KJ_MAP(trace, info.traces) -> jsg::JsValue {
+    KJ_IF_SOME(name, trace.scriptName) {
+      return js.str(name);
+    }
+    return js.null();
```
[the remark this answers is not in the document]
jasnell: It could be I prefer the readability the way it is.

src/workerd/util/completion-membrane.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
+
+#include <capnp/membrane.h>
+
+namespace workerd {
+
+// A membrane applied which detects when no capabilities are held any longer, at which point it
+// fulfills a fulfiller.
```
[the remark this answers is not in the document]
jasnell: To me these are the same. This is not a new comment, btw, this was just moved here from somewhere else.

src/workerd/io/trace-stream.c++
jasnell: Not currently. Can look at adding that later.

src/workerd/io/trace-stream.c++
jasnell: The `js.str(...)` values cannot be made constexpr. Those are js heap allocated strings.

src/workerd/util/BUILD.bazel
```diff
@@ -187,6 +187,13 @@ wd_cc_library(
     deps = ["@capnp-cpp//src/kj"],
 )
 
+wd_cc_library(
+    name = "completion-membrane",
+    hdrs = ["completion-membrane.h"],
+    visibility = ["//visibility:public"],
+    deps = ["@capnp-cpp//src/kj"],
```
fhanau, line 194: Nit: `<capnp/membrane.h>` is in `@capnp-cpp//src/capnp`, so that should be added too so that the target is self-sufficient/could be parsed on its own.

src/workerd/io/trace-stream.h
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  // TODO(streaming-tail-workers): Specify the correct type as specified in the
+  // internal capnp definition.
+  static constexpr uint16_t TYPE = 11;
+
+  rpc::TailStreamTarget::Client getCap() {
+    auto result = kj::mv(KJ_ASSERT_NONNULL(clientCap, "can only call getCap() once"));
+    clientCap = kj::none;
+    return result;  // Get the
```
fhanau, line 44: Nit: Comment appears incomplete

types/defines/trace.d.ts
```diff
@@ -0,0 +1,190 @@
+declare namespace TailStream {
```
fhanau, line 1: Non-blocking: Some of these are already defined in `src/workerd/api/trace.h`, do they need an explicit definition here too? Asking non-rhetorically, it looks like we have both `EmailEventInfo` and `TraceItemEmailEventInfo` now in the generated types and the only difference is having a `readonly type: "email"` variable, maybe we need both.

src/workerd/io/BUILD.bazel
```diff
@@ -50,6 +50,7 @@ wd_cc_library(
         "io-context.c++",
         "io-own.c++",
         "io-util.c++",
+        "trace-stream.c++",
```
fhanau, line 54: In the interest of not making the io target more bloated, this should be a separate target – just needs to depend on `:io` and completion-membrane, and the server needs to depend on the new target too.

src/workerd/server/server.c++
```diff
@@ -1447,9 +1449,183 @@ void Server::InspectorServiceIsolateRegistrar::registerIsolate(
 
 // =======================================================================================
 namespace {
+
+// The TailStreamWriterState holds the current client-side state for a collection
+// of streaming tail workers that a worker is reporting events to.
+struct TailStreamWriterState {
```
fhanau, line 1472: Non-blocking: Maybe there's some synergies with upstream here and this could live in a seperate file to avoid code duplication, that can be handled in the future though.

src/workerd/io/trace-stream.c++
```diff
@@ -0,0 +1,948 @@
...
+        .findOrCreate(value, [&]() -> decltype(cache)::Entry {
+      return {value, jsg::JsRef<jsg::JsValue>(js, js.strIntern(value))};
+    }).getHandle(js);
+  }
+
+ private:
+  kj::HashMap<kj::StringPtr, jsg::JsRef<jsg::JsValue>> cache;
+};
+
+// Why ToJS(...) functions and not JSG_STRUCT? Good question. The various tracing:*
```
fhanau, line 115: Thank you for explaining this! Makes it easier to understand that the boilerplate code here helps avoid interdependencies.

fhanau (approved):

types/defines/trace.d.ts
jasnell: The `TraceItemEmailEventInfo` variation is one that is automatically generated from the types `api/trace.h` which we're not using here. As we iterate on the API it's quite possible these will diverge more from the legacy schema/types. I'd rather keep them separate so that they can evolve independently.

src/workerd/server/server.c++
jasnell: Yep. Already in plan. I'll add a comment.

src/workerd/server/server.c++
```diff
@@ -1464,9 +1466,187 @@ void Server::InspectorServiceIsolateRegistrar::registerIsolate(
...
+  using Pending = kj::Array<kj::Own<WorkerInterface>>;
+
+  struct Active {
+    // Reference to keep the worker interface instance alive.
+    kj::Maybe<rpc::TailStreamTarget::Client> capability;
+
+    // Every active tail worker target will have a queue.
+    bool pumping = false;
+    bool onsetSeen = false;
+    std::list<tracing::TailEvent> queue;
```
fhanau, line 1490: Nit: This is referred to as a queue but std::list is used... would std::deque fit better here? (As you said performance is not important here at this stage, it's just confusing)

fhanau (approved):

fhanau:
LGTM otherwise.

PR #3229, opened 2024-12-10
Implements Response.replace_body and support for cf opts in Py fetch.

dom96:
While going through our examples I noticed that these were missing.

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -5,7 +5,7 @@ from collections.abc import Generator, Iterable, MutableMapping
 from contextlib import ExitStack, contextmanager
 from enum import StrEnum
 from http import HTTPMethod, HTTPStatus
-from typing import TypedDict, Unpack
+from typing import Any, TypedDict, Unpack
 
 import js
 
@@ -21,10 +21,32 @@ Body = "str | FormData | JSBody"
 Headers = dict[str, str] | list[tuple[str, str]]
 
 
+# https://developers.cloudflare.com/workers/runtime-apis/request/#the-cf-property-requestinitcfproperties
+class RequestInitCfProperties(TypedDict, total=False):
+    apps: bool | None
+    cacheEverything: bool | None
+    cacheKey: str | None
+    cacheTags: list[str] | None
+    cacheTtl: int
+    cacheTtlByStatus: dict[str, int]
+    image: (
+        Any | None
+    )  # TODO: https://developers.cloudflare.com/images/transform-images/transform-via-workers/
+    mirage: bool | None
+    polish: str | None
+    resolveOverride: str | None
+    scrapeShield: bool | None
+    webp: bool | None
+
+
+# This matches the Request options:
+# https://developers.cloudflare.com/workers/runtime-apis/request/#options
 class FetchKwargs(TypedDict, total=False):
     headers: Headers | None
     body: "Body | None"
     method: HTTPMethod = HTTPMethod.GET
+    redirect: str | None
+    cf: RequestInitCfProperties | None
 
 
 # TODO: Pyodide's FetchResponse.headers returns a dict[str, str] which means
@@ -65,6 +87,15 @@ class FetchResponse(pyodide.http.FetchResponse):
         except JsException as exc:
             raise _to_python_exception(exc) from exc
 
+    def replace_body(self, body: Body) -> "FetchResponse":
+        """
+        Returns a new Response object with the same options (status, headers, etc) as
+        the original but with an updated body.
+        """
+        b = body.js_object if isinstance(body, FormData) else body
+        js_resp = js.Response.new(b, self.js_response)
+        return FetchResponse(js_resp.url, js_resp)
+
 
 async def fetch(
     resource: str,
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/workers.py
```diff
@@ -117,15 +139,25 @@ def __init__(
...
+        status: HTTPStatus | int = HTTPStatus.OK,
+        status_text="",
+        headers: Headers = None,
     ):
         options = {
             "status": status.value if isinstance(status, HTTPStatus) else status,
         }
-        if len(statusText) > 0:
-            options["statusText"] = statusText
+        if len(status_text) > 0:
```
anonrig, line 159: This does the same thing afaict:

```suggestion
        if status_text:
```

danlapid: Explicit is always better than implicit

hoodmane: Well it's most common in python code to check if a container is non empty with `if container:`. I think you'll find this suggested in most style guides as the preferred way.

src/pyodide/internal/workers.py
```diff
@@ -117,15 +139,25 @@ def __init__(
         )
         super().__init__(js_resp.url, js_resp)
 
+    @classmethod
+    def from_response(cls, body: Body, response: "Response") -> "Response":
```
hoodmane, line 143: `from_response` is a bit of an odd name. Is there an existing api like this? I would think to make it a method called `replace_body`. e.g,
```
def replace_body(self, new_body: Body) -> "Response":
    """
    Returns a new Response object with the same headers but an updated body
    """"
```

src/pyodide/internal/workers.py
```diff
@@ -99,7 +121,7 @@ def __init__(
         self,
         body: Body,
         status: HTTPStatus | int = HTTPStatus.OK,
-        statusText="",
+        status_text="",
```
hoodmane, line 124: Formatting is inconsistent here, does our lint check not check formatting?

src/pyodide/internal/workers.py
```diff
@@ -117,15 +139,25 @@ def __init__(
         )
         super().__init__(js_resp.url, js_resp)
 
+    @classmethod
+    def from_response(cls, body: Body, response: "Response") -> "Response":
+        b = body.js_object if isinstance(body, FormData) else body
+        result = cls.__new__(cls)
+        js_resp = js.Response.new(b, response.js_object)
+        super(Response, result).__init__(js_resp.url, js_resp)
```
hoodmane, line 147: Can you explain the reasoning for this? Ideally we should switch to:
```suggestion
        js_resp = js.Response.new(b, response.js_object)
        return cls(js_resp.url, js_resp)
```
or `type(self)(js_resp.url, js_resp)` assuming we make it into an instance method. If this is different in some important way we'll need a comment. Particularly, I would expect `super(cls)` instead of `super(Response)` which skips the `__init__` functions of any subclasses.

src/workerd/server/tests/python/sdk/worker.py
```diff
@@ -51,8 +51,11 @@ async def on_fetch(request):
     elif request.url.endswith("/undefined_opts"):
         # This tests two things:
         #   * `Response.redirect` static method
-        #   * that other options can be passed into `fetch`
-        resp = await fetch("https://example.com/redirect", redirect="manual")
+        #   * that other options can be passed into `fetch` (so that we can support
+        #       new options without updating this code)
+        resp = await fetch(
+            "https://example.com/redirect", redirect="manual", foobarbaz=42
+        )
```
hoodmane, line 58: I guess we could test that we receive `foobarbaz` on the other side by mocking the JS fetch. In this case, we just test that we don't crash with this option but we could discard it without failing the test?

src/pyodide/internal/workers.py
dom96: This is emulating the `new Response(body, existing_resp)` pattern that a lot of Workers examples seem to use (https://developers.cloudflare.com/workers/runtime-apis/response/).

I don't have strong feelings about the name, so sure I can change it to your suggestion.

src/pyodide/internal/workers.py
dom96: IIRC I was doing it this way because the constructor of `Response` doesn't match that of `pyodide.http.FetchResponse`: https://github.com/cloudflare/workerd/blob/a99aa9788cb8067304a868fb5d7677e0c8318947/src/pyodide/internal/workers.py#L120

src/pyodide/internal/workers.py
dom96: it didn't seem to catch this at least, so I guess not

src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -30,6 +30,8 @@ py_wd_test("subdirectory")
 
 py_wd_test(
     "sdk",
+    # TODO: will need to rebuild bundle to change the `status_text` field name
```
hoodmane, line 33: Did you do this?

dom96: I was planning on merging this in first, but thinking about it again I suppose I can rebuild the bundle from this branch and update it in this PR

src/workerd/server/tests/python/sdk/worker.py
```diff
@@ -273,6 +307,23 @@ async def can_request_form_data_blob(env):
     assert text == "success"
 
 
+async def response_from_unit_tests(env):
```
hoodmane, line 310: Maybe update the name to `replace_body_unit_tests`?

PR #3230, opened 2024-12-10
[nfc] Roll Rust dependencies, clean up lol_html

fhanau:
This will enable/require some more cleanup in the upstream repo – they should be more closely aligned now.


anonrig (approved):

jasnell (approved):

fhanau:
I was wondering why we hadn't cleaned up cxxbridge earlier but this breaks on Windows as usual 🙄. Based on https://github.com/cloudflare/workerd-cxx/blob/master/tools/cargo/build.rs#L9 this is likely due to missing symlinks which we might be able to patch in but this is not important enough to fight Bazel about. I'll roll back that part of the change so that this can move forward.

mikea (approved):

PR #3231, opened 2024-12-10
Moar WritableStream memory optimization

jp4a50:
Make WritableStream controllers' `maybePendingAbort` members heap-allocated to reduce WritableStream's memory footprint when no abort is pending.

Reduces the footprint by a further ~100 bytes.


jasnell (approved):

anonrig (approved):
Q: How did you observe the memory usage?

PR #3233, opened 2024-12-11
add ability to create v8 objects more efficiently

anonrig:
Adds the ability to leverage a faster v8::Object creation

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -2608,6 +2608,8 @@ class Lock {
   JsSymbol symbolShared(kj::StringPtr) KJ_WARN_UNUSED_RESULT;
   JsSymbol symbolInternal(kj::StringPtr) KJ_WARN_UNUSED_RESULT;
   JsObject obj() KJ_WARN_UNUSED_RESULT;
+  JsObject obj(
+      kj::ArrayPtr<kj::StringPtr> keys, kj::ArrayPtr<jsg::JsValue> values) KJ_WARN_UNUSED_RESULT;
   JsObject objNoProto() KJ_WARN_UNUSED_RESULT;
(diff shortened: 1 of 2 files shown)


src/workerd/jsg/jsg.h
```diff
@@ -2608,6 +2608,8 @@ class Lock {
   JsSymbol symbolShared(kj::StringPtr) KJ_WARN_UNUSED_RESULT;
   JsSymbol symbolInternal(kj::StringPtr) KJ_WARN_UNUSED_RESULT;
   JsObject obj() KJ_WARN_UNUSED_RESULT;
+  JsObject obj(
+      kj::ArrayPtr<kj::StringPtr> keys, kj::ArrayPtr<kj::StringPtr> values) KJ_WARN_UNUSED_RESULT;
```
jasnell, line 2612: I would make the values an array of jsg::JsValue instead of requiring them to be strings.

erikcorry (approved):
This looks fine, but it's still going to be a lot slower than using an ObjectTemplate.  This routine looks up a bunch of strings in the V8 canonicalized string table on every object creation.

jasnell:
> This looks fine, but it's still going to be a lot slower than using an ObjectTemplate ...

This does make me wonder if our current uses of `JSG_STRUCT` could be improved by making use of `ObjectTemplate`. Right now the `StructWrapper` just calls `v8::Object::New(...)` and sets each field individually, which is fine if we're only going to create a couple of instances of a given object but if the thing is being created frequently it might benefit from use of `ObjectTemplate`. Not critical for now but something we might consider later.

jasnell (approved):

erikcorry:
Which following code?

I think ObjectTemplate is only feasible if we statically know the names of the object fields.  Creating a template for a one-time or rare object shape won't be worth it either for the programmer or in terms of runtime efficiency.

PR #3235, opened 2024-12-12
Add Windows Server 2022 Desktop Packer templates

harrishancock:
This commit adds a tools/devboxes/windows-2022 directory, which contains Packer templates and related scripts to aid in building a QEMU VM running a Windows Server 2022 Desktop guest OS, intended to help test workerd on Windows, when you do not have a Windows machine handy. At present, this has only been tested on a Linux host.

This was originally based on https://github.com/StefanScherer/packer-windows/blob/main/windows_2022.json, then heavily modified.

The build is split up into two separate Packer builds: an installation build, and a provisioning build which works on a copy-on-write image backed by the installation build's output. This is done so that working on the provisioning step doesn't require you to complete a full reinstallation of Windows first.

### DIFF
--- a/tools/devboxes/README.md
+++ b/tools/devboxes/README.md
@@ -1,0 +1,32 @@
+# Devboxes
+
+This directory contains tools to create virtual machine images to help you test workerd on different machines/OSes.
+
+Right now there is only a Windows devbox.
+
+## Pre-requisites
+
+This directory has been tested with:
+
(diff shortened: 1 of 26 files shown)


harrishancock:
I see we now have the https://github.com/cloudflare/workerd-tools/ repo. I wonder if that's a better place for this.

PR #3240, opened 2024-12-13
Undo tightening of rules around default entrypoints in #3193.

kentonv:
Say you have a module-syntax Worker that has no `export default`. You then refer to that worker from elsewhere in the workerd config, but you do not specify an entrypoint. What should happen?

In the midst of other changes, PR #3193 turned this into an error, becaues it seemed like obviously it should be. This turns out to have broken Miniflare tests, which often use an empty string as a placeholder Worker that is referenced but never invoked.

So, this PR restores the old behavior.

It turns out, incidentally, that the old behavior is crazier than you'd imagine. See the comments in the test case for details.

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -1630,7 +1630,19 @@ class Server::WorkerService final: public Service,
       name = entry.key;  // replace with more-permanent string
       handlers = &entry.value;
     } else {
-      handlers = &KJ_UNWRAP_OR_RETURN(defaultEntrypointHandlers, kj::none);
+      KJ_IF_SOME(d, defaultEntrypointHandlers) {
+        handlers = &d;
+      } else {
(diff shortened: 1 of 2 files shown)


danlapid (approved):

anonrig (approved):

src/workerd/server/server.c++
```diff
@@ -1630,7 +1630,19 @@ class Server::WorkerService final: public Service,
       name = entry.key;  // replace with more-permanent string
       handlers = &entry.value;
     } else {
-      handlers = &KJ_UNWRAP_OR_RETURN(defaultEntrypointHandlers, kj::none);
+      KJ_IF_SOME(d, defaultEntrypointHandlers) {
+        handlers = &d;
+      } else {
+        // It would appear that there is no default export, therefore this refers to an entrypoint
```
anonrig, line 1636: Should we log here, to inform the developer?

PR #3241, opened 2024-12-13
Add support for range queries in vectorize

garrettgu10:


src/cloudflare/internal/test/vectorize/vectorize-api-test.js
```diff
@@ -238,6 +238,30 @@ export const test_vector_search_vector_query = {
       };
       assert.deepStrictEqual(results, expected);
     }
+
+    {
+      const results = await IDX.query(new Float32Array(new Array(5).fill(0)), {
```
anonrig, line 243: Float32Array constructor takes a length argument and makes the default values 0 by default.

```suggestion
      const results = await IDX.query(new Float32Array(5), {
```

anonrig (approved):

PR #3242, opened 2024-12-13
Introducing better Workers AI types

thatsKevinJain:
With growing a growing list of supported models and their capabilities, we are making it simpler to infer input/output types from model names.

From users perspective, they will still be able to call `AI.run("<model-name>", {<inputs>}, {<options>})`, but we are modifying the type definitions to make it easier for future updates.

### DIFF
--- a/types/defines/ai.d.ts
+++ b/types/defines/ai.d.ts
@@ -50,10 +50,10 @@ export declare abstract class BaseAiSentenceSimilarity {
   inputs: AiSentenceSimilarityInput;
   postProcessedOutputs: AiSentenceSimilarityOutput;
 }
-export type AiSpeechRecognitionInput = {
+export type AiAutomaticSpeechRecognitionInput = {
   audio: number[];
 };
-export type AiSpeechRecognitionOutput = {
+export type AiAutomaticSpeechRecognitionOutput = {
   text?: string;
   words?: {
     word: string;
@@ -62,9 +62,9 @@ export type AiSpeechRecognitionOutput = {
   }[];
   vtt?: string;
 };
-export declare abstract class BaseAiSpeechRecognition {
-  inputs: AiSpeechRecognitionInput;
-  postProcessedOutputs: AiSpeechRecognitionOutput;
+export declare abstract class BaseAiAutomaticSpeechRecognition {
+  inputs: AiAutomaticSpeechRecognitionInput;
+  postProcessedOutputs: AiAutomaticSpeechRecognitionOutput;
 }
 export type AiSummarizationInput = {
   input_text: string;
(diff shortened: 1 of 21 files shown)


thatsKevinJain:

thatsKevinJain:

thatsKevinJain:
recheck

types/defines/ai.d.ts
```diff
@@ -148,15 +169,33 @@ export declare abstract class BaseAiTextGeneration {
...
+  width?: number;
   image?: number[];
+  image_b64?: string;
   mask?: number[];
   num_steps?: number;
   strength?: number;
   guidance?: number;
+  seed?: number;
 };
-export type AiTextToImageOutput = ReadableStream<Uint8Array>;
```
Cherry, line 159: It looks like this was intentionally changed in https://github.com/cloudflare/workerd/pull/2757 because that's what the API returned. Is this no longer the case?

types/defines/ai.d.ts
```diff
@@ -100,16 +100,31 @@ export declare abstract class BaseAiTextEmbeddings {
   postProcessedOutputs: AiTextEmbeddingsOutput;
 }
 export type RoleScopedChatInput = {
-  role: "user" | "assistant" | "system" | "tool";
+  role: "user" | "assistant" | "system" | "tool" | string;
```
Cherry, line 103: `"user" | "assistant" | "system" | "tool" | string` is equivalent to just `string`.

I would recommend:
```suggestion
  role: "user" | "assistant" | "system" | "tool" | (string & {});
```

This is a well-used TypeScript quirk that allows any string value, but doesn't reduce the entire type to just `string`, therefore allowing autocomplete of the common roles.

types/defines/ai.d.ts
```diff
@@ -100,16 +100,31 @@ export declare abstract class BaseAiTextEmbeddings {
...
+        type: string;
+        description?: string;
+      };
+    };
+    required: string[];
+  };
 };
 export type AiTextGenerationToolInput = {
-  type: "function";
+  type: "function" | string;
```
Cherry, line 122:
```suggestion
  type: "function" | (string & {});
```

types/defines/ai.d.ts
```diff
@@ -100,16 +100,31 @@ export declare abstract class BaseAiTextEmbeddings {
...
   function: {
     name: string;
     description: string;
     parameters?: {
-      type: "object";
+      type: "object" | string;
```
Cherry, line 127:
```suggestion
      type: "object" | (string & {});
```

types/defines/ai.d.ts
```diff
@@ -100,16 +100,31 @@ export declare abstract class BaseAiTextEmbeddings {
...
-  role: "user" | "assistant" | "system" | "tool";
+  role: "user" | "assistant" | "system" | "tool" | string;
   content: string;
+  name?: string;
+};
+export type AiTextGenerationToolLegacyInput = {
+  name: string;
+  description: string;
+  parameters?: {
+    type: "object" | string;
```
Cherry, line 111:
```suggestion
    type: "object" | (string & {});
```

types/defines/ai.d.ts
thatsKevinJain: This is great, TIL, updated it in the new commit!

mar-cf:
@G4brym could you please look this over?

G4brym (approved):

G4brym:
lgtm, I've been working with Kevin on these changes

mar-cf (approved):

muhajirdev:
is this released yet? trying to use llama 3.3

PR #3252, opened 2024-12-17
fix: Unknown encoding when undefined is used

mcandeia:
The behavior of `buffer.toString(encoding)` is not compliant with nodejs API. it throws an error `UNKNOWN_ENCODING: null` (or undefined) because it's being passed as a string "undefined" or "null"

### DIFF
--- a/src/node/internal/internal_utils.ts
+++ b/src/node/internal/internal_utils.ts
@@ -42,10 +42,12 @@ export function normalizeEncoding(enc?: string): Encoding | undefined {
     enc === 'UTF-8'
   )
     return UTF8;
-  return slowCases(enc);
+  return getEncodingOps(enc);
 }
 
-export function slowCases(enc: unknown): Encoding | undefined {
+export function getEncodingOps(enc: unknown): Encoding | undefined {
+  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+  enc += '';
   // @ts-expect-error TS18046 TS complains about unknown can not have length.
   switch (enc.length) {
     case 4:
--- a/src/node/internal/internal_utils.ts
+++ b/src/node/internal/internal_utils.ts
@@ -36,7 +36,6 @@ const { UTF8, UTF16LE, HEX, ASCII, BASE64, BASE64URL, LATIN1 } = bufferUtil;
 export function normalizeEncoding(enc?: string): Encoding | undefined {
   if (
     enc == null ||
-    enc === undefined ||
     enc === 'utf8' ||
     enc === 'utf-8' ||
     enc === 'UTF8' ||
--- a/src/node/internal/internal_utils.ts
+++ b/src/node/internal/internal_utils.ts
@@ -26,10 +26,10 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-import type { Encoding } from 'node-internal:buffer';
 import { default as bufferUtil } from 'node-internal:buffer';
-import { ERR_FALSY_VALUE_REJECTION } from 'node-internal:internal_errors';
+import type { Encoding } from 'node-internal:buffer';
 import { validateFunction } from 'node-internal:validators';
+import { ERR_FALSY_VALUE_REJECTION } from 'node-internal:internal_errors';
 
 const { UTF8, UTF16LE, HEX, ASCII, BASE64, BASE64URL, LATIN1 } = bufferUtil;
 
--- a/src/node/internal/internal_utils.ts
+++ b/src/node/internal/internal_utils.ts
@@ -26,16 +26,17 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-import { default as bufferUtil } from 'node-internal:buffer';
 import type { Encoding } from 'node-internal:buffer';
-import { validateFunction } from 'node-internal:validators';
+import { default as bufferUtil } from 'node-internal:buffer';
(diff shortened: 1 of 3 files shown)


src/node/internal/internal_utils.ts
```diff
@@ -26,16 +26,17 @@
...
+import { default as bufferUtil } from 'node-internal:buffer';
 import { ERR_FALSY_VALUE_REJECTION } from 'node-internal:internal_errors';
+import { validateFunction } from 'node-internal:validators';
 
 const { UTF8, UTF16LE, HEX, ASCII, BASE64, BASE64URL, LATIN1 } = bufferUtil;
 
 export function normalizeEncoding(enc?: string): Encoding | undefined {
   if (
     enc == null ||
+    enc === undefined ||
```
anonrig, line 39: Line 38 already checks for null and undefined. This line is not needed.

mcandeia: Oh, fixed, I missed the `==` thing, thanks

anonrig (approved):

mcandeia:
@anonrig fixed a test that was failing due to this expected behavior: https://github.com/cloudflare/workerd/blob/main/src/workerd/api/node/tests/buffer-nodejs-test.js#L5803

mcandeia:
@anonrig can you help me to decide one thing?

The NodeJs implementation does not fallback for `null` encoding - leading into a `TypeError` as this test covers (https://github.com/cloudflare/workerd/blob/main/src/workerd/api/node/tests/buffer-nodejs-test.js#L5823)

This test was failing before because of the wrong conversion to `null` (as a string) but now it isn't failing anymore because it is well-covered by the `encoding?.toString()`.

The nodejs implementation on the other hand does the following:

1. Buffer.prototype.toString implementation uses getEncodingOpts when `encoding` is `null`
2. https://github.com/nodejs/node/blob/a73c41c51e7711086b9801afd5e91ae2f96a7694/lib/buffer.js#L722C10-L722C24
3. it concats `null` with `''` https://github.com/nodejs/node/blob/a73c41c51e7711086b9801afd5e91ae2f96a7694/lib/buffer.js#L723
4. which results into a `undefined` value with length of `4` (omg)
image

5. The switch case will not match with any option https://github.com/nodejs/node/blob/a73c41c51e7711086b9801afd5e91ae2f96a7694/lib/buffer.js#L725
6. then it returns undefined and only then throws an error: https://github.com/nodejs/node/blob/a73c41c51e7711086b9801afd5e91ae2f96a7694/lib/buffer.js#L863


The options we have:

1. Remove the test case and let undefined and null have the same behavior
2. make sure null encoding is not allowed and explicitly throw an error

anonrig:
Hi @mcandeia,

Thank you for the awesome comment. I love it! Here's a suggestion to fix the issue:

```
diff --git a/src/node/internal/internal_buffer.ts b/src/node/internal/internal_buffer.ts
index cfde61ea..f57ba153 100644
--- a/src/node/internal/internal_buffer.ts
+++ b/src/node/internal/internal_buffer.ts
@@ -28,7 +28,10 @@ import {
   isUint8Array,
 } from 'node-internal:internal_types';
 
-import { normalizeEncoding } from 'node-internal:internal_utils';
+import {
+  normalizeEncoding,
+  getEncodingOps,
+} from 'node-internal:internal_utils';
 
 import { validateString } from 'node-internal:validators';
 
@@ -648,7 +651,7 @@ Buffer.prototype.toString = function toString(
     return '';
   }
 
-  const normalizedEncoding = normalizeEncoding(encoding?.toString());
+  const normalizedEncoding = getEncodingOps(encoding);
   if (normalizedEncoding === undefined) {
     throw new ERR_UNKNOWN_ENCODING(`${encoding}`);
   }
diff --git a/src/node/internal/internal_utils.ts b/src/node/internal/internal_utils.ts
index 7290f4f5..3a80b921 100644
--- a/src/node/internal/internal_utils.ts
+++ b/src/node/internal/internal_utils.ts
@@ -42,10 +42,12 @@ export function normalizeEncoding(enc?: string): Encoding | undefined {
     enc === 'UTF-8'
   )
     return UTF8;
-  return slowCases(enc);
+  return getEncodingOps(enc);
 }
 
-export function slowCases(enc: unknown): Encoding | undefined {
+export function getEncodingOps(enc: unknown): Encoding | undefined {
+  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+  enc += '';
   // @ts-expect-error TS18046 TS complains about unknown can not have length.
   switch (enc.length) {
     case 4:
diff --git a/src/workerd/api/node/tests/buffer-nodejs-test.js b/src/workerd/api/node/tests/buffer-nodejs-test.js
index 3dc88c3a..09c56f80 100644
--- a/src/workerd/api/node/tests/buffer-nodejs-test.js
+++ b/src/workerd/api/node/tests/buffer-nodejs-test.js
@@ -5815,7 +5815,8 @@ export const toStringRange = {
       },
       {
         name: 'TypeError',
-      }
+      },
+      'toString() with 0 and null as the encoding should have thrown'
     );
 
     throws(
@@ -5824,7 +5825,8 @@ export const toStringRange = {
       },
       {
         name: 'TypeError',
-      }
+      },
+      'toString() with null encoding should have thrown'
     );
   },
 };

```

anonrig:
By the way, I forgot. Can you add a test so we don't regress in the future? Thank you.

anonrig:
It seems there is a formatting issue on [warn] src/workerd/api/node/tests/buffer-nodejs-test.js

mcandeia:
fixed @anonrig

anonrig (approved):

mcandeia:
ok, had to fix it again @anonrig

anonrig (approved):

mcandeia:
Do you have any idea why these tests are failing @anonrig ?

mcandeia:
for the windows one:

image

mcandeia:
ping @anonrig

PR #3256, opened 2024-12-18
[o11y] Add remaining R2 span tags

fhanau:
Please consult the (internal) span definitions to review the added span tags. Input is welcome on API design for `r2GetClient()` and `getHttpClientWithCallback()`.

### DIFF
--- a/src/workerd/api/r2-bucket.c++
+++ b/src/workerd/api/r2-bucket.c++
@@ -24,6 +24,21 @@
 #include <regex>
 
 namespace workerd::api::public_beta {
+kj::Own<kj::HttpClient> r2GetClient(
+    IoContext& context, uint subrequestChannel, R2UserTracing user) {
+  kj::Vector<Span::Tag> tags;
+  tags.add("rpc.service"_kjc, kj::str("r2"_kjc));
+  tags.add(user.method.key, kj::str(user.method.value));
+  KJ_IF_SOME(b, user.bucket) {
+    tags.add("cloudflare.r2.bucket"_kjc, kj::str(b));
+  }
+  KJ_IF_SOME(tag, user.extraTag) {
+    tags.add(tag.key, kj::str(tag.value));
+  }
+
+  return context.getHttpClientWithSpans(subrequestChannel, true, kj::none, user.op, kj::mv(tags));
+}
+
 static bool isWholeNumber(double x) {
   double intpart;
   return modf(x, &intpart) == 0;
@@ -341,7 +356,8 @@ jsg::Promise<kj::Maybe<jsg::Ref<R2Bucket::HeadResult>>> R2Bucket::head(jsg::Lock
   return js.evalNow([&] {
     auto& context = IoContext::current();
 
-    auto client = context.getHttpClient(clientIndex, true, kj::none, "r2_get"_kjc);
+    auto client = r2GetClient(context, clientIndex,
+        {"r2_get"_kjc, {"rpc.method"_kjc, "GetObject"_kjc}, this->adminBucketName()});
 
     capnp::JsonCodec json;
     json.handleByAnnotation<R2BindingRequest>();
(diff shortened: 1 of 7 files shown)


src/workerd/io/io-context.h
```diff
@@ -744,6 +744,13 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
       kj::ConstString operationName,
       std::initializer_list<SpanTagParams> tags);
 
+  // As above, but with callback to add span tags to the trace context for the operation.
+  kj::Own<kj::HttpClient> getHttpClientWithCallback(uint channel,
```
mikea, line 748: not very big fan of callback version here.  why can't we make `WithSpans` work for us? (or maybe extend it?)

src/workerd/api/r2-bucket.c++
```diff
@@ -24,6 +24,25 @@
 #include <regex>
 
 namespace workerd::api::public_beta {
+kj::Own<kj::HttpClient> r2GetClient(IoContext& context,
+    uint subrequestChannel,
+    kj::LiteralStringConst op,
```
mikea, line 29: I feel that defining a simple `struct UserTracing` or something would go a long way for readability above (due to `.method = "r2_delete"_kjc`) and below because of maybe casts?

src/workerd/io/io-context.h
fhanau: A couple of thoughts here:
- We can't use the existing WithSpans() here since it takes an initializer list. Initializer lists are designed as a very lightweight interface, we can't append to them which makes them very inconvenient when there is a variable number of span tags.
- I pushed a PR that adds a WithSpans() version that takes a kj::Vector. The current approach is templated so that we still have the initializer list version available (which has less overhead, more syntactic sugar potential), but it looks like we might not even need the old version. Will push another change...

src/workerd/api/r2-bucket.c++
fhanau: To make sure I understand correctly – I assume you mean a struct like this:
```
struct UserTracing {
  kj::LiteralStringConst op;
  kj::LiteralStringConst method;
  kj::Maybe<kj::StringPtr> bucket;
  kj::Maybe<StringTagParams> extraTag;
};
```

Then we could pass parameters simply as `{"r2_get"_kjc, "GetObject"_kjc, this->adminBucketName()}`, but you prefer using this instead?
```
{
.op = "r2_get"_kjc,
.method = "GetObject"_kjc,
.bucket = this->adminBucketName()
}
```

FWIW I was able to clean up the Maybe\<StringTagParams> wrappers in the last push.

src/workerd/io/io-context.h
fhanau: Updated, much cleaner now and not using a callback.

src/workerd/io/io-context.h
```diff
@@ -91,6 +91,12 @@ class WarningAggregator final: public kj::AtomicRefcounted {
   kj::MutexGuarded<kj::Vector<kj::Own<WarningContext>>> warnings;
 };
 
+// Used to add span tags within user tracing.
+struct SpanTagParams {
+  kj::LiteralStringConst key;
+  kj::String value;
```
fhanau, line 97: Will be replaced with `Span::TagValue` in the next PR.

src/workerd/api/r2-bucket.c++
```diff
@@ -729,7 +750,20 @@ jsg::Promise<void> R2Bucket::delete_(jsg::Lock& js,
     const jsg::TypeHandler<jsg::Ref<R2Error>>& errorType) {
   return js.evalNow([&] {
     auto& context = IoContext::current();
-    auto client = context.getHttpClient(clientIndex, true, kj::none, "r2_delete"_kjc);
+    auto deleteKey = [&]() {
```
mikea, line 753: it is weird to store lambda in the case. I suggest you call it immediately: 
```
kj::String deleteKey = [&](){ ... }();
```

(all because we don't have case expressions)

src/workerd/api/r2-bucket.c++
mikea: I would prefer the version with names spelled out, yes. It is much easier to read and understand how `r2_get` is different from `GetObject`

mikea (approved):

jasnell (approved):

fhanau:
Merged via #3261.

PR #3260, opened 2024-12-19
Use std::list instead of std::deque for all buffers/queues in streams

jp4a50:
API implementation objects.

A very common use case is to create one or more stream API objects per request. These streams currently have a high native memory overhead even when they are empty. Switching to std::list (which doesn't pre-allocate space for items) significantly alleviates this problem.

For example, now ReadableStreams tend to have <50% of their original memory footprint.

### DIFF
--- a/src/workerd/api/streams/compression.c++
+++ b/src/workerd/api/streams/compression.c++
@@ -8,8 +8,8 @@
 
 #include <workerd/io/features.h>
 
-#include <deque>
 #include <iterator>
+#include <list>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


danlapid (approved):
LGTM, but I don't know the streams codebase well enough.
Up to you on whether to wait for James with this or not.

jasnell:
Not sure but I think it's worth the temporary performance hit now and we can revisit to see if more changes are needed here.

jasnell (approved):

PR #3261, opened 2024-12-20
[o11y] Add KV span tags

fhanau:
#3256 needs to land first.


src/workerd/api/kv.c++
```diff
@@ -105,19 +154,25 @@ kj::Own<kj::HttpClient> KvNamespace::getHttpClient(IoContext& context,
   return client;
 }
 
-jsg::Promise<KvNamespace::GetResult> KvNamespace::get(jsg::Lock& js,
-    kj::String name,
-    jsg::Optional<kj::OneOf<kj::String, GetOptions>> options,
-    CompatibilityFlags::Reader flags) {
+jsg::Promise<KvNamespace::GetResult> KvNamespace::get(
+    jsg::Lock& js, kj::String name, jsg::Optional<kj::OneOf<kj::String, GetOptions>> options) {
   return js.evalNow([&] {
```
fhanau, line 159: Not entirely sure if I'm handling async behavior correctly here? With the get/getWithMetadata implementation being moved to getWithMetadataImpl(), is the js.evalNow() closure still needed here? Should there be one added to getWithMetadata()?

js.evalNow

src/workerd/api/kv.h
```diff
@@ -50,10 +50,8 @@ class KvNamespace: public jsg::Object {
   using GetResult = kj::Maybe<
       kj::OneOf<jsg::Ref<ReadableStream>, kj::Array<byte>, kj::String, jsg::JsRef<jsg::JsValue>>>;
 
-  jsg::Promise<GetResult> get(jsg::Lock& js,
-      kj::String name,
-      jsg::Optional<kj::OneOf<kj::String, GetOptions>> options,
-      CompatibilityFlags::Reader flags);
```
fhanau, line 56: Minor cleanup: compat flags parameter is not needed here, that should simplify the generated code slightly.

src/workerd/api/r2-bucket.c++
```diff
@@ -439,7 +460,8 @@ jsg::Promise<kj::Maybe<jsg::Ref<R2Bucket::HeadResult>>> R2Bucket::put(jsg::Lock&
     });
 
     auto& context = IoContext::current();
-    auto client = context.getHttpClient(clientIndex, true, kj::none, "r2_put"_kjc);
+    auto client = r2GetClient(context, clientIndex, "r2_put"_kjc, "PutObject"_kjc,
```
mikea, line 463: I still think that calling the original function with initializer in-place wouldn't increase code amount much (maybe couple lines in total), but would be much more readable:

```
auto client = r2GetClient(context, clientIndex, { { "rpc.method"_kjc, "PutObject"_kjc} ... }
```

because currrently it is very hard to tell what is `r2_put` and what is `PutObject` and what is the difference.

mikea (approved):
sequences of very similar literals are not the best, otherwise lg.

jasnell (approved):

PR #3268, opened 2024-12-24
Add compatibility date for Pyodide 0.27.1

hoodmane:


src/workerd/io/compatibility-date.capnp
```diff
@@ -673,4 +673,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   tailWorkerUserSpans @69 :Bool
       $compatEnableFlag("tail_worker_user_spans")
       $experimental;
+
+  pythonWorkersV2 @70 :Bool
+      $compatEnableFlag("python_workers_v2")
+      $experimental
+      $pythonSnapshotRelease(pyodide = "0.26.4", pyodideRevision = "2024-12-18",
+          packages = "2024-12-18", backport = 0,
+          baselineSnapshotHash = "TODO");
```
danlapid, line 682: TODO

src/workerd/io/compatibility-date.capnp
```diff
@@ -673,4 +673,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   tailWorkerUserSpans @69 :Bool
       $compatEnableFlag("tail_worker_user_spans")
       $experimental;
+
+  pythonWorkersV2 @70 :Bool
```
danlapid, line 677: Again the V2 looks wrong to me, I don't know if that's the way we should go down.
We'll eventually have v2, v3, v4, v5 ... but that's not very descriptive of what they are and why the user should use one or the other.

hoodmane: I was thinking that we'll just add a bunch of `$impliedByAfterDate`'s and the user passes `python_workers` and a date.

hoodmane: In any case, I think it's okay to change the name as long as it's marked experimental. I also don't think "pyodide v0.26.4" means that much to the end users. Calendar versioning would be reasonable like `pythonWorkers_2024_12`.

src/workerd/io/compatibility-date.capnp
hoodmane: I wasn't expecting to do it in this PR.

src/workerd/io/compatibility-date.capnp
danlapid: Yeah calendar versioning seems better than pyodide versioning, I agree the pyodide version doesn't tell the user much

hoodmane: My only reluctance is that I don't see any compat flags with underscores in their names. But camel case doesn't really work for dates...

danlapid: We could just let the compatDate dictate the python version and not let user's choose.
So that you don't have a disabledByDate and you can't set the python_workers_v2 flag
It's just that getPyodideVersion is using the python_workers flag AND the compat date and we don't update this file at all.

hoodmane: Yeah I think that's the intent. The internal tests work by setting the CompatFlags directly but for the external tests I guess we could adjust the date instead of using an `enableFlag`. But we have to have a date in order to do that, for the pre-release ones we probably need an enable flag.

danlapid: I think I see what you mean.
Alright so let's document this above the python_workers flag.
And let's  rename the flag to pythonWorkers20241218.
And we'll document that this
1. pyodide versions are updated by pythonWorkersYYYYMMDD flags
2. These flags will remain experimental forever
3. Users update pyodide version by updating their compat date which will enable the flag via the impliedByAfterDate
4. Anything else?

hoodmane: Does `experimental` just stop the user from setting the flag directly with the enable flag and still allow it to be set by implies_by_after_date?

hoodmane: Anyways, if that doesn't work we can always remove both `experimental` and the `enable_flag` once we add an `impliedByAfterDate`.

danlapid: > Does experimental just stop the user from setting the flag directly
yes.

> still allow it to be set by implies_by_after_date?
I don't know, but if not then we need another approach here.

danlapid: > doesn't work we can always remove both `experimental`

Yeah that would work, add that to the documentation above

danlapid (approved):

jasnell (approved):

src/workerd/io/compatibility-date.capnp
dom96: The contract for `experimental` flags is that only accounts with a @cloudflare.com email can use those flags. Even if we can get around that mechanism by using `impliesByAfterDate` we shouldn't do so (and if that's possible then arguably we should fix it to avoid accidents). We should remove `experimental` and `enableFlag` instead as Hood suggested.

dom96 (approved):

PR #3278, opened 2024-12-28
Updated types for Workers AI

thatsKevinJain:
We are moving types for supported models under a new NPM package (https://www.npmjs.com/package/@cloudflare/ai-types).

This PR removes all the model specific types and converts AI.run() into a generic function.

### DIFF
--- a/types/defines/ai.d.ts
+++ b/types/defines/ai.d.ts
@@ -379,6 +379,26 @@ export declare abstract class Base_Ai_Cf_Openai_Whisper_Large_V3_Turbo {
   inputs: Ai_Cf_Openai_Whisper_Large_V3_Turbo_Input;
   postProcessedOutputs: Ai_Cf_Openai_Whisper_Large_V3_Turbo_Output;
 }
+export interface Ai_Cf_Black_Forest_Labs_Flux_1_Schnell_Input {
+  /**
+   * A text description of the image you want to generate.
+   */
+  prompt: string;
+  /**
+   * The number of diffusion steps; higher values can improve quality but take longer.
+   */
+  steps?: number;
+}
+export interface Ai_Cf_Black_Forest_Labs_Flux_1_Schnell_Output {
+  /**
+   * The generated image in Base64 format.
+   */
+  image?: string;
+}
+export declare abstract class Base_Ai_Cf_Black_Forest_Labs_Flux_1_Schnell {
+  inputs: Ai_Cf_Black_Forest_Labs_Flux_1_Schnell_Input;
+  postProcessedOutputs: Ai_Cf_Black_Forest_Labs_Flux_1_Schnell_Output;
+}
 export type Ai_Cf_Meta_Llama_3_2_11B_Vision_Instruct_Input = Prompt | Messages;
 export interface Prompt {
   /**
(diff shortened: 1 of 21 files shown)


celso (approved):

Cherry (commented):
While I implore the effort to continually improve Workers AI types, these changes would result in multiple major breaking changes to types of a GA product, as well as significant regressions in DX. I also have lots of other questions/concerns.

## Removal of `AiTextGenerationInput`, etc.

It is _very_ common in applications to build out your model inputs like this:

```
const input: AiTextGenerationInput = {...};
```
This lets you then make dynamic changes to them while adhering to types before passing to `AI.run()`, and prevents any human issues of typo'ing things like `messages`, etc. 
After this lands, this (and many other things like it) will immediately type-error when I upgrade my project to the latest `workers-types`, or run `npm update`.

## DX issues

### Types

With the above, making the `AI.run()` function a generic `any` essentially breaks all of the hard work from previous, and even the most recent fixes that landed just a short while ago with https://github.com/cloudflare/workerd/pull/3242.

Now, unless I realise that this breaking change has landed and needs me to install another package, I could pass anything to this and TS won't warn me of issues until I hit the runtime. Things that would have previously been caught before I deploy, now may break my production application because I don't have type warnings.

### Onboarding

Has this been communicated with the C3/create-cloudflare team? When initiating Workers, `workers-types` is installed and the `tsconfig.json` configured as appropriate. It sounds like additional steps are going to be needed here now for any AI workers.

### Usability Downgrade

Previously, `env.AI.run` was all I needed to get accurate types. I simply had to define my `env` which is handled mostly automagically for you today with `c3` and `wrangler`, and that was it - I had all the types I needed.

Now I have to do something like this:
```ts
import { AiModelTypes } from "@cloudflare/ai-types"
const ai = env.AI as Ai<AiModelTypes>;
```
which diverges from all other products, and adds additional complexity.

## Visibility

`workers-types` is a staple in the Cloudflare Workers ecosystem. It's open source, encourages so many third-party contributions, handles compatibility date changes, and even added recent snapshot checking to reduce regressions and other issues in https://github.com/cloudflare/workerd/pull/3188.

- Will the `@cloduflare/ai-types` package be OSS?
- Will it encourage third party contributions?
- Will it have a similar approach to versioning and regression handling?
- Where can people get support for it and/or report issues? This repo still?

## Conclusion

I would encourage you to reconsider this approach. It introduces a lot of DX regressions, and major breaking changes to the types of a GA product. If this is to land in the current state, I only hope it's at bare minimum in a new major version of `workers-types` so it's very clear to people when upgrading that there are breaking changes.

types/defines/ai.d.ts
```diff
@@ -1,291 +1,23 @@
...
-  "@cf/meta/llama-3.1-8b-instruct-fp8": BaseAiTextGeneration;
-  "@cf/meta/llama-3.1-8b-instruct-awq": BaseAiTextGeneration;
-  "@cf/meta/llama-3.2-3b-instruct": BaseAiTextGeneration;
-  "@cf/meta/llama-3.2-1b-instruct": BaseAiTextGeneration;
-  "@cf/meta/llama-3.3-70b-instruct-fp8-fast": BaseAiTextGeneration;
-  "@cf/meta/m2m100-1.2b": BaseAiTranslation;
-  "@cf/facebook/bart-large-cnn": BaseAiSummarization;
-  "@cf/unum/uform-gen2-qwen-500m": BaseAiImageToText;
-  "@cf/llava-hf/llava-1.5-7b-hf": BaseAiImageToText;
+  [key: string]: any;
```
penalosa, line 8: This shouldn't be `any`—at the very least can it be `Record<string, unknown>`?

penalosa (changes requested):
This is a (large) breaking change to workers-types, which will require additional changes across the release process

thatsKevinJain:
These were very valid points @Cherry , we have decided to continue adding all models to this repo instead of creating a new npm package. I have added new commits to add new models in the list as well.

thatsKevinJain:
This PR also adds the flux-schnell model with it's own custom types, so we can close this duplicate PR: https://github.com/cloudflare/workerd/pull/3314

types/defines/ai.d.ts
thatsKevinJain: Reverted the breaking changes and added new models to the list.

penalosa (approved):

PR #3280, opened 2024-12-30
Implement cross-language await syntax for KJ Promises and Rust Futures

harrishancock:
Associated pull request(s):
- https://github.com/capnproto/capnproto/pull/2216
- https://github.com/capnproto/capnproto/pull/2220
- https://github.com/capnproto/capnproto/pull/2223


danlapid:
Would it be reasonable to request a README.md on the src/rust/async directory explaining what we're doing, the approach and maybe pointers as to how to understand the code? Which part interacts with which and in what way?
Was trying to read it and while I have some context it's hard to grok how everything interacts 😢

harrishancock:
Definitely! I won't ask for review until code comments are fleshed out, and I have an explainer written.

jasnell:
The c++ parts are looking good. One concern I do have is this is a fairly large complicated bit of code... would it make sense / be possible for it to be split out into a separate dependency outside of workerd?

harrishancock:
@jasnell Yes, I agree. I think the capnproto repo is the most logical home for most of this, maybe in kj/compat. Or a separate repo, call it kj-rs, could also make sense.

I'll keep it here in draft form for now while I iterate, but will think about how to support this in capnproto's build.

PR #3286, opened 2025-01-02
Update WPT (use workerd-tools archive)

npaun:
We're tired of unstable hashes for the WPT archive when downloaded as a tarball from Github, so I created an action in cloudflare/workerd-tools#4 to automatically generate a WPT tarball for us on demand. As a bonus, we can exclude unused files from the archive which should speed up bazel.


src/workerd/api/wpt/urlpattern-test.js
```diff
@@ -4,37 +4,40 @@
...
-    'Component: protocol Left: {"protocol":"a"} Right: {"protocol":"b"}',
-    'Component: username Left: {"username":"a"} Right: {"username":"b"}',
-    'Component: password Left: {"password":"a"} Right: {"password":"b"}',
-    'Component: hostname Left: {"hostname":"a"} Right: {"hostname":"b"}',
-    'Component: search Left: {"search":"a"} Right: {"search":"b"}',
-    'Component: hash Left: {"hash":"a"} Right: {"hash":"b"}',
-  ],
-});
+export const urlpatternCompareTests = run(
+  'urlpattern-compare-tests.tentative.js',
```
npaun, line 8: The only actual change here is adding ".tentative." to the filename to match WPT's changes. The rest is auto-formatting noise.

fhanau (approved):

npaun:
Ugh this is annoying. The internal PR required thingy won't go away.

npaun:
Nope... we're now requiring all comments to be resolved before merging, so I have to close my own comment. y tho!?!?

PR #3294, opened 2025-01-06
Fix lifetimes in sendTracesToExportedHandler

danlapid:


anonrig (approved):
Is it easy to add a test to this? (So we don't regress)

danlapid:
> Is it easy to add a test to this? (So we don't regress)

Not really, that's why we get these so often.

PR #3298, opened 2025-01-07
Separate commonjs-related types from jsg/modules.h/c++

jasnell:
Further cleanup in preparation for finishing the new module registry.

### DIFF
--- a/src/workerd/jsg/commonjs.c++
+++ b/src/workerd/jsg/commonjs.c++
@@ -1,0 +1,163 @@
+#include "commonjs.h"
+
+#include "modules.h"
+
+namespace workerd::jsg {
+
+v8::Local<v8::Value> CommonJsModuleContext::require(jsg::Lock& js, kj::String specifier) {
+  auto modulesForResolveCallback = getModulesForResolveCallback(js.v8Isolate);
+  KJ_REQUIRE(modulesForResolveCallback != nullptr, "didn't expect resolveCallback() now");
+
+  if (isNodeJsCompatEnabled(js)) {
+    KJ_IF_SOME(nodeSpec, checkNodeSpecifier(specifier)) {
+      specifier = kj::mv(nodeSpec);
+    }
+  }
+
+  kj::Path targetPath = ([&] {
+    // If the specifier begins with one of our known prefixes, let's not resolve
+    // it against the referrer.
+    if (specifier.startsWith("node:") || specifier.startsWith("cloudflare:") ||
+        specifier.startsWith("workerd:")) {
+      return kj::Path::parse(specifier);
+    }
+    return path.parent().eval(specifier);
+  })();
+
+  // require() is only exposed to worker bundle modules so the resolve here is only
+  // permitted to require worker bundle or built-in modules. Internal modules are
+  // excluded.
+  auto& info = JSG_REQUIRE_NONNULL(modulesForResolveCallback->resolve(js, targetPath, path,
+                                       ModuleRegistry::ResolveOption::DEFAULT,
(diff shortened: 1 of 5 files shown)


src/workerd/jsg/commonjs.c++
```diff
@@ -0,0 +1,163 @@
...
+  auto modulesForResolveCallback = getModulesForResolveCallback(js.v8Isolate);
+  KJ_REQUIRE(modulesForResolveCallback != nullptr, "didn't expect resolveCallback() now");
+
+  if (isNodeJsCompatEnabled(js)) {
+    KJ_IF_SOME(nodeSpec, checkNodeSpecifier(specifier)) {
+      specifier = kj::mv(nodeSpec);
+    }
+  }
+
+  kj::Path targetPath = ([&] {
```
anonrig, line 17: Can we move this function outside? this way it won't get inlined, and I assume this will run a lot, right?

src/workerd/jsg/commonjs.c++
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  // TODO(cleanup): This implementation from here on is identical to the
+  // CommonJsModuleContext::require. We should consolidate these as the
+  // next step.
+
+  auto modulesForResolveCallback = jsg::getModulesForResolveCallback(js.v8Isolate);
+  KJ_REQUIRE(modulesForResolveCallback != nullptr, "didn't expect resolveCallback() now");
+
+  kj::Path targetPath = ([&] {
```
anonrig, line 84: This function is duplicate. We can reuse it if we move it outside.

src/workerd/jsg/commonjs.c++
```diff
@@ -0,0 +1,163 @@
...
+  JSG_REQUIRE(value->IsObject(), TypeError, "Invalid node:process implementation");
+  return value;
+}
+
+kj::String NodeJsModuleContext::getFilename() {
+  return path.toString(true);
+}
+
+kj::String NodeJsModuleContext::getDirname() {
+  return path.parent().toString(true);
```
anonrig, line 132: What if parent does not exist, and we are on root? How does this behave?

src/workerd/jsg/commonjs.h
```diff
@@ -0,0 +1,159 @@
...
+  v8::Local<v8::Value> getProcess(jsg::Lock& js);
+
+  // TODO(soon): Implement setImmediate/clearImmediate
+
+  jsg::Ref<NodeJsModuleObject> getModule(jsg::Lock& js);
+
+  v8::Local<v8::Value> getExports(jsg::Lock& js);
+  void setExports(jsg::Value value);
+
+  kj::String getFilename();
```
anonrig, line 134: Should this be const? If yes, I assume next line should be updated as well. 

```suggestion
  kj::String getFilename() const;
```

src/workerd/jsg/commonjs.c++
jasnell: For this PR I'm intentionally just moving things around, not changing the current impl. I'm planning some other changes here that will come soon in a different PR

src/workerd/jsg/commonjs.c++
jasnell: For this PR I'm intentionally just moving things around, not changing the current impl. I'm planning some other changes here that will come soon in a different PR

src/workerd/jsg/commonjs.c++
jasnell: I'm actually hoping to be able to remove this entire class soon so for now I don't want to make actual changes to this implementation. More changes are coming soon. I'd prefer to keep this focused only on moving the existing impl

anonrig (approved):

PR #3299, opened 2025-01-07
Clean up to asgi.py

hoodmane:
Remove some logs that Mike pointed out, also don't always raise a runtime error and remove fastapi import

### DIFF
--- a/src/pyodide/internal/asgi.py
+++ b/src/pyodide/internal/asgi.py
@@ -2,8 +2,6 @@ from asyncio import Future, Queue, ensure_future, sleep
 from contextlib import contextmanager
 from inspect import isawaitable
 
-from fastapi import Depends, Request
-
 ASGI = {"spec_version": "2.0", "version": "3.0"}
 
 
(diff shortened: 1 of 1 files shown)


src/pyodide/internal/asgi.py
```diff
@@ -217,3 +210,16 @@ async def fetch(app, req, env):
...
     return await process_websocket(app, req)
+
+
+def __getattr__(name):
+    if name == "env":
+        from fastapi import Depends, Request
+
+        @Depends
+        async def env(request: Request):
+            return request.scope["env"]
```
hoodmane, line 221: I guess we should move this into the sdk @dom96...

dom96: Hmm, what is the purpose of this? Isn't it fastapi-specific?

dom96 (approved):

src/pyodide/internal/asgi.py
hoodmane: Yes.

hoodmane: Sequestering it in the `__getattr__` function makes it so that as long as nobody tries to access `env`, it's possible to use asgi without importing fastapi.

PR #3301, opened 2025-01-07
Adds Sphinx config to generate Python SDK docs.

dom96:
Usage:

```
# set up venv with sphinx and sphinx-markdown-builder installed
make markdown
```

This will generate something like https://gist.github.com/dom96/4ce2594239d4672262eefed992402563.


danlapid (commented):
I don’t have any strong or smart opinions on how to generate docs for this.
I’ll delegate to @hoodmane to review this 🙏

src/pyodide/internal/sphinx/README.md
```diff
@@ -0,0 +1,15 @@
+## Python SDK docs markdown generator
+
+This directory contains config files for Sphinx, which can generate docs based on the exported methods in asgi.py and workers.py. These files are a part of the Python SDK.
+
+### Usage
+
+
+```
+# set up venv with sphinx and sphinx-markdown-builder installed
```
hoodmane, line 9: Could you add a `requirements-doc.txt` so that this can say:
```
python -m venv .venv
source .venv/bin/activate
pip install -r requirements-doc.txt
make markdown
```

src/pyodide/internal/sphinx/conf.py
```diff
@@ -0,0 +1,38 @@
...
+project = 'Python Workers API'
+copyright = '2025, Cloudflare'
+author = 'Cloudflare'
+
+# -- General configuration ---------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
+
+extensions = [
+    'sphinx.ext.autodoc',
+    'sphinx_markdown_builder'
```
hoodmane, line 18: We might want to add `sphinx.ext.autosummary` (can make a table with links to the individual documentation entries) and `myst_parser` so we can write the source files in markdown rather than rst. Also, `sphinx.ext.intersphinx` would be helpful, particularly if we can add xref links to cloudflare docs and mdn.
https://github.com/pyodide/pyodide/blob/main/docs/sphinx_pyodide/sphinx_pyodide/mdn_xrefs.py

None of these things have to happen in this PR though.

hoodmane:
sphinx-markdown-builder doesn't look super well maintained but it gets some commits. It would be nice if we could generate html directly and use that, but it wouldn't look like the rest of the cloudflare docs. I'm interested to see if this looks okay when rendered. Anyways, if we start out with this approach and find something else we like better later then we can just change it.

hoodmane (approved):

PR #3303, opened 2025-01-08
Save HibernatableWebSocketEventInfo events by value

danlapid:
These events are so small...
```C++
  struct Message final {};
  struct Close final {
    uint16_t code;
    bool wasClean;
  };
  struct Error final {};
```
That there shouldn't even be any performance penalty to saving them by value rather than by reference.
The nice thing about it is we don't need to save bare references that can easily cause UB or worse (such as a bug we're trying hopelessly to debug atm).

### DIFF
--- a/src/workerd/api/trace.h
+++ b/src/workerd/api/trace.h
@@ -455,11 +455,11 @@ class TraceItem::HibernatableWebSocketEventInfo final: public jsg::Object {
   class Error;
 
   explicit HibernatableWebSocketEventInfo(
-      const Trace& trace, const tracing::HibernatableWebSocketEventInfo::Message& eventInfo);
+      const Trace& trace, const tracing::HibernatableWebSocketEventInfo::Message eventInfo);
   explicit HibernatableWebSocketEventInfo(
-      const Trace& trace, const tracing::HibernatableWebSocketEventInfo::Close& eventInfo);
+      const Trace& trace, const tracing::HibernatableWebSocketEventInfo::Close eventInfo);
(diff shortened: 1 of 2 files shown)


anonrig:
> The nice thing about it is we don't need to save bare references that can easily cause UB or worse (such as a bug we're trying hopelessly to debug atm).

100 percent. I raise you and say: "not unless we use shared_ptr"

src/workerd/api/trace.h
```diff
@@ -534,7 +532,7 @@ class TraceItem::HibernatableWebSocketEventInfo::Error final: public jsg::Object
   }
 
  private:
-  const tracing::HibernatableWebSocketEventInfo::Error& eventInfo;
+  const tracing::HibernatableWebSocketEventInfo::Error eventInfo;
```
anonrig, line 535: Should we add a static assertion of the size of these structs? If in the future, it changes and increases significantly this will have a performance impact due to copying.

danlapid:
> > The nice thing about it is we don't need to save bare references that can easily cause UB or worse (such as a bug we're trying hopelessly to debug atm).
> 
> 100 percent. I raise you and say: "not unless we use shared_ptr"

Yeah shared ptr is great but for these structures specifically it’s probably cheaper to just copy them 😅

anonrig:
> Yeah shared ptr is great but for these structures specifically it’s probably cheaper to just copy them 😅

Or I can raise one more time and mention reference counting through kj. (I'm so happy to show off my KJ knowledge. Mom and dad are proud)

anonrig (approved):

PR #3304, opened 2025-01-08
Run Python tests under asan with leak san disabled.

dom96:
These tests are currently failing on shutdown due to a memory leak. This PR re-enables them on ASAN without the leak sanitizer to get some coverage under ASAN for them.

### DIFF
--- a/build/wd_test.bzl
+++ b/build/wd_test.bzl
@@ -63,6 +63,9 @@ def _wd_test_impl(ctx):
     # can't just specify some other executable with some args. OK, fine, we'll use a script that
     # just execs its args.
     if is_windows:
+        if len(ctx.attr.env):
+            fail("`env` attribute to _wd_test not supported on Windows")
+
         # Batch script executables must end with ".bat"
         executable = ctx.actions.declare_file("%s_wd_test.bat" % ctx.label.name)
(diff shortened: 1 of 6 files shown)


build/wd_test.bzl
```diff
@@ -63,6 +63,9 @@ def _wd_test_impl(ctx):
     # can't just specify some other executable with some args. OK, fine, we'll use a script that
     # just execs its args.
     if is_windows:
+        if len(ctx.attr.env):
```
fhanau, line 66: This will just make the Windows build fail right? Instead, you could use `target_compatible_with` to mark targets as incompatible with Windows if env is set

fhanau:
Looks like the tests are just segfaulting instead unfortunately.

build/wd_test.bzl
fhanau: Well, just passing env on non-Windows is probably easier and avoids just skipping the tests on Windows (assuming we won't have windows-asan in the future).

dom96:
Huh, that's interesting. It doesn't happen when run via `bazel test @workerd//src/cloudflare/internal/test/vectorize:python-vectorize-api-test_0.26.0a2@ --config=asan`

PR #3305, opened 2025-01-08
Fixes to external package loading: use external lock and load stdlib.

dom96:
Ready for review. Still needs a new bundle generated which I will do after review.

----

Implements changes to use the external package lock (rather than the built-in one) and to fix how we load the stdlib when external package loading is enabled.

### DIFF
--- a/src/pyodide/internal/loadPackage.ts
+++ b/src/pyodide/internal/loadPackage.ts
@@ -18,6 +18,7 @@ import {
 } from 'pyodide-internal:metadata';
 import {
   SITE_PACKAGES,
+  STDLIB_PACKAGES,
   getSitePackagesPath,
 } from 'pyodide-internal:setupPackages';
 import { parseTarInfo } from 'pyodide-internal:tar';
@@ -113,7 +114,12 @@ async function loadPackagesImpl(
   let loadPromises: Promise<[string, Reader]>[] = [];
   let loading = [];
   for (const req of requirements) {
-    if (SITE_PACKAGES.loadedRequirements.has(req)) continue;
+    if (req === 'test') {
+      continue; // Skip the test package, it is only useful for internal Python regression testing.
+    }
+    if (SITE_PACKAGES.loadedRequirements.has(req)) {
+      continue;
+    }
     loadPromises.push(loadBundle(req).then((r) => [req, r]));
     loading.push(req);
   }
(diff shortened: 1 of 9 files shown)


src/pyodide/internal/loadPackage.ts
```diff
@@ -135,9 +141,10 @@ async function loadPackagesImpl(
 }
 
 export async function loadPackages(Module: Module, requirements: Set<string>) {
+  const pkgsToLoad = requirements.union(new Set(STDLIB_PACKAGES));
```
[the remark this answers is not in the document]
hoodmane: They should really add an `addAll` function so that we don't need to create an extra temporary to add all of the elements of an iterable.

hoodmane (approved):

src/pyodide/internal/setupPackages.ts
```diff
@@ -23,17 +23,21 @@ function canonicalizePackageName(name: string): string {
...
 
 // The "name" field in the lockfile is not canonicalized
-const STDLIB_PACKAGES: string[] = Object.values(LOCKFILE.packages)
+export const STDLIB_PACKAGES: string[] = Object.values(LOCKFILE.packages)
   .filter(({ install_dir }) => install_dir === 'stdlib')
   .map(({ name }) => canonicalizePackageName(name));
 
+// Each item in the list is an element of the file path, for example
+// `folder/file.txt` -> `["folder", "file.txt"]
+export type FilePath = string[];
```
hoodmane, line 32: Thanks for documenting this.

src/pyodide/internal/snapshot.ts
```diff
@@ -102,6 +103,46 @@ function loadDynlib(
...
+    } else if (path.length == 1 && path[0] == sslFile) {
+      hasSsl = true;
+    } else {
+      result.push(path);
+    }
+  }
+
+  // JS might handle sorting lists of lists fine, but I'd rather be explicit here and make it compare
+  // strings.
+  result.sort((a, b) => a.join('/').localeCompare(b.join('/')));
```
hoodmane, line 135: This is a really poor performance sort since it creates new strings on every comparison. Either do:
```js
result.map(x => x.join('/')).sort(...).map(x => x.split('/'));
```
or even better performance would be to use:
```js
function pathCompare(p, q) {
  for (let idx = 0; idx < p.length && x < q.length; idx++) {
    let cmp = p[idx].localeCompare(q[idx]);
    if (cmp) {
      return cmp;
    }
  }
  return p.length - q.length;
}
result.sort(pathCompare);
```

PR #3311, opened 2025-01-09
Provide mechanism for autopopulating node.js process.env

jasnell:
Autopopulates the process.env from bindings in local dev. A similar PR will be needed internally to enable it there as it won't be automatic.

See caveats: https://github.com/cloudflare/workerd/pull/3311#issuecomment-2586475148

### DIFF
--- a/src/node/internal/util.d.ts
+++ b/src/node/internal/util.d.ts
@@ -120,6 +120,7 @@ export function isBoxedPrimitive(
   value: unknown
 ): value is number | string | boolean | bigint | symbol;
 
+export function getEnvObject(): Record<string, string>;
 export function getBuiltinModule(id: string): any;
 export function getCallSites(frames?: number): Record<string, string>[];
 export function processExitImpl(code: number): void;
--- a/src/workerd/api/node/tests/process-nodejs-test.js
+++ b/src/workerd/api/node/tests/process-nodejs-test.js
@@ -6,3 +6,59 @@ export const processPlatform = {
     assert.ok(['darwin', 'win32', 'linux'].includes(process.platform));
   },
 };
+
+process.env.QUX = 1;
+const pEnv = { ...process.env };
+
+export const processEnv = {
+  async test(ctrl, env) {
+    assert.notDeepStrictEqual(env, process.env);
+
+    assert.strictEqual(pEnv.FOO, 'BAR');
+
+    // It should be possible to mutate the process.env at runtime.
+    // All values manually set are coerced to strings.
+    assert.strictEqual(pEnv.QUX, '1');
+
+    // JSON bindings that do not parse to strings come through as
+    // raw unparsed JSON strings....
+    assert.strictEqual(pEnv.BAR, '{}');
+    JSON.parse(pEnv.BAR);
+
+    // JSON bindings that parse as strings come through as the
+    // parsed string value.
+    assert.strictEqual(pEnv.BAZ, 'abc');
+
+    // Throws because althought defined as a JSON binding, the value
+    // that comes through to process.env in this case is not a JSON
+    // parseable string because it will already have been parsed.
+    // This assertion is only true when the JSON bindings value
+    // happens to parse as a string.
+    assert.throws(() => JSON.parse(pEnv.BAZ));
+
+    // JSON bindings that parse as strings that happen to be double
+    // escaped might be JSON parseable.
+    assert.strictEqual(JSON.parse(pEnv.DUB), 'abc');
+
+    // Test that imports can see the process.env at the top level
+    const { FOO } = await import('mod');
+    assert.strictEqual(FOO, 'BAR');
+
+    // Mutating the env argument does not change process.env, and
+    // vis versa. The reason for this is that just mutations may
+    // be entirely surprising and unexpected for users.
+    assert.strictEqual(env.QUX, undefined);
+    env.ZZZ = 'a';
+    assert.strictEqual(process.env.ZZZ, undefined);
+
+    env.FOO = ['just some other value'];
+    assert.strictEqual(process.env.FOO, 'BAR');
+
+    // Other kinds of bindings will be on env but not process.env
+    assert.strictEqual(pEnv.NON, undefined);
+    assert.deepStrictEqual(
+      new Uint8Array(env.NON),
+      new Uint8Array([97, 98, 99, 100, 101, 102])
+    );
+  },
+};
--- a/src/workerd/api/node/tests/process-nodejs-test.wd-test
+++ b/src/workerd/api/node/tests/process-nodejs-test.wd-test
@@ -5,10 +5,21 @@ const unitTests :Workerd.Config = (
     ( name = "nodejs-process-test",
       worker = (
         modules = [
-          (name = "worker", esModule = embed "process-nodejs-test.js")
+          (name = "worker", esModule = embed "process-nodejs-test.js"),
+          (name = "mod", esModule = "export const { FOO } = process.env;")
         ],
-        compatibilityDate = "2024-10-11",
-        compatibilityFlags = ["nodejs_compat"],
+        compatibilityDate = "2024-12-28",
+        compatibilityFlags = [
+          "nodejs_compat",
+          "nodejs_compat_populate_process_env"
+        ],
+        bindings = [
+          (name = "FOO", text = "BAR"),
+          (name = "BAR", json = "{}"),
+          (name = "BAZ", json = "\"abc\""),
+          (name = "DUB", json = "\"\\\"abc\\\"\""),
+          (name = "NON", data = "abcdef")
+        ],
       )
     ),
   ],
--- a/src/workerd/api/node/util.c++
+++ b/src/workerd/api/node/util.c++
@@ -241,6 +241,10 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
   return js.undefined();
 }
 
+jsg::JsObject UtilModule::getEnvObject(jsg::Lock& js) {
+  return js.getEnv(true);
+}
+
 namespace {
 [[noreturn]] void handleProcessExit(jsg::Lock& js, int code) {
   // There are a few things happening here. First, we abort the current IoContext
--- a/src/workerd/api/node/util.h
+++ b/src/workerd/api/node/util.h
@@ -244,6 +244,8 @@ class UtilModule final: public jsg::Object {
     return processPlatform;
   }
 
+  jsg::JsObject getEnvObject(jsg::Lock& js);
+
   JSG_RESOURCE_TYPE(UtilModule) {
     JSG_NESTED_TYPE(MIMEType);
     JSG_NESTED_TYPE(MIMEParams);
@@ -263,6 +265,9 @@ class UtilModule final: public jsg::Object {
     JSG_METHOD(previewEntries);
     JSG_METHOD(getConstructorName);
     JSG_METHOD(getCallSites);
+    // TODO(cleanup): It might be about time to separate some of these out
+    // to a different module.
+    JSG_METHOD(getEnvObject);
 
 #define V(Type) JSG_METHOD(is##Type);
     JS_UTIL_IS_TYPES(V)
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -714,4 +714,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # tasks to complete.
   # This intentionally doesn't have a compatEnableDate yet until so we can let some users opt-in to
   # try it before enabling it for all new scripts, but will eventually need one.
+
+  populateProcessEnv @76 :Bool
+      $compatEnableFlag("nodejs_compat_populate_process_env")
+      $compatDisableFlag("nodejs_compat_do_not_populate_process_env")
+      $impliedByAfterDate(name = "nodeJsCompat", date = "2025-04-01");
+  # Automatically populate process.env from text bindings only
+  # when nodejs_compat is being used.
 }
--- a/src/workerd/jsg/setup.h
+++ b/src/workerd/jsg/setup.h
@@ -267,6 +267,9 @@ class IsolateBase {
   // object with 2 internal fields.
   v8::Global<v8::FunctionTemplate> opaqueTemplate;
 
+  // Object that is used as the underlying target of process.env when nodejs-compat mode is used.
+  v8::Global<v8::Object> envObj;
+
   // Polyfilled Symbol.asyncDispose.
   v8::Global<v8::Symbol> symbolAsyncDispose;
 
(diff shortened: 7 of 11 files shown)


src/workerd/server/workerd-api.c++
```diff
@@ -626,6 +626,9 @@ static v8::Local<v8::Value> createBindingValue(JsgWorkerdIsolate::Lock& lock,
     KJ_CASE_ONEOF(json, Global::Json) {
       v8::Local<v8::String> string = lock.wrap(context, kj::mv(json.text));
       value = jsg::check(v8::JSON::Parse(context, string));
+      if (featureFlags.getPopulateProcessEnv() && featureFlags.getNodeJsCompat()) {
+        lock.setEnvField(lock.str(global.name), jsg::JsValue(string));
```
kentonv, line 630: I guess this will place the JSON-encoded string value into process.env. However, sometimes "JSON"-typed bindings are actually used to deliver simple string values already -- the JSON evaluates to a single string. In that case, we end up with a string containing a quoted/escaped string value here, which may be surprising.

If we really want to maintain the property that `process.env` contains only strings, then I think what I'd do here is only set the env field if the value is a string, and then set it to that string.

src/node/internal/util.d.ts
```diff
@@ -120,6 +120,7 @@ export function isBoxedPrimitive(
   value: unknown
 ): value is number | string | boolean | bigint | symbol;
 
+export function getEnvObject(): object;
```
anonrig, line 123:
```suggestion
export function getEnvObject(): Record<string, string>;
```

src/workerd/api/node/util.h
```diff
@@ -259,6 +261,8 @@ class UtilModule final: public jsg::Object {
     JSG_METHOD(getConstructorName);
     JSG_METHOD(getCallSite);
 
+    JSG_METHOD(getEnvObject);
```
anonrig, line 264: I think we should move this implementation to a new file preferably called process. This util is node:util, and I'm afraid this file can easily become a place where we dump random functions.

danlapid (approved):

src/workerd/api/node/util.h
anonrig: Can you at least add a TODO so we don't forget?

anonrig (approved):

src/workerd/api/node/util.h
```diff
@@ -259,6 +261,8 @@ class UtilModule final: public jsg::Object {
     JSG_METHOD(getConstructorName);
     JSG_METHOD(getCallSite);
```
jasnell, line 263:
```suggestion
    // TODO(cleanup): It might be about time to separate some of these out to a different module.
```

vicb (changes requested):
@irvinebroque can we discuss this? 

I oppose this change of it exposed secrets to the world via process.env

jasnell:
> I oppose this change of it exposed secrets to the world via process.ENV

To be clear... the change puts any `TEXT` binding on process.env. We have no way of knowing if a a `TEXT` binding contains something that is considered a secret or not. If tooling insists on adding secrets to the worker configuration using a `TEXT` binding then this would mean we just cannot populate `process.env` automatically at all right now because we have zero ability to differentiate if the value of the `TEXT` binding is secret or not.

src/workerd/io/compatibility-date.capnp
```diff
@@ -679,4 +679,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("cache_no_cache_disabled")
       $experimental;
   # Enables the use of cache: no-cache in the fetch api.
+
+  populateProcessEnv @71 :Bool
+      $compatEnableFlag("nodejs_compat_populate_process_env")
+      $compatDisableFlag("nodejs_compat_dot_not_populate_process_env");
+  # Automatically populate process.env from text and json bindings
```
vicb, line 686:
```suggestion
  # Automatically populate process.env from text bindings only
```

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -6,3 +6,22 @@ export const processPlatform = {
...
+    assert.strictEqual(env.FOO, 'BAR');
+
+    // It should be possible to mutate the process.env at runtime.
+    assert.strictEqual(env.BAZ, '1');
+
+    // Any binding that is not explicitly a text binding should be
+    // ignored.
+    assert.strictEqual(env.BAR, undefined);
+
+    const { FOO } = await import('mod');
```
vicb, line 24: what's the purpose of this? 
_maybe add a comment?_

src/workerd/jsg/setup.h
```diff
@@ -665,6 +668,24 @@ class Isolate: public IsolateBase {
       }
     }
 
+    // Sets an env value that will be expressed on the process.env
+    // if/when nodejs-compat mode is used.
+    void setEnvField(const JsValue& name, const JsValue& value) override {
+      getEnv().set(*this, name, value);
+    }
+
+    // Returns the env base object.s
```
vicb, line 677: is the trailing "s" a typo?

src/workerd/api/node/tests/process-nodejs-test.js
jasnell, line 24:
```suggestion
    // Test that imports can see the process.env at the top level
    const { FOO } = await import('mod');
```

src/workerd/jsg/setup.h
jasnell:
```suggestion
    // Returns the env base object.
```

src/workerd/server/workerd-api.c++
```diff
@@ -711,6 +711,9 @@ static v8::Local<v8::Value> createBindingValue(JsgWorkerdIsolate::Lock& lock,
 
     KJ_CASE_ONEOF(text, kj::String) {
       value = lock.wrap(context, kj::mv(text));
+      if (featureFlags.getPopulateProcessEnv() && featureFlags.getNodeJsCompat()) {
+        lock.setEnvField(lock.str(global.name), jsg::JsValue(value));
```
kentonv, line 715: We also need to call this for JSON vars that parse to a plain string, and any other type of variable whose type is a plain string, otherwise we're still piercing the abstraction here.

kentonv: It might be easier to just give the entire `env` object over to the proxy, and let the proxy filter for fields which are strings.

jasnell: > We also need to call this for JSON vars that parse to a plain string,..

As discussed.. I'm strongly -1 on special casing JSON fields that parse to a string vs. other js types, largely because it sets up an inconsistency.

It is extremely common in Node.js for env vars to be defined as JSON strings such that node.js code will unconditionally do `JSON.parse(process.env.FOO)`. If sometimes a JSON binding is ignored, sometimes it comes through as a parsed json string, or sometimes it comes through as an encoded JSON string, then this extremely common Node.js pattern breaks. User code would end up having to be changed to inspect the value of `process.env.FOO` in advance or wrap the `JSON.parse(process.env.FOO)` in a `try/catch`, both of which are unlikely when we're talking about ecosystem modules folks are pulling off npm.

Yes, I understand that we have edge cases that represent what should be plaintext text bindings as JSON bindings but that's an implementation detail/quirk/edge case that I don't think we should be optimizing for.

> ... and any other type of variable whose type is a plain string

Such as?

`process.env` should be limited to specifically the things we call "environment variables" and not other types of bindings.

> It might be easier to just give the entire env object over to the proxy

I disagree. This mechanism also works for old service worker syntax workers in which the `env` object is the `globalThis`. It would make things rather more complicated if we had to special case these options whereas the current implementation keeps things rather simple.

jasnell: Further, if we did decide to propagate other types of bindings to `process.env` we can do so in separate PRs. That shouldn't be blocking for this initial PR

src/workerd/api/node/util.c++
```diff
@@ -234,6 +234,10 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
   return js.undefined();
 }
 
+jsg::JsObject UtilModule::getEnvObject(jsg::Lock& js) {
+  return js.getEnv(true);
```
vicb, line 238: what about:
getEnvObject -> getProcessEnvObj
getEnv -> getProcessEnv

src/workerd/jsg/setup.h
```diff
@@ -665,6 +668,24 @@ class Isolate: public IsolateBase {
       }
     }
 
+    // Sets an env value that will be expressed on the process.env
```
vicb, line 671: expressed -> exposed?
maybe expand on "when nodejs-compat mode is used" (to refer to the flag)

src/workerd/server/workerd-api.c++
vicb: Note: `wrangler dev/miniflare` use JSON bindings for now so the PR will not work as is.

jasnell: It does so for local development at least. A quick test of `wrangler deploy` shows that it appropriately uses `TEXT` bindings for env vars. So the limitation here would appear to be limited to local dev. Should be fixed, yes, but I don't think it's a blocker for merging.

jasnell: PR to fix `wrangler dev` so that it matches production and uses `TEXT` as appropriate https://github.com/cloudflare/workers-sdk/pull/7738

vicb:
@jasnell I think the following info needs to be added to the PR description (I can not edit it):

This PR
- is gated by the `nodejs_compat_populate_process_env` compatibility flag
- depends on https://github.com/cloudflare/workers-sdk/pull/7738
- publishes string environment variables and secrets to `process.env`

```
# wrangler.toml
[vars]
PUBLISHED="foo"
PUBLISHED_INT="1"
PUBLISHED_BOOL="true"
PUBLISHED_JSON = "{key: value}"
PUBLISHED_ARRAY = "[1]"

NOT_PUBLISHED_INT = 120
NOT_PUBLISHED_BOOL = true
NOT_PUBLISHED_JSON = {key = value}
NOT_PUBLISHED_ARRAY = [1]

// .dev.vars
PUBLISHED_SECRET = "secret"
```

Once the description has been added, it would be nice to create an issue on https://github.com/cloudflare/cloudflare-docs to ask this to be documented when merged

src/workerd/server/workerd-api.c++
kentonv: Sorry but the wrangler issue just proves my point: The existing abstraction is that a JSON binding containing just a string is equivalent to a text binding, and people use this.

I cannot accept this PR unless it treats string-typed JSON bindings the same as text.

kentonv (dismissed):
Sorry I really insist on this: All bindings which render to a plain string must be treated equivalently. We cannot have the behavior change based on the way the value was originally specified in config, because this is changing the abstraction in a way that has already been seen to break multiple tools and adds new constraints to our implementation going forward. The abstraction is that `env` contains JavaScript values; whether they were transmitted as simple text, JSON, v8 serilaization, etc. is not revealed to the application.

src/workerd/server/workerd-api.c++
jasnell: @mikenomitch has indicated that while the suggestion of special casing JSON bindings in this way is not ideal, it is acceptable to meet the product needs. PR is updated with comments and expanded tests.

src/workerd/api/node/util.c++
vicb: If feel strongly then.
Workerd has env and process.env.
They are different and it causes confusion. 
Unless I'm mistaken, the references here is to process.env?

src/workerd/api/node/tests/process-nodejs-test.wd-test
```diff
@@ -5,10 +5,21 @@ const unitTests :Workerd.Config = (
...
+        compatibilityFlags = [
+          "nodejs_compat",
+          "nodejs_compat_populate_process_env"
+        ],
+        bindings = [
+          (name = "FOO", text = "BAR"),
+          (name = "BAR", json = "{}"),
+          (name = "BAZ", json = "\"abc\""),
+          (name = "DUB", json = "\"\\\"abc\\\"\""),
+          (name = "NON", data = "abcdef")
```
vicb, line 21: maybe add json for number, boolean and null

src/workerd/server/workerd-api.c++
```diff
@@ -625,8 +625,68 @@ static v8::Local<v8::Value> createBindingValue(JsgWorkerdIsolate::Lock& lock,
...
+        //    console.log(typeof enf.BAZ);                      // 'string'
+        //    console.log(typeof process.env.BAZ);              // 'string'
+        //    console.log(env.BAZ);                             // '"abc"'
+        //    console.log(process.env.BAZ);                     // '"abc"
+        //    console.log(typeof JSON.parse(process.env.BAZ));  // 'string'
+        //
+        // JSON.parse(process.env.FOO) works because the value is a parseable
+        // JSON string. JSON.parse(process.env.BAR) throws an error because
+        // the value is not a parseable JSON string, even tho the binding uses
+        // type JSON. JSON.parse(process.env.BAZ)` works because the original
```
vicb, line 673: unbalanced back quote?

jasnell, line 673:
```suggestion
        // type JSON. JSON.parse(process.env.BAZ) works because the original
```

anonrig (approved):

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -6,3 +6,56 @@ export const processPlatform = {
...
+    assert.strictEqual(FOO, 'BAR');
+
+    // Mutating the env argument does not change process.env, and
+    // vis versa
+    assert.strictEqual(env.QUX, undefined);
+    env.ZZZ = 'a';
+    assert.strictEqual(pEnv.ZZZ, undefined);
+
+    env.FOO = ['just some other value'];
+    assert.strictEqual(pEnv.FOO, 'BAR');
```
jasnell, line 52:
```suggestion
    assert.strictEqual(process.env.FOO, 'BAR');
```

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -6,3 +6,56 @@ export const processPlatform = {
...
+
+    // Test that imports can see the process.env at the top level
+    const { FOO } = await import('mod');
+    assert.strictEqual(FOO, 'BAR');
+
+    // Mutating the env argument does not change process.env, and
+    // vis versa
+    assert.strictEqual(env.QUX, undefined);
+    env.ZZZ = 'a';
+    assert.strictEqual(pEnv.ZZZ, undefined);
```
jasnell, line 49:
```suggestion
    assert.strictEqual(process.env.ZZZ, undefined);
```

anonrig (approved):

jasnell:
Moved back to draft due to continued backend discussion on how the implementation intersects with non-nodejs-compat related aspects of the runtime.

jasnell:
Rebased the PR and expanded comments. Maintained the existing implementation approach. If it becomes necessary it ought to be able to further change the underlying implementation details without breaking changes. The hypothetical case around how to handle potential new binding types that evaluate to strings can be addressed separately later when they are no longer hypothetical. An internal PR is necessary.

irvinebroque (commented):
👍 let's find a way to ship

anonrig (approved):

vicb:
Could a doc PR be prepared for this change?

jasnell:
There already is a doc PR https://github.com/cloudflare/cloudflare-docs/pull/19187

vicb:
What are the plans regarding: 

> Autopopulates the process.env from bindings in local dev. A similar PR will be needed internally to enable it there as it won't be automatic.

When this will be activated on EW? When can this be enabled?

Does wrangler needs to be updated to support the flag/CLI options, is there a tracking issue?

PR #3319, opened 2025-01-10
Use python packages date from compat date.

dom96:
Changes necessary to avoid hardcoding the package bundle date.


hoodmane (approved):
Thanks!

PR #3320, opened 2025-01-10
remove eslint-prettier-config

anonrig:
Upon recommendation from @joshuakgoldberg (https://github.com/typescript-eslint/typescript-eslint/issues/10635#issuecomment-2583009860), removing invalid typescript-eslint rule, and removing eslint-prettier-config because it takes too much time running eslint with prettier.

We already have a linter check using `just f` no need to do the same work twice.

PS: This decreases eslint test to 4.1s seconds from 8 on a 88 core machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "esbuild": "^0.21.5",
     "eslint": "^9.10.0",
     "eslint-plugin-import": "^2.30.0",
-    "eslint-plugin-prettier": "^5.2.1",
     "expect-type": "^0.20.0",
     "typescript-eslint": "^8.4.0"
   },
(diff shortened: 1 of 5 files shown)


danlapid (approved):
LBTM

fhanau (approved):

JoshuaKGoldberg (commented):
Awesome, love to see the speedup! 🚀

fhanau:
> PS: This decreases eslint test to 4.1s seconds from 8 on a 88 core machine.

LGTM, but the `@eslint` job is single-threaded so the number of cores doesn't really matter right?

anonrig:
> > PS: This decreases eslint test to 4.1s seconds from 8 on a 88 core machine.
> 
> LGTM, but the `@eslint` job is single-threaded so the number of cores doesn't really matter right?

Yes, I was just showing the ridiculousness of the amount of time it takes to lint and format our JS files...

PR #3321, opened 2025-01-10
Enable external pkg loading by default for tests run against Python dev.

dom96:


src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -30,6 +30,7 @@ py_wd_test("subdirectory")
 
 py_wd_test(
     "sdk",
+    python_flags = ["0.26.0a2"],
```
hoodmane, line 33: Can you add a comment here explaining why development is disabled?

hoodmane (approved):

src/workerd/server/tests/python/BUILD.bazel
dom96: Will do in separate PR

PR #3324, opened 2025-01-10
Convert WPT test harness to TS

npaun:
Converts our WPT test harness to TypeScript.

### DIFF
--- a/src/wpt/BUILD.bazel
+++ b/src/wpt/BUILD.bazel
@@ -1,8 +1,44 @@
-filegroup(
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+load("@npm//:eslint/package_json.bzl", eslint_bin = "bin")
+
+srcs = glob(
+    [
+        "*.ts",
+    ],
+)
+
+ts_project(
     name = "wpt-test-harness",
-    srcs = glob(
-        include = ["**/*"],
-        allow_empty = True,
-    ),
+    testonly = False,
+    srcs = srcs,
+    allow_js = True,
+    declaration = True,
+    tsconfig = "tsconfig.json",
     visibility = ["//visibility:public"],
+    deps = [
+        "//:node_modules/@types/node",
+    ],
+)
+
+eslint_bin.eslint_test(
+    name = "wpt-test-harness@eslint",
+    size = "large",
+    args = [
+        "--config $(location {})".format("eslint.config.mjs"),
+        "--parser-options project:$(location {})".format("tsconfig.json"),
+        "-f stylish",
+        "--report-unused-disable-directives",
+    ] + ["$(location " + src + ")" for src in srcs],
+    data = srcs + [
+        "eslint.config.mjs",
+        "tsconfig.json",
+        "//:node_modules/@types/node",
+        "//:prettierrc",
+        "//tools:base-eslint",
+    ],
+    tags = ["lint"],
+    target_compatible_with = select({
+        "@platforms//os:windows": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    }),
 )
--- a/src/wpt/eslint.config.mjs
+++ b/src/wpt/eslint.config.mjs
@@ -1,0 +1,3 @@
+import { baseConfig } from '../../tools/base.eslint.config.mjs';
+
+export default [...baseConfig({ tsconfigRootDir: import.meta.dirname })];
--- a/src/wpt/harness.ts
+++ b/src/wpt/harness.ts
@@ -24,56 +24,182 @@
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import {
-  AssertionError,
   strictEqual,
   notStrictEqual,
   deepStrictEqual,
   ok,
   throws,
+  type AssertPredicate,
 } from 'node:assert';
 
-function OptionalFeatureUnsupportedError(message) {
+type TestRunnerOptions = {
+  expectedFailures?: string[];
+  verbose?: boolean;
+  skippedTests?: string[];
+};
+
+type TestRunnerFn = (callback: TestFn | PromiseTestFn, message: string) => void;
+type TestFn = () => void;
+type PromiseTestFn = () => Promise<void>;
+type ThrowingFn = () => unknown;
+
+declare global {
+  // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
+  var errors: Error[];
+  // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
+  var testOptions: TestRunnerOptions;
+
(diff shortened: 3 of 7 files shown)


src/wpt/harness.ts
```diff
@@ -32,12 +32,64 @@ import {
...
+  function assert_throws_dom(
+    type: any,
+    funcOrConstructor: any,
+    descriptionOrFunc: any,
+    maybeDescription: any
+  ): void;
 }
-OptionalFeatureUnsupportedError.prototype = Object.create(
-  AssertionError.prototype
-);
```
npaun, line 40: I did this to calm TS down. Does this have any bad side-effects?

src/wpt/harness.ts
```diff
@@ -67,13 +119,7 @@ globalThis.subsetTestByKey = (_key, testType, testCallback, testMessage) => {
   return testType(testCallback, testMessage);
 };
 
-globalThis.promise_test = async (func, name, properties) => {
-  if (typeof func !== 'function') {
-    properties = name;
-    name = func;
-    func = null;
-  }
```
npaun, line 75: I'm not sure what's going on here. If this branch is taken, we'd eventually get `null.call(this)` on line 128, which I don't think is good?

src/wpt/harness.ts
```diff
@@ -230,6 +276,7 @@ globalThis.assert_throws_dom = (
     func = descriptionOrFunc;
     description = maybeDescription;
   } else {
+    // @ts-ignore
     constructor = this.DOMException;
```
npaun, line 280: Any suggestions to do it better?

src/wpt/BUILD.bazel
```diff
@@ -1,8 +1,18 @@
...
+        [
+            "*.ts",
+        ],
     ),
+    allow_js = True,
+    declaration = True,
+    tsconfig = "tsconfig.json",
     visibility = ["//visibility:public"],
+    deps = [
+        "//:node_modules/@types",
```
anonrig, line 16:
```suggestion
        "//:node_modules/@types/node",
```

src/wpt/eslint.config.mjs
```diff
@@ -0,0 +1,3 @@
+import { baseConfig } from '../../tools/base.eslint.config.mjs';
+
+export default [...baseConfig({ tsconfigRootDir: import.meta.dirname })];
```
anonrig, line 3: It seems that ts_project for WPT does not use this eslint config. We have to add eslint_bin.eslint_test just like in wd_test_bundle if we want to run eslint in this folder. Otherwise, only your IDE will complain, and it won't fail on CI.

src/wpt/harness.ts
```diff
@@ -32,12 +32,68 @@ import {
...
+    description?: string
+  ): void;
+  function assert_throws_dom(
+    type: any,
+    funcOrConstructor: any,
+    descriptionOrFunc: any,
+    maybeDescription: any
+  ): void;
 }
-OptionalFeatureUnsupportedError.prototype = Object.create(
```
anonrig, line 38: Let's just add a ts-expect-error here and keep it as it is. I took the code from the actual wpt harness script in web-platform-tests repository.

src/wpt/harness.ts
anonrig: This code was taken from the actual web-platform-tests WPT harness. There might be a bug somewhere, or probably a porting error.

src/wpt/harness.ts
```diff
@@ -140,7 +190,7 @@ globalThis.assert_implements = (condition, description) => {
  */
 globalThis.assert_implements_optional = (condition, description) => {
   if (!condition) {
-    throw new OptionalFeatureUnsupportedError(description);
+    throw new OptionalFeatureUnsupportedError({ message: description });
```
anonrig, line 193: Since it extends AssertionError, the first parameter needs to be a string. I don't follow why this change was necessary.

src/wpt/harness.ts
```diff
@@ -230,6 +280,7 @@ globalThis.assert_throws_dom = (
     func = descriptionOrFunc;
     description = maybeDescription;
   } else {
+    // @ts-ignore
```
anonrig, line 283: It seems eslint is not running on CI for this file. If it did, it would give an error about replacing ts-ignore with ts-expect-error

src/wpt/harness.ts
```diff
@@ -285,12 +336,12 @@ function shouldRunTest(message) {
   return true;
 }
 
-function prepare(options) {
+function prepare(options: Options) {
```
anonrig, line 339: when eslint is run on this file, we will receive errors saying that each function return type should be explicitly defined.

src/wpt/harness.ts
```diff
@@ -32,12 +32,68 @@ import {
   throws,
 } from 'node:assert';
 
-function OptionalFeatureUnsupportedError(message) {
-  AssertionError.call(this, message);
+type Options = {
```
anonrig, line 35:
```suggestion
type TestRunnerOptions = {
```

src/wpt/harness.ts
npaun: I am a bit puzzled too, but TS is telling me: `error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type '{ message?: string | undefined; actual?: unknown; expected?: unknown; operator?: string | undefined; stackStartFn?: Function | undefined; } | undefined`

npaun: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5250f7a4bd6d7ff68f0ebb93da35f3b75aa22bdc/types/node/assert.d.ts#L38

anonrig: It seems web-platform-tests implementation of AssertionError is different from Node.js - https://github.com/web-platform-tests/wpt/blob/038c19b8529b374bb40b6b01cf062577f0dc39dc/resources/testharness.js#L4554 - Can you add a comment that this is indeed correct, and it is different from WPT AssertionError.

anonrig: Actually this means we shouldn't use AssertionError from Node.js and have a similar API just like WPT. Because some tests might call this error with a string as a parameter.

src/wpt/eslint.config.mjs
npaun: Thanks for the pointer! I have eslint on now and have fixed some stuff. Still have some issues to get through.

npaun:
@anonrig eslint issues should all be addressed now

jasnell (approved):

anonrig (approved):

anonrig:
It seems there are some eslint errors

npaun:
They only happen in CI, not on my machine. Going to look into what's up with that and get rid of the errors.

npaun:
The error was caused by not passing "//:node_modules/@types/node" to the eslint_test rule. Fixed now.

src/wpt/harness.ts
npaun: I can confirm it's in the original code. I'll leave it out for now until we find an example of a non-function being passed, and then we can upgrade our harness to deal with it.

PR #3329, opened 2025-01-13
Auto-discover new WPT as they are added

npaun:
WPT regularly adds new tests. To make sure we don't miss any, we automatically generate a test case for each WPT test file discovered within a group. 

The manually written test files now contain only a default export containing config for test cases. One new TestRunnerOption was added `skipAllTests: true` will avoid processing a file if we know there are no test cases we are able to run in that file.


src/workerd/api/wpt/url-test.js
```diff
@@ -2,140 +2,136 @@
...
-);
+  },
+  'urlsearchparams-sort.any.js': {
+    expectedFailures: ['Parse and sort: ﬃ&🌈', 'URL parse and sort: ﬃ&🌈'],
+  },
+  'idlharness.any.js': {
+    skipAllTests: true,
+  },
+  // TODO(soon): Implement globalThis.document
+  'a-element-origin.js': {
```
anonrig, line 130: Can you enable this lint rule for these kinds of files? https://eslint.org/docs/latest/rules/sort-keys

anonrig (commented):
One problem we should address in the future is, in node.js, we update wpt by module, meaning, you can update URL tests but not URLPattern. Right now, we have a single dependency, and whenever we update it to latest it will always update everything.

anonrig:
> I think we could split the dependency but it might make a mess in `deps.jsonc` if we eventually enable lots of wpt modules.

We can do something similar to Node.js and just clone/copy the necessary directories in the workerd repository.

src/workerd/api/wpt/url-test.js
npaun: It's all sortitudinated now.

build/wpt_test.bzl
```diff
@@ -9,34 +9,88 @@
...
+        result.append(entry)
+
+    return "\n".join(result)
+
+def test_case_name(filename):
+    words = (filename
+        .removesuffix(".js")
+        .removesuffix(".any")
+        .replace(".", "-")
+        .split("-"))
```
anonrig, line 74:
```suggestion
        .split("."))
```

anonrig (commented):
I'll review more tomorrow:

build/wpt_test.bzl
npaun: If I literally applied this, I'd end up with invalid JS identifiers I think

anonrig: you are replacing . with `-` and later splitting with `-`. No need to traverse the string twice, where you can just split by `.` unless there is something I'm missing, which in that case i think it requires adding a comment.

src/workerd/api/wpt/url-test.js
```diff
@@ -2,140 +2,136 @@
...
-});
-export const urlSearchParamsAppend = run('urlsearchparams-append.any.js');
-export const urlSearchParamsConstructor = run(
-  'urlsearchparams-constructor.any.js',
-  {
+export default {
+  // TODO(soon): Implement globalThis.document
+  'a-element-origin.js': {
+    skipAllTests: true,
+  },
```
anonrig, line 9: I have a suggestion, let's change the type of this. If skipAllTests is true, let's require `comment: string` to exist, so we force the developer to write comment.

```suggestion
  'a-element-origin.js': {
    comment: 'Implement globalThis.document',
    skipAllTests: true,
  },
```

build/wpt_test.bzl
npaun: Ah! The original file names contain both `-` and `.` like `'url-setters-a-area.window.js`. I'm trying to turn this into a nice-ish JS identifier like `urlSettersAAreaWindow`. I'll add a comment to clarify.

src/workerd/api/wpt/url-test.js
npaun: Love it!

build/wpt_test.bzl
```diff
@@ -9,34 +9,97 @@
...
+
+def generate_external_cases(files):
+    result = []
+
+    for file in files.to_list():
+        if file.extension == "js":
+            entry = """export const {} = run(config, '{}');""".format(test_case_name(file.basename), file.basename)
+        else:
+            continue
+        result.append(entry)
```
anonrig, line 68:
```suggestion
        if file.extension == "js":
            entry = """export const {} = run(config, '{}');""".format(test_case_name(file.basename), file.basename)
            result.append(entry)
```

build/wpt_test.bzl
```diff
@@ -9,34 +9,97 @@
...
+            test_config = ctx.file.test_config.basename,
+            cases = generate_external_cases(ctx.attr.wpt_directory.files),
+        ),
+    )
+
+    return DefaultInfo(
+        files = depset([src]),
+    )
+
+def generate_external_cases(files):
```
anonrig, line 60: Can you add a comment to all functions

src/workerd/api/wpt/BUILD.bazel
```diff
@@ -2,10 +2,35 @@
...
+    args = [
+        "--config $(location {})".format("eslint.config.mjs"),
+        "--parser-options project:$(location {})".format("tsconfig.json"),
+        "-f stylish",
+        "--report-unused-disable-directives",
+    ] + ["$(location " + src + ")" for src in srcs],
+    data = srcs + [
+        "eslint.config.mjs",
+        "tsconfig.json",
+        "//:prettierrc",
```
anonrig, line 28: We don't need prettierrc since we don't run prettier with eslint anymore.

src/workerd/api/wpt/url-test.ts
```diff
@@ -0,0 +1,144 @@
...
+    expectedFailures: [
+      'URLSearchParams constructor, DOMException as argument',
+      'Construct with 2 unpaired surrogates (no trailing)',
+      'Construct with 3 unpaired surrogates (no leading)',
+      'Construct with object with NULL, non-ASCII, and surrogate keys',
+    ],
+  },
+  'urlsearchparams-sort.any.js': {
+    comment: 'Figure out why we sort differently than the WPT tests',
+    expectedFailures: ['Parse and sort: ﬃ&🌈', 'URL parse and sort: ﬃ&🌈'],
```
anonrig, line 142: This is probably related to how sort() function is implemented in Ada https://github.com/ada-url/ada/blob/main/include/ada/url_search_params-inl.h#L165

Good opportunity to contribute to Ada

anonrig (approved):

PR #3331, opened 2025-01-13
dns: do not parse JSON when response is not ok

vicb:
/cc @anonrig

### DIFF
--- a/src/node/internal/internal_dns_client.ts
+++ b/src/node/internal/internal_dns_client.ts
@@ -71,10 +71,15 @@ export async function sendDnsRequest(
       },
       method: 'GET',
     });
+    if (!response.ok) {
+      throw new DnsError(name, errorCodes.BADRESP, syscall);
+    }
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     json = await response.json();
-  } catch {
-    throw new DnsError(name, errorCodes.BADQUERY, syscall);
+  } catch (e) {
+    throw e instanceof DnsError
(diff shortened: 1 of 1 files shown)


src/node/internal/internal_dns_client.ts
```diff
@@ -71,6 +71,11 @@ export async function sendDnsRequest(
       },
       method: 'GET',
     });
+    if (!response.ok) {
+      throw new Error(
```
anonrig, line 75: Can you throw DnsError rather than Error here?

anonrig (approved):

src/node/internal/internal_dns_client.ts
```diff
@@ -71,6 +71,11 @@ export async function sendDnsRequest(
...
+        `response not ok ${String(response.status)} ${response.statusText}`
```
vicb, line 76: @anonrig a DNSError will be thrown in the catch below, whatever we throw here doesn't really matter.

Does that sound ok?

vicb: Or maybe we should throw a BAD_RESP instead. The error is visible to the user, right?

anonrig: This error is visible to the user yes. I suppose BAD_RESP is a better error code.

jasnell (approved):

vicb:
@anonrig I have addressed the feedback, could you please merge if this looks good.

Thanks

PR #3338, opened 2025-01-15
Enable python external package loading by default.

dom96:
**Not yet ready to merge**

Removes the pyodide_extra embedded capnp file so nothing from Pyodide is embedded in the binaries anymore.


hoodmane:
I think this was superseded by #3678.

PR #3341, opened 2025-01-15
CACHE-11937: add compat flag to support cache rules overrides

yj7o5:
Adding a new flag to support overriding cache rules via `request.cf` object. Currently, if a resource is fetched via the fetch API and have cache settings such as `cacheEverything: true` specified in the 'cf' object, while there is also cache rules defined against that resource, the cache settings in request cf object isn't respected. Cache team is working on fixing this behavior to give precedence over to request cf object first and then to cache rules, however, we are hiding this behavior behind a compatibility flag until we're confident the change works as intended for everyone.


yj7o5:
recheck

PR #3346, opened 2025-01-16
add node:timers module

anonrig:
Once this lands, we can directly connect to mysql using "mysql" npm package from workers.

### DIFF
--- a/src/node/internal/internal_timers.ts
+++ b/src/node/internal/internal_timers.ts
@@ -1,0 +1,191 @@
+// Copyright (c) 2017-2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+//
+// Adapted from Node.js. Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import { validateFunction } from 'node-internal:validators';
+
+let clearTimeoutImpl: (obj: Timeout) => void;
+
+export class Timeout {
+  #timer: number;
+  #callback: (...args: unknown[]) => unknown;
+  #after: number;
+  #args: unknown[];
+  #isRepeat: boolean;
+  #isRefed: boolean;
+
+  public constructor(
+    callback: (...args: unknown[]) => unknown,
+    after: number = 1,
+    args: unknown[] = [],
+    isRepeat: boolean = false,
+    isRefed: boolean = false
+  ) {
+    this.#callback = callback;
+    // Left it as multiply by 1 due to make the behavior as similar to Node.js
+    // as possible.
+    this.#after = after * 1;
+    this.#args = args;
+    this.#isRepeat = isRepeat;
+    this.#isRefed = isRefed;
+    this.#timer = this.#constructTimer();
+  }
+
+  #constructTimer(): number {
+    if (this.#isRepeat) {
+      // @ts-expect-error TS2322 Due to difference between Node.js and globals
+      return globalThis.setInterval(this.#callback, this.#after, ...this.#args);
+    } else {
+      // @ts-expect-error TS2322 Due to difference between Node.js and globals
+      return globalThis.setTimeout(this.#callback, this.#after, ...this.#args);
+    }
+  }
+
+  #clearTimeout(): void {
+    if (this.#isRepeat) {
+      globalThis.clearInterval(this.#timer);
+    } else {
+      globalThis.clearTimeout(this.#timer);
+    }
+  }
+
+  public refresh(): this {
+    this.#clearTimeout();
+    this.#constructTimer();
+    return this;
+  }
+
+  public unref(): this {
+    // Intentionally left as no-op.
+    this.#isRefed = false;
+    return this;
+  }
+
+  public ref(): this {
+    // Intentionally left as no-op.
+    this.#isRefed = true;
+    return this;
+  }
+
+  public hasRef(): boolean {
+    return this.#isRefed;
+  }
+
+  public close(): this {
+    this.#clearTimeout();
+    return this;
+  }
+
+  public [Symbol.dispose](): void {
+    this.#clearTimeout();
+  }
+
+  public [Symbol.toPrimitive](): number {
+    return this.#timer;
+  }
+
+  static {
+    clearTimeoutImpl = (obj: Timeout): void => {
+      obj.#clearTimeout();
+    };
+  }
+}
+
+export function setTimeout(
+  callback: (...args: unknown[]) => unknown,
+  after: number,
+  ...args: unknown[]
+): Timeout {
+  validateFunction(callback, 'callback');
+
+  return new Timeout(
+    callback,
+    after,
+    args,
+    /* isRepeat */ false,
+    /* isRefed */ true
+  );
+}
+
+export function clearTimeout(timer: unknown): void {
+  if (timer instanceof Timeout) {
+    clearTimeoutImpl(timer);
+  } else if (typeof timer === 'number') {
+    globalThis.clearTimeout(timer);
+  }
+}
+
+export const setImmediate = globalThis.setImmediate.bind(globalThis);
+
+export const clearImmediate = globalThis.clearImmediate.bind(globalThis);
+
+export function setInterval(
+  callback: (...args: unknown[]) => void,
+  repeat: number,
+  ...args: unknown[]
+): Timeout {
+  validateFunction(callback, 'callback');
+  return new Timeout(
+    callback,
+    repeat,
+    args,
+    /* isRepeat */ true,
+    /* isRefed */ true
+  );
+}
+
+export function clearInterval(timer: unknown): void {
+  if (timer instanceof Timeout) {
+    clearTimeoutImpl(timer);
+  } else if (typeof timer === 'number') {
+    globalThis.clearInterval(timer);
+  }
+}
+
+/**
+ * @deprecated Please use timeout.refresh() instead.
+ */
+export function active(timer: unknown): void {
+  if (timer instanceof Timeout) {
+    timer.refresh();
+  }
+}
+
+/**
+ * @deprecated Please use clearTimeout instead.
+ */
+export function unenroll(timer: unknown): void {
+  if (timer instanceof Timeout) {
+    clearTimeoutImpl(timer);
+  } else if (typeof timer === 'number') {
+    globalThis.clearTimeout(timer);
+  }
+}
+
+/**
+ * @deprecated Please use setTimeout instead.
+ */
+export function enroll(_item: unknown, _msecs: number): void {
+  throw new Error('Not implemented. Please use setTimeout() instead.');
+}
--- a/src/node/internal/internal_timers_promises.ts
+++ b/src/node/internal/internal_timers_promises.ts
@@ -1,0 +1,229 @@
+// Copyright (c) 2017-2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+//
+// Adapted from Node.js. Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import * as timers from 'node-internal:internal_timers';
+import { ERR_INVALID_THIS, AbortError } from 'node-internal:internal_errors';
+import {
+  validateNumber,
+  validateAbortSignal,
+  validateBoolean,
+  validateObject,
+} from 'node-internal:validators';
+
+const kScheduler = Symbol.for('kScheduler');
+
+export async function setTimeout(
+  after: number | undefined,
+  value?: unknown,
+  options: { signal?: AbortSignal; ref?: boolean } = {}
+): Promise<typeof value> {
+  if (after !== undefined) {
+    validateNumber(after, 'delay');
+  }
+
+  validateObject(options, 'options');
+
+  // Ref options is a no-op.
+  const { signal, ref } = options;
+
+  if (signal !== undefined) {
+    validateAbortSignal(signal, 'options.signal');
+  }
+
+  // This is required due to consistency/compat reasons, even if it's no-op.
+  if (ref !== undefined) {
+    validateBoolean(ref, 'options.ref');
+  }
+
+  if (signal?.aborted) {
+    throw new AbortError(undefined, { cause: signal.reason });
+  }
+
+  const { promise, resolve, reject } = Promise.withResolvers<typeof value>();
+
+  const timer = timers.setTimeout(() => {
+    resolve(value);
+  }, after ?? 0);
+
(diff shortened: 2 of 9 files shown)


src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  public refresh(): this {
+    this.#clearTimeout();
+    this.#constructTimer();
+    return this;
+  }
+
+  public unref(): this {
+    // TODO(soon): Implement this
```
anonrig, line 67: @jasnell Any suggestions/ideas on how to handle unref() and ref()?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
+import { validateFunction, validateNumber } from 'node-internal:validators';
+import { ERR_OUT_OF_RANGE } from 'node-internal:internal_errors';
+
+// Timeout values > TIMEOUT_MAX are set to 1.
+const TIMEOUT_MAX = 2 ** 31 - 1;
```
vicb, line 5: perf/nit:

```suggestion
const TIMEOUT_MAX = 2147483647; //2 ** 31 - 1
```

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+        () => this.#callback(...this.#args),
+        this.#after
+      );
+    }
+  }
+
+  #clearTimeout(): void {
+    if (this.#isRepeat) {
+      globalThis.clearInterval(this.#timer);
+    }
```
vicb, line 72: missing an else here?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+
+function getTimerDuration(msecs: unknown, name: string): number {
+  validateNumber(msecs, name);
+  if (msecs < 0 || !Number.isFinite(msecs)) {
+    throw new ERR_OUT_OF_RANGE(name, 'a non-negative finite number', msecs);
+  }
+
+  // Ensure that msecs fits into signed int32
+  if (msecs > TIMEOUT_MAX) {
+    return TIMEOUT_MAX;
```
vicb, line 18: Does't do what the comment says: `// Timeout values > TIMEOUT_MAX are set to 1.`

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+  #isRefed: boolean;
+
+  public constructor(
+    callback: (...args: unknown[]) => unknown,
+    after?: number,
+    args: unknown[] = [],
+    isRepeat: boolean = false,
+    isRefed: boolean = false
+  ) {
+    if (after === undefined) {
```
vicb, line 39: nit
```ts
this.#after = after === undefined ? 1 : after * 1;
```

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+    /* isRepeat */ false,
+    /* isRefed */ true
+  );
+}
+
+export function clearTimeout(timer: unknown): void {
+  if (timer instanceof Timeout) {
+    clearTimeoutImpl(timer);
+    return;
+  } else if (typeof timer === 'number') {
```
vicb, line 205: nit: no else after return ?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+    timer.refresh();
+  }
+}
+
+/**
+ * @deprecated Please use clearTimeout instead.
+ */
+export function unenroll(timer: unknown): void {
+  if (timer instanceof Timeout) {
+    clearTimeoutImpl(timer);
```
vicb, line 308: Do we need `else if (typeof timer === 'number')` here?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+  }
+}
+
+/**
+ * @deprecated Please use setTimeout instead.
+ */
+export function enroll(_item: unknown, msecs: number): void {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  msecs = getTimerDuration(msecs, 'msecs');
```
vicb, line 317:
```suggestion
export function enroll(_item: unknown, _msecs: number): void {
```

src/workerd/api/node/tests/timers-nodejs-test.js
```diff
@@ -0,0 +1,26 @@
+import timers from 'node:timers';
+import { throws } from 'node:assert';
+
+// Tests are taken from:
+// https://github.com/nodejs/node/blob/3fe8027cbb197953bd524115295d7dd13b9ee1fe/test/parallel/test-timers-enroll-invalid-msecs.js
+export const enrollInvalidMsecs = {
+  async test() {
+    [{}, [], 'foo', () => {}, Symbol('foo')].forEach((val) => {
+      throws(() => timers.enroll({}, val), {
```
vicb, line 9: This would not work with the change I suggested above...

But do we really need to test enroll which is unsupported anyway?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+    this.#clearTimeout();
+    return this;
+  }
+
+  public [Symbol.dispose](): void {
+    this.#clearTimeout();
+  }
+
+  public [Symbol.toPrimitive](): number {
+    // @ts-expect-error TS2322 Timer is actually an ID.
```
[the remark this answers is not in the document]
jasnell: Looks like `this.#timer` is a number. Not sure what you're asking here @vicb

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  static {
+    clearTimeoutImpl = (obj: Timeout): void => {
+      obj.#clearTimeout();
+    };
+  }
+}
+
+class Immediate {
```
jasnell, line 121: hmm... we already have the `Immediate` class implemented in c++ in support of the existing `setImmediate()` impl... is there really a need to implement it again here? That's likely to cause issues I think

src/node/internal/internal_timers.ts
anonrig: I'm fixing this, but most of the time I intentionally doesn't make changes like this since the original code is taken from Node.js and any divergence from the original code might make it harder to keep our implementation as sync as possible.

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,320 @@
...
+  callback: (...args: unknown[]) => unknown,
+  after: number,
+  arg1?: unknown,
+  arg2?: unknown,
+  arg3?: unknown
+): Timeout {
+  validateFunction(callback, 'callback');
+
+  let i, args;
+  switch (arguments.length) {
```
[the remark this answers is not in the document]
anonrig: This is from the Node.js implementation. I try to reuse as much as code possible to keep track of our implementation being sync with Node.js'. 

I'll change the implementation.

src/node/internal/internal_timers.ts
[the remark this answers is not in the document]
vicb: Oh I was thinking Node (i.e. `Timeout`) and forgot that `globalThis.Set...` returns a `number` in workerd

So what I am asking is to change `#timer: string | number | this | undefined;` ;)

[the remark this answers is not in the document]
vicb: And BTW if we switch `globalThis.setInterval` (and `nodeTimer.setInterval`) to returning a `Timeout` (from a `number`) - is this considered a breaking change and needs a compat flag ?

src/node/internal/internal_timers.ts
anonrig: It doesn't seem to implement Symbol.dispose(). Do we implement SymbolDispose in anywhere in workerd in C++?

src/node/internal/internal_timers.ts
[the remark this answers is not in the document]
anonrig: I made it simpler now.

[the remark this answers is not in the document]
vicb: > This only exposes the APIs through node:timers module. The globals are not changed.

Isn't that a problem?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,252 @@
...
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import { validateFunction } from 'node-internal:validators';
+
+let clearTimeoutImpl: (obj: Timeout) => void;
+let clearImmediateImpl: (obj: Immediate) => void;
+
+class Timeout {
+  // @ts-expect-error TS2564 It has a default value, but TS fails to detect it.
```
vicb, line 32: maybe you can return `this.#timer` from `#constructTimer()` and call it as `this.#timer = this.#constructTimer()` in the ctor?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,252 @@
...
+    this.#after = after * 1;
+    this.#args = args;
+    this.#isRepeat = isRepeat;
+    this.#isRefed = isRefed;
+    this.#constructTimer();
+  }
+
+  #constructTimer(): void {
+    if (this.#isRepeat) {
+      // @ts-expect-error TS2322 Due to difference between Node.js and globals
```
vicb, line 59: Is this`@ts-expect-error` still needed after the type change?

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,252 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
```
vicb, line 1: it's 2025 Yagiz :)

anonrig: I'm an old fashioned guy

src/node/internal/internal_timers_promises.ts
```diff
@@ -0,0 +1,82 @@
...
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+/* eslint-disable @typescript-eslint/require-await,@typescript-eslint/no-deprecated,prefer-spread */
+
+import * as timers from 'node-internal:internal_timers';
+
+export async function setTimeout(
+  ...args: Parameters<typeof timers.setTimeout>
+): Promise<ReturnType<typeof timers.setTimeout>> {
+  return timers.setTimeout.apply(timers, args);
```
vicb, line 33: I fail to understand how this is a promise version?

vicb: Maybe there should be tests :)

vicb: The `timers/promises` API is quite different: https://nodejs.org/api/timers.html#timers-promises-api

vicb: Maybe ask this guy https://github.com/jasnell/awaitable-timers ;)

(I found the repo when I was thinking about implementing it in unenv)

vicb (commented):
Should we consider changing the global functions in node compat mode?

It weird that `globalThis.setTimeout !== nodeTimer.setTimeout` and they return a different type. IIUC `globalThis.clearTimeour(nodeTimer.setTimeout(...))` will not work.

Edit: found my [original comment](https://github.com/cloudflare/workerd/pull/3346#discussion_r1919097963)

src/node/internal/internal_timers_promises.ts
anonrig: Ah, yeah it seems I forgot certain features. This basically mimics promises behavior by surrounding it with async.

anonrig: Fixed. Thanks!

src/node/internal/internal_timers.ts
anonrig: Yes if not, eslint throws an error

src/node/internal/internal_timers.ts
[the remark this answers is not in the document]
anonrig: This requires a bigger discussion, and a breaking change. Let's do it after this PR lands.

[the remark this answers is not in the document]
vicb: I would prefer to discuss that **before** this PR lands.
I landed 2 PRs in unenv that should fix the mysql issue so I don't think we need to rush this one in before having a proper plan.

jasnell:
> Should we consider changing the global functions in node compat mode?

No, that would end up being a breaking change. I can foresee possibly adding a separate compat flag to do so but the ship has already sailed.

src/node/internal/internal_timers.ts
jasnell: Yep. Search for `JSG_DISPOSE`

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  static {
+    clearTimeoutImpl = (obj: Timeout): void => {
+      obj.#clearTimeout();
+    };
+  }
+}
+
+export class Immediate {
```
jasnell, line 125: Everything else here looks good. I just don't want to introduce a second implementation of `Immediate`. Let's use the existing `globalThis.Immediate` instead.

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,252 @@
...
+  }
+}
+
+export function setImmediate(
+  callback: (...args: unknown[]) => void,
+  ...args: unknown[]
+): Immediate {
+  validateFunction(callback, 'callback');
+  return new Immediate(callback, args);
+}
```
jasnell, line 196: This should just be `export const setImmediate = globalThis.setImmediate;`

src/node/internal/internal_timers.ts
```diff
@@ -0,0 +1,252 @@
...
+
+export function clearImmediate(immediate?: Immediate): void {
```
jasnell, line 198: `export const clearImmediate = globalThis.clearImmediate;`

src/node/internal/internal_timers_promises.ts
```diff
@@ -0,0 +1,218 @@
...
+class Scheduler {
+  public [kScheduler] = true;
+
+  public yield(): Promise<void> {
+    if (!this[kScheduler]) throw new ERR_INVALID_THIS('Scheduler');
+    // @ts-expect-error TS2555 Following Node.js implementation
+    return setImmediate();
+  }
+
+  public wait(
```
jasnell, line 208: We already implement the global `scheduler.wait(...)` in the runtime. (see `globalThis.scheduler`). This should defer to that instead of being a separate impl.

src/node/internal/internal_timers_promises.ts
```diff
@@ -0,0 +1,218 @@
...
+    validateNumber(after, 'delay');
+  }
+
+  validateObject(options, 'options');
+
+  if (options.signal !== undefined) {
+    validateAbortSignal(options.signal, 'options.signal');
+  }
+
+  if (options.ref !== undefined) {
```
jasnell, line 52: Add a comment here explaining that `options.ref` is not used but we are going to validate it still for consistency/compat reasons.

src/node/internal/internal_timers_promises.ts
```diff
@@ -0,0 +1,218 @@
...
+  after: number | undefined,
+  value?: unknown,
+  options: { signal?: AbortSignal; ref?: boolean } = {}
+): Promise<typeof value> {
+  if (after !== undefined) {
+    validateNumber(after, 'delay');
+  }
+
+  validateObject(options, 'options');
+
```
jasnell, line 47:
```js
    const {
      signal,
      ref,
    } = options;
```

Then you can simplify a few of the checks below.

vicb:
> > Should we consider changing the global functions in node compat mode?
> 
> No, that would end up being a breaking change. I can foresee possibly adding a separate compat flag to do so but the ship has already sailed...
> 
> That said, we could probably make it so that `globalThis.clearTimeout(nodeTimer.setTimeout())` (and vis versa) would work without too much difficulty

In node we can do:

```js
const timeout = setTimeout(...);
timeout.refresh(); // crashes, timeout is a number
```

We can not do it in worked (both before or after this PR).

One thing that we can do in workerd after this PR is:

```
import nodeTimers from `node:timers`;
const timeout = nodeTimers.setTimeout(...);
timeout.refresh();
```

Meaning `globalThis.setTimeout` is not the same as `nodeTimers.setTimeout` which I don't think is the right thing to do.

> I can foresee possibly adding a separate compat flag to do so but the ship has already sailed...

I don't agree here. I consider this a bug fix that we have to do. That is  I think `globalThis.setTimeout` must return an instance of `Timeout` as Node does and not a number as it does currently - and I agree there should be a flag.

I think there is no rush to merge this PR at least before we discuss an agree on what the behavior should be.

I think the behavior introduced in this PR can be implemented in unenv for now. In fact `unenv@2.0.0-rc.0` already has the changes enabling mysql, there is a [pending PR](https://github.com/cloudflare/workers-sdk/pull/7806) to integrate it into wrangler - so ETA is next Tuesday.

vicb (changes requested):
I think we need more discussion before landing this.

anonrig:
> I think we need more discussion before landing this.

@vicb Whether or not changing global setTimeout/setImmediate is not within this pull-request. We already have most of the node:timers through the polyfill, hence, this pull-request is not changing anything that isn't out there. Since this enables mysql and there are teams/customers waiting for this pull-request to land, I propose postponing the discussion of globals once this PR lands, and proceed with this PR.

src/node/internal/internal_timers.ts
anonrig: Immediate class doesn't implement Symbol.Dispose. I asked on another review comment of your in this pull-request but it seems it got lost in the process. Any suggestions on how to add SymbolDispose and whether or not it will be a breaking change?

src/node/internal/internal_timers.ts
anonrig: globalThis.setImmediate returns an ID, whereas we want to return an Immediate class in here. This wouldn't be compatible with nodejs, if I'm not missing anything here?

src/node/internal/internal_timers.ts
anonrig: This won't work if setImmediate returns an Immediate class instance and not an ID.

jasnell:
> Meaning globalThis.setTimeout is not the same as nodeTimers.setTimeout which I don't think is the right thing to do.

These are different APIs. Workers is Web compatible first and we're not going to change the behavior of `globalThis.setTimeout(...)` to match Node.js. Perhaps we can do so at some point with an explicitly opt-in compatibility flag but by default our `globalThis.setTimeout(...)` should follow the web platform spec while `require('node:timer').setTimeout(...)` will follow Node.js' spec. The fact that these have different returns and different behaviors is just something we'll have to accept.

> I don't agree here. I consider this a bug fix that we have to do. That is I think globalThis.setTimeout must return an instance of Timeout as Node does and not a number as it does currently - and I agree there should be a flag.

Strong -1 on this. It's not a bug that `globalThis.setTimeout(...)` conforms to the web platform API specification for `setTimeout` and it's not a bug that `nodeTimer.setTimeout(...)` conforms to the Node.js version of `setTimeout(...)`. The fact that these work differently is just something we'll need to put up with.

We would be absolutely no different than Deno with this. In Deno, like the implementation here, has `globalThis.setTimeout(...)` conforming the web platform spec and the `node:timers` version conforming to Node.js.

image

vicb:
@anonrig you said:

> Since this enables mysql and there are teams/customers waiting for this pull-request to land, I propose postponing the discussion of globals once this PR lands, and proceed with this PR.

What I said is:

> I think the behavior introduced in this PR can be implemented in unenv for now. In fact unenv@2.0.0-rc.0 already has the changes enabling mysql, there is a https://github.com/cloudflare/workers-sdk/pull/7806 to integrate it into wrangler - so ETA is next Tuesday.

So my understanding is that "teams/customers waiting for this pull-request to land" will be satisfied with the unenv update alone, but please correct me if I am wrong.

jasnell:
@vicb ... I acknowledge that the situation is unfortunate. Node.js should never have implemented this API in a non-web compatible way, but it did. This is not a reason to block and this is blocking progress on other things. I'm going to dismiss your block so this can proceed.

jasnell (approved):

irvinebroque:
Nice one. Ship it!

PR #3347, opened 2025-01-16
Fixes Python SDK tests when external package loading is enabled.

dom96:
Some explanation of the failure here: The `internet` outbound in workerd is the one that `fetch` targets, so since we fetch packages in workerd and this test points `fetch` at a custom _Python_ server, it goes into an infinite loop due to the Python server attempting to load packages from ... itself.

Solution here is to create a JS proxy that either sends the requests onto the Python server or to the internet.

### DIFF
--- a/src/workerd/server/tests/python/sdk/proxy.js
+++ b/src/workerd/server/tests/python/sdk/proxy.js
@@ -1,0 +1,12 @@
+export default {
+  async fetch(req, env) {
+    const url = new URL(req.url);
+    if (url.hostname == 'pyodide-packages.runtime-playground.workers.dev') {
+      return env.INTERNET.fetch(req);
+    } else if (url.hostname == 'example.com') {
+      return env.PYTHON.fetch(req);
+    }
+
+    throw new Error('Invalid url: ' + url);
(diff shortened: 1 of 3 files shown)


hoodmane (approved):

src/workerd/server/tests/python/sdk/proxy.js
```diff
@@ -0,0 +1,12 @@
+export default {
+  async fetch(req, env) {
+    const url = new URL(req.url);
+    if (url.hostname == 'pyodide-packages.runtime-playground.workers.dev') {
+      return env.INTERNET.fetch(req);
+    } else if (url.hostname == 'example.com') {
+      return env.PYTHON.fetch(req);
+    }
+
+    throw new Error('Invalid url: ' + req.url);
```
anonrig, line 10: New URL can normalize/encode the string. It will help us a lot if we log the parsed value. 

```suggestion
    throw new Error('Invalid url: ' + url);
```

anonrig: Right now this line is req, rather than url.

PR #3353, opened 2025-01-17
WPT: Run tests with unsafeEval so they can use IOContext

npaun:
WPT tests are now declared as text bindings and executed with unsafeEval. This way they can do IOContext things. 
This PR also fixes promise_test to store each promise created and await them all before evaluating the test results.

### DIFF
--- a/src/wpt/harness.ts
+++ b/src/wpt/harness.ts
@@ -57,8 +57,13 @@ export type TestRunnerConfig = {
   [key: string]: TestRunnerOptions;
 };
 
+type Env = {
+  unsafe: { eval: (code: string) => void };
+  [key: string]: unknown;
+};
+
 type TestCase = {
-  test(): Promise<void>;
+  test(_: unknown, env: Env): Promise<void>;
 };
 
 type TestRunnerFn = (callback: TestFn | PromiseTestFn, message: string) => void;
@@ -67,13 +72,13 @@ type PromiseTestFn = () => Promise<void>;
 type ThrowingFn = () => unknown;
 
 declare global {
-  // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
+  /* eslint-disable no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis */
   var errors: Error[];
-  // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
   var testOptions: TestRunnerOptions;
-
-  // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
   var GLOBAL: { isWindow(): boolean };
+  var env: Env;
+  var promises: { [name: string]: Promise<void> };
+  /* eslint-enable no-var */
 
   function test(func: TestFn, name: string): void;
   function done(): undefined;
@@ -82,12 +87,12 @@ declare global {
     testType: TestRunnerFn,
     testCallback: TestFn | PromiseTestFn,
     testMessage: string
-  ): void | Promise<void>;
+  ): void;
   function promise_test(
     func: PromiseTestFn,
     name: string,
     properties?: unknown
-  ): Promise<void>;
+  ): void;
   function assert_equals(a: unknown, b: unknown, message?: string): void;
   function assert_not_equals(a: unknown, b: unknown, message?: string): void;
   function assert_true(val: unknown, message?: string): void;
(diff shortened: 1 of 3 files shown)


src/wpt/harness.ts
```diff
@@ -57,8 +57,13 @@ export type TestRunnerConfig = {
   [key: string]: TestRunnerOptions;
 };
 
+type Env = {
+  unsafe: { eval: (code: string) => void };
+  [key: string]: unknown;
+};
```
anonrig, line 63:
```suggestion
type Environment = {
  unsafe: { eval: (code: string) => void };
};
```

src/wpt/harness.ts
```diff
@@ -71,9 +76,12 @@ declare global {
   var errors: Error[];
   // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
   var testOptions: TestRunnerOptions;
-
   // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
   var GLOBAL: { isWindow(): boolean };
+  // eslint-disable-next-line no-var -- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis
```
anonrig, line 81: Can you disable no-var in the beginning of this file rather than every single line?

src/wpt/harness.ts
npaun: The second member `[key: string]: unknown;` is needed for the actual JS and JSON files themselves. I'll apply the rest though.

anonrig: unknown will be preferred if you leave it as it is, and every key's value will be unknown. you need something like this, if you want to correctly pass the types for unsafe key.

```
type ComboObject = {
  known: boolean;
  field: number;
} & Record<Exclude<string, "known" | "field">, string | number | boolean>;
```

anonrig:
What does this PR does that we couldn't do in the past? Can you add those changes to this pull-request as well. It's not possible to justify these changes without an example/test that comes with it.

src/wpt/harness.ts
npaun: Please let me know if I'm missing something:

The type I'm using works correctly in Typescript playground:
https://www.typescriptlang.org/play/?#code/C4TwDgpgBAogdgNwIxQLxQN4Cgq6gVzgGcBDAMwgC5MIESAbagCgGMB7AEyqiOACcAlnADmASjQA+HvyHCAvgBoceANoBrCCGq9BIgLrVCauGwDucANxY5WLAHo7UAAp82kPqCgByQqQpeoASIoAFsgollAuChQSG8MOQCAI3xgKD4IAEd8AQyOKJjwaC94ZC8AOix2YjSSalKUdASre0cAFSLvHVkAoKgTWqII4TgSJPpoYDZCuK8MKFoGZnYubRkRcVQpbpELKESqthqoJPrERsxfcm4AIhBpjhJTYRu5FocoDtmd4V7ggagJCGAhGYwmMWmsWKrE43B+m2261+lWqvCgLDOyDQl2I12o80WjCgdweTxe+32LVRaQ4mIuGCuFHxCzoRMJ+zetg+ACl8GipuiABYQFhqGLCmZUVqfTpzFlLKAw1bSXRiSQq2R7RKBf5sQbDUbjSaQ2VwWgQPgVQ7HbhmhAW7EccqMiDvRy8-nTIiCsyBNKmNh8NRELBOl3lQlMG6CgQ3UQWIA

There are also examples like this in the TS manual: https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures

jasnell:
Many of the wpt test files rely on the need to perform io at the top level scope, as you know we want to be able to run those wpt files automatically as is, without porting those to workerd syntax. This PR is the solution @npaun and I came up with to support the case. We cannot use regular `eval` since that is disabled, and we cannot use `require(...)` since those run outside of the IoContext... turns out just adding these scripts as text bindings and passing them to the `unsafe eval` binding makes it work.

jasnell (approved):

anonrig:
> Many of the wpt test files rely on the need to perform io at the top level scope, as you know we want to be able to run those wpt files automatically as is, without porting those to workerd syntax. This PR is the solution @npaun and I came up with to support the case. We cannot use regular `eval` since that is disabled, and we cannot use `require(...)` since those run outside of the IoContext... turns out just adding these scripts as text bindings and passing them to the `unsafe eval` binding makes it work.

Yes, but this PR doesn't add any of those new tests. It would be helpful if we at least enabled 1 test that uses this new code.

npaun:
It's a particularly awesome demo but there is a simple way to check this change is working. 

Whereas before `urlencoded-parser.any.js`'s Response tests would fail with:

`Error: Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope. To fix this error, perform this operation within a handler. https://developers.cloudflare.com/workers/runtime-apis/handlers/`

Now they fail with:
`TypeError: Non-utf-8 application/x-www-form-urlencoded body.`

Which shows we can now run that part of the test code.

anonrig:
Next time please add a test before merging that asserts the intended behavior.

PR #3354, opened 2025-01-17
Initial implementation of DO-attached containers API

kentonv:
Implemented here:
- Basic lifecycle (start/monitor/destroy/signal)
- Connecting to a TCP port in the container, either as a raw socket or HTTP.

Not implemented yet:
- Intercepting outgoing connections from container (`listenTcp()`).
- Starting the DO with an already-running container.
- Hibernating while container is running.
- vsocks
- stdio interception
- Any sort of local testing with workerd (only the production runtime is fully wired up for now).

### DIFF
--- a/src/workerd/api/actor-state.h
+++ b/src/workerd/api/actor-state.h
@@ -8,6 +8,7 @@
 // See actor.h for APIs used by other Workers to talk to Actors.
 
 #include <workerd/api/actor.h>
+#include <workerd/api/container.h>
 #include <workerd/io/actor-cache.h>
 #include <workerd/io/actor-id.h>
 #include <workerd/io/compatibility-date.capnp.h>
@@ -462,7 +463,9 @@ class WebSocketRequestResponsePair: public jsg::Object {
 // The type passed as the first parameter to durable object class's constructor.
 class DurableObjectState: public jsg::Object {
  public:
-  DurableObjectState(Worker::Actor::Id actorId, kj::Maybe<jsg::Ref<DurableObjectStorage>> storage);
+  DurableObjectState(Worker::Actor::Id actorId,
+      kj::Maybe<jsg::Ref<DurableObjectStorage>> storage,
+      kj::Maybe<rpc::Container::Client> container);
 
   void waitUntil(kj::Promise<void> promise);
 
@@ -470,6 +473,10 @@ class DurableObjectState: public jsg::Object {
 
   jsg::Optional<jsg::Ref<DurableObjectStorage>> getStorage() {
     return storage.map([&](jsg::Ref<DurableObjectStorage>& p) { return p.addRef(); });
+  }
+
+  jsg::Optional<jsg::Ref<Container>> getContainer() {
+    return container.map([](jsg::Ref<Container>& c) { return c.addRef(); });
   }
 
   jsg::Promise<jsg::JsRef<jsg::JsValue>> blockConcurrencyWhile(
@@ -536,6 +543,7 @@ class DurableObjectState: public jsg::Object {
     JSG_METHOD(waitUntil);
     JSG_READONLY_INSTANCE_PROPERTY(id, getId);
     JSG_READONLY_INSTANCE_PROPERTY(storage, getStorage);
+    JSG_READONLY_INSTANCE_PROPERTY(container, getContainer);
     JSG_METHOD(blockConcurrencyWhile);
     JSG_METHOD(acceptWebSocket);
     JSG_METHOD(getWebSockets);
@@ -574,6 +582,7 @@ class DurableObjectState: public jsg::Object {
  private:
   Worker::Actor::Id id;
   kj::Maybe<jsg::Ref<DurableObjectStorage>> storage;
+  kj::Maybe<jsg::Ref<Container>> container;
 
   // Limits for Hibernatable WebSocket tags.
 
--- a/src/workerd/api/actor-state.h
+++ b/src/workerd/api/actor-state.h
@@ -541,9 +541,9 @@ class DurableObjectState: public jsg::Object {
 
   JSG_RESOURCE_TYPE(DurableObjectState, CompatibilityFlags::Reader flags) {
     JSG_METHOD(waitUntil);
-    JSG_READONLY_INSTANCE_PROPERTY(id, getId);
-    JSG_READONLY_INSTANCE_PROPERTY(storage, getStorage);
-    JSG_READONLY_INSTANCE_PROPERTY(container, getContainer);
+    JSG_LAZY_INSTANCE_PROPERTY(id, getId);
+    JSG_LAZY_INSTANCE_PROPERTY(storage, getStorage);
+    JSG_LAZY_INSTANCE_PROPERTY(container, getContainer);
     JSG_METHOD(blockConcurrencyWhile);
     JSG_METHOD(acceptWebSocket);
     JSG_METHOD(getWebSockets);
--- a/src/workerd/api/container.c++
+++ b/src/workerd/api/container.c++
@@ -1,0 +1,14 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "container.h"
+
+#include <workerd/io/io-context.h>
+
+namespace workerd::api {
+
+Container::Container(rpc::Container::Client rpcClient)
+    : rpcClient(IoContext::current().addObject(kj::heap(kj::mv(rpcClient)))) {}
+
+}  // namespace workerd::api
--- a/src/workerd/api/container.c++
+++ b/src/workerd/api/container.c++
@@ -4,11 +4,239 @@
 
 #include "container.h"
 
+#include <workerd/api/http.h>
 #include <workerd/io/io-context.h>
 
 namespace workerd::api {
 
+// =======================================================================================
+// Basic lifecycle methods
+
 Container::Container(rpc::Container::Client rpcClient)
     : rpcClient(IoContext::current().addObject(kj::heap(kj::mv(rpcClient)))) {}
+
+void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions) {
+  JSG_REQUIRE(!running, Error, "start() cannot be called on a container that is already running.");
+
+  StartupOptions options = kj::mv(maybeOptions).orDefault({});
+
+  auto req = rpcClient->startRequest();
+  KJ_IF_SOME(entrypoint, options.entrypoint) {
+    auto list = req.initEntrypoint(entrypoint.size());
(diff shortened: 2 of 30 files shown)


src/workerd/api/actor-state.h
```diff
@@ -536,6 +543,7 @@ class DurableObjectState: public jsg::Object {
     JSG_METHOD(waitUntil);
     JSG_READONLY_INSTANCE_PROPERTY(id, getId);
     JSG_READONLY_INSTANCE_PROPERTY(storage, getStorage);
+    JSG_READONLY_INSTANCE_PROPERTY(container, getContainer);
```
jasnell, line 546: Nit: these could probably be `JSG_LAZY_READONLY_INSTANCE_PROPERTY`

src/workerd/api/container.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include <workerd/io/container.capnp.h>
+#include <workerd/io/io-own.h>
+#include <workerd/jsg/jsg.h>
+
+namespace workerd::api {
+
+class Fetcher;
+
+class Container: public jsg::Object {
```
jasnell, line 18: Can you add code comments explaining what a Container is

src/workerd/api/container.h
```diff
@@ -0,0 +1,71 @@
...
+  IoOwn<rpc::Container::Client> rpcClient;
+
+  // TODO(containers): Actually check if the container is already running when the DO starts.
+  bool running = false;
+
+  kj::Maybe<jsg::Value> destroyReason;
+
+  void visitForGc(jsg::GcVisitor& visitor) {
+    visitor.visit(destroyReason);
+  }
```
jasnell, line 63: Nit: implement `visitForMemoryInfo(...)` for the `destroyReason` field.

src/workerd/api/container.c++
```diff
@@ -0,0 +1,242 @@
...
+  auto req = rpcClient->startRequest();
+  KJ_IF_SOME(entrypoint, options.entrypoint) {
+    auto list = req.initEntrypoint(entrypoint.size());
+    for (auto i: kj::indices(entrypoint)) {
+      list.set(i, entrypoint[i]);
+    }
+  }
+  req.setEnableInternet(options.enableInternet);
+
+  IoContext::current().addTask(req.send().ignoreResult());
```
jasnell, line 32: Nit: Might be just a slight bit better to `auto& ioContext = IoContext::current()` before  the `auto req = rpcClient->startRequest()` just in case the `IoContext::current()` fails. Not crtical and definitely an edge case tho.

src/workerd/api/container.c++
```diff
@@ -0,0 +1,242 @@
...
+
+  running = true;
+}
+
+jsg::Promise<void> Container::monitor(jsg::Lock& js) {
+  JSG_REQUIRE(running, Error, "monitor() cannot be called on a container that is not running.");
+
+  return IoContext::current()
+      .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).send().ignoreResult())
+      .then(js, [this](jsg::Lock& js) {
```
jasnell, line 42: safer to capture `self=JSG_THIS` rather than `this`? It's unlikely that the container will disappear while this is pending but likely best to be safe?

src/workerd/api/container.c++
```diff
@@ -0,0 +1,242 @@
...
+    }
+  }, [this](jsg::Lock& js, jsg::Value&& error) {
+    running = false;
+    destroyReason = kj::none;
+    js.throwException(kj::mv(error));
+  });
+}
+
+jsg::Promise<void> Container::destroy(jsg::Lock& js, jsg::Optional<jsg::Value> error) {
+  if (!running) return js.resolvedPromise();
```
jasnell, line 57: If destroy is called multiple times it looks like this will end up sending the destroyRequest multiple times? Also, it ends up creating and returning multiple promises. Once it has been called, it likely should be arranged to return the same js promise instance back.

src/workerd/api/container.c++
```diff
@@ -0,0 +1,242 @@
...
+  }
+  req.setEnableInternet(options.enableInternet);
+
+  IoContext::current().addTask(req.send().ignoreResult());
+
+  running = true;
+}
+
+jsg::Promise<void> Container::monitor(jsg::Lock& js) {
+  JSG_REQUIRE(running, Error, "monitor() cannot be called on a container that is not running.");
```
jasnell, line 38: What, if anything, should happen if this is called after "destroy()" is called but the destroy promise is still pending?

src/workerd/api/container.c++
```diff
@@ -0,0 +1,242 @@
...
+  if (destroyReason == kj::none) {
+    destroyReason = kj::mv(error);
+  }
+
+  return IoContext::current().awaitIo(
+      js, rpcClient->destroyRequest(capnp::MessageSize{4, 0}).send().ignoreResult());
+}
+
+void Container::signal(jsg::Lock& js, int signo) {
+  JSG_REQUIRE(signo > 0 && signo <= 64, TypeError, "Invalid signal number.");
```
jasnell, line 68: Nit: should be a `RangeError`?

src/workerd/api/container.c++
```diff
@@ -0,0 +1,242 @@
...
+jsg::Promise<void> Container::monitor(jsg::Lock& js) {
+  JSG_REQUIRE(running, Error, "monitor() cannot be called on a container that is not running.");
+
+  return IoContext::current()
+      .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).send().ignoreResult())
+      .then(js, [this](jsg::Lock& js) {
+    running = false;
+    KJ_IF_SOME(d, destroyReason) {
+      jsg::Value error = kj::mv(d);
+      destroyReason = kj::none;
```
jasnell, line 46: Non-blocking: this pattern still bugs. It's so easy for forget to set the var to `kj::none` after moving.

src/workerd/api/actor-state.h
kentonv: That does seem like a win, since none of these can actually change value.

Additionally, there's probably no reason for them to be read-only. In fact, making the new member read-only is arguably a breaking change. So I'm adding a commit to make all three be `JSG_LAZY_INSTANCE_PROPERTY`, for consistency.

src/workerd/api/container.c++
kentonv: `rpcClient` is an `IoOwn` so dereferencing it already does a more-specific check.

src/workerd/api/container.c++
kentonv: The JSG layer already guarantees that the JSG object cannot be GC'd until the returned Promise is still running.

src/workerd/api/container.c++
kentonv: Yes, I think this is fine as-is. If the RPC gets to the container engine and it turns out the container is already dead from a previous destroy() call, the container engine will just return without doing anything, which achieves the intended behavior.

src/workerd/api/container.c++
kentonv: Should be fine as-is. When the monitor() RPC reaches the container engine, the engine will see the container is already shut down, and will just return immediately with the same result as if monitor() were called before destroy().

src/workerd/api/container.c++
jasnell: Ok. I do think it would be slightly better to prevent multiple calls but if you're certain its being handled correctly below it's unlikely to be an issue.

kentonv:
Fixups:
* 543bac342c20abe59b546cf1fd8e2585cad9a8c9
* e3a66d1dd8ba177400b42487ea5784103bb8ea29
* 6fcd2f3f4a3f1d869194ebc8775c1fcee018bc5e

New commits:
* 263525eed3a89a2e5ffbce20a9b6f558c217404f
* 3eca402ce07d705fa1aa1ca5f4bb2b8f8763f600

src/workerd/api/container.c++
kentonv: 6fcd2f3f4a3f1d869194ebc8775c1fcee018bc5e

src/workerd/api/container.h
kentonv: e3a66d1dd8ba177400b42487ea5784103bb8ea29

src/workerd/api/container.h
kentonv: 543bac342c20abe59b546cf1fd8e2585cad9a8c9

src/workerd/api/actor-state.h
kentonv: 263525eed3a89a2e5ffbce20a9b6f558c217404f

shrima-cf (approved):

src/workerd/api/container.h
```diff
@@ -0,0 +1,79 @@
...
+// etc.
+class Container: public jsg::Object {
+ public:
+  Container(rpc::Container::Client rpcClient);
+
+  struct StartupOptions {
+    jsg::Optional<kj::Array<kj::String>> entrypoint;
+    bool enableInternet = false;
+
+    // TODO(containers): stdio interception
```
jasnell, line 30: Nit: would be helpful if you could expand this comment to provide more detail about what is intended for someone else coming into this code later on.

src/workerd/api/container.h
```diff
@@ -0,0 +1,79 @@
...
+  bool getRunning() {
+    return running;
+  }
+
+  // Methods correspond closely to the RPC interface in `container.capnp`.
+  void start(jsg::Lock& js, jsg::Optional<StartupOptions> options);
+  jsg::Promise<void> monitor(jsg::Lock& js);
+  jsg::Promise<void> destroy(jsg::Lock& js, jsg::Optional<jsg::Value> error);
+  void signal(jsg::Lock& js, int signo);
+  jsg::Ref<Fetcher> getTcpPort(jsg::Lock& js, int port);
```
jasnell, line 44: Nit: The naming of this is a bit weird and could use a code comment about what is intended here. The `get{whatever}` pattern is most typically seen for property getters but this is a method that takes a port? Just feels a bit awkward.

jasnell:
Couple of nits remaining but otherwise LGTM

src/workerd/api/container.h
kentonv: I think there's plenty of precedent for functions that start with `get` but are not simple property getters. E.g. `DO_NAMESPACE.get(id)`.

Per the comment above this code block, all these methods directly correspond to methods on the capnp interface where they are documented in detail. I didn't want to copy all the comments and have to keep them in sync.

kentonv: That said if you have a better naming idea I'm totally open to suggestions. This interface won't be public for a while.

jasnell: No renaming suggestions but a code comment explaining that would be worthwhile for future folks coming into this code.

PR #3358, opened 2025-01-18
Add crc32 algorithm to DigestStream

danlapid:
Since openssl does not support crc32 as a digest algorithm I extended our support through a new self implemented class and took the initial implementation out to it's own class.

Solves internal ticket #8999

### DIFF
--- a/src/workerd/api/crypto/crc-impl.c++
+++ b/src/workerd/api/crypto/crc-impl.c++
@@ -1,0 +1,85 @@
+// Copyright (c) 2017-2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "crc-impl.h"
+
+#include <array>
+#include <type_traits>
+
+namespace {
+constexpr auto crcTableSize = 256;
+
+template <typename T>
+concept Uint64OrUint32 = std::unsigned_integral<T> && (sizeof(T) == 8 || sizeof(T) == 4);
+
+template <Uint64OrUint32 T>
+constexpr T reverse(T value) {
+  if constexpr (sizeof(T) == 4) {
+    value = ((value & 0xaaaaaaaa) >> 1) | ((value & 0x55555555) << 1);
+    value = ((value & 0xcccccccc) >> 2) | ((value & 0x33333333) << 2);
+    value = ((value & 0xf0f0f0f0) >> 4) | ((value & 0x0f0f0f0f) << 4);
+    value = ((value & 0xff00ff00) >> 8) | ((value & 0x00ff00ff) << 8);
+    value = (value >> 16) | (value << 16);
+    return value;
+  } else {
+    value = ((value & 0xaaaaaaaaaaaaaaaa) >> 1) | ((value & 0x5555555555555555) << 1);
+    value = ((value & 0xcccccccccccccccc) >> 2) | ((value & 0x3333333333333333) << 2);
+    value = ((value & 0xf0f0f0f0f0f0f0f0) >> 4) | ((value & 0x0f0f0f0f0f0f0f0f) << 4);
+    value = ((value & 0xff00ff00ff00ff00) >> 8) | ((value & 0x00ff00ff00ff00ff) << 8);
+    value = ((value & 0xffff0000ffff0000) >> 16) | ((value & 0x0000ffff0000ffff) << 16);
+    value = (value >> 32) | (value << 32);
+    return value;
+  }
+}
+
+template <Uint64OrUint32 T>
+constexpr std::array<T, crcTableSize> gen_crc_table(T polynomial, bool reflectIn, bool reflectOut) {
+  constexpr auto numIterations = sizeof(polynomial) * 8;  // number of bits in polynomial
+  auto crcTable = std::array<T, crcTableSize>{};
+
+  for (T byte = 0u; byte < crcTableSize; ++byte) {
+    T crc = (reflectIn ? (reverse(T(byte)) >> (numIterations - 8)) : byte);
+
+    for (int i = 0; i < numIterations; ++i) {
+      if (crc & (static_cast<T>(1) << (numIterations - 1))) {
+        crc = (crc << 1) ^ polynomial;
+      } else {
+        crc <<= 1;
+      }
+    }
+
+    crcTable[byte] = (reflectOut ? reverse(crc) : crc);
+  }
+
+  return crcTable;
+}
+
+// https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-32-iscsi
+constexpr auto crc32c_table = gen_crc_table(static_cast<uint32_t>(0x1edc6f41), true, true);
+// https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-64-nvme
+constexpr auto crc64nvme_table =
+    gen_crc_table(static_cast<uint64_t>(0xad93d23594c93659), true, true);
+}  // namespace
+
+uint32_t crc32c(uint32_t crc, const uint8_t *data, unsigned int length) {
+  if (data == nullptr) {
+    return 0;
+  }
+  crc ^= 0xffffffff;
+  while (length--) {
+    crc = crc32c_table[(crc ^ *data++) & 0xffL] ^ (crc >> 8);
+  }
+  return crc ^ 0xffffffff;
+}
+
+uint64_t crc64nvme(uint64_t crc, const uint8_t *data, unsigned int length) {
+  if (data == nullptr) {
+    return 0;
+  }
+  crc ^= 0xffffffffffffffff;
+  while (length--) {
+    crc = crc64nvme_table[(crc ^ *data++) & 0xffL] ^ (crc >> 8);
+  }
+  return crc ^ 0xffffffffffffffff;
+}
--- a/src/workerd/api/crypto/crc-impl.h
+++ b/src/workerd/api/crypto/crc-impl.h
@@ -1,0 +1,14 @@
+// Copyright (c) 2017-2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#pragma once
+#include <stdint.h>
+
+// crc32-c implementation according to the spec:
+// https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-32-iscsi
+uint32_t crc32c(uint32_t crc, const uint8_t *data, unsigned int length);
+
+// crc64-nvme implementation according to the spec:
+// https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-64-nvme
+uint64_t crc64nvme(uint64_t crc, const uint8_t *data, unsigned int length);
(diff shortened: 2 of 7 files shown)


src/workerd/api/crypto/crypto.h
```diff
@@ -666,9 +666,15 @@ class SubtleCrypto: public jsg::Object {
...
+class DigestContext {
+ public:
+  virtual void write(kj::ArrayPtr<kj::byte> buffer) = 0;
+  virtual kj::Array<kj::byte> close() = 0;
+};
+
 class DigestStream: public WritableStream {
  public:
-  using DigestContextPtr = kj::Own<EVP_MD_CTX>;
+  using DigestContextPtr = kj::Own<DigestContext>;//, kj::Own<EVP_MD_CTX>>;
```
danlapid, line 677: Vestigial comment, I'll clean up before merging.

src/workerd/api/crypto/crypto.c++
```diff
@@ -14,13 +14,26 @@
...
+#include <zlib.h>
 
 #include <algorithm>
 #include <array>
 #include <limits>
 #include <set>
 #include <typeinfo>
 
+#ifndef htonl
+uint32_t htonl(uint32_t x) {
```
danlapid, line 26: Better than including different heavy annoying headers per OS imo.

src/workerd/api/crypto/crypto.c++
```diff
@@ -676,13 +689,61 @@ kj::String Crypto::randomUUID() {
...
+  CRC32DigestContext(): value(crc32(0, Z_NULL, 0)) {}
+  virtual ~CRC32DigestContext() = default;
+
+  void write(kj::ArrayPtr<kj::byte> buffer) {
+    value = crc32(value, buffer.begin(), buffer.size());
+  }
+
+  kj::Array<kj::byte> close() {
+    auto beValue = htonl(value);
+    return kj::heapArray<kj::byte>((kj::byte*)&beValue, sizeof(value));
```
danlapid, line 703: Should be sizeof(beValue) though I could also statically assert these sizes are equal

src/workerd/api/crypto/crypto.c++
```diff
@@ -676,13 +689,61 @@ kj::String Crypto::randomUUID() {
...
+    auto checkErrorsOnFinish = webCryptoOperationBegin(__func__, algorithm);
+    uint size = 0;
+    auto digest = kj::heapArray<kj::byte>(EVP_MD_CTX_size(context.get()));
+    OSSLCALL(EVP_DigestFinal_ex(context.get(), digest.begin(), &size));
+    KJ_ASSERT(size, digest.size());
+    return kj::mv(digest);
+  }
+
+ private:
+  kj::StringPtr algorithm;
```
danlapid, line 737: Arguably better to copy than to save a StringPtr, wdyt? @jsnell

anonrig:
@danlapid zlib supports crc32 that we already use in node:zlib: https://github.com/cloudflare/workerd/blob/34477c19ea75969fb554bd82b3b65e3b203b3d9f/src/workerd/api/node/zlib-util.c%2B%2B#L34

I think we can just use it rather than custom implementation.

src/workerd/api/crypto/crypto.c++
```diff
@@ -676,13 +689,61 @@ kj::String Crypto::randomUUID() {
...
+  kj::Array<kj::byte> close() {
+    auto beValue = htonl(value);
+    return kj::heapArray<kj::byte>((kj::byte*)&beValue, sizeof(value));
+  }
+
+ private:
+  uint32_t value;
+};
+
+class OpenSSLDigestContext: public DigestContext {
```
anonrig, line 710:
```suggestion
class OpenSSLDigestContext: public virtual DigestContext {
```

src/workerd/api/crypto/crypto.c++
jasnell: I'd copy here, yes.

src/workerd/api/crypto/crypto.c++
danlapid: DigestContext is a pure abstract class with no members so this won't really do anything but sure 😅🤷‍♂️

danlapid: Maybe I'll just make it final instead
That actually makes more sense.

danlapid:
@anonrig @jasnell fixed comments.
Any opposition to merging?

anonrig (approved):

src/workerd/api/tests/crypto-streams-test.js
```diff
@@ -107,6 +108,16 @@ export const digeststream = {
       deepStrictEqual(digest, check);
     }
 
+    {
+      const check = new Uint8Array([176, 224, 34, 147]);
+      const digestStream = new crypto.DigestStream('crc32');
+      const writer = digestStream.getWriter();
+      await writer.write(new Uint32Array([1, 2, 3]));
```
jasnell, line 115: This is fine but I'd like to see a test with multiple large writes.

jasnell (approved):

src/workerd/api/crypto/crypto.h
```diff
@@ -666,9 +666,15 @@ class SubtleCrypto: public jsg::Object {
 // DigestStream is a non-standard extension that provides a way of generating
 // a hash digest from streaming data. It combines Web Crypto concepts into a
 // WritableStream and is compatible with both APIs.
+class DigestContext {
+ public:
+  virtual void write(kj::ArrayPtr<kj::byte> buffer) = 0;
```
anonrig, line 671: Virtual destructor is required to prevent undefined behavior. 

(I think we should enable clang-tidy for workerd)

Ref: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/virtual-class-destructor.html

danlapid: This isn't really a problem with pure base classes with no data members but I will add a virtual destructor anyway 😄

src/workerd/api/crypto/crc-impl.c++
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-32-iscsi
+constexpr auto crc32c_table = gen_crc_table(static_cast<uint32_t>(0x1edc6f41), true, true);
+// https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-64-nvme
+constexpr auto crc64nvme_table =
+    gen_crc_table(static_cast<uint64_t>(0xad93d23594c93659), true, true);
+}  // namespace
+
+uint32_t crc32c(uint32_t crc, const uint8_t *data, unsigned int length) {
```
fhanau, line 65: We should really use the crc32c instructions available on x86_64 and arm64 [as in capnproto](https://github.com/capnproto/capnproto/blob/f600d249496c55289fa07fd6a21cadeeb340edb9/c%2B%2B/src/kj/hash.h#L226) here to have a hardware accelerated version (there's more room for improvement beyond that but according to https://dougallj.wordpress.com/2022/05/22/faster-crc32-on-the-apple-m1/ this should run at ~8GB/s, "good enough"). In capnproto this is only used to hash 4-byte values so we'd need to be using a loop here, not blocking if we want to land this quickly.

danlapid: Feel free to submit that in a followup PR.
I'd rather not deal with optimizations that I can't test on all of these different platforms and configurations to validate that they are outputting the correct result.

src/workerd/api/tests/crypto-streams-test.js
danlapid: Is this blocking or a nit?

danlapid:
@jasnell @anonrig quite a bit has changed so I reset your approvals.

Windows is passing now though sadly it required a whole endianness file for it to actually work reliably.

src/workerd/api/crypto/crc-impl.h
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017-2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#pragma once
+#include <stdint.h>
+
+uint32_t crc32c(uint32_t crc, const uint8_t *data, unsigned int length);
+uint64_t crc64nvme(uint64_t crc, const uint8_t *data, unsigned int length);
```
jasnell, line 9: Nit: code comments explaining the differences between these two would be helpful. The `nvme` is not going to be that meaningful to new folks looking at this.

src/workerd/api/crypto/endianness.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2017-2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#pragma once
+#include <stdint.h>
+
+uint16_t htobe16(uint16_t x);
```
jasnell, line 8: Some code comments explaining this would be helpful

src/workerd/api/tests/crypto-streams-test.js
jasnell: Not blocking but something that would be very helpful

anonrig (approved):

src/workerd/api/crypto/crc-impl.c++
```diff
@@ -0,0 +1,85 @@
...
+#include "crc-impl.h"
+
+#include <array>
+#include <type_traits>
+
+namespace {
+constexpr auto crcTableSize = 256;
+template <typename T>
+constexpr T reverse(T value) {
+  static_assert(std::is_same<T, uint32_t>::value || std::is_same<T, uint64_t>::value,
```
anonrig, line 14: You can change the template to use concepts to get rid of this assert call.

src/workerd/api/crypto/crc-impl.c++
```diff
@@ -0,0 +1,85 @@
...
+    value = ((value & 0xff00ff00ff00ff00) >> 8) | ((value & 0x00ff00ff00ff00ff) << 8);
+    value = ((value & 0xffff0000ffff0000) >> 16) | ((value & 0x0000ffff0000ffff) << 16);
+    value = (value >> 32) | (value << 32);
+    return value;
+  }
+}
+
+template <typename T>
+constexpr std::array<T, crcTableSize> gen_crc_table(T polynomial, bool reflectIn, bool reflectOut) {
+  static_assert(std::is_same<T, uint32_t>::value || std::is_same<T, uint64_t>::value,
```
anonrig, line 36: This can be removed with C++ concepts

src/workerd/api/crypto/crypto.c++
```diff
@@ -676,13 +679,113 @@ kj::String Crypto::randomUUID() {
 // =======================================================================================
 // Crypto Streams implementation
 
+class CRC32DigestContext final: public DigestContext {
+ public:
+  CRC32DigestContext(): value(crc32(0, Z_NULL, 0)) {}
+  virtual ~CRC32DigestContext() = default;
```
anonrig, line 685: Shouldn't this be in the parent class, not in here?

```suggestion
  ~CRC32DigestContext() = default;
```

danlapid: As I said previously, it really really doesn't matter for a pure abstract base class with no members 😅

src/workerd/api/tests/crypto-streams-test.js
danlapid: Added one, we could add more later if you have more test cases in mind there.

PR #3361, opened 2025-01-20
fix(node:timers) calling refresh breaks the timers

vicb:
Calling `TImeout#refresh` does not update `#timer` making it impossible to cancel
/refresh the timer.


jasnell (approved):

vicb:
Thanks for the review!

PR #3368, opened 2025-01-20
Add a new subcommand to workerd for making a Python snapshot

hoodmane:
For now this only makes baseline snapshots but I'll extend it to make package snapshots when I need it.

### DIFF
--- a/src/workerd/server/workerd.c++
+++ b/src/workerd/server/workerd.c++
@@ -703,6 +703,9 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
           .addSubCommand("test", KJ_BIND_METHOD(*this, getTest), "run unit tests")
           .addSubCommand("pyodide-lock", KJ_BIND_METHOD(*this, getPyodideLock),
               "outputs the package lock file used by Pyodide")
+          .addSubCommand("make-pyodide-baseline-snapshot",
+              KJ_BIND_METHOD(*this, getMakePyodideBaselineSnapshot),
+              "Make a Pyodide baseline memory snapshot")
           .build();
       // TODO(someday):
       // "validate": Loads the config and parses all the code to report errors, but then exits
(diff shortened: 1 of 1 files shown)


src/workerd/server/workerd.c++
```diff
@@ -703,6 +703,7 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
           .addSubCommand("test", KJ_BIND_METHOD(*this, getTest), "run unit tests")
           .addSubCommand("pyodide-lock", KJ_BIND_METHOD(*this, getPyodideLock),
               "outputs the package lock file used by Pyodide")
+          .addSubCommand("make-pyodide-snapshot", KJ_BIND_METHOD(*this, getMakePyodideSnapshot), "Make a Pyodide memory snapshot")
```
hoodmane, line 706: If it's possible to suppress this from the `--help` text that might be a good idea since I don't expect it to be particularly useful for downstream users.

dom96:
We already have `python-save-baseline-snapshot` (https://github.com/cloudflare/workerd/blob/dd3effdffb89dd7fdd5b1bdfa3b6425d97af9cf4/src/workerd/server/workerd.c%2B%2B#L779) for this which I think is the same as what you're adding here, right?

src/workerd/server/workerd.c++
```diff
@@ -899,6 +900,16 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
         .build();
   }
 
+  kj::MainFunc getMakePyodideSnapshot() {
+    server->allowExperimental();
+    server->setPythonCreateBaselineSnapshot();
+    auto builder = kj::MainBuilder(context, getVersionString(), "Make a Pyodide memory snapshot", "");
```
dom96, line 906: Describe here what kind of snapshot it creates

dom96 (approved):
Ahh okay, you want a nicer command that doesn't require a config.

PR #3369, opened 2025-01-20
py_ew_test: Add skip_flags argument

hoodmane:
This allows us to say which combinations don't work, which is more explicit than saying which ones do work.

Release note: None

### DIFF
--- a/src/workerd/server/tests/python/py_wd_test.bzl
+++ b/src/workerd/server/tests/python/py_wd_test.bzl
@@ -13,12 +13,14 @@ def py_wd_test(
         data = None,
         name = None,
         python_flags = "all",
+        skip_python_flags = [],
         args = [],
         size = "enormous",
         tags = [],
         **kwargs):
     if python_flags == "all":
         python_flags = FEATURE_FLAGS.keys()
+    python_flags = [flag for flag in python_flags if flag not in skip_python_flags]
     if data == None and directory != None:
         data = native.glob(
             [
--- a/src/workerd/server/tests/python/BUILD.bazel
+++ b/src/workerd/server/tests/python/BUILD.bazel
@@ -20,8 +20,7 @@ py_wd_test("env-param")
 
 py_wd_test(
     "asgi",
-    # TODO: update bundle to include asgi fix and test on 0.26.0a2
-    python_flags = ["development"],
+    skip_python_flags = ["0.27.1"],
 )
 
 py_wd_test("random")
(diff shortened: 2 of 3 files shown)


src/workerd/server/tests/python/py_wd_test.bzl
```diff
@@ -13,12 +13,14 @@ def py_wd_test(
...
         name = None,
         python_flags = "all",
+        skip_python_flags = [],
         args = [],
         size = "enormous",
         tags = [],
         **kwargs):
     if python_flags == "all":
         python_flags = FEATURE_FLAGS.keys()
+    python_flags = [flag for flag in python_flags if flag not in skip_python_flags]
```
anonrig, line 23: Nit:

Something like this might be more readable:

Python_flags.difference(skip_python_flags)

hoodmane: is `list.difference` a thing in starlark?

anonrig: It seems it's only available on sets, not lists. My bad.

hoodmane: But I don't think starlark has sets at all. I tried to do `python_flags = set(python_flags).difference(skip_python_flags)` like I'd do in Python but I get `ERROR: name 'set' is not defined`

anonrig: It seems - is the correct operator

> The - operation on two sets returns the difference of the two sets: a set containing the elements found in the left-hand side set but not the right-hand side set.

set([1, 2]) - set([2, 3])  # set([1])
set([1, 2]) - set([3, 4])  # set([1, 2])

hoodmane: But that's still Python, right? Not starlark?

danlapid (approved):

src/workerd/server/tests/python/py_wd_test.bzl
anonrig: It is Starlark - https://github.com/bazelbuild/starlark/blob/master/spec.md#sets

jclee: Hmm... looks like sets are a fairly recent addition that maybe haven't made it into that part of the docs yet: https://github.com/bazelbuild/starlark/pull/290

PR #3370, opened 2025-01-20
Update release.yml

danlapid:
Release failed in the following workflow:
https://github.com/cloudflare/workerd/actions/runs/12872327281

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-22.04, ubuntu-22.04-arm, macos-15, windows-2022]
+        os: [ubuntu-22.04, macos-15, windows-2022]
         target-arch: [ X64 ]
         include:
           - os-name: linux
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -164,6 +167,9 @@ jobs:
             cpu: ARM64
           - arch: windows-64
             name: Windows-X64
+        exclude:
```
fhanau, line 170: This exclude is not needed – only 3 default configurations are created, and linux-arm64 is not in the list.

.github/workflows/release.yml
```diff
@@ -80,6 +80,9 @@ jobs:
           - os-name: windows
             os: windows-2022
             bazel-config: release_windows
+        exclude:
```
fhanau, line 83: Nit: This works, but it's easier to remove `ubuntu-22.04-arm` from the os list instead (the `linux-arm64` configuration still gets created manually, with this we're just creating an entry for ubuntu-22.04-arm with X64 only to manually exclude it later)

fhanau (approved):

PR #3373, opened 2025-01-20
Make SQLite pass through VFS exceptions to the caller.

kentonv:
This way, if the VFS throws a JSG error, it can propagate to the application, instead of the application getting a mysterious SQLITE_IOERROR message.

### DIFF
--- a/src/workerd/util/sqlite.c++
+++ b/src/workerd/util/sqlite.c++
@@ -92,6 +92,58 @@ kj::String dbErrorMessage(int errorCode, sqlite3* db) {
   return msg.flatten();
 }
 
+// If a VFS call throws an exception, and vfsErrorListener is non-null, the exception will
+// be placed there, otherwise it will be logged. This is used to implement pass-through of KJ
+// exceptions through SQLite.
+static thread_local kj::Maybe<kj::Exception>* vfsErrorListener = nullptr;
+
+// Report that in a sqlite VFS callback, an exception was caught, and SQLITE_IOERROR is being
+// returned to SQLite.
+//
+// The exception must be caught using `catch (kj::Exception& e)`, NOT using `catch (...)` followed
+// by kj::getCaughtExceptionAsKj(). This is because the latter truncates the stack trace to show
+// only the frames between the throw and the catch. We actually want to retain the full trace
+// through SQLite.
+void reportVfsErrorCaught(kj::Exception&& e) {
+  if (vfsErrorListener != nullptr) {
+    // Only capture the first error; assume subsequent errors are side effects.
+    if (*vfsErrorListener == kj::none) {
+      *vfsErrorListener = kj::mv(e);
+    }
+  } else {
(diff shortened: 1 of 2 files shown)


kentonv:
(Had originally assigned @Frederik-Baetens to code review because I misremembered who I had recently been discussing this issue with, it was actually @jclee, so I changed it -- but will accept stamps from anyone)

src/workerd/util/sqlite.c++
```diff
@@ -92,6 +92,58 @@ kj::String dbErrorMessage(int errorCode, sqlite3* db) {
...
+  SqliteCallScope() {
+    KJ_DASSERT(vfsErrorListener == nullptr);
+    vfsErrorListener = &error;
+  }
+  ~SqliteCallScope() {
+    vfsErrorListener = nullptr;
+  }
+
+  void rethrowVfsError() {
+    KJ_IF_SOME(e, error) {
```
justin-mp, line 130: Is there ever a case where we expect error to be non-null?  (The callers all call this method `if  (_ec == SQLITE_IOERR)`, so I would expect that this should always be non-null.)  If there's no such case, perhaps we should use KJ_REQUIRE_NONNULL(error).

justin-mp (approved):

src/workerd/util/sqlite.c++
kentonv: I am not sure whether sqlite could cough up a SQLITE_IOERR error in some case other than when the VFS fails. I figured if that does happen it's probably better to report a generic SQLITE_IOERR than some KJ null assertion failure.

src/workerd/util/sqlite.c++
```diff
@@ -92,6 +92,58 @@ kj::String dbErrorMessage(int errorCode, sqlite3* db) {
...
+ public:
+  SqliteCallScope() {
+    KJ_DASSERT(vfsErrorListener == nullptr);
+    vfsErrorListener = &error;
+  }
+  ~SqliteCallScope() {
+    vfsErrorListener = nullptr;
+  }
+
+  void rethrowVfsError() {
```
jasnell, line 129: Nit/Bikeshed: `maybeRethrowVfsError()` since there might not be anything to rethrow.

src/workerd/util/sqlite.c++
```diff
@@ -92,6 +92,58 @@ kj::String dbErrorMessage(int errorCode, sqlite3* db) {
...
+}
+
+// Implements SQLITE_CALL_SCOPE.
+class SqliteCallScope {
+ public:
+  SqliteCallScope() {
+    KJ_DASSERT(vfsErrorListener == nullptr);
+    vfsErrorListener = &error;
+  }
+  ~SqliteCallScope() {
```
jasnell, line 125: Nit: It's likely going to be easy for someone to forget to call `rethrowVfsError`. Should the destructor do it for the user it is hasn't already been done? Or maybe emit a warning if `error` is not `kj::none` and rethrow hasn't been called?

PR #3379, opened 2025-01-21
Embeds Python package locks in the binary.

dom96:


build/python_metadata.bzl
```diff
@@ -0,0 +1,5 @@
+# The below is a list of pyodide-lock.json files for each package bundle version that we support.
+# Each of these gets embedded in the workerd and EW binary.
+PYTHON_LOCKFILES = {
+    "20240829.4": "c2d9c67ea55a672b95a3beb8d66bfbe7df736edb4bb657383b263151e7e85ef4",
```
hoodmane, line 4: What is this value here? Presumably we should update the package automation to put it in the automatically generated `pyodide_bucket.bzl` and import it from there.

src/pyodide/BUILD.bazel
```diff
@@ -44,21 +55,17 @@ capnp_embed(
...
     srcs = ["generated/pyodide_extra.capnp"],
     visibility = ["//visibility:public"],
     deps = [
         ":pyodide_extra_file_embed",
-        ":pyodide_lock_file_embed",
         ":pyodide_packages_archive_embed",
         ":python_entrypoint_file_embed",
+    ] + [
+        ":pyodide_lock_" + package_date + "_file_embed"
+        for package_date in PYTHON_LOCKFILES.keys()
```
hoodmane, line 68: Why don't we make a lockfiles `filegroup`?

hoodmane:
Generally looks good to me.

build/python_metadata.bzl
dom96: I'll add a comment to explain what the value is but it is just the sha256 hash of the lock file.

Do we want the package automation to enumerate all the old package bundle releases so it can generate the full list of package dates here?

src/pyodide/BUILD.bazel
dom96: What's the advantage of doing this?

hoodmane (approved):

Issue #3364, opened 2025-01-20
node:timers/promises does not `signal.removeEventListener('abort')` consistently with the node implementation


vicb:
The [node implementation of `node:timers/promises`](https://github.com/nodejs/node/blob/main/lib/timers/promises.js) calls `signal.removeEventListener('abort')`
- when the promise returned by `setTimeout` is settled
- when the promise returned by `setImmediate` is settled

Our current implementation
- does never call `signal.removeEventListener('abort')` for `setTimeout`
- only calls `signal.removeEventListener('abort')` when the promise rejects (via the AbortSignal) for `setImmediate`

Should we consider implementing that?

Issue #3363, opened 2025-01-20
Tests Request: add tests for node:timers/promises


vicb:
https://github.com/cloudflare/workerd/pull/3346 adds `node:timers` and `node:timers/promises` - it only has tests for `node:timers`, it would be nice to add tests for `node:timers/promises` too.

PR #3380, opened 2025-01-21
fix node:timers abort callback reset

anonrig:
Fixes https://github.com/cloudflare/workerd/issues/3364
Fixes https://github.com/cloudflare/workerd/issues/3363

### DIFF
--- a/src/node/internal/internal_timers_promises.ts
+++ b/src/node/internal/internal_timers_promises.ts
@@ -34,6 +34,8 @@ import {
 
 const kScheduler = Symbol.for('kScheduler');
 
+type OnCancelCallback = (() => void) | undefined;
+
 export async function setTimeout<T = void>(
   delay: number | undefined,
   value?: T,
@@ -62,17 +64,21 @@ export async function setTimeout<T = void>(
   }
 
   const { promise, resolve, reject } = Promise.withResolvers<T>();
+  let onCancel: OnCancelCallback;
 
   const timer = timers.setTimeout(() => {
     resolve(value as T);
-  }, delay ?? 0);
+    if (onCancel) {
+      signal?.removeEventListener('abort', onCancel);
+    }
+  }, delay ?? 1);
 
   if (signal) {
-    function onCancel(): void {
+    onCancel = (): void => {
       timers.clearTimeout(timer);
-      reject(new AbortError(undefined, { cause: signal?.reason }));
-    }
-    signal.addEventListener('abort', onCancel);
+      reject(new AbortError(undefined, { cause: signal.reason }));
+    };
+    signal.addEventListener('abort', onCancel, { once: true });
   }
 
   return promise;
@@ -101,18 +107,21 @@ export async function setImmediate<T>(
   }
 
   const { promise, resolve, reject } = Promise.withResolvers<T>();
+  let onCancel: OnCancelCallback;
 
   const timer = timers.setImmediate(() => {
     resolve(value as T);
+    if (onCancel) {
+      signal?.removeEventListener('abort', onCancel);
+    }
   });
 
   if (signal) {
-    function onCancel(): void {
+    onCancel = (): void => {
       timers.clearImmediate(timer);
-      signal?.removeEventListener('abort', onCancel);
-      reject(new AbortError(undefined, { cause: signal?.reason }));
-    }
-    signal.addEventListener('abort', onCancel);
+      reject(new AbortError(undefined, { cause: signal.reason }));
+    };
+    signal.addEventListener('abort', onCancel, { once: true });
   }
 
   return promise;
@@ -144,7 +153,7 @@ export async function* setInterval<T = void>(
     throw new AbortError(undefined, { cause: signal.reason });
   }
 
-  let onCancel: (() => void) | undefined;
+  let onCancel: OnCancelCallback;
   let interval: timers.Timeout;
   try {
     let notYielded = 0;
@@ -164,14 +173,12 @@ export async function* setInterval<T = void>(
     if (signal) {
       onCancel = (): void => {
         timers.clearInterval(interval);
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        signal.removeEventListener('abort', onCancel!);
         callback?.(
           Promise.reject(new AbortError(undefined, { cause: signal.reason }))
         );
         callback = undefined;
       };
-      signal.addEventListener('abort', onCancel);
+      signal.addEventListener('abort', onCancel, { once: true });
     }
 
     while (!signal?.aborted) {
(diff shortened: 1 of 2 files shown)


src/node/internal/internal_timers_promises.ts
```diff
@@ -62,16 +64,23 @@ export async function setTimeout<T = void>(
...
 
   const { promise, resolve, reject } = Promise.withResolvers<T>();
+  let onCancel: OnCancelCallback;
 
   const timer = timers.setTimeout(() => {
     resolve(value as T);
+    if (onCancel) {
+      signal?.removeEventListener('abort', onCancel);
+    }
   }, delay ?? 0);
```
vicb, line 74: Not directly related to this PR but looks like Node Timeout use 1 when undefined

```suggestion
    }
  }, delay ?? 1);
```

src/node/internal/internal_timers_promises.ts
```diff
@@ -62,16 +64,23 @@ export async function setTimeout<T = void>(
...
 
   if (signal) {
-    function onCancel(): void {
+    onCancel = (): void => {
       timers.clearTimeout(timer);
-      reject(new AbortError(undefined, { cause: signal?.reason }));
-    }
+      if (onCancel) {
+        signal.removeEventListener('abort', onCancel);
+      }
```
vicb, line 81: this code can only be excuted when onCancel is defined
```suggestion
      signal.removeEventListener('abort', onCancel);
```

src/node/internal/internal_timers_promises.ts
```diff
@@ -62,16 +64,23 @@ export async function setTimeout<T = void>(
   }
 
   const { promise, resolve, reject } = Promise.withResolvers<T>();
+  let onCancel: OnCancelCallback;
 
   const timer = timers.setTimeout(() => {
     resolve(value as T);
+    if (onCancel) {
+      signal?.removeEventListener('abort', onCancel);
+    }
```
vicb, line 73: I believe that's equivalent 

```suggestion
    signal?.removeEventListener('abort', onCancel);
```

so you probably don't need `let onCancel`

src/node/internal/internal_timers_promises.ts
```diff
@@ -101,17 +110,23 @@ export async function setImmediate<T>(
   }
 
   const { promise, resolve, reject } = Promise.withResolvers<T>();
+  let onCancel: OnCancelCallback;
```
vicb, line 113: Same comments as for setTimeout

src/node/internal/internal_timers_promises.ts
```diff
@@ -144,7 +159,7 @@ export async function* setInterval<T = void>(
     throw new AbortError(undefined, { cause: signal.reason });
   }
 
-  let onCancel: (() => void) | undefined;
+  let onCancel: OnCancelCallback;
```
vicb, line 162: I believe you can drop `let onCancel` and `if (onCancel)` below ?

jasnell:
Can you add tests?

vicb:
> Can you add tests?

Agreed, see https://github.com/cloudflare/workerd/issues/3363

Note: I just updated my comments, you do need the `let onCancel`

src/node/internal/internal_timers_promises.ts
anonrig: This won't work with since delay can still be undefined. I replaced it with ?? 1

src/node/internal/internal_timers_promises.ts
anonrig: This won't work because it can be undefined if no signal is provided. Changing this line requires adding non-null assertion, which also requires adding a eslint-ignore-next-line comment

src/node/internal/internal_timers_promises.ts
anonrig: Second parameter of removeEVentListener should not be undefined. This won't work.

anonrig:
> Can you add tests?

Added. This also closes https://github.com/cloudflare/workerd/issues/3363

src/workerd/api/node/tests/timers-nodejs-test.js
```diff
@@ -155,3 +156,343 @@ export const testRefresh = {
...
+
+// Tests are taken from
+// https://github.com/nodejs/node/blob/7bc2946293757389468f1fa09714860f8e1147b7/test/parallel/test-timers-immediate-promisified.js
+export const timersImmediatePromise = {
+  async test() {
+    {
+      const ac = new AbortController();
+      const signal = ac.signal;
+      const { promise, resolve } = Promise.withResolvers();
+      rejects(timersPromises.setImmediate(10, { signal }), /AbortError/).then(
```
jasnell, line 168: Would this work? Might be a bit nicer.

```
queueMicrotask(() => ac.abort());
await rejects(timersPromises.setImmediate(10, { signal }));
```

Or perhaps

```
await rejects(timersPromises.setImmediate(10, { signal: AbortSignal.timeout(1) });
```

src/node/internal/internal_timers_promises.ts
```diff
@@ -62,16 +64,23 @@ export async function setTimeout<T = void>(
...
 
   const { promise, resolve, reject } = Promise.withResolvers<T>();
+  let onCancel: OnCancelCallback;
 
   const timer = timers.setTimeout(() => {
     resolve(value as T);
-  }, delay ?? 0);
+    if (onCancel) {
+      signal?.removeEventListener('abort', onCancel);
+    }
```
jasnell, line 73: Another change that might be worthwhile... on the `addEventListener('abort', ...)` call, it would probably be a good idea to add the `once: true` option..

src/workerd/api/node/tests/timers-nodejs-test.js
anonrig: Yeah, I just converted existing Node.js tests.

jasnell (approved):

src/node/internal/internal_timers_promises.ts
anonrig: Yeah agreed. Applying it now.

vicb (approved):
Thanks for adding test 🎉

vicb:
There seems to be one legit test failure

vicb:
@anonrig IMO it's better to avoid force pushed after review so that it's easier to see what has changed

anonrig:
> @anonrig IMO it's better to avoid force pushes after reviews so that it's easier to see what has changed

There was a git conflict.

src/workerd/api/node/tests/timers-nodejs-test.js
anonrig: This update broke ASAN tests

anonrig:
I reverted one of the recommendations since it broke ASAN. Here's the correct test (as same as Node.js) that didn't cause ASAN errors:

```js
    {
      const signal = ac.signal;
      rejects(timersPromises.setImmediate(10, { signal }), /AbortError/).then(
        resolve
      );
      ac.abort();
      await promise;
    }
```

PR #3381, opened 2025-01-21
update util.getCallSites() implementation to follow node.js

anonrig:
This is an experimental API and just got recently changed.

Latest update: https://github.com/nodejs/node/pull/56584

### DIFF
--- a/src/node/util.ts
+++ b/src/node/util.ts
@@ -206,9 +206,12 @@ export function deprecate(
   return fn;
 }
 
-export function getCallSite(frames: number = 10) {
-  return utilImpl.getCallSite(frames);
-}
+// Node.js originally introduced the API with the name `getCallSite()` as an experimental
+// API but then renamed it to `getCallSites()` soon after. We had already implemented the
+// API with the original name in a release. To avoid the possibility of breaking, we export
+// the function using both names.
+export const getCallSite = utilImpl.getCallSites.bind(utilImpl);
+export const getCallSites = utilImpl.getCallSites.bind(utilImpl);
 
 export function isDeepStrictEqual(a: unknown, b: unknown): boolean {
   return _isDeepStrictEqual(a, b);
@@ -247,6 +250,7 @@ export default {
   transferableAbortController,
   transferableAbortSignal,
   getCallSite,
+  getCallSites,
   isDeepStrictEqual,
   isArray,
 };
--- a/src/workerd/api/node/util.c++
+++ b/src/workerd/api/node/util.c++
@@ -177,22 +177,29 @@ jsg::Name UtilModule::getResourceTypeInspect(jsg::Lock& js) {
   return js.newApiSymbol("kResourceTypeInspect"_kj);
 }
 
-kj::Array<UtilModule::CallSiteEntry> UtilModule::getCallSite(jsg::Lock& js, int frames) {
-  JSG_REQUIRE(
-      frames >= 1 && frames <= 200, Error, "Frame count should be between 1 and 200 inclusive."_kj);
-  auto stack = v8::StackTrace::CurrentStackTrace(js.v8Isolate, frames + 1);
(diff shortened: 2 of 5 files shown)


src/workerd/api/node/util.h
```diff
@@ -212,11 +212,11 @@ class UtilModule final: public jsg::Object {
     kj::String functionName;
     kj::String scriptName;
     int lineNumber;
-    int column;
```
jasnell, line 215: Did this already go out in a release? If it did, unfortunately we'll need to support both `column` and `columnNumber` as we have no way of knowing if this will break anyone.

src/workerd/api/node/util.h
```diff
@@ -212,11 +212,11 @@ class UtilModule final: public jsg::Object {
...
+    int columnNumber;
 
     JSG_STRUCT(functionName, scriptName, lineNumber, column);
```
jasnell, line 217: `columnNumber` is missing from the `JSG_STRUCT`

src/workerd/api/node/util.h
```diff
@@ -212,11 +212,11 @@ class UtilModule final: public jsg::Object {
...
   };
-  kj::Array<CallSiteEntry> getCallSite(jsg::Lock& js, int frames);
+  kj::Array<CallSiteEntry> getCallSites(jsg::Lock& js, int frames);
```
jasnell, line 219: If this already went out in a release we'll need to continue to support both `getCallSite()` and `getCallSites()` unfortunately. You can introduce a compatibility flag to expose one or the other or just export both in parallel as aliases.

src/workerd/api/node/util.h
anonrig: Should we add a compat_date for this, or just add a new value?

src/node/util.ts
```diff
@@ -206,8 +206,13 @@ export function deprecate(
   return fn;
 }
 
-export function getCallSite(frames: number = 10) {
-  return utilImpl.getCallSite(frames);
+// We leave this implementation for compat reasons.
```
jasnell, line 209:
```suggestion
// Node.js originally introduced the API with the name `getCallSite()` as an experimental
// API but then renamed it to `getCallSites()` soon after. We had already implemented the
// API with the original name in a release. To avoid the possibility of breaking, we export
// the function using both names.
```

src/node/util.ts
```diff
@@ -206,8 +206,13 @@ export function deprecate(
...
+export function getCallSite(frames: number = 10): Record<string, string>[] {
+  return utilImpl.getCallSites(frames);
+}
```
jasnell, line 212: Perhaps instead...

```js
export const getCallSite = getCallSites;
```

So it's really just an alias.

src/workerd/api/node/util.c++
```diff
@@ -193,6 +193,7 @@ kj::Array<UtilModule::CallSiteEntry> UtilModule::getCallSite(jsg::Lock& js, int
       .functionName = js.toString(stack_frame->GetFunctionName()),
       .scriptName = js.toString(stack_frame->GetScriptName()),
       .lineNumber = stack_frame->GetLineNumber(),
+      .columnNumber = stack_frame->GetColumn(),
```
jasnell, line 196:
```suggestion
      // Node.js originally implemented the experimental API using the "column" field
      // then later renamed it to columnNumber. We had already implemented the API
      // using column. To ensure backwards compat without the complexity of a compat
      // flag, we just export both.
      .columnNumber = stack_frame->GetColumn(),
```

src/node/util.ts
anonrig: This requires changing the function signature to have a default value of 10, since C++ doesn't handle it right now. Doing it now. Just FYI.

jasnell (approved):

Issue #3377, opened 2025-01-21
node:timers breaks for Workers without the `nodejs_compat_v2` compatibility flag


CarmenPopoviciu:
During our regular `workerd` [version bump](https://github.com/cloudflare/workers-sdk/pull/7836) in `workers-sdk`, some of our unit/fixture tests started [failing](https://github.com/cloudflare/workers-sdk/actions/runs/12882249162/job/35914162612?pr=7836) with

```
TypeError: Cannot read properties of undefined (reading 'bind')
 ❯ node-internal:internal_timers:107:53
```

After some investigation, we realised that changing the compatibility date of the Workers under test to `2024-09-23` (from `2024-01-01`) fixes the issue. After further investigation (thank you @vicb ♥ ), we found that the root cause is that the new implementation of node:timers relies on globalThis.setImmediate[1] which is only available with nodejs_compat_v2[2]

This means that the current implementation of node:timers might break users having a compat date before 2024-09-23.

[1] https://github.com/cloudflare/workerd/blob/b982f8bce58e4f0b713e19ad328fa3471047d75c/src/node/internal/internal_timers.ts#L137-L139
[2] https://github.com/cloudflare/workerd/blob/b982f8bce58e4f0b713e19ad328fa3471047d75c/src/workerd/api/global-scope.h#L387-L388

vicb:
Updating https://github.com/cloudflare/workerd/blob/2bef052863347425c7a806399aad3adb18cb93ac/src/workerd/api/node/tests/timers-nodejs-test.wd-test#L10 to a date before 2024-09-23 will make the test fail.

anonrig:
Thanks for reporting. I've opened a PR that addresses this issue.

PR #3382, opened 2025-01-21
use internal setImmediate/clearImmediate fns for node:timers

anonrig:
Fixes https://github.com/cloudflare/workerd/issues/3377

### DIFF
--- a/src/node/internal/internal_timers_promises.ts
+++ b/src/node/internal/internal_timers_promises.ts
@@ -102,13 +102,13 @@ export async function setImmediate<T>(
 
   const { promise, resolve, reject } = Promise.withResolvers<T>();
 
-  const timer = globalThis.setImmediate(() => {
+  const timer = timers.setImmediate(() => {
     resolve(value as T);
   });
 
   if (signal) {
     function onCancel(): void {
-      globalThis.clearImmediate(timer);
+      timers.clearImmediate(timer);
       signal?.removeEventListener('abort', onCancel);
       reject(new AbortError(undefined, { cause: signal?.reason }));
     }
--- a/src/workerd/api/node/timers.c++
+++ b/src/workerd/api/node/timers.c++
@@ -1,0 +1,27 @@
+#include "timers.h"
+
+#include <workerd/api/global-scope.h>
+#include <workerd/jsg/jsg.h>
+
+namespace workerd::api::node {
+
+// The setImmediate/clearImmediate methods are only exposed on globalThis if the
+// node_compat_v2 flag is set. However, we want them exposed via `node:timers`
+// generally when just the original node_compat is enabled. Therefore, we provide
+// this alternative route to the implementations on ServiceWorkerGlobalScope.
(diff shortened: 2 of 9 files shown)


src/workerd/api/node/timers.c++
```diff
@@ -0,0 +1,23 @@
...
+  return global.setImmediate(js, kj::mv(function), kj::mv(args));
+}
+
+void TimersUtil::clearImmediate(jsg::Lock& js, kj::Maybe<jsg::Ref<Immediate>> maybeImmediate) {
+  auto context = js.v8Context();
+  auto& global =
+      jsg::extractInternalPointer<ServiceWorkerGlobalScope, true>(context, context->Global());
+  global.clearImmediate(kj::mv(maybeImmediate));
+}
+}  // namespace workerd::api::node
```
jasnell, line 23:
```suggestion
}  // namespace workerd::api::node

```

src/workerd/api/node/timers.c++
```diff
@@ -0,0 +1,23 @@
+#include "timers.h"
+
+#include <workerd/api/global-scope.h>
+#include <workerd/jsg/jsg.h>
+
+namespace workerd::api::node {
+
```
jasnell, line 7:
```suggestion
// The setImmediate/clearImmediate methods are only exposed on globalThis if the
// node_compat_v2 flag is set. However, we want them exposed via `node:timers`
// generally when just the original node_compat is enabled. Therefore, we provide
// this alternative route to the implementations on ServiceWorkerGlobalScope.
```

src/node/internal/internal_timers_promises.ts
```diff
@@ -24,6 +24,7 @@
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import * as timers from 'node-internal:internal_timers';
+import { default as timersUtil } from 'node-internal:timers';
```
vicb, line 27: alternative: use `timers.setImmediate()` and `timers.clearImmediate()`

vicb (commented):
Could you please update the compat date in the test to the oldest possible compat date (https://github.com/cloudflare/workerd/issues/3377#issuecomment-2604460864)

jasnell:
> Could you please update the compat date in the test to the oldest possible compat date (https://github.com/cloudflare/workerd/issues/3377#issuecomment-2604460864)

Either this or the `nodejs_compat_v2` compat flag can be explicitly disabled. The point would be that the test needs to show these methods work via `node:timers` without `nodejs_compat_v2` enabled.

vicb:
> > Could you please update the compat date in the test to the oldest possible compat date ([#3377 (comment)](https://github.com/cloudflare/workerd/issues/3377#issuecomment-2604460864))
> 
> Either this or the `nodejs_compat_v2` compat flag can be explicitly disabled. The point would be that the test needs to show these methods work via `node:timers` without `nodejs_compat_v2` enabled.

But using the oldest compat date has the benefit to test all the flags and doing it everywhere (as possible) could prevent the issue we had here

PR #3389, opened 2025-01-21
Add diagnostic logging to determine import assertion use

jasnell:
The specification for import assertions strongly recommends that embedders reject import attributes and types they do not understand/implement. This is because import attributes can alter the interpretation of a module and are considered to be part of the unique key for caching a module. Throwing an error for things we do not understand is the safest thing to do. However, historically we have not followed this guideline in the spec and it's not clear if enforcing that constraint would be breaking so let's first try to determine if anyone is making use of import assertions. If we're lucky, we won't receive any hits on this and we can start enforcing the rule without a compat flag.


anonrig (approved):

jasnell:
https://github.com/tc39/proposal-import-attributes

Reference: 
image

npaun (approved):

PR #3390, opened 2025-01-22
Improve releasing by merging npm and release jobs.

danlapid:
Merge npm.yml and npm-types.yml into the release workflow. 
This will remove the necessity to run the npm and npm-types jobs manually.

There can be further work done to remove duplication of bazel setup inside of this file and with the _bazel.yml file.
That improvement does not need to block this PR as this PR does not make this problem any worse than it was before.

There are other similar improvements that can honestly be done in this file that I also punted from achieving in this PR.

The fact that this entire file runs only on release makes it really hard to iterate and I'd rather change as little as possible.


.github/workflows/release.yml
```diff
@@ -228,3 +249,113 @@ jobs:
...
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Use Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
anonrig, line 266: This should be 20, but I think we should go ahead on a different PR and use `node-version-file` and make sure our bazel configuration uses that too. Right now, it's too hard to maintain the same node version in CI, bazel etc.

.github/workflows/release.yml
```diff
@@ -228,3 +249,113 @@ jobs:
...
+        run: |
+          bazel build --disk_cache=~/bazel-disk-cache --strip=always --remote_cache=https://bazel:${{ secrets.BAZEL_CACHE_KEY }}@bazel-remote-cache.devprod.cloudflare.dev --config=ci --config=release_linux //types:types_worker
+
+      - name: Modify package.json version
+        run: node npm/scripts/bump-version.mjs npm/workerd/package.json
+        env:
+          WORKERD_VERSION: ${{ needs.version.outputs.version }}
+          LATEST_COMPATIBILITY_DATE: ${{ needs.version.outputs.date }}
+      - run: mkdir -p npm/workerd/lib
+      - run: mkdir -p npm/workerd/bin
```
anonrig, line 296: TIL: It seems mkdir can have multiple folder arguments. If I'm not mistaken this should work. 

```suggestion
      - run: mkdir -p npm/workerd/lib npm/workerd/bin
```

.github/workflows/release.yml
```diff
@@ -228,3 +249,113 @@ jobs:
...
+      - name: Build node-install
+        run: npx esbuild npm/lib/node-install.ts --outfile=npm/workerd/install.js --bundle --target=node18 --define:LATEST_COMPATIBILITY_DATE="\"${LATEST_COMPATIBILITY_DATE}\"" --define:WORKERD_VERSION="\"${WORKERD_VERSION}\"" --platform=node --external:workerd --log-level=warning
```
anonrig, line 298: Nit:

This installs esbuild's latest version. I'm surprised how one of our build flow didn't break until this point.

.github/workflows/release.yml
```diff
@@ -228,3 +249,113 @@ jobs:
...
+          WORKERD_VERSION: ${{ needs.version.outputs.version }}
+          LATEST_COMPATIBILITY_DATE: ${{ needs.version.outputs.date }}
+      - name: Build package
+        run: node npm/scripts/build-shim-package.mjs
+        env:
+          WORKERD_VERSION: ${{ needs.version.outputs.version }}
+          LATEST_COMPATIBILITY_DATE: ${{ needs.version.outputs.date }}
+      - run: echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > npm/workerd/.npmrc
+      - run: cd npm/workerd && npm publish --access public --tag 'latest'
+        env:
```
anonrig, line 319: Nit:

We can simplify all similar things

```suggestion
      - run: npm publish --access public --tag 'latest'
        working-directory: 'npm/workerd'
        env:
```

.github/workflows/release.yml
danlapid: Yeah agreed, as I stated in the PR description, this PR on purpose does not change any configuration option, it was 18 in the source file, I kept it 18 here as well.

.github/workflows/release.yml
danlapid: Hah, no clue 😅

.github/workflows/release.yml
anonrig: Yes definitely. I'm just leaving comments to create awareness to this problem.

fhanau:
@penalosa

.github/workflows/release.yml
```diff
@@ -228,3 +249,112 @@ jobs:
...
+          NPM_TOKEN: ${{ secrets.NPM_ACCESS_TOKEN }}
+
+  build-and-publish-types:
+    runs-on: ubuntu-22.04
+    needs: version
```
fhanau, line 323: This should be `needs: [version, upload-artifacts]` or perhaps `needs: [version, upload-artifacts, publish-wrapper]` – we should only publish types on tagged releases, requiring upload-artifacts allows us to skip this if there isn't a tagged release. Adding publish-wrapper means only running this if publishing workerd itself has succeeded – this reduces parallelism but we really shouldn't publish types if publishing workerd itself failed

anonrig (approved):

PR #3391, opened 2025-01-22
v8 13.2

mikea:


build/deps/v8.bzl
```diff
@@ -0,0 +1,76 @@
...
+        patch_args = ["-p1"],
+        patches = ["//:patches/v8/" + p for p in PATCHES],
+        strip_prefix = "v8-" + VERSION,
+        url = "https://github.com/v8/v8/archive/refs/tags/" + VERSION + ".tar.gz",
+    )
+
+    git_repository(
+        name = "com_googlesource_chromium_icu",
+        build_file = "@v8//:bazel/BUILD.icu",
+        commit = "9408c6fd4a39e6fef0e1c4077602e1c83b15f3fb",
```
fhanau, line 58: Nit: Based on `git diff origin/13.1-lkgr..origin/13.2-lkgr DEPS`, icu was updated to 4239b1559d11d4fa66c100543eda4161e060311e.

mikea: good catch. also updated perfetto.

fhanau:
Changes themselves look good. Unfortunately, V8 [reverted](https://chromium-review.googlesource.com/c/v8/v8/+/5953132) Erik's patch to add shared linkage support for V8 that landed in 13.1, so we'll have to bring back `patches/v8/0008-Enable-V8-shared-linkage.patch` from when we had 13.0 (shared linkage is only supported in Linux but helps keep binary sizes under control, especially relevant for asan and debug jobs which might experience disk space exhaustion otherwise).

fhanau (approved):

PR #3400, opened 2025-01-23
Compute WORKERD_INDEX_URL instead of embedding from pyodide_bucket.bzl.

dom96:
This ensures that we will get the right package versions in workerd. Before these changes it would always get the same compat date, no matter what is specified by the user.

### DIFF
--- a/src/pyodide/internal/metadata.ts
+++ b/src/pyodide/internal/metadata.ts
@@ -1,5 +1,4 @@
 import { default as MetadataReader } from 'pyodide-internal:runtime-generated/metadata';
-import { default as PYODIDE_BUCKET } from 'pyodide-internal:generated/pyodide-bucket.json';
 import { default as ArtifactBundler } from 'pyodide-internal:artifacts';
 
 export const IS_WORKERD = MetadataReader.isWorkerd();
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/metadata.ts
```diff
@@ -1,17 +1,19 @@
...
 
 export const IS_WORKERD = MetadataReader.isWorkerd();
 export const IS_TRACING = MetadataReader.isTracing();
 export const SHOULD_SNAPSHOT_TO_DISK = MetadataReader.shouldSnapshotToDisk();
 export const IS_CREATING_BASELINE_SNAPSHOT =
   MetadataReader.isCreatingBaselineSnapshot();
-export const WORKERD_INDEX_URL = PYODIDE_BUCKET.PYODIDE_PACKAGE_BUCKET_URL;
+export const PACKAGES_VERSION = MetadataReader.getPackagesVersion();
+export const WORKERD_INDEX_URL =
+  'https://pyodide-packages.runtime-playground.workers.dev/' +
```
dom96, line 11: We should transition this to non-playground

dom96:
Superseeded by https://github.com/cloudflare/workerd/pull/3451

PR #3372, opened 2025-01-20
Continue work on new module registry

jasnell:
First of a series of PRs that'll finish up the new module registry impl and testing ... there's a lot more to do here, just splitting into reasonably consumable/reviewable chunks. The focus here is on completing functionality, afterwards I'll make a pass at improving basic implementation details and perf nits.

PR #3401, opened 2025-01-23
Revert "Continue work on new module registry"

mikea:
Breaks things downstream, not sure why.

Reverts cloudflare/workerd#3372


jasnell:
Specifically, breaks internal fuzzing tests. Will investigate in the next day or two then try again.

jasnell (approved):

jasnell:
Investigation is actually pointing to the internal fuzzer-asan issue is likely a false positive. The same error occurs with or without this change in various tests. It would suggest that the internal bazel build configuration is not being appropriately applied to workerd when testing such that parts of the code are instrumented and other parts are not. Still digging but that's the current hypothesis. I believe this PR just surfaced the issue in testing but did not actually introduce any actual bugs.

PR #3402, opened 2025-01-23
POC of different destinations for rpc and fetch

penalosa:
**TLDR:** This is a quick POC of a way to point RPC calls and HTTP requests to different network addresses.

## The problem

Wrangler/Miniflare uses a `workerd` `ExternalServer` to represent service bindings between Workers in local development. This is because each Worker runs in a different `workerd` process configured by `wrangler dev` (ignoring the experimental multiworker support we've been landing recently). The caller will configure:

```
external: {
	address: address,
	http: {
		style: HttpOptions_Style.PROXY,
	},
},
```

as the service binding, where `address` is a direct address to a `capnpConnectHost` configured `Socket` from the callee.

This means that all communication over that service binding will go to the service linked to the `capnpConnectHost` `Socket` on the callee `workerd`.

However, for Workers Assets this presents a problem. Communication over service bindings needs to go to a different Worker depending on whether it's RPC or HTTP (to the user worker and router worker respectively). We need a way to connect to different addresses (Miniflare exposes a `Socket` for every configured worker) based on the type of communication. 

## The proposed solution

This PR implements (with bugs, caveats, and build failures) a proposed solution which adds an additional config property to `ExternalServer`, `addressFetch`. When workerd is making a connection to an `ExternalServer`, it should use the address specified in `addressFetch` for HTTP communication and the address specified in `address` for RPC communication. If `addressFetch` is _not_ specified it should use `address` for both RPC and HTTP communication.

Wrangler/Miniflare can then leverage this by pointing `address` at the user worker's `Socket`, and `addressFetch` at the router worker's `Socket`

### DIFF
--- a/src/workerd/server/workerd.capnp
+++ b/src/workerd/server/workerd.capnp
@@ -663,6 +663,8 @@ struct ExternalServer {
   # `HttpOptions`.
 
   address @0 :Text;
+
+  addressFetch @7 :Text;
   # Address/port of the server. Optional; if not specified, then you will be required to specify
   # the address on the command line with with `--external-addr <name>=<addr>`.
   #
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -597,14 +597,21 @@ class Server::ExternalHttpService final: public Service, private kj::TaskSet::Er
       kj::Timer& timer,
       kj::EntropySource& entropySource,
       capnp::ByteStreamFactory& byteStreamFactory,
-      capnp::HttpOverCapnpFactory& httpOverCapnpFactory)
+      capnp::HttpOverCapnpFactory& httpOverCapnpFactory,
+      kj::Own<kj::NetworkAddress> addrFetchParam)
       : addr(kj::mv(addrParam)),
+        addrFetch(kj::mv(addrFetchParam)),
         inner(kj::newHttpClient(timer,
             headerTable,
             *addr,
             {.entropySource = entropySource,
               .webSocketCompressionMode = kj::HttpClientSettings::MANUAL_COMPRESSION})),
-        serviceAdapter(kj::newHttpService(*inner)),
+        innerFetch(kj::newHttpClient(timer,
+            headerTable,
+            *addrFetch,
+            {.entropySource = entropySource,
+              .webSocketCompressionMode = kj::HttpClientSettings::MANUAL_COMPRESSION})),
+        serviceAdapterFetch(kj::newHttpService(*innerFetch)),
         rewriter(kj::mv(rewriter)),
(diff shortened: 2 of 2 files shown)


src/workerd/server/workerd.capnp
```diff
@@ -663,6 +663,8 @@ struct ExternalServer {
   # `HttpOptions`.
 
   address @0 :Text;
+
+  addressFetch @7 :Text;
```
danlapid, line 667: comment required, and I think you moved the comment of the above field?

danlapid:
We'll need a test as well

danlapid:
Can't we do the differentiation in the proxy worker instead?

penalosa:
> Can't we do the differentiation in the proxy worker instead?
I'm guessing this means Wrangler's proxy worker? As far as I know, there's no way to forward on RPC requests from a worker. The proxy worker would have to have logic along the lines of "if HTTP send to worker a, if RPC send to worker b", but I don't know where that could be written. It couldn't go in `fetch() {}` since that would miss RPC methods, and I don't think there's a good way to define a catchall RPC method (AFAIK it's based on the methods available on the prototype, so exporting a Proxy as a WorkerEntrypoint wouldn't work)

danlapid:
> > Can't we do the differentiation in the proxy worker instead?
> 
> 
> 
> I'm guessing this means Wrangler's proxy worker? As far as I know, there's no way to forward on RPC requests from a worker. The proxy worker would have to have logic along the lines of "if HTTP send to worker a, if RPC send to worker b", but I don't know where that could be written. It couldn't go in `fetch() {}` since that would miss RPC methods, and I don't think there's a good way to define a catchall RPC method (AFAIK it's based on the methods available on the prototype, so exporting a Proxy as a WorkerEntrypoint wouldn't work)

Are you certain a proxy wouldn't work? 
I think you would only need to define the entry point and a proxy might work 
I remember taking to Rahul about it and he seemed to believe it's possible

penalosa:
@danlapid Fairly sure—I think this codepath would fail: https://github.com/cloudflare/workerd/blob/9e915ed637d65adb3c57522607d2cd8b8d692b6b/src/workerd/io/worker.c%2B%2B#L1920-L1950

I can definitely give it a quick go though

penalosa:
@danlapid Having tried it, I can't figure out how to make a Proxy work—not to say it can't, but I can't figure out how. Additionally, I'm not sure doing this in userland is really the right solution here—it feels like this _should_ be a runtime concern (as it is when deployed). If we do it in userland, then anyone else who comes along and wants to build something similar to Workers Assets needs to copy a fairly convoluted proxy worker, compared to just adding a single config property.

PR #3408, opened 2025-01-24
Node.js compat guidelines... favor explicit throws for intentionally …

jasnell:
…unsupported apis


anonrig (commented):
Can you be explicit about what that error is and what error message we should use?

jasnell:
> Can you be explicit about what that error is and what error message we should use?

In the general guideline here I'd prefer not (this intentionally remains vague about implementation specifics). We can add a utility to the code to make it consistent.

PR #3409, opened 2025-01-24
Don't fail at upload time when module modifies sys.path

hoodmane:
We try to execute all the top level imports but we don't execute changes to the path so we can get crashes.


danlapid:
Why not just execute the top level scope?

hoodmane:
Well the idea here is that we share the package snapshots between workers with the same top level imports. If you exec'd the top level scope, you'd have to use a dedicated snapshot just for that one worker.

danlapid:
> Well the idea here is that we share the package snapshots between workers with the same top level imports. If you exec'd the top level scope, you'd have to use a dedicated snapshot just for that one worker.

I see, so that's why you went with a try and ignore failures approach 👍

hoodmane:
Also, this avoids the need to encrypt snapshots which makes debugging a lot easier.

src/pyodide/internal/snapshot.ts
```diff
@@ -262,8 +262,13 @@ function memorySnapshotDoImports(Module: Module): string[] {
...
   // Import the modules list so they are included in the snapshot.
-  if (deduplicatedModules.length > 0) {
-    simpleRunPython(Module, 'import ' + deduplicatedModules.join(','));
+  for (const mod of deduplicatedModules) {
+    try {
+      simpleRunPython(Module, `import ${mod}`);
+    } catch (e) {
+      continue;
+    }
+    simpleRunPython(Module, `delete ${mod.split(".", 1)[0]}`);
```
danlapid, line 271: Why delete?
And why the split?
I don't understand this we should probably have a comment here if there's a good reason.

hoodmane: Delete because we are importing the module for its side effects into a different context than it is actually imported into at runtime. If `a.py` has `import some_mod` here we are importing it into `__main__`. We just want to execute `some_mod` without actually adding symbols.

The reason for the split is if we want to execute submodules. So for instance if `some_mod` has a lazy submodule and we find `import some_mod.submod`, we need to import this dotted path to instantiate the submodule. But `delete some_mod.submod` will remove `submod` from `some_mod`'s namespace which is broken. We want `delete some_mod` to remove `some_mod` from the `__main__` namespace. I can add these as comments.

hoodmane: Perhaps it would be simpler to do 
```js
simpleRunPython(Module, `exec("import ${mod}", {})`);
```
and then we wouldn't need to delete it...

danlapid: I'm just confused why this delete suddenly appeared with this change.
Doesn't seem like what you're saying is new or specific to the title of this PR.

hoodmane: It should have been there before. See `toDelete` above on line 232.

danlapid (approved):

src/pyodide/internal/snapshot.ts
```diff
@@ -262,8 +262,13 @@ function memorySnapshotDoImports(Module: Module): string[] {
...
 
   // Import the modules list so they are included in the snapshot.
-  if (deduplicatedModules.length > 0) {
-    simpleRunPython(Module, 'import ' + deduplicatedModules.join(','));
+  for (const mod of deduplicatedModules) {
+    try {
+      simpleRunPython(Module, `import ${mod}`);
+    } catch (e) {
+      continue;
+    }
```
dom96, line 270: This feels fragile. Could we `continue` on only a specific kind of exception?

Also can you write a comment to explain exactly what kind of failure mode this solves?

dom96 (commented):
A test for the case that this solves would be good

hoodmane:
Yeah we'll need a validator test for this.

hoodmane:
I guess we should add a test that triggers the validation code path from workerd so we don't have to rely on edgeworker test to exercise this code path.

src/pyodide/internal/snapshot.ts
hoodmane: Fixed.

hoodmane:
Okay I am no longer reproducing any bug so I'll close this.

PR #3412, opened 2025-01-27
update prettier and esbuild

anonrig:
It's never the correct time to update prettier and esbuild I guess. Let's do it.


mikea (approved):

fhanau (approved):

anonrig:
It seems internal-pr is not required but a nice to have. I've removed the flag while opening an internal PR as well.

PR #3413, opened 2025-01-28
Make `PyodideMetadataReader::getWorkerFiles` not fail when it filters files

hoodmane:
ArrayBuilder assumes the array we are constructing has the exact size given. For constructing variable size arrays, we need a `kj::Vector`.

- [x] Add a test


dom96 (approved):

hoodmane:
I'll wait until we merge some version of #3415 so that we can get test coverage.

danlapid (approved):

danlapid:
Why do so many python tests fail with "EQ was unexpected at this time."?

hoodmane:
I dunno it's a pretty small change. It would be strange if it were a real regression...

hoodmane:
Oh it's because #3415 is wrong. Opened #3504 to fix.

PR #3415, opened 2025-01-28
Create and use snapshot by default in all `wd_py_test`s

hoodmane:
This isn't as good as having actual validator coverage but it should help catch a lot of the problems we're having at validation time.

On top of #3476.

### DIFF
--- a/src/workerd/server/tests/python/BUILD.bazel
+++ b/src/workerd/server/tests/python/BUILD.bazel
@@ -20,6 +20,8 @@ py_wd_test("env-param")
 
 py_wd_test(
     "asgi",
+    # TODO: investigate failure!
+    make_snapshot = False,
     skip_python_flags = ["0.27.1"],
 )
 
--- a/build/wd_test.bzl
+++ b/build/wd_test.bzl
@@ -6,6 +6,7 @@ def wd_test(
         name = None,
         args = [],
         ts_deps = [],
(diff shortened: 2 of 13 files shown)


src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -21,6 +21,8 @@ py_wd_test("env-param")
 py_wd_test(
     "asgi",
     skip_python_flags = ["0.27.1"],
+    # TODO: investigate failure!
+    snapshot_variant = False,
```
dom96, line 25: Just to make it a bit clearer:

```suggestion
    snapshot_generating_variant = False,
```

dom96 (approved):

danlapid:
As we spoke over chat.
Let's instead add a variant that generates the snapshot and then uses it to run the test to actually test e2e behaviour.

danlapid (changes requested):
As we spoke over chat.
Let's instead add a variant that generates the snapshot and then uses it to run the test to actually test e2e behaviour.

dom96:
You all probably know this already, but just to make sure we're on the same page: we already have test coverage for this in EW. The fastest way forward would be to just add the vendored worker sample in there.

Of course, creating test coverage for this in workerd is still useful and we should do it (assuming the effort required isn't too massive).

PR #3417, opened 2025-01-28
Roll V8 to 13.3

erikcorry:


WORKSPACE
```diff
@@ -144,6 +144,12 @@ git_repository(
...
 
+git_repository(
+    name = "hwy",
+    commit = "00fe003dac355b979f36157f9407c7c46448958e",
+    remote = "https://chromium.googlesource.com/external/github.com/google/highway.git",
+)
+
 # Bindings for abseil libraries used by V8
 [
     bind(
```
fhanau, line 155: Add bindings like this for the new hwy dependency. That will make them available as `"//external:hwy"` for V8.

fhanau:
Versions of zlib, absl and ICU (in `build/deps/v8.bzl`) should be updated based on `git diff origin/13.2-lkgr origin/13.3-lkgr -- DEPS`

WORKSPACE
erikcorry: It was not clear to me what you meant, but I think I managed to find an incantation that worked by trial and error.

fhanau:
Dropped the `build --@v8//:v8_enable_pointer_compression_shared_cage=False` – it caused the macOS build to fail.

mar-cf (approved):

PR #3418, opened 2025-01-28
add missing node:util methods

anonrig:
Last remaining methods so we can remove the polyfills on node:util and node:sys. 

Ref: https://github.com/cloudflare/workers-sdk/pull/7942

### DIFF
--- a/src/node/util.ts
+++ b/src/node/util.ts
@@ -221,6 +221,28 @@ export function isArray(a: unknown): boolean {
   return Array.isArray(a);
 }
 
+export function parseEnv(): void {
+  // We don't implement this function yet because it is currently under
+  // active development. We should implement it when it's stable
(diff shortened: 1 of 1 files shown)


src/node/util.ts
```diff
@@ -221,6 +221,28 @@ export function isArray(a: unknown): boolean {
...
+export function getSystemErrorName(): void {
+  throw new Error('node:util getSystemErrorName is not implemented');
+}
+
+export function getSystemErrorMessage(): void {
+  throw new Error('node:util getSystemErrorMessage is not implemented');
+}
+
+export function styleText(): void {
+  throw new Error('node:util styleText is not implemented');
```
jasnell, line 243: Just noting... there's really no reason not to implement this one... it's not likely super useful in the common workers case but there's also no other reason not to implement it.

jasnell (approved):

PR #3419, opened 2025-01-28
add remaining async_hooks methods

anonrig:
Adds remaining async_hooks methods so we can remove the unenv polyfill.

### DIFF
--- a/src/node/async_hooks.ts
+++ b/src/node/async_hooks.ts
@@ -4,10 +4,38 @@
 //
 
 import { default as async_hooks } from 'node-internal:async_hooks';
+import { ERR_METHOD_NOT_IMPLEMENTED } from 'node-internal:internal_errors';
 
 export const { AsyncLocalStorage, AsyncResource } = async_hooks;
+
+// We don't add all the async wrap providers since we don't use them
+// and will not expose any APIs that use them.
+export const asyncWrapProviders: Record<string, number> = {
+  NONE: 0,
+};
+
+export function createHook(): void {
+  throw new ERR_METHOD_NOT_IMPLEMENTED('createHook');
+}
+
+export function executionAsyncId(): void {
+  throw new ERR_METHOD_NOT_IMPLEMENTED('executionAsyncId');
+}
+
+export function executionAsyncResource(): void {
+  throw new ERR_METHOD_NOT_IMPLEMENTED('executionAsyncResource');
+}
+
+export function triggerAsyncId(): void {
+  throw new ERR_METHOD_NOT_IMPLEMENTED('triggerAsyncId');
+}
 
 export default {
   AsyncLocalStorage,
   AsyncResource,
+  asyncWrapProviders,
+  createHook,
+  executionAsyncId,
+  executionAsyncResource,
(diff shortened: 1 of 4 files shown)


src/node/async_hooks.ts
```diff
@@ -7,7 +7,105 @@ import { default as async_hooks } from 'node-internal:async_hooks';
 
 export const { AsyncLocalStorage, AsyncResource } = async_hooks;
 
+export const asyncWrapProviders: Record<string, number> = {
```
jasnell, line 10: I don't think we should list any of these, actually. We don't use these and won't expose any APIs that use these and I'd rather not have the burden of keeping a useless list up to date ;-)

jasnell: I'd remove everything except `NONE: 0`

src/node/async_hooks.ts
```diff
@@ -7,7 +7,105 @@ import { default as async_hooks } from 'node-internal:async_hooks';
...
+  RANDOMBYTESREQUEST: 57,
+  RANDOMPRIMEREQUEST: 58,
+  SCRYPTREQUEST: 59,
+  SIGNREQUEST: 60,
+  TLSWRAP: 61,
+  VERIFYREQUEST: 62,
+};
+
+export function createHook(): void {
+  throw new Error('createHook is not implemented');
```
jasnell, line 77: If we're going to go ahead and add stubs like this directly in the runtime, let's make these proper Node.js ERR codes using `ERR_METHOD_NOT_IMPLEMENTED`. Also, it may be worth expanding the error message for these in particular to indicate that they are not only not implemented but there is no intention of implementing these in order to better discourage folks from opening issues about them requesting them

anonrig: I converted them to ERR_METHOD_NOT_IMPLEMENTED but IMHO all of them can be implemented in the future. I don't see any technical blockers, if I'm not missing something.

jasnell (approved):
LGTM but please also include a test that verifies the `ERR_METHOD_NOT_IMPLEMENTED` is reported correctly.

anonrig:
> LGTM but please also include a test that verifies the `ERR_METHOD_NOT_IMPLEMENTED` is reported correctly.

Added extra tests. Thanks for the review @jasnell

src/workerd/api/node/tests/async_hooks-nodejs-test.wd-test
```diff
@@ -0,0 +1,15 @@
+using Workerd = import "/workerd/workerd.capnp";
```
jasnell, line 1: As a todo for later... I wonder if we can simplify adding these tests by having bazel auto-generate the wd-test files for us based on wd_test rule in the bazel. We could easily extend that to include compat flag, compat flags, etc. Would be nice not to have to add multiple files every time we add a single test.

jasnell (approved):

src/workerd/api/node/tests/async_hooks-nodejs-test.wd-test
anonrig: +1

PR #3420, opened 2025-01-28
add remaining node:module methods

anonrig:
I've purposefully didn't add the undocumented methods (even though unenv adds them), and purposefully didn't add experimental or in active development methods.

After this lands, we can remove the unenv polyfill, but I'm not quite sure what we will do with functions that are undocumented but implemented in the polyfill (even though they throw notImplemented() errors)

cc @irvinebroque @jasnell @danlapid

### DIFF
--- a/src/node/module.ts
+++ b/src/node/module.ts
@@ -3,7 +3,10 @@
 //     https://opensource.org/licenses/Apache-2.0
 
 import { default as moduleUtil } from 'node-internal:module';
-import { ERR_INVALID_ARG_VALUE } from 'node-internal:internal_errors';
+import {
+  ERR_INVALID_ARG_VALUE,
+  ERR_METHOD_NOT_IMPLEMENTED,
+} from 'node-internal:internal_errors';
 
 export function createRequire(
   path: string | URL
@@ -21,16 +24,24 @@ export function createRequire(
     );
   }
 
-  return moduleUtil.createRequire(normalizedPath);
+  // TODO(soon): We should move this to C++ land.
+  // Ref: https://nodejs.org/docs/latest/api/modules.html#requireid
+  return Object.assign(moduleUtil.createRequire(normalizedPath), {
+    // We don't throw ERR_METHOD_NOT_IMPLEMENTED because it's too disruptive.
+    resolve(): void {
+      return undefined;
+    },
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    cache: Object.create(null),
+    main: undefined,
+  });
 }
 
 // Indicates only that the given specifier is known to be a
 // Node.js built-in module specifier with or with the the
 // 'node:' prefix. A true return value does not guarantee that
 // the module is actually implemented in the runtime.
-export function isBuiltin(specifier: string): boolean {
-  return moduleUtil.isBuiltin(specifier);
-}
+export const isBuiltin = moduleUtil.isBuiltin.bind(moduleUtil);
 
 // Intentionally does not include modules with mandatory 'node:'
 // prefix like `node:test`.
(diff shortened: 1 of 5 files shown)


vicb (approved):
> After this lands, we can remove the unenv polyfill, but I'm not quite sure what we will do with functions that are undocumented but implemented in the polyfill (even though they throw notImplemented() errors)

I think we should not remove the `unenv` polyfill until we have evidence that all the functions it implements are unused. My understanding is that unenv is tree-shakeable so there is no benefit in dropping the polyfill if functions are unused. However dropping functions that are actually used will break users.

Maybe I miss something?

src/node/module.ts
```diff
@@ -21,16 +24,27 @@ export function createRequire(
...
+  return Object.assign(moduleUtil.createRequire(normalizedPath), {
+    resolve(): void {
+      throw new ERR_METHOD_NOT_IMPLEMENTED('require.resolve()');
+    },
+    get cache(): void {
+      throw new ERR_METHOD_NOT_IMPLEMENTED('require.cache');
+    },
+    get main(): void {
+      throw new ERR_METHOD_NOT_IMPLEMENTED('require.main');
+    },
```
jasnell, line 39: For better readability, it would be helpful to move these function bodies out and simply refer to them here.

src/node/module.ts
```diff
@@ -110,8 +124,40 @@ export const builtinModules = [
...
+export function syncBuiltinESMExports(): void {
+  // Not meaningful to implement in the context of workerd.
+  throw new ERR_METHOD_NOT_IMPLEMENTED('module.syncBuiltinESMExports');
+}
+
+//
+// IMPORTANT NOTE!
+//
+// We are deliberately not including any of these functions below because
+// they are either experimental or in active development as of January 2025.
```
jasnell, line 146: I would say "either experimental, in active development, or not relevant..."

For instance, I do not imagine that would would ever implement `registerHooks()`, `module.constants.compileCacheStatus`, `flushCompileCache`, or `getCompileCacheDir` in workers.

jasnell (approved):
LGTM but please add a test to ensure the expected `ERR_METHOD_NOT_IMPLEMENTED` errors are properly thrown.

src/node/module.ts
```diff
@@ -21,16 +24,24 @@ export function createRequire(
...
 
-  return moduleUtil.createRequire(normalizedPath);
+  // TODO(soon): We should move this to C++ land.
+  // Ref: https://nodejs.org/docs/latest/api/modules.html#requireid
+  return Object.assign(moduleUtil.createRequire(normalizedPath), {
+    // We don't throw ERR_METHOD_NOT_IMPLEMENTED because it's too disruptive.
+    resolve(): void {
+      return undefined;
+    },
+    extensions: undefined,
```
vicb, line 34: That does not seem to adheres to NodeJS types?

Also can a test be added?

jasnell: I personally think it's fine if the types don't match on apis we intentionally do not support and that have been deprecated in node.js. If we want to be strict about it, however, these can return an empty object instead. e.g. `return {};`

anonrig: I'll just remove extensions since it's deprecated and we're have a policy of not implementing deprecated/experimental features.

vicb: what about resolve?

anonrig: unenv polyfill throws an error afaict. i'm not sure what we consider a "breaking change" at this point. @jasnell

```
resolve: Object.assign(notImplemented("module.require.resolve"), {
      paths: notImplemented("module.require.resolve.paths"),
    }),
    ```

Comment (jasnell) [reply src/node/module.ts]: If `resolve` in the polyfill exists as a function but throws an error when called, that's what we should do also.

Generally speaking, for all of these, we have to generally align with whatever the polyfill is currently doing. We can tweak from there using compat flags/dates later, but if the goal is to replace the polyfill with the built-in non-op stub, then the behavior of the built-in non-op stub should match the polyfill or we risk breaking users.

Comment (vicb) [reply src/node/module.ts]: Sorry I might be wrong, not really have time tonight.

I think it's a breaking change because currently workerd returns a string for `createRequire(root).resolve()` - IIRC @jasnell implemented that.

Also @anonrig I think you are looking at the wrong file, [unenv code](https://github.com/unjs/unenv/blob/main/src/runtime/node/module/%24cloudflare.ts#L64-L72) is

```ts
export const createRequire: typeof nodeModule.createRequire = (
  file: string | URL,
) => {
  return Object.assign(workerdModule.createRequire(file), {
    resolve: Object.assign(notImplemented("module.require.resolve"), {
      paths: notImplemented("module.require.resolve.paths"),
    }),
    cache: Object.create(null),
    extensions: _extensions,
    main: undefined,
  });
};
```

Hope this helps

jasnell: Hmmm, good point about implementnig `createRequire(...).resolve(...)` ... we may be able to implement resolve here reasonably by having it defer to `createRequire(...).resolve(...)` so we don't have to leave this stubbed out and things stay more consistent.

vicb: > Wait... actually _did I_ implement `createRequire(...).resolve(...)`

Yes in https://github.com/cloudflare/workerd/pull/2636 :)

PR #3422, opened 2025-01-28
[build] Update Bazel to 7.5.0, update development docs

fhanau:
The Bazel 8 update is currently blocked on a TypeScript build issue, so let's update to the latest maintenance release instead. This includes a few remote cache bug fixes, so we enable Build Without The Bytes for Windows again (disabling effectively only applied to release CI runs, for tests this gets overwritten by the ci-test configuration).
Also update the development documentation since we're here – the section on code formatting was outdated and suggested setup steps that are no longer needed.  (Thank you to @jmorrell for suggesting this).


.bazelversion
```diff
@@ -1 +1 @@
-7.4.1rc2
+7.5.0rc3
```
fhanau, line 1: Changes: https://github.com/bazelbuild/bazel/compare/7.4.0...7.5.0rc3

- [ ] Wait until 7.5.0 stable is released before merging this.

danlapid (approved):

Issue #3425, opened 2025-01-29
`node:net` generates a `TypeError: You must call read() on a "byob" reader with a positive-sized TypedArray object`


vicb:
I was investigating an [issue](https://github.com/cloudflare/workerd/discussions/2721#discussioncomment-11992887) - I was not able to reproduce.

But I notice this log:

```bash
bazel test //src/workerd/api/node:tests/net-nodejs-test --test_output=all

workerd/server/server.c++:4336: debug: [ TEST ] net-nodejs-test:testNetOnReadStaticBuffer
workerd/io/worker.c++:1048: info: uncaught exception; desc = TypeError: You must call read() on a "byob" reader with a positive-sized TypedArray object.
    at startRead (node:net:811:50)
```

Not sure if that's expected?

PR #3427, opened 2025-01-29
stop reading empty buffers on node:net

anonrig:
Fixes https://github.com/cloudflare/workerd/issues/3425

### DIFF
--- a/src/node/net.ts
+++ b/src/node/net.ts
@@ -139,7 +139,7 @@ type SocketClass = SocketType & {
     | null
     | undefined
     | ((len?: number, buf?: Buffer) => boolean | Uint8Array);
-  [kBufferGen]: null | (() => null | boolean | Uint8Array);
+  [kBufferGen]: null | (() => undefined | Uint8Array);
   [kSocketInfo]: null | {
     address?: string;
     port?: number;
@@ -311,7 +311,7 @@ export const Socket = function Socket(
       this[kBufferGen] = onread.buffer;
     } else {
       this[kBuffer] = onread.buffer;
-      this[kBufferGen] = (): Uint8Array | boolean | null => this[kBuffer];
+      this[kBufferGen] = (): Uint8Array | undefined => onread.buffer;
     }
     this[kBufferCb] = onread.callback;
   } else {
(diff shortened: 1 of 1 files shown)


src/node/net.ts
```diff
@@ -1110,10 +1110,21 @@ async function startRead(socket: SocketClass): Promise<void> {
   const reader = socket._handle.reader;
   try {
     while (socket._handle.reading === true) {
+      const generatedBuffer = socket[kBufferGen]?.();
+
+      if (
+        generatedBuffer instanceof Uint8Array &&
+        generatedBuffer.length === 0
```
vicb, line 1117: I don't know this code well enough but would something like that work?

```suggestion
        generatedBuffer?.length === 0
```

That is, I can not tell if `generatedBuffer instanceof Uint8Array` is ok or too restrictive here?

src/node/net.ts
```diff
@@ -1110,10 +1110,21 @@ async function startRead(socket: SocketClass): Promise<void> {
   const reader = socket._handle.reader;
   try {
     while (socket._handle.reading === true) {
+      const generatedBuffer = socket[kBufferGen]?.();
+
+      if (
```
jasnell, line 1115: Likely worthwhile mentioning the conditions under which kBufferGen will return a zero-length buffer.

jasnell (approved):

src/node/net.ts
jasnell: hmm... actually, if `generatedBuffer` can be `null` or `boolean` then this is still buggy as we can't pass `null` or `boolean` as the argument to `read()`

anonrig: Type-wise it can be null or boolean, but it will never be called with null or boolean. If it did Node.js implementation would not work. I'll make sure to handle null or boolean in here as well.

src/node/net.ts
```diff
@@ -1110,10 +1110,23 @@ async function startRead(socket: SocketClass): Promise<void> {
...
     while (socket._handle.reading === true) {
+      const generatedBuffer = socket[kBufferGen]?.();
+
+      // Let's be extra cautious here and handle nullish values.
+      if (generatedBuffer == null || generatedBuffer.length === 0) {
+        // When reading a static buffer with fixed length, it's highly likely to
+        // read the whole buffer in a single take, which will make the second
+        // operation to read an empty buffer.
+        //
+        // Workerd throws the following exception when reading empty buffers
```
jasnell, line 1121: fwiw, throwing here is mandated by the spec. It's not something specific to our implementation as this comment may suggest.

jasnell (approved):
Is there any reasonable way to add a test for this?

anonrig:
> Is there any reasonable way to add a test for this?

There is a test for it already in node:net which was printing an error on the log, but it's not at the moment. I think it's unnecessary since the following function has a similar check in 10 lines below.

src/node/net.ts
anonrig: Agreed. AFAICT, node.js uses nodejs streams which doesn't require throwing an error, whereas workerd uses web streams in here.

PR #3429, opened 2025-01-29
Fix Windows build issues relating to wpt test harness

npaun:


src/workerd/api/wpt/tsconfig.json
```diff
@@ -26,6 +26,6 @@
       "wpt:*": ["../../../wpt/*"]
     }
   },
-  "include": ["**/*.ts"],
+  "include": ["**/*.ts", "../../../wpt/*.ts"],
```
npaun, line 29: I'm not yet able to figure out this one locally, but the error @anonrig had seems to indicate that this is needed.

anonrig (approved):

src/workerd/api/wpt/tsconfig.json
npaun: Yep, it works.

PR #3430, opened 2025-01-29
update nodejs and add compile cache

anonrig:
Passing NODE_COMPILE_CACHE environment variable enables compile cache on node.js

### DIFF
--- a/types/BUILD.bazel
+++ b/types/BUILD.bazel
@@ -28,6 +28,9 @@ js_binary(
         ":types_lib",
     ],
     entry_point = "scripts/build-types.js",
+    env = {
+        "NODE_COMPILE_CACHE": "/tmp/workerd-types-compile-cache",
+    },
 )
 
 js_run_binary(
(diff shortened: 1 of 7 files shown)


types/BUILD.bazel
```diff
@@ -28,6 +28,9 @@ js_binary(
         ":types_lib",
     ],
     entry_point = "scripts/build-types.js",
+    env = {
+        "NODE_COMPILE_CACHE": "/tmp/workerd-types-compile-cache",
```
anonrig, line 32: @fhanau @mikea any suggestions on what to pass here?

fhanau: I mean... do we need this? This action is only used to generate types right?

jasnell: Not strictly needed, no, but results in a significant performance improvement when used correctly.

anonrig: It can be any temp directory, but something that can be preserved in remote cache as well would be really good.

mikea: you'd need
- add data dependency to the target
- use `$(location)` to set env var

PR #3434, opened 2025-01-29
Further clarify some details around polyfills in node.js compat guide

jasnell:


danlapid (approved):
Feels like we should just delete all polyfilling references from this doc at this point.

But as it stands this PR is a clear improvement on what is currently there.

Goddessnunu22:
> 

Yes

PR #3436, opened 2025-01-29
WPT: Add dom/abortsignal tests

npaun:
Enables tests from the dom/abortsignal module from WPT.

The following features were added to the harness to support these tests:
- The `Test` class, including step, step_func, step_timeout, step_func_done and done
- The `async_test` test type


src/workerd/api/wpt/dom/abort-test.ts
```diff
@@ -0,0 +1,26 @@
...
+      '(1, 2) Target should be set to signal. (3) Should be investigated.',
+    expectedFailures: [
+      'AbortSignal.any() follows a single signal (using AbortController)',
+      'AbortSignal.any() follows multiple signals (using AbortController)',
+      'Abort events for AbortSignal.any() signals fire in the right order (using AbortController)',
+    ],
+  },
+  'event.any.js': {
+    comment: 'Target should be set to signal',
+    expectedFailures: ['the abort event should have the right properties'],
```
npaun, line 20: The event has `target = undefined` and `currentTarget = signal`. I'm not sure if I'm doing something wrong in the harness or if we have a bug.

src/workerd/api/wpt/dom/abort-test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default {
+  'abort-signal-any.any.js': {
+    includeFile: 'abort-signal-any-tests.js',
+    comment:
+      '(1, 2) Target should be set to signal. (3) Should be investigated.',
+    expectedFailures: [
+      'AbortSignal.any() follows a single signal (using AbortController)',
+      'AbortSignal.any() follows multiple signals (using AbortController)',
+      'Abort events for AbortSignal.any() signals fire in the right order (using AbortController)',
```
npaun, line 15: Quick link to the relevant bit for the "signals fire in the right order" test:

https://github.com/web-platform-tests/wpt/blob/d46d9abdea3ea1eb0f1eebdc4430f21a13afaae3/dom/abort/resources/abort-signal-any-tests.js#L164

npaun:
@anonrig 

Bad news: I have a Windows build failure too now.
Good news: It'll be convenient for me to fix it now that it's on my own branch lol.

jasnell:
Great work, will review tomorrow.

src/workerd/api/wpt/url-test.ts
```diff
@@ -36,7 +35,7 @@ export default {
     ],
   },
   'percent-encoding.window.js': {
-    comment: 'Implement `async_test`',
+    comment: 'Implement document',
```
anonrig, line 38: We don't need to implement document but simply need the functionality to mutate the code before running them:

Example: https://github.com/nodejs/node/blob/main/test/wpt/test-url.js#L9

src/workerd/api/wpt/dom/abort-test.ts
```diff
@@ -0,0 +1,26 @@
...
+    ],
+  },
+  'event.any.js': {
+    comment: 'Target should be set to signal',
+    expectedFailures: ['the abort event should have the right properties'],
+  },
+  'timeout-shadowrealm.any.js': {
+    comment: "We don't support shadowrealm",
```
anonrig, line 23:
```suggestion
    comment: "Enable when ShadowRealm is implemented",
```

src/wpt/harness.ts
```diff
@@ -66,8 +68,208 @@ type TestCase = {
...
+/**
+ * @class
+ *
+ * A single subtest. A Test is not constructed directly but via the
+ * :js:func:`test`, :js:func:`async_test` or :js:func:`promise_test` functions.
+ *
+ * @param {string} name - This must be unique in a given file and must be
+ * invariant between runs.
+ *
+ */
```
anonrig, line 90: We can remove the unnecessary parts from jsdoc, since we are using typescript. (such as line 82)

src/wpt/harness.ts
```diff
@@ -66,8 +68,208 @@ type TestCase = {
   test(_: unknown, env: Env): Promise<void>;
 };
 
+type UnknownFunc = (...args: unknown[]) => unknown;
+
+enum TestPhase {
```
anonrig, line 73: enum is not recommended to be used since it has no direct translation to javascript. Can you convert this to an object?

src/wpt/harness.ts
```diff
@@ -66,8 +68,208 @@ type TestCase = {
...
+   *
+   * @example
+   * let t = async_test("Example");
+   * onload = t.step_func(e => {
+   *   assert_equals(e.name, "load");
+   *   // Mark the test as complete.
+   *   t.done();
+   * })
+   *
+   * @param {string} func - Function to run as a step. If this
```
anonrig, line 164: is func a string or UnknownFunc? Can we remove the jsdoc params

src/wpt/harness.ts
```diff
@@ -66,8 +68,208 @@ type TestCase = {
...
+   * throws an :js:func:`AssertionError`, or any other exception,
+   * the :js:class:`Test` status is set to ``FAIL``.
+   * @param {Object} [this_obj] - The object to use as the this
+   * value when calling ``func``. Defaults to the :js:class:`Test` object.
+   */
+  public step_func(func: UnknownFunc, this_obj?: object): UnknownFunc {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias -- We could get rid of this this alias but the code is copied from WPT
```
anonrig, line 171: Can you add this to the beginning of this file, so we don't need to disable everywhere

src/wpt/harness.ts
```diff
@@ -66,8 +68,208 @@ type TestCase = {
...
+
+  /**
+   * Return a function that automatically sets the current test to
+   * ``FAIL`` if it's called.
+   *
+   * @param {string} [description] - Error message to add to assert
+   * in case of failure.
+   *
+   */
+  public unreached_func(description: string): UnknownFunc {
```
anonrig, line 221: this should be optional. in jsdoc, surrounding the name of the variable with square brackets make it optional. ex: [description]

Ref: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

```js
public unreached_func(description?: string): UnknownFunc {
```

src/wpt/harness.ts
```diff
@@ -219,15 +427,72 @@ globalThis.subsetTestByKey = (
...
+  if (promise instanceof Promise) {
+    globalThis.promises.push(
+      promise
+        .catch((err: unknown) => {
+          globalThis.errors.push(new AggregateError([err], name));
+        })
+        .then(() => {
+          if (testCase.error) {
+            globalThis.errors.push(testCase.error);
+          }
```
anonrig, line 451: line 438 and 449 does the same exact thing. doesn't this duplicate the errors?

src/wpt/harness.ts
```diff
@@ -219,15 +427,72 @@ globalThis.subsetTestByKey = (
...
-  try {
-    globalThis.promises[name] = func.call(this);
-  } catch (err) {
-    globalThis.errors.push(new AggregateError([err], name));
+  const testCase = new Test(name, properties);
+  const promise = testCase.step(func, testCase, testCase);
+
+  if (testCase.error) {
+    globalThis.errors.push(testCase.error);
+  }
```
anonrig, line 440: How can a test you just created can have an error property?

src/wpt/harness.ts
```diff
@@ -439,21 +705,14 @@ function prepare(env: Env, options: TestRunnerOptions): void {
...
 ): Promise<void> {
-  for (const [name, promise] of Object.entries(globalThis.promises)) {
-    try {
-      await promise;
-    } catch (err) {
-      globalThis.errors.push(new AggregateError([err], name));
-    }
-  }
-
+  await Promise.all(globalThis.promises);
```
anonrig, line 715: If this will never throw (since you're catching them while creating), can you add a comment here?

src/wpt/harness.ts
```diff
@@ -480,6 +739,13 @@ async function validate(
   }
 }
 
+function invokeFile(env: Env, file: string): void {
+  if (typeof env[file] !== 'string') {
+    throw new Error(`Unable to run ${file}. Code is not a string`);
```
anonrig, line 744: Is this possible?

src/workerd/api/wpt/dom/abort-test.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import { type TestRunnerConfig } from 'wpt:harness';
+
+export default {
+  'abort-signal-any.any.js': {
```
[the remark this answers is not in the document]
npaun: If they're not listed it means it's ok to run all the tests in the file. I can list them with a `{}` empty config if it's clearer though.

src/wpt/harness.ts
npaun: Not really but it makes TS happy.

src/workerd/api/wpt/dom/abort-test.ts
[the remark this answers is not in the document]
anonrig: How can we validate that it runs all of them? The problem that I see with this approach is, if we introduce a bug it will be extremely hard to realize it. Can we enforce to list all of these files, or should we write a test that validates that it will run such files? (so we don't regress)

build/BUILD.wpt
```diff
@@ -2,19 +2,21 @@
...
-    exclude_directories = 0,
-)
+)]
 
 [filegroup(
     name = dir,
     srcs = glob(["{}/**/*".format(dir)]),
     visibility = ["//visibility:public"],
-) for dir in directories]
+) for dir in wpt_get_directories("dom")]
```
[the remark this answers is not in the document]
npaun: I want to have a hierarchy under dom because having one huge `dom-test.ts` seems hard to work with. So this creates a `dom/` directory so we can have `dom/abort-test.ts`. dom is the only module that I'm breaking out so far. But we'll see if we need a different hierarchy as we enable more modules.

src/workerd/api/wpt/dom/abort-test.ts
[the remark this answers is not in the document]
npaun: I could make it a runtime error if we discover files not listed in the config.

src/wpt/harness.ts
anonrig: You can `String(env[file])` as well

src/wpt/harness.ts
npaun: The function passed to `promise_test` needs to return a `Promise` but is not required to be an `async function`, so if some exception is thrown before it returns the Promise that exception would be caught immediately in `step`.

src/wpt/harness.ts
npaun: Yeah this is a bit messy. I'll try to rework this.

src/wpt/harness.ts
anonrig: Can you add this as a comment?

build/BUILD.wpt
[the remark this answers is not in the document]
anonrig: I mean previously this was checking all directories. Right now it's checking only for dom?

src/workerd/api/wpt/dom/abort-test.ts
npaun: @jasnell I'm marking this as resolved for now to prepare for PR merge, but here were the failing cases we talked about last time.

src/workerd/api/wpt/dom/abort-test.ts
npaun: @jasnell I'm marking this as resolved for now to prepare for PR merge, but here were the failing cases we talked about last time.

build/BUILD.wpt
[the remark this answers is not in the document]
npaun: Ah, I think the parameter names were just unclear. I think I fixed it.

src/workerd/api/wpt/dom/abort-test.ts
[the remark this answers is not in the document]
anonrig: Did we make it a runtime error? I couldn't find it.

anonrig (approved):

src/wpt/harness.ts
```diff
@@ -480,22 +739,34 @@ async function validate(
...
+): (file: string) => TestCase {
+  return (file: string): TestCase => {
+    return {
+      async test(_: unknown, env: Env): Promise<void> {
+        const options = config[file];
+        if (!options) {
+          throw new Error(
+            `Missing test configuration for ${file}. Specify '${file}': {} for default options.`
+          );
+        }
```
npaun, line 753: @anonrig Here's the bit that requires us to list every test.

src/workerd/api/wpt/dom/abort-test.ts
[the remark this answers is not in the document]
npaun: https://github.com/cloudflare/workerd/pull/3436#discussion_r1940246914

src/wpt/harness.ts
anonrig: Thanks!

fhanau:
This breaks the downstream build – please fix or revert

PR #3438, opened 2025-01-30
Refactor: Make ValidationErrorReporter interface better.

kentonv:
This interface evolved from early beginnings where we didn't really know where we were going, and as a result was quite warty in the way that it reported entrypoints:
* It would report each _method_ of stateless entrypoints with a separate call, rather than reporting the whole entrypoint at once. All of the implementations had to group these back into entrypoints using HashMaps.
* It reported a Durable Object class as if it were an entrypoint with a single method called "class". This sort of made sense back when there was a fixed list of handler names, but the introduction of RPC made this very weird.

Evidence of why this was a bad interface can be seen in the first commit of this PR, which introduces tests for three different bugs that I discovered just while refactoring this interface.

In any case, this changes the interface to just make sense.

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -2144,20 +2144,22 @@ void Worker::Lock::validateHandlers(ValidationErrorReporter& errorReporter) {
     ignoredHandlers.insert("rejectionhandled"_kj);
 
     KJ_IF_SOME(c, worker.impl->context) {
+      // Service workers syntax.
       auto handlerNames = c->getHandlerNames();
-      bool foundAny = false;
+      kj::Vector<kj::String> handlers;
       for (auto& name: handlerNames) {
         if (!ignoredHandlers.contains(name)) {
-          errorReporter.addHandler(kj::none, name);
-          foundAny = true;
+          handlers.add(kj::str(name));
         }
       }
-      if (!foundAny) {
+      if (handlers.size() == 0) {
         errorReporter.addError(
             kj::str("No event handlers were registered. This script does nothing."));
       }
+      errorReporter.addEntrypoint(kj::none, handlers.releaseAsArray());
     } else {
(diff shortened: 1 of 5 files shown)


src/workerd/io/worker.c++
```diff
@@ -2196,7 +2192,19 @@ void Worker::Lock::validateHandlers(ValidationErrorReporter& errorReporter) {
...
+        KJ_IF_SOME(entrypointName, getEntrypointName(entry.key)) {
+          errorReporter.addActorClass(entrypointName);
+        } else {
+          // Hmm, it appears someone tried to export a Durable Object class as a default
+          // entrypoint. This doesn't actually work: the runtime will not allow this DO class
+          // to be used, either for actors or as an entrypoint.
+          //
+          // TODO(someday): Make this a hard error. I'm hesitant to do it in my current change
+          //   for fear that it'll break someone somewhere forcing a rollback. For now we log.
+          LOG_PERIODICALLY(ERROR,
```
jasnell, line 2204: Nit: log periodically is nice but still might be better to make this skip sentry. It is easy enough to query kibana directly for it.

jasnell (approved):

jasnell:
Looks like you're missing a `virtual` keyword in there somewhere. Signed off because the overall change looks good once the compile bugs are addressed.

src/workerd/io/worker.c++
kentonv: I think it'll be fine as I don't think it's even possible to deploy a worker with this problem, but if it's possible, it doesn't accomplish anything.

Don't like using NOSENTRY as there's a big risk of just not ever noticing it at all.

jasnell: heh, I think I'm probably the only one who regularly searches kibana for the NOSENTRY logs we do happen to add ;-)

PR #3439, opened 2025-01-30
[o11y] Fix segfault with tail worker self loop

fhanau:
To detect the main worker having a self loop, we need to compare the current WorkerService with the WorkerService corresponding to the tail worker. If service refers to an EntrypointService, that means we have to compare to the underlying worker instead.

As reported by @penalosa. Tracked this down together with @jmorrell.

### DIFF
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -212,6 +212,12 @@ class Server::Service {
 
   // Returns true if the service exports the given handler, e.g. `fetch`, `scheduled`, etc.
   virtual bool hasHandler(kj::StringPtr handlerName) = 0;
+
+  // Return the service itself, or the underlying service if this instance wraps another service as
+  // with EntrypointService.
+  virtual Service* service() {
+    return this;
+  }
 };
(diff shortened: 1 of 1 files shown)


danlapid:
Can we add a test?

jasnell (approved):
Nice catch

fhanau:
> Can we add a test?

Testing streaming tail worker is not really amenable to the testing framework we have in workerd (i.e. `wd_test`) since we don't define a number of tests in JS, but rather need to look at if communication between two worker stages works properly. This is easier to test within the downstream test environment.

danlapid:
> > Can we add a test?
> 
> 
> 
> Testing streaming tail worker is not really amenable to the testing framework we have in workerd (i.e. `wd_test`) since we don't define a number of tests in JS, but rather need to look at if communication between two worker stages works properly. This is easier to test within the downstream test environment.

Sadly the downstream test environment is unable to test the specific modified file.

Perhaps we need to add some testing infrastructure to workerd.

PR #3442, opened 2025-01-30
Fixes "promise will never complete" when exceeding memory.

dom96:
When we grow WebAssembly memory in a specific way, we can get into a situation where the isolate gets terminated and the termination waiter signal doesn't reach the isolate. This resulted in a promise being rejected with a "Promise will never complete" exception. This PR fixes this so that a more descriptive exception is used.

### DIFF
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -16,6 +16,7 @@
 #include <workerd/io/trace.h>
 #include <workerd/jsg/async-context.h>
 #include <workerd/jsg/jsg.h>
+#include <workerd/util/exception.h>
 #include <workerd/util/uncaught-exception-source.h>
 #include <workerd/util/weak-refs.h>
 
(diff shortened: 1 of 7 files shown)


src/workerd/io/io-context.h
```diff
@@ -1257,23 +1257,40 @@ kj::_::ReducePromises<RemoveIoOwn<T>> IoContext::awaitJs(jsg::Lock& js, jsg::Pro
...
+        : fulfiller(kj::mv(fulfiller)),
+          isolate(kj::mv(isolate)) {}
 
     ~RefcountedFulfiller() noexcept(false) {
       if (!isDone) {
-        // The JavaScript resolver was garbage collected, i.e. JavaScript will never resolve
-        // this promise.
-        fulfiller->reject(JSG_KJ_EXCEPTION(FAILED, Error, "Promise will never complete."));
+        auto hasExcessivelyExceededHeapLimit =
+            isolate->getLimitEnforcer().hasExcessivelyExceededHeapLimit();
```
jasnell, line 1271: hmm.. this *should* be ok I think? But does make me somewhat nervous. The references to the `RefcountedFulfiller` *should* be dropped before the isolate is destroyed but we should just double check and confirm that having the RefcountedFulfiller hold a strong reference to the isolate won't cause any problems. I'm probably being needlessly paranoid here tho.

src/workerd/io/io-context.h
```diff
@@ -1257,23 +1257,40 @@ kj::_::ReducePromises<RemoveIoOwn<T>> IoContext::awaitJs(jsg::Lock& js, jsg::Pro
...
+        if (hasExcessivelyExceededHeapLimit) {
+          fulfiller->reject(
+              JSG_KJ_EXCEPTION(OVERLOADED, Error, "jsg.Error: Worker has exceeded memory limit."));
```
jasnell, line 1274: Not something to solve for now but this does make me wonder if there are other possible similar cases where we know why the promise will never complete... something to look at later.

src/workerd/io/io-context.h
```diff
@@ -1257,23 +1257,40 @@ kj::_::ReducePromises<RemoveIoOwn<T>> IoContext::awaitJs(jsg::Lock& js, jsg::Pro
...
+        }
       }
     }
 
    private:
     kj::Maybe<kj::StringPtr> finalize() override {
       if (!isDone) {
-        fulfiller->reject(JSG_KJ_EXCEPTION(FAILED, Error, "Promise will never complete."));
+        auto hasExcessivelyExceededHeapLimit =
+            isolate->getLimitEnforcer().hasExcessivelyExceededHeapLimit();
```
jasnell, line 1287: Nit: the duplicated logic here has always bothered me. Any chance we could combine these into a single utility method and eliminate/reduce the duplication?

src/workerd/io/io-context.h
dom96: Yeah, I've been meaning to look into this too. Is there any other way I could access the IsolateLimitEnforcer here?

danlapid: If the isolate was already dropped returning a jsg.Error has no meaning.
Why not hold a WeakRef and only run the hasExcessivelyExceededHeapLimit specialization if we're able to get a stong ref?

danlapid: isolate.getWeakRef() is an easy way to get a weak ref btw.

dom96: Nice idea, done!

src/workerd/io/io-context.h
dom96: Hmm, it doesn't seem simple to find similar cases, but maybe I'm missing something.

src/workerd/io/io-context.h
```diff
@@ -1257,23 +1257,38 @@ kj::_::ReducePromises<RemoveIoOwn<T>> IoContext::awaitJs(jsg::Lock& js, jsg::Pro
...
 
     ~RefcountedFulfiller() noexcept(false) {
       if (!isDone) {
+        reject();
+      }
+    }
+
+   private:
+    void reject() {
+      auto hasExcessivelyExceededHeapLimit = maybeIsolate->tryAddStrongRef()
```
danlapid, line 1276: I'd love a comment here explaining why the usage of a WeakRef makes sense in this situation and why returning a false default is fine.
I know I suggested it, I know it makes sense, I just think a comment would be nice for future generations 😄

danlapid (approved):
LGTM, but I would love to get that comment in 😄

src/workerd/io/io-context.h
jasnell: Yeah we'd have to investigate to see. As I said, not something to solve for now but something we may want to explore as a later step

jasnell (approved):

anonrig (approved):

src/workerd/io/io-context.h
dom96: Done!

PR #3446, opened 2025-01-31
Prevent exception when passing directory instead of file

jmorrell-cloudflare:
If you accidentally pass a directory path instead of a capnp config file to `workerd serve $path` you get an uncaught exception:

```
*** Uncaught exception ***
kj/filesystem-disk-unix.c++:365: failed: mmap: Invalid argument
```

This change adds an assertion that the provided path is a file type. Invoking `workerd serve` with a directory now results in the following error:

```
> workerd serve samples/tail-workers/
workerd serve: samples/tail-workers/: Config path is not a file.
workerd serve --help' for more information.
```

### DIFF
--- a/src/workerd/server/workerd.c++
+++ b/src/workerd/server/workerd.c++
@@ -1094,15 +1094,13 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
     } else {
       // Read file from disk.
       auto path = fs->getCurrentPath().evalNative(pathStr);
-
-      // make sure the path represents a file
-      KJ_IF_SOME(fsMetadata, fs->getRoot().tryLstat(path)) {
-        if (fsMetadata.type != kj::FsNode::Type::FILE) {
-          CLI_ERROR("Path is not a file.");
-        }
-      }
-
       auto file = KJ_UNWRAP_OR(fs->getRoot().tryOpenFile(path), CLI_ERROR("No such file."));
+
+      // Use stat() to check that we have a file vs a directory which will fail to mmap
+      auto metadata = file->stat();
+      if (metadata.type != kj::FsNode::Type::FILE) {
+        CLI_ERROR("Config path is not a file.");
+      }
 
       if (binaryConfig) {
         // Interpret as binary config.
--- a/src/workerd/server/workerd.c++
+++ b/src/workerd/server/workerd.c++
@@ -1094,6 +1094,14 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
     } else {
       // Read file from disk.
       auto path = fs->getCurrentPath().evalNative(pathStr);
+
+      // make sure the path represents a file
+      KJ_IF_SOME(fsMetadata, fs->getRoot().tryLstat(path)) {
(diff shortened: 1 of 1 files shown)


jmorrell-cloudflare:

src/workerd/server/workerd.c++
```diff
@@ -1094,6 +1094,14 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
...
       auto path = fs->getCurrentPath().evalNative(pathStr);
+
+      // make sure the path represents a file
+      KJ_IF_SOME(fsMetadata, fs->getRoot().tryLstat(path)) {
+        if (fsMetadata.type != kj::FsNode::Type::FILE) {
+          CLI_ERROR("Path is not a file.");
+        }
+      }
+
       auto file = KJ_UNWRAP_OR(fs->getRoot().tryOpenFile(path), CLI_ERROR("No such file."));
```
jmorrell-cloudflare, line 1105: Alternatively, perhaps `tryOpenFile` should check for this? https://github.com/capnproto/capnproto/blob/v2/c%2B%2B/src/kj/filesystem.c%2B%2B#L1079

kentonv (approved):

kentonv:
Hmm, the code looks right to me but CI suggests it is not, somehow.

danlapid:
Not sure I understand what's going on here, perhaps a rebase will solve it? 🤔

src/workerd/server/workerd.c++
```diff
@@ -1094,6 +1094,14 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
     } else {
       // Read file from disk.
       auto path = fs->getCurrentPath().evalNative(pathStr);
+
+      // make sure the path represents a file
+      KJ_IF_SOME(fsMetadata, fs->getRoot().tryLstat(path)) {
+        if (fsMetadata.type != kj::FsNode::Type::FILE) {
```
kentonv, line 1100: Oh, I think this doesn't work if the path refers to a symbolic link, because it intentionally doesn't follow the link. You could allow `Type::SYMLINK` here to fix it -- except then if it's a symlink to a directory, it'll still have the old problem.

I guess that means there isn't actually a way to probe whether a symlink points at a file or a directory via the KJ filesystem API as it stands, except by catching exceptions.

jmorrell-cloudflare: 🤔 Should I modify this to catch the exception directly where its thrown then?

kentonv: Oh wait... based on the error message in the PR description, I guess the error only happens when you try to mmap the directory, not when you try to open it as a file. That makes sense: it's perfectly valid to open a directory as a file, at least on Unix.

What you could do, then, is open it, and then call `file->stat()` on the opened file to check if it is actually a file vs. a directory.

jmorrell-cloudflare: Updated to check the opened file

kentonv (approved):

danlapid (approved):

PR #3447, opened 2025-02-01
Implement automatic self-referential bindings via ctx.exports.

kentonv:
For every top-level export, `ctx.exports` will contain a corresponding property which acts as a binding to that export. If the export is a simple object or `WorkerEntrypoint`, then the `ctx.exports` property is a service binding. If the export is a Durable Object class (and it has been configured with storage), the `ctx.exports` property is a DO namespace binding.

So you can do: `ctx.exports.MyEntrypoint.someRpc()`

And it calls `someRpc()` on the entrypoint named `MyEntrypoint`, as if you had called it over a service binding.

This is marked experimental for now since it won't work on the edge until further changes are made to our control plane.

### DIFF
--- a/src/workerd/api/actor-state.c++
+++ b/src/workerd/api/actor-state.c++
@@ -826,9 +826,11 @@ kj::OneOf<jsg::Ref<DurableObjectId>, kj::StringPtr> ActorState::getId() {
 }
 
 DurableObjectState::DurableObjectState(Worker::Actor::Id actorId,
+    jsg::JsRef<jsg::JsValue> exports,
     kj::Maybe<jsg::Ref<DurableObjectStorage>> storage,
     kj::Maybe<rpc::Container::Client> container)
     : id(kj::mv(actorId)),
+      exports(kj::mv(exports)),
       storage(kj::mv(storage)),
       container(container.map(
           [&](rpc::Container::Client& cap) { return jsg::alloc<Container>(kj::mv(cap)); })) {}
--- a/src/workerd/api/actor-state.h
+++ b/src/workerd/api/actor-state.h
@@ -464,10 +464,15 @@ class WebSocketRequestResponsePair: public jsg::Object {
 class DurableObjectState: public jsg::Object {
  public:
   DurableObjectState(Worker::Actor::Id actorId,
+      jsg::JsRef<jsg::JsValue> exports,
       kj::Maybe<jsg::Ref<DurableObjectStorage>> storage,
       kj::Maybe<rpc::Container::Client> container);
 
   void waitUntil(kj::Promise<void> promise);
+
+  jsg::JsValue getExports(jsg::Lock& js) {
+    return exports.getHandle(js);
+  }
 
   kj::OneOf<jsg::Ref<DurableObjectId>, kj::StringPtr> getId();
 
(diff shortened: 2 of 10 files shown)


kentonv:
cc @geelen @irvinebroque

geelen:
I think we need to extend the typedefs for `WorkerEntrypoint` and `DurableObject` and `ExecutionContext`:

```ts
type ExportBindings<T> = {
  [K in keyof T]: T[K] extends typeof DurableObject 
    ? DurableObjectNamespace<InstanceType<T[K]>>
    : T[K] extends typeof WorkerEntrypoint
      ? Service<InstanceType<T[K]>>
      : never;
  }

  export abstract class WorkerEntrypoint<Env = unknown, Exports extends Record<string, unknown>>
    implements Rpc.WorkerEntrypointBranded
  {
    [Rpc.__WORKER_ENTRYPOINT_BRAND]: never;
    protected ctx: ExecutionContext<ExportBindings<Exports>>;
    // ...
```

['tested' here](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEAiCuAnMAjEBTA8igVu4ALgHJgC26EADmMOgDwAq06AHgegHYAmMCyaWXPgIA+aAG8AUNBnQuAewDC4KAEIAXNAbToAXx2hIMAMrpEANwCWtAEKXu9gOaNmbTj2gB1eYgDWZgFEOAkQAT0p5e1EJHRkAd3RlIw0tHX0ZQyg4JFQMbDxCGNloSkR5dkJ0LmhEdDAFDhBQ6AB9FoJw9E0AcgVu6ABeaF75boBuNIMVGG8-QOCwiKii2VLy4Sqauoam1vbOnoT+oe6jiZl06B1WCMQCaEyYAFlQ2ExXdm5eHIF84QlLjcfPdHtAXkEQuFIsEPu4Zj5-IgIYtofdxJcdB1KOhoAEWLcCHYHBxHBBGGIhlIZABtADS0Hs0H8oXkADMtABdTQMOkc2FfaBY9Bs7L8PJCQqxaAAflFuUEBWIZAo1FodAAkhwIAQwBxaAxOoxeSIRFLubz+R4hSLZojkVColKZLLTBZrOgiVwnBqtTq9egDdijbSOSandBNBx0OYzOc9DoDPJfbj8cCYJSwa9MAAaTP2pYw3RxnSs+B6giWJM1MsACkILE04lcBIgmjxBM9TjJ1vZ7bTYl0AEoVsAkxB5BgAHQgeSOOsEFiToF3CCTl5vScKJJQYfFYoAen3mbeDN9utoia1E-Q09n88Xy4Iq-BCwdwUnCW3EGHh7zr4L9z2Nq57oJMMhAA)

Similar change required for `DurableObject` base class

Usage:

```
import { WorkerEntrypoint, DurableObject } from "cloudflare:workers";

// These can be autogenerated by `wrangler types`
type Env = {}
type Exports = { MyEntrypoint, AnotherEntrypoint, MyActor, UnconfiguredActor }

export default {
  async fetch(request: Request, env: Env, ctx: ExecutionContext<Exports>) {
    // ...
  }
}
export class MyEntrypoint extends WorkerEntrypoint<Env, Exports> {
  foo(i) { return `foo: ${i}` }
}
export class AnotherEntrypoint extends WorkerEntrypoint<Env, Exports> {
  bar(i) { return `bar: ${i}` }
}
export class MyActor extends DurableObject<Env, Exports> {
  setValue(i) { this.value = i; }
  baz() { return `baz: ${this.value}` }
}
export class UnconfiguredActor extends DurableObject<Env, Exports> {
  qux(i) { return `qux: ${i}` }
}
```

Can bump this to a second PR when we make this non-experimental, if you like? I just wanted to jot down the definition of `ExportBindings<Exports>` while I thought of it

kentonv:
@jasnell Sorry to always assign you on everything but when I looked at the history of the files I'm modifying it's basically all me and you!

jasnell (approved):

kentonv:
(Added a test for `ctx.exports.default`.)

kentonv:
(Added commit to add `ctx.exports` to Durable Objects' `ctx` as well.)

tewaro:
How do I copy the `ctx` after this change for `ExportedHandler::clone`. `addRef` to the contents in `getCtx` doesn't seem to work.

PR #3448, opened 2025-02-01
Add type predicate to node:util#isArray

vicb:


jasnell (approved):

anonrig (approved):
If I'm not mistaken: This is not really necessary because in production people rely on types/node not cloudflare types for nodejs functions.

vicb:
> If I'm not mistaken: This is not really necessary because in production people rely on types/node not cloudflare types for nodejs functions.

It can help if the function is used internally and more generally it is the right thing to do - but yeah, it doesn't affect users of the runtime.

PR #3451, opened 2025-02-03
Run import tests against each package bundle version.

dom96:
Now that we have multiple package bundle versions and we plan to get more, it's important we test the packages in each bundle. This PR generates import tests for packages in each of the bundles and also ensures that new package bundles contain the older bundle's packages.


build/pyodide_bucket.bzl
```diff
@@ -5,170 +5,3 @@ PYODIDE_PACKAGE_BUCKET_URL = "https://pyodide-packages.runtime-playground.worker
 PYODIDE_GITHUB_RELEASE_URL = "https://github.com/cloudflare/pyodide-build-scripts/releases/download/20240513.2/"
 PYODIDE_PACKAGES_TAR_ZIP_SHA256 = "b71d4c3cee3b6bd12969a788545f4159fb1eb984a7ca5de2493c4fa8479beeec"
 PYODIDE_ALL_WHEELS_ZIP_SHA256 = "c17feb45fdcb4b41eab9c719e69c9e062a8fc88344fcb6bbd7de0de92c3ae660"
-
-# IMPORTANT: when updating this file in git, check the diff to make sure none of the imports below are being removed unexpectedly
-PYODIDE_IMPORTS_TO_TEST = {
```
hoodmane, line 10: I think the idea of keeping these in a generated file made sense.

dom96: The problem is that the code that generates this doesn't know about any previous versions that have already been generated, so generating it for all versions isn't possible. I think generating just the dictionary and asking for it to be copied in here is fine.

hoodmane: We can put multiple copies of the generated file with the version in their names and then import `PYODIDE_IMPORTS_TO_TEST` from each one. I don't think it's good to have a large block of auto generated data in the middle of a human-written file.

dom96: Sure, I can do that.

dom96: Fixed

dom96:
Split out the many fixes in this PR into:

* https://github.com/cloudflare/workerd/pull/3477
* https://github.com/cloudflare/workerd/pull/3478
* https://github.com/cloudflare/workerd/pull/3479
* https://github.com/cloudflare/workerd/pull/3480

hoodmane:
I moved the last few of these changes onto the main branch as #3508.

PR #3452, opened 2025-02-03
Add "wderrid" error IDs to opaque "internal error" exceptions

jclee:
...to make it easier to correlate internal error logs with user error reports.

### DIFF
--- a/src/workerd/jsg/util.c++
+++ b/src/workerd/jsg/util.c++
@@ -17,6 +17,7 @@
 #include <cxxabi.h>
 #endif
 
+#include <workerd/util/autogate.h>
 #include <workerd/util/sentry.h>
 
 namespace workerd::jsg {
@@ -125,12 +126,11 @@ InternalErrorId makeInternalErrorId() {
 }
 
 kj::String renderInternalError(InternalErrorId& internalErrorId) {
-  // TODO(now): put "internal error" change behind autogate or compatibility flag?
-  //
-  // It's possible that existing user error handling systems could rely on an exact match with the
-  // existing "internal error" string.  On the other hand, feature flags need a jsg::Lock to read,
-  // which may not be available in all contexts that generate internal errors?
-  return kj::str("internal error; reference = ", internalErrorId);
+  if (util::Autogate::isEnabled(util::AutogateKey::INTERNAL_ERROR_ID)) {
+    return kj::str("internal error; reference = ", internalErrorId);
+  } else {
+    return kj::str("internal error");
+  }
 }
 
 }  // namespace
--- a/src/workerd/jsg/util.c++
+++ b/src/workerd/jsg/util.c++
@@ -6,6 +6,8 @@
 #include "ser.h"
 #include "setup.h"
 
+#include <openssl/rand.h>
+
 #include <kj/debug.h>
 
 #include <cstdlib>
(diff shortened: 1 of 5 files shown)


src/workerd/jsg/util.c++
```diff
@@ -90,9 +92,53 @@ kj::String typeName(const std::type_info& type) {
...
+  return id;
+}
+
+kj::String renderInternalError(InternalErrorId& wdErrId) {
+  // TODO(now): put "internal error" change behind autogate or compatibility flag?
+  //
+  // It's possible that existing user error handling systems could rely on an exact match with the
+  // existing "internal error" string.  On the other hand, feature flags need a jsg::Lock to read,
+  // which may not be available in all contexts that generate internal errors?
+  return kj::str("internal error (wderrid ", wdErrId, ")");
```
kentonv, line 133: I'd suggest a format like:

`internal error; reference = 0123456789abcdefghijklmn`

Because:
* `wderrid` looks almost like a random string of letters, I don't think most people will figure out that it means "workerd error ID". "reference" feels like it has an intuitive meaning, though I'd be open to other suggestions.
* If somehow the "internal error" itself gets reported to Sentry, the format `; identifier = value` will be parsed as a parameter so that the value is not part of the fingerprint.

jclee: Ah, OK... I wasn't sure if it was maybe preferable to include a relatively unique term, to make it easier to search for it in logs, or find information about online.  Sure, I can update the format; I'd expect user-facing errors like "internal error" to rarely go through the sentry handler, but it's possible there could be edge cases.

src/workerd/jsg/util.c++
```diff
@@ -240,6 +292,7 @@ DecodedException decodeTunneledException(
 kj::StringPtr extractTunneledExceptionDescription(kj::StringPtr message) {
   auto tunneledError = tunneledErrorType(message);
   if (tunneledError.isInternal) {
+    // TODO(soon): Include an internal error ID in message, and also return the id.
```
jclee, line 295: This is another potential source of "internal error" messages, but the places where we produce the external exception and internal log differ enough that it will be a more involved change.

Looking at how this function is used, it looks like it's mostly for either startup exceptions -- which should largely be caught during worker validation -- or for exceptions reported in preview, so it's probably a less-frequent source of "internal error"s than the others, too.

src/workerd/jsg/util.c++
jclee: I updated the PR to use the suggested format for the user-facing exception.  I'm preserving the "wderrid" name in internal logs to make it easier to search for, assuming there's not a preferable term.

src/workerd/jsg/util.c++
```diff
@@ -90,9 +92,53 @@ kj::String typeName(const std::type_info& type) {
...
+    KJ_ASSERT(RAND_bytes(id.asPtr().asBytes().begin(), id.size()) == 1);
+  }
+  for (auto i: kj::indices(id)) {
+    id[i] = BASE32_DIGITS[static_cast<unsigned char>(id[i]) % 32];
+  }
+  return id;
+}
+
+kj::String renderInternalError(InternalErrorId& internalErrorId) {
+  // TODO(now): put "internal error" change behind autogate or compatibility flag?
```
justin-mp, line 128: Do we make any guarantees about the error messages we send back?  If we don't, and I don't think we do, I think it's fine to change what we return.  (And hope that we don't lose this round with Hyrum's Law...)

src/workerd/jsg/util.c++
```diff
@@ -90,9 +92,53 @@ kj::String typeName(const std::type_info& type) {
...
+}
+
+}  // namespace
+
 v8::Local<v8::Value> makeInternalError(v8::Isolate* isolate, kj::StringPtr internalMessage) {
-  KJ_LOG(ERROR, internalMessage);
-  return v8::Exception::Error(v8StrIntern(isolate, "internal error"));
+  auto wdErrId = makeInternalErrorId();
+  KJ_LOG(ERROR, internalMessage, wdErrId);
+  return v8::Exception::Error(v8StrIntern(isolate, renderInternalError(wdErrId)));
```
justin-mp, line 141: Given that every internal error is a brand new string, do we want to intern them?  With high probability, we're only going to use them once!

jclee: Oh, good catch...  I'd noticed and fixed this for the instance in `decodeTunneledException()`, but I missed it here.

jclee:
Updated to fix `v8StrIntern()` -> `v8Str()` occurrence, and also fix lint.

https://github.com/cloudflare/workerd/compare/b68407db..82d3fed42b5f2e8b29889426d12fd4acb766d834

src/workerd/jsg/util.c++
jclee: (fixed)

src/workerd/jsg/util.c++
jclee: Heh... yeah, that's kind of what I was wondering.  I do know that the custom `.retryable` and `.overloaded` exception properties were introduced in part to reduce the necessity of inspecting exception strings, as was observed in some internal workers.

But IIRC, those workers didn't look for bare `internal error` messages...  which I guess makes sense, since they should be pretty rare, since they go to internal logs.

kentonv: Hmm, anyone who depends on string-matching `"internal error"` is on extremely precarious ground since it always represents some sort of bug in our system (where often the bug is simply "we forgot to assign a better error message")...

I could imagine that someone is doing it, but what I'm not sure of is whether it's likely that we'll outright break them by changing the message. More likely we'll just slightly degrade their own error handling?

I'm inclined to say we don't need a compat flag here... I think...

src/workerd/jsg/util.c++
```diff
@@ -90,9 +92,53 @@ kj::String typeName(const std::type_info& type) {
...
+  return id;
+}
+
+kj::String renderInternalError(InternalErrorId& internalErrorId) {
+  // TODO(now): put "internal error" change behind autogate or compatibility flag?
+  //
+  // It's possible that existing user error handling systems could rely on an exact match with the
+  // existing "internal error" string.  On the other hand, feature flags need a jsg::Lock to read,
+  // which may not be available in all contexts that generate internal errors?
+  return kj::str("internal error; reference = ", internalErrorId);
```
kentonv, line 133: I wonder if we should use the opportunity to improve this error message, like...

"The Workers Runtime suffered an internal error, which has already been reported to the Workers team."

Although that text would be false when people are running workerd locally...

justin-mp: Personally, I much prefer the terse "internal error".

justin-mp (approved):

src/workerd/jsg/util.c++
kentonv: Users often don't understand what it means, though...

kentonv (approved):

jclee:
Put new behavior behind autogate (51123f6), rebased

PR #3463, opened 2025-02-04
Add ncrypto dependency, implement key handling operations

jasnell:
Still very much a work in progress.

This is gonna be a big one.


src/node/internal/crypto_keys.ts
```diff
@@ -276,7 +279,24 @@ export function createSecretKey(
...
   validateKeyData(key, 'key');
-  if (typeof key === 'string') key = Buffer.from(key as string, encoding);
+  if (typeof key === 'string') {
+    key = Buffer.from(key as string, encoding);
+  } else if (isAnyArrayBuffer(key)) {
+    // We want the key to be a copy of the original buffer, not a view.
+    key = Buffer.from(new Uint8Array(key));
+  } else if (isArrayBufferView(key)) {
+    // We want the key to be a copy of the original buffer, not a view.
+    key = Buffer.from(key as any);
```
anonrig, line 289: This will fail with eslint. I think we need to fix our types for Buffer.from. Maybe a good TODO comment here would be sufficient for now.

src/workerd/api/crypto/scrypt.c++
```diff
@@ -18,20 +24,17 @@ kj::Maybe<jsg::BufferSource> scrypt(jsg::Lock& js,
...
-    // are lots of places in the API currently where the errors do not match.
-    if (clearErrorOnReturn.peekError()) {
-      throwOpensslError(__FILE__, __LINE__, "crypto::scrypt");
-    }
-    return kj::none;
+  ncrypto::ClearErrorOnReturn clearErrorOnReturn;
+  auto backing = jsg::BackingStore::alloc<v8::ArrayBuffer>(js, length);
+  ncrypto::Buffer<kj::byte> buf(backing.asArrayPtr().begin(), backing.size());
+  ncrypto::Buffer<const char> passbuf{
+    .data = reinterpret_cast<const char*>(pass.begin()),
```
anonrig, line 31: is there a method that just returns const char*? I remember something like asChars().begin()?

src/workerd/api/crypto/spkac.c++
```diff
@@ -1,102 +1,63 @@
...
-    KJ_IF_SOME(key, tryOwnPkey(spki)) {
-      return NETSCAPE_SPKI_verify(spki.get(), key.get()) > 0;
-    }
-  }
-  return false;
+
+  ncrypto::Buffer<const char> buf{
+    .data = reinterpret_cast<const char*>(input.begin()),
+    .len = input.size(),
+  };
```
anonrig, line 33: Nit:

Can we just have a conversion function or a static method that converts to ncrypto::Buffer? It seems there are lots of repetitions.

src/workerd/api/crypto/spkac.c++
```diff
@@ -1,102 +1,63 @@
...
-      }
-    }
+  ncrypto::Buffer<const char> buf{
+    .data = reinterpret_cast<const char*>(input.begin()),
+    .len = input.size(),
+  };
+  if (auto bio = ncrypto::ExportPublicKey(buf)) {
+    BUF_MEM* bptr = bio;
+    auto buf = jsg::BackingStore::alloc(js, bptr->length);
+    auto aptr = kj::arrayPtr(bptr->data, bptr->length);
```
anonrig, line 45: Do you need to check for length/emptiness for safety?

src/workerd/api/node/crypto-keys.c++
```diff
@@ -20,24 +20,38 @@ namespace {
...
+    keyData.setToZero();
+  }
 
   kj::StringPtr getAlgorithmName() const override {
     return "secret"_kj;
   }
   CryptoKey::AlgorithmVariant getAlgorithm(jsg::Lock& js) const override {
-    return CryptoKey::KeyAlgorithm{.name = "secret"_kj};
+    return CryptoKey::ArbitraryKeyAlgorithm{
+      .name = "secret"_kj,
```
anonrig, line 35: Nit:

It would be good to keep these in sync

```suggestion
      .name = getAlgorithmName(),
```

src/workerd/api/node/crypto.c++
```diff
@@ -10,8 +10,13 @@
 #include <workerd/api/crypto/spkac.h>
 #include <workerd/jsg/jsg.h>
 
+#include <ncrypto.h>
+
 namespace workerd::api::node {
 
+// ======================================================================================
+#pragma region KDF
```
anonrig, line 18: What's the benefit of having pragma region?

src/workerd/api/node/crypto.h
```diff
@@ -252,3 +252,8 @@ class CryptoImpl final: public jsg::Object {
       api::node::CryptoImpl::KeyExportOptions, api::node::CryptoImpl::GenerateKeyPairOptions,      \
       api::node::CryptoImpl::CreateAsymmetricKeyOptions, EW_CRYPTO_X509_ISOLATE_TYPES
 }  // namespace workerd::api::node
+
+// Known todos
```
anonrig, line 256:
```suggestion
// TODO(soon): Known todos 
```

src/workerd/jsg/buffersource.h
```diff
@@ -382,10 +390,24 @@ class BufferSource {
...
+  inline BufferSource copy(jsg::Lock& js) {
+    KJ_IF_SOME(backing, maybeBackingStore) {
+      return BufferSource(js, backing.copy<T>(js));
+    }
+    return BufferSource(js, BackingStore::alloc<T>(js, 0));
+  }
+
+  inline void setToZero() {
+    KJ_IF_SOME(backing, maybeBackingStore) {
+      memset(backing.asArrayPtr().begin(), 0, backing.size());
```
anonrig, line 407: I think we should also check for size here before calling memset.

src/workerd/jsg/buffersource.h
```diff
@@ -208,6 +208,14 @@ class BackingStore {
     return BackingStore(backingStore, byteLength, byteOffset, elementSize, ctor, integerType);
   }
 
+  template <typename T = v8::Uint8Array>
```
anonrig, line 211: Nit:

Any particular reason for not using class?

```suggestion
  template <class T = v8::Uint8Array>
```

src/workerd/api/node/crypto.c++
jasnell: vscode makes these sections collapsible/foldable in the UI

src/node/internal/crypto_keys.ts
jasnell: I'll worry about lint fixups when the PR is mostly finished.

src/node/internal/crypto_keys.ts
```diff
@@ -81,6 +92,12 @@ import { validateObject, validateString } from 'node-internal:validators';
 // existed first. We're, however, going to layer our KeyObject on top of
 // CryptoKey with a few augmentations.
 
+function isStringOrBuffer(val: any) {
```
anonrig, line 95:
```suggestion
function isStringOrBuffer(val: any): asserts val is string | Buffer {
```

src/node/internal/crypto_keys.ts
```diff
@@ -148,6 +166,8 @@ export abstract class KeyObject {
       }
     }
 
+    options.format ??= 'buffer';
```
anonrig, line 169: This can just move to the else statement of line 149.

src/node/internal/crypto_keys.ts
```diff
@@ -190,6 +210,10 @@ export abstract class KeyObject {
   }
 }
 
+export function isKeyObject(obj: any): obj is KeyObject {
```
anonrig, line 213:
```suggestion
export function isKeyObject(obj: any): asserts obj is KeyObject {
```

src/node/internal/crypto_keys.ts
```diff
@@ -276,101 +338,112 @@ export function createSecretKey(
...
-//     }
-//   }
-//   return inner;
-// }
+function prepareAsymmetricKey(
+  key: CreateAsymmetricKeyOptions | KeyData,
+  ctx: KeyContext
+): InnerCreateAsymmetricKeyOptions {
+  // Safety check... key should not be undefined or null here.
+  if ((key as any) == null) {
```
anonrig, line 367: If we are doing this, we should change key type...

src/node/internal/crypto_keys.ts
```diff
@@ -276,101 +338,112 @@ export function createSecretKey(
...
+    key: data,
+    encoding = 'utf8',
+    format = 'pem',
+    type,
+    passphrase,
+  } = normalized;
+
+  // The key data must be specified. The value has to be one of either a
+  // string, an ArrayBuffer, an ArrayBufferView, or a JWK object.
+  if ((data as any) == null || isKeyObject(data) || data instanceof CryptoKey) {
```
anonrig, line 396: This (ref: as any) indicates that our types are wrong

src/workerd/api/crypto/prime.c++
```diff
@@ -45,63 +44,57 @@ jsg::BufferSource randomPrime(jsg::Lock& js,
...
-    if (rem) {
-      // This would definitely lead to an infinite loop if allowed since
-      // OpenSSL does not check this condition.
-      JSG_REQUIRE(
-          BN_cmp(add, rem) == 1, RangeError, "options.rem must be smaller than options.add");
-    }
   }
 
-  // BN_generate_prime_ex() calls RAND_bytes_ex() internally.
+  // Generating random primes uses the prng internally.
```
anonrig, line 70: nit

```suggestion
  // Generating random primes uses the PRNG internally.
```

src/workerd/api/crypto/scrypt.c++
```diff
@@ -3,13 +3,19 @@
...
 
-#include <workerd/api/crypto/kdf.h>
-
-#include <openssl/evp.h>
+#include <ncrypto.h>
 
 namespace workerd::api {
 
+namespace {
+template <typename T = const kj::byte>
```
anonrig, line 13: I see that this function is defined in multiple files. I recommend moving to a common denominator

src/workerd/api/node/BUILD.bazel
```diff
@@ -115,7 +116,54 @@ wd_test(
 wd_test(
     src = "tests/crypto_keys-test.wd-test",
     args = ["--experimental"],
-    data = ["tests/crypto_keys-test.js"],
+    data = [
```
anonrig, line 119: We can simplify this and just use glob()

src/workerd/api/crypto/scrypt.c++
jasnell: It will be.

src/workerd/api/node/BUILD.bazel
jasnell: It will be when done. Keeping it like this for now while making sure things are working.

src/workerd/api/crypto/spkac.c++
jasnell: For these, we shouldn't. The copy operation that follows should be doing the right thing.

src/node/internal/crypto_keys.ts
jasnell: That doesn't work here. adding the `asserts ...` is not compatible with the function returning a boolean.

src/node/internal/crypto_keys.ts
jasnell: Given that the function returns boolean, adding the asserts doesn't work.

src/node/internal/crypto_keys.ts
jasnell: This is intentional. Even when using typescript it is still possible for the value to be passed as null/undefined at runtime so this check adds an additional fallback.

WORKSPACE
```diff
@@ -80,6 +80,14 @@ http_archive(
     url = "https://github.com/nodejs/nbytes/archive/refs/tags/v0.1.1.tar.gz",
 )
 
+http_archive(
+    name = "ncrypto",
+    sha256 = "c8284e4719130916cbce340989a59249e605dfbdecd496910d6ed2d1ecaf69ad",
+    strip_prefix = "ncrypto-main",
```
anonrig, line 86: Let's make sure we release a new version before merging this PR.

jasnell: Yep, there's likely another PR or two that I'll be adding as I work on a few things. Doing an actual release is one of the necessary next steps.

src/workerd/jsg/buffersource.h
```diff
@@ -382,10 +390,24 @@ class BufferSource {
...
 
+  template <class T = v8::Uint8Array>
+  inline BufferSource copy(jsg::Lock& js) {
+    KJ_IF_SOME(backing, maybeBackingStore) {
+      return BufferSource(js, backing.copy<T>(js));
+    }
+    return BufferSource(js, BackingStore::alloc<T>(js, 0));
+  }
+
+  inline void setToZero() {
```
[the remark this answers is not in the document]
jasnell: Not required but also doesn't hurt anything.

src/workerd/api/crypto/jwk.c++
```diff
@@ -0,0 +1,303 @@
...
+      case EVP_PKEY_EC:
+        return jwkFromEcKey(key, keyType);
+      case EVP_PKEY_RSA:
+        return jwkFromRsaKey(key, keyType);
+      case EVP_PKEY_RSA2:
+        return jwkFromRsaKey(key, keyType);
+      case EVP_PKEY_RSA_PSS:
+        return jwkFromRsaKey(key, keyType);
+      case EVP_PKEY_DSA: {
+        // DSA keys are not supported for JWK export.
```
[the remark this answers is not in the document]
jasnell: Addressed these locally so marking resolved. The actual updates may not be pushed until the next round of updates tho

src/node/internal/crypto_keys.ts
```diff
@@ -204,6 +253,28 @@ abstract class AsymmetricKeyObject extends KeyObject {
   get asymmetricKeyType(): AsymmetricKeyType {
     return cryptoImpl.getAsymmetricKeyType(this[kHandle]);
   }
+
+  toCryptoKey() {
+    // TODO(soon): Implement the toCryptoKey API (added in Node.js 23.0.0)
+    throw new ERR_METHOD_NOT_IMPLEMENTED('toCryptoKey');
+  }
+
+  [kInspect](depth: number, options: any) {
```
anonrig, line 262: Can we use proper types for options?

src/node/internal/crypto_keys.ts
```diff
@@ -238,6 +309,22 @@ export class SecretKeyObject extends KeyObject {
   get type(): KeyObjectType {
     return 'secret';
   }
+
+  [kInspect](depth: number, options: any) {
```
anonrig, line 313: Ditto for any usage

src/node/internal/crypto_keys.ts
anonrig: if you change line 388 to `| null | undefined`, we won't need this as any check

src/node/internal/crypto_keys.ts
anonrig: I think the main functionality is done to worry about types :-))

src/node/internal/crypto_keys.ts
```diff
@@ -380,48 +478,69 @@ export function createPrivateKey(
...
+  // Passing a KeyObject or a CryptoKey allows deriving the public key
+  // from an existing private key.
+
+  if (isKeyObject(key)) {
+    if (key.type !== 'private') {
+      throw new ERR_INVALID_ARG_TYPE('key', 'PrivateKeyObject', key);
+    }
+    return KeyObject.from(
+      cryptoImpl.createPublicKey({
+        key: (key as KeyObject)[kHandle],
```
anonrig, line 509: If line 503 isKeyObject call assertion works, we won't need this key as KeyObject check. I think there is a bug here...

anonrig (approved):

anonrig:
Amazing work as always @jasnell. There are some TS related comments that I'm not happy about, but we can fix it afterwards. No need to block progress!

jasnell:
While this is generally ready to go for workerd, the production project uses a different version of boringssl that will require some additional tweaks in ncrypto to get working. This PR will remain in draft until those are done.

WORKSPACE
```diff
@@ -80,6 +80,14 @@ http_archive(
     url = "https://github.com/nodejs/nbytes/archive/refs/tags/v0.1.1.tar.gz",
 )
 
+http_archive(
```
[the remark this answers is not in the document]
jasnell: Can/should be done separately

vicb:
Does this unblock some of https://github.com/cloudflare/workerd/blob/186fa11d99d570d72d1d243406f7525ee9db8ce0/src/node/crypto.ts#L250 or is this PR only adding thing that will be used later to implement those?

jasnell:
@vicb ... it adds implementations for the following:

* `createPrivateKey()`
* `createPublicKey()`
* `generateKeyPair()`
* `generateKeyPairSync()`
* `PublicKeyObject`
* `PrivateKeyObject`

With caveats, of course. Key generation of DSA and DH key pairs do not work, for instance, due to current limitations in boringssl. There will be several followup PRs soon with more implementation detail.

jasnell:
@panva ... fyi ... just a heads up that this is landing.

panva:
@jasnell thanks for the heads up. I think it's a move in the right direction for compatibility.

That being said I'm switching jose fully to WebCryptoAPI with a new major and ESM-only as soon as 20.19.0 releases with unflagged require(esm). It will still support KeyObject as inputs (and will use [KeyObject.prototype.toCryptoKey](https://github.com/nodejs/node/pull/55262) to crossover KeyObject to CryptoKey efficiently) and will also use `globalThis?.process?.getBuiltinModule?.(id)` for small optimizations when available here and there, e.g. `new X509Certificate(pem)`).

If you don't support [KeyObject.prototype.toCryptoKey](https://github.com/nodejs/node/pull/55262) yet I'd suggest getting that in.

jasnell:
Yep, `KeyObject.prototype.toCryptoKey(...)` will be coming soon (hopefully in the next couple of weeks).

vicb:
> use `globalThis?.process?.getBuiltinModule?.(id)`

@panva could you expand how you will use that? FYI when using node compatibility with workerd, the return value could be different from an import.

panva:
@vicb here are all three instances in the v6.x-wip state at the moment

https://github.com/panva/jose/blob/151000e84b94551e7408f988e3d52f8ccf530a64/src/lib/asn1.ts#L253-L266
https://github.com/panva/jose/blob/151000e84b94551e7408f988e3d52f8ccf530a64/src/lib/buffer_utils.ts#L6-L21
https://github.com/panva/jose/blob/151000e84b94551e7408f988e3d52f8ccf530a64/src/lib/decrypt.ts#L11-L33

vicb:
> @vicb here are all three instances in the v6.x-wip state at the moment
> 
> https://github.com/panva/jose/blob/151000e84b94551e7408f988e3d52f8ccf530a64/src/lib/asn1.ts#L253-L266 https://github.com/panva/jose/blob/151000e84b94551e7408f988e3d52f8ccf530a64/src/lib/buffer_utils.ts#L6-L21 (used in other files for base64url handling, this will be replaced with https://github.com/tc39/proposal-arraybuffer-base64 when available) https://github.com/panva/jose/blob/151000e84b94551e7408f988e3d52f8ccf530a64/src/lib/decrypt.ts#L11-L33
> 
> I am specifically not interested in using an import() here because of how bundlers tend to either error out or insert shitty polyfills when they detect node modules in non-node runtime targets.

Thanks - I'll look more deeply next week.

One thing that could be better for workers is that you have an adapter that could be implemented for different platforms.

i.e.

```ts
import type { X509Certificate } from "node:crypto";

interface CryptoAdapter {
  X509Certificate: X509Certificate;
  // ...
}
```

You can have a default impl based on get `getBuiltinModule` but it leaves an option for platforms not (fully) supporting it.

For workers `getBuiltinModule` will return the workerd module which might not implement all the methods. However when you `import crypto from 'node:crypto';` we return an hybrid implementation (some APIs are implemented in polyfill). 

(Note that we do not polyfill `getBuiltinModule` to return the polyfills because it would mean that we can not detect the unused polyfills at build time and tree shake them).

panva:
@vicb how can i enable node-compat in this capnp definition [here](https://github.com/panva/jose/blob/ff8ea69f5d569c2ffa69a9329a1b74215f2dc3eb/tap/.workerd.sh)?

vicb:
> compatibilityFlags = ["nodejs_compat_v2"]

That should do.

However the recommended way it to use `compatibilityFlags = ["nodejs_compat"]` as long as your compatibility date is >= `2024-09-23` (because `nodejs_compat` will also activate v2 after this date).

Also note that when using wrangler, we add polyfills when the node compatibility is activated (`v2` or `nodejs_compat` with a date >= `2024-09-23`).

If you want to test with the polyfill, you would have to use `wrangler`. 

For example, instead of invoking `esbuild` in your script, you can invoke `wrangler deploy --dry-run -outdir=/path/to/dist` (You can also pass it the entry point, the compat date, the the compat flags - because the entry point and the flags aren't dynamic, they could also be in your `wrangler.{toml,json}`). That's similar to [this test](https://github.com/cloudflare/workers-sdk/blob/aaa9cca4deeaab3155dcedeca0177339b24246af/fixtures/nodejs-hybrid-app/package.json#L5) in the workers-sdk.

panva:
@vicb i get the same failure regardless of the compat flag version

vicb:
Yep, that's expected because they should be the same :)

To debug that, it's probably easier to create a `wrangler.json` in this directory, point to a `worker.ts` that should look something like:

```ts
// worker.ts
export default {
    // call your test from here (i.e. run.ts)
    await test();
    return new Response('Hello World!');
  },
};
```

Then you can `wrangler dev`, open the dev tools, add breakpoints or at least break on exception and then open the worker URL in a browser / curl it.

Let me know if you need more info (feel free to email me at the address on my GH profile).

PR #3464, opened 2025-02-05
Use new V8 v2 write APIs

danlapid:


danlapid:
This seems to not really work.
The new APIs aren't really compatible with the old ones.
Two problems from what I can see: 
1. Missing the read_nchars variable means it's hard to tell how much of the original string was read.
2. `WriteOneByteV2` does `DCHECK(IsOneByte());` which we fail in some of our tests that expect the value you be brokenly encoded successfully. We can patch that check probably

fhanau:
> This seems to not really work. The new APIs aren't really compatible with the old ones. Two problems from what I can see:
> 
> 1. Missing the read_nchars variable means it's hard to tell how much of the original string was read.
> 2. `WriteOneByteV2` does `DCHECK(IsOneByte());` which we fail in some of our tests that expect the value you be brokenly encoded successfully. We can patch that check probably
> 3. `WriteUtf8V2` works differently in some way I don't quite understand.

At least the 2) can be worked around well: WriteV2() is identical to WriteOneByteV2() except that it doesn't have the DCHECK that is causing us problems.
One difference in options based on the implementation of these in api.cc: `NO_NULL_TERMINATION` behavior may have changed – in the old mode we would add a null byte only if the input buffer was shorter than the output buffer. In the new mode we write even if they are the same size.

anonrig:
Upon talking with @danlapid, I'll take over this pull-request. I'll make some progress tomorrow.

PR #3465, opened 2025-02-05
fix: Pass ctx.props to default handlers

tewaro:


src/workerd/io/worker.c++
```diff
@@ -1971,6 +1971,10 @@ kj::Maybe<kj::Own<api::ExportedHandler>> Worker::Lock::getExportedHandler(
 
   kj::StringPtr n = name.orDefault("default"_kj);
   KJ_IF_SOME(h, worker.impl->namedHandlers.find(n)) {
+    jsg::Lock& js = *this;
+    if (FeatureFlags::get(js).getCtxPropsToDefault()) {
+      h.ctx = jsg::alloc<api::ExecutionContext>(js, props.toJs(js));
```
kentonv, line 1976: This is overwriting the `ctx` on the singleton `ExecutionContext` which is shared by all events. This seems unsafe -- a race condition between concurrent events could cause one event to receive the other one's props.

tewaro: Where is the appropriate place to do this assignment?

kentonv: Oops, I meant singleton `ExportedHandler`, not `ExecutionContext`. (You are creating a new `ExecutionContext`, but attaching it to the shared `ExportedHandler`, possibly overwriting a previous value that is still in use.)

We need to refactor the code to create a new `ExportedHandler` instance for each invocation, like we do for `WorkerEntrypoint` entrypoints.

tewaro: would it be as simple as copying h and then returning an owned version of that?

kentonv: `ExportedHandler` currently doesn't have a simple way to copy it. You may need to add a `clone()` method which calls each member's `addRef()` method (`jsg::Function`, `jsg::Value`, `jsg::Ref`, and `jsg::SelfRef` should all have versions of this method).

src/workerd/io/worker.c++
```diff
@@ -1971,7 +1971,15 @@ kj::Maybe<kj::Own<api::ExportedHandler>> Worker::Lock::getExportedHandler(
 
   kj::StringPtr n = name.orDefault("default"_kj);
   KJ_IF_SOME(h, worker.impl->namedHandlers.find(n)) {
-    return fakeOwn(h);
+    kj::Own<api::ExportedHandler> ret = fakeOwn(h);
+    jsg::Lock& js = *this;
+    if (FeatureFlags::get(js).getCtxPropsToDefault()) {
+      api::ExportedHandler constructedHandler = h.clone(js);
+      constructedHandler.ctx = jsg::alloc<api::ExecutionContext>(js, props.toJs(js));
+      kj::Own<api::ExportedHandler> handlerPtr = fakeOwn(constructedHandler);
```
tewaro, line 1979: @kentonv doing this feels off, how do I get a proper allocation for an ExportedHandler?

kentonv: As written this results in use-after-free -- definitely bad.

You can replace `fakeOwn(constructedHandler)` with `kj::heap(mv(constructedHandler))` to move it to a heap-allocated object.

src/workerd/io/compatibility-date.capnp
```diff
@@ -692,4 +692,9 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $experimental
       $neededByFl;
   # Enables cache settings specified request in fetch api cf object to override cache rules. (only for user owned or grey-clouded sites)
+
+  ctxPropsToDefault @73: Bool
+      $compatEnableFlag("ctx_props_to_default_enabled")
```
kentonv, line 697: This isn't about default vs. non-default entrypoints, it's about class vs. non-class. The default entrypoint can be class-based, and named entrypoints can be non-class-based.

Perhaps we can name these:
* enable: `unique_ctx_per_invocation`
* disable: `nonclass_entrypoint_reuses_ctx_across_invocations`

src/workerd/io/compatibility-date.capnp
```diff
@@ -692,4 +692,9 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatDisableFlag("ctx_props_to_default_disabled")
+      $experimental;
```
kentonv, line 699: I think there's no need to mark this experimental, in fact let's go ahead and give it an enable date of `2025-03-03`.

src/workerd/api/global-scope.h
```diff
@@ -382,6 +382,15 @@ struct ExportedHandler {
   jsg::Optional<jsg::Ref<ExecutionContext>> getCtx() {
     return ctx.map([&](jsg::Ref<ExecutionContext>& p) { return p.addRef(); });
   }
+
+  ExportedHandler clone(jsg::Lock& js) {
+    return ExportedHandler{
```
kentonv, line 387: You need to clone _all_ of the members, e.g. `fetch`, `trace`, etc.

Please make sure all of the tests pass with your compat flag forced to be enabled.

src/workerd/io/compatibility-date.capnp
```diff
@@ -692,4 +692,9 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $experimental
       $neededByFl;
   # Enables cache settings specified request in fetch api cf object to override cache rules. (only for user owned or grey-clouded sites)
+
+  uniqueCtxPerInvocation @73: Bool
```
danlapid, line 696: Add a comment for this flag please

src/workerd/api/global-scope.h
tewaro: How do I do that?

jasnell:
What exactly is the intent here? The PR description needs some explanation and detail. If the idea is to give every invocation it's own unique context, then rather than cloning the entire thing the way this PR does, could it make sense to simply wrap the context in a JS Proxy object for every request and either CoW the underlying context or just intercept mutations so that the original remains unmodified? Wrapping in a proxy could likely reduce much of the complexity here.

src/workerd/api/global-scope.h
```diff
@@ -382,6 +382,35 @@ struct ExportedHandler {
   jsg::Optional<jsg::Ref<ExecutionContext>> getCtx() {
     return ctx.map([&](jsg::Ref<ExecutionContext>& p) { return p.addRef(); });
   }
+
+  template <typename T>
+  static jsg::LenientOptional<T> ADDREF(jsg::LenientOptional<T>& function, jsg::Lock& js) {
```
kentonv, line 387: We don't use all-caps naming for functions. Maybe we could call this `mapAddRef`, which is a name we've used for similar helpers elsewhere.

src/workerd/api/global-scope.h
```diff
@@ -382,6 +382,35 @@ struct ExportedHandler {
...
+    return function.map([&](T& a) { return a.addRef(js); });
+  }
+
+#define CLONE(x)                                                                                   \
```
kentonv, line 391: Defining this macro means that anything in the codebase which uses the identifier `CLONE` will now be broken.

Generally, there are two ways to solve this:
* Use a namespace prefix on the macro so that it is unlikely to collide, like `WORKERD_EXPORTED_HANDLER_CLONE`.
* `#undef` the macro when you're done using it. (This is best for cases like this, where the macro is only meant to be used locally.)

However, I would recommend just _not_ using a macro here. It doesn't seem worth it, the initializers are not that verbose:

```
.fetch = mapAddRef(js, fetch),
```

src/workerd/api/global-scope.h
kentonv, line 387: As a rule whenever we have a parameter `jsg::Lock& js`, we always make it the first parameter.

kentonv, line 387: Please make this private. It doesn't belong in the public interface.

src/workerd/io/worker.c++
```diff
@@ -1971,7 +1971,15 @@ kj::Maybe<kj::Own<api::ExportedHandler>> Worker::Lock::getExportedHandler(
 
   kj::StringPtr n = name.orDefault("default"_kj);
   KJ_IF_SOME(h, worker.impl->namedHandlers.find(n)) {
-    return fakeOwn(h);
+    kj::Own<api::ExportedHandler> ret = fakeOwn(h);
```
kentonv, line 1974: Why declare this variable `ret` here when it's only used in one case below? Can the last line just be `return fakeOwn(h);`?

src/workerd/io/worker.c++
```diff
@@ -1971,7 +1971,15 @@ kj::Maybe<kj::Own<api::ExportedHandler>> Worker::Lock::getExportedHandler(
...
+    jsg::Lock& js = *this;
+    if (FeatureFlags::get(js).getUniqueCtxPerInvocation()) {
+      api::ExportedHandler constructedHandler = h.clone(js);
+      constructedHandler.ctx = jsg::alloc<api::ExecutionContext>(js, props.toJs(js));
+      kj::Own<api::ExportedHandler> handlerPtr = kj::heap(kj::mv(constructedHandler));
```
kentonv, line 1979: Why declare `handlePtr` when you're just returning it on the next line?

```suggestion
      return kj::heap(kj::mv(constructedHandler));
```

kentonv:
> What exactly is the intent here? The PR description needs some explanation and detail.

The goal of the change is to fix a known bug in which `ctx.props` is not available for non-class-based entrypoints because such entrypoints have always reused the same `ctx` object for all invocations. There's a comment on line 1674 describing the issue:

```
                        // TODO(cleanup): Unfortunately, for non-class-based handlers, we have
                        //   always created only a single `ctx` object and reused it for all
                        //   requests. This is weird and obviously wrong but changing it probably
                        //   requires a compat flag. Until then, connection properties will not be
                        //   available for non-class handlers.
```

BTW @tewaro we should update that comment to something like:

"Historically, non-class-based handlers reused the same `ctx` object for all requests. This was an accident, but some Workers depend on it. Newer worker with the `unique_ctx_per_invocation` will allocate a new `ctx` for every request."

> Wrapping in a proxy could likely reduce much of the complexity here.

That sounds way more complicated to me. This change as-is is really not very complex at all.

src/workerd/api/global-scope.h
```diff
@@ -382,6 +382,32 @@ struct ExportedHandler {
...
     return ctx.map([&](jsg::Ref<ExecutionContext>& p) { return p.addRef(); });
   }
+
+ private:
+  template <typename T>
+  static jsg::LenientOptional<T> mapAddRef(jsg::Lock& js, jsg::LenientOptional<T>& function) {
+    return function.map([&](T& a) { return a.addRef(js); });
+  }
+
+ public:
```
kentonv, line 392: Please put all private stuff at the end of the struct/class, don't interleave private and public. Private stuff adds noise for people who are just trying to read the public interface, so it's best to keep it clearly separate.

tewaro:
@kentonv how do I force the feature flag on all tests? do I need to modify each one?

tewaro:
> @tewaro I'm just proposing that you temporarily change your code locally to enable the flag by default and make sure the tests actually pass.

I've already done a test where I bump supported-compatibility-date.txt locally, is there anything more I need to do?

kentonv:
Yes, please add a unit test of some sort, where the test directly enables the flag and verifies that ctx.props is then present when expected. You can perhaps modify or duplicate one of the existing tests for ctx.props in server-test.c++.

kentonv:
> I've already done a test where I bump supported-compatibility-date.txt locally, is there anything more I need to do?

Oh, this doesn't actually accomplish what I asked for earlier.

To be clear, you need to do two things:
1. Include a unit test in the PR itself, which enables the flag and verifies it does what you wanted it to do. In general, you should _always_ do this, in _every_ PR, _before_ you post the PR for review.
2. Unique to this PR: I recommend forcing the flag to be on and running all the existing tests. We would not expect this flag to break any of our tests, so this would help shake out any weird edge cases that we might have missed. Bumping supported-compatibility-date.txt does not cause the flag to be turned on. What I'm suggesting is you temporarily set the flag default value to true, or even just temporarily remove the `if` statement so that the true branch always runs.

tewaro:
The above commit is to show that the tests pass with the change.

src/workerd/io/worker.c++
```diff
@@ -1977,6 +1975,12 @@ kj::Maybe<kj::Own<api::ExportedHandler>> Worker::Lock::getExportedHandler(
 
   kj::StringPtr n = name.orDefault("default"_kj);
   KJ_IF_SOME(h, worker.impl->namedHandlers.find(n)) {
+    jsg::Lock& js = *this;
+    // if (FeatureFlags::get(js).getUniqueCtxPerInvocation()) {
```
danlapid, line 1979: this shouldn't be commented

danlapid:
This should still have a unit test though.

tewaro:
Unit tests added.
Commit f14be28 shows that all tests pass with the new behavior. 
Commit f0e7b21 adds back the compatability flags checks.

src/workerd/api/global-scope.h
```diff
@@ -396,6 +396,31 @@ struct ExportedHandler {
   jsg::Optional<jsg::Ref<ExecutionContext>> getCtx() {
     return ctx.map([&](jsg::Ref<ExecutionContext>& p) { return p.addRef(); });
   }
+
+  ExportedHandler clone(jsg::Lock& js) {
+    return ExportedHandler{
```
jasnell, line 401: Nit: would generally prefer the body of this to be in `global-scope.c++`.

src/workerd/api/ctx-props-test.wd-test
```diff
@@ -0,0 +1,53 @@
...
+          (name = "worker-props-disabled", esModule =
+            `import assert from 'node:assert';
+            `export default {
+            ` async fetch(request, env, ctx) {
+            `   return new Response(JSON.stringify(ctx.props));
+            ` },
+            ` async test(request, env, ctx) {
+            `   const resp = await env.SERVICE.fetch(new Request("https://placeholder/hello"));
+            `   const body = await resp.json();
+            `   assert(JSON.stringify(body) === '{}');
```
jasnell, line 17:
```suggestion
            `   assert.strictEqual(JSON.stringify(body), '{}');
```

jasnell: Or use `deepStrictEqual` and skip the JSON.stringify

kentonv (approved):

kentonv:
Internal build failure looks like a flake.

PR #3467, opened 2025-02-05
D1 Sessions API - Rename commit token to bookmark for API and comments

lambrospetrou:
As title says, simple renames of comments and the `getCommitToken()` to be `getBookmark()`.

The HTTP header sent and used by the D1 API (not public REST API), just between the D1 binding and our API remains `x-cf-d1-session-commit-token`. Its rename will be done in a separate PR when our internal API supports both headers.

### DIFF
--- a/src/cloudflare/internal/d1-api.ts
+++ b/src/cloudflare/internal/d1-api.ts
@@ -54,18 +54,21 @@ type SQLError = {
 
 type ResultsFormat = 'ARRAY_OF_OBJECTS' | 'ROWS_AND_COLUMNS' | 'NONE';
 
-type D1SessionCommitTokenOrConstraint = string;
-type D1SessionCommitToken = string;
+type D1SessionBookmarkOrConstraint = string;
+type D1SessionBookmark = string;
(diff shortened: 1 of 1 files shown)


src/cloudflare/internal/d1-api.ts
```diff
@@ -208,13 +210,9 @@ class D1DatabaseSession {
       init.headers = h;
     }
     return this.fetcher.fetch(input, init).then((resp) => {
-      const newCommitToken = resp.headers.get(
-        D1_SESSION_COMMIT_TOKEN_HTTP_HEADER
-      );
-      // TODO(soon): Add validation of the received commit token, in case we sent a commit token,
```
justin-mp, line 214: Just noting that it's OK to lose this TODO because the internal DO methods for `waitForBookmark` should validate the token for D1.

justin-mp (approved):
Given that this is all behind the experimental `enableD1WithSessionsAPI` flag, this breaking change should be fine.

vy-ton:
@lambrospetrou lgtm

PR #3468, opened 2025-02-05
Fix uses of it's vs its and a few other spelling things.

erikcorry:


jasnell (approved):
Why was the doctor unhappy with the hospital's annual blood donation report?

Because it was full of Type O's.

anonrig (approved):
lol

jclee (approved):

PR #3470, opened 2025-02-05
Support .tar in Python's loadBundleFromR2.

dom96:
This PR improves decompressArrayBuffer so that it:

* detects uncompressed tarballs and doesn't attempt to decompress them again
* throws a good error when the filename points to an unsupported format

### DIFF
--- a/src/pyodide/internal/loadPackage.ts
+++ b/src/pyodide/internal/loadPackage.ts
@@ -53,23 +53,24 @@ function getFilenameOfPackage(requirement: string): string {
 async function loadBundleFromR2(requirement: string): Promise<Reader> {
   // first check if the disk cache has what we want
   const filename = getFilenameOfPackage(requirement);
-  const cached = DiskCache.get(filename);
-  if (cached) {
-    const decompressed = await decompressArrayBuffer(cached);
-    const reader = new ArrayBufferReader(decompressed);
-    return reader;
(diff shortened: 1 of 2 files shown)


src/pyodide/internal/loadPackage.ts
```diff
@@ -55,7 +65,7 @@ async function loadBundleFromR2(requirement: string): Promise<Reader> {
   const filename = getFilenameOfPackage(requirement);
   const cached = DiskCache.get(filename);
   if (cached) {
-    const decompressed = await decompressArrayBuffer(cached);
+    const decompressed = await decompressArrayBuffer(cached, filename);
```
hoodmane, line 68: Let's leave `decompressArrayBuffer` alone and do:
```js
async function loadBundleFromR2(requirement: string): Promise<Reader> {
  // first check if the disk cache has what we want
  const filename = getFilenameOfPackage(requirement);
  let original = DiskCache.get(filename);
  if (!original) {

    original = await response.arrayBuffer();
    DiskCache.put(filename, original);
  }

  const uncompressed = (filename.endsWith('.tar.gz')) ? await decompressArrayBuffer(original) : original;
  const reader = new ArrayBufferReader(uncompressed);
  return reader;
}
```

dom96: Done

hoodmane (approved):

PR #3471, opened 2025-02-05
Link doc for internal-build workflow

ketanhwr:


tools/cross/internal_build.py
```diff
@@ -61,6 +61,9 @@ def parse_args():
             status = resp.json()["status"]
             if status == "errored":
                 print("Build failed")
+                print(
+                    "For more information, please visit: https://cflare.co/workerd-internal-build"
```
danlapid, line 65: This is hidden behind cloudflare identity and links to an internal site, this cannot be our solution/documentation for a CI step on an open source repository.

danlapid: We need to make this Open Source friendly.

ketanhwr: This workflow is anyway restricted to internal members - I'm adding this log just so that CF folks who encounter this workflow for the first time get some sort of an explanation about it.

ketanhwr: IMO I doubt we'll ever turn this on for external contributors - there's no way they can debug internal builds regardless

danlapid: If we make it a required check then it's required for everyone. I don't see how you control who it's enabled for when it's required.

PR #3478, opened 2025-02-06
Don't embed pyodide-bucket into Python Worker.

dom96:
The pyodide-bucket was only used to get the package URL, but we can hardcode this in metadata.ts instead as it won't be changing.

### DIFF
--- a/src/pyodide/internal/metadata.ts
+++ b/src/pyodide/internal/metadata.ts
@@ -1,5 +1,4 @@
 import { default as MetadataReader } from 'pyodide-internal:runtime-generated/metadata';
-import { default as PYODIDE_BUCKET } from 'pyodide-internal:generated/pyodide-bucket.json';
 import { default as ArtifactBundler } from 'pyodide-internal:artifacts';
 
 export const IS_WORKERD = MetadataReader.isWorkerd();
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/metadata.ts
```diff
@@ -1,17 +1,25 @@
...
 export const PACKAGES_VERSION = MetadataReader.getPackagesVersion();
+export const WORKERD_INDEX_URL =
+  PACKAGES_VERSION == '20240829.4'
+    ? // TODO: The URL below points at a worker which redirects requests to the public R2 bucket URL
+      // at pub-45d734c4145d4285b343833ee450ef38.r2.dev. We should remove this worker and point at
+      // our prod bucket.
+      'https://pyodide-packages.runtime-playground.workers.dev/' +
+      PACKAGES_VERSION +
+      '/'
+    : 'https://python-packages.edgeworker.net/' + PACKAGES_VERSION + '/';
```
hoodmane, line 21: I think this is a bit neater:
```suggestion
// TODO: pyodide-packages.runtime-playground.workers.dev points at a worker which redirects requests to the public R2 bucket URL
// at pub-45d734c4145d4285b343833ee450ef38.r2.dev. We should remove this worker and point at
// our prod bucket.
export const WORKERD_INDEX_URL =
  PACKAGES_VERSION == '20240829.4'
    ? `https://pyodide-packages.runtime-playground.workers.dev/${PACKAGES_VERSION}/`
    : `https://python-packages.edgeworker.net/${PACKAGES_VERSION}/'`;
```

hoodmane (approved):
Looks good to me aside from a nit.

PR #3479, opened 2025-02-06
Fix issues loading stdlib packages on newer Python versions.

dom96:
This fixes two problems:

* the `install_dir` no longer being the source of truth of where a package belongs, now using `package_type`
* dependencies of stdlib packages not being loaded properly

### DIFF
--- a/src/pyodide/internal/loadPackage.ts
+++ b/src/pyodide/internal/loadPackage.ts
@@ -15,6 +15,7 @@ import {
   LOAD_WHEELS_FROM_R2,
   LOAD_WHEELS_FROM_ARTIFACT_BUNDLER,
   PACKAGES_VERSION,
+  USING_OLDEST_PACKAGES_VERSION,
 } from 'pyodide-internal:metadata';
 import {
   SITE_PACKAGES,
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/loadPackage.ts
```diff
@@ -141,10 +141,9 @@ async function loadPackagesImpl(
 }
 
 export async function loadPackages(Module: Module, requirements: Set<string>) {
```
hoodmane, line 143: We should add a comment warning that `loadPackages` does not resolve dependencies because the corresponding Pyodide function does resolve dependencies and this is super confusing to me. 

No action needed in this PR for any of the following:

Also, maybe in a followup we should rename this function to be more descriptive.

It's called after the runtime is set up but it isn't clear that it does anything that requires the runtime to be set up? It can't happen at top level because it needs access to the module, but it only needs at the very end when we want to mount the fs. Also, I guess it makes requests in workerd and so there can't happen at top level but can happen before any other part of the runtime is set up?

Anyways, it would be great to clean some of this up so that we can understand our code a bit better.

hoodmane: No action needed in this PR for any of the following:

Also, maybe in a followup we should rename this function to be more descriptive.

It's called after the runtime is set up but it isn't clear that it does anything that requires the runtime to be set up? It can't happen at top level because it needs access to the module, but it only needs at the very end when we want to mount the fs. Also, I guess it makes requests in workerd and so there can't happen at top level but can happen before any other part of the runtime is set up?

Anyways, it would be great to clean some of this up so that we can understand our code a bit better.

src/pyodide/internal/loadPackage.ts
```diff
@@ -140,8 +140,19 @@ async function loadPackagesImpl(
...
+ * do any dependency resolution, it just installs the requirements that are specified. See
+ * `getTransitiveRequirements` for the code that deals with this.
+ */
 export async function loadPackages(Module: Module, requirements: Set<string>) {
-  const pkgsToLoad = requirements.union(new Set(STDLIB_PACKAGES));
+  let pkgsToLoad = requirements;
+  // TODO: Package snapshot created with '20240829.4' needs the stdlib packages to be added here.
+  // We should remove this check once the next Python and packages versions are rolled
+  // out.
+  if (PACKAGES_VERSION == '20240829.4') {
```
hoodmane, line 153: Let's add to `metadata.ts`:
```
export const USING_OLDEST_PACKAGES_VERSION = PACKAGES_VERSION === '20240829.4';
```
and then use that in the three places we have this comparison.

PR #3480, opened 2025-02-06
Fixes to load dynlib on newer Python versions correctly.

dom96:
While working on https://github.com/cloudflare/workerd/pull/3451 I ran into issues with the newer Python version and package bundle. libssl is now expecting libcrypto and other dynlibs to be present in /usr/lib/ so we need to overlay the openssl tarball in that directory, this is what this PR implements.

### DIFF
--- a/src/pyodide/BUILD.bazel
+++ b/src/pyodide/BUILD.bazel
@@ -121,6 +121,14 @@ if (typeof FinalizationRegistry === "undefined") {
         unregister(){}
     };
 }
+
+function patchDynlibLookup(Module, libName) {
+    try {
+        return Module.FS.readFile("/usr/lib/" + libName);
+    } catch(e) {
+        console.error("Failed to read ", libName, e);
+    }
+}
 """
 
 REPLACEMENTS = [
@@ -168,6 +176,10 @@ REPLACEMENTS = [
     [
         "eval(UTF8ToString(ptr))",
         "(() => {throw new Error('Internal Emscripten code tried to eval, this should not happen, please file a bug report with your requirements.txt file\\'s contents')})()",
+    ],
+    [
+        "!libData&&flags.fs",
+        "!(libData ??= patchDynlibLookup(Module, libName))&&flags.fs",
     ],
 ]
 
--- a/src/pyodide/internal/loadPackage.ts
+++ b/src/pyodide/internal/loadPackage.ts
@@ -18,7 +18,8 @@ import {
   USING_OLDEST_PACKAGES_VERSION,
 } from 'pyodide-internal:metadata';
 import {
-  SITE_PACKAGES,
+  DYNLIB_PATH,
+  VIRTUALIZED_DIR,
   STDLIB_PACKAGES,
   getSitePackagesPath,
 } from 'pyodide-internal:setupPackages';
@@ -40,20 +41,20 @@ async function decompressArrayBuffer(
   }
 }
 
-function getFilenameOfPackage(requirement: string): string {
+function getPackageMetadata(requirement: string): PackageDeclaration {
   const obj = LOCKFILE['packages'][requirement];
   if (!obj) {
     throw new Error('Requirement ' + requirement + ' not found in lockfile');
   }
 
-  return obj.file_name;
+  return obj;
 }
 
 // loadBundleFromR2 loads the package from the internet (through fetch) and uses the DiskCache as
 // a backing store. This is only used in local dev.
 async function loadBundleFromR2(requirement: string): Promise<Reader> {
   // first check if the disk cache has what we want
-  const filename = getFilenameOfPackage(requirement);
+  const filename = getPackageMetadata(requirement).file_name;
   let original = DiskCache.get(filename);
   if (!original) {
     // we didn't find it in the disk cache, continue with original fetch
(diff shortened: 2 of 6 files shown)


src/pyodide/internal/loadPackage.ts
```diff
@@ -129,15 +133,21 @@ async function loadPackagesImpl(
...
   }
 
   console.log('Loaded ' + loading.join(', '));
 
   const tarFS = createTarFS(Module);
   const path = getSitePackagesPath(Module);
-  const info = SITE_PACKAGES.rootInfo;
-  Module.FS.mount(tarFS, { info }, path);
+  Module.FS.mount(tarFS, { info: SITE_PACKAGES.rootInfo }, path);
+  Module.FS.mount(tarFS, { info: SITE_PACKAGES.dynlibTarFs }, DYNLIB_PATH);
```
hoodmane, line 150: Maybe `SITE_PACKAGES` is misnamed now that it includes both `SITE_PACKAGES` and the system lib directory.

src/pyodide/internal/setupPackages.ts
```diff
@@ -31,26 +31,32 @@ export const STDLIB_PACKAGES: string[] = Object.values(LOCKFILE.packages)
 // `folder/file.txt` -> `["folder", "file.txt"]
 export type FilePath = string[];
 
+function createTarFs() {
```
hoodmane, line 34: I don't really appreciate the fact that we now have two functions called `createTarFs()`. Maybe this one could be `makeEmptyTarFSRoot()`? Or you could just use `structuredClone()` in the `SitePackagesDir` constructor, but maybe factoring it out is better.

src/pyodide/internal/setupPackages.ts
```diff
@@ -31,26 +31,32 @@ export const STDLIB_PACKAGES: string[] = Object.values(LOCKFILE.packages)
...
-      type: '5',
-      modtime: 0,
-      size: 0,
-      path: '',
-      name: '',
-      parts: [],
-      reader: null,
-    };
+    this.rootInfo = createTarFs();
+    this.dynlibTarFs = createTarFs();
```
hoodmane, line 59: Would be possible to do:
```suggestion
    this.dynlibTarFs = structuredClone(this.rootInfo);
```
and keep the inline definition of `rootInfo`. Not sure that it's better.

src/pyodide/internal/setupPackages.ts
```diff
@@ -61,33 +67,38 @@ class SitePackagesDir {
    * If a file or directory already exists, an error is thrown.
    * @param {TarInfo} overlayInfo The directory that is to be "copied" into site-packages
    */
-  mountOverlay(overlayInfo: TarFSInfo): void {
+  mountOverlay(overlayInfo: TarFSInfo, dir: InstallDir): void {
```
hoodmane, line 70: Can we make this a static method or top level function and directly pass `dest` as the argument?

src/pyodide/internal/setupPackages.ts
dom96: Oh I didn't even realise that a function with this name already existed. I'll name it something else.

src/pyodide/internal/setupPackages.ts
dom96: I guess we can, but I prefer this. Any reason you prefer a static method here?

src/pyodide/internal/loadPackage.ts
dom96: Done.

src/pyodide/BUILD.bazel
```diff
@@ -122,6 +122,14 @@ if (typeof FinalizationRegistry === "undefined") {
         unregister(){}
     };
 }
+
+function patchDynlibLookup(Module, libName) {
+    try {
+        return Module.FS.readFile("/usr/lib/" + libName);
+    } catch(e) {
+        console.log("Failed", e);
```
anonrig, line 130:
```suggestion
        console.error("Failed to patch dynlib lookup", e);
```

src/pyodide/internal/loadPackage.ts
```diff
@@ -63,6 +64,9 @@ async function loadBundleFromR2(requirement: string): Promise<Reader> {
   // we didn't find it in the disk cache, continue with original fetch
   const url = new URL(WORKERD_INDEX_URL + filename);
   const response = await fetch(url);
+  if (response.status != 200) {
+    throw new Error('Could not fetch package at url: ' + url);
```
anonrig, line 68: Maybe logging/throwing the status will help as well

src/pyodide/internal/setupPackages.ts
```diff
@@ -31,26 +31,33 @@ export const STDLIB_PACKAGES: string[] = Object.values(LOCKFILE.packages)
 // `folder/file.txt` -> `["folder", "file.txt"]
 export type FilePath = string[];
 
+function createTarFsInfo() {
```
anonrig, line 34: The return type seems to be missing here

src/pyodide/internal/setupPackages.ts
hoodmane: A method can access or modify anything in `this`, whereas a static method can't. So it's clearer to write static methods whenever it's possible.

dom96: That's true, but I think it's less clean to expose the fields that are owned by ~~SitePackagesDir~~`VirtualizedDir`. If anything, I would refactor to make this class own the mounting logic so that the `TarFsInfo` fields are not public.

hoodmane (approved):

PR #3481, opened 2025-02-06
feat: add Workflows to ValidationErrorReporter

LuisDuarte1:
Partially closes WOR-486

In order to have proper validation for Workflows I think we will need a similar setup as DOs: ValidationErrorReporter now has a `addWorkflowClass` that behaves similarily as `addActorClass`. (This will need a internal PR in order to work properly).

This will be useful in cases that a user exports a class but doesn't extend from `WorkflowEntrypoint` (meaning they won't get `ctx` and `env` and see weird errors).


a-robinson (approved):

a-robinson:
LGTM, but I'd assume this needs to wait for the corresponding internal change before merge.

LuisDuarte1:
Closed in favor of https://github.com/cloudflare/workerd/pull/3509 (I need it to not be from a fork so that I can check out it in EW)

PR #3482, opened 2025-02-06
Fix workflow if-check and add internal-build workflow doc link

ketanhwr:


tools/cross/internal_build.py
```diff
@@ -60,19 +60,27 @@ def parse_args():
...
+                )
+                print(
+                    "If you are an external contributor, please ask the auto-assigned"
+                    " reviewers to check the internal build failure."
+                )
                 sys.exit(1)
             elif status == "complete":
                 break
 
-            print("Waiting for build to finish..")
```
danlapid, line 68: Keep this line and add `python -u` to the command line to enable streaming.

PR #3483, opened 2025-02-06
sqlite: allow pragmas and statements needed to run the query optimizer

justin-mp:
https://www.sqlite.org/lang_analyze.html recommends a bunch of things one can do to optimize the performance of the query planner.  Most folks using Durable Objects will probably be best served by just calling `PRAGMA optimize` after making schema changes.

A couple of notes about this commit:

* `PRAGMA optimize` may call `ANALYZE` on all tables, including the internal `_cf_` tables, so we have to allow `ANALYZE` even on otherwise disallowed table names.

* Calls to `ANALYZE`, either directly or via `PRAGMA optimize`, cause the creation of the `sqlite_stat1` table.  This caused some noise in sql-test.js as I had to update tests that depend on (a) the particular set of tables in the database and (b) the particular size of the database.

### DIFF
--- a/src/workerd/util/sqlite.c++
+++ b/src/workerd/util/sqlite.c++
@@ -409,6 +409,7 @@ enum class PragmaSignature {
   BOOLEAN,
   OBJECT_NAME,
   OPTIONAL_OBJECT_NAME,
+  NULL_OR_NUMBER,
   NULL_NUMBER_OR_OBJECT_NAME
 };
 struct PragmaInfo {
@@ -438,7 +439,10 @@ static constexpr PragmaInfo ALLOWED_PRAGMAS[] = {{"data_version"_kj, PragmaSigna
   {"index_xinfo"_kj, PragmaSignature::OBJECT_NAME},
 
   // Takes an argument of table name/index name OR a max number of results, or nothing
-  {"quick_check"_kj, PragmaSignature::NULL_NUMBER_OR_OBJECT_NAME}};
+  {"quick_check"_kj, PragmaSignature::NULL_NUMBER_OR_OBJECT_NAME},
+
+  // Takes a number representing a bit mask or nothing to use the default mask.
+  {"optimize"_kj, PragmaSignature::NULL_OR_NUMBER}};
 
 }  // namespace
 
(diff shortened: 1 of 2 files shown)


src/workerd/util/sqlite.c++
```diff
@@ -899,13 +903,19 @@ bool SqliteDatabase::isAuthorized(int actionCode,
...
     case SQLITE_CREATE_VIEW:  /* View Name       NULL            */
     case SQLITE_DROP_VIEW:    /* View Name       NULL            */
     case SQLITE_REINDEX:      /* Index Name      NULL            */
       KJ_ASSERT(param2 == kj::none);
       return regulator.isAllowedName(KJ_ASSERT_NONNULL(param1));
 
+    case SQLITE_ANALYZE: /* Table Name      NULL            */
+      KJ_ASSERT(param2 == kj::none);
+      // We allow all names (including names where isAllowedName() would return false) because a
+      // SQLite ANALYZE statement with no parameters will analyze all tables, including otherwise
```
kentonv, line 915: If I understand correctly, `ANALYZE` doesn't return any information to the caller, it just makes notes for future query planning. Is that right? Therefore, it certainly doesn't reveal nor modify the contents of hidden tables, therefore there's no need to restrict it? Might be worth spelling that out in the comment here.

kentonv (commented):
I wonder if we should _automatically_ run `PRAGMA optimize` when DOs shut down? Similar to how we take care of enabling autovacuum automatically, etc.

src/workerd/util/sqlite.c++
justin-mp: Your understanding is correct and I've updated the comment.

One non-obvious thing is that the notes that `ANALYZE` makes are available in a table that users *can modify*.  This means that customers can make the query planner do suboptimal things.  It's likely that the worst thing a customer could do is waste CPU, but it is a rather large surface that we open up by allowing `ANALYZE`.

justin-mp:
> I wonder if we should _automatically_ run `PRAGMA optimize` when DOs shut down? Similar to how we take care of enabling autovacuum automatically, etc.

I do think we want either D1 or DO to automatically run `PRAGMA optimize` and I have an internal ticket filed to do so.  This was more of a get something out quick until we did the automatic version.

The pros of doing the automatic version now are (1) that we probably wouldn't need to allowlist `PRAGMA optimize` and the `ANALYZE` statement and (2) customers don't have to worry about doing this.

The downsides are: (1) more work to figure out when we can call `PRAGMA optimize` on shutdown (probably in the SqliteDatabase dtor, as long as we can still write to the DB when the dtor runs) and (2) the fact that the sqlite_stat1 table would appear in all DOs even if they don't use `ANALYZE`.

I'm happy to drop this in favor of the automatic approach if we think that's worth waiting for compared to getting something out sooner.

kentonv (commented):
I'd say it's your call.

I agree with everything you say. It feels cleaner to me if we just make  this happen automatically and never have to allowlist it, but that needs a bit more work. I don't know exactly who is asking for this or how urgent it is. So if you feel it's better to just allow it immediately for their sake, go for it.

Also we _could_ block writes to the sqlite_stat1 table if we're at all worried about users messing with it... but yeah it's just CPU time I guess, which is already policed directly.

kentonv (approved):
(Approved if you decide to proceed.)

Issue #3033, opened 2024-10-30
Feature proposal: Signal client disconnect using `Request.signal`


brettwillis:
As I understand it, when a client disconnects, the context/"thread" for a request is aborted as is. And if `ctx.waitUntil()` was called, then those promises are allowed some time to resolve.

When you are running async callbacks using `ctx.waitUntil(promise)`, there is no easy way to know that the client has disconnected.

Conceptually, if you responded with a `ReadableStream`, then [controller.enqueue(chunk)](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultController/enqueue) should throw or [`writer.write(chunk)`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStreamDefaultWriter/write) of the writable side should reject on the next write. This would require keeping the context alive with `ctx.waitUntil(writable.closed)` or something like that, otherwise the context would just die immediately anyway. However I've found that these don't seem to reliably throw/reject when the client disconnects. Perhaps that is a separate issue.

Much of what I'm describing above is from experience/testing or examining `workerd` rather than documentation, so feel free to correct me.

### Proposal

The [`Request.signal`](https://developer.mozilla.org/en-US/docs/Web/API/Request/signal) API is an existing standard API which reflects the `AbortSignal` passed when constructing the request. However the `signal` property of the incoming request currently does nothing.

I propose that the Workers runtime utilise this existing API by "aborting" the `signal` of the incoming request when a client disconnects. Any scripts that are interested in reacting when the client has disconnected can utilise this signal.

If an abort listener has been registered using `request.signal.addEventListener('abort', ...)` then the runtime will call those handler(s) when the client disconnects, before aborting the context. If the handler(s) synchronously call `ctx.waitUntil(promise)` then that will keep the context alive for a further period of time.

_Credit to uri2 [here](https://community.cloudflare.com/t/how-to-catch-handle-cancelled-requests/583797/3?u=arrowheadapps) for this inspiration of this idea, as other proposals I had entertained were new non-standard APIs such as `ctx.addEventListener('end', ...)` etc._

I'm happy to contribute if this proposal makes it through.

### Use cases

The use-cases I require this for are scenarios when the Worker is generating a long-running response stream on the fly, for example [Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events).

When the client disconnects from the stream, it is useful to react to perform logging, or do cleanup such as signalling user presence in a database.

[...]

PR #3488, opened 2025-02-06
Implement Request.signal to detect client disconnects

npaun:
How it works:
- An AbortController is added as a member of WorkerEntrypoint. 
- Its AbortSignal is passed through via ServiceWorkerGlobalScope and Request through to JS, so it can be accessed via `Request.signal`.
- If the client disconnects, and we've both not produced a response or thrown an exception, the AbortSignal is triggered.

### DIFF
--- a/src/workerd/api/basics.h
+++ b/src/workerd/api/basics.h
@@ -511,7 +511,7 @@ class EventTarget: public jsg::Object {
 // An implementation of the Web Platform Standard AbortSignal API
 class AbortSignal final: public EventTarget {
  public:
-  enum class Flag { NONE, NEVER_ABORTS };
+  enum class Flag { NONE, NEVER_ABORTS, IGNORE_FOR_SUBREQUESTS };
 
   AbortSignal(kj::Maybe<kj::Exception> exception = kj::none,
       jsg::Optional<jsg::JsRef<jsg::JsValue>> maybeReason = kj::none,
@@ -602,9 +602,14 @@ class AbortSignal final: public EventTarget {
     tracker.trackField("reason", reason);
   }
 
+  bool isIgnoredForSubrequests() const {
+    return flag == Flag::IGNORE_FOR_SUBREQUESTS;
+  }
+
  private:
   IoOwn<RefcountedCanceler> canceler;
   Flag flag;
+
   kj::Maybe<jsg::JsRef<jsg::JsValue>> reason;
   kj::Maybe<jsg::JsRef<jsg::JsValue>> onAbortHandler;
 
@@ -616,7 +621,9 @@ class AbortSignal final: public EventTarget {
 // An implementation of the Web Platform Standard AbortController API
 class AbortController final: public jsg::Object {
  public:
-  explicit AbortController(): signal(jsg::alloc<AbortSignal>()) {}
+  explicit AbortController(AbortSignal::Flag abortSignalFlag = AbortSignal::Flag::NONE)
+      : signal(jsg::alloc<AbortSignal>(
+            kj::none /* exception */, kj::none /* maybeReason */, abortSignalFlag)) {}
 
   static jsg::Ref<AbortController> constructor() {
     return jsg::alloc<AbortController>();
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -145,7 +145,8 @@ kj::Promise<DeferredProxy<void>> ServiceWorkerGlobalScope::request(kj::HttpMetho
     kj::HttpService::Response& response,
     kj::Maybe<kj::StringPtr> cfBlobJson,
     Worker::Lock& lock,
-    kj::Maybe<ExportedHandler&> exportedHandler) {
+    kj::Maybe<ExportedHandler&> exportedHandler,
+    kj::Maybe<jsg::Ref<AbortSignal>> abortSignal) {
   TRACE_EVENT("workerd", "ServiceWorkerGlobalScope::request()");
   // To construct a ReadableStream object, we're supposed to pass in an Own<AsyncInputStream>, so
   // that it can drop the reference whenever it gets GC'ed. But in this case the stream's lifetime
@@ -216,7 +217,17 @@ kj::Promise<DeferredProxy<void>> ServiceWorkerGlobalScope::request(kj::HttpMetho
 
   auto jsRequest = jsg::alloc<Request>(method, url, Request::Redirect::MANUAL, kj::mv(jsHeaders),
       jsg::alloc<Fetcher>(IoContext::NEXT_CLIENT_CHANNEL, Fetcher::RequiresHostAndProtocol::YES),
-      kj::none /** AbortSignal **/, kj::mv(cf), kj::mv(body));
+      /* signal */ kj::none, kj::mv(cf), kj::mv(body),
+      /* thisSignal */ kj::mv(abortSignal), Request::CacheMode::NONE);
+
+  // signal vs thisSignal
+  // --------------------
+  // The fetch spec definition of Request has a distinction between the
+  // "signal" (which is an optional AbortSignal passed in with the options), and "this' signal",
+  // which is an AbortSignal that is always available via the request.signal accessor.
+  //
+  // redirect
+  // --------
   // I set the redirect mode to manual here, so that by default scripts that just pass requests
   // through to a fetch() call will behave the same as scripts which don't call .respondWith(): if
   // the request results in a redirect, the visitor will see that redirect.
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -458,7 +458,8 @@ class ServiceWorkerGlobalScope: public WorkerGlobalScope {
       kj::HttpService::Response& response,
       kj::Maybe<kj::StringPtr> cfBlobJson,
       Worker::Lock& lock,
-      kj::Maybe<ExportedHandler&> exportedHandler);
+      kj::Maybe<ExportedHandler&> exportedHandler,
+      kj::Maybe<jsg::Ref<AbortSignal>> abortSignal);
   // TODO(cleanup): Factor out the shared code used between old-style event listeners vs. module
   //   exports and move that code somewhere more appropriate.
 
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1051,6 +1051,7 @@ jsg::Ref<Request> Request::constructor(
           // explicitly say `signal: null`, they must want to drop the signal that was on the
           // original request.
           signal = kj::mv(s);
+          initDict.signal = kj::none;
         }
 
         KJ_IF_SOME(newCf, initDict.cf) {
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -995,7 +995,7 @@ jsg::Ref<Request> Request::constructor(
       cacheMode = oldRequest->getCacheMode();
       redirect = oldRequest->getRedirectEnum();
       fetcher = oldRequest->getFetcher();
-      signal = oldRequest->getSignal();
+      signal = oldRequest->getThisSignal();
     }
   }
 
@@ -1093,7 +1093,7 @@ jsg::Ref<Request> Request::constructor(
         redirect = otherRequest->redirect;
         cacheMode = otherRequest->cacheMode;
         fetcher = otherRequest->getFetcher();
-        signal = otherRequest->getSignal();
+        signal = otherRequest->getThisSignal();
         headers = jsg::alloc<Headers>(*otherRequest->headers);
         cf = otherRequest->cf.deepClone(js);
         KJ_IF_SOME(b, otherRequest->getBody()) {
@@ -1112,7 +1112,8 @@ jsg::Ref<Request> Request::constructor(
 
   // TODO(conform): If `init` has a keepalive flag, pass it to the Body constructor.
   return jsg::alloc<Request>(method, url, redirect, KJ_ASSERT_NONNULL(kj::mv(headers)),
-      kj::mv(fetcher), kj::mv(signal), kj::mv(cf), kj::mv(body), cacheMode);
+      kj::mv(fetcher), kj::mv(signal), kj::mv(cf), kj::mv(body), /* thisSignal */ kj::none,
+      cacheMode);
 }
 
 jsg::Ref<Request> Request::clone(jsg::Lock& js) {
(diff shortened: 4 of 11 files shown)


src/workerd/io/worker-entrypoint.c++
```diff
@@ -332,6 +341,14 @@ kj::Promise<void> WorkerEntrypoint::request(kj::HttpMethod method,
       failOpenService = context.getSubrequestChannelNoChecks(
           IoContext::NEXT_CLIENT_CHANNEL, false, kj::mv(cfBlobJson));
     }
+
+    if (proxyTask == kj::none && !loggedExceptionEarlier) {
+      KJ_IF_SOME(ctrl, abortController) {
+        context.addWaitUntil(context.run(
```
npaun, line 347: TODO: I should avoid adding this task unless we know there is an event handler registered to receive the abort. I'm not quite sure where to do this though.

src/workerd/api/global-scope.c++
```diff
@@ -190,7 +191,8 @@ kj::Promise<DeferredProxy<void>> ServiceWorkerGlobalScope::request(kj::HttpMetho
 
   auto jsRequest = jsg::alloc<Request>(method, url, Request::Redirect::MANUAL, kj::mv(jsHeaders),
       jsg::alloc<Fetcher>(IoContext::NEXT_CLIENT_CHANNEL, Fetcher::RequiresHostAndProtocol::YES),
-      kj::none /** AbortSignal **/, kj::mv(cf), kj::mv(body));
+      /* signal */ kj::none, kj::mv(cf), kj::mv(body), Request::CacheMode::NONE,
+      kj::mv(abortSignal));
```
jasnell, line 195: Given that there's already an `AbortSignal` argument to the request constructor, some comments here explaining the additional argument would be helpful

src/workerd/api/http.h
```diff
@@ -805,19 +805,16 @@ class Request final: public Body {
   Request(kj::HttpMethod method, kj::StringPtr url, Redirect redirect,
           jsg::Ref<Headers> headers, kj::Maybe<jsg::Ref<Fetcher>> fetcher,
           kj::Maybe<jsg::Ref<AbortSignal>> signal, CfProperty&& cf,
-          kj::Maybe<Body::ExtractedBody> body, CacheMode cacheMode = CacheMode::NONE)
+          kj::Maybe<Body::ExtractedBody> body, CacheMode cacheMode = CacheMode::NONE, kj::Maybe<jsg::Ref<AbortSignal>> thisSignal = kj::none)
```
jasnell, line 808:
```suggestion
          kj::Maybe<Body::ExtractedBody> body, CacheMode cacheMode = CacheMode::NONE, 
          kj::Maybe<jsg::Ref<AbortSignal>> thisSignal = kj::none)
```

src/workerd/api/http.h
```diff
@@ -805,19 +805,16 @@ class Request final: public Body {
...
     : Body(kj::mv(body), *headers), method(method), url(kj::str(url)),
       redirect(redirect), headers(kj::mv(headers)), fetcher(kj::mv(fetcher)),
       cacheMode(cacheMode), cf(kj::mv(cf)) {
     KJ_IF_SOME(s, signal) {
-      // If the AbortSignal will never abort, assigning it to thisSignal instead ensures
-      // that the cancel machinery is not used but the request.signal accessor will still
-      // do the right thing.
-      if (s->getNeverAborts()) {
```
jasnell, line 816: This likely needs to remain for the uses on the client side?

src/workerd/api/http.h
npaun: I'm probably reading this wrong but that just looks like a clang-format thing?

src/workerd/io/worker-entrypoint.c++
jasnell: A mechanism for this does not currently exist but we can look at adding it.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -298,9 +300,16 @@ kj::Promise<void> WorkerEntrypoint::request(kj::HttpMethod method,
     TRACE_EVENT("workerd", "WorkerEntrypoint::request() run", PERFETTO_FLOW_FROM_POINTER(this));
     jsg::AsyncContextFrame::StorageScope traceScope = context.makeAsyncTraceScope(lock);
 
+    abortController = jsg::alloc<api::AbortController>();
+    kj::Maybe<jsg::Ref<api::AbortSignal>> signal;
+
+    KJ_IF_SOME(ctrl, abortController) {
+      signal = ctrl->getSignal();
+    }
```
kentonv, line 308: I think it's misleading to use `kj::Maybe` and `KJ_IF_SOME` in cases where the value can never actually be null -- it adds a lot of noise and cognitive overhead for the reader (and some redundant checks at runtime).

In this case you could avoid it by restructuring the code like this:

```suggestion
    jsg::Ref<api::AbortSignal> signal =
        abortController.emplace(jsg::alloc<api::AbortController>()).getSignal();
```

src/workerd/io/worker-entrypoint.c++
kentonv: Tricky part here is that you are trying to avoid taking the isolate lock, but the data structures you need to examine are protected by the lock. So we need to add some other mechanism where when an event handler is first added, we set some atomic boolean true, and you can check that atomic boolean here without the lock.

src/workerd/api/http.h
kentonv: Yes. clang-format should have wrapped this line since it's over 100 chars long. But I see the lint check passed, so that's weird. Is clang-format failing to wrap here for some reason?

src/workerd/io/worker-entrypoint.c++
```diff
@@ -332,6 +341,14 @@ kj::Promise<void> WorkerEntrypoint::request(kj::HttpMethod method,
       failOpenService = context.getSubrequestChannelNoChecks(
           IoContext::NEXT_CLIENT_CHANNEL, false, kj::mv(cfBlobJson));
     }
+
+    if (proxyTask == kj::none && !loggedExceptionEarlier) {
```
kentonv, line 345: Probably worth a comment to say what we're checking here: We only want to run this logic if the request was canceled from the client side, not if it completed successfully or failed with an exception.

src/workerd/io/worker-entrypoint.c++
kentonv: Hmm, given that we only run this code when the request is actually canceled (not in the common case of completing successfully), maybe it's OK to just always take the lock after all. I was worried earlier that we'd be adding an extra lock on _every_ request, but that's not actually the case.

src/workerd/api/global-scope.h
```diff
@@ -442,7 +442,8 @@ class ServiceWorkerGlobalScope: public WorkerGlobalScope {
       kj::HttpService::Response& response,
       kj::Maybe<kj::StringPtr> cfBlobJson,
       Worker::Lock& lock,
-      kj::Maybe<ExportedHandler&> exportedHandler);
+      kj::Maybe<ExportedHandler&> exportedHandler,
+      kj::Maybe<jsg::Ref<AbortSignal>> abortSignal = kj::none);
```
kentonv, line 446: I think the call site in worker-entrypoint is the _only_ call site for this method? In that case, there's no need to supply a default for this new argument, and in fact it doesn't even need to be `Maybe`.

If there are other call sites, it's probably worth updating them to pass `kj::none` explicitly -- this provides a chance to review whether maybe one or more of them should actually be passing an AbortSignal.

kentonv:
In #3033 I made an observation which I don't think has been covered here.

> One gotcha regarding `request.signal` that we'll need to work out: What happens in the common case that the request is passed along to an outgoing `fetch()`? If the response from that fetch is immediately returned by the worker, then it makes sense that if the client disconnects, the outgoing fetch should be canceled. But what if the response of that fetch is going to be used in some other logic, maybe involving a `waitUntil()` that writes to cache, etc.? Then, if we cancel the subrequest based on the client disconnecting, we might break that logic.
>
> I think we might have to say: If the `signal` on an outgoing `fetch()` is the exact same `signal` from the original incoming request that started the Worker, it is ignored. If that feels wrong, then we would have to introduce a compat flag to fix it without breaking anyone. But I think it's fine really... if the client disconnects and there are no `waitUntil()`s, then all subrequests are canceled implicitly. So this only makes a difference if there are `waitUntil()`s, in which case you probably don't actually want to cancel the subrequest?

src/workerd/io/worker-entrypoint.c++
npaun: I turned this into just a TODO comment for now

src/workerd/api/http.h
npaun: I changed this line for other reasons and it's now under the length limit.

src/workerd/api/global-scope.h
npaun: There was one other call site (TestFixture::runRequest). I also made the parameter mandatory in the Request class too which had only a handful of callsites too.

src/workerd/api/http.h
npaun: Reverted this change.

src/workerd/api/tests/request-client-disconnect.js
```diff
@@ -0,0 +1,126 @@
+import { DurableObject, WorkerEntrypoint } from 'cloudflare:workers';
+import assert from 'node:assert';
+export class AbortTracker extends DurableObject {
+  async getAborted(key) {
+    return this.ctx.storage.get(key);
+  }
+  async setAborted(key, value) {
+    this.ctx.storage.put(key, value);
```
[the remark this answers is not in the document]
npaun: Yeah... oops.

kentonv:
Still need to address the passthrough issue? (The big blockquote in my earlier comment.)

npaun:
Yes, still to address:

- Passthrough
- ASan issue

src/workerd/api/tests/request-client-disconnect.js
```diff
@@ -0,0 +1,128 @@
+import { DurableObject, WorkerEntrypoint } from 'cloudflare:workers';
+import assert from 'node:assert';
+import { scheduler } from 'node:timers/promises';
```
kentonv, line 3: FWIW, `scheduler.wait` is actually available without node compat.

npaun:
The latest push takes care of the passthrough subrequest case.

src/workerd/api/http.c++
```diff
@@ -1121,8 +1122,13 @@ jsg::Ref<Request> Request::clone(jsg::Lock& js) {
   auto cfClone = cf.deepClone(js);
   auto bodyClone = Body::clone(js);
 
-  return jsg::alloc<Request>(method, url, redirect, kj::mv(headersClone), getFetcher(), getSignal(),
-      kj::mv(cfClone), kj::mv(bodyClone));
+  return jsg::alloc<Request>(method, url, redirect, kj::mv(headersClone), getFetcher(),
+      /* signal */ getThisSignal(), kj::mv(cfClone), kj::mv(bodyClone), /* thisSignal */ kj::none);
```
npaun, line 1126: I'm treating `req.clone()` and `new Request(req)` differently in that former always clones the associated signal, while the latter will respect `ignoreForSubrequests` and ignore the signal of the incoming fetch. Let me know if that doesn't sound right.

src/workerd/api/tests/request-client-disconnect.js
npaun: VSCode just helpfully keeps importing it. I'll get rid of it before I merge.

src/workerd/api/http.c++
kentonv: Hmm, this is not the solution I was expecting. I thought that the implementation of `fetch()` would be modified to ignore the signal if it came from the original top-level request. Why should cloning the request (whether via `.clone()` or the constructor) drop the signal?

kentonv: Actually if someone just does `fetch(originalRequest)`, the request is not cloned at all, so doesn't this not even solve that case?

npaun: It does work because fetch calls the Request constructor again, but yeah this isn't the right way to accomplish this. I fixed it.

npaun:
...

src/workerd/api/basics.h
```diff
@@ -602,9 +602,21 @@ class AbortSignal final: public EventTarget {
     tracker.trackField("reason", reason);
   }
 
+  bool isIgnoredForSubrequests() {
```
jasnell, line 605:
```suggestion
  bool isIgnoredForSubrequests() const {
```

src/workerd/io/worker-entrypoint.c++
```diff
@@ -324,10 +326,13 @@ kj::Promise<void> WorkerEntrypoint::request(kj::HttpMethod method,
     TRACE_EVENT_END("workerd", PERFETTO_TRACK_FROM_POINTER(&context));
     TRACE_EVENT("workerd", "WorkerEntrypoint::request() run", PERFETTO_FLOW_FROM_POINTER(this));
     jsg::AsyncContextFrame::StorageScope traceScope = context.makeAsyncTraceScope(lock);
+    jsg::Ref<api::AbortSignal> signal =
+        abortController.emplace(jsg::alloc<api::AbortController>())->getSignal();
+    signal->setIgnoredForSubrequests();
```
jasnell, line 331: I'd generally prefer if this were set as an argument via the `api::AbortController` constructor as there's really not much reason to change it once set.

src/workerd/api/http.h
```diff
@@ -777,20 +777,25 @@ class Request final: public Body {
...
           kj::Maybe<jsg::Ref<AbortSignal>> signal, CfProperty&& cf,
-          kj::Maybe<Body::ExtractedBody> body, CacheMode cacheMode = CacheMode::NONE)
+          kj::Maybe<Body::ExtractedBody> body, kj::Maybe<jsg::Ref<AbortSignal>> thisSignal,
+          CacheMode cacheMode = CacheMode::NONE)
     : Body(kj::mv(body), *headers), method(method), url(kj::str(url)),
       redirect(redirect), headers(kj::mv(headers)), fetcher(kj::mv(fetcher)),
       cacheMode(cacheMode), cf(kj::mv(cf)) {
     KJ_IF_SOME(s, signal) {
-      // If the AbortSignal will never abort, assigning it to thisSignal instead ensures
+     // If the AbortSignal will never abort, assigning it to thisSignal instead ensures
```
jasnell, line 786:
```suggestion
      // If the AbortSignal will never abort, assigning it to thisSignal instead ensures
```

src/workerd/api/http.h
```diff
@@ -843,7 +848,10 @@ class Request final: public Body {
   // used on this request.
   kj::Maybe<jsg::Ref<AbortSignal>> getSignal();
 
-  jsg::Ref<AbortSignal> getThisSignal(jsg::Lock& js);
+  jsg::Ref<AbortSignal> getThisSignal();
+
+  void clearSignalIfIgnoredForSubrequest();
```
jasnell, line 853: A comment explaining this here would be helpful.

src/workerd/api/tests/request-client-disconnect.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  async hang() {
+    for (;;) {
+      await scheduler.wait(86400);
+    }
+  }
+
+  async hangAfterSendingSomeData() {
+    const { readable, writable } = new TransformStream();
```
jasnell, line 49: Will be slightly more efficient... yes I know it's just a test but still

```suggestion
    const { readable, writable } = new IdentityTransformStream();
```

jasnell (approved):
Overall LGTM but let's also have @kentonv review. Given the complexity of the change it's easy to miss something in this

kentonv:
Sorry, I am not likely to have time to thoroughly review this but I see the ignore-for-subrequests behavior is implemented which was my only concern before. I do think this behavior should be controlled by a compat flag as it's a bit weird and only done for compatibility reasons, but the compat flag could be added in a future PR. Feel free to merge based on James' review. Sorry for not getting to this earlier.

PR #3492, opened 2025-02-07
Enable use_plus_in_repo_names for common options

ujohnny:
Bazel 7 uses ~ in external repo names and it leads to various issues like bazelbuild/bazel#23127 or some tools consider such paths as ignored because tilde could mean a backup copy of the file. Bazel 8 addresses these issues with use of + by default, but Bazel 7 must be supplied with an incompatible argument to enable this behavior

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,6 +13,10 @@ build --test_timeout=1,15,60,240
 # Not using bzlmod for dependencies yet
 common --noenable_bzlmod
 
+# Use plus in repo names for Bazel 7 as long as it's used to address some resolution issues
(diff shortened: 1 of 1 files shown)


ujohnny:

fhanau (approved):
I've personally never encountered difficulties with this (as I understand it mostly matters on Windows?), but while the Bazel 8 update is blocked I can see some developers having issues with this – LGTM

fhanau:
Cherry-picked the change into https://github.com/cloudflare/workerd/pull/3627 and merged that – CI was malfunctioning here for some reason. Thank you for your contribution!

PR #3495, opened 2025-02-07
Python workers: Better error handling when the handler is not defined

hoodmane:


src/pyodide/python-entrypoint-helper.ts
```diff
@@ -131,9 +131,14 @@ function makeHandler(pyHandlerName: string): Handler {
         'prep_python',
         async () => await preparePython()
       );
-
+      const handler = mainModule[pyHandlerName];
+      if (!handler) {
+        throw new Error(
+          `Python entrypoint "${MAIN_MODULE_NAME}" does not export a handler named "${pyHandlerName}"`
```
hoodmane, line 137: We should check what the corresponding error for JS workers is and update this to match.

dom96 (approved):
Note that there is some logic for this in the validator, so gotta make sure this doesn't impact that. Our internal tests should hopefully cover that.

dom96:
Actually looking at python-validation-test I don't see a test case for when `on_fetch` is missing, might be worth adding it as part of this change.

PR #3498, opened 2025-02-07
Add some echos to the modified wd_test

hoodmane:
Helps with debugging to have these in the log


fhanau (approved):
LGTM (I assume that these are not logged when running bazel test the regular way, only when using --test_output=streamed or similar, while always being available in the test logs.)

PR #3499, opened 2025-02-07
Make sure we don't import pyodide into memory snapshot.

hoodmane:
Importing pyodide before `finalizeBootstrap()` is called causes mayhem.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -22,6 +22,7 @@ select = [
 ]
 ignore = [
   "E402",  # module import not at top of file
+  "E501", # line too long
   "PLR0912", # too many branches
   "PLR0915", # too many statements
   "PLR2004", # Magic value used in comparison
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -335,31 +335,43 @@ kj::String importToModuleFilename(kj::StringPtr pkgImport) {
 }
 
 kj::Array<kj::String> ArtifactBundler::filterPythonScriptImports(
-    kj::HashSet<kj::String> locals, kj::Array<kj::String> imports) {
-  kj::HashSet<kj::StringPtr> snapshotImportsSet;
+    kj::HashSet<kj::String> workerModules, kj::Array<kj::String> imports) {
+  auto baselineSnapshotImportsSet = kj::HashSet<kj::StringPtr>();
   for (auto& pkgImport: ArtifactBundler::getSnapshotImports()) {
-    snapshotImportsSet.insert(kj::mv(pkgImport));
+    baselineSnapshotImportsSet.insert(kj::mv(pkgImport));
   }
+  // baselineSnapshotImportsSet.insertAll(kj::mv(ArtifactBundler::getSnapshotImports()));
 
-  kj::HashSet<kj::String> filteredImports;
+  kj::HashSet<kj::String> filteredImportsSet;
+  filteredImportsSet.reserve(imports.size());
   for (auto& pkgImport: imports) {
-    if (filteredImports.contains(pkgImport)) [[unlikely]] {
-      continue;  // Skip duplicates.
+    // Skip duplicates
+    if (filteredImportsSet.contains(pkgImport)) [[unlikely]] {
+      continue;
     }
 
-    auto moduleFilename = importToModuleFilename(pkgImport);
-    if (!locals.contains(moduleFilename) && pkgImport != "js" &&
-        !snapshotImportsSet.contains(pkgImport)) {
-      filteredImports.insert(kj::mv(pkgImport));
+    // don't include js or pyodide.
+    if (pkgImport == "js" || pkgImport == "pyodide") {
+      continue;
     }
+
+    // Don't include anything that went into the baseline snapshot
+    if (baselineSnapshotImportsSet.contains(pkgImport)) {
+      continue;
+    }
+
+    // Don't include imports from worker files
+    if (workerModules.contains(importToModuleFilename(pkgImport))) {
+      continue;
+    }
+    filteredImportsSet.insert(kj::mv(pkgImport));
   }
 
-  kj::Vector<kj::String> filteredImportsVec;
-  for (auto& pkgImport: filteredImports) {
-    filteredImportsVec.add(kj::mv(pkgImport));
+  auto filteredImportsBuilder = kj::heapArrayBuilder<kj::String>(filteredImportsSet.size());
+  for (auto& pkgImport: filteredImportsSet) {
+    filteredImportsBuilder.add(kj::mv(pkgImport));
   }
-
-  return filteredImportsVec.releaseAsArray();
+  return filteredImportsBuilder.finish();
 }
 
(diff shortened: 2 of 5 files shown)


src/workerd/server/tests/python/py_wd_test.bzl
```diff
@@ -83,6 +83,6 @@ def py_wd_test(
             snapshot = make_snapshot,
             data = data,
             args = args,
-            size = size,
-            tags = tags,
+            # size = size,
```
danlapid, line 86: Seems unintentional

hoodmane: Indeed.

src/workerd/server/tests/python/dont-snapshot-pyodide/worker.py
```diff
@@ -0,0 +1,2 @@
+async def test():
```
danlapid, line 1: How does this test the issue?

src/pyodide/internal/snapshot.ts
```diff
@@ -264,8 +264,13 @@ function memorySnapshotDoImports(Module: Module): string[] {
         MetadataReader.getWorkerFiles('py')
       )
     );
-
-  const deduplicatedModules = [...new Set(importedModules)];
+  const moduleSet = new Set(importedModules);
+  // Importing pyodide before `finalizeBootstrap()` is called causes mayhem
+  // Make sure we don't do that!
+  // TODO: in Pyodide 0.27 we can call finalizeBootstrap() before taking the snapshot. Doing so will
```
danlapid, line 270: Perhaps we need a way to check pyodide version when generating the bundles so we could make things like this conditional

hoodmane: Indeed we do.

danlapid:
LGTM, only problem is I don’t understand the test

src/workerd/server/tests/python/dont-snapshot-pyodide/worker.py
hoodmane: To trigger the bug we need to do two things:
1. import `pyodide` at top level
2. ensure that there is some package requirement

The `numpy` import here is unnecessary, but if someone removed `(name = "numpy", pythonRequirement = "")` from the `wd-test` file it would no longer reproduce the problem.

hoodmane: Oh shoot ruff removed the imports. I keep having this problem.

danlapid: But where is the pyodide import? 😅

danlapid: > Oh shoot ruff removed the imports. I keep having this problem.

Hahaha that explains it
There’s probably some comments that disable ruff on a file right?

hoodmane: I put a whole explanation in now. Also #3415 is such a huge improvement to our test suite I am so happy we merged it.

danlapid (approved):

src/pyodide/internal/snapshot.ts
```diff
@@ -264,8 +264,13 @@ function memorySnapshotDoImports(Module: Module): string[] {
...
+  // remove the need for this.
+  moduleSet.delete('pyodide');
```
dom96, line 272: We already have some filtering logic in https://github.com/cloudflare/workerd/blob/07cb474032c16c09ca8750afcec07414b00dc44e/src/workerd/api/pyodide/pyodide.c%2B%2B#L351

Seems to only be filtering out `js` directly right now, but best to do the filtering in one place.

src/pyodide/internal/snapshot.ts
dom96, line 270: Since this is only necessary for 0.26, should we gate this behind a version check?

dom96 (approved):
Are there any other imports that might cause issues that should be filtered out? Should we add a test for other Pyodide modules that can be imported?

src/pyodide/internal/snapshot.ts
hoodmane: It's necessary for both until we move `finalizeBootstrap()`.

hoodmane:
There's nothing else that is publicly allowed to import. There are a few implementation detail imports that could cause trouble but I think that'd be a won't-fix since people really shouldn't touch that stuff.

hoodmane:
Doing the filtering in `pyodide.c++` also avoids any need to rebuild JS bundles.

.ruff.toml
```diff
@@ -22,6 +22,7 @@ select = [
 ]
 ignore = [
   "E402",  # module import not at top of file
+  "E501", # line too long
```
danlapid, line 25: We shouldn’t be ignoring this globally, we can maybe remove this for a specific line but I think we should revert this change @hoodmane

src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -66,3 +66,12 @@ py_wd_test(
         "0.27.1",
     ],
 )
+
+py_wd_test(
+    "dont-snapshot-pyodide",
+    # TODO: Requires a new bundle deploy
+    skip_python_flags = [
+        "0.26.0a2",
```
danlapid, line 74: Does this still require a new bundle? Seems like by moving the logic to pyodide.c++ you can enable these variants

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -335,31 +335,43 @@ kj::String importToModuleFilename(kj::StringPtr pkgImport) {
...
 kj::Array<kj::String> ArtifactBundler::filterPythonScriptImports(
-    kj::HashSet<kj::String> locals, kj::Array<kj::String> imports) {
-  kj::HashSet<kj::StringPtr> snapshotImportsSet;
+    kj::HashSet<kj::String> workerModules, kj::Array<kj::String> imports) {
+  auto baselineSnapshotImportsSet = kj::HashSet<kj::StringPtr>();
   for (auto& pkgImport: ArtifactBundler::getSnapshotImports()) {
-    snapshotImportsSet.insert(kj::mv(pkgImport));
+    baselineSnapshotImportsSet.insert(kj::mv(pkgImport));
   }
+  // baselineSnapshotImportsSet.insertAll(kj::mv(ArtifactBundler::getSnapshotImports()));
```
danlapid, line 343: Vestigial comment

danlapid:
If the PR’s implementation changes please ask for a re-review instead of merging with stale approvals

.ruff.toml
hoodmane: There's a separate `ruff fmt` that formats code. If the formatter makes a long line I think it's very frustrating for `ruff`'s linter to then complain about it.

hoodmane:
Sorry, I wasn't paying enough attention and forgot this had unreviewed code when I merged it.

src/workerd/server/tests/python/BUILD.bazel
hoodmane: No.

PR #3502, opened 2025-02-08
Cleanup some stuff accidentally commited in #3499

hoodmane:
cc @danlapid. If you want I can turn the ruff "line too long" lint back on, but it is common to disable this because ruff also comes with a formatter that itself wraps long lines and does not always agree with the ruff linter about when a line should be wrapped.


danlapid (approved):
Thanks!

PR #3503, opened 2025-02-09
rust-python-parser tidy up

hoodmane:
1. Tests move into a different compilation unit from logic
2. Put the `parseImports` function into the same statib library as the rust objects.

### DIFF
--- a/src/rust/python-parser/BUILD
+++ b/src/rust/python-parser/BUILD
@@ -1,18 +1,35 @@
 load("//:build/kj_test.bzl", "kj_test")
+load("//:build/wd_cc_library.bzl", "wd_cc_library")
 load("//:build/wd_rust_crate.bzl", "rust_cxx_include", "wd_rust_crate")
+
+wd_cc_library(
+    name = "cxx-bridge",
+    srcs = [
+        "cxx-bridge.c++",
+    ],
+    hdrs = [
+        "cxx-bridge.h",
+    ],
+    linkstatic = True,
+    deps = [
+        "//src/rust/cxx-integration",
+        "@capnp-cpp//src/kj",
+    ],
+)
 
 wd_rust_crate(
     name = "python-parser",
     cxx_bridge_src = "lib.rs",
     visibility = ["//visibility:public"],
     deps = [
+        ":cxx-bridge",
         "@crates_vendor//:ruff_python_ast",
         "@crates_vendor//:ruff_python_parser",
     ],
 )
 
 kj_test(
-    src = "import_parsing.c++",
+    src = "import-parsing-test.c++",
     deps = [
         ":python-parser",
         "//deps/rust:runtime",
--- a/src/rust/python-parser/cxx-bridge.c++
+++ b/src/rust/python-parser/cxx-bridge.c++
@@ -1,0 +1,26 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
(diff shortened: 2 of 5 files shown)


src/rust/python-parser/cxx-bridge.c++
```diff
@@ -0,0 +1,28 @@
...
+
+// #include <workerd/rust/python-parser/lib.rs.h>
+#include "workerd/rust/cxx-integration/lib.rs.h"
+
+#include <kj/test.h>
+
+namespace workerd {
+namespace rust {
+namespace python_parser {
+  ::rust::Vec<::rust::String> get_imports(::rust::Slice<::rust::Str const> sources);
```
hoodmane, line 13: I don't love declaring this manually, but at least C++ name mangling means there is a very high probability of a link error rather than generating a bad executable if the declaration is wrong. The problem is this is declared in `workerd/rust/python-parser/lib.rs.h` which is generated by `wd_rust_crate` but that is a dependent of the rule that compiles `cxx-bridge.c++`. This probably means I'm doing something wrong in Bazel? cc @mikea

src/rust/python-parser/import_parsing.h
```diff
@@ -0,0 +1,5 @@
+#include "kj/string.h"
```
danlapid, line 1: convention is to call this file `cxx-bridge.h`

src/rust/python-parser/BUILD
```diff
@@ -1,18 +1,36 @@
...
+    srcs = [
+        "cxx-bridge.c++",
+    ],
+    hdrs = [
+        "import_parsing.h",
+    ],
+    linkstatic = True,
+    deps = [
+        "//src/rust/cxx-integration",
+        "//src/workerd/jsg",
```
danlapid, line 16: why is jsg required here?

src/rust/python-parser/test_import_parsing.c++
```diff
@@ -1,25 +1,7 @@
-// Copyright (c) 2017-2022 Cloudflare, Inc.
-// Licensed under the Apache 2.0 license found in the LICENSE file or at:
-//     https://opensource.org/licenses/Apache-2.0
+#include "import_parsing.h"
```
danlapid, line 1: test file name conventions are <module>-test.c++
So this should be `python-parser-test.c++`

src/rust/python-parser/cxx-bridge.c++
fhanau: Try using the cxx_bridge_src bazel parameter to set this up, like in src/rust/cxx-integration/.

src/rust/python-parser/cxx-bridge.c++
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+// #include <workerd/rust/python-parser/lib.rs.h>
+#include "workerd/rust/cxx-integration/lib.rs.h"
+
+#include <kj/test.h>
```
fhanau, line 8: Nit: kj/test.h appears unused here, the commented include above can be dropped too

src/rust/python-parser/cxx-bridge.h
```diff
@@ -0,0 +1,5 @@
+#include "kj/string.h"
```
fhanau, line 1: Nit: In KJ-style, includes to paths that depend on include directives (i.e. are not a relative path from the current directory) should use angled brackets. (As opposed to Google style, where using quotes is preferred).

PR #3505, opened 2025-02-10
Python: also call `memorySnapshotDoImports` in local development

hoodmane:
There are several bugs in `memorySnapshotDoImports` that need to be fixed and tricky behaviors that need to be worked out. This doesn't fix any of the problems but at least it will surface them to users in local development and not when they deploy their worker.

### DIFF
--- a/src/pyodide/internal/snapshot.ts
+++ b/src/pyodide/internal/snapshot.ts
@@ -295,18 +295,12 @@ function checkLoadedSoFiles(dsoJSON: DylinkInfo): void {
  * are initialized in the linear memory snapshot and then saving a copy of the
  * linear memory into MEMORY.
  */
-function makeLinearMemorySnapshot(
-  Module: Module
-): ArtifactBundler.MemorySnapshotResult {
-  const importedModulesList = memorySnapshotDoImports(Module);
+function makeLinearMemorySnapshot(Module: Module): Uint8Array {
   const dsoJSON = recordDsoHandles(Module);
   if (IS_CREATING_BASELINE_SNAPSHOT) {
     // checkLoadedSoFiles(dsoJSON);
   }
-  return {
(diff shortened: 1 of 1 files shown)


src/pyodide/internal/snapshot.ts
```diff
@@ -296,9 +296,8 @@ function checkLoadedSoFiles(dsoJSON: DylinkInfo): void {
  * linear memory into MEMORY.
  */
 function makeLinearMemorySnapshot(
```
danlapid, line 298: No reason to take importedModulesList as an argument only to return it in the result but then ignore that result object 🤔

hoodmane: What a mess.

src/pyodide/internal/snapshot.ts
```diff
@@ -296,9 +296,8 @@ function checkLoadedSoFiles(dsoJSON: DylinkInfo): void {
...
-  Module: Module
+  Module: Module, importedModulesList: string[]
 ): ArtifactBundler.MemorySnapshotResult {
-  const importedModulesList = memorySnapshotDoImports(Module);
   const dsoJSON = recordDsoHandles(Module);
   if (IS_CREATING_BASELINE_SNAPSHOT) {
```
danlapid, line 302: Is this `if` and `checkLoadedSoFiles` deprecated? should we remove them?

danlapid (approved):

src/pyodide/internal/snapshot.ts
hoodmane: Yes I think we should remove them. I originally wrote `checkLoadedSoFiles` when we didn't have proper test coverage, but now any relevant problems would cause a test failure even without the added assertions.

hoodmane: I will delete this in a separate cleanup PR after this.

PR #3507, opened 2025-02-10
Test filterPythonScriptImports and fix some problems

hoodmane:
For most of these checks, we should only be comparing on the first component of a dotted import name. We also need to handle folders with `__init__.py` which are also modules.

This fixes the test failure in #3505.

We could consider removing the `dont-snapshot-pyodide` test, since this gives us unit testing coverage of that logic. Or perhaps we should just combine `filter-non-py-files` and `dont-snapshot-pyodide` into a single integration test covering this file.

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -317,40 +317,60 @@ kj::Array<kj::String> ArtifactBundler::parsePythonScriptImports(kj::Array<kj::St
   return result.releaseAsArray();
 }
 
-// This is equivalent to `pkgImport.replace('.', '/') + ".py"`.
-kj::String importToModuleFilename(kj::StringPtr pkgImport) {
-  auto result = kj::heapString(pkgImport.size() + 3);
-  for (auto i = 0; i < pkgImport.size(); i++) {
-    if (pkgImport[i] == '.') {
-      result[i] = '/';
-    } else {
-      result[i] = pkgImport[i];
-    }
-  }
+const kj::Array<kj::StringPtr> snapshotImports = kj::arr("_pyodide"_kj,
+    "_pyodide.docstring"_kj,
+    "_pyodide._core_docs"_kj,
+    "traceback"_kj,
+    "collections.abc"_kj,
+    // Asyncio is the really slow one here. In native Python on my machine, `import asyncio` takes ~50
+    // ms.
+    "asyncio"_kj,
+    "inspect"_kj,
+    "tarfile"_kj,
+    "importlib",
+    "importlib.metadata"_kj,
+    "re"_kj,
+    "shutil"_kj,
+    "sysconfig"_kj,
+    "importlib.machinery"_kj,
+    "pathlib"_kj,
+    "site"_kj,
+    "tempfile"_kj,
+    "typing"_kj,
+    "zipfile"_kj);
 
-  result[pkgImport.size()] = '.';
-  result[pkgImport.size() + 1] = 'p';
-  result[pkgImport.size() + 2] = 'y';
-  return result;
+kj::Array<kj::StringPtr> ArtifactBundler::getSnapshotImports() {
+  return kj::heapArray(snapshotImports.begin(), snapshotImports.size());
+}
+
+bool isWorkerModuleImport(
+    kj::HashSet<kj::String>& workerModules, kj::ArrayPtr<const char> firstComponent) {
+  // TODO: handling for:
+  // 1. namespace packages
+  // 2. shared libraries
+  return workerModules.contains(kj::str(firstComponent, ".py")) ||
+      workerModules.contains(kj::str(firstComponent, "/__init__.py"));
 }
 
 kj::Array<kj::String> ArtifactBundler::filterPythonScriptImports(
-    kj::HashSet<kj::String> workerModules, kj::Array<kj::String> imports) {
+    kj::HashSet<kj::String> workerModules, kj::ArrayPtr<kj::String> imports) {
   auto baselineSnapshotImportsSet = kj::HashSet<kj::StringPtr>();
-  for (auto& pkgImport: ArtifactBundler::getSnapshotImports()) {
+  for (auto& pkgImport: snapshotImports) {
     baselineSnapshotImportsSet.insert(kj::mv(pkgImport));
   }
 
   kj::HashSet<kj::String> filteredImportsSet;
   filteredImportsSet.reserve(imports.size());
   for (auto& pkgImport: imports) {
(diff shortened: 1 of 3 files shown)


src/workerd/api/pyodide/pyodide.c++
```diff
@@ -374,22 +372,22 @@ kj::Array<kj::String> ArtifactBundler::filterPythonScriptImports(
...
+    kj::Array<kj::String> workerModules, kj::Array<kj::String> imports) {
+  kj::HashSet<kj::String> workerModulesSet;
+  for (auto& workerModule: workerModules) {
+    workerModulesSet.insert(kj::mv(workerModule));
   }
-  return ArtifactBundler::filterPythonScriptImports(kj::mv(localsSet), kj::mv(imports));
+  return filterPythonScriptImports(kj::mv(workerModulesSet), kj::mv(imports));
 }
 
 kj::Array<kj::StringPtr> ArtifactBundler::getSnapshotImports() {
```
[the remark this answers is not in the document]
hoodmane: And this one has a static length so probably should be an `ArrayBuilder` anyways. But I'm not sure what specifically you are mean.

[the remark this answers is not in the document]
hoodmane: Maybe use `kj::arr` for instance?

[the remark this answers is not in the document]
hoodmane: Yeah `kj::arr` works fine.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -317,40 +317,64 @@ kj::Array<kj::String> ArtifactBundler::parsePythonScriptImports(kj::Array<kj::St
...
+    "tempfile"_kj,
+    "typing"_kj,
+    "zipfile"_kj);
 
-  result[pkgImport.size()] = '.';
-  result[pkgImport.size() + 1] = 'p';
-  result[pkgImport.size() + 2] = 'y';
-  return result;
+kj::Array<kj::StringPtr> ArtifactBundler::getSnapshotImports() {
+  return kj::heapArray((kj::ArrayPtr<kj::StringPtr>)snapshotImports);
```
hoodmane, line 343: @danlapid this seems to be another place where it can't figure out what to do: `kj::heapArray()` accepts an `ArrayPtr` and I'm passing an `Array` but it doesn't seem to automatically decay to `ArrayPtr`.

danlapid: Just return an ArrayPtr instead

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -317,40 +317,64 @@ kj::Array<kj::String> ArtifactBundler::parsePythonScriptImports(kj::Array<kj::St
...
-kj::String importToModuleFilename(kj::StringPtr pkgImport) {
-  auto result = kj::heapString(pkgImport.size() + 3);
-  for (auto i = 0; i < pkgImport.size(); i++) {
-    if (pkgImport[i] == '.') {
-      result[i] = '/';
-    } else {
-      result[i] = pkgImport[i];
-    }
-  }
+kj::Array<kj::StringPtr> snapshotImports = kj::arr("_pyodide"_kj,
```
danlapid, line 320: this needs to be const to be safe

src/workerd/api/pyodide/pyodide.c++
hoodmane: Well this is passing out to JavaScript so I think if I combined making it `const` with this change, it should hopefully be a compile error? Because there isn't a way to prove that it is thread safe?

danlapid: Either that or jsg will do the copy in the backend? 😁

hoodmane: Not only does it not work, it doesn't seem possible to cast `const kj::Array<...>` to `const kj::ArrayPtr<...>` so instead I have to do:
```C++
kj::heapArray(snapshotImports.begin(), snapshotImports.size())
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -317,40 +317,64 @@ kj::Array<kj::String> ArtifactBundler::parsePythonScriptImports(kj::Array<kj::St
...
-  result[pkgImport.size() + 1] = 'p';
-  result[pkgImport.size() + 2] = 'y';
-  return result;
+kj::Array<kj::StringPtr> ArtifactBundler::getSnapshotImports() {
+  return kj::heapArray(snapshotImports.begin(), snapshotImports.size());
 }
 
-kj::Array<kj::String> ArtifactBundler::filterPythonScriptImports(
-    kj::HashSet<kj::String> workerModules, kj::Array<kj::String> imports) {
+kj::ArrayPtr<const char> strToArrayPtr(kj::StringPtr k) {
```
danlapid, line 346: `k.asArray()`

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -317,40 +317,60 @@ kj::Array<kj::String> ArtifactBundler::parsePythonScriptImports(kj::Array<kj::St
...
-  result[pkgImport.size() + 1] = 'p';
-  result[pkgImport.size() + 2] = 'y';
-  return result;
+bool isWorkerModuleImport(
+    kj::HashSet<kj::String>& workerModules, kj::ArrayPtr<const char> firstComponent) {
+  // TODO: handling for:
+  // 1. namespace packages
+  // 2. shared libraries
+  return workerModules.contains(kj::str(firstComponent, ".py")) ||
+      workerModules.contains(kj::str(firstComponent, "/__init__.py"));
```
danlapid, line 352: Aren't `__init__.py` files no longer required with new python versions to declare a module?

danlapid (approved):

src/workerd/api/pyodide/pyodide.c++
hoodmane: You may be thinking of "namespace packages":
https://peps.python.org/pep-0420/#specification

hoodmane: The relevant detail is that a namespace packages are never preferred over normal packages. So if there is a folder called fastapi with no `__init__.py` but also a package called fastapi, then the package will be preferred. If there is no package called fastapi, then the folder gets turned into a namespace package.

So ideally to handle this we need to tell this algorithm the list of imports defined by the installed packages. Since it's a bit more complicated, I figured it was okay to leave as a TODO for now.

dom96:
I think this might be breaking our validator-upload test in EW

PR #3508, opened 2025-02-10
Run import tests against each package bundle version

hoodmane:
This is the last part of #3451.

I think I understand the reason for the crash and I _think_ the change to `snapshot.ts` fixes it. The problem is that `_preloadDynamicLibs` is a bad decision. However, if I remove it entirely it causes another crash which I haven't understood yet.


hoodmane:
Well the change to `preloadDynamicLibs()` definitely gets rid of the openssl problem but three tests segfault:
```
python-import-annotated-types-tests
python-import-langchain_openai
dont-snapshot-pyodide
```
I know why `dont-snapshot-pyodide` is segfaulting and how to fix it but it's a bit complicated. I don't know why `annotated-types` or `langchain_openai` are segfaulting.

Various other import tests fail for probably less worrisome reasons. Including a lot of `File/folder bs4 being written by multiple packages`, some packages are misconfigured, cryptography needs to be added to the `topLevelEntropy` management.

dom96:
See above for new PR

PR #3509, opened 2025-02-10
feat: add Workflows to ValidationErrorReporter

LuisDuarte1:
In order to have proper validation for Workflows I think we will need a similar setup as DOs: ValidationErrorReporter now has a `addWorkflowClass` that behaves similarily as `addActorClass`. (This will need a internal PR in order to work properly).

This will be useful in cases that a user exports a class but doesn't extend from `WorkflowEntrypoint` (meaning they won't get `ctx` and `env` and see weird errors).

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -947,4 +947,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # This flag is intended to be a roll-up flag. That is, as additional APIs
   # reach EOL, new compat flags will be added for those that will have
   # `impliedByAfterDate(name = "removeNodeJsCompatEOL", ...` annotations.
+
+  enableWorkflowScriptValidation @109 :Bool
+      $compatEnableFlag("enable_validate_workflow_entrypoint")
+      $compatDisableFlag("disable_validate_workflow_entrypoint")
+      $compatEnableDate("2025-09-20");
+  # This flag enables additional checks in the control plane to validate that workflows are
+  # defined and used correctly
 }
--- a/src/workerd/server/server.c++
+++ b/src/workerd/server/server.c++
@@ -1784,6 +1784,17 @@ struct Server::ErrorReporter: public Worker::ValidationErrorReporter {
   void addActorClass(kj::StringPtr exportName) override {
     actorClasses.insert(kj::str(exportName));
   }
+
+  void addWorkflowClass(kj::StringPtr exportName, kj::Array<kj::String> methods) override {
+    // At runtime, we need to add it into the normal namedEntrypoints for Workflows to appear
+    // in `WorkerService`. This is a different method compared to `addEntrypoint` because we need to
+    // check for `WorkflowEntrypoint` inheritance at validation time.
+    kj::HashSet<kj::String> set;
+    for (auto& method: methods) {
+      set.insert(kj::mv(method));
+    }
+    namedEntrypoints.insert(kj::str(exportName), kj::mv(set));
+  }
 };
 
 // Implementation of ErrorReporter specifically for reporting errors in the top-level workerd
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -510,6 +510,7 @@ struct Worker::Impl {
   kj::HashMap<kj::String, api::ExportedHandler> namedHandlers;
   kj::HashMap<kj::String, ActorClassInfo> actorClasses;
   kj::HashMap<kj::String, EntrypointClass> statelessClasses;
+  kj::HashMap<kj::String, EntrypointClass> workflowClasses;
 
   // If set, then any attempt to use this worker shall throw this exception.
   kj::Maybe<kj::Exception> permanentException;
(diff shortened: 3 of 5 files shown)


a-robinson (approved):
Approving since this is identical to #3481, but like that one this should wait to merge until the internal change is ready.

src/workerd/server/server.c++
```diff
@@ -3048,6 +3048,43 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
   TRACE_EVENT("workerd", "Server::makeWorker()", "name", name.cStr());
   auto& localActorConfigs = KJ_ASSERT_NONNULL(actorConfigs.find(name));
 
+  struct ErrorReporter: public Worker::ValidationErrorReporter {
```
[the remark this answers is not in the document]
LuisDuarte1: yeah, that was duplicated - just a leftover from rebasing :sweat_smile:

src/workerd/server/server.c++
```diff
@@ -1698,6 +1698,10 @@ struct ErrorReporter: public Worker::ValidationErrorReporter {
   void addActorClass(kj::StringPtr exportName) override {
     actorClasses.insert(kj::str(exportName));
   }
+
+  void addWorkflowClass(kj::StringPtr exportName, kj::Array<kj::String> methods) override {
+    // This is only used for validation and has no runtime implications, at least for now.
```
danlapid, line 1703: Wasn't the logic of addEntrypoint called until now on workflow classes?
Shouldn't it have been?

LuisDuarte1: > Wasn't the logic of addEntrypoint called until now on workflow classes? Shouldn't it have been?

Yhep, but that only checks for the methods that exist in a given exported class but doesn't check if it inherits `WorkflowEntrypoint` - I've done the same approach as Durable Objects, since they also do that validation

src/workerd/io/compatibility-date.capnp
```diff
@@ -947,4 +947,9 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # This flag is intended to be a roll-up flag. That is, as additional APIs
   # reach EOL, new compat flags will be added for those that will have
   # `impliedByAfterDate(name = "removeNodeJsCompatEOL", ...` annotations.
+  
+  enableWorfkflowScriptValidation @109 :Bool
+      $compatEnableFlag("validate_workflow_entrypoint")
```
jasnell: If there's going to be a default on date, then we likely need to have a disable flag here also.

```suggestion
      $compatEnableFlag("validate_workflow_entrypoint")
      $compatDisableFlag("no_validate_workflow_entrypoint")
```

src/workerd/io/compatibility-date.capnp
```diff
@@ -947,4 +947,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatDisableFlag("no_validate_workflow_entrypoint")
+      $compatEnableDate("2025-08-20");
```
danlapid:
```suggestion
      $compatEnableDate("2025-09-20");
```

src/workerd/io/compatibility-date.capnp
```diff
@@ -947,4 +947,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # This flag is intended to be a roll-up flag. That is, as additional APIs
   # reach EOL, new compat flags will be added for those that will have
   # `impliedByAfterDate(name = "removeNodeJsCompatEOL", ...` annotations.
+  
+  enableWorfkflowScriptValidation @109 :Bool
```
danlapid: So is this just unused??

src/workerd/io/compatibility-date.capnp
```diff
@@ -947,4 +947,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatEnableFlag("validate_workflow_entrypoint")
+      $compatDisableFlag("no_validate_workflow_entrypoint")
+      $compatEnableDate("2025-08-20");
+  # enables validation for workflow entrypoint without breaking workflows deployed without validation
```
danlapid:
```suggestion
  # This flag enables additional checks in the control plane to validate that workflows are 
  # defined and used correctly.
```

src/workerd/io/compatibility-date.capnp
```diff
@@ -947,4 +947,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # This flag is intended to be a roll-up flag. That is, as additional APIs
   # reach EOL, new compat flags will be added for those that will have
   # `impliedByAfterDate(name = "removeNodeJsCompatEOL", ...` annotations.
+  
+  enableWorfkflowScriptValidation @109 :Bool
```
dom96:
```suggestion
  enableWorkflowScriptValidation @109 :Bool
```

src/workerd/io/compatibility-date.capnp
```diff
@@ -947,4 +947,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatEnableFlag("validate_workflow_entrypoint")
```
dom96: Nit: but I think it's good to keep the flag names consistent:

```suggestion
      $compatEnableFlag("enable_workflow_script_validation")
```

PR #3511, opened 2025-02-10
update ada to v3.0.1

anonrig:
Updates ada to latest version. @mikea added BAZEL support to Ada, so we don't have to maintain our bazel file anymore...


fhanau (approved):
Compiling the individual source files instead of the amalgamated source feels less convenient, but should be good enough still.

anonrig:
> Compiling the individual source files instead of the amalgamated source feels less convenient, but should be good enough still.

We can still use the old way, if it's more convenient. I don't have any strong opinions here...

fhanau:
> > Compiling the individual source files instead of the amalgamated source feels less convenient, but should be good enough still.
> 
> We can still use the old way, if it's more convenient. I don't have any strong opinions here...

Including ada would be cleaner (as noted in the ada BUILD.bazel) but I don't think this matters too much... can go ahead as-is.

PR #3513, opened 2025-02-10
simplify clangcl usage on windows builds

anonrig:
It seems we have a simpler way of using clang-cl compiler now.


fhanau:
Will have to think about this... I'm pretty sure this is incompatible with the former `--incompatible_enable_cc_toolchain_resolution` which is always-on in Bazel 7

fhanau:
Closing this, we depend on platform resolution so this won't work in the workerd build.

PR #3517, opened 2025-02-11
AI: remove redundant return statement

thatsKevinJain:
Explicitly returning the response body on `input.stream` is redundant as we already do that in the final return statement.


anonrig (commented):
This changes the behavior of this function. If res.body is undefined, and inputs['stream'] exists, previously it was returning undefined, right now it throws.

Is this a breaking change, or a bug fix?

anonrig (approved):

G4brym (approved):

PR #3518, opened 2025-02-11
remove unnecessary url tests

anonrig:
Since we are already testing them on WPT and pass them, let's remove it. I didn't remove others, tests that start with w3c because we don't fully pass them on WPT harness yet, due to harness.


jasnell (approved):
I don't really see much of a need to remove them but not blocking

PR #3528, opened 2025-02-11
remove dependabot

anonrig:
too much is too much


danlapid:
Why is it too much? theoretically we merge all of the updates and then dependabot shouldn't create too much spam

fhanau:
> Why is it too much? theoretically we merge all of the updates and then dependabot shouldn't create too much spam

See internal discussion in Workers-OSS – dependabot opened ~8 PRs which caused lots of queueing of CI actions for workerd and workers-sdk based on the org-wide CI resource limitations.

danlapid (approved):
Yeah looks horrible, so sad, let's add this as soon as possible with new plan.

PR #3529, opened 2025-02-11
use ada-url's url_pattern implementation

anonrig:
Since we moved URLPattern implementation to Ada and shared it with Node.js, we can now use it as a dependency and get rid of our implementation. This will allow us to have higher compatibility with the web-platform tests and spec, and reduce our codebase by 3800 lines...

Additionally with the help of @npaun, we now run web-platform tests and can get rid of our custom URLPattern tests for better web compatibility and maintenance, since WPT is always changing!

## Changes

- Adds `get hasRegExpGroups(): boolean;` field

## TODO

- [x] Put it behind an autogate so we can roll this safely.


src/workerd/api/urlpattern.c++
```diff
@@ -4,289 +4,191 @@
...
-        }
-        KJ_CASE_ONEOF(pattern, jsg::UrlPattern) {
-          return create(js, kj::mv(pattern));
-        }
+  }
+
+  KJ_IF_SOME(b, maybeBase) {
+    KJ_SWITCH_ONEOF(b) {
+      KJ_CASE_ONEOF(str, kj::String) {
+        *base = std::string_view(str.begin(), str.size());
```
danlapid, line 113: I don't see how this is not a null pointer deref, why even use raw pointers in the first place?

src/workerd/api/urlpattern.c++
```diff
@@ -4,289 +4,191 @@
...
-      index++;
-    }
-
-    return URLPattern::URLPatternComponentResult{
-      .input = kj::str(input),
-      .groups = Groups{.fields = fields.releaseAsArray()},
-    };
+  for (const auto& [key, value]: other.groups) {
+    result.groups.set(js, js.str(kj::heapArray(key.c_str(), key.size())),
+        js.str(kj::heapArray(key.c_str(), key.size())));
```
danlapid, line 55: afaik this is a double copy, can we use StringPtr or ArrayPtr instead?

src/workerd/api/urlpattern.c++
```diff
@@ -4,289 +4,191 @@
...
-#undef V
+  KJ_IF_SOME(b, baseURL) {
+    init.base_url = std::string(b.begin(), b.size());
+  }
+  return init;
 }
+URLPattern::URLPatternComponentResult create(
+    jsg::Lock& js, const ada::url_pattern_component_result& other) {
+  auto result = URLPattern::URLPatternComponentResult{
+    .input = kj::String(kj::heapArray(other.input.c_str(), other.input.size())),
```
danlapid, line 49: Likewise for double copy, there has to be better constructors for this

src/workerd/api/urlpattern.c++
```diff
@@ -4,289 +4,191 @@
...
+#include "ada.h"
+
 #include <kj/vector.h>
 
 namespace workerd::api {
+URLPattern::URLPatternInit create(const ada::url_pattern_init& other) {
+  URLPattern::URLPatternInit result{};
+#define V(_, name)                                                                                 \
+  if (auto v = other.name) {                                                                       \
+    result.name = kj::String(kj::heapArray(v->c_str(), v->size()));                                \
```
danlapid, line 16: likewise etc.

src/workerd/api/urlpattern.c++
```diff
@@ -4,289 +4,191 @@
...
-}
-kj::StringPtr URLPattern::getPathname() {
-  return inner.getPathname().getPattern();
-}
-kj::StringPtr URLPattern::getSearch() {
-  return inner.getSearch().getPattern();
-}
-kj::StringPtr URLPattern::getHash() {
-  return inner.getHash().getPattern();
+  auto vecInputs = kj::Vector<URLPattern::URLPatternInput>(2);
```
danlapid, line 68: Should this just be a `kj::heapArray(other.inputs.size());`?

src/workerd/api/urlpattern.c++
```diff
@@ -4,289 +4,191 @@
...
-        .port = i.port.map([](kj::String& str) { return kj::str(str); }),
-        .pathname = i.pathname.map([](kj::String& str) { return kj::str(str); }),
-        .search = i.search.map([](kj::String& str) { return kj::str(str); }),
-        .hash = i.hash.map([](kj::String& str) { return kj::str(str); }),
-        .baseURL = i.baseURL.map([](kj::String& str) { return kj::str(str); }),
-      });
+  }
+
+  KJ_IF_SOME(b, maybeBase) {
+    *base = std::string_view(b.begin(), b.size());
```
danlapid, line 152: likewise for null deref

src/workerd/io/BUILD.bazel
```diff
@@ -351,13 +352,17 @@ kj_test(
...
+        ":io",
+        "//src/workerd/api:urlpattern",
+    ],
 )
 
 kj_test(
     src = "compatibility-date-test.c++",
     deps = [
         ":io",
+        "//src/workerd/api:urlpattern",
```
danlapid, line 365: Why does `compatibility-date-test` depend on urlpattern? 🤔

src/workerd/io/BUILD.bazel
```diff
@@ -351,13 +352,17 @@ kj_test(
 
 kj_test(
     src = "promise-wrapper-test.c++",
-    deps = [":io"],
+    deps = [
+        ":io",
+        "//src/workerd/api:urlpattern",
```
danlapid, line 357: Why does promise-wrapper-test depend on jsgpattern?

anonrig: I have no idea. I am investigating this at the moment.

src/workerd/api/urlpattern.c++
anonrig: ada::parse_url_pattern takes const std::string_view* pointer because it is more efficient compared to std::optional<std::string_view>

src/workerd/api/urlpattern.c++
anonrig: yes all of these needs to change.

src/workerd/api/urlpattern.c++
anonrig: how come?

src/workerd/io/BUILD.bazel
anonrig: I have no idea but I'm investigating the weird linker errors right now

src/workerd/api/urlpattern.c++
danlapid: That's fine, but the code is still a null deref as is.

src/workerd/api/urlpattern.c++
danlapid: base is defined as:
`std::string_view* base = nullptr;`
Then you have
`*base = ...`

anonrig:
Other than putting this behind a flag like @jasnell recommended, this is ready to review.

src/workerd/api/wpt/urlpattern-test.ts
```diff
@@ -17,151 +17,21 @@ export default {
...
-  },
+  'urlpattern-hasregexpgroups-tests.js': {},
   'urlpattern-hasregexpgroups.any.js': {},
   'urlpattern.any.js': {},
   'urlpattern.https.any.js': {},
   'urlpatterntests.js': {
-    comment: 'urlpattern implementation will soon be replaced with ada-url',
+    comment: 'skipping invalid tests',
     expectedFailures: [
-      // Each of these *ought* to pass. They are included here because we
```
fhanau, line 34: I assume that this comment is no longer needed since the tests themselves are invalid here.

src/workerd/api/urlpattern.h
```diff
@@ -4,8 +4,12 @@
 
 #pragma once
 
+#include "ada.h"
```
fhanau, line 7: As per KJ style, this should be <ada.h> (some people on the team have strong opinions about this).

src/workerd/api/BUILD.bazel
```diff
@@ -249,6 +249,7 @@ wd_cc_library(
     kj_test(
         src = f,
         deps = [
+            ":urlpattern",
```
fhanau, line 252: There is no such target; it should be worthwhile to create one by isolating it from io as noted in the TODO at the top of this file. It is not necessary to do so in this PR though as doing so may require downstream changes. Furthermore, urlpattern.h is not used by any of these tests (it is needed for api-rtti-test so it would be correct to add it there).

justfile
```diff
@@ -21,7 +21,7 @@ build *args="//...":
   bazel build {{args}}
 
 build-asan *args="//...":
-  just build {{args}} --config=asan --sandbox_debug
```
fhanau, line 24: Changes to this file are already in #3551, let's avoid duplicating them here

src/workerd/api/urlpattern.h
```diff
@@ -4,8 +4,12 @@
...
+
 #include <workerd/jsg/jsg.h>
-#include <workerd/jsg/url.h>
+
+#include <optional>
+#include <string_view>
```
npaun, line 12:
```suggestion
#include <kj/array.h>
#include <kj/one-of.h>
#include <kj/string.h>

#include <string>
#include <string_view>
#include <utility> // For std::move
#include <vector>
```

For include-what-you-use reasons.

src/workerd/api/urlpattern.c++
```diff
@@ -4,309 +4,252 @@
...
+  }
+  return std::nullopt;
 }
 
-jsg::Ref<URLPattern> create(jsg::Lock& js, jsg::UrlPattern pattern) {
-  bool ignoreCase = pattern.getIgnoreCase();
+ada::url_pattern_options URLPattern::URLPatternOptions::toAdaType() const {
+  ada::url_pattern_options options{};
+  options.ignore_case = ignoreCase.orDefault(false);
+  return options;
```
npaun, line 54:
```suggestion
  return {.ignore_case = ignoreCase.orDefault(false)};
```

src/workerd/api/BUILD.bazel
anonrig: I created urlpattern target on a different PR

anonrig: It exists: https://github.com/cloudflare/workerd/pull/3546

src/workerd/api/urlpattern.c++
```diff
@@ -4,309 +4,252 @@
...
-    jsg::Lock::RegExpFlags flags = jsg::Lock::RegExpFlags::kUNICODE;
-    if (ignoreCase) {
-      flags = static_cast<jsg::Lock::RegExpFlags>(
-          flags | static_cast<int>(jsg::Lock::RegExpFlags::kIGNORE_CASE));
+std::optional<std::vector<std::optional<std::string>>> URLPattern::URLPatternRegexEngine::
+    regex_search(std::string_view input, const regex_type& pattern) {
+  jsg::Lock& js = jsg::Lock::from(v8::Isolate::GetCurrent());
+  KJ_IF_SOME(matches, pattern.getHandle(js)(js, kj::StringPtr(input.data(), input.size()))) {
+    std::vector<std::optional<std::string>> results;
+    results.reserve(matches.size());
```
erikcorry, line 35: I think we are reserving one more slot than we need here?

src/workerd/api/urlpattern.c++
```diff
@@ -4,309 +4,252 @@
...
+      auto obj = std::get<ada::url_pattern_init>(input);
+      vecInputs[i] = createURLPatternInit(obj);
+    }
+    i++;
+  }
+  result.inputs = mv(vecInputs);
+  return result;
 }
 
+#define V(uppercase, lowercase)                                                                    \
```
erikcorry, line 134: Code like this is clearer if you give V a meaningful name like DEFINE_GETTER.

erikcorry (approved):

src/workerd/api/urlpattern.c++
anonrig: Yes but adding -1 adds a branch (checking if size is bigger than 0), which reduces the impact of this optimization

src/workerd/api/urlpattern.c++
anonrig: Agreed. Updating now. Thanks for the review @erikcorry

src/workerd/api/urlpattern.h
anonrig: This fails

```
src/workerd/api/urlpattern.h:9:10: error: 'ada.h' file not found with <angled> include; use "quotes" instead
    9 | #include <ada.h>
      |          ^~~~~~~
      |          "ada.h"
1 error generated.
```

npaun (approved):

anonrig:
I'll put this behind an autogate before merging.

jasnell:
Probably a good idea to have this PR be a draft until the autogate it added.

src/workerd/api/tests/url-test.js
```diff
@@ -9035,448 +9035,6 @@ export const urlPatternBasics = {
   },
 };
 
-export const urlPatternMdn1 = {
```
danlapid, line 9038: Would appreciate a review-comment noting why this deletion is OK.
Assuming it's because these tests are duplicates of WPT?

src/workerd/api/urlpattern-standard.h
```diff
@@ -0,0 +1,153 @@
...
+#include <workerd/jsg/jsg.h>
+
+#include <kj/array.h>
+#include <kj/one-of.h>
+#include <kj/string.h>
+
+#include <optional>
+#include <string>
+#include <string_view>
+#include <utility>  // For std::move
```
fhanau, line 18: kj::mv expressly exists as a shorter replacement for std::move that doesn't require pulling in <utility>: https://github.com/capnproto/capnproto/blob/v1.1.0/c%2B%2B/src/kj/common.h#L692

src/workerd/api/global-scope.h
```diff
@@ -708,7 +713,13 @@ class ServiceWorkerGlobalScope: public WorkerGlobalScope {
       JSG_NESTED_TYPE(URL);
       JSG_NESTED_TYPE(URLSearchParams);
     }
-    JSG_NESTED_TYPE(URLPattern);
+
+    // We conditionally enable a more spec compliant URLPattern to avoid any breakages.
+    if (util::Autogate::isEnabled(util::AutogateKey::URLPATTERN)) {
+      JSG_NESTED_TYPE(urlpattern::URLPattern);
```
jasnell, line 719: You'll want to use the `JSG_NESTED_TYPE_NAMED` variant here, otherwise the `urlpattern:` bit will end up part of the name here.

src/workerd/api/tests/url-test.js
danlapid, line 1: As-is afaik workerd tests don't run on all-autogates variants which is leading me to ask - is there any test coverage for this change included in this PR at all?

src/workerd/api/tests/url-test.js
jasnell, line 9038: Is it necessary to remove these existing tests? I know they might be duplicative of the web platform tests, but I still think there's value in them.

src/workerd/api/urlpattern-standard.h
```diff
@@ -0,0 +1,153 @@
...
+
+  JSG_RESOURCE_TYPE(URLPattern) {
+#define V(Name, name) JSG_READONLY_PROTOTYPE_PROPERTY(name, get##Name);
+    URL_PATTERN_COMPONENTS(V)
+#undef V
+    JSG_READONLY_PROTOTYPE_PROPERTY(hasRegExpGroups, getHasRegExpGroups);
+    JSG_METHOD(test);
+    JSG_METHOD(exec);
+
+    JSG_TS_OVERRIDE({
```
danlapid, line 131: formatting feels ugly, maybe just inline to
`JSG_TS_OVERRIDE({ get hasRegExpGroups(): boolean; });` ?

src/workerd/api/urlpattern-standard.h
anonrig: TIL. Thanks!

src/workerd/api/urlpattern-standard.h
danlapid: Do you even need a TS_OVERRIDE?

src/workerd/api/urlpattern-standard.h
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  // A struct providing the URLPattern matching results for a single
+  // URL component. The URLPatternComponentResult is only ever used
+  // as a member attribute of a URLPatternResult struct. The
+  // URLPatternComponentResult API is defined as part of the URLPattern
+  // specification.
+  struct URLPatternComponentResult final {
+    jsg::JsString input;
+    jsg::JsObject groups;
```
danlapid, line 74: Previously groups was well typed:
```
jsg::Dict<kj::String, kj::String> groups;
```
Why did we change that?

src/workerd/api/tests/url-test.js
anonrig: I recommend removing them and relying on web-platform tests since MDN is not a good source for correctness.

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+  KJ_IF_SOME(b, maybeBase) {
+    base_url_base = std::string(b.begin(), b.size());
+    base_url = std::string_view(base_url_base);
+  }
+
+  KJ_IF_SOME(mi, maybeInput) {
+    KJ_SWITCH_ONEOF(mi) {
+      KJ_CASE_ONEOF(str, kj::String) {
+        result = inner.exec(
+            std::string_view(str.begin(), str.size()), base_url ? &(*base_url) : nullptr);
```
danlapid, line 233:
```suggestion
            std::string_view(str.begin(), str.size()), base_url ? base_url.data() : nullptr);
```
as a general find and replace.
the &* syntax is horrible imho.

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+kj::Maybe<URLPattern::URLPatternResult> URLPattern::exec(jsg::Lock& js,
+    jsg::Optional<kj::OneOf<kj::String, URLPatternInit>> maybeInput,
+    jsg::Optional<kj::String> maybeBase) {
+  ada::result<std::optional<ada::url_pattern_result>> result;
+  std::optional<std::string_view> base_url{};
+  std::string base_url_base;
+
+  KJ_IF_SOME(b, maybeBase) {
+    base_url_base = std::string(b.begin(), b.size());
+    base_url = std::string_view(base_url_base);
```
danlapid, line 226: Why can't you create a string_view straight from a kj::String? you do it below for inner.exec's first argument)

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+      KJ_CASE_ONEOF(pi, URLPattern::URLPatternInit) {
+        result = inner.exec(pi.toAdaType(), base_url ? &(*base_url) : nullptr);
+      }
+    }
+  } else {
+    result = inner.exec(ada::url_pattern_init{}, base_url ? &(*base_url) : nullptr);
+  }
+
+  // If result does not exist, we should throw.
+  JSG_REQUIRE(result.has_value(), TypeError, "Failed to exec URLPattern");
```
danlapid, line 244: Does ada provide us with any more information regarding the failure type?

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+      KJ_CASE_ONEOF(str, kj::String) {
+        result = inner.exec(
+            std::string_view(str.begin(), str.size()), base_url ? &(*base_url) : nullptr);
+      }
+      KJ_CASE_ONEOF(pi, URLPattern::URLPatternInit) {
+        result = inner.exec(pi.toAdaType(), base_url ? &(*base_url) : nullptr);
+      }
+    }
+  } else {
+    result = inner.exec(ada::url_pattern_init{}, base_url ? &(*base_url) : nullptr);
```
danlapid, line 240: is ada exec noexcept? are there exceptions that should be converted to js errors?

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  // If result does not exist, we should throw.
+  JSG_REQUIRE(result.has_value(), TypeError, "Failed to exec URLPattern");
+
+  if (result->has_value()) {
```
danlapid, line 246: Kinda odd that you JSG_REQUIRE a line earlier and now you are checking for value existence else returning none

src/workerd/api/urlpattern-standard.c++
danlapid, line 233: Maybe extract this ternary operator to another variable similar to what you do in test function as well.
Looks a tiny bit nicer.

src/workerd/api/urlpattern-standard.c++
danlapid: Also odd that you do do that in the test function but not here

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  std::string_view* base_opt = base ? &(*base) : nullptr;
+  auto result =
+      ada::parse_url_pattern<URLPatternRegexEngine>(input, base_opt, options ? &*options : nullptr);
+  if (result.has_value()) {
+    return jsg::alloc<URLPattern>(std::move(*result));
+  }
+
+  JSG_FAIL_REQUIRE(TypeError, "Failed to construct URLPattern");
```
danlapid, line 185: different error handling compared to the other functions, why not be consistent with
`JSG_REQUIRE(result.has_value(), ...);`

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+
+  if (!options.has_value()) {
+    KJ_IF_SOME(o, maybeOptions) {
+      options = o.toAdaType();
+    }
+  }
+
+  std::string_view* base_opt = base ? &(*base) : nullptr;
+  auto result =
+      ada::parse_url_pattern<URLPatternRegexEngine>(input, base_opt, options ? &*options : nullptr);
```
danlapid, line 180: You extracted the ternary operator for base but not for options, why?

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+    }
+    i++;
+  }
+  result.inputs = mv(vecInputs);
+  return result;
+}
+
+#define DEFINE_GETTER(uppercase, lowercase)                                                        \
+  kj::StringPtr URLPattern::get##uppercase() const {                                               \
+    auto value = inner.get_##lowercase();                                                          \
```
danlapid, line 134: Is value an owned object here? is it a string_view?
Just gotta make sure that the underlying string outlives the return StringPtr.

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+
+URLPattern::URLPatternResult URLPattern::createURLPatternResult(
+    jsg::Lock& js, const ada::url_pattern_result& other) {
+  URLPatternResult result{
+#define V(_, name) .name = URLPattern::createURLPatternComponentResult(js, other.name),
+    URL_PATTERN_COMPONENTS(V)
+#undef V
+  };
+
+  auto vecInputs = kj::heapArray<kj::OneOf<jsg::JsString, URLPatternInit>>(other.inputs.size());
```
danlapid, line 115: This could be refactored to use KJ_MAP which would look much better IMHO

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+bool URLPattern::URLPatternRegexEngine::regex_match(
+    std::string_view input, const regex_type& pattern) {
+  jsg::Lock& js = jsg::Lock::from(v8::Isolate::GetCurrent());
+  return pattern.getHandle(js).match(js, kj::StringPtr(input.data(), input.size()));
+}
+
+std::optional<std::vector<std::optional<std::string>>> URLPattern::URLPatternRegexEngine::
+    regex_search(std::string_view input, const regex_type& pattern) {
+  jsg::Lock& js = jsg::Lock::from(v8::Isolate::GetCurrent());
+  KJ_IF_SOME(matches, pattern.getHandle(js)(js, kj::StringPtr(input.data(), input.size()))) {
```
danlapid, line 33: Is this safe if input happens to not be null terminated?

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+    std::vector<std::optional<std::string>> results;
+    results.reserve(matches.size());
```
danlapid, line 35: This can be inlined into the above line as a constructor argument.

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+    for (size_t i = 1; i < matches.size(); i++) {
```
danlapid, line 36: Shouldn't this be `<=`?

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+
+  return js.tryCatch([&]() -> std::optional<regex_type> {
+    return jsg::JsRef(js, js.regexp(kj::StringPtr(pattern.data(), pattern.size()), flags));
+  }, [&](auto reason) -> std::optional<regex_type> { return std::nullopt; });
+}
+
+bool URLPattern::URLPatternRegexEngine::regex_match(
+    std::string_view input, const regex_type& pattern) {
+  jsg::Lock& js = jsg::Lock::from(v8::Isolate::GetCurrent());
+  return pattern.getHandle(js).match(js, kj::StringPtr(input.data(), input.size()));
```
danlapid, line 27: Is this safe if input happens to not be null terminated?

src/workerd/api/urlpattern-standard.c++
```diff
@@ -0,0 +1,253 @@
...
+    create_instance(std::string_view pattern, bool ignore_case) {
+  jsg::Lock& js = jsg::Lock::from(v8::Isolate::GetCurrent());
+  jsg::Lock::RegExpFlags flags = jsg::Lock::RegExpFlags::kUNICODE_SETS;
+  if (ignore_case) {
+    flags = static_cast<jsg::Lock::RegExpFlags>(
+        flags | static_cast<int>(jsg::Lock::RegExpFlags::kIGNORE_CASE));
+  }
+
+  return js.tryCatch([&]() -> std::optional<regex_type> {
+    return jsg::JsRef(js, js.regexp(kj::StringPtr(pattern.data(), pattern.size()), flags));
```
danlapid, line 20: Is this safe if pattern happens to not be null terminated?

src/workerd/jsg/jsvalue.c++
```diff
@@ -327,10 +327,15 @@ kj::Maybe<JsArray> JsRegExp::operator()(Lock& js, const JsString& input) const {
 
 kj::Maybe<JsArray> JsRegExp::operator()(Lock& js, kj::StringPtr input) const {
   auto result = check(inner->Exec(js.v8Context(), js.str(input)));
-  if (result->IsNullOrUndefined()) return kj::none;
```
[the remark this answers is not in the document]
jasnell: This *should* be fine really. There's only one other place in the code using this (the original URLPattern impl) and this change shouldn't break that.

src/workerd/api/urlpattern-standard.h
jasnell: Changing is fine, I think, might just need a type override tho.

src/workerd/api/urlpattern-standard.h
```diff
@@ -0,0 +1,153 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
```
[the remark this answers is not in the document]
anonrig: This follows the existing pattern of url.h and url-standard.h where url-standard.h uses ada.

src/workerd/jsg/jsvalue.c++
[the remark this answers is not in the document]
anonrig: regexp exec returns null or value according to the spec and the v8. it never returned undefined. this is just a correction to use the correct type. it has no visible side-effects or behavior effects. if insisted, i'm more than happy to remove this.

src/workerd/jsg/jsvalue.c++
```diff
@@ -327,10 +327,15 @@ kj::Maybe<JsArray> JsRegExp::operator()(Lock& js, const JsString& input) const {
...
+  if (result->IsNull()) return kj::none;
   return JsArray(result.As<v8::Array>());
 }
 
+bool JsRegExp::match(Lock& js, kj::StringPtr input) {
+  auto result = check(inner->Exec(js.v8Context(), js.str(input)));
+  return !result->IsNull();
```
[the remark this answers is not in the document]
anonrig: I didn't quite get this comment. Would you mind explaining?

regexp has match() and exec() methods in JS. we implement exec() with operator(), and this just implements match() which returns a bool.

src/workerd/api/urlpattern-standard.h
```diff
@@ -0,0 +1,153 @@
...
+
+#include "ada.h"
+
+#include <workerd/jsg/jsg.h>
+
+#include <kj/array.h>
+#include <kj/one-of.h>
+#include <kj/string.h>
+
+#include <optional>
```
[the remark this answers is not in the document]
anonrig: some folks recommended including all headers that is used. they're not needed at all.

src/workerd/api/urlpattern-standard.h
```diff
@@ -0,0 +1,153 @@
...
+#include <string>
+#include <string_view>
+#include <utility>  // For std::move
+#include <vector>
+
+namespace workerd::api {
+
+namespace urlpattern {
```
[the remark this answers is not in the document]
anonrig: This follows the pattern of url-standard.h where the implementation is moved into workerd::api::url, and here to workerd::api::urlpattern

src/workerd/api/tests/url-test.js
anonrig: Yes, @jasnell raised this concern as well in his previous review.

src/workerd/api/tests/url-test.js
anonrig: I'm waiting on https://github.com/cloudflare/workerd/pull/3665 to land in order to have test coverage on the autogated feature.

src/workerd/api/urlpattern-standard.h
anonrig: This actually made the code a lot more simpler and easier to implement. Also, I used a v8::Object::New constructor that takes size, key and values array to speed up the javascript object creation whereas the existing jsg::Dict doesn't use that constructor.

src/workerd/api/urlpattern-standard.c++
anonrig: This parameter has `std::string_view*` type. I don't think this would work because base_url.data() doesn't return std::string_view* nor it can be casted into string_view*. Please correct me if i'm wrong

src/workerd/api/urlpattern-standard.c++
anonrig: Yeah this is a good catch. I was experimenting which resulted in this inconsistency. I'll fix it. Thanks Dan

src/workerd/api/urlpattern-standard.c++
anonrig: URLPattern spec only returns TypeError and doesnt' mention any other error type. We can change the implementation and have unique error messages but it's not standardized, unfortunately.

src/workerd/api/urlpattern-standard.c++
anonrig: the function isn't but due to the limitation of node.js and how ada is implemented, it never throws an error. (ps: node.js has -fno-exceptions enabled)

src/workerd/api/urlpattern-standard.c++
anonrig: It is std::string_view where the underlying data is owned by ada::url_pattern class

src/workerd/api/urlpattern-standard.c++
anonrig: Because this function can throw, return null or return a value. The underlying type is std::expected<std::optional<T>>. It looks odd because std::expected and std::optional both has has_value() function

src/workerd/api/urlpattern-standard.h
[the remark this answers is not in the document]
npaun: I'm (one of) the guys that recommended adding them. Each of these headers corresponds to a type that's referenced in the header file. Forward-declaring STL stuff is kind of errrh...

src/workerd/api/urlpattern-standard.c++
anonrig: No because regexp exec() method returns the original input as the first argument. We want to NOT include that in the returning vector.

src/workerd/api/urlpattern-standard.c++
anonrig: Input is always null terminated, since it is called from Ada, and the underlying data is std::string.

src/workerd/api/urlpattern-standard.h
anonrig: I don't know why but we need it... The previous implementation doesn't have hasRegExpGroups getter. (which is a non-spec compliant bug)

src/workerd/api/urlpattern-standard.c++
anonrig: Pattern is always null terminated, since it is called from Ada, and the underlying data is std::string.

src/workerd/api/urlpattern-standard.c++
danlapid: std::strings are not always null terminated

src/workerd/api/urlpattern-standard.h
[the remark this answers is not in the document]
danlapid: You should include everything you use
My comment was merely to validate that you use everything you are including.

src/workerd/api/urlpattern-standard.h
danlapid: What's the generated types if you don't set the override?

src/workerd/api/urlpattern-standard.h
danlapid: Seems like we can amend jsg::Dict to use whatever api is best but I am indifferent to this if you think it's too much of a lift

src/workerd/api/urlpattern-standard.h
anonrig: Basically it removes the getter.

src/workerd/api/urlpattern-standard.c++
danlapid: Nit: I would add that as a comment

src/workerd/api/urlpattern-standard.h
anonrig: If you are comfortable with this: I can follow up after this PR lands?

src/workerd/api/urlpattern-standard.c++
danlapid: Yeah sorry, didn't notice it's an optional 
Then you could do
'&base_url.value()' 
Still nice imho then the current syntax

src/workerd/api/urlpattern-standard.c++
danlapid: Sorry I still don't see it
You are calling result->has_value() twice

danlapid: Oh
One time it's a method call on result and one time it's on a deref of result
There has to be a more explicit way to write this

src/workerd/api/urlpattern-standard.h
danlapid: Does does rtti not support JSG_PROTOTYPE_READOLY_PROPERTY? Seems like we should solve that at the source!

src/workerd/api/urlpattern-standard.c++
anonrig: Agreed

src/workerd/api/urlpattern-standard.h
anonrig: I've added this to my list. I'll take a look.

src/workerd/api/urlpattern-standard.c++
anonrig: Resolving for now. We can go back to this in the future. cc @jasnell

src/workerd/api/urlpattern-standard.c++
anonrig: I wish...

src/workerd/api/urlpattern-standard.c++
anonrig: Agreed.

PR #3531, opened 2025-02-12
[o11y] Implement Cache API span tags

fhanau:
Adds support for url.full tag, scaffolding is added to support for cache_control tags but this is not yet complete.

Also see the downstream PR.

### DIFF
--- a/src/workerd/api/cache.c++
+++ b/src/workerd/api/cache.c++
@@ -89,7 +89,8 @@ jsg::Promise<jsg::Optional<jsg::Ref<Response>>> Cache::match(
       return js.resolvedPromise(jsg::Optional<jsg::Ref<Response>>());
     }
 
-    auto httpClient = getHttpClient(context, jsRequest->serializeCfBlobJson(js), "cache_match"_kjc);
+    auto httpClient = getHttpClient(
+        context, jsRequest->serializeCfBlobJson(js), "cache_match"_kjc, jsRequest->getUrl());
     auto requestHeaders = kj::HttpHeaders(context.getHeaderTable());
     jsRequest->shallowCopyHeadersTo(requestHeaders);
     requestHeaders.set(context.getHeaderIds().cacheControl, "only-if-cached");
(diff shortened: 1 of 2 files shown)


src/workerd/api/cache.c++
```diff
@@ -516,17 +520,29 @@ jsg::Promise<bool> Cache::delete_(
...
+  // cacheControl to support other revalidation, cacheability and expiration directives too.
+  KJ_IF_SOME(c, cacheControl) {
+    if (c == "must-revalidate") {
+      userSpan.setTag("cache_control.revalidation"_kjc, kj::str("must-revalidate"));
+    }
+  }
   auto cacheClient = context.getCacheClient();
   auto httpClient = cacheName
                         .map([&](kj::String& n) {
     return cacheClient->getNamespace(n, kj::mv(cfBlobJson), span);
```
fhanau, line 543: Perhaps these should also need to be updated to propagate the user span, but since we attach the user span to httpClient below we should be fine in terms of span lifetime which is what we need for now.

src/workerd/api/cache.c++
```diff
@@ -249,6 +250,8 @@ jsg::Promise<void> Cache::put(jsg::Lock& js,
         "Cannot cache response to a range request (206 Partial Content).");
 
     auto responseHeadersRef = jsResponse->getHeaders(js);
+    auto cacheControl = responseHeadersRef->get(jsg::ByteString(kj::str("Cache-Control")));
```
mikea, line 253: does jsg::bytestring really need a kj::String or will _kj do?

src/workerd/api/cache.c++
```diff
@@ -516,17 +520,29 @@ jsg::Promise<bool> Cache::delete_(
   });
 }
 
-kj::Own<kj::HttpClient> Cache::getHttpClient(
-    IoContext& context, kj::Maybe<kj::String> cfBlobJson, kj::LiteralStringConst operationName) {
+kj::Own<kj::HttpClient> Cache::getHttpClient(IoContext& context,
+    kj::Maybe<kj::String> cfBlobJson,
+    kj::LiteralStringConst operationName,
+    kj::StringPtr url,
+    kj::Maybe<jsg::ByteString> cacheControl) {
```
mikea, line 527: why is it a ByteString and not a StringPtr? Isn't copy-free conversion at call site possible?

mikea (approved):

src/workerd/api/cache.c++
fhanau: Looks like it is required – jsg::ByteString inherits from kj::String, passing a StringPtr results in a `no matching constructor for initialization of 'kj::String'` compile error. We also use kj::str() for this purpose below.

PR #3537, opened 2025-02-13
Move Python SDK from `cloudflare.workers` to `workers`.

dom96:
The SDK is also injected at the old location for backwards compatibility, but only on the oldest Pyodide version.


danlapid (approved):

danlapid (approved):

dom96:
It seems that rebuilding the pyodide.capnp.bin bundle succeeded, but the new bundle doesn't contain the changes in this PR? Confused by what is causing this, the CI run seems to have checked out the correct commit https://github.com/cloudflare/workerd/actions/runs/13312732440/job/37179125654.

PR #3538, opened 2025-02-13
Refactor to use C++ implementation of requirement gathering for Python.

dom96:
The current state of things is that the logic for gathering package requirements is duplicated in TS and C++. This PR moves it into workerd so that the same implementation can be used in workerd and EW.

### DIFF
--- a/src/pyodide/internal/setupPackages.ts
+++ b/src/pyodide/internal/setupPackages.ts
@@ -6,11 +6,11 @@ import {
   LOAD_WHEELS_FROM_R2,
   LOCKFILE,
   LOAD_WHEELS_FROM_ARTIFACT_BUNDLER,
-  PACKAGES_VERSION,
   USING_OLDEST_PACKAGES_VERSION,
 } from 'pyodide-internal:metadata';
 import { simpleRunPython } from 'pyodide-internal:util';
 import { default as EmbeddedPackagesTarReader } from 'pyodide-internal:packages_tar_reader';
+import { default as MetadataReader } from 'pyodide-internal:runtime-generated/metadata';
 
 const canonicalizeNameRegex = /[-_.]+/g;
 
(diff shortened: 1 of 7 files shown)


src/pyodide/internal/setupPackages.ts
```diff
@@ -340,5 +312,7 @@ function addPackageToLoad(
 }
 
 export { REQUIREMENTS };
-export const TRANSITIVE_REQUIREMENTS = getTransitiveRequirements();
+export const TRANSITIVE_REQUIREMENTS = new Set(
```
danlapid, line 315: Can't we return sets from c++ instead of converting this to a Set here?

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -117,6 +119,18 @@ int PyodideMetadataReader::readMemorySnapshot(int offset, kj::Array<kj::byte> bu
   return readToTarget(KJ_REQUIRE_NONNULL(memorySnapshot), offset, buf);
 }
 
+kj::Array<kj::String> PyodideMetadataReader::getTransitiveRequirements() {
+  auto packages = parseLockFile(packagesLock);
+  auto depMap = getDepMapFromPackagesLock(*packages);
+
+  auto allRequirements = getPythonPackageNames(*packages, depMap, requirements, packagesVersion);
```
danlapid, line 126: I really wish we could instead add to jsg the option to return HashSet as JsSet. we are converting a set to an array only to convert it back to a set.
Is it really that difficult to add native jsg support for sets?

src/workerd/api/pyodide/requirements.h
```diff
@@ -0,0 +1,33 @@
...
+namespace workerd::api::pyodide {
+
+capnp::json::Value::Reader getField(
+    capnp::List<::capnp::json::Value::Field, capnp::Kind::STRUCT>::Reader &object,
+    kj::StringPtr name);
+
+kj::String canonicalizePythonPackageName(kj::StringPtr name);
+
+typedef kj::HashMap<kj::String, kj::Vector<kj::String>>
+    DepMap;  // map from requirement to list of dependencies
```
danlapid, line 20: Nit: can we move the comment to above the typedef so that the type name doesn't get newlined?

src/workerd/api/pyodide/pyodide.c++
dom96: I haven't tried, but indeed it does sound like it should be fairly simple. I'll have a look.

dom96: This should do it https://github.com/cloudflare/workerd/pull/3550

src/workerd/api/pyodide/requirements.c++
```diff
@@ -0,0 +1,132 @@
...
+
+// addWithRecursiveDependencies adds a requirement along with all its dependencies (according to the dependency map) to the requirements set
+void addWithRecursiveDependencies(
+    kj::StringPtr requirement, const DepMap &depMap, kj::HashSet<kj::String> &requirementsSet) {
+  auto normalizedName = canonicalizePythonPackageName(requirement);
+  if (requirementsSet.contains(normalizedName)) {
+    return;
+  }
+
+  requirementsSet.insert(kj::str(normalizedName));
```
jasnell, line 72: Are you certain there won't be conflicts? This should likely use upsert or findOrCreate

dom96: There is some code just above that checks for duplicates and early exits. This code is just being moved around, it already exists in EW.

But yeah, I suppose it should use findOrCreate to be more efficient.

dom96:
For now I've reverted back to using an Array, let's get this merged in and we can use a HashSet later if we agree on merging https://github.com/cloudflare/workerd/pull/3550.

src/workerd/api/pyodide/requirements.c++
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+kj::Own<capnp::List<capnp::json::Value::Field>::Reader> parseLockFile(
+    kj::StringPtr lockFileContents) {
+  capnp::JsonCodec json;
+  capnp::MallocMessageBuilder message;
+
+  auto lock = message.initRoot<capnp::JsonValue>();
+  json.decodeRaw(lockFileContents, lock);
```
hoodmane, line 87: Ideally it would be good to use a schema. But we can do that later if someone gets to it.

danlapid:
> For now I've reverted back to using an Array, let's get this merged in and we can use a HashSet later if we agree on merging https://github.com/cloudflare/workerd/pull/3550.

Wasn't following too closely but it seems like you do have agreement on merging that 🤔

hoodmane (approved):
Generally looks good to me, but I have one nit I'd like addressed.

PR #3539, opened 2025-02-13
Fixup the streaming tail worker implementation

jasnell:
* Fixes UAF in server.c++ handling
* Simplifies trace-stream implementation
* Ensures that the outcome event is sent in fetch request traces
* Still needs to ensure that outcome is sent in all other types of invocations
* Still needs tests, tests, and more tests.
  * Tests to verify that throwing an error in the tail worker handler works as expected
  * Tests to verify that all of the various ways a worker invocation can complete will appropriately trigger an outcome event
  * Tests to verify that all internal state is cleaned up and not leaked

@fhanau .... opening this as a draft and handing it over to you for next steps..

### DIFF
--- a/src/workerd/api/tail-worker-test.js
+++ b/src/workerd/api/tail-worker-test.js
@@ -1,0 +1,69 @@
+// Copyright (c) 2017-2023 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+import * as assert from 'node:assert';
+
+let resposeMap = new Map();
+
+export default {
+  // https://developers.cloudflare.com/workers/observability/logs/tail-workers/
+  tailStream(...args) {
+    // Onset event, must be singleton
+    resposeMap.set(args[0].traceId, JSON.stringify(args[0].event));
+    return (...args) => {
+      // TODO(streaming-tail-worker): Support several queued elements
+      let cons = resposeMap.get(args[0].traceId);
+      resposeMap.set(args[0].traceId, cons + JSON.stringify(args[0].event));
+    };
+  },
+};
+
+export const test = {
+  async test() {
+    // HACK: The prior tests terminates once the scheduled() invocation has returned a response, but
+    // propagating the outcome of the invocation may take longer. Wait briefly so this can go ahead.
+    await scheduler.wait(100);
+    // This test verifies that we're able to receive tail stream events for various handlers.
+
+    // Recorded streaming tail worker events, in insertion order.
+    let response = Array.from(resposeMap.values());
+
+    let expected = [
+      // http-test.js: fetch and scheduled events get reported correctly.
+      '{"type":"onset","info":{"type":"fetch","method":"POST","url":"http://placeholder/body-length","cfJson":"{}","headers":[]}}{"type":"outcome","outcome":"ok","cpuTime":0,"wallTime":0}',
+      '{"type":"onset","info":{"type":"fetch","method":"POST","url":"http://placeholder/body-length","cfJson":"{}","headers":[]}}{"type":"outcome","outcome":"ok","cpuTime":0,"wallTime":0}',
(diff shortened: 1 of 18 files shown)


src/workerd/io/trace-stream.c++
```diff
@@ -670,77 +698,134 @@ class TailStreamTargetBase: public rpc::TailStreamTarget::Server {
...
-          return LOG_STR;
-        }
-        KJ_CASE_ONEOF(_, tracing::Return) {
-          return RETURN_STR;
+
+    // Invoke the tailStream handler function.
+    v8::Local<v8::Function> fn = maybeFn.As<v8::Function>();
+    v8::Local<v8::Value> obj = ToJs(js, event, stringCache);
+
+    try {
```
fhanau, line 742: This try-catch block seems to fix the issue being tested for in `tail-worker-test-invalid.wd-test` – it seems like we fail to setStop() when the tail worker throws an exception. I doubt it's the best way to address it, so input is welcome.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -568,11 +589,15 @@ kj::Promise<WorkerInterface::AlarmResult> WorkerEntrypoint::runAlarmImpl(
   KJ_IF_SOME(t, incomingRequest->getWorkerTracer()) {
     t.setEventInfo(context.now(), tracing::AlarmEventInfo(scheduledTime));
   }
-  context.getMetrics().reportTailEvent(context, [&] {
+  context.getMetrics().reportTailEvent(context.getInvocationSpanContext(), [&] {
     return tracing::Onset(
         tracing::AlarmEventInfo(scheduledTime), tracing::Onset::WorkerInfo{}, kj::none);
   });
 
+  // TODO(streaming-tail): Ensure that lifetime of outcomeObserver is handled correctly.
```
fhanau, line 597: Feels like it's easy to get outcome lifetimes wrong, so those should be reviewed closely.

src/workerd/api/tail-worker-test.js
```diff
@@ -0,0 +1,69 @@
...
+  async test() {
+    // HACK: The prior tests terminates once the scheduled() invocation has returned a response, but
+    // propagating the outcome of the invocation may take longer. Wait briefly so this can go ahead.
+    await scheduler.wait(100);
+    // This test verifies that we're able to receive tail stream events for various handlers.
+
+    // Recorded streaming tail worker events, in insertion order.
+    let response = Array.from(resposeMap.values());
+
+    let expected = [
```
fhanau, line 31: This approach feels reasonably elegant – we test that onset/outcome events work for many different event types by reusing existing tests (much less overhead). By inserting into the Map based on the traceId, we keep all events sent to a given tail worker invocation together. There is some risk of race conditions affecting the order here, but since the individual workers are invoked by tests that are running sequentially this appears to work well in practice.

Note that this only checks the "event" portion of the TailEvent – the other components are less interesting/non-deterministic.

src/workerd/io/observer.h
```diff
@@ -160,6 +163,37 @@ class RequestObserver: public kj::Refcounted {
   }
 };
 
+// At the conclusion of an invocation we need to dispatch the outcome
+// event to the streaming tail worker (if any). We essentially have a race
+// between the completion of waitUntil tasks and the completion of the
+// deferred proxy (if any). The request is done whenever both of these are
+// finished. However, those end up being two separate (and unrelated) branches
+// of the promise tree set up by an invocation.
```
jasnell, line 171: FWIW, this comment is mostly relevant to just the fetch handler. Other handlers *might* have similar characteristics and all handlers do need to wait for `waitUntil` tasks but the exact placement of the outcome event varies from handler to handler. It might make sense to move this comment back into the fetch handler where this is used.

src/workerd/api/tail-worker-test.js
jasnell: Sounds reasonable to me

jasnell:
@fhanau ... once this passes internal CI, I'll do a final review on the parts you added/modified.

jasnell:
@fhanau ... I cannot sign off on my own PR but your updates LGTM.

Issue #1346, opened 2023-10-25
crypto: 100,000 iterations of PBKDF2 is insecure


ghost:
Hello,

`workerd` (and Cloudflare Workers) limits the number of `PBKDF2` iterations to 100,000 to avoid DoS.

We can see the code here: https://github.com/cloudflare/workerd/blob/d1a3c89716591753406923e177b54da3475dd01c/src/workerd/api/crypto-impl-pbkdf2.c%2B%2B#L45C31-L45C37

While I understand the reason, especially in a multi-tenant environment like Cloudflare Workers, 100,000 iterations is far from secure, as per [OWASP guidelines](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html), which, in 2023, recommends at least 210,000 iterations for `PBKDF2-SHA-512` and 600,000 iterations for `PBKDF2-SHA-256`, making any application hashing passwords from Cloudflare Workers insecure.

This is why I'm requesting an increase to at least 400,000 iterations.

Warfields:
@dom96

fhanau:
I'll take a look at this later today.

kentonv:
As discussed internally, moving the computation out of the isolate lock would add a lot of new complexity.

What we want to do here is permit a higher iteration count for people who have higher CPU limits. The original 100k limit was put in place when Workers had a 50ms total time limit -- that's about how much you could do in 50ms. Since our CPU time-limiting code cannot interrupt BoringSSL in the middle of running PBKDF, we have to limit the iterations upfront. But, these days most Workers have a 30s time limit so we should be able to increase the limit on PBKDF2 much higher for workers with such a higher limit.

Or if we can find an easy way to be able to interrupt the PBKDF2 loop when the CPU time limit is hit, then we could make the iterations unlimited...

sjc5:
Just ran into this unfortunately after converting a Node/Prisma app into Workers/D1 structure. I wish I had known of this limitation or I wouldn't have started down this path.

I know these kinds of comments can be a bit insufferable (sorry), but any idea on a timeline here or a soft commitment on this being raised? It's a dealbreaker and I would love to know if I just need to stop going down the Worker-based infra path for my use case.

Thanks much.

kentonv:
@irvinebroque how should we prioritize this?

irvinebroque:
We see a path to allow a higher iteration count, but this change won't happen this year.

jasnell:
[...]

PR #1471, opened 2023-12-04
Introduce configurable pbkdf2 iteration limit

jasnell:
Introduces the ability for workerd embedders to configure the max iteration limit for KDFs. Removes the limit for workerd by default. Keeps the current limit of 100,000 when not configured.

Refs: https://github.com/cloudflare/workerd/issues/1346

PR #3541, opened 2025-02-13
Increase PBKDF2 iterations to OWASP recommended value

Leon338:
This PR increases the PBKDF2 iterations to align with the OWASP recommended value for better security. The change is minimal and does not affect other parts of the codebase.

- Updated PBKDF2 iterations to 600'000
- Verified tests pass successfully
- Discussion reference: #1346 

This change follows security best practices as recommended by OWASP. Let me know if any further modifications are needed.

### DIFF
--- a/src/workerd/io/limit-enforcer.h
+++ b/src/workerd/io/limit-enforcer.h
@@ -12,7 +12,7 @@ namespace workerd {
 struct ActorCacheSharedLruOptions;
 class IoContext;
 
-static constexpr size_t DEFAULT_MAX_PBKDF2_ITERATIONS = 100'000;
+static constexpr size_t DEFAULT_MAX_PBKDF2_ITERATIONS = 600'000;
 
 // Interface for an object that enforces resource limits on an Isolate level.
 //


Leon338:
A similar pull request (#1471 ) has been attempting to address this issue since 2023 but has not been merged. That PR includes broader changes, while this one is a much smaller and focused fix that simply increases the PBKDF2 iterations to align with OWASP recommendations.

Since this is a critical security improvement, it would be great to get this merged independently while discussions on the larger PR continue.

Leon338:

Leon338:
I ran some benchmarks on Cloudflare Workers, and based on the cpuTimeMs observed in production, it should be possible to run PBKDF2 with 600,000 iterations in under 250ms on prod. Given these results, this implementation should not pose an issue for the live system

danlapid (changes requested):
This change is backwards incompatible as far as I can tell.

aperture147:
> I ran some benchmarks on Cloudflare Workers, and based on the cpuTimeMs observed in production, it should be possible to run PBKDF2 with 600,000 iterations in under 250ms on production. On average, this takes about 6× longer than running 100,000 iterations. Given these results, this implementation should not pose an issue for the live system.
> 
> Of course, this doesn’t work on Workers Free, but 100,000 iterations didn’t work there either.

In my experience Workers does not immediately terminate your code once it hits 10ms CPU (because I think implementing a timer and terminating the workers on every request is not efficient in the V8 environment), but it measures the aggregated data (like p50) and limit the invocation if the aggregated metric pass the 10ms limit. So theoretically you could still do this in Workers Free tier, as long as there are not much PBKDF2 job runs.

Of course when running in production, we still need Workers Paid anyway, but instead of putting the PBKDF code inside the main server code, I suggest to decouple it into a separated crypto worker and make the server call it via the Worker RPC. This has two benefits:

- Isolate the heavy number crunching outside of the main business, so the consumed CPU time is not counted in the main server worker, because the RPC wait time is not counted in the CPU time. Even when your PBKDF code does not work the logged in user could still do their job without being rate limited.
- You could set the CPU time limit separately. The crypto worker could have higher limit (like 500ms) and your server worker could keep the limit low (such as 50ms).
 
I'm implementing this in my production, setting server code CPU limit to 50ms and crypto code to 200ms and rarely hit the limit. You should put some kind of rate limit mechanism (like making use of built-in worker rate limit) in the server code to limit the crypto workers to be crunched and prevent the chance of blowing up your wallet.

PR #3543, opened 2025-02-13
Change PipelineTransform to PipelineTransformationEntrypoint

hhoughgg:
…plementation

### DIFF
--- a/src/cloudflare/internal/pipeline-transform.ts
+++ b/src/cloudflare/internal/pipeline-transform.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 // Copyright (c) 2024 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
@@ -57,26 +58,39 @@ enum Format {
   JSON_STREAM = 'json_stream', // jsonl
 }
 
-export class PipelineTransformImpl extends entrypoints.WorkerEntrypoint {
+type PipelineBatchMetadata = {
+  pipelineId: string;
+  pipelineName: string;
+};
+
+type PipelineRecord<T> = {
+  body: T;
+  readonly metadata: object;
+};
+
+export class PipelineTransformImpl<I, O> extends entrypoints.WorkerEntrypoint {
   #batch?: Batch;
   #initalized: boolean = false;
 
-  // stub overridden on the sub class
+  // stub overridden on the subclass
   // eslint-disable-next-line @typescript-eslint/require-await
-  public async transformJson(_data: object[]): Promise<object[]> {
+  public async run(
+    _records: PipelineRecord<I>[],
+    _metadata: PipelineBatchMetadata
+  ): Promise<PipelineRecord<O>[]> {
     throw new Error('should be implemented by parent');
   }
 
-  // called by the dispatcher which then calls the subclass methods
+  // called by the dispatcher to validate that run is properly implemented by the subclass
   // @ts-expect-error thinks ping is never used
   private _ping(): Promise<void> {
     // making sure the function was overridden by an implementing subclass
-    if (this.transformJson !== PipelineTransformImpl.prototype.transformJson) {
+    if (this.run !== PipelineTransformImpl.prototype.run) {
       return Promise.resolve();
     } else {
       return Promise.reject(
         new Error(
-          'the transformJson method must be overridden by the PipelineTransform subclass'
+          'the run method must be overridden by the PipelineTransformationEntrypoint subclass'
         )
       );
     }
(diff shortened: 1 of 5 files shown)


src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -57,26 +58,39 @@ enum Format {
...
 }
 
-export class PipelineTransformImpl extends entrypoints.WorkerEntrypoint {
+type PipelineBatchMetadata = {
+  pipelineId: string;
+  pipelineName: string;
+};
+
+type PipelineRecord<T> = {
+  body: T;
```
cmackenzie1, line 67: We can probably drop the generic here for `Record<string,unknown>` (which is a tighter type than `object` since `Array` is also `object`. 

```suggestion
  body: Record<string, unknown>;
```

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -57,26 +58,39 @@ enum Format {
...
+  readonly metadata: object;
```
cmackenzie1, line 68: Maybe I forgot the discussion on this, but what are we adding metadata for each record here?

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -85,55 +99,63 @@ export class PipelineTransformImpl extends entrypoints.WorkerEntrypoint {
...
-        const data = await this.#readJsonStream();
-        const transformed = await this.transformJson(data);
-        return this.#sendJson(transformed);
-      }
-      default:
-        throw new Error('unsupported batch format');
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (this.#batch.format === Format.JSON_STREAM) {
+      const data = await this.#readJsonStream();
+      const records: PipelineRecord<I>[] = data.map((obj) => {
```
cmackenzie1, line 117: We can remove this if we drop `metadata` from each record.

types/defines/pipelines.d.ts
```diff
@@ -2,12 +2,13 @@
...
   /**
-   * send takes an array of javascript objects which are
-   * then received by the pipeline for processing
+   * The Pipeline interface represents the type of a binding to a Pipeline
    *
-   * @param data The data to be sent
+   * @param records The records to send to the pipeline
    */
-  send(data: object[]): Promise<void>
+  send(records: PipelineRecord<T>[]): Promise<void>
```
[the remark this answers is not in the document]
cmackenzie1: Type should be `Record<string, unknown>` in my opinion since `object` also allows `[]`

 ```js
> typeof []
'object'
```

**Review by cmackenzie1 — CHANGES_REQUESTED** [review]

File: src/cloudflare/internal/pipeline-transform.ts (hunk 2 above)
Comment (maheshwarip) [reply src/cloudflare/internal/pipeline-transform.ts:68]: IMO metadata shouldn't be here. 

We might add other fields here in the future (eg: `eventReceivedTimestamp`). So, PipelineRecord should be an object, with the body being the payload of data that the user had originally sent to the Pipeline. 

`metadata` inside the record feels wrong, especially since we're passing metadata about the batch of records in the metadata argument.

File: types/defines/pipelines.d.ts [hunk 5]
```diff
@@ -2,12 +2,11 @@
...
   /**
-   * send takes an array of javascript objects which are
-   * then received by the pipeline for processing
+   * The Pipeline interface represents the type of a binding to a Pipeline
    *
-   * @param data The data to be sent
+   * @param records The records to send to the pipeline
    */
-  send(data: object[]): Promise<void>
+  send(records: T[]): Promise<void>
```
cmackenzie1, line 11: We can drop the generic and probably do `PipelineRecord` here?

```suggestion
  send(records: PipelineRecord[]): Promise<void>
```

oliy: Should probably be `Pipeline<T extends PipelineRecord>`.  Seems like we want a way to allow the end implementor to enforce type safety in their code.  So if they specify for their environment:

```
interface MyEnv {
   MYPIPE: Pipeline<MyRecord>
}
```

It will enforce the shape of the send method at the call site:

```
  env.MYPIPE.send([{...}])
```

oliy (approved):

maheshwarip:
lgtm, excited for this :)

cmackenzie1 (approved):

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -57,26 +58,39 @@ enum Format {
   JSON_STREAM = 'json_stream', // jsonl
 }
 
-export class PipelineTransformImpl extends entrypoints.WorkerEntrypoint {
+type PipelineBatchMetadata = {
+  pipelineId: string;
+  pipelineName: string;
+};
+
+type PipelineRecord = Record<string, unknown>;
```
a-robinson, line 66: I assume our service is responsible for ensuring all records align with this specific format (i.e. of being an object with strings mapped to arbitrary values)? Because the JSON.parse() call in `readJsonStream` below won't guarantee that IIUC.

a-robinson (approved):

src/cloudflare/internal/pipeline-transform.ts
hhoughgg: Yeah exactly. If there is a problem here we need to address in the internal service.

a-robinson:
@hhoughgg are you able to view the failing CI checks and fix them up? I can merge for you once they're green, but given the number of failures it looks like some fixing is needed.

a-robinson:
Looks like there's [still an issue](https://github.com/cloudflare/workerd/actions/runs/13445430952/job/37569574210?pr=3543)?

```
Run diff -r bazel-bin/types/definitions/ types/generated-snapshot > types.diff
  diff -r bazel-bin/types/definitions/ types/generated-snapshot > types.diff
  shell: /usr/bin/bash -e {0}
Error: Process completed with exit code 1.
```

PR #3545, opened 2025-02-14
Implement delete operation for memory cache.

mar-cf:
Add functionality to remove entries from memory cache.

### DIFF
--- a/src/workerd/api/memory-cache.c++
+++ b/src/workerd/api/memory-cache.c++
@@ -452,6 +452,7 @@ jsg::Promise<jsg::JsRef<jsg::JsValue>> MemoryCache::read(jsg::Lock& js,
 void MemoryCache::delete_(jsg::Lock& js, jsg::NonCoercible<kj::String> key) {
   // Ignore operations on keys exceeding key max size.
   if (key.value.size() > MAX_KEY_SIZE) {
+    js.throwException(js.rangeError("Key too large."_kj));
     return;
   }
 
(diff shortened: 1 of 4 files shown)


src/workerd/api/memory-cache.c++
```diff
@@ -444,6 +449,15 @@ jsg::Promise<jsg::JsRef<jsg::JsValue>> MemoryCache::read(jsg::Lock& js,
   }
 }
 
+void MemoryCache::delete_(jsg::Lock& js, jsg::NonCoercible<kj::String> key) {
+  // Ignore operations on keys exceeding key max size.
+  if (key.value.size() > MAX_KEY_SIZE) {
+    return;
```
danlapid, line 455: We can still throw a jsg.Error exception here that will be translated to a JS error to indicate that we ignored the operation.
A silent ignore seems wrong to me.

danlapid:
We probably want to add tracing to this, see ref in `readSpan`, `userReadSpan`

kentonv:
Is this actually useful? Being able to purge from _all_ caches across the network could be useful, but would obviously be much more complicated to implement. If you're only deleting from one cache, how useful can it actually be, considering there's no way to target requests at a specific cache?

jasnell (changes requested):
Let's please only add this under a separate / additional experimental compat flag. We need to soon make MemoryCache public API, lifting the compat flag for it and it is not clear that we will (ever?) want the delete exposed as part of the regular user-facing API.

src/workerd/api/memory-cache.c++
mar-cf: Neither seem wrong. Throwing is defensive. However, it's not possible to insert an oversized key, so it wouldn't need to be deleted.

Changed to throw.

jasnell (approved):

PR #3548, opened 2025-02-14
skip all unnecessary urlpattern tests

anonrig:
URLPattern.compareComponent is not part of the spec and we shouldn't run WPT on them.

This actually points out a bug in our web-platform tests runner: `'urlpattern-compare.tentative.https.any.js': {},` this shouldn't succeed. This means that the first 2 lines in WPT is not followed. @npaun Can you take a look at this?

----

PS: the first 2 lines are:

```
// META: global=window,worker
// META: script=resources/urlpattern-compare-tests.tentative.js
```

### DIFF
--- a/src/workerd/api/wpt/urlpattern-test.ts
+++ b/src/workerd/api/wpt/urlpattern-test.ts
@@ -6,39 +6,17 @@ import { type TestRunnerConfig } from 'wpt:harness';
 
 export default {
   'urlpattern-compare-tests.tentative.js': {
-    comment: 'urlpattern implementation will soon be replaced with ada-url',
-    expectedFailures: [
-      // Each of these *ought* to pass. They are included here because we
-      // know they currently do not. Each needs to be investigated.
-      'Component: pathname Left: {"pathname":"/foo/a"} Right: {"pathname":"/foo/b"}',
-      'Component: pathname Left: {"pathname":"/foo/b"} Right: {"pathname":"/foo/bar"}',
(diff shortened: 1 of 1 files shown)


danlapid (approved):

npaun:
@anonrig `urlpattern-compare.tentative.https.any.js` gets treated as containing zero tests, but we process `urlpattern-compare-tests.tentative.js` and work with the tests within it. It is a bit clunky, but I feel like actually parsing the JS files and figuring out that they're empty isn't worth it.

npaun (approved):

anonrig:
> @anonrig `urlpattern-compare.tentative.https.any.js` gets treated as containing zero tests, but we process `urlpattern-compare-tests.tentative.js` and work with the tests within it. It is a bit clunky, but I feel like actually parsing the JS files and figuring out that they're empty isn't worth it.

@npaun 

There are tests that depend on this syntax in order to execute. For example: idl_tests. We should support running them: https://web-platform-tests.org/writing-tests/idlharness.html

npaun:
@anonrig Until we need it for a specific test we want, maybe I could just make the harness reject any files that seem like they contain no tests? 

The format is not too hard to parse but it's currently not going to help us for any test we have right now, I feel. 

Let me know what you think.

anonrig:
> Until we need it for a specific test we want, maybe I could just make the harness reject any files that seem like they contain no tests?

I think we shouldn't hide them, but eventually support them if we want a higher spec compliance. You never know how impactful they are without going over them(which can also change at anytime)

I think the test runner shouldn't include all tests but rather follow them through the meta descriptions, just like the actual WPT does. This would allow us to easily rely on the WPT.

PR #3550, opened 2025-02-14
Implements kj::HashSet support for jsg.

dom96:
### Test Plan

```
bazel run @workerd//src/workerd/jsg:value-test
```

### DIFF
--- a/src/workerd/jsg/value-test.c++
+++ b/src/workerd/jsg/value-test.c++
@@ -908,6 +908,37 @@ KJ_TEST("Array Values") {
 
 // ========================================================================================
 
+struct SetContext: public ContextGlobalObject {
+  kj::HashSet<kj::String> takeSet(kj::HashSet<kj::String> set) {
+    KJ_ASSERT(set.contains("42"_kj));
+    return kj::mv(set);
+  }
+  kj::HashSet<kj::String> returnStrings(int i, int j, int k) {
+    auto result = kj::HashSet<kj::String>();
+    result.insert(kj::str(i));
+    result.insert(kj::str(j));
+    result.insert(kj::str(k));
+    return kj::mv(result);
+  }
+  JSG_RESOURCE_TYPE(SetContext) {
+    JSG_METHOD(takeSet);
+    JSG_METHOD(returnStrings);
+  }
+};
+JSG_DECLARE_ISOLATE_TYPE(SetIsolate, SetContext);
+
+KJ_TEST("Set Values") {
+  Evaluator<SetContext, SetIsolate> e(v8System);
+  e.expectEval("m = new Set(); m.add('42'); takeSet(m).has('42')", "boolean", "true");
+  e.expectEval(
+      "const toString = () => '42'; takeSet(new Set([{ toString }, { toString }])).has('42')",
+      "throws", "TypeError: Duplicate values in the set after unwrapping.");
+
+  e.expectEval("returnStrings(123, 1024, 456).has('1024')", "boolean", "true");
+}
+
+// ========================================================================================
+
 struct SequenceContext: public ContextGlobalObject {
   Sequence<kj::String> testSequence(Sequence<kj::String> sequence) {
     KJ_ASSERT(sequence.size() == 2);
--- a/src/workerd/jsg/value.h
+++ b/src/workerd/jsg/value.h
@@ -852,6 +852,62 @@ class ArrayWrapper {
 };
 
 // =======================================================================================
+// Sets
+
+// TypeWrapper mixin for sets.
+template <typename TypeWrapper>
+class SetWrapper {
+ public:
+  static auto constexpr MAX_STACK = 64;
+  template <typename U>
+  static constexpr const char* getName(kj::HashSet<U>*) {
+    return "Set";
+  }
+
+  template <typename U>
+  v8::Local<v8::Value> wrap(v8::Local<v8::Context> context,
+      kj::Maybe<v8::Local<v8::Object>> creator,
+      kj::HashSet<U> set) {
+    v8::Isolate* isolate = context->GetIsolate();
+    v8::EscapableHandleScope handleScope(isolate);
+
+    auto out = v8::Set::New(isolate);
+    for (auto& item: set) {
+      v8::HandleScope scope(isolate);
+      check(
+          out->Add(context, static_cast<TypeWrapper*>(this)->wrap(context, creator, kj::mv(item))));
+    }
+
+    return handleScope.Escape(out);
+  }
+
+  template <typename U>
+  kj::Maybe<kj::HashSet<U>> tryUnwrap(v8::Local<v8::Context> context,
+      v8::Local<v8::Value> handle,
+      kj::HashSet<U>*,
+      kj::Maybe<v8::Local<v8::Object>> parentObject) {
+    if (!handle->IsSet()) {
+      return kj::none;
+    }
+
+    auto set = handle.As<v8::Set>();
+    auto array = set->AsArray();
+    auto length = array->Length();
+    auto builder = kj::HashSet<U>();
+    builder.reserve(length);
+    for (auto i: kj::zeroTo(length)) {
+      v8::Local<v8::Value> element = check(array->Get(context, i));
+      auto value = static_cast<TypeWrapper*>(this)->template unwrap<U>(
+          context, element, TypeErrorContext::other());
+      builder.upsert(kj::mv(value), [&](U& existing, U&& replacement) {
+        JSG_FAIL_REQUIRE(TypeError, "Duplicate values in the set after unwrapping.");
+      });
+    }
+    return kj::mv(builder);
+  }
+};
+
+// =======================================================================================
 // ArrayBuffers / ArrayBufferViews
 //
 // This wrapper implements the following wrapping conversions:
--- a/src/workerd/jsg/README.md
+++ b/src/workerd/jsg/README.md
@@ -170,6 +170,21 @@ void doSomething(kj::Array<kj::String> strings) {
 doSomething(['a', 'b', 'c']);
 ```
 
+### Set type (`kj::HashSet<T>`)
+
+The `kj::HashSet<T>` type maps to JavaScript sets. The `T` can be any value, though there are
+currently some restrictions. For example, you cannot have a `kj::HashSet<CustomStruct>`.
+
+```cpp
+void doSomething(kj::HashSet<kj::String> strings) {
+  KJ_DBG(strings.has("a"));
+}
+```
+
+```js
+doSomething(new Set(['a', 'b', 'c']));
+```
+
 ### Sequence types (`jsg::Sequence<T>`)
 
 A [Sequence][] is any JavaScript object that implements `Symbol.iterator`. These are
--- a/src/workerd/jsg/rtti.h
+++ b/src/workerd/jsg/rtti.h
@@ -379,6 +379,7 @@ FOR_EACH_MAYBE_TYPE(DECLARE_MAYBE_TYPE)
 #define FOR_EACH_ARRAY_TYPE(F)                                                                     \
   F(kj::Array)                                                                                     \
   F(kj::ArrayPtr)                                                                                  \
+  F(kj::HashSet)                                                                                   \
   F(jsg::Sequence)                                                                                 \
   F(jsg::AsyncGenerator)
 
--- a/src/workerd/jsg/type-wrapper.h
+++ b/src/workerd/jsg/type-wrapper.h
@@ -378,6 +378,7 @@ class TypeWrapper: public DynamicResourceTypeMap<Self>,
                    public MaybeWrapper<Self>,
                    public OneOfWrapper<Self>,
                    public ArrayWrapper<Self>,
+                   public SetWrapper<Self>,
                    public SequenceWrapper<Self>,
                    public GeneratorWrapper<Self>,
                    public ArrayBufferWrapper<Self>,
@@ -435,6 +436,7 @@ class TypeWrapper: public DynamicResourceTypeMap<Self>,
   USING_WRAPPER(MaybeWrapper<Self>);
   USING_WRAPPER(OneOfWrapper<Self>);
   USING_WRAPPER(ArrayWrapper<Self>);
+  USING_WRAPPER(SetWrapper<Self>);
   USING_WRAPPER(SequenceWrapper<Self>);
   USING_WRAPPER(GeneratorWrapper<Self>);
   USING_WRAPPER(ArrayBufferWrapper<Self>);


anonrig (approved):

kentonv (changes requested):
What's the intended use case here?

IMO `HashSet` is not the right C++ representation to correspond to a JS `Set`. The problem is, in order to convert between JS `Set` and `HashSet`, you necessarily have to enumerate the whole set, which is O(n), which probably defeats the purpose of using a Set vs. just using an array or something.

Probably all use cases fall into one of two categories:
- The C++ consumer is going to iterate the whole set. In this case, building the hash index is a waste of time. It would be better to simply deliver an array.
- The C++ consumer only wants to look up certain values in the set. In this case, what is really wanted is some sort of a wrapper that does _not_ iterate the whole set upfront, but instead looks up specific values on-demand.

Or conversely for the C++ -> JS direction:
- The C++ producer is going to construct a one-off set to pass to JS. There's again no reason to construct a hash index, the C++ representation should be an array or vector.
- The C++ producer is passing some long-lived set to JS. In this case, JS should really get a view into the set that doesn't convert the contents until requested, otherwise it's wasteful to re-convert the whole set every time it is requested.

dom96:
The use case is https://github.com/cloudflare/workerd/pull/3538#discussion_r1954509060 where a C++ function is producing a HashSet and the JS side also wants a Set. My initial solution was to convert the set to an array and return that from the C++ side, then convert to a Set in JS. Dan suggested that jsg should support this so I implemented it.

kentonv:
Hmm, I see, this does appear to be an unusual use case where you:
- Built a one-off `kj::HashSet` on the C++ side.
- But you actually did need the hash index, because the code to build it is walking a DAG and needs to detect duplicates.
- And then you actually want to pass this HashSet off to JS as a Set.

So converting a HashSet to a JS Set is actually what you want here, and there's no wasted index-building.

In the future we might want to add an alternative representation like:

```
template <typename T>
struct JsSet: public kj::Vector<T> {};
```

Kind of like how `jsg::Optional<T>` is an alias for `kj::Maybe<T>` that instructs JSG on how to represent it to JS.

But I guess we could _also_ support `HashSet<T>`, and you use whichever one is best for the use case. I just worry about people using `HashSet` "because it makes sense" without thinking through problems with wasted indexing or differing equality functions...

kentonv:
I think we still do need an answer for the duplicate question when converting a Set to a HashSet. Possible answers are:

* First instance wins.
* Last instance wins.
* Detect and throw a JSG error instead of "internal error".
* Don't implement JS Set -> kj::HashSet for now. Your use case only needs the opposite direction anyway, it's not necessary to provide both.

jasnell:
> ... The use case is https://github.com/cloudflare/workerd/pull/3538#discussion_r1954509060 where a C++ function is producing a HashSet and the JS side also wants a Set.

It does seem rather unfortunate that what this conversion pattern ends up with is:

1. Iterate through a collection to construct the kj::HashSet (not entirely low-cost due to the maintenance of the hash index)...
2. Create the JS Set
3. Iterate through the kj::HashSet to add items to the JS Set which has its own costs for its own internal hash index

It would seem to be more efficient to skip the creation of the kj::HashSet entirely if possible and just have the C++ code just directly create the v8::Set (wrapped as a `jsg::JsSet`) and return that directly.

Is the `kj::HashSet` itself being used for other things?

dom96:
> Is the kj::HashSet itself being used for other things?

Yeah, the function is also called from C++ in edgeworker to pre-load the correct Python packages.

dom96:
Okay, I fixed the duplicate issue and added a test for it.

src/workerd/jsg/value.h
```diff
@@ -851,6 +851,67 @@ class ArrayWrapper {
...
+
+    auto set = handle.As<v8::Set>();
+    auto array = set->AsArray();
+    auto length = array->Length();
+    auto builder = kj::HashSet<U>();
+    for (auto i: kj::zeroTo(length)) {
+      v8::Local<v8::Value> element = check(array->Get(context, i));
+      auto value = static_cast<TypeWrapper*>(this)->template unwrap<U>(
+          context, element, TypeErrorContext::arrayElement(i));
+      if (!builder.contains(value)) {
```
kentonv, line 907: Please use either `findOrCreate()` or `upsert()` here so that you're not doing the same hash lookup twice in a row, once for `contains()` and one for `insert()`.

kentonv, line 907: I guess this implements "first element wins" behavior, which is _an_ option but I wonder if it's the right option. It's hard to say for sure without knowing what the future use case will be but I sort of think an exception might be better?

You can implement that by using `upsert()` and then throwing from the update callback...

jasnell: Honestly I would suggest an exception is better here. Silently choosing the first or the last and causing another entry to be dropped just smells like it will lead to subtle bugs at some point.

jasnell (approved):
I'm not entirely sold on this just yet. If the hashsets are small and the objects in the set are simple enough then it's likely to be ok but there's likely a number of performance issues associated with model.

Consider, for instance, if I have a kj::HashSet<kj::String>. Let's say those strings end up being fairly non-trivial in size. To get that hash set out to JavaScript would mean multiple allocations and copies for each string in addition to iterating over the entire set.

```
kj::HashSet<kj::String> set;
set.add(kj::str("here is a string"));  // allocation
set.add(kj::str("here is another string"));  // allocation
```

Converting this to a JS set means we perform an *additional* allocation in the v8 heap and do a utf8 write to copy the strings in addition to v8 having to calculate the hashes of those for it's own internal hash index. That just feels super wasteful at scale when the C++ code could just *start* with the JS Set in the first place.

kentonv:
@jasnell you are raising the same objections I did initially, but then Dominik pointed to his use case and indeed in that use case this conversion makes sense. He hash a `HashSet<String>` which he legitimately had to build as a `HashSet` (he couldn't have built it as a `Vector`), and he wants it to become a JS Set.

It's debatable whether that is a common enough case to build into JSG, but it does check out in this case.

jasnell:
Ok well, I won't block it but I'm still not convinced.

src/workerd/jsg/value-test.c++
```diff
@@ -908,6 +908,37 @@ KJ_TEST("Array Values") {
 
 // ========================================================================================
 
+struct SetContext: public ContextGlobalObject {
+  kj::HashSet<kj::String> takeSet(kj::HashSet<kj::String> set) {
+    KJ_ASSERT(set.contains("42"_kj));
+    return kj::mv(set);
+  }
```
jasnell, line 915: Hmm... there appear to be some issues with using some values of `T` with `kj::HashSet<T>` still with this... for example... if I modify this test to add a new `JSG_STRUCT` `Foo` that works just fine with `kj::HashSet<Foo>` and try to return it (see the `makeSet()` call below, then `value-test.c++` does not compile successfully.

```
struct SetContext: public ContextGlobalObject {
  struct Foo {
    kj::String a;
    JSG_STRUCT(a);
    uint hashCode() const { return kj::hashCode(a); }
    bool operator==(const Foo& other) const { return a == other.a; }
  };

  kj::HashSet<Foo> makeSet() {
    auto result = kj::HashSet<Foo>();
    result.insert(Foo{kj::str("42")});
    auto check = Foo { kj::str("42") };
    KJ_DBG(result.find(check) != kj::none);
    return result;
  }

  JSG_RESOURCE_TYPE(SetContext) {
    JSG_METHOD(makeSet);
  }
};
JSG_DECLARE_ISOLATE_TYPE(SetIsolate, SetContext, SetContext::Foo);

KJ_TEST("Set Values") {
  Evaluator<SetContext, SetIsolate> e(v8System);
  e.expectEval("const test = makeSet(); test instanceof Set", "boolean", "true");
}
```

I think it's fine if the type mapping does not work for certain kinds of types so I'm not saying this necessarily needs to be fixed, but the limitations should be documented at least

src/workerd/jsg/value.h
```diff
@@ -851,6 +851,67 @@ class ArrayWrapper {
...
+  v8::Local<v8::Value> wrap(v8::Local<v8::Context> context,
+      kj::Maybe<v8::Local<v8::Object>> creator,
+      kj::HashSet<U> set) {
+    v8::Isolate* isolate = context->GetIsolate();
+    v8::EscapableHandleScope handleScope(isolate);
+
+    auto out = v8::Set::New(isolate);
+    for (const auto& item: set) {
+      out = check(out->Add(context,
+          static_cast<TypeWrapper*>(this)->wrap(context, creator, item).template As<v8::Value>()));
```
jasnell, line 877: Should this actually `kj::mv(item)`?

src/workerd/jsg/value.h
dom96: I don't have strong feelings here, but I opted for "first element wins" because it seemed least surprising to me. That said, I suppose jsg conversions can fail for all sorts of reasons so it does make sense to throw in this context and it's better to be loud about it.

src/workerd/jsg/value.h
dom96: hm, do you mean that I should pass an owned `U` to `wrap`? I don't think I can do that from a HashSet, certainly in the case of kj::String it doesn't work. Isn't passing a ref fine here?

dom96:
@kentonv gentle ping for another review

src/workerd/jsg/value.h
kentonv: > certainly in the case of kj::String it doesn't work

I think it only doesn't work because you declared `item` as `const auto&` instead of `auto&`. If you change that it should work fine. It will leave the hash index inconsistent, but that doesn't matter since it isn't used again before the HashSet is destroyed at the end of the function.

> Isn't passing a ref fine here?

Depends on the type. It may be that the wrapper for `U` requires that you pass ownership, or is more efficient if you pass ownership (e.g. `Array<byte>`). For `String` it doesn't matter since the string has to be translated to UTF-16 anyway when passed to JavaScript, so a copy is unavoidable.

But since you've taken ownership of the `HashSet` from the caller, it would be appropriate to pass away ownership of all the elements, since they're otherwise going to be destroyed anyway.

(Ironically, `HashSet<HashSet<T>>` wouldn't work because the wrapper for `HashSet` itself (this function) requires transferring ownership.)

src/workerd/jsg/value.h
```diff
@@ -851,6 +851,67 @@ class ArrayWrapper {
...
+    }
+
+    auto set = handle.As<v8::Set>();
+    auto array = set->AsArray();
+    auto length = array->Length();
+    auto builder = kj::HashSet<U>();
+    for (auto i: kj::zeroTo(length)) {
+      v8::Local<v8::Value> element = check(array->Get(context, i));
+      auto value = static_cast<TypeWrapper*>(this)->template unwrap<U>(
+          context, element, TypeErrorContext::arrayElement(i));
```
kentonv, line 906: Passing `arrayElement(i)` as the error context may be weird because the caller doesn't perceive that they're passing an array in the first place... they are passing a `Set`.

src/workerd/jsg/value.h
```diff
@@ -851,6 +851,67 @@ class ArrayWrapper {
...
+      kj::HashSet<U>*,
+      kj::Maybe<v8::Local<v8::Object>> parentObject) {
+    if (!handle->IsSet()) {
+      return kj::none;
+    }
+
+    auto set = handle.As<v8::Set>();
+    auto array = set->AsArray();
+    auto length = array->Length();
+    auto builder = kj::HashSet<U>();
```
kentonv, line 902: Might be worth adding:

```
builder.reserve(length);
```

src/workerd/jsg/value.h
jasnell: @dom96 fwiw, I believe this is why https://github.com/cloudflare/workerd/pull/3550#discussion_r1956777211 does not work.

src/workerd/jsg/value.h
```diff
@@ -851,6 +851,67 @@ class ArrayWrapper {
...
+  template <typename U>
+  v8::Local<v8::Value> wrap(v8::Local<v8::Context> context,
+      kj::Maybe<v8::Local<v8::Object>> creator,
+      kj::HashSet<U> set) {
+    v8::Isolate* isolate = context->GetIsolate();
+    v8::EscapableHandleScope handleScope(isolate);
+
+    auto out = v8::Set::New(isolate);
+    for (const auto& item: set) {
+      out = check(out->Add(context,
```
kentonv, line 876: Looks like `Add()` returns a new handle pointing at the same set. You don't really need to assign `out`.

In fact you should probably add a `v8::HandleScope` inside the loop here so that these handles don't accumulate when converting a large set.

src/workerd/jsg/value.h
```diff
@@ -851,6 +851,67 @@ class ArrayWrapper {
...
+    }
+
+    return handleScope.Escape(out);
+  }
+
+  template <typename U>
+  v8::Local<v8::Value> wrap(v8::Local<v8::Context> context,
+      kj::Maybe<v8::Local<v8::Object>> creator,
+      kj::HashSet<U>& set) {
+    return static_cast<TypeWrapper*>(this)->wrap(context, creator, set.asPtr());
```
kentonv, line 887: Does this overload actually work? I don't think `set` has an `asPtr()` method?

src/workerd/jsg/value.h
jasnell:
```
js.withinHandleScope([&] {
  //...
});
```

kentonv: since we're deep in JSG here, and there is no `jsg::Lock` in the local scope already, I'd just use `v8::HandleScope` directly.

jasnell: Fair :-)

src/workerd/jsg/value.h
dom96: Cool, that works, thanks for the help!

src/workerd/jsg/value.h
dom96: Hm, maybe not, I'll just remove this overload.

src/workerd/jsg/value-test.c++
```diff
@@ -908,6 +908,37 @@ KJ_TEST("Array Values") {
...
+    JSG_METHOD(returnStrings);
+  }
+};
+JSG_DECLARE_ISOLATE_TYPE(SetIsolate, SetContext);
+
+KJ_TEST("Set Values") {
+  Evaluator<SetContext, SetIsolate> e(v8System);
+  e.expectEval("m = new Set(); m.add('42'); takeSet(m).has('42')", "boolean", "true");
+  e.expectEval(
+      "m = new Set(); m.add({ toString: () => '42' }); m.add({ toString: () => '42' }); takeSet(m).has('42')",
```
jasnell, line 934: fwiw, I think you can simplify this slightly to be just...

```js
const toString = () => '42'; takeSet(new Set([{ toString }, { toString }]));
```

jasnell (approved):
Still not overly excited about it but code LGTM

src/workerd/jsg/value.h
```diff
@@ -851,6 +851,64 @@ class ArrayWrapper {
...
+    v8::Isolate* isolate = context->GetIsolate();
+    v8::EscapableHandleScope handleScope(isolate);
+
+    auto out = v8::Set::New(isolate);
+    for (auto& item: set) {
+      v8::HandleScope scope(isolate);
+      check(out->Add(context,
+          static_cast<TypeWrapper*>(this)
+              ->wrap(context, creator, kj::mv(item))
+              .template As<v8::Value>()));
```
kentonv, line 880: Hmm, this `As<v8::Value>` looks redundant. Doesn't `wrap()` always return a `v8::Local`, and since `v8::Value` is the base type of all values, it should implicitly convert?

I only ask because `handle.As<T>()` is a bit of an assertion -- I believe it's UB if the value isn't actually of type `T`. It makes me uncomfortable to see it without some sort of an explicit check nearby showing that it's safe.

What is the error message you get without this?

kentonv (approved):
Approving because my comment is probably just paranoid and there's no real problem, but I am still interested in the answer.

src/workerd/jsg/value.h
dom96: well, it seems to work without it too, so I'll remove it. I based this code on the ArrayWrapper above and it does the same thing, is it redundant in that code too? https://github.com/cloudflare/workerd/blob/main/src/workerd/jsg/value.h#L852

kentonv: Weird, yeah, that looks redundant too.

PR #3551, opened 2025-02-14
add new recipes to justfile

anonrig:
Add new recipes to justfile

### DIFF
--- a/justfile
+++ b/justfile
@@ -21,7 +21,7 @@ build *args="//...":
   bazel build {{args}}
 
 build-asan *args="//...":
-  just build {{args}} --config=asan --sandbox_debug
+  just build {{args}} --config=asan
 
 test *args="//...":
   bazel test {{args}}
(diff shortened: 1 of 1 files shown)


npaun (approved):

justfile
```diff
@@ -37,6 +37,16 @@ stream-test args:
 node-test test_name:
   just stream-test //src/workerd/api/node:tests/{{test_name}}-nodejs-test
 
+# e.g. just wpt-test urlpattern
+wpt-test test_name:
+  just stream-test //src/workerd/api/wpt:{{test_name}}
+
+lldb-wpt-test test_name: build
+  cd bazel-bin/src/workerd/api/wpt/{{test_name}}.runfiles/workerd/src/workerd/api/wpt; lldb ../../../workerd/server/workerd  -- test {{test_name}}.wd-test --experimental --directory-path=TEST_TMPDIR=/tmp
```
ketanhwr, line 45: [nit] `working-directory` (https://just.systems/man/en/working-directory.html) might be a better option than `cd`

anonrig: nice catch! i love it. thank you for the review.

jasnell (approved):

npaun:
...

justfile
npaun: Unfortunately, it doesn't seem like you can use arguments like `{{test_name}}` with the working-directory directive.

npaun:
🦥 This CI sponsored by sloths.

anonrig:
@danlapid @fhanau I think we have a bug. This PR does not trigger any test CI because those CI jobs are ignored on justfile changes, but still it is required to land it...

PR #3552, opened 2025-02-14
do not trigger ci if you don't need it

anonrig:
let's limit our scope for triggering ci

### DIFF
--- a/.github/workflows/internal-build.yml
+++ b/.github/workflows/internal-build.yml
@@ -8,6 +8,8 @@ on:
       - '.devcontainer'
       - '**/*.md'
       - '.gitignore'
+      - '.github/*'
+      - '!.github/workflows/internal-build.yml'
 
 concurrency:
   # Cancel existing builds for the same PR.
--- a/.github/workflows/internal-build.yml
+++ b/.github/workflows/internal-build.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     paths-ignore:
       - 'doc/**'
+      - 'justfile'
+      - '.devcontainer'
+      - '**/*.md'
+      - '.gitignore'
 
 concurrency:
   # Cancel existing builds for the same PR.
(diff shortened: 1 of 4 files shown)


.github/workflows/internal-build.yml
```diff
@@ -2,8 +2,15 @@ name: Run internal build
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
fhanau, line 5: What PR actions exactly are we excluding here? The commit message suggests that it's draft PRs but are there more? I find it useful to have CI run on draft PRs – when I open a PR with experimental changes I might not have run all tests locally and I have no idea if it works on Windows, I'd want to know if CI passes before marking the PR as ready and having reviewers look at it.

anonrig: My goal was to remove draft PRs. Personally, whenever I open a draft PR I expect it to not work and just open to share the code with someone, but it's open to discussion I guess. If you prefer this way, I don't have any strong opinions and just remove this commit.

fhanau: Yeah, I think that would be better. As a specific example, the Bazel 8 update PR is in the draft stage, I've been able to push to it and see Windows CI fail a number of times there. If I had to make it a non-draft PR to see if CI passes, that would cause lots of spam for reviewers and falsely suggest that it is ready for review.

danlapid: Strong agree with Felix here, draft PRs should have CI running by default.

anonrig: Removed this change.

fhanau (approved):
Changes LGTM. Also noticed today that `doc/**` needs to be `docs/**`  instead, maybe we can fit that in here.

anonrig:
It was on auto-merge. We can fix it on a new PR

mikea:
This makes it impossible to submit doc-only change, which imo defeats the purpose of this change to land such PRs faster?

I suggest reverting this, since this is not the first time we hit it. I don't think protected branch configuration is flexible enough to conditionally require builds, is it @danlapid ?

@fhanau

PR #3554, opened 2025-02-14
Fix some WPT url tests failing due to invalid surrogates in UTF8

npaun:
* Adds jsg::USVString, which is guaranteed not to contain invalid characters in UTF8. This type is intended to make it convenient and consistent to implement Web APIs which say they take a USVString without having to deal with this issue every time.
* Updates url-standard to use USVStrings as specified by the WHATWG URL spec.

### DIFF
--- a/src/workerd/api/url-standard.h
+++ b/src/workerd/api/url-standard.h
@@ -35,12 +35,12 @@ class URLSearchParams: public jsg::Object {
     }
   };
 public:
-  using StringPair = jsg::Sequence<kj::String>;
+  using StringPair = jsg::Sequence<jsg::USVString>;
   using StringPairs = jsg::Sequence<StringPair>;
 
   using Initializer = kj::OneOf<StringPairs,
-                                jsg::Dict<kj::String, kj::String>,
-                                kj::String>;
+                                jsg::Dict<jsg::USVString, jsg::USVString>,
+                                jsg::USVString>;
 
   // Constructor called by the static constructor method.
   URLSearchParams(Initializer init);
@@ -50,12 +50,12 @@ public:
 
   static jsg::Ref<URLSearchParams> constructor(jsg::Optional<Initializer> init);
 
-  void append(kj::String name, kj::String value);
-  void delete_(jsg::Lock& js, kj::String name, jsg::Optional<kj::String> value);
-  kj::Maybe<kj::ArrayPtr<const char>> get(kj::String name);
-  kj::Array<kj::ArrayPtr<const char>> getAll(kj::String name);
-  bool has(jsg::Lock& js, kj::String name, jsg::Optional<kj::String> value);
-  void set(kj::String name, kj::String value);
+  void append(jsg::USVString name, jsg::USVString value);
+  void delete_(jsg::Lock& js, jsg::USVString name, jsg::Optional<jsg::USVString> value);
+  kj::Maybe<kj::ArrayPtr<const char>> get(jsg::USVString name);
+  kj::Array<kj::ArrayPtr<const char>> getAll(jsg::USVString name);
+  bool has(jsg::Lock& js, jsg::USVString name, jsg::Optional<jsg::USVString> value);
+  void set(jsg::USVString name, jsg::USVString value);
   void sort();
 
   JSG_ITERATOR(EntryIterator, entries,
@@ -162,9 +162,9 @@ public:
   URL(kj::StringPtr url, kj::Maybe<kj::StringPtr> base = kj::none);
   ~URL() noexcept(false) override;
 
-  static jsg::Ref<URL> constructor(kj::String url, jsg::Optional<kj::String> base);
+  static jsg::Ref<URL> constructor(jsg::USVString url, jsg::Optional<jsg::USVString> base);
 
-  static kj::Maybe<jsg::Ref<URL>> parse(jsg::Lock& js, kj::String url, jsg::Optional<kj::String> base) {
+  static kj::Maybe<jsg::Ref<URL>> parse(jsg::Lock& js, jsg::USVString url, jsg::Optional<jsg::USVString> base) {
     // Method should not throw if the parse fails
     return js.tryCatch([&]() -> kj::Maybe<jsg::Ref<URL>> {
       return constructor(kj::mv(url), kj::mv(base));
@@ -172,36 +172,36 @@ public:
   }
 
   kj::ArrayPtr<const char> getHref();
-  void setHref(jsg::Lock& js, kj::String value);
+  void setHref(jsg::Lock& js, jsg::USVString value);
 
   kj::Array<const char> getOrigin();
 
   kj::ArrayPtr<const char> getProtocol();
-  void setProtocol(kj::String value);
+  void setProtocol(jsg::USVString value);
 
   kj::ArrayPtr<const char> getUsername();
-  void setUsername(kj::String value);
+  void setUsername(jsg::USVString value);
 
   kj::ArrayPtr<const char> getPassword();
-  void setPassword(kj::String value);
+  void setPassword(jsg::USVString value);
 
   kj::ArrayPtr<const char> getHost();
-  void setHost(kj::String value);
+  void setHost(jsg::USVString value);
 
   kj::ArrayPtr<const char> getHostname();
-  void setHostname(kj::String value);
+  void setHostname(jsg::USVString value);
 
   kj::ArrayPtr<const char> getPort();
-  void setPort(kj::String value);
+  void setPort(jsg::USVString value);
 
   kj::ArrayPtr<const char> getPathname();
-  void setPathname(kj::String value);
+  void setPathname(jsg::USVString value);
 
   kj::ArrayPtr<const char> getSearch();
-  void setSearch(kj::String value);
+  void setSearch(jsg::USVString value);
 
   kj::ArrayPtr<const char> getHash();
-  void setHash(kj::String value);
+  void setHash(jsg::USVString value);
 
   jsg::Ref<URLSearchParams> getSearchParams();
 
@@ -215,9 +215,9 @@ public:
   //   'not a url'
   // ].filter((test) => URL.canParse(test));
   //
-  static bool canParse(kj::String url, jsg::Optional<kj::String> base = kj::none);
+  static bool canParse(jsg::USVString url, jsg::Optional<jsg::USVString> base = kj::none);
   static jsg::JsString createObjectURL(jsg::Lock& js, kj::OneOf<jsg::Ref<File>, jsg::Ref<Blob>> object);
-  static void revokeObjectURL(jsg::Lock& js, kj::String object_url);
+  static void revokeObjectURL(jsg::Lock& js, jsg::USVString object_url);
 
   JSG_RESOURCE_TYPE(URL) {
     JSG_READONLY_PROTOTYPE_PROPERTY(origin, getOrigin);
--- a/src/workerd/jsg/value.h
+++ b/src/workerd/jsg/value.h
@@ -412,8 +412,10 @@ class StringWrapper {
   static constexpr const char* getName(kj::String*) {
     return "string";
   }
-  // TODO(someday): This conflates USVStrings, which must have valid code points, with DOMStrings,
-  //   which needn't have valid code points.
+
+  // TODO(someday): The conversion to kj::String doesn't explicitly consider the distinction
+  // between DOMString (~ WTF-8; could contain invalid code points) and USVString (invalid code
+  // points are always replaced with U+FFFD). Code should make an explict choice between the two.
 
   static constexpr const char* getName(kj::ArrayPtr<const char>*) {
     return "string";
@@ -426,6 +428,14 @@ class StringWrapper {
     return "ByteString";
   }
   // TODO(cleanup): Move to a HeaderStringWrapper in the api directory.
+
+  static constexpr const char* getName(USVString*) {
+    return "USVString";
+  }
+
+  static constexpr const char* getName(DOMString*) {
+    return "DOMString";
+  }
 
   v8::Local<v8::String> wrap(v8::Local<v8::Context> context,
       kj::Maybe<v8::Local<v8::Object>> creator,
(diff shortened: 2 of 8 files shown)


anonrig:
Why can't we change kj::String to use correct utf-8 values? wouldn't it be safer? @kentonv @jasnell

src/workerd/jsg/value.h
```diff
@@ -487,6 +499,18 @@ class StringWrapper {
...
     return kj::mv(result);
   }
+
+  kj::Maybe<USVString> tryUnwrap(v8::Local<v8::Context> context,
+      v8::Local<v8::Value> handle,
+      USVString*,
+      kj::Maybe<v8::Local<v8::Object>> parentObject) {
+    v8::Local<v8::String> str = check(handle->ToString(context));
+    v8::Isolate* isolate = context->GetIsolate();
+    auto buf = kj::heapArray<char>(str->Utf8LengthV2(isolate) + 1);
```
anonrig, line 509: technically we can check for str->ContainsOneByteOnly() and omit utf8 calls for one byte strings

npaun:
@anonrig Some APIs specify that a value is a DOMString which allows those invalid characters.

anonrig:
> @anonrig Some APIs specify that a value is a DOMString which allows those invalid characters.

I'm thinking out loud:

Maybe we can tackle the problem from a different perspective. Rather than allowing invalid characters for all inputs, we can just disallow by default, and introduce DOMString() rather than USVString().

Allowing invalid utf8 values by default can cause security issues.

kentonv:
> Why can't we change kj::String to use correct utf-8 values? wouldn't it be safer?

kj::String accepts whatever you give it. It contains an array of chars. By convention it is supposed to contain UTF-8 chars, but it is not kj::String's job to validate this. Validating in kj::String would be a bad idea because:

1. It would be incredibly slow as kj::String is used everywhere, so presumably any validation checks would happen everywhere.
2. Over-validation can lead to serious problems. For example, most systems that are based around UTF-16 do _not_ validate surrogate pairs. This includes Java, JavaScript, Windows NT, etc. On Windows in particular, filenames can contain invalid surrogate pairs. If kj::String aggressively validated surrogate pairs, then people would be unable to use the KJ filesystem APIs to list and manipulate filenames on Windows that contain invalid surrogate pairs. This would be bad! In fact, this could be a DoS attack vector: drop a file with a bad name in the right place and break any KJ program trying to look at that directory. Yes that's right: aggressive validation can _create_ security problems; it is _not_ safer.

src/workerd/jsg/value.h
```diff
@@ -1021,17 +1045,22 @@ class DictWrapper {
     auto object = handle.As<v8::Object>();
     v8::Local<v8::Array> names = check(object->GetOwnPropertyNames(context));
     auto length = names->Length();
-    auto builder = kj::heapArrayBuilder<typename Dict<V, K>::Field>(length);
+
+    // Use a HashMap during construction because the conversion to the string type K could
```
kentonv, line 1049: This adds a bunch of overhead, and in the common case that the key type `K` is `kj::String` then this overhead is completely wasted. We need to gate this to apply only when the type is `USVString`. (Even then, I'm really dubious about whether anyone cares about this.)

src/workerd/api/url-standard.h
```diff
@@ -50,12 +50,12 @@ class URLSearchParams: public jsg::Object {
 
   static jsg::Ref<URLSearchParams> constructor(jsg::Optional<Initializer> init);
 
-  void append(kj::String name, kj::String value);
-  void delete_(jsg::Lock& js, kj::String name, jsg::Optional<kj::String> value);
-  kj::Maybe<kj::ArrayPtr<const char>> get(kj::String name);
-  kj::Array<kj::ArrayPtr<const char>> getAll(kj::String name);
-  bool has(jsg::Lock& js, kj::String name, jsg::Optional<kj::String> value);
-  void set(kj::String name, kj::String value);
+  void append(jsg::USVString name, jsg::USVString value);
```
kentonv, line 53: Isn't this a breaking change? Don't we need a compat flag?

Maybe we could even have a compat flag that enables the strict checks, and just never even bother turning it on by default... but use it for the tests?

kentonv (changes requested):
I feel like this is one of those things that nobody cares about except for some pedantic test-writers. Do we really have to do this? What if we just didn't?

npaun:
@kentonv The main motivation I had for fixing this is that the underlying library we use for URL stuff, ada-url, says the caller is responsible for never passing invalid UTF-8 characters.

I feel like this test is pretty pedantic but it does seem a bit unpleasant that we'd parse a URL differently than a browser or Node.

src/workerd/jsg/value.h
jasnell: Definitely agree with Kenton here. Less so about whether anyone cares about it or not but about the added overhead the extra book keeping here adds.

src/workerd/jsg/value.h
```diff
@@ -1048,10 +1077,15 @@ class DictWrapper {
           throwTypeError(context->GetIsolate(), TypeErrorContext::dictField(strName.cStr()),
               TypeWrapper::getName((V*)nullptr));
         }
-        builder.add(typename Dict<V, K>::Field{
-          KJ_ASSERT_NONNULL(kj::mv(unwrappedName)), KJ_ASSERT_NONNULL(kj::mv(unwrappedValue))});
+        entries.upsert(
+            KJ_ASSERT_NONNULL(kj::mv(unwrappedName)), KJ_ASSERT_NONNULL(kj::mv(unwrappedValue)));
```
jasnell, line 1081: I'm not sure if silently ignoring the potentially duplicate entries makes sense and is what we'd want.

npaun: It's what URLSearchParams is expected to do; I could move the behaviour over there if it makes more sense.

src/workerd/api/url-standard.h
```diff
@@ -50,12 +50,12 @@ class URLSearchParams: public jsg::Object {
...
+  void delete_(jsg::Lock& js, jsg::USVString name, jsg::Optional<jsg::USVString> value);
+  kj::Maybe<kj::ArrayPtr<const char>> get(jsg::USVString name);
+  kj::Array<kj::ArrayPtr<const char>> getAll(jsg::USVString name);
+  bool has(jsg::Lock& js, jsg::USVString name, jsg::Optional<jsg::USVString> value);
+  void set(jsg::USVString name, jsg::USVString value);
```
jasnell, line 58: Note: you'll likely need to also update the type generation code to handle `USVString`

src/workerd/api/url-standard.h
jasnell: If we're going to do that it might be easier to have a compatibility flag that updates the `kj::String` type mapping in general to always perform the invalid character replacement. I think the only web platform APIs we have that would *not* be USVString are the `Header` and we already have `jsg::ByteString` to cover that case. Either way, I agree that this likely does need a compat flag.

kentonv: Well, if we change it everywhere, then cases where an API round-trips text back to JavaScript now have a side effect that might be unexpected...

jasnell: is't that already the case when unpaired surrogates are used tho? For instance, if I have a test like...

```cpp
struct StringsContext: public ContextGlobalObject {
  kj::String str(kj::String value) {
    return kj::mv(value);
  }

  JSG_RESOURCE_TYPE(StringsContext) {
    JSG_METHOD(str);
  }
};
JSG_DECLARE_ISOLATE_TYPE(StringsIsolate, StringsContext);

KJ_TEST("jsg::Name works") {
  Evaluator<StringsContext, StringsIsolate> e(v8System);
  e.expectEval("const m = '\\ud83c'; str(m) === m", "boolean", "false");
}
```

The `kj::String` is not going to roundtrip successfully since while the unpaired surrogate will not be replaced with the FFFD replacement character on the inbound type mapping, it *is* replaced on the outbound type mapping.

jasnell: In fact, it's even worse than that... with the current type mapping, if I change the above test to do `str(str(m))` the resulting kj::String that is received by the second call to `str(...)` ends up with the byte sequence `239, 191, 189, 239, 191, 189, 239, 191, 189` ...  If we modify the `kj::String` type mapping to always replace the invalid character then the double round trip produces a reasonable result.

```cpp
struct StringsContext: public ContextGlobalObject {
  kj::String str(kj::String value) {
    KJ_DBG(value.asBytes());
    return kj::mv(value);
  }

  JSG_RESOURCE_TYPE(StringsContext) {
    JSG_METHOD(str);
  }
};
JSG_DECLARE_ISOLATE_TYPE(StringsIsolate, StringsContext);

KJ_TEST("...") {
  Evaluator<StringsContext, StringsIsolate> e(v8System);
  e.expectEval("const m = '\\ud83c'; str(str(m)).codePointAt(0)", "number", "65533");
}
```

Results without invalid char replacement on the inbound type mapping:

```
workerd/jsg/value-test.c++:1216: value.asBytes() = 237, 160, 188
workerd/jsg/value-test.c++:1216: value.asBytes() = 239, 191, 189, 239, 191, 189, 239, 191, 189
```

Results *with* invalid char replacement on the inbound tuype mapping:

```
workerd/jsg/value-test.c++:1216: value.asBytes() = 239, 191, 189
workerd/jsg/value-test.c++:1216: value.asBytes() = 239, 191, 189
```

This seems like far more reasonable behavior.

jasnell:
> ... kj::String accepts whatever you give it. It contains an array of chars. By convention it is supposed to contain UTF-8 chars, but it is not kj::String's job to validate this

Couldn't we have a variation of the `kj::String` construction that performs the validation and replacement? It wouldn't incur the additional costs and would be explicitly opt in.

```
auto wellformed = kj::usvstr("...");

auto notWellformed = kj::str("...");
auto wellformed = kj::mv(notWellformed).toWellFormed();
```

(the "toWellFormed" terminology is borrowed by the recent API additions in JS to accomplish the same)

jasnell:
> ... On Windows in particular, filenames can contain invalid surrogate pairs. If kj::String aggressively validated surrogate pairs, then people would be unable to use the KJ filesystem APIs to list and manipulate filenames on Windows that contain invalid surrogate pairs

Side note... stuff like this is why Node.js' file system apis accept string *or* Buffer for filenames. Years ago I had a consulting client in Japan who was hitting up against the fact that their filesystem contained files and directory names that used multiple text encodings (since on some OS's file paths are just byte sequences). Node.js originally treated all file paths as UTF8. Sometimes we do have to just let invalid sequences be invalid sequences without validation or replacement

src/workerd/api/url-standard.h
kentonv: > it is replaced on the outbound type mapping.

WTF!

I totally assumed that if V8 was happy to produce WTF-8, it would also accept it, but apparently not. Apparently V8's UTF-8 parser, when presented with a UTF-8 sequence representing an unpaired surrogate -- which is a three-byte sequence -- just replaces _each_ of the three bytes with a replacement character?!?!

kentonv: Well I guess that means no one is relying on us round-tripping unpaired surrogates after all. But why the heck does V8 only do this validation in one direction and not the other?

npaun: Yikes. The inbound case is a mess.

jasnell: So how about this as a path forward.... Let's

* introduce a compat flag to treat `kj::String` inbound as USVString (replacing invalid sequences).
* introduce `jsg::UsvString` as a more explicit indicator to use replacement that works with or without the compat flag
* introduce `jsg::DomString` as an explicit indicator to use the current inbound type mapping and we can find a way to ensure that it works as expected on the outbound too (that is, that it produces WTF and round trips correctly).
* Incrementally move our existing direct uses of `kj::String` over to either one of these explicit string types. 

I think most of our APIs are going to handle the switch to USVString semantics easily. It's more likely that we have a far smaller set of APIs that may rely on the current behavior.

kentonv: IMO feels like a lot of work to solve a problem that no one is actually hitting. But I won't stop you.

src/workerd/jsg/value.h
npaun: Given that we're not sure we want to ignore duplicate keys, I backed out the changes to Dict - it'll just produce a list of pairs and callers can decide what to do.

src/workerd/jsg/value.h
npaun: Moved to the caller

npaun:
...

npaun:
Hey guys, here's an update on this PR:
- Creates USVString and DOMString to allow code to explicitly specify what behaviour it intends.
- Reverted the behaviour where jsg::Dict<K = USVString> would silently ignore duplicate keys. Instead jsg::Dict is always a list of pairs and the caller must decide what to do in this case.

I propose we merge this PR at this point and follow-on with further PRs.

Here's what's left to discuss and implement:
- Add a compat flag to make kj::String act as a USVString.
- Fix the inbound case where V8 replaces a unpaired surrogate with three replacement chars (this might involve fixing stuff on their side).

src/workerd/jsg/value.h
jasnell: Let's look at that separately since that also holds true for several other places where we call `WriteUtf8...`

src/workerd/jsg/value.h
```diff
@@ -487,6 +509,29 @@ class StringWrapper {
...
+    return USVString(kj::mv(buf));
+  }
+
+  kj::Maybe<DOMString> tryUnwrap(v8::Local<v8::Context> context,
+      v8::Local<v8::Value> handle,
+      DOMString*,
+      kj::Maybe<v8::Local<v8::Object>> parentObject) {
+    v8::Local<v8::String> str = check(handle->ToString(context));
+    v8::Isolate* isolate = context->GetIsolate();
+    auto buf = kj::heapArray<char>(str->Utf8LengthV2(isolate) + 1);
```
jasnell, line 531: This needs tests added to value-test.c++

src/workerd/api/url-standard.h
npaun: We'll handle the remaining aspects of this discussion in a follow-up PR.

src/workerd/api/url-standard.h
npaun: As we suspected, there aren't any changes to the generated types - it all comes out as "string".

npaun:
@jasnell It's ready now.

jasnell (approved):

jasnell:
I don't think a compat flag is necessary in this case.. other places we likely need to be more careful of but given the fairly extensive testing we were already doing with URL I think we're safe on this one. Updating other existing `kj::String` uses, however, will be a different story.

PR #3556, opened 2025-02-14
WPT harness: parse meta comments, raise error if config provided for nonexistent test

npaun:
* Parse WPT meta comments found at the beginning of tests, in the form of "// META key=value"
* Implement the "script" directive which includes another script
   - Because we're using unsafeEval to run the test code, I had to add a guard to make sure each script is included only once.
   - Also gets rid of our ad-hoc `TestOptions.includeFile` feature
 * Throw an error at the beginning of a test suite if a file listed in config wasn't actually discovered in the WPT repo. This way if you thought we were running a certain test but it wasn't working, there won't be a silent failure.


npaun:
...

npaun:
@anonrig Let me know if this makes things better for CLion too...

npaun:
Damn. Still no windows. At this point the tsconfig is so simple.

build/wpt_test.bzl
```diff
@@ -133,14 +140,20 @@ def _wpt_wd_test_gen_impl(ctx):
     file, WPT test scripts, associated JSON resources.
     """
     src = ctx.actions.declare_file("{}.wd-test".format(ctx.attr.test_name))
+
+    # Both the .js and the .js.map file are included, and we need to tell them apart.
```
anonrig, line 144: How come? Can you add a link to WPT documentation or an example?

build/wpt_test.bzl
```diff
@@ -183,6 +196,13 @@ def wd_relative_path(label, target):
     """
     return "../" * (label.package.count("/") + label.name.count("/") + 1) + target.short_path
 
+def is_resource(file):
```
anonrig, line 199:
```suggestion
def is_in_resources_folder(file):
```

src/wpt/BUILD.bazel
```diff
@@ -20,7 +33,7 @@ ts_project(
 )
 
 eslint_bin.eslint_test(
-    name = "wpt-test-harness@eslint",
+    name = "all@eslint",
```
anonrig, line 36:
```suggestion
    name = "wpt-all@eslint",
```

src/wpt/harness/harness.ts
```diff
@@ -685,6 +685,7 @@ globalThis.assert_not_own_property = (
   );
 };
 
+globalThis.includedFiles = new Set();
```
anonrig, line 688: Can you add a comment to what this set does?

src/wpt/harness/harness.ts
```diff
@@ -714,38 +715,122 @@ async function validate(
...
+      continue;
+    }
+
+    switch (groups.key) {
+      case 'global':
+        meta.global = groups.value.split(',');
+        break;
+
+      case 'script': {
+        // Replace leading "./" if found
```
anonrig, line 779: What happens if it starts with ../?

src/wpt/urlpattern-test.ts
```diff
@@ -164,4 +157,9 @@ export default {
       'Pattern: [{"pathname":"/([\\\\d&&[0-1]])"}] Inputs: [{"pathname":"/3"}]',
     ],
   },
+  'urlpattern.https.any.js': {
+    comment:
+      "We don't currently distinguish between the HTTP and HTTPS variants of tests",
```
anonrig, line 162: I'm following this comment? Why are we skipping this?

build/wpt_test.bzl
npaun: That's actually down to Bazel ts_project. I'll reword for clarity.

npaun: Actually I'm trying another trick that might work better on Windows. We'll see.

src/wpt/harness/harness.ts
npaun: I moved some stuff around in the harness. Here's where the comment was added:
https://github.com/cloudflare/workerd/blob/2cab4bb9dd84ab2b3d0ab6142e6669177aca8748/src/wpt/harness/harness.ts#L852

src/wpt/harness/harness.ts
npaun: I switched to `path.normalize` which is a bit nicer, but we still have limitations on what files can be accessed in a wd-test. We'll at least get an error message if a file can't be loaded.
https://github.com/cloudflare/workerd/blob/2cab4bb9dd84ab2b3d0ab6142e6669177aca8748/src/wpt/harness/harness.ts#L829

src/wpt/urlpattern-test.ts
npaun: I tried improving the wording. Let me know if it makes more sense now.

anonrig:
We need to fix Windows before landing this PR but otherwise it looks good to me

anonrig (approved):

PR #3560, opened 2025-02-17
Enable FinalizationRegistry

ketanhwr:


samples/helloworld_esm/worker.js
```diff
@@ -4,6 +4,82 @@
 
 export default {
   async fetch(req, env) {
-    return new Response("Hello World\n");
+    let resp = await fetch("https://example.com");
```
jasnell, line 7: Don't forget to revert this sample back to its original state when this is done. Most likely it would make sense to have this as a separate example.

src/workerd/api/global-scope.c++
```diff
@@ -247,6 +247,7 @@ kj::Promise<DeferredProxy<void>> ServiceWorkerGlobalScope::request(kj::HttpMetho
       return DeferredProxy<void>{promise.attach(kj::mv(adapter), kj::mv(client))};
     }
   } else KJ_IF_SOME(promise, event->getResponsePromise(lock)) {
+
```
jasnell, line 250:
```suggestion
```

src/workerd/io/io-context.c++
```diff
@@ -1182,6 +1197,10 @@ bool IoContext::hasCurrent() {
   return threadLocalRequest != nullptr;
 }
 
+bool IoContext::isCurrentNull() {
+  return threadLocalRequest == nullptr;
+}
```
jasnell, line 1202: There is already an `IoContext::hasCurrent()` that this duplicates

src/workerd/io/io-context.c++
```diff
@@ -1155,6 +1159,17 @@ void IoContext::runImpl(Runnable& runnable,
...
   });
+
+  if (!isCurrentNull()) {
+    KJ_LOG(ERROR, "IoContext not-null before running PumpMessageLoop()");
+  } else {
+    worker->runInLockScope(lockType, [&](Worker::Lock& lock) {
+        jsg::Lock& js = lock;
+        auto& system = const_cast<jsg::V8System&>(js.getV8System());
+        KJ_DBG(js.v8Isolate);
+        while (v8::platform::PumpMessageLoop(&system.getDefaultPlatform(), js.v8Isolate, v8::platform::MessageLoopBehavior::kDoNotWait)) {}
```
jasnell, line 1170: hmm... what is it you're looking to accomplish here? Manually pumping the message loop is quite unusual. Is the idea here to try to force the finalization registry callback to run more predictably?

src/workerd/jsg/jsg.h
```diff
@@ -2276,6 +2276,8 @@ class ExternalMemoryAdjustment final {
 // Isolate<TypeWrapper>::Lock. Usually this is only done in top-level code, and the Lock is
 // passed down to everyone else from there. See setup.h for details.
 
+class V8System;
```
jasnell, line 2279: I don't think we should be exposing `V8System` like this. The intent of JSG is to abstract V8 specifics away as much as possible and this just exposes them more. If we have to expose something here, which I'm unsure about, then the functionality should be folded into a new `jsg::Lock` method... like `js.pumpMessageLoop()` so that the details of interfacing with the v8 APIS do not need to leak more out to other areas.

src/workerd/jsg/resource.h
```diff
@@ -1431,7 +1431,7 @@ class ResourceWrapper {
     // We do not allow use of WeakRef or FinalizationRegistry because they introduce
     // non-deterministic behavior.
     check(global->Delete(context, v8StrIntern(isolate, "WeakRef"_kj)));
-    check(global->Delete(context, v8StrIntern(isolate, "FinalizationRegistry"_kj)));
+    //check(global->Delete(context, v8StrIntern(isolate, "FinalizationRegistry"_kj)));
```
jasnell, line 1434: We'd be enabling `WeakRef` too right?

src/workerd/io/io-context.c++
ketanhwr: We don't pump the message loop anywhere yet, and the Finalizers will never by invoked unless we manually pump it.
Deno had the same issue sometime back: https://github.com/denoland/deno/issues/7674#issuecomment-868925712

src/workerd/jsg/jsg.h
ketanhwr: Yeah, I agree. `js.pumpMessageLoop()` was the intended plan before I sent this out for review :sweat_smile:

src/workerd/jsg/resource.h
ketanhwr: The current plan is to only enable `FinalizationRegistry`, given the immediate memory cleanup benefits for wasm users. `WeakRef` is something we can probably think about as a followup.

src/workerd/io/io-context.c++
ketanhwr: I actually saw `hasCurrent` and then still wrote this function for some reason :facepalm:

src/workerd/io/io-context.c++
kentonv: Our failure to pump the message loop is also what led to the memory leak with liftoff code flushing in a recent V8 update. Apparently we were supposed to be doing this all along and somehow got away without it.

src/workerd/io/io-context.c++
```diff
@@ -1164,6 +1159,17 @@ void IoContext::runImpl(Runnable& runnable,
...
```
kentonv, line 1170: It feels wrong for this code to be aware of the v8::Platform. Knowledge of how the platform is set up (and even the fact that we use the default platform) should be encapsulated inside JSG.

To that end, perhaps `jsg::Lock` should have a `pumpMessageLoop()` method, which internally calls `v8::platform::PumpMessageLoop` appropriately?

kentonv, line 1170: Is this actually the right time to call PumpMessageLoop -- after every single `ctx.run()`?

Arguably it might be preferable if we invoked `PumpMessageLoop` asynchronously later on when we're not actively responding to a request.

src/workerd/jsg/resource.h
kentonv: Is there actually anything more needed to enable WeakRef? If it's just a matter of removing the line above, I think we should do it in this change.

kentonv, line 1434: This probably needs a compat flag as I suspect there are people that check for the existence of these APIs and use them if they are available -- making them suddenly available would therefore cause such workers to start taking a new code path which could end up being broken.

ketanhwr: Yes it should work by just removing the line, although @harrishancock was concerned about `WeakRef` giving immediate notification of GC collection instead of `FinalizationRegistry` which is more non-deterministic.

jasnell: `WeakRef` does not give immediate notification thankfully. I suppose someone could poll it as quickly as possible to approximate immediate notification but thankfully it does not provide a notification api.

src/workerd/io/io-context.c++
jasnell: I agree that we likely need to find a better place for this. We can end up running ctx.run quite a few times during a request.

src/workerd/jsg/resource.h
jasnell: This makes me sad, but yes, compat flag absolutely needed. I've seen such checks in the wild a few times now.

harrishancock: My concern isn't immediate notification, but lack of control over when notifications occur. With FinalizationRegistry, we can control exactly when finalization callbacks are scheduled. With WeakRef, we have no control over when they appear empty -- that's entirely up to the GC. So, if controlling the timing of GC observation is important for risk mitigation, then enabling WeakRef is strictly higher risk than enabling FinalizationRegistry.

That said, I'd love to just accept the risk and enable WeakRef and be done with it.

src/workerd/io/compatibility-date.capnp
```diff
@@ -692,4 +692,9 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $experimental
       $neededByFl;
   # Enables cache settings specified request in fetch api cf object to override cache rules. (only for user owned or grey-clouded sites)
+
+  jsWeakRef @73 :Bool
+      $compatEnableFlag("enable_weak_ref")
+      $compatDisableFlag("disable_weak_ref");
+  # Enables WeakRefs and FinalizationRegistry API. WebAssembly based projects often rely on this API for wasm memory cleanup
```
ketanhwr, line 699: Should we add a compatibility date as well (I think we should?)? How exactly do we decide what goes behind a date vs flag vs both?

ketanhwr (commented):
TODO: tests (was testing adhoc as of now)

src/workerd/io/compatibility-date.capnp
jasnell: If the intention is for `FinalizationRegistry` to just be available everywhere then yes, this will need a date where it is default switched on. We exclude the date for any flags that should always be explicitly opt-in.

src/workerd/jsg/jsg.h
jasnell: Would it be possible to make `V8System` entirely hidden here so that it's not exposed at all from jsg.h?

src/workerd/server/tests/weakref/test.mjs
```diff
@@ -0,0 +1,48 @@
...
+  workerd = null;
+});
+
+// FinalizationRegistry callbacks only run after the exported handler returns
+// So we should see its effects in a follow-up request
+test('JS FinalizationRegistry', async () => {
+  let httpPort = await workerd.getListenPort('http');
+  for (let i = 0; i < 3; ++i) {
+    const response = await fetch(`http://localhost:${httpPort}`);
+    assert.equal(await response.text(), i);
```
jasnell, line 46:
```suggestion
    assert.strictEqual(await response.text(), i);
```

src/workerd/server/tests/weakref/test.mjs
```diff
@@ -0,0 +1,48 @@
...
+
+  await workerd.start();
+
+  await workerd.getListenPort('http');
+});
+
+// Stop workerd.
+afterEach(async () => {
+  const [code, signal] = await workerd.stop();
+  assert(code === 0 || signal === 'SIGTERM');
```
jasnell, line 36: would prefer if this were somehow reworked to use `assert.strictEqual(...)` to ensure a more useful stack output if the assertion fails

src/workerd/server/tests/weakref/test.mjs
```diff
@@ -0,0 +1,48 @@
+import { env } from 'node:process';
+import { beforeEach, afterEach, test } from 'node:test';
+import assert from 'node:assert';
+import { WorkerdServerHarness } from '../server-harness.mjs';
```
jasnell, line 4: This file appears to be expected to be run by node.js? Some comments in here explaining what this file is and how it is used would be helpful. At first it looked like just another wd-test file but then I realized it was a node.js script and not meant for workerd.

src/workerd/jsg/resource.h
kentonv: I noticed [the MDN docs for WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) say:

> If your code has just created a WeakRef for a target object, or has gotten a target object from a WeakRef's deref method, that target object will not be reclaimed until the end of the current JavaScript [job](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#job) (including any promise reaction jobs that run at the end of a script job). That is, you can only "see" an object get reclaimed between turns of the event loop. This is primarily to avoid making the behavior of any given JavaScript engine's garbage collector apparent in code — because if it were, people would write code relying on that behavior, which would break when the garbage collector's behavior changed. (Garbage collection is a hard problem; JavaScript engine implementers are constantly refining and improving how it works.)

Does this help with your concerns, @harrishancock?

For my part, my main concern is that we can think of how we'd implement record/replay deterministically. I think it's pretty straightforward: in record mode, we intercept WeakRef.deref() calls and make a recording that the call happened, and whether the object was still live. In replay mode, we replace WeakRef with something that isn't really a weak ref, but holds onto the object until the last recorded `deref()` call that was recorded as returning the live value, then drops the reference at that point and returns undefined after that.

src/workerd/io/io-context.c++
```diff
@@ -447,9 +448,11 @@ kj::Promise<void> IoContext::IncomingRequest::drain() {
     // For non-actor requests, apply the configured soft timeout, typically 30 seconds.
     timeoutPromise = context->limitEnforcer->limitDrain();
   }
+  context->pumpMessageLoop();
   return context->waitUntilTasks.onEmpty()
       .exclusiveJoin(kj::mv(timeoutPromise))
-      .exclusiveJoin(context->abortPromise.addBranch().then([] {}, [](kj::Exception&&) {}));
+      .exclusiveJoin(context->abortPromise.addBranch().then(
+          [] {}, [](kj::Exception&& e) { KJ_LOG(INFO, "uncaught exception", e); }));
```
danlapid, line 455: Is this added log intended? I'm afraid it will spam sentry

src/workerd/jsg/jsg.c++
```diff
@@ -4,6 +4,7 @@
 
 #include "jsg.h"
 
+#include "libplatform/libplatform.h"
```
danlapid, line 7:
```suggestion
#include <libplatform/libplatform.h>
```

src/workerd/jsg/jsg.c++
```diff
@@ -270,6 +271,11 @@ void Lock::terminateExecution() {
   v8Isolate->TerminateExecution();
 }
 
+void Lock::pumpMessageLoop() {
+  auto platform = IsolateBase::from(v8Isolate).getDefaultPlatform();
+  while (v8::platform::PumpMessageLoop(platform, v8Isolate)) {}
```
danlapid, line 276: This seems to be using a class method as a static method, why not use `platform.PumpMessageLoop(v8Isolate)`?
On that note, another Nit of mine is I would want to explicitly pass `MessageLoopBehavior::kDoNotWait` rather than rely on default function argument.

Another note, the `while(...)` is concerning. I guess we are still covered by the thread's watchdog..
Essentially the while loop is to solve for if someone registers another finalizer inside of the finalizers right?
Wonder if we should limit this to some const amount of loops.

src/workerd/jsg/setup.c++
```diff
@@ -77,11 +77,19 @@ static kj::Own<v8::Platform> userPlatform(v8::Platform& platform) {
...
+    v8::Platform* defaultPlatformPtr)
+    : V8System(userPlatform(platformParam), flags, defaultPlatformPtr) {}
+V8System::V8System(kj::Own<v8::Platform> platformParam,
+    kj::ArrayPtr<const kj::StringPtr> flags,
+    v8::Platform* defaultPlatformPtr)
     : platformInner(kj::mv(platformParam)),
-      platformWrapper(*platformInner) {
+      platformWrapper(*platformInner),
+      defaultPlatformPtr_{defaultPlatformPtr} {
+  if (!defaultPlatformPtr_) {
```
danlapid, line 90: Calls to getDefaultPlatform would be a null deref if this branch was not entered - seems unsafe to me.
This should probably be `KJ_ASSERT(defaultPlatformPtr_ != nullptr)` instead.

src/workerd/jsg/setup.h
```diff
@@ -56,19 +56,28 @@ class V8System {
...
+  explicit V8System(v8::Platform& platform,
+      kj::ArrayPtr<const kj::StringPtr> flags,
+      v8::Platform* defaultPlatformPtr = nullptr);
 
   ~V8System() noexcept(false);
 
   typedef void FatalErrorCallback(kj::StringPtr location, kj::StringPtr message);
   static void setFatalErrorCallback(FatalErrorCallback* callback);
 
+  auto& getDefaultPlatform() {
```
danlapid, line 68: Let's use an explicit type here please.

danlapid: I wonder if it's better to return a pointer instead of a reference here 🤔

src/workerd/jsg/setup.h
```diff
@@ -56,19 +56,28 @@ class V8System {
...
  private:
   kj::Own<v8::Platform> platformInner;
   V8PlatformWrapper platformWrapper;
   friend class IsolateBase;
+  v8::Platform* defaultPlatformPtr_;
 
-  explicit V8System(kj::Own<v8::Platform>, kj::ArrayPtr<const kj::StringPtr>);
+  explicit V8System(kj::Own<v8::Platform>,
+      kj::ArrayPtr<const kj::StringPtr>,
+      v8::Platform* defaultPlatformPtr = nullptr);
```
danlapid, line 80: As noted above, I'm not sure as is this argument is truly optional.

src/workerd/io/io-context.c++
kentonv: It's an INFO log so it'll only show up in tests, but it looks unnecessary to me as any abort exception should have been reported elsewhere already, hence why it was being ignored here.

ketanhwr: Yup, this is only for tests. However I didn't see `Worker exceeded CPU time limit` log (if a FinalizationRegistry callback exceeds it) unless I added this log line here

src/workerd/jsg/jsg.c++
ketanhwr: There's no `platform.PumpMessageLoop(v8Isolate)` method as part of `v8::Platform` unfortunately, it's part of `v8::DefaultPlatform` which isn't exported. 

The `while(...)` is to execute each "cleanup" task, not necessarily for the recursive use case. Although I should test how that behaves as well

src/workerd/jsg/setup.c++
ketanhwr: This is kind-of a workaround due to the default constructor. Let me refactor this a bit

ketanhwr:
@jasnell gentle ping - whenever you get time

src/workerd/jsg/setup.h
```diff
@@ -42,21 +42,17 @@ kj::Own<v8::Platform> defaultPlatform(uint backgroundThreadCount);
...
-  // Use a possibly-custom v8::Platform implementation. Use this if you need to override any
-  // functionality provided by the v8::Platform API.
-  explicit V8System(v8::Platform& platform);
+  explicit V8System(kj::ArrayPtr<const kj::StringPtr> flags = nullptr);
 
   // Use a possibly-custom v8::Platform implementation, and apply flags.
-  explicit V8System(v8::Platform& platform, kj::ArrayPtr<const kj::StringPtr> flags);
+  explicit V8System(v8::Platform& platform,
+      kj::ArrayPtr<const kj::StringPtr> flags,
+      v8::Platform* defaultPlatformPtr);
```
kentonv, line 55: This needs some commentary explaining that `platform` is expected to be some sort of wrapper around a default platform, and you must pass in the default platform pointer so that the message loop can be pumped.

Although arguably it would be more reusable if instead the caller passed in a callback function to pump the message loop, so that the platform doesn't actually have to wrap the default platform at all.

src/workerd/io/io-context.c++
```diff
@@ -1377,6 +1381,73 @@ kj::Promise<void> IoContext::startDeleteQueueSignalTask(IoContext* context) {
...
+        if (gotTermination) {
+          js.terminateExecution();
+        }
+      });
+
+      v8::TryCatch tryCatch(workerLock.getIsolate());
+      try {
+        jsg::Lock& js = workerLock;
+        js.pumpMessageLoop();
+      } catch (const jsg::JsExceptionThrown&) {
```
kentonv, line 1418: I don't think there's any way for `jsg::JsExceptionThrown` to be thrown by `pumpMessageLoop()`, so can this whole catch block be deleted?

(And also the KJ_DEFER, etc.)

src/workerd/io/io-context.c++
```diff
@@ -447,9 +448,11 @@ kj::Promise<void> IoContext::IncomingRequest::drain() {
     // For non-actor requests, apply the configured soft timeout, typically 30 seconds.
     timeoutPromise = context->limitEnforcer->limitDrain();
   }
+  context->pumpMessageLoop();
```
kentonv, line 451: Hmm, so you are pumping the message loop at `drain()` time. This means that it'll be pumped once per request, some time after the response is returned, but possibly before the request is totally done (if the request has waitUntil tasks).

Is this... actually right? If someone has a long-running request (and no concurrent requests), then I guess the message loop will never be pumped, so they'll never get any finalization callbacks.

Sorry, I know I already asked "is this the right place?" once before, you changed it, and now I'm asking again... I should have probably made a more concrete recommendation in the first place.

The right answer really depends on whether the message loop tasks are expected to be cheap or expensive. I'm not really clear on this, though I'd guess they are mostly cheap -- the only reason the message loop is used at all is so that the execution can be asynchronous.

If they are cheap, then the right place to pump the message loop would be right before releasing the isolate lock, every time we release the lock. This should actually be implemented in `Worker::Lock`, I think, not in `IoContext`.

If they are expensive, then we want to make sure they run asynchronously without blocking any I/O. To do that, just before releasing a lock, we would not run the loop, but _check_ if it has any messages in it and, if so, schedule an asynchronous task to run them. This async task should probably use `kj::yieldUntilWouldSleep()` to defer until there's nothing else to do.

I would probably assume for now that they are cheap, and implement the first option.

danlapid: Now that I think about it,
A key use case we need to support are DOs written in workers-rs which run out of memory.
I'm not actually sure when drain is called on those but seems right to run this more than once per request.

src/workerd/jsg/jsg.c++
```diff
@@ -270,6 +271,11 @@ void Lock::terminateExecution() {
   v8Isolate->TerminateExecution();
 }
 
+void Lock::pumpMessageLoop() {
+  auto platform = IsolateBase::from(v8Isolate).getDefaultPlatform();
+  while (v8::platform::PumpMessageLoop(platform, v8Isolate, v8::platform::MessageLoopBehavior::kDoNotWait)) {}
```
jasnell, line 276: Oy, it feels gnarly to add a while loop here. I'm assuming that we are assured enough that this won't go into an infinite loop at all? A comment here would probably be helpful.

src/workerd/io/io-context.c++
kentonv: Oh but I guess the problem with completely separating this from IoContext is that you need to apply the CPU limiter.

But the reason you don't want to put this right next to `js.runMicrotasks()` in `IoContext::runImpl()` is because you don't want the current IoContext to be set during the finalization callbacks?

I think the right answer there might actually be to... set `threadLocalRequest = nullptr` early? I think you could do that, right after `js.runMicrotasks()`, because it's what's going to happen anyway when you return.

```
js.runMicrotasks();
threadLocalRequest = nullptr;
js.pumpMessageLoop();
```

BTW, what happens if a finalizer queues microtasks? Does `pumpMessageLoop()` run them or do we need another call to `runMicrotasks()` to make them run?

jasnell: This might help for the not-wanting-the-IoContext-to-be-set now that it has landed: https://github.com/cloudflare/workerd/pull/3712/files#diff-b6ab0724c7e9c7f5b57a1f51b4ebd8480e3240c7823f3e2f09e9dbc8b31ed12aR1000

```c++
{
  SuppressIoContextScope noIoScope;
  js.pumpMessageLoop();
}
```

jasnell: > BTW, what happens if a finalizer queues microtasks? Does pumpMessageLoop() run them or do we need another call to runMicrotasks() to make them run?

A quick test shows that any microtasks queued during the callback are *not* drained by `v8::platform::PumpMessageLoop`

kentonv: > A quick test shows that any microtasks queued during the callback are not drained by v8::platform::PumpMessageLoop

Yeah that's not good -- those microtasks will end up running whenever the next event is delivered. We will need to drain them, and I suppose we may then have to pump the message loop _again_, and so on.

Perhaps this actually belongs inside `Lock::pumpMessageLoop()` -- the `while()` loop there should have a body that runs `runMicrotasks()`?

kentonv: Of course meanwhile in this loop we need to be able to check whether the isolate has been terminated... this is tricky.

src/workerd/io/compatibility-date.capnp
```diff
@@ -733,4 +733,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("allow_importable_env");
   # When allowed, `import { env } from 'cloudflare:workers'` will provide access
   # to the per-request environment/bindings.
+
+  jsWeakRef @79 :Bool
+      $compatEnableFlag("enable_weak_ref")
+      $compatEnableDate("2025-03-24")
```
harrishancock, line 739: Before merging, we should postpone this a bit. I typically use a Tuesday ~3 weeks out. (I see this is a Monday, which is fine, too.)

src/workerd/io/compatibility-date.capnp
```diff
@@ -733,4 +733,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatDisableFlag("disable_weak_ref");
+  # Enables WeakRefs and FinalizationRegistry API. WebAssembly based projects often rely on this API for wasm memory cleanup
```
harrishancock, line 741: Nit here and elsewhere: we use a 100 column limit in source files

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,19 @@
+let counter = 0;
+const fr = new FinalizationRegistry(() => {
```
harrishancock, line 2: Could this test exercise WeakRef, too, since we're enabling it?

src/workerd/server/workerd-api.c++
```diff
@@ -342,9 +343,8 @@ CompatibilityFlags::Reader WorkerdApi::getFeatureFlags() const {
   return *impl->features;
 }
 jsg::JsContext<api::ServiceWorkerGlobalScope> WorkerdApi::newContext(jsg::Lock& lock) const {
-  jsg::NewContextOptions options{
-    .newModuleRegistry = impl->tryGetModuleRegistry(),
-  };
+  jsg::NewContextOptions options{.newModuleRegistry = impl->tryGetModuleRegistry(),
+    .enableWeakRef = getFeatureFlags().getJsWeakRef()};
```
harrishancock, line 347: clang-format will format this a bit better if you leave a trailing comma after `.getJsWeakRef(),`.

src/workerd/io/compatibility-date.capnp
jasnell: Yeah, this is going to need a bit more time. 3-weeks out from when it is expected to land in workerd is typically reasonable. For this, it is a sizeable enough of a change that I might even bump it 4-5 weeks to be safe but 3-24 is *way* too soon

src/workerd/jsg/setup.h
ketanhwr: Passing a callback sounds better, since `DefaultPlatform` isn't exported by v8 as of now.

src/workerd/io/io-context.c++
ketanhwr: Yup will remove it (https://github.com/cloudflare/workerd/pull/3560#discussion_r1971921585)

src/workerd/jsg/jsg.c++
ketanhwr: Will add a comment - it is (theoretically) possible to keep adding FinalizationRegistry cleanup tasks from inside the cleanup callback. But, the while loop does run under cpu limiter, so it will get terminated at some point.

src/workerd/io/compatibility-date.capnp
jasnell, line 739:
```suggestion
      $compatEnableDate("2025-04-28")
```

src/workerd/io/io-context.c++
jasnell: hmm.. I'm wondering if it would be easiest to implement our own task queue similar to what node.js is doing?

ketanhwr:
(just a rebase, no changes yet)

src/workerd/io/io-context.c++
ketanhwr: > > A quick test shows that any microtasks queued during the callback are not drained by v8::platform::PumpMessageLoop
> 
> Yeah that's not good -- those microtasks will end up running whenever the next event is delivered. We will need to drain them, and I suppose we may then have to pump the message loop _again_, and so on.

Thinking about this a bit: should we care about this case at all? Given that FinalizationRegistry callbacks can be executed at some arbitrary point in the future, and the user has no control over when they actually execute, does it really matter that if they enqueue a microtask in the callback which gets executed at the next event?

kentonv: I think it would be quite weird to defer the microtasks to the next I/O context because that means if someone declared their finalization callback async and did `await Promise.resolve()` as the first line, the rest of the callback would actually run in the next I/O context. This could lead to all sorts of surprises I think.

So I think we need to pump the microtask queue until it is empty.

ketanhwr:
TODO: WeakRef tests

ketanhwr:
(rebasing now)

src/workerd/io/io-context.c++
ketanhwr: I've tried to handle this a bit better now (running PumpMessageLoop and MicrotaskCheckpoint in a loop) - and handle isolate termination, but there could be edge cases I might've missed.

src/workerd/io/io-context.c++
```diff
@@ -1101,19 +1098,32 @@ void IoContext::runImpl(Runnable& runnable,
...
+          // Ensure we don't pump the message loop in this case
+          gotTermination = true;
+        }
+      } while ([&]() {
+        if (gotTermination) {
+          // TerminateExecution() doesn't drain any tasks from platform's task queue
+          // so we want to avoid pumping the message loop in this case
+          return false;
+        } else {
+          SuppressIoContextScope noIoCtxt;
```
kentonv, line 1123: At present if `js.pumpMsgLoop()` queues microtasks, it looks like the IoContext will be restored before running those microtasks. I think we don't actually want that, though? Seems like instead any microtasks created by finalization callbacks should themselves run with no IoContext.

jasnell: Oy... that might be tricky. There's a chance that the task queue could still contain other continuations not scheduled by the finalization callbacks. I think we'd want it so that only these specific callbacks moved the IoContext, if it exists, out of the way.

ketanhwr: Hmm, yeah that's true. So it would look something like this:
```
js.runMicrotasks()

{
  SuppressIoContextScope noIoCtxt;
  while (js.pumpMsgLoop()) {
    js.runMicrotasks();
  }
}
```
Don't understand the microtask queue semantics too well at this point - is it possible that some "legitimate" microtasks (i.e. ones which expect valid IoContext) somehow end up running without an IoContext in this case? (Is it possible that a promise created in the fetch handler gets resolved _while_ we're executing a cleanup callback?)

ketanhwr: Would it be simpler to not support this use case entirely? We can document that any microtasks enqueued during FinalizationRegistry cleanup callbacks won't execute deterministically.

> if someone declared their finalization callback async and did await Promise.resolve() as the first line, the rest of the callback would actually run in the next I/O context. This could lead to all sorts of surprises I think.

As per your earlier comment, I personally don't think we should try to accomodate such use of the API (more of a footgun really).

As a side note, I tried asking Claude about how users rely on FinalizationRegistry in the wild (atleast from the code it has been trained on) and are there any legitime uses of supporting the above usecase and this is what it responded with:

> ● In all the code I've been trained on, I've rarely seen purposeful use of microtasks or Promises within FinalizationRegistry callbacks. This is an uncommon pattern for several good reasons:
> 
>   1. FinalizationRegistry callbacks are meant to be simple cleanup functions
>     - Their main purpose is to free resources that GC can't automatically manage
>     - Examples: closing file handles, releasing WebGL textures, etc.
>   2. Asynchronous operations in finalizers are generally considered an anti-pattern
>     - The timing is unpredictable (GC can happen at any time)
>     - There's no guarantee the callback will ever run
>     - The original context may be long gone
>   3. Alternative patterns are preferred
>     - Using explicit cleanup methods (like close(), dispose())
>     - Using the Disposable pattern (which is now standardized with .dispose() and using statements)

jasnell: Hmm... I think even that is a bit unfortunate. I wonder if some extension to the cross-IoContext promise following mechanism might work here? Will have to think about it and stew on it more but...

Currently every IoContext has a unique tag. All promises that are created while an IoContext is current will be tagged with that specific tag. All promises created when an IoContext is not current won't be tagged. We use this as a way to ensure that promise continuations (microtasks) are scheduled while the correct IoContext is active and since we drain the task queue when we exit the IoContext then that typically means the tasks run in the correct context... but we could take this mechanism one step further and tag the *continuation* (the microtask) with the same mechanism. When the microtask runs, if it has no tag, then we move the IoContext (if any) aside to ensure that that the task runs without the IoContext. If it has a tag, then we arrange to enter the correct IoContext (if it still exists) and then run the task. This would probably even make our cross-context promise mechanism even safer, tho it likely has a performance penalty. It would also require another set of floated v8 patches to add the additonal tagging... we'd likely want to implement our own task queue to avoid having to modify v8's. Let me stew over it for a bit more tho.

danlapid: Seems like we're having a really hard time solving it the "right" way.
Can we agree to a more partial more limited solution in the meantime and consider extensions later on as Ketan suggested?

We have real customers, many customers waiting for FinalizationRegistry to be available and they are all looking for the most simple usage of it.

I really feel like we're bike shedding a bit too much on this and possibly dare I say, over engineering?

jasnell: Unfortunately this really isn't a bike shed. It's fairly trivial to get this wrong and introduce a range of bugs.
 
Here's an example case:

```js
const weakReg = new FinalizationRegistry((heldValue) => {
  // new AbortController();  <-- would fail if not commented
  Promise.resolve().then(() => {
    const ac = new AbortController();  // works
    ac.abort();  // works
  });
});

export default {
  async fetch(req, env) {

    globalThis.gc();

    let m = {};
    weakReg.register(m, "foo");

  }
};
```

In the first request we register a value in the `FinalizationRegistry`. On the next request that gets gc'd and the `FinalizationRegistry` callback is invoked. At the top level of that callback, we are not able to successfully create the `AbortController` because there is no `IoContext` and the `AbortController` requires an active `IoContext`.

We schedule the promise continuation, however, which *does* end up running in an `IoContext` except it ends up being the in the `IoContext` for the second request when the gc was called. This ends up being very similar to the cross-request promise continuation problem all over again and can cause similar issues. It also means that the promise continuation can schedule I/O like a `fetch(...)`, `setTimeout(...)`, etc which really shouldn't ever be allowed here.

I think this is likely something we absolutely need to address and it's better to do so now.

danlapid: I think you missed what I meant, I didn't mean that the problem you are suggesting exists here isn't real, but rather was trying to probe whether disabling functionality can let us get around the problem.

For example, disabling queuing microtasks in the FinalizationRegistry

It would eliminate the concerns here afaiu.
If a customer has a good reason to require microtasks in FinalizationRegistry callbacks we can revisit this laterz
But as I said, most of our customers need the most basic of FinalizationRegistry usage.

kentonv: > When the microtask runs, if it has no tag, then we move the IoContext (if any) aside to ensure that that the task runs without the IoContext.

Can we just introduce a tag that means "no IoContext", and when we see that, we schedule the thing in a suppressed-iocontext scope?

But I don't think we need to solve this edge case today. Specifically, the edge case being:
* Create a promise-resolver pair inside a finalization callback.
* Await the promise.
* Resolve the resolver in an I/O context.
* Now the continuation that was supposed to be part of the finalization callback runs with an I/O context when it shouldn't have. Oops.

I think this is an extreme corner case and probably not a big deal. We can fix it later if we want to (with a compat flag).

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,70 @@
...
+        let obj = {};
+        fr.register(obj, '');
+        obj = undefined;
+      })();
+
+      // Ensure obj gets GC'd
+      gc();
+
+      if (frCounter > 0) {
+        await wait(10);
```
jasnell, line 27:
```suggestion
        await scheduler.wait(10);
```

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,70 @@
+let frCounter = 0;
+let weakRefState = { isDereferenced: false, value: null };
+
+const fr = new FinalizationRegistry(() => {
+  ++frCounter;
+});
+
+let wait = (delay) => new Promise((res) => setTimeout(res, delay));
```
jasnell, line 8:
```suggestion
```

This isn't needed. Use `await scheduler.wait(...)` instead

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,70 @@
...
+          JSON.stringify({
+            created: true,
+            value: weakRefState.value,
+          })
+        );
+      }
+      if (url.searchParams.has('gc')) {
+        // Force garbage collection and check if the WeakRef has been cleared
+        gc();
+        await wait(10); // Give GC time to clean up
```
jasnell, line 52:
```suggestion
        await scheduler.wait(10); // Give GC time to clean up
```

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,70 @@
...
+      const value = weakRefState.ref?.deref()?.message || null;
+      weakRefState.isDereferenced = value === null;
+      weakRefState.value = value;
+
+      return new Response(
+        JSON.stringify({
+          isDereferenced: weakRefState.isDereferenced,
+          value: weakRefState.value,
+        })
+      );
```
jasnell, line 65:
```suggestion
      return Response.json({
          isDereferenced: weakRefState.isDereferenced,
          value: weakRefState.value,
      });
```

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,70 @@
...
+        weakRefState.ref = ref;
+        weakRefState.isDereferenced = false;
+        weakRefState.value = ref.deref()?.message || null;
+
+        return new Response(
+          JSON.stringify({
+            created: true,
+            value: weakRefState.value,
+          })
+        );
```
jasnell, line 47:
```suggestion
        return Response.json({
          created: true,
          value: weakRefState.value,
        });
```

src/workerd/io/io-context.c++
jasnell: Sounds good.

src/workerd/io/io-context.c++
```diff
@@ -1114,16 +1115,40 @@ void IoContext::runImpl(Runnable& runnable,
...
-          return js.pumpMsgLoop();
+      }
+
+      // Run FinalizationRegistry cleanup tasks without an IoContext
+      {
+        SuppressIoContextScope noIoCtxt;
+        while (!gotTermination && js.pumpMsgLoop()) {
+          // Check if FinalizationRegistry cleanup callbacks have not breached our limits
+          try {
+            limitEnforcer->requireLimitsNotExceeded();
```
kentonv, line 1126: Use `limitEnforcer->getLimitsExceeded()` rather than throwing an exception and catching it here.

ketanhwr:
(will squash this change later on when rebasing)

src/workerd/jsg/jsg.c++
```diff
@@ -272,6 +274,16 @@ void Lock::terminateExecution() {
   v8Isolate->TerminateExecution();
 }
 
+bool Lock::pumpMsgLoop() {
+  bool tasksRun = false;
+  // Ensure that this runs under limit enforcer scope
+  // We currently disable Minor GC, so this while loop will only ever run at max one iteration
+  while (IsolateBase::from(v8Isolate).pumpMsgLoop()) {
```
kentonv, line 281: I guess this loop is kind of redundant now, as the immediate caller also runs their own loop.

src/workerd/jsg/jsg.c++
```diff
@@ -11,6 +11,8 @@
 #include <workerd/jsg/util.h>
 #include <workerd/util/thread-scopes.h>
 
+#include <libplatform/libplatform.h>
```
kentonv, line 14: New include is no longer used.

kentonv (approved):

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,62 @@
+let frCounter = 0;
+let weakRefState = { isDereferenced: false, value: null };
+
+const fr = new FinalizationRegistry(() => {
+  ++frCounter;
+});
```
jasnell, line 6: It doesn't need to be done in this PR but it would be good to expand the test cases a bit more.

Specifically:

1. Throwing an error from the FinalizationRegistry callback

```js
const fr = new FinalizationRegistry(() => {
  throw new Error('boom');
});
```

2. Scheduling microtasks to ensure that they are not run within an IoContext

```js
const fr = new FinalizationRegistry(() => {
  Promise.resolve().then(() => {
    new AbortController();
  }).catch((err) => {
    // ... should reject because we cannot create an AbortController
    // when we don't have an IoContext
  });
});

const fr = new FinalizationRegistry(() => {
  queueMicrotask(() => {
    // this should also throw if there's no IoContext, which is what we want
    new AbortController();
  })
});
```

3. It should be disallowed to follow a promise made during a request within the handler...

```js
let promise;
const fr = new FinalizationRegistry(() => {
  if (promise) promise.then(() => {});  // this should throw
  // doesn't matter that it won't be resolved, we should not be able to follow it.
});

export default {
  fetch() {
    fr.register({}, 'foo');
    promise = Promise.withResolvers().promise;
    return new Response("ok");
  }
}
```

jasnell: The test should also make sure that the `heldValue` that the callback receives is the expected value.

src/workerd/server/tests/weakref/index.mjs
```diff
@@ -0,0 +1,62 @@
...
+export default {
+  async fetch(request, env, ctx) {
+    const url = new URL(request.url);
+    const test = url.searchParams.get('test') || 'fr';
+
+    if (test === 'fr') {
+      // Test FinalizationRegistry
+      (function () {
+        fr.register({}, '');
+      })();
```
jasnell, line 17: Nit: this can be just a regular block,, but not critical.

```suggestion
      {
        fr.register({}, '');
      }
```

jasnell (approved):

src/workerd/server/tests/weakref/index.mjs
ketanhwr: From my observations: regular block changes when the object will get GC'd (it will get garbage collected during the next request even with explicit `gc()`, v8 is probably a bit conservative with locally scoped variables). If we want to observe immediate GC and FinalizationRegistry execution, we need a function call (or an IIFE like this one)

PR #3565, opened 2025-02-18
Run import tests against each package bundle version.

dom96:
New version of https://github.com/cloudflare/workerd/pull/3451 and https://github.com/cloudflare/workerd/pull/3508

Now that we have multiple package bundle versions and we plan to get more, it's important we test the packages in each bundle. This PR generates import tests for packages in each of the bundles and also ensures that new package bundles contain the older bundle's packages.

Some tests are failing, these have been disabled for now so that we can resolve them via separate PRs.


src/workerd/server/tests/python/import_tests.bzl
```diff
@@ -31,12 +31,34 @@ const unitTests :Workerd.Config = (
...
+    "beautifulsoup4",
+    "cffi",
+    "cryptography",
+    "pycparser",
+    "python-multipart",
+    "soupsieve",
+    "sqlalchemy",
+    "starlette",
+    "tblib",
+]
```
hoodmane, line 45: Can we move this into BUILD.bazel and not duplicate the mechanism for skipping? I think it makes sense to organize them like this rather than 
```
            "micropip": [
                "0.27.1",
            ],
```
which is more verbose.

dom96: Ahh sorry, forgot about this mechanism. Done!

src/workerd/server/tests/python/BUILD.bazel
```diff
@@ -39,24 +39,58 @@ py_wd_test(
...
-        ],
-    },
-)
+[
+    gen_import_tests(
+        PYTHON_IMPORTS_TO_TEST[package_version],
+        python_version,
+        # TODO: Micropip version mismatch for 0.27.0
+        pkg_skip_versions = {
+            "micropip": [
```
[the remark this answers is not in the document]
dom96: Done

src/workerd/server/tests/python/import_tests.bzl
```diff
@@ -32,11 +32,16 @@ def generate_wd_test_file(requirement):
     return WD_FILE_TEMPLATE.format(requirement, requirement)
 
 # to_test is a dictionary from library name to list of imports
-def gen_import_tests(to_test, pkg_skip_versions = {}):
+def gen_import_tests(to_test, python_version, pkg_skip_versions = {}):
     for lib in to_test.keys():
+        if lib.endswith("-tests"):
+            # TODO: Skip testing the import of `-tests` packages like `regex-tests`, they are failing
+            # for now and we need to fix them.
+            continue
```
hoodmane, line 40: Not clear that we ever want to be testing this actually. I think that `pyodide-build-scripts` should be updated to not emit them.

hoodmane (approved):
Thanks!

PR #3568, opened 2025-02-18
Implement just enough to make D1’s Sessions API work in local dev

justin-mp:
D1’s Sessions API for read replicas relies on the `getCurrentBookmark` and `waitForBookmark` methods doing something reasonable rather than throwing an exception.

### DIFF
--- a/src/workerd/io/actor-sqlite.c++
+++ b/src/workerd/io/actor-sqlite.c++
@@ -662,6 +662,50 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::onNoPendingFlush() {
   return outputGate.wait();
 }
 
+kj::Promise<kj::String> ActorSqlite::getCurrentBookmark() {
+  // This is an ersatz implementation that's good enough for local dev with D1's Session API.
+  //
+  // The returned bookmark satisfies the properties that D1 cares about:
+  //
+  // * Later bookmarks sort after earlier bookmarks.  We implement this by incrementing the bookmark
+  // * whenever getCurrentBookmark() is called.
+  //
+  // * Bookmarks from the current workerd session sort after bookmarks from previous sessions.  We
+  //   implement this by saving an ersatz bookmark in the SqliteMetadata table.
+
+  requireNotBroken();
+  uint64_t bookmark = 0;
+  KJ_IF_SOME(b, metadata.getLocalDevelopmentBookmark()) {
+    bookmark = b + 1;
+  }
+  metadata.setLocalDevelopmentBookmark(bookmark);
+
+  // TODO(cleanup): Left-padded number stringification should maybe be in KJ?
+  auto paddedHex = [](uint32_t n) {
+    kj::FixedArray<char, 8> result;
+    for (auto i = 0; i < result.size(); i++) {
+      char digit = n % 16;
+      n /= 16;
+      digit += digit < 10 ? '0' : ('a' - 10);
+      result[result.size() - 1 - i] = digit;
+    }
+    return result;
+  };
+
+  // Turn the bookmark into a format matching what Cloudflare's production returns.
+  constexpr uint32_t uint32_max = kj::maxValue;
+  kj::FixedArray<char, 32> pad;
+  pad.fill('0');
+  return kj::str(paddedHex(bookmark / uint32_max), '-', paddedHex(bookmark % uint32_max), '-',
+      paddedHex(0), '-', pad);
+}
+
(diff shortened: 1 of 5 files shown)


src/workerd/io/actor-sqlite.c++
```diff
@@ -662,6 +662,45 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::onNoPendingFlush() {
...
+  // The returned bookmark satisfies the properties that D1 cares about:
+  //
+  // * Later bookmarks sort after earlier bookmarks.  We implement this by incrementing the bookmark
+  // * whenever getCurrentBookmark() is called.
+  //
+  // * Bookmarks from the current workerd session sort after bookmarks from previous sessions.  We
+  //   implement this by saving an ersatz bookmark in the SqliteMetadata table.
+  //
+  // Note this method returns a bookmark that's in a different format than the one returned by
+  // Cloudflare's production to explicitly discourage parsing bookmarks.
```
kentonv, line 677: Why stop them? If the goal is "we want to be able to change the bookmark format whenever we feel like it", I think realistically this is not sufficient to achieve that: people can and will write code to parse the production format even if it differs from the local format.

Meanwhile, I think using different formats in local vs. production may cause frustration, as someone might write a bunch of code to parse the local format only to find that it breaks the moment they deploy it. Sure, people _should_ deploy to a staging environment first and test there, but realistically many people do assume that if it works locally then it's safe to push to prod. (I certainly do, when working on my own personal projects.)

So I think it might actually be better to generate a format that looks like the production format?

src/workerd/io/actor-sqlite.c++
```diff
@@ -662,6 +662,45 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::onNoPendingFlush() {
   return outputGate.wait();
 }
 
+kj::Promise<kj::String> ActorSqlite::getCurrentBookmark() {
+  // This is an ersatz implementation that's good enough for local dev with D1's Session API.
```
kentonv, line 666: I had to look up the word "ersatz". I had assumed it was the name of some algorithm or something, but apparently it just means "fake". Maybe the word "fake" would be more approachable?

src/workerd/io/actor-sqlite.c++
justin-mp: The code now generates a bookmark in the format of production.

src/workerd/io/actor-sqlite.c++
justin-mp: I decided to call a spade a spade and use localDevelopment instead of ersatz.

jasnell: "ersatz" is such a beautiful word tho. I mourn its loss here. But yeah, probably one that's best to avoid lol.

src/workerd/io/actor-sqlite.c++
```diff
@@ -662,6 +662,50 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::onNoPendingFlush() {
...
+      result[result.size() - 1 - i] = digit;
+    }
+    return result;
+  };
+
+  // Turn the bookmark into a format matching what Cloudflare's production returns.
+  constexpr uint32_t uint32_max = kj::maxValue;
+  kj::FixedArray<char, 32> pad;
+  pad.fill('0');
+  return kj::str(paddedHex(bookmark / uint32_max), '-', paddedHex(bookmark % uint32_max), '-',
```
kentonv, line 699: I guess with this algorithm people will probably never see a non-zero value for the first component.

Arguably we could instead increment the first component of the bookmark every time we reload the DO, and increment the second component for each call to `getBookmark()` within one session?

I'll leave it up to you though whether you think it's worth the effort.

src/workerd/io/actor-sqlite.c++
```diff
@@ -662,6 +662,50 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::onNoPendingFlush() {
...
+      paddedHex(0), '-', pad);
```
kentonv, line 700: We could make these components actually match their meanings in production but again... up to you... and I suppose we can always change it later.

kentonv (approved):

src/workerd/io/actor-sqlite.c++
justin-mp: I'm going to merge it as is since we can always change it if we think it matters.

PR #3569, opened 2025-02-19
Check container status before invoking actor constructor.

kentonv:
This way we can construct the `Container` API object with a correct value for the `.running` property.

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -3439,47 +3439,49 @@ Worker::Actor::Actor(const Worker& worker,
 }
 
 void Worker::Actor::ensureConstructed(IoContext& context) {
-  KJ_IF_SOME(info, impl->classInstance.tryGet<Worker::ActorClassInfo*>()) {
-    context.addWaitUntil(context
-                             .run([this, &info = *info](Worker::Lock& lock) {
-      jsg::Lock& js = lock;
-
-      kj::Maybe<jsg::Ref<api::DurableObjectStorage>> storage;
-      KJ_IF_SOME(c, impl->actorCache) {
-        storage = impl->makeStorage(lock, worker->getIsolate().getApi(), *c);
-      }
-      auto handler = info.cls(lock,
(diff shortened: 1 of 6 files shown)


src/workerd/io/worker.c++
```diff
@@ -3444,49 +3439,67 @@ Worker::Actor::Actor(const Worker& worker,
...
   }
 }
 
+kj::Promise<void> Worker::Actor::ensureConstructedImpl(IoContext& context, ActorClassInfo& info) {
+  InputGate::Lock inputLock = co_await impl->inputGate.wait();
+
+  bool containerRunning = false;
+  KJ_IF_SOME(c, impl->container) {
+    // We need to do an RPC to check if the container is running.
+    // TODO(perf): It would be nice if we could have started this RPC earlier, e.g. in parallel
```
anonrig, line 3454: How impactful is this? I expect nothing major but it's good to know to prioritize it.

src/workerd/io/worker.c++
```diff
@@ -3444,49 +3439,67 @@ Worker::Actor::Actor(const Worker& worker,
...
+    //   with starting the script, and also if we could save the status across hibernations. But
+    //   that would require some refactoring, and this RPC should (eventally) be local, so it's
+    //   not a huge deal.
+    auto status = co_await c.statusRequest(capnp::MessageSize{4, 0}).send();
```
anonrig, line 3458: Can we get away with delaying the await to the first read rather than blocking while this request is completing?

src/workerd/io/worker.c++
kentonv: As the comment says, it's not a huge deal.

src/workerd/io/worker.c++
kentonv: No because it's a synchronous property, so it has to already be there when the app reads it.

Almost all apps will read it immediately anyway.

gabivlj (approved):

jasnell (approved):

anonrig (approved):

PR #3572, opened 2025-02-19
Fixes Python pkg import processing.

dom96:
* Fixes a regression which caused breakage in validator-upload test.
* Fixes another bug: only .py filenames are considered in pkg imports now in JS (this was inconsistent with the validator)

@hoodmane it's possible I missed something here, but I think we need the `importToModuleFilename` check so I brought it back.

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -67,12 +67,18 @@ int ReadOnlyBuffer::read(jsg::Lock& js, int offset, kj::Array<kj::byte> buf) {
   return readToTarget(source, offset, buf);
 }
 
-kj::Array<jsg::JsRef<jsg::JsString>> PyodideMetadataReader::getNames(jsg::Lock& js) {
-  auto builder = kj::heapArrayBuilder<jsg::JsRef<jsg::JsString>>(this->names.size());
+kj::Array<jsg::JsRef<jsg::JsString>> PyodideMetadataReader::getNames(
+    jsg::Lock& js, jsg::Optional<kj::String> maybeExtFilter) {
+  auto builder = kj::Vector<jsg::JsRef<jsg::JsString>>(this->names.size());
   for (auto i: kj::zeroTo(builder.capacity())) {
+    KJ_IF_SOME(ext, maybeExtFilter) {
+      if (!this->names[i].endsWith(ext)) {
+        continue;
+      }
+    }
     builder.add(js, js.str(this->names[i]));
   }
-  return builder.finish();
+  return builder.releaseAsArray();
 }
 
 kj::Array<jsg::JsRef<jsg::JsString>> PyodideMetadataReader::getWorkerFiles(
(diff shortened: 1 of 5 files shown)


hoodmane:
I really don't believe that the logic in `importToModuleFilename` is correct. We'll need to discuss more.

dom96:
Okay, curious to learn why you think it's wrong. But in any case, would be good to merge this asap so we can fix edgeworker CI.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -67,9 +67,15 @@ int ReadOnlyBuffer::read(jsg::Lock& js, int offset, kj::Array<kj::byte> buf) {
   return readToTarget(source, offset, buf);
 }
 
-kj::Array<jsg::JsRef<jsg::JsString>> PyodideMetadataReader::getNames(jsg::Lock& js) {
+kj::Array<jsg::JsRef<jsg::JsString>> PyodideMetadataReader::getNames(
+    jsg::Lock& js, kj::Maybe<kj::String> maybeExtFilter) {
```
anonrig, line 71: Why don't we go with jsg::Optional and use `maybeExtFilter?: string` as the type? 

```suggestion
    jsg::Lock& js, jsg::Optional<kj::String> maybeExtFilter) {
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -343,13 +349,32 @@ kj::Array<kj::StringPtr> ArtifactBundler::getSnapshotImports() {
...
+// This is equivalent to `pkgImport.replace('.', '/') + ".py"`.
+kj::String importToModuleFilename(kj::StringPtr pkgImport) {
+  auto result = kj::heapString(pkgImport.size() + 3);
+  for (auto i = 0; i < pkgImport.size(); i++) {
+    if (pkgImport[i] == '.') {
+      result[i] = '/';
+    } else {
+      result[i] = pkgImport[i];
+    }
+  }
```
anonrig, line 361: Nit:

Instead of copying one by one, we can copy the whole string and later to a replace on the character which would make this code a little bit more readable.

I'm not sure if there is a kj method for this:


```
std::replace(s.begin(), s.end(), '.', '/');
```

anonrig (commented):
Can we add a test so we don't regress once again?

dom96:
> Can we add a test so we don't regress once again?

We have a test in EW. This fixes that test, currently EW CI fails with workerd `main`.

src/workerd/api/pyodide/pyodide.c++
dom96: I'd rather have the caller be explicit when calling this method.

danlapid:
> Okay, curious to learn why you think it's wrong. But in any case, would be good to merge this asap so we can fix edgeworker CI.

We shouldn't be rushing to merge stuff
If there's anything more than a plain simple fix up then let's revert whichever PR broke EW CI and work on re-landing that with internal CI passing

dom96:
Sure, I can do that.

dom96:
https://github.com/cloudflare/workerd/pull/3574

hoodmane:
Okay since the #3574 revert fails ci, I'd be okay to merge this one if you add unit tests here that cover the behavior expected by the validator test.

dom96:
okay, added unit test

hoodmane (approved):
Okay to merge this, but will definitely need more work to refine the added requirement.

dom96:
Need an approval on the internal PR too

dom96:
No worries, got someone else to approve

PR #3575, opened 2025-02-19
Be able to pass environment variables in containers start()

gabivlj:
We are just defining in the container capnp schema the `environmentVariables` property, and the `env` property in the start options. 
The user will be able to pass environment variables like this:
```
this.ctx.container.start({ 
  env: {
    A: "C",
    C: "A",
  },
});
```

If the container image is already configured to have `A=C`, it will override it and set `A` to `B`. New environment variables are just appended. It's up to the container runtime to implement this behaviour correctly.

### DIFF
--- a/src/workerd/api/container.c++
+++ b/src/workerd/api/container.c++
@@ -30,6 +30,23 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
   }
   req.setEnableInternet(options.enableInternet);
 
+  KJ_IF_SOME(env, options.env) {
+    auto list = req.initEnvironmentVariables(env.fields.size());
+    for (auto i: kj::indices(env.fields)) {
+      auto field = &env.fields[i];
+      JSG_REQUIRE(field->name.findFirst('=') == kj::none, Error,
+          "Environment variable names cannot contain '=': ", field->name);
+
+      JSG_REQUIRE(field->name.findFirst('\0') == kj::none, Error,
+          "Environment variable names cannot contain '\\0': ", field->name);
+
+      JSG_REQUIRE(field->value.findFirst('\0') == kj::none, Error,
+          "Environment variable values cannot contain '\\0': ", field->name);
+
+      list.set(i, str(field->name, "=", field->value));
+    }
+  }
+
   IoContext::current().addTask(req.send().ignoreResult());
 
   running = true;
--- a/src/workerd/api/container.h
+++ b/src/workerd/api/container.h
@@ -26,10 +26,11 @@ class Container: public jsg::Object {
   struct StartupOptions {
     jsg::Optional<kj::Array<kj::String>> entrypoint;
     bool enableInternet = false;
+    jsg::Optional<jsg::Dict<kj::String>> env;
 
     // TODO(containers): Allow intercepting stdin/stdout/stderr by specifying streams here.
 
-    JSG_STRUCT(entrypoint, enableInternet);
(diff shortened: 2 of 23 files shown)


src/workerd/api/container.h
```diff
@@ -26,10 +26,11 @@ class Container: public jsg::Object {
   struct StartupOptions {
     jsg::Optional<kj::Array<kj::String>> entrypoint;
     bool enableInternet = false;
+    jsg::Optional<kj::Array<kj::String>> environmentVariables;
```
kentonv, line 29: I think this name should be shortened to `environ` or `env`. I'm ambivalent about which one. Is this the same concept as Workers env, or not? It does seem like the same concept as `process.env` from node, so I guess that leans towards `env`.

I think the format should be an object, like:

```
start({
  env: {
    "PATH": "/usr/bin:/bin",
    "FOO": "bar",
  }
});
```

I suppose we'd need to throw an exception if the name contains `=`, or if either name or value contains `\0`. (Are any other characters technically invalid to include in an env var?)

kentonv: To accept the object format, use `jsg::Optional<jsg::Dict<kj::String>>` as the type.

gabivlj: completely agreed on the JS API changes!

src/workerd/api/container.c++
```diff
@@ -30,6 +30,30 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
   }
   req.setEnableInternet(options.enableInternet);
 
+  KJ_IF_SOME(env, options.env) {
+    auto list = req.initEnvironmentVariables(env.fields.size());
+    for (auto i: kj::indices(env.fields)) {
+      auto field = &env.fields[i];
+      JSG_REQUIRE(!field->name.contains("="), Error,
```
kentonv, line 37: `contains` looks for a substring, but since it's just one character, I would use `findFirst`.

```suggestion
      JSG_REQUIRE(field->name.findFirst('=') == kj::none, Error,
```

src/workerd/api/container.c++
```diff
@@ -30,6 +30,30 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+          "environment variable names cannot include '=' as seen in name of env variable ",
+          field->name);
+
+      for (auto c: field->name) {
```
kentonv, line 41: I'd use `findFirst` for these NUL character checks too.

src/workerd/api/container.c++
```diff
@@ -30,6 +30,30 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
   }
   req.setEnableInternet(options.enableInternet);
 
+  KJ_IF_SOME(env, options.env) {
+    auto list = req.initEnvironmentVariables(env.fields.size());
+    for (auto i: kj::indices(env.fields)) {
+      auto field = &env.fields[i];
+      JSG_REQUIRE(!field->name.contains("="), Error,
+          "environment variable names cannot include '=' as seen in name of env variable ",
```
kentonv, line 38: JavaScript errors should be complete sentences starting with a capital letter. I'd also shorten this a bit:

```suggestion
          "Environment variable names cannot contain '=': ",
```

kentonv (approved):
Looks good but you need to run the formatter.

gabivlj:
rebased

PR #3580, opened 2025-02-20
WPT: Fix harness and enable tests in url/historical.any.js

npaun:
* Fixes `assert_throws_dom`. We copy-pasted some code from WPT that uses `constructor = this.DOMException`, but we could just use `constructor = DOMException` which both works properly and gets rid of a TypeScript error.
* Also gave a better explanation for why we don't enable `url-setters-a-area.window.js` -- it's redundant.


npaun:
Hmm that implementation of assert_throws_dom is still not right. Converting to draft to do another pass on it.

anonrig (approved):

npaun:
OK fixed assert_throws_dom

PR #3581, opened 2025-02-20
add a todo for IsolateBase::getUuid()

anonrig:
https://chromium-review.googlesource.com/c/v8/v8/+/6286748 has landed. We can get rid of getUuid() and use GetHashSeed() for uniqueness check.

We go from 128 bits to 64 bits but since this is only used for testing, I think we can get away with collisions.

### DIFF
--- a/src/workerd/jsg/setup.c++
+++ b/src/workerd/jsg/setup.c++
@@ -731,6 +731,8 @@ kj::Maybe<kj::StringPtr> getJsStackTrace(void* ucontext, kj::ArrayPtr<char> scra
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


fhanau (approved):

jasnell:
Hmm not actually sure this makes sense. getUuid was not added to enable uniqueness checks but for diagnostics and logging purposes. I believe we rely on the UUID in external logging and correlation of some events -- unless that changed at some point. We might be able to use the new API for the same purpose but we'd need to verify

kentonv:
Yeah I'm not sure it would be appropriate to use the hash seed as an identifier... revealing the hash seed might defeat the security it's supposed to provide.

PR #3582, opened 2025-02-20
WPT dom/abort: Fix missing target on events

npaun:
This fix is so simple I'm worried I'm missing something. 
It looks like we just... forgot to implement `Event.target` at all. 
Given that we don't support hierarchical events, we can use exactly the same logic we use for `Event.currentTarget`.
Fixes three WPT tests for AbortSignal.


anonrig (approved):
Let's wait for @jasnell's review before landing this

jasnell (approved):

PR #3590, opened 2025-02-22
Add Response.prototype.type property

jasnell:
WinterTC recently agreed that conformant implementations should, at the very least, support the `'default'` value for `Response.prototype.type` property and, if the runtime implements `Response.error()`, the `'error'` value -- even if the implementation otherwise does not implement CORS.

### DIFF
--- a/types/generated-snapshot/2021-11-03/index.d.ts
+++ b/types/generated-snapshot/2021-11-03/index.d.ts
@@ -1591,6 +1591,8 @@ interface Response extends Body {
   readonly url: string;
   readonly webSocket: WebSocket | null;
   readonly cf?: any;
+  /* [MDN Reference](https://developer.mozilla.org/docs/Web/API/Response/type) */
+  type: "default" | "error";
 }
 interface ResponseInit {
   status?: number;
--- a/types/generated-snapshot/2022-10-31/index.ts
+++ b/types/generated-snapshot/2022-10-31/index.ts
@@ -5694,18 +5694,6 @@ export type PagesPluginFunction<
 // Copyright (c) 2022-2023 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
-export declare abstract class PipelineTransform {
-  /**
-   * transformJson recieves an array of javascript objects which can be
-   * mutated and returned to the pipeline
-   * @param data The data to be mutated
-   * @returns A promise containing the mutated data
-   */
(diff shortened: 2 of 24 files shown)


anonrig (approved):

types/generated-snapshot/2021-11-03/index.d.ts
```diff
@@ -5620,6 +5622,18 @@ declare module "assets:*" {
 // Copyright (c) 2022-2023 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
+declare abstract class PipelineTransform {
+  /**
+   * transformJson recieves an array of javascript objects which can be
```
anonrig, line 5627:
```suggestion
   * transformJson receives an array of javascript objects which can be
```

types/generated-snapshot/2021-11-03/index.d.ts
```diff
@@ -5620,6 +5622,18 @@ declare module "assets:*" {
 // Copyright (c) 2022-2023 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
+declare abstract class PipelineTransform {
```
anonrig, line 5625: how are these types generated? I don't see any code related to this

anonrig (approved):

types/generated-snapshot/2022-10-31/index.ts
```diff
@@ -5692,6 +5694,18 @@ export type PagesPluginFunction<
...
+   * transformJson recieves an array of javascript objects which can be
+   * mutated and returned to the pipeline
+   * @param data The data to be mutated
+   * @returns A promise containing the mutated data
+   */
+  public transformJson(data: object[]): Promise<object[]>;
+}
+// Copyright (c) 2022-2023 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
```
jasnell, line 5708:
```suggestion
```

jasnell:
@irvinebroque ... for the changelog entry once this and https://github.com/cloudflare/workerd/pull/3592 are merged and are released: We are adding the `Response.prototype.type` and `Response.error()` standard fetch APIs to improve spec compliance and interop between WinterTC-aligned runtimes. The value of `Response.prototype.type` will always be `'default'` unless the `Response` is created using `Response.error()`, in which case it will be `'error'`.

PR #3594, opened 2025-02-23
implement TLSSocket and connect from node:tls

anonrig:
Adds `node:tls` module as well as `_tls_wrap`. Includes the following methods/classes:

1. `TLSSocket`
2. `connect()`


danlapid:
Nice!
Glad to see this progressing!

src/node/internal/internal_errors.ts
```diff
@@ -703,6 +703,23 @@ export class ERR_CRYPTO_SIGN_KEY_REQUIRED extends NodeError {
...
 }
 
+export class ERR_TLS_HANDSHAKE_TIMEOUT extends NodeError {
+  constructor() {
+    super('ERR_TLS_HANDSHAKE_TIMEOUT', 'TLS handshake timeout');
+  }
+}
+
+export class ConnResetException extends NodeError {
+  path?: string | undefined;
```
npaun, line 713: For learning purposes: given that the parameter is already optional (`?`), what does adding `| undefined` change in TS?

anonrig: ?: means the parameter can be "not passed" the the function. Example: fn() will work "hello?: string". 

On the other hand if you want to pass undefined as the value hello?: string | undefined handles the case of passing undefined as an argument. example: fn(undefined). 

This doesn't really matter for js because if you don't pass a variable it will be undefined by default but if you're dealing with types that uses "| undefined" syntax, this forces you to be explicit about this too. (Shoutout to types/node package maintainers)

src/node/tls.ts
```diff
@@ -0,0 +1,31 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import { TLSSocket, connect } from 'node-internal:internal_tls_wrap';
+export { TLSSocket, connect } from 'node-internal:internal_tls_wrap';
```
jasnell, line 27: This should work right?
```suggestion
export { TLSSocket, connect };
```

anonrig: Yes

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,714 @@
...
+    // Not yet supported. Can be implemented in the future.
+    throw new ERR_OPTION_NOT_IMPLEMENTED('options.requestOCSP');
+  }
+
+  if (tlsOptions.secureContext) {
+    // TODO(soon): Investigate supporting this.
+    throw new ERR_OPTION_NOT_IMPLEMENTED('options.secureContext');
+  }
+
+  if (tlsOptions.pskCallback) {
```
jasnell, line 245: I would make these `if ({option} !== undefined` so that no value is accepted.

anonrig: This can be too much restrictive since some use default values (such as false). Even in tls_wrap, Node.js uses default values and passes false. For example, this will fail on `rejectUnauthorized: false` which is used a lot afaict.

jasnell: Yes, but things like { pskCallback: false} does not make any sense. Nor does tlsOptions.ALPNProtocols = ''.. etc. The options whose values are not booleans should not just ignore any falsy value.

anonrig: Yes. It definitely makes sense to be restrictive on them.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,714 @@
...
+  // immediately after onocspresponse(). Ignore it.
+  if (!this._handle) return;
+
+  this._secureEstablished = true;
+  if (
+    this._tlsOptions.handshakeTimeout &&
+    this._tlsOptions.handshakeTimeout > 0
+  )
+    this.setTimeout(0, this._handleTimeout.bind(this));
+  this.emit('secure');
```
jasnell, line 493: Node.js will often emit events using `process.nextTick(...)` to defer execution. Is that needed on these?

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,714 @@
...
+  const cb = args[1] as VoidFunction | undefined;
+
+  // If args[0] was options, then normalize dealt with it.
+  // If args[0] is port, or args[0], args[1] is host, port, we need to
+  // find the options and merge them in, normalize's options has only
+  // the host/port/path args that it knows about, not the tls options.
+  // This means that options.host overrides a host arg.
+  if (listArgs[1] !== null && typeof listArgs[1] === 'object') {
+    Object.assign(options, listArgs[1]);
+  } else if (listArgs[2] !== null && typeof listArgs[2] === 'object') {
```
jasnell, line 601: Fun fact... you don't need the `!== null` check here. `Object.assign(options, null)` is handled just fine. In fact, you could get by without any type check here.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,714 @@
...
+
+function onConnectSecure(this: TLSSocket): void {
+  this.authorized = true;
+  this.secureConnecting = false;
+  this.emit('secureConnect');
+
+  this[kIsVerified] = true;
+  const session = this[kPendingSession];
+  this[kPendingSession] = null;
+  if (session) this.emit('session', session);
```
jasnell, line 616: Another one... should these be wrapped in `process.nextTick`/`queueMicrotask`?

src/workerd/api/node/tests/tls-nodejs-test.js
```diff
@@ -0,0 +1,163 @@
...
+      signal,
+    });
+
+    const assertAbort = async (socket, testName) => {
+      try {
+        await once(socket, 'close');
+        fail(`close ${testName} should have thrown`);
+      } catch (err) {
+        strictEqual(err.name, 'AbortError');
+      }
```
jasnell, line 45: Why not use `assert.rejects()`?

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,714 @@
...
+
+    // To prevent assertion in afterConnect() and properly kick off readStart
+    this.connecting = socket.connecting || !socket._handle;
+    socket.once('connect', () => {
+      this.connecting = false;
+      this.emit('connect');
+    });
+  }
+
+  wrap?.on('error', (err: Error): void => {
```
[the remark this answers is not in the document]
anonrig: This should be on.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,714 @@
...
+  this: TLSSocket,
+  socket?: Socket,
+  wrap?: Socket | null
+): void {
+  const options = this._tlsOptions;
+
+  // Let's disable floating promises error.
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  socket?._handle?.socket.opened.then(this._finishInit.bind(this));
+  socket?.on('error', onerror.bind(this));
```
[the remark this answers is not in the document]
anonrig: This should be on as well. Since it can emit error event, and still needs to listen to it.

src/node/internal/internal_tls_wrap.ts
anonrig: This function is called on `socket.opened.then()` no need.

src/node/internal/internal_tls_wrap.ts
anonrig: I don't think it is needed. Will look into this once we have more test coverage.

src/workerd/api/node/tests/tls-nodejs-test.js
```diff
@@ -0,0 +1,163 @@
...
+        lookup() {},
+      });
+      strictEqual(socket.allowHalfOpen, false);
+    }
+
+    {
+      const { promise, resolve } = Promise.withResolvers();
+      const socket = tls.connect(
+        {
+          port: 8887,
```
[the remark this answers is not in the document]
anonrig: thank you

src/workerd/api/node/tests/tls-nodejs-test.js
jasnell: you marked this resolved but it's still not using `assert.rejects`?

src/workerd/api/node/tests/tls-nodejs-test.js
```diff
@@ -0,0 +1,210 @@
...
+      setImmediate(() => ac.abort());
+      await assertAbort(socket, 'testConstructorPostTick');
+    }
+
+    await postAbort();
+    await preAbort();
+    await tickAbort();
+    await testConstructor();
+    await testConstructorPost();
+    await testConstructorPostTick();
```
jasnell, line 103: I'd generally prefer these to be their own individual tests

src/workerd/api/node/tests/tls-nodejs-test.js
```diff
@@ -0,0 +1,210 @@
...
+  },
+};
+
+// Tests are taken from
+// https://github.com/nodejs/node/blob/304743655d5236c2edc39094336ee2667600b684/test/parallel/test-tls-connect-allow-half-open-option.js
+export const connectAllowHalfOpenOption = {
+  async test() {
+    {
+      const socket = tls.connect({ port: 42, lookup() {} });
```
jasnell, line 112: Is lookup actually going to be called?

src/workerd/api/node/tests/tls-nodejs-tcp-server.js
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+// This file is used as a sidecar for the net-nodejs-test tests.
+// It creates 2 TCP servers to act as a source of truth for the node:net tests.
```
jasnell, line 6: This says it creates two tcp servers but there's only one created in this script?

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+    this[kIsVerified] = false;
+  });
+};
+
+TLSSocket.prototype._init = function _init(this: TLSSocket): void {
+  const options = this._tlsOptions;
+
+  this.on('error', onerror.bind(this));
+
+  // This is thrown from net.Socket class.
```
jasnell, line 323: Nit "thrown" is typically used only for errors. "emitted" is likely the better word choice.

src/workerd/api/node/tests/tls-nodejs-test.js
anonrig: This is directly from node.js tests. I deliberately didn't update them to follow up for changes easily. I can split them if you insist.

src/workerd/api/node/tests/tls-nodejs-tcp-server.js
anonrig: Outdated comment... I'll remove it.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+  // so even after .destroy(), the handshake done info callback occurs
+  // immediately after onocspresponse(). Ignore it.
+  if (!this._handle) return;
+
+  this._secureEstablished = true;
+  if (
+    this._tlsOptions.handshakeTimeout &&
+    this._tlsOptions.handshakeTimeout > 0
+  )
+    this.setTimeout(0, this._handleTimeout.bind(this));
```
jasnell, line 422:
```suggestion
  this.setTimeout(0, this._handleTimeout.bind(this));
```

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+TLSSocket.prototype._start = function _start(this: TLSSocket): void {
+  if (this.connecting) {
+    this.once('connect', this._start.bind(this));
+    return;
+  }
+
+  // Socket was destroyed before the connection was established
+  if (!this._handle) return;
+
+  // this._handle.start();
```
jasnell, line 435: Why is this commented out?

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+};
+
+TLSSocket.prototype.setServername = function setServername(
+  this: TLSSocket,
+  name: string
+): void {
+  validateString(name, 'name');
+  // Does not apply to Cloudflare Workers.
```
jasnell, line 443: Heh, interesting... this appears to have never been documented. https://nodejs.org/docs/latest/api/tls.html

That said, we actually *do* eventually here need to have a way of specifying both the SNI and ALPN extensions. Pipefitter currently does not provide us a way on the internal system and I'm not sure if kj's TLS impl provides a way, but it is something we will need sooner than later.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+  // Do nothing. We don't support setting session.
+};
+
+// @ts-expect-error TS2322 Inconsistencies between @types/node
+TLSSocket.prototype.getPeerCertificate = function (
+  _detailed?: boolean
+): ReturnType<TLSSocketType['getPeerCertificate']> {
+  // Returns an object representing a peer certificate.
+  // This function is not supported.
+  throw new Error('getPeerCertificate is not implemented');
```
jasnell, line 456: Feels weird having `getCertificate()` return `null` but  `getPeerCertificate()` throws.

jasnell: Especially since `getPeerX509Certificate` returns `undefined` and not `null` or throwing lol

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+  // This function is not supported.
+  return undefined;
+};
+
+TLSSocket.prototype.setKeyCert = function (
+  this: TLSSocket,
+  _context: unknown
+): void {
+  // Changing private key and certificate to be used with TCP connection
+  // is not supported due to the limitations of connect() api.
```
jasnell, line 489: Should this throw?

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+  'getProtocol',
+  'getSession',
+  'getTLSTicket',
+  'isSessionReused',
+  'enableTrace',
+].forEach((method) => {
+  // @ts-expect-error TS7053 Omitting...
+  TLSSocket.prototype[method] = function (): null {
+    // None of these functions are supported by connect() api.
+    return null;
```
jasnell, line 507: It's not clear why these are handled like this but the `getX509Certificate`, etc cases above are separated out.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+}
+
+// Arguments: [port,] [host,] [options,] [cb]
+export function connect(...args: unknown[]): TLSSocket {
+  args = normalizeConnectArgs(args);
+  let options = args[0] as NormalizedConnectionOptions;
+  const cb = args[1] as VoidFunction | undefined;
+
+  options = {
+    minDHSize: 1024,
```
jasnell, line 578: We don't support setting the `minDHSize` ... perhaps we shouldn't default it and should either throw or ignore it?

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+    // be accessible when the `tlsClientError` event is emitted.
+    this.destroy(err);
+  } else {
+    // Emit error
+    this._emitTLSError(err);
+  }
+}
+
+// @ts-expect-error TS2323 Cannot redeclare error.
+export function TLSSocket(
```
jasnell, line 163: Add a small comment explaining that we're using old style function classes for node.js compat

src/node/internal/internal_tls_wrap.ts
anonrig: getPeerCertificate() doesn't return null. It always returns a value in TLS (ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L312) but whereas in getCertificate() can return null. That's why I didn't throw an error on getCertificate().

Yes, getPeerCertificate() returns undefined. I explicitly tried to work around with types/node since most people will use types/node in their code. REf: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L437

src/node/internal/internal_tls_wrap.ts
```diff
@@ -0,0 +1,644 @@
...
+TLSSocket.prototype.setServername = function setServername(
+  this: TLSSocket,
+  name: string
+): void {
+  validateString(name, 'name');
+  // Does not apply to Cloudflare Workers.
+};
+
+TLSSocket.prototype.setSession = function (_session: string | Buffer): void {
+  // Do nothing. We don't support setting session.
```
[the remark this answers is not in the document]
anonrig: Yeah this is kind of open to discussion I guess. We can throw...

src/node/internal/internal_tls_wrap.ts
anonrig: Agreed

src/node/internal/internal_tls_wrap.ts
anonrig: I was waiting for the correct test where we actually needed this, but it seems we can just make this function a noop for now.

src/node/internal/internal_tls_wrap.ts
anonrig: This won't work because previous line is an if statement. I guess someone forgot to add curly brackets when adding this to Node.js.

src/node/internal/internal_tls_wrap.ts
anonrig: Node.js does validate minDHSize, and that's why I added a default value here.

src/node/internal/internal_net.ts
```diff
@@ -1303,7 +1325,7 @@ function addClientAbortSignalOption(
   }
 
   if (signal.aborted) {
-    queueMicrotask(onAbort);
+    onAbort();
```
jasnell, line 1328: This might be a limitation in Node.js also but... there is `signal.reason` which provides access to the error (or other value) that the `AbortSignal` was aborted with. If the `signal` is aborted then we really ought to be propagating that `reason`

src/node/internal/internal_tls_wrap.ts
anonrig: This is directly from Node.js. It is separated originally in the codebase because they are all properties of this._handle[xxx]

src/node/internal/internal_tls_wrap.ts
anonrig: I'm not sure. This might break stuff, or maybe not...

src/workerd/api/node/tests/tls-nodejs-test.js
jasnell: Node.js tests are awful. There's an opportunity to improve both there and here. Not going to insist on it tho

src/node/internal/internal_tls_wrap.ts
jasnell: Ah, didn't realize it was an if block. Yeah, please add curly's around it so it is clearer.

src/node/internal/internal_tls_wrap.ts
jasnell: My point here is that this ends up being very inconsistent. We don't implement any of these but one throws, one returns `undefined`, one returns `null`... if this is intended to make it work like Node.js because Node.js happens to have the same quirk, then please add a comment documenting that. Otherwise, I think it's better to err on the side of consistency here for all of the APIs we do not implement.

src/node/internal/internal_tls_wrap.ts
jasnell: A comment to that effect would be helpful

src/node/internal/internal_tls_wrap.ts
jasnell: Yes, but Node.js also makes use of the option which has a very real impact on the TLS handshake. We are not using it and I really don't want to give the impression we are by silently ignoring it. I would throw if the option is provided.

src/workerd/api/node/tests/tls-nodejs-test.js
anonrig: Resolve this for now, and I'll try to open a PR to node.js for this soon.

src/workerd/api/node/tests/tls-nodejs-test.js
anonrig: Yes it will be called, unless there is a bug with node:net tls.connect() calls net.connect() and passes options.lookup to the net.connect() which handles and checks and calls the lookup function.

src/node/internal/internal_tls_wrap.ts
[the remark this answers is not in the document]
anonrig: I'm keeping it as it is, and we can revisit later.

src/node/internal/internal_tls_wrap.ts
anonrig: I'm keeping it as it is for now, and we can revisit this later.

src/node/internal/internal_tls_wrap.ts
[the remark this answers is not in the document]
jasnell: Just keep in mind that adding or removing a throw ends up likely being a breaking change.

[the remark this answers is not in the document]
anonrig: This could potentially make other packages unusable if we throw. In that case, either we can go with a compat_flag for TLS, or not throw right now. Which one do you prefer?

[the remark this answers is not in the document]
jasnell: Since node.js does not appear to be throwing currently when this is invalid, let's not throw for now. But we likely need to have a discussion in Node.js about what to do with this undocumented API and what behaviors it should actually implement. Silently ignoring an invalid session feels less than ideal to me.

jasnell (approved):

PR #3596, opened 2025-02-24
Add compat flag for queue consumers to not block on waitUntil'ed work

a-robinson:
The intent is that the event can be considered done when the
queue() handler's returned promise resolves, allowing acked/nacked
message info to be returned back to the queues infrastructure sooner,
rather than requiring all waitUntil'ed work to finish first. This has
two important benefits:

1. This will keep consumer invocations from getting slowed down by
   slow/hung waitUntil'ed tasks.
1. This will keep consumer invocations from failing due to failing
   waitUntil tasks.

This notably makes queue() handlers behave more like fetch() handlers, which should make them more intuitive for most users.

### DIFF
--- a/src/workerd/api/queue.c++
+++ b/src/workerd/api/queue.c++
@@ -356,9 +356,7 @@ QueueMessage::QueueMessage(
       attempts(message.attempts),
       result(result) {}
 
-jsg::JsValue QueueMessage::getBody(jsg::Lock& js) {
-  return body.getHandle(js);
-}
+jsg::JsValue QueueMessage::getBody(jsg::Lock& js) { return body.getHandle(js); }
 
 void QueueMessage::retry(jsg::Optional<QueueRetryOptions> options) {
   if (result->ackAll) {
@@ -550,7 +548,7 @@ kj::Promise<WorkerInterface::CustomEvent::Result> QueueCustomEventImpl::run(
   // It's a little ugly, but the usage of waitUntil (and finishScheduled) down below are here so
   // that users can write queue handlers in the old addEventListener("queue", ...) syntax (where we
   // can't just wait on their addEventListener handler to resolve because it can't be async).
-  context.addWaitUntil(context.run(
+  auto runProm = context.run(
       [this, entrypointName = entrypointName, &context, queueEvent = kj::addRef(*queueEventHolder),
           &metrics = incomingRequest->getMetrics(),
           props = kj::mv(props)](Worker::Lock& lock) mutable {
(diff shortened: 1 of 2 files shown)


src/workerd/api/queue.c++
```diff
@@ -504,10 +512,25 @@ jsg::Ref<QueueEvent> startQueueEvent(EventTarget& globalEventTarget,
     }
     globalEventTarget.dispatchEventImpl(lock, event.addRef());
     event->setCompletionStatus(QueueEvent::CompletedSuccessfully{});
+    if (FeatureFlags::get(js).getQueueConsumerNoWaitForWaitUntil()) {
+      // TODO(now): It doesn't really make sense for event-listener scripts NOT to wait on
```
a-robinson, line 516: This is the biggest question here, and needs a decision before releasing this.

src/workerd/api/queue.c++
```diff
@@ -544,76 +567,113 @@ kj::Promise<WorkerInterface::CustomEvent::Result> QueueCustomEventImpl::run(
...
+      KJ_IF_SOME(handlerProm, queueEvent->exportedHandlerProm) {
+        return handlerProm.then([]() { return EventOutcome::OK; });
       }
+      return EventOutcome::OK;
+    }).catch_([](kj::Exception&& e) {
+      return EventOutcome::EXCEPTION;
+    }).exclusiveJoin(timeoutPromise.then([] {
+      return EventOutcome::EXCEEDED_CPU;
+    })).exclusiveJoin(context.onAbort().then([] {
+      // TODO(now): It seems like EXCEPTION might be better? But this is what we return today in
```
a-robinson, line 608: This is less critical, but now seems like a fine time for us to start reporting "aborted" invocations as exceptions rather than exceeded CPU errors. That would bring us inline with how fetch handlers behave (whereas now we're copying how cron triggers behave).

jbwcloudflare (approved):

src/workerd/api/queue.c++
a-robinson: (an important subtext here being that it's unlikely the `addEventListener` format is used much, so minimizing the effort involved in this is valuable)

src/workerd/api/queue.c++
jbwcloudflare: agreed -- this seems like a more accurate "outcome" and now is a good time, given the compat flag

src/workerd/api/queue.c++
jbwcloudflare: I don't have strong feelings on this one FWIW -- I tend to agree that the `addEventListener` format is very stale/unlikey-to-be-used (for Queue consumers at least) and not worth too much of your effort!

jasnell: hmm.. not sure I have much of an opinion on this one. Maybe @kentonv ?

a-robinson: Thanks for chiming in!

FWIW the consensus amongst the team this morning was just to "ignore the compat flag and run with the old behavior" since we've never heard of anyone actually using service worker syntax for queues.

src/workerd/api/queue.c++
a-robinson: Done - we now record these invocations as exceptions rather than cpu exceeded errors.

a-robinson:
Fixed up the code to implement the decisions for the two questions above. There's a lot of test coverage in the corresponding internal PR, so from my perspective this is good to go.

a-robinson:
Ughh the linter doesn't like something about the formatting -- forgive any small force pushes for a little bit while I try to fix it since I guess the version of the formatter that I'm running locally isn't up to date anymore, but I've never been able to get the tooling in this repo to work :|

a-robinson:
> Ughh the linter doesn't like something about the formatting -- forgive any small force pushes for a little bit while I try to fix it since I guess the version of the formatter that I'm running locally isn't up to date anymore, but I've never been able to get the tooling in this repo to work :|

Ok, looks like it was just the one line that the linter was unhappy with -- it's passing now.

So this just needs review from someone on the runtime team, since @jbwcloudflare doesn't count.

src/workerd/api/queue.c++
```diff
@@ -544,76 +560,124 @@ kj::Promise<WorkerInterface::CustomEvent::Result> QueueCustomEventImpl::run(
...
+      return EventOutcome::OK;
+    }).catch_([](kj::Exception&& e) {
+      return EventOutcome::EXCEPTION;
+    }).exclusiveJoin(timeoutPromise.then([] {
+      return EventOutcome::EXCEEDED_CPU;
+    })).exclusiveJoin(context.onAbort().then([] {
+      // This is a change from the outcome we returned on abort before the compat flag, but better
+      // matches the behavior of fetch() handlers and the semantics of what's actually happening.
+      return EventOutcome::EXCEPTION;
+    }, [](kj::Exception&&) { return EventOutcome::EXCEPTION; }));
```
jasnell, line 606: nit... just noting, following these promise chains melts the brain a bit.  A comment at the top explaining the flow here a bit more would be helpful for future newcomers looking into this code.

src/workerd/io/compatibility-date.capnp
```diff
@@ -703,4 +703,13 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("nonclass_entrypoint_reuses_ctx_across_invocations")
       $compatEnableDate("2025-03-10");
   # Creates a unique ExportedHandler for each call to `export default` thus allowing a unique ctx per invocation
+
+  queueConsumerNoWaitForWaitUntil @75 :Bool
+      $compatEnableFlag("queue_consumer_no_wait_for_wait_until")
+      $compatDisableFlag("queue_consumer_wait_for_wait_until");
```
jasnell, line 709: Is it intentional that this does not have a default on date?

a-robinson: It was intentional since PM had wanted to see how this worked for some specific users before making it the default, but I'll double check.

Assuming that hasn't changed, I've left a comment to that effect for now.

src/workerd/api/queue.c++
a-robinson: Yeah, very fair. Added some comments to try to break it down.

a-robinson:
Thanks for the review!

anonrig (approved):

PR #3602, opened 2025-02-25
Adds Pyodide version check to prevent incorrect bundle release

dom96:
Just a simple check to ensure an incorrectly build version of the pyodide bundle cannot be released.

- [x] Rebuild 0.26.0 bundle to ensure this works.


hoodmane (approved):
Thanks!

dom96:
Not going to push the rebuilt bundle because it changes jaeger tracing and I'd rather avoid having to create a EW PR to fix that. The important part is that the rebuilt bundle passes the other tests.

PR #3603, opened 2025-02-25
Implement node:crypto sign and verify APIs

jasnell:
Implements the `node:crypto` `createSign(...)`/`createVerify(...)`/`crypto.sign(...)`/`crypto.verify(...)` APIs. RSA and ed25519 are currently test to work, other key types aren't verified yet. DSA likely does not work, for instance. Will be working on expanding tests next but this ought to be able to land.

Reviewers: some familiarity with the Node.js implementation is likely necessary to review this for correctness.

### DIFF
--- a/src/node/internal/crypto_sign.ts
+++ b/src/node/internal/crypto_sign.ts
@@ -1,0 +1,440 @@
+// Copyright (c) 2017-2023 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+//
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+/* TODO: the following is adopted code, enabling linting one day */
+/* eslint-disable */
+
+import {
+  default as cryptoImpl,
+  type SignHandle,
+  type VerifyHandle,
+  type CreateAsymmetricKeyOptions,
+} from 'node-internal:crypto';
+
+import { validateString } from 'node-internal:validators';
+
+import { Buffer } from 'node-internal:internal_buffer';
+
+import {
+  KeyObject,
+  isKeyObject,
+  getKeyObjectHandle,
+  createPrivateKey,
+  createPublicKey,
+} from 'node-internal:crypto_keys';
+
+import {
+  ERR_CRYPTO_SIGN_KEY_REQUIRED,
+  ERR_CRYPTO_INVALID_KEY_OBJECT_TYPE,
+  ERR_INVALID_ARG_TYPE,
+  ERR_INVALID_ARG_VALUE,
+} from 'node-internal:internal_errors';
+
+import { Writable } from 'node-internal:streams_writable';
+import { isArrayBufferView } from 'node-internal:internal_types';
+
+export interface SignOptions {}
+
+export interface Sign extends Writable {
+  [kHandle]: SignHandle;
+  update(data: any): void;
+  sign(privateKey: any): any;
+}
+export interface Verify extends Writable {
+  [kHandle]: VerifyHandle;
+  update(data: any): void;
+  verify(publicKey: any, signature: any): any;
+}
+
+const kHandle = Symbol('kHandle');
+
+// Uses old-style class syntax for Node.js compatibility
+export const Sign = function (this: Sign, algorithm: string, options: any) {
+  if (!(this instanceof Sign)) return new Sign(algorithm, options);
+  validateString(algorithm, 'algorithm');
+  this[kHandle] = new cryptoImpl.SignHandle(algorithm);
+  Writable.call(this, options);
+  return this;
+} as any as {
+  new (algorithm: string, options?: SignOptions): Sign;
+};
+Object.setPrototypeOf(Sign.prototype, Writable.prototype);
+Object.setPrototypeOf(Sign, Writable);
+
+Sign.prototype._write = function _write(
+  chunk: string | ArrayBufferView,
+  encoding: string | undefined | null,
+  callback: (err?: any) => void
+) {
+  this.update(chunk, encoding);
+  callback();
+};
+
+Sign.prototype.update = function (
+  this: Sign,
+  data: string | ArrayBufferView,
+  encoding?: string
+) {
+  if (typeof data === 'string') {
+    data = Buffer.from(data, encoding);
+  }
+  if (!isArrayBufferView(data)) {
+    throw new ERR_INVALID_ARG_TYPE(
+      'data',
+      ['string', 'Buffer', 'TypedArray', 'DataView'],
+      data
+    );
+  }
+  this[kHandle].update(data);
+  return this;
+};
+
+function getIntOption(name: string, options: any): number | undefined {
+  const value = options[name];
+  if (value !== undefined) {
+    if (value === value >> 0) {
+      return value;
+    }
+    throw new ERR_INVALID_ARG_VALUE(`options.${name}`, value);
+  }
+  return undefined;
+}
+
+function getDSASignatureEncoding(options: any): number {
+  if (typeof options === 'object') {
+    const { dsaEncoding = 'der' } = options;
+    if (dsaEncoding === 'der')
+      return 0; // kSigEncDER;
+    else if (dsaEncoding === 'ieee-p1363') return 1; // kSigEncP1363
+    throw new ERR_INVALID_ARG_VALUE('options.dsaEncoding', dsaEncoding);
+  }
+
+  return 0;
+}
+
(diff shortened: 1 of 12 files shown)


src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,444 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+/* TODO: the following is adopted code, enabling linting one day */
+/* eslint-disable */
```
anonrig, line 27:
```suggestion
```

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,444 @@
...
+      'data',
+      ['string', 'Buffer', 'TypedArray', 'DataView'],
+      data
+    );
+  }
+  this[kHandle].update(data);
+  return this;
+};
+
+function getIntOption(name: string, options: any): number | undefined {
```
anonrig, line 113:
```suggestion
function getIntOption(name: string, options: Record<string, unknown>): number | undefined {
```

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,444 @@
...
+  if (value !== undefined) {
+    if (value === value >> 0) {
+      return value;
+    }
+    throw new ERR_INVALID_ARG_VALUE(`options.${name}`, value);
+  }
+  return undefined;
+}
+
+function getDSASignatureEncoding(options: any) {
```
anonrig, line 124:
```suggestion
function getDSASignatureEncoding(options: unknown): number {
```

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,444 @@
...
+    if (dsaEncoding === 'der')
+      return 0; // kSigEncDER;
+    else if (dsaEncoding === 'ieee-p1363') return 1; // kSigEncP1363
+    throw new ERR_INVALID_ARG_VALUE('options.dsaEncoding', dsaEncoding);
+  }
+
+  return 0;
+}
+
+function getPrivateKey(options: any): CryptoKey {
```
anonrig, line 136:
```suggestion
function getPrivateKey(options: unknown): CryptoKey {
```

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,444 @@
...
+  } else {
+    return getKeyObjectHandle(
+      createPrivateKey(options as CreateAsymmetricKeyOptions)
+    );
+  }
+}
+
+Sign.prototype.sign = function (
+  this: Sign,
+  options: any,
```
anonrig, line 165: Can you replace all any with options please?

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,444 @@
...
+} as any as {
+  new (algorithm: string, options?: SignOptions): Verify;
+};
+Object.setPrototypeOf(Verify.prototype, Writable.prototype);
+Object.setPrototypeOf(Verify, Writable);
+
+Verify.prototype._write = function _write(
+  chunk: string | ArrayBufferView,
+  encoding: string | undefined | null,
+  callback: (err?: any) => void
```
anonrig, line 208:
```suggestion
  callback: (err?: unknown) => void
```

vicb:
@jasnell it would be great to open docs PR whenever new API are implemented.

I don't see https://github.com/cloudflare/workerd/pull/3463 documented at https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/

It's probably even more important if only some algos are tested/supported.

Also what "does not work" mean in:

> RSA and ed25519 are currently test to work, other key types aren't verified yet. DSA likely does not work, for instance.

Does it throws or could it return incorrect results? (if the later maybe we should throw for all that is not tested?)

Thanks

src/node/internal/streams_writable.d.ts
```diff
@@ -0,0 +1,164 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE. */
+
+// Derived from DefinitelyTyped https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/stream.d.ts#L1060
+
+/* TODO: the following is adopted code, enabling linting one day */
+/* eslint-disable */
```
anonrig, line 24:
```suggestion
```

jasnell:
> I don't see https://github.com/cloudflare/workerd/pull/3463 documented at [developers.cloudflare.com/workers/runtime-apis/nodejs/crypto](https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/)

Yep. Haven't gotten to that yet. It hadn't yet been deployed to production until some time last week while I was at TC-39. Opening the PR to add the docs is on my todo list for this week.

> Also what "does not work" mean in:

Likely throws because I don't think the key types are supported. It should not return incorrect results. It should error noisily.

src/node/internal/crypto_sign.ts
jasnell: Going to keep this until I'm sure it's all working correctly.

src/node/internal/crypto_sign.ts
jasnell: It's not a `Record<string, unknown>`. It might be a `KeyObject`, and `CryptoKey`, etc.

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,441 @@
...
+        data,
+        rsaPadding,
+        pssSaltLength,
+        dsaSigEnc
+      )
+    );
+  }
+
+  try {
+    callback(
```
anonrig, line 335: queueMicrotask is missing from here. we have queueMicrotask in catch()

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,441 @@
...
+          rsaPadding,
+          pssSaltLength,
+          dsaSigEnc
+        )
+      )
+    );
+  } catch (err) {
+    queueMicrotask(() => callback(err));
+  }
+  return; // explicit return is necessary to squelch typescript warning.
```
anonrig, line 351: This is actually due to a eslint rule. We should just remove that rule and make this function return `Buffer | void`

Ref: https://github.com/typescript-eslint/typescript-eslint/issues/5752

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,441 @@
...
+}
+
+export function verify(
+  algorithm: string | null | undefined,
+  data: BufferSource,
+  options: any,
+  signature: BufferSource,
+  callback?: VerifyCallback
+): boolean | undefined {
+  if (algorithm != null) {
```
anonrig, line 361: Without this, we allow algorithm to be string | null | undefined

```suggestion
  if (algorithm !== undefined) {
```

src/node/internal/crypto_sign.ts
```diff
@@ -0,0 +1,441 @@
...
+    );
+  }
+
+  if (!options) {
+    throw new ERR_CRYPTO_SIGN_KEY_REQUIRED();
+  }
+
+  let key: CryptoKey;
+  if (options instanceof CryptoKey) {
+    key = options as CryptoKey;
```
anonrig, line 389: Since you have the instanceof check, this is not needed

```suggestion
    key = options;
```

src/node/internal/streams_writable.d.ts
```diff
@@ -0,0 +1,164 @@
...
+
+import { EventEmitter } from 'node:events';
+
+interface WritableOptions {
```
anonrig, line 28: We don't need to define this here. We can just use types/node

Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/stream.d.ts#L991

src/workerd/api/node/tests/crypto_sign-test.js
```diff
@@ -0,0 +1,146 @@
...
+  '27bcfe1b306291188f97ced0e3ceaa5ded1ae1406ec30e46a18434e55dc6' +
+  'c9f237e26b124bf7ec77e54483d7782b805aa9a74bbe0f4aa8658d7620e4' +
+  'd3a305777b5dc8262c675bf23c8dc5acfe8c4fa1ca8acdd956cdcbdf1fb2' +
+  'f879b37dc0ed8ec51815ff02b98eefde44ac79f886902ea69e5ed2561e9e' +
+  'eb2a74ec1de677b285f8108f25e9f34cc826fbbd1ad091d231dc73eaf28a' +
+  'e02f09ad84ec9a38d8a7e28bea26fb7d4db20eecd075b5b261ca7320af94' +
+  'cd58ba4b26d895df4fd6f68bd4d82acdcb35557012e2f69739ce8cf4a66e' +
+  'bf4550ee50f6c9cec642d66fef71495a';
+
+export const rsaSignVerifyObjects = {
```
anonrig, line 23: If these are from nodejs tests, can you add link to the test file as a comment?

src/node/internal/crypto_sign.ts
jasnell: Yes, that is intentional. Node.js accepts either `null` or `undefined` for this option.

src/node/internal/streams_writable.d.ts
jasnell: Well, it was needed to make this compile ;-) ... Likely need to rework the streams types definitions separately. For now let's leave this.

src/workerd/api/node/tests/crypto_sign-test.js
jasnell: The fixtures are from node.js' tests but the actual test I rolled myself. I haven't started porting Node.js' actual tests yet.

src/node/internal/crypto_sign.ts
jasnell: It also does not hurt tho right?

anonrig:
@jasnell There are conflicts. Can you resolve them?

src/node/internal/crypto_sign.ts
jasnell: Added a comment to clarify.

anonrig (approved):
Other than the typescript issues and eslint disables, LGTM

jasnell:
@irvinebroque @mikenomitch ... just a heads up, I'm going to start labeling workerd PRs that should be highlighted in the changelog using the `changelog` label.

PR #3604, opened 2025-02-25
Add new AI.toMarkdown() method

G4brym:


src/cloudflare/internal/ai-api.ts
```diff
@@ -221,6 +229,68 @@ export class Ai {
...
+    const input = Array.isArray(files) ? files : [files];
+
+    const processedFiles = [];
+    for (const file of input) {
+      processedFiles.push({
+        name: file.name,
+        mimeType: file.blob.type,
+        data: btoa(
+          String.fromCharCode(...new Uint8Array(await file.blob.arrayBuffer()))
+        ),
```
jasnell, line 251: This feels extremely convoluted. I believe you ought to be able to make use of `internal-node:internal-buffer` here and just use `Buffer.from(await file.blob.arrayBuffer()).toString('base64'))` here.

src/cloudflare/internal/ai-api.ts
```diff
@@ -221,6 +229,68 @@ export class Ai {
...
+          throw new AiInternalError(
+            'Internal Error Converting files into Markdown'
+          );
+        }
+
+        if (Array.isArray(files)) {
+          return data.result;
+        }
+
+        return data.result[0] as ConversionResponse;
```
anonrig, line 284: I recommend removing this "as" and add an assertion for length or [0] to exist rather than forcing compiler

src/cloudflare/internal/ai-api.ts
```diff
@@ -221,6 +229,68 @@ export class Ai {
...
+      case 200: {
+        const data = (await res.json()) as { result: ConversionResponse[] };
+
+        if (!data.result || data.result.length === 0) {
+          throw new AiInternalError(
+            'Internal Error Converting files into Markdown'
+          );
+        }
+
+        if (Array.isArray(files)) {
```
anonrig, line 280: I'm not following. Where is this defined?

src/cloudflare/internal/ai-api.ts
```diff
@@ -221,6 +229,68 @@ export class Ai {
...
+          return data.result;
+        }
+
+        return data.result[0] as ConversionResponse;
+      }
+      default: {
+        const data = (await res.json()) as { errors: { message: string }[] };
+
+        throw new AiInternalError(data.errors[0]?.message || 'Internal Error');
```
anonrig, line 289: Using .at() returns nullable response which is better than index based access which doesn't return nullable value in typescript

```suggestion
        throw new AiInternalError(data.errors.at(0)?.message || 'Internal Error');
```

src/cloudflare/internal/ai-api.ts
G4brym: this import as you are suggesting `import { Buffer } from 'internal-node:internal-buffer';` 
throws this error: `src/cloudflare/internal/ai-api.ts(6,24): error TS2307: Cannot find module 'internal-node:internal-buffer' or its corresponding type declarations.`
I didn't find any other `cloudflare/*` file importing anything from `internal-node` so it might be a simple configuration missing

src/cloudflare/internal/ai-api.ts
G4brym: I didn't understand the question, `files` is the function parameter, and this condition is just checking if the user sent one or multiple files as input

anonrig: Ah I see. Can you add a comment to this line so the intended behavior is documented?

src/cloudflare/internal/ai-api.ts
```diff
@@ -221,6 +229,77 @@ export class Ai {
     }
   }
 
+  public async toMarkdown(
+    files: { name: string; blob: Blob }[]
+  ): Promise<ConversionResponse[]>;
+  public async toMarkdown(files: {
+    name: string;
+    blob: Blob;
+  }): Promise<ConversionResponse>;
```
anonrig, line 238: I don't think these are necessary since the output value doesn't change according to the parameter type.

```suggestion
```

src/cloudflare/internal/ai-api.ts
anonrig: The current implementation limits the length of the string to 65k. Using spread operator means that you are bounded by the V8's argument limitation.

G4brym: i would prefer to use the ìnternal-node` version, but I'm unable to import it

src/cloudflare/internal/ai-api.ts
G4brym: the output is dependent on the input, if the user sends a single file, the function returns an object, if the user sends a list, then an array is return
See `src/cloudflare/internal/test/ai/ai-api-test.js` for example usage

src/cloudflare/internal/ai-api.ts
G4brym: I wasn't able to import the Buffer class, so i changed the implementation to stop using the spread operator, this should bypass the 65k limit

G4brym:
Hey @jasnell or @anonrig can you take a look at this pr again?

src/cloudflare/internal/ai-api.ts
jasnell: hmm.. yeah it looks like there's a bit more we need to do in the runtime to make importing `node-internal:*` stuff available to `cloudflare-internal:*` ... dang. Ok, will have to fix that then we can update this soon.

src/cloudflare/internal/ai-api.ts
```diff
@@ -74,6 +82,22 @@ export class AiInternalError extends Error {
   }
 }
 
+async function blobToBase64(blob: Blob): Promise<string> {
+  const arrayBuffer = await blob.arrayBuffer();
```
jasnell, line 86: Let's add a todo comment here...
```suggestion
  // TODO(soon): This is better implemented using the node::buffer API
  // but we cannot get to that from here currently. Once the node:buffer
  // API (actually, `node-internal:internal_buffer`) is available to be imported
  // here we should update this code to use it instead.
  const arrayBuffer = await blob.arrayBuffer();
```

src/cloudflare/internal/ai-api.ts
```diff
@@ -221,6 +245,82 @@ export class Ai {
...
+      },
+    };
+
+    const endpointUrl =
+      'https://workers-binding.ai/to-everything/markdown/transformer';
+
+    const res = await this.fetcher.fetch(endpointUrl, fetchOptions);
+
+    switch (res.status) {
+      case 200: {
```
jasnell, line 291: A better practice would be to check for `if (res.ok) { ...` which covers the full range of 2xx status codes, not just 200, but this is fine if you know your service would only ever produce 200 codes.

jasnell (approved):

src/cloudflare/internal/ai-api.ts
G4brym: this service only produces 200
I used the switch statement, because other methods in this file produce several codes and use it already

G4brym:
Can someone merge this pr? i don't have permissions

jasnell (approved):

PR #3605, opened 2025-02-25
SQLite: Don't bill for internal queries.

kentonv:
We bill for:
- SQL queries issued via the JavaScript SQL API.
- DO old-style key/value storage operations.

Anything else, we assume is an internal query, and not billed.

### DIFF
--- a/src/workerd/util/sqlite-kv.h
+++ b/src/workerd/util/sqlite-kv.h
@@ -15,6 +15,11 @@ namespace workerd {
 // of KJ exceptions which become internal errors.
 class SqliteKvRegulator: public SqliteDatabase::Regulator {
   void onError(kj::Maybe<int> sqliteErrorCode, kj::StringPtr message) const override;
+
+  // We bill for KV operations as rows read/written.
+  virtual bool shouldAddQueryStats() const override {
+    return true;
+  }
 };
 
 // Class which implements KV storage on top of SQLite. This is intended to be used for Durable
@@ -123,9 +128,7 @@ class SqliteKv: private SqliteDatabase::ResetListener {
       ORDER BY key DESC
       LIMIT ?
     )");
-    // We don't pass in the regulator here because this query doesn't take user input, so
-    // it failing is always our fault.
-    SqliteDatabase::Statement stmtCountKeys = db.prepare(R"(
+    SqliteDatabase::Statement stmtCountKeys = db.prepare(regulator, R"(
       SELECT count(*) FROM _cf_KV
     )");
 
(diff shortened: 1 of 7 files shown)


justin-mp (approved):

shrima-cf:
LGTM

src/workerd/util/sqlite-kv.h
```diff
@@ -15,6 +15,11 @@ namespace workerd {
 // of KJ exceptions which become internal errors.
 class SqliteKvRegulator: public SqliteDatabase::Regulator {
   void onError(kj::Maybe<int> sqliteErrorCode, kj::StringPtr message) const override;
+
+  // We bill for KV operations as rows read/written.
+  virtual bool shouldAddQueryStats() const override {
```
jclee, line 20: (I guess the `stmtCountKeys` prepared statement doesn't use this regulator yet, so would be excluded from stats, but not sure if that query represents notable overhead.)

jclee:
Nice... When I was thinking about how to implement something similar, I was considering adding a "reportingId" value to queries, which could then be passed as an additional parameter to `addQueryStats()`... That might have the advantage of allowing us to separately track both system query overhead (maybe for internal metrics?) and user query overhead.  But not sure if we need that, and if not, this approach seems simpler and more performant.

I guess I also don't know enough about the query stats recording to know if running "ROLLBACK" as a system query -- after running some user queries -- results in significant overhead that we'd want included in user metrics.  But maybe that's not a big deal, as long as the stats of the initial execution get recorded and are roughly proportional to what the rollback metrics would be.

src/workerd/util/sqlite-kv.h
kentonv: Ah good catch, probably should fix that.

kentonv:
> I guess I also don't know enough about the query stats recording to know if running "ROLLBACK" as a system query -- after running some user queries -- results in significant overhead that we'd want included in user metrics.

The overhead of ROLLBACK is pretty minimal, the WAL just gets truncated back to an earlier point. In fact, any writes that end up rolled back never even leave the local machine, so you could argue a rollback should actually roll back the billing counters! But that would be complicated to implement and awkward to explain.

kentonv:
(rebased)

kentonv:
(resolved Joe's comment)

src/workerd/util/sqlite-kv.h
kentonv: Fixed.

There was a comment saying:

```
    // We don't pass in the regulator here because this query doesn't take user input, so
    // it failing is always our fault.
```

I decided this concern doesn't really mater as there is simply no way the query would ever fail, except for deeper I/O errors that could affect any of these queries, so it might as well use the same regulator as everything else.

kentonv:
ok this time the internal build failure looks legit, I guess I will need to look into it.

PR #3609, opened 2025-02-26
Python: Add vendor directory to the path automatically if present

hoodmane:
This won't work yet with top level imports from the vendor directory because of snapshot problems. I am leaving that to a followup.


src/pyodide/internal/python.ts
```diff
@@ -110,6 +110,18 @@ export async function loadPyodide(
...
+      from pathlib import Path
 
+      VENDOR_PATH = "/session/metadata/vendor"
+      print(VENDOR_PATH, Path(VENDOR_PATH).is_dir())
+      if Path(VENDOR_PATH).is_dir():
+        sys.path.append(VENDOR_PATH)
+
+    _tmp()
+    del _tmp
+  `);
```
dom96, line 125: Nit: stick this into a separate function and then call it from loadPyodide (just to keep it cleaner)

src/pyodide/internal/python.ts
```diff
@@ -110,6 +110,18 @@ export async function loadPyodide(
     (msg) => console.log(msg),
     (msg) => console.error(msg)
   );
+  pyodide.runPython(`
+    def _tmp():
+      import sys
+      from pathlib import Path
 
+      VENDOR_PATH = "/session/metadata/vendor"
+      print(VENDOR_PATH, Path(VENDOR_PATH).is_dir())
```
dom96, line 119:
```suggestion
```

dom96 (approved):

PR #3610, opened 2025-02-26
Implement importable env

jasnell:
PR marked as a draft until we're certain this is what we want....

```js
import { env } from 'cloudflare:workers';

console.log(env.FOO);  // Undefined, outside request io context

export default {
  fetch(req) {
    console.log(env.FOO);  // 'BAR', inside request io context
    // ...
  }
}
```

~~Gated by a compat flag that is off by default~~ The flag is now on by default.

  * `allow_importable_env`
  * `disallow_importable_env`

Mutations of the env properties are passed through to the underlying object so will be reflected on the `env` argument passed into the handlers.

Note that this is *not* directly tied to `process.env`... this exposes the regular `env`/`bindings` via the import when accessed within the context of a request. As such, any modules used by your worker can access the bindings without having them explicitly passed down. The `process.env` is populated from the same original source but mutations on `process.env` will not be reflected on the importable `env` and vis versa. At least for the time being.

### DIFF
--- a/src/cloudflare/internal/env.d.ts
+++ b/src/cloudflare/internal/env.d.ts
@@ -1,0 +1,3 @@
+// Get the current environment, if any
+export function getCurrent(): Record<string, unknown> | undefined;
+export function withEnv(newEnv: unknown, fn: () => unknown): unknown;
--- a/src/cloudflare/workers.ts
+++ b/src/cloudflare/workers.ts
@@ -6,9 +6,86 @@
 //   wrapper module that simply re-exports the classes from the built-in module.
 
 import entrypoints from 'cloudflare-internal:workers';
+import innerEnv from 'cloudflare-internal:env';
 
 export const WorkerEntrypoint = entrypoints.WorkerEntrypoint;
 export const DurableObject = entrypoints.DurableObject;
 export const RpcStub = entrypoints.RpcStub;
 export const RpcTarget = entrypoints.RpcTarget;
 export const WorkflowEntrypoint = entrypoints.WorkflowEntrypoint;
+
+export function withEnv(newEnv: unknown, fn: () => unknown): unknown {
+  return innerEnv.withEnv(newEnv, fn);
+}
+
+// A proxy for the workers env/bindings. The proxy is io-context
+// aware in that it will only return values when there is an active
+// IoContext. Mutations to the env via this proxy propagate to the
+// underlying env and do not follow the async context.
+export const env = new Proxy(
+  {},
+  {
+    get(_: unknown, prop: string | symbol): unknown {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.get(inner, prop);
+      }
+      return undefined;
+    },
+
+    set(_: unknown, prop: string | symbol, newValue: unknown): boolean {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.set(inner, prop, newValue);
+      }
+      return true;
+    },
+
+    has(_: unknown, prop: string | symbol): boolean {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.has(inner, prop);
+      }
+      return false;
+    },
+
+    ownKeys(_: unknown): ArrayLike<string | symbol> {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.ownKeys(inner);
+      }
+      return [];
+    },
+
+    deleteProperty(_: unknown, prop: string | symbol): boolean {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.deleteProperty(inner, prop);
+      }
+      return true;
+    },
+
+    defineProperty(
+      _: unknown,
+      prop: string | symbol,
+      attr: PropertyDescriptor
+    ): boolean {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.defineProperty(inner, prop, attr);
+      }
+      return true;
+    },
+
+    getOwnPropertyDescriptor(
+      _: unknown,
+      prop: string | symbol
+    ): PropertyDescriptor | undefined {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.getOwnPropertyDescriptor(inner, prop);
+      }
+      return undefined;
+    },
+  }
+);
--- a/src/workerd/api/modules.h
+++ b/src/workerd/api/modules.h
@@ -10,12 +10,30 @@
 #include <workerd/api/sockets.h>
 #include <workerd/api/unsafe.h>
 #include <workerd/api/worker-rpc.h>
+#include <workerd/io/features.h>
 #include <workerd/io/worker.h>
 #include <workerd/jsg/modules-new.h>
 
 #include <cloudflare/cloudflare.capnp.h>
 
 namespace workerd::api {
+
+class EnvModule final: public jsg::Object {
+ public:
+  EnvModule() = default;
+  EnvModule(jsg::Lock&, const jsg::Url&) {}
+
+  kj::Maybe<jsg::JsObject> getCurrent(jsg::Lock& js);
+
+  // Arranges to propagate the given newEnv in the async context.
+  jsg::JsRef<jsg::JsValue> withEnv(
+      jsg::Lock& js, jsg::Value newEnv, jsg::Function<jsg::JsRef<jsg::JsValue>()> fn);
+
+  JSG_RESOURCE_TYPE(EnvModule) {
+    JSG_METHOD(getCurrent);
+    JSG_METHOD(withEnv);
+  }
+};
 
 template <class Registry>
 void registerModules(Registry& registry, auto featureFlags) {
@@ -30,6 +48,8 @@ void registerModules(Registry& registry, auto featureFlags) {
   registerSocketsModule(registry, featureFlags);
   registry.addBuiltinBundle(CLOUDFLARE_BUNDLE);
   registerRpcModules(registry, featureFlags);
+  registry.template addBuiltinModule<EnvModule>(
+      "cloudflare-internal:env", workerd::jsg::ModuleRegistry::Type::INTERNAL);
 }
 
 template <class TypeWrapper>
@@ -58,6 +78,7 @@ void registerBuiltinModules(jsg::modules::ModuleRegistry::Builder& builder, auto
   {
     jsg::modules::ModuleBundle::BuiltinBuilder builtinsBuilder(
         jsg::modules::ModuleBundle::BuiltinBuilder::Type::BUILTIN);
+    builtinsBuilder.addObject<EnvModule, TypeWrapper>("cloudflare-internal:env"_url);
     jsg::modules::ModuleBundle::getBuiltInBundleFromCapnp(builtinsBuilder, CLOUDFLARE_BUNDLE);
     builder.add(builtinsBuilder.finish());
   }
--- a/src/workerd/api/node/util.c++
+++ b/src/workerd/api/node/util.c++
@@ -242,7 +242,7 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
 }
 
 jsg::JsObject UtilModule::getEnvObject(jsg::Lock& js) {
-  return js.getEnv(true);
+  return js.getProcessEnv(true);
 }
 
 namespace {
--- a/src/workerd/api/tests/disable-importable-env-test.js
+++ b/src/workerd/api/tests/disable-importable-env-test.js
@@ -1,0 +1,25 @@
+import { env, withEnv } from 'cloudflare:workers';
+
+// This test runs with the disallow-importable-env flag set, meaning that
+// the env imported from cloudflare:workers exists but is not-populated
+// with the primary environment. Using withEnv, however, would still work.
+
+import { strictEqual, notDeepStrictEqual } from 'node:assert';
+
+strictEqual(env.FOO, undefined);
+
+export const test = {
+  async test(_, argEnv) {
+    strictEqual(env.FOO, undefined);
+    notDeepStrictEqual(argEnv, env);
+
+    // withEnv still works as expected tho
+    const { env: otherEnv2 } = await withEnv({ BAZ: 1 }, async () => {
+      await scheduler.wait(0);
+      return import('child');
+    });
+    strictEqual(otherEnv2.FOO, undefined);
+    strictEqual(otherEnv2.BAZ, 1);
+    strictEqual(argEnv.BAZ, undefined);
+  },
+};
--- a/src/workerd/api/tests/disable-importable-env-test.wd-test
+++ b/src/workerd/api/tests/disable-importable-env-test.wd-test
@@ -1,0 +1,22 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "disable-importable-env-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "disable-importable-env-test.js"),
+          (name = "child", esModule = "import {env as live} from 'cloudflare:workers'; export const env = {...live};"),
+        ],
+        compatibilityDate = "2025-02-01",
+        compatibilityFlags = [
+          "nodejs_compat_v2",
+          "disallow_importable_env",
+        ],
+        bindings = [
+          (name = "FOO", text = "BAR"),
+        ],
+      )
+    ),
+  ],
+);
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -727,4 +727,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $experimental
       $neededByFl;
   # Enables cache settings specified request in cache api cf object to override cache rules. (only for user owned or grey-clouded sites)
+
+  disableImportableEnv @78 :Bool
+      $compatEnableFlag("disallow_importable_env")
+      $compatDisableFlag("allow_importable_env");
+  # When allowed, `import { env } from 'cloudflare:workers'` will provide access
+  # to the per-request environment/bindings.
 }
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -19,6 +19,7 @@
 #include <workerd/jsg/jsg.h>
 #include <workerd/jsg/modules-new.h>
 #include <workerd/jsg/script.h>
+#include <workerd/jsg/setup.h>
 #include <workerd/jsg/util.h>
 #include <workerd/util/batch-queue.h>
 #include <workerd/util/color-util.h>
@@ -1626,6 +1627,9 @@ Worker::Worker(kj::Own<const Script> scriptParam,
             if (script->isModular()) {
               // Use `env` variable.
               bindingsScope = v8::Object::New(lock.v8Isolate);
+              if (!FeatureFlags::get(js).getDisableImportableEnv()) {
+                lock.setWorkerEnv(lock.v8Ref(bindingsScope.As<v8::Value>()));
+              }
             } else {
               // Use global-scope bindings.
               bindingsScope = context->Global();
--- a/src/workerd/jsg/setup.h
+++ b/src/workerd/jsg/setup.h
@@ -210,6 +210,10 @@ class IsolateBase {
     return externalMemoryAccounter;
   }
 
+  AsyncContextFrame::StorageKey& getEnvAsyncContextKey() {
+    return *envAsyncContextKey;
+  }
+
  private:
   template <typename TypeWrapper>
   friend class Isolate;
@@ -274,11 +278,17 @@ class IsolateBase {
   // Object that is used as the underlying target of process.env when nodejs-compat mode is used.
   v8::Global<v8::Object> envObj;
 
+  // Object used as the underlying storage for a workers environment.
+  v8::Global<v8::Object> workerEnvObj;
+
   // Polyfilled Symbol.asyncDispose.
   v8::Global<v8::Symbol> symbolAsyncDispose;
 
   // Used to account for external memory
   v8::ExternalMemoryAccounter externalMemoryAccounter;
+
+  // A shared async context key for accessing env
+  kj::Own<AsyncContextFrame::StorageKey> envAsyncContextKey;
 
   // We expect queues to remain relatively small -- 8 is the largest size I have observed from local
   // testing.
@@ -683,12 +693,12 @@ class Isolate: public IsolateBase {
 
     // Sets an env value that will be expressed on the process.env
     // if/when nodejs-compat mode is used.
-    void setEnvField(const JsValue& name, const JsValue& value) override {
-      getEnv().set(*this, name, value);
+    void setProcessEnvField(const JsValue& name, const JsValue& value) override {
+      getProcessEnv().set(*this, name, value);
     }
 
     // Returns the env base object.
-    JsObject getEnv(bool release = false) override {
+    JsObject getProcessEnv(bool release = false) override {
       KJ_DEFER({
         if (release) jsgIsolate.envObj.Reset();
       });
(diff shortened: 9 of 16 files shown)


jasnell:
Summary of internal discussion:

1. We want to go ahead and make `env` also populated at the top-level scope (PR updated to do so)
2. It has been suggested that we need a way of propagating a modified/custom context only down specific async contexts (PR updated to do so using `withEnv(...)`). This uses async context propagation under the covers.
3. Test case has been updated to illustrate the updated functionality.

src/workerd/api/modules.h
```diff
@@ -10,13 +10,46 @@
...
+#include <workerd/jsg/setup.h>
 
 #include <cloudflare/cloudflare.capnp.h>
 
 namespace workerd::api {
 
+class EnvModule final: public jsg::Object {
+ public:
+  kj::Maybe<jsg::JsObject> getCurrent(jsg::Lock& js) {
+    if (!FeatureFlags::get(js).getImportableEnv()) {
```
kentonv, line 25: Can we make it so when the flag is disabled, `withEnv` still works? The flag just decides whether the async context is initialized with the real env vs. an empty object.

src/workerd/io/compatibility-date.capnp
```diff
@@ -721,4 +721,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $impliedByAfterDate(name = "nodeJsCompat", date = "2025-04-01");
   # Automatically populate process.env from text bindings only
   # when nodejs_compat is being used.
+
+  importableEnv @77 :Bool
+      $compatEnableFlag("allow_importable_env")
```
kentonv, line 726: Since this isn't a breaking change, perhaps the meanings of enable/disable should be reversed here. The flag would exist _only_ for people like me to block populating of env. But for everyone else, it's exposed by default, even for past compat dates.

src/workerd/api/global-scope.c++
```diff
@@ -230,6 +231,8 @@ kj::Promise<DeferredProxy<void>> ServiceWorkerGlobalScope::request(kj::HttpMetho
   bool useDefaultHandling;
   KJ_IF_SOME(h, exportedHandler) {
     KJ_IF_SOME(f, h.fetch) {
+      jsg::AsyncContextFrame::StorageScope envStorageScope(
```
kentonv, line 234: Hmm, you've instrumented every event entrypoint in global-scope, but I think this doesn't cover custom events, including RPC?

Instrumenting every entrypoint feels error-prone. Can we make the initial assignment just apply to the _whole worker_, until such a point as `withEnv()` is used to set a different value?

jasnell: That's a bit difficult given where we initialize the bindingScope in worker.c++. I did try to minimize it but wasn't able to get it working. Before this is moved from draft, however, I plan to get that part figured out so we can minimize the instrumentation.

jasnell: Yep, considered that. Once we're sure of the overall behavior we want I'll go back and clean this bit up.

src/workerd/api/modules.h
jasnell: Done.

src/workerd/io/compatibility-date.capnp
jasnell: Done.

jasnell:
Next round of updates:

1. Reversed the polarity of the compat flag. The importable `env` is now populated/enabled by default with a compat flag to switch it off
2. `withEnv(...)` will work even if the disable flag is turned on.

Next known remaining todos:

1. Simplify the instrumentation to avoid having to instrument multiple times. https://github.com/cloudflare/workerd/pull/3610#discussion_r1972631045

src/workerd/api/global-scope.c++
jasnell: Btw, given just a jsg::Lock and no IoContext, what is the idiomatic way of getting to the `Worker::Impl::env`?

kentonv: Uhhhhhhh.... good question. I guess I'm not sure there is one. Might need to add something to JSG.

jasnell: Ok, I've got something in mind but it'll take a bit to work up. Will do that tomorrow.

jasnell: Ok, updated the PR to store the env on jsg::Lock rather than in Worker::Impl and it works just fine. Was able to remove all of the instrumentation points and have the lookup fallback as needed.

src/workerd/api/global-scope.c++
```diff
@@ -20,6 +20,7 @@
 #include <workerd/io/io-context.h>
 #include <workerd/jsg/async-context.h>
 #include <workerd/jsg/ser.h>
+#include <workerd/jsg/setup.h>
```
kentonv, line 23: no longer needed?

src/workerd/api/modules.h
```diff
@@ -10,13 +10,51 @@
...
 #include <workerd/jsg/modules-new.h>
+#include <workerd/jsg/setup.h>
 
 #include <cloudflare/cloudflare.capnp.h>
 
 namespace workerd::api {
 
+class EnvModule final: public jsg::Object {
+ public:
+  kj::Maybe<jsg::JsObject> getCurrent(jsg::Lock& js) {
```
kentonv, line 24: Let's define these methods out-of-line, otherwise we're piling onto the compile time for workerd-api.c++.

src/workerd/io/worker.c++
```diff
@@ -1626,6 +1628,7 @@ Worker::Worker(kj::Own<const Script> scriptParam,
             if (script->isModular()) {
               // Use `env` variable.
               bindingsScope = v8::Object::New(lock.v8Isolate);
+              lock.setWorkerEnv(lock.v8Ref(bindingsScope.As<v8::Value>()));
```
kentonv, line 1631: Could we check the compat flag here, and not even call `setWorkerEnv()` (or call it with an empty object) if the flag is set? Rather than checking it on every access.

It doesn't make a difference currently, but if we start adding other ways to get an env ambiently, the compat flag applies to all of them, which is what I'd want as a user of this flag.

jasnell: If you look further down, you'll see that I've changed from storing `env` in `Worker::Impl` and instead store it in the lock here, so this sets it for the existing uses also, replacing the `impl->env = ` line was down below.

jasnell: If we end up adding other ways to get to the env, then we can move the code currently in `EnvModule::getCurrent()` into a more central location since that does everything that is needed

kentonv (approved):

src/workerd/io/worker.c++
kentonv: Wait, I think this breaks dynamic bindings: There can be multiple `Worker` objects in the same isolate, each with a different `env`, if they have bindings that change from request to request. Dynamic bindings were only ever exposed internally and we want to get rid of them, but a lot of internal workers are still relying on them today, so if you start storing the `env` on the isolate, concurrent requests will stomp each other.

kentonv (changes requested):

src/workerd/io/worker.c++
kentonv: Note that we don't need imported env to support dynamic bindings. For people using dynamic bindings, we can just say: don't use imported env. So I think imported env can just reflect the _first_ env ever received. But the env delivered to event handlers as a parameter still needs to be the per-request env, which we still need to store in `Worker::Impl`.

kentonv:
See my last comment above:

> Note that we don't need imported env to support dynamic bindings. For people using dynamic bindings, we can just say: don't use imported env. So I think imported env can just reflect the first env ever received. But the env delivered to event handlers as a parameter still needs to be the per-request env, which we still need to store in Worker::Impl.

src/workerd/io/worker.c++
jasnell: hmm... I think there's a likelihood of weird bugs with that approach... let me play with it a bit to see what approaches may work.

src/cloudflare/workers.ts
```diff
@@ -6,9 +6,92 @@
...
+import innerEnv from 'cloudflare-internal:env';
 
 export const WorkerEntrypoint = entrypoints.WorkerEntrypoint;
 export const DurableObject = entrypoints.DurableObject;
 export const RpcStub = entrypoints.RpcStub;
 export const RpcTarget = entrypoints.RpcTarget;
 export const WorkflowEntrypoint = entrypoints.WorkflowEntrypoint;
+
+// We turn off these specific linting rules here intentionally as mixing
+// typescript and proxies can just be... awkward. The use of any in these
```
vicb, line 18: Outdated comment?

src/workerd/api/tests/disable-importable-env-test.wd-test
```diff
@@ -0,0 +1,22 @@
...
+  services = [
+    ( name = "disable-importable-env-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "disable-importable-env-test.js"),
+          (name = "child", esModule = "import {env as live} from 'cloudflare:workers'; export const env = {...live};"),
+        ],
+        compatibilityDate = "2025-02-01",
+        compatibilityFlags = [
+          "nodejs_compat_v2",
```
vicb, line 13: Does this need nodejs compat?

vicb: Oh... Assert

vicb:
The PR description mentions that the flag is off by default. I think this is not the current state. 

Would be nice to read about security considerations in the description too. IIUC this could be used by any library to get access to the bindings.

vicb:
One last thing: is "(dis)allow_importable_env" really reflecting what this does? 

IIUC the PR does not (dis)allow anything. Env is always importable - sometimes not populated. 

Would be nice to have a better description on the PR description and maybe a name reflecting the actual behavior. 

Thanks

jasnell:
Happy to accept suggestions for alternative names for the compat flags. I feel they are descriptive of what they are doing but opinions may vary

vicb:
"Note that this is not related to process.env..."

It actually is somehow related now that process.env is populated from the env.
IMO it's important to document that mutating an importable env will not affect process.env (and probably vice-versa while we're are at it)

src/workerd/api/tests/disable-importable-env-test.js
```diff
@@ -0,0 +1,24 @@
+import { env, withEnv } from 'cloudflare:workers';
+
+// This test runs with the disallow-importable-env flag set, meaning that
+// the env imported from cloudflare:workers exists but is not-populated
+// with the primary environment. Using withEnv, however, would still work.
+
+import { strictEqual, notDeepStrictEqual } from 'node:assert';
+
+strictEqual(env.FOO, undefined);
```
vicb, line 9: maybe add a test setting `env.BAR` not reflected on `argEnv.BAR`

src/workerd/io/compatibility-date.capnp
```diff
@@ -721,10 +721,16 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
       $compatEnableFlag("cache_api_request_cf_overrides_cache_rules")
       $experimental
       $neededByFl;
   # Enables cache settings specified request in cache api cf object to override cache rules. (only for user owned or grey-clouded sites)
+
+  disableImportableEnv @78 :Bool
+      $compatEnableFlag("disallow_importable_env")
+      $compatDisableFlag("allow_importable_env");
+  # When enabled, `import { env } from 'cloudflare:workers'` will provide access
+  # the per-request environment/bindings.
```
vicb, line 735:
```suggestion
  # When enabled, `import { env } from 'cloudflare:workers'` will provide access
  # to the per-request environment/bindings.
```

src/cloudflare/workers.ts
```diff
@@ -6,9 +6,92 @@
...
+      }
+      return [];
+    },
+
+    deleteProperty(_: any, prop: string | symbol): boolean {
+      const inner = innerEnv.getCurrent();
+      if (inner) {
+        return Reflect.deleteProperty(inner, prop);
+      }
+      return true;
```
vicb, line 71: Should this return `false` here ?

src/workerd/api/tests/disable-importable-env-test.js
```diff
@@ -0,0 +1,24 @@
...
+    strictEqual(env.FOO, undefined);
+    notDeepStrictEqual(argEnv, env);
+
+    // withEnv still works as expected tho
+    const { env: otherEnv2 } = await withEnv({ BAZ: 1 }, async () => {
+      await scheduler.wait(0);
+      return import('child');
+    });
+    strictEqual(otherEnv2.FOO, undefined);
+    strictEqual(otherEnv2.BAZ, 1);
```
jasnell, line 22:
```suggestion
    strictEqual(otherEnv2.BAZ, 1);
    strictEqual(argEnv.BAZ, undefined);
```

src/cloudflare/workers.ts
jasnell, line 71:
```suggestion
      return false;
```

jasnell: Wait... actually no. `true` is correct here. The reason is that this is effectively deleting a non-existent property. If, for instance, you have `const n = {}; delete n.abc`, that delete will return `true` because effectively the property *is* deleted.

vicb: Ok, I was looking at a MDN [snippet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#a_complete_traps_list_example):

```ts
const docCookies = new Proxy(docCookies, {
// ...
  deleteProperty(target, key) {
    if (!(key in target)) {
      return false;
    }
    return target.removeItem(key);
  },
```

jasnell: There's some disagreement in the ecosystem about exactly what should be returned in which cases but I tend to err on duplicating what I see v8 doing which is returning `true` when deleting a non-existent property.

vicb: It looks like returning `false` might throw in strict mode, which we do not want.

Maybe a simpler impl for all method could be:

```ts
// before
      const inner = innerEnv.getCurrent();
      if (inner) {
        return Reflect.deleteProperty(inner, prop);
      }
      return true;
      
// after
return Reflect.deleteProperty( innerEnv.getCurrent() ?? {}, prop);
```

jasnell: I think I prefer the more verbose version as I think it's a bit clearer / easier to read. It will also be a bit faster as it avoids the extraneous property lookup on the empty object.

src/cloudflare/workers.ts
```diff
@@ -6,9 +6,92 @@
...
 export const DurableObject = entrypoints.DurableObject;
 export const RpcStub = entrypoints.RpcStub;
 export const RpcTarget = entrypoints.RpcTarget;
 export const WorkflowEntrypoint = entrypoints.WorkflowEntrypoint;
+
+// We turn off these specific linting rules here intentionally as mixing
+// typescript and proxies can just be... awkward. The use of any in these
+// is intentional.
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
anonrig, line 21:
```suggestion
```

Please apply this patch

```diff
diff --git a/src/cloudflare/internal/env.d.ts b/src/cloudflare/internal/env.d.ts
index 1ad1cd02..e30cf92b 100644
--- a/src/cloudflare/internal/env.d.ts
+++ b/src/cloudflare/internal/env.d.ts
@@ -1,3 +1,3 @@
 // Get the current environment, if any
-export function getCurrent(): object | undefined;
+export function getCurrent(): Record<string, unknown> | undefined;
 export function withEnv(newEnv: unknown, fn: () => unknown): unknown;
diff --git a/src/cloudflare/workers.ts b/src/cloudflare/workers.ts
index f71bc591..5ab958e4 100644
--- a/src/cloudflare/workers.ts
+++ b/src/cloudflare/workers.ts
@@ -14,12 +14,6 @@ export const RpcStub = entrypoints.RpcStub;
 export const RpcTarget = entrypoints.RpcTarget;
 export const WorkflowEntrypoint = entrypoints.WorkflowEntrypoint;
 
-// We turn off these specific linting rules here intentionally as mixing
-// typescript and proxies can just be... awkward. The use of any in these
-// is intentional.
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
 export function withEnv(newEnv: unknown, fn: () => unknown): unknown {
   return innerEnv.withEnv(newEnv, fn);
 }
@@ -28,10 +22,10 @@ export function withEnv(newEnv: unknown, fn: () => unknown): unknown {
 // aware in that it will only return values when there is an active
 // IoContext. Mutations to the env via this proxy propagate to the
 // underlying env and do not follow the async context.
-export const env: object = new Proxy(
+export const env = new Proxy(
   {},
   {
-    get(_: any, prop: string | symbol): unknown {
+    get(_: unknown, prop: string | symbol): unknown {
       const inner = innerEnv.getCurrent();
       if (inner) {
         return Reflect.get(inner, prop);
@@ -39,7 +33,7 @@ export const env: object = new Proxy(
       return undefined;
     },
 
-    set(_: any, prop: string | symbol, newValue: any): boolean {
+    set(_: unknown, prop: string | symbol, newValue: unknown): boolean {
       const inner = innerEnv.getCurrent();
       if (inner) {
         return Reflect.set(inner, prop, newValue);
@@ -47,7 +41,7 @@ export const env: object = new Proxy(
       return true;
     },
 
-    has(_: any, prop: string | symbol): boolean {
+    has(_: unknown, prop: string | symbol): boolean {
       const inner = innerEnv.getCurrent();
       if (inner) {
         return Reflect.has(inner, prop);
@@ -55,7 +49,7 @@ export const env: object = new Proxy(
       return false;
     },
 
-    ownKeys(_: any): any {
+    ownKeys(_: unknown): ArrayLike<string | symbol> {
       const inner = innerEnv.getCurrent();
       if (inner) {
         return Reflect.ownKeys(inner);
@@ -63,7 +57,7 @@ export const env: object = new Proxy(
       return [];
     },
 
-    deleteProperty(_: any, prop: string | symbol): boolean {
+    deleteProperty(_: unknown, prop: string | symbol): boolean {
       const inner = innerEnv.getCurrent();
       if (inner) {
         return Reflect.deleteProperty(inner, prop);
@@ -72,7 +66,7 @@ export const env: object = new Proxy(
     },
 
     defineProperty(
-      _: any,
+      _: unknown,
       prop: string | symbol,
       attr: PropertyDescriptor
     ): boolean {
@@ -84,7 +78,7 @@ export const env: object = new Proxy(
     },
 
     getOwnPropertyDescriptor(
-      _: any,
+      _: unknown,
       prop: string | symbol
     ): PropertyDescriptor | undefined {
       const inner = innerEnv.getCurrent();

```

src/workerd/api/modules.h
```diff
@@ -61,6 +76,8 @@ void registerBuiltinModules(jsg::modules::ModuleRegistry::Builder& builder, auto
     jsg::modules::ModuleBundle::getBuiltInBundleFromCapnp(builtinsBuilder, CLOUDFLARE_BUNDLE);
     builder.add(builtinsBuilder.finish());
   }
+
+  // TODO(later): Add the internal env module also.
```
anonrig, line 80: Can you be more descriptive with this comment? I'm not following

src/workerd/api/modules.h
```diff
@@ -10,13 +10,26 @@
...
 
 #include <cloudflare/cloudflare.capnp.h>
 
 namespace workerd::api {
 
+class EnvModule final: public jsg::Object {
+ public:
+  kj::Maybe<jsg::JsObject> getCurrent(jsg::Lock& js);
+
+  jsg::JsValue withEnv(jsg::Lock& js, jsg::Value newEnv, jsg::Function<jsg::JsValue()> fn);
```
anonrig, line 25: Can you add a comment to both of these functions and explain what they do? It's not really visible the difference between getCurrent and withEnv by looking at the function names

src/workerd/server/workerd-api.c++
```diff
@@ -705,9 +706,9 @@ static v8::Local<v8::Value> createBindingValue(JsgWorkerdIsolate::Lock& lock,
         // parsed in one context (env) and unparsed in another (process.env).
 
         if (value->IsString()) {
-          lock.setEnvField(lock.str(global.name), jsg::JsValue(value));
+          lock.setProcessEnvField(lock.str(global.name), jsg::JsValue(value));
```
anonrig, line 709: This is getting confusing. process comes from node.js process, but it's extremely hard for a non-nodejs contributor to understand what the "process" prefix is coming from. I recommend picking a more verbose function name like: "setNodeProcessEnvField"

src/workerd/api/node/util.c++
```diff
@@ -242,7 +242,7 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
 }
 
 jsg::JsObject UtilModule::getEnvObject(jsg::Lock& js) {
-  return js.getEnv(true);
+  return js.getProcessEnv(true);
```
anonrig, line 245: This function name should also change. We should select a more verbose function name to indicate the context: node.js

src/cloudflare/internal/env.d.ts
```diff
@@ -0,0 +1,3 @@
+// Get the current environment, if any
+export function getCurrent(): object | undefined;
+export function withEnv(newEnv: unknown, fn: () => unknown): unknown;
```
anonrig, line 3: Can you add a comment to this function as well, just as you did for getCurrent()

src/cloudflare/workers.ts
jasnell: Can you use suggestions instead please. A lot easier that way.

src/workerd/server/workerd-api.c++
jasnell: eh, I'd rather not make the name longer. I think `ProcessEnv` is fairly descriptive that it applies to `process.env`

src/workerd/api/node/util.c++
jasnell: Ditto. I'd prefer not to rename further.

src/workerd/api/modules.h
jasnell, line 80:
```suggestion
  // TODO(new-module-registry): Add the internal env module also. This relates to the new module
  // registry implementation.
```

src/workerd/api/modules.h
jasnell, line 25:
```suggestion
  // Arranges to propagate the given newEnv in the async context.
  jsg::JsValue withEnv(jsg::Lock& js, jsg::Value newEnv, jsg::Function<jsg::JsValue()> fn);
```

src/cloudflare/internal/env.d.ts
jasnell, line 3:
```suggestion
// Arrange to have the given newEnv propagated along the async context.
export function withEnv(newEnv: unknown, fn: () => unknown): unknown;
```

src/cloudflare/workers.ts
jasnell: Done

anonrig (approved):

src/workerd/jsg/setup.h
```diff
@@ -683,12 +693,12 @@ class Isolate: public IsolateBase {
...
     // if/when nodejs-compat mode is used.
-    void setEnvField(const JsValue& name, const JsValue& value) override {
-      getEnv().set(*this, name, value);
+    void setProcessEnvField(const JsValue& name, const JsValue& value) override {
+      getProcessEnv().set(*this, name, value);
     }
 
     // Returns the env base object.
-    JsObject getEnv(bool release = false) override {
+    JsObject getProcessEnv(bool release = false) override {
```
kentonv, line 701: Could we actually change the `process.env` implementation to be based on the new `getWorkerEnv()`, rather than hooking into `workerd-api`? It seems like this should be functionally identical, except if `disallow_importable_env`, but I think that's a feature, not a bug: I would like this flag to prevent my env from being globally available by any means. I'd rather not have to set multiple flags.

I'm OK with that being a separate PR but it would need to be done soon.

jasnell:
There is a segmentation fault in the internal repo when running the new tests here. Will need to investigate before this lands

Issue #3410, opened 2025-01-27
Ai_Cf_Openai_Whisper_Large_V3_Turbo_Output["segments"] should be typed as an array to match the response


danamajid:
The actual returned segments is an array on [Ai_Cf_Openai_Whisper_Large_V3_Turbo_Output]( https://github.com/cloudflare/workerd/blob/fa6bf311407b36305595ec726baadc130db61c8f/types/defines/ai.d.ts#L329-L372)'s `segments` type.

fhanau:
Fixed in https://github.com/cloudflare/workerd/pull/3652

PR #3623, opened 2025-02-27
fix(types/defines/ai.d.ts): correct the type of the `segments` property to be an array

Lqm1:
Correct the incorrect type of Output in “@cf/openai/whisper-large-v3-turbo”.
I think segments is array.

### DIFF
--- a/types/defines/ai.d.ts
+++ b/types/defines/ai.d.ts
@@ -430,7 +430,7 @@ export interface Ai_Cf_Openai_Whisper_Large_V3_Turbo_Output {
        */
       end?: number;
     }[];
-  };
+  }[];
   /**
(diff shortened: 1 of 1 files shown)


fhanau:
@thatsKevinJain

Lqm1:

Lqm1:
After checking the issue, we found a issue, so we will link it for management purposes.

https://github.com/cloudflare/workerd/issues/3410

fhanau:
@thatsKevinJain can you review this change? Looks like CI is passing, except that generated type definitions need to be updated.

thatsKevinJain:
This is correct, segments should be an array. I will add a commit with updated generated-snapshots files as well.

fhanau:
Covered by #3652, closing this.

PR #3628, opened 2025-02-28
feat: add new option for bulk gets in KV

teresalves:
Creating a new part in the binding for new bulk gets, along with tests for regular gets and the bulk gets
We consider we have a POST endpoint in SGW and we send the keys in a json format as such:
```
{
   "keys": ["key1", "key2"]
}
```

However, we decided that the binding here will only be a passthrough and everything will processed in the endpoint. On top of that, since everything is processed in the binding and comes as a single response, we don't need extra logic to process metadata, we just need to make sure that the binding responds what we what. Because of these two things, we added two more parameters in the request body:
```
{
   "keys": ["key1", "key2"],
   "type": "text",
   "withMetadata": true
}
```

We added tests for:
Regular gets:
- 400s
- 500s
- 200s with text, json, stream and arrayBuffer

Bulk Gets:
- 500s
- 200s with text and json, as these will be the only supported types for now.

Note that we do not test for 400s because we will not have them. If none of the keys exist, we return a 200 but with empty values. Example:
```
{
   "key1": null,
   "key2": null,
 }
 ```

### DIFF
--- a/src/workerd/api/kv-test.js
+++ b/src/workerd/api/kv-test.js
@@ -6,16 +6,16 @@ import assert from 'node:assert';
 export default {
   // Producer receiver (from `env.NAMESPACE`)
   async fetch(request, env, ctx) {
-    let result = "example";
+    let result = 'example';
     const { pathname } = new URL(request.url);
     if (pathname === '/fail-client') {
-      return new Response(null, {status: 404})
-    } else if (pathname == "/fail-server") {
-      return new Response(null, {status: 500})
-    } else if (pathname == "/get-json") {
-      result = JSON.stringify({ example: "values" });
-    } else if (pathname == "/bulk/get") {
-      let r = "";
+      return new Response(null, { status: 404 });
+    } else if (pathname == '/fail-server') {
+      return new Response(null, { status: 500 });
+    } else if (pathname == '/get-json') {
+      result = JSON.stringify({ example: 'values' });
+    } else if (pathname == '/bulk/get') {
+      let r = '';
       const decoder = new TextDecoder();
       for await (const chunk of request.body) {
         r += decoder.decode(chunk, { stream: true });
@@ -24,52 +24,55 @@ export default {
       const parsedBody = JSON.parse(r);
       const keys = parsedBody.keys;
       if (keys.length < 1 || keys.length > 100) {
-        return new Response(null, {status: 400})
+        return new Response(null, { status: 400 });
       }
-      result = {}
-      if(parsedBody.type == "json") {
-        for(const key of keys) {
-          if(key == "key-not-json") {
-            return new Response(null, {status: 500})
+      result = {};
+      if (parsedBody.type == 'json') {
+        for (const key of keys) {
+          if (key == 'key-not-json') {
+            return new Response(null, { status: 500 });
           }
-          const val = { example: `values-${key}`};
+          const val = { example: `values-${key}` };
           if (parsedBody.withMetadata) {
-            result[key] = {value: val, metadata: "example-metadata"};
+            result[key] = { value: val, metadata: 'example-metadata' };
           } else {
             result[key] = val;
           }
         }
-      } else if (!parsedBody.type || parsedBody.type == "text") {
-        for(const key of keys) {
-          const val = JSON.stringify({ example: `values-${key}` });;
-          if(key == "not-found") {
+      } else if (!parsedBody.type || parsedBody.type == 'text') {
+        for (const key of keys) {
+          const val = JSON.stringify({ example: `values-${key}` });
+          if (key == 'not-found') {
             result[key] = null;
           } else if (parsedBody.withMetadata) {
-            result[key] = {value: val, metadata: "example-metadata"};
+            result[key] = { value: val, metadata: 'example-metadata' };
           } else {
             result[key] = val;
           }
         }
-      } else { // invalid type requested
-        return new Response(null,{status: 500});
-
+      } else {
+        // invalid type requested
+        return new Response(null, { status: 500 });
       }
       result = JSON.stringify(result);
-    } else { // generic success for get key
-      result = "value-"+pathname.slice(1);
+    } else {
+      // generic success for get key
+      result = 'value-' + pathname.slice(1);
     }
 
-    let response =  new Response(result, {status: 200});
-    response.headers.set("CF-KV-Metadata", '{"someMetadataKey":"someMetadataValue","someUnicodeMeta":"🤓"}');
+    let response = new Response(result, { status: 200 });
+    response.headers.set(
+      'CF-KV-Metadata',
+      '{"someMetadataKey":"someMetadataValue","someUnicodeMeta":"🤓"}'
+    );
 
     return response;
   },
 
-
   async test(ctrl, env, ctx) {
     // Test .get()
-    let response = await env.KV.get('success',{});
-    assert.strictEqual(response, "value-success");
+    let response = await env.KV.get('success', {});
+    assert.strictEqual(response, 'value-success');
 
     response = await env.KV.get('fail-client');
     assert.strictEqual(response, null);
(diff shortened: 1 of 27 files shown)

### CODE REVIEW

**Comment by teresalves** [comment]

**Comment by teresalves** [comment]
recheck

File: src/workerd/api/kv-test.js [hunk 1]
```diff
@@ -0,0 +1,170 @@
...
+      // assert(message.includes("invalid")) // this message is not processed, should it?
+      assert.ok(true);
+    }
+    try{
+      var response = await env.KV.get(["key-not-json", "key2"], {type: "banana"});
+      assert.ok(false); // not reached
+    } catch ({ name, message }){
+      // assert(message.includes("invalid")) // this message is not processed, should it?
+      assert.ok(true);
+    }
```
sesteves, line 145: what about adding the following test cases:
1. a key is not found
2. the array of keys is empty (or some of the key names are empty strings)
3. fetching the maximum number of keys that we will allow

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,170 @@
...
+    try{
+      var response = await env.KV.get(["key-not-json", "key2"], "json");
+      assert.ok(false); // not reached
+    } catch ({ name, message }){
+      assert(message.includes("500"))
+      assert.ok(true);
+    }
+    // requested type is invalid for bulk get
+    try{
+      var response = await env.KV.get(["key-not-json", "key2"], "arrayBuffer");
```
sesteves, line 133: Does this even get compiled?

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,170 @@
...
+      result = ""
+    } else if (pathname == "/get-json") {
+      result = JSON.stringify({ example: "values" });
+    } else if (pathname == "/get-bulk") {
+      var body = request.body;
+      const reader = body.getReader();
+      const decoder = new TextDecoder(); // UTF-8 by default
+      let r = "";
+      while (true) {
+          const { done, value } = await reader.read();
```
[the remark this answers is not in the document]
teresalves: Tried some other ways and it was not reading the stream properly. Both here and in the type "stream" test below.

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,170 @@
...
+        result += decoder.decode(value, { stream: true });
+    }
+    assert.strictEqual(result, "value-success");
+
+    var response = await env.KV.get('success', "arrayBuffer");
+    assert.strictEqual(new TextDecoder().decode(response), "value-success");
+
+
+    // Testing .get bulk
+    var response = await env.KV.get(["key1", "key2"],{});
```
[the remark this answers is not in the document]
teresalves: We do not, I can add a test without them. But at some point, it was failing when we had options but they were empty, hence the test.

src/workerd/api/kv-test.js
teresalves: 1. added
2. added but I am returning an empty object and not throwing an error in that case - confirming that this is the intended behaviour and getting back to you
3. have not added that one since that will be validated in the endpoint and not here. Should it be in both? I get it so, but wondering if it might be redundant

src/workerd/api/kv-test.js
teresalves: I am sending the type that is invalid here. But the format itself is acceptable, so it compiles but throws an error.

src/workerd/api/kv-test.js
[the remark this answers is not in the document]
teresalves: added another test that does have that second parameter

src/workerd/api/kv-test.js
teresalves: Changed 2 to throw a 400

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+  async fetch(request, env, ctx) {
+    const options = {
+      status: 200,
+      statusText: "Success!",
+      headers: new Headers({
+        'Content-Type': 'application/json'
+      })
+    };
+
+    var result = "example";
```
jasnell, line 17:
```suggestion
    let result = "example";
```

src/workerd/api/kv-test.js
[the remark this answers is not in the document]
jasnell: Can you expand on what "not reading the stream properly" means? ;-)

Also... you can do the following to simplify this a bit:

```js
for await (const chunk of request.body) {
  r += decoder.decode(value, { stream: true });
}
// But also remember to grab the final buffered chunk if any.
r += decoder.decode();
```

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+      const parsedBody = JSON.parse(r);
+      const keys = parsedBody.keys;
+      if (keys.length < 1) {
+        return new Response("Array needs at least one key", {status: 400})
+      }
+      result = {}
+      if(parsedBody.type == "json") {
+        for(const key of keys) {
+          if(key == "key-not-json") {
+            return new Response(null, {status: 500})
```
jasnell, line 46: fwiw, We have `Response.error()` now that would end up sending a 500 response back to the eyeball.

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+    }
+    let response =  new Response(result, {status: 200});
+    response.headers.set("CF-KV-Metadata", '{"someMetadataKey":"someMetadataValue","someUnicodeMeta":"🤓"}');
+    return response;
+  },
+
+
+  async test(ctrl, env, ctx) {
+    // Test .get()
+    var response = await env.KV.get('success',{});
```
jasnell, line 82: generally... just avoid using `var` 
```suggestion
    let response = await env.KV.get('success',{});
```

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+    assert.strictEqual(response, "value-success");
+
+    response = await env.KV.get('fail-client');
+    assert.strictEqual(response, "");
+    try {
+      response = await env.KV.get('fail-server');
+      assert.ok(false);
+    } catch {
+      assert.ok(true);
```
jasnell, line 91: If the expectation is that the env.KV.get promise will reject, use

```
await assert.rejects(env.KV.get('fail-server'), {
  message: '... whatever the expected error message is',
});
```

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+
+    var response = await env.KV.get('success', "stream");
+    const reader = response.getReader();
+    const decoder = new TextDecoder(); // UTF-8 by default
+    let result = "";
+    while (true) {
+        const { done, value } = await reader.read();
+        if (done) break;
+        result += decoder.decode(value, { stream: true });
+    }
```
jasnell, line 109:
```suggestion
    let result = "";
    for await (const chunk of response) {
      result += decoder.decode(value, { stream: true });
    }
    result += decoder.decode();
```

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+    assert.strictEqual(result, "value-success");
+
+    var response = await env.KV.get('success', "arrayBuffer");
```
jasnell, line 112: `let response...`

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+    assert.strictEqual(new TextDecoder().decode(response), "value-success");
+
+
+    // Testing .get bulk
+    var response = await env.KV.get(["key1", "key2"]);
+    var expected = { key1: '{\"example\":\"values-key1\"}', key2: '{\"example\":\"values-key2\"}' };
+    assert.deepEqual(response, expected);
```
jasnell, line 119:
```suggestion
    assert.deepStrictEqual(response, expected);
```

src/workerd/api/kv-test.js
```diff
@@ -0,0 +1,186 @@
...
+
+    var response = await env.KV.get(["key1", "key2"],{});
+    var expected = { key1: '{\"example\":\"values-key1\"}', key2: '{\"example\":\"values-key2\"}' };
+    assert.deepEqual(response, expected);
```
jasnell, line 123:
```suggestion
    assert.deepStrictEqual(response, expected);
```

jasnell: here and elsewhere....

src/workerd/api/kv-test.js
jasnell: Use `await assert.rejects(...)` here instead of the try/catch

src/workerd/api/kv.c++
```diff
@@ -165,9 +167,131 @@ jsg::Promise<KvNamespace::GetResult> KvNamespace::get(
...
+      return getSingle(js, kj::mv(str), kj::mv(options));
+    }
+  }
+  KJ_UNREACHABLE;
+};
+
+jsg::Promise<KvNamespace::GetWithMetadataResult> KvNamespace::getWithMetadataSingle(
+  jsg::Lock& js, kj::String name, jsg::Optional<kj::OneOf<kj::String, GetOptions>> options) {
+return getWithMetadataImpl(js, kj::mv(name), kj::mv(options), LimitEnforcer::KvOpType::GET_WITH);
+}
```
jasnell, line 282: Be sure to check the linting. Looks like the indentation on a few of these is out of whack.

src/workerd/api/kv-test.js
teresalves: For 3, added the scenario where we have 100 keys and 101 keys, throwing an error on that last one.

src/workerd/api/kv.c++
teresalves: That line in particular was already here, but I agree that it seems generally funky. Tried to adjust it a bit. Is there a linter I can run on top of this?
Thank you!

src/workerd/api/kv-test.js
teresalves: This is so much cleaner. Thanks a bunch!
And thanks for all the suggestions :)

src/workerd/api/kv-test.js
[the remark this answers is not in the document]
teresalves: I did not try this one in particular. Added it, thank you!

src/workerd/api/kv.h
```diff
@@ -144,6 +157,12 @@ class KvNamespace: public jsg::Object {
       get(key: Key, options?: KVNamespaceGetOptions<"arrayBuffer">): Promise<ArrayBuffer | null>;
       get(key: Key, options?: KVNamespaceGetOptions<"stream">): Promise<ReadableStream | null>;
 
+      get<ExpectedValue = unknown>(key: Array<Key>, type: "text"): Promise<ExpectedValue | null>;
+      get<ExpectedValue = unknown>(key: Array<Key>, type: "json"): Promise<ExpectedValue | null>;
+      get<ExpectedValue = unknown>(key: Array<Key>, options?: Partial<KVNamespaceGetOptions<undefined>>): Promise<ExpectedValue | null>;
+      get<ExpectedValue = unknown>(key: Array<Key>, options?: KVNamespaceGetOptions<"text">): Promise<ExpectedValue | null>;
+      get<ExpectedValue = unknown>(key: Array<Key>, options?: KVNamespaceGetOptions<"json">): Promise<ExpectedValue | null>;
```
radev, line 164:
```suggestion
      get(key: Array<Key>, type: "text"): Promise<Map<string, string | null>>;
      get<ExpectedValue = unknown>(key: Array<Key>, type: "json"): Promise<Map<string, ExpectedValue | null>>;
      get(key: Array<Key>, options?: Partial<KVNamespaceGetOptions<undefined>>): Promise<Map<string, string | null>>;
      get(key: Array<Key>, options?: KVNamespaceGetOptions<"text">): Promise<Map<string, string | null>>;
      get<ExpectedValue = unknown>(key: Array<Key>, options?: KVNamespaceGetOptions<"json">): Promise<Map<string, ExpectedValue | null>>;
```

src/workerd/api/kv.h
```diff
@@ -158,6 +177,12 @@ class KvNamespace: public jsg::Object {
       getWithMetadata<Metadata = unknown>(key: Key, options: KVNamespaceGetOptions<"arrayBuffer">): Promise<KVNamespaceGetWithMetadataResult<ArrayBuffer, Metadata>>;
       getWithMetadata<Metadata = unknown>(key: Key, options: KVNamespaceGetOptions<"stream">): Promise<KVNamespaceGetWithMetadataResult<ReadableStream, Metadata>>;
 
+      getWithMetadata<ExpectedValue = unknown>(key: Array<Key>, type: "text"): Promise<ExpectedValue | null>;
+      getWithMetadata<ExpectedValue = unknown>(key: Array<Key>, type: "json"): Promise<ExpectedValue | null>;
+      getWithMetadata<ExpectedValue = unknown>(key: Array<Key>, options?: Partial<KVNamespaceGetOptions<undefined>>): Promise<ExpectedValue | null>;
+      getWithMetadata<ExpectedValue = unknown>(key: Array<Key>, options?: KVNamespaceGetOptions<"text">): Promise<ExpectedValue | null>;
+      getWithMetadata<ExpectedValue = unknown>(key: Array<Key>, options?: KVNamespaceGetOptions<"json">): Promise<ExpectedValue | null>;
+
```
radev, line 185:
```suggestion
      getWithMetadata<Metadata = unknown>(key: Array<Key>, type: "text"): Promise<Map<string, KVNamespaceGetWithMetadataResult<string, Metadata>>;
      getWithMetadata<ExpectedValue = unknown, Metadata = unknown>(key: Array<Key>, type: "json"): Promise<Map<string, KVNamespaceGetWithMetadataResult<ExpectedValue, Metadata>>;
      getWithMetadata<Metadata = unknown>(key: Array<Key>, options?: Partial<KVNamespaceGetOptions<undefined>>): Promise<Map<string, KVNamespaceGetWithMetadataResult<string, Metadata>>;
      getWithMetadata<Metadata = unknown>(key: Array<Key>, options?: KVNamespaceGetOptions<"text">): Promise<Map<string, KVNamespaceGetWithMetadataResult<string, Metadata>>;
      getWithMetadata<ExpectedValue = unknown, Metadata = unknown>(key: Array<Key>, options?: KVNamespaceGetOptions<"json">): Promise<Map<string, KVNamespaceGetWithMetadataResult<ExpectedValue, Metadata>>;

```

radev (commented):
Fix Typescript definitions for Array<Key> variants.

PR #3630, opened 2025-02-28
Roll V8 to 13.4

erikcorry:


jasnell (approved):
These are always difficult to review. I scanned the patch set and didn't spot anything questionable but should ask... I assume there were no non-trivial changes required to any of the existing patches? Overall LGTM tho.

patches/v8/0003-Allow-Windows-builds-under-Bazel.patch
```diff
@@ -1,113 +1,9 @@
...
--        ],
--        "//conditions:default": [],
-     }) + select({
-         "@v8//bazel/config:is_windows_64bit": [
-             "src/diagnostics/unwinding-info-win64.cc",
-@@ -3833,6 +3833,9 @@ filegroup(
-         "@v8//bazel/config:is_msvc_asm_ia32": ["src/heap/base/asm/ia32/push_registers_masm.asm"],
-         "@v8//bazel/config:is_msvc_asm_x64": ["src/heap/base/asm/x64/push_registers_masm.asm"],
-         "@v8//bazel/config:is_msvc_asm_arm64": ["src/heap/base/asm/arm64/push_registers_masm.S"],
-+        "@v8//bazel/config:is_windows_clang_asm_ia32": ["src/heap/base/asm/ia32/push_registers_asm.cc"],
```
fhanau, line 73: I was able to upstream parts of this patch, but didn't try to land potentially controversial aspects like clang-cl support – we still need this part of the patch.

patches/v8/0003-Allow-Windows-builds-under-Bazel.patch
```diff
@@ -1,113 +1,9 @@
...
-+    }),
- )
- 
- genrule(
-@@ -4581,7 +4586,7 @@ py_test(
-         ":noicu/d8",
-         ":noicu/v8_build_config",
-         "//testing/pybase",
--    ] + glob(["test/**"]) + glob(["tools/**/*.js"]) + glob(["tools/**/*.mjs"]),
-+    ] + glob(["test/**"], exclude = ["test/message/unicode-filename-*"],) + glob(["tools/**/*.js"]) + glob(["tools/**/*.mjs"]),
```
fhanau, line 98: Not sure if this patch is still needed – it hasn't been ported over, we'll see if Windows still has problems with the `test/message/unicode-filename-🎅🎄.js` filename.

patches/v8/0003-Allow-Windows-builds-under-Bazel.patch
```diff
@@ -1,113 +1,9 @@
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Brendan Coll <bcoll@cloudflare.com>
-Date: Thu, 16 Mar 2023 11:56:10 +0000
+From 37b9b4b1195d733b6cc10abc2b148383c84c8e82 Mon Sep 17 00:00:00 2001
+From: Erik Corry <erikcorry@chromium.org>
+Date: Wed, 26 Feb 2025 18:30:24 +0100
```
fhanau, line 3: nit: We should preserve patch authors/dates if possible

patches/v8/0005-Speed-up-V8-bazel-build-by-always-using-target-cfg.patch
```diff
@@ -1,18 +1,11 @@
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Felix Hanau <felix@cloudflare.com>
-Date: Wed, 7 Jun 2023 21:40:54 -0400
+From 1906956a221634250ed2ae32d2dae9d4ef2642a9 Mon Sep 17 00:00:00 2001
+From: Erik Corry <erikcorry@chromium.org>
+Date: Wed, 26 Feb 2025 18:33:33 +0100
 Subject: Speed up V8 bazel build by always using target cfg
 
-See the workerd build cfg changes for rationale. This provides a significant
```
fhanau, line 6: nit: Preserve PR descriptions (the second paragraph here is technically obsolete but the first still applies)

fhanau:
- [ ] While we're here, let's clean up the absl components in WORKSPACE that are no longer required now.

patches/v8/0010-Set-torque-generator-path-to-external-v8.-This-allow.patch
```diff
@@ -1,4 +1,4 @@
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
fhanau, line 1: This patch is no longer needed – the approach in https://chromium-review.googlesource.com/c/v8/v8/+/6111432 works with and without bzlmod, when building V8 directly and as a dependency.

fhanau (approved):
Couple smaller things:
- Looks like the bind() for highway is no longer needed either, we can drop it from WORKSPACE too
- [As discussed] Update ancillary dependencies (icu, fp16, highway) based on V8 DEPS as needed. For perfetto, V8 recently switched to using the latest main commit, we're currently using the last tagged release instead which we can keep IMHO.
- [As discussed] zero out commit hashes again

LGTM otherwise.

erikcorry:
Looks like ICU did not change - still bbccc2f6efc1b825de5f2c903c48be685cd0cf22
Nor did FP16 - still 0a92994d729ff76a58f692d3028ca1b64b145d91
Nor did highway - still cb1d42aaa1e14b09e1452cfdef373d051b8c02a4

For perfetto there is no newer tag than 0.49 which we already have.

fhanau:
Thank you for applying the suggested changes – LGTM, let's ship this 🚢

PR #3631, opened 2025-02-28
fix(bot): detectionIds default json

thiskevinwang:
## What

This aims to resolve a runtime discrepancy in `botManagement.detectionIds`.

## Why

We're observing two different runtime types `{}` and `number[]`, but expect only `number[]` based on [HTTP Requests Docs](https://developers.cloudflare.com/logs/reference/log-fields/zone/http_requests/#botdetectionids) — `array[int]` — as well as the [TypeScript types](https://github.com/cloudflare/workerd/blob/main/types/generated-snapshot/oldest/index.ts#L4810)

---

CloudFlare dashboard logs
| own `cURL` request | non bot traffic  |
|--------|--------|
| CleanShot 2025-02-28 at 09 58 22@2x |  |

### DIFF
--- a/src/workerd/api/cf-property.c++
+++ b/src/workerd/api/cf-property.c++
@@ -13,7 +13,7 @@ static constexpr kj::StringPtr kDefaultBotManagementValue = R"DATA({
   "verifiedBot": false,
   "jsDetection": { "passed": false },
   "staticResource": false,
-  "detectionIds": {},
+  "detectionIds": [],
   "score": 99
 })DATA"_kjc;
 


ketanhwr:
Thanks for your contribution! 
Your change makes sense to me logically, but I can see that it'll break an internal test (which might itself be wrong since the start).
cc: @jasnell who might have more context

kentonv:
This change seems correct, however, I think this value is only used if your compat date is set before `2023-08-01`. I believe this default value is being injected to work around a bug where bot management was previously passing bot management info on every request. If you instead set your compat date to `2023-08-01` or later (or set the compat flag `no_cf_botmanagement_default`) then I think you won't see this value at all.

Would it make more sense to update your compat date, or set the flag?

thiskevinwang:
@kentonv Thanks for getting back - The compatibility flag I have is:
```toml
# wrangler.toml
compatibility_date = "2024-04-03"
```

...and to clarify, I'd like to have `botManagement.detectionIds` present (the account in question is entitled to Bot Management), but the surprise/undocumented `{}` json value seen by the worker is the problem. (I expect to only be `int[]`)

And neither of the [Bot management flags](https://developers.cloudflare.com/workers/configuration/compatibility-flags/#bot-management-data) seem appropriate (if i'm reading the docs correctly)

Does this make sense? Does it add some actionable direction to this PR?

kentonv:
I see, I think this PR does not actually fix the underlying problem. In this PR you are modifying a placeholder value that is used for compatibility when people don't have bot management enabled at all (and have an old compat date). The problem seems to be that the real bot management service is actually generating `{}` when this list is empty. That is a bug in the bot management service, not in workerd.

thiskevinwang:
@kentonv Insightful! 

> That is a bug in the bot management service, not in workerd.

Would this still be the case given -> I never see occurrences of `{}` in [HTTP Request - BotDetectionIDs](https://developers.cloudflare.com/logs/reference/log-fields/zone/http_requests/#botdetectionids) log push data?

At any rate - I'll defer to you for closing this PR (it sounds like that is the direction).
And thanks for the feedback here!

kentonv:
Maybe @irvinebroque can route this to the right people on bots management to fix things there?

cysp:
💯 I have just encountered what I assume is this issue when attempting to access `.botManagement.detectionIds` in a Cloudflare Snippet.
(Acknowledging the comment above indicating that this isn't actually a Workers (or Snippets) bug, but commenting here to help others who have encountered this issue.)

PR #3636, opened 2025-02-28
Add TS support for importable env

penalosa:
This adds TypeScript support for importable env:

```ts
import { env } from "cloudflare:workers"
```

It works by typing `env` as the empty interface `Env`. This will then be merged with user `Env` declarations in `worker-configuration.d.ts` files generated by `wrangler types`, resulting in `env` being typed correctly.

As a future augmentation, we should (in `wrangler types`) modify the generated runtime types to respect the user-provided name for the `Env` interface.

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -1,3 +1,6 @@
+// Extend this in your apps to properly type Env
+interface Env {}
+
 // Namespace for RPC utility types. Unfortunately, we can't use a `module` here as these types need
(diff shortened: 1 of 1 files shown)


jasnell:
Doh... saw this after doing some cleanup and rebasing on my branch. It will need to be rebased to eliminate the conflicts.

vicb:
I'm not sure how types are generated? But does this code need to depend on the `disallow_importable_env` flag?

PR #3646, opened 2025-03-03
[o11y] Refactor Streaming Tail Worker to facilitate downstream usage

fhanau:
This is mostly moving code around to make it easier to support STW downstream.

- Disable instrumentation for STW itself for now to avoid infinite recursion
- I was convinced that adding maybeTailStreamWriter to PipelineTracer would require moving it and TailStreamWriter to a new file to avoid crazy include cycles. While that appears to not be the case, factoring some tracers out to tracer.c++/.h still seems like a good choice, we move more code there later on to keep the trace.h header small (for easier maintenance, and because lots of code includes it).
- A lifetime issue in trace-stream.c++ merits close attention.


src/workerd/io/trace-stream.c++
```diff
@@ -826,6 +826,14 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
     // the stream should be stopped and will fulfill the done promise.
     bool finishing = false;
 
+    // When we have a bona fide tail worker that receives its outcome event, we need to make sure
+    // that the final tail worker invocation is completed before destroying the tail worker
+    // customEvent and incomingRequest. To achieve this, we only fulfill the doneFulfiller after
+    // JS execution has completed.
+    // TODO(streaming-tail): I doubt this is the best possible approach here. We should be able to
```
fhanau, line 833: Not sure how to handle this well so that the JS call gets to complete before wrapping up the customEvent, but without stalling other actions unnecessarily – "option 2" below solves the problems as far as tests are concerned, but it's certainly inelegant. "option 1" results in some errors but should be cleaner. (We probably need to discuss this out-of-band).

jasnell (approved):

src/workerd/io/trace-stream.c++
fhanau: Ended up getting option 1 to work – turns out JSG didn't like me returning `jsg::Promise<nullptr>`, removing that fixes error messages during the test. Also added an error handler to reject the fulfiller if needed.

PR #3647, opened 2025-03-04
Fix new queue waitUntil drain() logic

a-robinson:
I thought that deferring the drain() in all cases could potentially be useful (e.g. what if an exception gets thrown from somewhere in this method without getting caught?), but that was dumb. This change (A) moves it to only the code path where we aren't already calling finishScheduled (which waits on waitUntilTasks), and (B) properly attaches references to the necessary state that can be accessed from JS.

Follow-up to #3596

### DIFF
--- a/src/workerd/api/queue.c++
+++ b/src/workerd/api/queue.c++
@@ -532,7 +532,6 @@ kj::Promise<WorkerInterface::CustomEvent::Result> QueueCustomEventImpl::run(
     kj::TaskSet& waitUntilTasks) {
   incomingRequest->delivered();
   auto& context = incomingRequest->getContext();
-  KJ_DEFER({ waitUntilTasks.add(incomingRequest->drain().attach(kj::mv(incomingRequest))); });
 
   kj::String queueName;
   uint32_t batchSize;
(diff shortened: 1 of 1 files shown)


jasnell (approved):
LGTM but one thing, if I may suggest, that would be helpful (for more than just queue) would be more code comment documentation explaining what the actual intended flow of these requests is expected to be.  It's non-obvious and non-trivial to understand the lifespan, expected flow, etc for these making it difficult to understand the correctness of changes like this.

anonrig (approved):

anonrig:
Can we add a test?

a-robinson:
> LGTM but one thing, if I may suggest, that would be helpful (for more than just queue) would be more code comment documentation explaining what the actual intended flow of these requests is expected to be. It's non-obvious and non-trivial to understand the lifespan, expected flow, etc for these making it difficult to understand the correctness of changes like this.

Yeah, you're right. I'll add that in soon (within the next day or two unless something very strange is going on), but I'm going to merge this for now since builds are green and it fixes a real problem.

> Can we add a test?

Perhaps it's past time to move some of the tests out here, but all the test coverage for this is in the internal repo. I'll cc you on the corresponding PR there.

a-robinson:
> I'll add that in soon

#3655

PR #3651, opened 2025-03-05
[build] 03-05-25 Bazel cleanup

fhanau:
[build] 03-05-25 Bazel cleanup
- It turns out that Bazel never uses the CXX env variable – when we need to specify the right path to clang, providing CC is sufficient
- For some V8 dependencies, replace googlesource mirror with the GitHub repository itself
- Drop superfluous repo_mapping and Windows overrides that are no longer needed for V8

[nfc] Fix -Wdangling-field warning
- LLVM 20 produces a warning since the type ptr is initialized with a constructor
argument that will go out of scope. Initializing it with the ownType class
variable instead resolves the warning.


danlapid:
LGTM

PR #3655, opened 2025-03-06
Add comments explaining the intended flow of QueueCustomEventImpl::run

a-robinson:
As requested on #3647. It's possible that this would benefit from being split up into multiple functions, but I don't really want to mess with the actual code any more until I'm back from my time off unless absolutely needed.

@jbwcloudflare @jasnell


jasnell (approved):
thank you!

PR #3656, opened 2025-03-06
Fixes asgi import at top-level failing.

dom96:
This fixes a regression introduced by https://github.com/cloudflare/workerd/pull/3505.

I've looked into moving when setupPatches gets called but it depends on pyodide being initialised... whereas the memory snapshot stuff seems to require being called before pyodide initialisation is finalised.

Anyway, I'd love @hoodmane's thoughts on this when he's back.

### DIFF
--- a/src/pyodide/python-entrypoint-helper.ts
+++ b/src/pyodide/python-entrypoint-helper.ts
@@ -80,6 +80,9 @@ async function setupPatches(pyodide: Pyodide): Promise<void> {
     pyodide.site_packages = `/lib/python${pymajor}.${pyminor}/site-packages`;
 
     // Inject modules that enable JS features to be used idiomatically from Python.
+    //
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
Makes sense. We can remove all these exceptions on 0.27 once we rearrange to `finalizeBootstrap()` before taking the snapshot. But we'll need to keep them for 0.26.

hoodmane:
To be clear though, this is not a regression introduced by #3505 but a problem that always existed and #3505 helps us to notice it in workerd tests.

dom96:
Well, unfortunately it broke a script that was deployed and working fine (Garrett's pet AI worker), so it wasn't just that it caused workerd tests to fail.

hoodmane:
I think this needs a test.

dom96:
There is a test in EW

PR #3666, opened 2025-03-07
Implement node:crypto getCipherInfo(...)

jasnell:


src/workerd/api/node/crypto.c++
```diff
@@ -1066,6 +1066,79 @@ jsg::BufferSource CryptoImpl::privateEncrypt(jsg::Lock& js,
...
 }
 
+namespace {
+ncrypto::Cipher getCipher(kj::OneOf<kj::String, int>& nameOrNid) {
+  KJ_SWITCH_ONEOF(nameOrNid) {
+    KJ_CASE_ONEOF(nid, int) {
+      return ncrypto::Cipher::FromNid(nid);
+    }
+    KJ_CASE_ONEOF(name, kj::String) {
+      std::string_view nameStr(name.begin(), name.size());
```
anonrig, line 1076: Shouldn't this be cStr?

```suggestion
      std::string_view nameStr(name.cStr(), name.size());
```

src/workerd/api/node/crypto.c++
```diff
@@ -1066,6 +1066,79 @@ jsg::BufferSource CryptoImpl::privateEncrypt(jsg::Lock& js,
...
+            if (len != ivLength) return kj::none;
+            break;
+        }
+        ivLength = len;
+      }
+    }
+
+    auto nameView = cipher.getName();
+    auto modeView = cipher.getModeLabel();
+    kj::String name = kj::str(kj::heapArray<char>(nameView.data(), nameView.size()));
```
anonrig, line 1127: At this point I feel like we should add a static method to kj string to initialize from string_view

src/workerd/api/node/crypto.h
```diff
@@ -385,6 +385,26 @@ class CryptoImpl final: public jsg::Object {
       jsg::BufferSource buffer,
       PublicPrivateCipherOptions options);
 
+  struct CipherInfo {
+    kj::String name;
+    int nid;
+    jsg::Optional<int> blockSize;
+    jsg::Optional<int> ivLength;
+    int keyLength;
+    kj::String mode;  // 'cbc', 'ccm', 'cfb', 'ctr', 'ecb', 'gcm', 'ocb',
```
anonrig, line 394: Maybe a todo to make this an enum to avoid serialization?

src/workerd/api/node/crypto.h
```diff
@@ -385,6 +385,26 @@ class CryptoImpl final: public jsg::Object {
...
+                      // 'ofb', 'stream', 'wrap', 'xts'
+    JSG_STRUCT(name, nid, blockSize, ivLength, keyLength, mode)
+  };
+
+  struct GetCipherInfoOptions {
+    jsg::Optional<int> keyLength;
```
anonrig, line 400: Should these be unsigned?

src/workerd/api/node/tests/crypto_cipher-test.js
```diff
@@ -299,3 +301,47 @@ export const missingArgChecks = {
...
+      strictEqual(typeof info.keyLength, 'number');
+      strictEqual(typeof info.mode, 'string');
+    });
+  },
+};
+
+export const test_cipher_info = {
+  test() {
+    const info = getCipherInfo('aes-128-cbc', {
+      ivLength: 16,
```
anonrig, line 325: Can you add a test that validates negative ivLength?

src/workerd/api/node/crypto.h
jasnell: eh, this isn't that much of a hot path method in any use I've ever seen it in. Not convinced it is worth the added complexity of an enum.

src/workerd/api/node/crypto.h
jasnell: They are ints in openssl and node.js.

src/workerd/api/node/crypto.c++
jasnell: Having a `KJ_STRINGIFY(std::string_view&)` somewhere should be enough. Can be done in a separate PR.

jasnell: Can be done with a KJ_STRINGIFY(...) function. Doesn't need to be in kj. Can be done in a separate PR.

PR #3677, opened 2025-03-10
Improve error message when SQL not enabled for Durable Objects

irvinebroque:
- The previous error messaged first mentioned workerd. But most people don't configure things directly through workerd, they use Wrangler.
- The previous error message only mentioned wrangler.toml, even though now wrangler.jsonc is the default

### DIFF
--- a/src/workerd/api/actor-state.c++
+++ b/src/workerd/api/actor-state.c++
@@ -722,10 +722,11 @@ SqliteDatabase& DurableObjectStorage::getSqliteDb(jsg::Lock& js) {
       // We're presumably running local workerd, which always uses SQLite for DO storage, but we're
       // trying to simulate a non-SQLite DO namespace for testing purposes.
       JSG_FAIL_REQUIRE(Error,
-          "SQL is not enabled for this Durable Object class. To enable it, set "
-          "`enableSql = true` in your workerd config for the class. If using wrangler, "
-          "under `[[migrations]]` in wrangler.toml, change `new_classes` to "
-          "`new_sqlite_classes`. Note that this change cannot be made after the class is "
+          "SQL is not enabled for this Durable Object class. To enable it, change "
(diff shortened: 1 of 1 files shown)


MellowYarker (approved):
Thanks for catching this :slightly_smiling_face:.

MellowYarker:
Just a heads up @irvinebroque, the `enableSql = true` part isn't wrapped in double quotes, which is why the build is failing.

byule:
Nit: I don't love that workerd has to be concerned with wrangler or it's config format, but I don't initially have a better idea.

src/workerd/api/actor-state.c++
```diff
@@ -722,10 +722,11 @@ SqliteDatabase& DurableObjectStorage::getSqliteDb(jsg::Lock& js) {
...
       // trying to simulate a non-SQLite DO namespace for testing purposes.
       JSG_FAIL_REQUIRE(Error,
-          "SQL is not enabled for this Durable Object class. To enable it, set "
-          "`enableSql = true` in your workerd config for the class. If using wrangler, "
-          "under `[[migrations]]` in wrangler.toml, change `new_classes` to "
-          "`new_sqlite_classes`. Note that this change cannot be made after the class is "
+          "SQL is not enabled for this Durable Object class. To enable it, change "
+          "`new_classes` to `new_sqlite_classes` within the 'migrations' field in "
+          "your wrangler.jsonc or wrangler.toml file. If using workerd directly,"
+          "set "`enableSql = true` in your workerd config for the class. Note "
```
kentonv, line 728:
```suggestion
          "set `enableSql = true` in your workerd config for the class. Note "
```

vy-ton (approved):

irvinebroque:
If we _really_ wanted to wrangler/miniflare/vite could take error message and transform it, present its own. But feels like lot of coordination

kentonv (approved):

PR #3684, opened 2025-03-10
Support x25519 for stateless diffieHellman

jasnell:
Other named curves using ec keys still not supported and will remain unsupported for the time being.

### DIFF
--- a/src/node/internal/crypto_dh.ts
+++ b/src/node/internal/crypto_dh.ts
@@ -313,12 +313,6 @@ export function diffieHellman(options: DiffieHellmanKeyPair): Buffer {
       'must equal privateKey.asymmetricKeyType'
     );
   }
-  // if (
-  //   publicKey.asymmetricKeyType !== 'dh' ||
-  //   privateKey.asymmetricKeyType !== 'dh'
-  // ) {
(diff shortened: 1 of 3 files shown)


src/node/internal/crypto_dh.ts
```diff
@@ -300,12 +301,19 @@ export function diffieHellman(options: DiffieHellmanKeyPair): Buffer {
...
+      publicKey.asymmetricKeyType,
+      'must equal privateKey.asymmetricKeyType'
+    );
   }
+  // if (
+  //   publicKey.asymmetricKeyType !== 'dh' ||
+  //   privateKey.asymmetricKeyType !== 'dh'
+  // ) {
+  //   throw new ERR_INVALID_ARG_TYPE('options', 'DiffieHellman keys', options);
+  // }
```
anonrig, line 316:
```suggestion
```

anonrig (approved):

src/node/internal/crypto_dh.ts
```diff
@@ -305,12 +306,19 @@ export function diffieHellman(options: DiffieHellmanKeyPair): Buffer {
...
```
jasnell, line 321:
```suggestion
```

PR #3685, opened 2025-03-10
Use prod bucket URL for pyodide-capnp-bin.

dom96:
Small cleanup to use our prod bucket for these.

Verified that I can access https://pyodide-capnp-bin.edgeworker.net/pyodide_0.26.0a2_2024-03-01_19.capnp.bin


hoodmane (approved):
Thanks!

PR #3688, opened 2025-03-10
Implement node:crypto ECDH

jasnell:


src/node/internal/crypto_dh.ts
```diff
@@ -312,3 +317,169 @@ export function diffieHellman(options: DiffieHellmanKeyPair): Buffer {
...
+  ): void;
+}
+
+export const ECDH = function (this: ECDH, curveName: string) {
+  if (!(this instanceof ECDH)) {
+    return new ECDH(curveName);
+  }
+  if (typeof curveName !== 'string') {
+    throw new ERR_INVALID_ARG_TYPE('curveName', 'string', curveName);
+  }
```
anonrig, line 345: Why not call validateString()?

src/workerd/api/node/crypto.c++
```diff
@@ -1141,4 +1141,176 @@ jsg::Optional<CryptoImpl::CipherInfo> CryptoImpl::getCipherInfo(
...
+
+  ncrypto::ClearErrorOnReturn clear_error_on_return;
+
+  JSG_REQUIRE(key_.checkKey(), Error, "Invalid keypair");
+
+  auto pub = bufferToPoint(group_, otherPublicKey);
+  JSG_REQUIRE(pub, Error, "Invalid to set ECDH public key");
+
+  int field_size = EC_GROUP_get_degree(group_);
+  size_t out_len = (field_size + 7) / 8;
```
anonrig, line 1223: How come no static cast is required here? Field_size is int, and casting it into size_t would remove the unsigned attribute.

anonrig (approved):

src/workerd/api/node/crypto.c++
jasnell: converting from signed to unsigned here is generally safe. Plus this is copied directly from node.js

anonrig (approved):

PR #3689, opened 2025-03-10
Implements Request class in Python SDK.

dom96:
Implements a Python native `Request` class based on https://developer.mozilla.org/en-US/docs/Web/API/Request/Request.

The API tries to match the `Response` API as closely as possible.

### Test Plan

```
$ bazel run @workerd//src/workerd/server/tests/python:sdk_development@
```

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -1,6 +1,8 @@
 # This module defines a Workers API for Python. It is similar to the API provided by
 # JS Workers, but with changes and additions to be more idiomatic to the Python
 # programming language.
+import http.client
+import json
 from collections.abc import Generator, Iterable, MutableMapping
 from contextlib import ExitStack, contextmanager
 from enum import StrEnum
@@ -18,7 +20,7 @@ JSBody = (
     "js.ReadableStream | js.URLSearchParams"
 )
 Body = "str | FormData | JSBody"
-Headers = dict[str, str] | list[tuple[str, str]]
+Headers = "dict[str, str] | list[tuple[str, str]] | js.Headers"
 
 
 # https://developers.cloudflare.com/workers/runtime-apis/request/#the-cf-property-requestinitcfproperties
@@ -42,7 +44,7 @@ class RequestInitCfProperties(TypedDict, total=False):
 # This matches the Request options:
 # https://developers.cloudflare.com/workers/runtime-apis/request/#options
 class FetchKwargs(TypedDict, total=False):
-    headers: Headers | None
+    headers: "Headers | None"
     body: "Body | None"
     method: HTTPMethod = HTTPMethod.GET
     redirect: str | None
@@ -53,16 +55,14 @@ class FetchKwargs(TypedDict, total=False):
 #       duplicates are lost, we should fix that so it returns a http.client.HTTPMessage
 class FetchResponse(pyodide.http.FetchResponse):
     # TODO: Consider upstreaming the `body` attribute
+    # TODO: Behind a compat flag make this return a native stream (StreamReader?), or perhaps
+    #       behind a different name, maybe `stream`?
     @property
-    def body(self) -> Body:
+    def body(self) -> "js.ReadableStream":
         """
-        Returns the body from the JavaScript Response instance.
+        Returns the body as a JavaScript ReadableStream from the JavaScript Response instance.
         """
-        b = self.js_response.body
-        if b.constructor.name == "FormData":
-            return FormData(b)
-        else:
-            return b
+        return self.js_response.body
 
     @property
     def js_object(self) -> "js.Response":
@@ -74,14 +74,16 @@ class FetchResponse(pyodide.http.FetchResponse):
     Some methods are implemented by `FetchResponse`, these include `buffer`
     (replacing JavaScript's `arrayBuffer`), `bytes`, `json`, and `text`.
 
-    Some methods are intentionally not implemented, these include `blob`.
-
     There are also some additional methods implemented by `FetchResponse`.
     See https://pyodide.org/en/stable/usage/api/python-api/http.html#pyodide.http.FetchResponse
     for details.
     """
 
-    async def formData(self) -> "FormData":
+    async def formData(self) -> "FormData":  # TODO: Remove after certain compat date.
+        return await self.form_data()
+
+    async def form_data(self) -> "FormData":
+        self._raise_if_failed()
         try:
             return FormData(await self.js_response.formData())
         except JsException as exc:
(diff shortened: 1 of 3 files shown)


jasnell:
@anonrig ... given that this is a Python implementation of the class, conformance with the wpt is a bit secondary.

jasnell (approved):

anonrig:
> @anonrig ... given that this is a Python implementation of the class, conformance with the wpt is a bit secondary.

Even if it's a Python implementation we might introduce a major behavior difference between the spec and our implementation that I'm afraid we might have to live with for a really long time.

dom96:
Admittedly I don't have a lot of context on WPT, but my understanding is that they are all JS tests for the browser (or other web platform), so I don't think they can be run on a Python worker easily. But the Python code is mostly wrapping the JS anyway, so behaviour should be spec compliant in those cases, where we don't simply wrap we try to make the API more idiomatic to Python which WPT wouldn't have test coverage for.

In the long-term we'll probably want to create standards for Python on the web platform and create a test suite specifically for it, maybe that can even become part of WPT, but for now I don't think we should worry about it.

src/pyodide/internal/workers.py
```diff
@@ -53,16 +55,14 @@ class FetchKwargs(TypedDict, total=False):
...
     @property
-    def body(self) -> Body:
+    def body(self) -> "js.ReadableStream":
         """
-        Returns the body from the JavaScript Response instance.
+        Returns the body as a JavaScript ReadableStream from the JavaScript Response instance.
         """
-        b = self.js_response.body
-        if b.constructor.name == "FormData":
-            return FormData(b)
```
hoodmane, line 63: Why did we have the special case for `FormData` before? And why are we removing it now?

src/pyodide/internal/workers.py
```diff
@@ -457,3 +466,137 @@ def name(self) -> str:
...
+    def credentials(self) -> str:
+        return self.js_object.credentials
+
+    @property
+    def destination(self) -> str:
+        return self.js_object.destination
+
+    @property
+    def headers(self) -> http.client.HTTPMessage:
+        result = http.client.HTTPMessage()
```
hoodmane, line 511: Is there a reason to use `HTTPMessage` here instead of a dict? I'm pretty sure that the JavaScript `Headers` object will never give us multiple copies of the same key, instead it combines them with a comma. Or perhaps we should split the values on commas and turn them back into multiple entries?

src/pyodide/internal/workers.py
```diff
@@ -457,3 +466,137 @@ def name(self) -> str:
...
+
+        for entry in self.js_object.headers:
+            result[entry[0]] = entry[1]
```
hoodmane, line 514:
```suggestion
        for k, v in self.js_object.headers:
            result[k] = v
```

src/pyodide/internal/workers.py
```diff
@@ -96,6 +95,10 @@ def replace_body(self, body: Body) -> "FetchResponse":
         js_resp = js.Response.new(b, self.js_response)
         return FetchResponse(js_resp.url, js_resp)
 
+    async def blob(self) -> "Blob":
+        self._raise_if_failed()
+        return _js_value_to_py(await self.js_object.blob())
```
hoodmane, line 100: Unless it's possible for `blob()` to return objects other than `Blob`s, I'd think we should do:
```suggestion
        return Blob(await self.js_object.blob())
```

src/pyodide/internal/workers.py
```diff
@@ -457,3 +466,137 @@ def name(self) -> str:
...
+    async def formData(self) -> "FormData":
+        self._raise_if_failed()
+        try:
+            return FormData(await self.js_object.formData())
+        except JsException as exc:
+            raise _to_python_exception(exc) from exc
+
+    async def blob(self) -> Blob:
+        self._raise_if_failed()
+        return _js_value_to_py(await self.js_object.blob())
```
hoodmane, line 582:
```suggestion
        return Blob(await self.js_object.blob())
```

hoodmane (approved):
I made some comments but it generally looks good to me.

src/pyodide/internal/workers.py
dom96: I was wondering the same thing and I couldn't recall why this was added. I don't think `body` can return anything other than ReadableStream so this code does nothing.

src/pyodide/internal/workers.py
dom96: The main reason is because it supports multiple header values with the same key.

But good catch, I thought the iterator in JS would yield all the combinations without the commas but that doesn't seem to be the case.

anonrig:
> Admittedly I don't have a lot of context on WPT, but my understanding is that they are all JS tests for the browser (or other web platform), so I don't think they can be run on a Python worker easily. But the Python code is mostly wrapping the JS anyway, so behaviour should be spec compliant in those cases, where we don't simply wrap we try to make the API more idiomatic to Python which WPT wouldn't have test coverage for.
> 
> In the long-term we'll probably want to create standards for Python on the web platform and create a test suite specifically for it, maybe that can even become part of WPT, but for now I don't think we should worry about it.

We can at least port the tests from WPT to Python to ensure we are compliant.

dom96:
We can do so, but I don't think that will give us much signal. We should focus on testing where we diverge from the JS and let the existing JS tests verify that the underlying platform that we wrap is compliant.

anonrig (approved):

anonrig:
> We can do so, but I don't think that will give us much signal. We should focus on testing where we diverge from the JS and let the existing JS tests verify that the underlying platform that we wrap is compliant.

For correction: I'm not blocking, but I still think we should have some of test (preferably from WPT) to validate the Request class API surface is same as the web spec dictates. It is fine if we do it later, but I strongly believe we should do it. 

It is also fine not to do it, and don't be spec compliant even though we implement a web spec. 

I approved the PR because it is in a good state that I'm comfortable with approving.

anonrig:
@danlapid thats a really good solution and a path forward in my honest opinion.

hoodmane:
Once we move to Pyodide 0.27 as the default, we can encourage our users to use the requests package directly to make requests. I am not sure whether there are any features that are missing when using the requests package that would be present when using the fetch API directly. Perhaps we should wait until we find out that there are specific needs before adding anything specific to the sdk.

dom96:
I think that in the same way we have defined https://docs.rs/worker/latest/worker/struct.Request.html in Rust, we want the same in Python. I don't think we want to tie ourselves to any particular third-party library for our API. We can certainly offer nice interoperability with `requests` in the same way that we do for fastapi, but it shouldn't be the default way to write Python Workers.

hoodmane:
Well in our case, it would be helpful not to have something distinct from `pyodide.http` in the long term. Unless there is a reason for it to be cloudflare-specific.

PR #3693, opened 2025-03-11
Revert "Add TS support for importable env"

penalosa:
Reverts cloudflare/workerd#3636 after reports of breakage for users who use type rather than interface to define their env type

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -1,6 +1,3 @@
-// Extend this in your apps to properly type Env
-interface Env {}
-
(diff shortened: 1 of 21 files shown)


jahands:
Example: my types that use `import()` broke after this change:
```ts
type Env = import('@repo/hono-helpers/src/types').SharedHonoBindings & {
	CF_ACCESS_CLIENT_ID: string
	CF_ACCESS_CLIENT_SECRET: string
}
```

vicb (approved):

PR #3696, opened 2025-03-11
Move pipelines types to module cloudflare:pipelines

hhoughgg:
Inherit Env from WorkerEntrypoint in PipelineEntrypoint


types/defines/pipelines.d.ts
```diff
@@ -1,30 +1,33 @@
...
-   */
-  send(records: T[]): Promise<void>
+    public run(records: I[], metadata: PipelineBatchMetadata): Promise<O[]>;
+  }
+  export type PipelineRecord = Record<string, unknown>
+  export type PipelineBatchMetadata = {
+    pipelineId: string;
+    pipelineName: string;
+  }
+  export interface Pipeline<T extends PipelineRecord> {
```
oliy, line 25: I assume this does mean that it will break existing bindings that will now need to import the type to add it to their Env

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -65,8 +65,10 @@ type PipelineBatchMetadata = {
 type PipelineRecord = Record<string, unknown>;
 
 export class PipelineTransformImpl<
-  I extends PipelineRecord,
-  O extends PipelineRecord,
+  // @ts-ignore
```
oliy, line 68: why is this needed?

hhoughgg: It give a warning that Env is not used

types/defines/pipelines.d.ts
hhoughgg: Yes - I think we are ok with this since only a few internal users and we have other breaking changes - such as db schema.

oliy (approved):

cmackenzie1:
LGTM!

a-robinson (approved):
LGTM

PR #3699, opened 2025-03-11
Rethrow abort exception on IoContext::run

ObsidianMinor:
We're seeing errors getting reported from actors trying to run promises without an active request. We've determined this is from tasks trying to run after an actor's IO context has been aborted. This change stores the abort exception in the `IoContext` and rethrows it in `IoContext::run`.

### DIFF
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -336,6 +336,7 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler 
 
   // Force context abort now.
   void abort(kj::Exception&& e) {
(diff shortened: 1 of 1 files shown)


tewaro (approved):

anonrig:
Can we add a test to not regress in the future?

ObsidianMinor:
> Can we add a test to not regress in the future?

Maybe...? It'd have to be incredibly hacky since this is just turning an assertion failure that was only observable as a log message into another log message that isn't reported as an internal error...

ObsidianMinor:
I'll continue looking into adding a test for this but I don't think it's worth blocking since this is one of our highest recorded issues.

PR #3700, opened 2025-03-12
Make the D1 Sessions API default allowed for everyone

lambrospetrou:
We no longer need the `enable_d1_with_sessions_api` experimental compatibility flag since the `withSession()` method should be exposed to everyone.

Internally: CFSQL-1220

## Tests

```
bazel run //src/cloudflare/internal/test/d1:d1-api-test-with-sessions
bazel run //src/cloudflare/internal/test/d1:d1-api-test
```

### DIFF
--- a/src/cloudflare/internal/d1-api.ts
+++ b/src/cloudflare/internal/d1-api.ts
@@ -2,8 +2,6 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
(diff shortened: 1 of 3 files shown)


danlapid (approved):

lambrospetrou:
This caused some issues with other codebases doing `.name` comparisons on the returned class (vitest pool workers), and things like Rust workers that try to statically cast the returned binding class into the `D1Database` and fail.

https://github.com/cloudflare/workerd/pull/3726 fixes this.

PR #3703, opened 2025-03-12
Importable env types, take 2

penalosa:
This adds a namespaced type for importable env, `Cloudflare.Env`. This can be provided by a user `.d.ts` file via something like:

```ts
declare namespace Cloudflare {
 	interface Env {
 		string: 'hello world';
 		json: { key: 'value' };
 		SECRET: string;
	}
}
```

This will be generated by `wrangler types` in the near future

### DIFF
--- a/types/defines/rpc.d.ts
+++ b/types/defines/rpc.d.ts
@@ -151,6 +151,10 @@ declare namespace Rpc {
   };
 }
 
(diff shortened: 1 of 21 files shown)


hoodmane (approved):
Makes sense to me. What was take 1?

danlapid (approved):

PR #3706, opened 2025-03-12
Repro of issue with RPC & promise pipelining

penalosa:
Reproduces what we think is a bug in the runtime, that is related to promise pipelining.

The sample sets up three Workers that service bind in the following way: `worker-a` -> `proxy-worker` -> `user-worker`. The proxy Worker acts as a "proxy", by forwarding the `foo()` RPC calls to the `user-worker` 

```
foo(emoji) {
    return this.env.USER.foo(emoji);
}
```

Running this sample will result in `worker-a` throwing an error: `RPC stub used after being disposed.`.

### DIFF
--- a/samples/promise-pipelining/proxy-worker.js
+++ b/samples/promise-pipelining/proxy-worker.js
@@ -1,0 +1,7 @@
+import { WorkerEntrypoint } from 'cloudflare:workers';
+export default class Worker extends WorkerEntrypoint {
+  fetch() {}
+  foo(emoji) {
+    return this.env.USER.foo(emoji);
+  }
+}
--- a/samples/promise-pipelining/config.capnp
+++ b/samples/promise-pipelining/config.capnp
@@ -1,0 +1,39 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const tailWorkerExample :Workerd.Config = (
+  services = [
+    (name = "worker-a", worker = .workerA),
+    (name = "proxy-worker", worker = .proxyWorker),
+    (name = "user-worker", worker = .userWorker),
+  ],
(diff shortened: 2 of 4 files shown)


mhart:
Just to clarify – the promise pipelining works without the intermediate `proxy-worker` right? It's the introduction of the proxy that's (presumably) making workerd think the stub is done and disposing of it too early.

samples/promise-pipelining/proxy-worker.js
```diff
@@ -0,0 +1,7 @@
+import { WorkerEntrypoint } from 'cloudflare:workers';
+export default class Worker extends WorkerEntrypoint {
+  fetch() {}
+  foo(emoji) {
```
mhart, line 4: Does the error still happen if this is declared async or if the result below is wrapped in a Promise? I expect it would, but just checking

penalosa: Yeah—this reproduces with an async function (with and without await)

CarmenPopoviciu:
> Just to clarify – the promise pipelining works without the intermediate `proxy-worker` right? It's the introduction of the proxy that's (presumably) making workerd think the stub is done and disposing of it too early.

yes that is correct. without that "middle layer" proxy Worker, everything works fine. Once we add that proxy Worker, things break

CarmenPopoviciu:
@mhart also to clarify, I have also tested this setup in production, by deploying three workers with this config, and am seeing the same error

CarmenPopoviciu:
one more detail here is that the following works as expected, though it kind of defeats the purpose of promise pipelining

```
export default {
  async fetch(request, env) {
    const foo = await env.PROXY.foo('🐜');
    return new Response(await foo.bar.buzz());
  },
};
```

PR #3708, opened 2025-03-12
Fix minor undefined behavior in modules-new-test

fhanau:
When running under high load, a race condition in modules-new-test.c++ could lead to the V8 isolate being destroyed too late, leading to a member access within null pointer during isolate deallocation.

As suggested by @danlapid


jasnell (approved):
Heh, I just spotted this yesterday. To avoid some merge conflicts, if you don't mind, let me incorporate this change into the other PR that I've been in the process of preparing that I'll likely open tomorrow. I'm good with this landing on its own if it must but I can just as easily just pull it in with the other PR also.

fhanau:
> Heh, I just spotted this yesterday. To avoid some merge conflicts, if you don't mind, let me incorporate this change into the other PR that I've been in the process of preparing that I'll likely open tomorrow. I'm good with this landing on its own if it must but I can just as easily just pull it in with the other PR also.

The change itself is not that important – not sure to what extent this is causing downstream tests to be flaky though

anonrig (approved):

fhanau:
Merged as part of #3712.

PR #3712, opened 2025-03-12
More work on new module registry

jasnell:
More work on the new module registry. This is just an interim PR to keep the change set relatively small. There are more PRs to come. As such, unless they are breaking issues I likely will make any follow on changes from code review in the next PR.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -872,7 +872,7 @@ void ServiceWorkerGlobalScope::reportError(jsg::Lock& js, jsg::JsValue error) {
 
 jsg::JsValue ServiceWorkerGlobalScope::getBuffer(jsg::Lock& js) {
   KJ_ASSERT(FeatureFlags::get(js).getNodeJsCompatV2());
-  static const auto kSpecifier = "node:buffer"_kj;
+  constexpr auto kSpecifier = "node:buffer"_kj;
   KJ_IF_SOME(module, js.resolveModule(kSpecifier)) {
     auto def = module.get(js, "default"_kj);
     auto obj = KJ_ASSERT_NONNULL(def.tryCast<jsg::JsObject>());
@@ -889,7 +889,7 @@ jsg::JsValue ServiceWorkerGlobalScope::getBuffer(jsg::Lock& js) {
 
 jsg::JsValue ServiceWorkerGlobalScope::getProcess(jsg::Lock& js) {
   KJ_ASSERT(FeatureFlags::get(js).getNodeJsCompatV2());
-  static const auto kSpecifier = "node:process"_kj;
+  constexpr auto kSpecifier = "node:process"_kj;
   KJ_IF_SOME(module, js.resolveModule(kSpecifier)) {
     auto def = module.get(js, "default"_kj);
     JSG_REQUIRE(def.isObject(), TypeError, "Invalid node:process implementation");
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -870,47 +870,36 @@ void ServiceWorkerGlobalScope::reportError(jsg::Lock& js, jsg::JsValue error) {
   }
 }
 
-namespace {
-jsg::JsValue resolveFromRegistry(jsg::Lock& js, kj::StringPtr specifier) {
-  auto moduleRegistry = jsg::ModuleRegistry::from(js);
-  if (moduleRegistry == nullptr) {
-    // TODO: Return the known built-in instead? This gets a bit tricky as we currently
-    // have no mechanism defined for accessing and caching the built-in module without
-    // the module registry. Should we even support this case at all? Without the module
-    // registry the user can't access the other importable modules anyway. For now, just
-    // returning undefined in this case seems the most appropriate.
-    return js.undefined();
-  }
-
-  auto spec = kj::Path::parse(specifier);
-  auto& info = JSG_REQUIRE_NONNULL(
-      moduleRegistry->resolve(js, spec), Error, kj::str("No such module: ", specifier));
-  auto module = info.module.getHandle(js);
-
-  jsg::instantiateModule(js, module);
-  return jsg::JsValue(js.v8Get(module->GetModuleNamespace().As<v8::Object>(), "default"_kj));
-}
-}  // namespace
-
 jsg::JsValue ServiceWorkerGlobalScope::getBuffer(jsg::Lock& js) {
   KJ_ASSERT(FeatureFlags::get(js).getNodeJsCompatV2());
-  auto value = resolveFromRegistry(js, "node:buffer"_kj);
-  auto obj = JSG_REQUIRE_NONNULL(
-      value.tryCast<jsg::JsObject>(), TypeError, "Invalid node:buffer implementation");
-  // Unlike the getProcess() case below, this getter is returning an object that
-  // is exported by the node:buffer module and not the module itself, so we need
-  // this additional get to the grab the reference to the thing we're actually
-  // returning.
-  auto buffer = obj.get(js, "Buffer"_kj);
-  JSG_REQUIRE(buffer.isFunction(), TypeError, "Invalid node:buffer implementation");
-  return buffer;
+  static const auto kSpecifier = "node:buffer"_kj;
+  KJ_IF_SOME(module, js.resolveModule(kSpecifier)) {
+    auto def = module.get(js, "default"_kj);
+    auto obj = KJ_ASSERT_NONNULL(def.tryCast<jsg::JsObject>());
+    auto buffer = obj.get(js, "Buffer"_kj);
+    JSG_REQUIRE(buffer.isFunction(), TypeError, "Invalid node:buffer implementation");
+    return buffer;
+  } else {
+    // If we are unable to resolve the node:buffer module here, it likely
+    // means that we don't actually have a module registry installed. Just
+    // return undefined in this case.
+    return js.undefined();
+  }
 }
 
 jsg::JsValue ServiceWorkerGlobalScope::getProcess(jsg::Lock& js) {
   KJ_ASSERT(FeatureFlags::get(js).getNodeJsCompatV2());
-  auto value = resolveFromRegistry(js, "node:process"_kj);
-  JSG_REQUIRE(value.isObject(), TypeError, "Invalid node:process implementation");
-  return value;
+  static const auto kSpecifier = "node:process"_kj;
+  KJ_IF_SOME(module, js.resolveModule(kSpecifier)) {
+    auto def = module.get(js, "default"_kj);
+    JSG_REQUIRE(def.isObject(), TypeError, "Invalid node:process implementation");
+    return def;
+  } else {
+    // If we are unable to resolve the node:process module here, it likely
+    // means that we don't actually have a module registry installed. Just
+    // return undefined in this case.
+    return js.undefined();
+  }
 }
 
 double Performance::now() {
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -995,7 +995,7 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler 
       jsg::Lock& js, IoContext::ExceptionOr<Result>&& exceptionOrResult);
 };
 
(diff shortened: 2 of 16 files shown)


jasnell:
@penalosa .. can you help me understand the workers-sdk test failure here?

src/workerd/api/global-scope.c++
```diff
@@ -870,47 +870,36 @@ void ServiceWorkerGlobalScope::reportError(jsg::Lock& js, jsg::JsValue error) {
...
-  // returning.
-  auto buffer = obj.get(js, "Buffer"_kj);
-  JSG_REQUIRE(buffer.isFunction(), TypeError, "Invalid node:buffer implementation");
-  return buffer;
+  static const auto kSpecifier = "node:buffer"_kj;
+  KJ_IF_SOME(module, js.resolveModule(kSpecifier)) {
+    auto def = module.get(js, "default"_kj);
+    auto obj = KJ_ASSERT_NONNULL(def.tryCast<jsg::JsObject>());
+    auto buffer = obj.get(js, "Buffer"_kj);
+    JSG_REQUIRE(buffer.isFunction(), TypeError, "Invalid node:buffer implementation");
```
anonrig, line 880: What is the purpose of this function? Why can't we trust the output of js.resolveModule() method, and have to implement such checks on this function?

src/workerd/api/global-scope.c++
```diff
@@ -870,47 +870,36 @@ void ServiceWorkerGlobalScope::reportError(jsg::Lock& js, jsg::JsValue error) {
...
+    return buffer;
+  } else {
+    // If we are unable to resolve the node:buffer module here, it likely
+    // means that we don't actually have a module registry installed. Just
+    // return undefined in this case.
```
anonrig, line 885: How is this possible? This is quite vague to understand for me. Would you mind expanding this comment and explain how this scenario could occur?

src/workerd/io/io-context.h
```diff
@@ -992,6 +992,15 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
       jsg::Lock& js, IoContext::ExceptionOr<Result>&& exceptionOrResult);
 };
 
+// The SuppressIoContextScope utilty is used to temporarily suppress the active IoContext
```
anonrig, line 995:
```suggestion
// The SuppressIoContextScope utility is used to temporarily suppress the active IoContext
```

src/workerd/io/worker.c++
```diff
@@ -1884,7 +1875,8 @@ void Worker::handleLog(jsg::Lock& js,
     auto colors =
         COLOR_MODE == ColorMode::ENABLED || (COLOR_MODE == ColorMode::ENABLED_IF_TTY && tty);
 
-    auto inspectModule = resolveNodeInspectModule(js);
+    static constexpr auto kSpecifier = "node-internal:internal_inspect"_kj;
```
anonrig, line 1878: 3-4 months ago I think either @fhanau or @mikea mentioned that static constexpr is unnecessary and constexpr is sufficient. I don't remember why tho...

src/workerd/jsg/BUILD.bazel
```diff
@@ -86,6 +83,7 @@ wd_cc_library(
         "jsg.c++",
         "jsvalue.c++",
         "modules.c++",
+        "modules-new.c++",
```
anonrig, line 86: Can't we make "modules-new" a new target?

src/workerd/jsg/BUILD.bazel
```diff
@@ -121,6 +120,7 @@ wd_cc_library(
         ":meta",
         ":modules_capnp",
         ":observer",
+        ":url",
```
anonrig, line 123: This is the old url implementation. Any reason why we are not using the new one?

src/workerd/jsg/jsg.h
```diff
@@ -2718,6 +2718,14 @@ class Lock {
   // Retrieve the worker environment.
   virtual kj::Maybe<Value> getWorkerEnv() = 0;
 
+  // Resolve an internalk module namespace from the given specifier.
```
anonrig, line 2721:
```suggestion
  // Resolve an internal module namespace from the given specifier.
```

src/workerd/jsg/modules-new.c++
```diff
@@ -378,41 +375,62 @@ class IsolateModuleRegistry final {
       auto module = entry.key.getHandle(js);
       auto status = module->GetStatus();
 
-      if (status == v8::Module::kEvaluated) {
-        return module->GetModuleNamespace().As<v8::Object>();
+      // If status is kErrored, that means a prior attempt to evaluate the module
```
anonrig, line 378: Q: Does the new module registry uses the compile cache that is generated from gen-compile-cache step?

src/workerd/jsg/modules-new.c++
```diff
@@ -378,41 +375,62 @@ class IsolateModuleRegistry final {
...
-      if (status == v8::Module::kErrored) {
-        js.throwException(js.v8Ref(module->GetException()));
+      // TODO(new-module-registry): Circular dependencies should be fine
+      // when we are talking strictly about CJS/Node.js style modules.
+      // For ESM, it becomes more problematic because v8 will not allow
+      // us to grab the default export while the module is still evaluating.
+
+      if (entry.module.isEsm() && status == v8::Module::kEvaluating) {
+        JSG_FAIL_REQUIRE(Error, "Circular dependency when resolving module: ", specifier);
       }
```
anonrig, line 391: I guess this can be simplified to JSG_REQUIRE(entry.module.isEsm() ...)

src/workerd/server/workerd-api.c++
```diff
@@ -266,6 +264,9 @@ struct WorkerdApi::Impl final {
         jsgIsolate(v8System, Configuration(*this), kj::mv(observerParam), kj::mv(createParams)),
         memoryCacheProvider(memoryCacheProvider),
         pythonConfig(pythonConfig) {
+    if (maybeOwnedModuleRegistry != kj::none) {
```
anonrig, line 267: Would you mind adding a comment here explaining this?

src/workerd/server/workerd-api.c++
```diff
@@ -1064,13 +1041,14 @@ kj::Own<jsg::modules::ModuleRegistry> WorkerdApi::initializeBundleModuleRegistry
...
 
+  // Add the built-in module bundles that support python workers/pyodide.
   if (hasPythonModules) {
     jsg::modules::ModuleBundle::BuiltinBuilder pyodideBundleBuilder;
-    auto metadataSpecifier = "pyodide-internal:runtime-generated/metadata"_url;
-    auto artifactsSpecifier = "pyodide-internal:artifacts"_url;
-    auto internalJaegerSpecifier = "pyodide-internal:internalJaeger"_url;
-    auto diskCacheSpecifier = "pyodide-internal:disk_cache"_url;
-    auto limiterSpecifier = "pyodide-internal:limiter"_url;
+    const auto metadataSpecifier = "pyodide-internal:runtime-generated/metadata"_url;
```
anonrig, line 1047: Can these be constexpr?

src/workerd/io/worker.c++
fhanau: Yeah, there's a clang-tidy lint that triggers for it (can't have a storage class for constexpr). Plain constexpr should work well

src/workerd/api/global-scope.c++
jasnell: You mean what's the purpose of the `getBuffer(...)` function? This is the accessor that makes `globalThis.Buffer` accessible when nodejs_compat is enabled. The `js.resolveModule(...)` just returns the module namespace but what we need is the `Buffer` export from that. Since user code can override the implementation of `node:buffer` we have to verify that the export is what we expect it to be.

src/workerd/api/global-scope.c++
jasnell: The original service worker syntax has no module registry but may still have the `nodejs_compat` flag enabled in which case the `globalThis.Buffer` accessor will still be defined. I'll add a comment, but note this behavior is not new, I've just modified it in this PR.

src/workerd/jsg/BUILD.bazel
jasnell: generally no. It is needed in all the same places `modules.c++|h` are needed. I will look at cleaning this part up later (both for the original modules impl and the new one) in a separate follow up PR (there will be several PRs before this work is complete)

src/workerd/jsg/BUILD.bazel
jasnell: No, this is the JSG url package.

src/workerd/jsg/modules-new.c++
jasnell: Not yet. The original implementation here predated the `gen-compile-cache` stuff and hasn't yet been updated for that. That will be one of the upcoming PRs to ensure that is all used.

src/workerd/api/global-scope.c++
anonrig: > Since user code can override the implementation of node:buffer

How come?

jasnell: That's just how it works. For historical reasons we allow the worker bundle to provide their own overriding implementations of the node:* modules. The module registry will pull `node:*` modules from the runtime only if the worker bundle does not have its own modules with those names.

src/workerd/server/workerd-api.c++
jasnell: Not currently. The `_url` operator does not support it at this time. We can do that separately.

src/workerd/jsg/modules-new.c++
jasnell: Yes but I find this form more readable, to be honest. To use `JSG_REQUIRE(...)` would involve inverting the check here and honestly this just *reads* better to me than the alternative. I'd rather keep it as is.

src/workerd/server/server.h
```diff
@@ -112,6 +112,10 @@ class Server final: private kj::TaskSet::ErrorHandler {
   class InspectorService;
   class InspectorServiceIsolateRegistrar;
 
+  void handleReportConfigError(kj::String error) {
```
danlapid, line 115: Why not just use reportConfigError?

src/workerd/server/workerd-api.c++
```diff
@@ -266,6 +264,12 @@ struct WorkerdApi::Impl final {
         jsgIsolate(v8System, Configuration(*this), kj::mv(observerParam), kj::mv(createParams)),
         memoryCacheProvider(memoryCacheProvider),
         pythonConfig(pythonConfig) {
+    // maybeOwnedModuleRegistry is only set when using the
+    // new module registry implementation. When that is the
+    // case we also need to tell JSG.
+    if (maybeOwnedModuleRegistry != kj::none) {
+      jsgIsolate.setUsingNewModuleRegistry();
```
danlapid, line 271: Honestly kinda seems like an odd way to do this.
maybeOwnedModuleRegistry is available before constructing jsgIsolate.
usingNewModuleRegistry can more logically be defined as a boolean argument to jsgIsolate's constructor

jasnell: I *really* dislike boolean arguments to constructors and this follows the pattern we have with other flag-controlled behaviors.

danlapid: Honestly I don't see how this is better

src/workerd/server/server.h
jasnell: `reportConfigError(...)` is private and the callsite for this is in an anonymous namespace in a utility class. I could move `reportConfigError(...)`, I suppose but this seems just as easy.

danlapid (approved):

src/workerd/server/server.h
danlapid: Yeah I mean, creating a public method that exports a private method is kinda funny.

jasnell: Well, `reportConfigError` is not a private method. It's a private field member whose type happens to be a function...

```
kj::Function<void(kj::String)> reportConfigError;
```

jasnell:
Internal CI run was successful with no internal changes necessary.

PR #3715, opened 2025-03-13
Fix exporting a class instance as an entrypoint.

kentonv:
This should work:

```
class Foo {
  async fetch(req, env, ctx) { ... }
}

export default new Foo();
```

But it had a slight bug: `validateHandlers()` did not report the `fetch` method as existing. This meant you couldn't actually deploy a worker defined this way as the upload validator on the server side would reject it for not implementing `fetch`.

I asked Claude Code to fix it (as a test... I knew what the problem was). Transcript: https://claude-workerd-transcript.pages.dev/validate-handlers

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -2156,6 +2156,44 @@ void Worker::Lock::validateHandlers(ValidationErrorReporter& errorReporter) {
     ignoredHandlers.insert("unhandledrejection"_kj);
     ignoredHandlers.insert("rejectionhandled"_kj);
 
+    // Helper function to collect methods from a prototype chain
+    auto collectMethodsFromPrototypeChain = [&](jsg::JsValue startProto,
+                                                kj::HashSet<kj::String>& seenNames) {
+      // Find the prototype for `Object` by creating one.
+      auto obj = js.obj();
+      jsg::JsValue prototypeOfObject = obj.getPrototype(js);
+
+      // Walk the prototype chain.
+      jsg::JsValue proto = startProto;
+      for (;;) {
+        auto protoObj = JSG_REQUIRE_NONNULL(proto.tryCast<jsg::JsObject>(), TypeError,
+            "Exported value's prototype chain does not end in Object.");
+        if (protoObj == prototypeOfObject) {
+          // Reached the prototype for `Object`. Stop here.
+          break;
+        }
+
+        // Awkwardly, the prototype's members are not typically enumerable, so we have to
+        // enumerate them rather directly.
+        jsg::JsArray properties = protoObj.getPropertyNames(js, jsg::KeyCollectionFilter::OWN_ONLY,
+            jsg::PropertyFilter::SKIP_SYMBOLS, jsg::IndexFilter::SKIP_INDICES);
+        for (auto i: kj::zeroTo(properties.size())) {
+          auto name = properties.get(js, i).toString(js);
+          if (name == "constructor"_kj) {
+            // Don't treat special method `constructor` as an exported handler.
+            continue;
+          }
+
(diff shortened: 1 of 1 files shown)


src/workerd/io/worker.c++
```diff
@@ -2182,15 +2219,28 @@ void Worker::Lock::validateHandlers(ValidationErrorReporter& errorReporter) {
...
           for (auto& field: dict.fields) {
             if (!ignoredHandlers.contains(field.name)) {
-              methods.add(kj::mv(field.name));
+              methodSet.upsert(kj::mv(field.name), [&](auto&, auto&&) {});
             }
           }
+
+          // Then, check for methods in the prototype chain (like a class instance)
+          js.withinHandleScope([&]() {
+            collectMethodsFromPrototypeChain(jsg::JsObject(handle).getPrototype(js), methodSet);
```
geelen, line 2238: So this is the meaningful line change in this PR, yeah? Previously you weren't walking the prototype chain for default exports, is that right? I guess I'm not super clear why default is special, I suppose historical reasons?

I presume `worker.impl->statelessClasses` includes both `WorkerEntrypoint` and `ExportedHandler` style exports, right? Or is `ExportedHandler` handled elsewhere?

kentonv: `statelessClasses` is strictly `WorkerEntrypoint`. `ExportedHandler` style is what we're handling here.

kentonv: To be clear this code here is used for all plain-object exports, whether default or not.

src/workerd/io/worker.c++
```diff
@@ -2156,6 +2156,44 @@ void Worker::Lock::validateHandlers(ValidationErrorReporter& errorReporter) {
     ignoredHandlers.insert("unhandledrejection"_kj);
     ignoredHandlers.insert("rejectionhandled"_kj);
 
+    // Helper function to collect methods from a prototype chain
+    auto collectMethodsFromPrototypeChain = [&](jsg::JsValue startProto,
```
jasnell, line 2160: minor nit...

```suggestion
    auto collectMethodsFromPrototypeChain = [&](const jsg::JsValue& startProto,
```

src/workerd/io/worker.c++
```diff
@@ -2156,6 +2156,44 @@ void Worker::Lock::validateHandlers(ValidationErrorReporter& errorReporter) {
...
+          break;
+        }
+
+        // Awkwardly, the prototype's members are not typically enumerable, so we have to
+        // enumerate them rather directly.
+        jsg::JsArray properties = protoObj.getPropertyNames(js, jsg::KeyCollectionFilter::OWN_ONLY,
+            jsg::PropertyFilter::SKIP_SYMBOLS, jsg::IndexFilter::SKIP_INDICES);
+        for (auto i: kj::zeroTo(properties.size())) {
+          auto name = properties.get(js, i).toString(js);
+          if (name == "constructor"_kj) {
```
jasnell, line 2182: hmm, might be more efficient to create a "constructor" JS string before the loop and compare equality of `properties.get(js, i)` to it rather than doing the `toString(...)` allocation-and-copy for each property.

jasnell (approved):

src/workerd/io/worker.c++
kentonv: A `jsg::JsValue` is just a `v8::Local` which is ultimately just a single pointer, so passing it by value is likely more efficient than passing it by reference.

src/workerd/io/worker.c++
kentonv: We need to convert each property name into a kj::String anyway for the code below.

PR #3722, opened 2025-03-13
Use IoContext::abort more consistently

ObsidianMinor:
Also ignore future aborts if abort has already been called once. This was originally implied by the behavior of fulfiller since calling reject more than once will ignore future calls.

### DIFF
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -337,6 +337,9 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler 
 
   // Force context abort now.
   void abort(kj::Exception&& e) {
+    if (abortException != kj::none) {
+      return;
+    }
     abortException = kj::cp(e);
     abortFulfiller->reject(kj::mv(e));
   }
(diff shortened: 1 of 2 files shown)


anonrig (approved):

src/workerd/io/io-context.h
```diff
@@ -337,6 +337,9 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
 
   // Force context abort now.
   void abort(kj::Exception&& e) {
+    if (abortException != kj::none) {
+      return;
```
jasnell, line 341: Hmm... I'm always a bit hesitant to silently ignore an error like this. I understand the intention but I just need to ask: should we at least log the exception in this case or is it truly ok to ignore this case?

jasnell (approved):
Code change overall LGTM with a clarifying question

src/workerd/io/io-context.h
kentonv: We've always only paid attention to the first abort, since `abortFulfiller->reject()` will only pay attention to the first call. @ObsidianMinor added the `abortException = kj::cp(e)` line yesterday which made things inconsistent if abort() was called multiple times, so this just brings us back to consistency.

I think it makes sense to only report the first abort() error as any later errors are probably just side effects or duplicates.

kentonv (approved):

ObsidianMinor:
Not sure why the internal build is failing. Base must be too old.

src/workerd/io/io-context.h
jasnell: Ok, works for me. Still feels a bit gnarly to not at least log the error but can definitely understand the reasoning

PR #3723, opened 2025-03-13
Create Secrets Store Proxy Worker types

jvaughan-cloudflare:
Description: 
This PR setups up an interface Secrets Store secret binding. The objective is to have an interface in a worker that can retrieve a secret from from the Secrets Store with a call like `await env.SECRET_KEY.get()`.

### DIFF
--- a/types/defines/secrets-store.d.ts
+++ b/types/defines/secrets-store.d.ts
@@ -1,0 +1,7 @@
+interface SecretsStoreSecret {
+  /**
+   * Get a secret from the Secrets Store, returning a string of the secret value
+   * if it exists, or throws an error if it does not exist
+   */
+  get(): Promise<string>;
+}
--- a/types/generated-snapshot/2021-11-03/index.d.ts
+++ b/types/generated-snapshot/2021-11-03/index.d.ts
@@ -6199,6 +6199,13 @@ declare module "cloudflare:workers" {
   }
   export const env: Cloudflare.Env;
 }
+interface SecretsStoreSecret {
+  /**
+   * Get a secret from the Secrets Store, returning a string of the secret value
+   * if it exists, or throws an error if it does not exist
+   */
+  get(): Promise<string>;
+}
 declare module "cloudflare:sockets" {
   function _connect(
     address: string | SocketAddress,
--- a/types/generated-snapshot/2021-11-03/index.ts
+++ b/types/generated-snapshot/2021-11-03/index.ts
@@ -6080,6 +6080,13 @@ export declare namespace Rpc {
 export declare namespace Cloudflare {
   interface Env {}
 }
+export interface SecretsStoreSecret {
+  /**
+   * Get a secret from the Secrets Store, returning a string of the secret value
+   * if it exists, or throws an error if it does not exist
+   */
+  get(): Promise<string>;
+}
 export declare namespace TailStream {
   interface Header {
     readonly name: string;
--- a/types/generated-snapshot/2022-01-31/index.d.ts
+++ b/types/generated-snapshot/2022-01-31/index.d.ts
@@ -6225,6 +6225,13 @@ declare module "cloudflare:workers" {
   }
   export const env: Cloudflare.Env;
 }
+interface SecretsStoreSecret {
+  /**
+   * Get a secret from the Secrets Store, returning a string of the secret value
+   * if it exists, or throws an error if it does not exist
+   */
+  get(): Promise<string>;
+}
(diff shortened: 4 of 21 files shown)


jasnell (approved):

jvaughan-cloudflare:
recheck

src/cloudflare/internal/secrets-store-api.ts
```diff
@@ -0,0 +1,32 @@
+type Fetcher = {
+  fetch: typeof fetch;
+};
+
+export default function makeBinding(env: { fetcher: Fetcher }): SecretsStore {
+  return new SecretsStoreBindingImpl(env.fetcher);
+}
+
+type SecretsStoreProxyWorkerResponse = {
```
mitalirawat, line 9: We can just call this SecretsStoreResponse

src/cloudflare/internal/secrets-store-api.ts
```diff
@@ -0,0 +1,32 @@
...
+  value: string;
+};
+
+class SecretsStoreBindingImpl implements SecretsStore {
+  public constructor(private readonly fetcher: Fetcher) {}
+
+  public async get(): Promise<string> {
+    try {
+      const response = await this.fetcher.fetch(
+        new Request('https://secrets-store-proxy-worker.cloudflare.com')
```
mitalirawat, line 20: we can rename this to say `https://secrets-store-service.cloudflare.com`

kentonv (changes requested):
New bindings should use RPC. There's no longer any need to implement a wrapper that lives in workerd. With RPC, you can implement your `get()` method directly in your own worker. This way you don't need to modify workerd code whenever you want to change your API.

src/cloudflare/internal/secrets-store-api.ts
```diff
@@ -0,0 +1,32 @@
...
+  name: string;
+  value: string;
+};
+
+class SecretsStoreBindingImpl implements SecretsStore {
+  public constructor(private readonly fetcher: Fetcher) {}
+
+  public async get(): Promise<string> {
+    try {
+      const response = await this.fetcher.fetch(
```
mitalirawat, line 19: i believe we'd need something like this example binding here (without the fetchOptions) 
```
const endpointUrl =
      'https://workers-binding.ai/to-everything/markdown/transformer';

    const res = await this.fetcher.fetch(endpointUrl, fetchOptions);
    ```

**Review by jasnell — APPROVED** [review]

File: types/defines/secrets-store.d.ts [hunk 4]
```diff
@@ -0,0 +1,7 @@
+export declare class SecretsStoreEntrypoint extends WorkerEntrypoint {
+  /**
+   * Get a secret from the Secrets Store, returning a string of the secret value
+   * if it exists, or throws an error if it does not exist
+   */
+  get(): Promise<string>;
+}
```
penalosa, line 7: This isn't the right type for the user-facing binding. Instead, you should have something more like https://github.com/cloudflare/workerd/blob/main/types/defines/ratelimit.d.ts. This type represents the binding, not the implementation of it in your proxy worker.

```suggestion
interface SecretsStoreSecret {
  /**
   * Get a secret from the Secrets Store, returning a string of the secret value
   * if it exists, or throws an error if it does not exist
   */
  get(): Promise<string>;
}
```

penalosa (changes requested):

types/defines/secrets-store.d.ts
jvaughan-cloudflare: Ah, I see. Am I correct in understand that with this interface definition here, a customer's worker will be able to do something like:

```
# wrangler.toml

[[secrets_store_secret]]
binding = 'API_KEY'
store_id= abc123
secret_name = API_KEY
```

```
# index.ts

interface Env {
  API_KEY: Service<SecretsStoreSecret>;
}

env.API_KEY.get()
```

penalosa: No, they'd do:

```
interface Env {
  API_KEY: SecretsStoreSecret;
}

env.API_KEY.get()
```

The fact that this binding is implemented as a JSRPC service binding is hidden from the user

kentonv (approved):

types/defines/secrets-store.d.ts
jvaughan-cloudflare: Ah ok, got it. I was confused on this after reading the JSRPC blog post, since it referred to the Service<Interface> approach.

penalosa (approved):

PR #3724, opened 2025-03-14
Implements decorator for Python event handlers.

dom96:
For now a simple decorator that transforms any `request` argument to the Python `Request` class. In the future we will make it do more things like transform the `env` argument for bindings.

Happy to bikeshed the name, I chose `event` because that's what the Rust workers use.

### Test Plan

```
$ bazel run @workerd//src/workerd/server/tests/python:sdk_development@
```

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -1,6 +1,7 @@
 # This module defines a Workers API for Python. It is similar to the API provided by
 # JS Workers, but with changes and additions to be more idiomatic to the Python
 # programming language.
+import functools
 import http.client
 import json
 from collections.abc import Generator, Iterable, MutableMapping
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/workers.py
```diff
@@ -604,3 +605,24 @@ async def json(self, **kwargs: Any) -> Any:
     async def text(self) -> str:
         self._raise_if_failed()
         return await self.js_object.text()
+
+
+def event(func):
```
hoodmane, line 610: I think `handler` would be a better name? What does our documentation call these things?

src/pyodide/internal/workers.py
```diff
@@ -604,3 +605,24 @@ async def json(self, **kwargs: Any) -> Any:
...
+    """
+    When applied to handlers such as `on_fetch` it will rewrite arguments passed in to native Python
+    types defined in this module. For example, the `request` argument to `on_fetch` gets converted
+    to an instance of the Request class defined in this module.
+    """
+
+    @functools.wraps(func)
```
hoodmane, line 617: I guess `functools.wraps` preserves the signature for introspection, so we can still tell whether the function wanted 3, 2, or 1 argument. In your example, the handler only takes one argument so we can see that this is working.

hoodmane (approved):
Thanks! Looks great, other than the name. I don't think `event` is the right name.

dom96:
Happy to bikeshed a different name, I chose `event` because that's what Rust Workers use https://github.com/cloudflare/workers-rs?tab=readme-ov-file#example-usage

`handler` sounds good to me too, indeed it is what we call these https://developers.cloudflare.com/workers/runtime-apis/handlers/

PR #3725, opened 2025-03-14
ctx and env on RPC targets shouldn't be protected

threepointone:
We mark ctx and env as `protected` on Durable Objects / WorkerEntrypoints / Workerflows so that they don't show in the dropdown for available methods/props to be called when RPCing them. This unfortunately means that when passing an instance of an agent to anything else, ctx and env show as inaccessible on the instance, even though they actually are. This PR removes the protected qualifier, and explicitly adds them to a Provider's excluded fields.


threepointone:
have read the CLA Document and I hereby sign the CLA

PR #3726, opened 2025-03-14
CFSQL-1220: Rename the D1 binding class to the previous name to fix vitest tests

lambrospetrou:
The recent changes to make `D1DatabaseWithSessionAPI` the default class returned for the D1 binding (see https://github.com/cloudflare/workerd/pull/3700) broke some vitest pool tests that use the returned class name and were expecting `D1Database`.

This PR just merges `D1DatabaseWithSessionAPI` with its parent class `D1Database` since now we will roll out the `withSession()` method to everyone, and returns the `D1Database` class.

### DIFF
--- a/src/cloudflare/internal/d1-api.ts
+++ b/src/cloudflare/internal/d1-api.ts
@@ -94,19 +94,6 @@ class D1Database {
     return this.alwaysPrimarySession.exec(query);
   }
 
-  /**
-   * @deprecated
-   */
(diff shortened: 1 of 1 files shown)


danlapid (approved):

lambrospetrou:
This fix is needed also because some things like Rust workers try to statically cast the returned binding class from the `makeBinding()` into `D1Database`, and it was failing before this.

PR #3727, opened 2025-03-14
Put Request.signal ignoreForSubrequests behaviour behind a compat flag

npaun:
When I implemented Request.signal (#3488), we wanted to avoid introducing any surprising new behaviour. As a result, any subrequests created from the incoming request have an `ignoreForSubrequests` flag applied to them. This flag makes it so these outgoing subrequests won't be cancelled even if the incoming request is cancelled. But this is, itself, peculiar behaviour, so this PR adds a compat flag to prevent this behaviour.

- Enable: `abort_signal_not_ignored_for_subrequests`
- Disable/default: `abort_signal_ignored_for_subrequests`

There's a whole lot of negation in this one, so a sanity check on the code would be kindly appreciated.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -732,4 +732,12 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("allow_importable_env");
   # When allowed, `import { env } from 'cloudflare:workers'` will provide access
   # to the per-request environment/bindings.
+
+  requestSignalPassthrough @79 :Bool
+      $compatEnableFlag("request_signal_passthrough")
+      $compatDisableFlag("no_request_signal_passthrough")
+      $compatEnableDate("2025-04-14");
+    # When enabled, the AbortSignal of the incoming request is not passed through to subrequests.
+    # As a result, outgoing subrequests will not be cancelled when the incoming request is.
+
 }
--- a/src/workerd/io/worker-entrypoint.c++
+++ b/src/workerd/io/worker-entrypoint.c++
@@ -7,6 +7,7 @@
 #include <workerd/api/basics.h>
 #include <workerd/api/global-scope.h>
 #include <workerd/api/util.h>
+#include <workerd/io/features.h>
 #include <workerd/io/io-context.h>
 #include <workerd/jsg/jsg.h>
 #include <workerd/util/autogate.h>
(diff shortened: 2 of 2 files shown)


jasnell (approved):

src/workerd/io/compatibility-date.capnp
```diff
@@ -732,4 +732,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("allow_importable_env");
   # When allowed, `import { env } from 'cloudflare:workers'` will provide access
   # to the per-request environment/bindings.
+
+  abortSignalNotIgnoredForSubrequests @79 :Bool
+      $compatEnableFlag("abort_signal_not_ignored_for_subrequests")
+      $compatDisableFlag("abort_signal_ignored_for_subrequests");
```
jasnell, line 738: I'm assuming it is intentional that this has no default on date? If so, might be good to expand the doc comment to indicate so, but good without that also.

npaun: I was copying your "disableImportableEnv" -- I'm still a bit shaky on compat flags.

 I'd say the reasoning is that we won't enable "abortSignalNotIgnoredForSubrequests" because it's potentially not backwards compatible, and we haven't yet seen a clear use case that needs "abortSignalNotIgnoredForSubrequests". Does that make sense to you?

kentonv: I think this is exactly the sort of change where we want a default-enable date. 99.99% of people won't be affected and the new behavior is "better".

Setting a default-enable date doesn't impact existing workers -- they have to actually update their compatibility date to be past the default-enable date before it takes effect.

Of course, we need to make sure all flags with default-on dates are documented here, so people can review what's changing when they update their compat date: https://developers.cloudflare.com/workers/configuration/compatibility-flags/

npaun: How do I decide the default enable date - is it just today, then?

src/workerd/io/compatibility-date.capnp
```diff
@@ -732,4 +732,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("allow_importable_env");
   # When allowed, `import { env } from 'cloudflare:workers'` will provide access
   # to the per-request environment/bindings.
+
+  abortSignalNotIgnoredForSubrequests @79 :Bool
+      $compatEnableFlag("abort_signal_not_ignored_for_subrequests")
```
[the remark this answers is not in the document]
kentonv: `abort_signal_ignored_for_subrequests` doesn't really capture the fact that this only affects _passthrough_ of the signal from the incoming request to the outgoing.

PR #3728, opened 2025-03-14
Break actor on errors that rollback transactions

shrima-cf:
We should break output gates and reset actors in the event that we catch the following exceptions because they implicitly roll back any transaction:

[SQLITE_FULL](https://www.sqlite.org/rescode.html#full): database or disk full
[SQLITE_IOERR](https://www.sqlite.org/rescode.html#ioerr): disk I/O error
[SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy): database in use by another process
[SQLITE_NOMEM](https://www.sqlite.org/rescode.html#nomem): out of memory
https://www.sqlite.org/lang_transaction.html#response_to_errors_within_a_transaction

### DIFF
--- a/src/workerd/io/actor-sqlite.c++
+++ b/src/workerd/io/actor-sqlite.c++
@@ -197,11 +197,11 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
   }
 }
 
-void ActorSqlite::onCriticalError(kj::StringPtr message) {
-  auto exception =
-      kj::Exception(kj::Exception::Type::FAILED, __FILE__, __LINE__, kj::heapString(message));
-  broken.emplace(kj::mv(exception));
-  requireNotBroken();
+void ActorSqlite::onCriticalError(kj::Exception exception) {
+  // If we have already experienced a terminal exception, no need to replace it
+  if (broken == kj::none) {
+    broken.emplace(kj::mv(exception));
+  }
 }
 
 void ActorSqlite::onWrite() {
--- a/src/workerd/io/actor-sqlite.c++
+++ b/src/workerd/io/actor-sqlite.c++
@@ -5,6 +5,7 @@
 #include "actor-sqlite.h"
 
 #include "io-gate.h"
+#include "kj/function.h"
 
 #include <workerd/jsg/exception.h>
 #include <workerd/util/sentry.h>
@@ -37,6 +38,7 @@ ActorSqlite::ActorSqlite(kj::Own<SqliteDatabase> dbParam,
       metadata(*db),
       commitTasks(*this) {
   db->onWrite(KJ_BIND_METHOD(*this, onWrite));
+  db->onCriticalError(KJ_BIND_METHOD(*this, onCriticalError));
   lastConfirmedAlarmDbState = metadata.getAlarm();
 
   // Because we preserve an invariant that scheduled alarms are always at or earlier than
@@ -192,6 +194,21 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
     alarmDirty = p->alarmDirty;
   } else {
     alarmDirty = false;
+  }
+}
+
+void ActorSqlite::onCriticalError(int sqliteErrorCode, kj::StringPtr message) {
+  switch (sqliteErrorCode) {
+    case SQLITE_FULL:
+      JSG_FAIL_REQUIRE(Error, "Database or disk full: ", message);
+    case SQLITE_IOERR:
+      JSG_FAIL_REQUIRE(Error, "Disk I/O error: ", message);
+    case SQLITE_BUSY:
+      JSG_FAIL_REQUIRE(Error, "Database in use by another process: ", message);
+    case SQLITE_NOMEM:
+      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
+    case SQLITE_INTERRUPT:
+      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
   }
 }
 
--- a/src/workerd/io/actor-sqlite.c++
+++ b/src/workerd/io/actor-sqlite.c++
@@ -197,19 +197,11 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
   }
 }
 
-void ActorSqlite::onCriticalError(int sqliteErrorCode, kj::StringPtr message) {
-  switch (sqliteErrorCode) {
-    case SQLITE_FULL:
-      JSG_FAIL_REQUIRE(Error, "Database or disk full: ", message);
-    case SQLITE_IOERR:
-      JSG_FAIL_REQUIRE(Error, "Disk I/O error: ", message);
-    case SQLITE_BUSY:
-      JSG_FAIL_REQUIRE(Error, "Database in use by another process: ", message);
-    case SQLITE_NOMEM:
-      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
-    case SQLITE_INTERRUPT:
-      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
-  }
+void ActorSqlite::onCriticalError(kj::StringPtr message) {
+  auto exception =
+      kj::Exception(kj::Exception::Type::FAILED, __FILE__, __LINE__, kj::heapString(message));
+  broken.emplace(kj::mv(exception));
+  requireNotBroken();
 }
 
 void ActorSqlite::onWrite() {
--- a/src/workerd/util/sqlite-test.c++
+++ b/src/workerd/util/sqlite-test.c++
@@ -1487,13 +1487,11 @@ KJ_TEST("SQLite critical error handling for SQLITE_IOERR") {
 
   // Create a tracker to verify our callback is called
   bool criticalErrorCallbackCalled = false;
-  int capturedErrorCode = 0;
   kj::String capturedErrorMessage = nullptr;
 
   // Register a critical error callback
-  db.onCriticalError([&](int sqliteErrorCode, kj::StringPtr message) {
+  db.onCriticalError([&](kj::StringPtr message) {
     criticalErrorCallbackCalled = true;
-    capturedErrorCode = sqliteErrorCode;
     capturedErrorMessage = kj::str(message);
   });
 
@@ -1513,12 +1511,10 @@ KJ_TEST("SQLite critical error handling for SQLITE_IOERR") {
   )")));
 
   KJ_EXPECT(criticalErrorCallbackCalled);
-  KJ_EXPECT(capturedErrorCode == SQLITE_IOERR);
   KJ_EXPECT(capturedErrorMessage.startsWith("disk I/O error"));
 }
 
-void testCriticalError(int expectedErrorCode,
-    const char* expectedErrorMessage,
+void testCriticalError(const char* expectedErrorMessage,
     kj::Function<void(SqliteDatabase&, SqliteDatabase::Vfs& vfs)> triggerErrorFn) {
   auto dir = kj::newInMemoryDirectory(kj::nullClock());
   SqliteDatabase::Vfs vfs(*dir);
@@ -1526,26 +1522,22 @@ void testCriticalError(int expectedErrorCode,
 
   // Create a tracker to verify our callback is called
   bool criticalErrorCallbackCalled = false;
-  int capturedErrorCode = 0;
   kj::String capturedErrorMessage = nullptr;
 
   // Register a critical error callback
-  db.onCriticalError([&](int sqliteErrorCode, kj::StringPtr message) {
+  db.onCriticalError([&](kj::StringPtr message) {
     criticalErrorCallbackCalled = true;
-    capturedErrorCode = sqliteErrorCode;
     capturedErrorMessage = kj::str(message);
   });
 
   KJ_EXPECT_THROW_MESSAGE(expectedErrorMessage, triggerErrorFn(db, vfs));
 
   KJ_EXPECT(criticalErrorCallbackCalled);
-  KJ_EXPECT(capturedErrorCode == expectedErrorCode);
   KJ_EXPECT(capturedErrorMessage.startsWith(expectedErrorMessage));
 }
 
 KJ_TEST("SQLite critical error handling for SQLITE_BUSY") {
-  testCriticalError(
-      SQLITE_BUSY, "database is locked", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
+  testCriticalError("database is locked", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
     // Create a second database connection to hold an exclusive lock
     SqliteDatabase lockDb(vfs, kj::Path({"foo"}), kj::WriteMode::MODIFY);
 
@@ -1563,15 +1555,13 @@ KJ_TEST("SQLite critical error handling for SQLITE_BUSY") {
 }
 
 KJ_TEST("SQLite critical error handling for SQLITE_FULL") {
-  testCriticalError(
-      SQLITE_FULL, "database or disk is full", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
+  testCriticalError("database or disk is full", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
     // Set up a database with limited size
     db.run("PRAGMA max_page_count = 10");
     db.run("CREATE TABLE IF NOT EXISTS test_full (id INTEGER PRIMARY KEY, data BLOB)");
 
     // Create a large blob to quickly fill the database
-    auto largeData = kj::heapArray<byte>(100000);  // 100KB
-    memset(largeData.begin(), 'X', largeData.size());
+    auto largeData = kj::heapArray<byte>(100000, 'X');  // 100KB
 
     // This should eventually trigger SQLITE_FULL
     db.run(SqliteDatabase::TRUSTED, "INSERT INTO test_full VALUES (?, ?)", 1, largeData.asPtr());
@@ -1579,16 +1569,14 @@ KJ_TEST("SQLite critical error handling for SQLITE_FULL") {
 }
 
 KJ_TEST("SQLite critical error handling for SQLITE_NOMEM") {
-  testCriticalError(
-      SQLITE_NOMEM, "out of memory", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
+  testCriticalError("out of memory", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
     db.run("CREATE TABLE test_nomem (id INTEGER PRIMARY KEY, data BLOB)");
 
     // Set SQLite's memory limit very low to trigger SQLITE_NOMEM
     db.run("PRAGMA hard_heap_limit=8192");  // 8KB limit
 
     // Create data that will exceed the memory limit
-    auto largeData = kj::heapArray<byte>(50000);  // 50KB
-    memset(largeData.begin(), 'X', largeData.size());
+    auto largeData = kj::heapArray<byte>(50000, 'X');  // 50KB
 
     // This operation should trigger SQLITE_NOMEM when SQLite tries to allocate
     // memory for the large parameter
--- a/src/workerd/util/sqlite-test.c++
+++ b/src/workerd/util/sqlite-test.c++
@@ -2,9 +2,12 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
+#include "kj/mutex.h"
 #include "sqlite.h"
+#include "workerd/io/io-gate.h"
 
 #include <fcntl.h>
+#include <sqlite3.h>
 
 #include <kj/refcount.h>
 #include <kj/test.h>
@@ -14,6 +17,7 @@
 #include <cerrno>
 #include <cstdint>
 #include <cstdlib>
+#include <thread>
 
 #if _WIN32
 #include <io.h>
@@ -1476,10 +1480,22 @@ class ErrorInjectableDirectory final: public kj::Directory, public kj::AtomicRef
   }
 };
 
-KJ_TEST("I/O exceptions pass through SQLite") {
+KJ_TEST("SQLite critical error handling for SQLITE_IOERR") {
   auto dir = kj::atomicRefcounted<ErrorInjectableDirectory>();
   SqliteDatabase::Vfs vfs(*dir);
   SqliteDatabase db(vfs, kj::Path({"db"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+  // Create a tracker to verify our callback is called
+  bool criticalErrorCallbackCalled = false;
+  int capturedErrorCode = 0;
+  kj::String capturedErrorMessage = nullptr;
+
+  // Register a critical error callback
+  db.onCriticalError([&](int sqliteErrorCode, kj::StringPtr message) {
+    criticalErrorCallbackCalled = true;
+    capturedErrorCode = sqliteErrorCode;
+    capturedErrorMessage = kj::str(message);
+  });
 
   db.run(SqliteDatabase::TRUSTED, kj::str(R"(
     CREATE TABLE IF NOT EXISTS things (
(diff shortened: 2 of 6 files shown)


shrima-cf:
@kentonv For some reason, I don't have permissions to edit the reviewers list to add you explicitly, would be great if you could review this PR

src/workerd/util/sqlite.c++
```diff
@@ -174,16 +181,24 @@ class SqliteCallScope {
     /* SQLITE_MISUSE doesn't put error info on the database object, so check it separately */      \
     KJ_ASSERT(_ec != SQLITE_MISUSE, "SQLite misused: " #code, ##__VA_ARGS__);                      \
     if (_ec == SQLITE_IOERR) sqliteCallScope.rethrowVfsError();                                    \
+    if (isCriticalError(code)) {                                                                   \
+      regulator.onCriticalError(_ec, dbErrorMessage(code, db));                                    \
```
kentonv, line 185: We need to actually verify whether a transaction was rolled back.
- There might not have been a transaction open. In this case `db.savepoints` will be empty.
- If there was a transaction open, SQLite did not necessarily roll it back. [Per the docs](https://www.sqlite.org/lang_transaction.html#response_to_errors_within_a_transaction), we need to call `sqlite3_get_autocommit()` to find out.

We don't want to do all this in a macro, of course, so we probably want a method like `db.checkForCriticalError(code)`.

kentonv, line 185: We need to actually verify whether a transaction was rolled back.
- There might not have been a transaction open. In this case `db.savepoints` will be empty.
- If there was a transaction open, SQLite did not necessarily roll it back. [Per the docs](https://www.sqlite.org/lang_transaction.html#response_to_errors_within_a_transaction), we need to call `sqlite3_get_autocommit()` to find out.

We don't want to do all this in a macro, of course, so we probably want a method like `db.checkForCriticalError(code)`.

kentonv, line 185: We need to actually verify whether a transaction was rolled back.
- There might not have been a transaction open. In this case `db.savepoints` will be empty.
- If there was a transaction open, SQLite did not necessarily roll it back. [Per the docs](https://www.sqlite.org/lang_transaction.html#response_to_errors_within_a_transaction), we need to call `sqlite3_get_autocommit()` to find out.

We don't want to do all this in a macro, of course, so we probably want a method like `db.checkForCriticalError(code)`.

src/workerd/util/sqlite.c++
```diff
@@ -174,16 +181,24 @@ class SqliteCallScope {
     /* SQLITE_MISUSE doesn't put error info on the database object, so check it separately */      \
     KJ_ASSERT(_ec != SQLITE_MISUSE, "SQLite misused: " #code, ##__VA_ARGS__);                      \
     if (_ec == SQLITE_IOERR) sqliteCallScope.rethrowVfsError();                                    \
+    if (isCriticalError(code)) {                                                                   \
```
kentonv, line 184: I would move this check into `SQLITE_REQUIRE` rather than reproduce it in both variants of `SQLITE_CALL`.

src/workerd/util/sqlite.c++
```diff
@@ -174,16 +181,24 @@ class SqliteCallScope {
...
+      regulator.onCriticalError(_ec, dbErrorMessage(code, db));                                    \
+      KJ_FAIL_REQUIRE("SQLite faield with critical error", dbErrorMessage(_ec, db), ##__VA_ARGS__);\
```
kentonv, line 186: No need to throw a unique message here, just let the flow continue and throw the exception it would have thrown otherwise.

src/workerd/io/actor-sqlite.c++
```diff
@@ -721,6 +721,25 @@ void ActorSqlite::TxnCommitRegulator::onError(
   // as we don't know for sure that the problem is the application's fault.
 }
 
+void ActorSqlite::TxnCommitRegulator::onCriticalError(
```
kentonv, line 724: This regulator is only used when committing a transaction. But the error could happen on any statement within the transaction.

I think this shouldn't be a regulator callback. Instead we need something like `SqliteDatabase::onCriticalError()` which lets you specify a callback to call in this case, similar to how `onWrite()` works.

src/workerd/util/sqlite.c++
```diff
@@ -174,16 +181,24 @@ class SqliteCallScope {
...
-    KJ_ASSERT(error != SQLITE_MISUSE, "SQLite misused: " code, ##__VA_ARGS__);                     \
-    if (error == SQLITE_IOERR) sqliteCallScope.rethrowVfsError();                                  \
-    SQLITE_REQUIRE(error == SQLITE_OK, error, dbErrorMessage(error, db), ##__VA_ARGS__);           \
+#define SQLITE_CALL_FAILED(code, error, ...)                                                         \
+  do {                                                                                               \
+    KJ_ASSERT(error != SQLITE_MISUSE, "SQLite misused: " code, ##__VA_ARGS__);                       \
+    if (error == SQLITE_IOERR) sqliteCallScope.rethrowVfsError();                                    \
+    if (isCriticalError(error)) {                                                                    \
+      regulator.onCriticalError(error, dbErrorMessage(error, db));                                   \
+      KJ_FAIL_REQUIRE("SQLite faield with critical error", dbErrorMessage(error, db), ##__VA_ARGS__);\
```
MellowYarker, line 199: `SQLite faield with` -> `SQLite failed with`

src/workerd/util/sqlite-test.c++
```diff
@@ -1498,5 +1501,156 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+                        SqliteDatabase::Vfs& vfs,
+                        const SqliteDatabase::Regulator&)> triggerErrorFn) {
+  kj::EventLoop loop;
+  kj::WaitScope ws(loop);
+
+  auto dir = kj::newInMemoryDirectory(kj::nullClock());
+  SqliteDatabase::Vfs vfs(*dir);
+  SqliteDatabase db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+  OutputGate gate;
```
kentonv, line 1533: I don't think incorporating `OutputGate` into this test accomplishes much.

You show in this test that when you use `gate.lockWhile()` to wrap some code that causes a critical error to be thrown, then it breaks the output gate. But of course that's true -- if _any_ exception is thrown inside `gate.lockWhile()`, it breaks the gate.

This doesn't actually demonstrate that output gate has been hooked up correctly to critical exceptions in the actual runtime. Indeed, as you have it implemented right now, critical exceptions don't necessarily break the output gate. They break it if the exception is thrown specifically from the `COMMIT` statement committing an implicit transaction -- but it has always been the case that exceptions thrown from that statement break the output gate.

I would recommend removing the `OutputGate` dependency from this test as we can't really show anything interesting about it here. We need a higher-level test that actually tests that the runtime has integrated these correctly, probably by running JavaScript code that triggers a critical error.

src/workerd/util/sqlite-test.c++
```diff
@@ -1492,10 +1508,92 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+    // Create a large blob to quickly fill the database
+    auto largeData = kj::heapArray<byte>(100000);  // 100KB
+    memset(largeData.begin(), 'X', largeData.size());
+
+    // This should eventually trigger SQLITE_FULL
+    db.run(SqliteDatabase::TRUSTED, "INSERT INTO test_full VALUES (?, ?)", 1, largeData.asPtr());
+  });
+}
+
+KJ_TEST("SQLite critical error handling for SQLITE_NOMEM") {
```
shrima-cf, line 1581: I tried hard to come up with a scenario to throw SQLITE_INTERRUPT but was not able to do so. Suggestions appreciated on how to throw one

kentonv: You would need to start a long-running query and then call `sqlite3_interrupt()` from a separate thread.

kentonv: Hmm, but SQLITE_INTERRUPT is not actually documented as being a critical error?

src/workerd/io/actor-sqlite.c++
```diff
@@ -195,6 +197,21 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
...
+    case SQLITE_IOERR:
+      JSG_FAIL_REQUIRE(Error, "Disk I/O error: ", message);
+    case SQLITE_BUSY:
+      JSG_FAIL_REQUIRE(Error, "Database in use by another process: ", message);
+    case SQLITE_NOMEM:
+      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
+    case SQLITE_INTERRUPT:
+      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
+  }
+}
```
justin-mp, line 213: Probably good to put a KJ_UNREACHABLE down here to make sure that the code never silently falls off the end of the function.

src/workerd/util/sqlite.c++
```diff
@@ -173,7 +174,6 @@ class SqliteCallScope {
     int _ec = code;                                                                                \
     /* SQLITE_MISUSE doesn't put error info on the database object, so check it separately */      \
     KJ_ASSERT(_ec != SQLITE_MISUSE, "SQLite misused: " #code, ##__VA_ARGS__);                      \
-    if (_ec == SQLITE_IOERR) sqliteCallScope.rethrowVfsError();                                    \
```
justin-mp, line 176: Removing the calls to `rethrowVfsError` breaks the VFS exception passthrough.  That seems like a bad idea to me!

It seems we're doing this because `handleCriticalError` wants to throw its own JSG I/O error.  Is that the correct thing to do?  Previously, things like I/O error would get turned into internal errors, but now that they're JSG errors, I think they'll get passed directly to the user and that's probably more confusing than getting a straight "internal error".

I'd prefer to keep the old behavior, but if that's not possible, at the very least we should log the VFS error so that we can find out why we got an I/O error.

src/workerd/util/sqlite-test.c++
```diff
@@ -1492,10 +1508,92 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+KJ_TEST("SQLite critical error handling for SQLITE_FULL") {
+  testCriticalError(
+      SQLITE_FULL, "database or disk is full", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
+    // Set up a database with limited size
+    db.run("PRAGMA max_page_count = 10");
+    db.run("CREATE TABLE IF NOT EXISTS test_full (id INTEGER PRIMARY KEY, data BLOB)");
+
+    // Create a large blob to quickly fill the database
+    auto largeData = kj::heapArray<byte>(100000);  // 100KB
+    memset(largeData.begin(), 'X', largeData.size());
```
justin-mp, line 1574: nit: there's also `kj::heapArray<T>(size, val)` that will initialize all of the values to `val`.  Or if you want to do it in two steps, you can do `largeData.fill('X')`.

Applies below too.

src/workerd/io/actor-sqlite.c++
```diff
@@ -195,6 +197,21 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
   }
 }
 
+void ActorSqlite::onCriticalError(int sqliteErrorCode, kj::StringPtr message) {
```
kentonv, line 200: This method needs to break the output gate. Throwing an exception does not break the output gate. (It actually doesn't need to throw an exception, if it just returns then SqliteDatabase already takes care of throwing an exception.)

kentonv: I think if you simply set `broken` (a member of `ActorSqlite`) to an appropriate exception, that accomplishes what we need.

In `ActorSqlite::onWrite()`, you see it says `commitTasks.add(outputGate.lockWhile(...))`, and the callback code there begins with `requireNotBroken()` before trying to commit the transaction. So if `broken` has been assigned, than code will fail, breaking the output gate, and the implicit transaction is never committed. That's what we want.

src/workerd/io/actor-sqlite.c++
```diff
@@ -195,6 +197,21 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
...
+  switch (sqliteErrorCode) {
```
kentonv, line 201: You don't really need to inspect the error code to generate a message. `message` already describes the error.

In fact, maybe there's no need for this callback to receive the error code at all.

src/workerd/io/actor-sqlite.c++
kentonv: At least, that is correct when `currentTxn` is `ImplicitTxn`.

When it's `ExplicitTxn`, we might have to do something a little different. In that case there's no existing `lockWhile()` to interrupt. But we also don't necessarily have to break the output gate in this case anyway, because the application doesn't _expect_ that an explicit transaction is committed until the app actually tries to commit it...

_In theory_, we could just make sure that all further actions on the explicit transaction fail, and the commit fails, and then the transaction gets rolled back and everything is fine.

_However_ this may be too dangerous to actually try to do, because sqlite is in a weird state where it has actually rolled back our transaction, so it would be easy for us to mess up and accidentally execute a statement that the application _expected_ was going to be part of the explicit transaction, but then it ends up treated as auto-commit. Hmm.

So maybe the safest thing to do in this case is in fact to mark ourselves broken anyway. Perhaps we should just do `outputGate.lockWhile(exception)` to force the gate to immediately break, even (in addition to assigning `broken`).

I guess that's also the right thing to do for `NoTxn` case (which should never happen anyway).

src/workerd/util/sqlite.c++
```diff
@@ -533,6 +532,22 @@ void SqliteDatabase::notifyWrite() {
   }
 }
 
+void SqliteDatabase::handleCriticalError(kj::Maybe<int> errorCode, kj::StringPtr errorMessage) {
+  KJ_IF_SOME(code, errorCode) {
+    if (code == SQLITE_FULL || code == SQLITE_IOERR || code == SQLITE_BUSY ||
+        code == SQLITE_NOMEM || code == SQLITE_INTERRUPT) {
+
+      sqlite3* db = &KJ_ASSERT_NONNULL(maybeDb, "previous reset() failed");
+      // The transaction was rolledback, re-enabling the auto commit mode, so we should fail
```
kentonv, line 541: Note that we might not have been in a transaction to start with, in which case there's no need to treat this as a critical error.

You can check if we were in a transaction before the error by checking `inTransaction || !savepoints.empty()`.

We should probably then do `inTransaction = false; savepoints.clear();` here since we need these to track what sqlite thinks the state is... although this feels precarious. I kind of wonder whether `SqliteDatabase` itself should refuse to execute any more statements after a critical error... not sure.

src/workerd/util/sqlite-test.c++
shrima-cf: I tried doing that, but all those attempts resulted in a SQLITE_ERROR for me.
SQLITE_INTERRUPT is not mentioned as one of the errors here - https://www.sqlite.org/lang_transaction.html#response_to_errors_within_a_transaction 
But is is mentioned here - https://www.sqlite.org/c3ref/get_autocommit.html

src/workerd/io/actor-sqlite.c++
```diff
@@ -195,6 +197,21 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
...
+    case SQLITE_FULL:
+      JSG_FAIL_REQUIRE(Error, "Database or disk full: ", message);
+    case SQLITE_IOERR:
+      JSG_FAIL_REQUIRE(Error, "Disk I/O error: ", message);
+    case SQLITE_BUSY:
+      JSG_FAIL_REQUIRE(Error, "Database in use by another process: ", message);
+    case SQLITE_NOMEM:
+      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
+    case SQLITE_INTERRUPT:
+      JSG_FAIL_REQUIRE(Error, "Out of memory: ", message);
```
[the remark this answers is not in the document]
shrima-cf: Ah, copy pasta error, will update this

src/workerd/util/sqlite.c++
shrima-cf: I was not very sure of this change in behavior either, but decided to go with the change since the IOERR should not result in breaking the output gate as the transaction automatically got rolled back.

src/workerd/io/actor-sqlite.c++
shrima-cf: Hmm, in that case, would it be better to set the broken exception onCriticalError and then call requireNotBroken on commit for Expliticit and Implicit txn?

src/workerd/io/actor-sqlite.c++
shrima-cf: Yeah, if I am going to set the broken exception, I can just use the message and do it

src/workerd/util/sqlite.c++
shrima-cf: I had the check for `inTransaction || !savepoints.empty()` , but got rid of it as it felt checking if the db is in auto_commit mode was sufficient.
I was also not able to get into the code path of `if(inTransaction || !savepoints.empty())` through my test - any idea why this could be the case with the test I have?

src/workerd/util/sqlite-test.c++
shrima-cf: Given that including SQLITE_INTERRUPT is causing a change in behavior (pointed out by Justin below), I am leaning towards removing it from the list of critical errors here, but still not sure about why there is inconsistency in the list of errors in the two links above

src/workerd/util/sqlite.c++
shrima-cf: I am removing the SQLITE_INTERRUPT from the list of critical errors (there is some confusion if it is actually a critical error or not, see discussion above). With that, this change will go away

src/workerd/io/actor-sqlite.c++
shrima-cf: This code has changed, this is no longer needed - https://github.com/cloudflare/workerd/pull/3728/commits/567cf308fdd05af33e3265118445c24452ff84e9

src/workerd/util/sqlite-test.c++
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3728/commits/567cf308fdd05af33e3265118445c24452ff84e9

src/workerd/io/actor-sqlite.c++
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3728/commits/567cf308fdd05af33e3265118445c24452ff84e9

src/workerd/util/sqlite.c++
shrima-cf: Discussed with Justin, will use the error message from the vfs error that we got so we don't get a blanket disk i/o error here.

kentonv: Please be sure to reuse not just the error message, but the entire Exception object. It has a stack trace, and we need that stack trace for debugging.

src/workerd/util/sqlite.h
```diff
@@ -178,6 +178,12 @@ class SqliteDatabase {
     onWriteCallback = kj::mv(callback);
   }
 
+  void handleCriticalError(kj::Maybe<int> errorCode, kj::StringPtr errorMessage);
+
+  void onCriticalError(kj::Function<void(kj::StringPtr message)> callback) {
```
kentonv, line 183: We may want to pass a whole `kj::Exception` to the callback instead of just a string, so that it can capture a stack trace, especially in the case of passthrough I/O errors from the VFS layer.

src/workerd/io/actor-sqlite.c++
```diff
@@ -195,6 +197,13 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
   }
 }
 
+void ActorSqlite::onCriticalError(kj::StringPtr message) {
+  auto exception =
+      kj::Exception(kj::Exception::Type::FAILED, __FILE__, __LINE__, kj::heapString(message));
+  broken.emplace(kj::mv(exception));
+  requireNotBroken();
```
kentonv, line 204: Again, I don't think we need to throw here. The caller is already going to throw if we return normally. So this line can be removed.

src/workerd/io/actor-sqlite.c++
```diff
@@ -195,6 +197,13 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
   }
 }
 
+void ActorSqlite::onCriticalError(kj::StringPtr message) {
+  auto exception =
+      kj::Exception(kj::Exception::Type::FAILED, __FILE__, __LINE__, kj::heapString(message));
+  broken.emplace(kj::mv(exception));
```
kentonv, line 203: If `broken` is already non-null we probably should not rewrite it here.

src/workerd/util/sqlite.c++
kentonv: `inTransaction || !savepoints.empty()` determines if we _were_ in a transaction before the error.

`sqlite3_get_autocommit()` tells us if we _are_ in a transaction now.

This is how we tell if the critical error caused a rollback -- we have to compare these two things.

If we were not in a transaction before, then we don't need to consider the error critical, because it only rolled back the immediate statement that caused it, which is what we expect from any error. The critical problem is if past statements were rolled back since then the database state may no longer be consistent with the expectations of the application.

src/workerd/util/sqlite-test.c++
kentonv: I think it makes sense that this is in the list. It's weird that you can't trigger it. But there's no harm in keeping it in the list.

In practice we only ever interrupt sqlite when the CPU or hard memory limit is hit, in which case we're blowing up the actor anyway.

src/workerd/util/sqlite.h
shrima-cf: Agreed! That's what I discussed with Justin as well, changing this

src/workerd/util/sqlite-test.c++
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3728/commits/2a8ca60281c5bbb516ab5f42e3e290296eb9d73e#diff-3009a3da17d76a94673aa828ce6c364fc24e34d889b4995adfc249a0e2a3c754

src/workerd/util/sqlite.c++
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3728/commits/2a8ca60281c5bbb516ab5f42e3e290296eb9d73e#diff-3009a3da17d76a94673aa828ce6c364fc24e34d889b4995adfc249a0e2a3c754

src/workerd/util/sqlite.h
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3728/commits/2a8ca60281c5bbb516ab5f42e3e290296eb9d73e#diff-3009a3da17d76a94673aa828ce6c364fc24e34d889b4995adfc249a0e2a3c754

src/workerd/io/actor-sqlite.c++
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3728/commits/2a8ca60281c5bbb516ab5f42e3e290296eb9d73e#diff-e0f2e99f58d22f6880187a9de379c249811338370351ce2bea41b905865193d9

src/workerd/io/actor-sqlite.c++
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3728/commits/2a8ca60281c5bbb516ab5f42e3e290296eb9d73e#diff-e0f2e99f58d22f6880187a9de379c249811338370351ce2bea41b905865193d9

src/workerd/util/sqlite.c++
shrima-cf: I fixed the check here - https://github.com/cloudflare/workerd/pull/3728/commits/9b726ce5659af631ea91e12059a065636926423e

But two tests are failing: 
KJ_TEST("SQLite critical error handling for SQLITE_BUSY")
KJ_TEST("SQLite critical error handling for SQLITE_NOMEM")

Both are failing because the commit is not auto-rolledback when run inside a transaction. Appreciate help to debug this

shrima-cf:
I am going to merge this into my personal branch in cf repo (shrima/STOR-3841) and will open a new PR from there.
This is the branch from CF repo - https://github.com/cloudflare/workerd/tree/shrima/STOR-3841

Will keep this PR open as well since the comments need to be present.

I had to do this update since something changed in the the main cf repo's build generation logic, that I am not able to reproduce here. There is no way to sync those two repos without discarding these commits either.

shrima-cf:
Pull request moved - https://github.com/cloudflare/workerd/pull/3832

PR #3729, opened 2025-03-14
fix types eslint config and reenable more strict checks

anonrig:
cc @penalosa Appreciate if you could review this more in depth.

cc @mikea @fhanau Would you mind reviewing the bazel changes in this pull-request?


npaun (approved):
LGTM and will make things more convenient

types/BUILD.bazel
```diff
@@ -12,8 +12,19 @@ wd_ts_project(
         ],
         exclude = ["src/worker/**/*"],
     ),
+    data = [
```
fhanau, line 15: Maybe we can just have the eslint_test use the wd_ts_project deps so that we don't have to add a data attribute/duplicate the list of dependencies here?

build/wd_ts_project.bzl
```diff
@@ -1,15 +1,41 @@
...
-        tsconfig = "//types:tsconfig.json",
+        tsconfig = tsconfig,
         allow_js = True,
         composite = True,
         source_map = True,
         testonly = testonly,
     )
+
+    if eslintrc_json:
+        eslint_bin.eslint_test(
```
fhanau, line 21: We'd want this to consolidate this with the other places where we call eslint_test() to factor them all out into one function, but that's not needed for now

fhanau:
LGTM otherwise

fhanau (approved):

types/BUILD.bazel
anonrig: I feel like I'm missing something but, this doesn't work?

We get bunch of eslint errors because the types are not there...

```diff
diff --git a/build/wd_ts_project.bzl b/build/wd_ts_project.bzl
index e4ccb70d..2e0b518d 100644
--- a/build/wd_ts_project.bzl
+++ b/build/wd_ts_project.bzl
@@ -1,7 +1,7 @@
 load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
 load("@npm//:eslint/package_json.bzl", eslint_bin = "bin")
 
-def wd_ts_project(name, srcs, deps, data = [], eslintrc_json = None, testonly = False):
+def wd_ts_project(name, srcs, deps, eslintrc_json = None, testonly = False):
     """Bazel rule for a workerd TypeScript project, setting common options"""
 
     tsconfig = "//types:tsconfig.json"
@@ -27,7 +27,7 @@ def wd_ts_project(name, srcs, deps, data = [], eslintrc_json = None, testonly =
                 "-f stylish",
                 "--report-unused-disable-directives",
             ] + ["$(location " + src + ")" for src in srcs],
-            data = srcs + data + [
+            data = srcs + [
                 eslintrc_json,
                 tsconfig,
                 "//tools:base-eslint",
diff --git a/types/BUILD.bazel b/types/BUILD.bazel
index c4babc9f..6b36d8c7 100644
--- a/types/BUILD.bazel
+++ b/types/BUILD.bazel
@@ -12,15 +12,6 @@ wd_ts_project(
         ],
         exclude = ["src/worker/**/*"],
     ),
-    data = [
-        "//:node_modules/@types/debug",
-        "//:node_modules/@types/node",
-        "//:node_modules/@workerd/jsg",
-        "//:node_modules/capnp-ts",
-        "//:node_modules/esbuild",
-        "//:node_modules/prettier",
-        "//:node_modules/typescript",
-    ],
     eslintrc_json = "eslint.config.mjs",
     deps = [
         "//:node_modules/@types/debug",

```

anonrig: You're absolutely right.

build/wd_ts_project.bzl
anonrig: Agreed. I can follow up in a different pull-request.

anonrig: I went ahead and implemented this on a new commit. Would you mind taking a look @fhanau?

fhanau: Thank you for doing so!

Couple smaller things:
- Load statements in 1st line of eslint_test.bzl are unused, can be dropped (if you use the VSCode Bazel extension it will warn about stuff like this automatically)
- `//:prettierrc` is no longer being passed in the data argument – I assume it was unused before?
- Ideally we'd also fold the `eslint_bin.eslint_test` target in `src/wpt/BUILD.bazel` into this (haven't checked if there'd be additional complexity with it)

anonrig: > //:prettierrc is no longer being passed in the data argument – I assume it was unused before?

Yes prettier is not used at all.

> Ideally we'd also fold the eslint_bin.eslint_test target in src/wpt/BUILD.bazel into this (haven't checked if there'd be additional complexity with it)

Yes, I did another take. Nice catch.

fhanau: LGTM.

types/.prettierrc
penalosa, line 1: Why are we removing this prettier config>

types/src/generator/index.ts
```diff
@@ -46,42 +45,44 @@ function collectIncluded(map: StructureMap, root?: string): Set<string> {
...
+        { visitFunction(member.getMethod()); return; }
       case Member_Which.PROPERTY:
-        return visitType(member.getProperty().getType());
+        { visitType(member.getProperty().getType()); return; }
       case Member_Which.NESTED:
-        return visitStructure(member.getNested().getStructure());
+        { visitStructure(member.getNested().getStructure()); return; }
       case Member_Which.CONSTRUCTOR:
-        return member.getConstructor().getArgs().forEach(visitType);
+        { member.getConstructor().getArgs().forEach(visitType); return; }
```
penalosa, line 85: In my mind this is less clear, but happy to go with this for the sake of consistency

penalosa (approved):
This looks good from a workers-types perspective

types/src/generator/index.ts
anonrig: This function is returning a void. I guess .forEach() doesn't return void. This was done automatically by the formatter/linter.

types/.prettierrc
anonrig: It wasn't used at all. tools/cross/format.py uses the prettier config inside the root directory.

mikea (approved):

PR #3732, opened 2025-03-17
update ruff to 0.11

anonrig:
Update ruff to latest version

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,6 +26,9 @@ jobs:
       - uses: actions/checkout@v4
         with:
           show-progress: false
+      - name: Configure git hooks
+        # Configure git to quell an irrelevant warning for runners (they never commit / push).
+        run: git config core.hooksPath githooks
       - name: Lint
         run: |
           bazel info output_base # Ensure bazel is initialized before procedding
--- a/build/deps/gen/dep_ruff_darwin_arm64.bzl
+++ b/build/deps/gen/dep_ruff_darwin_arm64.bzl
@@ -2,10 +2,10 @@
 
 load("@//:build/http.bzl", "http_archive")
 
-TAG_NAME = "0.8.3"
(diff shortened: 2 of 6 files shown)


npaun (approved):

dom96 (approved):

anonrig:
> I guess we should apply the changes from the lint job before reviewing. From reading them out of the CI job they look like improvements to me.

Unfortunately, running the formatter script doesn't fix these issues...

anonrig:
OK I think I got it working. I had to somehow update rules_python because it was returning 404 on the existing URL we were using.

hoodmane:
> Unfortunately, running the formatter script doesn't fix these issues...

The problem is here:
```
    if not tool_path.exists():
        build_target = (
            f"@{tool_target}//:file" if is_archive else f"@{tool_target}//file"
        )
        subprocess.run(["bazel", "build", build_target])
```
If you ran it before and didn't explicitly `bazel clean` it will keep using the old version.

fhanau:
> OK I think I got it working. I had to somehow update rules_python because it was returning 404 on the existing URL we were using.

Those 404s tend to be temporary (just need to re-run the affected build).
I ran into issues updating rules_python before, looks like CI is having problems with it (the downstream build doesn't need to use rules_python's pip_parse, so we don't have that issue and already use the newest version there). Consider keeping it at the old version for now to unblock CI.

fhanau:
> > Unfortunately, running the formatter script doesn't fix these issues...
> 
> The problem is here:
> 
> ```
>     if not tool_path.exists():
>         build_target = (
>             f"@{tool_target}//:file" if is_archive else f"@{tool_target}//file"
>         )
>         subprocess.run(["bazel", "build", build_target])
> ```
> 
> If you ran it before and didn't explicitly `bazel clean` it will keep using the old version.

That's unfortunate – I guess we'd have to use `bazel fetch <target>` on every script invocation to work around that?

hoodmane (approved):
Thanks @anonrig!

anonrig:
It seems the checksum is failing for python_rules...

hoodmane:
I think it is also failing now on the main branch.

PR #3733, opened 2025-03-17
Pass featureFlagsForFl to Cache API

ObsidianMinor:
We probably should've been doing this before, but it wasn't really necessary.

### DIFF
--- a/src/workerd/api/cache.c++
+++ b/src/workerd/api/cache.c++
@@ -70,8 +70,10 @@ jsg::Unimplemented Cache::addAll(kj::Array<Request::Info> requests) {
   return {};
 }
 
-jsg::Promise<jsg::Optional<jsg::Ref<Response>>> Cache::match(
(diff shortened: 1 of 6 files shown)


dom96 (approved):
Nice, looks good.

kentonv:
Could this have any backwards-incompatible side effects?

kentonv:
> It shouldn't. Cache is only now starting to use compat flags to change behavior with the most recent one added.

But since the Cache API goes out through FL, won't the existing `neededByFl` feature flags potentially affect its behavior? I kind of suspect they will but I don't know what the impact would be. @harrishancock might have a better idea, though this is all distant past stuff for both of us...

ObsidianMinor:
FL isn't 100% sure so we'll just add another flag to turn _sending_ the flags on or off.

kentonv (approved):

PR #3738, opened 2025-03-17
Always rebuild bazel tool in format.py

hoodmane:
Otherwise, when using `format.py` locally we won't pick up updates to the tools.

See discussion here:
https://github.com/cloudflare/workerd/pull/3732

### DIFF
--- a/tools/cross/format.py
+++ b/tools/cross/format.py
@@ -121,11 +121,8 @@ def run_bazel_tool(
     else:
         tool_path = init_external_dir() / tool_target / "file" / "downloaded"
 
-    if not tool_path.exists():
-        build_target = (
-            f"@{tool_target}//:file" if is_archive else f"@{tool_target}//file"
-        )
-        subprocess.run(["bazel", "build", build_target])
+    build_target = f"@{tool_target}//:file" if is_archive else f"@{tool_target}//file"
+    subprocess.run(["bazel", "build", build_target])
 
     return subprocess.run([tool_path, *args])
 


npaun (approved):
It slows down the tool a bit given that we have to call into bazel always but it fixes a bug so we should do this for now.

fhanau:
`bazel fetch <target>` should achieve the same thing, and I _think_ that it is more lightweight since it can forego the Bazel analysis phase. This should reduce the performance impact of always running bazel.

npaun:
@fhanau The difference between fetch vs build caused some issues in the past. Let's not change it for now until I can look into why that happened.

hoodmane:
> GitHub archives are supposed to remain static

I am pretty sure you can just delete a github archive and add a new one with the same name.

anonrig:
@hoodmane Can you rebase?

anonrig (approved):

tools/cross/format.py
```diff
@@ -121,11 +121,10 @@ def run_bazel_tool(
...
 
-    if not tool_path.exists():
-        build_target = (
-            f"@{tool_target}//:file" if is_archive else f"@{tool_target}//file"
-        )
-        subprocess.run(["bazel", "build", build_target])
+    build_target = (
+        f"@{tool_target}//:file" if is_archive else f"@{tool_target}//file"
+    )
+    subprocess.run(["bazel", "build", build_target])
```
mikea, line 127: why not simply `bazel run {target} -- {args}`? This is what we do for rustfmt

npaun: We tried to make this tool faster by not invoking bazel unless we knew we needed to download a tool. But now that we have to do it unconditionally in order to handle upgrades, I guess there's not much point and we can just do bazel run.

I guess we should suppress all the verbose messages it prints though.

hoodmane: I'm going to keep the current change set as the simplest fix for the problem. Maybe someone can make the various suggested improvements in a followup.

PR #3739, opened 2025-03-17
Fix inconsistent hash on rules_python

npaun:
I've frozen rules_python to 0.4.0 to minimize the effect of this PR. All it does is pin rules_python to a file that will have a consistent hash, because it's an uploaded asset and not a github tarball.

I'll open a folllow-up PR to fix update-deps.py and unfreeze rules_python.


anonrig:
cc @hoodmane

build/deps/gen/dep_rules_python.bzl
```diff
@@ -3,9 +3,9 @@
 load("@//:build/http.bzl", "http_archive")
 
 TAG_NAME = "0.40.0"
-URL = "https://api.github.com/repos/bazelbuild/rules_python/tarball/0.40.0"
-STRIP_PREFIX = "bazelbuild-rules_python-1944874"
-SHA256 = "3e8396b0741231f07e175d0e721969ea3456b4330a4daf32d1520ce427a09937"
+URL = "https://github.com/bazel-contrib/rules_python/releases/download/0.40.0/rules_python-0.40.0.tar.gz"
```
anonrig, line 6: This now uses a user uploaded file rather than the tarball provided by Github. On theory, this is unsafe.

fhanau (approved):

build/deps/gen/dep_rules_python.bzl
npaun: I've verified that the file content is the same, using both `diff -r` and `sha256sum`

mar-cf (approved):

build/deps/gen/dep_rules_python.bzl
npaun: It's also not the mtime btw, Github sensibly sets that to `0`.

anonrig (approved):

PR #3740, opened 2025-03-17
kj::Arc<Worker>

mikea:
To demonstrate the power of Arc<T> I have converted one of the most important use-cases.

This makes ownership of Worker more explicit and easier to understand. I left comments where the code was forced to be different (for better imo).

downstream 10086

### DIFF
--- a/src/workerd/io/io-context.h
+++ b/src/workerd/io/io-context.h
@@ -87,7 +87,7 @@ class WarningAggregator final: public kj::AtomicRefcounted {
   using Map = kj::HashMap<const Key&, kj::Own<WarningAggregator>>;
 
  private:
-  kj::Own<const Worker> worker;
+  kj::Arc<Worker> worker;
   kj::Own<RequestObserver> requestMetrics;
   EmitCallback emitter;
   kj::MutexGuarded<kj::Vector<kj::Own<WarningContext>>> warnings;
@@ -224,7 +224,7 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler 
 
   // Construct a new IoContext. Before using it, you must also create an IncomingRequest.
   IoContext(ThreadContext& thread,
-      kj::Own<const Worker> worker,
+      kj::Arc<Worker> worker,
       kj::Maybe<Worker::Actor&> actor,
       kj::Own<LimitEnforcer> limitEnforcer);
 
@@ -233,8 +233,8 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler 
 
   using IncomingRequest = IoContext_IncomingRequest;
 
-  const Worker& getWorker() {
-    return *worker;
+  const kj::Arc<Worker>& getWorker() {
+    return worker;
   }
   Worker::Lock& getCurrentLock() {
     return KJ_REQUIRE_NONNULL(currentLock);
@@ -835,7 +835,7 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler 
 
   kj::Own<WeakRef> selfRef = kj::refcounted<WeakRef>(kj::Badge<IoContext>(), *this);
 
-  kj::Own<const Worker> worker;
+  kj::Arc<Worker> worker;
   kj::Maybe<Worker::Actor&> actor;
   kj::Own<LimitEnforcer> limitEnforcer;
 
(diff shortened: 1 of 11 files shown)


src/workerd/tests/test-fixture.c++
```diff
@@ -368,7 +368,7 @@ TestFixture::TestFixture(SetupParams&& params)
         return jsg::alloc<api::DurableObjectStorage>(
             IoContext::current().addObject(actorCache), /*enableSql=*/false);
       };
-      actor = kj::refcounted<Worker::Actor>(*worker, /*tracker=*/kj::none, kj::mv(id),
+      actor = kj::refcounted<Worker::Actor>(worker.addRef(), /*tracker=*/kj::none, kj::mv(id),
```
mikea, line 371: made addRef explicit

src/workerd/server/server.c++
```diff
@@ -2081,16 +2081,15 @@ class Server::WorkerService final: public Service,
             co_return;
           }
           KJ_IF_SOME(m, manager) {
-            auto& worker = a->getWorker();
-            auto workerStrongRef = kj::atomicAddRef(worker);
+            auto worker = a->getWorker();
```
mikea, line 2084: simplified/made the code less bug-prone

src/workerd/io/worker.h
```diff
@@ -849,8 +849,8 @@ class Worker::Actor final: public kj::Refcounted {
   // Only needs to be called when allocating a HibernationManager!
   kj::Maybe<uint16_t> getHibernationEventType();
 
-  inline const Worker& getWorker() {
-    return *worker;
+  inline kj::Arc<Worker> getWorker() {
```
mikea, line 852: turns out that we always use the result of this `getWorker()` to hold the reference, so doing addRef here makes sense. Otherwise we can return `kj::Arc<Worker>&` and force call-sites to do `addRef`.

src/workerd/io/io-context.h
```diff
@@ -233,8 +233,8 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
 
   using IncomingRequest = IoContext_IncomingRequest;
 
-  const Worker& getWorker() {
-    return *worker;
+  kj::Arc<Worker>& getWorker() {
```
mikea, line 236: this getWorker() is often used as a temporory to fetch compat flags or script id. It makes sense to return a reference (which simple `auto` won't accept).

src/workerd/io/worker-entrypoint.c++
```diff
@@ -753,10 +753,10 @@ kj::Promise<WorkerInterface::CustomEvent::Result> WorkerEntrypoint::customEvent(
 }
 
 #ifdef KJ_DEBUG
-void requestGc(const Worker& worker) {
+void requestGc(kj::Arc<Worker> worker) {
```
mikea, line 756: passing Arc instead of & makes ownership more explicit and less error-prone in async situations

src/workerd/io/worker-entrypoint.c++
```diff
@@ -753,10 +753,10 @@ kj::Promise<WorkerInterface::CustomEvent::Result> WorkerEntrypoint::customEvent(
...
   TRACE_EVENT("workerd", "Debug: requestGc()");
-  jsg::runInV8Stack([&](jsg::V8StackScope& stackScope) {
-    auto& isolate = worker.getIsolate();
+  jsg::runInV8Stack([worker = kj::mv(worker)](jsg::V8StackScope& stackScope) {
```
kentonv, line 758: Nitpick: This runs synchronously so the move is redundant, I would stick with the `[&]` capture.

src/workerd/io/worker-entrypoint.c++
kentonv: Well, this method isn't async, it's entirely synchronous, so making the caller increment the refcount is redundant.

But it's only used in debug so I don't care that much.

jasnell: Hm.. given the usage here and how it is called, I agree with @kentonv that the additional addRef feels unnecessary here. Seems to me this could possible be a `const kj::Arc<Worker>&` ref instead and the capture below can remain `[&]`

jasnell (approved):

src/workerd/io/worker-entrypoint.c++
kentonv: I think @mikea is right that the only call site has to addRef anyway, and is able to move-in that reference, so we're not adding any addrefs after all.

PR #3741, opened 2025-03-17
`@cloudflare/workers-types` v5

penalosa:
After the stabilisation of [`wrangler types`-based runtime type generation](https://developers.cloudflare.com/workers/languages/typescript/) in Wrangler v4, the confusing entrypoint system in `@cloudflare/workers-types` is no longer needed. As such, this PR changes `@cloudflare/workers-types` to only include the _latest_ version of the runtime types (previously this was the `/experimental` entrypoint).

After some analysis of open source repos with Sourcegraph, it's clear that the usage of the alternative entrypoints to `@cloudflare/workers-types` is very low to 0, and so this shouldn't have a practical impact on users. However, it's still a breaking change, and so this will be released as `@cloudflare/workers-types@5`. There are two migration paths available to users:
* Continue using `@cloudflare/workers-types`, but be aware that the typing environment may not be 100% compatible with runtime behaviour, depending on your compatibility date + flags, or
* [Migrate to `wrangler types`](https://developers.cloudflare.com/workers/languages/typescript/#migrating)

cc @anonrig this will also cut the build time of workers-types by 90%

cc @irvinebroque @korinne for visibility


DaniFoldi:
Hi @penalosa :wave:,

Would it perhaps make more sense to have it defined as an alias to the "latest stable" date? Right now, that would be 2023-07-01, which more closely maps to types available to users. Experimental has some methods which aren't yet accessible to users, but if they show up in types, it might cause confusion.

dario-piotrowicz (approved):

penalosa:
(converting to draft so that this doesn't get accidentally merged while still under discussion)

penalosa:
@DaniFoldi The problem with the "latest stable" compatibility date is that it doesn't really represent anything in the runtime. It's just the last time the typings changed in a significant way, but it also means users miss out on types added since then. Using the experimental entrypoint has a slight risk of experimental APIs being available in user code, but it also means that everything else is always up to date.

irvinebroque:
So basically this means — if you want types that match a specific compat date — then you need to use `wrangler types`?

types/scripts/build-types.ts
```diff
@@ -10,28 +10,7 @@ import { SourcesMap, createMemoryProgram } from "../src/program.js";
...
-  // https://developers.cloudflare.com/workers/platform/compatibility-dates/#compliant-transformstream-constructor
-  { compatDate: "2022-11-30" },
-  // https://github.com/cloudflare/workerd/blob/fcb6f33d10c71975cb2ce68dbf1924a1eeadbd8a/src/workerd/io/compatibility-date.capnp#L275-L280 (http_headers_getsetcookie)
-  { compatDate: "2023-03-01" },
-  // https://github.com/cloudflare/workerd/blob/fcb6f33d10c71975cb2ce68dbf1924a1eeadbd8a/src/workerd/io/compatibility-date.capnp#L307-L312 (urlsearchparams_delete_has_value_arg)
-  { compatDate: "2023-07-01" },
-  // Latest compatibility date with experimental features
-  { compatDate: "experimental" },
-];
+const ENTRYPOINTS = [{ compatDate: "experimental", name: "" }];
```
anonrig, line 13: Can we move the spawning of workerd to bazel now?

DaniFoldi:
> but it also means users miss out on types added since then

I don't think it does, the only types that are unavailable are ones that haven't been default on that date, such as some stream related ones in the oldest date.

types/scripts/build-types.ts
```diff
@@ -11,27 +11,9 @@ import { getFilePath } from "../src/utils";
...
-  { compatDate: "2022-11-30" },
-  // https://github.com/cloudflare/workerd/blob/fcb6f33d10c71975cb2ce68dbf1924a1eeadbd8a/src/workerd/io/compatibility-date.capnp#L275-L280 (http_headers_getsetcookie)
-  { compatDate: "2023-03-01" },
-  // https://github.com/cloudflare/workerd/blob/fcb6f33d10c71975cb2ce68dbf1924a1eeadbd8a/src/workerd/io/compatibility-date.capnp#L307-L312 (urlsearchparams_delete_has_value_arg)
-  { compatDate: "2023-07-01" },
-  // Latest compatibility date with experimental features
-  { compatDate: "2999-12-31", name: "latest" },
-  // Latest compatibility date with experimental features
+  // Latest compatibility date
+  { compatDate: "2999-12-31", name: "" },
```
jamesopstad, line 18: Does this result in a separate export to `experimental` below or are they the same?

penalosa: It'll end up with `@cloudflare/workers-types` (flags on as of 2999-12-31) and `@cloudflare/workers-types/experimental` (which will have experimental flags turned on)

jamesopstad: Thanks. That seems a little different to what it says in the release plan document.

> As such, we’re releasing @cloudflare/workers-types v5, which only includes the latest version of the runtime types (previously this was the /experimental entrypoint).

Does that need updating?

anonrig (approved):

NuroDev:

NuroDev (approved):

types/scripts/build-types.ts
penalosa: @NuroDev I don't think this is resolved—it seems like we've removed experimental types from the generation and diffing, which we'll need to keep

penalosa: (I think I removed it in my initial commit 😅). e.g. the change in `justfile`

NuroDev: Good catch @penalosa! I've pushed a fix for this in 8d44a46 that restores generation and diffing for experimental types.
The `justfile` now explicitly copies the generated types in the top directory. Experimental types get copied as a directory like they did before.

NuroDev:
@penalosa AI just pointed out to me, we also probably get rid of [this line](https://github.com/cloudflare/workerd/blob/main/.github/workflows/release.yml#L329) too since `oldest/` types won't exist anymore, correct?

penalosa:
> @penalosa AI just pointed out to me, we also probably get rid of [this line](https://github.com/cloudflare/workerd/blob/main/.github/workflows/release.yml#L329) too since `oldest/` types won't exist anymore, correct?

Yes, sounds good

justfile
```diff
@@ -112,7 +112,8 @@ clang-tidy target="//...":
 
 generate-types:
   bazel build //types
-  cp -r bazel-bin/types/definitions/latest types/generated-snapshot/
+  cp bazel-bin/types/definitions/index.d.ts types/generated-snapshot/
+  cp bazel-bin/types/definitions/index.ts types/generated-snapshot/
```
dario-piotrowicz, line 116: why do here we copy the files specifically instead of `cp -r`?

dario-piotrowicz (approved):

PR #3742, opened 2025-03-18
JSRPC: Fix multiple bugs with proxy pipelining

kentonv:
See commits.

I tried using Claude Code to help debug this, fully expecting it would be too hard for it. Indeed, it flailed around a fair amount, proposing several non-fixes that weren't anywhere near the right location. But much to my surprise it then suddenly identified the problem fixed in the first commit -- which I hadn't actually found myself yet. I was pretty shocked because it requires a pretty deep understanding of the RPC system to get what's going on there... Mostly though this change is my own work rather than Claude's.


jasnell (approved):

kentonv:
Rebased on main in hopes of the workers-sdk test failure going away.

kentonv:
The internal build failure is spurious and we want to get this out, so I'm ignoring it.

PR #3748, opened 2025-03-18
Extend D1 meta information - served_by_region/served_by_primary and sql_duration_ms

lambrospetrou:
These fields are already returned by the D1 API so adding them in the D1 Binding official types.

Example of response (see the `result.meta` field):
```
{
  "result": [
    {
      "results": [
        {
          "1": 1
        }
      ],
      "success": true,
      "meta": {
        ...
        "served_by_region": "WNAM",
        "served_by_primary": true,
        "timings": {
          "sql_duration_ms": 0.2084
        },
        "duration": 0.2084,
        ...
      }
    }
  ],
  "errors": [],
  "messages": [],
  "success": true
}
```

### DIFF
--- a/types/defines/d1.d.ts
+++ b/types/defines/d1.d.ts
@@ -11,6 +11,7 @@ interface D1Meta {
    * The region of the database instance that executed the query.
    */
   served_by_region?: string;
+
   /**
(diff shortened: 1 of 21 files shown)


danlapid (approved):
Would likely need the type snapshot generated as per failed CI job

PR #3750, opened 2025-03-19
03-18-25 assorted build cleanup

fhanau:
- Only force Rust static linkage on Windows, Linux should still use shared link when possible
- Fix typos, prefer bazel name over bazelisk
- Add ncrypto to compile_flags.txt to fix clangd includes
- Port --nobuild_runfile_links bazel option from downstream, reducing CI I/O overhead.


npaun (approved):
All here seems good to me

PR #3752, opened 2025-03-19
remove secureContext() not implemented error to unblock mysql2

anonrig:
This will unblock connecting and using mysql2 with SSL.

cc @vicb @thomasgauvin

### DIFF
--- a/src/node/internal/internal_tls_wrap.ts
+++ b/src/node/internal/internal_tls_wrap.ts
@@ -207,10 +207,14 @@ export function TLSSocket(
     throw new ERR_OPTION_NOT_IMPLEMENTED('options.requestOCSP');
   }
 
-  if (tlsOptions.secureContext !== undefined) {
-    // TODO(soon): Investigate supporting this.
-    throw new ERR_OPTION_NOT_IMPLEMENTED('options.secureContext');
-  }
+  // In order to unblock mysql2 SSL connections, we omit the validation
+  // of secureContext option. When we implement createSecureContext(),
+  // we will visit this behavior.
+  // TODO(soon): Implement createSecureContext() to support secureContext option.
+  //
+  // if (tlsOptions.secureContext !== undefined) {
+  //   throw new ERR_OPTION_NOT_IMPLEMENTED('options.secureContext');
+  // }
 
   if (tlsOptions.pskCallback !== undefined) {
     // Used for TLS-PSK negotiation. We do not support it.


src/node/internal/internal_tls_wrap.ts
```diff
@@ -207,10 +207,14 @@ export function TLSSocket(
     throw new ERR_OPTION_NOT_IMPLEMENTED('options.requestOCSP');
   }
 
-  if (tlsOptions.secureContext !== undefined) {
-    // TODO(soon): Investigate supporting this.
-    throw new ERR_OPTION_NOT_IMPLEMENTED('options.secureContext');
-  }
+  // In order to unblock mysql2 SSL connections, we omit the validation
```
danlapid, line 210: So mysql2 is setting it but then not using it?
Seems really weird to silently fail here, can confuse our users

anonrig: mysql calls createSecureContext() function which is not yet implemented right now, but @vicb will make sure that it will not throw "unimplemented error" if you just call it with empty options. in that scenario, it will later fail at this case. Even if you don't set custom SSL certificates, it seems mysql2 team decided to call this function regardless.

This won't cause friction, since createSecureContext() will handle the validation.

danlapid: Maybe instead of checking tlsOptions.secureContext !== undefined we can check that secureContext doesn't have any field set to anything other than what we can handle and otherwise throw?

anonrig: It will be handled in: https://github.com/cloudflare/workers-sdk/pull/8577

danlapid: We should handle it in the runtime on usage rather than in the polyfill on creation.

vicb: That's what we initially discuss but doesn't really look possible: https://github.com/nodejs/node/blob/109c817db63197870c72d8f8e25e151684fa1277/lib/_tls_common.js#L72-L94

danlapid: Not sure what I'm seeing in the file linked but from the runtime's perspective we shouldn't silently fail.
If we support some usage of SecureContext we can lax our error handling to allow that usage and disallow whatever parameters of SecureContext that we don't support.

PR #3754, opened 2025-03-19
add createSecureContext and checkServerIdentity to node:tls

anonrig:
The following pull-request:

- Adds createSecureContext() function but only supports options with no values. this is done to unblock mysql2 from working in cloudflare workers where no custom ssl options are passed.
- Adds checkServerIdentity() for validating a PeerCertificate according to the host.

### DIFF
--- a/src/node/internal/internal_errors.ts
+++ b/src/node/internal/internal_errors.ts
@@ -732,9 +732,9 @@ export class ERR_TLS_INVALID_CONTEXT extends NodeTypeError {
 export class ERR_TLS_CERT_ALTNAME_INVALID extends NodeError {
   reason: string;
   host: string;
-  cert: PeerCertificate;
+  cert?: Partial<PeerCertificate> | undefined;
 
-  constructor(reason: string, host: string, cert: PeerCertificate) {
+  constructor(reason: string, host: string, cert?: Partial<PeerCertificate>) {
     super('ERR_TLS_CERT_ALTNAME_INVALID', reason);
     this.reason = reason;
     this.host = host;
--- a/src/node/internal/internal_errors.ts
+++ b/src/node/internal/internal_errors.ts
@@ -709,6 +709,12 @@ export class ERR_TLS_HANDSHAKE_TIMEOUT extends NodeError {
   }
 }
 
+export class ERR_TLS_INVALID_CONTEXT extends NodeTypeError {
+  constructor(field: string) {
+    super('ERR_TLS_INVALID_CONTEXT', `${field} must be a SecureContext`);
+  }
+}
+
 export class ConnResetException extends NodeError {
   path?: string | undefined;
   host?: string | undefined;
--- a/src/node/internal/internal_errors.ts
+++ b/src/node/internal/internal_errors.ts
@@ -30,6 +30,7 @@
 // TODO(soon): Fill in more of the internal/errors implementation
 
 import { inspect } from 'node-internal:internal_inspect';
+import type { PeerCertificate } from 'node:tls';
 
 const classRegExp = /^([A-Z][a-z0-9]*)+$/;
 
@@ -83,6 +84,19 @@ export class NodeTypeError extends NodeErrorAbstraction implements TypeError {
   constructor(code: string, message: string) {
     super(TypeError.prototype.name, code, message);
     Object.setPrototypeOf(this, TypeError.prototype);
+    this.toString = function () {
+      return `${this.name} [${this.code}]: ${this.message}`;
+    };
+  }
+}
+
+export class NodeSyntaxError
+  extends NodeErrorAbstraction
+  implements SyntaxError
+{
+  constructor(code: string, message: string) {
+    super(SyntaxError.prototype.name, code, message);
+    Object.setPrototypeOf(this, SyntaxError.prototype);
     this.toString = function () {
       return `${this.name} [${this.code}]: ${this.message}`;
     };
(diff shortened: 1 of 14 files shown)


danlapid:
LGTM, let's block until @jasnell reviews though 😅

src/workerd/api/node/tests/tls-nodejs-test.js
```diff
@@ -282,3 +282,16 @@ export const tlsConnectGivenSocket = {
     await Promise.all(promises);
   },
 };
+
+export const testSecureContext = {
+  async test() {
+    throws(() => tls.connect({ secureContext: {} }), {
```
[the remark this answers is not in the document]
anonrig: Yes

```
> const tls = require('tls')
undefined
> tls.connect({ secureContext: {} })
Uncaught TypeError [ERR_TLS_INVALID_CONTEXT]: context must be a SecureContext
    at TLSSocket._wrapHandle (node:_tls_wrap:696:11)
    at new TLSSocket (node:_tls_wrap:570:18)
    at Object.connect (node:_tls_wrap:1766:19) {
  code: 'ERR_TLS_INVALID_CONTEXT'
}
```

mikea (approved):

src/node/internal/internal_tls_common.ts
```diff
@@ -0,0 +1,65 @@
...
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import type tls from 'node:tls';
+import { ERR_OPTION_NOT_IMPLEMENTED } from 'node-internal:internal_errors';
+import { validateInteger } from 'node-internal:validators';
+
+export class SecureContext {
```
jasnell, line 30: It pains me to say it but in Node.js, `const sc  = tls.SecureContext()` works .. this might need to be an old-style function class.

src/node/internal/internal_errors.ts
```diff
@@ -89,6 +90,19 @@ export class NodeTypeError extends NodeErrorAbstraction implements TypeError {
...
+export class NodeSyntaxError
+  extends NodeErrorAbstraction
+  implements SyntaxError
+{
+  constructor(code: string, message: string) {
+    super(SyntaxError.prototype.name, code, message);
+    Object.setPrototypeOf(this, SyntaxError.prototype);
+    this.toString = function () {
+      return `${this.name} [${this.code}]: ${this.message}`;
+    };
```
npaun, line 102: Why does toString need to be injected like this, instead of being declared as a method?

src/rust/net/lib.rs
```diff
@@ -0,0 +1,16 @@
+use std::net::IpAddr;
+use std::str::FromStr;
+
+#[cxx::bridge(namespace = "workerd::rust::net")]
```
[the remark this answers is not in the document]
anonrig: I'm not sure. Also there is no direct way of validating if a host is ipv4 or ipv6 unless you parse the whole string, which makes it a lot more complex than this 3 lines.

src/node/internal/internal_errors.ts
anonrig: No specific way. We can do the other way around as well.

src/rust/net/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+    extern "Rust" {
+        fn canonicalize_ip(input: &str) -> String;
+    }
+}
+
+#[must_use]
+pub fn canonicalize_ip(input: &str) -> String {
+    IpAddr::from_str(input)
+        .map(|ip| ip.to_string())
+        .unwrap_or(String::new())
```
npaun, line 15: This returns empty string for invalid IPs. Do we check for this anywhere?

anonrig: No empty string means it will match for anything. This is imitating nodejs behavior unfortunately.

npaun: Just in case someone needs to use this function for something else in the future, I'd suggest adding a comment about this on the C++ wrapper function.

npaun (approved):

src/node/tls.ts
```diff
@@ -23,9 +23,13 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+import { checkServerIdentity } from 'node-internal:internal_tls';
```
vicb, line 26: should `SecureContext` be export too?

If not `SecureContext` will come from the polyfill and `createSecureContext() instanceof SecureContext` will be false

anonrig: Yes you're right.

src/node/internal/internal_tls_common.ts
```diff
@@ -0,0 +1,78 @@
...
+  public constructor(
+    _secureProtocol?: string,
+    secureOptions?: number,
+    minVersion?: string,
+    maxVersion?: string
+  );
+}
+
+// @ts-expect-error TS2323 Redeclare error.
+export function SecureContext(
```
jasnell, line 42: Nit: it is likely worth adding a comment that this is intentionally not fully compatible with Node.js' implementation since creating and customizing an actually equivalent `SecureContext` is not supported.

src/workerd/api/node/tests/tls-nodejs-test.js
```diff
@@ -282,3 +283,356 @@ export const tlsConnectGivenSocket = {
...
 };
+
+export const testSecureContext = {
+  async test() {
+    throws(() => tls.connect({ port: 42, secureContext: {} }), {
+      code: 'ERR_TLS_INVALID_CONTEXT',
+    });
+
+    doesNotThrow(() => {
+      const secureContext = tls.createSecureContext({});
```
jasnell, line 294: Should have a test that also covers the `tls.SecureContext()` without new or using the `createSecureContext()` method so we make sure we don't regress.

jasnell (approved):
LGTM with a couple of nits

jasnell (approved):

PR #3759, opened 2025-03-20
Implements durable object support for Python Workers in workerd.

dom96:
This PR implements basic Durable Object support in workerd. 

As workerd currently relies on Python code not being run at the top-level, we cannot introspect the Python code looking for DurableObject classes. So instead we grab the durable object classes defined in the workerd config and use those. For EW we'll implement introspection of the Python code to find any classes that are defined in the main Python module (this will come in a separate PR).

The Durable Object classes are defined in a named export which is only read when Python Workers are enabled.

### TODOs before merging

- [x] RPC calls seem to fail currently
- [x] Ensure that only primitive types can be sent/received via RPC for now 

### Test plan

```
```


src/workerd/api/pyodide/pyodide.c++
```diff
@@ -514,6 +514,11 @@ jsg::Ref<PyodideMetadataReader> makePyodideMetadataReader(Worker::Reader conf,
   }
   auto lock = KJ_ASSERT_NONNULL(getPyodideLock(pythonRelease),
       kj::str("No lock file defined for Python packages release ", pythonRelease.getPackages()));
+  kj::Vector<kj::String> durableObjectClasses =
```
danlapid, line 517: How about
```c++
auto durableObjectClasses = KJ_MAP(objectNamespace, conf.getDurableObjectNamespaces()) {
    return kj::str(objectNamespace.getClassName());
}
```
?

src/workerd/io/worker.c++
```diff
@@ -1675,6 +1675,32 @@ Worker::Worker(kj::Own<const Script> scriptParam,
...
+                                    });
+                              }
+                              KJ_CASE_ONEOF(obj, api::ExportedHandler) {
+                                KJ_FAIL_ASSERT("Expected EntrypointClass");
+                              }
+                            }
+                          }
+                        }
+
                         impl->namedHandlers.insert(kj::mv(handler.name), kj::mv(obj));
```
danlapid, line 1704: Shouldn't we either add the handler as python DO OR add the handler as is rather than both?

src/workerd/server/tests/python/durable-object/worker.py
```diff
@@ -0,0 +1,39 @@
...
+class DurableObjectExample(DurableObject):
+    def __init__(self, state, env):
+        self.state = state
+        self.counter = 0
+
+    async def on_fetch(self, req, env, ctx):
+        self.counter += 1
+        return Response(f"hello from python {self.counter}")
+
+    async def custom_method(self, arg):
```
danlapid, line 17: untested?

danlapid:
NamedEntrypoints excluded on purpose I assume?

Would love to see JSRPC to DOs work, very important feature.
Had a chance to talk to Kenton about it yet?

dom96:
Yep, we talked with Kenton. It turns out that JS RPC doesn't support serialising custom data types in JavaScript, so no need to support that in Python. Instead we will make sure the types that we do support have Python equivalents (which is mainly just types that can be structured cloned, https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) and we'll convert to/from Python types where necessary automatically. This should work very nicely and allow us to interoperate with any language so long as it supports those core structured cloneable types.

In any case, won't do that in this PR because of the way Pyodide calls JS methods causes a failure. Something that will need to be fixed in JS RPC or in Pyodide itself (that's the TODO here: https://github.com/cloudflare/workerd/pull/3759#discussion_r2008219172).

danlapid:
> Yep, we talked with Kenton. It turns out that JS RPC doesn't support serialising custom data types in JavaScript, so no need to support that in Python. Instead we will make sure the types that we do support have Python equivalents (which is mainly just types that can be structured cloned, https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) and we'll convert to/from Python types where necessary automatically. This should work very nicely and allow us to interoperate with any language so long as it supports those core structured cloneable types.
> 
> 
> 
> In any case, won't do that in this PR, because the way Pyodide calls JS methods currently causes a failure. Something that will need to be fixed in JS RPC or in Pyodide itself (that's the TODO here: https://github.com/cloudflare/workerd/pull/3759#discussion_r2008219172).

SGTM
Great answers
To be clear, is that change(the Python JSRPC fix) not part of this PR but is intended to be resolved before allowing access to DOs for customers? (In other words, is it still part of the Python DOs workstream?)

dom96:
> To be clear, is that change(the Python JSRPC fix) not part of this PR but is intended to be resolved before allowing access to DOs for customers? (In other words, is it still part of the Python DOs workstream?)

yep, the plan is to resolve it before releasing DOs to customers.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -162,26 +163,98 @@ function makeHandler(pyHandlerName: string): Handler {
...
+
+    constructor(...args: any[]) {
+      super(...args);
+      // Initialise a Python instance of the class.
+      this.pyInstance = this.initPyInstance(args);
+    }
+
+    async initPyInstance(args: any[]) {
+      const mainModule = await getMainModule();
+      const constructor = mainModule[className] as unknown as (
```
hoodmane, line 179: Can we call this `pyClass` maybe? It's confusing with the `constructor()` function above.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -162,26 +163,98 @@ function makeHandler(pyHandlerName: string): Handler {
...
+      const instance = new target(...constrArgs);
+      return new Proxy(instance, {
+        get(target, prop, receiver) {
+          if (typeof prop !== 'string') {
+            return null;
+          }
+          const isKnownHandler = SUPPORTED_HANDLER_NAMES.includes(prop);
+          if (isKnownHandler) {
+            prop = 'on_' + prop;
+          }
```
hoodmane, line 203: Rather than using a proxy for this we should dynamically generate the class.

hoodmane: Let me see if I can work something out that is reasonable.

src/pyodide/python-entrypoint.js
```diff
@@ -6,4 +6,23 @@
...
+  makeDurableObjectClass,
+} from 'pyodide:python-entrypoint-helper';
+
+let pythonDurableObjects = {};
+for (const className of pythonDurableObjectClasses) {
+  pythonDurableObjects[className] = makeDurableObjectClass(
+    className,
+    DurableObject
+  );
+}
```
hoodmane, line 25: Nit:
```suggestion
const pythonDurableObjects = Object.fromEntries(
  pythonDurableObjectClasses.map((className) => [
    className,
    makeDurableObjectClass(className, DurableObject),
  ]),
);
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -162,26 +163,98 @@ function makeHandler(pyHandlerName: string): Handler {
...
+let pythonDurableObjectClasses: string[] = [];
+
 try {
   // Do not setup anything to do with Python in the global scope when tracing. The Jaeger tracing
   // needs to be called inside an IO context.
   if (IS_WORKERD || IS_TRACING) {
-    handlers.fetch = makeHandler('on_fetch');
+    for (const className of DURABLE_OBJECT_CLASSES ?? []) {
+      pythonDurableObjectClasses.push(className);
+    }
```
hoodmane, line 246:
```suggestion
  pythonDurableObjectClasses.push(...(DURABLE_OBJECT_CLASSES ?? []));
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -162,26 +163,98 @@ function makeHandler(pyHandlerName: string): Handler {
...
+            return null;
+          }
+          const isKnownHandler = SUPPORTED_HANDLER_NAMES.includes(prop);
+          if (isKnownHandler) {
+            prop = 'on_' + prop;
+          }
+          const asyncWrapper = async function (...args: any[]) {
+            const pyInstance = await instance.pyInstance;
+            if (typeof pyInstance[prop] === 'function') {
+              return pyInstance[prop](...args).then((res: any) => {
```
hoodmane, line 207: Nit: Use `await` instead of `.then()`.

src/pyodide/python-entrypoint-helper.ts
hoodmane: Okay, since it's necessary we just need a comment explaining why.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -162,26 +163,98 @@ function makeHandler(pyHandlerName: string): Handler {
...
+                if (res.js_object && isKnownHandler) {
```
hoodmane, line 208:
```suggestion
                if (isKnownHandler && res?.js_object) {
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -162,26 +163,98 @@ function makeHandler(pyHandlerName: string): Handler {
...
+                  return res.js_object;
+                }
```
hoodmane, line 210:
```suggestion
                if (isKnownHandler) {
                  return res?.js_object ?? res;
                }
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -162,26 +163,98 @@ function makeHandler(pyHandlerName: string): Handler {
...
+      return new Proxy(instance, {
+        get(target, prop, receiver) {
+          if (typeof prop !== 'string') {
+            return null;
+          }
+          const isKnownHandler = SUPPORTED_HANDLER_NAMES.includes(prop);
+          if (isKnownHandler) {
+            prop = 'on_' + prop;
+          }
+          const asyncWrapper = async function (...args: any[]) {
```
hoodmane, line 204:
```suggestion
          return async function (...args: any[]) {
```

src/workerd/api/kv-test.js
hoodmane, line 1: Maybe this should be reverted since it's not really relevant to this PR?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -164,26 +165,90 @@ function makeHandler(pyHandlerName: string): Handler {
...
+    pyInstance: Promise<PyModule>;
+
+    constructor(...args: any[]) {
+      super(...args);
+      // Initialise a Python instance of the class.
+      this.pyInstance = this.initPyInstance(args);
+      // We do not know the methods that are defined on the RPC class, so we need a proxy to
+      // support any possible method name.
+      return new Proxy(this, {
+        get(target, prop, receiver) {
```
hoodmane, line 180: Not clear to me that this respects [the invariants required of a get handler](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/get#invariants). But we can probably fix any problems later as they come up.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -164,26 +165,90 @@ function makeHandler(pyHandlerName: string): Handler {
...
+          if (typeof prop !== 'string') {
+            return null;
```
hoodmane, line 182: Wouldn't one of `return Reflect.get(target, prop, receiver);` or `return undefined;` make more sense here? Why `null`? Did you hit this code path?

hoodmane (approved):

src/workerd/api/kv-test.js
dom96: Somehow the formatter decided to autoformat this, let's see if reverting won't break the lint CI...

PR #3771, opened 2025-03-21
Migrate from deprecated v8::Write interfaces

ketanhwr:
We need this v8 patch to be merged for this to work: https://chromium-review.googlesource.com/c/v8/v8/+/6369297

### DIFF
--- a/src/workerd/api/node/buffer.c++
+++ b/src/workerd/api/node/buffer.c++
@@ -83,8 +83,7 @@ uint32_t writeInto(jsg::Lock& js,
     return 0;
   }
 
-  static constexpr jsg::JsString::WriteOptions flags = static_cast<jsg::JsString::WriteOptions>(
-      jsg::JsString::NO_NULL_TERMINATION | jsg::JsString::REPLACE_INVALID_UTF8);
+  static constexpr jsg::JsString::WriteOptions flags = jsg::JsString::WriteOptions::REPLACE_INVALID_UTF8;
 
   switch (encoding) {
     case Encoding::ASCII:
@@ -120,10 +119,7 @@ uint32_t writeInto(jsg::Lock& js,
     }
     case Encoding::HEX: {
       KJ_STACK_ARRAY(kj::byte, buf, string.length(js), 1024, 536870888);
-      static constexpr jsg::JsString::WriteOptions options =
-          static_cast<jsg::JsString::WriteOptions>(
-              jsg::JsString::NO_NULL_TERMINATION | jsg::JsString::REPLACE_INVALID_UTF8);
-      string.writeInto(js, buf, options);
+      string.writeInto(js, buf, flags);
       auto backing = decodeHexTruncated(js, buf, false);
       auto bytes = backing.asArrayPtr();
       auto amountToCopy = kj::min(bytes.size(), dest.size());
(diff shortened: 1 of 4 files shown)


src/workerd/api/node/buffer.c++
```diff
@@ -120,10 +119,7 @@ uint32_t writeInto(jsg::Lock& js,
     }
     case Encoding::HEX: {
       KJ_STACK_ARRAY(kj::byte, buf, string.length(js), 1024, 536870888);
-      static constexpr jsg::JsString::WriteOptions options =
-          static_cast<jsg::JsString::WriteOptions>(
-              jsg::JsString::NO_NULL_TERMINATION | jsg::JsString::REPLACE_INVALID_UTF8);
-      string.writeInto(js, buf, options);
+      string.writeInto(js, buf, flags);
```
anonrig, line 122: Previously we had REPLACE_INVALID_UTF8 in this path, but now we don't. Is this intentional?

src/workerd/jsg/jsvalue.h
```diff
@@ -250,10 +250,9 @@ class JsString final: public JsBase<v8::String, JsString> {
   static JsString concat(Lock& js, const JsString& one, const JsString& two) KJ_WARN_UNUSED_RESULT;
 
   enum WriteOptions {
```
anonrig, line 252: Nit: I think we should rename this to WriteFlags, since the new implementation changes the name from options to flags and it might cause some confusion until you look at the actual declaration of this enum.

patches/v8/0027-Remove-DCHECK-from-WriteOneByteV2-to-skip-v8-fatal.patch
```diff
@@ -0,0 +1,22 @@
...
+
+diff --git a/src/api/api.cc b/src/api/api.cc
+index 43540a968a5b7e94c3e883d1bf8b5f51072bae05..da393be16944058da30acc9d0e3d94de0203257a 100644
+--- a/src/api/api.cc
++++ b/src/api/api.cc
+@@ -6157,7 +6157,8 @@ void String::WriteV2(Isolate* v8_isolate, uint32_t offset, uint32_t length,
+ 
+ void String::WriteOneByteV2(Isolate* v8_isolate, uint32_t offset,
+                             uint32_t length, uint8_t* buffer, int flags) const {
+-  DCHECK(IsOneByte());
```
erikcorry, line 17: I think the assert should be ContainsOnlyOneByte instead, and I think we can uncomment it again if you do that. If not it seems like a bug in workerd that it is calling the API with non-Latin1 text.

The comment is wrong I think.

Probably a simple thing to upstream too.

erikcorry (approved):
LGTM

patches/v8/0027-Remove-DCHECK-from-WriteOneByteV2-to-skip-v8-fatal.patch
erikcorry: OK after off-line discussion, let's just commit this as is, then file bugs against both node and workerd to handle high bytes better.  Currently, node ignores the high bytes, while we incorrectly pass them to V8 and then redo all the work afterwards, which is wasteful but follows the spec (but not node).

erikcorry (approved):

anonrig (approved):

PR #3772, opened 2025-03-21
upgrade rules_python

mikea:


fhanau (approved):
Couldn't figure it out when I tried this before – great work!

PR #3773, opened 2025-03-21
[o11y] Consolidate adding trace events for legacy and streaming tail workers

fhanau:
First commit is very simple refactoring to facilitate cloning trace events. Second commit removes reportTailEvent() and instead uses WorkerTracer to report events to the tail stream writer. This means duplication in adding trace events is abstracted away.

- Note that Outcome event reporting is not folded into this class at present.
- Note that this can cause the legacy trace size limit to the Streaming model – this will be cleaned up once we transmit all event data over the stream layer.
- As a side effect, streaming tail worker now supports headers in fetch handler onset events, update tests accordingly
- This will make it easy to support Return and SpanOpen/SpanClose events.

### DIFF
--- a/src/workerd/io/tracer.c++
+++ b/src/workerd/io/tracer.c++
@@ -74,23 +74,28 @@ kj::Own<WorkerTracer> PipelineTracer::makeWorkerTracer(PipelineLogLevel pipeline
     kj::Maybe<kj::Own<ScriptVersion::Reader>> scriptVersion,
     kj::Maybe<kj::String> dispatchNamespace,
     kj::Array<kj::String> scriptTags,
-    kj::Maybe<kj::String> entrypoint) {
+    kj::Maybe<kj::String> entrypoint,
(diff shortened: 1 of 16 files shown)


src/workerd/io/tracer.c++
```diff
@@ -117,6 +125,9 @@ void WorkerTracer::addLog(kj::Date timestamp, LogLevel logLevel, kj::String mess
     return;
   }
   trace->bytesUsed = newSize;
+  KJ_IF_SOME(writer, maybeTailStreamWriter) {
+    writer->report(context, tracing::Mark(tracing::Log(timestamp, logLevel, kj::str(message))));
+  }
```
jasnell, line 130: Perhaps consider adding `TODO` comments in here to make it clear that additional refactors are expected to handle the case where a worker has a streaming tail worker but no legacy tail worker.

jasnell (approved):

PR #3774, opened 2025-03-21
add missing translatePeerCertificate function

anonrig:
This is the last remaining function provided in "_tls_common" module. We can now remove the polyfill from it.

cc @vicb would you mind updating the polyfill for `_tls_common`?

### DIFF
--- a/src/node/_tls_common.ts
+++ b/src/node/_tls_common.ts
@@ -26,7 +26,8 @@
 import {
   createSecureContext,
   SecureContext,
+  translatePeerCertificate,
 } from 'node-internal:internal_tls_common';
 
-export { createSecureContext, SecureContext };
-export default { createSecureContext, SecureContext };
+export { createSecureContext, SecureContext, translatePeerCertificate };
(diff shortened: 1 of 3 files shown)


npaun (approved):

vicb:
> cc @vicb would you mind updating the polyfill for `_tls_common`?

Sure.

We also need to sync with https://github.com/cloudflare/workerd/pull/3754 which has already been merged. Should we batch both the changes in a single PR? It would make sense if both the changes are required to move forward but it would also delay the change as we would need to wait for the current PR to be merged and released.

Let me know how to proceed here!

jasnell:
We already have a variation on this implemented that might be reusable here? Specifically, look at the toLegacyObject method on the X509Certificate class. In Node.js, that method uses the `translatePeerCertificate` function. I ended up implementing it a different way when writing X509Certificate because I didn't think we were going to be doing TLSSocket. It doesn't need to be done in this PR but it would be ideal to consolidate the code so we don't have duplication.

PR #3775, opened 2025-03-21
add node:sqlite to NODEJS_BUILTINS

anonrig:


npaun (approved):

jasnell (approved):

jasnell:
Btw, it's likely possible to implement the `node:sqlite` API (for in-memory dbs only) in workerd... especially given that we already have sqlite support for DOs

anonrig:
> Btw, it's likely possible to implement the `node:sqlite` API (for in-memory dbs only) in workerd... especially given that we already have sqlite support for DOs

Yes, it's on my agenda but I'm waiting for SQLite to be stable.

PR #3776, opened 2025-03-21
HTMLRewriter: no more write promise chaining

npaun:
This PR follows up on #3758 to provide a more solid fix. 

The chained `writePromise` is no longer used. Instead, every time lol_html generates output, we store it in `outputBuffer` (just a `kj::Vector`). `outputBuffer` is flushed to the output sink after each handler thunk. If more than 128 MB accumulates in `outputBuffer`, we poison the `Rewriter`, preventing users from causing arbitrary memory usage.

As a result of this PR, the test `stackOverflow1` now passes (lots of output from one handler), and `stackOverflow2` continues to pass.

### DIFF
--- a/src/workerd/api/html-rewriter.c++
+++ b/src/workerd/api/html-rewriter.c++
@@ -325,9 +325,17 @@ class Rewriter final: public WritableStreamSink {
   // this (buildRewriter()) modifies that vector.
   kj::Own<lol_html_HtmlRewriter> rewriter;
 
-  kj::Own<WritableStreamSink> inner;
+  // Stores data written by lol-html, which will be periodically flushed to inner.
+  kj::Vector<kj::byte> outputBuffer;
 
-  kj::Maybe<kj::Promise<void>> writePromise;
+  // Used to ensure memory usage is reported to V8 and cannot grow arbritrarily
+  jsg::ExternalMemoryAdjustment externalMemoryAdjustment;
+
+  // True if we are currently flushing from outputBuffer to inner (in flushWrite)
+  bool flushing = false;
+
+  // The destination for the output from lol-html
+  kj::Own<WritableStreamSink> inner;
 
   kj::Maybe<kj::Exception> maybeException;
 
@@ -422,6 +430,7 @@ Rewriter::Rewriter(jsg::Lock& js,
     kj::ArrayPtr<const char> encoding,
     kj::Own<WritableStreamSink> inner)
     : rewriter(buildRewriter(js, unregisteredHandlers, encoding, *this)),
+      externalMemoryAdjustment(js.getExternalMemoryAdjustment()),
       inner(kj::mv(inner)),
       ioContext(IoContext::current()),
       maybeAsyncContext(jsg::AsyncContextFrame::currentRef(js)) {}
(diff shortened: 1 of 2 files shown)


src/workerd/api/html-rewriter.c++
```diff
@@ -325,9 +325,11 @@ class Rewriter final: public WritableStreamSink {
   // this (buildRewriter()) modifies that vector.
   kj::Own<lol_html_HtmlRewriter> rewriter;
 
-  kj::Own<WritableStreamSink> inner;
+  // Stores data written by lol-html, which will be periodically flushed to inner.
+  kj::Vector<kj::byte> outputBuffer;
```
danlapid, line 329: Instead of an arbitrary 128MB limit per rewriter why not use ExternalMemoryAccounter to wrap the vector to correctly account for it's memory?
That will let us handle memory overallocation correctly.

src/workerd/api/tests/htmlrewriter-test.js
```diff
@@ -968,10 +968,6 @@ export const svgNamespace = {
   },
 };
 
-// This test is commented out because it will cause a stack overflow in workerd if enabled.
-// It is included for reference.
-
-/*
 export const stackOverflow1 = {
```
danlapid, line 971: Test name likely need to be changed

src/workerd/api/html-rewriter.c++
```diff
@@ -516,14 +519,15 @@ kj::Promise<void> Rewriter::flushWrite() {
...
   };
 
-  KJ_IF_SOME(wp, writePromise) {
-    KJ_DEFER(writePromise = kj::none);
-    return wp.then([checkException]() { return checkException(); });
+  if (!outputBuffer.empty()) {
+    auto bufferCopy = kj::heapArray(outputBuffer.asPtr());
+    outputBuffer.clear();
+
+    return inner->write(bufferCopy).then([checkException]() { return checkException(); });
```
harrishancock, line 526: I think this is a UaF, because we don't attach `bufferCopy` to the returned promise.

But, I think we probably don't have to make a copy anyway, since we don't expect anyone to add more data to the buffer until we finish flushing it here. I also think we could convert this function to a coroutine to simplify it a bit (it was written before we implemented KJ coroutines). Something like this:

```cpp
kj::Promise<void> Rewriter::flushWrite() {
  if (!outputBuffer.empty()) {
    KJ_DEFER(outputBuffer.clear());
    co_await inner->write(outputBuffer);
  }
  KJ_IF_SOME(exception, maybeException) {
    inner->abort(kj::cp(exception));
    kj::throwFatalException(kj::cp(exception));
  }
}
```

src/workerd/api/html-rewriter.c++
```diff
@@ -754,15 +758,12 @@ void Rewriter::outputImpl(kj::ArrayPtr<const byte> buffer) {
...
-  auto bufferCopy = kj::heapArray(buffer);
-
-  KJ_IF_SOME(wp, writePromise) {
-    writePromise = wp.then([this, bufferCopy = kj::mv(bufferCopy)]() mutable {
-      return inner->write(bufferCopy.asPtr()).attach(kj::mv(bufferCopy));
-    });
-  } else {
-    writePromise = inner->write(bufferCopy.asPtr()).attach(kj::mv(bufferCopy));
+  if (outputBuffer.size() + buffer.size() > MAX_OUTPUT_BUFFER_SIZE) {
+    maybePoison(KJ_EXCEPTION(FAILED, kj::str(JSG_EXCEPTION(RangeError) ": Memory limit exceeded")));
```
harrishancock, line 762: If this exceptional code path still exists after resolving @danlapid's comment about memory accounting (https://github.com/cloudflare/workerd/pull/3776/files#r2008768150), it'd be nice to have a test which exercises it. That is, a test case which expects this RangeError to be thrown.

harrishancock (approved):

npaun:
rebasin'

src/workerd/api/html-rewriter.c++
```diff
@@ -505,25 +510,17 @@ kj::Promise<void> Rewriter::finishWrite() {
...
-
-    KJ_IF_SOME(exception, maybeException) {
-      inner->abort(kj::cp(exception));
-      return kj::cp(exception);
-    }
-
-    return kj::READY_NOW;
-  };
+  if (!outputBuffer.empty()) {
+    KJ_DEFER(outputBuffer.clear());
```
jclee, line 514: Not super familiar with this code, but I assume that something ensures that a `flushWrite()` call must complete before more content is added to `outputBuffer` -- e.g. either something in the fiber code or an invariant in lolhtml?  Otherwise, I'd wonder if it might be possible to lose output via a sequence of events, like:

- `flushWrite()` calls `inner->write()` to write the initial content of `outputBuffer`
- more output gets added to `outputBuffer` while waiting for the write to complete
- the write succeeds, resolving the `inner->write()` promise, which then runs `outputBuffer.clear()` to clean up the scope, before the content that was added to `outputBuffer` can be used

If the code ensures this can't happen, it might be possible to assert against it to make that clear -- maybe by asserting that only one `flushWrite()` invocation can be active at once, and that none are active when adding content to `outputBuffer`?

harrishancock: I think it's currently safe because `thunkImpl()` `.wait()`s its promises in serial, but you are right, we should assert this.

Perhaps we could add a `bool flushing` and assert that there's no concurrent `flushWrite()` operation whenever we add to `outputBuffer`?

npaun: Would it have to be an `atomic_bool`?

harrishancock: No, I don't think there's any possible thread (or signal handler) synchronization concerns.

npaun:
rebased

PR #3778, opened 2025-03-22
Change process.env implementation to be based on importable env.

kentonv:
This means that `disallow_importable_env` will disable both importable env and `process.env`, which is likely what anyone using this flag wants. (Source: Me, the only known person who even wants `disallow_importable_env`.)

This also allows us to delete a lot of code.

Although this is techincally a breaking change since these fetaures have already shipped, it seems highly unlikely that someone has managed to deploy code which both sets `disallow_importable_env` and also `nodejs_compat_populate_process_env` and depends on `process.env`. This will get harder once `nodejs_compat_populate_process_env` becomes the default on `2025-04-01`.

### DIFF
--- a/src/node/internal/process.ts
+++ b/src/node/internal/process.ts
@@ -20,13 +20,105 @@ export function nextTick(cb: Function, ...args: unknown[]) {
   });
 }
 
-// Note that there is no process-level environment in workers so the process.env
-// object is initially empty. This is different from Node.js where process.env
-// picks up values from the operating system environment. The configured bindings
-// for the worker are accessible from the env argument passed into the fetch
-// handler and have no impact here.
+// Decide if a value can round-trip to JSON without losing any information.
+function isJsonSerializable(
+  value: any,
+  seen: Set<Object> = new Set()
+): boolean {
+  switch (typeof value) {
+    case 'boolean':
+    case 'number':
+    case 'string':
+      return true;
 
-export const env = new Proxy(utilImpl.getEnvObject(), {
+    case 'object': {
+      if (value === null) {
+        return true;
+      }
+
+      if (seen.has(value)) {
+        // Don't allow cycles or aliases. (Non-cyclic aliases technically could be OK, but a
+        // round trip to JSON would lose the fact that they are aliases.)
+        return false;
+      }
+      seen.add(value);
+
+      if (typeof value.toJSON === 'function') {
+        // This type is explicitly designed to be JSON-serialized so we'll accept it.
+        return true;
+      }
+
+      // We only consider objects to be serializable if they are plain objects or plain arrays.
+      // Technically, JSON can serialize any subclass of Object (as well as objects with null
+      // prototypes), but the round trip would lose information about the original type. Hence,
+      // we assume that any env var containing such things is not intended to be appear as JSON
+      // in process.env. For example, we wouldn't want a KV namespace to show up in process.env as
+      // "{}" -- this would be weird.
+      switch (Object.getPrototypeOf(value)) {
+        case Object.prototype:
+          // Note that Object.values() only returns string-keyed values, not symbol-keyed.
+          return Object.values(value).every((prop) =>
+            isJsonSerializable(prop, seen)
+          );
+        case Array.prototype:
+          return (<Array<unknown>>value).every((elem) =>
+            isJsonSerializable(elem, seen)
+          );
+        default:
+          return false;
+      }
+    }
+
+    default:
+      return false;
+  }
+}
+
+function getInitialEnv() {
+  let env: Record<string, string> = {};
+  for (let [key, value] of Object.entries(utilImpl.getEnvObject())) {
+    // Workers environment variables can have a variety of types, but process.env vars are
+    // strictly strings. We want to convert our workers env into process.env, but allowing
+    // process.env to contain non-strings would probably break Node apps.
+    //
+    // As a compromise, we say:
+    // - Workers env vars that are plain strings are unchanged in process.env.
+    // - Workers env vars that can be represented as JSON will be JSON-stringified in process.env.
+    // - Anything else will be omitted.
+    //
+    // Note that you might argue that, at the config layer, it's possible to differentiate between
+    // plain strings and JSON values that evaluated to strings. Wouldn't it be nice if we could
+    // check which way the binding was originally configured in order to decide whether to
+    // represent it plain or as JSON here. However, there is no way to tell just by looking at
+    // the `env` object inside a Worker whether a particular var was originally configured as
+    // plain text, or as JSON that evaluated to a string. Either way, you just get a string. And
+    // indeed, the Workers Runtime itself does not necessarily know this. In many cases it does
+    // know, but in general the abstraction the Runtime intends to provide is that `env` is just
+    // a JavaScript object, and how exactly the contents were originally represented is not
+    // intended to be conveyed. This is important because, for example, we could extend dynamic
+    // dispatch bindings in the future such that the caller can specify `env` directly, and in
+    // that case the caller would simply specify a JS object, without JSON or any other
+    // serialization involved. In this case, there would be no way to know if a string var was
+    // "supposed to be" raw text vs. JSON.
+    //
+    // So, we have to do the best we can given just what we know -- the JavaScript object that is
+    // `env`.
+    //
+    // As a consolation, this is consistent with how variables are defined in wrangler.toml: you
+    // do not explicitly specify whether a variable is text or JSON. If you define a variable with
+    // a simple string value, it gets configured as a text var. If you specify an object, then it's
+    // configured as JSON.
+
+    if (typeof value === 'string') {
+      env[key] = value;
+    } else if (isJsonSerializable(value)) {
+      env[key] = JSON.stringify(value);
+    }
+  }
+  return env;
+}
+
+export const env = new Proxy(getInitialEnv(), {
   // Per Node.js rules. process.env values must be coerced to strings.
   // When defined using defineProperty, the property descriptor must be writable,
   // configurable, and enumerable using just a falsy check. Getters and setters
--- a/src/workerd/api/node/util.c++
+++ b/src/workerd/api/node/util.c++
@@ -250,7 +250,17 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
 }
 
 jsg::JsObject UtilModule::getEnvObject(jsg::Lock& js) {
-  return js.getProcessEnv(true);
+  if (FeatureFlags::get(js).getPopulateProcessEnv()) {
+    KJ_IF_SOME(env, js.getWorkerEnv()) {
+      jsg::JsValue val(env.getHandle(js));
+      KJ_IF_SOME(obj, val.tryCast<jsg::JsObject>()) {
+        return obj;
+      }
+    }
+  }
+
+  // Default to empty object.
+  return js.obj();
 }
 
 namespace {
--- a/src/workerd/api/node/util.c++
+++ b/src/workerd/api/node/util.c++
@@ -252,10 +252,7 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
 jsg::JsObject UtilModule::getEnvObject(jsg::Lock& js) {
   if (FeatureFlags::get(js).getPopulateProcessEnv()) {
     KJ_IF_SOME(env, js.getWorkerEnv()) {
-      jsg::JsValue val(env.getHandle(js));
-      KJ_IF_SOME(obj, val.tryCast<jsg::JsObject>()) {
-        return obj;
-      }
+      return jsg::JsObject(env.getHandle(js));
     }
   }
 
--- a/src/workerd/api/modules.c++
+++ b/src/workerd/api/modules.c++
@@ -20,10 +20,8 @@ kj::Maybe<jsg::JsObject> EnvModule::getCurrent(jsg::Lock& js) {
   if (FeatureFlags::get(js).getDisableImportableEnv()) return kj::none;
 
   // Otherwise, fallback to provide the stored environment.
-  return js.getWorkerEnv().map([&](const jsg::Value& val) -> jsg::JsObject {
-    auto handle = val.getHandle(js);
-    JSG_REQUIRE(handle->IsObject(), TypeError, "Expected environment to be an object.");
-    return jsg::JsObject(handle.As<v8::Object>());
+  return js.getWorkerEnv().map([&](const jsg::V8Ref<v8::Object>& val) -> jsg::JsObject {
+    return jsg::JsObject(val.getHandle(js));
   });
 }
 
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -1618,7 +1618,7 @@ Worker::Worker(kj::Own<const Script> scriptParam,
               // Use `env` variable.
               bindingsScope = v8::Object::New(lock.v8Isolate);
               if (!FeatureFlags::get(js).getDisableImportableEnv()) {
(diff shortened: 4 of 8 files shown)


src/workerd/api/node/util.c++
```diff
@@ -242,7 +243,17 @@ jsg::JsValue UtilModule::getBuiltinModule(jsg::Lock& js, kj::String specifier) {
 }
 
 jsg::JsObject UtilModule::getEnvObject(jsg::Lock& js) {
-  return js.getProcessEnv(true);
+  if (FeatureFlags::get(js).getPopulateProcessEnv()) {
+    KJ_IF_SOME(env, js.getWorkerEnv()) {
+      jsg::JsValue val(env.getHandle(js));
+      KJ_IF_SOME(obj, val.tryCast<jsg::JsObject>()) {
```
anonrig, line 249: Whats the reason for not having an assertion here, but a tryCast?

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+    }
+
+    default:
+      return false;
+  }
+}
+
+function getInitialEnv() {
+  let env: Record<string, string> = {};
+  for (let [key, value] of Object.entries(utilImpl.getEnvObject())) {
```
anonrig, line 64: Why don't we move this to C++?

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+        case Object.prototype:
+          for (let prop of Object.values(value)) {
+            if (!isJsonSerializable(prop)) return false;
+          }
+          return true;
+        case Array.prototype:
+          for (let elem of value) {
+            if (!isJsonSerializable(elem)) return false;
+          }
+          return true;
```
anonrig, line 51: Nit: This can be simplified with an array.prototype.some() call

src/workerd/api/node/util.c++
jasnell: We know this will always be an object or nothing since we're in control of it. An assert would be unnecessary

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+
+      // We only consider objects to be serializable if they are plain objects or plain arrays.
+      // Any subclass is not serializable because serializing it would be lossy.
+      //
+      // (Practically speaking, we don't want, say, a KV namespace to show up in process.env as
+      // "{}" -- this would be weird.)
+      switch (Object.getPrototypeOf(value)) {
+        case Object.prototype:
+          for (let prop of Object.values(value)) {
+            if (!isJsonSerializable(prop)) return false;
```
jasnell, line 44: If the object has circular references this recursion will be a problem. You'll likely want to memoize and bail out

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+          }
+          return true;
+        case Array.prototype:
+          for (let elem of value) {
+            if (!isJsonSerializable(elem)) return false;
```
jasnell, line 49: Likewise here, this will be a problem if there are circular references

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+function isJsonSerializable(value: any): boolean {
+  switch (typeof value) {
+    case "boolean":
+    case "number":
+    case "string":
+      return true;
 
-export const env = new Proxy(utilImpl.getEnvObject(), {
+    case "object": {
+      if (value === null) {
```
jasnell, line 32: What if the object contains a toJSON method?

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+      }
+
+      // We only consider objects to be serializable if they are plain objects or plain arrays.
+      // Any subclass is not serializable because serializing it would be lossy.
+      //
+      // (Practically speaking, we don't want, say, a KV namespace to show up in process.env as
+      // "{}" -- this would be weird.)
+      switch (Object.getPrototypeOf(value)) {
+        case Object.prototype:
+          for (let prop of Object.values(value)) {
```
jasnell, line 43: This should probably ignore entries whose keys are symbols since json stringify ignores those.

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+      if (value === null) {
+        return true;
+      }
+
+      // We only consider objects to be serializable if they are plain objects or plain arrays.
+      // Any subclass is not serializable because serializing it would be lossy.
+      //
+      // (Practically speaking, we don't want, say, a KV namespace to show up in process.env as
+      // "{}" -- this would be weird.)
+      switch (Object.getPrototypeOf(value)) {
```
jasnell, line 41: An object with a null prototype is also safe to json stringify

src/node/internal/process.ts
```diff
@@ -20,13 +20,90 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+    if (typeof value === "string") {
+      env[key] = value;
+    } else if (isJsonSerializable(value)) {
+      env[key] = JSON.stringify(value);
+    }
+  }
+  return env;
+}
+
+export const env = new Proxy(getInitialEnv(), {
```
jasnell, line 106: One key difference with this approach is that the process.env will be populated lazily when first imported, correct? That means if there are manual changes to env those will be picked up, which is an inconsistency since changes to env after that point won't be picked up

src/node/internal/process.ts
jasnell, line 64: This should ignore symbol keys

src/workerd/api/node/util.c++
kentonv: I don't really know why the env handle is of type JsValue and not JsObject, but I figure if it's not an object, we probably what process.env to be empty.

src/node/internal/process.ts
kentonv: It's much easier to write in JavaScript. It may also be faster in JS, since the C++ APIs for manipulating JS values aren't particularly efficient and in JS the compiler may optimize it.

src/node/internal/process.ts
kentonv: Ugh.

At present there's no way to express a cyclic object as an env var, and actually I would like to back out support of cycles in JSRPC because it's a security issue (because people tend to forget about cycles when writing code exactly like this).

But I suppose it would be good to not make assumptions here.

Two options:
* Use a Set to track what we've already seen.
* Don't actually do a deep scan. I was actually considering this anyway: Maybe we just say, if the top-level value is a simple object or simple array, we go ahead and JSON-serialize it. This does mean that if there's any nested non-JSON types, they may stringify as `{}`. But that doesn't seem so bad, and we don't have any extant case where that happens anyway.

Which do you prefer?

kentonv: Wait never mind, if we don't detect cycles then JSON.stringify() will actually throw an exception which will break everything.

So I guess we have to use the first option.

src/node/internal/process.ts
kentonv: Well, we don't have any case where that would happen today, but I can check for it if you think we should.

src/node/internal/process.ts
kentonv: `Object.values()` only returns values with string keys. I was actually surprised by this myself, but that's what the docs and my tests say.

src/node/internal/process.ts
kentonv: Technically objects with _any_ prototype are safe to JSON-stringify. What this function is testing for is whether the object looks like it came from JSON / can round-trip to JSON without losing its type.

src/node/internal/process.ts
kentonv: `Object.entries()` also returns only string keys.

src/node/internal/process.ts
kentonv: I suppose that true. Is it worth worrying about? I think it's unlikely to affect anyone, and the only way to avoid it would be to make a copy early on in C++ code which seems like a lot of work.

jasnell: Well, given that the user can introduce objects with circular references, we will, at the very least, need to address that.

src/node/internal/process.ts
jasnell: Ah! that's right, forgot about that.

src/node/internal/process.ts
jasnell: Yeah I totally forgot that symbol properties were omitted from those... Yet Another Fun TC-39 Quirk.

src/node/internal/process.ts
jasnell: As with null prototypes, if we're going to allow `process.env` to be built lazily after user code has an opportunity to modify `env` (https://github.com/cloudflare/workerd/pull/3778#discussion_r2008634277) then we probably should account for it.

src/node/internal/process.ts
jasnell: Yep, using the set and bailing out (returning false) if cycles are detected would be the best option here.

src/workerd/api/node/util.c++
jasnell: It could be changed to a JsObject.

kentonv: I added a commit which enforces through the type system that this is always an object, so we don't need a runtime check at all.

src/node/internal/process.ts
kentonv: Done.

src/node/internal/process.ts
kentonv: Updated to check for `toJSON` and accept anything that has it.

src/node/internal/process.ts
kentonv: Sorry, I think you didn't understand my comment. I do not want to permit null prototypes here, for the same reason that I don't want to permit prototypes of arbitrary subclasses of Object: they would be lost in a round-trip, therefore the type is not really JSON-serializable.

src/node/internal/process.ts
kentonv: Cycles have now been addressed.

src/node/internal/process.ts
kentonv: Matter of opinion but honestly I find the loop easier to read.

src/node/internal/process.ts
jasnell: I would suggest adding a comment then to that effect since a null-prototype object is technically JSON stringify-able. Something like, "Ignore null prototype objects because, while they can be JSON stringified, they cannot be parsed with a null prototype".

jasnell: heh, just saw the updated comment :-) ... so nevermind! :-)

src/node/internal/process.ts
```diff
@@ -20,13 +20,107 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+      }
+
+      if (seen.has(value)) {
+        // Don't allow cycles or aliases. (Non-cyclic aliases technically could be OK, but a
+        // round trip to JSON would lose the fact that they are aliases.)
+        return false;
+      }
+      seen.add(value);
+
+      if ('toJSON' in value) {
```
jasnell, line 46:
```suggestion
      if (typeof value.toJSON === 'function') {
```

Just to prevent against silly user code bugs (I've actually seen things like `foo.toJSON = {}` in the wild.... yes, I know, it's awful)

anonrig (approved):

vicb:
A couple questions:

- "Although this is techincally a breaking change" -> how hard would it be to not make this a breaking change 
- Are updated to `process.env` reflected on `env` after this change?

jasnell (approved):

src/node/internal/process.ts
```diff
@@ -20,13 +20,107 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+    // Note that you might argue that, at the config layer, it's possible to differentiate between
+    // plain strings and JSON values that evaluated to strings. Wouldn't it be nice if we could
+    // check which way the binding was originally configured in order to decide whether to
+    // represent it plain or as JSON here. However, there is no way to tell just by looking at
+    // the `env` object inside a Worker whether a particular var was originally configured as
+    // plain text, or as JSON that evaluated to a string. Either way, you just get a string. And
+    // indeed, the Workers Runtime itself does not necessarily know this. In many cases it does
+    // know, but in general the abstraction the Runtime intends to provide is that `env` is just
+    // a JavaScript object, and how exactly the contents were originally represented is not
+    // intended to be conveyed. This is important becasue, for example, we could extend dynamic
```
vicb, line 100:
```suggestion
    // intended to be conveyed. This is important because, for example, we could extend dynamic
```

src/node/internal/process.ts
```diff
@@ -20,13 +20,107 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+      // we assume that any env var containing such things is not intended to be appear as JSON
+      // in process.env. For example, we wouldn't want a KV namespace to show up in process.env as
+      // "{}" -- this would be weird.
+      switch (Object.getPrototypeOf(value)) {
+        case Object.prototype:
+          // Note that Object.values() only returns string-keyed values, not symbol-keyed.
+          for (let prop of Object.values(value)) {
+            if (!isJsonSerializable(prop, seen)) return false;
+          }
+          return true;
```
vicb, line 63:
```suggestion
          return Object.values(value).every(v => isJsonSerializable(v, seen));
```

src/node/internal/process.ts
```diff
@@ -20,13 +20,107 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+        case Array.prototype:
+          for (let elem of value) {
+            if (!isJsonSerializable(elem, seen)) return false;
+          }
+          return true;
+        default:
+          return false;
```
vicb, line 70:
```suggestion
          return value.every(elem => isJsonSerializable(elem, seen));
```

kentonv:
> "Although this is techincally a breaking change" -> how hard would it be to not make this a breaking change

That would mean not fixing the bug. The current behavior is not what we wanted.

> Are updated to process.env reflected on env after this change?

No, nor were they before the change.

src/node/internal/process.ts
kentonv: Didn't know about `every()`. That does look nicer, but not worth another CI cycle + rebase of the internal PR. I'll apply it if some other issue forces me to push further updates to this PR.

src/node/internal/process.ts
```diff
@@ -20,13 +20,107 @@ export function nextTick(cb: Function, ...args: unknown[]) {
...
+      // in process.env. For example, we wouldn't want a KV namespace to show up in process.env as
+      // "{}" -- this would be weird.
+      switch (Object.getPrototypeOf(value)) {
+        case Object.prototype:
+          // Note that Object.values() only returns string-keyed values, not symbol-keyed.
+          for (let prop of Object.values(value)) {
+            if (!isJsonSerializable(prop, seen)) return false;
+          }
+          return true;
+        case Array.prototype:
```
vicb, line 64: nit: sparse arrays are not json serializable.

jasnell: well, they are but they don't roundtrip as sparse arrays. For instance,

```js
const m = [1,,,,5];
// [1, < 3 empty items>, 5]

const str = JSON.stringify(m);
// '[1,null,null,null,5]'

const n = JSON.parse(str);
// [ 1, null, null, null, 5 ]
```

jasnell (approved):

PR #3780, opened 2025-03-23
plumb specific internal request type into limit enforcer newSubrequest

Frederik-Baetens:
I made these changes to enable the limit enforcer to enforce free tier limits for various products. This goes paired with an internal EW PR that actually does the enforcement.

### DIFF
--- a/src/workerd/io/internal-subrequest-type.h
+++ b/src/workerd/io/internal-subrequest-type.h
@@ -1,0 +1,7 @@
+#pragma once
+
+#include <kj/one-of.h>
+
+struct GenericInternalSubrequest {};
+struct DOSubrequest {};
+typedef kj::OneOf<GenericInternalSubrequest, DOSubrequest> InternalSubrequestType;
--- a/src/workerd/api/actor.c++
+++ b/src/workerd/api/actor.c++
@@ -38,7 +38,7 @@ class LocalActorOutgoingFactory final: public Fetcher::OutgoingFactory {
       return KJ_REQUIRE_NONNULL(actorChannel)
           ->startRequest({.cfBlobJson = kj::mv(cfStr), .tracing = tracing});
     },
-        {.inHouse = true,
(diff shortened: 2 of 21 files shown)


src/workerd/io/internal-subrequest-type.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <kj/one-of.h>
+
+struct GenericInternalSubrequest{};
+struct DOSubrequest{};
+typedef kj::OneOf<GenericInternalSubrequest, DOSubrequest> InternalSubrequestType;
```
Frederik-Baetens, line 7: This could be a flat enum as it stands, but in the future KV might use this system too for their free tier limit enforcement, and they might want to further subdivide their operations. It's possible to collapse those operations into a flat enum nonetheless, but I thought that that would be a bit messy in the future, so I went for a kj::OneOf.

Can change this to an enum if that's deemed to more tasteful.

Frederik-Baetens:
Build fails with `function is not marked const`, but so does main :face_with_diagonal_mouth:

kentonv:
Hmm, this is a lot of churn and I'm not sure it's necessary. Couldn't we do the enforcement in `ActorChannelImpl::startRequest()` instead?

I don't think this mechanism would work well for a future implementation of KV free tier as KV is also in the process of converting to an RPC API, in which case the C++ wrapper would go away and they'd need some other way to identify KV requests (probably a setting on the binding itself rather than based on the interface being used).

Frederik-Baetens:
:+1: That sounds like it'd work. I'll look how to plumb that in & just keep the approach focused on DO then without paying much attention to the needs of KV. 

Do you think that'd be better than adding a separate DO-specific method to limit-enforcer.h? Or do you think a similar amount of plumbing will be necessary when using new methods on LimitEnforcer as is seen in this PR which tries to re-use the limiting in newSubrequest?

PR #3782, opened 2025-03-24
Constructing a `WorkerEntrypoint` from userland

penalosa:
This PR adds a sample that demonstrates that it's not possible to construct a `WorkerEntrypoint` in userland with a user-provided `ctx` object. Instead, this error appears:

```
workerd/server/server.c++:3880: error: Uncaught exception: workerd/jsg/_virtual_includes/iterator/workerd/jsg/value.h:1473: failed: remote.jsg.TypeError: Failed to construct 'WorkerEntrypoint': constructor parameter 1 is not of type 'ExecutionContext'.
```

This is important for the Vitest integration because we need to be able to (within a single "real" execution context):
- Construct a WorkerEntrypoint with a custom execution context
- wait for all `waitUntil`s in the custom execution context to be settled
- Do some assertions

This works with default export-style workers, because the `fetch()` method can just be called directly with the custom execution context, but that's not possible with `WorkerEntrypoint`-style workers.

---

Or at least that _was_ the case! Thanks to the changes now in this PR, the above is possible—and `WorkerEntrypoint` can be constructed with any object, allowing for custom execution contexts.

### DIFF
--- a/src/workerd/api/tests/js-rpc-test.js
+++ b/src/workerd/api/tests/js-rpc-test.js
@@ -1731,3 +1731,19 @@ export let proxiedRpcTarget = {
     }
   },
 };
+
+// Test that we can construct a WorkerEntrypoint
+export class MyEntrypoint extends WorkerEntrypoint {
+  rpcFunc() {
+    return 'hello from entrypoint';
+  }
+}
+function constructEntrypoint(cls, env) {
+  return new cls({ waitUntil: () => {} }, env);
+}
+export let testConstructEntrypoint = {
+  async test(controller, env, ctx) {
+    const constructed = constructEntrypoint(MyEntrypoint, env);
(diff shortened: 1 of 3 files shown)


kentonv:
I guess this is just because the constructor is declared to take `ExecutionContext`:

https://github.com/cloudflare/workerd/blob/1dfb4472c3b4139a4b749611b927e796127b2826/src/workerd/api/worker-rpc.h#L493-L500

I think literally just changing the type of `ctx` from `jsg::Ref<ExecutionContext>` to `jsg::JsObject` would solve the problem, no other code changes.

kentonv:
Could add this to js-rpc-test pretty easily.

danlapid:
@kentonv added test and cleaned up sample

kentonv (approved):

PR #3784, opened 2025-03-24
CACHE-11937: enable cache rules request cf overrides compat flag

yj7o5:
This PR enables the compat flag which had been experimental and tested in FL for overriding cache rules when cache settings specified in the request cf object.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -688,7 +688,8 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
 
   requestCfOverridesCacheRules @72 :Bool
       $compatEnableFlag("request_cf_overrides_cache_rules")
-      $experimental
+      $compatDisableFlag("no_request_cf_overrides_cache_rules")
+      $compatEnableDate("2025-04-02")
       $neededByFl;
   # Enables cache settings specified request in fetch api cf object to override cache rules. (only for user owned or grey-clouded sites)
 


src/workerd/io/compatibility-date.capnp
```diff
@@ -688,7 +688,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
 
   requestCfOverridesCacheRules @72 :Bool
       $compatEnableFlag("request_cf_overrides_cache_rules")
-      $experimental
+      $compatEnableDate("2025-03-26")
```
danlapid, line 691: Let's set this a bit further in the future

src/workerd/io/compatibility-date.capnp
```diff
@@ -688,7 +688,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
 
   requestCfOverridesCacheRules @72 :Bool
       $compatEnableFlag("request_cf_overrides_cache_rules")
```
danlapid, line 690: If we intend to make this the default we'll also need a compatDisableFlag

yj7o5: how does the disable flag works? can someone set both `request_cf_overrides_cache_rules` and `no_ request_cf_overrides_cache_rules` at the same time? because in FL I just check for the presence of `request_cf_overrides_cache_rules` to enable the feature

danlapid: It will either be enabled or disabled, in FL if you find `request_cf_overrides_cache_rules` then it is enabled, otherwise it is disabled.

src/workerd/io/compatibility-date.capnp
yj7o5: updated; pushed a week back

PR #3787, opened 2025-03-24
[ci] Make wrangler team code owners of types/, src/cloudflare/

fhanau:
As discussed with @danlapid. The runtime team often has limited context on changes to various bindings resulting in changes to src/cloudflare or type definitions, these directories should be owned by the wrangler team.
Let me know if you have any concerns/further suggestions here.

CC @penalosa


penalosa (approved):
I'm onboard with this. cc @lrapoport-cf for visibility

anonrig (approved):

PR #3789, opened 2025-03-24
Don't preload dynamic libraries in newer Python compat dates

hoodmane:
This fixes some but not all of the test failures with the new package set and also should improve performance because it avoids loading unneeded dynamic libraries.


src/pyodide/BUILD.bazel
```diff
@@ -155,10 +153,6 @@ REPLACEMENTS = [
         "Date.now",
         "monotonicDateNow",
     ],
-    [
-        "reportUndefinedSymbols()",
-        "reportUndefinedSymbolsNoOp()",
```
hoodmane, line 160: It turns out that `__heap_base` is only set to its correct value if we either preload at least one dynamic library (which moves the heap base) or if we call `reportUndefinedSymbols()`. This problem was masked before because we always loaded at least one dynamic library.

src/pyodide/BUILD.bazel
```diff
@@ -178,9 +172,13 @@ REPLACEMENTS = [
         "(() => {throw new Error('Internal Emscripten code tried to eval, this should not happen, please file a bug report with your requirements.txt file\\'s contents')})()",
     ],
     [
-        "!libData&&flags.fs",
-        "!(libData ??= patchDynlibLookup(Module, libName))&&flags.fs",
+        "!libData",
```
hoodmane, line 175: The updated version works with debug builds.

danlapid (approved):

petebacondarwin (approved):

PR #3790, opened 2025-03-24
Remove the NodeJsCompatModule type

jasnell:
This was never fully deployed and we've opted to go a different direction with things. I've confirmed there are no uses in production. Let's simplify and remove.


anonrig (approved):

src/workerd/server/workerd-api.c++
```diff
@@ -530,6 +516,10 @@ kj::Maybe<jsg::ModuleRegistry::ModuleInfo> WorkerdApi::tryCompileModule(jsg::Loc
       // Nothing to do. Handled in compileModules.
       return kj::none;
     }
+    default: {
```
danlapid, line 519: I think we should catch the `config::Worker::Module::OBSOLETE` specifically and throw rather than define default.
Not having defaults is generally good practice and makes extending the enum easier as compile errors direct you to more locations where the enum is used.

src/workerd/server/workerd-api.c++
```diff
@@ -1039,6 +1020,10 @@ kj::Own<jsg::modules::ModuleRegistry> WorkerdApi::initializeBundleModuleRegistry
         // Handled separately
         break;
       }
+      default: {
```
danlapid, line 1023: likewise

danlapid (approved):

jasnell:
@penalosa ... just a heads up here. I think there are some tests in wrangler that will need to be updated?

jasnell:
Internal PR is good.

PR #3795, opened 2025-03-25
Avoid taking isolate lock recursively in ResponseStreamWrapper

jasnell:
The destructor for `ResponseStreamWrapper` would unconditionally take the isolate lock even if it was already held, leading to "recusive isolate lock" warnings spamming sentry. This changes to check to see if the lock is already held (indirectly via checking to see if there is a current IoContext). If there is a current IoContext, do not take the lock.

### DIFF
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -3947,28 +3947,28 @@ class Worker::Isolate::ResponseStreamWrapper final: public kj::AsyncOutputStream
   }
 
   ~ResponseStreamWrapper() noexcept(false) {
-    jsg::runInV8Stack([&](jsg::V8StackScope& stackScope) {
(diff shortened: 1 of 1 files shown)


src/workerd/io/worker.c++
```diff
@@ -3948,7 +3948,27 @@ class Worker::Isolate::ResponseStreamWrapper final: public kj::AsyncOutputStream
 
   ~ResponseStreamWrapper() noexcept(false) {
     jsg::runInV8Stack([&](jsg::V8StackScope& stackScope) {
```
kentonv, line 3950: If we already have a lock, then we're already in a V8 stack, s owe shouldn't call `runInV8Stack` at all I think.

kentonv (approved):

PR #3796, opened 2025-03-25
add node:fs and node:fs/promises stub methods

anonrig:
> Work in progress.

- [x] Add sync and callback APIs
- [x] Put it behind an experimental flag


jasnell (approved):

src/node/internal/validators.ts
```diff
@@ -333,6 +333,28 @@ export function validatePort(
...
 
+const octalReg = /^[0-7]+$/;
+const modeDesc = 'must be a 32-bit unsigned integer or an octal string';
+
+export function parseFileMode(
+  value: unknown,
+  name: string,
+  def?: number
+): number {
+  if (def) {
```
vicb, line 344: what if `def === 0`
```suggestion
  if (def !== undefined) {
```

src/node/internal/internal_fs_promises.ts
```diff
@@ -0,0 +1,324 @@
...
+}
+
+export async function appendFile(
+  _path: FilePath,
+  _data: unknown,
+  options: Record<string, unknown>
+): Promise<void> {
+  options = getOptions(options, { encoding: 'utf8', mode: 0o666, flag: 'a' });
+  options = copyObject(options);
+  options.flag ||= 'a';
```
vicb, line 63: Do we need to do some processing if we throw at the end anyway?

anonrig: Well this is going to change, since I'll be implementing node:fs. This is just a stub and some input validation to prepare for the actual work.

vicb: Should this be a DRAFT PR at the moment?

anonrig: There are still todos on the description. But I don't plan to add the actual implementation on this PR.

jasnell (dismissed):

jasnell:
This can be rebased on built on top of https://github.com/cloudflare/workerd/pull/4029 will a number of refactors.

Also, I think we should initially put this behind an experimental compatibility flag that we can switch to default on later.

anonrig:
> This can be rebased on built on top of #4029 will a number of refactors.
> 
> Also, I think we should initially put this behind an experimental compatibility flag that we can switch to default on later.

Since #4029 is a moving target, I prefer to wait for it to land to update this pull-request.

jasnell:
PR updated with a range of APIs implemented. More to come.

src/node/internal/internal_fs.ts
```diff
@@ -0,0 +1,115 @@
...
+import { getOptions } from 'node-internal:internal_fs_utils';
+import { validateFunction, validateUint32 } from 'node-internal:validators';
+import { ERR_MISSING_ARGS } from 'node-internal:internal_errors';
+
+const kType = Symbol('type');
+
+export class Dirent {
+  public name: string;
+  public parentPath: string;
+  public [kType]: number;
```
anonrig, line 19: Is this exposed in nodejs implementation as well?

src/node/internal/internal_fs_callback.ts
```diff
@@ -0,0 +1,1373 @@
...
+
+function callWithSingleArgCallback<T>(
+  fn: () => T,
+  callback: undefined | SingleArgCallback<T>
+): void {
+  if (typeof callback !== 'function') {
+    throw new ERR_INVALID_ARG_TYPE('callback', ['function'], callback);
+  }
+  try {
+    const result = fn();
```
anonrig, line 66: Shouldn't we execute this on the queue micro task, rather than doing sync work and returning the result in queue micro task?

src/node/internal/internal_fs_callback.ts
```diff
@@ -0,0 +1,1373 @@
...
+    mode = modeOrCallback;
+  }
+  // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
+  callWithErrorOnlyCallback(() => fssync.accessSync(path, mode), callback);
+}
+
+export type ExistsCallback = (result: boolean) => void;
+
+export function exists(path: FilePath, callback: ExistsCallback): void {
+  // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
anonrig, line 100: This can be disabled across the whole file rather than one by one.

src/node/internal/internal_fs_callback.ts
```diff
@@ -0,0 +1,1373 @@
...
+    callback
+  );
+}
+
+export function chmod(
+  path: FilePath,
+  mode: number,
+  callback: ErrorOnlyCallback
+): void {
+  // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
anonrig, line 144: Is there no validation for input, in all these functions? What about adding validate number to mode?

src/node/internal/internal_fs_callback.ts
jasnell: Yeah, I know. I did it this way so that I can easily search for them later.

src/node/internal/internal_fs_callback.ts
jasnell: The `fssync.chmodSync` will apply input validation. If we also performed input validation here we'd be duplicating.

src/node/fs.ts
```diff
@@ -0,0 +1,25 @@
...
+export * from 'node-internal:internal_fs_sync';
+export { constants, promises, Dirent, Dir, WriteStream, ReadStream };
+
+export default {
+  constants,
+  promises,
+  Dirent,
+  Dir,
+  WriteStream,
+  ReadStream,
```
guybedford, line 22: What's the situation with `FileReadStream` and `FileWriteStream`? It might be worth adding a comment about these.

src/node/internal/internal_fs_sync.ts
```diff
@@ -0,0 +1,1066 @@
...
+  path: FilePath,
+  mode: number = F_OK,
+  followSymlinks: boolean
+): void {
+  validateUint32(mode, 'mode');
+
+  // If the X_OK flag is set we will always throw because we don't
+  // support executable files.
+  if (mode & X_OK) {
+    throw new Error('access denied'); // not executable.
```
guybedford, line 55: It would be great to see the error system developed here, as it's critical to most FS workflows. Left some notes on the C++ side as well but completely understood if it's a follow-on.

src/node/internal/internal_fs_sync.ts
```diff
@@ -0,0 +1,1066 @@
...
+    return time;
+  }
+  throw new ERR_INVALID_ARG_TYPE(
+    'time',
+    ['string', 'number', 'bigint', 'Date'],
+    time
+  );
+}
+
+export function futimesSync(
```
guybedford, line 299: What about `glob` / `globSync`? At the very least could be worth having a comment / placeholder.

src/node/internal/internal_fs_sync.ts
```diff
@@ -0,0 +1,1066 @@
...
+  validateObject(options, 'options');
+  const { encoding = 'utf8', bufferSize = 32, recursive = false } = options;
+  validateEncoding(encoding, 'options.encoding');
+  validateUint32(bufferSize, 'options.bufferSize');
+  validateBoolean(recursive, 'options.recursive');
+  path = normalizePath(path);
+  throw new Error('Not implemented');
+}
+
+export function openSync(
```
guybedford, line 451: What about `openAsBlob`? Again comment or placeholder might help.

src/node/internal/internal_fs_sync.ts
```diff
@@ -0,0 +1,1066 @@
...
+    throw new Error('failed to create temporary directory');
+  }
+  return ret;
+}
+
+export type OpenDirOptions = {
+  encoding?: string;
+  bufferSize?: number;
+  recursive?: boolean;
+};
```
guybedford, line 432:
```suggestion
export type OpenDirOptions = {
  encoding?: string | undefined;
  bufferSize?: number | undefined;
  recursive?: boolean;
};
```

src/node/internal/internal_fs_utils.ts
```diff
@@ -0,0 +1,572 @@
...
+  // issues that real file system paths have and don't need to
+  // worry quite as much about strictly checking for null bytes
+  // in the path. The URL parsing will take care of those details.
+  // We do, however, need to be sensitive to the fact that there
+  // are two different URL impls in the runtime that are selected
+  // based on compat flags. A worker that is using the legacy URL
+  // implementation will end up seeing slightly different behavior
+  // here but that's not something we need to worry about for now.
+  if (typeof path === 'string') {
+    return new URL(path, 'file://');
```
guybedford, line 95: See the note about normalization in the C++ section.

src/node/internal/internal_fs_utils.ts
```diff
@@ -0,0 +1,572 @@
...
+  return options;
+}
+
+// converts Date or number to a fractional UNIX timestamp
+export function toUnixTimestamp(
+  time: string | number | Date,
+  name: string = 'time'
+): string | number | Date {
+  // @ts-expect-error TS2367 number to string comparison error.
+  if (typeof time === 'string' && +time == time) {
```
guybedford, line 238: It might be nice to be more explicit here, as it's difficult to follow the logic. `Number(time).toString() === time` perhaps?

src/node/internal/internal_fs_promises.ts
```diff
@@ -0,0 +1,323 @@
...
+}
+
+export async function watch(): Promise<void> {
+  throw new Error('Not implemented');
+}
+
+export async function writeFile(
+  _path: FilePath,
+  _data: unknown,
+  options: Record<string, unknown>
```
guybedford, line 309: Maybe use `WriteFileOptions` from `import type { WriteFileOptions } from 'node:fs'`?

src/node/internal/internal_fs_sync.ts
guybedford: Alternatively - `import type { OpenDirOptions } from 'node:fs'`?

Or even using `Partial<OptionDirOptions>` on the function params.

src/node/internal/internal_fs_sync.ts
```diff
@@ -0,0 +1,1066 @@
...
+  // We do not implement the maxRetries or retryDelay options in any meaningful
+  // way. We just validate them.
+  validateUint32(maxRetries, 'options.maxRetries');
+  validateBoolean(recursive, 'options.recursive');
+  validateUint32(retryDelay, 'options.retryDelay');
+
+  cffs.rm(normalizePath(path), { recursive, force: false, dironly: true });
+}
+
+export type RmSyncOptions = {
```
guybedford, line 735: This can just be `RmOptions`, or it can be `import type { RmOptions } from 'node:fs'`.

Again, partial support would be good (`force?: boolean | undefined` etc).

src/node/internal/internal_fs_utils.ts
```diff
@@ -0,0 +1,572 @@
...
+
+    if (stat.device) {
+      this.mode |= S_IFCHR;
+    } else {
+      switch (stat.type) {
+        case 'file':
+          this.mode |= S_IFREG;
+          break;
+        case 'directory':
+          this.mode |= S_IFDIR;
```
guybedford, line 488: Should we also add execution permissions to directories here? Seems like that would usually be expected on a unix filesystem.

guybedford (approved):
Made an initial pass. Great start! The bigger comments about path encoding and errors can certainly be follow-ons, but I also think it's important to get these right early too.

src/node/internal/internal_fs_utils.ts
jasnell: We explicitly are not supporting execution permissions. If you call `access(...)` specifying X_OK it'll throw. If we ever do add any notion of execution permissions we can revisit this.

src/node/internal/internal_fs_utils.ts
jasnell: This method isn't likely to be retained actually.

src/node/internal/internal_fs_sync.ts
jasnell: I'll be adding glob/globSync separately. Since this isn't going to be exposed to production yet (it requires the experimental flag) we don't need the full API surface present here yet.

src/node/internal/internal_fs_sync.ts
jasnell: Planning to open that as a separate PR

src/node/internal/internal_fs_utils.ts
jasnell: It will be removed in the next set of changes that update the promises file. So for now going to mark this resolved

src/node/internal/internal_fs_sync.ts
jasnell: Will be working on this incrementally as we go. This PR has some of these fixed up but most likely most of these will be done in separate PRs before this is made public.

src/node/fs.ts
jasnell: These will be added in a separate PR before this goes public.

src/workerd/api/filesystem.c++
```diff
@@ -1,11 +1,816 @@
...
+            // If a symbolic link is returned here then the options.followSymLinks
+            // must have been set to false.
+            return Stat(link->stat(js));
+          }
+        }
+        KJ_UNREACHABLE;
+      }
+    }
+    KJ_CASE_ONEOF(fd, int) {
+      auto& opened = JSG_REQUIRE_NONNULL(vfs.tryGetFd(js, fd), Error, "Bad file descriptor"_kj);
```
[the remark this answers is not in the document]
jasnell: Yep, the intention is to implement all of this before it goes public. I've done some here but will do more in follow on PRs

src/workerd/api/filesystem.c++
```diff
@@ -1,11 +1,816 @@
...
+      KJ_CASE_ONEOF(standard, jsg::Ref<url::URL>) {
+        return *standard;
+      }
+    }
+    KJ_UNREACHABLE;
+  }
+
+  operator const kj::Path() const {
+    const jsg::Url& url = *this;
+    auto path = kj::str(url.getPathname().slice(1));
```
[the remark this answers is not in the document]
jasnell: Yep, planning to tackle this before removing fs from experimental. Likely in a separate PR thol

src/node/internal/internal_fs.ts
jasnell: It's visible when the Dirent is inspected but otherwise no because the symbol is not exported. I made it private.

jasnell:
There is still a ton of work to do around this but I'd like to get this bit in then keep iterating in another PR.

src/node/internal/internal_fs_utils.ts
guybedford: I specifically meant for directories here - as in execution permissions to mean `readdir` permissions.

PR #3799, opened 2025-03-26
Add query level observability for SRS DO

ramyabygari-cf:


ramyabygari-cf:

ramyabygari-cf:
recheck

src/workerd/util/sqlite.h
```diff
@@ -522,11 +544,64 @@ class SqliteDatabase::Query final: private ResetListener {
     }
   }
 
+  struct QueryEvent {
```
MellowYarker, line 547: This should be a class. Keep in mind everything is public in structs by default, so the setter methods aren't really enforceable unless your properties are designated `private`.

src/workerd/util/sqlite.h
```diff
@@ -522,11 +544,64 @@ class SqliteDatabase::Query final: private ResetListener {
...
+          dbWalSizeBefore(observer.getDbWalSize()),
+          startTime(observer.now()) {}
+
+    ~QueryEvent() {
+      uint64_t dbWalSizeAfter = observer.getDbWalSize();
+      uint64_t dbWalBytesWritten = (dbWalSizeAfter - dbWalSizeBefore);
+      kj::Duration queryLatency = observer.now() - startTime;
+
+      observer.reportQueryEvent(/*queryStatement=*/queryStatement,
+          /*queryRowsRead=*/rowsRead,
```
MellowYarker, line 560: Not sure the comments give us much info here. The variable names are already descriptive enough.

src/workerd/util/sqlite.c++
```diff
@@ -1296,6 +1298,14 @@ void SqliteDatabase::Query::destroy() {
     db.sqliteObserver.addQueryStats(rowsRead, rowsWritten);
   }
 
+  queryEvent.setQueryEventStats(rowsRead, rowsWritten, !(regulator.shouldAddQueryStats()));
+  try {
+    auto& statementAndEffect = getStatementAndEffect();
```
MellowYarker, line 1303: nit: Do we need the temporary `statementAndEffect` variable?

src/workerd/util/sqlite.h
```diff
@@ -522,11 +544,58 @@ class SqliteDatabase::Query final: private ResetListener {
...
   }
 
+  class QueryEvent {
+   public:
+    explicit QueryEvent(Query& q)
+        : observer(q.db.sqliteObserver),
+          dbWalSizeBefore(observer.getDbWalSize()),
+          startTime(observer.now()) {}
+
+    ~QueryEvent() {
```
jclee, line 554: Destructor should probably be tagged `noexcept(false)`?

src/workerd/util/sqlite.c++
```diff
@@ -1296,6 +1298,13 @@ void SqliteDatabase::Query::destroy() {
     db.sqliteObserver.addQueryStats(rowsRead, rowsWritten);
   }
 
+  queryEvent.setQueryEventStats(rowsRead, rowsWritten, !(regulator.shouldAddQueryStats()));
+  try {
+    queryEvent.setQueryStatement(sqlite3_sql(getStatementAndEffect().statement));
+  } catch (kj::Exception& e) {
+    queryEvent.setQueryErrorDescription(e.getDescription());
+  }
```
jclee, line 1306: It looks like the current implementations of `setQueryStatement()` and `setQueryErrorDescription()` store their StringPtr arguments, so they are assuming that the values being pointed to outlive the queryEvent object.  I'm not sure about the return value of `sqlite3_sql()`, but I think the `e.getDescription()` value at least will become invalid at the end of the catch block.

So, I think `setQueryStatement()` and `setQueryErrorDescription()` should be modified to copy their parameters into `kj::String` members, instead.

src/workerd/util/sqlite.h
```diff
@@ -522,11 +544,58 @@ class SqliteDatabase::Query final: private ResetListener {
     }
   }
 
+  class QueryEvent {
+   public:
```
MellowYarker, line 548: TIL we half-indent `public/private`, apparently we've done this for 5 months and I literally never noticed :laughing:

MellowYarker (approved):

src/workerd/util/sqlite.h
jclee: Heh... yeah.  IIRC, we do that because it makes git output more relevant context lines at the top of diff blocks, like `@@ -522,11 +544,58 @@ class SqliteDatabase::Query [...]` above.

src/workerd/util/sqlite.h
```diff
@@ -522,11 +544,58 @@ class SqliteDatabase::Query final: private ResetListener {
...
+    }
+
+    void setQueryResult(int res) {
+      queryResult = res;
+    }
+
+   private:
+    SqliteObserver& observer;
+    kj::Maybe<kj::String> queryStatement = kj::none;
+    bool isInternalQuery;
```
jclee, line 584: I think in some cases the compiler is allowed to leave primitive fields like `int` and `bool` uninitialized, if there is not an explicit value given in either the constructor or a member initializer. So it might not be a bad idea to initialize `isInternalQuery = false` here.

src/workerd/util/sqlite.h
```diff
@@ -522,11 +544,58 @@ class SqliteDatabase::Query final: private ResetListener {
     }
   }
 
+  class QueryEvent {
```
jclee, line 547: I think, as currently implemented, `QueryEvent` can actually be moved to the private section of `Query`?  Not sure if we have any style preferences around ordering, but if not, might just be sufficient to move the `private:` label from below the class to above it.

I guess we also have the option of just refactoring `QueryEvent`'s implementation into `Query`, but I assume they are currently separate for conceptual purposes, which seems fine, too.

src/workerd/util/sqlite.c++
```diff
@@ -1413,6 +1422,7 @@ void SqliteDatabase::Query::nextRow(bool first) {
 
   SQLITE_CALL_SCOPE {
     int err = sqlite3_step(statement);
+    queryEvent.setQueryResult(err);
```
jclee, line 1425: It looks like there's one other place in the code that also runs `sqlite3_step()`, as part of multistatement stored procedure setup...  I'm not sure of the exact circumstances of when that case gets hit.  But from the surrounding comments, it sounds like we run it for all but the last query in the sequence.  So I assume we end up relying on the last query being run in order to record the metrics for the sequence, and that we're OK with maybe not capturing query-level observability for the intermediate queries.

src/workerd/util/sqlite.h
```diff
@@ -26,11 +26,33 @@ using kj::uint;
...
   virtual void addQueryStats(uint64_t rowsRead, uint64_t rowsWritten) {}
   // The method is not used by the SqliteDatabase, it is added here for convenience
   virtual void setSqliteStoredBytes(uint64_t sqliteStoredBytes) {}
 
+  virtual void reportQueryEvent(kj::Maybe<kj::String> queryStatement,
+      uint64_t queryRowsRead,
+      uint64_t queryRowsWritten,
+      kj::Duration queryLatency,
+      uint64_t dbWalBytesWritten,
+      int queryError,
```
jclee, line 47: It looks like this parameter value is referred to as `queryResult` in other code?

src/workerd/util/sqlite.c++
ramyabygari-cf: Yes thank you for this!
Emitted an at the other sqlite3_step(), because it was a multi-statement
Updated the test to cover this!

src/workerd/util/sqlite.h
ramyabygari-cf: Moved the private above!

src/workerd/util/sqlite.h
```diff
@@ -523,10 +545,66 @@ class SqliteDatabase::Query final: private ResetListener {
...
+  class QueryEvent {
+   public:
+    explicit QueryEvent(Query& q)
+        : observer(q.db.sqliteObserver),
+          dbWalSizeBefore(observer.getDbWalSize()),
+          startTime(observer.now()),
+          reported(false) {}
+
+    ~QueryEvent() noexcept(false) {
+      if (!reported) reportEvent();
```
ramyabygari-cf, line 557: Added this because of https://github.com/cloudflare/workerd/blob/main/src/workerd/util/sqlite.h#L565

jclee: Does this end up reporting an (empty?) event, in the case where a Query is set up, but then destroyed without calling `nextRow()` (and thus setQueryEventStats()/setQueryStatement()/setQueryErrorDescription()/setQueryResult()`)?

ramyabygari-cf: But prior to this chane as well, Query would be destroyed and an empty event would be emitted?

src/workerd/util/sqlite.c++
```diff
@@ -720,7 +720,30 @@ SqliteDatabase::StatementAndEffect SqliteDatabase::prepareSql(const Regulator& r
...
+            kj::Duration queryLatency = sqliteObserver.now() - start;
+            auto dbWalBytesWritten = sqliteObserver.getDbWalSize() - dbWalSizeBefore;
+            auto rowsRead = sqlite3_stmt_status(result, LIBSQL_STMTSTATUS_ROWS_READ, 0);
+            auto rowsWritten = sqlite3_stmt_status(result, LIBSQL_STMTSTATUS_ROWS_WRITTEN, 0);
+
+            kj::Maybe<kj::String> queryStatement;
+            kj::Maybe<kj::String> queryErrorDescription;
+
+            try {
+              queryStatement = kj::str(sqlite3_sql(result));
```
jclee, line 737: I guess this might benefit from similar truncation of the query and error as is being done in `SqliteDatabase::Query::destroy()`.

(It might also be possible to factor them together in some way?)

src/workerd/util/sqlite.h
ramyabygari-cf: I think this should be : Because Query::destroy() is responsible for reporting QueryEvent after setQueryEventStats(), setQueryStatement(), setQueryErrorDescription(), or setQueryResult()
~QueryEvent() noexcept(false) { it
  KJ_DASSERT(reported, "QueryEvent was destroyed without reporting!");
}

src/workerd/util/sqlite.c++
ramyabygari-cf: Oh yes! Forgot about it

src/workerd/util/sqlite.h
jclee: Oh, that's right, I suppose it would have also done it prior to the change.  But if capturing empty metrics is intentional, that's fine...  Just wanted to be sure.

Yeah, that `KJ_ON_SCOPE_FAILURE(destroy());` line in the constructor(s) is a bit unusual, but I suspect that your original code would have worked OK, too...  As I understand it, the order of execution during construction of a class instance is:
- the constructor of each class member, from top to bottom
- the constructor body

If at any point during construction something throws, execution unwinds out of the throwing scope, then destructors of everything that was successfully constructed are run in reverse order.  So, yeah, it's possible for something in the Query constructor body to throw, such that its destructor body doesn't run.  But by the time execution has reached the Query constructor body, the member variables of Query have already been successfully constructed, including QueryEvent.  So if something later throws in the Query constructor body, it will run `Query::destroy()` as part of the constructor's scope cleanup, but then the destructors of the member variables will run in reverse order, including `~QueryEvent()`, as part of unwinding the exception.  So if the `QueryEvent` was successfully constructed, we're guaranteed that `~QueryEvent` will also be run as part of cleaning up `Query`, even if `Query`'s destructor body is not run.

jclee: > This could be : Because Query::destroy() is responsible for reporting QueryEvent after setQueryEventStats(), setQueryStatement(), setQueryErrorDescription(), or setQueryResult() ~QueryEvent() noexcept(false) { it KJ_DASSERT(reported, "QueryEvent was destroyed without reporting!"); }

Sure, that would work, too.

ramyabygari-cf: Oh yes, this makes sense
Reverted it to how it was prior!

src/workerd/util/sqlite.c++
```diff
@@ -83,6 +83,8 @@ kj::String namedErrorCode(int errorCode) {
 #undef LITERAL
 }
 
+constexpr ssize_t RA_MAX_METRICS_QUERY_SIZE = 1024;
```
jclee, line 86: Huh... I actually never noticed we had a `ssize_t` available, which I guess is the signed version of `size_t`.  Was there something that required using it?  I think the general convention is to use `size_t` for sizes.

src/workerd/util/sqlite.c++
```diff
@@ -720,7 +722,33 @@ SqliteDatabase::StatementAndEffect SqliteDatabase::prepareSql(const Regulator& r
...
+
+            kj::Maybe<kj::String> queryStatement;
+            kj::Maybe<kj::String> queryErrorDescription;
+            try {
+              kj::StringPtr statement = sqlite3_sql(result);
+              queryStatement = kj::heapString(
+                  statement.slice(0, kj::min(statement.size(), RA_MAX_METRICS_QUERY_SIZE)));
+            } catch (kj::Exception& e) {
+              kj::StringPtr errorDescription = e.getDescription();
+              queryErrorDescription = kj::heapString(errorDescription.slice(
```
jclee, line 743: Oh, good catch!  I could have sworn `kj::StringPtr::slice()` had a built-in `min()` behavior, but looking through the implementation, I see it just has debug-only bounds checks.

src/workerd/util/sqlite.h
```diff
@@ -523,10 +545,57 @@ class SqliteDatabase::Query final: private ResetListener {
   }
 
  private:
+  class QueryEvent {
+   public:
+    explicit QueryEvent(Query& q)
+        : observer(q.db.sqliteObserver),
```
jclee, line 551: I guess that since the class now only uses the `sqliteObserver` field, it could be made slightly more generic by having the constructor take `SqliteObserver&` directly instead of `Query&`.  Not a big deal, though, either way.

jclee (approved):

src/workerd/util/sqlite.c++
ramyabygari-cf: Changed it because vscode highlighted including a new header. Now looking at the headers #include<unistd.h>, where ssize_t is defined, is conditionally included for non-Windows platforms. Probably thats why the windows build kept failing?
Should you have just ignored the warning on vscode

jclee: If I'm understanding correctly, you tried to add a use of `size_t`, and vscode tried to suggest including an additional header to support it?

In that case, I'd expect that it's trying to enforce a code style guideline that might be helpful in other codebases -- maybe for portability?  But, yeah, I think that particular guideline can be ignored here in favor of the workerd/capnproto code style conventions.  Pretty much any `kj` header (maybe all?) will transitively include `kj/common.h`, which includes `stddef.h`, which defines `size_t`, so it should be OK to use pretty much everywhere in the codebase, without separately including `stddef.h`.

vscode probably isn't aware of the transitive inclusion, or it's trying to enforce some variation of "include-what-you-use" that the kj code style doesn't adhere closely to.

PR #3802, opened 2025-03-26
[o11y] Expand and refactor STW support

fhanau:
- Add rudimentary user tracing support
- Add support for fetch response info, ancillary fields like script tags
- Extend lifetime of outcome event for fetch handler so that exception outcome can be reported correctly.

Upstream PR to follow.

### DIFF
--- a/src/workerd/io/tracer.c++
+++ b/src/workerd/io/tracer.c++
@@ -4,6 +4,8 @@
 
 #include <workerd/io/tracer.h>
 
+#include <capnp/message.h>  // for capnp::clone()
+
 namespace workerd {
 
 namespace {
@@ -55,7 +57,7 @@ void PipelineTracer::addTracesFromChild(kj::ArrayPtr<kj::Own<Trace>> traces) {
 }
 
 void PipelineTracer::addTailStreamWriter(kj::Own<tracing::TailStreamWriter>&& writer) {
-  maybeTailStreamWriter = kj::mv(writer);
+  tailStreamWriters.add(kj::mv(writer));
 }
 
 kj::Promise<kj::Array<kj::Own<Trace>>> PipelineTracer::onComplete() {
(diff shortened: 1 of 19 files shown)


src/workerd/io/tracer.c++
```diff
@@ -278,8 +312,19 @@ void WorkerTracer::setEventInfo(
   trace->bytesUsed = newSize;
 
   KJ_IF_SOME(writer, maybeTailStreamWriter) {
-    writer->report(
-        context, tracing::Onset(cloneEventInfo(info), tracing::Onset::WorkerInfo{}, kj::none));
+    // Provide WorkerInfo to the streaming tail worker if available. This data is provided when the
+    // WorkerTracer is created, but the actual onset event is the best time to send it.
+    auto workerInfo = tracing::Onset::WorkerInfo{
+      .executionModel = trace->executionModel,
+      .scriptName = trace->scriptName.map([](kj::String& s) { return kj::str(s); }),
```
fhanau, line 319: Is there a more elegant way to map a `kj::Maybe<kj::String>`? Also do we have a way to clone capnp readers?

danlapid: We do this in a bunch of places, I think it's a good idea to add this to own-utils as another overload to `mapCopyString`

src/workerd/io/tracer.h
```diff
@@ -91,7 +91,8 @@ class PipelineTracer final: public kj::Refcounted, public kj::EnableAddRefToThis
   friend class WorkerTracer;
 
  public:
-  kj::Maybe<kj::Own<tracing::TailStreamWriter>> maybeTailStreamWriter;
+  // tail stream writers for worker stages in the given pipeline.
+  kj::Vector<kj::Own<tracing::TailStreamWriter>> tailStreamWriters;
```
danlapid, line 95: I gotta say that this feels odd as a vector, should it really grow and shrink in size throughout the lifetime of the PipelineTracer?
Seems to me like this should be a kj::Array that is passed at construction time.

jasnell (approved):
Generally lgtm but I agree with @danlapid's comment about the vector vs. array.

src/workerd/io/tracer.h
fhanau: The vector is needed here since tail stream writers are added to the pipeline tracer after it is created, and if there are several workers to be added in the downstream implementation we may not know how many workers need to be added in advance.

PR #3808, opened 2025-03-27
Python: Move a lot of bazel setup into helpers.bzl

hoodmane:
This just moves the code from one file to another, split from https://github.com/cloudflare/workerd/pull/3807 to make the diff easier to read.


src/pyodide/BUILD.bazel
```diff
@@ -50,236 +50,4 @@ cc_capnp_library(
...
-        "$(location :pyodide.capnp)",
-        "pyodideBundle",
-        "-I $$INCLUDE",
-        "-o binary",
-        "> $@",
-    ]),
-    tools = ["@capnp-cpp//src/capnp:capnp_tool"],
-    visibility = ["//visibility:public"],
-)
+main()
```
dom96, line 53: Maybe if we could separate everything that is in that `main` function into multiple functions, but otherwise I think it's better if it's just inside this file.

hoodmane: See https://github.com/cloudflare/workerd/pull/3807 where it becomes useful that this is a function.

PR #3811, opened 2025-03-27
Worker::numRequestsInFlight

mikea:
We do scheduling decision based on number requests in flight. We currently use the number of references for that but it is very fiddly and easy to break.

Track the number of requests in flight directly.

I was considering to make on* methods private, but that would require me to move WorkerEntrypoint out of anonymous namespace. LMK if you still prefer me to do that.

### DIFF
--- a/src/workerd/io/worker-entrypoint.c++
+++ b/src/workerd/io/worker-entrypoint.c++
@@ -57,6 +57,10 @@ class WorkerEntrypoint final: public WorkerInterface {
       kj::Maybe<kj::String> cfBlobJson,
       kj::Maybe<tracing::InvocationSpanContext> maybeTriggerInvocationSpan);
 
+  virtual ~WorkerEntrypoint() {
+    worker->onRequestFinished();
+  }
+
   kj::Promise<void> request(kj::HttpMethod method,
       kj::StringPtr url,
       const kj::HttpHeaders& headers,
@@ -78,6 +82,7 @@ class WorkerEntrypoint final: public WorkerInterface {
 
   // Members initialized at startup.
 
+  kj::Arc<Worker> worker;
   ThreadContext& threadContext;
   kj::TaskSet& waitUntilTasks;
   kj::Maybe<kj::Own<IoContext::IncomingRequest>> incomingRequest;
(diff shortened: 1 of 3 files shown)


src/workerd/io/worker-entrypoint.c++
```diff
@@ -174,46 +179,53 @@ kj::Own<WorkerInterface> WorkerEntrypoint::construct(ThreadContext& threadContex
...
+void WorkerEntrypoint::init(kj::Maybe<kj::Own<Worker::Actor>> actor,
     kj::Own<LimitEnforcer> limitEnforcer,
     kj::Own<void> ioContextDependency,
     kj::Own<IoChannelFactory> ioChannelFactory,
     kj::Own<RequestObserver> metrics,
     kj::Maybe<kj::Own<WorkerTracer>> workerTracer,
     tracing::InvocationSpanContext invocationSpanContext) {
   TRACE_EVENT("workerd", "WorkerEntrypoint::init()");
+
+  worker->onRequestStarted();
```
jp4a50, line 219: So I think originaly the initial `Worker` refcount increment for a request always happened when the correpsonding worker set shard was locked. If so, that means that whenever we checked `isShared()` whilst also holding a shard lock, we could guarantee that that `Worker` not only had no requests in-flight but also wouldn't be handed out to any requests until we release the lock. We could then remove the `WorkerInfo` from the worker set and destroy it and later (once we'd released the shard lock) drop what should be the last ref to the `Worker`, ensuring the worker is destroyed on the GC thread. I don't know how important it is that we continue to destroy the Worker on the GC thread but I think this behaviour was intentional and I vaguely remember talking to @harrishancock about it.

By moving this refcounting to here (by which time we will have dropped the shard lock), we can no longer guarantee this. I don't *think* that makes the code incorrect, per se, but I think it does mean that we may end up evicting a worker from the worker set for which a request is just starting (probably fine as an edge case?), and the worker object itself will be deleted on that connection thread instead of the GC thread.

kentonv:
Left a more detailed comment on the internal PR, but I don't think this is the right approach. In particular it misses Actors holding Workers with no requests currently in flight, and maybe it misses other things too that I'm not thinking of.

PR #3812, opened 2025-03-27
simplify websocket constructor url validation

anonrig:
Throwing and later catching is not necessary due to the existence of tryParse method.


danlapid (approved):
LGTM

PR #3813, opened 2025-03-27
Implement MessageChannel/MessagePort

jasnell:
This is an initial sketch of a MessageChannel/MessagePort impl. Why? Thank you for asking.. two reasons:

1. Node.js compat. There are a number of node.js ecosystem modules/frameworks that use `MessageChannel`.
2. WinterTC conformance. MessageChannel/MessagePort are being added to the WinterTC minimum common API.

This is a work in progress. PR will remain draft until all of the design/impl questions are answered but I'm opening this to start getting review.

```js
const { port1, port2 } = new MessageChannel();
port1.onmessage = (event) => console.log(event.data);
port2.postMessage('hello');
```

Or when used with JSRPC...

```js
// ...
   const { port1, port2 } = new MessageChannel();
   port1.postMessage('hello');
   port1.onmessage = console.log;
   await env.RPCThing.setup(port2);

// then on the other side...

export default {
  async setup(port) {
    port.onmessage = (event) => console.log(event.data);
    port.postMessage('world');
  }
}
```

There are a range of open questions here that need to be figured out before this lands:

1. Currently, when sending a port over rpc, the local version of that port is still usable. This is because jsrpc does not necessarily imply transfer semantics. So in the above jsrpc example, port2 is still usable locally and we actually have a one-to-many relationship now from port1 to port2 and port1 to the remote port on the rpc remote. That feels a bit weird.

4. I've no idea if the underlying capnp rpc mechanism here is entirely correct. Is the output gate usage correct? Do we need a registerPendingEvent in here somewhere, etc. Specifically, should both sides of the jsrpc be kept alive so long as the message ports are connected and not closed? There are a range of behavioral things we should consider here. Simply... how do we expect things to behave where there are active ports spread across the rpc connection?

5. Right now the message cloning does not support the extended types like ReadableStream and WritableStream. It probably should? But, should that also be supported if we're not cloning across a jsrpc boundary and everything is just local. Specifically, what should the cloning behavior be in the local-to-local case vs. the local-to-remote-rpc case? 

6. This does not implement the transfer list at all, is there a way to do so reasonably?

This still needs a lot more tests, so the PR will remain draft until the tests are expanded and more of these questions are answered.


src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// Deliver the message to this port, buffering if necessary if the port
+// has not been started. Buffered messages will be delivered when the
+// port is started later.
+void MessagePort::deliverMessage(jsg::Lock& js, rpc::JsValue::Reader value) {
+  KJ_SWITCH_ONEOF(state) {
+    KJ_CASE_ONEOF(pending, Pending) {
+      // We have not yet started the port so buffer to message.
+      pending.add(capnp::clone(value));
```
jasnell, line 107: TODO: Tie this into isolate memory accounting so we don't over-buffer.

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+  });
+}
+
+// Deliver the message to all the jsrpc remotes we have
+kj::Promise<void> MessagePort::sendToRpc(kj::Own<rpc::JsValue::Reader> message) {
+  KJ_IF_SOME(outputLocks, ioContext.waitForOutputLocksIfNecessary()) {
+    co_await outputLocks;
+  }
+  kj::Vector<kj::Promise<void>> promises;
+  for (rpc::JsMessagePort::Client cap: rpcClients) {
```
jasnell, line 91: Why multiple here? It's because currently sending a port over jsrpc does not have transfer semantics... that is, a single MessagePort could be passed to multiple rpc endpoints but still needs to remain usable... consider for instance the following example:

```
function foo(port) {
  // ...
}

function bar(port) {
  // ...
}

const { port1, port2 } = new MessageChannel();

foo(port2);
bar(port2);
port2.postMessage('3');
```

In this case, `port2` is passed by reference to each of `foo(...)` and `bar(...)` and remains usable after those calls. Now imagine if we refactored this so that `foo` and `bar` were jsrpc calls...

```
await env.FOO.foo(port2);
await env.BAR.bar(port2);
port2.postMessage('3');
```

Should the same assumption be held here that `port2` is still usable after the jsrpc call?  If these were streams, then passing the readable stream or writable stream to the jsrpc call as an argument ends up locking the local objects because those are being fully consumed by the jsrpc mechanism but it's not clear if the same rule would hold for a `MessagePort`.

If we are going to allow a `MessagePort` to be usable after being passed as a jsrpc argument, then we need to support multiple jsrpc capabilities and end up with a one-to-many relationship with the ports, which feels weird. But if we transfer the `MessagePort` the example above becomes impossible because `port2` would not be usable any longer after the call to `env.FOO.foo(...)`. Is that ok?

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+      KJ_CASE_ONEOF(opts, PostMessageOptions) {
+        KJ_IF_SOME(list, opts.transfer) {
+          JSG_REQUIRE(list.size() == 0, Error, "Transfer list is not supported");
+        }
+      }
+    }
+  }
+
+  KJ_IF_SOME(o, other) {
+    // TODO(message-port): Set up the external handler to support more types.
```
jasnell, line 148: This is currently not set up to allow for other types we would send over jsrpc but it can be. Should it be? What if there are no jsrpc targets and the `MessagePort` is just cloning locally? Should we still use the jsrpc external handler in that case? It's not clear... so initially this will only take regularly serializable objects.

src/workerd/api/messagechannel.c++
anonrig: MessagePort is a transferable object. My guess is after jsrpc call, MessagePort becomes detached and can not be reusable. IMHO this shouldn't change regardless of the underlying implementation details

Referring to the spec:

> [Platform objects](https://webidl.spec.whatwg.org/#dfn-platform-object) that are [transferable objects](https://html.spec.whatwg.org/multipage/structured-data.html#transferable-objects) have a [[Detached]] internal slot. This is used to ensure that once a platform object has been transferred, it cannot be transferred again.

Ref: https://html.spec.whatwg.org/multipage/structured-data.html#transferable

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+// Deliver the message to all the jsrpc remotes we have
+kj::Promise<void> MessagePort::sendToRpc(kj::Own<rpc::JsValue::Reader> message) {
+  KJ_IF_SOME(outputLocks, ioContext.waitForOutputLocksIfNecessary()) {
+    co_await outputLocks;
+  }
+  kj::Vector<kj::Promise<void>> promises;
+  for (rpc::JsMessagePort::Client cap: rpcClients) {
+    auto req = cap.callRequest();
+    req.setData(*message);
+    // TODO(message-port): Removing the port when dropped?
```
anonrig, line 94: You're referencing to the case where MessagePort is removed and the existing promises needs to be cancelled right?

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+    global.reportError(js, jsg::JsValue(exception.getHandle(js)));
+  });
+}
+
+// Deliver the message to all the jsrpc remotes we have
+kj::Promise<void> MessagePort::sendToRpc(kj::Own<rpc::JsValue::Reader> message) {
+  KJ_IF_SOME(outputLocks, ioContext.waitForOutputLocksIfNecessary()) {
+    co_await outputLocks;
+  }
+  kj::Vector<kj::Promise<void>> promises;
```
anonrig, line 90: Maybe reserve the size?

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+  for (rpc::JsMessagePort::Client cap: rpcClients) {
+    auto req = cap.callRequest();
```
anonrig, line 92: Question for educational purposes: Why don't we use prewarmRequest() in here?

src/workerd/api/messagechannel.c++
anonrig: Does this cover transfer property of postMessage() function? if we transfer, we should move rather than clone.

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+    KJ_CASE_ONEOF(_, Closed) {
+      // Nothing to do in this case. Drop the message on the floor.
+    }
+  }
+}
+
+// Binds two ports to each other such that messages posted to one
+// are delivered on the other.
+void MessagePort::entangle(MessagePort& port1, MessagePort& port2) {
+  port1.other = port2.addRef();
```
anonrig, line 122: This has a bug. If this port is entangled with another port, we should disentangle first. If not, port3 that is entangled with port1 will still be entangled with port1, even though port1 is entangled with port2.

Step 1 of entangle spec step: https://html.spec.whatwg.org/multipage/web-messaging.html#entangle

anonrig: Also disentangle should emit "close" event which in this case it never emits.

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+
+    capnp::MallocMessageBuilder builder;
+    rpc::JsValue::Builder val = builder.initRoot<rpc::JsValue>();
+    val.setV8Serialized(kj::mv(released.data));
+    o->deliverMessage(js, val.asReader());
+  }
+}
+
+// Should close this port, the entangle port, and any known rpc clients.
+void MessagePort::close() {
```
anonrig, line 164: We should also emit "close" event

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+  }
+}
+
+// Start delivering messages on this port. Any messages that are
+// buffered will be drained immediately.
+void MessagePort::start(jsg::Lock& js) {
+  KJ_SWITCH_ONEOF(state) {
+    KJ_CASE_ONEOF(pending, Pending) {
+      auto list = kj::mv(pending);
+      state = Started{};
```
anonrig, line 180: Nit: We should have a changeState() function that validates:

> A [port message queue](https://html.spec.whatwg.org/multipage/web-messaging.html#port-message-queue) can be enabled or disabled, and is initially disabled. Once enabled, a port can never be disabled again (though messages in the queue can get moved to another queue or removed altogether, which has much the same effect)

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+ public:
+  JsMessagePortImpl(IoContext& ctx, jsg::Ref<MessagePort> port)
+      : port(kj::mv(port)),
+        weakIoContext(ctx.getWeakRef()) {}
+  ~JsMessagePortImpl() {
+    port->close();
+  }
+
+  kj::Promise<void> call(CallContext context) override {
+    IoContext& ctx = JSG_REQUIRE_NONNULL(weakIoContext->tryGet(), Error,
```
anonrig, line 229: I think this should be a DOMException

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+
+void MessagePort::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
+  auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
+      "MessagePort can only be serialized for RPC.");
+  auto externalHandler = dynamic_cast<RpcSerializerExternalHander*>(&handler);
+  JSG_REQUIRE(
+      externalHandler != nullptr, DOMDataCloneError, "MessagePort can only be serialized for RPC.");
+
+  // Don't send a MessagePort that has been closed already.
+  JSG_REQUIRE(!state.is<Closed>(), DOMDataCloneError, "MessagePort is closed");
```
anonrig, line 261: This should be DOMException not DOMDataCloneError

> If [this](https://webidl.spec.whatwg.org/#this)'s [closed flag](https://html.spec.whatwg.org/multipage/web-messaging.html#concept-broadcastchannel-closed) is true, then throw an ["InvalidStateError"](https://webidl.spec.whatwg.org/#invalidstateerror) [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException).

anonrig (commented):
I think we should have a subset of MessagePort web-platform tests that validate this. If not, I'm afraid we're going to introduce yet another implementation that is not fully spec compliant.

src/workerd/api/basics.c++
```diff
@@ -269,9 +269,16 @@ void EventTarget::addEventListener(jsg::Lock& js,
 
     auto& set = getOrCreate(type);
 
+    KJ_DEFER({
+      KJ_IF_SOME(listener, maybeListenerCallback) {
+        listener(js, type, getHandlerCount(type));
+      } else {
+      }
```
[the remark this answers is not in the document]
jasnell: Squash compiler warning

src/workerd/api/messagechannel.c++
jasnell: no, if the capability is dropped on the remote side then we should drop the local one too. Right now I think this leaks. Will need to resolve this before this is ready to land.

src/workerd/api/messagechannel.c++
jasnell: That shouldn't ever happen. If anything, this should just assert if the port is already entangled, but in every use of this currently there's no way for the ports to already be entangled.

src/workerd/api/messagechannel.c++
jasnell: Yep, I forgot to add a comment but the `'close'` event is still an open TODO

src/workerd/api/messagechannel.c++
jasnell: `DOMDataCloneError` *is* a `DOMException`

src/workerd/api/messagechannel.c++
jasnell: I don't know what you mean by "cover transfer property of postMessage()"

jasnell:
> I think we should have a subset of MessagePort web-platform tests that validate this. If not, I'm afraid we're going to introduce yet another implementation that is not fully spec compliant.

That's a given. We won't be able to run the full web-platform test suite since it requires web workers in many cases. Before this is taken out of draft the tests will be greatly expanded and there will be more verifications that we are as close to spec conformance as possible.

src/workerd/api/messagechannel.c++
jasnell: This is following the pattern used elsewhere for other types.

src/workerd/api/messagechannel.h
```diff
@@ -0,0 +1,193 @@
...
+      JSG_READONLY_PROTOTYPE_PROPERTY(lastEventId, getLastEventId);
+      JSG_READONLY_PROTOTYPE_PROPERTY(source, getSource);
+      JSG_READONLY_PROTOTYPE_PROPERTY(ports, getPorts);
+      // The standard also defines the origin property, but we don't
+      // implement origin in this case so we leave it out.
+    }
+
+   private:
+    jsg::JsRef<jsg::JsValue> data;
+    kj::String lastEventId;
```
danlapid, line 69: Nit: why not just have this be a JsString?
Just for the good measure of trying to remove kj types from our JSG resources

jasnell: Yeah, I'm coming along more and more to the idea of this more and more...

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+  // will switch to the Started state if necessary.
+  setEventListenerCallback([&](jsg::Lock& js, kj::StringPtr name, size_t count) {
+    if (name == "message"_kj) {
+      KJ_SWITCH_ONEOF(state) {
+        KJ_CASE_ONEOF(pending, Pending) {
+          // If we are in the pending state, start the port if we have listeners.
+          // This is technically not spec compliant, but it is what Node.js
+          // supports. Specifically, adding a new message listener using the
+          // addEventListener method is *technically* not supposed to start
+          // the port but we're going to do what Node.js does.
```
kentonv, line 43: Why would we do what Node.js does and not what browsers do? Shouldn't we follow standards first?

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  kj::Promise<void> call(CallContext context) override {
+    IoContext& ctx = JSG_REQUIRE_NONNULL(weakIoContext->tryGet(), Error,
+        "The destination object for this message port no longer exists.");
+
+    KJ_IF_SOME(other, port->getOther()) {
+      // We can only dispatch messages under the isolate lock, so acquire
+      // that here then deliver...
+      auto lock = co_await ctx.getWorker()->takeAsyncLockWithoutRequest(nullptr);
```
kentonv, line 235: Use `ctx.run` to enter the isolate. The code you have here is manually doing a lot of what `ctx.run` does, and is missing important things like setting the current `IoContext`, honoring input locks, etc.

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+}
+
+namespace {
+// The jsrpc handler that receives messages posted from the remote and
+// delivers them to the local port.
+class JsMessagePortImpl final: public rpc::JsMessagePort::Server {
+ public:
+  JsMessagePortImpl(IoContext& ctx, jsg::Ref<MessagePort> port)
+      : port(kj::mv(port)),
+        weakIoContext(ctx.getWeakRef()) {}
```
kentonv, line 223: You also need to call `ctx.registerPendingEvent()` and hold onto it as long as this object is alive, since this is a source of events that could cause JavaScript to run in the IoContext.

ALTERNATIVELY, you might consider a "pull" approach instead of a "push" approach: Have JsMessagePortImpl actually implement a producer-consumer queue. When messages arrive, just put them in the queue. Have a method `Promise<Message> next()` that returns the next message. Don't hold a reference to the IoContext at all.

Then, inside `MessagePort::start()`, you set up a loop that repeatedly calls `next()` and then delivers the event. This loop would use `ctx.awaitIo()` for each `next()` call.

The nice part about doing it this way is that you don't have to mess around with weak references and pending events. `awaitIo()` just does the right thing. You also don't need any complexity to wait for `start()` to be called before delivering events. The code ends up looking more like the WebSocket implementation, which I think fits pretty well here.

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,317 @@
...
+// port is started later.
+void MessagePort::deliverMessage(jsg::Lock& js, rpc::JsValue::Reader value) {
+  KJ_SWITCH_ONEOF(state) {
+    KJ_CASE_ONEOF(pending, Pending) {
+      // We have not yet started the port so buffer to message.
+      pending.add(capnp::clone(value));
+    }
+    KJ_CASE_ONEOF(started, Started) {
+      ioContext.addTask(sendToRpc(capnp::clone(value)));
+      deliver(js, capnp::clone(value), addRef());
```
kentonv, line 111: You're making a clone to pass into `deliver()`, but it appears `deliver()` does not hold onto the object beyond its own scope -- could it just accept a `Reader` instead of an `Own<Reader>`?

src/workerd/io/worker-interface.capnp
```diff
@@ -480,6 +486,16 @@ struct JsValue {
...
 }
 
+interface JsMessagePort $Cxx.allowCancellation {
+  struct Params {
+    data @0 :JsValue;
+  }
+
+  struct Results {}
+
+  call @0 Params -> Results;
```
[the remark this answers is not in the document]
kentonv: Hmm, but meanwhile it occurs to me that as written, it won't be possible to pass a `ReadableStream` -- or another `MessagePort` -- in a message, because the method you define here doesn't support `StreamSink` the way `JsRpcTarget.call` does.

This seems like a big problem! But supporting `StreamSink` is quite complicated and you probably don't want to be re-implementing it. You really want to share the implementation from worker-rpc.

I wonder if it would actually be easier to implement MessagePort in pure JavaScript on top of JS RPC. Barring that, it seems like we need to refactor worker-rpc to make the implementation more reusable for these kinds of use cases.

kentonv (commented):
I had started reviewing some code details but then at some point realized that unfortunately the overall design here might have a fatal flaw -- I don't think you can send a new MesasgePort across an existing MessagePort as designed, but I believe this is a requirement. See my first comment thread below... this might require a totally different design.

So... You can probably ignore the implementation comments since I suspect much of the implementation will have to change to solve this anyway...

src/workerd/api/messagechannel.c++
jasnell: A key motivator for this is node.js compat... specifically, there are some modules/frameworks that we want to be able to support that are currently written to Node.js' implementation of `MessageChannel` that we would need to reconcile.

src/workerd/io/worker-interface.capnp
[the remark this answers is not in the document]
jasnell: I'm going to mark these reviews resolved for now. Decided to remove the serialization/deserialization/jsrpc stuff from this for now so we can at least get the api available (we have customers asking that don't need the jsrpc stuff yet)

src/workerd/api/messagechannel.h
jasnell: Will address this in a separate follow up PR since the MessageEvent is now shared across multiple apis...

src/workerd/api/messagechannel.c++
jasnell: We can revisit this later if necessary.

anonrig (changes requested):
Can we put this behind an experimental flag or include the web platform tests, before landing it?

jasnell:
See the last commit in the set for discussion of web platform tests. We don't want to block on that indefinitely as we have a customer that needs this as soon as possible and the minimum of what we need is what they need to support their use of via undici. As for putting this behind an experimental, same. We will need this to be available outside of experimental.

anonrig:
> See the last commit in the set for discussion of web platform tests. We don't want to block on that indefinitely as we have a customer that needs this as soon as possible and the minimum of what we need is what they need to support their use of via undici. As for putting this behind an experimental, same. We will need this to be available outside of experimental.

If we're going to add web platform tests, which will require a compat flag to fix any WPT bug, I'm not sure if we should land this as is.

Regardless, of WPT, I think we put it behind a compat flag. This will likely break people since this is a global variable. And any application that checks if MessagePort includes will behave differently.

danlapid:
Does undici work now that we have this?

jasnell:
> Does undici work now that we have this?

That'll be the next validation point. We need to have the customer test it...

FWIW, undici does not actually *use* `MessagePort` directly, which is fun. It does implement it's own version of `MessageEvent` which checks for the existence of `MessagePort` when figuring out the webidl conversion to ensure that the `ports` property is implemented correctly, so I believe this should actually work just fine with undici... which does not *ACTUALLY* need it lol.

Also, keep in mind that there might be other reasons undici might not work, this is just the one we know for sure based on what the customer has asked for.

jasnell:
> If we're going to add web platform tests, which will require a compat flag to fix any WPT bug, I'm not sure if we should land this as is.

-1 on this. We can use a `pedantic_wpt*` flag later to fix up variances as needed. For now the customer need is going to trump that.

I will look at adding a compat flag for the global exposure.

anonrig (approved):
Thank you

jasnell:
While this should otherwise be ready to land, I want to hold off just a bit until we get confirmation from the customer that it's going to meet the need they have. (I also think I might need to tweak the internal branch to get internal tests working but I'm going to hold off on that until I get confirmation on the customer question).

src/workerd/api/messagechannel.c++
```diff
@@ -0,0 +1,200 @@
...
+      KJ_IF_SOME(list, opts.transfer) {
+        hasTransfer = list.size() > 0;
+      }
+    }
+  }
+  JSG_REQUIRE(!hasTransfer, Error, "Transfer list is not supported");
+
+  // If the port is closed, other will be kj::none and we will just drop the message.
+  other->runIfAlive([&](MessagePort& o) {
+    jsg::Serializer ser(js);
```
kentonv, line 112: At present this won't support sending RPC stubs. Perhaps that's fine? It would certainly add a lot of complexity which maybe we don't want to deal with?

We should make sure to test what happens if you try to send an RPC stub, or a derivative of RpcTarget. Need to make sure these cases fail, rather than, say, serializing them as plain objects dropping the prototype (which is normally the default behavior when serializing a class with structured clone).

Annoyingly, I suppose for regular user-defined classes that _don't_ inherit `RpcTarget`, MessagePort will need to maintain the standard behavior of serializing the object ignoring the prototype. In JSRPC, of course, we made this case an error because the standard behavior is confusing and unexpected, but MessagePort I suppose needs to follow the standard. :/ But since `RpcTarget` itself is non-standard, I think we can safely make it an error to try to send anything derived from `RpcTarget`, and we should do that because otherwise the standard behavior will be _very_ confusing for people who are expecting RPC semantics.

src/workerd/io/worker-interface.capnp
```diff
@@ -458,6 +460,10 @@ struct JsValue {
         }
       }
 
+      messagePort :group {
```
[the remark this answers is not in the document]
kentonv: If this code is staying here it still needs a doc comment. If we're abandoning the RPC support, let's delete this?

src/workerd/io/worker-interface.capnp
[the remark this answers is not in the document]
kentonv: Can we delete the changes to worker-interface.capnp then?

src/workerd/api/messagechannel.c++
jasnell: Yes, that's intentional here. I'm intentionally making this really just support regular structuredClone like behavior and taking all jsrpc stuff from this. Currently this will not support passing a `MessagePort` over a jsrpc and won't support anything other than regular ole cloneable objects. The idea is to just make the API available as a first step, even if it can only be used within a single isolate/worker (which obviously makes it fairly useless but believe it or not I've actually seen modules in the ecosystem using it this way... which is,.... weird, but ok).

As a separate follow-up step I plan to open a later PR that adds the jsrpc support.

src/workerd/io/worker-interface.capnp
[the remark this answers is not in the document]
jasnell: ah, yes, forgot about these, good catch :-)

src/workerd/api/messagechannel.c++
kentonv: OK but can you please verify that attempting to send a subclass of `RpcTarget` will throw?

Otherwise adding RPC support later will be a breaking change.

src/workerd/io/worker-interface.capnp
```diff
@@ -470,6 +470,10 @@ struct JsValue {
       # the one that will later on send the abort signal. This external will have an associated
       # stream in the corresponding `StreamSink` with type `AbortTrigger`.
 
+      messagePort :group {
+        out @8 :JsMessagePort;
+      }
+
```
jasnell:
```suggestion
```

src/workerd/io/worker-interface.capnp
```diff
@@ -509,6 +513,16 @@ interface AbortTrigger $Cxx.allowCancellation {
...
+interface JsMessagePort $Cxx.allowCancellation {
+  struct Params {
+    data @0 :JsValue;
+  }
+
+  struct Results {}
+
+  call @0 Params -> Results;
+}
+
```
jasnell:
```suggestion
```

src/workerd/api/messagechannel.c++
jasnell: Will do :-)

src/workerd/io/worker-interface.capnp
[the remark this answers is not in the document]
jasnell: Removed!

emily-shen (approved):

PR #3814, opened 2025-03-27
add types for event-subscriptions

sdnts:


sdnts:
@jbwcloudflare @maheshwarip

jbwcloudflare:
this looks pretty solid to me!

sdnts:

sdnts:
recheck

types/defines/event-subscriptions.d.ts
```diff
@@ -0,0 +1,80 @@
+type EventSubscriptionEvent<
+	Service extends string,
+	Type extends string,
+	Source,
+	Payload = {},
```
penalosa, line 5: I think this should probably default to `Record<string, never>` instead? (assuming this is meant to be an empty object)

types/defines/event-subscriptions.d.ts
```diff
@@ -0,0 +1,80 @@
...
+	  >
+	| EventSubscriptionEvent<"superSlurper", "jobAborted", { jobId: string }>;
+
+// Workflows
+type SubscriptionEventSourceWorkflows = {
+	workflowId: string;
+	workflowName: string;
+	versionId: string;
+	instanceId: string;
+};
```
penalosa, line 42: It's a bit weird that workflows has a helper for this, but superSlurper doesn't. Since this type will be exposed to users, I wonder if we should inline it?

PR #3815, opened 2025-03-27
remove authorization properties from url in websocket constructor

anonrig:
This is not a breaking change since we already disallow it but rather throw an Internal Error whenever a user passes a URL with username/password state.

After this change, we now throw a JSG error which will be propagated to the end-user in a clear way.

### DIFF
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -5,6 +5,7 @@
 #include "web-socket.h"
 
 #include "events.h"
+#include "util.h"
 
 #include <workerd/io/features.h>
 #include <workerd/io/io-context.h>
@@ -184,9 +185,14 @@ jsg::Ref<WebSocket> WebSocket::constructor(jsg::Lock& js,
   auto& context = IoContext::current();
 
   // Check if we have a valid URL
+  constexpr auto urlOptions = kj::Url::Options{.percentDecode = false, .allowEmpty = true};
   constexpr auto wsErr = "WebSocket Constructor: "_kj;
-  kj::Url urlRecord =
-      JSG_REQUIRE_NONNULL(kj::Url::tryParse(url), DOMSyntaxError, wsErr, "The url is invalid.");
+
+  // To be compatible with fetch() implementation:
+  // - First parse the URL with REMOTE_HREF which requires hostname.
+  // - Then stringify the URL with HTTP_PROXY_REQUEST which omits the userinfo.
+  kj::Url urlRecord = JSG_REQUIRE_NONNULL(kj::Url::tryParse(url, kj::Url::REMOTE_HREF, urlOptions),
+      DOMSyntaxError, wsErr, "The url is invalid.");
 
   JSG_REQUIRE(urlRecord.scheme == "ws" || urlRecord.scheme == "wss", DOMSyntaxError, wsErr,
       "The url scheme must be ws or wss.");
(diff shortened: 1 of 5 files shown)


src/workerd/api/web-socket.c++
```diff
@@ -185,8 +185,10 @@ jsg::Ref<WebSocket> WebSocket::constructor(jsg::Lock& js,
 
   // Check if we have a valid URL
   constexpr auto wsErr = "WebSocket Constructor: "_kj;
-  kj::Url urlRecord =
-      JSG_REQUIRE_NONNULL(kj::Url::tryParse(url), DOMSyntaxError, wsErr, "The url is invalid.");
+  // Any authority state (username and/or password) members should be removed.
+  // Users should use Authorization header for this purpose.
+  kj::Url urlRecord = JSG_REQUIRE_NONNULL(kj::Url::tryParse(url, kj::Url::HTTP_PROXY_REQUEST),
```
danlapid, line 190: Should we also have 
```suggestion
  auto urlOptions = kj::Url::Options{.percentDecode = false, .allowEmpty = true};
  kj::Url urlRecord = JSG_REQUIRE_NONNULL(kj::Url::tryParse(url, kj::Url::HTTP_PROXY_REQUEST, urlOptions),
```
here?

jasnell (approved):

src/workerd/api/web-socket.c++
anonrig: In reality, allowEmpty should not make a difference but for the sake of completeness with fetch, yes we should add.

Although, we should percent decode. I don't see any harm in doing it?

kentonv, line 190: This isn't correct. Please make this constructor do what fetch() does.

kentonv: Specifically it first parses as REMOTE_HREF and then re-stringifies the URL as HTTP_PROXY_REQUEST, with a bunch of other logic in between. Please actually look carefully at it and make sure we're doing the same logic here.

kentonv (changes requested):
This isn't actually what fetch() does. Let's carefully look at Fetcher::parseUrl() and how the URL is later re-serialized for the outgoing KJ request.

Also we might want to check the WebSocket API spec to see exactly how it is supposed to treat URLs with things like username:password in them.

anonrig:
> This isn't actually what fetch() does. Let's carefully look at Fetcher::parseUrl() and how the URL is later re-serialized for the outgoing KJ request.

I've updated the implementation to follow fetch() implementation.

> Also we might want to check the WebSocket API spec to see exactly how it is supposed to treat URLs with things like username:password in them.

The websocket API constructor doesn't check authority state at all. I believe what we are doing is not spec compliant and should be replaced. Most importantly, WHATWG Websocket implementation uses WHATWG URL whereas we are using RFC 3986. 

Also we should run web-platform tests for the WebSocket implementation: https://github.com/web-platform-tests/wpt/tree/master/websockets (cc @npaun)

Constructor ref: https://websockets.spec.whatwg.org/#dom-websocket-websocket

jasnell:
We already know that our `WebSocket` API implementation diverges from the standard WHATWG API in a number of ways. The web platform tests would be helpful in understanding just how far we diverge but it's not going to tell us a lot that we don't already know.

> The websocket API constructor doesn't check authority state at all. I believe what we are doing is not spec compliant and should be replaced. Most importantly, WHATWG Websocket implementation uses WHATWG URL whereas we are using RFC 3986.

More specifically, the WHATWG standard `WebSocket` constructor simply defers to the standard `Request` object's handling of the request URL.

Because we know we are not spec compliant already, just improving the error that is thrown in the current case should be good enough here. Any changes that make us more spec compliant will need to be done behind a compat flag.

jasnell:
Since the standard definition of `WebSocket` also essentially defers to fetch, I think that makes sense.

kentonv (approved):

PR #3818, opened 2025-03-27
Skip empty lines when calling JSON.parse in Pipeline Transformations

cmackenzie1:
This commit adds a safety check that the line we are going to JSON.parse
is not an empty line, which will cause JSON.parse to throw an error.

I also updated the test cases to use more data to test the chunking of
readLines function, which was rewritten to use the TextDecoderStream and
skip iterating over each character.

### DIFF
--- a/src/cloudflare/internal/test/pipeline-transform/transform-test.js
+++ b/src/cloudflare/internal/test/pipeline-transform/transform-test.js
@@ -15,7 +15,7 @@ const customTransform = class MyEntrypoint extends PipelineTransformationEntrypo
   async run(records, _) {
     for (const record of records) {
       record.dispatcher = 'was here!';
-      await new Promise((resolve) => setTimeout(resolve, 50));
+      await new Promise((resolve) => setTimeout(resolve, 1));
       record.wait = 'happened!';
     }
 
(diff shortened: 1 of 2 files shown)


cmackenzie1:

cmackenzie1:
recheck

src/cloudflare/internal/test/pipeline-transform/transform-test.js
```diff
@@ -15,19 +15,22 @@ const customTransform = class MyEntrypoint extends PipelineTransformationEntrypo
...
 
-const lines = [
-  `${JSON.stringify({ name: 'jimmy', age: '42' })}\n`,
-  `${JSON.stringify({ name: 'jonny', age: '9' })}\n`,
-  `${JSON.stringify({ name: 'joey', age: '108' })}\n`,
-];
+const lines = [];
+const jimmy = `${JSON.stringify({ name: 'jimmy', age: '42', bio: 'hello\nworld\n\n' })}\n`;
+const jonny = `${JSON.stringify({ name: 'jonny', age: '9' })}\n`;
+const joey = `${JSON.stringify({ name: 'joey', age: '108' })}\n\n`;
```
hhoughgg, line 29: We control the data that is sent to the readlines function meaning this should never happen. It seems like the appropriate fix is to guard against this by rejecting json containing newlines sent into the pipeline in the first place. The ndjson spec seems to say this is invalid.

cmackenzie1: This is valid NDJSON as an empty lines are typically accepted and ignored while parsing it. In the runtime, the code we write should be defensive and guard against things like this instead of having other services promise not to send stuff like this:

```
{"id": 1}

{"id": 2}
{"id": 3}
```

hhoughgg: I mean https://github.com/ndjson/ndjson-spec?tab=readme-ov-file#31-serialization
This error does not happen with other transforms. I had assumed that in this case the JSON itself contains newlines which then causes json parse to fail since we end up parsing partial json. This should be blocked at ingest time not during transformation.

penalosa (approved):

PR #3819, opened 2025-03-27
Add daily release automation

danlapid:
## Summary
- Adds a GitHub Actions workflow that automatically updates the \ file daily at 00:30 UTC
- This update triggers the existing release workflow, creating daily releases
- The workflow includes a manual trigger option for testing and only pushes when the date has actually changed

## Test plan
- Manually trigger the workflow to verify it correctly updates the compatibility date file
- Verify the workflow runs automatically at the scheduled time

### DIFF
--- a/.github/workflows/daily-release.yml
+++ b/.github/workflows/daily-release.yml
@@ -1,0 +1,50 @@
+name: Daily Release
+
+on:
+  schedule:
+    # Run at 00:30 UTC every day
+    - cron: '30 0 * * *'
+  # Allow manual triggering for testing
+  workflow_dispatch:
+
+jobs:
+  update-compatibility-date:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          
+      - name: Get current date
+        id: date
+        run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT
+        
+      - name: Update compatibility date
+        run: |
+          echo "${{ steps.date.outputs.date }}" > src/workerd/io/supported-compatibility-date.txt
+          
+      - name: Check for changes
+        id: git-check
+        run: |
+          if [[ $(git status --porcelain src/workerd/io/supported-compatibility-date.txt) ]]; then
+            echo "changed=true" >> $GITHUB_OUTPUT
+          else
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-release.yml
```diff
@@ -0,0 +1,50 @@
...
+            echo "changed=false" >> $GITHUB_OUTPUT
+          fi
+          
+      - name: Set up Git identity
+        if: steps.git-check.outputs.changed == 'true'
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          
+      - name: Commit and push changes
```
anonrig, line 45: This won't open a pull request, right?

danlapid: Nope, should push straight to main every day.

dom96 (approved):
The only question I have is: are there ever any situations where we don't want to make a new release so frequently? I can't think of any scenarios but maybe others have experienced them, may be worth asking the rest of our team.

Otherwise LGTM.

anonrig:
Should we exclude weekends?

fhanau:
Do we anticipate any to hit any sort of limits from releasing daily, perhaps in terms of the volume of binary artifacts being produced? (I think it should be fine since WPT releases very frequently too, but maybe different limits apply to them)

It feels wasteful to cut a release when the last commit was already a release commit, which will be common on weekends and holidays – can we add a check so that releases are cut only if the last commit wasn't a release? We can simply check if the output of `git show --pretty=format:%ae` starts with the Github bot email and bail out early in that case.

danlapid:
> Should we exclude weekends?

Why?

danlapid:
> question I have is: are there ever any situations where we don't want to make a new release



> Do we anticipate any to hit any sort of limits from releasing daily, perhaps in terms of the volume of binary artifacts being produced? (I think it should be fine since WPT releases very frequently too, but maybe different limits apply to them)
> 
> It feels wasteful to cut a release when the last commit was already a release commit, which will be common on weekends and holidays – can we add a check so that releases are cut only if the last commit wasn't a release? We can simply check if the output of `git show --pretty=format:%ae` starts with the Github bot email and bail out early in that case.

I really don't see why we should care, it's not like every release of workerd breaks the internet, a few extra megabytes will fill up github and npm's servers, I am certain we are far from the worst offenders even if we cut more releases than are useful.

anonrig (approved):

PR #3823, opened 2025-03-28
use llvm 18

mikea:
I hit https://github.com/llvm/llvm-project/issues/67449 similar to https://github.com/dtolnay/cxx/issues/1436 in #3822 

This was fixed in llvm-18: https://github.com/llvm/llvm-project/commit/078651b6de4b767b91e3e6a51e5df11a06d7bc4f

clang-18 is available since Debian stable (bookworm) and Ubuntu LTS (24.04)

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,5 +1,3 @@
-# TODO(later): llvm-toolchain-16 has been backported to Bullseye with the 11.9 point release –
-# switch to using the regular packages instead of LLVM's APT repo here.
 FROM node:bullseye AS builder
 
 WORKDIR /workerd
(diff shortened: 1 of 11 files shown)


Dockerfile.release
```diff
@@ -1,4 +1,4 @@
-# TODO(later): llvm-toolchain-16 has been backported to Bullseye with the 11.9 point release –
+# TODO(later): llvm-toolchain-18 has been backported to Bullseye with the 11.9 point release –
```
fhanau, line 1: This is the one place where auto-replace isn't helpful – 19 was only backported to Bookworm, not Bullseye. I think we can just remove the TODO here, the slightly easier setup process isn't that important.

doxyfile
```diff
@@ -1222,8 +1222,8 @@ CLANG_OPTIONS          = -std=c++20     \
                          -I/usr/include/c++/11/bits \
                          -I/usr/include/x86_64-linux-gnu \
                          -I/usr/include/x86_64-linux-gnu/c++/11 \
-                         -I/usr/lib/llvm-16/include/c++/v1 \
-                         -I/usr/lib/llvm-16/lib/clang/16/include
+                         -I/usr/lib/llvm-18/include/c++/v1 \
+                         -I/usr/lib/llvm-18/lib/clang/16/include
```
fhanau, line 1226: Nit: The second 16 needs to be updated too

compile_flags.txt
```diff
@@ -18,8 +18,8 @@
 -Iexternal/ncrypto/include
 -isystembazel-bin/external/sqlite3
 -Isrc
--isystem/usr/lib/llvm-16/include/c++/v1
--isystem/usr/lib/llvm-16/lib/clang/16/include
+-isystem/usr/lib/llvm-18/include/c++/v1
+-isystem/usr/lib/llvm-18/lib/clang/16/include
```
fhanau, line 22: Nit: The second 16 needs to be updated too

fhanau (approved):
- Last time we updated deps we already ran into the underlying issue here – I recommended freezing cxxbridge-cmd since the deps update was large enough without the compiler upgrade. We can unfreeze this in `build/deps/build_deps.jsonc` now.
- The following files still have one reference to clang or LLVM 16 each:
`README.md`
`.devcontainer/Dockerfile`
`docs/development.md`

LGTM otherwise.

anonrig (approved):

PR #3826, opened 2025-03-28
Improve crypto memory accounting 

jasnell:
Avoid use of the `kj::Array<kj::byte>` type mapping when not needed. Use jsg::BufferSource instead to better ensure that memory is accounted for.

This does not cover all of crypto but gets a sizeable chunk of it.

### DIFF
--- a/src/workerd/api/crypto/rsa.c++
+++ b/src/workerd/api/crypto/rsa.c++
@@ -70,12 +70,12 @@ jsg::BufferSource Rsa::getPublicExponent(jsg::Lock& js) {
   return KJ_REQUIRE_NONNULL(bignumToArray(js, *e));
 }
 
-CryptoKey::AsymmetricKeyDetails Rsa::getAsymmetricKeyDetail() const {
+CryptoKey::AsymmetricKeyDetails Rsa::getAsymmetricKeyDetail(jsg::Lock& js) const {
   CryptoKey::AsymmetricKeyDetails details;
 
   details.modulusLength = BN_num_bits(n);
   details.publicExponent =
-      JSG_REQUIRE_NONNULL(bignumToArrayPadded(*e), Error, "Failed to extract public exponent");
+      JSG_REQUIRE_NONNULL(bignumToArrayPadded(js, *e), Error, "Failed to extract public exponent");
 
   // TODO(soon): Does BoringSSL not support retrieving RSA_PSS params?
   // if (type == EVP_PKEY_RSA_PSS) {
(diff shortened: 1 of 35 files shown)


src/workerd/api/crypto/rsa.c++
```diff
@@ -1021,12 +1022,19 @@ kj::Own<CryptoKey::Impl> CryptoKey::Impl::importRsaRaw(jsg::Lock& js,
...
+      JSG_REQUIRE_NONNULL(Rsa::tryGetRsa(key.get()), DOMDataError, "Input was not an RSA key");
+
   return kj::heap<RsassaPkcs1V15Key>(AsymmetricKeyData{.evpPkey = kj::mv(key),
                                        .keyType = KeyType::PUBLIC,
                                        .usages = CryptoKeyUsageSet::decrypt() |
                                            CryptoKeyUsageSet::sign() | CryptoKeyUsageSet::verify()},
-      CryptoKey::RsaKeyAlgorithm{.name = "RSA"_kj}, true);
+      CryptoKey::RsaKeyAlgorithm{
+        .name = "RSA"_kj,
+        .publicExponent = KJ_REQUIRE_NONNULL(bignumToArray(js, *rsa.getE())),
```
jp4a50, line 1035: So `.publicExponent` wasn't being initialized before? Is this a functional change?

jp4a50 (approved):

src/workerd/api/crypto/rsa.c++
jasnell: Yes but a minor one. This is method is only used, as far as I can remember, in the node.js compat `X509Certificate` impl, in which case this won't have any meaningful impact.

anonrig (approved):

fhanau (approved):

anonrig (approved):

types/generated-snapshot/2021-11-03/index.d.ts
```diff
@@ -1182,26 +1182,26 @@ interface RsaOtherPrimesInfo {
 }
 interface SubtleCryptoDeriveKeyAlgorithm {
   name: string;
-  salt?: ArrayBuffer;
+  salt?: ArrayBuffer | ArrayBufferView;
```
petebacondarwin, line 1185: @jasnell - does loosening this type actually constitute a breaking change?
For example a user might previously have been relying on `salt` only being an `ArrayBuffer` but now has to handle the case where it might be an `ArrayBufferView`. This is the case in Miniflare where we had the following code:

https://github.com/cloudflare/workers-sdk/blob/7427004d45e52c0ef6e6e8dbe3ed5b79dc985d55/packages/miniflare/src/workers/r2/bucket.worker.ts#L117

This is now broken because the `stream.digest` can no longer simply be passed to `Buffer.from()`.

We have worked around this by:

```ts
const digest = await hashes[i].stream.digest;
if (digest instanceof ArrayBuffer) {
  result.set(algorithms[i], Buffer.from(digest));
} else {
  result.set(algorithms[i], Buffer.from(digest.buffer));
}
```

jasnell: The `salt` and `stream.digest` are two different things here. For `stream.digest` I just missed updating the type definition. The actual value returned will always be an `ArrayBuffer` so this is just the type declaration that is wrong. I'll open a fixup PR for that.

PR #3827, opened 2025-03-28
Allow ExternalMemoryAdjustment when isolate lock isn't held

npaun:
The initial implementation of `jsg::ExternalMemoryAdjustment` allowed for some changes in external memory usage to be made without an isolate lock being held. For example, an adjuster could be moved or destroyed at any time. If the isolate is locked, the change is reported immediately to V8. Otherwise, the change can be _deferred_, by storing an atomic value in the jsg Isolate keeping track of pending memory adjustments. The next time the isolate lock is taken, the pending adjustment is reported to V8.

This PR extends this functionality by allowing all memory adjustments, including construction, `adjust()`, and `set()`, to be deferred if needed.

I also created `jsg::ExternalMemoryTarget` which wraps a reference to an isolate, providing a `getAdjustment()` method to construct an `ExternalMemoryAdjustment` on demand. This class isn't strictly needed - you could just store an isolate pointer, but I think it's helpful because it makes intentions clear when it appears in code. I'm not 100% sure about the name though.

Lastly, I updated `compression.c++` and `zlib-util.c++` to use the new functionality.

Future question: what if a very large amount of memory is used while we're not holding the isolate? Should we implement our own maximum memory limits for pending adjustments eventually?

### DIFF
--- a/src/workerd/jsg/jsg-test.c++
+++ b/src/workerd/jsg/jsg-test.c++
@@ -423,37 +423,99 @@ KJ_TEST("jsg::Lock getUuid") {
 KJ_TEST("External memory adjustment") {
   IsolateUuidIsolate isolate(v8System, kj::heap<IsolateObserver>());
   isolate.runInLockScope([&](IsolateUuidIsolate::Lock& lock) {
-    // Creating with a specific amount works as expected
-    auto adjuster = lock.getExternalMemoryAdjustment(100);
-    KJ_ASSERT(adjuster.getAmount() == 100);
+    // Creating an inner scope to check the case where the adjustment object does not outlive the isolate
+    {
+      // Creating with a specific amount works as expected
+      auto adjuster = lock.getExternalMemoryAdjustment(100);
+      KJ_ASSERT(adjuster.getAmount() == 100);
 
-    // Adjusting up works as expected
-    adjuster.adjust(lock, 10);
-    KJ_ASSERT(adjuster.getAmount() == 110);
+      // Adjusting up works as expected
+      adjuster.adjust(10);
+      KJ_ASSERT(adjuster.getAmount() == 110);
 
-    // Adjusting down works as expected
-    adjuster.adjust(lock, -10);
-    KJ_ASSERT(adjuster.getAmount() == 100);
+      // Adjusting down works as expected
+      adjuster.adjust(-10);
+      KJ_ASSERT(adjuster.getAmount() == 100);
 
-    // Setting an explicit value just works
-    adjuster.set(lock, 50);
-    KJ_ASSERT(adjuster.getAmount() == 50);
+      // Setting an explicit value just works
+      adjuster.set(50);
+      KJ_ASSERT(adjuster.getAmount() == 50);
 
-    // Decrementing by more than the amount just sets to 0
-    adjuster.adjust(lock, -200);
-    KJ_ASSERT(adjuster.getAmount() == 0);
+      // Decrementing by more than the amount will throw an exception
+      try {
+        adjuster.adjust(-200);
+      } catch (...) {
+        auto exc = kj::getCaughtExceptionAsKj();
+        KJ_ASSERT(exc.getDescription() ==
+            "expected amount >= -static_cast<ssize_t>(this->amount) [-200 >= -50]; Memory usage may not be decreased below zero");
+      }
 
-    adjuster.set(lock, 100);
-    auto adjuster2 = kj::mv(adjuster);
-    KJ_ASSERT(adjuster2.getAmount() == 100);
-    KJ_ASSERT(adjuster.getAmount() == 0);
+      KJ_ASSERT(adjuster.getAmount() == 50);
 
+      adjuster.set(100);
+      auto adjuster2 = kj::mv(adjuster);
+      KJ_ASSERT(adjuster2.getAmount() == 100);
+      KJ_ASSERT(adjuster.getAmount() == 0);
+    }
     // Note that we are not testing the actual effect on the isolate itself here.
     // While we have added a getExternalMemory() API to the isolate via a patch in
     // the internal repo, we have not added that patch to workerd so testing the
     // specific external memory reported by the isolate is possible but a bit
     // more cumbersome here.
   });
+}
+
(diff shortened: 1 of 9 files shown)


src/workerd/jsg/jsg.h
```diff
@@ -2249,22 +2249,18 @@ class ExternalMemoryAdjustment final {
  public:
   ExternalMemoryAdjustment(
       v8::ExternalMemoryAccounter& externalMemoryAccounter, v8::Isolate* isolate, size_t amount);
-  ExternalMemoryAdjustment(v8::Isolate* isolate, size_t amount);
+  //ExternalMemoryAdjustment(v8::Isolate* isolate, size_t amount);
```
jasnell, line 2252: If this is meant to be left commented out then just remove it.
```suggestion
```

jasnell:
Generally LGTM but, as discussed in chat, I'd like this to also include the protections against trying to adjust the external memory on a v8::Isolate that has already been disposed.

jasnell:
Also tagging in @jp4a50 for a review for visibility.

src/workerd/jsg/setup.h
```diff
@@ -60,7 +60,7 @@ class V8System {
 
   ~V8System() noexcept(false);
 
-  typedef void FatalErrorCallback(kj::StringPtr location, kj::StringPtr message);
+  typedef void FatalErrorCallback(kj::StringPtr location, kj::StringPtr messagex);
```
jasnell, line 63:
```suggestion
  typedef void FatalErrorCallback(kj::StringPtr location, kj::StringPtr message);
```

Is the addition of the `x` intentional?

src/workerd/jsg/jsg.h
```diff
@@ -2246,42 +2246,72 @@ JS_TYPE_CLASSES(V)
...
+  kj::Maybe<Impl> maybeInner;
+
+  friend class ExternalMemoryAdjustment;
+  friend class IsolateBase;
+};
+
+// Used to save a reference to an isolate that is responsible for external memory usage.
+// getAdjustment() can be invoked at any time to create a new RAII adjustment object
+// pointing to this isolate
+using ExternalMemoryTarget = kj::Own<ExternalMemoryPtr>;
```
jasnell, line 2283: This alias may be a bit confusing/counter-productive. Whenever we have a `kj::Own<T>` it's best for that to be explicit so that folks know how to handle it. Just seeing `ExternalMemoryTarget` they may not realize that it's actually an own.

jasnell:
Couple of initial nits. I'll need to run some local tests to verify the changes in the behavior here. Will complete my review after I do so.

src/workerd/jsg/setup.h
npaun: I'll change it back ~to Twitter~

src/workerd/jsg/jsg-test.c++
```diff
@@ -456,6 +458,47 @@ KJ_TEST("External memory adjustment") {
...
+      KJ_ASSERT(target->getPendingMemoryUpdate() == 0);
+      KJ_ASSERT(target->isIsolateAlive());
+    });
+  }
+
+  KJ_ASSERT(!target->isIsolateAlive());
+  // Making an adjustment anyway won't do anything but also won't crash
+  auto adjuster3 = target->getAdjustment(500);
+  KJ_ASSERT(target->getPendingMemoryUpdate() == 0);
+}
```
jasnell, line 500: As suggested in chat... let's add a case to this where we allocate a `kj::Array<kj::byte>` with an attached memory adjustment that ends up outliving the isolate to ensure that the order of destruction is safe.

jasnell (approved):
LGTM with the additional test suggestion and other nits resolved.

jasnell:
The new test is failing internal CI

image

npaun:
Looking at the code, V8 does not allow destroying an ExternalMemoryAccounter that is holding a non-zero amount of memory. I think that given that we're disposing the Isolate, we no longer care and I can zero out the memory usage as part of destruction.

jasnell (approved):
LGTM so long as internal debug CI runs are good.

jasnell:
Still appears to be failing in internal CI. 

```
V8 fatal error; location = ../../src/api/api.cc:12423; message = amount_of_external_memory_ == 0U (100 vs. 0)
```

npaun:
Gah sorry for all the back and forth on this error.  I'm at a medical appointment but I'll fix it when I get back home.

jasnell:
Looks like the failure might be a bit flaky. Sometimes it works, most times it appears to fail.

npaun:
I'm going to see what's up locally. Hopefully it's not some kind of race condition.

src/workerd/jsg/jsg-test.c++
```diff
@@ -423,31 +423,33 @@ KJ_TEST("jsg::Lock getUuid") {
...
+
+      // Adjusting down works as expected
+      adjuster.adjust(-10);
+      KJ_ASSERT(adjuster.getAmount() == 100);
+
+      // Setting an explicit value just works
+      adjuster.set(50);
+      KJ_ASSERT(adjuster.getAmount() == 50);
+
+      // Decrementing by more than the amount just sets to 0
```
danlapid, line 444: This seems really error prone
Why not throw an exception? We should find out when our code messes up memory accounting

npaun: @jasnell Any objections if I switch to an exception?

anonrig: +1 on throwing

npaun:
Haha it looks like it's gonna be a 1 byte fix.

I forgot the semi-colon after a KJ_ASSERT, and because macros are evil, the call on the next line that actually decreases memory usage in V8 was interpreted as being part of a different block (the for loop that fires the exception on a failed assert) and never runs. Did I say I hate macros?

jasnell:
Hah! Love that one. That has hit us before!

src/workerd/jsg/jsg-test.c++
jasnell: No objection

jasnell (approved):

npaun:
Failures are spurious. Re-running.

PR #3828, opened 2025-03-28
Avoid heap allocations where possible in crypto and node apis

jasnell:


src/workerd/api/crypto/aes.c++
```diff
@@ -408,7 +408,8 @@ class AesCbcKey final: public AesKeyBase {
     int plainSize = 0;
     auto blockSize = EVP_CIPHER_CTX_block_size(cipherCtx.get());
 
-    auto plainText = kj::heapArray<kj::byte>(cipherText.size() + ((blockSize > 1) ? blockSize : 0));
+    KJ_STACK_ARRAY(
+        kj::byte, plainText, cipherText.size() + ((blockSize > 1) ? blockSize : 0), 1024, 4096);
```
anonrig, line 412: Can you move the magical numbers to their unique declarations with appropriate comments?

src/workerd/api/crypto/impl.c++
```diff
@@ -364,16 +364,14 @@ kj::Maybe<kj::Array<kj::byte>> simdutfBase64UrlDecode(kj::StringPtr input) {
 
 kj::Maybe<jsg::BufferSource> simdutfBase64UrlDecode(jsg::Lock& js, kj::StringPtr input) {
   auto size = simdutf::maximal_binary_length_from_base64(input.begin(), input.size());
-  ;
-  auto buf = kj::heapArray<kj::byte>(size);
+  KJ_STACK_ARRAY(kj::byte, buf, size, 1024, 4096);
```
anonrig, line 367: Ditto: magic numbers

src/workerd/api/crypto/aes.c++
jasnell: As much as I generally agree with you, I think it's better to keep usage consistent and in every other place we use `KJ_STACK_ARRAY` we inline the sizes directly the same way.

danlapid (approved):

src/workerd/api/crypto/aes.c++
anonrig: At least can you add comments explaining what those magical numbers are?

danlapid: That's explained in the doctoring of KJ_STACK_ARRAY though

anonrig: Sorry if I'm not clear enough: I know what does numbers represent. I don't know why we chose those numbers. That's what I meant by the magical number usage.

PR #3829, opened 2025-03-28
Remove superflous throw from web-socket handling

danlapid:
See internal PR 10195 for reasoning 😅

### DIFF
--- a/src/workerd/api/web-socket.c++
+++ b/src/workerd/api/web-socket.c++
@@ -529,7 +529,6 @@ void WebSocket::startReadLoop(jsg::Lock& js, kj::Maybe<kj::Own<InputGate::Critic
       } else {
         native.closedIncoming = true;
         reportError(js, kj::cp(e));
-        kj::throwFatalException(kj::mv(e));
       }
     }
   })));


jasnell (approved):
The change LGTM but this has been there for a while and I can't really recall why. Let's also have either @kentonv or @harrishancock take a look

src/workerd/api/web-socket.c++
```diff
@@ -529,7 +529,6 @@ void WebSocket::startReadLoop(jsg::Lock& js, kj::Maybe<kj::Own<InputGate::Critic
       } else {
         native.closedIncoming = true;
         reportError(js, kj::cp(e));
-        kj::throwFatalException(kj::mv(e));
```
danlapid, line 532: This throw doesn't have much purpose in a context waitUntil task outside of failing the invocation and causing an exception trace on an otherwise successful request.

The error can and should be caught by the user worker with 
```js
ws.addEventListener("error", e => {
		...
      });
```

anonrig (approved):

anonrig:
@danlapid Is it easy to add a test to workerd as well?

danlapid:
> @danlapid Is it easy to add a test to workerd as well?

The functionality is currently not supported in workerd at all causing all WS errors to be treated as internal errors
This differs from production behaviour in a bad and inconsistent way.

My intention is to fix production behaviour and then migrate the solution over to workerd along with tests.

harrishancock (approved):

PR #3830, opened 2025-03-28
Avoid unnecessary extraneous copy in blob/file

jasnell:
The concat operation returning a kj::Array was causing an extraneous data copy. Eliminate the extra copy and simplify a few other bits

### DIFF
--- a/src/workerd/api/blob.h
+++ b/src/workerd/api/blob.h
@@ -15,7 +15,6 @@ class ReadableStream;
 class Blob: public jsg::Object {
  public:
   Blob(jsg::Lock& js, jsg::BufferSource data, kj::String type);
-  Blob(jsg::Lock& js, kj::Array<byte> data, kj::String type);
   Blob(jsg::Ref<Blob> parent, kj::ArrayPtr<const byte> data, kj::String type);
 
   kj::ArrayPtr<const byte> getData() const KJ_LIFETIMEBOUND;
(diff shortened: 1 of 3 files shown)


src/workerd/api/blob.h
```diff
@@ -121,7 +120,8 @@ class File: public Blob {
   // lock. This is currently only the case when parsing FormData outside of running
   // JavaScript (such as in the internal fiddle service).
   File(kj::Array<byte> data, kj::String name, kj::String type, double lastModified);
```
danlapid, line 122: Why not change this constructor as well to consolidate the `else` clause here: https://github.com/cloudflare/workerd/pull/3830/files#diff-d1203721eeacfc2a660056488a1e47d12e63017b164f6b1c7b01290c47dd3678R185 ?

danlapid (approved):

src/workerd/api/blob.h
jasnell: We can't really. In this case we don't have the isolate lock so we end up having to keep the data as a kj::Array rather than an v8::ArrayBuffer/v8::BackingStore. There's a case where we end up parsing FormData outside of the isolate lock and end up creating a `File` object even tho it is never actually passed off to JS. This would require an additional refactor that I'd prefer to keep separate.

anonrig (approved):

PR #3832, opened 2025-03-28
Break actor on errors that rollback transactions

shrima-cf:
We should break output gates and reset actors in the event that we catch the following exceptions because they implicitly roll back any transaction:

[SQLITE_FULL](https://www.sqlite.org/rescode.html#full): database or disk full
[SQLITE_IOERR](https://www.sqlite.org/rescode.html#ioerr): disk I/O error
[SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy): database in use by another process
[SQLITE_NOMEM](https://www.sqlite.org/rescode.html#nomem): out of memory
https://www.sqlite.org/lang_transaction.html#response_to_errors_within_a_transaction

### DIFF
--- a/src/workerd/api/sql-test.js
+++ b/src/workerd/api/sql-test.js
@@ -1420,6 +1420,51 @@ export class DurableObjectExample extends DurableObject {
     }
     return bookmark;
   }
+
+  async createTableForTestingAutoRollBackOnCriticalError() {
+    this.state.storage.sql.exec(
+      'CREATE TABLE IF NOT EXISTS test_full (id INTEGER PRIMARY KEY, data BLOB)'
+    );
+  }
+
+  async createCriticalErrorThatLeadsToAutoRollback() {
+    // Limit size of db so we can trigger a SQLITE_FULL error
+    this.state.storage.sql.setMaxPageCountForTest(10);
+
+    try {
+      // Create large data to fill the database quickly
+      const largeData = new Uint8Array(1000000).fill(42);
+      this.state.storage.sql.exec(
+        'INSERT INTO test_full VALUES (?, ?)',
+        2,
+        largeData
+      );
+      throw new Error(
+        'should have thrown SQLITE_FULL exception before we reach here'
+      );
+    } catch (err) {
+      if (err.message !== 'database or disk is full: SQLITE_FULL') {
+        throw err;
+      }
+    }
+
+    const smallData = new Uint8Array(10).fill(1);
+    this.state.storage.sql.exec(
+      'INSERT INTO test_full VALUES (?, ?)',
+      1,
+      smallData
+    );
+  }
+
+  async verifyAutoRollbackAfterCriticalError() {
+    if (
+      [...this.state.storage.sql.exec('SELECT * FROM test_full')].length != 0
+    ) {
+      throw new Error(
+        'found data that was committed when a critical error happened'
+      );
+    }
+  }
 }
 
 export default {
@@ -1529,5 +1574,26 @@ export let testSessionsAPIBookmark = {
     for (let i = 0; i < 20; ++i) {
       bookmark = await stub.testSessionsAPIBookmark(bookmark);
     }
+  },
+};
+
+export let testAutoRollBackOnCriticalError = {
+  async test(ctrl, env, ctx) {
+    let id = env.ns.idFromName('auto-rollback-on-critical-error-test');
+    let stub = env.ns.get(id);
+    await stub.createTableForTestingAutoRollBackOnCriticalError();
+    try {
+      await stub.createCriticalErrorThatLeadsToAutoRollback();
+      throw new Error(
+        'should have thrown SQLITE_FULL exception before we reach here'
+      );
+    } catch (err) {
+      if (!err.message.startsWith('internal error; reference =')) {
+        throw err;
+      }
+    }
+    // Get a new stub since the old stub is broken due to critical error
+    stub = env.ns.get(id);
+    await stub.verifyAutoRollbackAfterCriticalError();
   },
 };
--- a/src/workerd/io/actor-sqlite.c++
+++ b/src/workerd/io/actor-sqlite.c++
@@ -5,6 +5,7 @@
 #include "actor-sqlite.h"
 
 #include "io-gate.h"
+#include "kj/function.h"
 
 #include <workerd/jsg/exception.h>
 #include <workerd/util/autogate.h>
@@ -38,6 +39,7 @@ ActorSqlite::ActorSqlite(kj::Own<SqliteDatabase> dbParam,
       metadata(*db),
       commitTasks(*this) {
   db->onWrite(KJ_BIND_METHOD(*this, onWrite));
+  db->onCriticalError(KJ_BIND_METHOD(*this, onCriticalError));
   lastConfirmedAlarmDbState = metadata.getAlarm();
 
   // Because we preserve an invariant that scheduled alarms are always at or earlier than
@@ -144,6 +146,7 @@ void ActorSqlite::ExplicitTxn::setAlarmDirty() {
 }
 
 kj::Maybe<kj::Promise<void>> ActorSqlite::ExplicitTxn::commit() {
+  actorSqlite.requireNotBroken();
   KJ_REQUIRE(!hasChild,
       "critical sections should have prevented committing transaction while "
       "nested txn is outstanding");
@@ -193,6 +196,13 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
     alarmDirty = p->alarmDirty;
   } else {
     alarmDirty = false;
+  }
+}
+
+void ActorSqlite::onCriticalError(kj::Exception exception) {
+  // If we have already experienced a terminal exception, no need to replace it
+  if (broken == kj::none) {
+    broken.emplace(kj::mv(exception));
   }
 }
 
--- a/src/workerd/util/sqlite-test.c++
+++ b/src/workerd/util/sqlite-test.c++
@@ -1569,5 +1569,114 @@ KJ_TEST("I/O exceptions pass through SQLite") {
   )")));
 }
 
+void testCriticalError(const char* expectedErrorMessage,
+    kj::Function<void(SqliteDatabase&, SqliteDatabase::Vfs& vfs)> triggerErrorFn) {
+  auto dir = kj::newInMemoryDirectory(kj::nullClock());
+  SqliteDatabase::Vfs vfs(*dir);
+  SqliteDatabase db(vfs, kj::Path({"foo"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+  // Create a tracker to verify our callback is called
+  bool criticalErrorCallbackCalled = false;
+  kj::Maybe<kj::Exception> capturedException = kj::none;
+
+  // Register a critical error callback
+  db.onCriticalError([&](kj::Maybe<kj::Exception> exception) {
+    criticalErrorCallbackCalled = true;
+    capturedException = exception;
+  });
+
+  KJ_EXPECT_THROW_MESSAGE(expectedErrorMessage, triggerErrorFn(db, vfs));
+
+  KJ_EXPECT(criticalErrorCallbackCalled);
+  KJ_ASSERT(capturedException != kj::none);
+  KJ_IF_SOME(exception, capturedException) {
+    KJ_EXPECT(exception.getDescription().contains(expectedErrorMessage));
+  }
+}
+
+KJ_TEST("SQLite critical error handling for SQLITE_IOERR") {
+  auto dir = kj::atomicRefcounted<ErrorInjectableDirectory>();
+  SqliteDatabase::Vfs vfs(*dir);
+  SqliteDatabase db(vfs, kj::Path({"db"}), kj::WriteMode::CREATE | kj::WriteMode::MODIFY);
+
+  // Create a tracker to verify our callback is called
+  bool criticalErrorCallbackCalled = false;
+  kj::Maybe<kj::Exception> capturedException;
+  // Register a critical error callback
+  db.onCriticalError([&](kj::Exception exception) {
+    criticalErrorCallbackCalled = true;
+    capturedException = exception;
+  });
+
+  // Use a small cache size to force flushing to disk on even a small write
+  db.run(SqliteDatabase::TRUSTED, "PRAGMA cache_size = 1");  // 1 page cache
+
+  db.run(SqliteDatabase::TRUSTED, kj::str(R"(
+    CREATE TABLE IF NOT EXISTS things (
+      id INTEGER PRIMARY KEY AUTOINCREMENT,
+      value INTEGER
+    );
+    INSERT INTO things(value) VALUES (123);
+  )"));
+
+  db.run("BEGIN TRANSACTION");
+
+  // Now arrange for an error on write().
+  KJ_ASSERT_NONNULL(dir->dbFile)->error = KJ_EXCEPTION(FAILED, "test-vfs-error");
+
+  KJ_EXPECT_THROW_MESSAGE("test-vfs-error", db.run(SqliteDatabase::TRUSTED, kj::str(R"(
+    INSERT INTO things(value) VALUES (456);
+  )")));
+
+  KJ_EXPECT(criticalErrorCallbackCalled);
+  KJ_ASSERT(capturedException != kj::none);
+  KJ_IF_SOME(exception, capturedException) {
+    KJ_EXPECT(exception.getDescription().startsWith("test-vfs-error"));
+  }
+}
+
+// No test for SQLITE_BUSY as a critical error because we haven't been able to figure out how to
+// trigger it in a way that causes an auto-rollback. It seems like an auto-rollback would only
+// happen if the transaction had already included other writes before hitting SQLITE_BUSY,
+// but if a transaction is open and has performed writes, then obviously the caller must hold
+// the lock, and so would not be expected to see SQLITE_BUSY.
+
+KJ_TEST("SQLite critical error handling for SQLITE_FULL") {
+  testCriticalError("database or disk is full", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
+    // Set up a database with limited size
+    db.run("PRAGMA max_page_count = 10");
+    db.run("CREATE TABLE IF NOT EXISTS test_full (id INTEGER PRIMARY KEY, data BLOB)");
+
+    db.run("BEGIN TRANSACTION");
+
+    // Create a large blob to quickly fill the database
+    auto largeData = kj::heapArray<byte>(100000, 'X');  // 100KB
+
+    // This should eventually trigger SQLITE_FULL
+    db.run(SqliteDatabase::TRUSTED, "INSERT INTO test_full VALUES (?, ?)", 1, largeData.asPtr());
+  });
+}
+
+KJ_TEST("SQLite critical error handling for SQLITE_NOMEM") {
+  testCriticalError("out of memory", [](SqliteDatabase& db, SqliteDatabase::Vfs& vfs) {
+    db.run("CREATE TABLE test_nomem (id INTEGER PRIMARY KEY, data BLOB)");
+    db.run(
+        "CREATE TABLE test_refs (id INTEGER PRIMARY KEY, ref_id INTEGER, FOREIGN KEY(ref_id) REFERENCES test_nomem(id) ON DELETE CASCADE)");
+
+    db.run("BEGIN TRANSACTION");
+
+    db.run("INSERT INTO test_nomem VALUES (1, 'small data')");
+    db.run("INSERT INTO test_refs VALUES (1, 1)");
+
+    // Set SQLite's memory limit very low to trigger SQLITE_NOMEM
+    db.run("PRAGMA hard_heap_limit=8192");  // 8KB limit
+
+    // Create data that will exceed the memory limit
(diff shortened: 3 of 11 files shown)


shrima-cf:
See https://github.com/cloudflare/workerd/pull/3832 for context

src/workerd/io/actor-sqlite.c++
```diff
@@ -195,6 +197,13 @@ void ActorSqlite::ExplicitTxn::rollbackImpl() noexcept(false) {
   }
 }
 
+void ActorSqlite::onCriticalError(kj::Exception exception) {
+  // If we have already experienced a terminal exception, no need to replace it
+  if (broken == kj::none) {
+    broken.emplace(kj::mv(exception));
```
kentonv, line 203: Let's add `requireNotBroken()` to `ExplicitTxn::commit()` to make sure that this actually forces the txn commit to fail.

src/workerd/util/sqlite-test.c++
```diff
@@ -2,9 +2,12 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
+#include "kj/mutex.h"
 #include "sqlite.h"
+#include "workerd/io/io-gate.h"
```
kentonv, line 7: No longer used?

src/workerd/util/sqlite-test.c++
```diff
@@ -2,9 +2,12 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
+#include "kj/mutex.h"
```
kentonv, line 5: No longer used?

src/workerd/util/sqlite-test.c++
```diff
@@ -14,6 +17,7 @@
 #include <cerrno>
 #include <cstdint>
 #include <cstdlib>
+#include <thread>
```
kentonv, line 20: No longer used?

src/workerd/util/sqlite-test.c++
```diff
@@ -2,9 +2,12 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
+#include "kj/mutex.h"
 #include "sqlite.h"
+#include "workerd/io/io-gate.h"
 
 #include <fcntl.h>
+#include <sqlite3.h>
```
kentonv, line 10: No longer used?

src/workerd/util/sqlite-test.c++
```diff
@@ -1476,11 +1480,23 @@ class ErrorInjectableDirectory final: public kj::Directory, public kj::AtomicRef
   }
 };
 
-KJ_TEST("I/O exceptions pass through SQLite") {
```
kentonv, line 1479: Can we keep this test as it was, and add a new test? I think the existing test should still work as-is, because the I/O error happens while in autocommit mode, so shouldn't be treated as a critical error.

src/workerd/util/sqlite-test.c++
```diff
@@ -1489,13 +1505,103 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+    lockDb.run("BEGIN EXCLUSIVE TRANSACTION");
+    lockDb.run("UPDATE lock_test SET id = 2 WHERE id = 1");
+
+    db.run("BEGIN TRANSACTION");
+
+    // Attempt to start another exclusive transaction in the main database
+    // This should trigger SQLITE_BUSY
+    // TODO: This should ideally trigger an auto-rollback, but it doesn't, need to figure out why
+    // It does trigger the auto-rollback if not executed inside a transaction
+    db.run("BEGIN EXCLUSIVE TRANSACTION");
```
kentonv, line 1567: SQLite doesn't support nested transactions, so I'd expect this statement to fail simply because it's executed inside another transaction on the same connection, regardless of whether the database is also locked by another transaction. Is this actually testing the right thing?

I would think what you actually want to do here is try to perform some sort of write.

src/workerd/util/sqlite-test.c++
```diff
@@ -1489,13 +1505,103 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+    // Set SQLite's memory limit very low to trigger SQLITE_NOMEM
+    db.run("PRAGMA hard_heap_limit=8192");  // 8KB limit
+
+    // Create data that will exceed the memory limit
+    auto largeData = kj::heapArray<byte>(50000, 'X');  // 50KB
+
+    // This operation should trigger SQLITE_NOMEM when SQLite tries to allocate
+    // memory for the large parameter
+    // TODO: This should ideally trigger an auto-rollback, but it doesn't, need to figure out why
+    // It does trigger the auto-rollback if not executed inside a transaction
```
kentonv, line 1602: I don't understand this comment. If the statement is not triggered inside a transaction, then there's nothing to roll back, so what does it mean to "trigger the auto-rollback" in that case?

src/workerd/util/sqlite.c++
```diff
@@ -137,6 +137,10 @@ class SqliteCallScope {
     }
   }
 
+  kj::Maybe<kj::Exception> getException() {
+    return error;
```
kentonv, line 141: This is making a copy of the exception, which may be expensive. It looks like a second copy is made when calling `handleCriticalError()`.

Perhaps this should return `kj::Maybe<const kj::Exception&>` instead, or maybe there's some better way to avoid creating these copies.

src/workerd/util/sqlite.c++
```diff
@@ -173,17 +178,17 @@ class SqliteCallScope {
     int _ec = code;                                                                                \
     /* SQLITE_MISUSE doesn't put error info on the database object, so check it separately */      \
     KJ_ASSERT(_ec != SQLITE_MISUSE, "SQLite misused: " #code, ##__VA_ARGS__);                      \
-    if (_ec == SQLITE_IOERR) sqliteCallScope.rethrowVfsError();                                    \
```
kentonv, line 176: Reminder that the call to `rethrowVfsError()` is important. It looks like we'll now end up throwing a generic error instead of the pass-through error, and you changed the test which checked for this so it no longer checks for it, but we actually do want to keep that behavior as it was.

src/workerd/util/sqlite.c++
```diff
@@ -150,8 +154,9 @@ class SqliteCallScope {
 // the return value of sqlite3_errmsg() or a string literal containing a similarly
 // application-approriate error message. A reference called `regulator` must be in-scope.
 // sqliteErrorCode is a kj::Maybe<int> and represents the error code from sqlite.
-#define SQLITE_REQUIRE(condition, sqliteErrorCode, errorMessage, ...)                              \
+#define SQLITE_REQUIRE(condition, sqliteErrorCode, errorMessage, maybeException, ...)              \
```
kentonv, line 157: Adding this `maybeException` param to the macro seems to add a lot of noise to the places that use the macro directly, none of which actually pass an error code and thus they cannot actually trigger critical errors.

Maybe we should put the `handleCriticalError()` call inside of `SQLITE_CALL`/`SQLITE_CALL_FAILED` after all. Plus that way there's always a `SqliteCallScope`...

src/workerd/util/sqlite-test.c++
shrima-cf: Doing a write here instead of another exclusive transaction also results in same behavior.
This is what I see: 
1. if I do the write inside the begin transaction, the inTransaction=true check passes but autorollback does not happen
2. if I skip the begin transaction and directly go to write or do the write inside an exclusive transaction, the inTransaction=true check fails, but autorollback happens

shrima-cf: I am not able to formulate a scenario that forces both inTransaction=true and auto rollback=true in order to verify the onCritical error works as expected.
Maybe it is hard to simulate this deterministically because SQLITE_BUSY is not a forever error, the db will recover from this if the DB is not longer busy after sometime?

src/workerd/util/sqlite-test.c++
shrima-cf: When I say trigger the auto rollback, I mean `sqlite3_get_autocommit(db) != 0`
I believe SQLITE creates an implicit transaction in this case and that is rolls back.

src/workerd/io/actor-sqlite.c++
shrima-cf: Fixed in https://github.com/cloudflare/workerd/pull/3832/commits/abe4e51551ca8c6d36811926f1957f26d4989e98

src/workerd/util/sqlite-test.c++
shrima-cf: Fixed in https://github.com/cloudflare/workerd/pull/3832/commits/abe4e51551ca8c6d36811926f1957f26d4989e98

src/workerd/util/sqlite-test.c++
shrima-cf: Fixed in https://github.com/cloudflare/workerd/pull/3832/commits/abe4e51551ca8c6d36811926f1957f26d4989e98

src/workerd/util/sqlite-test.c++
shrima-cf: Fixed in https://github.com/cloudflare/workerd/pull/3832/commits/abe4e51551ca8c6d36811926f1957f26d4989e98

src/workerd/util/sqlite.c++
shrima-cf: Fixed in https://github.com/cloudflare/workerd/pull/3832/commits/abe4e51551ca8c6d36811926f1957f26d4989e98

src/workerd/util/sqlite-test.c++
shrima-cf: I fixed the test for SQLITE_NOMEM, we have to have enough complex statements in the transaction before we hit SQLITE_NOMEM error for Sqlite to decide to auto-rollback entire transaction despite being in an explicit transaction.

src/workerd/util/sqlite-test.c++
shrima-cf: It is not possible for us to test this scenario due to the way we have identified transactions in our code. I have written an explanation on why we cannot write a test for this.

shrima-cf: See https://github.com/cloudflare/workerd/pull/3832/files#diff-2b573b1c6fa7749e40c7da2c29f37f0f4accf53f2406bdab6ca5378755260fec

kentonv (commented):
OK, at this point I just have straightforward comments. Yay!

But, what about an end-to-end test? I don't see any here nor in the internal repo. Can we test this somehow? Maybe using an SRS testDirective to inject an error?

src/workerd/util/sqlite-test.c++
shrima-cf: Uhh, I will remove this

src/workerd/util/sqlite-test.c++
```diff
@@ -1498,5 +1499,124 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+  )")));
+
+  KJ_EXPECT(criticalErrorCallbackCalled);
+  KJ_ASSERT(capturedException != kj::none);
+  KJ_IF_SOME(exception, capturedException) {
+    KJ_EXPECT(exception.getDescription().startsWith("test-vfs-error"));
+  }
+}
+
+KJ_TEST("SQLite critical error handling for SQLITE_BUSY") {
```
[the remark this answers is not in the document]
shrima-cf: Fine to leave a comment, it might just be difficult to search for this comment in a 1500+ line test file.

src/workerd/util/sqlite-test.c++
```diff
@@ -1498,5 +1499,124 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+  /*
+  In order for this test to succeed, we have to create a scenario where we are in an explicit
+  transaction and the sqlite has decided to auto-rollback the statement inside the transaction
+  that caused the error.
+  It is not straightforward to simulate this because Sqlite will only auto-rollback
+  an explicit transaction if it determines it has complex operations preceeding the point at
+  which it encountered a SQLITE_BUSY error.
+  With the current sqlite.c++ code we have, we only conside BEGIN TRANSACTION as a transaction,
```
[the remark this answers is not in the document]
shrima-cf: Removed this line. Its probably too much details for the comment here, but what I was trying to say was sqlite defers a db level lock in case of just BEGIN TRANSACTION till it hits the first statement that needs a lock, whereas in case of EXCLUSIVE or IMMEDIATE transactions, it takes the lock immediately.

src/workerd/util/sqlite.c++
```diff
@@ -4,6 +4,8 @@
 
 #include "sqlite.h"
 
+#include "kj/exception.h"
```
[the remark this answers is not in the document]
shrima-cf: And you are right about the transitive dependency, I removed this include.

src/workerd/util/sqlite-test.c++
shrima-cf: Removed - https://github.com/cloudflare/workerd/pull/3832/commits/a40f9d7dd2684fa4f60d8a3eb5f14c2adc98b9ea#diff-2b573b1c6fa7749e40c7da2c29f37f0f4accf53f2406bdab6ca5378755260fec

src/workerd/util/sqlite-test.c++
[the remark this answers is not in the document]
shrima-cf: https://github.com/cloudflare/workerd/pull/3832/commits/a40f9d7dd2684fa4f60d8a3eb5f14c2adc98b9ea#diff-2b573b1c6fa7749e40c7da2c29f37f0f4accf53f2406bdab6ca5378755260fec

src/workerd/util/sqlite.c++
```diff
@@ -533,6 +541,34 @@ void SqliteDatabase::notifyWrite() {
...
+
+      sqlite3* db = &KJ_ASSERT_NONNULL(maybeDb, "previous reset() failed");
+      // We are in a transaction
+      if (inTransaction || !savepoints.empty()) {
+        // The transaction was auto-rolledback, re-enabling the auto commit mode, so we should fail
+        if (sqlite3_get_autocommit(db) != 0) {
+          KJ_IF_SOME(cb, onCriticalErrorCallback) {
+            KJ_IF_SOME(e, maybeException) {
+              kj::Exception exception(e);
+              cb(kj::mv(exception));
```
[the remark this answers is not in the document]
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3832/commits/a40f9d7dd2684fa4f60d8a3eb5f14c2adc98b9ea#diff-3009a3da17d76a94673aa828ce6c364fc24e34d889b4995adfc249a0e2a3c754

src/workerd/util/sqlite.c++
```diff
@@ -533,6 +541,34 @@ void SqliteDatabase::notifyWrite() {
...
+            } else {
+              auto exception = kj::Exception(
+                  kj::Exception::Type::FAILED, __FILE__, __LINE__, kj::heapString(errorMessage));
+              cb(kj::mv(exception));
```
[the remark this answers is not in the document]
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3832/commits/a40f9d7dd2684fa4f60d8a3eb5f14c2adc98b9ea#diff-3009a3da17d76a94673aa828ce6c364fc24e34d889b4995adfc249a0e2a3c754

src/workerd/util/sqlite.h
```diff
@@ -178,6 +178,14 @@ class SqliteDatabase {
     onWriteCallback = kj::mv(callback);
   }
 
+  void handleCriticalError(kj::Maybe<int> errorCode,
```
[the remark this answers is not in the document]
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3832/commits/a40f9d7dd2684fa4f60d8a3eb5f14c2adc98b9ea#diff-3f8ec143f000a9ec0c36f05ac809dc91de6b65742cebbfcd1d8198437e0e4533

src/workerd/util/sqlite.h
```diff
@@ -522,7 +531,14 @@ class SqliteDatabase::Query final: private ResetListener {
     }
   }
 
+  void handleCriticalError(kj::Maybe<int> errorCode,
```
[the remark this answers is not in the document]
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3832/commits/a40f9d7dd2684fa4f60d8a3eb5f14c2adc98b9ea#diff-3f8ec143f000a9ec0c36f05ac809dc91de6b65742cebbfcd1d8198437e0e4533

src/workerd/util/sqlite-test.c++
```diff
@@ -1569,5 +1569,117 @@ KJ_TEST("I/O exceptions pass through SQLite") {
...
+
+  KJ_EXPECT(criticalErrorCallbackCalled);
+  KJ_ASSERT(capturedException != kj::none);
+  KJ_IF_SOME(exception, capturedException) {
+    KJ_EXPECT(exception.getDescription().startsWith("test-vfs-error"));
+  }
+}
+
+/*
+There is no test for SQLITE_BUSY error because for such a test to succeed, we have to create a
```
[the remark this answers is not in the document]
shrima-cf: Fixed - https://github.com/cloudflare/workerd/pull/3832/commits/8ee51d4adabafe6eb0e8c37a2407e18a14174630#diff-2b573b1c6fa7749e40c7da2c29f37f0f4accf53f2406bdab6ca5378755260fec

kentonv:
LGTM, just needs the fixups squashed and the internal test.

src/workerd/api/sql-test.js
```diff
@@ -1420,6 +1420,35 @@ export class DurableObjectExample extends DurableObject {
...
     return bookmark;
   }
+
+  async createTableForTestingAutoRollBackOnCriticalError() {
+    this.state.storage.sql.exec("CREATE TABLE IF NOT EXISTS test_full (id INTEGER PRIMARY KEY, data BLOB)");
+  }
+
+  async createCriticalErrorThatLeadsToAutoRollback() {
+    // Limit size of db so we can trigger a SQLITE_FULL error
+    // this.state.storage.sql.exec('PRAGMA max_page_count = 10');
```
shrima-cf, line 1430: Need a way to run this without hitting SQLITE_AUTH error. Without this, the test will fail

shrima-cf: Thanks! I got it to work - https://github.com/cloudflare/workerd/pull/3832/commits/d9466ba16adb299c8251980459ee0ab87c113c6c#diff-cafe694298518daa2033c3d3343dae4de7cbb5dd8aaf37f98ec8f5c0f3feb597

src/workerd/api/sql-test.js
```diff
@@ -1420,6 +1420,51 @@ export class DurableObjectExample extends DurableObject {
...
+    this.state.storage.sql.exec(
+      'INSERT INTO test_full VALUES (?, ?)',
+      1,
+      smallData
+    );
+  }
+
+  async verifyAutoRollbackAfterCriticalError() {
+    if (
+      [...this.state.storage.sql.exec('SELECT * FROM test_full')].length != 0
```
kentonv, line 1461: As currently written the very first INSERT will fail because it tries to insert a 1MB value when the database is limited to 40kb (10 pages, 4k per page). So of course we expect this query won't return any results, but that doesn't prove anything about critical error handling.

What we really need to test here is:
1. Perform an insert that succeeds.
2. Without doing any awaits in between, perform an insert that causes the critical error.
3. After the object restarts, verify that the insert in step 1 was rolled back.

The successful insert in step 1 should be a small value.

shrima-cf: The reason I had the test this way was to verify that, with the fix, the actor will be broken after the large write, so the subsequent small write will not succeed. Without the fix, the actor will not be broken, so the small write will go through.

The rollback will happen regardless of the fix, it is the output gate broken behavior that is changing with the fix?

shrima-cf: When I do what you are suggesting, the test actually passes with and without my fix, I am guessing this is because, the transaction gets rolled back completely including the successful write (since they are both inside the same implicit transaction).

Unless I understood what you mean by #1 here, which is maybe creating a separate method for doing that successful write ?

shrima-cf: We discussed offline, the test is correct.

kentonv: Ugh, sorry, I misread your test and forgot what we're testing for!

That's right, the idea it to do an additional insert _after_ the critical error, to prove that it doesn't incorrectly go through. And that's exactly what you are doing. So your test is right.

kentonv (approved):

shrima-cf:
@cloudflare/wrangler Could someone help review this PR so I can merge it?

kentonv:
I bypassed codeowner check and merged.

@danlapid why are our merges being blocked on review from the wrangler team?

danlapid:
> I bypassed codeowner check and merged.
> 
> 
> 
> @danlapid why are our merges being blocked on review from the wrangler team?

As a follow up to an incident of D1 a few weeks ago we set wrangler as a required CODEOWNER on types changes.

The ANT team has historically been the one to maintain and own the types generation and the generated types and they have a better grasp of the possible implications of those changes.

They are usually very quick to review but I suspect that due to Friday and Monday being bank holidays in the UK this pull request didn't get much attention.

PR #3834, opened 2025-03-30
[build] Clean up pyodide target deps and headers, resolving workerd.capnp dependency

fhanau:
Spiritual successor to #2811. Resolves downstream's dependency on workerd.capnp.h by moving workerd-only python code into server target.

=================
Decided to do some cleanup to kill time on a flight. This will eliminate workerd.capnp as a superfluous linker input upstream. We also manage to no longer compile requirements.c++ twice.

### DIFF
--- a/src/workerd/api/BUILD.bazel
+++ b/src/workerd/api/BUILD.bazel
@@ -18,6 +18,7 @@ filegroup(
             "hyperdrive.c++",
             "pyodide/pyodide.c++",
             "pyodide/setup-emscripten.c++",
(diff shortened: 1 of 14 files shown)


dom96 (approved):
Looks good, but internal CI seems to be failing, do make sure that passes before merging.

Thank you for working on this!

fhanau:
> Looks good, but internal CI seems to be failing, do make sure that passes before merging.
> 
> Thank you for working on this!

The Run internal build CI job is still unstable in my experience – tests pass when running downstream CI manually.

PR #3838, opened 2025-03-31
Split up several Bazel targets

joshthoward:


jasnell (approved):

fhanau:
Ready to review this when out of draft stage – also note that this will need a downstream PR with non-trivial changes as targets like `@workerd//src/workerd/util` disappear.

joshthoward:
> Ready to review this when out of draft stage – also note that this will need a downstream PR with non-trivial changes as targets like `@workerd//src/workerd/util` disappear.

@fhanau To clarify, you mean an internal PR replacing references to this target in edgeworker? That makes sense, but why would you expect that to be any less trivial than anything else I've done here?

fhanau:
> > Ready to review this when out of draft stage – also note that this will need a downstream PR with non-trivial changes as targets like `@workerd//src/workerd/util` disappear.
> 
> @fhanau To clarify, you mean an internal PR replacing references to this target in edgeworker? That makes sense, but why would you expect that to be any less trivial than anything else I've done here?

With non-trivial, I was referring to the effort needed/number of targets that need to be updated – need to replace the split-up dependencies, and then dependents of targets that used to use util may need to have more dependencies added to make a different part of util available.

PR #3842, opened 2025-03-31
Update release workflow to create incremental patch versions

danlapid:
This change enables the release workflow to automatically create a new release on every merge to main by incrementing the patch version number for a given date-based release. Previously, the workflow always used ".0" as the patch version, which meant only one release per day was possible.

With this change:
- When a new date is used, it starts with patch version ".0"
- For each subsequent merge on the same day, the patch version is incremented
- The workflow searches for existing tags with the same YYYYMMDD date part
- Finds the highest current patch version and increments it by 1
- The manual patch version input parameter has been removed
- Both automatic and manual triggers now use the same auto-incrementing logic

This ensures we can have multiple releases per day, which enables continuous delivery for workerd where each merge to main results in a new release with an automatically incremented version number.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,10 +6,6 @@ on:
     - main
   workflow_dispatch:
     inputs:
-      patch:
-        description: 'Patch Version'
-        required: true
-        default: '0'
       prerelease:
         description: 'Is Prerelease'
         type: boolean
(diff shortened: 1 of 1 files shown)


danlapid:
Total cost:            $0.68

danlapid:
Moving back to draft to see if this is still relevant after enabling daily release workflow.

danlapid:
Closing for now but we can merge it at any point if we want this.

PR #3844, opened 2025-03-31
Implements introspection of Python classes for DO support in EW.

dom96:
This PR implements code necessary to support Durable Objects in EW. The user worker's main module is analysed to determine any classes that are exported.

### Test Plan

See PR in EW repo.


src/pyodide/python-entrypoint-helper.ts
```diff
@@ -104,6 +104,9 @@ async function setupPatches(pyodide: Pyodide): Promise<void> {
     if (TRANSITIVE_REQUIREMENTS.has('fastapi')) {
       await injectSitePackagesModule(pyodide, 'asgi', 'asgi');
     }
+
+    // TODO: ideally we shouldn't need to inject this permanently.
+    await injectSitePackagesModule(pyodide, 'introspection', 'introspection');
```
dom96, line 109: @hoodmane what's the best API to use here to avoid allowing the user to import this module?

src/pyodide/internal/introspection.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Iterates through the defined symbols in the input module. Returns any classes which extend
+    DurableObject.
+    """
+    result = []
+    for [key, val] in user_mod.__dict__.items():
+        if key.startswith("_"):
+            continue
+        if isclass(val) and issubclass(val, DurableObject) and val != DurableObject:
+            result.append(key)
```
hoodmane, line 16: We should respect `__all__` if it's defined:
```
if hasattr(user_mod, "__all__"):
    keys = user_mod.__all__
else:
    keys = (key for key in user_mod.__dict__ if not key.startswith("_"))
for key in keys:
    ....
```

hoodmane: And probably we should do:
```
return to_js([key for key in keys if filter(key)])
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -246,14 +254,22 @@ try {
...
     }
+
+    // In order to get the durable object classes exported by the worker, we use a Python module
+    // to introspect the user's main module. So we are effectively using Python to analyse the
+    // classes exported by the user worker here. The class names are then exported and used to
+    // create the equivalent JS classes via makeDurableObjectClass.
+    const pyodide = await getPyodide();
+    const introspectionMod = pyodide.pyimport('introspection');
+    const classesList = introspectionMod.collect_classes(mainModule);
+    pythonDurableObjectClasses.push(...classesList);
```
hoodmane, line 272: Why not:
```suggestion
    pythonDurableObjectClasses  = introspectionMod.collect_classes(mainModule);
```

hoodmane (approved):

hoodmane:
https://github.com/cloudflare/workerd/pull/3859

PR #3847, opened 2025-03-31
Add exit code as a return value in containers monitor

gabivlj:


kentonv:
I think `monitor()` should default to throwing an exception on a non-zero exit code. Otherwise it's too easy to screw up and write code that treats non-zero exits as non-errors.

But I think we could consider adding an opt-in way to handle the exit code without an exception. Maybe you can optionally specify an error handler function:

```
container.monitor({
  onExit(code) {
    if (coed == 2) {
      // exit code 2 is not an error
      return {};
    }

    // If we return undefined, default handling throws an error.
  }

  onTerminateSignal(signo) {
    // handle exit due to signal
  }

  onError(err) {
    // handle arbitrary other reasons the container might shut down
  }
})
```

idk if that's the best interface though.

gabivlj:
should we instead pass an option to monitor that is like `monitor({ 2: 'none' })` or a list of accepted exit codes? `monitor({ expected: [2]})`? I imagine that it might be a bit verbose to the user to write code that checks for exit codes. I think it also could be fine handling the non exit 0 in an exception if the error message does not ever change tbf.

jasnell:
Could the `exitCode` "simply" be added to the error object?

```
try {
  await container.monitor();
} catch (err) {
  console.log(err.exitCode);
}
```

kentonv:
@jasnell It's possible, but as you know I don't like when exceptions become "part of the API". I think if the application is expecting to handle non-zero error codes in some particular way, it should really be receiving them without the use of exceptions.

PR #3856, opened 2025-04-02
Fix AbortController creation in global scope

### LINKED ISSUE #3657: `AbortController` outside of request context
Creating an `AbortController` outside of the request context will throw an exception even if not used `Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope. To fix this error, perform this operation within a handler`

It could cause issue, especially while bundling other lib (for example https://github.com/opennextjs/opennextjs-cloudflare/pull/441). It can even fail silently, like in the linked PR, making it very hard to debug

Not sure if this is the expected behavior or not.

```js
const a = new AbortController()
addEventListener("fetch", event => {
  event.respondWith(new Response("Hello World"));
});
```

anonrig:
## Summary
- Makes AbortSignal check its aborted state without requiring IoContext, allowing AbortController/AbortSignal to be created in global scope without throwing an exception
- AbortSignal only requires IoContext when actually used for I/O operations, avoiding binding to a specific request context
- Adds a test that verifies AbortController creation in global scope works properly

## Test plan
- Added a test case in abortsignal-test.js that creates an AbortController in global scope (outside any request context)
- Verified the test passes with `bazel test //src/workerd/api:tests/abortsignal-test`

Fixes #3657

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -548,7 +548,7 @@ jsg::Ref<AbortSignal> AbortSignal::abort(jsg::Lock& js, jsg::Optional<jsg::JsVal
 }
 
 void AbortSignal::throwIfAborted(jsg::Lock& js) {
-  if (getCanceler().isCanceled()) {
+  if (getAborted()) {
     KJ_IF_SOME(r, reason) {
       js.throwException(r.getHandle(js));
     } else {
@@ -645,10 +645,12 @@ RefcountedCanceler& AbortSignal::getCanceler() {
 void AbortSignal::triggerAbort(
     jsg::Lock& js, jsg::Optional<kj::OneOf<kj::Exception, jsg::JsValue>> maybeReason) {
   KJ_ASSERT(flag != Flag::NEVER_ABORTS);
-  if (getCanceler().isCanceled()) {
+  if (getAborted()) {
     return;
   }
   auto exception = AbortSignal::abortException(js, maybeReason);
+  pendingException = kj::cp(exception);
+
   KJ_IF_SOME(r, maybeReason) {
     KJ_SWITCH_ONEOF(r) {
       KJ_CASE_ONEOF(value, jsg::JsValue) {
@@ -661,7 +663,11 @@ void AbortSignal::triggerAbort(
   } else {
     reason = js.exceptionToJsValue(kj::mv(exception));
   }
-  getCanceler().cancel(kj::cp(exception));
+
+  // If we already have a canceler, cancel it
+  KJ_IF_SOME(c, canceler) {
+    c->cancel(kj::cp(exception));
+  }
 
   // This is questionable only because it goes against the spec but it does help prevent
   // memory leaks. Once the abort signal has been triggered, there's really nothing else
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -495,8 +495,7 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
 AbortSignal::AbortSignal(kj::Maybe<kj::Exception> exception,
     jsg::Optional<jsg::JsRef<jsg::JsValue>> maybeReason,
     Flag flag)
-    : canceler(
-          IoContext::current().addObject(kj::refcounted<RefcountedCanceler>(kj::cp(exception)))),
+    : pendingException(kj::mv(exception)),
       flag(flag),
       reason(kj::mv(maybeReason)) {}
 
@@ -549,7 +548,7 @@ jsg::Ref<AbortSignal> AbortSignal::abort(jsg::Lock& js, jsg::Optional<jsg::JsVal
 }
 
 void AbortSignal::throwIfAborted(jsg::Lock& js) {
-  if (canceler->isCanceled()) {
+  if (getCanceler().isCanceled()) {
     KJ_IF_SOME(r, reason) {
       js.throwException(r.getHandle(js));
     } else {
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -33,3 +33,5 @@ package-lock.json
 # Bazel plugin for Intellij paths
 .clwb
 .aswb
+
+**/CLAUDE.local.md
```
anonrig, line 37:
```suggestion
```

src/workerd/api/tests/abortsignal-test.js
```diff
@@ -188,3 +188,30 @@ export const onabortPrototypeProperty = {
     strictEqual(ac.signal.onabort, handler);
   },
 };
+
+// Create a global AbortController instance at module scope to verify
+// that it's possible to create AbortControllers in the global scope
+// This test validates that the fix for https://github.com/cloudflare/workerd/issues/3657 works properly
+const globalAbortController = new AbortController();
```
jasnell, line 195: There are a number of issues with this that make this a bit iffy... While this does allow creating the `AbortController` at the global scope, it will end up binding it to the first `IoContext` that accesses it, so any attempt access it from other requests will fail, and any attempt to access it at the global scope will fail. Because of that I'm really not sure if this change is worthwhile. Interested in @kentonv's thoughts.

anonrig: According to AI: 

A better approach might be to:

  1. Completely decouple the AbortSignal's "aborted" state from the IoContext
  2. Only require IoContext when the signal is actually being used to abort I/O operations
  3. Keep track of abort state using simple boolean flags and reason variables
  4. Create the IoContext-dependent canceler only when the signal is attached to I/O

  This would allow AbortController/AbortSignal to be fully usable in global scope for its core functionality, and only throw
  exceptions when actually trying to integrate with I/O operations.

anonrig:
One possible change on top of this is 

```diff
diff --git a/src/workerd/api/basics.c++ b/src/workerd/api/basics.c++
index a56c880b..5fdb058b 100644
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -548,7 +548,7 @@ jsg::Ref<AbortSignal> AbortSignal::abort(jsg::Lock& js, jsg::Optional<jsg::JsVal
 }
 
 void AbortSignal::throwIfAborted(jsg::Lock& js) {
-  if (getCanceler().isCanceled()) {
+  if (getAborted()) {
     KJ_IF_SOME(r, reason) {
       js.throwException(r.getHandle(js));
     } else {
@@ -645,10 +645,12 @@ RefcountedCanceler& AbortSignal::getCanceler() {
 void AbortSignal::triggerAbort(
     jsg::Lock& js, jsg::Optional<kj::OneOf<kj::Exception, jsg::JsValue>> maybeReason) {
   KJ_ASSERT(flag != Flag::NEVER_ABORTS);
-  if (getCanceler().isCanceled()) {
+  if (getAborted()) {
     return;
   }
   auto exception = AbortSignal::abortException(js, maybeReason);
+  pendingException = kj::cp(exception);
+
   KJ_IF_SOME(r, maybeReason) {
     KJ_SWITCH_ONEOF(r) {
       KJ_CASE_ONEOF(value, jsg::JsValue) {
@@ -661,7 +663,11 @@ void AbortSignal::triggerAbort(
   } else {
     reason = js.exceptionToJsValue(kj::mv(exception));
   }
-  getCanceler().cancel(kj::cp(exception));
+
+  // If we already have a canceler, cancel it
+  KJ_IF_SOME(c, canceler) {
+    c->cancel(kj::cp(exception));
+  }
 
   // This is questionable only because it goes against the spec but it does help prevent
   // memory leaks. Once the abort signal has been triggered, there's really nothing else
diff --git a/src/workerd/api/basics.h b/src/workerd/api/basics.h
index 4c806f7e..f7408380 100644
--- a/src/workerd/api/basics.h
+++ b/src/workerd/api/basics.h
@@ -523,7 +523,10 @@ class AbortSignal final: public EventTarget {
   static jsg::Ref<AbortSignal> constructor() = delete;
 
   bool getAborted() {
-    return getCanceler().isCanceled();
+    KJ_IF_SOME(c, canceler) {
+      return c->isCanceled();
+    }
+    return pendingException != kj::none;
   }
 
   jsg::JsValue getReason(jsg::Lock& js);
@@ -577,8 +580,7 @@ class AbortSignal final: public EventTarget {
   // Allows this AbortSignal to also serve as a kj::Canceler
   template <typename T>
   kj::Promise<T> wrap(kj::Promise<T> promise) {
-    JSG_REQUIRE(
-        !getCanceler().isCanceled(), TypeError, "The AbortSignal has already been triggered");
+    JSG_REQUIRE(!getAborted(), TypeError, "The AbortSignal has already been triggered");
     return getCanceler().wrap(kj::mv(promise));
   }
```

src/workerd/api/basics.c++
```diff
@@ -631,16 +630,27 @@ void AbortSignal::visitForGc(jsg::GcVisitor& visitor) {
 }
 
 RefcountedCanceler& AbortSignal::getCanceler() {
```
jasnell, line 632: This approach just is not going to work. To understand why, consider the following case:

```js
// Global scope

export default {
  fetch(req) {
    ac.signal.throwIfAborted();
  }
}
```

Now let's say you throw two separate requests at this worker. The `AbortController` will end up being bound to the first received requests `IoContext` via the create `IoOwn`. When the subsequent request comes in, it will cause an error because of that internal `IoOwn` belonging to a different `IoContext`. 

What would be a better approach, if we think `AbortController` really needs to be supported at the global scope, would be to remove the underlying use of `kj::Canceler` in its current form and instead of relying on that, cancel via a registered event listener.

Ultimately I don't think Claudes approach here is at all correct.

PR #3857, opened 2025-04-02
[nfc] 04-01-25 build and V8 cleanup

fhanau:
- Drop obsolete V8 patch
- Update compile_flags.txt to support Apple Clang 17 (ships with latest Xcode minor release, based on LLVM 19)
- Silence V8 unused function warning as found with LLVM19, drop warning overrides that are no longer needed.


jasnell (approved):

danlapid:
Internal PR seems to be a real failure

danlapid (approved):

fhanau:
> Internal PR seems to be a real failure

None of the files being modified here are being used in the downstream build, so I took a closer look – based on the internal PR pipeline, a downstream test fails to build as this branch does not have #3826, which the downstream repo was already expecting to have. Maybe we can make the internal build CI job "smarter" so that it can detect such cases, or we'd need to make sure that workerd PRs are in sync with the workerd commit used by downstream.

danlapid:
> > Internal PR seems to be a real failure
> 
> None of the files being modified here are being used in the downstream build, so I took a closer look – based on the internal PR pipeline, a downstream test fails to build as this branch does not have #3826, which the downstream repo was already expecting to have. Maybe we can make the internal build CI job "smarter" so that it can detect such cases, or we'd need to make sure that workerd PRs are in sync with the workerd commit used by downstream.

Maybe github has an option to run CI only on your-branch-rebased-on-main?

PR #3858, opened 2025-04-02
Wait for PendingEvents to clear when draining actors

ObsidianMinor:
With JS RPC we've noticed tasks getting scheduled after all incoming requests have been dropped and tried draining waitUntilTasks. This happens because often times JS RPC sessions get canceled prematurely and don't end normally.

In a normal request, we wait until the request, or "session", is "done". When a request is "done", all capabilities between the caller and callee have been dropped. In that case, each capability linked to a stub would have been deconstructed and scheduled any stub disposal tasks already through waitUntilTasks. In a canceled request, we immediately try to drain the current set of waitUntilTasks. However, at this point, there might not be any tasks to run, but there will be tasks in the future. Once the drain process is finished, the IncomingRequest is dropped. But as the stubs from the request are released by the RPC system, as the whole request unwinds, the stubs will try to schedule disposal tasks. But there is no IncomingRequest at this point! So the task fails.

With this, actors now create PendingEvents but they still don't get finalized like a normal stateless request. This also updates the PendingEvent system to support exporting a Promise that resolves when there are no more PendingEvents, much like `TaskSet::onEmpty()`. This is used when draining actor requests through a new function, `drainEventsAndTasks()`. This runs on a loop to wait until there are no more PendingEvents and waitUntilTasks.

This on its own breaks WebSocket hibernation, since a PendingEvent is unconditionally attached to the "upstream" promise made by coupling a WebSocket by returning it in a Response. So this changes it to only register a pending event if it's not a hibernatable WebSocket.

### DIFF
--- a/src/workerd/io/io-context.c++
+++ b/src/workerd/io/io-context.c++
@@ -431,6 +431,35 @@ void IoContext::addWaitUntil(kj::Promise<void> promise) {
   waitUntilTasks.add(kj::mv(promise));
 }
 
+kj::Promise<void> IoContext::drainPendingEvents() {
+  // Shamelessly copied from TaskSet::onEmpty
+  KJ_IF_SOME(fulfiller, finishedPendingEventsFulfiller) {
+    if (fulfiller->isWaiting()) {
+      KJ_FAIL_REQUIRE("onEmpty() can only be called once at a time");
+    }
+  }
+
+  if (pendingEvent == kj::none) {
+    return kj::READY_NOW;
+  } else {
+    auto paf = kj::newPromiseAndFulfiller<void>();
+    finishedPendingEventsFulfiller = kj::mv(paf.fulfiller);
+    return kj::mv(paf.promise);
+  }
+}
+
+kj::Promise<void> IoContext::drainEventsAndTasks() {
+  if (pendingEvent != kj::none) {
+    return drainPendingEvents().then([this]() { return drainEventsAndTasks(); });
+  }
+
+  if (!waitUntilTasks.isEmpty()) {
+    return waitUntilTasks.onEmpty().then([this]() { return drainEventsAndTasks(); });
+  }
+
+  return kj::READY_NOW;
+}
+
 // Mark ourselves so we know that we made a best effort attempt to wait for waitUntilTasks.
 kj::Promise<void> IoContext::IncomingRequest::drain() {
   waitedForWaitUntil = true;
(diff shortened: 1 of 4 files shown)


src/workerd/server/server.c++
```diff
@@ -4322,6 +4322,9 @@ kj::Promise<bool> Server::test(jsg::V8System& v8System,
       ++failCount;
     }
     KJ_LOG(DBG, kj::str(result ? "[ PASS ] "_kj : "[ FAIL ] "_kj, name));
+
+    // Wait until everything on this thread is wrapped up.
+    co_await kj::yieldUntilWouldSleep();
```
ObsidianMinor, line 4327: Is this ok? I'm not sure if maybe `finishedPendingEventsFulfiller` should actually be fulfilled before waiting for the thread to become idle. Thoughts @kentonv?

src/workerd/io/io-context.c++
```diff
@@ -513,36 +551,43 @@ IoContext::PendingEvent::~PendingEvent() noexcept(false) {
...
+      return kj::READY_NOW;
+    }
+
     return context.worker->takeAsyncLock(context.getMetrics())
         .then([&context](Worker::AsyncLock asyncLock) { context.runFinalizers(asyncLock); });
   }).eagerlyEvaluate(nullptr);
 }
 
 kj::Own<void> IoContext::registerPendingEvent() {
-  if (actor != kj::none) {
```
kentonv, line 528: Oh man, I forgot that actors don't even using pending events.

Now this makes me question whether this is the right thing to do.

One alternative hack that comes to mind... What if `TransientJsRpcTarget`, actually used `awaitIo` to await the destructor being called, and then invoke the disposer?

Like:

```
auto droppedPaf = kj::newPromiseAndFulfiller<void>();
ctx.awaitIo(droppedPaf.promise, [](jsg::Lock& js) {
  // call the disposer here
});
```

And then you'd fulfill `droppedPaf.fulfiller` in `TransientJsRpcTarget`'s destructor?

This is actually really clean I think:
- No need for a pending event; awaitIo() takes care of that itself.
- Tasks added by addTask() already delay drain() for actors (but not for non-actors).
- Destructor no longer needs to mess around with `weakIoContext`.
- Generally seems to cleanly model this as I/O, which now that I think about it, seems right.

What do you think?

Sorry for sending you on another wrong path here...

ObsidianMinor: Sounds like it'd work. Though I wonder why we specifically don't use pending event at all? I realize it's mostly because actors don't use finalization, but having a way to just say "don't count me idle until this object goes away" sounds useful.

ObsidianMinor:
I'm going to close this and open a new PR because I think there could be value in this patch in the future.

PR #3865, opened 2025-04-03
Add dateAlwaysUseUtc compatibility flag to make Date API use UTC

### LINKED ISSUE #2328: 🐛 Bug: workerd uses computer's local time for date parsing; production uses UTC
When running workerd for local testing, date-related APIs that are defined to use the local timezone actually do so, using the time zone of the machine on which workerd is running.

However, in production, Cloudflare Workers always use UTC as the time zone. This makes the local simulation inaccurate (and leads to embarrassing things like me inviting my friends to a LAN party supposedly starting at 2AM when in fact it was meant to start at noon (true story)).

workerd should strictly use UTC, to match production. Since workerd is strictly meant for servers, there's never a reason to use the machine's local time zone.

anonrig:
## Summary
- Adds a new dateAlwaysUseUtc compatibility flag that makes JavaScript Date API methods always use UTC timezone instead of local timezone
- This fixes an issue where workerd's local development environment behaves differently from production Workers runtime
- In production Cloudflare Workers, Date methods always use UTC, but locally they use the system timezone
- Implementation applies v8::Isolate::TimeZoneDetection::kSkip to enforce UTC for all Date methods
- Adds tests to verify behavior with and without the flag

Fixes #2328

---
_Created with the assistance of [Claude Code](https://claude.ai/code)_


kentonv:
I don't think this needs a compat flag since the incorrect behavior occurred in local testing only, not prod. Hence, this change cannot possibly break Workers in production on Cloudflare, which is the bar for compat flags.

npaun (approved):

jclee:
IIRC, I ran into this when testing some alternate edgeworker configs, which is why I ended up adding `setenv("TZ", "UTC", true);` to the test harness.  Might be able to remove it now, or at least update the comment.

I guess there's also LC_ALL and LANG environment vars, which [can apparently affect the format of `new Date().toLocaleString()`](https://github.com/nodejs/node/issues/23301) and maybe other stuff under node, at least, via ICU.

dom96 (approved):

src/workerd/server/workerd-api.c++
```diff
@@ -368,6 +368,13 @@ struct WorkerdApi::Impl final {
     if (maybeOwnedModuleRegistry != kj::none) {
       jsgIsolate.setUsingNewModuleRegistry();
     }
+
+    // Apply UTC timezone for Date methods
+    jsgIsolate.runInLockScope([&](JsgWorkerdIsolate::Lock& lock) {
+      lock.v8Isolate->DateTimeConfigurationChangeNotification(
+          v8::Isolate::TimeZoneDetection::kSkip);
+    });
```
jasnell, line 376: This could probably be set in `jsg/setup.h` when the isolate is created? Especially if we just want this to be the mode all isolates run in.

src/workerd/server/workerd-api.c++
```diff
@@ -368,6 +368,7 @@ struct WorkerdApi::Impl final {
     if (maybeOwnedModuleRegistry != kj::none) {
       jsgIsolate.setUsingNewModuleRegistry();
     }
+
```
jasnell, line 371: Extraneous whitespace change now...

```suggestion
```

jasnell (approved):

PR #3867, opened 2025-04-03
EW-9182 Add assertion to detect timeout ID generator mismatches

harrishancock:
Add instrumentation to verify that the TimeoutId::Generator being used belongs to the ServiceWorkerGlobalScope associated with the current IoContext. This will help diagnose potential timeout ID collisions caused by generators being used with the wrong IoContext.

- Add getTimeoutIdGenerator() method to Worker::Lock
- Add assertion in TimeoutManagerImpl::setTimeout to validate generator
- This will fail early with a descriptive error if the invariant is violated

🤖 Generated with Claude, then lightly edited by Harris Hancock.


anonrig (approved):

anonrig:
Couple of tests seem to be failing

PR #3872, opened 2025-04-03
Add sendEvent on Workflows instance

sidharthachatterjee:
Adds a `sendEvent` function on instance returned from Workflows binding


LuisDuarte1 (approved):

penalosa (approved):
This looks good, but will need the type snapshots to be updated

PR #3874, opened 2025-04-03
Remove toJSON check on new process.env population logic

jasnell:


kentonv (approved):
Would be nice to add a test that defines an RPC service binding and verifies it doesn't appear in `process.env`.

But fine with merging this as-is to fix the bug and punting the test if there isn't appetite.

PR #3875, opened 2025-04-03
do not throw on rejectUnauthorized option

anonrig:
This is making things harder for users that depend on libraries that does not expose default configurations.

### DIFF
--- a/src/node/internal/internal_tls_wrap.ts
+++ b/src/node/internal/internal_tls_wrap.ts
@@ -190,9 +190,8 @@ export function TLSSocket(
     throw new ERR_OPTION_NOT_IMPLEMENTED('options.requestCert');
   }
 
-  if (tlsOptions.rejectUnauthorized) {
-    // Used by servers to reject unauthorized connections.
-    // Does not apply to Cloudflare Workers.
+  if (tlsOptions.rejectUnauthorized === false) {
+    // TODO(soon): We don't support rejectUnauthorized=false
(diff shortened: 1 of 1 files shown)


jasnell (approved):
LGTM but I wonder if we shouldn't at least emit a warning in cases like this... of course, we don't currently have a way of emitting such warnings from JavaScript currently so that can be done separately if it is desirable.

anonrig:
> LGTM but I wonder if we shouldn't at least emit a warning in cases like this... of course, we don't currently have a way of emitting such warnings from JavaScript currently so that can be done separately if it is desirable.

This unblocks mysql2. I recommend doing this separately. I can add a TODO for now.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -598,8 +597,8 @@ export function connect(...args: unknown[]): TLSSocket {
     throw new ERR_OPTION_NOT_IMPLEMENTED('options.minDHSize');
   }
 
-  if (options.rejectUnauthorized) {
-    // Not supported.
+  if (options.rejectUnauthorized === false) {
+    // TODO(soon): We don't support rejectUnauthorized=false
```
danlapid, line 601:
```suggestion
    // TODO(soon): We don't support rejectUnauthorized=true
```

danlapid (approved):
Thanks for fixing this 🙏

jasnell (approved):

PR #3876, opened 2025-04-03
[build] Run tests previously disabled under ASan

fhanau:
We explicitly enable V8's ASan/LSan support here, and add a LSan suppression from the canonical GN build.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -199,6 +199,8 @@ build:asan --config=sanitizer-common
 build:asan --copt="-fsanitize=address" --linkopt="-fsanitize=address"
 build:asan --test_env=ASAN_OPTIONS=abort_on_error=true
 build:asan --test_env=KJ_CLEAN_SHUTDOWN=1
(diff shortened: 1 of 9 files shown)


npaun (approved):

danlapid (approved):

danlapid (approved):

dom96 (approved):

fhanau:
@mikea this is the PR I mentioned (LSan violation which seemed to be fixed but was actually just being suppressed in the GN build). This approach is preferable IMHO as we can enable ASan on the tests and just need to add the LSan suppression, with a TODO to investigate later.

fhanau:
@penalosa need another wrangler team approval here (small change to src/cloudflare BUILD files again)

penalosa:
@fhanau I think it would probably make sense to re-add the runtime team to src/cloudflare—Wrangler doesn't have a whole lot of context here

penalosa (approved):

PR #3879, opened 2025-04-04
add navigator.hardwareConcurrency

anonrig:
Adding navigator.hardwareConcurrency because it also exists on Node.js. There are libraries that depend on this to run tests etc.

### DIFF
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -84,12 +84,21 @@ class Navigator: public jsg::Object {
 
   bool sendBeacon(jsg::Lock& js, kj::String url, jsg::Optional<Body::Initializer> body);
 
+  kj::uint getHardwareConcurrency() {
+    // Workers does not expose hardware concurrency to users.
+    // From the user code perspective there's only one core.
+    return 1;
+  }
+
   JSG_RESOURCE_TYPE(Navigator) {
     JSG_METHOD(sendBeacon);
     JSG_READONLY_INSTANCE_PROPERTY(userAgent, getUserAgent);
+    JSG_READONLY_INSTANCE_PROPERTY(hardwareConcurrency, getHardwareConcurrency);
 #ifdef WORKERD_EXPERIMENTAL_ENABLE_WEBGPU
     JSG_READONLY_INSTANCE_PROPERTY(gpu, getGPU);
 #endif
+
+    JSG_TS_OVERRIDE({ readonly hardwareConcurrency: number; });
   }
 };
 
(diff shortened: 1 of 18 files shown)


npaun (approved):

danlapid (approved):

anonrig:
cc @cloudflare/wrangler can you review?

src/workerd/api/global-scope.h
```diff
@@ -84,12 +84,19 @@ class Navigator: public jsg::Object {
...
+
   JSG_RESOURCE_TYPE(Navigator) {
     JSG_METHOD(sendBeacon);
     JSG_READONLY_INSTANCE_PROPERTY(userAgent, getUserAgent);
+    JSG_READONLY_INSTANCE_PROPERTY(hardwareConcurrency, getHardwareConcurrency);
 #ifdef WORKERD_EXPERIMENTAL_ENABLE_WEBGPU
     JSG_READONLY_INSTANCE_PROPERTY(gpu, getGPU);
 #endif
+
+    JSG_TS_OVERRIDE({ readonly hardwareConcurrency: number; });
```
jasnell, line 99: Why does this need an override?

anonrig: There is a bug with type generation logic. You can see the change history of the type bot comment: https://github.com/cloudflare/workerd/pull/3879#issuecomment-2777290994

jasnell: A comment would be helpful then to keep track.

danlapid: So this is not actually a bug with the type generation logic, it's a bug with the workflow of type generation and it's actually that you didn't commit the type changes.

In type-snapshot.yml you can see `diff -r bazel-bin/types/definitions/ types/generated-snapshot > types.diff` this is actually incorrect, it's the wrong way around, it's comparing new to old instead of old to new so the type snapshot comment shows additions as deletions and vice versa.

When the bot posted that the types mismatch it was because you have not yet committed the types (as can be seen by a later change in which you committed the types).

When the bot commented that it would have "removed" the hardware concurrency type it actually meant that it would add(!) it.

So types generation is working right!
You can remove this override.

danlapid: PR to fix the workflow:
https://github.com/cloudflare/workerd/pull/3883

penalosa (approved):
Approved from a types perspective

src/workerd/api/global-scope.h
```diff
@@ -84,12 +84,21 @@ class Navigator: public jsg::Object {
...
```
danlapid, line 101: Safe to assume that a followup to remove this is coming right?

PR #3886, opened 2025-04-04
add process.features* to node:process

anonrig:
This adds non-experimental and non-deprecated features of `process.features` object for nodejs_compat.

### DIFF
--- a/src/node/internal/process.ts
+++ b/src/node/internal/process.ts
@@ -191,10 +191,27 @@ export function exit(code: number) {
 
 export const platform = utilImpl.processPlatform;
 
+// The following features does not include deprecated or experimental flags mentioned in
+// https://nodejs.org/docs/latest/api/process.html
+export const features = Object.freeze({
+  // A boolean value that is true if the current Node.js build is caching builtin modules.
+  cached_builtins: true,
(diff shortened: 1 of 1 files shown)


src/node/internal/process.ts
```diff
@@ -191,10 +191,27 @@ export function exit(code: number) {
...
+  // A boolean value that is true if the current Node.js build is caching builtin modules.
+  cached_builtins: true,
+  // A boolean value that is true if the current Node.js build is a debug build.
+  debug: false,
+  // A boolean value that is true if the current Node.js build includes the inspector.
+  inspector: false,
+  // A boolean value that is true if the current Node.js build supports loading ECMAScript modules using require().
+  // TODO(soon): Update this when we support ESM modules through require().
+  require_module: false,
+  // A boolean value that is true if the current Node.js build includes support for TLS.
```
jasnell, line 206: Are these actually being used by anything that users are trying to use in workers?

jasnell (approved):

src/node/internal/process.ts
anonrig: I'm not sure. People might depend on these functionality and lack of having them might break. For example any package that calls `process.features.require_module` in workerd will throw with require_module of undefined error without this change.

PR #3889, opened 2025-04-05
[build] 04-05-25 Build maintenance

fhanau:
[build] 04-05-25 Build maintenance

- Build simdutf using SSE4.2 to avoid having to include fallback code path, reducing code size. We already require SSE4.2 for zlib, so this won't break compatibility with any systems.
- Minor Bazel/include cleanup based on IWYU
- Clean up redundant Bazel arguments
- Clean up comments based on no longer using Dawn

[build] Build using lld 20 on macOS
lld20 introduces a symbol string deduplication pass, reducing binary sizes on macOS (https://github.com/llvm/llvm-project/pull/123874). For debug binaries with linker optimizations turned on, binary size improves by 930KB to 235.4MB, gains for release binaries should be similar.
Run `brew update` on CI to make this version available.


src/workerd/api/node/BUILD.bazel
```diff
@@ -9,28 +9,26 @@ wd_cc_library(
...
-        "//src/rust/cxx-integration",
-        "//src/rust/dns",
-        "//src/workerd/io",
         "@capnp-cpp//src/kj/compat:kj-gzip",
     ],
     visibility = ["//visibility:public"],
     deps = [
         ":node-core",
         "//src/node",
+        "//src/workerd/io",
```
anonrig, line 31: Any particular reason to why we are moving workerd/io from implementation_deps to deps? Can you add a comment?

.github/workflows/release.yml
```diff
@@ -124,6 +124,7 @@ jobs:
         run: |
           # Install lld and link it to /usr/local/bin. We overwrite any existing link, which may
           # exist from an older pre-installed LLVM version on the runner image.
+          brew update
```
anonrig, line 127: Q: Why do we do a brew update?

src/workerd/api/node/BUILD.bazel
anonrig: I think we were following "include what you use" pattern? Why not keep rust/dns here?

anonrig (approved):

src/workerd/api/node/BUILD.bazel
fhanau: 1) The "node" target has headers that depend on io target headers, for example src/workerd/api/node/crypto.h including workerd/api/crypto/crypto.h. Thus targets that depend on node need to have headers from the io target available, and we need to have io as a full dependency.
2) Absolutely – dns.c++ moves to the "node-core" target since it does not depend on io, thus rust/dns should also move based on IWYU.

danlapid (approved):

PR #3893, opened 2025-04-07
Add platform-agnostic formatter targets

danlapid:
Create a simplified bazel setup for code formatters that:
1. Adds platform-agnostic formatter targets under build/deps/formatters
2. Updates the format.py script to use the new targets
3. Adds VS Code integration for the formatters in settings.json
4. Uses copy_file to create physical files in bazel-bin

🤖 Generated in collaboration with [Claude Code](https://claude.ai/code)

### DIFF
--- a/build/deps/formatters/BUILD
+++ b/build/deps/formatters/BUILD
@@ -1,0 +1,79 @@
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+
+config_setting(
+    name = "linux_amd64",
+    constraint_values = [
+        "@platforms//os:linux",
+        "@platforms//cpu:x86_64",
+    ],
+)
+
+config_setting(
+    name = "linux_arm64",
+    constraint_values = [
+        "@platforms//os:linux",
+        "@platforms//cpu:arm64",
+    ],
+)
+
+config_setting(
+    name = "darwin_amd64",
+    constraint_values = [
+        "@platforms//os:macos",
+        "@platforms//cpu:x86_64",
+    ],
+)
+
+config_setting(
+    name = "darwin_arm64",
+    constraint_values = [
+        "@platforms//os:macos",
+        "@platforms//cpu:arm64",
+    ],
+)
+
+config_setting(
+    name = "windows_amd64",
+    constraint_values = [
+        "@platforms//os:windows",
+        "@platforms//cpu:x86_64",
+    ],
+)
+
+copy_file(
+    name = "clang-format@rule",
+    src = select({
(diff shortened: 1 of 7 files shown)


danlapid:
I still haven't figure out the internal PR for this one

build/deps/formatters/formatters.bzl
```diff
@@ -0,0 +1,13 @@
+# Platform-agnostic formatter targets for use in WORKSPACE
+
+def formatter_aliases():
```
fhanau, line 3: This file does nothing.

npaun (approved):
Other than the useless file Felix mentioned, this looks nice

build/deps/shared_deps.jsonc
```diff
@@ -49,23 +49,23 @@
       "owner": "astral-sh",
       "repo": "ruff",
       "file_regex": "^ruff-aarch64-apple-darwin.tar.gz$",
-      "build_file_content": "filegroup(name='file', srcs=glob(['**']))"
+      "build_file_content": "filegroup(name='file', srcs=['ruff'], visibility=['//visibility:public'])"
```
fhanau, line 52: Nit: I think it's cleaner to continue to still use a glob and capture all files here, the actual name of the binary feels like an implementation detail here that the dependency definition doesn't need to know about.

build/deps/formatters/BUILD
```diff
@@ -0,0 +1,82 @@
+# OS x Architecture config settings for formatter aliases
```
fhanau, line 1: Nit: misleading comment

build/deps/shared_deps.jsonc
danlapid: If we keep using a glob then copy_file fails on there being more than one file 😢

build/deps/formatters/BUILD
```diff
@@ -0,0 +1,79 @@
...
+config_setting(
+    name = "windows_amd64",
+    constraint_values = [
+        "@platforms//os:windows",
+        "@platforms//cpu:x86_64",
+    ],
+)
+
+copy_file(
+    name = "clang-format@rule",
```
fhanau, line 44: Nit: I think of "@rule" target names as an antipattern as they don't tell us anything about what the rule actually does. We can just drop that suffix for these targets and drop it in format.py and things should still work right?

danlapid: There's a bazel warning hating that the name and the output are identical.

fhanau: 🤔 I thought it'd be fine, that's what we do in other cases too. Either way, it's non-blocking.

PR #3894, opened 2025-04-07
Switch to std::list from std::dequeue in web socket

jasnell:
These will end up being very minor impact but worth doing. Trades the less memory-efficient std::deque for the slightly less performant but more memory-efficient std::list in a couple of cases.


danlapid (approved):

jasnell:
`internal-build` CI check failed for an unrelated reason. Manually tested and verified that it works on the internal repo.

PR #3896, opened 2025-04-07
Use v8::LocalVector where appropriate

jasnell:
KJ_MAP will allocate the array of v8::Local<T>'s on the heap which is technically not allowed. While most of these places are not necessarily problematic as is, the general pattern is. Let's prefer to use `v8::LocalVector` where appropriate instead.


anonrig:
Can you also double check if we can enable this flag now Ref: https://github.com/cloudflare/workerd/pull/2752

jasnell:
I don't believe we can just yet. There are still a few of these out there.

anonrig (approved):

PR #3897, opened 2025-04-07
[o11y] Use BaseTracer class to support tracer polymorphism

fhanau:
Also see the downstream PR.


src/workerd/io/tracer.h
```diff
@@ -101,8 +101,13 @@ class PipelineTracer final: public kj::Refcounted, public kj::EnableAddRefToThis
 // of the add* methods below should make more sense: The invocation span context below is currently
 // only being used in the streaming model, when we have switched the legacy model to streaming
 // there will be plenty of cleanup potential.
-class BaseTracer {
+class BaseTracer: public kj::Refcounted {
  public:
+  BaseTracer(): self(kj::refcounted<WeakRef<BaseTracer>>(kj::Badge<BaseTracer>{}, *this)) {}
+  ~BaseTracer() {
```
danlapid, line 107: This should be virtual

jasnell, line 107:
```suggestion
  virtual ~BaseTracer() {
```

jasnell (approved):
LGTM with the one virtual destructor nit fixed.

danlapid (approved):

src/workerd/io/trace-stream.c++
```diff
@@ -971,6 +971,8 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
     rpc::EventDispatcher::Client dispatcher) {
   auto revokePaf = kj::newPromiseAndFulfiller<void>();
 
+  // TODO(streaming-tail): Session is currently being reported as canceled when using
```
fhanau, line 974: The tail stream gets reported as canceled when running the downstream STW test via RPC – this code is largely copied from JSRPC, unfortunately I do not understand it well enough to investigate this properly. Note that this is an entirely different stream than the TailStreamTarget stream where previous issues have been fixed, even though the error message is identical. Tests are still passing, so I'm just adding a TODO here for now.

danlapid: Likely the waitUntilTasks

PR #3898, opened 2025-04-07
Add new AsyncLocalStorage constructor options

jasnell:
These were just today added to Node.js implementation. We likely should hold off merging this until the new options go out in a Node.js release. Just opening this now to get ready. I don't plan to merge this until the first Node.js release with the options is cut.

### DIFF
--- a/src/workerd/api/node/async-hooks.h
+++ b/src/workerd/api/node/async-hooks.h
@@ -29,12 +29,25 @@ namespace workerd::api::node {
 //   console.log(als.getStore());  // undefined
 class AsyncLocalStorage final: public jsg::Object {
  public:
-  AsyncLocalStorage(): key(kj::refcounted<jsg::AsyncContextFrame::StorageKey>()) {}
+  struct AsyncLocalStorageOptions {
+    jsg::Optional<jsg::JsRef<jsg::JsValue>> defaultValue;
+    jsg::Optional<kj::String> name;
+    JSG_STRUCT(defaultValue, name)
+  };
+
+  AsyncLocalStorage(jsg::Optional<AsyncLocalStorageOptions> options = kj::none)
+      : key(kj::refcounted<jsg::AsyncContextFrame::StorageKey>()) {
+    KJ_IF_SOME(opt, options) {
+      defaultValue = kj::mv(opt.defaultValue), name = kj::mv(opt.name);
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


src/workerd/api/node/async-hooks.h
```diff
@@ -29,12 +29,25 @@ namespace workerd::api::node {
...
+  struct AsyncLocalStorageOptions {
+    jsg::Optional<jsg::JsRef<jsg::JsValue>> defaultValue;
+    jsg::Optional<kj::String> name;
+    JSG_STRUCT(defaultValue, name)
+  };
+
+  AsyncLocalStorage(jsg::Optional<AsyncLocalStorageOptions> options = kj::none)
+      : key(kj::refcounted<jsg::AsyncContextFrame::StorageKey>()) {
+    KJ_IF_SOME(opt, options) {
+      defaultValue = kj::mv(opt.defaultValue), name = kj::mv(opt.name);
```
npaun, line 41: Nit: I don't like using the comma operator in cases like this but that is just my personal take.

```suggestion
      defaultValue = kj::mv(opt.defaultValue);
      name = kj::mv(opt.name);
```

npaun (approved):

src/workerd/api/node/async-hooks.h
jasnell: huh, I don't remember actually typing it that way. Wonder if the formatter changed it. Either way, yep, I agree with the suggestion!

src/workerd/api/node/async-hooks.h
```diff
@@ -29,12 +29,26 @@ namespace workerd::api::node {
 //   console.log(als.getStore());  // undefined
 class AsyncLocalStorage final: public jsg::Object {
  public:
-  AsyncLocalStorage(): key(kj::refcounted<jsg::AsyncContextFrame::StorageKey>()) {}
+  struct AsyncLocalStorageOptions {
+    jsg::Optional<jsg::JsRef<jsg::JsValue>> defaultValue;
+    jsg::Optional<kj::String> name;
```
anonrig, line 34: Any reason to why we are not storing this as JsString since it has a getter and it would avoid unnecessary copy whenever the getter is called?

jasnell: We have so many of these that I'll just update them at once or in a series of large batches when the time comes.

PR #3899, opened 2025-04-07
Replace kj::Array<kj::byte> args with jsg::BufferSource

jasnell:
More work moving away from the kj::Array<kj::byte> type mapping.

While this won't have any impact on memory accounting, I'm on a mission to eliminate the `kj::Array<kj::byte>` type mapping entirely....


anonrig (approved):

jasnell:
looks like there's a relevant failing test. Will look into that tomorrow.

PR #3902, opened 2025-04-08
Schedule JS RPC target disposal on construction

ObsidianMinor:
Rather than wait until the target is dropped, we schedule a task on construction and fulfill it on destruction.


src/workerd/api/worker-rpc.c++
```diff
@@ -1544,17 +1530,12 @@ class TransientJsRpcTarget final: public JsRpcTargetBase {
  private:
   struct Handles {
     jsg::JsRef<jsg::JsObject> object;
-    kj::Maybe<jsg::V8Ref<v8::Function>> dispose;
 
     // Discposers for stubs or other things inside `object` which need to be dropped as soon as
     // the pipeline is dropped.
     kj::Vector<kj::Own<void>> stubDisposers;
```
kentonv, line 1536: Hmm, I think we need to make sure the `stubDisposers` are destroyed after calling the main disposer, like was done before. Otherwise, when the main disposer is called it'll find that the object it is called on has already been hollowed out (all the disposers contained within are already dead).

Unfortunately this will be slightly annoying since the `stubDisposers` are currently initialized _after_ initial construction of the `TransientJsRpcTarget` (but immediately after).

ObsidianMinor: I refactored it so we initialize them on construction. But it's mildly messy so I'm tempted to introduce some helper functions for making clients.

kentonv (approved):

PR #3906, opened 2025-04-08
Code owners: Add workers-runtime team back to src/cloudflare

fhanau:
As suggested at https://github.com/cloudflare/workerd/pull/3876#issuecomment-2786890320 – some src/cloudflare changes are not relevant to the Runtime team. Hopefully the wrangler team will still get assigned preferentially based on being listed first?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,7 +6,7 @@ RELEASE.md @cloudflare/wrangler @cloudflare/workers-runtime-1
 package.json @cloudflare/wrangler @cloudflare/workers-runtime-1
 pnpm-lock.yaml @cloudflare/wrangler @cloudflare/workers-runtime-1
 /types/ @cloudflare/wrangler
-/src/cloudflare/ @cloudflare/wrangler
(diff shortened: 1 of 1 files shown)


penalosa (approved):

danlapid (changes requested):
Let's be more specific if we have to, or just wait for their review, it's not that bad.

We would get auto assigned because our team is configured to auto assigned while theirs isn't.

penalosa:
@danlapid I don't think any `src/cloudflare` changes are really relevant to the Wrangler team—are there specific sections you're thinking of? AFAIK these contain code that's part of workerd and are unrelated to type definitions

danlapid:
> @danlapid I don't think any `src/cloudflare` changes are really relevant to the Wrangler team—are there specific sections you're thinking of? AFAIK these contain code that's part of workerd and are unrelated to type definitions

All of the types in https://github.com/cloudflare/workerd/tree/main/src/cloudflare/internal are better off reviewed by the wrangler team imo.
The runtime team can't really do a good job reviewing those.

penalosa:
@danlapid those types aren't exposed to users though (the ones in types/defines are, which admittedly are usually basically a duplicate, I think).

PR #3907, opened 2025-04-08
Bump jsrpc message limit to 32 MB

danlapid:


src/workerd/api/worker-rpc.h
```diff
@@ -28,7 +28,7 @@ namespace workerd::api {
 // as well as put too much memory pressure on the isolate. Applications which need to move
 // large amounts of data should split the data into several smaller chunks transmitted through
 // separate calls.
```
kentonv, line 30: I suppose the comment needs updating.

The new limit is based on the fact that Cap'n Proto's default total message size limit is 64MB, and we want to stay clear of that. (Also with the isolate total limit being 128MB, getting much larger without blowing heap limits would be hard.)

jasnell (approved):

PR #3908, opened 2025-04-08
improve clearTimeout and clearInterval compatibility

anonrig:
Browsers do not throw an error when passed NaN value to clearTimeout and clearInterval. On workerd, this throws an error under UBsan builds.

```
src/workerd/io/_virtual_includes/io-helpers/workerd/io/io-timers.h:30:32: runtime error: nan is outside the range of representable values of type 'unsigned long'
```

This fixes this issue.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -794,14 +794,18 @@ TimeoutId::NumberType ServiceWorkerGlobalScope::setTimeout(jsg::Lock& js,
     function(js, kj::mv(args));
   };
   auto timeoutId = IoContext::current().setTimeoutImpl(timeoutIdGenerator,
-      /* repeats = */ false, [function = kj::mv(fn)](jsg::Lock& js) mutable { function(js); },
+      /* repeat */ false, [function = kj::mv(fn)](jsg::Lock& js) mutable { function(js); },
       msDelay.orDefault(0));
   return timeoutId.toNumber();
 }
 
-void ServiceWorkerGlobalScope::clearTimeout(kj::Maybe<TimeoutId::NumberType> timeoutId) {
-  KJ_IF_SOME(id, timeoutId) {
-    IoContext::current().clearTimeoutImpl(TimeoutId::fromNumber(id));
+void ServiceWorkerGlobalScope::clearTimeout(jsg::Lock& js, kj::Maybe<jsg::JsNumber> timeoutId) {
+  KJ_IF_SOME(rawId, timeoutId) {
+    // Browsers does not throw an error when "unsafe" integers are passed to the clearTimeout method.
+    // Let's make sure we ignore those values, just like browsers and other runtimes.
+    KJ_IF_SOME(id, rawId.toSafeInteger(js)) {
+      IoContext::current().clearTimeoutImpl(TimeoutId::fromNumber(id));
+    }
   }
 }
 
@@ -818,9 +822,13 @@ TimeoutId::NumberType ServiceWorkerGlobalScope::setInterval(jsg::Lock& js,
     function(js, jsg::Arguments(kj::mv(argv)));
   };
   auto timeoutId = IoContext::current().setTimeoutImpl(timeoutIdGenerator,
-      /* repeats = */ true, [function = kj::mv(fn)](jsg::Lock& js) mutable { function(js); },
+      /* repeat */ true, [function = kj::mv(fn)](jsg::Lock& js) mutable { function(js); },
       msDelay.orDefault(0));
   return timeoutId.toNumber();
+}
+
+void ServiceWorkerGlobalScope::clearInterval(jsg::Lock& js, kj::Maybe<jsg::JsNumber> timeoutId) {
+  clearTimeout(js, kj::mv(timeoutId));
 }
 
 jsg::Ref<Crypto> ServiceWorkerGlobalScope::getCrypto(jsg::Lock& js) {
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -537,7 +537,7 @@ class ServiceWorkerGlobalScope: public WorkerGlobalScope {
       jsg::Function<void(jsg::Arguments<jsg::Value>)> function,
       jsg::Optional<double> msDelay,
       jsg::Arguments<jsg::Value> args);
-  void clearTimeout(kj::Maybe<TimeoutId::NumberType> timeoutId);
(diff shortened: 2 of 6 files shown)


src/workerd/jsg/jsvalue.c++
```diff
@@ -200,11 +200,28 @@ JsValue JsObject::getPrototype(Lock& js) {
...
+  if (std::abs(value) <= static_cast<double>(MAX_SAFE_INTEGER)) return true;
+  return false;
+}
+
 bool JsValue::isTruthy(Lock& js) const {
   KJ_ASSERT(!inner.IsEmpty());
   return inner->BooleanValue(js.v8Isolate);
 }
 
+double JsValue::toNumber(Lock& js) const {
```
jasnell, line 219: Would recommend `toSafeInteger(...)` as the name and enforcing the `isSafeInteger(...)` assert.

src/workerd/api/global-scope.h
```diff
@@ -537,17 +537,15 @@ class ServiceWorkerGlobalScope: public WorkerGlobalScope {
       jsg::Function<void(jsg::Arguments<jsg::Value>)> function,
       jsg::Optional<double> msDelay,
       jsg::Arguments<jsg::Value> args);
-  void clearTimeout(kj::Maybe<TimeoutId::NumberType> timeoutId);
+  void clearTimeout(jsg::Lock& js, kj::Maybe<jsg::JsValue> timeoutId);
```
jasnell, line 540: This could probably be `jsg::JsNumber`? That might avoid the need for the type override.

src/workerd/api/global-scope.c++
```diff
@@ -794,14 +794,17 @@ TimeoutId::NumberType ServiceWorkerGlobalScope::setTimeout(jsg::Lock& js,
...
       msDelay.orDefault(0));
   return timeoutId.toNumber();
 }
 
-void ServiceWorkerGlobalScope::clearTimeout(kj::Maybe<TimeoutId::NumberType> timeoutId) {
-  KJ_IF_SOME(id, timeoutId) {
-    IoContext::current().clearTimeoutImpl(TimeoutId::fromNumber(id));
+void ServiceWorkerGlobalScope::clearTimeout(jsg::Lock& js, kj::Maybe<jsg::JsValue> timeoutId) {
+  KJ_IF_SOME(rawId, timeoutId) {
+    if (rawId.isSafeInteger(js)) {
```
jasnell, line 804: A comment here would be helpful.

src/workerd/jsg/jsvalue.c++
anonrig: if I enforce isSafeInteger assert in here, then in an ideal scenario we check the safety twice. For example

```
void ServiceWorkerGlobalScope::clearTimeout(jsg::Lock& js, kj::Maybe<jsg::JsValue> timeoutId) {
  KJ_IF_SOME(rawId, timeoutId) {
    if (rawId.isSafeInteger(js)) {
      TimeoutId::NumberType id = rawId.toSafeInteger(js);
      IoContext::current().clearTimeoutImpl(TimeoutId::fromNumber(id));
    }
  }
}
```

isSafeInteger() make a check and toSafeInteger() also checks for isSafeInteger(). I didn't want to rename it as toSafeInteger() to make the assumption that the value is "safe" to avoid checking the same thing twice.

I'm open to suggestions for the naming.

jasnell: We can simplify this pattern using `kj::Maybe` semantics... That is, have `toSafeInteger(...)` return a `kj::Maybe<...>` ... such that if `kj::none` is returned the assumption is that it is not a safe integer.

```cpp
void ServiceWorkerGlobalScope::clearTimeout(jsg::Lock& js, kj::Maybe<jsg::JsValue> timeoutId) {
  KJ_IF_SOME(rawId, timeoutId) {
    KJ_IF_SOME(id, rawId.toSafeInteger(js)) {
      IoContext::current().clearTimeoutImpl(TimeoutId::fromNumber(id));
    }
  }
}
```

Or simpler yet... make `toSafeInteger` a static and allow it to accept `const kj::Maybe<jsg::JsValue>&` so we can just do...

```cpp
void ServiceWorkerGlobalScope::clearTimeout(jsg::Lock& js, kj::Maybe<jsg::JsValue> timeoutId) {
  KJ_IF_SOME(id, jsg::JsNumber::toSafeInteger(timeoutId)) {
    IoContext::current().clearTimeoutImpl(TimeoutId::fromNumber(id));
  }
}
```

src/workerd/jsg/jsvalue.c++
```diff
@@ -200,6 +200,34 @@ JsValue JsObject::getPrototype(Lock& js) {
   return JsValue(inner->GetPrototypeV2());
 }
 
+kj::Maybe<double> JsNumber::value(Lock& js) const {
+  KJ_ASSERT(!inner.IsEmpty());
+  double value;
+  if (inner->NumberValue(js.v8Context()).To(&value)) {
```
jasnell, line 206:
```suggestion
  // The NumberValue(...) operation can fail with a JS exception, in which case
  // we return kj::none and the error should be allowed to propagate.
  if (inner->NumberValue(js.v8Context()).To(&value)) {
```

src/workerd/jsg/jsvalue.h
```diff
@@ -325,13 +325,21 @@ class JsProxy final: public JsBase<v8::Proxy, JsProxy> {
...
 #undef V
 
+class JsNumber final: public JsBase<v8::Number, JsNumber> {
+ public:
+  kj::Maybe<double> value(Lock& js) const KJ_WARN_UNUSED_RESULT;
+  bool isSafeInteger(Lock& js) const KJ_WARN_UNUSED_RESULT;
+  kj::Maybe<double> toSafeInteger(Lock& js) const KJ_WARN_UNUSED_RESULT;
+
+  using JsBase<v8::Number, JsNumber>::JsBase;
+};
```
jasnell, line 341: The definition of this should be moved into jsvalue.h, just like the others like `JsBoolean`

anonrig: what if i told you, this file is jsvalue.h?

jasnell: Ha!! Wow... I totally misread the header name and could have sworn it said `jsg.h` ... sigh. I think that means I get to take the rest of the week off? Right? Right? ...... Right?

jasnell (approved):
LGTM as long as the type generation looks good with the switch to using `JsNumber` (I honestly can't remember if that was ever updated since I don't know if anything is actually using `JsNumber` yet)

PR #3909, opened 2025-04-08
Update workers-sdk tests in CI to use built types

emily-shen:
Previously, miniflare and co weren't being built with types built from the PR changes, but rather using published workers-types. This meant some errors weren't surfacing until these types were released and reached workers-sdk.

Also adds includes vitest-pool-workers tests and vite-plugin-tests which should hopefully be relatively quick but useful signals.

Fixes CFSQL-1245, DEVX-1778


.github/workflows/test.yml
```diff
@@ -111,10 +178,27 @@ jobs:
...
+          cat > /tmp/test-types/package.json << EOF
+          {
+              "name": "@cloudflare/workers-types",
+              "version": "4.20250405.0"
+            }
+          EOF
+          pnpm add /tmp/test-types -w
+
+      - name: Build Wrangler, Miniflare, Vitest, Vite & dependencies
+        run: pnpm build --filter miniflare --filter wrangler --filter @cloudflare/vitest-pool-workers --filter @cloudflare/vite-plugin
```
petebacondarwin, line 199: This step should not be necessary because Turbo should build the dependencies as part of the test:ci step below.

.github/workflows/test.yml
```diff
@@ -81,10 +81,77 @@ jobs:
       BAZEL_CACHE_KEY: ${{ secrets.BAZEL_CACHE_KEY }}
       WORKERS_MIRROR_URL: ${{ secrets.WORKERS_MIRROR_URL }}
 
+  check-snapshot:
```
petebacondarwin, line 84: There is some heavy duty scripting going on here.
Did this come from some other workflow or did you author it from scratch?
Looks generally OK to me but I haven't looked closely.

petebacondarwin (approved):
LGTM assuming the snapshot bits are good, which I didn't really check

.github/workflows/test.yml
penalosa: I believe this is copied from the type-snapshot.yml file, which has been deleted?

.github/workflows/test.yml
```diff
@@ -81,10 +81,77 @@ jobs:
...
+          only_update: true
+          message: |
+            The generated output of `@cloudflare/workers-types` matches the snapshot in `types/generated-snapshot` :tada:
+
   workers-sdk-test:
-    needs: test
+    needs: [test, check-snapshot]
     name: Run workers-sdk tests
-    runs-on: ubuntu-22.04-arm
+    runs-on: ubuntu-22.04
```
penalosa, line 154: Why this change? The arm runners seem to generally be faster?

.github/workflows/test.yml
emily-shen: yep i moved it from type-snapshot.yml, with no changes other than to upload the generated types as an artifact unconditionally.

.github/workflows/test.yml
emily-shen: ah that was an accident, changed back

.github/workflows/test.yml
```diff
@@ -81,8 +81,75 @@ jobs:
...
+          {
+            echo 'TYPES_DIFF<<EOF'
+            cat types.diff
+            echo EOF
+          } >> "$GITHUB_ENV"
+      - uses: actions/upload-artifact@v4
+        id: artifact-upload-step
+        with:
+          name: generated-snapshot
+          path: bazel-bin/types/definitions/
```
emily-shen, line 127: this should be the only thing thats changed - it used run on failure only.

needed to move this job into the same workflow so that the workers-sdk tests could depend on this completing and uploading the artifacts

.github/workflows/test.yml
```diff
@@ -111,10 +178,24 @@ jobs:
...
+          name: generated-snapshot
+          path: /tmp/test-types
+
+      - name: Create and install workers-types package
+        run: |
+          cp /tmp/test-types/oldest/* /tmp/test-types
+          cat > /tmp/test-types/package.json << EOF
+          {
+              "name": "@cloudflare/workers-types",
+              "version": "4.20250405.0"
```
danlapid, line 193: What's this hard coded version here?

PR #3911, opened 2025-04-09
EventTarget: Recheck handler map on each handler iteration

jclee:
Fixes potential use-after-free, if the underlying kj::HashMap grows enough during iteration to trigger reallocation.


dom96 (approved):
Nice find!

PR #3913, opened 2025-04-09
Add isolate memory accounting for jsg string types, Headers object

jasnell:
Add methods to jsg::Lock to create instances of jsg string specializations that include external memory accounting.

* jsg::ByteString
* jsg::DOMString
* jsg::USVString

The accounted for `jsg::ByteString`s are then used in the `Headers` object to ensure that all headers are accounted for towards the isolate limit.

While I was at it, removed a number of extraneous string allocations.


patches/v8/0022-Add-GetExternalMemory.patch
```diff
@@ -0,0 +1,49 @@
...
+Subject: Add methods to get heap and external memory sizes directly.
+
+`GetHeapStatistics()` exists for this, but also collects a lot of other info and apparently performs non-trivial computation. These new accessors are intended to be very fast.
+
+Signed-off-by: James M Snell <jsnell@cloudflare.com>
+
+diff --git a/include/v8-isolate.h b/include/v8-isolate.h
+index 3a261b4ad1aaa8f424a2a7e5d6dbad7b4ad8d7bf..3d3781ed2d60f12d8ec70ea6091f32b07e4ae9ad 100644
+--- a/include/v8-isolate.h
++++ b/include/v8-isolate.h
```
danlapid, line 13: Why not just use isolate->GetHeapStatistics()?

danlapid: Can we improve the documentation on the patch for the exact reason and differences for posterity?

jasnell: For those reading just the comments here tho... `GetHeapStatistics()` actually does not provide the correct result here so adding the `GetExternalMemory()` patch is required to get the most accurate view that matches what we use internally.

src/workerd/jsg/value.h
```diff
@@ -532,9 +536,10 @@ class StringWrapper {
       kj::Maybe<v8::Local<v8::Object>> parentObject) {
     v8::Local<v8::String> str = check(handle->ToString(context));
     v8::Isolate* isolate = context->GetIsolate();
+    auto& js = Lock::from(isolate);
     auto buf = kj::heapArray<char>(str->Utf8LengthV2(isolate) + 1);
     str->WriteUtf8V2(isolate, buf.begin(), buf.size(), v8::String::WriteFlags::kNullTerminate);
-    return DOMString(kj::mv(buf));
+    return js.accountedDOMString(kj::mv(buf));
```
npaun, line 542: I noticed that DOMString/USVString are now accounted whenever they're unwrapped, but it looks like there won't be accounting if kj::String is used. Is that intentional / is it worth leaving a comment about?

jasnell: Intentional for now. I'm going to look at the various uses of the `kj::String` type mapping next but this PR was already big and I wanted to focus primarily on the `Headers` case

npaun (approved):
I like the super clear explanations in `headers-test.c++`

src/workerd/api/global-scope.c++
```diff
@@ -207,14 +207,14 @@ kj::Promise<DeferredProxy<void>> ServiceWorkerGlobalScope::request(kj::HttpMetho
     // JavaScript headers object, ignoring the REQUEST guard that usually makes them immutable.
     KJ_IF_SOME(l, requestBody.tryGetLength()) {
       jsHeaders->setUnguarded(
-          jsg::ByteString(kj::str("Content-Length")), jsg::ByteString(kj::str(l)));
+          js, js.accountedByteString("Content-Length"_kj), js.accountedByteString(kj::str(l)));
```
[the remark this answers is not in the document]
jasnell: I prefer the accounted prefix because it is easier to reason about.. i.e. has this allocation been accounted for or not?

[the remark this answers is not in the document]
anonrig: Why not make the other way around? Like js.unaccountedByte... rather than js.accounted. We need to make it easier to type for the function we want to promote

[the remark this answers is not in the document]
jasnell: We can bikeshed names once we have the accounting in place ;-)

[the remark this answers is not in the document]
anonrig: Or we can do it now? :-)

PR #3917, opened 2025-04-09
Improve memory accounting for FormData

jasnell:
Similar to the Headers PR. Adds memory accounting for FormData.

### DIFF
--- a/samples/helloworld_esm/worker.js
+++ b/samples/helloworld_esm/worker.js
@@ -4,6 +4,23 @@
 
 export default {
   async fetch(req, env) {
+
+    const headers = new Headers();
+    headers.set('content-type', 'application/x-www-form-urlencoded');
+    const req2 = new Request('http://example.org', {
+      method: 'POST',
+      headers,
+      body: [
+        'field0=part0',
+        'field1=part1',
+        'field0=part2',
+        'field1=part3',
+      ].join('&')
+    });
+
+    const fd = await req2.formData();
+    for (const [k,v] of fd) console.log(k,v);
+
     return new Response("Hello World\n");
   }
 };
--- a/src/workerd/api/form-data-memory-test.c++
+++ b/src/workerd/api/form-data-memory-test.c++
@@ -1,0 +1,117 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "actor-state.h"
+#include "actor.h"
+#include "basics.h"
+#include "blob.h"
+#include "form-data.h"
+#include "http.h"
+#include "queue.h"
+#include "sockets.h"
+#include "sql.h"
+#include "streams.h"
+#include "web-socket.h"
+
+#include <workerd/io/promise-wrapper.h>
+#include <workerd/jsg/jsg-test.h>
+#include <workerd/jsg/jsg.h>
+
+namespace workerd::api {
+namespace {
+
+jsg::V8System v8System;
+
+static constexpr auto kBody = R"(--+
+Content-Disposition: form-data; name="field0"
+
+part0
+--+
+CONTENT-DISPOSITION: form-data;name="field1"
+
+part1
+--+
+content-disposition:form-data;name="field0"
+
+part2
+--+
+CoNTent-dIsposiTIOn: form-data; name="field1"
+
+part3
+--+--)"_kj;
+
+static constexpr auto kUrlData = "field0=part0&field1=part1&field0=part2&field1=part3"_kj;
+
+struct HeadersContext: public jsg::Object, public jsg::ContextGlobal {
+
+  bool test(jsg::Lock& js) {
+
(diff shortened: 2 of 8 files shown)


src/workerd/api/form-data.c++
```diff
@@ -348,8 +350,8 @@ void FormData::parse(jsg::Lock& js,
       data.reserve(query.size());
       for (auto& param: query) {
         data.add(Entry{
-          .name = kj::mv(param.name),
-          .value = kj::mv(param.value),
+          .name = js.accountedKjString(param.name),
```
npaun, line 353: Here comes my annoying question again.  Do we intend to make FormData use DOMStrings eventually, or is the concept of DOMString irrelevant to FormData?

samples/helloworld_esm/worker.js
```diff
@@ -4,6 +4,23 @@
 
 export default {
   async fetch(req, env) {
+
+    const headers = new Headers();
```
npaun, line 8: Should not have been committed.

jasnell: doh

npaun (approved):

src/workerd/api/form-data.c++
jasnell: Eventually maybe? I'd rather tackle that separately

samples/helloworld_esm/worker.js
```diff
@@ -4,6 +4,23 @@
...
+        'field0=part0',
+        'field1=part1',
+        'field0=part2',
+        'field1=part3',
+      ].join('&')
+    });
+
+    const fd = await req2.formData();
+    for (const [k,v] of fd) console.log(k,v);
+
```
jasnell, line 23:
```suggestion
```

src/workerd/api/form-data.c++
```diff
@@ -318,9 +318,10 @@ void FormData::parse(jsg::Lock& js,
               kj::Maybe<kj::StringPtr> maybeType, kj::ArrayPtr<const kj::byte> message) {
         KJ_IF_SOME(filename, maybeFilename) {
           if (convertFilesToStrings) {
+            auto messageData = kj::heapArray<char>(message.asChars());
             data.add(FormData::Entry{
-              .name = kj::str(name),
-              .value = kj::str(message.asChars()),
+              .name = js.accountedKjString(name),
+              .value = js.accountedKjString(kj::str(kj::mv(messageData))),
```
fhanau, line 324: Nit: I think the kj::str is redundant here, `accountedKjString(kj::Array<char>&& str)` can handle it directly right?

src/workerd/jsg/jsg.h
```diff
@@ -2365,6 +2365,15 @@ class Lock {
     return Ref<T>(kj::refcounted<T>(kj::fwd<Params>(params)...));
   }
 
+  // Returns a kj::String with an external memory adjustment attached.
+  kj::String accountedKjString(kj::Array<char>&& str);
+  kj::String accountedKjString(kj::String&& str) {
+    return accountedKjString(str.releaseArray());
+  }
+  kj::String accountedKjString(kj::StringPtr str) {
```
anonrig, line 2373: Before we go with more refactorings, I strongly think we should drop "accounted" from this. 

I feel like Sean Parker.

src/workerd/api/form-data-memory-test.c++
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  JSG_RESOURCE_TYPE(HeadersContext) {
+    JSG_METHOD(test);
+  }
+};
+
+JSG_DECLARE_ISOLATE_TYPE(HeadersIsolate,
+    HeadersContext,
+    // It's unfortunate but we have to pull in all of these sets of
```
fhanau, line 93: I agree that this is a really inconvenient issue – if we're willing to use the types directly instead of the collected macros, we can get away with a much smaller set though. This works for me locally.
```
JSG_DECLARE_ISOLATE_TYPE(HeadersIsolate,
    HeadersContext,
    // It's unfortunate but we have to pull in all of these sets of
    // types just for the test to build, even tho they aren't actually
    // used by or relevant to the test.
    EW_FORMDATA_ISOLATE_TYPES,
    workerd::api::Blob,
    workerd::api::Blob::Options,
    workerd::api::File,
    workerd::api::File::Options);
```
If you prefer the macros instead of this fine-grained approach, this can stay as-is. I do think this approach might be lower maintenance though as with the larger macro set there's more of a risk of needing to pull in more types later as there are more of them, so this would both reduce code size and maintenance effort here.

fhanau (approved):

src/workerd/api/form-data.c++
jasnell: Unfortunately no. The array is not null terminated. I'm going to look at it more later but passing it through kj::str was necessary to make it work.

src/workerd/jsg/jsg.h
jasnell: The naming of these is far less important than making these improvements. If we decide that a rename is necessary then it'll be super trivial to search for and find these all at once.

PR #3921, opened 2025-04-10
Update R2Bucket to use jsg::BufferSource

jasnell:
As we move closer to being able to enable the V8 Sandbox hopefully in
the not too distant future, our `kj::Array<kj::byte>` type wrapper impl
will end up either needing to be phased or changed due to the requirement
that all ArrayBuffer/TypedArrays will need to be allocated from within
the isolate rather than using external memory. This change will mean
any APIs that return a `kj::Array<kj::byte>` will cause the array to
be copied into a freshly allocated `ArrayBuffer` rather than wrapped.
Because the `Checksums` property accessors are already copying, if we
kept these methods as is we would end up allocating and copying twice
on every call. Let's no do such silly things. This PR modifies the
accessors to return jsg::BufferSource instances that are allocated within
the isolate, ensuring that we only copy these once when v8 sandbox is
enabled.

~~Most likely the type snapshots will need to be regenerated. Going to let the github bot decide that tho.~~ Done

### DIFF
--- a/types/generated-snapshot/2021-11-03/index.ts
+++ b/types/generated-snapshot/2021-11-03/index.ts
@@ -2014,11 +2014,11 @@ export interface R2PutOptions {
   onlyIf?: R2Conditional | Headers;
   httpMetadata?: R2HTTPMetadata | Headers;
   customMetadata?: Record<string, string>;
-  md5?: ArrayBuffer | string;
-  sha1?: ArrayBuffer | string;
-  sha256?: ArrayBuffer | string;
-  sha384?: ArrayBuffer | string;
-  sha512?: ArrayBuffer | string;
+  md5?: (ArrayBuffer | ArrayBufferView) | string;
+  sha1?: (ArrayBuffer | ArrayBufferView) | string;
+  sha256?: (ArrayBuffer | ArrayBufferView) | string;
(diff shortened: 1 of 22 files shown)


types/generated-snapshot/2021-11-03/index.ts
```diff
@@ -2014,11 +2014,11 @@ export interface R2PutOptions {
   onlyIf?: R2Conditional | Headers;
   httpMetadata?: R2HTTPMetadata | Headers;
   customMetadata?: Record<string, string>;
-  md5?: ArrayBuffer | string;
-  sha1?: ArrayBuffer | string;
-  sha256?: ArrayBuffer | string;
-  sha384?: ArrayBuffer | string;
-  sha512?: ArrayBuffer | string;
+  md5?: (ArrayBuffer | ArrayBufferView) | string;
```
anonrig, line 2017: Isn't this a breaking change?

jasnell: Nope. It already accepted both. The original type generation for kj::Array<kj::byte> is incorrect

fhanau (approved):

danlapid (approved):

PR #3922, opened 2025-04-10
Make AbortSignal cloneable

npaun:
Step right up folks, that's right, we've got abort signals that you can pass over RPC.

Now you can create an abort controller, then pass the signal to a function over RPC so you can tell it to abort later.

```js

export class VeryUseful extends WorkerEntrypoint {
  async countToInfinity(signal) {
    for (let i = 0; ; i++) {
      if (signal.aborted) {
        return i;
    }
  }

const promise = env.VeryUseful.countToInfinity(ac.signal);
while (Math.random() <= .9)
  ;

ac.abort();
console.log("We got to", await promise);
```

Please let me know if you can think of any more tests this PR is missing.

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -522,6 +522,8 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
   return dispatchEventImpl(js, kj::mv(event));
 }
 
+// A wrapper for the AbortTrigger jsrpc client, that automatically sends a release() message once
+// the client is destroyed, informing the server that an abort will not be triggered in the future.
 class AbortTriggerRpcClient final {
  public:
   AbortTriggerRpcClient(rpc::AbortTrigger::Client&& client): client(kj::mv(client)) {}
@@ -531,6 +533,20 @@ class AbortTriggerRpcClient final {
     auto field = req.initReason();
     field.setV8Serialized(reason);
     return req.send().ignoreResult();
+  }
+
+  ~AbortTriggerRpcClient() noexcept(false) {
+    auto req = client.releaseRequest(capnp::MessageSize{4, 0});
+    // We call detach() on the resulting promise so that we can perform RPC in a destructor
+    req.send().ignoreResult().detach([](kj::Exception exc) {
+      if (exc.getType() == kj::Exception::Type::DISCONNECTED) {
+        // It's possible we can't send the release message because we're already disconnected.
+        return;
+      };
+
+      // Other exceptions could be more interesting
+      LOG_EXCEPTION("abortTriggerReleaseRpc", exc);
+    });
   }
 
  private:
@@ -555,9 +571,31 @@ class AbortTriggerRpcServer final: public rpc::AbortTrigger::Server {
     return kj::READY_NOW;
   }
 
+  kj::Promise<void> release(ReleaseContext releaseCtx) override {
+    released = true;
+    return kj::READY_NOW;
+  }
+
+  ~AbortTriggerRpcServer() noexcept(false) {
+    if (pendingReason->getWrapped() != nullptr) {
+      // Already triggered
+      return;
+    }
+
+    if (!released) {
+      pendingReason->getWrapped() = JSG_KJ_EXCEPTION(FAILED, DOMAbortError,
+          "An AbortSignal received over RPC was implicitly aborted because the connection back to "
+          "its trigger was lost.");
+    }
+
+    // Always fulfill the promise in case the AbortSignal was waiting
+    fulfiller->fulfill();
+  }
+
  private:
   kj::Own<kj::PromiseFulfiller<void>> fulfiller;
   kj::Own<AbortSignal::PendingReason> pendingReason;
+  bool released = false;
 };
 }  // namespace
 
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -852,7 +852,7 @@ jsg::Ref<AbortSignal> AbortSignal::deserialize(
     jsg::Lock& js, rpc::SerializationTag tag, jsg::Deserializer& deserializer) {
   auto& handler = KJ_REQUIRE_NONNULL(
       deserializer.getExternalHandler(), "got AbortSignal on non-RPC serialized object?");
-  auto externalHandler = dynamic_cast<RpcDeserializerExternalHander*>(&handler);
+  auto externalHandler = dynamic_cast<RpcDeserializerExternalHandler*>(&handler);
   KJ_REQUIRE(externalHandler != nullptr, "got AbortSignal on non-RPC serialized object?");
 
   auto isCanceled = static_cast<bool>(deserializer.readRawUint32());
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -9,6 +9,7 @@
 
 #include <workerd/io/io-context.h>
 
+#include <capnp/message.h>
 #include <kj/async.h>
 #include <kj/vector.h>
 
@@ -265,7 +266,7 @@ void EventTarget::addEventListener(jsg::Lock& js,
       KJ_IF_SOME(signal, maybeSignal) {
         // If the AbortSignal has already been triggered, then we need to stop here.
         // Return without adding the event listener.
-        if (signal->getAborted()) {
+        if (signal->getAborted(js)) {
           return;
         }
       }
@@ -521,6 +522,45 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
   return dispatchEventImpl(js, kj::mv(event));
 }
 
+class AbortTriggerRpcClient final {
+ public:
+  AbortTriggerRpcClient(rpc::AbortTrigger::Client&& client): client(kj::mv(client)) {}
+
+  kj::Promise<void> abort(kj::ArrayPtr<kj::byte> reason) {
+    auto req = client.abortRequest(capnp::MessageSize{reason.size() / sizeof(capnp::word) + 8, 0});
+    auto field = req.initReason();
+    field.setV8Serialized(reason);
+    return req.send().ignoreResult();
+  }
+
+ private:
+  rpc::AbortTrigger::Client client;
+};
+
+namespace {
+// The jsrpc handler that receives aborts from the remote and triggers them locally
+class AbortTriggerRpcServer final: public rpc::AbortTrigger::Server {
+ public:
+  AbortTriggerRpcServer(kj::Own<kj::PromiseFulfiller<void>> fulfiller,
+      kj::Own<AbortSignal::PendingReason>&& pendingReason)
+      : fulfiller(kj::mv(fulfiller)),
+        pendingReason(kj::mv(pendingReason)) {}
+
+  kj::Promise<void> abort(AbortContext abortCtx) override {
+    auto params = abortCtx.getParams();
+    auto reason = params.getReason().getV8Serialized();
+
+    pendingReason->getWrapped() = kj::heapArray(reason.asBytes());
+    fulfiller->fulfill();
+    return kj::READY_NOW;
+  }
+
+ private:
+  kj::Own<kj::PromiseFulfiller<void>> fulfiller;
+  kj::Own<AbortSignal::PendingReason> pendingReason;
+};
+}  // namespace
+
 AbortSignal::AbortSignal(kj::Maybe<kj::Exception> exception,
     jsg::Optional<jsg::JsRef<jsg::JsValue>> maybeReason,
     Flag flag)
@@ -540,16 +580,34 @@ void AbortSignal::setOnAbort(jsg::Lock& js, jsg::Optional<jsg::JsValue> handler)
   KJ_IF_SOME(h, handler) {
     if (h.isFunction() || h.isObject()) {
       onAbortHandler = jsg::JsRef(js, h);
+      subscribeToRpcAbort(js);
       return;
     }
   }
   onAbortHandler = kj::none;
 }
 
+void AbortSignal::addEventListener(jsg::Lock& js,
+    kj::String type,
+    jsg::Identified<Handler> handler,
+    jsg::Optional<AddEventListenerOpts> maybeOptions) {
+  EventTarget::addEventListener(js, kj::mv(type), kj::mv(handler), kj::mv(maybeOptions));
+  subscribeToRpcAbort(js);
+}
+
+bool AbortSignal::getAborted(jsg::Lock& js) {
+  return canceler->isCanceled() || hasPendingReason();
+}
+
 jsg::JsValue AbortSignal::getReason(jsg::Lock& js) {
   KJ_IF_SOME(r, reason) {
     return r.getHandle(js);
   }
+
+  KJ_IF_SOME(r, deserializePendingReason(js)) {
+    return r;
+  }
+
   return js.undefined();
 }
 
@@ -582,8 +640,12 @@ void AbortSignal::throwIfAborted(jsg::Lock& js) {
     KJ_IF_SOME(r, reason) {
       js.throwException(r.getHandle(js));
     } else {
-      js.throwException(abortException(js));
+      js.throwException(abortException(js, kj::none));
     }
+  }
+
+  KJ_IF_SOME(r, deserializePendingReason(js)) {
+    js.throwException(r);
   }
 }
 
@@ -619,7 +681,7 @@ jsg::Ref<AbortSignal> AbortSignal::any(jsg::Lock& js,
   // Let's check to see if any of the signals are already aborted. If it is, we can
   // optimize here by skipping the event handler registration.
   for (auto& sig: signals) {
-    if (sig->getAborted()) {
+    if (sig->getAborted(js)) {
       return AbortSignal::abort(js, sig->getReason(js));
     }
   }
@@ -681,9 +743,24 @@ void AbortSignal::triggerAbort(
       }
     }
   } else {
-    reason = js.exceptionToJsValue(kj::mv(exception));
+    reason = js.exceptionToJsValue(kj::cp(exception));
   }
-  canceler->cancel(kj::cp(exception));
+
+  canceler->cancel(kj::mv(exception));
+
+  // 1. Dispatch to RPC clients
+  if (!rpcClients.empty()) {
+    IoContext& ioContext = IoContext::current();
+    jsg::Serializer ser(js);
+    KJ_IF_SOME(r, reason) {
+      ser.write(js, r.getHandle(js));
+    }
+
+    auto released = ser.release();
+    ioContext.addTask(sendToRpc(kj::mv(released.data)));
+  }
+
+  // 2. Dispatch to local listeners
 
   // This is questionable only because it goes against the spec but it does help prevent
   // memory leaks. Once the abort signal has been triggered, there's really nothing else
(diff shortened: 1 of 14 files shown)


src/workerd/api/tests/abortsignal-test.js
```diff
@@ -1,8 +1,52 @@
-import { strictEqual, ok, throws } from 'node:assert';
+import { strictEqual, ok, throws, strict } from 'node:assert';
+import { WorkerEntrypoint } from 'cloudflare:workers';
+import { scheduler } from 'node:timers/promises';
```
jasnell, line 3: Let's add a test that tries caching a local abort signal in the global scope in one request then tries triggering it in a different request. It should fail with an error and shouldn't crash anything.

```
let ac;

export default {
  fetch(req, env) {
    if (ac === undefined) {
      ac = new AbortController();
      env.FOO.rpcCall(ac.signal);  // send the signal over
    } else {
      ac.abort(new Error('boom?'));
    }
  }
}
```

The `ac.abort()` should fail with an error complaining about cross request I/O. Let's just double check that nothing weird happens with the rpc clients.

src/workerd/api/basics.c++
```diff
@@ -666,6 +709,97 @@ void AbortSignal::triggerAbort(
...
+
+  auto streamCap =
+      externalHandler->writeStream([&](rpc::JsValue::External::Builder builder) mutable {
+    auto field = builder.initAbortSignal();
+    field.setSignal(kj::heap<AbortSignalImpl>(ioContext, JSG_THIS));
+  });
+
+  // Keep track of every AbortSignal cloned from this one.
+  // If this->triggerAbort(...) is called, each rpcClient will be informed.
+  rpcClients.add(kj::mv(streamCap).castAs<rpc::AbortSignal>());
```
jasnell, line 737: One thing that my `MessageChannel` PR did not cover yet was removing `rpcClients` from this collection when the capability is dropped remotely. You might need to make sure this doesn't become a memory leak or problematic.

npaun:
It seems like this PR exploded RPC serialization for other types, like RpcPromise, but I didn't change them at all.

src/workerd/api/basics.c++
npaun: What method do I need to implement to get notified about this situation?

jasnell: heh, I hadn't quite worked that out yet for the MessageChannel PR. Still fairly new to this mechanism myself. @kentonv likely can say.

src/workerd/io/worker-interface.capnp
```diff
@@ -460,6 +462,10 @@ struct JsValue {
         }
       }
 
+      abortSignal :group {
```
kentonv, line 465: You don't need a group if it just wraps one field.

src/workerd/io/worker-interface.capnp
```diff
@@ -482,6 +488,16 @@ struct JsValue {
   }
 }
 
+interface AbortSignal $Cxx.allowCancellation {
+    struct Params {
```
kentonv, line 492: Nitpick: Indentation is off here.

src/workerd/io/worker-interface.capnp
```diff
@@ -482,6 +488,16 @@ struct JsValue {
...
+      reason @0 :JsValue;
+    }
+
+    struct Results {}
+
+    triggerAbort @0 Params -> Results;
```
kentonv, line 498: Why are `Params` and `Results` being factored out as separate structs here? You can just write:

```suggestion
    triggerAbort @0 (reason :JsValue) -> ();
```

This is exactly equivalent -- in fact, it's binary-compatible, so you could change it to structs later if desired.

src/workerd/api/basics.c++
```diff
@@ -669,6 +712,97 @@ void AbortSignal::triggerAbort(
...
+  auto& handler = KJ_REQUIRE_NONNULL(
+      deserializer.getExternalHandler(), "got AbortSignal on non-RPC serialized object?");
+  auto externalHandler = dynamic_cast<RpcDeserializerExternalHander*>(&handler);
+  KJ_REQUIRE(externalHandler != nullptr, "got AbortSignal on non-RPC serialized object?");
+
+  auto isCanceled = static_cast<bool>(deserializer.readRawUint32());
+  auto flag = static_cast<Flag>(deserializer.readRawUint32());
+  auto reason = deserializer.readValue(js);
+
+  auto reader = externalHandler->read();
```
kentonv, line 754: The `reader` contains an `AbortSignal` capability but it appears you don't ever use it here? You just drop it?

src/workerd/api/basics.c++
```diff
@@ -495,6 +496,44 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
...
+        });
+      });
+    });
+
+    triggerAbortCtx.initResults();
+  }
+
+ private:
+  jsg::Ref<AbortSignal> signal;
+  kj::Own<IoContext::WeakRef> weakIoContext;
```
kentonv, line 533: Instead of using `weakIoContext` and `ctx.run()`, I'd highly recommend that you create a promise-fulfiller pair. `triggerAbort` rejects the fulfiller. Elsewhere, use `ctx.awaitIo()` to wait on the promise side and deliver the signal on receipt. This is important because otherwise the IoContext doesn't know that it's waiting on anything and may prematurely terminate or hibernate.

However! I think it's important that we do not do this `awaitIo()` unless something is actually listening on the AbortSignal. If no one actually registers to listen on the "abort" event, then we do _not_ want the IoContext to think that it's waiting for something, because that could block hibernation inappropriately.

Also, I think using a promise-fulfiller pair fixes a cyclic reference problem that exists currently: the two AbortSignals in the two isolates end holding strong references to each other over RPC, and so neither one will ever be able to be GC'd, and they will presumably hold the RPC session open forever. A promise-fulfiller is essentially a weak reference.

src/workerd/api/basics.c++
```diff
@@ -495,6 +496,44 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
   return dispatchEventImpl(js, kj::mv(event));
 }
 
+namespace {
+// The jsrpc handler that receives aborts from the remote and triggers them locally
+class AbortSignalImpl final: public rpc::AbortSignal::Server {
```
kentonv, line 501: A name like `AbortSignalRpcImpl` would make it clearer how this class differs from the `AbortSignal` API type.

src/workerd/api/basics.c++
```diff
@@ -669,6 +712,97 @@ void AbortSignal::triggerAbort(
...
+  jsg::Serializer ser(js);
+  KJ_IF_SOME(r, reason) {
+    ser.write(js, r.getHandle(js));
+  }
+
+  auto released = ser.release();
+  capnp::MallocMessageBuilder builder;
+  rpc::JsValue::Builder val = builder.initRoot<rpc::JsValue>();
+  val.setV8Serialized(kj::mv(released.data));
+  return sendToRpc(capnp::clone(val.asReader()));
```
kentonv, line 786: This `capnp::clone` seems redundant as `sendToRpc()` does not need the reader to outlive the call. You could just pass the `JsValue::Reader` itself in without cloning.

But even better yet, you could pass in the serialized data (`released.data`) instead of a whole Reader here. Then you could set it directly on each outgoing call. Then you don't need to create a `MallocMessageBuilder` at all. Two copies (and a bunch of allocation) avoided!

src/workerd/api/basics.c++
```diff
@@ -669,6 +712,97 @@ void AbortSignal::triggerAbort(
...
+}
+
+kj::Promise<void> AbortSignal::sendToRpc(kj::Own<rpc::JsValue::Reader> reason) {
```
kentonv, line 789: In seems like the one and only caller of this is the other `sendToRpc` overload? Why split them at all? Why not have just one method?

kentonv (commented):
The RPC interface is missing doc comments. Please consider this a general rule: ALWAYS explain RPC interfaces in full directly in the capnp file. It should never be necessary to look at the code to determine how to use the RPC interface.

Reverse-engineering from the code, it appears the `rpc::AbortSignal` interface is used in both directions: as a regular capability, and as a stream. This is a b it confusing and needs documentation explaining and justifying the design.

src/workerd/io/worker-interface.capnp
npaun: Is there a capnp autoformatter? I would love to put once in CI if it exists.

src/workerd/api/basics.c++
npaun: I did this split to avoid being hit by `kj::IsDisallowedInCoroutine`.

src/workerd/io/worker-interface.capnp
kentonv: Sorry, there isn't.

src/workerd/api/basics.c++
npaun: Safe to say I have no idea what I'm doing. Please let me know if there's a better way to encode the AbortSignal interface in terms of RPC.

npaun: AbortSignals only go one way, there's no need for the cloned signal to send anything back to its parent.

kentonv: Actually hold on, back up.

I was assuming the idea was that, once an `AbortSignal` has been sent over the wire, then either end can listen on the signal _and_ either end can _trigger_ the signal, so both ends are listening on each other.

But actually does that make sense? An `AbortSignal` JS object lets you _receive_ the signal. It doesn't let you trigger it -- that's what `AbortController` is for.

So you actually _only_ want to use a stream slot here (pointing caller -> callee)? You don't need any callee -> caller capability at all?

npaun: Correct

src/workerd/io/worker-interface.capnp
```diff
@@ -460,6 +462,10 @@ struct JsValue {
         }
       }
 
+      abortSignal :group {
+        signal @7 :AbortSignal;
```
kentonv, line 466: (Per my previous comment...)

This is sending an `rpc::AbortSignal` capability to the callee, which can be used to send an abort _back_ to the caller. But I think that's actually not what you want? The JavaScript `AbortSignal` type only lets you receive signals, not send them.

You seem to have correctly used StreamSink to go the other direction.

kentonv: Perhaps calling the RPC interface `AbortSignal` is creating confusion here, since the RPC interface actually more closely resembles `AbortController`?

npaun: Yeah I think I will rename to AbortController

src/workerd/api/basics.c++
```diff
@@ -669,6 +712,97 @@ void AbortSignal::triggerAbort(
...
+  } else {
+    serializer.write(js, js.undefined());
+  }
+
+  auto& ioContext = IoContext::current();
+
+  auto streamCap =
+      externalHandler->writeStream([&](rpc::JsValue::External::Builder builder) mutable {
+    auto field = builder.initAbortSignal();
+    field.setSignal(kj::heap<AbortSignalImpl>(ioContext, JSG_THIS));
```
npaun, line 735: There's one part I still have a mental block on - how do I change the code to avoid passing the capability the wrong way around?

npaun: @kentonv Do I just get rid of these two lines?

kentonv: Change `JsValue.External.abortSignal` to have type `Void`. Then you do `builder.setAbortSignal()` here to set it, but it's just empty, because there's nothing you need to send here.

npaun:
AbortSignal currently fails Asan checks with

```
*** Fatal uncaught kj::Exception: kj/async.c++:142: failed: KJ async object being destroyed when not allowed: JavaScript heap objects must not contain KJ I/O objects without a IoOwn
```

Do you know at a glance which field is not being handled correctly?

kentonv:
> Do you know at a glance which field is not being handled correctly?

The stack trace should contain some hints, but my guess is this:

`kj::Vector<rpc::AbortSignal::Client> rpcClients;`

A capnp `Client` is a KJ I/O object, you can't hold it inside a JSG object except with an `IoOwn` wrapper.

npaun:
Was my guess too. The stack trace is unhelpful on my box for some reason.

src/workerd/api/basics.c++
danlapid: Sounds to me like you need a membrane

npaun: I've seen that word before. Do you remember a good file for me to look at?

kentonv: I actually don't think you need to do anything here.

These are outgoing capabilities, that is, this object is holding pointers to remote objects. They don't get "dropped remotely", because this end is the end that can be "dropped". Arguably you could try to detect if the connection has been lost, but I don't think there's any need to do that. If the capabilities are disconnected, then calling them will fail, but you can just ignore that. I don't think someone is going to send the _same_ AbortSignal over a huge number of connections in order to cause a memory leak.

src/workerd/io/worker-interface.capnp
```diff
@@ -482,6 +488,14 @@ struct JsValue {
   }
 }
 
+interface AbortSignal $Cxx.allowCancellation {
```
kentonv, line 491: It occurs to me that we may want to think carefully about what happens if the AbortSignal is dropped without calling `triggerAbort()`.

The intuitive answer, from a JavaScript perspective, is that nothing should happen. The signal never aborts.

But the standard answer from a Cap'n Proto perspective is that we should assume an abort. This is because one reason the capability might be dropped is due to a network disconnect. In this case, the client has lost the ability to communicate aborts to us, but very likely, if they've lost the connection, they'd prefer to cancel the operation. "Fail closed" rather than "fail open".

To this end we may actually want this RPC interface to have two methods: `abort()` and `release()`. Calling `release()` means "the AbortController was GC'd on the client side, indicating the client never intends to abort it, so please do NOT assume an abort when this capability is dropped, which it will be shortly". Only if we didn't receive a `release()`, we would assume an abort when dropped.

src/workerd/api/basics.c++
```diff
@@ -669,6 +706,107 @@ void AbortSignal::triggerAbort(
...
+    auto req = cap->triggerAbortRequest();
+    auto field = req.initReason();
+    field.setV8Serialized(reason);
+    promises.add(req.send().ignoreResult());
+  }
+
+  co_await kj::joinPromises(promises.releaseAsArray());
+}
+
+void AbortSignal::awaitRpcAbort(jsg::Lock& js) {
```
npaun, line 798: > However! I think it's important that we do not do this awaitIo() unless something is actually listening on the AbortSignal. If no one actually registers to listen on the "abort" event, then we do not want the IoContext to think that it's waiting for something, because that could block hibernation inappropriately.

Here's my first crack at this problem. `awaitRpcAbort` is called when an event listener is added.  On the first listener added, I call `awaitIo` to set up a handler if a triggerAbort is received over RPC.

However this doesn't deal with `removeEventListener`. Nor something like add, remove, add. Do we care? Does there need to be some way to cancel this `awaitIo`?

npaun: I'm also wondering if that behaviour is right. Even without registering an event handler, code can synchronously check properties like `signal.aborted`, `signal.reason` or use `singal.throwIfAborted()`

src/workerd/api/tests/abortsignal-test.js
```diff
@@ -188,3 +251,257 @@ export const onabortPrototypeProperty = {
...
+      {
+        name: 'Error',
+        message:
+          "Cannot perform I/O on behalf of a different request. I/O objects (such as streams, request/response bodies, and others) created in the context of one request handler cannot be accessed from a different request's handler. This is a limitation of Cloudflare Workers which allows us to improve overall performance. (I/O type: RefcountedCanceler)",
+      }
+    );
+  },
+};
+
+export const rpcRemoteCanIgnoreSignal = {
```
npaun, line 489: @danlapid Re: our discussion this morning. I added a test showing that just passing an AbortSignal over RPC, then aborting it, won't cause anything to happen by itself on the other end, unless some kind of handler is registered.

Maybe this will help a bit towards convincing ourselves that re-enabling Request.signal won't break existing usage.

npaun:
- rebase -

npaun:
Ready for another look. I switched to promise-and-fulfiller, and I tried implementing release. They don't quite work right but hopefully I can fix them.

jasnell:
There are enough subtle nuances with the jsrpc interface here that I'm not quite sure I'm comfortable just yet reviewing this. I'll have to leave it to @kentonv for the primary review.

npaun:
Have to push another rebase...

src/workerd/io/worker-interface.capnp
```diff
@@ -482,6 +488,18 @@ struct JsValue {
...
```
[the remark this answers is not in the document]
npaun: I think `AbortTrigger` is the best name, because the semantics don't match the JS AbortController interface either. An AbortController has a signal as a member.

src/workerd/api/basics.h
```diff
@@ -518,11 +519,18 @@ class EventTarget: public jsg::Object {
 // An implementation of the Web Platform Standard AbortSignal API
 class AbortSignal final: public EventTarget {
  public:
+  using RpcReasonT = kj::OneOf<kj::Array<kj::byte>, /* V8 serialized JS value */
```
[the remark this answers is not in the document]
npaun: Fine I'll stop putting my PIN number into the ATM machine.

src/workerd/api/basics.c++
```diff
@@ -692,6 +808,133 @@ void AbortSignal::triggerAbort(
...
+  if (*pendingReason == nullptr) {
+    // Pending reason not initialized; abort not yet triggered
+    return;
+  }
+
+  KJ_SWITCH_ONEOF(*pendingReason) {
+    KJ_CASE_ONEOF(v8Serialized, kj::Array<kj::byte>) {
+      jsg::Deserializer des(js, v8Serialized);
+      auto reasonJs = des.readValue(js);
+      canceler->cancel(kj::cp(abortException(js, reasonJs)));
```
npaun, line 916: What should we do in cases where there are no registered event listeners, the synchronous  JS methods of AbortSignal aren't used, but something is using the canceller?

src/workerd/api/tests/abortsignal-test.js
```diff
@@ -212,3 +276,273 @@ export const onabortPrototypeProperty = {
...
+    ac.signal.skipReleaseForTest();
+    ac = null;
+    gc();
+    await scheduler.wait(10);
+
+    const res = await resPromise;
+    // This isn't what I'd expect - I can't seem to force the dropped capability logic to run when I want it
+    strictEqual(res.reason.message, 'The operation was aborted due to timeout');
+  },
+};
```
npaun, line 548: I'm not having a good time trying to simulate the dropped capability case. This doesn't seem to actually GC the signal.

src/workerd/api/basics.h
```diff
@@ -524,13 +527,15 @@ class AbortSignal final: public EventTarget {
       jsg::Optional<jsg::JsRef<jsg::JsValue>> maybeReason = kj::none,
       Flag flag = Flag::NONE);
 
+  class PendingReason: public kj::OneOf<kj::Array<kj::byte> /* v8Serialized*/,
+                           kj::Exception> /* if capability is dropped*/,
+                       public kj::Refcounted {};
```
npaun, line 532: Just confirming that this is a safe way for the AbortTrigger rpc server and the AbortSignal object to share data.

src/workerd/api/basics.c++
```diff
@@ -521,13 +522,93 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
...
+          __LINE__, kj::str(JSG_ERROR_DOMAbortError ": Dropped capability, assuming abort")));
+    }
+
+    // Always fulfill the promise in case the AbortSignal was waiting
+    fulfiller->fulfill();
+  }
+
+ private:
+  kj::Own<kj::PromiseFulfiller<void>> fulfiller;
+  kj::Own<AbortSignal::PendingReason> pendingReason;
```
jasnell, line 599: Can these be combined into a single `kj::Own<kj::PromiseFulfiller<kj::Own<AbortSignal::PendingReason>>>`? Then you can attach a continuation on the `signal->maybeRpcAbortPromise` that triggers the abort signal.

npaun: Is there a way to do a non-blocking check for the reason being set, if someone calls AbortSignal.reason or AbortSignal.throeUfAborted() etc?

src/workerd/api/basics.c++
```diff
@@ -681,10 +781,26 @@ void AbortSignal::triggerAbort(
...
 
+  dispatchEvent(js);
+}
+
+void AbortSignal::dispatchEvent(jsg::Lock& js) {
+  // 1. Dispatch to RPC clients
+  IoContext& ioContext = IoContext::current();
+  jsg::Serializer ser(js);
+  KJ_IF_SOME(r, reason) {
+    ser.write(js, r.getHandle(js));
```
kentonv, line 796: Can we skip this serialization if there are no RPC clients?

src/workerd/api/basics.h
kentonv: I don't think we should use inheritance this way.

There's actually a utility class that does exactly what you need here: `kj::RefcountedWrapper<OneOf<...>>`

src/workerd/api/basics.h
```diff
@@ -609,15 +620,53 @@ class AbortSignal final: public EventTarget {
...
+  // Trigger an abort on all associated clients
+  kj::Promise<void> sendToRpc(kj::Array<kj::byte>&& reason);
+
+  /** RPC server functionality. Used if this signal was deserialized. **/
+
+  // A promise that is fulfilled if an abort() message is received over RPC.
+  kj::Maybe<kj::Promise<void>> maybeRpcAbortPromise;
+
+  // A refcounted object used to receive a serialized abort reason
+  kj::Own<PendingReason> pendingReason;
```
kentonv, line 662: You will need this to be `IoOwn` since `PendingReason` may be modified on the RPC's I/O thread without holding the isolate lock.

src/workerd/api/basics.c++
kentonv: I think we also need to register the wait as soon as anything registers itself with the canceler. So if anyone calls `wrap()` or `getCanceler()`, we need to start waiting.

And if we do that, then `checkForRpcAbort()` can get simpler -- it doesn't need to call cancel->cancel() at all, since it can assume that that will happen separately if anyone is waiting.

Relatedly, it's weird that calling `getAbort()` has side effects. I think we can get rid of that. `getAbort()` should, if `reason` is empty, check `pendingReason` directly, and if it's filled in, use it, _without_ modifying `reason`, nor triggering the canceler, nor scheduling any events.

I think this also means you can get rid of the new `dispatchEvent()` method and put its logic back into `triggerAbort`. In `subscribeToRpcAbort()`, you should be invoking `triggerAbort` directly with the abort reason.

kentonv:
NOTE: GitHub UI hid 4 review comments behind this thingy:

Screenshot from 2025-04-21 15-50-22

Make sure to click on it and expand them!

(This may sound obvious, but I've frequently had people miss "hidden" comments in the past.)

src/workerd/api/basics.c++
```diff
@@ -521,13 +522,93 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
...
+    });
+  }
+
+  bool skipReleaseForTest = false;
+
+ private:
+  rpc::AbortTrigger::Client client;
+};
+
+namespace {
```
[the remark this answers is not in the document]
npaun: I think I cannot do this because `AbortSignal` needs to have a member variable (`rpcClients`) that holds `AbortTriggerRpcClient`s, so I have to forward declare `AbortTriggerRpcClient`.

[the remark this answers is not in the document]
kentonv: Ah, missed that. (Note that you can forward-declare things in anonymous namespaces, but you can't use anonymous namespaces in a header file, so yeah, that wouldn't work.)

Another option, though, would be to make it a nested class in AbortSignal.

But I guess I don't really care.

src/workerd/api/basics.c++
```diff
@@ -681,10 +787,25 @@ void AbortSignal::triggerAbort(
       }
     }
   } else {
-    reason = js.exceptionToJsValue(kj::mv(exception));
+    reason = js.exceptionToJsValue(kj::cp(exception));
   }
+
   canceler->cancel(kj::cp(exception));
```
npaun, line 793: If I'm reading this right, the original code was trying to use a moved-from value at this point

kentonv: That does appear to be the case, yes. Which in practice probably meant its description appeared empty, if the exception ended up being reported anywhere...

src/workerd/api/basics.c++
```diff
@@ -521,13 +522,96 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
...
+    auto req = client.releaseRequest(capnp::MessageSize{4, 0});
+    // We call detach() on the resulting promise so that we can perform RPC in a destructor
+    req.send().ignoreResult().detach([](kj::Exception exc) {
+      if (exc.getType() == kj::Exception::Type::DISCONNECTED) {
+        // It's possible we can't send the release message because we're already disconnected.
+        return;
+      };
+
+      // Other exceptions could be more interesting
+      LOG_EXCEPTION("~AbortTriggerRpcClient", exc);
```
kentonv, line 552: I think Sentry might dislike `~` appearing here, so...

```suggestion
      LOG_EXCEPTION("abortTriggerReleaseRpc", exc);
```

src/workerd/api/basics.c++
kentonv: That said, since this line _is_ the last use of `exception`, you could use `kj::mv` on this line.

src/workerd/api/basics.h
```diff
@@ -609,15 +619,54 @@ class AbortSignal final: public EventTarget {
...
+  // Trigger an abort on all associated clients
+  kj::Promise<void> sendToRpc(kj::Array<kj::byte>&& reason);
+
+  /** RPC server functionality. Used if this signal was deserialized. **/
+
+  // A promise that is fulfilled if an abort() message is received over RPC.
+  kj::Maybe<kj::Promise<void>> maybeRpcAbortPromise;
+
+  // A refcounted object used to receive a serialized abort reason
+  IoOwn<PendingReason> pendingReason;
```
kentonv, line 661: I think you should make this `Maybe<IoOwn<PendingReason>>`, and not set it for AbortSignals that didn't originate from RPC.

Otherwise I think if anyone today is using `getAborted` or `getReason` on an AbortSignal created by a different I/O context, they would be broken by this change. (People probably shouldn't be doing that anyway, but I could imagine someone out there is...)

src/workerd/api/basics.c++
```diff
@@ -696,6 +817,132 @@ void AbortSignal::triggerAbort(
...
+    KJ_CASE_ONEOF(exception, kj::Exception) {
+      return kj::some(js.exceptionToJsValue(kj::cp(exception)).getHandle(js));
+    }
+  }
+
+  KJ_UNREACHABLE;
+}
+
+void AbortSignal::subscribeToRpcAbort(jsg::Lock& js) {
+  KJ_IF_SOME(abortPromise, maybeRpcAbortPromise) {
```
kentonv, line 935: Maybe worth a comment like:

> For an AbortSignal received over RPC, the first time someone registers an event on the signal, we want to arrange to awaitIo() for the underlying RPC signal. If no one is actually listening, though, we don't want to awaitIo() since it blocks hibernation in actors.

src/workerd/api/basics.c++
```diff
@@ -521,13 +522,96 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
...
+
+  ~AbortTriggerRpcServer() noexcept(false) {
+    if (pendingReason->getWrapped() != nullptr) {
+      // Already triggered
+      return;
+    }
+
+    if (!released) {
+      pendingReason->getWrapped() = kj::Exception(kj::Exception::Type::FAILED, __FILE__, __LINE__,
+          kj::str(JSG_ERROR_DOMAbortError ": Dropped capability, assuming abort"));
```
kentonv, line 593: This message probably isn't understandable to most Workers developers. Perhaps:

> An AbortSignal received over RPC was implicitly aborted because the connection back to its trigger was lost.

src/workerd/api/basics.c++
kentonv: You can't synchronously check a promise, no. I agree these probably have to stay separate like this. Maybe add a comment explaining the motivation, though.

src/workerd/api/basics.c++
```diff
@@ -521,6 +522,88 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
...
+
+  ~AbortTriggerRpcServer() noexcept(false) {
+    if (pendingReason->getWrapped() != nullptr) {
+      // Already triggered
+      return;
+    }
+
+    if (!released) {
+      pendingReason->getWrapped() = JSG_KJ_EXCEPTION(FAILED, DOMAbortError,
+          "An AbortSignal received over RPC was implicitly aborted because the connection back to its trigger was lost.");
```
kentonv, line 593: Need to wrap at 100 columns.

```suggestion
          "An AbortSignal received over RPC was implicitly aborted because the connection back to "
          "its trigger was lost.");
```

Disappointing that clang-format isn't doing this automatically.

src/workerd/api/basics.h
```diff
@@ -609,15 +619,56 @@ class AbortSignal final: public EventTarget {
...
   kj::Maybe<jsg::JsRef<jsg::JsValue>> onAbortHandler;
 
+  static kj::Exception abortException(
+      jsg::Lock& js, jsg::Optional<kj::OneOf<kj::Exception, jsg::JsValue>> reason);
+
   void visitForGc(jsg::GcVisitor& visitor);
 
   friend class AbortController;
+
+  /** RPC client functionality. Used if this signal was serialized. **/
```
kentonv, line 646: OK this is really nitpicky but we don't use `/*` block comments. "Section header" comments are usually styled like this:

```suggestion
  // -----------------------------------------------------------------
  // RPC client functionality. Used if this signal was serialized.
```

src/workerd/api/basics.h
```diff
@@ -609,15 +619,56 @@ class AbortSignal final: public EventTarget {
...
+
+  // A collection of rpcClients, which will be notified if this signal is triggered and when this
+  // signal is destroyed.
+  kj::Vector<IoOwn<AbortTriggerRpcClient>> rpcClients;
+
+  // Trigger an abort on all associated clients
+  kj::Promise<void> sendToRpc(kj::Array<kj::byte>&& reason);
+
+  /** RPC server functionality. Used if this signal was deserialized. **/
```
kentonv, line 655:
```suggestion
  // -----------------------------------------------------------------
  // RPC server functionality. Used if this signal was deserialized.

```

kentonv (approved):

src/workerd/api/basics.h
```diff
@@ -578,28 +588,28 @@ class AbortSignal final: public EventTarget {
     }
     JSG_PROTOTYPE_PROPERTY(onabort, getOnAbort, setOnAbort);
     JSG_METHOD(throwIfAborted);
+    JSG_METHOD(skipReleaseForTest);
```
kentonv, line 591: Oh, this still needs to be guarded behind "experimental".

npaun: I'm trying to clean up this test more deeply. I'll make sure any test only methods are protected before I finish.

npaun:
@kentonv I think this is now good to go.

Since last time:
- Fixed test failures (missing an IoOwn on rpcAbortPromise)
- Properly guarded skipReleaseForTest
- Made sure the tests for disconnection work reliably

npaun:
^^^ just pushed a rebase because Internal build failures look spurious

npaun:
Pushed again because I accidentally undid some work with a bad force push

npaun:
rebase; ew build bungled again

kentonv:
Are you rebasing as a way to make the tests re-run due to flakes? You don't need to do that, just re-run the tests, it's much faster.

kentonv (approved):

PR #3923, opened 2025-04-10
Re-merge Request.signal PR

npaun:
This PR re-enables the support for `Request.signal` on incoming requests that I originally added in #3488 and #3727. These changes were originally reviewed and approved in those PRs.

However, these changes caused a regression in production and had to be reverted in #3735. AbortSignals were never cloneable, and trying to clone any Request with the `signal` property set would fail. But with my PR, the `signal` property was now always set in a very common case, the request received by a worker in its fetch handler.

AbortSignal will now be cloneable once #3922 is merged. The second commit to this PR adds a regression test which now fails but will work once #3922 is merged.

### DIFF
--- a/src/workerd/api/basics.h
+++ b/src/workerd/api/basics.h
@@ -521,7 +521,7 @@ class AbortTriggerRpcClient;
 
 class AbortSignal final: public EventTarget {
  public:
-  enum class Flag { NONE, NEVER_ABORTS };
+  enum class Flag { NONE, NEVER_ABORTS, IGNORE_FOR_SUBREQUESTS };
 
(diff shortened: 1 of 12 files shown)


src/workerd/api/basics.h
```diff
@@ -603,9 +603,14 @@ class AbortSignal final: public EventTarget {
     tracker.trackField("reason", reason);
   }
 
+  bool isIgnoredForSubrequests() const {
```
anonrig, line 606: Would you mind writing a documentation for this?

anonrig (approved):

PR #3926, opened 2025-04-11
Python: Remove version info from devPyodide compat flag

hoodmane:
Instead, pass around a class with a pair of the SnapshotRelease::Reader and an `isDev` boolean. This should make it easier to run development tests against multiple versions of Python in local development.

### DIFF
--- a/src/workerd/io/compatibility-date.h
+++ b/src/workerd/io/compatibility-date.h
@@ -51,9 +51,33 @@ kj::Maybe<kj::String> normalizeCompatDate(kj::StringPtr date);
 // Returns the current date as a string formatted by CompatDate.
 kj::String currentDateStr();
 
-kj::Maybe<PythonSnapshotRelease::Reader> getPythonSnapshotRelease(
-    CompatibilityFlags::Reader featureFlags);
-kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease);
+class PythonRelease {
+ public:
+  PythonSnapshotRelease::Reader capnpRelease;
+  bool isDevelopment;
+  PythonRelease(PythonSnapshotRelease::Reader capnpRelease, bool isDevelopment)
+      : capnpRelease(capnpRelease),
+        isDevelopment(isDevelopment) {}
+
+  kj::StringPtr getPyodide() {
+    return capnpRelease.getPyodide();
+  }
+  kj::StringPtr getPyodideRevision() {
+    return capnpRelease.getPyodideRevision();
+  }
+  kj::StringPtr getPackages() {
+    return capnpRelease.getPackages();
+  }
+  kj::StringPtr getBaselineSnapshotHash() {
+    return capnpRelease.getBaselineSnapshotHash();
+  };
+  int64_t getBackport() {
+    return capnpRelease.getBackport();
+  }
+};
+
+kj::Maybe<PythonRelease> getPythonSnapshotRelease(CompatibilityFlags::Reader featureFlags);
+kj::String getPythonBundleName(PythonRelease& pyodideRelease);
 
 // These values come from src/workerd/io/compatibility-date.capnp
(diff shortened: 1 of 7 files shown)


danlapid (approved):
LGTM with that one comment.
This will likely require an internal PR.

src/workerd/io/compatibility-date.h
```diff
@@ -51,9 +51,33 @@ kj::Maybe<kj::String> normalizeCompatDate(kj::StringPtr date);
...
+  }
+  kj::StringPtr getBaselineSnapshotHash() {
+    return capnpRelease.getBaselineSnapshotHash();
+  };
+  int64_t getBackport() {
+    return capnpRelease.getBackport();
+  }
+};
+
+kj::Maybe<PythonRelease> getPythonSnapshotRelease(CompatibilityFlags::Reader featureFlags);
```
danlapid, line 79: There's a key semantic difference between the old version of this function and the new.
The old had a reference-counting based copy constructor, the new does not.
That forced you to replace many instances of `PythonSnapshotRelease::Reader` with `PythonRelease&`.
In most places this is actually unsafe.
Seeing as PythonRelease is relatively small it might be appropriate to just implement a copy constructor for it and keep passing it by value as we did before and not by reference.

hoodmane: I don't think anyone stores it ever?

hoodmane: But it does make sense to implement copy and not worry about.

src/workerd/io/compatibility-date.h
```diff
@@ -51,9 +51,33 @@ kj::Maybe<kj::String> normalizeCompatDate(kj::StringPtr date);
...
+  }
+  kj::StringPtr getPackages() {
+    return capnpRelease.getPackages();
+  }
+  kj::StringPtr getBaselineSnapshotHash() {
+    return capnpRelease.getBaselineSnapshotHash();
+  };
+  int64_t getBackport() {
+    return capnpRelease.getBackport();
+  }
```
dom96, line 76: Instead of doing this, why not just have a struct with two fields: `capnpRelease` and `isDevelopment`? I think that would also solve the UaF concerns, though `PythonSnapshotRelease::Reader` is a pointer too so whatever it's pointing to needs to be kept alive.

dom96 (changes requested):
Hmm, maybe I'm missing something, but why don't we just continue to use `pyodideRelease.getPyodide() == "dev"` as a check on whether a release is "dev"? In that case you can also set the other `pythonSnapshotRelease` fields to `dev` for consistency and it should be just as easy to implement logic to decide the version used for "dev" in local dev. All without having to introduce a wrapper class around PythonSnapshotRelease.

src/workerd/io/compatibility-date.h
dom96: But I don't think we have a need for this class. See my comment above.

hoodmane: I don't think that would change the UaF concerns in any way?

hoodmane:
> it should be just as easy to implement logic to decide the version used for "dev" in local dev.

Well currently it uses the `packages` and `baselineSnapshotHash` fields from the dev flag:
```
`packages = "20240829.4"`
          baselineSnapshotHash = "92859211804cd350f9e14010afad86e584bdd017dc7acfd94709a87f3220afae"
```
If I switch `dev` to being based on 0.27.5 I have to change these fields in `compatibility-date.capnp` which I don't like. But perhaps I should switch to generating all of the `PythonSnapshotRelease` data from bazel and not just for the development compat date.

hoodmane:
>  In that case you can also set the other pythonSnapshotRelease fields to dev for consistency and it should be just as easy to implement logic to decide the version used for "dev" in local dev.

Well the point of these changes is that the logic for those is "just pull them off a different compat flag". Which compat flag it should be pulled off also has to be generated somehow.

hoodmane:
After discussion with @dom96, closing this in favor of another approach.

PR #3927, opened 2025-04-11
Migrate to capnp-es

penalosa:
Continuation of https://github.com/cloudflare/workers-sdk/pull/7788, fixes https://jira.cfdata.org/browse/DEVX-1599. This migrates the type generation system in `workerd` to use `capnp-es`. Notably, this should have no changes to the actually generated types, which should give us confidence that this is working as expected.


patches/capnp-es.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/dist/compiler/main.mjs b/dist/compiler/main.mjs
+index 6999c0e0c50570efd51959661ffc611f561bd31b..c37f97d5d786e4fc316acd4f5780e8e3976b21dd 100755
+--- a/dist/compiler/main.mjs
++++ b/dist/compiler/main.mjs
+@@ -38,7 +38,7 @@ Options:
+ `;
+ const sources = [];
+ const options = [];
+-let outFormats = ["js"];
++let outFormats = ["ts"];
```
anonrig, line 10: Would you mind creating/linking to an upstream issue? Is it possible to get rid of this patch eventually?

types/src/generator/index.ts
```diff
@@ -44,69 +44,93 @@ function collectIncluded(map: StructureMap, root?: string): Set<string> {
...
-      case Type_Which.ONE_OF:
-        { type.getOneOf().getVariants().forEach(visitType); return; }
-      case Type_Which.FUNCTION:
-        { visitFunction(type.getFunction()); return; }
+        const dict = type.dict;
+        visitType(dict.key);
+        {
+          visitType(dict.value);
+          return;
+        }
```
anonrig, line 74:
```suggestion
        visitType(dict.value);
        return;
```

anonrig (approved):

patches/capnp-es.patch
penalosa: Yeah, I'm going to put up a PR today

fhanau (approved):

anonrig (approved):

WORKSPACE
```diff
@@ -18,7 +18,7 @@ deps_gen()
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
 
-NODE_VERSION = "22.11.0"
+NODE_VERSION = "22.14.0"
```
penalosa, line 21: This was needed to get `require()` working with `capnp-es` (which is ESM only) in our TS tests. I used `just update-deps` for Node.js, which I think is the correct approach (cc @anonrig ?)

WORKSPACE
```diff
@@ -212,13 +212,6 @@ load("@aspect_rules_js//npm:repositories.bzl", "npm_translate_lock")
 npm_translate_lock(
     name = "npm",
     npmrc = "//:.npmrc",
-    patch_args = {
-        "capnp-ts@0.7.0": ["-p1"],
-    },
-    # Patches required for `capnp-ts` to type-check
-    patches = {
-        "capnp-ts@0.7.0": ["//:patches/capnp-ts@0.7.0.patch"],
-    },
```
penalosa, line 221: It seems like `npm_translate_lock` can instead load from a `package.json` list of patched deps

WORKSPACE
anonrig: It's fine to ignore them for now. I'll take a look at once this lands. No need to block you.

anonrig:
We need to update node and types/node on EW as well

penalosa:
@anonrig what's the process for that?

anonrig:
> @anonrig what's the process for that?

Opening a branch on EW with the same branch name as this one, updating workerd git hash, and updating node and types/node. Once that build passes you can rerun the internal-build github action which will pick up your branches changes rather than an automated branch. Once this Pr lands you can land the EW change.

edmundhung (approved):

PR #3928, opened 2025-04-11
Temporarily disable string memory accounting

jasnell:
Impl of the external memory accounting is being looked at again, temporarily disable use.


npaun (approved):

danlapid (approved):

kentonv:
#3930 should make this OK again.

kentonv:
I wonder, though, if it should only apply to strings of non-negligible size? Like if a string is only a few bytes long, then this is wasting more memory than it is tracking.

PR #3929, opened 2025-04-11
Roll deps

npaun:


fhanau (approved):
All looks good to me.
Additional possible updates for "extra credit" (i.e. optional but nice to have):
- Update Rust -> 1.86.0 (now available with rules_rust 0.60.0)
- It seems unlikely that Rust crates didn't change, except for the new cargo_bazel version adding the rustc_env_files thing. Can you try to update them again?
- Since we're updating ruff, we should also update ruff_python_ast and ruff_python_parser in deps/rust/cargo.bzl to the latest tag (v0.11.5) so they are in sync. This should deduplicate some transitive dependencies.

.bazelrc
```diff
@@ -53,6 +53,9 @@ build --per_file_copt=external/com_google_tcmalloc@-DANNOTATE_MEMORY_IS_INITIALI
 build --per_file_copt=external/com_google_protobuf@-Wno-deprecated-this-capture
 build --host_per_file_copt=external/com_google_protobuf@-Wno-deprecated-this-capture
 
+## simdutf
+build --per_file_copt=external/simdutf@-Wno-unused-const-variable
```
anonrig, line 57: Do we need this? I though we didn't build with WError on workerd?

anonrig (approved):

.bazelrc
npaun: We do build with `-Werror`, but only in CI.

npaun:
@fhanau Did the extra credit, can I get a sticker now?

fhanau:
> @fhanau Did the extra credit, can I get a sticker now?

All looks good! 💯

PR #3930, opened 2025-04-11
Fix and clean up issues with external memory tracking

kentonv:
See commits for details.

### DIFF
--- a/src/workerd/jsg/jsg-test.c++
+++ b/src/workerd/jsg/jsg-test.c++
@@ -515,7 +515,7 @@ KJ_TEST("External memory adjustment - defered") {
 
   // Making an adjustment anyway won't do anything but also won't crash
   auto adjuster3 = target->getAdjustment(500);
-  KJ_ASSERT(target->getPendingMemoryUpdate() == 0);
+  KJ_ASSERT(target->getPendingMemoryUpdate() == 400);
 }
 
 }  // namespace
--- a/src/workerd/jsg/jsg-test.c++
+++ b/src/workerd/jsg/jsg-test.c++
@@ -480,27 +480,27 @@ KJ_TEST("External memory adjustment - defered") {
     // Adjustment to memory while not holding lock will be applied later
     auto adjuster1 = target->getAdjustment(1000);
     KJ_ASSERT(adjuster1.getAmount() == 1000);
-    KJ_ASSERT(target->getPendingMemoryUpdate() == 1000);
+    KJ_ASSERT(target->getPendingMemoryUpdateForTest() == 1000);
 
     {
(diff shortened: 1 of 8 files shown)


jasnell (approved):

src/workerd/jsg/jsg-test.c++
```diff
@@ -480,42 +480,42 @@ KJ_TEST("External memory adjustment - defered") {
...
-  KJ_ASSERT(!target->isIsolateAlive());
+  KJ_ASSERT(!target->isIsolateAliveForTest());
 
   // Delete the long-lived array, which will call the adjustment's destructor, to make sure it's safe.
   mem = nullptr;
 
   // Making an adjustment anyway won't do anything but also won't crash
   auto adjuster3 = target->getAdjustment(500);
-  KJ_ASSERT(target->getPendingMemoryUpdate() == 0);
+  KJ_ASSERT(target->getPendingMemoryUpdateForTest() == 400);
```
npaun, line 518: Why is it 400 at this point?

kentonv: Previously it was zero only because the count was no longer accessible after the isolate was destroyed. But I moved the counter to be inside the ExternalMemoryTarget itself, so now the counter is still there. Destroying `mem`, above, applied a -100 adjustment, and then `getAdjustment(500)` added 500, for a total of 400.

src/workerd/jsg/jsg.c++
```diff
@@ -418,7 +410,7 @@ void ExternalMemoryAdjustment::maybeDeferAdjustment(ssize_t amount) {
 }
 
 ExternalMemoryAdjustment::ExternalMemoryAdjustment(
-    kj::Own<ExternalMemoryTarget> externalMemory, size_t amount)
+    kj::Own<const ExternalMemoryTarget> externalMemory, size_t amount)
```
npaun, line 413:
```suggestion
    kj::Own<const ExternalMemoryTarget>&& externalMemory, size_t amount)
```

I guess?

npaun (approved):
This is really nice. Thank you for having a look at this stuff.

src/workerd/jsg/jsg.c++
kentonv: These are mostly functionally equivalent, but the `&&` adds noise, and sometimes can be confusing if the called function doesn't actually move away the value. Most of the time I think it's best not to declare parameters as rvalue references.

mrwhalebones:
You guys are awesome. I'm going to be working on a fully python based node with included utilities and a selection of theme mode. Thanks for all the inspiration from true coders for the new coders. Thanks for working out the bugs!!!

PR #3931, opened 2025-04-11
Implements Python RPC + test suite.

dom96:
Relatively large PR which introduces Workers RPC support for Python. The aim is to make it easy to call functions defined in other Python and JS Workers, to do this all arguments are converted to/from JS where appropriate.

The implementation relies on a number of wrappers as well as conversion functions:

* The binding `env` object is wrapped via the `handler` decorator, it gets wrapped in a class that intercepts calls to `Fetcher` bindings (i.e. `env.ServiceBinding`) and then another class that intercepts calls made on that binding. All arguments passed to these calls are wrapped in a call to `serialize_to_js` and the returned value from the call is wrapped in a call to `serialize_to_python`.
* The `serialize_to_python` and `serialize_to_js` functions convert Python and JS arguments to JS and Python respectively. These mostly use Pyodide's `to_js` and `to_py` but with some improvements and restrictions.
* Arguments received by WorkerEntrypoints also get converted. This is done in the code that manages the entrypoint classes and already intercepts calls.

All of this is supported by a test suite which attempts to send/receive both from and to Python, as well as from and to JS. 

### Test Plan
```
$ bazel run @workerd//src/workerd/server/tests/python:python-rpc_development@
```

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -1,6 +1,7 @@
 # This module defines a Workers API for Python. It is similar to the API provided by
 # JS Workers, but with changes and additions to be more idiomatic to the Python
 # programming language.
+import datetime
 import functools
 import http.client
 import json
@@ -8,6 +9,8 @@ from collections.abc import Generator, Iterable, MutableMapping
 from contextlib import ExitStack, contextmanager
 from enum import StrEnum
 from http import HTTPMethod, HTTPStatus
+from inspect import isawaitable
+from types import LambdaType
 from typing import Any, TypedDict, Unpack
 
 import js
@@ -99,7 +102,7 @@ class RequestInitCfProperties(TypedDict, total=False):
 class FetchKwargs(TypedDict, total=False):
     headers: "Headers | None"
     body: "Body | None"
-    method: HTTPMethod = HTTPMethod.GET
+    method: HTTPMethod | None
     redirect: str | None
     cf: RequestInitCfProperties | None
 
@@ -184,23 +187,36 @@ def _manage_pyproxies():
         destroy_proxies(proxies)
 
 
+def _is_js_instance(val, js_cls_name):
+    return hasattr(val, "constructor") and val.constructor.name == js_cls_name
+
+
 def _to_js_headers(headers: Headers):
     if isinstance(headers, list):
         # We should have a list[tuple[str, str]]
         return js.Headers.new(headers)
     elif isinstance(headers, dict):
         return js.Headers.new(headers.items())
-    elif hasattr(headers, "constructor") and headers.constructor.name == "Headers":
+    elif _is_js_instance(headers, "Headers"):
         return headers
     else:
         raise TypeError("Received unexpected type for headers argument")
 
 
 class Response(FetchResponse):
+    """
+    This class represents the response to an HTTP request, with a similar API to that of the web
+    `Response` API: https://developer.mozilla.org/en-US/docs/Web/API/Response.
+
+    This is a static class to enable defining `FetchResponse` using a constructor more similar to
+    JS. All non-static methods for this class should be defined on the `FetchResponse` class it
+    extends.
+    """
+
     def __init__(
         self,
-        body: Body,
-        status: HTTPStatus | int = HTTPStatus.OK,
+        body: "Body | js.Response | None" = None,
+        status: HTTPStatus | int | None = None,
         status_text="",
         headers: Headers = None,
     ):
@@ -220,12 +236,22 @@ class Response(FetchResponse):
                 "FormData",
                 "ReadableStream",
                 "URLSearchParams",
+                "Response",
             ):
                 raise TypeError(
                     f"Unsupported type in Response: {body.constructor.name}"
                 )
         elif not isinstance(body, str | FormData):
             raise TypeError(f"Unsupported type in Response: {type(body).__name__}")
+
+        # Handle constructing a Response from a JS Response.
+        if _is_js_instance(body, "Response"):
+            if status is not None or len(status_text) > 0 or headers is not None:
+                raise ValueError(
+                    "Expected no options when constructing Response from a js.Response"
+                )
+            super().__init__(body.url, body)
+            return
 
         options = self._create_options(status, status_text, headers)
 
@@ -238,13 +264,15 @@ class Response(FetchResponse):
 
     @staticmethod
     def _create_options(
-        status: HTTPStatus | int = HTTPStatus.OK,
+        status: HTTPStatus | int | None = HTTPStatus.OK,
         status_text="",
         headers: Headers = None,
     ):
-        options = {
-            "status": status.value if isinstance(status, HTTPStatus) else status,
-        }
+        options = {}
+        if status:
+            options["status"] = (
+                status.value if isinstance(status, HTTPStatus) else status
+            )
         if status_text:
             options["statusText"] = status_text
         if headers:
@@ -313,7 +341,10 @@ def _js_value_to_py(item: FormDataValue) -> "str | Blob | File":
 
 class FormData(MutableMapping[str, FormDataValue]):
     """
-    This API follows that of https://pypi.org/project/multidict/.
+    This class represents a set of key/value pairs for forms.
+
+    The API of this class follows that of https://pypi.org/project/multidict/ and
+    https://developer.mozilla.org/en-US/docs/Web/API/FormData.
     """
 
     def __init__(
@@ -329,10 +360,7 @@ class FormData(MutableMapping[str, FormDataValue]):
                 self._js_form_data.append(k, _py_value_to_js(v))
             return
 
-        if (
-            hasattr(form_data, "constructor")
-            and form_data.constructor.name == "FormData"
-        ):
+        if _is_js_instance(form_data, "FormData"):
             self._js_form_data = form_data
             return
 
@@ -416,6 +444,8 @@ def _make_blob_entry(e):
 
 
 def _is_iterable(obj):
+    if isinstance(obj, (str, bytes)):
+        return False
     try:
         iter(obj)
     except TypeError:
@@ -548,7 +578,17 @@ class File(Blob):
 
 
 class Request:
-    def __init__(self, input: "Request | str", **other_options: Unpack[FetchKwargs]):
+    def __init__(
+        self, input: "Request | str | js.Request", **other_options: Unpack[FetchKwargs]
+    ):
+        if _is_js_instance(input, "Request"):
+            if len(other_options) > 0:
+                raise ValueError(
+                    "Expected no options when constructing Request from a js.Request"
+                )
+            self._js_request = input
+            return
+
         if "method" in other_options and isinstance(
             other_options["method"], HTTPMethod
         ):
@@ -682,6 +722,156 @@ class Request:
         return await self.js_object.text()
 
 
+def _python_from_rpc_default_converter(value, convert, cache):
+    if not hasattr(value, "constructor"):
+        # Assume that the object doesn't need conversion as it's not a JS object.
+        return value
+
+    if value.constructor.name == "Response":
+        return Response(value)
+    elif value.constructor.name == "FormData":
+        return FormData(value)
+    elif value.constructor.name == "Blob":
+        return Blob(value)
+    elif value.constructor.name == "File":
+        return File(value)
+    elif value.constructor.name == "Request":
+        return Request(value)
+    elif value.constructor.name == "Date":
+        # TODO: Pyodide should gain support for this, we should upstream this.
+        return datetime.datetime.fromtimestamp(value.getTime() / 1000)
+    elif value.constructor.name == "Error":
+        return Exception(value.toString())
+    elif value.constructor.name == "Number":
+        return value.valueOf()
+
+    raise TypeError(
+        f"Couldn't convert object to Python type, got {value.constructor.name}"
+    )
+
+
+def python_from_rpc(obj: "JsProxy"):
+    """
+    Converts JS objects like Response, Request, Blob, etc. to equivalent Python objects defined in
+    this module and also other JS objects like Map, Set, etc. to equivalent Python stdlib objects.
+
+    This method is used for Workers RPC in Python to convert JavaScript objects to Python. As such
+    it does not support serializing all JS object types.
+    """
+
+    if not hasattr(obj, "constructor"):
+        return obj
+
+    result = obj.to_py(default_converter=_python_from_rpc_default_converter)
+
+    return result
+
+
+def _raise_on_disabled_type(value):
+    if _is_js_instance(value, "RegExp"):
+        raise TypeError(f"{value.constructor.name} cannot be sent over RPC.")
+
+    if isinstance(value, (tuple, bytearray, LambdaType)):
+        raise TypeError(f"{type(value)} cannot be sent over RPC.")
+
+    if _is_iterable(value):
+        if isinstance(value, dict):
+            for v in value.values():
+                _raise_on_disabled_type(v)
+        else:
+            for v in value:
+                _raise_on_disabled_type(v)
+
+
+def _python_to_rpc_default_converter(obj, convert, cache):
+    if obj is None:
+        return obj
+
+    if hasattr(obj, "js_object"):
+        return obj.js_object
+
+    if isinstance(obj, datetime.datetime):
+        # TODO: Pyodide should gain support for this, we should upstream this.
+        return js.Date.new(obj.timestamp() * 1000)
+
+    if isinstance(obj, Exception):
+        return js.Error.new(str(obj))
+
+    _raise_on_disabled_type(obj)
+
+    return obj
+
+
+def python_to_rpc(value) -> JsProxy:
+    """
+    Converts Python objects defined in this module (Response, Request, etc) and native Python types
+    like Map, Set, datetime to equivalent JavaScript types.
+
+    This method is used for Workers RPC in Python to convert Python objects to JavaScript. As such
+    it does not support serializing all Python object types.
+    """
+
+    # `to_js` won't always call the default_converter, for example when a list of tuples is passed
+    _raise_on_disabled_type(value)
+
+    result = to_js(
+        value,
+        default_converter=_python_to_rpc_default_converter,
+        dict_converter=js.Map.new,
+    )
+
+    return result
+
+
+class _FetcherWrapper:
+    def __init__(self, binding):
+        self._binding = binding
+
+    def _getattr_helper(self, name):
+        attr = getattr(self._binding, name)
+        if name == "fetch":
+            # TODO: Implement python native fetch for bindings.
+            return attr
+
+        if not callable(attr):
+            return attr
+
+        # Not using `@functools.wraps(attr)` here because `attr` is a JS proxy.
+        async def wrapper(*args, **kwargs):
+            js_args = [python_to_rpc(arg) for arg in args]
+            js_kwargs = {k: python_to_rpc(v) for k, v in kwargs.items()}
+            result = attr(*js_args, **js_kwargs)
+            if hasattr(result, "then") and callable(result.then):
+                return python_from_rpc(await result)
+            else:
+                return python_from_rpc(result)
+
+        return wrapper
+
+    def __getattr__(self, name):
+        result = self._getattr_helper(name)
+        setattr(self, name, result)
+        return result
+
+
+class _EnvWrapper:
+    def __init__(self, env):
+        self._env = env
+
+    def _getattr_helper(self, name):
+        binding = getattr(self._env, name)
+        if _is_js_instance(binding, "Fetcher"):
+            return _FetcherWrapper(binding)
+
+        # TODO: Implement APIs for bindings.
+        return binding
+
+    def __getattr__(self, name):
+        result = self._getattr_helper(name)
+        setattr(self, name, result)
+        return result
+
+
 def handler(func):
     """
     When applied to handlers such as `on_fetch` it will rewrite arguments passed in to native Python
@@ -691,16 +881,41 @@ def handler(func):
 
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
-        # TODO: wrap `env` so that bindings can be used without to_js.
-        if (
-            len(args) > 0
-            and hasattr(args[0], "constructor")
-            and args[0].constructor.name == "Request"
-        ):
+        # TODO: support transforming kwargs
+        if len(args) > 0 and _is_js_instance(args[0], "Request"):
             args = (Request(args[0]),) + args[1:]
+
+        # Wrap `env` so that bindings can be used without to_js.
+        if len(args) > 1:
+            args = (args[0], _EnvWrapper(args[1])) + args[2:]
+
         return func(*args, **kwargs)
 
     return wrapper
+
+
+def _wrap_attr(attr):
+    if not callable(attr):
+        return attr
+
+    @functools.wraps(attr)
+    async def wrapper(*args, **kwargs):
+        py_args = [python_from_rpc(arg) for arg in args]
+        py_kwargs = {k: python_from_rpc(v) for k, v in kwargs.items()}
+        result = attr(*py_args, **py_kwargs)
+        if isawaitable(result):
+            return python_to_rpc(await result)
+        else:
+            return python_to_rpc(result)
+
+    return wrapper
+
+
+def _wrap_subclass(cls):
+    for k, v in cls.__dict__.items():
+        if k.startswith("__"):
+            continue
+        setattr(cls, k, _wrap_attr(v))
 
 
 class DurableObject:
@@ -711,6 +926,9 @@ class DurableObject:
     def __init__(*_args, **_kwds):
         pass
 
+    def __init_subclass__(cls, **_kwargs):
+        _wrap_subclass(cls)
+
 
 class WorkerEntrypoint:
     """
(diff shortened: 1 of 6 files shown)


src/pyodide/internal/workers.py
```diff
@@ -1,6 +1,7 @@
 # This module defines a Workers API for Python. It is similar to the API provided by
 # JS Workers, but with changes and additions to be more idiomatic to the Python
 # programming language.
+import datetime
 import functools
 import http.client
 import json
```
hoodmane, line 7: Not 100% connected to this PR but we should consider some way to prevent people from doing `from workers import js` or `from workers import datetime`. There are two possible approaches:
1. `import datetime as _datetime`
2. move the implementation to `_workers.py` and then import just the public things into `workers.py`.

I favor option 2.

src/pyodide/internal/workers.py
```diff
@@ -607,6 +640,60 @@ async def text(self) -> str:
         return await self.js_object.text()
 
 
+def to_python(obj: "JsProxy", throw=True):
```
hoodmane, line 643: Maybe let's just call these `to_js` and `to_py`?

src/pyodide/internal/workers.py
```diff
@@ -607,6 +640,60 @@ async def text(self) -> str:
...
+    elif obj.constructor.name == "FormData":
+        return FormData(obj)
+    elif obj.constructor.name == "Blob":
+        return Blob(obj)
+    elif obj.constructor.name == "File":
+        return File(obj)
+    elif obj.constructor.name == "Request":
+        return Request(obj)
+    elif obj.constructor.name == "Date":
+        return datetime.datetime.fromtimestamp(obj.getTime() / 1000)
```
hoodmane, line 667: We'd want this stuff to apply also to for instance a list of responses. `to_py` and `to_js` have a `default_converter` argument for that. So I'd suggest putting all of this logic in a separate `default_converter` function and setting it to this by default. Then we'd also apply this logic to `[[[response, blob]]]`.

src/pyodide/internal/workers.py
dom96: I think we'll need to give them different names since they are a bit specialised for the serialisation. For example we don't allow converting tuples.

src/pyodide/internal/workers.py
```diff
@@ -607,6 +640,89 @@ async def text(self) -> str:
...
+
+    if hasattr(obj, "js_object"):
+        return obj.js_object
+
+    if isinstance(obj, datetime.datetime):
+        # XXX: Pyodide should gain support for this, we should upstream this.
+        return js.Date.new(obj.timestamp() * 1000)
+
+    if isinstance(obj, (tuple, bytearray)):
+        raise TypeError(f"DataCloneError: {type(obj)} cannot be sent over RPC.")
```
hoodmane, line 692: This only works for the outermost layer unfortunately. It will still convert `[(1,2)]` to an array of arrays.

hoodmane: Nope. I probably should add a way to adjust the conversion of stuff before we apply the default logic to it.

danlapid:
I'll review tomorrow
In the meantime, mind adding a description to the PR of your implementation approach?
Thanks!

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -210,18 +210,29 @@ function makeEntrypointClass(className: string, classKind: AnyClass): any {
...
+            // Check if the requested method exists.
             const pyInstance = await target.pyInstance;
             if (typeof pyInstance[prop] === 'function') {
-              const res = await pyInstance[prop](...args);
-              if (isKnownHandler) {
-                return res?.js_object ?? res;
-              }
-              return res;
+              // Convert any JS arguments to Python SDK objects.
+              const pyodide = await getPyodide();
```
danlapid, line 224: Not sure I understand this, we said we're going to convert the parameters using decorators, why are we converting in the entrypoint-helper?

src/workerd/server/tests/python/python-rpc/worker.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    const null_resp = await env.PythonRpc.identity(null);
+    assert.deepStrictEqual(null_resp, undefined);
+
+    // Web/API Types
+    const py_response = await env.PythonRpc.handle_response(
+      new Response('this is a response')
+    );
+    assert.deepStrictEqual(await py_response.text(), 'this is a response');
```
danlapid, line 52: This should be a js Response object but the test doesn't seem to make this clear

src/workerd/server/tests/python/python-rpc/worker.js
```diff
@@ -0,0 +1,59 @@
...
+
+    const py_response2 = await env.PythonRpc.handle_request(
+      new Request('https://test.com', { method: 'POST' })
+    );
+    assert.deepStrictEqual(py_response2.method, 'POST');
```
danlapid, line 57: same as above.

src/pyodide/python-entrypoint-helper.ts
dom96: We convert using decorator for top-level handlers, but we do it implicitly for entrypoint classes. I guess you could argue that this is inconsistent, I just really dislike having to write `@handler` everywhere, doing so for every RPC method would be annoying. Also these RPC methods aren't technically "handlers".

src/workerd/server/tests/python/python-rpc/worker.js
dom96: I can add some checks to make it clear, but it definitely will be a JS response, there is no other way this can work (you cannot send a PyProxy through the RPC)

src/pyodide/internal/workers.py
dom96: Will do in separate PR.

src/pyodide/python-entrypoint-helper.ts
danlapid: This has the same problem that drove the push for the decorator
Users can’t modify this logic in their local workers-py module and use that instead
Its pre-baked into the runtime
I really strongly feel like we should minimize what the runtime bakes in without any ability for user overrides

danlapid: We can have the decorator for entry points be a class decorator or a meta class so that you wouldn't need to decorate every method

dom96: Okay, if that's the concern then I think I see a way to do this in workers.py without the need for a decorator.

danlapid: I'd love to see the implementation but importing the workers.py from the user's bundle is not what I'd like to see either.
We should really keep the python entrypoint slim as possible.

I don't see the resistance to decorators, I really think this is the right call here.
We do the same for Rust Workers.

dom96: > but importing the workers.py from the user's bundle is not what I'd like to see either.

Shouldn't be necessary if my idea pans out

> I don't see the resistance to decorators, I really think this is the right call here.

I think it's worthwhile to avoid as much clutter as possible when defining workers.

danlapid: Alright, curious to see what you come up with 😄

dom96: Done! Essentially I moved the logic to workers.py, so each of our entrypoint classes implements it rather than it being implemented in the JS code.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -229,18 +229,17 @@ function makeEntrypointProxyHandler(
...
       const isKnownHandler = SUPPORTED_HANDLER_NAMES.includes(prop);
       if (isKnownHandler) {
         prop = 'on_' + prop;
       }
+
       return async function (...args: any[]): Promise<any> {
+        // Check if the requested method exists and if so, call it.
         const pyInstance = await pyInstancePromise;
         if (typeof pyInstance[prop] === 'function') {
-          const res = await doPyCall(pyInstance[prop], args);
```
danlapid, line 239: What's the reason for this change?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -217,7 +217,7 @@ async function initPyInstance(
       `There is no '${className}' class defined in the Python Worker's main module`
     );
   }
-  const res = await doPyCall(pyClassConstructor, args);
```
danlapid, line 220: I think this was required for some reason, @hoodmane I think did this.

src/pyodide/python-entrypoint-helper.ts
danlapid: Oh! I like that!
Really clean and simple 🤩

danlapid: This makes me want default Class entrypoints even more.
Get away from the on_fetch handlers and move towards full class-based handlers that are auto-wrapped.

src/pyodide/python-entrypoint-helper.ts
hoodmane: Yeah we need the `await` here since in 0.27 it will return a promise.

dom96: It looked to me like we are awaiting the result of a class constructor here, so it returning a promise didn't make sense to me. What did I miss?

src/pyodide/python-entrypoint-helper.ts
dom96: Looks like something I missed in the rebase

src/pyodide/python-entrypoint-helper.ts
dom96: And now Hood helped me make it more performant with some more advanced Python metaprogramming magic 🚀

src/pyodide/internal/workers.py
```diff
@@ -197,10 +198,19 @@ def _to_js_headers(headers: Headers):
 
 
 class Response(FetchResponse):
+    """
+    This class represents the response to an HTTP request, with a similar API to that of the web
+    `Response` API: https://developer.mozilla.org/en-US/docs/Web/API/Response.
+
+    This is a static class to enable defining `FetchResponse` using a constructor more similar to
+    JS. All non-static methods for this class should be defined on the `FetchResponse` class it
+    extends.
```
hoodmane, line 207: Why are `Response` and `FetchResponse` separate classes?

src/pyodide/internal/workers.py
```diff
@@ -416,6 +441,8 @@ def _make_blob_entry(e):
 
 
 def _is_iterable(obj):
+    if isinstance(obj, (str, bytes)):
+        return False
```
hoodmane, line 445: Yeah, it's unusual to want this. Probably it would have been better if they had been designed from the start to not be iterable but have a method that gives an iterable view.

src/pyodide/internal/workers.py
```diff
@@ -548,7 +575,17 @@ def last_modified(self) -> int:
 
 
 class Request:
-    def __init__(self, input: "Request | str", **other_options: Unpack[FetchKwargs]):
+    def __init__(
+        self, input: "Request | str | js.Request", **other_options: Unpack[FetchKwargs]
+    ):
+        if hasattr(input, "constructor") and (input.constructor.name == "Request"):
+            if len(other_options) > 0:
```
hoodmane, line 582: In the standard Python dialect this denoted as:
```suggestion
            if other_options:
```

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
         return await self.js_object.text()
 
 
+def serialize_to_python(obj: "JsProxy"):
```
hoodmane, line 722: Don't love the name -- if the wire format is JavaScript, then isn't this deserializing? What about `python_from_rpc` and `python_to_rpc`?

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+            return Request(obj)
+        elif obj.constructor.name == "Date":
+            # TODO: Pyodide should gain support for this, we should upstream this.
+            return datetime.datetime.fromtimestamp(obj.getTime() / 1000)
+        elif obj.constructor.name == "Error":
+            return Exception(obj.toString())
+        elif obj.constructor.name == "Number":
+            return obj.valueOf()
+
+        return obj
```
hoodmane, line 754: This shadows `obj` and doesn't close over anything, why not pull it out to a top level definition?

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+def serialize_to_python(obj: "JsProxy"):
+    """
+    Converts JS objects like Response, Request, Blob, etc. to equivalent Python objects defined in
+    this module and also other JS objects like Map, Set, etc. to equivalent Python stdlib objects.
+
+    This method is used for Workers RPC in Python to convert JavaScript objects to Python. As such
+    it does not support serializing all JS object types.
+    """
+
+    def convert(obj, convert, cache):
```
hoodmane, line 731: This shadows `obj` and doesn't close over anything, why not pull it out to a top level definition?

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+            return File(obj)
+        elif obj.constructor.name == "Request":
+            return Request(obj)
+        elif obj.constructor.name == "Date":
+            # TODO: Pyodide should gain support for this, we should upstream this.
+            return datetime.datetime.fromtimestamp(obj.getTime() / 1000)
+        elif obj.constructor.name == "Error":
+            return Exception(obj.toString())
+        elif obj.constructor.name == "Number":
+            return obj.valueOf()
```
hoodmane, line 752: I don't think this is reachable? Or necessary? Numbers already get converted. Also, even in JS if x is a number, `x.valueOf()` just returns `x`, so I don't think it would do anything as JavaScript code either.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,133 @@ async def text(self) -> str:
...
+    return result
+
+
+def _raise_on_disabled_type(value):
+    if isinstance(value, (tuple, bytearray)):
+        raise TypeError(f"{type(value)} cannot be sent over RPC.")
+
+    if _is_iterable(value):
+        for v in value:
+            _raise_on_disabled_type(v)
```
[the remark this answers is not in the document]
hoodmane: This still needs to be fixed.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+        if isinstance(obj, datetime.datetime):
+            # TODO: Pyodide should gain support for this, we should upstream this.
+            return js.Date.new(obj.timestamp() * 1000)
+
+        if isinstance(obj, Exception):
+            return js.Error.new(str(obj))
+
+        _raise_on_disabled_type(obj)
+
+        return obj
```
hoodmane, line 803: Also move this to a top level definition since it doesn't close over anything.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+def serialize_to_js(value) -> JsProxy:
+    """
+    Converts Python objects defined in this module (Response, Request, etc) and native Python types
+    like Map, Set, datetime to equivalent JavaScript types.
+
+    This method is used for Workers RPC in Python to convert Python objects to JavaScript. As such
+    it does not support serializing all Python object types.
+    """
+
+    def converter(obj, convert, cache):
```
hoodmane, line 787: Also move this to a top level definition since it doesn't close over anything.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+        if name == "fetch":
+            # TODO: Implement python native fetch for bindings.
+            return attr
+
+        if callable(attr):
+
+            async def wrapper(*args, **kwargs):
+                js_args = [serialize_to_js(arg) for arg in args]
+                js_kwargs = {k: serialize_to_js(v) for k, v in kwargs.items()}
+                result = await attr(*js_args, **js_kwargs)
```
hoodmane, line 828: Is `attr` necessarily an async method? If it isn't, awating it will raise.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+
+    def _getattr_helper(self, name):
+        attr = getattr(self._binding, name)
+        if name == "fetch":
+            # TODO: Implement python native fetch for bindings.
+            return attr
+
+        if callable(attr):
+
+            async def wrapper(*args, **kwargs):
```
hoodmane, line 825: When wrapping a function, you should decorate the wrapper with `@functools.wraps(original_function)`

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+    def __init__(self, binding):
+        self._binding = binding
+
+    def _getattr_helper(self, name):
+        attr = getattr(self._binding, name)
+        if name == "fetch":
+            # TODO: Implement python native fetch for bindings.
+            return attr
+
+        if callable(attr):
```
hoodmane, line 823: Maybe:
```suggestion
        if not callable(attr):
            return attr
```
and then we can reduce the indentation level by one for the wrapper.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+        result = self._getattr_helper(name)
+        setattr(self, name, result)
+        return result
+
+
+class _EnvWrapper:
+    def __init__(self, env):
+        self._env = env
+
+    def __getattr__(self, name):
```
hoodmane, line 845: I think this should also go in `_getattr_helper` and have the same `__getattr__` definition as `_FetcherWrapper`.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+        binding = getattr(self._env, name)
+        if binding.constructor.name == "Fetcher":
```
hoodmane, line 847: Can't this fail if the environment variable value is a string?

src/pyodide/internal/workers.py
```diff
@@ -691,18 +860,44 @@ def handler(func):
...
+        if len(args) > 1:
+            args = (args[0], _EnvWrapper(args[1])) + args[2:]
+
         return func(*args, **kwargs)
 
     return wrapper
 
 
+def _wrap_attr(attr):
+    if callable(attr):
```
hoodmane, line 881:
```suggestion
    if not callable(attr):
       return attr
```

src/pyodide/internal/workers.py
```diff
@@ -691,18 +860,44 @@ def handler(func):
...
+
+        async def wrapper(*args, **kwargs):
```
hoodmane, line 883: `@functools.wraps`

src/pyodide/internal/workers.py
```diff
@@ -691,18 +860,44 @@ def handler(func):
...
+            py_args = [serialize_to_python(arg) for arg in args]
+            py_kwargs = {k: serialize_to_python(v) for k, v in kwargs.items()}
+            result = await attr(*py_args, **py_kwargs)
```
hoodmane, line 886: Again, are we sure that the result of `attr` is awaitable?

src/pyodide/python-entrypoint-helper.ts
hoodmane: JSPI makes all calls into webassembly code return promises. The class constructor can stack switch, which looks synchronous from inside of webassembly but looks async from JS.

src/pyodide/internal/workers.py
dom96: Hmm, I don't recall why. I guess we can try and merge them in a separate PR.

src/pyodide/internal/workers.py
dom96: Yeah, I know it's more idiomatic to do it this way in Python, but I like to be explicit in these checks.

src/pyodide/internal/workers.py
dom96: Because it's only used by `serialize_to_python`, I can rename `obj` so it's not shadowed. I think keeping it inside this method makes the code easier to reason about though.

src/pyodide/internal/workers.py
dom96: Closing over variables isn't the only reason to create a nested function, I'd rather keep it in here if you don't mind. Grouping the code like this makes it easier to reason about IMO

src/pyodide/internal/workers.py
dom96: This is for `new Number("123123")` which doesn't get converted, or at least I had to add this to make the test that sends this via RPC

src/pyodide/internal/workers.py
dom96: This should have a different constructor name then, but will add a test for this.

src/pyodide/internal/workers.py
hoodmane: I don't believe this. If it's necessary, please make sure there is a test case that breaks when it is removed.

src/pyodide/internal/workers.py
```diff
@@ -682,6 +719,138 @@ async def text(self) -> str:
...
+
+        return obj
+
+    if not hasattr(obj, "constructor"):
+        return obj
+
+    result = obj.to_py(default_converter=convert)
+
+    if isinstance(result, JsProxy):
+        raise TypeError(
```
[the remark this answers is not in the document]
hoodmane: I think the correct thing to do would be to raise an error from `convert` in the fall through case instead of `return obj`. Because you're only checking the outermost level of the thing here, if `x` is a thing we can't convert then this check will let through `[x]`.

src/pyodide/internal/workers.py
hoodmane: What about moving it to top level as `serialize_to_python_default_converter`? I don't see why it's easier to reason about it when it's a nested function. I guess:

Pros of nested functions:
* clear that it isn't used anywhere else
* definition won't get separated by other functions from the function that uses it.

Cons of nested functions:
* unclear whether it might close over variables
* extra indentation

I only have to worry about whether there are other call sites to a top level function when I want to change it. Whereas I find myself very frequently checking what variables a function is closing over just to understand what the code means. So I think the readability disadvantages of nested functions outweigh the benefits.

src/pyodide/internal/workers.py
hoodmane: I think if `binding` is a string, then `binding.constructor` will raise an `AttributeError`.

src/pyodide/internal/workers.py
dom96: FWIW nested functions aren't the only ones that can access variables outside of their scope, non-nested functions can access global variables too. So I don't really see that as a unique concern for nested functions.

It would be nice if Python had the ability to force nested functions not to close over any args.

But alright, I'm not that strongly opposed to moving this into its own top-level function, so I'll just do it.

src/pyodide/internal/workers.py
dom96: It is definitely necessary, the test suite does test this path and it does fail if I remove it.

Also I just tried it in Pyodide REPL:

```
>>> x = js.Number.new("1234")
>>> isinstance(x, JsProxy)
True
>>> y = x.to_py()
>>> isinstance(y, JsProxy)
True
>>> y == 1234
False
```

src/pyodide/internal/workers.py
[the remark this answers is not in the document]
dom96: Ahh, good point. I was expecting `default_converter` to be called first, but it's actually only called when Pyodide cannot perform the conversion.

src/pyodide/internal/workers.py
hoodmane: ohhhh it's so weird that `new Number("10")` and `Number("10")` both work and do really different things... I tried `js.Number("1234")` and that works the way I expected.

src/workerd/server/tests/python/python-rpc/worker.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+async def assertRaises(func):
+    try:
+        await func()
+        raise ValueError("assertRaises failed")  # noqa: TRY301
+    except TypeError:
+        return
+    except Exception as e:
+        assert "DataCloneError" in str(e) or "TypeError" in str(e)
```
hoodmane, line 49: Maybe you should use:
```py
from unittest import TestCase
assertRaises = TestCase().assertRaises
```

src/workerd/server/tests/python/python-rpc/worker.py
```diff
@@ -0,0 +1,229 @@
...
+    )
+    assert js_response.method == "POST"
+    assert isinstance(js_response, Request)
+
+    # - Verify that a JS type can be sent.
+    py_response2 = await env.PythonRpc.handle_response(js.Response.new("a JS response"))
+    assert await py_response2.text() == "a JS response"
+
+    # Verify that sending unsupported types fails.
+    await assertRaises(lambda: env.PythonRpc.one_arg(CustomType(42)))
```
hoodmane, line 217: If you use `assertRaises` from `unittest` you can do:
```suggestion
    with assertRaises(ExpectedErrorType):
       env.PythonRpc.one_arg(CustomType(42))
```

hoodmane (approved):
Looks good to me other than suggestion for using unittest's `assertRaises`.

PR #3932, opened 2025-04-11
Sidecar port assignment: let bazel manage timeouts

npaun:
Some CI runners are so slow (looking at you, TeamCity) that we're pre-emptively failing the test while the sidecar is still setting up. This PR removes the custom timeout of 10 attempts to connect to the sidecar. Instead, we'll keep retrying forever until Bazel times out the test itself.

### DIFF
--- a/src/workerd/api/node/tests/sidecar-supervisor.mjs
+++ b/src/workerd/api/node/tests/sidecar-supervisor.mjs
@@ -8,7 +8,6 @@ import { scheduler } from 'node:timers/promises';
 
 const ANY_PORT = 0;
 const CONNECT_POLL_INTERVAL_MS = 500;
-const CONNECT_POLL_MAX_ATTEMPTS = 10;
 
 function getListeningServer() {
   const { promise, resolve } = Promise.withResolvers();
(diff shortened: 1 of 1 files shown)


anonrig (approved):

anonrig:
I'm fine with this but in the long term I think we should depend on Bazel for this.

anonrig:
Can you change the PR title and description?

src/workerd/api/node/tests/sidecar-supervisor.mjs
```diff
@@ -51,12 +50,6 @@ function waitForListening(port) {
         resolve();
       })
       .once('error', (err) => console.log('waiting for sidecar...', err.code));
-
-    if (attempts > CONNECT_POLL_MAX_ATTEMPTS) {
-      reject(
-        `Sidecar failed to listen to port ${port} after ${attempts} attempts.`
```
anonrig, line 57: You don't need attempts now I guess?

npaun: True

npaun: Ah no I see what you mean, I'm getting rid of the local variable now.

PR #3935, opened 2025-04-11
support upgrading insecure tcp connection in nodejs_compat

anonrig:
Enables node:net streams to be passed into node:tls. The following code now works:

```typescript
import { createConnection } from "mysql2/promise";

export default {
    // Create a connection to the MySQL database
    const connection = await createConnection({
      host: env.MYSQL_HOST,
      user: env.MYSQL_USER,
      password: env.MYSQL_PASSWORD,
      database: "mysql",
      port: 3306,
      disableEval: true,
      debug: true,
      ssl: {}
    });

    // Query the products table
    const [rows] = await connection.query("SHOW tables;");

    // Close the connection
    await connection.end();

    // Return the result as JSON
    const resp = new Response(JSON.stringify(rows), {
      headers: { "Content-Type": "application/json" },
    });

    return resp;
  },
```

### DIFF
--- a/src/node/internal/internal_tls_wrap.ts
+++ b/src/node/internal/internal_tls_wrap.ts
@@ -27,6 +27,8 @@ import {
   Socket,
   SocketOptions,
   _normalizeArgs,
+  onConnectionOpened,
+  onConnectionClosed,
 } from 'node-internal:internal_net';
 import { checkServerIdentity } from 'node-internal:internal_tls';
 import type {
@@ -251,7 +253,6 @@ export function TLSSocket(
   if (socket) {
     if (socket instanceof Socket && socket._handle) {
       wrap = socket;
-      // TODO(soon): Check TLS connection type and call starttls() if needed.
     } else {
       // Cloudflare Workers does not support any other socket type.
       throw new ERR_OPTION_NOT_IMPLEMENTED('options.socket');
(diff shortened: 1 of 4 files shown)


src/node/internal/internal_tls_wrap.ts
```diff
@@ -442,7 +443,35 @@ TLSSocket.prototype._finishInit = function _finishInit(this: TLSSocket): void {
...
+      socket: socket,
+      writer: socket.writable.getWriter(),
+      reader: socket.readable.getReader({ mode: 'byob' }),
+      bytesRead: 0,
+      bytesWritten: 0,
+    };
+
+    startRead(this);
+  } catch (error) {
+    // Ignore errors.
```
jasnell, line 472: Hmmm... why ignore errors here? Should errors at this point instead cause the socket to be destroyed?

jasnell (approved):

src/node/internal/internal_tls_wrap.ts
anonrig: The catch is required for this particular scenario. I've removed it but I will add it back

```
JSG_REQUIRE(!isSecureSocket, TypeError, "Cannot startTls on a TLS socket.");
```

jasnell (changes requested):
Actually.. going to block this we should not land that compat flag.

src/node/internal/internal_tls_wrap.ts
```diff
@@ -637,7 +684,10 @@ export function connect(...args: unknown[]): TLSSocket {
   tlssock[kConnectOptions] = options;
 
   if (cb) {
+    console.log('listening for event');
```
dario-piotrowicz, line 687: are the console logs intentional? 😮

dario-piotrowicz (commented):
Looks good to me, besides the `console.log`s 🙂 
(unless those are intentional for some reason 🤔)

jasnell (approved):

anonrig:
cc @cloudflare/wrangler would you mind reviewing this pull-request?

PR #3936, opened 2025-04-11
Update queueMicrotask to use jsg::Function

jasnell:
Working on incrementally adopting more use of the jsg abstractions where appropriate.

In this case, queueMicrotask really has no reason to use v8::Function directly, and the work to compile the args list was unnecessary since we know the function will be called with no arguments.

### DIFF
--- a/src/workerd/api/global-scope.c++
+++ b/src/workerd/api/global-scope.c++
@@ -721,20 +721,14 @@ jsg::JsString ServiceWorkerGlobalScope::atob(jsg::Lock& js, kj::String data) {
   return js.str(decoded.asBytes());
 }
 
-void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, v8::Local<v8::Function> task) {
-  auto fn = js.wrapReturningFunction(js.v8Context(),
-      JSG_VISITABLE_LAMBDA((this, fn = js.v8Ref(task)), (fn),
-          (jsg::Lock & js, const v8::FunctionCallbackInfo<v8::Value>& args)->v8::Local<v8::Value> {
-            return js.tryCatch([&]() -> v8::Local<v8::Value> {
-              auto function = fn.getHandle(js);
-
-              v8::LocalVector<v8::Value> argv(js.v8Isolate, args.Length());
-              for (int n = 0; n < args.Length(); n++) {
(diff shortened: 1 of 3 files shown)


src/workerd/api/tests/global-scope-test.js
```diff
@@ -134,7 +134,7 @@ export const queueMicrotask = {
       throws(() => globalThis.queueMicrotask(i), {
         message:
           "Failed to execute 'queueMicrotask' on 'ServiceWorkerGlobalScope': " +
-          "parameter 1 is not of type 'Function'.",
+          "parameter 1 is not of type 'function'.",
```
jasnell, line 137: This appears to just be an artifact of the type mapping system that uses `Function` for v8::Function and `function` for `jsg::Function`. Ah well.

src/workerd/api/global-scope.c++
```diff
@@ -721,20 +721,14 @@ jsg::JsString ServiceWorkerGlobalScope::atob(jsg::Lock& js, kj::String data) {
...
-              for (int n = 0; n < args.Length(); n++) {
-              argv[n] = args[n];
-              }
-
-              return jsg::check(
-                  function->Call(js.v8Context(), js.v8Null(), argv.size(), argv.data()));
+void ServiceWorkerGlobalScope::queueMicrotask(jsg::Lock& js, jsg::Function<void()> task) {
+  auto fn = js.wrapSimpleFunction(js.v8Context(),
+      JSG_VISITABLE_LAMBDA((this, fn = kj::mv(task)), (fn),
+          (jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>& args) {
```
npaun, line 727:
```suggestion
          (jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>&) {
```

If we're going to disregard a param in C++ we can leave it nameless.

npaun (approved):

src/workerd/api/global-scope.c++
jasnell: Will change that the next time I'm mucking around in this part of the code. It's not worth restarting CI over.

npaun: it's been so slow today...

anonrig (approved):

PR #3937, opened 2025-04-12
Convert websocket exceptions to JS exceptions.

danlapid:
This already works in production and is tested internally.
This commit ports the solution to workerd so that it works in local
development as well.
Websockets can be used with Workers in two main ways:
1. Eyeball connecting to a Worker acting as a websocket server
2. Worker connecting to an origin websocket server
This commit adds test for both of these cases.

### DIFF
--- a/src/workerd/api/tests/websocket-client-error-test.js
+++ b/src/workerd/api/tests/websocket-client-error-test.js
@@ -1,0 +1,61 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+// This test verifies that WebSocket protocol errors are properly converted to JavaScript errors
+// using the JsgifyWebSocketErrors handler.
+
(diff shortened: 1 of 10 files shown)


danlapid:
This PR is still in draft stage as I can't for the life of me figure out how to get server-test to pass.
I don't really understand what is going wrong.
If I simply comment out the logging line https://github.com/cloudflare/workerd/blob/main/src/workerd/server/server.c%2B%2B#L1296 the test passes consistently.
I am very confused, maybe someone will have a guess.

src/workerd/api/tests/websocket-client-error-test.js
```diff
@@ -0,0 +1,62 @@
...
+      }, 5000);
+
+      // Handle errors
+      ws.addEventListener('error', (event) => {
+        clearTimeout(timeoutId);
+        receivedError = true;
+
+        if (
+          event.message &&
+          event.message ===
```
anonrig, line 30:
```suggestion
          event.message ===
```

anonrig (approved):

PR #3942, opened 2025-04-14
Support SSE in asgi.py

danlapid:
For SSE to work we need to quickly return the Response and then start streaming responses.  
We also need the `receive` call to block until the response has finished streaming.


src/pyodide/internal/asgi.py
```diff
@@ -99,53 +99,118 @@ async def send(got):
...
+                    "text/event-stream"
+                ):
+                    is_sse = True
+
+                    # For SSE, create and return the response immediately after http.response.start
+                    resp = Response.new(
+                        readable, headers=Object.fromEntries(headers), status=status
+                    )
+                    result.set_result(resp)
+                    break
```
hoodmane, line 160: I think the following is clearer:
```suggestion
                    break

            if is_sse:
                # For SSE, create and return the response immediately after http.response.start
                resp = Response.new(
                    readable, headers=Object.fromEntries(headers), status=status
                )
                result.set_result(resp)

```

src/pyodide/internal/asgi.py
```diff
@@ -99,53 +99,118 @@ async def send(got):
...
+
+    # Create task to run the application in the background
+    app_task = create_task(run_app())
+
+    # Wait for the result (the response)
+    response = await result
+
+    # For non-SSE responses, we need to wait for the application to complete
+    if not is_sse:
+        await app_task
```
hoodmane, line 212: Should we `ctx.waitUntil(app_task)` in the `sse` case?

src/pyodide/internal/asgi.py
```diff
@@ -99,53 +99,118 @@ async def send(got):
...
+                finished_response.set()
+                result.set_exception(
+                    RuntimeError("The application did not generate a response")
+                )
+        except Exception as e:
+            # Handle any errors in the application
+            if not result.done():
+                await writer.close()  # Close the writer
+                finished_response.set()
+                result.set_exception(e)
```
hoodmane, line 202: A minor simplification:
```suggestion
            if not result.done():
                raise RuntimeError("The application did not generate a response")
        except Exception as e:
            # Handle any errors in the application
            if not result.done():
                await writer.close()  # Close the writer
                finished_response.set()
                result.set_exception(e)
```

src/pyodide/internal/asgi.py
```diff
@@ -99,53 +99,118 @@ async def send(got):
...
+                # If this is the last chunk, close the writer
+                if not more_body:
+                    await writer.close()
+                    finished_response.set()
+            else:
+                resp = Response.new(
+                    buf.data, headers=Object.fromEntries(headers), status=status
+                )
+                result.set_result(resp)
+                finished_response.set()
```
hoodmane, line 183: Is there a reason we don't do `writer.close()` here?

src/pyodide/internal/asgi.py
danlapid: I don't think so? Our SSE examples don't do that.

hoodmane (approved):
Approving to unblock since there's a test now. At some point we should try to restructure this to be less confusing.

src/pyodide/internal/asgi.py
```diff
@@ -98,54 +98,123 @@ async def send(got):
...
+
+    # Wait for the result (the response)
+    response = await result
+
+    # For non-SSE responses, we need to wait for the application to complete
+    if not is_sse:
+        await app_task
+    else:  # noqa: PLR5501
+        if ctx is not None:
+            ctx.waitUntil(create_once_callable(app_task))
```
danlapid, line 212: Adding a comment to block this PR until @hoodmane approves this line.

hoodmane: This seems odd don't you `waitUntil` a promise and not a callable?

danlapid: I pinged you on the internal chat
There was an error trying to wait until a Python promise
I thought this would convert it to a JavaScript promise 
It seemed to work
The blocking comment is exactly because I have no honest clue how to convert a Python promise to a JavaScript promise but this seemed to work

jasnell (approved):

src/pyodide/internal/asgi.py
hoodmane: I think this should be `create_proxy()` here and not `create_once_callable()`. Otherwise it looks good enough for now.

danlapid: How would I use create_proxy in this scenario?

danlapid: FWIW create_once_callable is working well both here and deployed
Why is it not appropriate
What's the difference?
The documentation is very very scarce

hoodmane: Well `create_once_callable` is generally only for callables. Promises are not callables. I would think that `ctx.waitUntil(create_once_callable(p))` would be roughly a no-op, same as if you hadn't called `ctx.waitUntil()` at all, because the callable is not a thenable, and so when it awaits the non-thenable it resolves immediately to the callable.

`create_once_callable()` should type check the argument to ensure that it is a callable and throw in cases like this where it isn't. Anyways `ctx.waitUntil(create_proxy(app_task))` should do what you want.

danlapid: Updated

PR #3944, opened 2025-04-14
Improves handling of unsupported types in Python SDK's Response.

dom96:
Ran into these while implementing some Python Workers for my personal projects.

This PR:

* Adds support for `list` to Response.json
* Ensures a Python Response is returned from Response.json rather than a JS response object.
* Verifies the input data types in Response() and Response.json so that passing invalid types throws


src/pyodide/internal/workers.py
```diff
@@ -198,23 +215,41 @@ def redirect(url: str, status: HTTPStatus | int = HTTPStatus.FOUND):
...
+        js_resp = None
         with _manage_pyproxies() as pyproxies:
             try:
-                return js.Response.json(
-                    to_js(
-                        data, dict_converter=js.Object.fromEntries, pyproxies=pyproxies
-                    ),
-                    **options,
-                )
+                if isinstance(data, list):
```
hoodmane, line 227: I think if `data` isn't a JsProxy the best think to do might be to use `json.dumps()` to serialize it in python and then use `Response.text()`.

src/pyodide/internal/workers.py
```diff
@@ -158,6 +158,23 @@ def __init__(
...
+                "TypedArray",
+                "DataView",
+                "FormData",
+                "ReadableStream",
+                "URLSearchParams",
+            ):
+                raise TypeError(
+                    f"Unsupported type in Response: {body.constructor.name}"
+                )
+        elif not isinstance(body, str) and not isinstance(body, FormData):
```
hoodmane, line 175: I think you can do:
```suggestion
        elif not isinstance(body, str | FormData):
```

src/workerd/server/tests/python/sdk/worker.py
```diff
@@ -395,6 +395,45 @@ async def can_use_event_decorator(env):
     assert text == "success"
 
 
+async def response_unit_tests(env):
+    response_json = Response.json([1, 2, 3])
```
hoodmane, line 399: We should check the content type at least once.

hoodmane (approved):
Looks good to me aside from a few comments.

src/pyodide/internal/workers.py
```diff
@@ -198,22 +215,22 @@ def redirect(url: str, status: HTTPStatus | int = HTTPStatus.FOUND):
...
-                    **options,
-                )
-            except JsException as exc:
-                raise _to_python_exception(exc) from exc
+        js_resp = None
+        try:
+            if isinstance(data, JsProxy):
+                js_resp = js.Response.json(data, **options)
+            else:
+                js_resp = js.Response.new(json.dumps(data), **options)
```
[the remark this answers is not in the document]
dom96: oof, good catch

PR #3945, opened 2025-04-14
add upgraded attribute to socket class

anonrig:
This enables developers to understand why their closed promise is resolving if they call socket.starttls()

### DIFF
--- a/src/workerd/api/node/tests/tls-nodejs-test.js
+++ b/src/workerd/api/node/tests/tls-nodejs-test.js
@@ -22,6 +22,7 @@
 // DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
+import { connect } from 'cloudflare:sockets';
 import tls from 'node:tls';
 import {
   strictEqual,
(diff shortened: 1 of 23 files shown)


danlapid:
Can you further elaborate here?
I don't understand why this makes sense over not resolving the closed promise on upgrade.
Nothing is actually being closed right?

anonrig:
> Can you further elaborate here? I don't understand why this makes sense over not resolving the closed promise on upgrade. Nothing is actually being closed right?

Referencing the sockets-api proposal:

```
In this secureTransport mode of operation the socket begins the connection in plain-text, with messages read and written without any encryption. Then once the startTls method is called on the socket, the following shall take place:

- the original socket is closed, though the original connection is kept alive
- a secure TLS connection is established over that connection
- a new socket is created and returned from the startTls call
```

jasnell:
@danlapid ... this is a change @anonrig @dom96 and I agreed to in order to help address some limitations in the Socket API that Yagiz is hitting. Specifically, consider the following:

```js
const originalSocket = connect('...', { secureTransport: 'secureTls' });
originalSocket.closed.then(() => console.log('closed'));

const secureSocket = originalSocket.startTls();
```

This is the established API pattern for Socket... essentially, the `startTls()` call leads the original socket to be closed and discarded and users should use the newly returned `secureSocket` to continue to use the connection. The underlying connection remains in tact but the `Socket` instance itself is replaced.

The key issue that Yagiz is hitting is that this means the `originalSocket.closed` promise gets resolved but there's no indication given *why* the socket was closed. The user code making use of startTls would need to do some additional state book keeping in order to know. This new property just provides the necessary detail.

jasnell (approved):

dom96 (approved):

danlapid:
SGTM, thanks for the explanation.
Would be nice to have had it in the PR description.

danlapid (approved):

anonrig:
@irvinebroque @mikenomitch FYI: Cloudflare documentation needs to be updated after this change. Particularly: https://developers.cloudflare.com/workers/runtime-apis/tcp-sockets/

anonrig:
cc @cloudflare/wrangler can you review these changes?

dario-piotrowicz (approved):
Looks good to me 🙂

PR #3946, opened 2025-04-14
add runtime team to types/generated-snapshot code owners

anonrig:
When generated-snapshots are changed (either through updating a type in a C++ file, or through adding a new function) we require Wrangler team input, but it is often not needed.

An example: https://github.com/cloudflare/workerd/pull/3945


jasnell (changes requested):
Per internal discussion, I don't believe this is what we actually want.

PR #3949, opened 2025-04-14
Resolve `opened` promise on new socket returned from startTls.

dom96:
This is a (simpleish) change to ensure that the `opened` promise on the socket returned by `startTls` gets resolved. 

The `opened` promise is resolved when the `tlsStarter` completes, which doesn't always mean that the TLS upgrade was successful. In particular, because we use a custom capsule protocol in EW and we don't eagerly handle the handshake through that protocol, the `opened` promise will resolve before the handshake is fully completed.

Fixing this is something we should possibly do but it will require more work. For future reference, in EW the `ProxyToHostHttpClientAdapter::connect` class handles the immediate TLS and startTls cases, it defines a TLS starter which begins the TLS handshake by sending the appropriate packet to OER (via CapsuleAsyncIoStream::startTls). This could potentially be changed to wait for the handshake in that code path and fail the TLS starter when appropriate, though it's unclear whether this is possible while other pending reads are in progress from the JS side.

### DIFF
--- a/src/workerd/api/sockets.c++
+++ b/src/workerd/api/sockets.c++
@@ -326,56 +326,68 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
   auto openedPrPair = js.newPromiseAndResolver<SocketInfo>();
   auto secureStreamPromise = context.awaitJs(js,
       writable->flush(js).then(js,
-          [this, domain = kj::heapString(domain), tlsOptions = kj::mv(tlsOptions),
-              tlsStarter = kj::mv(tlsStarter),
-              openedResolverResolve = openedPrPair.resolver.addRef(js),
-              openedResolverReject = openedPrPair.resolver.addRef(js),
-              remoteAddress = kj::str(remoteAddress)](jsg::Lock& js) mutable {
-    writable->detach(js);
-    readable = readable->detach(js, true);
+          // The openedResolver is a jsg::Promise::Resolver. It should be gc visited here in
+          // case the opened promise is resolves captures a circular references to itself in
+          // JavaScript (which is most likely). This prevents a possible memory leak.
+          // We also capture a strong reference to the original Socket instance that is being
+          // upgraded in order to prevent it from being GC'd while we are waiting for the
+          // flush to complete. While it is unlikely to be GC'd while we are waiting because
+          // the user code *likely* is holding a active reference to it at this point, we
+          // don't want to take any chances. This prevents a possible UAF.
+          JSG_VISITABLE_LAMBDA((self = JSG_THIS, domain = kj::heapString(domain),
+                                   tlsOptions = kj::mv(tlsOptions), tlsStarter = kj::mv(tlsStarter),
+                                   openedResolver = openedPrPair.resolver.addRef(js),
+                                   remoteAddress = kj::str(remoteAddress)),
+              (self, openedResolver), (jsg::Lock & js) mutable {
+                auto& context = IoContext::current();
 
-    // We should set this before closedResolver.resolve() in order to give the user
-    // the option to check if the closed promise is resolved due to upgrade or not.
-    upgraded = true;
-    closedResolver.resolve(js);
+                self->writable->detach(js);
+                self->readable = self->readable->detach(js, true);
 
-    auto acceptedHostname = domain.asPtr();
-    KJ_IF_SOME(s, tlsOptions) {
-      KJ_IF_SOME(expectedHost, s.expectedServerHostname) {
-        acceptedHostname = expectedHost;
-      }
-    }
+                // We should set this before closedResolver.resolve() in order to give the user
+                // the option to check if the closed promise is resolved due to upgrade or not.
+                self->upgraded = true;
+                self->closedResolver.resolve(js);
 
-    // All non-secure sockets should have a tlsStarter. Though since tlsStarter is an IoOwn, if
-    // the request's IoContext has ended then `tlsStarter` will be null. This can happen if the
-    // flush operation is taking a particularly long time (EW-8538), so we throw a JSG error if
-    // that's the case.
-    JSG_REQUIRE(
-        *tlsStarter != kj::none, TypeError, "The request has finished before startTls completed.");
+                auto acceptedHostname = domain.asPtr();
+                KJ_IF_SOME(s, tlsOptions) {
+                KJ_IF_SOME(expectedHost, s.expectedServerHostname) {
+                acceptedHostname = expectedHost;
+                } else {
+                }  // Needed to avoid compiler error/warning
+                } else {
+                }  // Needed to avoid compiler error/warning
 
-    auto& context = IoContext::current();
-    auto tlsStarterResult =
-        context.awaitIo(js, KJ_ASSERT_NONNULL(*tlsStarter)(acceptedHostname))
-            .then(js,
-                [openedResolverResolve = kj::mv(openedResolverResolve),
-                    remoteAddress = kj::mv(remoteAddress)](jsg::Lock& js) mutable {
-      openedResolverResolve.resolve(js,
-          SocketInfo{
-            .remoteAddress = kj::mv(remoteAddress),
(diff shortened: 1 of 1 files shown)


jasnell (approved):

anonrig (approved):

src/workerd/api/hyperdrive.c++
```diff
@@ -44,7 +44,8 @@ jsg::Ref<Socket> Hyperdrive::connect(jsg::Lock& js) {
   // some users may want it anyway.
   auto nullTlsStarter = kj::heap<kj::TlsStarterCallback>();
   auto sock = setupSocket(js, kj::mv(conn), kj::str(getHost(), ":", getPort()), kj::none,
-      kj::mv(nullTlsStarter), false, kj::str(this->randomHost), false);
+      kj::mv(nullTlsStarter), false, kj::str(this->randomHost), false,
+      kj::none /* maybeOpenedPrPair */);
```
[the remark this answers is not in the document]
dom96: We need the upstream PR to change the test anyway and I prefer to be explicit about arguments like this.

src/workerd/api/sockets.c++
```diff
@@ -346,9 +352,28 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
...
+          });
+    },
+                [openedResolverReject = kj::mv(openedResolverReject)](
+                    jsg::Lock& js, jsg::Value&& error) mutable {
+      openedResolverReject.reject(js, jsg::JsValue(error.getHandle(js)));
+    });
+
+    auto secureStream = context.awaitJs(js, kj::mv(tlsStarterResult))
                             .then([stream = connectionStream->addWrappedRef()]() mutable
                                 -> kj::Own<kj::AsyncIoStream> { return kj::mv(stream); });
```
jasnell, line 375: @kentonv ... wanted to see if you can double check the safety on this part of the change as @dom96 hit an asan issue on his first iteration. This one *appears* to be ok in that it's not throwing asan errors in bazel test but just want another set of eyes.

In Dom's original code... The asan issue was related to a use-after-free with the `connectionStream->addWrappedRef()` here. In Dom's initial iteration, he had the JS promise (the `context.awaitIo` at line 358) capturing the ref here which is a `kj::Own` of an `AsyncIoStream`. You can see the asan error here: https://github.com/cloudflare/workerd/actions/runs/14468777877/job/40577235976?pr=3949#step:9:101 . It's a fun one. The UAF was triggered in `~Server` when it is cleaning up the collection of services. It appears the `NetworkService` (which the `kj::<AsyncIoStream>` appears to hold a reference to is destroyed before the `WorkerService` ... specifically in that when the `WorkerService` is destroyed it destroys the `Isolate` which causes `dropWrappers` to run which in this test causes the things holding the `kj::Own<AsyncIoStream>` to be destroyed.... Basically, the `NetworkService` was being destroyed before it's reference... doh!

In this current iteration, however, the asan error disappears but I'm still not super confident that this is correct. The `kj::Own<AsyncIoStream>` returned by `connectionStream->addWrappedRef()` is held by the kj promise instead of the jsg promise. Dominik can likely best explain the logic of what he's trying to achieve with this code.

kentonv: > In Dom's initial iteration, he had the JS promise (the context.awaitIo at line 358) capturing the ref here which is a kj::Own of an AsyncIoStream.

Was there an IoOwn involved? If not I think this is a standard missing-IoOwn error? A JS promise cannot capture an AsyncIoStream except using IoOwn. Using the IoOwn ensures that the object is destroyed when the IoContext is destroyed, which should be before the server can be torn down. Without the IoOwn the object would have been kept by the JS heap until isolate teardown, which is too late to be destroying I/O objects.

Holding the I/O object on a KJ promise instead should be safe -- the KJ promise will presumably be canceled one way or another before the server can be torn down.

src/workerd/api/sockets.c++
```diff
@@ -346,9 +352,28 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
     // that's the case.
     JSG_REQUIRE(
         *tlsStarter != kj::none, TypeError, "The request has finished before startTls completed.");
-    auto secureStream = KJ_ASSERT_NONNULL(*tlsStarter)(acceptedHostname)
+
+    auto& context = IoContext::current();
+    auto tlsStarterResult =
+        context.awaitIo(js, KJ_ASSERT_NONNULL(*tlsStarter)(acceptedHostname))
+            .then(js,
```
kentonv, line 359: `context.awaitIo(js, promise).then(func)` should almost always be changed to `context.awaitIo(js, promise, func)` instead, as this limits the lifetime of `func` to the I/O operation and allows it to capture I/O objects. If you use `.then()`, then the func's lifetime is governed by garbage collection, which is almost always worse.

That said, in this particular case, your captures look safe as you aren't capturing any I/O objects.

However, you _are_ capturing GC objects which could create cycles, and you are not using `JSG_VISITABLE_LAMBDA` to make the lambdas GC-visitable, so this could lead to memory leaks. Again, this is a non-issue if you pass the func directly to `awaitIo`.

src/workerd/api/sockets.c++
```diff
@@ -346,9 +352,28 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
...
+                [openedResolverResolve = kj::mv(openedResolverResolve),
+                    remoteAddress = kj::mv(remoteAddress)](jsg::Lock& js) mutable {
+      openedResolverResolve.resolve(js,
+          SocketInfo{
+            .remoteAddress = kj::mv(remoteAddress),
+            .localAddress = kj::none,
+          });
+    },
+                [openedResolverReject = kj::mv(openedResolverReject)](
```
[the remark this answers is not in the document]
jasnell: ooo... I hadn't considered that! good idea.

src/workerd/api/sockets.c++
```diff
@@ -346,9 +352,28 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
...
+                    jsg::Lock& js, jsg::Value&& error) mutable {
+      openedResolverReject.reject(js, jsg::JsValue(error.getHandle(js)));
+    });
+
+    auto secureStream = context.awaitJs(js, kj::mv(tlsStarterResult))
```
kentonv, line 373: This awaitJs() is waiting on the result of an awaitIo(), which creates a weird need to jump from KJ event loop to JS microtask loop and back to KJ event loop. Could this be avoided by instead forking the TLS starter promise, passing one branch to awaitIo and waiting directly on the other branch here?

src/workerd/api/sockets.c++
jasnell: Yep, that was exactly the issue. Thank you!

src/workerd/api/sockets.c++
jasnell: oh good catch! There's actually a `JSG_VISITABLE_LAMBDA necessary higher up in this function (in the `writable->flush(js).then(js, ...` call since that lambda ends up capturing the `jsg::Promise::Resolver` and the opened promise most likely captures a reference to itself in it's continuations.

src/workerd/api/sockets.c++
```diff
@@ -321,43 +323,78 @@ jsg::Ref<Socket> Socket::startTls(jsg::Lock& js, jsg::Optional<TlsOptions> tlsOp
...
+                self->closedResolver.resolve(js);
+
+                auto acceptedHostname = domain.asPtr();
+                KJ_IF_SOME(s, tlsOptions) {
+                KJ_IF_SOME(expectedHost, s.expectedServerHostname) {
+                acceptedHostname = expectedHost;
+                } else {
+                }  // Needed to avoid compiler error/warning
+                } else {
+                }  // Needed to avoid compiler error/warning
```
dom96, line 359: Hmm, strange formatting here, formatter bug?

PR #3959, opened 2025-04-16
Fixes entropy issue in Python RPC.

dom96:
Test upstream.


hoodmane:
Is it not possible to test this in workerd?

dom96:
> Is it not possible to test this in workerd?

The issue only happened in prod afaik. But yeah, I can add a test in workerd too.

hoodmane:
If it can't be reproduced in workerd, there isn't much reason to add a test.

src/workerd/server/tests/python/worker-entrypoint/worker.py
```diff
@@ -11,6 +13,8 @@ def __init__(self, state, env):
         self.counter = 0
 
     async def rpc_trigger_func(self):
+        # Test that entropy works in WorkerEntrypoint...
+        os.urandom(4)
```
danlapid, line 17: This doesn't actually reproduce the problem but it doesn't hurt to have it here.
The internal test actually do.

PR #3962, opened 2025-04-16
Add isFloat16Array APIs for nodejs_compat

jasnell:
`Float16Array` has reached stage 4 in TC-39 and v8 is nearly done with the implementation. While it is still behind a flag, the C++ APIs are available and Node.js is about to add the `util.types.isFloat16Array` API. This commit adds the API under `nodejs_compat` so that we can prepare for it to be enabled by default in the very near future.

This also expands the testing and completes the implementation of `isKeyObject(...)` since that was still pending.

### DIFF
--- a/src/node/internal/internal_comparisons.ts
+++ b/src/node/internal/internal_comparisons.ts
@@ -42,6 +42,7 @@ import {
   isBooleanObject,
   isBigIntObject,
   isSymbolObject,
+  isFloat16Array,
   isFloat32Array,
   isFloat64Array,
 } from 'node-internal:internal_types';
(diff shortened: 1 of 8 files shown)


anonrig (approved):

npaun:
This PR ought to make a WPT test start passing. I don't know how to give you a fancy PR code suggestion on an unmodified file, but in 
`src/wpt/api-tests.ts` you may need to remove this line:

```
'Fetch with POST with Float16Array body',
```

jasnell:
This PR does not actually enable `Float16Array` availability in the JavaScript side yet so I don't think it will enable that test just yet.

npaun:
Ah I didn't realize there were more steps involved in that, sorry.

PR #3963, opened 2025-04-16
Fetch pedantry: do not set statusText unless spec explicitly says to

npaun:
Fetch spec:

> A [response](https://fetch.spec.whatwg.org/#concept-response) has an associated status message. Unless stated otherwise it is the empty byte sequence.

https://fetch.spec.whatwg.org/#concept-response-status-message

So actually it turns out we're not supposed to helpfully set a default status message to match the code in most cases. The exception when we're supposed to do so are when  `about`, `blob` and `data` URLs are fetched.

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1461,7 +1461,10 @@ jsg::Ref<Response> Response::constructor(jsg::Lock& js,
       }
     }
   } else {
-    statusText = kj::str(defaultStatusText(statusCode));
+    // A response has an associated status message. Unless stated otherwise it is the empty byte sequence.
+    // https://fetch.spec.whatwg.org/#concept-response-status-message
+
+    statusText.emplace(kj::str(""));
   }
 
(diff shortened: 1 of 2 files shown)


src/workerd/api/http.c++
```diff
@@ -1461,7 +1461,10 @@ jsg::Ref<Response> Response::constructor(jsg::Lock& js,
       }
     }
   } else {
-    statusText = kj::str(defaultStatusText(statusCode));
+    // A response has an associated status message. Unless stated otherwise it is the empty byte sequence.
+    // https://fetch.spec.whatwg.org/#concept-response-status-message
+
+    statusText.emplace(kj::str(""));
```
jasnell, line 1467: Hmm.. Unfortunately -- and this makes me sad -- there's a small possibility that this could be a breaking change for someone. I think we likely need a compat flag for this.

npaun:
This PR is valid but it probably contains more merge conflicts than actual content. Let's just close it and redo this fix with opencode.

PR #3974, opened 2025-04-17
WPT: Enable fetch/api/abort tests

npaun:
* Adds promise_rejects_dom to harness
* Enables the WPT tests for fetch abort.
  - Our implementation seems to be quite far off from the spec unfortunately.

### DIFF
--- a/src/wpt/harness/harness.ts
+++ b/src/wpt/harness/harness.ts
@@ -437,6 +437,14 @@ declare global {
     descriptionOrFunc: string | ThrowingFn,
     maybeDescription?: string
   ): void;
+  function promise_rejects_dom(
+    test: Test,
+    type: number | string,
+    promiseOrConstructor: Promise<unknown> | typeof DOMException,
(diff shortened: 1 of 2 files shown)


src/wpt/harness/harness.ts
```diff
@@ -881,6 +889,76 @@ globalThis.assert_throws_dom = (
...
+ * `assert_throws_dom <#assert_throws_dom>`_.
+ * @param promiseOrConstructor - Either the constructor
+ * for the expected exception (if the exception comes from another
+ * global), or the promise that's expected to reject (if the
+ * exception comes from the current global).
+ * @param descriptionOrPromise - Either the
+ * promise that's expected to reject (if the exception comes from
+ * another global), or the optional description of the condition
+ * being tested (if the exception comes from the current global).
+ * @param [description] - Description of the condition
```
anonrig, line 918: Can you also make sure jsdoc param names match with the actual function param names?

```suggestion
 * @param [maybeDescription] - Description of the condition
```

PR #3982, opened 2025-04-18
Speed up workers-sdk-test.

danlapid:
The check "workers-sdk-test" is a required check that depends on "test". 
That means it depends on the entire matrix to succeed. 
In reality it only actually depends on test linux-arm. 
Currently all of "test" is a required check because of it.
Another problem is that it currently only runs once all of "test" succeeds instead of right after it's actual dependencies succeed.

This commit changes that so workers-sdk-test depends only on test-linux. 
A new check that only runs the test on the linux with empty suffix configuration.
This should restore our ability to choose which test jobs are required and which aren't.

### DIFF
--- a/build/ci.bazelrc
+++ b/build/ci.bazelrc
@@ -57,8 +57,8 @@ build:ci-linux-common --copt='-Wno-error=deprecated-declarations'
 build:ci-linux-common --action_env=CC=/usr/lib/llvm-18/bin/clang
 build:ci-linux-common --host_action_env=CC=/usr/lib/llvm-18/bin/clang
 
-build:ci-linux --config=ci-linux-common
-build:ci-linux-arm --config=ci-linux --remote_download_regex=".*src/workerd/server/workerd.*"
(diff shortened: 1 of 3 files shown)


anonrig (approved):

build/ci.bazelrc
```diff
@@ -57,8 +57,8 @@ build:ci-linux-common --copt='-Wno-error=deprecated-declarations'
 build:ci-linux-common --action_env=CC=/usr/lib/llvm-18/bin/clang
 build:ci-linux-common --host_action_env=CC=/usr/lib/llvm-18/bin/clang
 
-build:ci-linux --config=ci-linux-common
-build:ci-linux-arm --config=ci-linux --remote_download_regex=".*src/workerd/server/workerd.*"
+build:ci-linux --config=ci-linux-common --remote_download_regex=".*src/workerd/server/workerd.*"
+build:ci-linux-arm --config=ci-linux
```
fhanau, line 61: Nit: make this line `--config=ci-linux-common` so that binary is only downloaded for the x64 job.

PR #3984, opened 2025-04-18
ignore certain values for nodejs redis and mssql client

anonrig:
These options break Redis and mssql client. 

Redis reproduction:

```typescript
import { createClient } from 'redis';

export default {
    const client = createClient({})

    await client.connect()

    await client.hSet('user-session:123', {
        name: 'John',
        surname: 'Smith',
        company: 'Redis',
        age: 29
    })

    const rows = await client.hGetAll('user-session:123');
    // Return the result as JSON
    const resp = new Response(JSON.stringify(rows), {
      headers: { "Content-Type": "application/json" },
    });

    return resp;
  },
```

Mssql reproduction:

```typescript
import sql from 'mssql';

export default {
	  const config = {}
		const conn = await sql.connect(config);

		const rows = await conn.request().query('select * from SYSOBJECTS')

    // Return the result as JSON
    const resp = new Response(JSON.stringify(rows), {
      headers: { "Content-Type": "application/json" },
    });

    return resp;
  },
```

### DIFF
--- a/src/node/internal/internal_net.ts
+++ b/src/node/internal/internal_net.ts
@@ -269,7 +269,8 @@ export function Socket(this: Socket, options?: SocketOptions): Socket {
     throw new ERR_OPTION_NOT_IMPLEMENTED('options.fd');
   }
 
-  // TODO(soon): Support both of these options.
+  // We do not support the noDelay and keepAlive options at this
+  // time and will just ignore them if they are passed.
   //
   // We shouldn't throw an error for the following validations,
   // because it breaks packages such as redis.
--- a/src/node/internal/internal_net.ts
+++ b/src/node/internal/internal_net.ts
@@ -245,7 +245,7 @@ export function Socket(this: Socket, options?: SocketOptions): Socket {
       'is not supported'
     );
   }
-  if (typeof options?.keepAliveInitialDelay !== 'undefined') {
+  if (options?.keepAliveInitialDelay !== undefined) {
     validateNumber(
       options.keepAliveInitialDelay,
       'options.keepAliveInitialDelay'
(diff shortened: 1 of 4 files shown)


src/node/internal/internal_tls_common.ts
```diff
@@ -56,12 +55,16 @@ export function SecureContext(
       maxVersion
     );
   }
-  if (minVersion !== undefined) {
-    throw new ERR_OPTION_NOT_IMPLEMENTED('minVersion');
-  }
-  if (maxVersion !== undefined) {
-    throw new ERR_OPTION_NOT_IMPLEMENTED('maxVersion');
-  }
+  // TODO(soon): Support this.
```
jasnell, line 58: I don't think "Support this" is a good todo here. We don't have any way of telling the underlying layers which TLS version to support. I would probably just replace this entire commented block with something like:

```js
// We do not support the minVersion and maxVersion options at this
// time and will just ignore them if they are passed.
```

src/node/internal/internal_net.ts
```diff
@@ -269,12 +269,18 @@ export function Socket(this: Socket, options?: SocketOptions): Socket {
...
   }
 
-  if (options.noDelay) {
-    throw new ERR_OPTION_NOT_IMPLEMENTED('truthy options.noDelay');
-  }
-  if (options.keepAlive) {
-    throw new ERR_OPTION_NOT_IMPLEMENTED('truthy options.keepAlive');
-  }
+  // TODO(soon): Support both of these options.
+  //
```
jasnell, line 273: I don't think we really have any intention of supporting these. I would just replace this whole block with a comment like:

```
// We do not support the noDelay and keepAlive options at this
// time and will just ignore them if they are passed.
```

src/workerd/api/node/tests/net-nodejs-test.js
```diff
@@ -457,8 +457,7 @@ export const testNetConnectImmediateFinish = {
 // something different than the original Node.js test
 export const testNetConnectKeepAlive = {
   async test() {
-    // Test that setKeepAlive call does not throw.
-    throws(() => new net.Socket({ keepAlive: true }));
+    // throws(() => new net.Socket({ keepAlive: true }));
```
npaun, line 460: Could you please add a comment explaining that we're choosing not to throw on keep alive?

npaun (approved):

jasnell (approved):

PR #3987, opened 2025-04-18
add support for duplex socket to node:net

anonrig:
It seems this is required for mssql to work.

### DIFF
--- a/src/node/internal/internal_tls_jsstream.ts
+++ b/src/node/internal/internal_tls_jsstream.ts
@@ -1,0 +1,308 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+//
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+/* eslint-disable @typescript-eslint/no-redundant-type-constituents */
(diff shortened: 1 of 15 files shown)


src/workerd/api/node/tests/tls-nodejs-test.js
```diff
@@ -776,6 +778,77 @@ export const testStartTlsBehaviorOnUpgrade = {
...
+// https://github.com/nodejs/node/blob/52d95f53e466016120048fb43b3732ff9089ecd7/test/parallel/test-tls-destroy-whilst-write.js
+export const testTlsDestroyWhilstWrite = {
+  async test() {
+    const { promise, resolve } = Promise.withResolvers();
+    const delay = new stream.Duplex({
+      read: function read() {},
+      write: function write(data, enc, cb) {
+        queueMicrotask(function () {
+          cb();
+        });
```
jasnell, line 791:
```suggestion
        queueMicrotask(cb);
```

src/node/internal/internal_tls_jsstream.ts
```diff
@@ -0,0 +1,308 @@
...
+      closed: closePromise.promise,
+      releaseLock: async (): Promise<void> => {},
+    });
+    const handle: Socket['_handle'] = {
+      reading: true,
+      bytesRead: 0,
+      bytesWritten: 0,
+      socket: {
+        startTls(): CloudflareSocket {
+          throw new Error('This should not have triggered');
```
jasnell, line 80: I know this is more of an internal error than a user facing one but the message could be improved more

src/node/internal/internal_tls_jsstream.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    stream.pause();
+    stream.on('error', (err) => this.emit('error', err));
+    const ondata = (chunk: string | Buffer): void => {
+      if (typeof chunk === 'string' || stream.readableObjectMode === true) {
+        // Make sure that no further `data` events will happen.
+        stream.pause();
+        stream.removeListener('data', ondata);
+
+        this.emit('error', new ERR_STREAM_WRAP());
```
jasnell, line 112: Generally in a Node stream you want to call destroy(error) rather than emitting error yourself

anonrig: This is done to be in par with nodejs behavior

jasnell:
The one immediate bug that needs to be addressed before anything else can be looked at here is you'll need to determine why the `net.Socket` implementation is not emitted the `readable` event. Ignore the rest of the tls stuff for now and investigate that. It's a bit non-trivial and you'll need to look into the implementation of `stream.Readable` to determine how and when the `readable` event is emitted. 

This is the test case to focus on:

```js
export const netReadableTest = {
  async test(_, env) {
    const socket = net.connect(env.JS_STREAM_SERVER_PORT);
    // comment out the following two lines and comment
   // the socket.on('readable',, ... block out and it works.
    // socket.on('data', console.log);
    // socket.on('end', resolve);
    socket.on('readable', () => {
      const chunk = socket.read();
      if (chunk === null) resolve();
    });
    await promise;
  }
};
```

The rest can be investigated further once thi case is working reliably.

danlapid:
Where is this?

anonrig:
> Where is this?

Waiting for @jasnell to take a look. The tests are failing because we're not emitting readable event at all due to our byob solution.

anonrig:
Rebased and force pushed.

anonrig:
cc @npaun @jasnell can you review this pull-request? I'd like to land this.

jasnell (approved):

src/node/internal/internal_tls_jsstream.ts
```diff
@@ -0,0 +1,308 @@
...
+  public finishWrite(
+    _handle: JSStreamSocket['_handle'],
+    _errCode: string | null
+  ): void {
+    // The write request might already have been cancelled.
+    if (this[kCurrentWriteRequest] === null) return;
+    // const req = this[kCurrentWriteRequest];
+    this[kCurrentWriteRequest] = null;
+
+    // handle.finishWrite(req, errCode);
```
npaun, line 279: Commented out code

npaun (approved):

PR #3990, opened 2025-04-19
Make JSPI work correctly

hoodmane:
This fixes two problems:
1. pools made `newWasmModule` not work at startup but JSPI relies on it.
2. can't stack switch if the handler is sync.

- [ ] Update bundle


src/pyodide/types/Pyodide.d.ts
```diff
@@ -5,6 +5,10 @@ declare type Handler = (...args: any[]) => any;
 declare type PyCallable = ((...args: any[]) => any) & {
   call: (...args: any[]) => any;
   callRelaxed: (...args: any[]) => any;
+  callWithOptions?: (
```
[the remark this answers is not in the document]
danlapid: Yeah I saw that
Still can’t see where it is actually used

jasnell (approved):

src/workerd/server/tests/python/jspi/worker.py
```diff
@@ -0,0 +1,23 @@
+from asyncio import sleep
+
+from pyodide import __version__
+
+
+def jspi_sleep():
+    if __version__ == "0.26.0a2":
+        print("JSPI not supported on 0.26.0a2, skipping")
+        return
```
dom96, line 9: You skip it already via the `skip_python_flags`, may be good to remove this so that it doesn't silently pass if we do want to run it against 0.26.0 in the future.

dom96 (approved):

src/workerd/server/tests/python/jspi/worker.py
hoodmane: It won't ever work against 0.26.0

dom96: oh, duh, I forgot about dev. Makes sense, ignore me.

PR #3994, opened 2025-04-21
add node:fs base classes

anonrig:
> Work in progress.

Example usage internal usage:

- fs.getResource() will return different classes that extend from "Node" class depending on the root path of the file.

```typescript
import { default as fs } from 'node-internal:filesystem';

const resource = fs.get('/bundle/hello.jpeg');
resource.getStat() // Returns: { name, ... }
resource.asyncOnly;
resource.syncOnly;
resource.readable;
resource.writable;

const fd = resource.getFd();
fd.close();
``

### DIFF
--- a/src/workerd/api/filesystem.h
+++ b/src/workerd/api/filesystem.h
@@ -1,0 +1,96 @@
+// Copyright (c) 2017-2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#pragma once
+
+#include <workerd/jsg/jsg.h>
+#include <workerd/jsg/memory.h>
+#include <workerd/jsg/url.h>
+
+#include <kj/filesystem.h>
+#include <kj/string.h>
+
+namespace workerd::api::filesystem {
+
+enum class NodeType : kj::uint { FILE, DIRECTORY };
+
+class Handle: public jsg::Object {
+ public:
+  Handle(kj::Path path): path(kj::mv(path)) {}
+
+  void close();
+
+  JSG_RESOURCE_TYPE(Handle) {
+    JSG_METHOD(close);
+  }
+
+ private:
+  kj::Path path;
+};
+
+class Node: public jsg::Object {
+ public:
+  Node(kj::String name, kj::Path path, kj::Date modifiedAt, kj::Date createdAt, NodeType type)
+      : name(kj::mv(name)),
+        path(kj::mv(path)),
+        modifiedAt(kj::mv(modifiedAt)),
+        createdAt(kj::mv(createdAt)),
+        type(type) {};
+
+  virtual bool isReadable() const {
+    return false;
+  }
+  virtual bool isWritable() const {
+    return false;
+  }
+  virtual bool isAsyncOnly() const {
+    return false;
+  }
+  virtual bool isSyncOnly() const {
+    return false;
+  }
+
+  virtual kj::Maybe<jsg::Ref<Handle>> getFd() {
+    return kj::none;
+  }
+
+  struct Stat {
+    kj::StringPtr name;
+    kj::StringPtr path;
+    uint64_t modifiedAt;
+    uint64_t createdAt;
+    kj::uint type;
+
+    JSG_STRUCT(name, path, modifiedAt, createdAt, type);
+  };
+
+  Stat getStat();
+
+  void visitForMemoryInfo(jsg::MemoryTracker& tracker) const {
+    tracker.trackFieldWithSize("name", name.size());
+    tracker.trackFieldWithSize("path", path.size());
+  }
+
+  JSG_RESOURCE_TYPE(Node) {
+    JSG_READONLY_PROTOTYPE_PROPERTY(readable, isReadable);
+    JSG_READONLY_PROTOTYPE_PROPERTY(writable, isWritable);
+    JSG_READONLY_PROTOTYPE_PROPERTY(asyncOnly, isAsyncOnly);
+    JSG_READONLY_PROTOTYPE_PROPERTY(syncOnly, isSyncOnly);
+
+    JSG_METHOD(getStat);
+    JSG_METHOD(getFd);
+  }
+
+ private:
+  kj::String name;
+  kj::Path path;
+  kj::Date modifiedAt;
+  kj::Date createdAt;
+  NodeType type;
+};
+
+}  // namespace workerd::api::filesystem
+
+#define EW_FILESYSTEM_ISOLATE_TYPES                                                                \
+  api::filesystem::Node, api::filesystem::Node::Stat, api::filesystem::Handle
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/workerd/api/node/filesystem.h [hunk 1]
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+// Copyright Joyent and Node contributors. All rights reserved. MIT license.
+#pragma once
```
jasnell, line 5: Given that there's a possibility of us implementing the WHATWG file system api on top of this also (https://fs.spec.whatwg.org/) it might make sense to move this into `workerd/api` and to not make any part of this C++ layer Node.js specific in any way.

src/workerd/api/node/filesystem.h
```diff
@@ -0,0 +1,109 @@
...
+    virtual bool isWritable() const = 0;
+    virtual bool isAsyncOnly() const = 0;
+    virtual bool isSyncOnly() const = 0;
+
+    kj::StringPtr getName() const {
+      return name;
+    }
+    kj::StringPtr getPath() const {
+      return path.toString();
+    }
```
jasnell, line 41: just for clarification, is `getName()` the same as basename?

anonrig: Yes, that was my intention. I can change this to base name.

src/workerd/api/node/filesystem.h
```diff
@@ -0,0 +1,109 @@
...
+    jsg::JsDate getCreatedAt() const {
+      return jsg::JsDate(createdAt);
+    }
+
+    bool isDirectory() const {
+      return type == NodeType::DIRECTORY;
+    }
+    bool isFile() const {
+      return type == NodeType::FILE;
+    }
```
jasnell, line 54: Rather than separate getters for each type can we use an enum here? Or, alternatively have a `getStat()` method that returns a struct that contains the modified, created, and type metadata?

src/workerd/api/node/filesystem.h
```diff
@@ -0,0 +1,109 @@
...
+      return false;
+    }
+
+    JSG_RESOURCE_TYPE(BundleNode) {
+      JSG_INHERIT(Node);
+    }
+  };
+
+  using PathLike = kj::OneOf<kj::String, jsg::Url, jsg::BufferSource>;
+  kj::Maybe<Node> getResource(jsg::Lock&, PathLike path);
```
jasnell, line 99: s/getResource/open ?

Also, this should return `kj::Maybe<jsg::Ref<Node>>` since `Node` is a `jsg::Object`

src/workerd/api/node/filesystem.h
```diff
@@ -0,0 +1,109 @@
...
+    bool isSyncOnly() const override {
+      return false;
+    }
+
+    JSG_RESOURCE_TYPE(BundleNode) {
+      JSG_INHERIT(Node);
+    }
+  };
+
+  using PathLike = kj::OneOf<kj::String, jsg::Url, jsg::BufferSource>;
```
jasnell, line 98: For this layer, we should normalize on a single path input type. The JavaScript layer can handle normalizing multiple input types. Just helps keep things simpler.

src/workerd/api/node/filesystem.h
```diff
@@ -0,0 +1,109 @@
...
+    kj::StringPtr getName() const {
+      return name;
+    }
+    kj::StringPtr getPath() const {
+      return path.toString();
+    }
+    jsg::JsDate getModifiedAt() const {
+      return jsg::JsDate(modifiedAt);
+    }
+    jsg::JsDate getCreatedAt() const {
```
jasnell, line 45: Might be easier at this level just to provide the numeric posix timestamp and let the JS layer handle converting to `Date` objects.

src/workerd/api/node/node.h
```diff
@@ -54,7 +55,8 @@ class CompatibilityFlags: public jsg::Object {
   V(ZlibUtil, "node-internal:zlib")                                                                \
   V(UrlUtil, "node-internal:url")                                                                  \
   V(DnsUtil, "node-internal:dns")                                                                  \
-  V(TimersUtil, "node-internal:timers")
```
jasnell, line 57: For now, the new internal module should be guarded by the `--experimental` flag.

src/workerd/api/node/filesystem.h
```diff
@@ -0,0 +1,109 @@
...
+        : name(kj::mv(name)),
+          path(kj::mv(path)),
+          modifiedAt(kj::mv(modifiedAt)),
+          createdAt(kj::mv(createdAt)),
+          type(type) {};
+
+    virtual bool isReadable() const = 0;
+    virtual bool isWritable() const = 0;
+    virtual bool isAsyncOnly() const = 0;
+    virtual bool isSyncOnly() const = 0;
```
jasnell, line 34: A `Node` will need to expose the "file descriptor" ... a `getFd()` accessor is needed.

src/workerd/api/node/filesystem.h
```diff
@@ -0,0 +1,109 @@
...
+    }
+    bool isFile() const {
+      return type == NodeType::FILE;
+    }
+
+    JSG_RESOURCE_TYPE(Node) {
+      JSG_READONLY_INSTANCE_PROPERTY(readable, isReadable);
+      JSG_READONLY_INSTANCE_PROPERTY(writable, isWritable);
+      JSG_READONLY_INSTANCE_PROPERTY(asyncOnly, isAsyncOnly);
+      JSG_READONLY_INSTANCE_PROPERTY(syncOnly, isSyncOnly);
```
jasnell, line 60: Since these are virtual you can't do it like this. Instead, you'll need anything that inherits from `Node` here to have these mapped in its own `JSG_RESOURCE_TYPE` block... e.g. something like

```cpp
 public:
  virtual void doFoo() = 0;
}

class MyFoo: public Foo {
 public:
  void doFoo() override {}

  JSG_RESOURCE_TYPE(MyFoo) {
    JSG_METHOD(doFoo);    
  }
}
```

src/workerd/api/node/filesystem.h
anonrig: Yeah nice touch. Agreed.

src/workerd/api/filesystem.h
```diff
@@ -0,0 +1,73 @@
...
+  void visitForMemoryInfo(jsg::MemoryTracker& tracker) const {
+    tracker.trackFieldWithSize("name", name.size());
+    tracker.trackFieldWithSize("path", path.size());
+  }
+
+  JSG_RESOURCE_TYPE(Node) {
+    JSG_READONLY_INSTANCE_PROPERTY(readable, isReadable);
+    JSG_READONLY_INSTANCE_PROPERTY(writable, isWritable);
+    JSG_READONLY_INSTANCE_PROPERTY(asyncOnly, isAsyncOnly);
+    JSG_READONLY_INSTANCE_PROPERTY(syncOnly, isSyncOnly);
```
jasnell, line 58: Make these `...PROTOTYPE_PROPERTY` instead.

PR #3999, opened 2025-04-22
Return abort exception error through promise in IoContext::run()

ObsidianMinor:
In WorkerEntrypoint::request, we set up draining as part of the returned promise. But if an exception is thrown by IoContext::run separately from the returned promise, the IncomingRequest is dropped during unwinding without draining! So with this we just wrap context.run in evalNow to make sure any thrown exceptions are caught.


dom96 (approved):

ObsidianMinor:
> Does this really only apply to HTTP events or do all the other event types also need updating? Maybe we should bake this directly into `run()`?

It only really applies to HTTP afaik because all the other events immediately set up the drain after calling delivered. But maybe you're right maybe we should just return the exception from `run()` instead of calling `throwFatalException()`.

kentonv:
As this is the `abortException` specifically? Yeah simply changing it to a return seems like a good idea.

kentonv (approved):

PR #4000, opened 2025-04-22
Make `ctx.props` optional in workers-types

penalosa:
Make `ctx.props` optional in workers-types, and allow users to extend it by defining `Cloudflare.Props`

### DIFF
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -238,6 +238,8 @@ class ExecutionContext: public jsg::Object {
     }
     JSG_LAZY_INSTANCE_PROPERTY(props, getProps);
 
+    JSG_TS_OVERRIDE({ props?: Cloudflare.Props })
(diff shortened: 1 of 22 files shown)


danlapid (approved):

kentonv (changes requested):
Can you please explain what this is about? I don't get it.

kentonv:
This seems incorrect to me:
1. `ctx.props` is _not_ optional. It is _always_ present, but may contain an empty object.
2. The type of `ctx.props` is expected to differ for every entrypoint, so giving it a global types seems wrong. We really need to update the types to make it a type parameter.

penalosa:
> The type of ctx.props is expected to differ for every entrypoint

Ah, I'd misunderstodd that! I'd thought it was per-worker, and so thought the aproach we took for importable env typing would match. Given that, I'll close this PR

PR #4003, opened 2025-04-22
avoid leading slashes on .gitignore

anonrig:
Editors like Zed doesn't support "/" prefix for the gitignore entries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,31 +2,31 @@
 .DS_Store
 
 # directory for developers to store local resources
-/.local/
+.local/
 
-/*.code-workspace
+*.code-workspace
 
 node_modules
-/npm/*/bin
-/npm/workerd/install.js
-/npm/workerd/lib/
+./npm/*/bin
+./npm/workerd/install.js
+./npm/workerd/lib/
 package-lock.json
 
 # The external link for compile_flags.txt: Differs on Windows vs macOS/Linux, so we can't check it in. The pattern needs to not have a trailing / because it's a symlink on macOS/Linux.
-/external
+external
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,31 +2,31 @@
...
-/*.code-workspace
+*.code-workspace
 
 node_modules
-/npm/*/bin
-/npm/workerd/install.js
-/npm/workerd/lib/
+./npm/*/bin
+./npm/workerd/install.js
+./npm/workerd/lib/
```
npaun, line 12: Why do these ones get an extra ./?

npaun (approved):
Even leaving aside zed, the leading slashes were weird. This makes more sense to me

.gitignore
anonrig, line 12:
```suggestion
npm/*/bin
npm/workerd/install.js
npm/workerd/lib/
```

jasnell (approved):

.gitignore
```diff
@@ -2,31 +2,31 @@
...
-/npm/workerd/install.js
-/npm/workerd/lib/
+npm/*/bin
+npm/workerd/install.js
+npm/workerd/lib/
 package-lock.json
 
 # The external link for compile_flags.txt: Differs on Windows vs macOS/Linux, so we can't check it in. The pattern needs to not have a trailing / because it's a symlink on macOS/Linux.
-/external
+external
```
hoodmane, line 16: This changes the meaning to ignoring all folders named `external` instead of just ones in the root directory of the repo. Maybe zed should be changed to understand gitignore syntax?

anonrig: I can add "./" prefix which will limit the scope to the root folder.

hoodmane: Yeah adding this to all of them would probably be a good idea.

anonrig:
Unfortunately, git doesn't seem to support "./" prefix (relative paths)

PR #4005, opened 2025-04-22
Add createHttpClient API for Socket objects

danlapid:
This change adds a new method to the Socket API that allows for creating an HTTP client from an existing Socket connection. This enables making standard HTTP requests over an already established socket connection, providing a higher-level interface for HTTP communication compared to raw socket operations.

Key components:
1. Added a new `createHttpClient` method to SocketsModule that takes a Socket object and returns a Fetcher that can perform HTTP requests over that socket
2. Implemented a new `StreamWorkerInterface` class that handles HTTP requests by forwarding them through the socket's connection stream
3. Created a `SocketOutgoingFactory` class that generates WorkerInterface instances from the underlying socket
4. Added TypeScript definitions for the new API in the cloudflare modules
5. Added comprehensive tests for the new API

The implementation uses KJ's existing HTTP client capabilities via `kj::newHttpClient` to create an HTTP client from the socket's stream, then wraps it in a WorkerInterface implementation that can be used by a Fetcher to provide a familiar fetch API.

Usage example:
```js
import { connect, createHttpClient } from 'cloudflare:sockets';

export default {
  async fetch(req, env) {
    let socket = connect("example.com:80");
    let httpClient = createHttpClient(socket);
    return await httpClient.fetch("/path");
  }
}
```

This makes it easier to work with various HTTP servers over socket connections without having to manually implement HTTP protocol handling.

ToDo:
- Fix the TypeScript return type for createHttpClient to return Fetcher
- Make the API experimental until fully tested

===================================================================
THIS CODE HAS NOT BEEN REVIEWED BY ME WELL ENOUGH YET!!!!!!!!!
PUTTING THIS OUT THERE BUT NOT READY FOR REVIEW
Convo: https://paste.cfdata.org/UkG1700vjiwm
===================================================================


kentonv:
There is some very similar code in `container.c++` for constructing an HTTP fetcher from a raw stream connection to a container. Probably should share code?

src/cloudflare/sockets.ts
```diff
@@ -10,3 +10,11 @@ export function connect(
 ): sockets.Socket {
   return sockets.connect(address, options);
 }
+
+interface Fetcher {
+  fetch: typeof fetch;
+}
+
+export function createHttpClient(socket: sockets.Socket): Fetcher {
+  return sockets.createHttpClient(socket);
```
jasnell, line 19: I can imagine that this might get a bit confusing *if* we ever do decide to implement `node:http`. I wonder if an alternative name like `getFetcher(...)` might be clearer as a way to indicate that the returned client implements the `fetch(...)` api.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +511,93 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+ public:
+  StreamWorkerInterface(kj::Own<kj::AsyncIoStream> stream, const kj::HttpHeaderTable& headerTable)
+      : stream(kj::mv(stream)),
+        headerTable(headerTable) {}
+
+  kj::Promise<void> request(kj::HttpMethod method,
+      kj::StringPtr url,
+      const kj::HttpHeaders& headers,
+      kj::AsyncInputStream& requestBody,
+      kj::HttpService::Response& response) override {
```
jasnell, line 526: This will need to pay attention to whether the request url uses `https://` vs. `http://` and whether the `Socket` was opened with TLS support or not.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +511,93 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+        .attach(socket.addRef());
+  }
+
+ private:
+  jsg::Ref<Socket> socket;
+};
+
+jsg::Ref<Fetcher> SocketsModule::createHttpClient(jsg::Lock& js, jsg::Ref<Socket> socket) {
+  // Create our custom factory that will create client instances from this socket
+  auto outgoingFactory = kj::heap<SocketOutgoingFactory>(socket.addRef());
```
jasnell, line 597: This should check to make sure the `socket`'s `readable` or `writable` are not locked, and possibly not disturbed. It especially needs to check to ensure the socket is not closed, and probably needs to wait for the opened promise to be resolved.

Locked indicates that there is `Reader` or `Writer` associated, in which case wrapping the socket should fail. Disturbed indicates that the socket has been written to or read from, indicating that it may not be in the correct state to be wrapped correctly.

jasnell: We should probably also arrange it such that the returned `Fetcher` completely takes ownership over the `Socket` such that the `readable` and `writable` streams get closed/detached or locked so that nothing on the JS side mistakenly tries to access them while the socket it bound... e.g.

```js
const socket = connect('...');
const fetcher = getHttpClient(socket);
socket.readable.getReader();  // should fail!
```

jasnell: As an additional bit of fun, since `fetch(...)` is cancelable using an AbortSignal, the tests need to verify that canceling a fetch via the returned Fetcher works as expected here. 

Also, is the `Socket` expected to support multiple http requests or should it be closed when the one http request is finished? Would it be expected to support pipelining (I'd expect no). This would need to ensure that we don't end up with multiple concurrent http requests going over the socket at a time... e.g. 

```js
const socket = connect('...');
const fetcher = getHttpClient(socket);
await Promise.all([ fetcher.fetch('...'), fetcher.fetch('...') ]);
```

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,113 @@
...
+
+import { connect, createHttpClient } from 'cloudflare:sockets';
+import { strict as assert } from 'node:assert';
+
+export default {
+  async test(ctrl, env) {
+    // Basic connectivity and GET test
+    {
+      const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+      const httpClient = createHttpClient(socket);
```
jasnell, line 13: What happens if I do `socket.readable.getReader()` after this point....

src/workerd/api/sockets.c++
danlapid: This is meant as an internal API specifically for a project yet to be announced, if we ever release it as a public API it might need to take different considerations but:
For this usecase, we do not intend to ever negotiate TLS at all, we just assume that TLS was already negotiated on the socket.

src/cloudflare/sockets.ts
danlapid: Sure, I don't mind the naming, the getFetcher name seemed overloaded but it has also crossed my mind.
I don't really think it would be confusing if we decide to implement `node:http` though as this is intended to stay as an experimental only API in the near future.

src/workerd/api/sockets.c++
danlapid: A few comments here.
Regarding the state of socket when receiving control and after. I left it a bit vague because this is intended to be an internal API only used by one worker really, I don't mind amending the implementation to make it a bit more safe in this regard though.

Socket is expected to support multiple http requests and there's even a test case for it.

src/workerd/api/tests/http-socket-test.js
danlapid: Hell breaks loose.

jasnell: `ETOOMUCHCHAOS`

danlapid:
Tried to share code with `container.c++` it actually has a few small intricacies that make it quite hard to share code in a way that makes sense. I think this version is good enough for now. Opening up for review.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +511,106 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+// Implementation of the custom factory for creating WorkerInterface instances from a socket
+class StreamOutgoingFactory final: public Fetcher::OutgoingFactory {
+ public:
+  StreamOutgoingFactory(kj::Own<kj::AsyncIoStream> stream, const kj::HttpHeaderTable& headerTable)
+      : stream(kj::mv(stream)),
+        headerTable(headerTable) {}
+
+  kj::Own<WorkerInterface> newSingleUseClient(kj::Maybe<kj::String> cfStr) override {
+    JSG_ASSERT(stream.get() != nullptr, Error,
+        "Fetcher created from convertSocketToFetcher can only be used once");
```
xortive, line 595: i'm rusty on how this interface works, but does this allow multiple requests on one stream?

the test cases don't make it clear, there is an assert for this error that would seem to indicate the fetcher is only usable once, but there's no `finally` clause that fails the test if the fetcher works for multiple requests

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,247 @@
...
+      assert.equal(response1.status, 200);
+      const text1 = await response1.text();
+      assert.equal(text1, 'pong');
+      try {
+        // Second request on same connection
+        const response2 = await httpClient.fetch('/json');
+        assert.equal(response2.status, 200);
+        const data = await response2.json();
+        assert.deepEqual(data, { message: 'Hello from HTTP socket server' });
+      } catch (error) {
```
xortive, line 103: I don't understand this try/catch -- we want to the fetcher to work for multiple requests

danlapid: The fetcher sadly can't really work on multiple requests as we can't be absolutely sure whether the previous request completed or are we multiplexing request which obviously isn't supported in http1

I don't see why your interface requires it.

You will be exposing a fetch interface to your clients

Your code for fetch would essentially be
fetch(request) {
  Let socket = await apollo.connect()
  Let fetcher = convertSocketToHttpClient(socket)
  Return await fetcher.fetch(request)
}

Meaning each connection is only used for a single request as expected

src/workerd/api/sockets.c++
danlapid: Nope, it is explicitly not supported

src/workerd/api/tests/http-socket-test.js
xortive: > The fetcher sadly can't really work on multiple requests as we can't be absolutely sure whether the previous request completed or are we multiplexing request which obviously isn't supported in http1

KJ's HTTP implementation should handle managing the state of the connection I would think? Connection keep-alive is a core part of the protocol...

> I don't see why your interface requires it.

Your code for fetch would essentially be
```
fetch(request) {
    let socket = await apollo.connect()
    let fetcher = convertSocketToHttpClient(socket)
    return await fetcher.fetch(request)
}
```

The worker isn't always going to act as a proxy, it might make multiple requests to a single endpoint, having to re-create the connection every time would be unfortunate.

I think we might need to expose a "fetcher capability" similar to a durable object stub, that can be re-used for multiple requests?

xortive: If this adds a significant amount of implementation complexity, I'm fine merging this as-is, and we can deal with it later, but if KJ handles the difficult part of managing the state of the underlying connection, I don't see why we can't make the fetcher multi-use?

danlapid: The above does enable fetch to be called multiple times though
It's just that every fetch would be a new connection

xortive: > The above does enable fetch to be called multiple times though
It's just that every fetch would be a new connection

I'd like to take a shot at supporting connection reuse in this PR. If not, we can live with new connection per request for prototyping by having the binding implement support for multiple requests by calling `createHttpClient` every time like you mentioned.

danlapid:
@dom96 @anonrig @jasnell reviews appreciated

src/workerd/api/tests/http-socket-test.js
kentonv: I agree with @xortive that we should support connection reuse. It should be no problem to support in this code as KJ HTTP already handles it.

Yes, there is the caveat that responses will be FIFO and if any response stream breaks, all later responses are also broken. We'll need to advise app developers that it's best to send only one request at a time, and if the fetch() or response stream read throws an exception, don't reuse the connection. But IMO connection reuse is too useful to simply disallow for fear of these footguns.

danlapid: @kentonv any way I tried to implement this either allows the user to create deadlocks or to throw KJ exceptions that will end up in sentry.
If I just try to reuse the connection without kj::mving it then there is a kj exception when the client does 2 concurrent requests
e.g
```js
        const responses = await Promise.all([
          httpClient.fetch('/ping'),
          httpClient.fetch('/json'),
        ]);
```
If I try to use the stream to create an HttpClient and then create an HttpService out of that and then pass that into the WorkerInterface then the above snippet just deadlocks and never returns.

Both of these seem like really bad possibilities.
If I'm doing something obviously wrong please refer me to the intended way to implement this.

xortive: FWIW, there's more factors at here for optimization than just connection reuse for our usecase. The service we will communicate with using this binding supports FD passing after connection establishment, and for the eventual product release we will likely need to implement support for this in our binding implementation.

I suspect building a JS API to allow receiving FDs from an `InternalHTTPService` may be more complex than just writing a C++ implementation for our binding when we want to move out of the prototyping phase.

With the above context in mind, I'd be perfectly content landing this as-is without connection reuse and marked as experimental, so we can unblock prototyping, considering we may end up moving to a C++ binding implementation later on regardless...

src/workerd/api/sockets.c++
```diff
@@ -511,4 +511,106 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+    return kj::heap<StreamWorkerInterface>(kj::mv(stream), headerTable);
+  }
+
+ private:
+  kj::Own<kj::AsyncIoStream> stream;
+  const kj::HttpHeaderTable& headerTable;
+};
+
+jsg::Ref<Fetcher> SocketsModule::convertSocketToFetcher(jsg::Lock& js, jsg::Ref<Socket> socket) {
+  auto& ioctx = IoContext::current();
```
jasnell, line 606: this really ought to check to see if either the readable or writable streams for the socket are locked or disturbed before proceeding. If either side is locked it means there are outstanding readers/writers and if it is disturbed it means the streams have already been partially used and may not be in a good state to be taken.

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,247 @@
...
+      const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+      const httpClient = convertSocketToFetcher(socket);
+
+      // First request
+      const response1 = await httpClient.fetch('/ping');
+      assert.equal(response1.status, 200);
+      const text1 = await response1.text();
+      assert.equal(text1, 'pong');
+      try {
+        // Second request on same connection
```
jasnell, line 98: This test is confusing. Are you expecting this to throw or succeed? If you are expecting a throw, using the `node:assert` `throws(...)` API would be a better way to construct this block

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,247 @@
...
+          'Expected the error message to contain "Fetcher created from convertSocketToFetcher can only be used once"'
+        );
+      }
+    }
+
+    // Test multiple concurrent requests on the same connection
+    {
+      const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+      const httpClient = convertSocketToFetcher(socket);
+      try {
```
jasnell, line 118: Use `assert.throws(...)`

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,247 @@
...
+      const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+      const httpClient = convertSocketToFetcher(socket);
+      // Now close the socket
+      await socket.close();
+
+      // Try to send a request with the HTTP client after closing the socket
+      // This should still work since the HTTP client should have its own stream
+      const response1 = await httpClient.fetch('/ping');
+      assert.equal(response1.status, 200);
+    }
```
jasnell, line 194: Additional test cases here should include:

1. What happens if the socket writable/readable are already locked to a `Writer` or `Reader`
2. What happens if the socket has already been partially consumed / written to
3. What happens if the socket has been errored
4. What happens if the socket is configured to potentially use tls upgrade (internally we use a packet protocol to communicate with the proxy layer when tls upgrade is expected...)
5. What happens if the remote end drops the connection prematurely

src/workerd/api/tests/http-socket-test.js
kentonv: @danlapid 

> If I just try to reuse the connection without kj::mving it then there is a kj exception when the client does 2 concurrent requests

What is the exception?

src/workerd/api/sockets.c++
dom96: This is calling `takeConnectionStream` which calls `detach` on the readable/writable streams which should fail when the stream is locked/disturbed. This is probably what we want to happen here, right?

src/workerd/api/tests/http-socket-test.js
dom96: Another test that may be worthwhile: what happens when a full URL is specified, it should fail.

Also can the Fetcher/kj::HttpClient follow redirects? Might be worth checking what happens when the HTTP server returns a 301.

dom96 (commented):
Agree that connection reuse would be nice, left a few other comments. Overall looks good to me, but I'll let others stamp.

src/workerd/api/tests/http-socket-test.js
kentonv: > Another test that may be worthwhile: what happens when a full URL is specified, it should fail.

Wait, what? `fetch()` always requires a full URL. I see as currently implemented this is accepting paths without a hostname or scheme. We should fix that. The hostname should be required and used to populate the `Host` header.

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,247 @@
...
+import { strict as assert } from 'node:assert';
+
+export default {
+  async test(ctrl, env) {
+    // Basic connectivity and GET test
+    {
+      const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+      const httpClient = convertSocketToFetcher(socket);
+
+      const response = await httpClient.fetch('/ping');
```
kentonv, line 15: Why does this accept a path with no hostname? How does this fill in the `Host` header, which is required under HTTP/1.1?

src/workerd/api/sockets.c++
danlapid: Yep

src/workerd/api/tests/http-socket-test.js
xortive: I discussed with Kenton and the product does not require no-host on the fetcher -- we should accept a full URL for consistency, such that the host component of the URL is used *only* to provide a value for the `Host` header. The provided `URL` is *not* used for routing.

tewaro: I'm taking over getting this PR merged. Just to clarify should we expect the fetch to look like:
```
httpClient.fetch (`localhost:${env.HTTP_SOCKET_SERVER_PORT}`'/ping');
```

Then the `Host` header should be added within the `newSingleUseClient` for `StreamOutgoingFactory`?

danlapid: No, Kenton just means that
`httpClient.fetch('/ping')` shouldn't work, instead the following should be required
`httpClient.fetch('https://example.com/ping')`

danlapid: Essentially changing the NO in https://github.com/cloudflare/workerd/pull/4005/files#diff-4e2699e03d914d2336177c99e165ed8e7f912f9bb61af226be284e62c3324f6eR613 to YES

kentonv: Right. The hostname in the URL can be anything, it'll just be shoved into the `Host` header, it won't be used for routing because we already have a socket. `Fetcher::RequiresHostAndProtocol::YES` should be all that's needed.

src/workerd/api/tests/http-socket-test.js
tewaro: Exception: 
```
workerd/jsg/util.c++:320: error: e = kj/async-io.c++:925: failed: can't read() again until previous read() completes
stack: src/workerd/server/workerd@9720a34 src/workerd/server/workerd@971e50f src/workerd/server/workerd@9748e9b src/workerd/server/workerd@971284b src/workerd/server/workerd@4f525f0 src/workerd/server/workerd@5015060 src/workerd/server/workerd@5022a62 src/workerd/server/workerd@4fc4460 src/workerd/server/workerd@5006970 src/workerd/server/workerd@2e5d7c0 src/workerd/server/workerd@4380f80 src/workerd/server/workerd@41762a0
kj::(anonymous namespace)::AsyncPipe::BlockedRead::tryRead(void*, unsigned long, unsigned long)
async-io.c++:0:0
```

tewaro: This is based on the commit titled: `AsyncIoStream reuse attempt 1`.

tewaro: @kentonv is this exception acceptable, or should we try to do something else?

src/workerd/api/tests/http-socket-server.js
```diff
@@ -0,0 +1,220 @@
+// Copyright (c) 2017-2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+// This file is used as a sidecar for the http-socket tests.
+// It creates an HTTP server that will respond to requests from the convertSocketToFetcher API.
+const http = require('http');
+const crypto = require('crypto');
+
```
anonrig:
```suggestion
const http = require('node:http');
const crypto = require('node:crypto');

```

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) 2017-2024 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+import { connect, convertSocketToFetcher } from 'cloudflare:sockets';
+import { strict as assert } from 'node:assert';
+
+export default {
+  async test(ctrl, env) {
```
anonrig: Can we have multiple test exports rather than a single one? This will make debugging errors a lot easier.

src/workerd/api/tests/http-socket-server.js
```diff
@@ -0,0 +1,220 @@
...
+// Handle upgrade requests to switch from HTTP to WebSocket protocol
+function upgradeToWebSocketConnection(req, socket, head) {
+  // Check if it's a WebSocket upgrade request
+  if (req.headers['upgrade'] !== 'websocket') {
+    socket.end('HTTP/1.1 400 Bad Request');
+    return;
+  }
+
+  console.log('WebSocket upgrade request received');
+  console.log('WebSocket headers:', req.headers);
```
anonrig: Can we remove the console.logs in this file?

src/workerd/api/tests/http-socket-test.js
kentonv: This exception is pretty bad -- it implies two things are trying to read from the stream at the same time, which could lead to corruption. In fact, overlapping calls to read() on a KJ stream is undefined behavior; this particular implementation is throwing an exception, but only as a courtesy.

Normally, though, KJ HTTP should automatically enforce queuing of requests and should not allow multiple overlapping reads from the underlying  stream. Something is going wrong in this case. I guess I'd have to study it to understand what.

danlapid: Can we settle on disallowing fetcher reuse to unblock Workers VPC and revisit this if you can figure out what's going wrong?

tewaro: After adding a redirect test as @dom96 mentioned I think this causes a problem if we don't allow multiple fetches, essentially when trying to follow the redirect another fetch is made on the same socket. This definitely limits server behavior but not sure if that matters @xortive do we need redirects?

src/workerd/api/sockets.c++
```diff
@@ -511,4 +512,106 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+    return service->request(method, noHostUrl, newHeaders, requestBody, response)
+        .attach(kj::mv(service), kj::mv(httpClient));
+  }
+
+  kj::Promise<void> connect(kj::StringPtr host,
+      const kj::HttpHeaders& headers,
+      kj::AsyncIoStream& connection,
+      ConnectResponse& response,
+      kj::HttpConnectSettings settings) override {
+    KJ_UNIMPLEMENTED("connect() not supported on StreamWorkerInterface");
```
kentonv: This is actually reachable if someone tries to call `connect()` on the resulting fetcher, so this should probably be a JSG_FAIL_REQUIRE with a friendly error message, rather than a Sentry error as it is now.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +512,106 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+        headerTable(headerTable) {}
+
+  kj::Promise<void> request(kj::HttpMethod method,
+      kj::StringPtr url,
+      const kj::HttpHeaders& headers,
+      kj::AsyncInputStream& requestBody,
+      kj::HttpService::Response& response) override {
+    // Parse the URL to extract the path
+    auto parsedUrl = KJ_REQUIRE_NONNULL(
+        kj::Url::tryParse(url, kj::Url::Context::HTTP_PROXY_REQUEST), "invalid url", url);
```
kentonv: You'll want to pass the options: `{.percentDecode = false, .allowEmpty = true}`

Otherwise this parse will potentially modify people's URLs.

You should also check that the scheme is `http`. We should reject `https` right now as we aren't prepared to negotiate TLS here.

Check out the very similar code in `container.c++`.

(Hmm, although, it's possible the underlying connection is already doing HTTPS. Should we permit, or even require, `https` URLs in that case? It sort of pierces the abstraction -- we should be treating the Socket as just an abstract byte stream, and not thinking about what encryption it may be applying...)

src/cloudflare/sockets.ts
```diff
@@ -10,3 +10,11 @@ export function connect(
 ): sockets.Socket {
   return sockets.connect(address, options);
 }
+
+interface Fetcher {
```
kentonv: Isn't the `Fetcher` type already declared elsewhere?

src/workerd/api/tests/http-socket-test.js
kentonv: Hmm, it's pretty unclear how automatic redirect following should work at all here. What if the redirect is to a different hostname? Should it really go to the same socket? It sort of depends on the use case.

Maybe we should require people to set `redirect: "manual"` for now. Although that's pretty tedious, as unfortunately "automatic" is the default.

src/workerd/api/sockets.c++
xortive: > You should also check that the scheme is http. We should reject https right now as we aren't prepared to negotiate TLS here.

Why is that? We do want to support http + https for requests made using a fetcher from `createHttpClient`.

> (Hmm, although, it's possible the underlying connection is already doing HTTPS. Should we permit, or even require, https URLs in that case? It sort of pierces the abstraction -- we should be treating the Socket as just an abstract byte stream, and not thinking about what encryption it may be applying...)

It's possible for the service implementing the underlying socket to negotiate TLS, though I had intended for this to be optional (for things like mTLS where we'd want to keep the client's private key out of the sandbox).

kentonv: My point is: If the socket _isn't_ using TLS, and the application tries to fetch at HTTPS URL, what do we do? The current code will just pass the request through without any encryption. This seems insecure.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +512,106 @@ jsg::Ref<Socket> SocketsModule::connect(
     jsg::Lock& js, AnySocketAddress address, jsg::Optional<SocketOptions> options) {
   return connectImpl(js, kj::none, kj::mv(address), kj::mv(options));
 }
+
+kj::Own<kj::AsyncIoStream> Socket::takeConnectionStream(jsg::Lock& js) {
+  writable->detach(js);
+  readable->detach(js);
+  // We should set this before closedResolver.resolve() in order to give the user
+  // the option to check if the closed promise is resolved due to upgrade or not.
+  upgraded = true;
```
xortive: I think this is copypasta from the code inside of Socket::startTls... I don't think we mean to set this here.

src/workerd/api/sockets.c++
xortive: > My point is: If the socket isn't using TLS, and the application tries to fetch at HTTPS URL, what do we do? The current code will just pass the request through without any encryption. This seems insecure.

Yeah, that's definitely bad.

What I'm getting at here is that I was under the impression that KJ's HTTP client could negotiate a TLS upgrade for us, and we would only have to handle establishing a connection in our service that provides the underlying socket and not do any TLS negotiation on behalf of Edgeworker.

We definitely need to support http and https schemes from the user's perspective, even if `convertSocketToFetcher` is http only, i.e same behavior as global `fetch()`. We need to figure out if it makes sense to do TLS negotiation inside of EW, or inside the service that provides the underlying socket, which I believe will require some Oxy framework changes to support TLS negotiation on egress when the incoming connection wasn't already using TLS.

danlapid: EW will not do TLS negotiation in prod.
Oxy already knows to upgrade an existing connection to TLS.
That's how TLS in our sockets API works today with OER.
That's what WVPC should use as well.

kentonv: > What I'm getting at here is that I was under the impression that KJ's HTTP client could negotiate a TLS upgrade for us, and we would only have to handle establishing a connection in our service that provides the underlying socket and not do any TLS negotiation on behalf of Edgeworker.

Regardless of whether we could do it, I suspect you don't actually want this.

Consider the use case where the socket points at your origin server, and you want to receive eyeball HTTP requests in your Worker and just forward them verbatim on to the origin. Do you want your origin server to have to present a valid TLS certificate matching whatever hostname the HTTP request is for? Probably not! You _do_ want to be sure that the origin server is the one the binding was designed to talk to, but as long as it is, then you do _not_ care whether it has a certificate matching the HTTP request URL, because you are only forwarding requests to it that you already know you want to go to that specific server.

Also if the underlying transport is already encrypted separately (because it's going over a VPN or Argo Tunnel or whatever), then adding a TLS layer on top is redundant and just wasting cycles and latency. Consider especially that the key exchange will require a round trip all the way from the Worker to the origin, even if we're travelling over many shorter hops that are individually already secured.

For these reasons we try to push any encryption concerns into the binding itself, so that bindings have flexibility to implement the best approach for the use case.

So when creating a Fetcher from a Socket, we'd really like to assume that whoever set up the Socket has already arranged whatever transport security they want. We just want to speak the HTTP protocol over it.

If only it were possible to call `fetch()` without specifying http vs. https, then there wouldn't be an issue. Unfortunately, that's part of the API. From a technical perspective, it would make the most sense to just ignore it, and use plaintext HTTP, saying "it's the app's responsibility to provide a Socket that represents a secure transport if that's what they want". But this would inevitably lead to mistakes where people _assume_ that using at HTTPS URL will cause TLS to happen, and then they'll think they are secure when they are not.

So all we really want, I think, is for the socket/binding to have some sort of "allowHttps" flag. For TCP sockets created by global connect(), this flag becomes true if the socket used TLS. For sockets that come from a binding, it's up to the binding to decide whether it should allow HTTPS -- but in practice I suspect essentially every binding will use some form of secure transport under the hood, and therefore should allow it.

xortive: We're going to do this in Oxy.

> So all we really want, I think, is for the socket/binding to have some sort of "allowHttps" flag. For TCP sockets created by global connect(), this flag becomes true if the socket used TLS. For sockets that come from a binding, it's up to the binding to decide whether it should allow HTTPS -- but in practice I suspect essentially every binding will use some form of secure transport under the hood, and therefore should allow it.

As an argument on `createFetcherFromSocket`? This makes sense to me, fyi @tewaro

src/cloudflare/sockets.ts
xortive: Yeah, I wouldn't think we need a new one?

src/workerd/api/tests/http-socket-test.js
xortive: After discussing this with @kentonv offline, I think the approach we want to take is to support redirects:

- Requiring a non-default option like `redirect: "manual"` is pretty weird
- If the redirect is to a different path on the same host, it should work fine
- If the redirect is to a different host, the request will likely 404. The solution to this problem is a "network" binding where multiple hosts are exposed in a single capability. This won't be needed for the WVPC MVP.

src/workerd/api/sockets.c++
kentonv: Well, I was thinking the Socket itself would have some sort of property indicating if it's already secure.

TCP sockets on which TLS has been initiated would have this flag set. Binding implementations would need some way to declare themselves secure. (All worker-to-worker service bindings would be presumed secure by default since they don't leave Cloudflare's network.)

src/cloudflare/sockets.ts
```diff
@@ -10,3 +10,11 @@ export function connect(
...
+  fetch: typeof fetch;
+}
+
+export function convertSocketToFetcher(socket: sockets.Socket): Fetcher {
```
kentonv: This feels non-idiomatic as a public API.

Something like `new Fetcher(socket)` would be idiomatic -- except I don't think we actually expose `Fetcher` as a concrete JS class anywhere, and if we decided to expose it, maybe we want to bikeshed the name first...

If we keep this as a freestanding function, maybe it could be called `newHttpClient()`.

src/cloudflare/sockets.ts
tewaro: It's constantly redefined everywhere, examples:
- internal/ai-api.ts
- internal/autorag-api.ts
- internal/d1-api.ts
- internal/vectorize-api.ts
- etc...
Should I be looking to clean that up at another time?

src/workerd/api/sockets.c++
tewaro: I'm a little confused what the concrete ask is here. Do we just want a property that is readable on the fetcher, that reflects if the socket was tls, is there something else?

src/workerd/api/tests/http-socket-test.js
tewaro: I plan to address redirection in the next PR, there are a few approaches to enable nonconcurrent requests on the socket.

src/cloudflare/internal/sockets.d.ts
```diff
@@ -30,3 +32,5 @@ export function connect(
   address: string | SocketAddress,
   options?: SocketOptions
 ): Socket;
+
+export function newHttpClient(socket: Socket): Promise<Fetcher>;
```
jasnell: We likely want to stress in the documentation that `newHttpClient` is not a part of the "standard" `Socket` API that we're proposing in WinterTC.

danlapid: There's not going to be any documentation, this is an internal only experimental feature.

src/workerd/api/sockets.c++
jasnell: Given some of the requirements for `node:http`, we might actually have another reason to add a property to `Socket` that indicates if it is secure. That's entirely secondary tho, for now, as that is based on `fetch` right now... but I can foresee a future where it can be needed.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +513,117 @@ jsg::Ref<Socket> SocketsModule::connect(
     jsg::Lock& js, AnySocketAddress address, jsg::Optional<SocketOptions> options) {
   return connectImpl(js, kj::none, kj::mv(address), kj::mv(options));
 }
+
+kj::Own<kj::AsyncIoStream> Socket::takeConnectionStream(jsg::Lock& js) {
+  writable->detach(js);
+  readable->detach(js);
```
jasnell, line 519: Both of these require that there aren't pending reads or writes in flight and will throw if there are. Just want to make sure that's accounted for here. When a socket is in the middle of a starttls upgrade, for instance, we make sure things are drained before proceeding with the upgrade and detaching the original streams. This is to ensure that the streams are in a good state to actually be detached.

This likely also needs to account for whether a starttls operation is underway, the socket is already closed, etc... basically just some additional guards around whether the socket is in a good state overall to have the underlying connection removed.

src/cloudflare/internal/sockets.d.ts
dom96: Since it's internal-only, should we give it a name that emphasises this? `internalNewHttpClient` or something?

src/workerd/api/sockets.c++
tewaro: I can add a test that checks what happens when starttls is called, and then the conversion.

src/workerd/api/sockets.c++
dom96: We already seem to have both an `upgraded` and a `secureTransport` property on the socket. So you might not need to add anything new here.

src/cloudflare/internal/sockets.d.ts
danlapid: That's what the experimental guard is for.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +513,117 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+        // Create our custom factory that will create client instances from this socket
+        kj::Own<Fetcher::OutgoingFactory> outgoingFactory = kj::heap<StreamOutgoingFactory>(
+            socket->takeConnectionStream(js), ioctx.getHeaderTable());
+
+        // Create a Fetcher that uses our custom factory
+        auto fetcher = js.alloc<Fetcher>(
+            ioctx.addObject(kj::mv(outgoingFactory)), Fetcher::RequiresHostAndProtocol::YES);
+
+        return kj::mv(fetcher);
+      }));
```
tewaro, line 628: @dom96 I think there is something wrong with flushing writes, can you take a look.

dom96: Looks like you figured out the issue since the CI is passing now.

This code seems fine, but two things:

* Should the flushing logic actually live in `takeConnectionStream`?
* I'm not very familiar with how JSG_VISITABLE_LAMBDA works, maybe @jasnell can comment on whether this usage of it is correct. In particular I wonder if this should capture `JSG_THIS` in the same way that the `JSG_VISITABLE_LAMBDA` captures it in `startTls`.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +513,118 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+
+// Implementation of the custom factory for creating WorkerInterface instances from a socket
+class StreamOutgoingFactory final: public Fetcher::OutgoingFactory {
+ public:
+  StreamOutgoingFactory(kj::Own<kj::AsyncIoStream> stream, const kj::HttpHeaderTable& headerTable)
+      : stream(kj::mv(stream)),
+        headerTable(headerTable) {}
+
+  kj::Own<WorkerInterface> newSingleUseClient(kj::Maybe<kj::String> cfStr) override {
+    JSG_ASSERT(stream.get() != nullptr, Error,
```
xortive, line 596: Will we be able to remove this restriction to support following redirects?

src/workerd/api/sockets.c++
jasnell: Let's also add a comment in here so folks coming into the code later can have a heads up about the requirements.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +513,118 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+    // Parse the URL to extract the path
+    auto parsedUrl = KJ_REQUIRE_NONNULL(kj::Url::tryParse(url, kj::Url::Context::HTTP_PROXY_REQUEST,
+                                            {.percentDecode = false, .allowEmpty = true}),
+        "invalid url", url);
+
+    // We need to convert the URL from proxy format (full URL in request line) to host format
+    // (path in request line, hostname in Host header).
+    auto newHeaders = headers.cloneShallow();
+    newHeaders.set(kj::HttpHeaderId::HOST, parsedUrl.host);
+    auto noHostUrl = parsedUrl.toString(kj::Url::Context::HTTP_REQUEST);
```
jasnell, line 546: Doesn't need to be changed in this PR, but for newer cases like this I'd really actually prefer if we could start relying instead on ada_url for standards-compliant parsing of URLs rather than using kj's known-to-be-not-spec-compliant parser.

src/workerd/api/sockets.c++
```diff
@@ -511,4 +513,118 @@ jsg::Ref<Socket> SocketsModule::connect(
...
+        .attach(kj::mv(service), kj::mv(httpClient));
+  }
+
+  kj::Promise<void> connect(kj::StringPtr host,
+      const kj::HttpHeaders& headers,
+      kj::AsyncIoStream& connection,
+      ConnectResponse& response,
+      kj::HttpConnectSettings settings) override {
+    JSG_FAIL_REQUIRE(TypeError,
+        "connect is not something that can be done on a fetcher converted from a socket");
```
jasnell, line 564: At least in theory, it should be possible tho... we might want to consider this later.

src/workerd/api/sockets.c++
jasnell: Given that the lambda is not using `this` at all I don't think it needs to capture `JSG_THIS`. This usage looks correct to me.

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,501 @@
...
+    assert.equal(text1, 'pong');
+
+    // Second request on same connection
+    // TODO(cleanup) we want this to fail during this initial implementation but later we should
+    // support mutiple fetches (perhaps behind a compat flag)
+    await assert.rejects(httpClient.fetch('https://example.com/json'), {
+      name: 'Error',
+      message:
+        'Fetcher created from internalNewHttpClient can only be used once',
+    });
```
jasnell, line 130: Hmm... yeah, this is a bit weird. Generally it should be possible to use this fetch multiple times but each call to fetch should create it's own one-time-use `kj::HttpClient` under the covers, just reusing the same stream. Then, of course, it should forbid allowing multiple fetches to be in flight at the same time.

Essentially a pattern like:

* Calling `internalNewHttpClient` returns a fetcher that takes ownership of the underlying connection stream.. but...
* Instead of creating an http client immediately, each call to fetch:
  * Checks to see if a client currently exists, if so, error
  * If not, create a new client over the stream, perform the fetch, release the stream when done to free it up for a new fetch (unless it errored, at which point the socket is errored)

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,501 @@
...
+export const multipleConcurrentRequests = {
+  async test(ctrl, env, ctx) {
+    const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+    const httpClient = await internalNewHttpClient(socket);
+
+    // TODO(cleanup) when multiple fetches are enabled make sure this message is changed
+    await assert.rejects(
+      Promise.all([
+        httpClient.fetch('https://example.com/ping'),
+        httpClient.fetch('https://example.com/json'),
```
jasnell, line 148: Once we update it to allow the fetcher to be reused, this would still be a failure if the fetches are expected to be active at the same time. If the fetches are queued then this should work so long as the second fetch is not actually initiated until the second completes.

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,501 @@
...
+  },
+};
+
+// Test that closing the socket does not affect the HTTP client
+export const socketCloseThenFetch = {
+  async test(ctrl, env, ctx) {
+    const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+    const httpClient = await internalNewHttpClient(socket);
+    // Now close the socket
+    await socket.close();
```
jasnell, line 198: Hmm, in this case I would expect the `httpClient` to have complete ownership over the socket such that `socket.close()` should be a non-op. If every other way the socket is unusable so this is a bit strange.

jasnell: Oh, wait... it helps if I read the entire test case, lol. The httpclient does take over ownership and the `socket.close()` here is actually a non-op? I'd almost prefer that it errors.

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,501 @@
...
+};
+
+// Test that queueing up writes before creating the stream throws an error.
+export const writeThenConvertFlushes = {
+  async test(ctrl, env, ctx) {
+    const socket = connect(`localhost:${env.FLUSH_HELLO_SOCKET}`);
+    const writer = socket.writable.getWriter();
+    const encoder = new TextEncoder();
+    await writer.write(encoder.encode('Hello'));
+    writer.releaseLock();
```
jasnell, line 239: you're not queuing up the write here, you're allowing it to finish completely before upgrading. Is the comment at the top incorrect or is the test incorrect?

src/workerd/api/sockets.c++
tewaro: Yes that's the goal of the next piece of work.

src/workerd/api/tests/http-socket-test.js
```diff
@@ -0,0 +1,501 @@
...
+
+import { connect, internalNewHttpClient } from 'cloudflare:sockets';
+import { strict as assert } from 'node:assert';
+import { setTimeout as sleep } from 'node:timers/promises';
+
+// Basic connectivity and GET test
+export const oneRequest = {
+  async test(ctrl, env, ctx) {
+    const socket = connect(`localhost:${env.HTTP_SOCKET_SERVER_PORT}`);
+    const httpClient = await internalNewHttpClient(socket);
```
jasnell, line 13: I may have missed it but there needs to be a test that checks if the socket cannot be established (e.g. connection refused) the `await internalNewHttpclient(...)` fails.

tewaro: Will addresss later.

src/workerd/api/tests/http-socket-test.js
tewaro: Will address in next PR.

src/workerd/api/tests/http-socket-test.js
tewaro: will discuss before seeing if we need to address this.

src/workerd/api/tests/http-socket-test.js
tewaro: I can try this for redirects...

PR #4006, opened 2025-04-22
remove prettierc bazel setting

anonrig:
It is not used at all.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -91,10 +91,3 @@ selects.config_setting_group(
         ":not_dbg_build",
     ],
 )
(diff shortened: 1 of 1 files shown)


fhanau (approved):
You're right in that the target here is unused. I assume the associated file is being used in some capacity?

anonrig:
> You're right in that the target here is unused. I assume the associated file is being used in some capacity?

I forgot to remove this while I was refactoring the eslint test files where it used "eslint-prettier-config" package previously. Right now, we depend on format.py file to check the prettier correctness

PR #4008, opened 2025-04-22
add support for v8 fast api

anonrig:
Adds v8 fast api support to type wrapper and JSG.

TODO:

- [x] Find a way to test this
- [x] Add benchmarks
- [x] Support v8:Local<v8::Value> and v8::Local<v8::Object> as parameters
- [x] Support throwing error
- [x] Support const functions
- [x] Add support for jsg::Lock& in first argument
- [x] Add support for custom JSG structs and objects as parameters
- [x] Add support for all kj types
- [x] Add support for optional types
- [x] Automatically use fast APIs for method signatures that support it.
- [x] Call registerFastMethod() from JSG_STATIC_METHOD, and other possible macros.

BONUS:
- [ ] Add both FastOneByteString and v8::Local<v8::Value> function overload for kj::String (FastOneByteString is faster)
- [ ] Add a cfunction overload if the last parameter of the function is optional.
- [ ] Add function overloads, if there are multiple trailing optional parameters.

---

Used AI here and there to solve certain errors. Please review with caution.

### DIFF
--- a/src/workerd/jsg/README.md
+++ b/src/workerd/jsg/README.md
@@ -10,6 +10,36 @@ the same interface on top of different JavaScript engines. However, as of today,
 quite the case. At present application code will still need to use V8 APIs directly in some
 cases. We would like to improve this in the future.
 
+## V8 Fast API
+
+V8 Fast API allows V8 to compile JavaScript code that calls native functions in a way that directly jumps to the C++ implementation without going through the usual JavaScript-to-C++ binding layers. This is accomplished by having V8 generate specialized machine code that knows how to call the C++ function directly, skipping the overhead of value conversion and JavaScript calling conventions.
+
+### Requirements for Fast API compatibility
+
+For a method to be compatible with Fast API, it must adhere to several constraints due to v8 itself, and may change as the fast api mechanism continues to evolve:
+
+1. **Return Type**: Must be one of these primitive types: `void`, `bool`, `int32_t`, `uint32_t`, `float`, or `double`.
+
+2. **Parameter Types**: Can be the primitive types listed above, or V8 handle types like `v8::Local<v8::Value>` or `v8::Local<v8::Object>`, or types that can be unwrapped from a V8 value using the TypeWrapper.
+
+3. **Method Structure**: Can be a regular instance method, a const instance method, or a method that takes `jsg::Lock&` as its first parameter.
+
+### Using Fast API in workerd
+
+By default, any `JSG_METHOD(name)` will execute fast path if the method signature is compatible with Fast API requirements. To explicitly force the function to use v8 Fast API in workerd, you can use `JSG_ASSERT_FASTAPI`.
+
+### How it works
+
+When a method is registered with the v8 fast api, workerd automatically:
+
+1. Checks if the method signature is compatible with Fast API requirements
+2. Registers both a regular (slow path) method handler and a Fast API handler
+3. Let's V8 optimize calls to this method when possible
+
+V8 determines at runtime whether to use the fast or slow path:
+- The fast path is used when the method is called from optimized code
+- The slow path is used when called from unoptimized code or when handling complex cases
+
 ## The Basics
 
 If you haven't done so already, I recommend reading through both the ["KJ Style Guide"][]
@@ -1430,11 +1460,11 @@ the following rules are applied:
    modifier will be added if it's not already present. In the special case that the override is a
    `type`-alias to `never`, the generated definition will be deleted.
 
-2. Otherwise, the override will be converted to a TypeScript class as follows: (where `<name>` is the
+2. Otherwise, the override will be converted to a TypeScript class as follows: (where `<n>` is the
    unqualified C++ type name of this resource type)
 
-   1. If an override starts with `extends `, `implements ` or `{`: `class <name> ` will be prepended
-   2. If an override starts with `<`: `class <name>` will be prepended
+   1. If an override starts with `extends `, `implements ` or `{`: `class <n> ` will be prepended
+   2. If an override starts with `<`: `class <n>` will be prepended
    3. Otherwise, `class ` will be prepended
 
    After this, if the override doesn't end with `}`, ` {}` will be appended.
--- a/src/workerd/jsg/fast-api-test.c++
+++ b/src/workerd/jsg/fast-api-test.c++
@@ -1,0 +1,409 @@
+#include "jsg-test.h"
+
+#include <workerd/util/autogate.h>
+
+namespace workerd::jsg::test {
+namespace {
+
+using jsg::CallCounter;
+
+struct WrappedInt {
+  int32_t i;
+
+  JSG_STRUCT(i);
+};
+
+class StaticMethodContainer: public jsg::Object {
+ public:
+  StaticMethodContainer() = default;
+
+  static int32_t staticMethod() {
+    return 42;
+  }
+
+  uint32_t getValue() const {
+    return value;
+  }
+
+  void setValue(uint32_t value_) {
+    value = value_;
+  }
+
+  JSG_RESOURCE_TYPE(StaticMethodContainer) {
+    JSG_STATIC_METHOD(staticMethod);
+    JSG_PROTOTYPE_PROPERTY(value, getValue, setValue);
+  }
+
+ private:
+  uint32_t value = 42;
+};
+
+class FastMethodContext: public jsg::Object, public jsg::ContextGlobal {
+ public:
+  int32_t add(int32_t a, int32_t b) {
+    return a + b;
+  }
+
+  int32_t processValue(v8::Local<v8::Value> value) {
+    KJ_ASSERT(value->IsNumber(), "Value must be a number");
+    return value.As<v8::Int32>()->Value();
+  }
+
+  // When arbitrary type unwrapping is supported, keep this test as it is.
+  int32_t processObject(v8::Local<v8::Object> obj) {
+    auto isolate = v8::Isolate::GetCurrent();
+    auto context = isolate->GetCurrentContext();
+    auto testKey = v8::String::NewFromUtf8(isolate, "test").ToLocalChecked();
+    v8::Local<v8::Value> value;
+    if (obj->Get(context, testKey).ToLocal(&value)) {
+      v8::String::Utf8Value type(isolate, value->TypeOf(isolate));
+      KJ_ASSERT(value->IsInt32(), "Received ", *type);
+      return value.As<v8::Int32>()->Value();
+    }
+    return 0;
+  }
+
+  void voidMethod(uint32_t a, uint32_t b) {
+    // Do nothing. This is testing void return type.
+  }
+
+  int32_t throwError(int32_t code) {
+    if (code > 0) {
+      JSG_FAIL_REQUIRE(TypeError, "Test error with code ", code);
+    }
+    return 0;
+  }
+
+  int32_t addWithLock(jsg::Lock& js, int32_t a, v8::Local<v8::Value> b) {
+    KJ_ASSERT(b->IsNumber(), "Second parameter must be a number");
+    int32_t bValue = b.As<v8::Int32>()->Value();
+    return a + bValue;
+  }
+
+  int32_t constAdd(int32_t a, int32_t b) {
+    return a + b;
+  }
+
+  int32_t constAddWithLock(jsg::Lock& js, int32_t a, int32_t b) {
+    return a + b;
+  }
+
+  int32_t unwrapStruct(jsg::Lock& js, WrappedInt w) {
+    return w.i;
+  }
+
+  int32_t unwrapUint(jsg::Lock& js, kj::uint u) {
+    return u;
+  }
+
+  int32_t unwrapString(jsg::Lock& js, kj::String str) {
+    return str.size();
+  }
+
+  int32_t unwrapBufferSource(jsg::Lock& js, jsg::BufferSource source) {
+    return source.size();
+  }
+
+  int32_t unwrapMaybe(jsg::Lock& js, kj::Maybe<kj::String> str) {
+    KJ_IF_SOME(s, str) {
+      return s.size();
+    } else {
+      return -1;
+    }
+  }
+
+  int32_t unwrapOptional(jsg::Lock& js, jsg::Optional<kj::String> str) {
+    KJ_IF_SOME(s, str) {
+      return s.size();
+    } else {
+      return -1;
+    }
+  }
+
+  int32_t unwrapLenientOptional(jsg::Lock& js, jsg::LenientOptional<kj::String> str) {
+    KJ_IF_SOME(s, str) {
+      return s.size();
+    } else {
+      return -1;
+    }
+  }
+
+  jsg::Ref<StaticMethodContainer> newContainer() {
+    return jsg::alloc<StaticMethodContainer>();
+  }
+
+  JSG_RESOURCE_TYPE(FastMethodContext) {
+    JSG_NESTED_TYPE(StaticMethodContainer);
+
+    JSG_METHOD(add);
+    JSG_METHOD(processValue);
+    JSG_METHOD(processObject);
+    JSG_METHOD(voidMethod);
+    JSG_METHOD(throwError);
+    JSG_METHOD(addWithLock);
+    JSG_METHOD(constAdd);
+    JSG_METHOD(constAddWithLock);
+    JSG_METHOD(unwrapStruct);
+    JSG_METHOD(unwrapUint);
+    JSG_METHOD(unwrapString);
+    JSG_METHOD(unwrapBufferSource);
+    JSG_METHOD(unwrapMaybe);
+    JSG_METHOD(unwrapOptional);
+    JSG_METHOD(unwrapLenientOptional);
+
+    JSG_METHOD(newContainer);
+  }
+};
+
+JSG_DECLARE_DEBUG_ISOLATE_TYPE(
+    FastMethodIsolate, FastMethodContext, WrappedInt, StaticMethodContainer);
+
+jsg::V8System v8System({"--allow-natives-syntax"});
+
+struct Test {
+  kj::LiteralStringConst expr;
+  kj::LiteralStringConst expectedReturnType;
+  kj::LiteralStringConst expectedReturnValue;
+  kj::LiteralStringConst target = ""_kjc;
+  kj::uint expectedSlowCount = 1;
+};
+
+CallCounter runTest(Test test) {
+  jsg::callCounter.reset();
+  jsg::test::Evaluator<FastMethodContext, FastMethodIsolate> e(v8System);
+
+  auto target = test.target == ""_kjc ? kj::str(test.expr) : kj::str(test.target, ".", test.expr);
+  e.expectEval(target, test.expectedReturnType, test.expectedReturnValue);
+  KJ_ASSERT(jsg::callCounter == CallCounter(test.expectedSlowCount, 0));
+
+  e.expectEval(kj::str("const fastCall = () => { return ", target,
+                   "; }; "
+                   "%PrepareFunctionForOptimization(fastCall); "
+                   "fastCall(); "
+                   "%OptimizeFunctionOnNextCall(fastCall); "
+                   "fastCall()"),
+      test.expectedReturnType, test.expectedReturnValue);
+  return jsg::callCounter;
+}
+
+KJ_TEST("v8::Local<v8::Value> and v8::Local<v8::Object> as fast method parameters") {
+  util::Autogate::initAutogateNamesForTest({"v8-fast-api"_kj});
+  KJ_ASSERT(runTest({"processValue(42)"_kjc, "number"_kjc, "42"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(
+      runTest({"processObject({test: 123})"_kjc, "number"_kjc, "123"_kjc}) == CallCounter(2, 1));
+}
+
+KJ_TEST("Lock& as the first parameter in fast method calls") {
+  KJ_ASSERT(runTest({"addWithLock(3, 4)"_kjc, "number"_kjc, "7"_kjc}) == CallCounter(2, 1));
+}
+
+KJ_TEST("Const methods in fast method calls") {
+  KJ_ASSERT(runTest({"constAdd(3, 4)"_kjc, "number"_kjc, "7"_kjc}) == CallCounter(2, 1));
+}
+
+KJ_TEST("Const methods with Lock& in fast method calls") {
+  KJ_ASSERT(runTest({"constAddWithLock(3, 4)"_kjc, "number"_kjc, "7"_kjc}) == CallCounter(2, 1));
+}
+
+KJ_TEST("type unwrapping arguments") {
+  KJ_ASSERT(runTest({"unwrapUint(4)"_kjc, "number"_kjc, "4"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(runTest({"unwrapStruct({i: 3})"_kjc, "number"_kjc, "3"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(runTest({"unwrapString('0123')"_kjc, "number"_kjc, "4"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(runTest({"unwrapBufferSource(new Uint8Array(256))"_kjc, "number"_kjc, "256"_kjc}) ==
+      CallCounter(2, 1));
+  KJ_ASSERT(runTest({"unwrapMaybe(undefined)"_kjc, "number"_kjc, "-1"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(runTest({"unwrapMaybe('foo')"_kjc, "number"_kjc, "3"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(
+      runTest({"unwrapOptional(undefined)"_kjc, "number"_kjc, "-1"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(runTest({"unwrapOptional('foo')"_kjc, "number"_kjc, "3"_kjc}) == CallCounter(2, 1));
+  KJ_ASSERT(runTest({"unwrapLenientOptional(undefined)"_kjc, "number"_kjc, "-1"_kjc}) ==
+      CallCounter(2, 1));
+  KJ_ASSERT(
+      runTest({"unwrapLenientOptional('foo')"_kjc, "number"_kjc, "3"_kjc}) == CallCounter(2, 1));
+
+  KJ_ASSERT(runTest({"StaticMethodContainer.staticMethod()"_kjc, "number"_kjc, "42"_kjc}) ==
+      CallCounter(2, 1));
+}
+
+KJ_TEST("Fast methods should work with getters/setters") {
+  KJ_ASSERT(
+      runTest({"value"_kjc, "number"_kjc, "42"_kjc, "newContainer()"_kjc, 2}) == CallCounter(5, 1));
+  KJ_ASSERT(runTest({"value = 12"_kjc, "number"_kjc, "12"_kjc, "newContainer()"_kjc, 2}) ==
+      CallCounter(5, 1));
+}
+
+KJ_TEST("Fast methods properly catch JSG_FAIL_REQUIRE errors") {
+  jsg::callCounter.reset();
+  jsg::test::Evaluator<FastMethodContext, FastMethodIsolate> e(v8System);
+
+  // Test that directly calling the method results in an error
+  e.expectEval("throwError(42)", "throws", "TypeError: Test error with code 42");
+  KJ_ASSERT(jsg::callCounter == CallCounter(1, 0));
+
+  // First run a non-throwing call to allow optimization
+  e.expectEval("throwError(0)", "number", "0");
+  KJ_ASSERT(jsg::callCounter == CallCounter(2, 0));
+
+  e.expectEval("function fastThrow(code) { return throwError(code); }; "
+               "%PrepareFunctionForOptimization(fastThrow); "
+               "fastThrow(0); "
+               "%OptimizeFunctionOnNextCall(fastThrow); "
+               "fastThrow(42)",
+      "throws", "TypeError: Test error with code 42");
+
+  // The counts should now include both the slow path and fast path calls
+  // 3 slow path calls (original direct call + non-throwing call + the initial optimization call that didn't throw)
+  // 1 fast path call (the optimized call that threw an error)
+  KJ_ASSERT(jsg::callCounter == CallCounter(3, 1));
+}
+
+KJ_TEST("isFastMethodCompatible Detection") {
+  static_assert(
+      FastApiWrappedObject<FastMethodIsolate_TypeWrapper, kj::String>, "should be compatible");
+
+  static_assert(FastApiParam<FastMethodIsolate_TypeWrapper, WrappedInt>, "should be compatible");
+
+  static_assert(FastApiParam<FastMethodIsolate_TypeWrapper, kj::OneOf<kj::String, WrappedInt>>,
+      "should be compatible");
+
+  static_assert(!FastApiParam<FastMethodIsolate_TypeWrapper, WrappedInt&>, "should be compatible");
+
+  struct Foo {};
+
+  static_assert(!FastApiWrappedObject<FastMethodIsolate_TypeWrapper, Foo>, "foo is not a jsg type");
+
+  static_assert(!FastApiWrappedObject<FastMethodIsolate_TypeWrapper, jsg::Lock&>,
+      "lock is not a wrapped type");
+
+  static_assert(!isFastMethodCompatible<FastMethodIsolate_TypeWrapper,
+                    void (FastMethodContext::*)(bool, jsg::Lock&)>,
+      "lock is accepted only as first argument");
+
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper,
+                    void (FastMethodContext::*)(jsg::Lock&, bool)>,
+      "lock is accepted only as first argument");
+
+  // Method type declarations
+  // -----------------------
+
+  // Compatible basic types
+  using VoidMethod = void (FastMethodContext::*)();
+  using IntMethod = int32_t (FastMethodContext::*)(int32_t);
+  using BoolMethod = bool (FastMethodContext::*)(double, bool);
+  using FloatMethod = float (FastMethodContext::*)(int32_t, float);
+
+  // V8 Local types as parameters - should be compatible
+  using V8ValueParamMethod = int32_t (FastMethodContext::*)(v8::Local<v8::Value>);
+  using V8ObjectParamMethod = int32_t (FastMethodContext::*)(v8::Local<v8::Object>);
+
+  // Const method variants - should be compatible
+  using ConstIntMethod = int32_t (FastMethodContext::*)(int32_t) const;
+  using ConstFloatMethod = float (FastMethodContext::*)(float) const;
+
+  // Methods with Lock& as first parameter - should be compatible
+  using LockFirstMethod = int32_t (FastMethodContext::*)(jsg::Lock&, int32_t);
+  using LockFirstVoidMethod = void (FastMethodContext::*)(jsg::Lock&, bool);
+  using LockFirstWithV8Local = int32_t (FastMethodContext::*)(jsg::Lock&, v8::Local<v8::Value>);
+  using ConstLockFirstMethod = int32_t (FastMethodContext::*)(jsg::Lock&, int32_t) const;
+
+  // ---- Non-compatible method types ----
+
+  // Pointer types - not compatible
+  using PointerParamMethod = void* (FastMethodContext::*)(void*);
+  using PointerReturnMethod = void* (FastMethodContext::*)();
+
+  // V8 Local as return type - not compatible
+  using V8ReturnMethod = v8::Local<v8::Value> (FastMethodContext::*)(int32_t);
+
+  // Non-primitive types - not compatible
+  using StringMethod = kj::String (FastMethodContext::*)(kj::String);
+  using ComplexMethod = Ref<FastMethodContext> (FastMethodContext::*)(jsg::Lock&);
+  using KjArrayMethod = kj::Array<int> (FastMethodContext::*)(int32_t);
+  using PromiseMethod = jsg::Promise<int> (FastMethodContext::*)(int32_t);
+  using MaybeVoidMethod = kj::Maybe<void> (FastMethodContext::*)();
+  using PointerMethod = void(v8::Isolate*);
+  using ReferenceMethod = void(v8::Isolate&);
+  using StaticMethodContainerReferenceMethod = void(StaticMethodContainer&);
+  using StaticMethodContainerMethod = void(StaticMethodContainer);
+
+  // Static assertions for compatible method types
+  // --------------------------------------------
+
+  // Basic primitives
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, VoidMethod>,
+      "Void methods should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, IntMethod>,
+      "Integer methods should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, BoolMethod>,
+      "Boolean methods should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, FloatMethod>,
+      "Float methods should be fast-method compatible");
+
+  // V8 Local parameters
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, V8ValueParamMethod>,
+      "Methods with v8::Local<v8::Value> parameters should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, V8ObjectParamMethod>,
+      "Methods with v8::Local<v8::Object> parameters should be fast-method compatible");
+
+  // Const methods
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, ConstIntMethod>,
+      "Const methods should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, ConstFloatMethod>,
+      "Const methods with float should be fast-method compatible");
+
+  // Lock& as first parameter
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, LockFirstMethod>,
+      "Methods with Lock& as first parameter should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, LockFirstVoidMethod>,
+      "Void methods with Lock& as first parameter should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, LockFirstWithV8Local>,
+      "Methods with Lock& and v8::Local params should be fast-method compatible");
+  static_assert(isFastMethodCompatible<FastMethodIsolate_TypeWrapper, ConstLockFirstMethod>,
+      "Const methods with Lock& as first parameter should be fast-method compatible");
+
+  // Complex methods
+  static_assert(FastApiParam<FastMethodIsolate_TypeWrapper, WrappedInt>,
+      "JSG_STRUCT as a parameter is supported");
+
+  // Static assertions for incompatible method types
+  // ----------------------------------------------
+
+  // Pointer types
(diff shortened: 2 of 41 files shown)


src/workerd/jsg/jsg.h
```diff
@@ -201,6 +201,18 @@ namespace workerd::jsg {
     registry.template registerMethod<NAME, decltype(&Self::name), &Self::name>();                  \
   } while (false)
 
+// Similar to JSG_METHOD, but also creates a V8 Fast API path for the method if its signature
+// is compatible with V8's Fast API Call conventions. This can significantly improve performance
+// for performance-critical methods with supported argument types.
+//
+// The same method will be used for both fast and slow paths, so it should be compatible with both.
```
kentonv, line 208: Could we just automatically use fast API for method signatures that support it, without the need for each method to be explicitly declared to use it?

anonrig: Yes, I'll add this to the TODO.

jasnell: I think ideally I'd prefer this to be addressed now as opposed to introducing the second macro here only to remove it later once the `JSG_METHOD` macro is updated.

anonrig: Yes, I'll address this before merging. I've added it to the TODO list in the pull-request description

src/workerd/jsg/resource.h
```diff
@@ -18,6 +18,7 @@
 #include <workerd/jsg/meta.h>
 #include <workerd/jsg/modules.capnp.h>
 
+#include <v8-fast-api-calls.h>
```
jasnell, line 21: I do have some concerns here with these specific limitations of fast apis from the v8-fast-api-calls header:

```
 *   - they should not allocate on the JS heap;
 *   - they should not trigger JS execution.
 * To enforce them, the embedder could use the existing
 * v8::Isolate::DisallowJavascriptExecutionScope and a utility similar to
 * Blink's NoAllocationScope:
```

Our existing `JSG_METHOD` mappings do not have these restrictions obviously, and lots of our `JSG_METHOD` implementations both allocate on the JS heap and trigger JS execution. Combining both fast and slow implementations behind a single method signature might be a problem as we would actually need to know within the body of the function if allocations and JS execution is permitted. Unfortunately just gating on the types is not necessarily enough here, for instance:

```cpp
class Foo : public jsg::Object {
 public:
  bool hasSafeTypesButIsNotSafe(bool b) {
    v8::Isolate* isolate = v8::Isolate::GetCurrent();
    v8::String::NewFromUtf(isolate, ...);
    return true;
  }
}
```

I think maybe your fast callback implementation should make sure to use the `DiallowJavascriptExecutionScope` utility etc to make things safer here.

Fortunately, the fast method aren't receiving anything like the `jsg::Lock&` and receiver arguments so the developer would need to jump through hoop to do the wrong thing but I still think it's worth adding some additional safety guards.

jasnell: Throwing exceptions here is also a concern:

```
 * Due to these limitations, it's not directly possible to report errors by
 * throwing a JS exception or to otherwise do an allocation. There is an
 * alternative way of creating fast calls that supports falling back to the
 * slow call and then performing the necessary allocation. When one creates
 * the fast method by using CFunction::MakeWithFallbackSupport instead of
 * CFunction::Make, the fast callback gets as last parameter an output variable,
 * through which it can request falling back to the slow call. So one might
 * declare their method like:
```

This limitation would mean that something like `JSG_REQUIRE(...)` will need to be handled differently in these APIS.

```cpp
 public:
  void Foo(bool bar) { JSG_FAIL_REQUIRE(Error, "how does this throw in fast api calls?"); }
  // ...
};
```

src/workerd/jsg/resource.h
```diff
@@ -191,6 +192,89 @@ template <typename TypeWrapper,
...
+        kj::_::Indexes<indexes...>>::callback(args);
+  }
+
+  static Ret fastCallback(
+      v8::Local<v8::Object> receiver, Args... fastArgs, v8::FastApiCallbackOptions& options) {
+    v8::Local<v8::Context> context = options.isolate->GetCurrentContext();
+
+    auto& self = extractInternalPointer<T, isContext>(context, receiver);
+
+    return (self.*method)(fastArgs...);
```
jasnell, line 269: This will need to catch and handle exceptions (e.g. `kj::Exception`, `jsg::JsExceptionThrown`) that are thrown by the method. Look for `liftKj` within `workerd/jsg` for details.

src/workerd/jsg/resource.h
anonrig: I've added `DiallowJavascriptExecutionScope`. I also need to handle the error reporting for sure.

src/workerd/jsg/resource.h
```diff
@@ -191,6 +192,94 @@ template <typename TypeWrapper,
...
+template <typename Class, typename Ret, typename... Args>
+struct IsFastMethodCompatible<Ret (Class::*)(Args...)> {
+  static constexpr bool value = IsFastApiCompatible<Ret, Args...>::value;
+};
+
+// Specialization for const non-static member functions
+template <typename Class, typename Ret, typename... Args>
+struct IsFastMethodCompatible<Ret (Class::*)(Args...) const> {
+  static constexpr bool value = IsFastApiCompatible<Ret, Args...>::value;
+};
```
jasnell, line 228: I think another key difference here that we need to be careful of is that the type mappings for fast api calls will *not* pass through our TypeWrapper mechanism. This is generally a good thing but might be surprising for folks. It will be worth documenting this.

src/workerd/jsg/resource.h
```diff
@@ -191,6 +192,94 @@ template <typename TypeWrapper,
...
+    MethodCallback<TypeWrapper, methodName, isContext, T, Ret (U::*)(Args...), method,
+        kj::_::Indexes<indexes...>>::callback(args);
+  }
+
+  static Ret fastCallback(
+      v8::Local<v8::Object> receiver, Args... fastArgs, v8::FastApiCallbackOptions& options) {
+    auto context = options.isolate->GetCurrentContext();
+
+    // Create a scope that disallows JavaScript execution while in the fast API callback
+    // This enforces the V8 Fast API constraint that the callback must not trigger JS execution
```
jasnell, line 268: This is good but I also think a utility that catches and prevents allocation is also important as suggested by the fast calls header.

src/workerd/jsg/resource.h
jasnell: the error reporting is likely to be tricky... specifically due to the fallback pattern. That is, in a fast-api call, if an action fails resulting in an error being thrown, we are supposed to fallback to the slow-api call, try the operation again, and then throw if it still fails... so that fast-api call is expected to be fully idempotent up to the point the error would be thrown so that the operation can be tried again. The slow-api call has no such restriction and I'm not sure we can definitely say that all our existing `JSG_METHOD`s are idempotent up to the point an error would be thrown. This fact alone might be enough to justify having the separate `JSG_FAST_METHOD` rather than reusing the `JSG_METHOD`. This would need to be carefully handled.

jasnell:
@anonrig ... have you done a survey to see just how many of our existing `JSG_METHOD` apis actually would be able to qualify as fast-api calls?

src/workerd/jsg/resource.h
kentonv: Ah yeah given the extra requirements it seems unrealistic to just automatically apply fastapi all methods whose signatures seem OK. We need methods to explicitly opt in to say that they follow these rules.

anonrig: It seems you can trigger GC, call JS and even throw errors from fast methods right now! So, we can potentially call fast api on all possible methods, if I'm not mistaken? https://github.com/v8/v8/commit/be03ed2efc10b71c89ce76ef89c5c5325b527e68

jasnell: I mentioned in direct chat but there's enough discrepancy between the current state and out of date comments and the fact that this appears to be a moving target that it actually makes me more nervous to land this. We need to make sure this is stable enough before we depend on it.

erikcorry:
Given the restrictions I would suggest identifying API calls that are heavily used/a performance bottleneck and making them fast on a case by case basis.

anonrig:
> Given the restrictions I would suggest identifying API calls that are heavily used/a performance bottleneck and making them fast on a case by case basis.

@erikcorry It seems those restrictions are gone, but V8 team didn't update the v8-fast-api-calls.h file. Referring: https://github.com/v8/v8/commit/be03ed2efc10b71c89ce76ef89c5c5325b527e68

src/workerd/jsg/resource.h
anonrig: This is not needed because we can make allocations.

erikcorry:
OK after looking at the changes linked from https://issues.chromium.org/issues/41492790 it's clear that all the restrictions are gone. There's no good reason not to use the fastAPI.

src/workerd/jsg/resource.h
anonrig: We can enable type mappings now, since all of the limitations are gone.

src/workerd/jsg/resource.h
anonrig: Yes, I'll add liftKj. That's a good recommendation!

npaun (approved):
Implementation code looks great to me. In terms of verifying the implications of turning this on, I'll leave that to the others to approve.

src/workerd/jsg/resource.h
```diff
@@ -966,6 +1086,28 @@ struct ResourceTypeBuilder {
             v8::Local<v8::Value>(), signature, 0, v8::ConstructorBehavior::kThrow));
   }
 
+  // Register a method with both slow and fast paths using the V8 Fast API mechanism
+  template <const char* name, typename Method, Method method, bool trackCallCount = false>
```
jasnell, line 1090: Perhaps instead of passing `trackCallCount` as a template option and having the additional jsg macro, can enabling the call count tracking be automatically enabled only when running in predicable mode for tests?

src/workerd/jsg/resource.h
```diff
@@ -966,6 +1086,28 @@ struct ResourceTypeBuilder {
...
+    // Create a v8::MemorySpan to hold the CFunction
+    // TODO(soon): Add support for multiple CFunctions.
+    v8::MemorySpan<const v8::CFunction> cfunctionSpan(&cFunction, 1);
+
+    // Create a function template with both slow and fast paths
+    auto tmpl = v8::FunctionTemplate::NewWithCFunctionOverloads(isolate,
+        &FastMethodCallback<TypeWrapper, name, isContext, Self, Method, method,
+            ArgumentIndexes<Method>>::template slowCallback<trackCallCount>,
+        v8::Local<v8::Value>(), signature, 0, v8::ConstructorBehavior::kThrow,
+        v8::SideEffectType::kHasNoSideEffect, cfunctionSpan);
```
jasnell, line 1106: hmm.. is `kHasNoSideEffect` correct? I'm not sure we can necessarily guarantee that a JSG_METHOD has no side effect.

src/workerd/jsg/resource.h
```diff
@@ -191,6 +192,125 @@ template <typename TypeWrapper,
...
+template <typename Class, typename Ret, typename... Args>
+struct IsFastMethodCompatible<Ret (Class::*)(Args...)> {
+  static constexpr bool value = IsFastApiCompatible<Ret, Args...>::value;
+};
+
+// Specialization for const non-static member functions
+template <typename Class, typename Ret, typename... Args>
+struct IsFastMethodCompatible<Ret (Class::*)(Args...) const> {
+  static constexpr bool value = IsFastApiCompatible<Ret, Args...>::value;
+};
```
jasnell, line 233: Nit: It would be helpful to have some `static_assert` tests in `fast-method-test.c++` that verify these work as expected for method signatures that are *not* fast-method compatible.

src/workerd/jsg/resource.h
```diff
@@ -191,6 +192,125 @@ template <typename TypeWrapper,
...
+
+    try {
+      if constexpr (std::is_same_v<Ret, void>) {
+        (self.*method)(fastArgs...);
+        return;
+      } else {
+        return (self.*method)(fastArgs...);
+      }
+    } catch (kj::Exception& exception) {
+      throwInternalError(isolate, kj::mv(exception));
```
jasnell, line 292: If you notice in the implementation of `liftKj(...)` we actually use a nested try catch pattern. Are you certain that the same pattern is not needed here? cc'ing @kentonv as the author of liftKj

src/workerd/jsg/jsg.h
```diff
@@ -201,6 +230,26 @@ namespace workerd::jsg {
...
+// See V8's v8-fast-api-calls.h for more details on Fast API requirements and limitations.
```
jasnell, line 238: Can we add a TODO comment here indicating that, long-term, the plan would be to try to merge `JSG_FAST_METHOD` into `JSG_METHOD` such that any compatible `JSG_METHOD` would just automatically be defined with a fast-mode alternative?

src/workerd/jsg/resource.h
anonrig: I updated the implementation. (used kHasSideEffect)

src/workerd/jsg/resource.h
anonrig: `isPredictableModeForTest` is not a constexpr method, this means that if we use it, registerFastMethod on production will always pay a toll for an unnecessary branch

src/workerd/jsg/resource.h
anonrig: Last commit extends these tests. Thank you for the review.

anonrig:
@jasnell With the current state of the implementation (which doesn't support typewrapper), we can replace the following functions to v8 fast API

  1. From Event class in /workerd/src/workerd/api/basics.h:
    - void stopImmediatePropagation()
    - void preventDefault()
    - void stopPropagation()
    - int getEventPhase()
    - bool getCancelBubble()
    - void setCancelBubble(bool stopped)
    - bool getComposed()
    - bool getBubbles()
    - bool getCancelable()
    - bool getDefaultPrevented()
    - bool getReturnValue()
    - bool getIsTrusted()
  2. From AbortSignal in basics.h:
    - bool getAborted()
  3. From Blob class in /workerd/src/workerd/api/blob.h:
    - int getSize()
  4. From File class in blob.h:
    - double getLastModified()
  5. From NodeJS Buffer in /workerd/src/workerd/api/node/buffer.h:
    - bool isView()
  6. From HTTP Response/Request classes:
    - Various getter methods like int getLength(), int getStatus()
    - bool hasBody()
  7. From Headers class:
    - bool has(kj::String name)
  8. From Stream classes:
    - void close() methods

src/workerd/jsg/jsg.h
```diff
@@ -48,6 +48,35 @@ kj::String KJ_STRINGIFY(v8::Local<T> value) {
 
 namespace workerd::jsg {
 
+// The following APIs are available to be used for testing purposes for
+// v8 fast API. Please take a look into v8-fast-api.h file in V8 for more
+// information.
+namespace fast_api {
```
kentonv, line 54: Nitpick: We don't use underscores in names. `fastApi` or `fastapi`...

src/workerd/jsg/jsg.h
```diff
@@ -48,6 +48,35 @@ kj::String KJ_STRINGIFY(v8::Local<T> value) {
...
+
+namespace {
```
kentonv, line 56: Anonymous namespaces do not work in header files. This will cause the content of the namespace to be cloned into every translation unit, creating code bloat.

src/workerd/jsg/resource.h
```diff
@@ -44,6 +45,28 @@ const auto kIllegalInvocation =
...
+  } catch (JsExceptionThrown&) {
+    // Exception already set
+  } catch (std::exception& exception) {
+    throwInternalError(isolate, exception.what());
+  } catch (...) {
+    throwInternalError(
+        isolate, kj::str("caught unknown exception of type: ", kj::getCaughtExceptionType()));
+  }
+
+  if constexpr (!std::is_same_v<Ret, void>) {
```
kentonv, line 65: Prefer `kj::isSameType` for consistency -- lots of stuff in JSG and the rest of our codebase uses `kj::isSameType`, it's confusing if we add a bunch of other code using `std::is_same_v` (which does the same thing).

src/workerd/jsg/resource.h
```diff
@@ -44,6 +45,28 @@ const auto kIllegalInvocation =
 class Serializer;
 class Deserializer;
 
+// Helper function for executing fast methods with consistent error handling.
+// Used by FastMethodCallback to propagate errors in a consistent manner.
+template <typename Ret, typename Func>
+Ret executeFastMethod(v8::Isolate* isolate, Func&& func) {
```
kentonv, line 51: This looks a lot like `liftKj`. Why is it separate? Given the similarity in structure, at the very least I think it should live with `liftKj` so that people changing error handling in `liftKj` know to update this function as well. Though if we can refactor `liftKj` to be able to handle both cases with the same code that would be ideal.

src/workerd/jsg/jsg.h
anonrig: I was planning on enabling it on this PR. No need to add a TODO to code since this PR already has a TODO in the pull-request description

src/workerd/jsg/resource.h
```diff
@@ -191,6 +214,308 @@ template <typename TypeWrapper,
...
+  }
+
+  template <bool trackCallCount = false>
+  static v8::CFunction makeCFunction() {
+    static_assert(IsFastApiCompatible<Ret, Args...>::value);
+    return v8::CFunction::Make(fastCallback<trackCallCount>);
+  }
+};
+
+// Specialization for const instance methods
```
kentonv, line 388: This doesn't seem related to fast API? Are you just adding support for const methods because they weren't supported before?

Do we actually need this? Can it be in a separate PR, to reduce the complexity of this PR?

src/workerd/jsg/fast-method-test.c++
```diff
@@ -0,0 +1,304 @@
...
+
+  int32_t processValue(v8::Local<v8::Value> value) {
+    // Return 42 if it's a number, 0 otherwise
+    if (value->IsNumber()) {
+      return 42;
+    }
+    return 0;
+  }
+
+  // Keep this test as it is, and don't change it to use kj internals to test this
```
kentonv, line 20: What does this comment mean?

src/workerd/jsg/fast-method-test.c++
```diff
@@ -0,0 +1,304 @@
...
+class FastMethodContext: public jsg::Object, public jsg::ContextGlobal {
+ public:
+  int32_t add(int32_t a, int32_t b) {
+    return a + b;
+  }
+
+  int32_t processValue(v8::Local<v8::Value> value) {
+    // Return 42 if it's a number, 0 otherwise
+    if (value->IsNumber()) {
+      return 42;
```
kentonv, line 15: Why not actually return the number? Wouldn't it be useful to verify we're getting the _correct_ value?

src/workerd/jsg/fast-method-test.c++
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  // Keep this test as it is, and don't change it to use kj internals to test this
+  // particular case.
+  int32_t processObject(v8::Local<v8::Object> obj) {
+    // Return 123 if object has a property 'test', 0 otherwise
+    auto context = v8::Isolate::GetCurrent()->GetCurrentContext();
+    auto v8_test = v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), "test").ToLocalChecked();
+    if (obj->Has(context, v8_test).FromMaybe(false)) {
+      return 123;
```
kentonv, line 27: Would be nice to return the content of the "test" property (would have to be a number, of course).

src/workerd/jsg/resource.h
```diff
@@ -191,6 +214,308 @@ template <typename TypeWrapper,
...
+template <typename T>
+struct IsFastApiCompatibleReturnType {
+  static constexpr bool value = std::is_same_v<T, void> || std::is_same_v<T, bool> ||
+      std::is_same_v<T, int32_t> || std::is_same_v<T, uint32_t> || std::is_same_v<T, float> ||
+      std::is_same_v<T, double>;
+};
+
+// Helper to determine if a type can be used as a parameter in V8 Fast API
+template <typename T>
+struct IsFastApiCompatibleParamType {
```
kentonv, line 227: If Fast API supports `v8::Local<v8::Value>`, that implies that Fast API actually supports _any_ parameter type. We just have to do the conversion.

Instead of requiring the target function to declare its members as purely numbers or v8::Locals, we should instead say that the target function can use any JSG-compatible type, and we should automatically apply the correct unwrapper to convert a `v8::Local` to the correct type for the function call.

We should, of course, special-case the primitive numeric types, to expose those directly to V8 rather than forcing it to provide us with a handle.

src/workerd/jsg/fast-method-test.c++
anonrig: TLDR: When type wrapper is supported, don't change line 22 to use kj::String because we're testing Local<Object>

src/workerd/jsg/fast-method-test.c++
anonrig: Agreed. That sounds better.

src/workerd/jsg/fast-method-test.c++
anonrig: Agreed. I'll change the tests.

src/workerd/jsg/resource.h
anonrig: Yes, agreed. I'll update this particular part, once I figure out the internals of type wrapper.

src/workerd/jsg/jsg.h
anonrig: Any recommendation on where to move this, or are you ok with the current place?

kentonv: I agree it shouldn't be in jsg.h, since jsg.h is supposed to be the public interface. I don't have a specific alternative suggestion.

src/workerd/jsg/fast-method-test.c++
kentonv: Ah, maybe add "When arbitrary type unwrapping is supported" to the comment.

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,131 @@
...
+#include <v8-fast-api-calls.h>
+#include <v8-local-handle.h>
+#include <v8-value.h>
+
+#include <kj/common.h>
+
+namespace workerd::jsg {
+
+template <typename T>
+struct IsFastApiCompatibleWrappedType1 {
```
npaun, line 14: Can we come up with a better name for this type? I'm not very keen on the "1" at the end.

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,131 @@
...
+};
+
+// Specialization for jsg::Lock& - this is compatible with Fast API.
+// This allows methods that take jsg::Lock& as their first parameter to be used with V8 Fast API.
+// The Lock will be extracted from the V8 isolate in the fast path, enabling methods that need
+// access to the V8 context and isolate to still benefit from the performance of Fast API.
+template <>
+struct IsFastApiCompatibleParamType<jsg::Lock&> {
+  static constexpr bool value = true;
+};
```
npaun, line 69: Is it not possible to handle this using kj::isSameType and merge into the earlier specialization?

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,131 @@
...
+    bool isContext,
+    typename T,
+    typename Method,
+    Method method,
+    typename Indexes>
+struct FastMethodCallback;
+
+template <typename T, typename Enable = void>
+struct FastApiJSGToV8 {
+  static_assert(sizeof(T) != sizeof(T), "Type is not supported in fast API");
```
npaun, line 108: When would `sizeof(T)`  not equal `sizeof(T)`? A comment would be helpful because I wouldn't know.

src/workerd/jsg/jsg-test.h
```diff
@@ -104,10 +104,13 @@ class Evaluator {
         v8::Local<v8::Value> result;
         if (script->Run(js.v8Context()).ToLocal(&result)) {
           v8::String::Utf8Value type(js.v8Isolate, result->TypeOf(js.v8Isolate));
-          v8::String::Utf8Value value(js.v8Isolate, result);
-
           KJ_EXPECT(*type == expectedType, *type, expectedType);
-          KJ_EXPECT(*value == expectedValue, *value, expectedValue);
+
+          // We ignore checking for the return value if the expected type is "undefined"
```
npaun, line 109: Can you add another line to this comment saying why we do that?

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,131 @@
...
+#include <v8-value.h>
+
+#include <kj/common.h>
+
+namespace workerd::jsg {
+
+template <typename T>
+struct IsFastApiCompatibleWrappedType1 {
+  static constexpr bool value = webidl::isDictionaryType<T> ||
+      webidl::isNonCallbackInterfaceType<T> || kj::isSameType<T, kj::String>() ||
```
npaun, line 16: Does this also work with DOMString/USVString, etc.? I'm a little bit intrigued that only kj::String appears here.

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,131 @@
...
+template <typename T>
+struct IsFastApiCompatibleWrappedType {
+  static constexpr bool value =
+      IsFastApiCompatibleWrappedType1<T>::value || IsFastApiMaybe<T>::value;
+};
+
+// Helper to determine if a type can be used as a return type in V8 Fast API
+template <typename T>
+struct IsFastApiCompatibleReturnType {
+  static constexpr bool value = kj::isSameType<T, void>() || kj::isSameType<T, bool>() ||
```
npaun, line 49: It seems like we cannot return wrapped types, etc. Is that a limitation of the fast API, or just not yet covered in this PR?

jasnell: Current limitation of the fast API. While it now accepts a broader range of input arguments it is strictly limited in the types of return values it can handle

npaun: Darn

src/workerd/jsg/fast-api.h
anonrig: This type will eventually need to be merged with type-wrapper. I'll remove this before landing.

src/workerd/jsg/fast-api.h
anonrig: No, this is not yet finished because we need to find a way to reuse the types in `type-wrapper.h` to avoid duplication. Particularly the following lines:

```

#define USING_WRAPPER(Name)                                                                        \
  using Name::getName;                                                                             \
  using Name::wrap;                                                                                \
  using Name::tryUnwrap

  USING_WRAPPER(PrimitiveWrapper);
  USING_WRAPPER(NameWrapper<Self>);
  USING_WRAPPER(StringWrapper);
  USING_WRAPPER(OptionalWrapper<Self>);
  USING_WRAPPER(LenientOptionalWrapper<Self>);
  USING_WRAPPER(MaybeWrapper<Self>);
  USING_WRAPPER(OneOfWrapper<Self>);
  USING_WRAPPER(ArrayWrapper<Self>);
  USING_WRAPPER(SetWrapper<Self>);
  USING_WRAPPER(SequenceWrapper<Self>);
  USING_WRAPPER(GeneratorWrapper<Self>);
  USING_WRAPPER(ArrayBufferWrapper<Self>);
  USING_WRAPPER(DictWrapper<Self>);
  USING_WRAPPER(DateWrapper<Self>);
  USING_WRAPPER(BufferSourceWrapper<Self>);
  USING_WRAPPER(FunctionWrapper<Self>);
  USING_WRAPPER(PromiseWrapper<Self>);
  USING_WRAPPER(NonCoercibleWrapper<Self>);
  USING_WRAPPER(MemoizedIdentityWrapper<Self>);
  USING_WRAPPER(IdentifiedWrapper<Self>);
  USING_WRAPPER(SelfRefWrapper<Self>);
  USING_WRAPPER(ExceptionWrapper<Self>);
  USING_WRAPPER(ObjectWrapper<Self>);
  USING_WRAPPER(V8HandleWrapper);
  USING_WRAPPER(UnimplementedWrapper);
  USING_WRAPPER(JsValueWrapper<Self>);
#undef USING_WRAPPER
```

npaun (approved):

danlapid:
Another note - workers-sdk is a good smoke test for some real world usage of workerd.
We should make sure it passes.

anonrig:
> Another note - workers-sdk is a good smoke test for some real world usage of workerd. We should make sure it passes.

@danlapid I re-run the test, it passed. I think they're just flaky.

src/workerd/jsg/README.md
```diff
@@ -10,6 +10,39 @@ the same interface on top of different JavaScript engines. However, as of today,
...
 
+## V8 Fast API
+
+V8 Fast API allows V8 to compile JavaScript code that calls native functions in a way that directly jumps to the C++ implementation without going through the usual JavaScript-to-C++ binding layers. This is accomplished by having V8 generate specialized machine code that knows how to call the C++ function directly, skipping the overhead of value conversion and JavaScript calling conventions.
+
+### Requirements for Fast API compatibility
+
+For a method to be compatible with Fast API, it must adhere to several constraints:
+
+1. **Return Type**: Must be one of these primitive types: `void`, `bool`, `int32_t`, `uint32_t`, `float`, or `double`.
```
jasnell, line 21: I would expand this just a bit to say that these restrictions are imposed by v8 itself and may change as the fast api mechanism continues to evolve.

src/workerd/jsg/fast-api-test.c++
```diff
@@ -0,0 +1,336 @@
...
+  int32_t unwrapLenientOptional(jsg::Lock& js, jsg::LenientOptional<kj::String> str) const {
+    KJ_IF_SOME(s, str) {
+      return s.size();
+    } else {
+      return -1;
+    }
+  }
+
+  JSG_RESOURCE_TYPE(FastMethodContext) {
+    JSG_TRACKABLE_FAST_METHOD(add);
```
jasnell, line 103: I'm still fairly unhappy with `JSG_TRACKABLE_FAST_METHOD` being a thing that has to exist in jsg.h given that it's only needed for tests.

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,103 @@
...
+#include <v8-fast-api-calls.h>
+#include <v8-local-handle.h>
+#include <v8-value.h>
+
+#include <kj/common.h>
+
+#include <type_traits>
+#include <utility>
+
+namespace workerd::jsg {
```
jasnell, line 14: Nit: I'd like us to get in the habit of including more leading documentation at the top of these headers to provide more context.

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,103 @@
...
+
+// these types are passed by fast api as is and do not require wrapping/unwrapping.
```
jasnell, line 16:
```suggestion
// These types are passed by fast api as is and do not require wrapping/unwrapping.
```

src/workerd/jsg/fast-api.h
```diff
@@ -0,0 +1,103 @@
...
+template <typename T>
+constexpr bool isFastApiPrimitiveType =
+    kj::isSameType<T, void>() || kj::isSameType<T, bool>() || kj::isSameType<T, int32_t>() ||
+    kj::isSameType<T, uint32_t>() || kj::isSameType<T, float>() || kj::isSameType<T, double>();
+
+template <typename T>
+concept FastApiPrimitive = isFastApiPrimitiveType<T>;
```
jasnell, line 23: This feels a bit weird. Shouldn't this be something like...

```cpp
template <typename T>
concept FastApiPrimitive = requires {
  kj::isSameType<T, void>() ||
  kj::isSameType<T, bool>() ||
  kj::isSameType<T, int32_t>() ||
  kj::isSameType<T, uint32_t>() ||
  kj::isSameType<T, double>();
};
```

Or something like that... then you can replace the various other templates with the `FastApiPrimitive` concept...

```cpp
template <typename TypeWrapper, FastApiPrimitive T>
constexpr bool isFastApiWrappedType = isWrappedType<TypeWrapper, T> ;
```

or something like that... with concepts you really ought to be able to streamline this.

src/workerd/jsg/jsg-test.h
```diff
@@ -104,10 +104,15 @@ class Evaluator {
...
+          // because without this check, there is no way to valite a function that
```
jasnell, line 110:
```suggestion
          // because without this check, there is no way to validate a function that
```

jasnell:
I have to agree with @danlapid about preferring that this is rolled out with an autogate.

src/workerd/jsg/jsg.h
```diff
@@ -201,6 +201,26 @@ namespace workerd::jsg {
...
+// The same method will be used for both fast and slow paths, so it should be compatible with both.
+// See V8's v8-fast-api-calls.h for more details on Fast API requirements and limitations.
+#define JSG_FAST_METHOD(name)                                                                      \
+  do {                                                                                             \
+    static const char NAME[] = #name;                                                              \
+    registry.template registerFastMethod<NAME, decltype(&Self::name), &Self::name>();              \
+  } while (false)
+
+// Similar to JSG_FAST_METHOD but tracks the call count of slow and fast path executions.
+#define JSG_TRACKABLE_FAST_METHOD(name)                                                            \
```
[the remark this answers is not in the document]
anonrig: I moved this to fast-api-test.c++ file

src/workerd/jsg/fast-api-test.c++
anonrig: I moved this to fast-api-test.c++ file

src/workerd/jsg/fast-api.h
anonrig: Altough, I like the idea, if you use "requires" on concepts, it will not be possible to test these with static assertions. This is because concepts work differently than traditional SFINAE and template specialization patterns. This occurs because concepts with requires are evaluated as constraints rather than as substitution failures. static_asserts with a concept is immediately evaluated whereas with template type traits, the assertion might only be evaluated when the template is instantiated.

tagging @danlapid since he asked for more concepts as well.

anonrig: For example, taking this example:

```
template <typename TypeWrapper, FastApiPrimitive T>
constexpr bool isFastApiWrappedType = isWrappedType<TypeWrapper, T> ;
```

The following code will not compile:

```
static_assert(!isFastApiWrappedType<kj::String>);
```

jasnell:
Overall LGTM but I'd feel safer if this were rolled out with an autogate.

anonrig:
@jasnell @danlapid put the implementation behind autogate.

anonrig:
Rebased due to conflict.

erikcorry (approved):
It looks fine to me.

jasnell (approved):

src/workerd/jsg/fast-api.h
kentonv: Yes, please.

A good header contains more comments than code.

src/workerd/jsg/resource.h
```diff
@@ -191,6 +194,254 @@ template <typename TypeWrapper,
...
+struct FastMethodCallback<TypeWrapper,
+    methodName,
+    isContext,
+    T,
+    Ret (U::*)(Args...),
+    method,
+    kj::_::Indexes<indexes...>> {
+
+  template <bool trackCallCount = false>
+  static void slowCallback(const v8::FunctionCallbackInfo<v8::Value>& args) {
```
kentonv, line 226: It seems like we could merge `FastMethodCallback` into `MethodCallback`. They have the same template parameters. `slowCallback()` just calls `MethodCallback::callback`, instead we could just drop it. `MethodCallback` would have `callback` and `fastCallback` methods. This would eliminate a lot of boilerplate?

src/workerd/jsg/rtti.h
```diff
@@ -578,7 +578,11 @@ struct MemberCounter {
   template <typename Type, typename GetNamedMethod, GetNamedMethod getNamedMethod>
   inline void registerWildcardProperty() { /* not a member */ }
 
-  template <const char* name, typename Method, Method method>
+  template <const char* name,
+      typename Method,
+      Method method,
+      bool failOnIncompatibleFastApi = false,
```
kentonv, line 584: Why do we need this? This adds a lot of noise to the code.

src/workerd/jsg/rtti.h
```diff
@@ -578,7 +578,11 @@ struct MemberCounter {
...
+      bool trackFastApiUsage = false>
```
kentonv, line 585: This too. This adds a lot of noise, I guess for some testing? Is there a less-invasive way we can support testing?

src/workerd/jsg/rtti.h
anonrig: JSG_FAST_XXX methods should fail if the function signature is not compatible with fast api. This was a request by @danlapid to understand/debug/enforce a method to validate if it's fast api compatible or not.

src/workerd/jsg/rtti.h
anonrig: If we want to have a single method for both fast and slow paths, unfortunately this is the least amount of code change required to support this

src/workerd/jsg/util.h
```diff
@@ -486,4 +522,29 @@ using WontImplement = Unimplemented;
...
+struct CallCounter {
+  uint32_t slow = 0;
+  uint32_t fast = 0;
+
+  void reset() {
+    slow = 0;
+    fast = 0;
+  }
+
+  bool operator==(const kj::Tuple<uint32_t, uint32_t>& rhs) {
```
kentonv, line 538: `kj::Tuple` should almost never be used. The only place `kj::Tuple` actually make sense is when dealing with vararg templates. Anywhere else, it's better to just use a struct.

In fact, if you instead changed this equality operator to take `const CallCounter&` as the rhs argument, then instead of writing:

```
KJ_ASSERT(jsg::fastApi::callCounter == kj::tuple(1, 0));
```

You could write:

```
KJ_ASSERT(jsg::fastApi::callCounter == {1, 0});
```

It's actually less code!

anonrig: Ah yes, that's definitely less code! Thanks @kentonv

src/workerd/jsg/rtti.h
kentonv: Do we think people are going to want to _assert_ that certain methods are fast API? I kind of feel like nobody will care to do that. But if we do think that, then maybe we can have a separate macro like `JSG_ASSERT_FASTAPI(method)` which performs a static_assert. This can work independently of the whole registration system.

anonrig: Yeah I like the idea of using JSG_ASSERT_FASTAPI. I moved the implementation to there.

src/workerd/jsg/resource.h
anonrig: Yeah you're right. Updating now.

src/workerd/jsg/resource.h
```diff
@@ -191,6 +194,141 @@ template <typename TypeWrapper,
     typename Indexes>
 struct MethodCallback;
 
+// Specialization for const instance methods
```
kentonv, line 197: Why do we need this? We haven't needed const API methods up until now, can we go on just not using them and avoid all this extra code?

anonrig:
Rebased from main branch, and force pushed.

src/workerd/jsg/jsg-test.h
```diff
@@ -104,10 +104,15 @@ class Evaluator {
...
           v8::String::Utf8Value type(js.v8Isolate, result->TypeOf(js.v8Isolate));
-          v8::String::Utf8Value value(js.v8Isolate, result);
-
           KJ_EXPECT(*type == expectedType, *type, expectedType);
-          KJ_EXPECT(*value == expectedValue, *value, expectedValue);
+
+          // We ignore checking for the return value if the expected type is "undefined"
+          // because without this check, there is no way to validate a function that
+          // has a return type of undefined.
+          if (*type != "undefined"_kj && expectedValue.size() != 0) {
```
kentonv, line 112: Won't this break checks that are intentionally checking for an empty string?

I don't understand why this is needed. If you want to check for undefined, can't you do `expectEval(code, "undefined", "undefined")`?

src/workerd/jsg/resource.h
```diff
@@ -191,6 +194,75 @@ template <typename TypeWrapper,
     typename Indexes>
 struct MethodCallback;
 
+// Specialization for const methods that take `Lock&` as their first parameter.
```
kentonv, line 197: I previously suggested removing the new specialization for const methods, you resolved my comment, but it seems like it's still here? Why do we need this?

src/workerd/jsg/resource.h
```diff
@@ -998,19 +1372,41 @@ struct ResourceTypeBuilder {
...
-    // properties because it will not properly handle the prototype chain when it comes
-    // to using setters... which is annoying. It means we end up having to use FunctionTemplates
-    // instead of the more convenient property callbacks.
-    auto getterFn = v8::FunctionTemplate::New(isolate, Gcb::callback);
-    auto setterFn = v8::FunctionTemplate::New(
-        isolate, &PropertySetterCallback<TypeWrapper, name, Setter, setter, isContext>::callback);
+    bool useSlowApi = true;
+    if constexpr (isFastMethodCompatible<TypeWrapper, Getter> &&
+        isFastMethodCompatible<TypeWrapper, Setter>) {
+      if (util::Autogate::isEnabled(util::AutogateKey::V8_FAST_API)) {
```
kentonv, line 1387: Maybe it's worth initializing a boolean in TypeWrapper based on the autogate and checking that, rather than calling `isEnabled()` everywhere. Normally `isEnabled()` is cheap enough, but _every_ JSG_METHOD is going to be generating these calls and they are all out-of-line calls, it could actually be non-negligible code bloat.

Also it would be kind of nice not to introduce the dependency on autogates from JSG, we could instead pass it down as a parameter from a higher level.

kentonv (commented):
Have we checked how much this increases binary size and compile time of the `*-api.c++` megafiles?

Not that I think it would be a showstopper, but useful to check on.

src/workerd/jsg/util.h
```diff
@@ -486,4 +522,24 @@ using WontImplement = Unimplemented;
 kj::Maybe<kj::String> checkNodeSpecifier(kj::StringPtr specifier);
 bool isNodeJsCompatEnabled(jsg::Lock& js);
 
+struct CallCounter {
```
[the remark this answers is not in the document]
anonrig: Yes! Definitely.

src/workerd/jsg/promise-test.c++
```diff
@@ -118,6 +118,7 @@ struct PromiseContext: public jsg::Object, public jsg::ContextGlobal {
 JSG_DECLARE_ISOLATE_TYPE(PromiseIsolate, PromiseContext);
 
 KJ_TEST("jsg::Promise<T>") {
+  util::Autogate::initAutogateNamesForTest({"v8-fast-api"_kj});
```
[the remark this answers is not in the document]
anonrig: I've looked. It's doable. Asked @npaun to take a look (ref: https://github.com/cloudflare/workerd/issues/4089). If it's a blocker, I can add that support in a different pull-request and later update this PR.

src/workerd/jsg/jsg-test.h
kentonv: > for empty string you should do: expectEval(code, "string", "")

But if I did that, then this `if` statement would skip the value check. Perhaps the `&&` should have been `||`?

src/workerd/jsg/promise-test.c++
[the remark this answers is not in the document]
kentonv: It should apply to KJ tests too I think (#4089 only mentions wd-test).

I'm a little worried that all these `initAutogateNamesForTest` additions mean that we're now only doing these tests with fastapi enabled, and may have lost coverage of the non-fastapi path. But as long as we plan to remove these lines later once we have all-autogates (and that will happen soon), then I guess this is fine for now.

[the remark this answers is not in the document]
anonrig: That's what I had in mind. I'll remove these lines once we roll out fast api 100%.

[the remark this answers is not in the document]
kentonv: Well... maybe we actually still want to run our tests both with fastapi off and on. Otherwise we might be losing test coverage of the slow path.

src/workerd/jsg/resource.h
anonrig: My latest commit (https://github.com/cloudflare/workerd/pull/4008/commits/a4336a0e81a35b4e9c29de927b49df3ac36e98f2) should address this. Thank you for the review.

src/workerd/jsg/resource.h
kentonv, line 1: Reminder to squash before merging.

kentonv (approved):

PR #4012, opened 2025-04-23
feat(KV): fix statusText in the Workers KV tests

teresalves:
Just improving the already existing KV binding tests

### DIFF
--- a/src/workerd/api/kv-test.js
+++ b/src/workerd/api/kv-test.js
@@ -23,14 +23,30 @@ export default {
       r += decoder.decode();
       const parsedBody = JSON.parse(r);
       const keys = parsedBody.keys;
-      if (keys.length < 1 || keys.length > 100) {
-        return new Response(null, { status: 400 });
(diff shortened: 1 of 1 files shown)


fhanau (approved):

src/workerd/api/kv-test.js
```diff
@@ -150,18 +172,21 @@ export default {
...
+      message:
+        'KV GET_BULK failed: 400 "arrayBuffer" is not a valid type. Use "json" or "text"',
     });
 
     await assert.rejects(
       env.KV.get(['key-not-json', 'key2'], { type: 'banana' }),
       {
-        message: 'KV GET_BULK failed: 500 Internal Server Error',
+        message:
+          'KV GET_BULK failed: 500 "banana" is not a valid type. Use "json" or "text"',
```
fhanau, line 189: Nit: This needs to be 400 instead of 500, that should get the test (and thus CI checks) to pass.

PR #4014, opened 2025-04-23
Speed up reviews for experimental type changes

npaun:
Based on previous discussions, I know we need to ask the Wrangler team to look at generated type changes. 

I'm wondering if it would be ok for us to be able to just change experimental types, though? It seems like we're using this for internal stuff that shouldn't affect users.

Happy to close this if it's not a good idea


kentonv:
IMO requiring Wrangler review of type changes doesn't actually make sense. I asked @danlapid yesterday about why we require this and it turns out we were thinking it would prevent a certain kind of incident, but on closer inspection the incident in question didn't actually change the types, so this wouldn't have helped.

On the other hand, though, we _should_ have increased scrutiny of type changes compared to other changes, since this means our public API is changing and such changes need to be supported forever. I'm not sure exactly who should be approvers for such changes but probably someone within the runtime team. Experimental changes could probably be exempted from this.

anonrig (approved):

danlapid (approved):

PR #4015, opened 2025-04-23
Python: Set up FFI before taking snapshot

hoodmane:


src/pyodide/internal/snapshot.ts
```diff
@@ -226,6 +234,7 @@ type DylinkInfo = {
   settings?: { baselineSnapshot?: boolean };
   loadOrder?: string[];
   soMemoryBases?: { [name: string]: number };
+  version?: undefined;
```
dom96, line 237: Should this be undefined? Could it be removed?

src/pyodide/internal/snapshot.ts
```diff
@@ -322,11 +334,20 @@ function memorySnapshotDoImports(Module: Module): string[] {
...
-  const dsoJSON = recordDsoHandles(Module);
+  let json: SnapshotJson = recordDsoHandles(Module);
+  if (Module.API.version !== '0.26.0a2') {
+    const hiwire = Module.API.serializeHiwireState(Module);
+    const tmp: NewSnapshotJson = {
+      version: 1,
+      dsos: json,
+      hiwire,
+    };
+    json = tmp;
```
dom96, line 345: Nit

```suggestion
    json = {
      version: 1,
      dsos: json,
      hiwire,
    } as NewSnapshotJson;
```

src/pyodide/internal/snapshot.ts
```diff
@@ -36,6 +36,14 @@ export let SHOULD_RESTORE_SNAPSHOT = false;
...
  */
 let DSO_METADATA: DylinkInfo = {};
+export let HIWIRE_STATE: SnapshotConfig;
+
+type NewSnapshotJson = {
+  hiwire: SnapshotConfig;
+  dsos: DylinkInfo;
+  version: 1;
+};
+type SnapshotJson = DylinkInfo | NewSnapshotJson;
```
dom96, line 46: Nit: maybe this should be called `SnapshotMetadata`?

src/pyodide/internal/snapshot.ts
```diff
@@ -322,11 +334,20 @@ function memorySnapshotDoImports(Module: Module): string[] {
  * linear memory into MEMORY.
  */
 function makeLinearMemorySnapshot(Module: Module): Uint8Array {
-  const dsoJSON = recordDsoHandles(Module);
+  let json: SnapshotJson = recordDsoHandles(Module);
```
dom96, line 337: Nit: It's not JSON yet, maybe call it `snapshotMeta` or just `metadata`

src/pyodide/internal/snapshot.ts
```diff
@@ -338,15 +359,15 @@ export let LOADED_SNAPSHOT_VERSION: number | undefined = undefined;
 /**
  * Encode heap and dsoJSON into the memory snapshot artifact that we'll upload
  */
-function encodeSnapshot(heap: Uint8Array, dsoJSON: object): Uint8Array {
-  const dsoString = JSON.stringify(dsoJSON);
-  let snapshotOffset = HEADER_SIZE + 2 * dsoString.length;
+function encodeSnapshot(heap: Uint8Array, json: object): Uint8Array {
```
dom96, line 362:
```suggestion
function encodeSnapshot(heap: Uint8Array, json: SnapshotJson): Uint8Array {
```

src/pyodide/internal/topLevelEntropy/entropy_import_context.py
```diff
@@ -159,6 +160,16 @@ def patched_Random():
...
 
 
+@contextmanager
+def aiohttp_connector_context(module):
+    try:
+        ssl = sys.modules["ssl"]
+        sys.modules["ssl"] = None
+        yield
+    finally:
+        sys.modules["ssl"] = ssl
```
dom96, line 170: What does this do? maybe add a comment in here to explain

src/pyodide/types/emscripten.d.ts
```diff
@@ -8,22 +8,34 @@ interface PyodideConfig {
...
+  ) => void;
   public_api: Pyodide;
   rawRun: (code: string) => [status: number, err: string];
   initializeStreams: (
     stdin?: any,
     stdout?: (a: string) => void,
     stderr?: (a: string) => void
   ) => void;
-  version: string;
+  version: '0.26.0a2' | '0.27.5';
```
dom96, line 34: What happens when we add new versions? It may be difficult to remember to update this, maybe best to keep it as a string and add some `throw` statements when the version isn't what we expect

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -412,11 +415,14 @@ kj::Array<kj::String> PythonModuleInfo::filterPythonScriptImports(
 
     // don't include modules that we provide and that are likely to be imported by most
     // workers.
-    if (firstComponent == "js"_kj.asArray() || firstComponent == "pyodide"_kj.asArray() ||
-        firstComponent == "asgi"_kj.asArray() || firstComponent == "workers"_kj.asArray() ||
-        firstComponent == "httpx"_kj.asArray() || firstComponent == "openai"_kj.asArray() ||
-        firstComponent == "starlette"_kj.asArray() || firstComponent == "urllib3"_kj.asArray()) {
-      continue;
+    KJ_LOG(INFO, "filterPythonScriptImports", version);
```
dom96, line 418:
```suggestion

```

dom96 (approved):
A few suggestions but overall LGTM

src/pyodide/internal/snapshot.ts
hoodmane: If you do:
```ts
const a: A = expr;
```
then typescript will check that expr is a subtype of A. If you do:
```ts
const a = expr as A;
```
then typescript will not check the type of `expr`. Thus the former is to be preferred whenever it works, and using `as` is a fallback for when we can't convince typescript to type check stuff properly.

src/pyodide/types/emscripten.d.ts
hoodmane: This will make the editor autocomplete the version everywhere which is really handy. Typescript will remind us to update this when we need to.

src/pyodide/internal/snapshot.ts
hoodmane: The point is to get type narrowing to work correctly for `type SnapshotJson = DylinkInfo | NewSnapshotJson`. In particular, if I do:
```
let a = {} as SnapshotJson;
if (a.version === 1) {
   // type of a should be narrowed to `NewSnapshotJson`
} else {
   // type of a should be narrowed to `DylinkInfo`.
}
```
If I leave this out, `a.version` is a type error and no type narrowing will happen. So it's just a type hack.

src/pyodide/internal/snapshot.ts
```diff
@@ -36,6 +36,14 @@ export let SHOULD_RESTORE_SNAPSHOT = false;
  * Record the dlopen handles that are needed by the MEMORY.
  */
 let DSO_METADATA: DylinkInfo = {};
+export let HIWIRE_STATE: SnapshotConfig;
```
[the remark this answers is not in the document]
hoodmane: There's a big mess here. I'll make a followup PR to clean the mess up.

But I don't think the globals themselves are so bad. This file is attached to the life time of the isolate and of the Pyodide runtime, so globals are okay. They aren't global to edgeworker, they are attached to a specific worker replica, and are effectively state of that worker's Python runtime.

src/pyodide/internal/snapshot.ts
hoodmane: Makes sense to me.

PR #4017, opened 2025-04-23
Local dev: allow tailing the same worker

danlapid:
This gets around the current limitation by allowing exactly one extra layer of depth. 
This should be enough for any real use case. 
Tail workers can now tail the same worker but they will not tail themselves. 
I believe this is what users would expect and is similar to what they get in prod.

### DIFF
--- a/src/workerd/api/BUILD.bazel
+++ b/src/workerd/api/BUILD.bazel
@@ -360,6 +360,14 @@ wd_test(
     ],
 )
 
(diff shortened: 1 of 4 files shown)


fhanau:
Will review later today. Also CC @jmorrell-cloudflare

fhanau (approved):
LGTM, reasonably elegant and it deduplicates the "Caution here..." comment which I had meant meant to do earlier.

PR #4025, opened 2025-04-25
[nfc] Use Request::sendIgnoringResult() to elide promise continuations

fhanau:
This eliminates some memory allocations for promise continuations and improves code size.
See capnp PR for context: https://github.com/capnproto/capnproto/pull/2286

### DIFF
--- a/build/deps/gen/dep_capnp_cpp.bzl
+++ b/build/deps/gen/dep_capnp_cpp.bzl
@@ -2,11 +2,11 @@
 
 load("@//:build/http.bzl", "http_archive")
 
-URL = "https://github.com/capnproto/capnproto/tarball/ff191c66c2bbff428c10923b5d4032087b592fe0"
-STRIP_PREFIX = "capnproto-capnproto-ff191c6/c++"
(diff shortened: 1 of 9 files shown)


fhanau:
> @fhanau I think there are a couple cases that can be changed in `basics.c++`

Can you elaborate? This PR focuses on folding promise continuations together, for example `send().ignoreResult()` is replaced by the new and more efficient `sendIgnoringResult()`, `then(<func A>).catch_(<func B>)` can be reduced to `then(<func A>, <func B>)`. Do you see candidates for that in basics.c++?

npaun:
@fhanau Yes, I introduced a couple of `send().ignoreResult()` there lately. I can just push a commit to take care of it.

npaun (approved):

fhanau:
> @fhanau Yes, I introduced a couple of `send().ignoreResult()` there lately. I can just push a commit to take care of it.

I see! I guess those probably just hadn't landed yet when I looked at converting to sendIgnoringResult(). Changes LGTM, and save another 0.02% from the downstream binary size 🤑

npaun:
> Changes LGTM, and save another 0.02% from the downstream binary size 🤑

A penny saved is a penny earned

PR #4026, opened 2025-04-25
log when workerEntrypoint waiting on context takes longer than 5s

Frederik-Baetens:


src/workerd/io/worker-entrypoint.c++
```diff
@@ -286,10 +290,22 @@ kj::Promise<void> WorkerEntrypoint::request(kj::HttpMethod method,
   TRACE_EVENT_BEGIN("workerd", "WorkerEntrypoint::request() waiting on context",
       PERFETTO_TRACK_FROM_POINTER(&context), PERFETTO_FLOW_FROM_POINTER(this));
 
+  // Using a DurationExceededLambda here instead of a DurationExceededLogger to avoid doing a string
+  // heap allocation in the happy path where the duration isn't exceeded.
```
Frederik-Baetens, line 294: Rationale for introducing the DurationExceededLambda is here, but let me know if this rationale doesn't make sense, then I'll use the simpler DurationExceededLogger here. Just wanted to avoid heap allocations here since this is a hot path.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -111,6 +113,8 @@ class WorkerEntrypoint final: public WorkerInterface {
       kj::Date scheduledTime,
       uint32_t retryCount);
 
+  const kj::MonotonicClock& monotonicClock = kj::systemPreciseMonotonicClock();
```
kentonv, line 116: Minor, but it doesn't seem worth wasting a member variable on this, it is used no more than once per instance.

kentonv (commented):
This basically measures the isolate lock wait time, but only for the first lock of an HTTP request -- not any subsequent locks, nor non-HTTP requests.

For Durable Objects specifically, though, I suspect this will never show a long wait, because all DOs for the same namespace run on the same thread, so there's no lock contention. Instead, messages queue up in the socket buffer waiting for the thread to process them.

So I'm not sure if placing this specific log in this specific spot will actually help much. And anyway, lock waits are already part of traces...

Frederik-Baetens:
> So I'm not sure if placing this specific log in this specific spot will actually help much. And anyway, lock waits are already part of traces...

I found traces with really long pre_responses but no significant lock wait. That usually indicates the user worker just executing for a long time, but we saw this even on basic debug endpoints that didn't really do anything. Will dm you.

src/workerd/io/worker-entrypoint.c++
Frederik-Baetens: :+1: I'll just get the clock when instantiating the durationExceededLambda then.

kentonv:
> I found traces with really long pre_responses but no significant lock wait.

Well, the logging you're adding here is measuring the time between context.run() starting, and the callback being invoked. Two things happen in that time:
* Acquire the actor's input lock.
* Acquire the isolate lock.

The isolate lock would be reflected in lock_wait, but you say your traces don't show that. The input lock could be blocked due to a slow CRDB query, or due to the app using blockConcurrencyWhile() and taking a long time.

Do we have trace spans for the input lock acquisition? If not, maybe we should add that? Wouldn't that be better than this ad hoc logging?

Frederik-Baetens:
> The input lock could be blocked due to a slow CRDB query

What query is that? I'd like to perhaps add specific logging there too.

kentonv:
> What query is that? Just the getOrCreate query? Maybe I can add specific logging there too.

I meant read operations performed by the application, which take the input lock to prevent concurrent requests in the meantime (unless the application requests otherwise).

PR #4028, opened 2025-04-25
Support JSPI in durable objects

hoodmane:


build/python_metadata.bzl
```diff
@@ -59,6 +59,9 @@ def make_bundle_version_info(versions):
         entry["feature_flags"] = [entry["flag"]]
         result[name] = entry
     dev = result["development"]
+
+    # Uncomment to test with development = 0.27.5
+    # dev["real_pyodide_version"] = "0.27.5"
```
danlapid, line 64: Remove?

hoodmane: I want to keep this because it will make development faster for me.

src/workerd/server/tests/python/durable-object/worker.py
```diff
@@ -20,6 +24,12 @@ async def no_args_method(self):
     async def args_method(self, arg):
         return "value from python " + arg
 
+    def jspi_method(self, arg):
```
danlapid, line 27: Async def

hoodmane: It's intentionally not async since it's testing JSPI. It should be supported to define synchronous methods on DO's right?

danlapid: Nope
All DO exported functions used over jsrpc should be async

danlapid: It probably works because we have some async wrapper around it

danlapid (approved):

PR #4029, opened 2025-04-25
Implement in-memory worker file system

jasnell:
So this is a fun one. In preparation for the implementation of node:fs (and eventually also the Web file system API), we need to implement a virtual file system for the worker. Consider this a WIP sketch for the time being. There are still a range of things to do and verify before this is ready.

There's quite a bit to unpack here so let's go through it a bit.

First, let's talk about the file system itself. Every worker instance will have its own root directory (`/`). In this root directory we will have at least three special directories, the "bundle" root, the "dev" root, and the "temp" root.

* The bundle root is where all of the modules that are included in the worker bundle will be accessible. Everything in this directory will be strictly read-only. The contents are populated by the worker configuration bundle. By default, the bundle root will be `/bundle` but this can be overridden.

* The dev root provides the equivalent to `/dev/null`, `/dev/zero`, `/dev/full`, and `/dev/random`.

* The temp root is where we will allow temporary files to be created. Everything in this directory is read-write but will be transient. By default, the temp root will be `/tmp` but this can also be overridden.

Let's imagine the following simple workerd configuration:

```
  modules = [
    (name = "worker",
     esModule = embed "worker.js"),
    (name = "foo", text = "Hello World!"),
  ],
  compatibilityDate = "2023-02-28",
);
```

Given this configuration, the worker fs will initially have the following structure:

```
/
├── bundle
│   ├── worker
│   └── foo
├── dev
│   ├── null
│   ├── zero
│   ├── full
│   └── random
└── tmp
```

We can then access the file system at the C++ level.

```cpp
jsg::Lock& js = ..
KJ_IF_SOME(vfs, workerd::VirtualFileSystem::tryGetCurrent(js)) {
  auto dir = vfs.getRoot(js);
  KJ_IF_SOME(item, vfs.resolve(js, "file:///a/b/c"_url")) {
    KJ_SWITCH_ONEOF(item) {
     KJ_CASE_ONEOF(file, kj::Rc<workerd::File>) { /* ... */ }
     KJ_CASE_ONEOF(dir, kj::Rc<workerd::Directory>) { /* ... */ }
    }
  }
}
```

The temporary file directory is a bit special in that the contents are fully transient based on whether there is or is not an active IoContext. We use a special RAII `TmpDirStoreScope` scope to manage the contents of the temporary directory. For example,

```cpp
    TempDirStoreScope temp_dir_scope;
    auto tmpFile = dir->tryOpen(js, path, workerd::FsType::FILE);
    KJ_ASSERT(tmpFile.write(js, 0, "Hello World!"_kjb) == 12);
    KJ_ASSERT_NONNULL(dir->stat(js, path));  // exists!
  }
  // The temp dir scope is destructed and the file is deleted
  KJ_ASSERT(dir->stat(js, path) == kj::none);
}
```

If there is an active IoContext, the temporary file wil be created within that IoContext's TmpDirStoreScope, and will be deleted when the IoContext is destructed. This allows us to have a single virtual file system whose temporary directories are either deleted immediately as soon as the execution scope is exited, or are specific to the IoContext and are deleted when the IoContext is destructed. This mechanism allows us to have multiple IoContexts active at the same time while still having a single virtual file system whose contents correctly reflect the current IoContext.

Temporary files can be created, copied, etc only within the temporary directory. All other directories are read-only.

When there is no active IoContext, all temporary files will have a timestamp set to the Unix epoch. When there is an active IoContext, the temporary files will acquire the current timestamp from the IoContext ensuring that the file system's view of time is consistent within a request.

The design here is intended to be extensible. We can add new root directories in the future with different semantics and implementations. For example, to support python workers we can introduce a new root directory that is backed, for instance, by a tar/zip file containing the python standard library, etc.

What isn't implemented yet? Great question! The following todos are remaining:

* Implementing "file descriptors". Currently the file system will return `kj::none` for all file descriptors. As a next step, the file system will implement a custom file descriptor counter that will ensure that all files are assigned a unique monotonically increasing file descriptor. When the counter reaches a maximum threshold of opened files, the worker will be condemned. We can also just track the number of active `kj::Own<const kj::File>` objects and use that internally for accounting but the `node:fs` impl does a lot with integer fds so we will need them at some point for that.

The implementation is split across several files/components:

`worker-fs.h/c++` - These provide the main interfaces for the worker file system.

`bundle-fs.h/c++` - These provide the interfaces for interfacing the workerd configuration with the worker file system. A similar interface will need to be provided for the internal repo since the two use different configuration schemas.


jasnell:
Although this is marked draft, it is ready for an initial round of reviews.

src/workerd/io/io-context.h
```diff
@@ -839,6 +863,9 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
 
   kj::Own<WeakRef> selfRef = kj::refcounted<WeakRef>(kj::Badge<IoContext>(), *this);
 
+  IoContextClock clock;
+  kj::Own<workerd::TempDirStoreScope> tempDirStoreScope;
```
danlapid, line 867:
```suggestion
  kj::Own<TempDirStoreScope> tempDirStoreScope;
```

src/workerd/io/worker-fs-test.c++
```diff
@@ -0,0 +1,120 @@
+#include "worker-fs.h"
+
+#include <kj/debug.h>
+#include <kj/test.h>
+
+namespace workerd {
```
danlapid, line 6: nit: using `workerd::` everywhere in this file even though you are in the `workerd` namespace

src/workerd/io/worker-fs-test.c++
```diff
@@ -0,0 +1,120 @@
...
+  KJ_EXPECT(knownRoot == workerd::FsMap::KnownRoot::BUNDLE);
+
+  knownRoot = fsMap->tryGetKnownRootForUrl("file:///mything/temp"_url);
+  KJ_EXPECT(knownRoot == workerd::FsMap::KnownRoot::TEMP);
+
+  knownRoot = fsMap->tryGetKnownRootForPath(kj::Path({"mything", "bundle"}));
+  KJ_EXPECT(knownRoot == workerd::FsMap::KnownRoot::BUNDLE);
+
+  knownRoot = fsMap->tryGetKnownRootForPath(kj::Path({"mything", "temp"}));
+  KJ_EXPECT(knownRoot == workerd::FsMap::KnownRoot::TEMP);
```
danlapid, line 43: What about tests for what should not exist, and for error cases

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,192 @@
...
+    kj::Own<FsMap> finish();
+
+   private:
+    kj::Own<FsMap> map;
+  };
+
+  static kj::Own<FsMap> createDefault();
+
+ private:
+  kj::HashMap<KnownRoot, jsg::Url> paths;
```
danlapid, line 65: If `KnownRoot` is likely to stay finite and small (and is already an int-based value starting with 0) you'll be better off with a simple `kj::Array<kj::Maybe<jsg::Url>>(KnownRoot::MAX_VAL)`

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,703 @@
...
+      for (auto& entry: *locked) {
+        names.add(kj::str(entry.key));
+      }
+      return names.releaseAsArray();
+    }
+
+    kj::Array<kj::ReadableDirectory::Entry> listEntries() const override {
+      auto locked = entries.lockShared();
+      // List the immediate contents of the directory.
+      kj::Vector<kj::ReadableDirectory::Entry> entriesList(locked->size());
```
danlapid, line 182:
```suggestion
      return KJ_MAP(entry, *locked) {
        return kj::ReadableDirectory::Entry{
          .type = kj::FsNode::Type::DIRECTORY,
          .name = kj::str(entry.key),
        };
      };
```

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,192 @@
...
+      .spaceUsed = 0,
+      .lastModified = kj::UNIX_EPOCH,
+      .linkCount = 1,
+      .hashCode = 0,
+    };
+  }
+
+  virtual bool exists(kj::PathPtr path) const = 0;
+
+  virtual kj::Array<kj::String> listNames() const = 0;
```
danlapid, line 106: `listNames` seems redundant and not significantly more efficient than `listEntries`

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,192 @@
...
+
+  virtual kj::Array<kj::ReadableDirectory::Entry> listEntries() const = 0;
+
+  virtual kj::Maybe<kj::FsNode::Metadata> tryLstat(kj::PathPtr path) const = 0;
+
+  virtual kj::Maybe<kj::Own<const kj::ReadableFile>> tryOpenFile(kj::PathPtr path) const = 0;
+
+  virtual kj::Maybe<kj::Own<const kj::ReadableDirectory>> tryOpenSubdir(kj::PathPtr path) const = 0;
+
+  virtual kj::Maybe<kj::Own<const kj::File>> tryOpenFile(
```
danlapid, line 116: This reads odd to me.
We can open a file in writable mode on a const DirectoryDelegate and then get a const File and write to that file?

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,192 @@
...
+      kj::PathPtr path, kj::WriteMode mode) const {
+    return kj::none;
+  }
+
+  virtual kj::Maybe<kj::Own<kj::AppendableFile>> tryAppendFile(
```
danlapid, line 121: How can this be const if WriteMode can be  CREATE

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,192 @@
...
+      kj::PathPtr path, kj::WriteMode mode) const {
+    return kj::none;
+  }
+
+  virtual kj::Maybe<kj::Own<const kj::Directory>> tryOpenSubdir(
+      kj::PathPtr path, kj::WriteMode mode) const {
+    return kj::none;
+  }
+
+  virtual bool tryRemove(kj::PathPtr path) const {
```
danlapid, line 131:
```suggestion
  virtual bool tryRemove(kj::PathPtr path) {
```

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,192 @@
...
+    return false;
+  }
+
+  virtual kj::Own<const kj::File> createTemporary() const {
```
danlapid, line 135:
```suggestion
  virtual kj::Own<const kj::File> createTemporary() {
```

src/workerd/io/worker.c++
```diff
@@ -1494,7 +1494,7 @@ kj::Maybe<jsg::JsObject> tryResolveMainModule(jsg::Lock& js,
     // call to tryResolveModuleNamespace because I intend to add some additional
     // logging/metrics logic around this call.
     KJ_IF_SOME(ns,
-        jsg::modules::ModuleRegistry::tryResolveModuleNamespace(js, mainModule.toString(true))) {
```
danlapid, line 1497: Why?

src/workerd/jsg/setup.h
```diff
@@ -345,7 +352,8 @@ class IsolateBase {
 
   explicit IsolateBase(V8System& system,
       v8::Isolate::CreateParams&& createParams,
-      kj::Own<IsolateObserver> observer);
+      kj::Own<IsolateObserver> observer,
+      kj::Maybe<VfsCallback> vfsCallback = kj::none);
```
danlapid, line 356: Any place where we need this to be none? if this is for tests maybe we can just create a mock callback that returns none for every path.
```suggestion
      kj::Maybe<VfsCallback> vfsCallback);
```

src/workerd/server/workerd-api.h
```diff
@@ -38,7 +38,8 @@ class WorkerdApi final: public Worker::Api {
       kj::Own<JsgIsolateObserver> observer,
       api::MemoryCacheProvider& memoryCacheProvider,
       const PythonConfig& pythonConfig,
-      kj::Maybe<kj::Own<jsg::modules::ModuleRegistry>> newModuleRegistry);
+      kj::Maybe<kj::Own<jsg::modules::ModuleRegistry>> newModuleRegistry,
+      kj::Maybe<jsg::VfsCallback> vfsCallback = kj::none);
```
danlapid, line 42:
```suggestion
      kj::Maybe<jsg::VfsCallback> vfsCallback);
```

src/workerd/io/worker-fs.h
jasnell: Modeled directly from kj/filesystem, with an implementation based on what is in kj also.

For instance, here's KJ's definition of `tryAppendFile...`

```
  virtual Maybe<Own<AppendableFile>> tryAppendFile(PathPtr path, WriteMode mode) const = 0;
```

src/workerd/io/worker-fs.h
jasnell: -1 ... again, modeled this directly from `kj/filesystem` to remain as consistent as possible.

src/workerd/io/worker-fs.h
jasnell: -1... again based off kj/filesystem.h

src/workerd/server/server.c++
```diff
@@ -3087,18 +3087,55 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
...
+  // This callback is used to resolve file system URLs to the actual
+  // underlying file system nodes.
+  auto fsMapProvider = [workerFs = kj::mv(workerFs)](jsg::Lock&,
+                           const jsg::Url& url) -> kj::Maybe<kj::Own<const kj::FsNode>> {
+    kj::Path root{};
+    kj::String pathname = kj::str(url.getPathname().slice(1));
+    kj::Path path = root.eval(pathname);
+
+    // First, try to find it as a file.
+    KJ_IF_SOME(file, workerFs->getRoot().tryOpenFile(path)) {
```
danlapid, line 3107: This means you're iterating the list of entries twice, seems like this might be a clue that the real API we need is `tryOpenNode` that returns a OneOf.

src/workerd/jsg/setup.h
jasnell: For now this is set as kj::none in order to keep it from failing on the internal repo where this is not yet implemented. Once we have the internal implementation then this can be updated.

src/workerd/io/worker.c++
jasnell: `mainModule.toString(true)` means the generated string will have a leading `/`, which we don't actually want here. This was a bug found while I was working in here.

src/workerd/server/workerd-api.h
jasnell: For now, until the internal repo PR is opened that provides this, I'm going to keep the default values here.

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,433 @@
...
+        .hashCode = 0,
+      };
+    }
+
+    void sync() const override {};
+    void datasync() const override {};
+
+    size_t read(uint64_t offset, kj::ArrayPtr<kj::byte> buffer) const override {
+      auto locked = content.lockShared();
+      if (buffer.size() == 0) return 0;
```
danlapid, line 383: This can be moved to before the lock

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,433 @@
...
+          src = str.asBytes();
+        }
+        KJ_CASE_ONEOF(arr, kj::ArrayPtr<const kj::byte>) {
+          src = arr.asBytes();
+        }
+      }
+
+      if (src.size() == 0 || offset >= src.size()) return 0;
+
+      auto view = src.slice(offset);
```
danlapid, line 396:
```suggestion
      auto view = src.slice(offset, kj::min(offset+buffer.size(), src.size()));
      if (view.size() == 0) return 0;
      buffer.copyFrom(view);
      return view.size();
```

src/workerd/io/worker-fs.h
danlapid: kj::Filesystem works under different constraints, if you are not inheriting from it I don't see a reason to keep an API surface that isn't a good fit for what we need to do here.

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,433 @@
...
+  BundleDirectoryDelegate(config::Worker::Reader conf)
+      : inner(capnp::clone(conf)),
+        root(kj::atomicRefcounted<DirectoryEntry>()) {
+
+    kj::Path rootPath{};
+    DirectoryEntry* current = root.get();
+
+    switch (inner->which()) {
+      case config::Worker::Which::MODULES: {
+        for (auto module: inner->getModules()) {
```
danlapid, line 25: This will increase the startup time of all scripts, even those who might never us the filesystem.
We should set up lazy initialization for this type and make sure nothing in our codebase eagerly initializes the bundle filesystem unless it is actually used.
Either that or figure out a different data structure that lets us have an O(1) initialization).

src/workerd/io/worker-fs.h
jasnell: Agreed but this is modeled directly after kj/filesystem which separates these. Since our implementation puts a `kj::ReadableDirectory` behind this API it's better to remain consistent.

src/workerd/io/worker-fs.h
jasnell: The `DirectoryDelegate` does not inherit from `kj::ReadableDirectory` but it does directly abstract it in at least the temp directory implementation. For now I'd rather keep these as is and we can refine later as we go.

src/workerd/server/bundle-fs.c++
jasnell: Given that this particular class is only used in workerd I'm not all that concerned with the additional cost. A separate version of this will need to be implemented in the internal repo using the different bundle configuration format. That one will have more performance optimizations.

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,433 @@
...
+// if the bundle contains a module with the name "foo/bar/baz", it will be
+// represented as a directory "foo" with a subdirectory "bar" and a file
+// "baz" inside it. The directory structure and files are read-only. All
+// timestamps are set to the Unix epoch.
+class BundleDirectoryDelegate final: public workerd::DirectoryDelegate,
+                                     public kj::AtomicRefcounted {
+ public:
+  BundleDirectoryDelegate(config::Worker::Reader conf)
+      : inner(capnp::clone(conf)),
+        root(kj::atomicRefcounted<DirectoryEntry>()) {
```
danlapid, line 18: Are we really expecting that certain DirectoryEntries can outlive the general BundleDirectoryDelegate?
Is it even safe? They don't have their own copy of the conf, merely a reader, a read from a file that outlived BundleDirectoryDelegate would read from an outlived data-source.

All I mean to say is that this data structure contains way too many ARCs in my opinion considering that everything in here is by definition const and immutable.

I think it should be enough to say that since `BundleDirectoryDelegate` is born with the isolate and gets destructed with the isolate then there's no need for its members to be ARCs and no need for anyone to ever have an owned reference to them, merely a bare reference should be enough.

src/workerd/server/bundle-fs.c++
danlapid: The internal and the external versions don't really need to be all that different, good code deduplication would have them be very thin wrappers around the same object.
I disagree with punting this question for later, a good implementation should have this thought through.

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    kj::Own<const kj::FsNode> cloneFsNode() const override {
+      return kj::atomicAddRef(*this);
+    }
+
+    bool exists(kj::PathPtr path) const override {
+      auto locked = entries.lockShared();
+      if (path.size() == 0) return true;
+      KJ_IF_SOME(found, locked->find(path[0])) {
```
danlapid, line 161: This path finding logic is duplicated in a lot of different places with only small modifications.
I think this can be generalized in a private method that just returns a Maybe<OneOf<FileEntry,DirectoryEntry>> or OneOf<void, FileEntry, DirectoryEntry> and each of these methods can use it and act appropriately  based on the result.

src/workerd/server/bundle-fs.c++
jasnell: While I generally agree, there is *nothing* in `workerd/server` that is reused in the internal repo. The two projects use two entirely separate configuration schemas. It's not about punting the question for later it's that I have yet to do that internal implementation. Once I do, there's an opportunity to separate out and deduplicate but right now there's nothing to deduplicate yet.

src/workerd/server/bundle-fs.c++
jasnell: Yep, already in plan. Will be done before this is moved out of draft.

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,703 @@
...
+  // If there is neither a current IoContext nor a TempDirStoreScope,
+  // then the delegate acts like a normal empty read-only directory.
+  class TempDirDelegate final: public DirectoryDelegate, public kj::AtomicRefcounted {
+   public:
+    kj::Own<const DirectoryDelegate> clone() const override {
+      return kj::atomicAddRef(*this);
+    }
+
+    bool exists(kj::PathPtr path) const override {
+      if (IoContext::hasCurrent()) {
```
danlapid, line 363: Seems like 
```
      if (IoContext::hasCurrent()) {
        return IoContext::current().getTempDirStoreScope();
      }

      if (TempDirStoreScope::hasCurrent()) {
        return TempDirStoreScope::current();
      }

      return kj::none;
```
can be deduplicated for readability.

danlapid:
I'll stop reviewing at this point as there seems to be too much up in the air at this stage.
My general sentiment is that this is quite a bit overcomplicated by the API surface it seems to mimic.
I don't know if it's actually serving us.

Also, considering we've already implemented everything in C++ anyway and left very little for the imagination. I would love to see this interfaced as an API type for easy interfacing from different languages (Node,Python,Rust etc.)

danlapid:
Another thing of note, I think we can do well with having the code split up to
1. Directory structure and delegates interfaces + implementation for finding the right delegate and the root fs.
2. Temp filesystem header + implementation.
3. Bundle filesystem header + implementation.

Then the bundle filesystem should find a good common interface between the internal and the external configs so that as much as possible can be shared and startup time isn't increased significantly.
Off the cuff I think that a possible interface would be receiving two arrays of the same size:
```
kj::Array<capnp::Text::Reader> paths;
kj::Array<kj::OneOf<capnp::Text::Reader, capnp::Bytes::Reader>> data;
```

This should be easy and quick to produce both internally and externally with a single iteration and pretty much zero copies.
Then we can build the tree structure on usage.

jasnell:
@danlapid et. al ... the fixup commit here significantly refactors and hopefully simplifies the implementation here, moving away from the dependence on kj::Filesystem

src/workerd/server/server.c++
```diff
@@ -3087,18 +3088,47 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
   auto observer = kj::atomicRefcounted<IsolateObserver>();
   auto limitEnforcer = kj::refcounted<NullIsolateLimitEnforcer>();
 
+  // Create the FsMap that will be used to map known file system
+  // roots to configurable locations.
+  // TODO(node-fs): This is set up to allow users to configure the "mount"
```
danlapid, line 3093: Nit: I would remove that from a TODO() and wait for someone to actually request the feature.
TODOs have a tendency to mount up and shouldn't document possible feature requests in my opinion but rather things that we think we should do eng-wise.

src/workerd/server/server.c++
```diff
@@ -3087,18 +3088,47 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
...
+  // points for known roots but we currently do not expose that in the
+  // config. So for now this just uses the defaults.
+  auto fsMap = kj::heap<FsMap>();
+
+  // This callback is used to resolve file system URLs to the actual
+  // underlying file system nodes.
+  auto vfsProvider = [workerFs = kj::Maybe<kj::Own<jsg::fs::VirtualFileSystem>>(), &fsMap = *fsMap,
```
danlapid, line 3100: This kinda looks funky and smells in my opinion.
I think this entire thing should just be a virtual method of the API base-class that we override in both workerd and internally. 
We can add a utility method to Worker::Lock to simplify the API.
Then anyone who needs to access the filesystem can access that.

I think setting it on the jsg isolate makes for an odd design with callbacks that take objects by reference and lazily allocated fields that may as well be members.

src/workerd/server/server.c++
```diff
@@ -3087,18 +3088,47 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
...
+    return KJ_ASSERT_NONNULL(workerFs)->getRoot(js).tryOpen(js, path);
+  };
+
   kj::Maybe<kj::Own<jsg::modules::ModuleRegistry>> newModuleRegistry;
   if (featureFlags.getNewModuleRegistry()) {
     KJ_REQUIRE(experimental,
         "The new ModuleRegistry implementation is an experimental feature. "
         "You must run workerd with `--experimental` to use this feature.");
+
+    // We use the same path for modules that the virtual file system users.
```
danlapid, line 3119: nit:
```suggestion
    // We use the same path for modules that the virtual file system uses.
```

src/workerd/server/server.c++
danlapid: To piggy back off of that.
I feel like this entire feature fits better in the api directory/concept rather than in jsg.
There is some interaction with the module registry but mostly as a user rather than co-dependencies.
I think we could do with moving moving the logic out of jsg and into the api layer.

src/workerd/server/server.c++
jasnell: Kenton has specifically indicated that he'd like to see this, so I think that qualifies here?

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+class Node: public kj::Refcounted {
+ public:
+  using Type = Node_Type;
+
+  Node() = default;
+  KJ_DISALLOW_COPY_AND_MOVE(Node);
+  virtual ~Node() noexcept(false) = default;
+
+  // Metadata about this filesystem node
+  struct Stat final {
```
danlapid, line 23: Nit: I prefer to avoid structs/classes in classes where possible.
Specifically here, you've already set the fs namespace.
I think stat even makes more sense as `fs::Stat` than `fs::Node::Stat`

src/workerd/jsg/jsg.h
```diff
@@ -2325,6 +2328,21 @@ class ExternalMemoryAdjustment final {
   void maybeDeferAdjustment(ssize_t amount);
 };
 
+// ======================================================================================
+// An abstraction of a virtual file system node.
+class Url;
+namespace fs {
+enum class Node_Type {
```
danlapid, line 2335: Only used in filesystem.h, why not define it there?

src/workerd/jsg/jsg.h
```diff
@@ -34,6 +34,9 @@ using kj::uint;
 typedef long long ssize_t;
 #endif
 
+namespace kj {
+class PathPtr;
```
danlapid, line 38: Redundant forward declaration?

src/workerd/jsg/jsg.h
```diff
@@ -2325,6 +2328,21 @@ class ExternalMemoryAdjustment final {
   void maybeDeferAdjustment(ssize_t amount);
 };
 
+// ======================================================================================
+// An abstraction of a virtual file system node.
+class Url;
```
danlapid, line 2333: Redundant forward declaration?

src/workerd/jsg/jsg.h
```diff
@@ -2325,6 +2328,21 @@ class ExternalMemoryAdjustment final {
...
+namespace fs {
+enum class Node_Type {
+  FILE,
+  DIRECTORY,
+};
+
+class File;
```
danlapid, line 2340: Redundant forward declaration?
esp if we move filesystem out of the already overloaded jsg layer.

src/workerd/jsg/jsg.h
```diff
@@ -2829,6 +2846,12 @@ class Lock {
   // This variation includes modules from the worker bundle.
   kj::Maybe<JsObject> resolveModule(kj::StringPtr specifier);
 
+  // Used for the underlying implementation of the file system API
+  // Only file:// URLs are supported. If the target does not exist
+  // and createAs is not set, kj::none is returned. If createAs is
+  // set, the target is created as the given type then returned.
+  kj::Maybe<kj::OneOf<kj::Own<fs::File>, kj::Own<fs::Directory>>> resolveVfsNode(const Url& path);
```
danlapid, line 2853: Since both fs::File and fs::Directory inherit from fs::Node, why not just return fs::Node?

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+    void addPath(kj::PathPtr path, kj::Own<Directory>&& dir);
+    void addPath(kj::PathPtr path, kj::Own<File>&& file);
+
+    // Finalizes and returns the directory. The directory will not be writable.
+    // The directory will not be initially included in a directory. To add it
+    // to a directory, use the transfer method.
+    kj::Own<Directory> finish();
+
+    using Map = kj::HashMap<kj::String,
+        kj::OneOf<kj::Own<File>, kj::Own<Directory>, kj::Own<Directory::Builder>>>;
```
danlapid, line 225: `kj::OneOf< kj::Own<File>, kj::Own<Directory>>` could be more cleanly represented as `kj::Own<Node>`

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+
+    void addPath(kj::PathPtr path, kj::Own<Directory>&& dir);
+    void addPath(kj::PathPtr path, kj::Own<File>&& file);
+
+    // Finalizes and returns the directory. The directory will not be writable.
+    // The directory will not be initially included in a directory. To add it
+    // to a directory, use the transfer method.
+    kj::Own<Directory> finish();
+
+    using Map = kj::HashMap<kj::String,
```
danlapid, line 224: Haven't read all the code yet but I really hope we can get around having DirectoryBuilder in this OneOf.
Makes sense to me that only finished directories can be added to a directory.
Though honestly, I don't really see why we need the builder at all anyway.
I don't see how it actually makes us any more efficient, from all I gather it's actually doing the opposite.

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+    // The creation time of the node.
+    kj::Date created = kj::UNIX_EPOCH;
+
+    // Indicates if the node is writable.
+    bool writable = false;
+  };
+
+  // Attempt to set the last modified time of this node. If the node is
+  // read-only, this will be a non-op.
+  virtual void setLastModified(Lock& js, kj::Date date = kj::UNIX_EPOCH) = 0;
```
danlapid, line 42: Nit: Doesn't feel like this should have a default argument

src/workerd/jsg/filesystem.h
jasnell: The difficulty comes in the fact that the "bundle" fs might include nested paths that are built-up incrementally.

For instance, imagine if I have a bundle configuration with the following:

* Name: `/a/b/c/foo` Text module
* Name: `/a/b/d/bar` Text module
* Name: `/a/b/c/baz` Text module

In our completed bundle directory structure we need:

```
/ 
 |--- a
       |--- b
             |--- c
             |     |-- foo
             |     |-- baz
             | 
             |--- d
                   |-- bar
```

All of the directories here (`/a`, `/a/b`, `/a/b/c`, and `/a/b/d`) are created as read-only and are built up incrementally using the builder... so the Builder acts as a stand-in until the full set of entries is complete and then it is flipped into the constructed read-only directory. 

That said, I agree that it's a bit awkward. I wanted to focus on the main model then come back to this to find a better solution, so you're certainly correct to call it out ;-)

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+// A file is writable if it owns its contents. In such cases, the memory
+// allocation is tracked by the corresponding isolate memory accounting.
+class File: public Node {
+ public:
+  kj::Own<File> addRef(jsg::Lock& js) {
+    return kj::addRef(*this);
+  }
+
+  // Reads all the contents of the file as a string. The operation will
+  // fail if the contents exceed a 16 MB limit.
```
danlapid, line 64: Why the limit? let the customers have their footguns.

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+  kj::String readAllText(Lock& js) const;
+
+  // Reads all the contents of the file as a Uint8Array. The operation will
+  // fail if the contents exceed a 16 MB limit.
```
danlapid, line 68: same, remove the limit

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+  // Writes data to the file at the given offset. Returns the number of bytes
+  // written.
+  virtual size_t write(Lock& js, size_t offset, kj::ArrayPtr<const kj::byte> data) = 0;
+
+  size_t write(Lock& js, size_t offset, kj::StringPtr data) {
+    return write(js, offset, data.asBytes());
+  }
+
+  // Zero out the contents of the file.
+  virtual void zero(Lock& js, kj::Maybe<size_t> offset = kj::none) = 0;
```
danlapid, line 96: seems redundant and if not, seems like it can be polyfilled with write()?

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+
+  // Copies this file into a new orphaned file that is writable.
+  virtual kj::Own<File> cloneAsWritable(Lock& js) const = 0;
```
danlapid, line 99: `cloneAsWritable` `newWritable` `newReadable` seem odd, is there any practical use-case for them you have in mind?
Can't we keep these as methods on a directory receiving a filename?

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // Return the number of entries in this directory. If typeFilter is provided,
+  // only entries matching the given type will be counted.
+  virtual size_t count(Lock& js, kj::Maybe<Type> typeFilter = kj::none) const = 0;
+
+  // Iterate over the entries in this directory. If typeFilter is provided,
+  // only entries matching the given type will be returned.
+  virtual void forEach(Lock& js,
+      kj::FunctionParam<bool(Lock& js, kj::StringPtr, const Stat stat)> fn,
```
danlapid, line 145: Why do we need this to receive a callback?
Can't we just as well turn Directory into an iterator and have a C++ idiomatic interface for iterating it?

danlapid: Should be as simple as begin() {return entries.begin()} end(){return entries.end()}

jasnell: Other than the fact that I despise implementing to the C++ idiomatic iterator API? No reason I suppose ;-)

src/workerd/jsg/filesystem.h
jasnell: I do but I'm not convinced it's worthwhile. KJ has this interesting concept of `createTemporary()` which creates a fully transient in-memory, anonymous file that can be used independently of a directory. It's actually quite nice for certain use cases. This was meant to emulate that same basic concept but it is rather odd and unfamiliar. tl;dr the idea it to be able to create and work with a file *before* committing to add it to a specific directory. But, node:fs doesn't allow for that idea so we likely don't actually need it.

src/workerd/jsg/filesystem.h
jasnell: That would require at least having an allocation holding the zeroes. Having this as a separate api is a lot more efficient for what is possibly a fairly common case.

src/workerd/jsg/filesystem.h
jasnell: The key challenge is overcommitting memory before our isolate heap limit has a chance to kick in. However, I think we can avoid that simply by skipping the step of writing to a kj::String and going straight to writing a JS string... similar to the readAllBytes() variant allocating a v8::BackingStore, that would ensure that the heap limit is enforced in the copy operation. Routing this through kj::String just means we are creating an extra allocation and copy anyway.

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,245 @@
...
+
+namespace {
+
+// Thread-local storage to track the current temp directory storage scope
+// on the stack.
+static thread_local TempDirStoreScope* tempDirStorageScope = nullptr;
+
+class TempDirectory final: public jsg::fs::Directory {
+ public:
+  void setLastModified(jsg::Lock& js, kj::Date date = kj::UNIX_EPOCH) override {
```
anonrig, line 19: I don't think this should have a default value, if it should than, we should have a separate function called setLastModifiedToNow()

```suggestion
  void setLastModified(jsg::Lock& js, kj::Date date) override {
```

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,203 @@
...
+#define KNOWN_VFS_ROOTS(V)                                                                         \
+  V(Bundle, "file:///bundle/")                                                                     \
+  V(Temp, "file:///tmp/")
+
+class FsMap final {
+ public:
+  FsMap() = default;
+  KJ_DISALLOW_COPY_AND_MOVE(FsMap);
+
+#define V(name, _)                                                                                 \
```
anonrig, line 131: Can you give this a meaningful name rather than V?

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,203 @@
...
+    auto url = KJ_REQUIRE_NONNULL(jsg::Url::tryParse(path, "file:///"_kj), "invalid path");        \
+    set##name##Root(kj::mv(url));                                                                  \
+  }
+
+  KNOWN_VFS_ROOTS(V)
+
+#undef V
+
+ private:
+#define V(name, _) kj::Maybe<jsg::Url> maybe##name##Root;
```
anonrig, line 154: Can you give this a meaningful name rather than V?

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,203 @@
...
+ public:
+  static bool hasCurrent();
+  static TempDirStoreScope& current();
+  TempDirStoreScope(kj::Maybe<kj::Badge<TempDirStoreScope>> guard = kj::none);
+  KJ_DISALLOW_COPY_AND_MOVE(TempDirStoreScope);
+  ~TempDirStoreScope() noexcept(false);
+
+  static kj::Own<TempDirStoreScope> create();
+
+  jsg::fs::Directory& getDirectory() {
```
anonrig, line 182:
```suggestion
  const jsg::fs::Directory& getDirectory() {
```

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,245 @@
...
+      return IoContext::current().getTempDirStoreScope().getDirectory();
+    }
+    if (TempDirStoreScope::hasCurrent()) {
+      return TempDirStoreScope::current().getDirectory();
+    }
+    return kj::none;
+  }
+};
+
+class LazyDirectory final: public jsg::fs::Directory {
```
anonrig, line 110: I couldn't find any documentation about what this class does? Can you add a comment explaining the difference, and possibly add a comment to directory class as well, mentioning the existence of this class?

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,245 @@
...
+kj::Own<jsg::fs::VirtualFileSystem> getWorkerFileSystem(
+    const FsMap& fsMap, kj::Own<jsg::fs::Directory> bundleDirectory) {
+  // Our root directory is a read-only directory
+  jsg::fs::Directory::Builder builder;
+  builder.addPath(fsMap.getBundlePath(), kj::mv(bundleDirectory));
+  builder.addPath(fsMap.getTempPath(), getTempDirectoryImpl());
+  return jsg::fs::newVirtualFileSystem(builder.finish());
+}
+
+kj::Own<jsg::fs::Directory> getTempDirectoryImpl() {
```
anonrig, line 205: I think we should follow posix naming style and call this tmp rather temp.

```suggestion
kj::Own<jsg::fs::Directory> getTmpDirectoryImpl() {
```

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,203 @@
...
+// directory. For example,
+//
+// ```cpp
+// KJ_IF_SOME(node,
+//     js.resolveVfsNode("file:///tmp"_url)) {
+//   auto& dir = KJ_ASSERT_NONNULL(node.tryGet<kj::Own<jsg::fs::Directory>>());
+//   TempDirStoreScope temp_dir_scope;
+//   kj::Path path("a/b/c/foo.txt")
+//   auto tmpFile = dir.tryOpen(js, path, jsg::fs::Node::FILE);
+//   KJ_ASSERT(tmpFile.write(js, 0, "Hello World!"_kjb) == 12);
```
anonrig, line 71: Can you add a comment about write() function returning the length of the input.

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,203 @@
...
+// IoContext).
+//
+// By default, all file instances are created using the UnixEpochClock,
+// which sets all timestamps to the Unix epoch. When creating a new
+// TempDirStoreScope, you can pass a different clock to set the
+// timestamps. This is particularly important when setting timestamp
+// for files created under an IoContext, in which case the we want
+// to make sure the times are set to the correct time as returned by
+// IoContext::now(), correctly reflecting time in for that request.
+class TempDirStoreScope final {
```
anonrig, line 172:
```suggestion
class TmpDirStoreScope final {
```

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
+#include "filesystem.h"
+
+#include "buffersource.h"
+
+namespace workerd::jsg::fs {
+
+namespace {
+constexpr size_t kMaxReadLength = 32 * 1024;
+
+void validatePathWithNoSeparators(kj::StringPtr path) {
```
anonrig, line 10: Can you add a comment to functions similar to this? what is the intended result etc. Whether or not it will throw.

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+  try {
+    auto parsed = kj::Path::parse(path);
+    JSG_REQUIRE(parsed.size() == 1, Error, "Invalid path: \"", path, "\"");
```
anonrig, line 13: I'm not sure I'm following, what does this validate?

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+      .size = 0,
+      .lastModified = lastModified,
+      .writable = Writable,
+    };
+  }
+
+  size_t count(Lock& js, kj::Maybe<Type> typeFilter = kj::none) const override {
+    KJ_IF_SOME(type, typeFilter) {
+      size_t counter = 0;
+      forEach(js, [&counter](Lock&, kj::StringPtr name, const Stat stat) {
```
anonrig, line 40: forEach will execute it's callbacks if getDirectory() succeeds. but this implementation hides that fact. maybe we should rename this function, and similar ones.

```suggestion
      tryForEach(js, [&counter](Lock&, kj::StringPtr name, const Stat stat) {
```

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+  void forEach(Lock& js,
+      kj::FunctionParam<bool(Lock&, kj::StringPtr, const Stat stat)> fn,
+      kj::Maybe<Type> typeFilter = kj::none) const override {
+    for (auto& entry: entries) {
+      KJ_SWITCH_ONEOF(entry.value) {
+        KJ_CASE_ONEOF(file, kj::Own<File>) {
+          KJ_IF_SOME(type, typeFilter) {
+            if (type != Node::Type::FILE) continue;
+          }
+          if (!fn(js, entry.key, file->stat())) {
```
anonrig, line 58: Can you add a comment

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+            break;
+          }
+        }
+        KJ_CASE_ONEOF(dir, kj::Own<Directory>) {
+          KJ_IF_SOME(type, typeFilter) {
+            if (type != Node::Type::DIRECTORY) continue;
+          }
+          if (!fn(js, entry.key, dir->stat())) {
```
anonrig, line 66: Can you add a comment about what this line does

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+      path = path.slice(1, path.size());
+      if (path.size() == 0) {
+        // We found the entry, return it.
+        KJ_SWITCH_ONEOF(found) {
+          KJ_CASE_ONEOF(file, kj::Own<File>) {
+            return kj::Maybe<kj::OneOf<kj::Own<File>, kj::Own<Directory>>>(file->addRef(js));
+          }
+          KJ_CASE_ONEOF(dir, kj::Own<Directory>) {
+            return kj::Maybe<kj::OneOf<kj::Own<File>, kj::Own<Directory>>>(dir->addRef(js));
+          }
```
anonrig, line 89: Since they have similar APIs, these lines can be reduced into 1, and be simplified.

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+      JSG_FAIL_REQUIRE(Error, "Cannot transfer a file into a read-only directory");
+    }
+  }
+
+  kj::Own<Directory> transfer(Lock& js, kj::StringPtr name, kj::Own<Directory>&& dir) override {
+    if constexpr (Writable) {
+      validatePathWithNoSeparators(name);
+      KJ_IF_SOME(_, entries.find(name)) {
+        JSG_FAIL_REQUIRE(Error, "File or directory already exists: \"", name, "\"");
+      }
```
anonrig, line 143: This line can be reduced into 1 line of JSG_REQUIRE

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+      auto ret = dir->addRef(js);
+      entries.upsert(
+          kj::str(name), kj::mv(dir), [&](auto& current, auto&& node) { return kj::mv(node); });
+      return kj::mv(ret);
+    } else {
+      JSG_FAIL_REQUIRE(Error, "Cannot transfer a directory into a read-only directory");
```
anonrig, line 149: I feel like this should be the first line in this function

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+ private:
+  kj::HashMap<kj::String, kj::OneOf<kj::Own<File>, kj::Own<Directory>>> entries;
+  kj::Date lastModified = kj::UNIX_EPOCH;
+
+  kj::Maybe<kj::OneOf<kj::Own<File>, kj::Own<Directory>>> tryCreate(
+      Lock& js, kj::PathPtr path, Type createAs) {
+    KJ_DASSERT(Writable);
+    KJ_DASSERT(path.size() > 0);
+
+    if (path.size() == 1) {
```
anonrig, line 228: Why do we have this?

src/workerd/jsg/filesystem.c++
```diff
@@ -0,0 +1,537 @@
...
+  KJ_DASSERT(info.type == Node::Type::FILE);
+  JSG_REQUIRE(info.size <= kMaxReadLength, Error, "File too large to read all at once");
+  if (info.size == 0) return kj::String();
+  auto data = kj::heapArray<char>(info.size);
+  auto size = read(js, 0, data.asBytes());
+  JSG_REQUIRE(size == info.size, Error, "failed to read all data");
+  return js.accountedKjString(kj::str(kj::mv(data)));
+}
+
+BufferSource File::readAllBytes(Lock& js) const {
```
anonrig, line 399: What about using function overloading the previous function and calling it `readAll` which depends on the return value to determine which function to overload.

src/workerd/jsg/filesystem.h
anonrig: Agree

src/workerd/jsg/filesystem.h
```diff
@@ -0,0 +1,244 @@
...
+  // fail if the contents exceed a 16 MB limit.
+  BufferSource readAllBytes(Lock& js) const;
+
+  // Reads data from the file at the given offset into the given buffer.
+  virtual size_t read(Lock& js, size_t offset, kj::ArrayPtr<kj::byte> buffer) const = 0;
+
+  // Replaces the full contents of the file with the given data.
+  // Equivalent to resize(js, data.size()) followed by write(js, 0, data).
+  void writeAll(Lock& js, kj::ArrayPtr<const kj::byte> data) {
+    resize(js, data.size());
```
anonrig, line 77: Can these implementations live in C++ file?

src/workerd/jsg/modules-new-test.c++
```diff
@@ -809,9 +810,9 @@ KJ_TEST("Basic types of modules work (text, data, json, wasm)") {
   PREAMBLE(([&](Lock& js) {
     ResolveObserver resolveObserver;
     CompilationObserver compilationObserver;
-    ModuleRegistry::Builder registryBuilder(resolveObserver);
+    ModuleRegistry::Builder registryBuilder(resolveObserver, BASE);
 
-    ModuleBundle::BundleBuilder bundleBuilder;
+    ModuleBundle::BundleBuilder bundleBuilder(BASE);
```
anonrig, line 815: I believe these changes can be done on a different PR. It's hard to keep track.

src/workerd/server/BUILD.bazel
```diff
@@ -179,6 +181,16 @@ kj_test(
     ],
 )
 
+kj_test(
+    src = "bundle-fs-test.c++",
+    deps = [
+        "//src/workerd/io",
+        "//src/workerd/server",
```
anonrig, line 188: Can you include bundle-fs only? Since server is a huge dependency

src/workerd/server/BUILD.bazel
```diff
@@ -179,6 +181,16 @@ kj_test(
     ],
 )
 
+kj_test(
+    src = "bundle-fs-test.c++",
+    deps = [
+        "//src/workerd/io",
```
anonrig, line 187: I don't think you need IO

src/workerd/server/BUILD.bazel
```diff
@@ -97,10 +97,12 @@ wd_cc_library(
 wd_cc_library(
     name = "server",
     srcs = [
+        "bundle-fs.c++",
         "server.c++",
         "workerd-api.c++",
     ],
     hdrs = [
+        "bundle-fs.h",
```
anonrig, line 105: Can you put this in a new dependency?

src/workerd/jsg/filesystem.c++
jasnell: Validates that the path has only a single component, e.g. `foo` and not `foo/bar`, etc

anonrig: Ah, I see. Maybe renaming the function to `validatePathHasSingleComponent` is better?

src/workerd/jsg/jsg.h
jasnell: I'll just remove `fs::Node`. Stylistically I have a preference towards using `KJ_SWITCH_ONEOF` rather than casting from `Node`.

src/workerd/jsg/filesystem.h
jasnell: Same... I'm removing `Node` and will keep the `kj::OneOf`

src/workerd/io/worker-fs.h
jasnell: In this case it does. What it actually returns is the number of bytes written, which depending on the underlying implementation may be <= length of input.

src/workerd/jsg/filesystem.c++
jasnell: Not following... this is in a `if consexpr(...)` else clause. It's only relevant to that branch.

src/workerd/jsg/filesystem.c++
jasnell: Much prefer having them separated like this and it's more consistent with every other place where we differentiate reading text vs. reading bytes.

src/workerd/io/worker-fs.h
jasnell: The directory returned is not `const` tho. This wouldn't make sense.

src/workerd/io/io-context.h
```diff
@@ -623,6 +624,10 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
   // Access the event loop's current time point. This will remain constant between ticks.
   kj::Date now();
 
+  TempDirStoreScope& getTempDirStoreScope() {
```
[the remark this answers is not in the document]
jasnell: That would require that the `getDirectory()` method on `TempDirStoreScope` is either const or returns a const Directory, which is not what we want here.

src/workerd/jsg/filesystem.c++
jasnell: I'm not sure that renaming it to that would really be much of an improvement. Added comments

src/workerd/jsg/filesystem.c++
jasnell: Not really. Since we're using the `kj::OneOf` and these are separate types, they'll need separate branches.

src/workerd/io/worker-fs.c++
jasnell: I have a mild preference for `Temp` but don't feel strongly enough about it to argue :-)

src/workerd/jsg/filesystem.h
jasnell: They could but they are small enough to efficiently inline. I'm not sure there's much benefit in moving them out.

src/workerd/server/BUILD.bazel
jasnell: bundle-fs depends on worker-fs which depends on io-context, which is part of io

src/workerd/server/server.c++
jasnell: Refactored to use `VirtualFileSystem::tryGetCurrent(js)` as the entry point and moved out of jsg.

src/workerd/jsg/filesystem.h
jasnell: I think for now I'll keep this as is. We can revisit later.

src/workerd/jsg/filesystem.h
jasnell: Changed this to `fill(...)` and added some comments. Intend to keep this as it is far more efficient than using write for the same purpose.

src/workerd/jsg/filesystem.h
jasnell: Decided to keep these (with the exception of `cloneAsWritable()`, which is removed).... as the pattern is useful. The typical path for creating files or directories will be to use the `tryOpen(...)` but there are scenarios where these are useful to have.

jasnell:
@danlapid and @anonrig .. updated based on review and handling additional todos. PTAL

src/workerd/server/server.c++
jasnell: For the time being, I'll leave this as a TODO

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,80 @@
+#include "bundle-fs.h"
+
+#include <capnp/message.h>
+
+namespace workerd::server {
+kj::Rc<Directory> getBundleDirectory(config::Worker::Reader conf) {
+  static kj::Path kRoot{};
```
danlapid, line 7: Not sure why this should be static or why this needs to be outside of the lambda, if this is just an optimization then I'd vote to just remove the opt and and move it into the lambda.
If you insist then let's at least make it const.

src/workerd/server/workerd-api.h
```diff
@@ -38,11 +39,14 @@ class WorkerdApi final: public Worker::Api {
       kj::Own<JsgIsolateObserver> observer,
       api::MemoryCacheProvider& memoryCacheProvider,
       const PythonConfig& pythonConfig,
-      kj::Maybe<kj::Own<jsg::modules::ModuleRegistry>> newModuleRegistry);
+      kj::Maybe<kj::Own<jsg::modules::ModuleRegistry>> newModuleRegistry,
+      kj::Maybe<kj::Own<VirtualFileSystem>> vfs);
```
danlapid, line 43: Is this really a Maybe? Seems like we'll have it always so it should probably just be unwrapped.

danlapid: PS: if this is just for tests then I think we can just add a fixture that's a no-op VFS instead.
Our code should signify what is possible in prod rather than what is useful in tests (in my opinion).

src/workerd/io/worker.h
```diff
@@ -572,6 +580,11 @@ class Worker::Api {
   virtual void setModuleFallbackCallback(kj::Function<ModuleFallbackCallback>&& callback) const {
     // By default does nothing.
   }
+
+  // Return the virtual file system for this worker, if any.
+  virtual kj::Maybe<const VirtualFileSystem&> getVirtualFileSystem() const {
```
danlapid, line 585: If the former question regarding the VFS being a maybe is accepted this can be changed to not return a maybe either and just be a pure virtual function.

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,935 @@
...
+kj::Rc<Directory> getLazyDirectoryImpl(kj::Function<kj::Rc<Directory>()> func) {
+  return kj::rc<LazyDirectory>(kj::mv(func));
+}
+
+kj::Maybe<const VirtualFileSystem&> tryGetCurrent(jsg::Lock&) {
+  // Note that the jsg::Lock& argument here is not actually used. We require
+  // that a jsg::Lock reference is passed in as proof that current() is called
+  // from within a valid isolate lock so that the Worker::Api::current()
+  // call below will work as expected.
+  return Worker::Api::current().getVirtualFileSystem();
```
danlapid, line 899: Bah, this won't work in python 😢 
`Api::current` uses `Worker::Isolate::Impl::Lock` which is not instantiated when we do python startups

danlapid: That's part of why I suggested we might have to go with the abstract base class that is passed into jsg

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,475 @@
...
+    // Adds a file or directory to the directory. The name must not contain any
+    // path separators. If the name already exists an exception is thrown.
+    void add(kj::StringPtr name, kj::OneOf<kj::Rc<File>, kj::Rc<Directory>> fileOrDirectory);
+
+    // Adds a directory builder to the directory. This is used to incrementally build
+    // up read-only directories. When finish is called, the directory will be
+    // finalized. The name must not contain any path separators.
+    void add(kj::StringPtr name, kj::Own<Directory::Builder> dir);
+
+    // Adds a directory at the given path. The path may contain path separators.
```
danlapid, line 317: Nit:
```suggestion
    // Adds a file or directory at the given path. The path may contain path separators.
```

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,475 @@
...
+  // Creates a new readable/writable in-memory directory. This directory will
+  // not be initially included in a directory. To add it to a directory, use the
+  // add method. If the directory is not added, it will be deleted
+  // when the handle is dropped.
+  static kj::Rc<Directory> newWritable();
+
+  // Used to build a new read-only directory. All files and directories added
+  // may or may not be writable. The directory will not be initially included
+  // in a directory. To add it to a directory, use the add method.
+  class Builder {
```
danlapid, line 303: Is the only reason we actually need a Builder is because Directory doesn't support adding files with path separators?
e.g. tryOpen("dir_a/dir_b/file", mode=Create) is not supported and will not recursively create directories as required

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,935 @@
...
+    guard.linksSeen.insert(link);
+  }
+
+ private:
+  kj::HashSet<SymbolicLink*> linksSeen;
+};
+
+// Thread-local storage to track the current temp directory storage scope
+// on the stack.
+static thread_local TmpDirStoreScope* tmpDirStorageScope = nullptr;
```
danlapid, line 50: I wonder if this needs to be a `kj::EventLoopLocal`

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,80 @@
...
+  // Note that we are using a lazy directory here. That means we won't actually
+  // build the directory structure out until it is actually accessed in order
+  // to avoid unnecessary operations in the case a worker never actually uses
+  // this part of the filesystem.
+  return getLazyDirectoryImpl([&] {
```
danlapid, line 12:
```suggestion
  return getLazyDirectoryImpl([conf] {
```

src/workerd/io/worker-fs.h
jasnell: Builder is only used when constructing a read only directory. It's not used at all with directories that are already built.

src/workerd/server/workerd-api.h
jasnell: Currently it's a maybe because nothing on the internal repo side sets this. Once the internal project is updated to provide the vfs instance this can be changed removing the kj::Maybe.

src/workerd/io/worker-fs.c++
jasnell: I don't believe so. We don't want it to be tied to a specific event loop. If it's not coming from the IoContext, then it truly is temporary on the current execution scope... for instance, in the top-level bootstrap of worker, or the top-level evaluation of a dynamically imported module, etc. I don't think `kj::EventLoopLocal` achieves anything here. Could be wrong tho.

jasnell:
Ok... updated the PR with a few more touch ups *and* largely just to demonstrate that it works as expected, I through in an initial rough implementation of the Web FileSystem API (would need more tests, namely web platform forms) .. https://fs.spec.whatwg.org/

src/workerd/io/worker-fs.h
danlapid: Can't we simply add a releaseAsReadOnly to a WritableDirectory then?

src/workerd/server/workerd-api.h
danlapid: Is this expected to be part of this PR before pulling out of draft?
If so, mind adding it as a TODO tag to the PR description?

src/workerd/io/worker-fs.h
jasnell: Sure but I'm not convinced there's much benefit in doing so as it really wouldn't reduce the complexity at all.

jasnell:
This is already massive but doesn't yet cover everything. The various pieces may still be incomplete and buggy but the implementation is disabled by default without the additional pieces to expose the internal modules. Therefore, rather than let this continue to get bigger, let's get this part merged and I'll keep iterating in a different PR.

Note... there are likely api and implementation nits that need to be fixed. Those can but don't really need to be fixed in this PR still this is still all not exposed by default.

src/workerd/api/filesystem.c++
```diff
@@ -0,0 +1,1213 @@
...
+}
+
+constexpr FsType getTypeForName(kj::StringPtr name) {
+  if (name == "file"_kj) return FsType::FILE;
+  if (name == "directory"_kj) return FsType::DIRECTORY;
+  if (name == "symlink"_kj) return FsType::SYMLINK;
+  KJ_UNREACHABLE;
+}
+
+jsg::Url filePathToUrl(FilePath& path) {
```
anonrig, line 34:
```suggestion
jsg::Url filePathToUrl(const FilePath& path) {
```

src/workerd/api/filesystem.c++
```diff
@@ -0,0 +1,1213 @@
...
+  KJ_SWITCH_ONEOF(path) {
+    KJ_CASE_ONEOF(str, kj::String) {
+      // For a string, we need to try parsing it as a URL first.
+      auto url = JSG_REQUIRE_NONNULL(jsg::Url::tryParse(str, "file:///"_kj), Error, "Invalid path");
+      JSG_REQUIRE(url.getProtocol() == "file:"_kjb, Error, "Only file: URLs are supported");
```
anonrig, line 39: it will always start with file: since the second argument of tryParse is file://. I think this should be a debug assertion rather than a JSG_REQUIRE

src/workerd/io/compatibility-date.capnp
```diff
@@ -750,4 +750,9 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatEnableDate("2025-05-19")
       $compatDisableFlag("disable_navigator_language");
   # Enables Navigator.language API.
+
+  webFileSystem @87 :Bool
+      $compatEnableFlag("enable_web_file_system")
+      $experimental;
+  # Enables the Web File System API.
```
anonrig, line 757:
```suggestion
  # Exposes the Web File System API.
```

src/workerd/server/bundle-fs-test.c++
```diff
@@ -0,0 +1,192 @@
+// Copyright (c) 2024-2029 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include <workerd/server/bundle-fs.h>
```
anonrig, line 5: I think all fs related files should have a prefix of `fs-` or it will be extremely hard to navigate through multiple fs related implementations.

```suggestion
#include <workerd/server/fs-bundle.h>
```

src/workerd/api/filesystem.c++
```diff
@@ -0,0 +1,1213 @@
...
+  // We are not yet implementing this and it needs to be done before we can ship this.
+  auto sharedState = kj::rc<FileSystemWritableFileStream::State>(inner.addRef());
+  auto stream =
+      js.alloc<FileSystemWritableFileStream>(newWritableStreamJsController(), sharedState.addRef());
+  stream->getController().setup(js,
+      UnderlyingSink{.type = kj::str("bytes"),
+        .write =
+            [state = sharedState.addRef()](
+                jsg::Lock& js, v8::Local<v8::Value> chunk, auto c) mutable {
+    // TODO(node-fs): The spec allows the write parameter to be a WriteParams struct
```
anonrig, line 1011: This seems like a new syntax?

```suggestion
    // TODO(soon): The spec allows the write parameter to be a WriteParams struct
```

jasnell: I'm using `node-fs` intentionally to make it easier to find and track these specific todos

src/workerd/api/filesystem.c++
```diff
@@ -0,0 +1,1213 @@
...
+    position = stat.size;
+    return 0;
+  }
+  auto ret = inner->read(js, offset, buffer);
+
+  position += ret;
+  return static_cast<double>(ret);
+}
+
+double FileSystemSyncAccessHandle::write(
```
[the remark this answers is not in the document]
jasnell: These are standard defined classes. `FileSystemSyncAccessHandle` is part of the whatwg fs spec.

src/workerd/api/filesystem.c++
jasnell: -1 ... that causes problems a bit further down ... Specifically `FilePath` here is a `kj::OneOf`... if we make this a const ref, then the individual union types also become const refs... which causes a problem when we try to access the `legacyUrl` case since those methods aren't marked `const`

image

I'd rather not have to go make changes to the `URL` class here so for now I'll keep this as a non-const reference and we can revisit later.

src/workerd/api/url-standard.h
```diff
@@ -251,6 +250,10 @@ class URL: public jsg::Object {
     tracker.trackField("searchParams", maybeSearchParams);
   }
 
+  const jsg::Url& getInner() const {
+    return inner;
```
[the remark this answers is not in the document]
jasnell: It's called here: 
image

in the `jsg::Ref<url::URL>` `KJ_CASE_ONEOF` block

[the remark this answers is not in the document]
jasnell: For the time being I'll leave this as is.

jasnell:
> I think this pull-request should land gradually, and possibly splitted into multiple pull-requests to review it in more detail. It's not easy to understand the data flow.

I can split it up but if anyone asks how bits and pieces fit together and why certain changes are needed in the individual small PRs I'm just going to point them back to this one.

jasnell:
Moving this back to draft while I split out chunks into separate PRs

PR #4030, opened 2025-04-27
[DRAFT][o11y] Fix tail stream cancellation warning

fhanau:
The revoker membrane was an artifact of trace-stream being proted from JSRPC and should not be needed here.

### DIFF
--- a/src/workerd/io/trace-stream.c++
+++ b/src/workerd/io/trace-stream.c++
@@ -967,22 +967,10 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
     capnp::HttpOverCapnpFactory& httpOverCapnpFactory,
     capnp::ByteStreamFactory& byteStreamFactory,
     rpc::EventDispatcher::Client dispatcher) {
-  auto revokePaf = kj::newPromiseAndFulfiller<void>();
-
-  // TODO(streaming-tail): Session is currently being reported as canceled when using
-  // TailStreamCustomEventImpl via RPC, investigate.
-  KJ_DEFER({
-    if (revokePaf.fulfiller->isWaiting()) {
-      revokePaf.fulfiller->reject(KJ_EXCEPTION(DISCONNECTED, "Streaming tail session canceled"));
-    }
-  });
-
   auto req = dispatcher.tailStreamSessionRequest();
   auto sent = req.send();
 
   rpc::TailStreamTarget::Client cap = sent.getTopLevel();
-
-  cap = capnp::membrane(kj::mv(cap), kj::refcounted<RevokerMembrane>(kj::mv(revokePaf.promise)));
 
   auto completionPaf = kj::newPromiseAndFulfiller<void>();
   cap = capnp::membrane(
@@ -993,11 +981,7 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
   try {
     co_await sent.ignoreResult().exclusiveJoin(kj::mv(completionPaf.promise));
   } catch (...) {
-    auto e = kj::getCaughtExceptionAsKj();
-    if (revokePaf.fulfiller->isWaiting()) {
-      revokePaf.fulfiller->reject(kj::cp(e));
-    }
-    kj::throwFatalException(kj::mv(e));
+    kj::throwFatalException(kj::getCaughtExceptionAsKj());
   }
 
   co_return WorkerInterface::CustomEvent::Result{.outcome = EventOutcome::OK};
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -8,6 +8,7 @@
 #include <workerd/io/features.h>
 #include <workerd/io/tracer.h>
 #include <workerd/jsg/ser.h>
+#include <workerd/util/completion-membrane.h>
 
 #include <capnp/membrane.h>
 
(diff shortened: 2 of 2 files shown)


src/workerd/io/trace-stream.c++
```diff
@@ -993,11 +981,7 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
   try {
     co_await sent.ignoreResult().exclusiveJoin(kj::mv(completionPaf.promise));
   } catch (...) {
-    auto e = kj::getCaughtExceptionAsKj();
-    if (revokePaf.fulfiller->isWaiting()) {
-      revokePaf.fulfiller->reject(kj::cp(e));
-    }
-    kj::throwFatalException(kj::mv(e));
+    kj::throwFatalException(kj::getCaughtExceptionAsKj());
```
danlapid, line 984: Just remove the try-catch statement. No point to catch only to rethrow

src/workerd/io/trace-stream.c++
```diff
@@ -967,23 +967,11 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
...
   auto req = dispatcher.tailStreamSessionRequest();
   auto sent = req.send();
 
   rpc::TailStreamTarget::Client cap = sent.getTopLevel();
 
-  cap = capnp::membrane(kj::mv(cap), kj::refcounted<RevokerMembrane>(kj::mv(revokePaf.promise)));
-
   auto completionPaf = kj::newPromiseAndFulfiller<void>();
   cap = capnp::membrane(
       kj::mv(cap), kj::refcounted<CompletionMembrane>(kj::mv(completionPaf.fulfiller)));
```
danlapid, line 977: Not sure if you really need the CompletionMembrane either

fhanau: @jasnell this is the PR I had in mind. I'm removing the RevokerMembrane here, getting rid of the cancellation warning that was happening when running the internal streaming tail worker tests over RPC so far.
My rationale is basically that we don't need the revoker membrane in the STW model since the JSRPC model uses membranes to be able to revoke capabilities, but this is under the assumption that "Normally, the server side doesn't return if any capabilities still exist" which is not the case in the STW model (and perhaps why we're getting the cancelation warnings). In TailStreamCustomEventImpl::run() we already forego wrapping the capFulfiller promise in a membrane as we do in the JSRPC model (see worker-rpc.c++). 

To me this feels like some add-on mechanism only needed for JSRPC, not something that indicates an actual problem. Dan suggested here that the CompletionMembrane isn't needed either – perhaps we can just go ahead with that and get rid of the membranes altogether.

jasnell (approved):
As mentioned on our call yesterday, this looks like the correct change ut would feel more comfortable if @kentonv could review as well

fhanau:
> As mentioned on our call yesterday, this looks like the correct change ut would feel more comfortable if @kentonv could review as well

@kentonv

src/workerd/io/trace-stream.c++
kentonv: I agree with Dan.

kentonv (changes requested):
I don't think this is right.

If you are seeing this membrane actually revoke anything, that implies that capabilities associated with the session were still open at the time that `sendRpc()` finished or was canceled. It's important that capabilities associated with the event don't remain live beyond the lifetime of the event itself. That's exactly what the revoker is enforcing.

Now, it _may_ be true that this revoker pattern isn't really needed: if the caller of `sendRpc()` always made sure to drop all capabilities before canceling it, and if it only completed normally after all capabilities had been dropped, then there's no need to revoke anything. For many event types, this is a reasonable assumption. JSRPC only needs the revoker because it lets the application send arbitrary capabilities, and since we can't force the application to clean up after itself, we have to forcefully revoke things.

HOWEVER, if the revoker _actually is_ revoking something in your case, that implies that you have _some other bug_ which the revoker is uncovering. Somehow, you're either keeping capabilities open too long, or sendRpc() is completing prematurely, or both. That's the real bug you need to fix -- deleting the revoker doesn't actually fix it.

fhanau:
Managed to fix the cancellation issue for the affected RPC-based test after some debugging and internal discussion. Some open questions remain:
- I assume that the other `doneFulfiller->fulfill()` calls used in error cases should also be converted to only fulfill the fulfiller later on?
- Based on Kenton's comment, I think removing the revocation membrane now is acceptable (assuming we are convinced that capabilities are cleaned up properly now), so I kept the removal for now. Keeping it doesn't seem crazy either.
- Based on Dan's suggestion, is the CompletionMembrane still needed based on keeping customEvent alive while there are live capabilities, or was it only really needed alongside the RevokerMembrane?

fhanau:
> Managed to fix the cancellation issue for the affected RPC-based test after some debugging and internal discussion. Some open questions remain:
> 
> * I assume that the other `doneFulfiller->fulfill()` calls used in error cases should also be converted to only fulfill the fulfiller later on?
> * Based on Kenton's comment, I think removing the revocation membrane now is acceptable (assuming we are convinced that capabilities are cleaned up properly now), so I kept the removal for now. Keeping it doesn't seem crazy either.
> * Based on Dan's suggestion, is the CompletionMembrane still needed based on keeping customEvent alive while there are live capabilities, or was it only really needed alongside the RevokerMembrane?

@kentonv is the CompletionMembrane still needed here?

src/workerd/io/trace-stream.c++
```diff
@@ -1011,39 +1012,18 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
...
-      revokePaf.fulfiller->reject(KJ_EXCEPTION(DISCONNECTED, "Streaming tail session canceled"));
-    }
-  });
-
   auto req = dispatcher.tailStreamSessionRequest();
   auto sent = req.send();
 
   rpc::TailStreamTarget::Client cap = sent.getTopLevel();
 
-  cap = capnp::membrane(kj::mv(cap), kj::refcounted<RevokerMembrane>(kj::mv(revokePaf.promise)));
```
kentonv, line 1029: IMO you should keep the revoker as a failsafe to make sure that capabilities cannot leak from the event.

kentonv: With the other changes, this should no longer be causing a problem, right?

kentonv: Odd. It looks like the `WorkerInterface::customEvent()` call is being placed in the global `waitUntilTasks`, so why would it ever be canceled?

fhanau: Resolved in the latest change set (which turns out to be fully disjoint from the last attempt), with help from Dan. Fixing waitUntil lifetimes and making sure that the capability doesn't get deallocated prematurely on the client side resolves all prior test issues.

src/workerd/io/trace-stream.c++
fhanau: The new version maintains the completion membrane to be safe.

PR #4033, opened 2025-04-28
[ci] Do not publish automated releases when there are no changes since last auto-release

fhanau:
This should avoid superfluous releases being published on weekends.


anonrig (approved):

danlapid:
What's the downside of these releases? Why not publish?

anonrig:
> What's the downside of these releases? Why not publish?

I think the only downside is: noise.

PR #4034, opened 2025-04-28
[o11y] Allow streaming tail worker tracing itself

fhanau:
This needs to be supported at service creation time, otherwise we already support event creation in the backend and just need to enable the existing test.

==============

This builds upon #4017, but unlike that PR assumes we do want to support tail workers being traced. Recursion is limited to a single such tracing step. This is unlikely to be supported in prod, but conceivably useful when debugging tail workers locally. We already have the tests and the setEventInfo() call in trace-stream.c++ that sets up tracing support for the STW event – I figured I'd open this to discuss.


dom96 (approved):

danlapid:
I don't think we should support this
In general we should be very careful with differences in local dev and prod
If we need to debug a streaming tail worker we can just trace it with another tail worker.
Same as legacy

fhanau:
> I don't think we should support this In general we should be very careful with differences in local dev and prod If we need to debug a streaming tail worker we can just trace it with another tail worker. Same as legacy

Yeah, tracing tail workers with a different worker might just be enough. Updated the PR to no longer add recursion support, but still enable the test for tracing with a different worker. Also added a small change to refuse tracing for the test event, this was causing type errors in self-logger-test when the tail stream handler receives an empty trace array (these happened after the test customEvent completed, so they didn't cause the test to fail)

danlapid (approved):

PR #4043, opened 2025-04-29
WPT test harness improvements

npaun:
* Support functions from several `/common/` files:
  - gc.js
  - sab.js (SharedArrayBuffer)
  - subset_tests.js
* Stub implementation of `format_value`
* `expectedFailures` and `skippedTests` can now contain regexes if WPT generates a pile of similar error messages.

Now that we're running everything in WebCryptoAPI, there are two effects:
1. We can remove the ad-hoc WPT-derived tests from the internal repo, as these will be easier to keep maintained.
2. The tests are now pretty slow, and take more than 60 s to run so they might time out in CI. I'm not sure what the best solution would be...

### DIFF
--- a/src/wpt/harness/harness.ts
+++ b/src/wpt/harness/harness.ts
@@ -506,6 +506,7 @@ declare global {
   function setup(func: UnknownFunc | Record<string, unknown>): void;
   function add_completion_callback(func: UnknownFunc): void;
   function garbageCollect(): void;
+  function format_value(val: unknown): string;
   function createBuffer(
     type: 'ArrayBuffer' | 'SharedArrayBuffer',
     length: number,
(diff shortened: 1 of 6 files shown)


src/wpt/harness/harness.ts
```diff
@@ -628,6 +653,15 @@ globalThis.subsetTestByKey = (
   return testType(testCallback, testMessage);
 };
 
+globalThis.subsetTest = (testType, testCallback, testMessage): void => {
```
anonrig, line 656: Shouldn't this be named testSubset?

```suggestion
globalThis.testSubset = (testType, testCallback, testMessage): void => {
```

anonrig (approved):

src/wpt/harness/harness.ts
npaun: Nope, I have to match the function name WPT is using which is `subsetTest`.

npaun:
It's still too slow. I'm going to have to disable a couple of subtests

PR #4046, opened 2025-04-29
Add source location for the last delivered() IncomingRequest

ObsidianMinor:
Debugging is hard if you don't know who's dropping requests. However, getting a stack trace on destruction is expensive. So to help with debugging missing `IncomingRequest`, we can instead pass the `SourceLocation` of the caller of `delivered()` and use that instead. That should at least help with identifying what the last request was that was handled in an agnostic way.

### DIFF
--- a/src/workerd/io/io-context.c++
+++ b/src/workerd/io/io-context.c++
@@ -222,7 +222,7 @@ IoContext::IncomingRequest::IoContext_IncomingRequest(kj::Own<IoContext> context
 // A call to delivered() implies a promise to call drain() later (or one of the other methods
 // that sets waitedForWaitUntil). So, we can now safely add the request to
 // context->incomingRequests, which implies taking responsibility for draining on the way out.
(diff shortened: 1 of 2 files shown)


src/workerd/io/io-context.c++
```diff
@@ -429,6 +431,11 @@ void IoContext::addWaitUntil(kj::Promise<void> promise) {
     }
   }
 
+  if (!incomingRequests.empty()) {
+    KJ_LOG(WARNING, "Adding task to IoContext with no current IncomingRequest",
```
ObsidianMinor, line 435: Changed to a warning since, if my hypothesis is correct, making it an assert might break prod otherwise.

kentonv (approved):
Seems reasonable. I think the stack trace of `addWaitUntil()` may be the more useful bit but I like the idea of storing the `delivered()` location for debugging in general.

PR #4048, opened 2025-04-30
[o11y] Support more STW features

fhanau:
- [o11y] Support STW event scheduledTime
- [o11y] Support env in tailStream
  The ctx argument will be suported in a follow-up.
- [o11y] Provide unique span IDs for streaming tail worker
- [o11y] Add initial timestamp support
  We might do this completely differently (possibly providing tracee time instead of tracer time), but having timestamps available aids evaluation.

### DIFF
--- a/src/workerd/io/trace-stream.c++
+++ b/src/workerd/io/trace-stream.c++
@@ -200,8 +200,14 @@ jsg::JsValue ToJs(jsg::Lock& js, const tracing::JsRpcEventInfo& info, StringCach
 jsg::JsValue ToJs(jsg::Lock& js, const tracing::ScheduledEventInfo& info, StringCache& cache) {
   auto obj = js.obj();
   obj.set(js, TYPE_STR, cache.get(js, SCHEDULED_STR));
-  // TODO (streaming-tail-worker): Make timestamp available again, except when running tests
-  obj.set(js, SCHEDULEDTIME_STR, js.date(kj::UNIX_EPOCH));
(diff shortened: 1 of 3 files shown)


jasnell (approved):

src/workerd/io/trace-stream.c++
```diff
@@ -738,10 +752,12 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
 
     // Invoke the tailStream handler function.
     v8::Local<v8::Function> fn = maybeFn.As<v8::Function>();
-    v8::Local<v8::Value> obj = ToJs(js, event, stringCache);
+    v8::LocalVector<v8::Value> handlerArgs(js.v8Isolate, 2);
```
fhanau, line 755: Out of the tailHandler(args, env, ctx) arguments, we now support args and env. Before we add ctx here and provide these arguments to the returned handler or function too, I was wondering if the approach to invoking the tail worker here is the right one – the jsg::check(fn->Call()) pattern is used in worker-rpc.c++, whereas the legacy tail worker is invoked differently in global-scope.c++ by calling the function from the handler directly, returning a promise (`f(lock, tailEvent->getEvents(), h.env.addRef(isolate), h.getCtx());`). This seems more convenient, but we'd need to handle the returned promise properly.

fhanau: Will raise this in a follow-up since the change in here is sound

PR #4050, opened 2025-04-30
Python: Add back in reportError

hoodmane:
This is a partial revert of #4001. Without this help, edgeworker tests show:
```
2.267743795s	1m10	2025.4.1200-4-gdecc02ecd-dev error   workerd/io/worker.c++:287
script startup threw exception; id = 79b9b0009762ad21278ebf245600582fefcdd203d9aeb3efaa2105eb948f6b2f;
description = jsg.TypeError: Cannot read properties of undefined (reading 'snapshotSize');
trace =   at pyodide-internal:snapshot:140:43; ownerId = 1212; zoneId = test-zone;
scriptId = m.python; cordon = paid; process = shared
```
AKA not a traceback, and not even the name of the function the error occurred in. With this help, we get a nice traceback in the edgeworker logs.

cc @anonrig

### DIFF
--- a/src/pyodide/internal/python.ts
+++ b/src/pyodide/internal/python.ts
@@ -31,6 +31,7 @@ import { default as SetupEmscripten } from 'internal:setup-emscripten';
 
 import { default as UnsafeEval } from 'internal:unsafe-eval';
 import { simpleRunPython } from 'pyodide-internal:util';
+import { reportError } from 'pyodide-internal:util';
 import { loadPackages } from 'pyodide-internal:loadPackage';
 import { default as MetadataReader } from 'pyodide-internal:runtime-generated/metadata';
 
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/python.ts
```diff
@@ -112,56 +113,60 @@ export async function loadPyodide(
...
+        console.log(msg);
+      },
+      (msg) => {
+        console.error(msg);
+      }
+    );
+    maybeAddVendorDirectoryToPath(pyodide);
+    return pyodide;
+  } catch (e) {
+    reportError(e);
```
anonrig, line 170: Would you mind adding a comment before this line, so we don't regress?

anonrig (approved):

src/pyodide/internal/util.ts
```diff
@@ -1,3 +1,8 @@
+export function reportError(e: any): never {
+  e.stack?.split('\n').forEach((s: any) => console.warn(s));
```
anonrig, line 2: Can you add a comment to here explaining why we are splitting the stack and printing it seperately? Particularly the limitation of test runner etc.

```suggestion
  e.stack?.split('\n').forEach((s: string) => console.warn(s));
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -158,10 +159,14 @@ function getMainModule(): Promise<PyModule> {
...
-  const mainModule = await getMainModule();
-  entropyBeforeRequest(pyodide._module);
-  return mainModule;
+  try {
+    const pyodide = await getPyodide();
+    const mainModule = await getMainModule();
+    entropyBeforeRequest(pyodide._module);
+    return mainModule;
+  } catch (e) {
+    reportError(e);
```
anonrig, line 168: Can you add a comment to here as well, so we don't regress

anonrig (approved):

PR #4057, opened 2025-04-30
Python: Clean up state in snapshot.ts, NFC

hoodmane:
This reduces the number of global variables in `snapshot.ts` from 8 to 2. The two new global variables are `LOADED_SNAPSHOT_META` which contains information about a snapshot we loaded and `CREATED_SNAPSHOT_META` which contains information about a snapshot we are in the process of creating.

### DIFF
--- a/src/pyodide/internal/snapshot.ts
+++ b/src/pyodide/internal/snapshot.ts
@@ -1,3 +1,4 @@
+import { enterJaegerSpan } from 'pyodide-internal:jaeger';
 import { default as ArtifactBundler } from 'pyodide-internal:artifacts';
 import { default as UnsafeEval } from 'internal:unsafe-eval';
 import { default as DiskCache } from 'pyodide-internal:disk_cache';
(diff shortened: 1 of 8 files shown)


src/pyodide/internal/snapshot.ts
```diff
@@ -8,40 +9,72 @@ import {
...
+// DsoLoadInfo is not present because we used to preload all dynamic libraries in a standard order.
+// The "version" field is never present in the old wire format, but at runtime accessing it will
+// give undefined. We need to include it here so that we can use the version field as the
+// descriminator for `OldSnapshotMeta | SnapshotMeta`.
+type OldSnapshotMeta = DsoHandles & {
+  readonly settings?: { readonly baselineSnapshot?: boolean };
+  readonly version?: undefined;
+};
+
+// The new wire format, with additional information about the hiwire state, the order that dsos were
```
dom96, line 43: What does the hiwire state represent? I asked this before so maybe I missed your explanation when I asked it previously.

dom96 (approved):
Seems fine... though this is a lot of changes and the code is rather tough to follow. Do we have good test coverage for this?

As a general comment, I also wonder if it's possible to avoid globals completely for this code. Can we have a `SnapshotState` class that is instantiated at the start and holds all of this state? Not something that needs to be done as part of this PR, but perhaps something to consider for future refactorings.

src/pyodide/internal/snapshot.ts
hoodmane: WebAssembly has this type `externref` which is an opaque reference to a host object. The main point of them is to allow passing a JavaScript object through WebAssembly frames. To use it, WebAssembly has to call back out to a JavaScript function.

An `externref` is not a number and it is impossible to store it into the WebAssembly linear memory, or for instance in a struct field. Instead, WebAssembly has a concept of a [Table](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/Table) and there are instructions to store and retrieve externrefs in tables.

So for a C struct to hold a reference to a JavaScript object, it has to put the object into a table and then the struct can hold the index into the table instead. [Hiwire](https://github.com/hoodmane/hiwire) is responsible for maintaining the table state and converting between table indexes and externrefs. Since the table entries are random JavaScript objects, we need some way to serialize them. This is the serialized table state.

hoodmane:
We do have quite good test coverage for this.

PR #4058, opened 2025-04-30
enable erasableSyntaxOnly on src/node

anonrig:
Enables erasableSyntaxOnly on src/node


npaun (approved):

anonrig:
This requires a review from @cloudflare/wrangler team

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -166,7 +167,8 @@ class VectorizeIndexImpl implements Vectorize {
       headers: {
         'content-type': contentType,
         'cf-vector-search-dim-width': String(
-          vectors.length ? vectors[0]?.values?.length : 0
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          vectors.length ? vectors.at(0)?.values?.length : 0
```
penalosa, line 171: Why does this need changed?

penalosa (approved):

src/cloudflare/internal/vectorize-api.ts
anonrig: it doesn't had to be changed, i just had to add a eslint disable next line there. somehow there is a bug, and with or without [0]/.at(0), it still fails. 

This will probably need to be done in a separate PR but we need to disallow using unsafe array index in our tsconfig, since array[123123] will not resolve into T | undefined, it will resolve to T, unles we use .at(123123)

fhanau (approved):

jasnell (approved):
If we are going to use erasable syntax moving forward then we should include some documentation to help folks out. What are the requirements for syntax to be erasable? What typescript features are not allowed, etc. We should also just look at the impact on the overall binary size here. I don't imagine it will be significant but given that the source won't be compacted down and the bundled sources will have more whitespace, I have to imagine the overall binary size will grow.

jasnell:
Can you expand the PR description with either a description of what erasable syntax is or links to where folks can find out.

vicb:
> If we are going to use erasable syntax moving forward then we should include some documentation to help folks out. What are the requirements for syntax to be erasable? What typescript features are not allowed, etc. We should also just look at the impact on the overall binary size here. I don't imagine it will be significant but given that the source won't be compacted down and the bundled sources will have more whitespace, I have to imagine the overall binary size will grow.

Good point for the documentation!

We should tell use to use the `erasableSyntaxOnly` TS option and can [link the TS website](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#the---erasablesyntaxonly-option)

anonrig:
> If we are going to use erasable syntax moving forward then we should include some documentation to help folks out. What are the requirements for syntax to be erasable? What typescript features are not allowed, etc.

So, everything related to the erasable syntax will be given as an error by the typescript compiler. For example, if you use "enum" in typescript, it will fail because there isn't a 1 to 1 correlation between TS and JS. I don't think this requires a documentation since other tsconfig's also doesn't have a documentation in workerd.

> We should also just look at the impact on the overall binary size here. I don't imagine it will be significant but given that the source won't be compacted down and the bundled sources will have more whitespace, I have to imagine the overall binary size will grow.

@jasnell I didn't update esbuild or how to build things. This will only disallow us to use typescript features that cannot be erased. So, no need.

> Can you expand the PR description with either a description of what erasable syntax is or links to where folks can find out.

Definitely. I added a link to the description. 

> We should tell users to use the erasableSyntaxOnly TS option and can [link the TS website](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#the---erasablesyntaxonly-option)

@vicb This doesn't have any user impact. This will allow workerd developers/contributors to have a better error stack trace, since we don't have sourcemaps enabled for our typescript builds.

anonrig:
Rebased due to broken workers-sdk tests.

vicb:
> This doesn't have any user impact. This will allow workerd developers/contributors to have a better error stack trace, since we don't have sourcemaps enabled for our typescript builds.

Is there any particular reason not to enable source maps then? 
It would be nice to add your comment and the answer to that question in the PR description - it is always good to have context when reviewing a PR.
Thx

anonrig:
> > This doesn't have any user impact. This will allow workerd developers/contributors to have a better error stack trace, since we don't have sourcemaps enabled for our typescript builds.
> 
> Is there any particular reason not to enable source maps then? It would be nice to add your comment and the answer to that question in the PR description - it is always good to have context when reviewing a PR. Thx

I think sourcemap support is there, but it doesn't work properly. Will need someone with more experience in build systems need to take a look at it in more detail.

PR #4062, opened 2025-05-01
[build] Add Windows arm64 runner

fhanau:


fhanau:
This is blocked for now – rules_js uses yq to parse the pnpm lock file, but doesn't set up a arm64 Windows toolchain for it. https://github.com/bazel-contrib/bazel-lib/blob/main/lib/private/yq_toolchain.bzl

anonrig (approved):

fhanau:
Managed to patch out the yq issue, but that only revealed several other issues:
- The rules_nodejs bazel toolchain (and possibly more toolchains in rules_js/ts) does not support arm64 Windows.
- clang-cl is present, but compiling fails based on an error about using an LLVM version < 20. This is not the case on the x64 image. Clearly, the image could benefit from better maintenance.
- https://github.com/actions/partner-runner-images/issues/94

This continues to be blocked unfortunately.

fhanau:
Superseded by #6486.

PR #4065, opened 2025-05-01
Update internal build to use merge commit sha instead of head ref

ketanhwr:


.github/workflows/internal-build.yml
```diff
@@ -33,7 +48,7 @@ jobs:
         run: |
           python3 -u ./tools/cross/internal_build.py \
             ${{github.event.pull_request.number}} \
-            ${{github.event.pull_request.head.sha}} \
+            ${{steps.get_sha.outputs.sha}} \
```
npaun, line 51:
```suggestion
            ${{github.event.pull_request.merge_commit_sha} \
```

The documentation's not the clearest but you may be able to save some work with this: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request

ketanhwr: That was my initial attempt, but `merge_commit_sha` was null for some reason.

danlapid (approved):
🔥🔥🔥🔥
So clean

PR #4066, opened 2025-05-01
Fix urlpattern standard test

ketanhwr:
The compat flag becomes available by default starting 2025-05-01

### DIFF
--- a/src/wpt/BUILD.bazel
+++ b/src/wpt/BUILD.bazel
@@ -49,7 +49,6 @@ wpt_test(
 
 wpt_test(
     name = "urlpattern-standard",
-    compat_flags = ["urlpattern_standard"],
     config = "urlpattern-standard-test.ts",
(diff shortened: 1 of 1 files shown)


danlapid (approved):

npaun (approved):

anonrig (approved):

src/wpt/BUILD.bazel
```diff
@@ -49,7 +49,6 @@ wpt_test(
 
 wpt_test(
     name = "urlpattern-standard",
-    compat_flags = ["urlpattern_standard"],
```
anonrig, line 52: Why is this needed?

ketanhwr: The test currently fails with "This compat flag is default starting 2025-05-01 so no longer needed". (this should probably be a warning instead of an error though?)

PR #4069, opened 2025-05-01
[build] Implement fast C23-based wd_cc_embed macro

fhanau:
This ports the wd_cc_embed macro to workerd to use #embed to embed files into C/C++ sources. LLVM19 is required, which is now available on macOS (via Xcode 16.3), Ubuntu 24.04 and Debian Bookworm. It seems reasonable to require that version soon, but overall this PR only provides limited cleanup opportunities as pyodide capnp embeds can't be ported to being embedded in C++ easily.
~~We'd also want to add a convenient way to rename the generated variables, `EMBED_workerd_io_trimmed_supported_compatibility_date_txt` feels inconvenient – this change set can wait for now.~~


fhanau:
Reworked this a bit from the internal version to require fewer source code changes. Upstream CI is failing so I assume some changes will be needed there.
Reviewers: This ports an internal mechanism to embed files into C++ code to workerd. This is much faster than the existing capnp embeds. I ported a few files over, although this is not trivial for cases where we currently embed several files together. I'd imagine that python embeds can benefit from this in the medium term.

- [x] Make any needed downstream changes before merge

mikea (approved):
nice! I bet this makes a build at least 1 minute faster.

fhanau:
Rebased, and added another commit to facilitate using the macro in a different workspace.

PR #4072, opened 2025-05-02
Fix Python RPC calls when there is a pool

hoodmane:
The problem is that when the Python pool is used, the `Function` constructor is used from the incorrect realm. This creates mysterious problems serializing the arguments to an RPC call. It was failing in `serializer.write(js, value);`. This fixes it by using the `Function` constructor from the correct realm.

Tested downstream.


src/pyodide/internal/python.ts
```diff
@@ -44,6 +44,7 @@ import { default as MetadataReader } from 'pyodide-internal:runtime-generated/me
 function prepareWasmLinearMemory(Module: Module): void {
   // Note: if we are restoring from a snapshot, runtime is not initialized yet.
   Module.noInitialRun = !SHOULD_RESTORE_SNAPSHOT;
+  Module.Function = Function;
```
danlapid, line 47: Why not use Module.globalThis.Function instead of defining another member on Module?

dom96 (approved):
Nice! Apart from Dan's question, LGTM

danlapid (approved):

PR #4074, opened 2025-05-02
feat: Add `WorkflowRetentionDuration` to `WorkflowInstanceCreateOptions`

Caio-Nogueira:
Related to WOR-638

We need to expose `InstanceRetention` as an additional field to `WorkflowInstanceCreateOptions`, thus allowing custom retention periods through binding-shim.

### DIFF
--- a/src/cloudflare/internal/workflows.d.ts
+++ b/src/cloudflare/internal/workflows.d.ts
@@ -50,6 +50,21 @@ declare abstract class Workflow<PARAMS = unknown> {
   ): Promise<WorkflowInstance[]>;
 }
 
+type WorkflowDurationLabel =
+  | 'second'
+  | 'minute'
+  | 'hour'
+  | 'day'
+  | 'week'
+  | 'month'
+  | 'year';
+
+type WorkflowSleepDuration =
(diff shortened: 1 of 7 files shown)


src/cloudflare/internal/workflows.d.ts
```diff
@@ -60,6 +60,10 @@ interface WorkflowInstanceCreateOptions<PARAMS = unknown> {
    * The event payload the Workflow instance is triggered with
    */
   params?: PARAMS;
+  /**
+   * The retention policy for Workflow instance
```
dario-piotrowicz, line 64: here and in the other d.ts file, could we specify what the default retention behaviors/values are?

Caio-Nogueira: We can't really specify default values here, as they depend on the account. However, we can (and probably should) provide clarification about what happens if no `retention` is given.
I added another line to fix this.

LuisDuarte1 (approved):
Approved from the workflows team

Caio-Nogueira:

danlapid (approved):

PR #4075, opened 2025-05-02
add navigator.language attribute

anonrig:
Opening another one since auto-merge feature was faster than me.

https://github.com/cloudflare/workerd/pull/4071

### DIFF
--- a/src/workerd/api/global-scope.h
+++ b/src/workerd/api/global-scope.h
@@ -83,11 +83,26 @@ class Navigator: public jsg::Object {
     return 1;
   }
 
-  JSG_RESOURCE_TYPE(Navigator) {
+  kj::StringPtr getLanguage() {
+    return language_;
+  }
+
+  void setLanguage(kj::String language) {
+    language_ = kj::mv(language);
+  }
+
+  JSG_RESOURCE_TYPE(Navigator, CompatibilityFlags::Reader reader) {
     JSG_METHOD(sendBeacon);
     JSG_READONLY_INSTANCE_PROPERTY(userAgent, getUserAgent);
     JSG_READONLY_INSTANCE_PROPERTY(hardwareConcurrency, getHardwareConcurrency);
+
+    if (reader.getEnableNavigatorLanguage()) {
+      JSG_INSTANCE_PROPERTY(language, getLanguage, setLanguage);
+    }
   }
+
+ private:
+  kj::String language_ = kj::str("en");
 };
 
 class Performance: public jsg::Object {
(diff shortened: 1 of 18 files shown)


src/workerd/api/global-scope.h
```diff
@@ -84,16 +84,25 @@ class Navigator: public jsg::Object {
...
   kj::StringPtr getLanguage() {
-    // Some packages depend on navigator.language being set to a specific value.
-    return "en"_kj;
+    return language_;
   }
 
-  JSG_RESOURCE_TYPE(Navigator) {
+  void setLanguage(kj::String language) {
+    language_ = kj::mv(language);
+  }
```
jasnell, line 92: I don't think this is actually correct either. It's not that the property is not read-only, it's that it is configurable.

```
> Object.getOwnPropertyDescriptor(navigator.__proto__, 'language')
{
  get: [Function: get language],
  set: undefined,
  enumerable: true,
  configurable: true
}
```

It's also worth noting that `navigator.language` that per the spec and all other implementations, navigator properties are actually on the prototype and not the instance... which is not something we've actually followed correctly :-( ...

anonrig: I can put every getter in navigator under the new compat flag and set them using `JSG_PROTOTYPE_PROPERTY`. Would that be sufficient?

anonrig: wdyt about

```
  JSG_RESOURCE_TYPE(Navigator, CompatibilityFlags::Reader reader) {
    JSG_METHOD(sendBeacon);

    if (reader.getEnableNavigatorLanguage()) {
      JSG_PROTOTYPE_PROPERTY(language, getLanguage, setLanguage);
      JSG_READONLY_PROTOTYPE_PROPERTY(userAgent, getUserAgent);
      JSG_READONLY_PROTOTYPE_PROPERTY(hardwareConcurrency, getHardwareConcurrency);
    } else {
      JSG_READONLY_INSTANCE_PROPERTY(userAgent, getUserAgent);
      JSG_READONLY_INSTANCE_PROPERTY(hardwareConcurrency, getHardwareConcurrency);
    }
  }
  ```

jasnell: For now, let's leave them as instance properties. We can address the instance vs prototype property issue separately. For here, I wouldn't add the setter, I would just confirm whether or not adding the property as read-only still allows `Object.defineProperty(...)` to work. It might just work already but I don't think we've verified it.

jasnell: we might need an additional macro... something like `JSG_CONFIGURABLE_READONLY_PROTOTYPE_PROPERTY`... yes yes, I know that's super terrible....

jasnell (approved):

PR #4076, opened 2025-05-02
[o11y] Support ctx attribute in tailStream(), clarify that only one event is passed at a time

fhanau:
- [o11y] Support ctx attribute in tailStream(), clarify that only one event is passed at a time
- [o11y] Provide parent span ID in SpanOpen event
  API design may still change here, so this is isolated in one commit to make reverting easier if needed.

### DIFF
--- a/src/workerd/io/trace-stream.c++
+++ b/src/workerd/io/trace-stream.c++
@@ -752,12 +752,17 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
 
     // Invoke the tailStream handler function.
     v8::Local<v8::Function> fn = maybeFn.As<v8::Function>();
-    v8::LocalVector<v8::Value> handlerArgs(js.v8Isolate, 2);
+    auto maybeCtx = KJ_ASSERT_NONNULL(handler->getCtx()).tryGetHandle(js);
+    v8::LocalVector<v8::Value> handlerArgs(js.v8Isolate, maybeCtx != kj::none ? 3 : 2);
     handlerArgs[0] = ToJs(js, event, stringCache);
     handlerArgs[1] = handler->env.getHandle(js);
+    KJ_IF_SOME(ctx, maybeCtx) {
+      handlerArgs[2] = ctx;
+    }
 
     try {
-      auto result = jsg::check(fn->Call(js.v8Context(), target, 2, handlerArgs.data()));
+      auto result =
+          jsg::check(fn->Call(js.v8Context(), target, handlerArgs.size(), handlerArgs.data()));
 
       // We need to be able to access the results builder from both the
       // success and failure branches of the promise we set up below.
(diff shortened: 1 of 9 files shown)


src/workerd/io/trace-stream.c++
```diff
@@ -752,12 +761,17 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
 
     // Invoke the tailStream handler function.
     v8::Local<v8::Function> fn = maybeFn.As<v8::Function>();
-    v8::LocalVector<v8::Value> handlerArgs(js.v8Isolate, 2);
-    handlerArgs[0] = ToJs(js, event, stringCache);
-    handlerArgs[1] = handler->env.getHandle(js);
+    v8::LocalVector<v8::Value> handlerArgs(js.v8Isolate);
```
danlapid, line 764: I would still preallocate vector size based on arg count
Ctx != none ? 3 : 2

src/workerd/io/trace.c++
```diff
@@ -1035,8 +1037,10 @@ kj::Maybe<tracing::SpanOpen::Info> readSpanOpenInfo(rpc::Trace::SpanOpen::Reader
 }  // namespace
 
 tracing::SpanOpen::SpanOpen(rpc::Trace::SpanOpen::Reader reader)
+    // TODO(streaming-tail): Propagate parentSpanId properly
     : operationName(readSpanOpenOperationName(reader)),
-      info(readSpanOpenInfo(reader)) {}
+      info(readSpanOpenInfo(reader)),
+      parentSpanId(0) {}
```
danlapid, line 1043: Shouldn't this be populated from the reader?

src/workerd/io/trace.h
```diff
@@ -640,15 +640,17 @@ struct SpanOpen final {
   // details of that subrequest.
   using Info = kj::OneOf<FetchEventInfo, JsRpcEventInfo, CustomInfo>;
 
-  explicit SpanOpen(
-      kj::Maybe<kj::String> operationName = kj::none, kj::Maybe<Info> info = kj::none);
+  explicit SpanOpen(uint64_t parentSpanId,
+      kj::Maybe<kj::String> operationName = kj::none,
```
danlapid, line 644: We wanted to make this non-optional, is this in a separate PR?

danlapid (approved):

src/workerd/io/trace.c++
fhanau: In practice, yes – there is no parent Span ID tag defined for the SpanOpen reader so far though. We certainly need to define this somewhere in the capnp representation eventually, but this PR is mostly intended to unblock evaluation, so I just wanted to make a parent span ID available in the common case. Where exactly the parent Span ID should be defined is still something we need to finalize.

PR #4077, opened 2025-05-02
Update node to 20.19.1 (version used by wrangler).

hoodmane:
This unbreaks CI on the main branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -181,7 +181,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           # Match workers-sdk node version
-          node-version: 18.20.5
+          node-version: 20.19.1
           cache: 'pnpm'
       - name: Install workers-sdk dependencies
         run: pnpm install


danlapid (approved):
Though I seem to recall seeing another such PR from @penalosa

hoodmane:
cc @penalosa could we add a note in workers-sdk to come and update the node version here when you update it in workers-sdk?

.github/workflows/test.yml
```diff
@@ -181,7 +181,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           # Match workers-sdk node version
-          node-version: 18.20.5
+          node-version: 20.19.1
```
anonrig, line 184: Can you update this to v22? We use v22 internally.

https://github.com/cloudflare/workerd/blob/e94e9d689591d02b4fe364d1a0c76bca0bc09335/WORKSPACE#L21

hoodmane: the comment says:

> # Match workers-sdk node version

Do you disagree with that?

hoodmane: If the goal is to avoid breaking wrangler's CI with workerd commits, this makes the most sense I'd say.

penalosa: Yeah, I'd vote for keeping this at 20. These tests also run with Node 22 before a workers-sdk release

anonrig (approved):

PR #4078, opened 2025-05-02
Fixes WorkerEntrypoint methods not being reported in validator.

dom96:
Turns out WorkerEntrypoint support for Python is broken due to the validator needing to know at least one method that it exports. This should fix it.

### DIFF
--- a/src/pyodide/internal/introspection.py
+++ b/src/pyodide/internal/introspection.py
@@ -6,21 +6,42 @@ from workers import DurableObject, WorkerEntrypoint, WorkflowEntrypoint
 from pyodide.ffi import to_js
 
 
-def collect_classes(user_mod, cls):
+def collect_methods(class_val):
+    """
+    Iterates through the methods in `class_val` and returns only public non-static/non-class methods
+    defined on that class.
+    """
+    return [
+        name
+        for name, value in class_val.__dict__.items()
+        if not isinstance(value, (classmethod, staticmethod))
+        and not name.startswith("_")
+    ]
+
+
+def collect_classes(user_mod, base_cls):
     """
     Iterates through the defined symbols in the input module. Returns any classes which extend
-    DurableObject.
+    `base_cls` (where `base_cls` is one of DurableObject, WorkerEntrypoint or WorkflowEntrypoint).
+
+    This method returns a list of JS objects like [{"className": "MyClass", "methodNames": ["foo"]}].]
     """
     if hasattr(user_mod, "__all__"):
         keys = user_mod.__all__
     else:
         keys = (key for key in dir(user_mod) if not key.startswith("_"))
 
-    def filter(key):
-        val = getattr(user_mod, key)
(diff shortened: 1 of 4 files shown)


src/pyodide/internal/introspection.py
```diff
@@ -1,3 +1,4 @@
+import types
```
hoodmane, line 1:
```suggestion
from types import FunctionType
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -275,6 +279,15 @@ function makeEntrypointClass(className: string, classKind: AnyClass): any {
       return new Proxy(this, makeEntrypointProxyHandler(pyInstancePromise));
     }
   };
+
+  // Add dummy functions to the class so that the validator can detect them.
+  for (let method of methods) {
+    if (SUPPORTED_HANDLER_NAMES.includes(method.slice(3))) {
+      method = method.slice(3);
```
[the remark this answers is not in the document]
hoodmane: I guess you're removing the `on_` prefix.

src/pyodide/python-entrypoint.js
```diff
@@ -32,10 +32,10 @@ async function doAnImport(name) {
 // Pass the import function to the helper
 setDoAnImport(doAnImport, cloudflareWorkersModule, cloudflareSocketsModule);
 
-function makeEntrypointClassFromNames(names, cls) {
-  return names.map((className) => [
+function makeEntrypointClassFromNames(classes, cls) {
```
hoodmane, line 35: Maybe let's rename the `cls` argument to `baseClass` for clarity?

hoodmane (approved):
Thanks!

src/pyodide/internal/workers.py
```diff
@@ -895,7 +895,7 @@ def wrapper(*args, **kwargs):
 
 
 def _wrap_attr(attr):
-    if not callable(attr):
+    if not callable(attr) or isinstance(attr, (classmethod, staticmethod)):
```
hoodmane, line 898: Of course, we learned that `isinstance(attr, classmethod)` implies `not callable(attr)` so checking that is redundant. But doing so is more readable. Maybe add a comment?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -270,6 +274,16 @@ function makeEntrypointClass(className: string, classKind: AnyClass): any {
       return new Proxy(this, makeEntrypointProxyHandler(pyInstancePromise));
     }
   };
+
+  // Add dummy functions to the class so that the validator can detect them.
```
hoodmane, line 278: Is the point that these will never get accessed on an instance because of the `Proxy`? If so could you explain in the comment?

dom96: Yep.

PR #4082, opened 2025-05-04
Refactor Worker::Api module loading

kentonv:
My goal is to be able to construct a `Worker::Script` from code that does _not_ originate from a workerd config file.

To that end, I've defined an intermediate representation of workerd modules as a `OneOf` of a bunch of structs (see `Worker::Script::Module`). `Worker::Script::ModulesSoruce` now returns this representation, instead of returning a callback used to compile the modules later.

`Wroker::Api` now has a virtual `compileModules()` method which replaces the old lambda callback, taking the intermediate representation as input.

Note: I am a little confused about how the "new module registry" stuff works, since it appears that the code is passed directly to it, bypassing the `Worker::Script` constructor. Is the code passed to the constructor not used anymore when using the new module registry? Well, with this change, the code in `WorkerdApi::initializeBundleModuleRegistry()` is no longer workerd-specific, so perhaps in a future change it could move somewhere else and become reusable in the internal codebase.


jasnell:
> ... I am a little confused about how the "new module registry" stuff works, since it appears that the code is passed directly to it, ...

Yeah, the new module registry set up works quite a bit differently. Really happy to see these cleanups as I think it'll simplify a number of things. Should be able to do some review here this afternoon

src/workerd/io/worker.h
```diff
@@ -228,12 +233,57 @@ class Worker::Script: public kj::AtomicRefcounted {
...
+  inline bool isPython() const {
+    return python;
+  }
 
   struct CompiledGlobal {
     jsg::V8Ref<v8::String> name;
     jsg::V8Ref<v8::Value> value;
   };
 
+  struct EsModule {
```
jasnell, line 245: Nit: a code comment here briefly explaining these structs would be nice. Nonblocking.

jasnell (approved):
Overall change LGTM. I'll look at the new module registry details to see if any of that needs to be updated or can be updated to benefit here.

src/workerd/io/worker.h
kentonv: Done in f2858bcd48ec761434f37108604e68cf90e523f2

PR #4085, opened 2025-05-04
enable v8's explicit-resource-management and float16array

anonrig:
These will be enabled by default starting from Node.js v24.

### DIFF
--- a/src/workerd/jsg/setup.c++
+++ b/src/workerd/jsg/setup.c++
@@ -149,6 +149,11 @@ void V8System::init(kj::Own<v8::Platform> platformParam,
   // more flags.)
   v8::V8::SetFlagsFromString("--noincremental-marking");
 
+  // These features are completed and enabled by default in Chrome, but not
+  // in V8. Follows Node.js: https://github.com/nodejs/node/pull/58154
+  v8::V8::SetFlagsFromString("--js-explicit-resource-management");
+  v8::V8::SetFlagsFromString("--js-float16array");
+
 #ifdef __APPLE__
   // On macOS arm64, we find that V8 can be collecting pages that contain compiled code when
   // handling requests in short succession. There are some specific differences for macOS arm64
--- a/src/wpt/compression-test.ts
+++ b/src/wpt/compression-test.ts
@@ -45,15 +45,7 @@ export default {
       'V8 assertion - Cannot construct ArrayBuffer with a BackingStore of SharedArrayBuffer',
     skipAllTests: true,
   },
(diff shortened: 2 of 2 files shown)


jasnell (changes requested):
We typically do not enable v8 flags like this. ERM, in particular, is still only at stage 3 in TC-39 so theoretically could still take breaking changes. We usually take a more conservative approach in workers and wait for v8 to enable these by default.

src/workerd/jsg/setup.c++
```diff
@@ -151,6 +151,11 @@ void V8System::init(kj::Own<v8::Platform> platformParam,
   // more flags.)
   v8::V8::SetFlagsFromString("--noincremental-marking");
 
+  // These features are completed and enabled by default in Chrome, but not
+  // in V8. Follows Node.js: https://github.com/nodejs/node/pull/58154
+  v8::V8::SetFlagsFromString("--js-explicit-resource-management");
+  v8::V8::SetFlagsFromString("--js-float16array");
```
jasnell, line 157: It would probably be helpful to add a couple of simple tests that verify the flags are properly set. I know the compression-test covers the `Float16Array` usage but we should also have one that covers ERM

jasnell (approved):

jasnell:
FWIW, this PR will likely need to wait for the v8 13.7 update to land in workerd in order for use to be able to drop the `Symbol.dispose` and `Symbol.asyncDispose` polyfill patch.

src/workerd/jsg/setup.c++
anonrig: Any recommendations on what to test?

anonrig: I'll do a follow up for js-explicit-resource-management tests.

jasnell: Simple test:

```js
import { strictEqual } from 'node:assert';
import { mock } from 'node:test';

// Test verifies that explicit resource management is, in fact, enabled.

export const disposeTest = {
  test() {
    const fn = mock.fn();
    {
      using _ = {
        [Symbol.dispose]() {
          fn();
        },
      };
    }
    strictEqual(fn.mock.callCount(), 1);
  },
};

export const asyncDisposeTest = {
  async test() {
    const fn = mock.fn(async () => {});
    {
      await using _ = {
        async [Symbol.asyncDispose]() {
          await fn();
        },
      };
    }
    strictEqual(fn.mock.callCount(), 1);
  },
};

```

PR #4088, opened 2025-05-05
Put Request.signal behind a compat flag

npaun:
* Puts `Request.signal` behind a compat flag so we can unblock releases and continue investigating it
  - `enable_request_signal` must be set for WorkerEntrypoint to initialize `abortController` and trigger it on cancellation.
  - Just in case, `enable_request_signal` must be set to switch from `getSignal()` to `getThisSignal()` in places in `Request` where I did that for Fetch spec compliance reasons.
* Adds a new test, `request-signal-disabled.js` to make sure things work fine with the flag disabled.
* I've added a commit where I intentionally mess up `AbortSignal`'s serialization. In this case: 
   - request-signal-disabled: PASS
   - request-client-disconnect.incomingRequestCanBeCloned: FAIL with DataCloneError
   - abortsignal-test: FAIL with DataCloneError
   - I'm doing a CI run with it in, and will just drop the commit after.

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1008,7 +1008,7 @@ jsg::Ref<Request> Request::constructor(
       cacheMode = oldRequest->getCacheMode();
       redirect = oldRequest->getRedirectEnum();
       fetcher = oldRequest->getFetcher();
-      signal = oldRequest->getThisSignal(js);
+      signal = oldRequest->getSignalForSubrequest(js);
     }
   }
 
@@ -1118,7 +1118,7 @@ jsg::Ref<Request> Request::constructor(
         cacheMode = otherRequest->cacheMode;
         responseBodyEncoding = otherRequest->responseBodyEncoding;
         fetcher = otherRequest->getFetcher();
-        signal = otherRequest->getThisSignal(js);
+        signal = otherRequest->getSignalForSubrequest(js);
         headers = js.alloc<Headers>(js, *otherRequest->headers);
         cf = otherRequest->cf.deepClone(js);
         KJ_IF_SOME(b, otherRequest->getBody()) {
@@ -1208,6 +1208,18 @@ jsg::Ref<AbortSignal> Request::getThisSignal(jsg::Lock& js) {
   auto newSignal = js.alloc<AbortSignal>(kj::none, kj::none, AbortSignal::Flag::NEVER_ABORTS);
   thisSignal = newSignal.addRef();
   return newSignal;
+}
+
+kj::Maybe<jsg::Ref<AbortSignal>> Request::getSignalForSubrequest(jsg::Lock& js) {
+  // For subrequests, the Fetch spec specifies that "this signal" should be taken from the original
+  // request, rather than "signal". To be certain backwards compatibility is not broken, this
+  // function preserves the old behaviour unless the 'enableRequestSignal' compat flag is set
+
+  if (FeatureFlags::get(js).getEnableRequestSignal()) {
+    return getThisSignal(js);
+  } else {
+    return getSignal();
+  }
 }
 
 void Request::clearSignalIfIgnoredForSubrequest() {
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1008,7 +1008,7 @@ jsg::Ref<Request> Request::constructor(
       cacheMode = oldRequest->getCacheMode();
       redirect = oldRequest->getRedirectEnum();
       fetcher = oldRequest->getFetcher();
-      signal = oldRequest->getSignal();
+      signal = oldRequest->getThisSignal(js);
     }
   }
 
@@ -1064,6 +1064,7 @@ jsg::Ref<Request> Request::constructor(
           // explicitly say `signal: null`, they must want to drop the signal that was on the
           // original request.
           signal = kj::mv(s);
+          initDict.signal = kj::none;
         }
 
         KJ_IF_SOME(newCf, initDict.cf) {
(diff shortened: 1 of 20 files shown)


src/workerd/io/compatibility-date.capnp
```diff
@@ -750,4 +750,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatEnableDate("2025-05-19")
       $compatDisableFlag("disable_navigator_language");
   # Enables Navigator.language API.
+
+  enableRequestSignal @87 :Bool
+    $compatEnableFlag("enable_request_signal")
+    $compatDisableFlag("disable_request_signal");
```
danlapid, line 756: +$experimental

src/workerd/io/worker-entrypoint.c++
```diff
@@ -296,12 +296,18 @@ kj::Promise<void> WorkerEntrypoint::request(kj::HttpMethod method,
...
-        ? api::AbortSignal::Flag::NONE
-        : api::AbortSignal::Flag::IGNORE_FOR_SUBREQUESTS;
-    jsg::Lock& js = lock;
-    jsg::Ref<api::AbortSignal> signal =
-        abortController.emplace(js.alloc<api::AbortController>(js, flags))->getSignal();
+    auto featureFlags = FeatureFlags::get(lock);
+
+    kj::Maybe<jsg::Ref<api::AbortSignal>> signal;
+
+    if (featureFlags.getEnableRequestSignal()) {
```
danlapid, line 303: Nit:
```suggestion
    if (FeatureFlags::get(lock).getEnableRequestSignal()) {
```

src/workerd/api/http.c++
```diff
@@ -1210,6 +1210,14 @@ jsg::Ref<AbortSignal> Request::getThisSignal(jsg::Lock& js) {
   return newSignal;
 }
 
+kj::Maybe<jsg::Ref<AbortSignal>> Request::getSignalForSubrequest(jsg::Lock& js) {
```
danlapid, line 1213: @jasnell You reviewed the original PR I think, can you validate this PR as well?

anonrig (approved):

anonrig:
@jasnell can you take a look at this pr?

src/workerd/api/http.c++
```diff
@@ -1210,6 +1210,14 @@ jsg::Ref<AbortSignal> Request::getThisSignal(jsg::Lock& js) {
...
+  if (FeatureFlags::get(js).getEnableRequestSignal()) {
```
jasnell, line 1214: Nit: some additional documentation here to provide context would be helpful... I know there's comments in the http.h but having it here with the code as well would be ideal

jasnell (approved):

jasnell:
Given the issues we've had with rollouts on this, I would also strongly recommend that we roll this out with an autogate.

npaun:
^^^ @jasnell, @kentonv

src/workerd/api/http.c++
```diff
@@ -1330,8 +1352,13 @@ void Request::serialize(jsg::Lock& js,
             //
             // Note we have to double-Maybe this, so that if no signal is present, the property is absent
             // instead of `null`.
-            .signal = signal.map([](jsg::Ref<AbortSignal>& s) -> kj::Maybe<jsg::Ref<AbortSignal>> {
-    return s.addRef();
+            .signal =
+                signal.map([&js](jsg::Ref<AbortSignal>& s) -> kj::Maybe<jsg::Ref<AbortSignal>> {
```
npaun, line 1356: This stuff gives me a headache. Maybe we should be serializing this signal instead of signal? Otherwise the incoming request's signal would never get passed over RPC in the first place.

src/workerd/api/http.c++
```diff
@@ -1352,8 +1352,13 @@ void Request::serialize(jsg::Lock& js,
...
+    if (FeatureFlags::get(js).getRequestSignalPassthrough()) {
+      return s.addRef();
+    } else {
+      return kj::none;
```
kentonv, line 1360: I think we should only be dropping the signal here if it came from the original incoming request (just like how `fetch()` ignores the signal if it came from the original incoming request).

Yes, it's true that `Request` objects explicitly constructed with a `signal` will fail to serialize, but that has been the case all along. The purpose of the `request_signal_passthrough` compat flag is to avoid causing breakage from the introduction of `request.signal` on incoming requests.

npaun: You're right. I'll change it.

src/workerd/api/http.c++
kentonv: I don't really understand what you mean here. Can you elaborate?

src/workerd/io/compatibility-date.capnp
```diff
@@ -749,4 +749,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatEnableDate("2025-05-19")
       $compatDisableFlag("disable_navigator_language");
   # Enables Navigator.language API.
+
+  enableRequestSignal @87 :Bool
```
npaun, line 753: @kentonv How do you feel about making the `enable_request_signal` flag not be experimental? So no passthrough, no RPC AbortSignal, just able to set a handler for client disconnect.

src/workerd/api/http.c++
npaun: I pushed a commit to move the incoming request's signal to the `signal` field instead of `thisSignal` as I think it made the code simpler.

danlapid:
@npaun ready to go?

npaun:
Yeah, let's do this.

PR #4092, opened 2025-05-06
Python: Remove FinalziationRegistry stub

hoodmane:
Since @ketanhwr has enabled it: #3560


danlapid:
We should leave the stub as is.
What if the compat flag is not enabled?

hoodmane:
Ah didn't realize it was

PR #4095, opened 2025-05-06
[o11y] Update Streaming Tail Worker Attributes definition

fhanau:
This should be easier to process in the tail worker. Note that only the JS side is updated and we still use a plain array internally, this makes this change simpler and avoids needing e.g. capnp definition changes.

Alongside this change, update the name of the handler for these events to "attributes", which is more descriptive.

### DIFF
--- a/types/defines/trace.d.ts
+++ b/types/defines/trace.d.ts
@@ -124,7 +124,7 @@ interface Hibernate {
 interface SpanOpen {
   readonly type: "spanOpen";
   readonly name: string;
-  readonly info?: FetchEventInfo | JsRpcEventInfo | Attribute[];
+  readonly info?: FetchEventInfo | JsRpcEventInfo | Attributes;
 }
 
 interface SpanClose {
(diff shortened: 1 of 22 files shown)


types/defines/trace.d.ts
```diff
@@ -165,23 +165,27 @@ interface Link {
 }
 
 interface Attribute {
-  readonly type: "attribute";
   readonly name: string;
   readonly value: string | string[] | boolean | boolean[] | number | number[] | bigint | bigint[];
 }
 
+interface Attributes {
```
fhanau, line 172: @jmorrell-cloudflare is this in line with what you had in mind? Note that I'm dropping the "attribute" type from individual Attribute objects (feels redundant if Attributes already has it).

jmorrell-cloudflare: Yep, this looks exactly like I would expect. Thank you!

jasnell (approved):

penalosa (approved):

danlapid (approved):

PR #4097, opened 2025-05-06
Initial impl of the internal worker-fs API

jasnell:
First batch of changes separated out from https://github.com/cloudflare/workerd/pull/4029

This is split off from a larger PR to hopefully make it easier to review. Unfortunately, splitting it off does mean that that there's not a lot of context available for how this is intended to be used. It is expected that this internal part may need to evolve further as the rest of the implementation is fleshed out. While an internal PR will be necessary in order for this to land, it is not expected that this initial PR will be accessible/usable by users until the rest of the implementation is complete. Please refer to the [larger PR](https://github.com/cloudflare/workerd/pull/4029) for context on how these new APIs would be intended to be used.


src/workerd/io/BUILD.bazel
```diff
@@ -52,6 +52,7 @@ wd_cc_library(
         "io-util.c++",
         "tracer.c++",
         "worker.c++",
+        "worker-fs.c++",
```
jasnell, line 55: Review note: for the time being I won't be trying to separate this out into a separate dep. That's something that can be done later in a separate PR

anonrig (approved):

src/workerd/server/bundle-fs.c++
```diff
@@ -0,0 +1,80 @@
...
+            case config::Worker::Module::OBSOLETE: {
+              // Just ignore it.
+              break;
+            }
+          }
+        }
+        break;
+      }
+      case config::Worker::Which::SERVICE_WORKER_SCRIPT: {
+        addToBundle("worker"_kj, conf.getServiceWorkerScript().asBytes());
```
danlapid, line 67: Maybe:
```suggestion
        addToBundle("worker.js"_kj, conf.getServiceWorkerScript().asBytes());
```
?
I'm not really sure but not having an extension feels odd to me.
This has to be a javascript file right?

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,638 @@
...
+  //
+  // Note that tryOpen doubles as both an existence check and a stat operation.
+  // When the createAs parameter is not specified, the method will return
+  // kj::none if the file or directory does not exist.
+  //
+  // If the path identifies a symbolic link, the symbolic link will be resolved
+  // and the target, if it exists, will be returned. If the target does not exist,
+  // kj::none will be returned.
+  virtual kj::Maybe<kj::OneOf<kj::Rc<File>, kj::Rc<Directory>, kj::Rc<SymbolicLink>>> tryOpen(
+      jsg::Lock& js, kj::PathPtr path, kj::Maybe<OpenOptions> options = kj::none) = 0;
```
danlapid, line 302: Nit:
The following is cleaner and conveys the purpose at the API level without having to go into the implementation.
It also simplifies the implementation a tiny tiny bit
```suggestion
      jsg::Lock& js, kj::PathPtr path, OpenOptions options = {}) = 0;
```

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,638 @@
...
+  // if the directory is not read only.
+  // If the node is a symlink, the symlink will be removed but the target will
+  // be left intact.
+  // The path must be relative to the current directory.
+  // Note that this method will only remove the node from this directory. If the
+  // node as references in other directories, those will not be removed.
+  // If the target is a symbolic link, the symbolic link will be removed but
+  // the target will not be removed.
+  virtual bool remove(
+      jsg::Lock& js, kj::PathPtr path, kj::Maybe<RemoveOptions> options = kj::none) = 0;
```
danlapid, line 329: Same
```suggestion
      jsg::Lock& js, kj::PathPtr path, RemoveOptions options = {}) = 0;
```

src/workerd/io/worker-fs.h
```diff
@@ -0,0 +1,638 @@
...
+    using Map = kj::HashMap<kj::String,
+        kj::OneOf<kj::Rc<File>, kj::Rc<Directory>, kj::Own<Directory::Builder>>>;
+    using Entry = Map::Entry;
+
+   private:
+    Map entries;
+  };
+};
+
+// The equivalent to a symbolic link. A symlink holds a reference to both the
```
danlapid, line 376: Nit:
The comment here could be clarified as the symlink does *not* hold a reference to the target node, it just holds its path.
This mimics linux' behaviour under which the target node can be deleted and recreated and the symlink will still be valid.

src/workerd/io/worker-fs.c++
```diff
@@ -0,0 +1,1319 @@
...
+  KJ_IF_SOME(ret, root->tryOpen(js, getTargetPath())) {
+    KJ_SWITCH_ONEOF(ret) {
+      KJ_CASE_ONEOF(file, kj::Rc<File>) {
+        return kj::Maybe<kj::OneOf<kj::Rc<File>, kj::Rc<Directory>>>(kj::mv(file));
+      }
+      KJ_CASE_ONEOF(dir, kj::Rc<Directory>) {
+        return kj::Maybe<kj::OneOf<kj::Rc<File>, kj::Rc<Directory>>>(kj::mv(dir));
+      }
+      KJ_CASE_ONEOF(link, kj::Rc<SymbolicLink>) {
+        // A symbolic link won't resolve to another symbolic link.
```
danlapid, line 1010: Why not?

jasnell: The `resolve(...)` method here follows all symbolic links in the path, so when it encounters a symlink as the path is being processed, it will attempt to resolve it into it's target or return `kj::none`. If you want the symlink itself, then use `tryOpen(...)` on the directory

src/workerd/io/worker-fs.h
jasnell: The comment here is that the symlink holds a reference to the VirtualFileSystem and a target path, not that it holds a reference to the target node.

src/workerd/io/worker-fs.c++
danlapid: Sweet, makes sense, would love to see this answer inlined into the comment next time you edit this code.

PR #4099, opened 2025-05-07
Add proper error message prefix when output gate is broken

shrima-cf:


src/workerd/io/actor-sqlite.c++
```diff
@@ -206,7 +206,9 @@ void ActorSqlite::onCriticalError(
     KJ_IF_SOME(e, maybeException) {
       broken.emplace(kj::mv(e));
     } else {
-      broken.emplace(JSG_KJ_EXCEPTION(FAILED, Error, errorMessage));
+      kj::Exception exception = JSG_KJ_EXCEPTION(FAILED, Error, errorMessage);
+      exception.setDescription(kj::str("broken.outputGateBroken; ", exception.getDescription()));
```
kentonv, line 210: Huh, I'm surprised this isn't added somewhere downstream but looking at the code I don't see where it would be, so I guess this makes sense.

Maybe we should add the prefix to `maybeException` as well, if it was provided?

kentonv (approved):

src/workerd/io/actor-sqlite.c++
shrima-cf: I had this question as well, but given the other exception is a pass through of VFS exception from before, was not sure if I should add the prefix or not.
Maybe it is not such a big difference since the actual description is persisted anyway. I will update

PR #4102, opened 2025-05-07
Fix the list of proxied special methods for DO/RPC in Python.

dom96:
Methods like `scheduled` aren't special in DO/RPC, so we do not proxy them to `on_scheduled`. But there are also other methods that we haven't been proxying.

This is just a quick PR to fix this.

### DIFF
--- a/src/pyodide/python-entrypoint-helper.ts
+++ b/src/pyodide/python-entrypoint-helper.ts
@@ -232,17 +232,30 @@ async function initPyInstance(
   return res as PyModule;
 }
 
+// https://developers.cloudflare.com/workers/runtime-apis/rpc/reserved-methods/
+const SPECIAL_HANDLER_NAMES = ['fetch', 'connect'];
+const SPECIAL_DO_HANDLER_NAMES = [
+  'alarm',
+  'webSocketMessage',
+  'webSocketClose',
+  'webSocketError',
+];
+
 function makeEntrypointProxyHandler(
-  pyInstancePromise: Promise<PyModule>
+  pyInstancePromise: Promise<PyModule>,
+  isDurableObject: boolean
 ): ProxyHandler<any> {
   return {
     get(target, prop, receiver): any {
       if (typeof prop !== 'string') {
         return Reflect.get(target, prop, receiver);
       }
+
       // Proxy calls to `fetch` to methods named `on_fetch` (and the same for other handlers.)
-      const isKnownHandler = SUPPORTED_HANDLER_NAMES.includes(prop);
-      if (isKnownHandler) {
+      const isKnownHandler = SPECIAL_HANDLER_NAMES.includes(prop);
+      const isKnownDoHandler =
+        isDurableObject && SPECIAL_DO_HANDLER_NAMES.includes(prop);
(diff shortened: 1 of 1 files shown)


src/pyodide/python-entrypoint-helper.ts
```diff
@@ -232,16 +232,28 @@ async function initPyInstance(
   return res as PyModule;
 }
 
+// https://developers.cloudflare.com/workers/runtime-apis/rpc/reserved-methods/
+const SPECIAL_HANDLER_NAMES = ['fetch', 'connect'];
+const SPECIAL_DO_HANDLER_NAMES = [
+  'alarm',
+  'webSocketMessage',
+  'webSocketClose',
+  'webSocketError',
```
hoodmane, line 241: should we also translate these to snake case e.g., `websocket_message`?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -265,14 +277,18 @@ function makeEntrypointProxyHandler(
 }
 
 function makeEntrypointClass(className: string, classKind: AnyClass): any {
+  const isDurableObject = classKind.name == 'DurableObject';
```
hoodmane, line 280:
```suggestion
  const isDurableObject = classKind.name === 'DurableObject';
```

src/pyodide/python-entrypoint-helper.ts
danlapid: Please keep it same as JS, even if it is a tiny bit less python idiomatic

danlapid: or special case both and fail startup if both are defined

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -232,17 +232,30 @@ async function initPyInstance(
...
         return Reflect.get(target, prop, receiver);
       }
+
       // Proxy calls to `fetch` to methods named `on_fetch` (and the same for other handlers.)
-      const isKnownHandler = SUPPORTED_HANDLER_NAMES.includes(prop);
-      if (isKnownHandler) {
+      const isKnownHandler = SPECIAL_HANDLER_NAMES.includes(prop);
+      const isKnownDoHandler =
+        SPECIAL_DO_HANDLER_NAMES.includes(prop) && isDurableObject;
+      if (isKnownHandler || isKnownDoHandler) {
```
hoodmane, line 258: It's a bit more efficient to combine it into one condition (so we get short circuit evaluation):
```suggestion
      if (SPECIAL_HANDLER_NAMES.includes(prop) || isDurableObject && SPECIAL_DO_HANDLER_NAMES.includes(prop)) {
```
If not this, at least it should be `isDurableObject && SPECIAL_DO_HANDLER_NAMES.includes(prop)`.

hoodmane (approved):

src/pyodide/python-entrypoint-helper.ts
dom96: Yeah... I prefer the clearer code in this case. But I will flip the order of `isKnownDoHandler`.

PR #4103, opened 2025-05-07
Log whenever AbortSignal is used over RPC

npaun:
We're going to have to put AbortSignal RPC support behind a compat flag ASAP, because it spuriously causes the outcome of a request to show `exception` even though things work fine. First we need to make sure no one is using this feature.

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -821,7 +821,7 @@ void AbortSignal::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
   auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
       "AbortSignal can only be serialized for RPC.");
 
-  LOG_PERIODICALLY(WARNING, "NOSENTRY ", "An AbortSignal was serialized for RPC");
+  LOG_PERIODICALLY(WARNING, "NOSENTRY An AbortSignal was serialized for RPC");
 
   auto externalHandler = dynamic_cast<RpcSerializerExternalHandler*>(&handler);
   JSG_REQUIRE(
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -820,6 +820,9 @@ void AbortSignal::triggerAbort(
 void AbortSignal::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
   auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
       "AbortSignal can only be serialized for RPC.");
+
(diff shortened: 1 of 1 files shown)


jasnell (approved):

src/workerd/api/basics.c++
```diff
@@ -820,6 +820,9 @@ void AbortSignal::triggerAbort(
 void AbortSignal::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
   auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
       "AbortSignal can only be serialized for RPC.");
+
+  KJ_LOG(WARNING, "An AbortSignal was serialized for RPC");
```
jasnell, line 824: Nit: should should probably be `NOSENTRY` and maybe even `LOG_PERIODICALLY`

src/workerd/api/basics.c++
```diff
@@ -852,6 +855,9 @@ jsg::Ref<AbortSignal> AbortSignal::deserialize(
     jsg::Lock& js, rpc::SerializationTag tag, jsg::Deserializer& deserializer) {
   auto& handler = KJ_REQUIRE_NONNULL(
       deserializer.getExternalHandler(), "got AbortSignal on non-RPC serialized object?");
+
+  KJ_LOG(WARNING, "An AbortSignal was deserialized from RPC");
```
kentonv, line 859: We can probably skip logging deserialization since there would necessarily already be a serialization just before it.

src/workerd/api/basics.c++
jasnell: If on the off chance there's already someone using this a bunch we could end up spamming sentry needlessly. These are most often best left as NOSENTRY as long as someone is actively watching for them

npaun: Made it both PERIODIC and NOSENTRY

src/workerd/api/basics.c++
```diff
@@ -820,6 +820,9 @@ void AbortSignal::triggerAbort(
 void AbortSignal::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
   auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
       "AbortSignal can only be serialized for RPC.");
+
+  LOG_PERIODICALLY(WARNING, "NOSENTRY ", "An AbortSignal was serialized for RPC");
```
kentonv, line 824: I'm not aware of any reason to split the string here. It's probably harmless but it could result in weird formatting, I'm not sure, so I'd not do that.

```suggestion
  LOG_PERIODICALLY(WARNING, "NOSENTRY An AbortSignal was serialized for RPC");
```

kentonv (approved):

PR #4106, opened 2025-05-07
R2-3031: add `.bytes()` support in `R2ObjectBody`

helloimalastair:
Reported by @DaniFoldi


jasnell (approved):
A test would also be welcome.

jasnell (approved):

helloimalastair:
@danlapid I have bumped the snapshot, so should be good now

src/workerd/api/r2-test.js
```diff
@@ -107,6 +107,15 @@ async function compareResponse(res, { head, body } = {}) {
...
+    if (bytes) {
+      const expected = new Uint8Array(new TextEncoder().encode(body));
+      const actual = await (await res).bytes();
+      assert.equal(expected.byteLength, actual.byteLength);
+      for (let i = 0; i < expected.byteLength; i++) {
+        assert.equal(expected[i], actual[i]);
+      }
+      return;
+    }
     assert.strictEqual(await (await res).text(), body);
```
petebacondarwin, line 119: Should this statement be in an `else` block? Doesn't calling `(await res).bytes()` drain the body so that calling `(await res).text()` will fail?

petebacondarwin (approved):
Approved for the typings on behalf of Wrangler team, but a question about the test...

src/workerd/api/r2-test.js
petebacondarwin: Doh! I missed the `return` statement. Cool

types/generated-snapshot/index.d.ts
danlapid, line 1: these shouldn't be here

PR #4108, opened 2025-05-08
Add clone method to ArtifactBundler::State

jasnell:


src/workerd/api/pyodide/pyodide.h
```diff
@@ -304,6 +304,12 @@ class ArtifactBundler: public jsg::Object {
           storedSnapshot(kj::none),
           existingSnapshot(kj::mv(existingSnapshot)),
           isValidating(isValidating) {};
+
+    kj::Own<State> clone() {
+      return kj::heap<State>(packageManager,
+          existingSnapshot.map(
+              [](kj::Array<const kj::byte>& data) { return kj::heapArray<const kj::byte>(data); }));
```
dom96, line 311: It's more than likely false here but I think for completeness it's worth including it in case there are future changes that require it.
```suggestion
              [](kj::Array<const kj::byte>& data) { return kj::heapArray<const kj::byte>(data); }), isValidating);
```

dom96 (approved):
Thanks!

PR #4109, opened 2025-05-08
Wrap `env` in Python DO/WorkerEntrypoint constructor.

dom96:
This `env` is passed via the constructor, this PR ensures that it is wrapped in our `EnvWrapper` which deals with nice RPC conversions.

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -917,6 +917,17 @@ def _wrap_subclass(cls):
             continue
         setattr(cls, k, _wrap_attr(v))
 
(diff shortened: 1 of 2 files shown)


hoodmane:
Can you resolve conflicts? Since we got rid of `_wrap_subclass` it's not clear how this should look anymore.

dom96:
The _wrap_subclass is still needed in this case afaik. We need to override the class constructor and that seems to be the way to do it.

hoodmane:
Conflicts still need to be resolved.

dom96:
Resolved.

hoodmane (approved):
Looks good to me.

PR #4116, opened 2025-05-08
Experimental implementation of the Web FileSystem api

jasnell:
Separated out from the larger #4029 PR. This Pr focuses specifically on the Web FileSystem API impl

This will remain behind an experimental compat flag until the impl is complete and there are web platform tests available for it.

As this is behind an experimental compat flag, this can be safely landed before having the appropriate web platform tests landed.

### DIFF
--- a/src/workerd/api/filesystem.c++
+++ b/src/workerd/api/filesystem.c++
@@ -1,0 +1,449 @@
+#include "filesystem.h"
+
+#include "blob.h"
+
+#include <workerd/api/streams/standard.h>
+
+namespace workerd::api {
+
+// =======================================================================================
+// Implementation of the Web File System API
+
+namespace {
+constexpr bool isValidFileName(kj::StringPtr name) {
+  return name.size() > 0 && name != "."_kj && name != ".."_kj && name.find("/"_kj) == kj::none;
+}
+}  // namespace
+
+jsg::Promise<jsg::Ref<FileSystemDirectoryHandle>> StorageManager::getDirectory(
+    jsg::Lock& js, const jsg::TypeHandler<jsg::Ref<jsg::DOMException>>& exception) {
+  KJ_IF_SOME(vfs, workerd::VirtualFileSystem::tryGetCurrent(js)) {
+    return js.resolvedPromise(
+        js.alloc<FileSystemDirectoryHandle>(jsg::USVString(), vfs.getRoot(js)));
+  }
+
+  auto ex = js.domException(kj::str("SecurityError"), kj::str("No current virtual file system"));
+  return js.rejectedPromise<jsg::Ref<FileSystemDirectoryHandle>>(exception.wrap(js, kj::mv(ex)));
+}
+
+FileSystemDirectoryHandle::FileSystemDirectoryHandle(
+    jsg::USVString name, kj::Rc<workerd::Directory> inner)
+    : FileSystemHandle(kj::mv(name)),
+      inner(kj::mv(inner)) {}
+
+jsg::Promise<bool> FileSystemDirectoryHandle::isSameEntry(
+    jsg::Lock& js, jsg::Ref<FileSystemHandle> other) {
+  KJ_IF_SOME(other, kj::dynamicDowncastIfAvailable<FileSystemDirectoryHandle>(*other)) {
+    return js.resolvedPromise(other.inner.get() == inner.get());
+  }
+  return js.resolvedPromise(false);
+}
+
+jsg::Promise<jsg::Ref<FileSystemFileHandle>> FileSystemDirectoryHandle::getFileHandle(jsg::Lock& js,
+    jsg::USVString name,
+    jsg::Optional<FileSystemGetFileOptions> options,
+    const jsg::TypeHandler<jsg::Ref<jsg::DOMException>>& exception) {
+  JSG_REQUIRE(isValidFileName(name), TypeError, "Invalid file name");
+  kj::Maybe<FsType> createAs;
+  KJ_IF_SOME(opts, options) {
+    if (opts.create) createAs = FsType::FILE;
+  }
+  KJ_IF_SOME(node,
+      inner->tryOpen(js, kj::Path({name}),
+          Directory::OpenOptions{
+            .createAs = createAs,
+          })) {
+    KJ_SWITCH_ONEOF(node) {
+      KJ_CASE_ONEOF(file, kj::Rc<workerd::File>) {
+        return js.resolvedPromise(js.alloc<FileSystemFileHandle>(kj::mv(name), kj::mv(file)));
+      }
+      KJ_CASE_ONEOF(dir, kj::Rc<workerd::Directory>) {
+        auto ex =
+            js.domException(kj::str("TypeMismatchError"), kj::str("File name is a directory"));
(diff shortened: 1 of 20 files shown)


src/workerd/api/filesystem.h
```diff
@@ -0,0 +1,333 @@
...
+  // We don't yet implement any form of serialization for these objects.
+  // Since each worker has its own file system, it might be a bit weird
+  // as the received file on the other side would not actually be in that
+  // destination worker's file system... that is, it would be an orphaned
+  // file handle. It would still likely be useful tho. As a follow up
+  // step, we will implement serialization/deserialization of these objects.
+ public:
+  FileSystemHandle(jsg::USVString name): name(kj::mv(name)) {}
+  jsg::USVString getName(jsg::Lock& js) {
+    return jsg::USVString(kj::str(name));
```
danlapid, line 33: Why do we need the double conversion?
Isn't it enough to just return name?

jasnell: `jsg::USVString` does not have an equivalent to `kj::StringPtr` and the type mapping requires `jsg::USVString` for it to get written correctly. It's an annoyance at this point that I plan to get back around to simplifying but not in this PR.

src/workerd/api/filesystem.h
```diff
@@ -0,0 +1,333 @@
...
+    auto& entry = state.entries[state.index++];
+
+    if constexpr (kj::isSameType<Type, EntryIteratorType>()) {
+      return EntryType(js.accountedUSVString(entry->getName(js)), entry.addRef());
+    } else if constexpr (kj::isSameType<Type, KeyIteratorType>()) {
+      return js.accountedUSVString(entry->getName(js));
+    } else if constexpr (kj::isSameType<Type, ValueIteratorType>()) {
+      return entry.addRef();
+    } else {
+      KJ_UNREACHABLE;
```
danlapid, line 224: this should be a static_assert(false) or similar to denote that it is compile time proven to never be the case.

src/workerd/api/filesystem.h
```diff
@@ -0,0 +1,333 @@
...
+    State(kj::Rc<workerd::File> file): file(kj::mv(file)) {}
+  };
+
+  FileSystemWritableFileStream(
+      kj::Own<WritableStreamController> controller, kj::Rc<State> sharedState);
+
+  static jsg::Ref<FileSystemWritableFileStream> constructor() = delete;
+
+  struct WriteParams {
+    kj::String type;  // one of: write, seek, truncate
```
danlapid, line 249: General nit:
Why not use JsStrings everywhere?
Non blocking of course

src/workerd/api/filesystem.h
```diff
@@ -0,0 +1,333 @@
...
+    jsg::Optional<double> size;
```
danlapid, line 250: I might be missing something but why is this double?
I think it should be uint64_t or even uint32_t

https://fs.spec.whatwg.org/#api-filesystemwritablefilestream
The SPEC seems to say unsigned long long which is not a double and double to size_t conversions in the implementation are rather ugly

src/workerd/api/filesystem.h
jasnell: those still end up needing to be converted to `kj::String` for comparison operations like what we see in `FileSystemWritableFileStream::write` (e.g. `if (params.type == "write"_kj) {` unless we created `JsStrings` to replace the `"write"_kj` part. I'd rather keep this as familiar as possible to our existing convention. If we want to convert more of the `kj::String` uses to `JsString` en masse then we can do so in a separate PR.

src/workerd/api/filesystem.h
jasnell: `uin64_t` maps to `bigint` in our type wrapper system. It could be `uing32_t` but the spec says these should be `unsigned long long` and not `unsigned long` which is what `uint32_t` maps to.  If we're going to map to `uint32_t` then we should change the places where we are using `size_t` to using `uint32_t` so that everything is consistent.

jasnell: Use of `uint32_t` ought to be fine throughout since that'll easily fit anything that fits without our memory limits unless we decide to support large sparse files at any point (which I wouldn't expect)

danlapid: Yeah going with uint32 makes sense to me considering our memory limitations 
Double feels wrong

src/workerd/api/filesystem.c++
```diff
@@ -0,0 +1,442 @@
...
+    jsg::USVString name,
+    jsg::Optional<FileSystemGetFileOptions> options,
+    const jsg::TypeHandler<jsg::Ref<jsg::DOMException>>& exception) {
+  JSG_REQUIRE(isValidFileName(name), TypeError, "Invalid file name");
+  kj::Maybe<FsType> createAs;
+  KJ_IF_SOME(opts, options) {
+    if (opts.create) createAs = FsType::FILE;
+  }
+  KJ_IF_SOME(node,
+      inner->tryOpen(js, kj::Path({name}),
```
danlapid, line 52: I regret the original decision to let tryOpen return `kj::Maybe<kj::OneOf<kj::Rc<File>, kj::Rc<Directory>, kj::Rc<SymbolicLink>>>` with the comment 
```
  // If the path identifies a symbolic link, the symbolic link will be resolved
  // and the target, if it exists, will be returned. If the target does not exist,
  // kj::none will be returned.
```
Is there any reason not to amend that to return `kj::Maybe<kj::OneOf<kj::Rc<File>, kj::Rc<Directory>>>` instead and skip all of the 
```
      KJ_CASE_ONEOF(link, kj::Rc<workerd::SymbolicLink>) {
        KJ_UNREACHABLE;
      }
```
These are counter-intuitive and confusing without real value.

src/workerd/io/compatibility-date.capnp
```diff
@@ -749,4 +749,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatEnableDate("2025-05-19")
       $compatDisableFlag("disable_navigator_language");
   # Enables Navigator.language API.
+
+  webFileSystem @87 :Bool
+      $compatEnableFlag("enable_web_file_system")
+      $experimental;
+  # Enables the experimental Web File System API.
```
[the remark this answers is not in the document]
jasnell: There's a link to the spec in the filesystem.h

src/workerd/api/filesystem.c++
jasnell: These are *currently* `KJ_UNREACHABLE` because we're not doing anything with symlinks yet. Once we fill out the full node:fs implementation these may not remain `KJ_UNREACHABLE`. Once the implementation is done I'll be going back and revisiting these.

anonrig (approved):
We can now implement URL.createObjectURL as well

anonrig:
@npaun I think we should enable relevant WPT as well

jasnell:
Making a few more small tweaks before this is fully ready. Namely taking care of a couple of the outstanding todos such as making the dir handle an async iterable rather than a sync iterable. And yeah, once this lands we'll need to look at the web platform tests. It'll be kept experimental until we're sure those are ready to go. We likely won't be able to be 100% compatible so it might take some iteration on the tests to lock down.

PR #4117, opened 2025-05-08
simplify type-wrapper by removing support from certain types

anonrig:
This pull-request removes support for using v8::Isolate* type as a parameter. Also starting from this pull-request, we no longer check for the parameter types in order to detect if a function is fast api compatible or not, because in reality everything is v8::Local<v8::Value>...

- v8::Isolate* - we can do the same thing via jsg::Lock&, and actually v8::Isolate* parameter is equal to js.v8Isolate parameter. so there is no need to have this

### DIFF
--- a/src/workerd/io/promise-wrapper-test.c++
+++ b/src/workerd/io/promise-wrapper-test.c++
@@ -56,9 +56,9 @@ struct CaptureThrowContext: public jsg::Object, public ContextGlobal {
     JSG_FAIL_REQUIRE(TypeError, "boom");
   }
 
-  static kj::Promise<void> staticTest3(v8::Isolate* isolate) {
+  static kj::Promise<void> staticTest3(jsg::Lock& js) {
     // Tests that JsExceptionThrown is handled properly.
-    jsg::throwTypeError(isolate, "boom"_kj);
+    jsg::throwTypeError(js.v8Isolate, "boom"_kj);
   }
 
   kj::Promise<void> getTest() {
--- a/src/workerd/jsg/fast-api-test.c++
+++ b/src/workerd/jsg/fast-api-test.c++
@@ -258,27 +258,6 @@ KJ_TEST("Fast methods properly catch JSG_FAIL_REQUIRE errors") {
 }
 
 KJ_TEST("isFastMethodCompatible Detection") {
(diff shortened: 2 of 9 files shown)


jasnell:
> `v8::FunctionCallbackInfo` 

We have a number of places where this is explicitly used. I need to dig into more detail here but not sure this PR is complete as is?

anonrig:
> > `v8::FunctionCallbackInfo` 
> 
> 
> 
> We have a number of places where this is explicitly used. I need to dig into more detail here but not sure this PR is complete as is?

@jasnell Except the unused variable errors, it builds and tests successfully locally.

jasnell:
No, I mean if the intent is to eliminate use of `FunctionCallbackInfo` then it seems incomplete. There are some constructors, for instance, that take `FunctionCallbackInfo` as an argument that aren't updated (see `WorkerEntrypoint` in `worker-rpc.h`). The PR description is not super clear what all is intended here. Also, it's confusing that this says it is simplifying type-wrapper but if we look at the changes it's not removing or changing existing stuff it's adding new stuff, so the impact here is not super clear/obvious. More description/comments would be helpful for context.

anonrig:
> I'm OK with removing the implicit fill-in of `v8::Isolate*` since `jsg::Lock&` makes it obsolete.
> 
> But `FunctionCallbackInfo` is still widely used. In theory it would be nice to remove the uses, but many of them will need some other abstraction that provides similar functionality.

Agreed. This pull-request still keeps the FunctionCallbackInfo usage, but only removes tests that call them from a method. We currently only need to support FunctionCallbackInfo on constructor method overrides (due to workerd-rpc and dom-exception constructor class usages). We can remove the rest of the implementation. Once this PR lands, no-one can use FunctionCallbackInfo in a JSG_METHOD.

kentonv:
Why would we want to support different things for constructors and methods? That's just going to cause confusion later. Let's keep it uniform.

anonrig:
> Why would we want to support different things for constructors and methods? That's just going to cause confusion later. Let's keep it uniform.

Removed those deletions and kept the change limited to the removal of v8::Isolate* @kentonv

kentonv (approved):

PR #4121, opened 2025-05-09
adopting downstream's rustfmt rules

mikea:
will keep everything consistent.

Also extracted rust_toolchains.bzl while I'm at it (separate commit)

### DIFF
--- a/build/rust_toolchains.bzl
+++ b/build/rust_toolchains.bzl
@@ -7,6 +7,7 @@ RUST_STABLE_VERSION = "1.86.0"  # LLVM 19
 
 RUST_NIGHTLY_VERSION = "nightly/2025-02-20"
 
+# List of additional triples to be configured on top of the local platform triple
 RUST_TARGET_TRIPLES = [
     # Add support for macOS cross-compilation
     "x86_64-apple-darwin",
@@ -17,7 +18,7 @@ RUST_TARGET_TRIPLES = [
 def rust_toolchains():
     rules_rust_dependencies()
     rust_register_toolchains(
-        edition = "2021",
+        edition = "2024",
         extra_target_triples = RUST_TARGET_TRIPLES,
         versions = [
             RUST_STABLE_VERSION,
--- a/build/rust_toolchains.bzl
+++ b/build/rust_toolchains.bzl
@@ -1,0 +1,41 @@
+load("@rules_rust//crate_universe:repositories.bzl", "crate_universe_dependencies")
+load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains", "rustfmt_toolchain_repository")
+load("@rules_rust//tools/rust_analyzer:deps.bzl", "rust_analyzer_dependencies")
+load("//deps/rust/crates:crates.bzl", "crate_repositories")
+
+RUST_STABLE_VERSION = "1.86.0"  # LLVM 19
+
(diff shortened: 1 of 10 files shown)


build/rust_toolchains.bzl
```diff
@@ -0,0 +1,41 @@
+load("@rules_rust//crate_universe:repositories.bzl", "crate_universe_dependencies")
+load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains", "rustfmt_toolchain_repository")
+load("@rules_rust//tools/rust_analyzer:deps.bzl", "rust_analyzer_dependencies")
+load("//deps/rust/crates:crates.bzl", "crate_repositories")
+
+RUST_STABLE_VERSION = "1.86.0"  # LLVM 19
+
+RUST_NIGHTLY_VERSION = "nightly/2025-02-20"
+
+RUST_TARGET_TRIPLES = [
```
fhanau, line 10: Nit: rename to RUST_EXTRA_TARGET_TRIPLES to clarify that this is the list of additional triples to be configured on top of the local platform triple, not the list of all platforms we support.

build/rust_toolchains.bzl
```diff
@@ -0,0 +1,41 @@
...
+        edition = "2021",
+        extra_target_triples = RUST_TARGET_TRIPLES,
+        versions = [
+            RUST_STABLE_VERSION,
+            RUST_NIGHTLY_VERSION,
+        ],
+    )
+
+    for t in RUST_TARGET_TRIPLES:
+        rustfmt_toolchain_repository(
```
fhanau, line 29: Not sure if this will work as intended: Here we set up toolchains for the two extra triples, but not for the local target triple. Would having one toolchain for just the local triple be sufficient here? If I interpret the docs correctly we might just be able to drop the exec_triple parameter then.

fhanau (approved):

build/rust_toolchains.bzl
mikea: hm. let me try it.

src/rust/cxx-integration-test/lib.rs
```diff
@@ -1,12 +1,15 @@
 //! Non-production crate to help test various aspects of rust/c++ integration.
 
-use std::{
-    io::{Error, ErrorKind},
-    pin::Pin,
-    time::Duration,
-};
-
-use tracing::{debug, error, info, trace, warn};
+use std::io::Error;
```
anonrig, line 3: Why do we prefer this?

build/rust_toolchains.bzl
mikea: `Error: rustfmt_toolchain_repository() missing 1 required positional argument: exec_triple`

mikea: frankly I am not sure how it picks up nightly rustfmt for me locally then since I don't specify x64 triple?

fhanau: Oof. So there's no way to set it up for just the current target? Maybe it's fine though, CI passed on this PR so maybe if we don't set up a toolchain for the host target rustfmt still works?

anonrig (approved):

src/rust/cxx-integration-test/lib.rs
mikea: less merge conflicts

PR #4122, opened 2025-05-09
get rid of v8 deprecation warning

anonrig:
Get's rid of the nasty deprecation warning I've been seeing for the past 9+ months...


anonrig:
Creates segfault on zlib...

kentonv:
This is something like the fourth or fifth time someone has tried to fix this. Every time leads to crashes in production and a rollback.

Issue #4089, opened 2025-05-05
wd_test should create a variant of each test with all autogates enabled


npaun:
We have something similar in the internal repo.

PR #4124, opened 2025-05-09
add `WORKERD_ENABLE_ALL_AUTOGATES` env variable

anonrig:
Ref: https://github.com/cloudflare/workerd/issues/4089

Adds an environment variable called "WORKERD_ENABLE_ALL_AUTOGATES" which enables all autogates regardless of what's set in the config. This is beneficial for testing.

### DIFF
--- a/build/kj_test.bzl
+++ b/build/kj_test.bzl
@@ -1,3 +1,5 @@
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+
 def kj_test(
         src,
         data = [],
@@ -6,8 +8,11 @@ def kj_test(
         size = "medium",
         **kwargs):
     test_name = src.removesuffix(".c++")
+    binary_name = test_name + "_binary"
+
     native.cc_test(
-        name = test_name,
+        name = binary_name,
+        testonly = True,
         srcs = [src],
         deps = [
             "@capnp-cpp//src/kj:kj-test",
@@ -18,6 +23,20 @@ def kj_test(
         }),
         data = data,
         tags = tags,
-        size = size,
         **kwargs
+    )
+    sh_test(
+        name = test_name,
+        size = size,
+        srcs = ["//build/fixtures:kj_test.sh"],
+        data = [binary_name],
+        args = ["$(location " + binary_name + ")"],
+    )
+    sh_test(
+        name = test_name + "@all-autogates-enabled",
+        size = size,
+        env = {"WORKERD_ALL_AUTOGATES": "1"},
+        srcs = ["//build/fixtures:kj_test.sh"],
+        data = [binary_name],
+        args = ["$(location " + binary_name + ")"],
     )
--- a/src/workerd/util/autogate.c++
+++ b/src/workerd/util/autogate.c++
@@ -5,6 +5,8 @@
 
 #include <workerd/util/sentry.h>
 
+#include <stdlib.h>
+
 #include <capnp/message.h>
 #include <kj/common.h>
 #include <kj/debug.h>
@@ -51,9 +53,9 @@ bool Autogate::isEnabled(AutogateKey key) {
   KJ_IF_SOME(a, globalAutogate) {
     return a.gates[(unsigned long)key];
   }
-  LOG_ERROR_PERIODICALLY(
-      kj::str("Autogates not initialised, check for ", key, " will have no effect"));
-  return false;
+
+  static const bool defaultResult = getenv("WORKERD_ALL_AUTOGATES") != nullptr;
+  return defaultResult;
 }
 
 void Autogate::initAutogate(capnp::List<capnp::Text>::Reader gates) {
--- a/build/fixtures/BUILD.bazel
+++ b/build/fixtures/BUILD.bazel
@@ -1,0 +1,1 @@
+exports_files(["kj_test.sh"])
--- a/build/fixtures/kj_test.sh
+++ b/build/fixtures/kj_test.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+set -euo pipefail
+"$@"
--- a/src/workerd/api/actor-state-test.c++
+++ b/src/workerd/api/actor-state-test.c++
@@ -29,7 +29,6 @@ struct ActorStateContext: public jsg::Object, public jsg::ContextGlobal {
 JSG_DECLARE_ISOLATE_TYPE(ActorStateIsolate, ActorStateContext);
 
 KJ_TEST("v8 serialization version tag hasn't changed") {
-  util::Autogate::initAutogateNamesForTest({"v8-fast-api"_kj});
   jsg::test::Evaluator<ActorStateContext, ActorStateIsolate> e(v8System);
   e.getIsolate().runInLockScope([&](ActorStateIsolate::Lock& isolateLock) {
     JSG_WITHIN_CONTEXT_SCOPE(isolateLock,
--- a/src/workerd/api/basics-test.c++
+++ b/src/workerd/api/basics-test.c++
@@ -87,7 +87,6 @@ JSG_DECLARE_ISOLATE_TYPE(BasicsIsolate,
     jsg::TypeWrapperExtension<PromiseWrapper>);
 
 KJ_TEST("EventTarget native listeners work") {
-  util::Autogate::initAutogateNamesForTest({"v8-fast-api"_kj});
   jsg::test::Evaluator<BasicsContext, BasicsIsolate, CompatibilityFlags::Reader> e(v8System);
   e.expectEval("testNativeListenersWork()", "boolean", "true");
 }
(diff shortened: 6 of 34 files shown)


npaun (approved):

jasnell:
I'm not entirely sure this is the right approach but could be convinced. The key reason is that this potentially misses the various combinations of some-autogates-on-some-autogates-off that could expose bugs and might be missed by this... For instance, the code might work if both autogates A and B are off or both are on, but does it work if only one of the two are on and the other is off? What I think might be better is a bazel configuration that automatically sets up variations of each test for each combination of autogates. That is, if we have three active autogates, then we would have eight test permutations covering each of the various combinations.

danlapid:
> I'm not entirely sure this is the right approach but could be convinced. The key reason is that this potentially misses the various combinations of some-autogates-on-some-autogates-off that could expose bugs and might be missed by this... For instance, the code might work if both autogates A and B are off or both are on, but does it work if only one of the two are on and the other is off? What I think might be better is a bazel configuration that automatically sets up variations of each test for each combination of autogates. That is, if we have three active autogates, then we would have eight test permutations covering each of the various combinations.

This PR mimics what we have internally, we lack the same detail with our internal tests.
It's a balance between coverage and test times.
It is reasonable to want better coverage but I think the request is out of scope for this PR.
This mimics correctly what we have internally and I think it is reasonable.

kentonv:
Should we consider using an environment variable instead, so that we can also run KJ tests with all-autogates?

> That is, if we have three active autogates, then we would have eight test permutations covering each of the various combinations.

Unfortunately this doesn't work because it doesn't scale. Even 8x is already probably too much, but in the internal codebase we have 21 autogates defined currently, in addition to the 3 in workerd, that's 24 autogates, so we'd be running each test 16 million times.

I think running one with all off and once with all on is a reasonable compromise. Autogates rarely interact with each other in practice.

anonrig:
> Should we consider using an environment variable instead, so that we can also run KJ tests with all-autogates?

I don't have any strong opinion. I'm fine with replacing the cli flag with an environment variable.

src/workerd/server/workerd.c++
```diff
@@ -754,6 +754,10 @@ class CliMain final: public SchemaFileImpl::ErrorReporter {
         "<const-name>", CLI_METHOD(setConstName));
   }
 
+  bool shouldEnableAllAutogates() {
```
danlapid, line 757: As is I don't see how this would work for KJ_TEST as Kenton requested.

danlapid: > Should we consider using an environment variable instead, so that we can also run KJ tests with all-autogates?

The idea of moving to env var was specifically so that it would be useful for KJ_TESTs

build/kj_test.bzl
```diff
@@ -6,18 +6,55 @@ def kj_test(
...
-        data = data,
-        tags = tags,
-        size = size,
-        **kwargs
+
+    # Create wrapper source file that will inject autogate initialization based on environment variable
+    wrapper_name = test_name + "_autogate_wrapper"
+    wrapper_file = wrapper_name + ".c++"
+
+    native.genrule(
```
kentonv, line 14: Wouldn't it be simpler to put some code in `autogate.c++` which checks for the env var when it finds the autogates uninitialized in `isEnabled()`?

build/kj_test.bzl
```diff
@@ -6,18 +6,55 @@ def kj_test(
...
+            }),
+            data = data,
+            tags = tags,
+            size = size,
+            env = env,
+            **kwargs
+        )
+
+    run_test(name = test_name)
+    run_test(name = test_name + "@all-autogates", env = {"WORKERD_ALL_AUTOGATES": "1"})
```
kentonv, line 60: Seems like each of these `run_test` calls will compile a separate copy of the binary, but with identical code... is there a way to compile one binary and run it twice?

anonrig: There are several problems in here.

1. Autogates doesn't initialize by default. This means if it's not enabled, calling isEnabled() will trigger a periodic error calling that autogate is not initialized. I think we need to change this. 
2. According to @mikea we can have a sh_test that helps us avoid unnecessary double compilation - https://bazel.build/reference/be/shell#sh_test

---

Regarding 1 - I recommend moving the initializing and checking for WORKERD_ALL_AUTOGATES to isEnabled static method.

```diff
index 3a924124..22bfc79c 100644
--- a/src/workerd/util/autogate.c++
+++ b/src/workerd/util/autogate.c++
@@ -53,9 +53,16 @@ bool Autogate::isEnabled(AutogateKey key) {
   KJ_IF_SOME(a, globalAutogate) {
     return a.gates[(unsigned long)key];
   }
-  LOG_ERROR_PERIODICALLY(
-      kj::str("Autogates not initialised, check for ", key, " will have no effect"));
-  return false;
+
+  if (getenv("WORKERD_ALL_AUTOGATES") != nullptr) {
+    Autogate::initAllAutogates();
+  } else {
+    Autogate::initAutogate({});
+  }
+
+  return globalAutogate.map([&key](auto& a) {
+    return a.gates[(unsigned long)key];
+  }).orDefault(false);
 }

 void Autogate::initAutogate(capnp::List<capnp::Text>::Reader gates) {
 ```

File: build/kj_test.bzl (hunk 2 above)
Comment (anonrig) [reply build/kj_test.bzl:14]: I just saw this. You're right.

File: src/workerd/util/autogate.c++ [hunk 4]
```diff
@@ -78,4 +100,18 @@ void Autogate::initAutogateNamesForTest(std::initializer_list<kj::StringPtr> gat
   Autogate::initAutogate(gates.asReader());
 }
 
+bool Autogate::isAllAutogatesEnabled() {
+  initializeIfNot();
```
anonrig, line 104: @kentonv I'm extremely unhappy about this. If the code doesn't check for isAllAutogatesEnabled() or isEnabled(), we won't initialize all autogates. I think we need to revert and use the first solution I've did, and initialize it on a separate cpp binary.

Referencing: https://github.com/cloudflare/workerd/commit/982740115da30613698a79b5787fd96d3604726a#diff-a66344587988d22c97a0e51d030a064225c81d3f12efe46b989b6c5483e96b5bR18-R22

src/workerd/util/autogate.c++
```diff
@@ -9,6 +9,8 @@
 #include <kj/common.h>
 #include <kj/debug.h>
 
+#include <cstdlib>
```
kentonv, line 12: Nitpick: Please use `stdlib.h`.

The entire purpose of `cstdlib` is to provide the symbols of `stdlib.h`, except declared in the `std::` namespace -- in order to avoid polluting the global namespace. But in practice it doesn't do this. In your code you called `getenv()` instead of `std::getenv()`, and it still worked. This is because all compilers have decided to have `cstdlib` (and similar headers) put the symbols in _both_ locations. But this defeats the purpose.

So KJ style is to just use `stdlib.h` and not expect symbols to be in the `std::` namespace.

src/workerd/util/autogate.c++
```diff
@@ -53,19 +57,41 @@ bool Autogate::isEnabled(AutogateKey key) {
   KJ_IF_SOME(a, globalAutogate) {
     return a.gates[(unsigned long)key];
   }
-  LOG_ERROR_PERIODICALLY(
-      kj::str("Autogates not initialised, check for ", key, " will have no effect"));
-  return false;
+
+  if (getenv("WORKERD_ALL_AUTOGATES") != nullptr) {
+    Autogate::initAllAutogates();
```
kentonv, line 62: Why do you need to init them at all?

Why not just do:

```
static const bool defaultResult = getenv("WORKERD_ALL_AUTOGATES") != nullptr;
return defaultResult;
```

Seems a lot simpler, and avoids thread-safety issues.

src/workerd/util/autogate.c++
anonrig: TIL. Thanks!

src/workerd/util/autogate.c++
anonrig: This is really nice. You're absolutely right. Thank you!

build/kj_test.bzl
```diff
@@ -18,6 +23,20 @@ def kj_test(
         }),
         data = data,
         tags = tags,
-        size = size,
         **kwargs
     )
+    sh_test(
```
fhanau, line 28: This doubles the amount of tests we will run. Have we considered the impact on CI performance this will have?

anonrig: I believe this is inevitable, if we want to increase our test coverage. (PS: EW is following the same pattern as well)

fhanau: It is, but it also has a lot more than 4 cores available 😄
I don't want to be blocking on this change, just saying that this means we'll have to think harder about CI performance with this change.

jasnell (approved):

build/kj_test.bzl
kentonv: The workerd tests take a tiny tiny fraction of the time of the internal codebase tests, so it's probably fine.

kentonv (approved):

anonrig:
(Rebased and force pushed)

PR #4127, opened 2025-05-11
Wrap top-level server-side RPC call in RevokerMembrane

ObsidianMinor:
Capabilities can be called past the end of an RPC session if that session is ended by the IoContext getting aborted. This doesn't result in any JS getting executed since the call to IoContext::run immediately returns the abort exception, but this might lead to calls attempting to schedule work after the IncomingRequest is getting torn down. This change updates ServerTopLevelMembrane to wrap the top-level capability in a RevokerMembrane on top of the existing CompletionMembrane.

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -526,8 +526,16 @@ bool EventTarget::dispatchEvent(jsg::Lock& js, jsg::Ref<Event> event) {
 // A wrapper for the AbortTrigger jsrpc client, that automatically sends a release() message once
 // the client is destroyed, informing the server that an abort will not be triggered in the future.
 class AbortTriggerRpcClient final {
+  struct Release {
+    rpc::AbortTrigger::Client client;
+    bool skipReleaseForTest;
+  };
+
  public:
-  AbortTriggerRpcClient(rpc::AbortTrigger::Client&& client): client(kj::mv(client)) {}
+  AbortTriggerRpcClient(
+      rpc::AbortTrigger::Client&& client, kj::Own<kj::PromiseFulfiller<Release>> releaseFulfiller)
+      : client(kj::mv(client)),
+        releaseFulfiller(kj::mv(releaseFulfiller)) {}
 
   kj::Promise<void> abort(kj::ArrayPtr<kj::byte> reason) {
     auto req = client.abortRequest(capnp::MessageSize{reason.size() / sizeof(capnp::word) + 8, 0});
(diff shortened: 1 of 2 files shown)


src/workerd/api/worker-rpc.c++
```diff
@@ -1870,7 +1871,10 @@ class JsRpcSessionCustomEventImpl::ServerTopLevelMembrane final: public capnp::M
     auto f = kj::mv(JSG_REQUIRE_NONNULL(
         doneFulfiller, Error, "Only one RPC method call is allowed on this object."));
     doneFulfiller = kj::none;
-    return capnp::membrane(kj::mv(target), kj::refcounted<CompletionMembrane>(kj::mv(f)));
+    auto completion = capnp::membrane(
+        kj::mv(target), kj::refcounted<CompletionMembrane>(kj::mv(f)));
+    return capnp::membrane(
```
ObsidianMinor, line 1876: Should the RevokerMembrane wrap the target before or after `CompletionMembrane`? I'd think it'd be after, but `JsRpcSessionCustomEventImpl::sendRpc` does it before.

src/workerd/api/basics.c++
```diff
@@ -536,27 +544,36 @@ class AbortTriggerRpcClient final {
...
-      LOG_EXCEPTION("abortTriggerReleaseRpc", exc);
-    });
+    releaseFulfiller->fulfill(Release{kj::mv(client), skipReleaseForTest});
   }
 
   bool skipReleaseForTest = false;
 
+  static IoOwn<AbortTriggerRpcClient> create(IoContext& ioctx, rpc::AbortTrigger::Client&& client) {
+    auto [promise, fulfiller] = kj::newPromiseAndFulfiller<Release>();
+    ioctx.addTask(promise.then([](Release release) -> kj::Promise<void> {
```
kentonv, line 554: Hmm, on one hand this strategy fixes a problem @npaun observed where, at IoContext destruction, the CustomEvent for the outgoing RPC session might be torn down before the AbortTrigger, causing the release call to fail.

On the other hand, though, I  think this `addTask` effectively means that after a Durable Object has sent an AbortSignal somewhere, it may be blocked from hibernating while it waits for the AbortTrigger to be GC'd? Or... maybe that problem exists either way, because the CustomEvent itself is registered as a task, and the AbortTrigger holds open the RPC session even after everything else has finished?

src/workerd/api/worker-rpc.c++
kentonv: Hmm, does this even belong inside `ServerTopLevelMembrane`, or should it be stacked on top?

I was expecting to see the `RevokerMembrane` added in `JsRpcSessionCustomEventImpl::run()`, stacked with `ServerTopLevelMembrane`. Though I guess I'm not sure if it makes a difference.

src/workerd/api/basics.c++
ObsidianMinor: Is this blocking or should I continue waiting for a review from @npaun?

PR #4130, opened 2025-05-12
Fixes regression in Python DO on_alarm handling.

dom96:
Specifically the error is:

```
workerd/io/worker.c++:2116: info: uncaught exception; source = Uncaught (in promise); stack = PythonError: Traceback (most recent call last):                                                                                                                            
  File "/lib/python3.12/site-packages/workers.py", line 905, in wrapper                                                                                                                                                                                                  
    py_args = [python_from_rpc(arg) for arg in args]                                                                                                                                                                                                                     
               ^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                      
  File "/lib/python3.12/site-packages/workers.py", line 765, in python_from_rpc                                                                                                                                                                                          
    result = obj.to_py(default_converter=_python_from_rpc_default_converter)                                                                                                                                                                                             
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                             
  File "pyodide-internal:generated/emscriptenSetup", line 691, in new_error                                                                                                                                                                                              
  File "/lib/python3.12/site-packages/workers.py", line 748, in _python_from_rpc_default_converter                                  
    raise TypeError(                                      
TypeError: Couldn't convert object to Python type, got AlarmInvocationInfo
```

This fixes it and adds a test case to reproduce this.

### Test Plan

```
```

### DIFF
--- a/src/workerd/server/tests/python/durable-object/worker.py
+++ b/src/workerd/server/tests/python/durable-object/worker.py
@@ -3,7 +3,9 @@
 #     https://opensource.org/licenses/Apache-2.0
 
 from asyncio import sleep
+from urllib.parse import urlparse
 
+from js import Date
 from workers import DurableObject, Response
 
 import pyodide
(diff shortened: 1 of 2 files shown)


src/workerd/server/tests/python/durable-object/worker.py
```diff
@@ -52,3 +69,11 @@ async def test(ctrl, env, ctx):
     if pyodide.__version__ != "0.26.0a2":
         res = await obj.jspi_method(9)
         assert res == 10
+
+    # Wait for alarm to get triggered.
+    while True:
```
hoodmane, line 74: Sleep 200 ms and then make the request only once?

dom96: I think that has a higher potential for being flaky. When running lots of tests at once the CI could easily make this alarm miss by multiple seconds.

Waiting indefinitely should prevent flaky false negatives at the expense of waiting a long time for the test to time out if the test is broken. That's preferable though.

hoodmane: What about adding a 200ms sleep between the attempts?

PR #4131, opened 2025-05-12
ignore llvm20 warnings

anonrig:
When using LLVM20, we receive certain errors. This follows the same pattern of EW changes.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -53,6 +53,9 @@ build --host_per_file_copt=external/com_google_protobuf@-Wno-deprecated-this-cap
 ## simdutf
 build --per_file_copt=external/simdutf@-Wno-unused-const-variable
 
+# TODO(cleanup): New LLVM 20 warning, fix and enable again
+build --copt=-Wno-nontrivial-memcall
+
 # Increasing the optimization level of some portions of V8 significantly speeds up Python tests in
 # GitHub Actions. This is an optional performance hack intended for CI, although it might also be
 # useful for local development.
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -283,6 +286,9 @@ build:linux --copt="-fdata-sections" --host_copt="-fdata-sections"
 # On Linux, use clang lld.
 build:linux --linkopt="-fuse-ld=lld"
 
+# LLVM 20 warning.
```
fhanau, line 289: Nit: rephrase the comment – this is not a LLVM 20 warning, I assume it's used to disable the warning about nontrivial-memcall not being known in older versions.

fhanau (approved):

.bazelrc
anonrig:
```suggestion
# Disable LLVM 20 warning about unused cli argument "-c".
```

.bazelrc
```diff
@@ -283,6 +286,9 @@ build:linux --copt="-fdata-sections" --host_copt="-fdata-sections"
 # On Linux, use clang lld.
 build:linux --linkopt="-fuse-ld=lld"
 
+# Disable LLVM 20 warning about unused cli argument "-c".
+build:linux --copt='-Wno-unused-command-line-argument' --host_copt='-Wno-unused-command-line-argument'
```
fhanau, line 290: Based on CI failing, I think you need -Wno-unknown-warning-option to support versions older than 20 still.

PR #4133, opened 2025-05-12
Fixes wrapping of Python to Python calls.

dom96:
Turns out we are incorrectly wrapping calls within RPC classes. This PR fixes that.

### Test Plan

```
```

### DIFF
--- a/src/pyodide/internal/workers.py
+++ b/src/pyodide/internal/workers.py
@@ -9,7 +9,6 @@ from collections.abc import Generator, Iterable, MutableMapping
 from contextlib import ExitStack, contextmanager
 from enum import StrEnum
 from http import HTTPMethod, HTTPStatus
-from inspect import isawaitable
 from types import LambdaType
 from typing import Any, TypedDict, Unpack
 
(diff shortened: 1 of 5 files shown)


src/pyodide/internal/workers.py
```diff
@@ -920,7 +921,23 @@ def _wrap_subclass(cls):
         setattr(cls, k, _wrap_attr(v))
 
 
-class DurableObject:
+class EntrypointBase:
+    async def _wrapper_func(self, relaxed, inst, prop, *args, **kwargs):
```
hoodmane, line 925: This should be a top level function, not a class method.

src/pyodide/internal/workers.py
```diff
@@ -920,37 +921,44 @@ def _wrap_subclass(cls):
...
+class WorkerEntrypoint(EntrypointBase):
     """
     Base class used to define a Worker Entrypoint.
     """
 
     def __init__(*_args, **_kwds):
         pass
 
-    def __init_subclass__(cls, **_kwargs):
-        _wrap_subclass(cls)
```
hoodmane, line 944: You should also delete the `_wrap_subclass()` function

src/pyodide/internal/workers.py
dom96: I put it in the class so that I can easily access it from python-entrypoint-helper. It seems preferable over having to call `getPyodide` and importing the function manually, do you disagree?

hoodmane: I think it's better to get it from top level so people can't break things by defining their own method with the same name.

src/pyodide/internal/workers.py
```diff
@@ -913,44 +914,44 @@ async def wrapper(*args, **kwargs):
     return wrapper
 
 
-def _wrap_subclass(cls):
-    for k, v in cls.__dict__.items():
-        if k.startswith("__"):
-            continue
-        setattr(cls, k, _wrap_attr(v))
```
hoodmane, line 920: We can get rid of `_wrap_attr()` too.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -255,18 +255,28 @@ function makeEntrypointProxyHandler(
...
-        if (typeof pyInstance[prop] === 'function') {
-          return await doPyCallHelper(isKnownHandler, pyInstance[prop], args);
-        } else {
+        if (typeof pyInstance[prop] !== 'function') {
           throw new TypeError(`Method ${prop} does not exist`);
         }
+
+        if ((isKnownHandler || isKnownDoHandler) && !isFetch) {
+          return await doPyCallHelper(true, pyInstance[prop], args);
+        }
```
hoodmane, line 272: Is this right? Don't we want to apply some wrapper logic to all RPC calls?

dom96: For `alert` at least we don't. I guess we could, but the object that's passed in just has two fields that can be easily read using Pyodide's FFI.

For the websocket methods we would need to implement wrappers for them, so keeping them unwrapped for now makes sense. But even for those I think the arguments are mostly usable via the FFI: https://developers.cloudflare.com/durable-objects/api/base/#websocketmessage. The WebSocket object is something we may want to wrap, but that will require further work.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -319,10 +333,22 @@ async function getIntrospectionMod(
...
   });
+
   return introspectionMod;
 }
 
+async function getIntrospectionMod(
+  pyodide: Pyodide
+): Promise<IntrospectionMod> {
+  if (introspectionModPromise === null) {
+    introspectionModPromise = loadIntrospectionMod(pyodide);
```
hoodmane, line 346: Since we call this like `getIntrospectionMod(await getPyodide())` in both spots, maybe we should do:
```js
introspectionModPromise = getPyodide().then(loadIntrospectionMod);
```
and make this take 0 arguments.

dom96: I'll do this in a separate PR, want to avoid another CI run for this

PR #4137, opened 2025-05-12
Setup codspeed for continuous benchmarking

art049:


build/deps/build_deps.jsonc
```diff
@@ -167,6 +167,12 @@
       "name": "com_google_benchmark",
       "owner": "google",
       "repo": "benchmark"
+    },
```
anonrig, line 170: Would you mind removing the previous "com_google_benchmark" named dependency?

art049: I removed it

src/workerd/tests/BUILD.bazel
```diff
@@ -84,3 +84,34 @@ wd_cc_benchmark(
...
+
+genrule(
+    name = "run_all_benchmarks",
+    outs = ["run_benchmarks.sh"],
+    cmd = """
+        echo '#!/bin/bash' > $@
+        echo 'set -e' >> $@
+        echo 'echo "Running all benchmarks..."' >> $@
+        for bench in %s; do
+            echo "bazel run //src/workerd/tests:$${bench#:}" >> $@
```
anonrig, line 112:
```suggestion
            echo "bazel run //src/workerd/tests:$${bench#:} --config=benchmark" >> $@
```

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,59 @@
...
+          sed -i '/apt-get install/d' llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh 18
+          sudo apt-get install -y --no-install-recommends clang-18 lld-18 libunwind-18 libc++abi1-18 libc++1-18 libc++-18-dev
+          echo "build:linux --action_env=CC=/usr/lib/llvm-18/bin/clang" >> .bazelrc
+          echo "build:linux --host_action_env=CC=/usr/lib/llvm-18/bin/clang" >> .bazelrc
+
+      - name: Build benchmarks
+        run: |
+          bazel build ${{ env.BAZEL_ARGS }} //src/workerd/tests:all_benchmarks
```
anonrig, line 51:
```suggestion
          bazel build ${{ env.BAZEL_ARGS }} //src/workerd/tests:all_benchmarks --config=benchmark
```

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,59 @@
...
+concurrency:
+  group: codspeed.yml-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  benchmarks:
+    name: Run benchmarks
+    runs-on: ubuntu-22.04
+    env:
+      BAZEL_ARGS: --config=ci --config=ci-linux --@codspeed_cpp//core:codspeed_mode=instrumentation --compilation_mode=dbg --copt=-O2 --remote_cache=https://bazel:${{ secrets.BAZEL_CACHE_KEY }}@bazel-remote-cache.devprod.cloudflare.dev
```
anonrig, line 25: Should we add --config=benchmark to here as well?

.github/workflows/codspeed.yml
anonrig: cc @fhanau Should this be here, or in the BAZEL_ARGS?

src/workerd/tests/BUILD.bazel
```diff
@@ -84,3 +84,34 @@ wd_cc_benchmark(
...
+    ":bench-global-scope",
+    ":bench-regex",
+    ":bench-fast-api",
+]
+
+filegroup(
+    name = "all_benchmarks",
+    srcs = BENCHMARKS,
+    visibility = ["//visibility:public"],
+)
```
anonrig, line 102:
```suggestion
filegroup(
    name = "all_benchmarks",
    srcs = glob(["bench-*"]),
    visibility = ["//visibility:public"],
)
```

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,59 @@
...
+      - main
+
+concurrency:
+  group: codspeed.yml-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  benchmarks:
+    name: Run benchmarks
+    runs-on: ubuntu-22.04
```
anonrig, line 23: @danlapid can we use large runners for this?

src/workerd/tests/BUILD.bazel
fhanau: I don't think this will work as written, globs refer to files not targets. You could build something based on the globs though if you cut off the extension (not sure if that would result in more elegant code).

src/workerd/tests/BUILD.bazel
fhanau: A target like this is not Bazel style at all. Why not just run the tests using `bazel test <args> //src/workerd/tests: all_benchmarks` instead of building a script?

.github/workflows/codspeed.yml
danlapid: Not sure how large runners work with github enterprise.
Perhaps @fhanau or @penalosa know.
In general I'd say why not use large runners for everything if they exist and work well.

.github/workflows/codspeed.yml
danlapid, line 1: I think this could be significantly simplified if we simple added a default false flag called `run_benchmarks` to `_bazel.yml` and enable that flag for one specific config in `test.yml`.
Would reduce duplication and also prevent double building the same target.

danlapid: We could just as well have the benchmark bazel label be configurable as another argument if that makes anything cleaner/easier.

danlapid:
Thanks for the contribution!

src/workerd/tests/BUILD.bazel
anonrig: I see. I'll resolve for now.

.github/workflows/codspeed.yml
fhanau: We should not be using dbg here or specify -O2 manually. Based on the configurations we currently have, `--config=benchmark` is a workable choice here, although this doesn't exactly give us full optimizations (but that's another topic, should be addressed in a wider bazelrc refactor)

.github/workflows/codspeed.yml
fhanau: Important question, but this is beyond the IC level and depends on our GitHub plan (is there a certain amount of paid runner usage included with our plan, or do we decide to pay for a given amount of usage, based on that making larger runners available needs to be decided on the org level). IMHO the most impactful change would be to use paid runners for Windows, but other jobs could benefit too.

src/workerd/tests/BUILD.bazel
anonrig: benchmarks are a cc_binary not a test. bazel test won't work.

PR #4138, opened 2025-05-12
Implement node:fs sync APIs

jasnell:
Not yet a full implementation but incrementally getting there. This adds impls for most of the fs sync APIs with a compat flag to enable. Wranglers polyfill overrides are still necessary until these are considered fully complete, tested, and optimized. It should be safe to land these as is, even if still incomplete and not fully tested as an experimental compat flag is required to enable them. I'm going to try to work on these over several PRs in order to keep the overall size of each PR down.


anonrig:
Why didn't you push this on top of https://github.com/cloudflare/workerd/pull/3796? I think we should keep all implementation in a single branch as discussed, before merging.

src/node/internal/internal_fs_sync.ts
```diff
@@ -0,0 +1,822 @@
+if (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any,@typescript-eslint/no-unsafe-member-access
+  !(globalThis as any).Cloudflare.compatibilityFlags['enable_node_file_system']
+) {
+  throw new Error('The experimental node:fs implementation is not enabled');
```
anonrig, line 5: Even though this is behind an experimental flag, we're now throwing a different error message, then we do in the current state. The correct way to do this is through module loader, and just act like "node:fs" doesn't exist (which is the prior behavior to this PR)

jasnell:
Will merge these changes into the other PR.

That said, there's no reason not to merge into main incrementally under the experimental compat flag so will still likely seek to do the implementation incrementally over a series of individual PRs.

PR #4141, opened 2025-05-13
IMAGES-1608: Allow base64-encoded input and output in Images Binding

ns476:


src/cloudflare/internal/images-api.ts
```diff
@@ -40,8 +40,16 @@ class TransformationResultImpl implements ImageTransformationResult {
...
+  public image(
+    options?: ImageTransformationOutputOptions
+  ): ReadableStream<Uint8Array> {
+    let stream = this.bindingsResponse.body || new ReadableStream();
+
+    if (options?.encoding === 'base64') {
+      stream = stream.pipeThrough(createBase64EncoderTransformStream());
+    }
+
+    return stream;
```
anonrig, line 52: Right now, we always create a new ReadableStream even though we override it in the next statement. This would avoid creating unnecessary readable streams.

```suggestion
    return options?.encoding === 'base64' ?
      stream.pipeThrough(createBase64EncoderTransformStream()) : (this.bindingsResponse.body || new ReadableStream());
```

src/cloudflare/internal/images-api.ts
```diff
@@ -203,10 +211,17 @@ class ImagesBindingImpl implements ImagesBinding {
...
+    stream: ReadableStream<Uint8Array>,
+    options?: ImageInputOptions
   ): Promise<ImageInfoResponse> {
     const body = new StreamableFormData();
-    body.append('image', stream, { type: 'file' });
+
+    let decodedStream = stream;
+    if (options?.encoding === 'base64') {
+      decodedStream = stream.pipeThrough(createBase64DecoderTransformStream());
+    }
```
anonrig, line 222: Can you use an inline if statement and make this a const?

src/cloudflare/internal/images-api.ts
```diff
@@ -235,8 +250,17 @@ class ImagesBindingImpl implements ImagesBinding {
...
-    return new ImageTransformerImpl(this.fetcher, stream);
+  public input(
+    stream: ReadableStream<Uint8Array>,
+    options?: ImageInputOptions
+  ): ImageTransformer {
+    let decodedStream = stream;
+
+    if (options?.encoding === 'base64') {
+      decodedStream = stream.pipeThrough(createBase64DecoderTransformStream());
+    }
```
anonrig, line 261: This can be a single line.

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
   return outputStream;
 }
 
+function concatUint8Arrays(a: Uint8Array, b: Uint8Array): Uint8Array {
```
anonrig, line 330: This function doesn't take care of different offsets. Also we can use resizable array buffers, as far as I know, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/resize

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+  const result = new Uint8Array(a.length + b.length);
+  result.set(a, 0);
+  result.set(b, a.length);
+  return result;
+}
+
+class Base64Error extends Error {
+  public constructor(cause: unknown) {
+    if (cause instanceof Error) {
```
anonrig, line 339: instanceof Error checks are incomplete:

```suggestion
    if (Error.isError(cause)) {
```

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+  if (maxEncodeChunkSize % 3 != 0) {
+    // Ensures that chunks don't require padding
+    throw new Error('maxChunkSize must be a multiple of 3');
+  }
+
+  let buffer: Uint8Array | null = null;
+  const asciiEncoder = new TextEncoder();
+
+  const toBase64: (buf: Uint8Array) => Uint8Array = (buf) => {
+    const binaryString = String.fromCharCode.apply(null, Array.from(buf));
```
anonrig, line 359: This will not work. Maximum argument length you can get in V8 is 65k. If buf has more than 65k this will fail.

Node.js way of doing this is:

Buffer.from(buf).toString('base64')

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+
+    const base64String = btoa(binaryString);
+    const base64Bytes = asciiEncoder.encode(base64String);
+
+    return base64Bytes;
```
anonrig, line 364:
```suggestion
    return asciiEncoder.encode(base64String);
```

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+
+function createBase64DecoderTransformStream(
+  maxChunkSize: number = 32 * 1024
+): TransformStream<Uint8Array, Uint8Array> {
+  if (maxChunkSize % 4 !== 0 || maxChunkSize <= 0) {
+    throw new Error('maxChunkSize must be a positive multiple of 4.');
+  }
+
+  let base64Buffer: Uint8Array | null = null;
+  const asciiDecoder = new TextDecoder('ascii');
```
anonrig, line 413: You can create this in the global scope. No need to create this everytime this method is executed.

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+
+  const decodeAndEnqueueSegment = (
+    base64SegmentBytes: Uint8Array,
+    controller: TransformStreamDefaultController<Uint8Array>
+  ): boolean => {
+    try {
+      const base64SegmentString = asciiDecoder.decode(base64SegmentBytes);
+
+      const binaryString = atob(base64SegmentString);
```
anonrig, line 422: atob is not recommended to be used in production environments. this will throw for utf8 values.

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+          }
+        }
+      }
+    },
+
+    flush(controller): void {
+      if (base64Buffer && base64Buffer.length > 0) {
+        if (!decodeAndEnqueueSegment(base64Buffer, controller)) {
+          // Error already handled within the helper
+        }
```
anonrig, line 482: Why do you need this?

src/cloudflare/internal/test/images/images-api-test.js
```diff
@@ -257,3 +275,79 @@ export const test_images_transform_consumed = {
...
+    for (const testName in inputs) {
+      const input = inputs[testName];
+
+      const result = await env.images
+        .input(inputStream(input))
+        .transform({ rotate: 90 })
+        .output({ format: 'image/avif' });
+
+      // Would be image/avif in real life, but mock always returns JSON
+      assert.equal(result.contentType(), 'application/json');
```
anonrig, line 340:
```suggestion
      assert.strictEqual(result.contentType(), 'application/json');
```

src/cloudflare/internal/images-api.ts
```diff
@@ -40,8 +40,16 @@ class TransformationResultImpl implements ImageTransformationResult {
     return contentType;
   }
 
-  public image(): ReadableStream<Uint8Array> {
-    return this.bindingsResponse.body || new ReadableStream();
+  public image(
+    options?: ImageTransformationOutputOptions
+  ): ReadableStream<Uint8Array> {
+    let stream = this.bindingsResponse.body || new ReadableStream();
```
jasnell, line 46: hmm... I know this isn't new but is creating the `new ReadableStream()` like this actually the correct thing to do? This stream looks like it would never be ended/closed

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+  }
+
+  let buffer: Uint8Array | null = null;
+  const asciiEncoder = new TextEncoder();
+
+  const toBase64: (buf: Uint8Array) => Uint8Array = (buf) => {
+    const binaryString = String.fromCharCode.apply(null, Array.from(buf));
+
+    const base64String = btoa(binaryString);
+    const base64Bytes = asciiEncoder.encode(base64String);
```
jasnell, line 362: Hmm... we should probably look at providing a `cloudflare-internal:...` API to do the base64 encoding more efficiently. If nodejs_compat is on then I we could just use `Buffer`s built-in base64 support. We could expose the same thing via a `cloudflare-internal` for times when `nodejs_compat` is not enabled.

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
   return outputStream;
 }
 
+function concatUint8Arrays(a: Uint8Array, b: Uint8Array): Uint8Array {
+  const result = new Uint8Array(a.length + b.length);
+  result.set(a, 0);
+  result.set(b, a.length);
+  return result;
```
jasnell, line 334: This is going to become quite expensive as the buffers get bigger. If nodejs_compat was enabled then we could use `Buffer.concat` to be far more efficient.

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+}
+
+class Base64Error extends Error {
```
jasnell, line 337: Do we really need to create an error subclass for this? Seems less than ideal.

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+        try {
+          controller.enqueue(toBase64(processChunk));
+        } catch (error) {
+          controller.error(new Base64Error(error));
+          buffer = null;
+          return;
+        }
+      }
+
+      buffer = offset < currentData.length ? currentData.slice(offset) : null;
```
jasnell, line 388: Something that most folks miss here is that `slice(...)` on a TypedArray actually creates a copy of the data for the slice. Is that what you're intending here or do you want a view (in which case you want `subarray(...)` not `slice(...)`.

src/cloudflare/internal/images-api.ts
```diff
@@ -303,6 +327,165 @@ function chainStreams<T>(streams: ReadableStream<T>[]): ReadableStream<T> {
...
+    },
+
+    flush(controller): void {
+      if (buffer && buffer.length > 0) {
```
jasnell, line 392:
```suggestion
      if (buffer?.length > 0) {
```

src/cloudflare/internal/images-api.ts
jasnell: While true, it's generally discouraged to create shared `TextDecoder` instances because they can contain state depending on what encoding is used and how they are utilized. Keeping this here makes sense I think.

src/cloudflare/internal/images-api.ts
jasnell: We probably either need to enable nodejs_compat for this case or expose similar APIs via a `cloudflare-internal` module.

src/cloudflare/internal/images-api.ts
ns476: `||` shortcircuits so we'll only create the empty stream if it is needed here (also changed to use `Blob` anyway due to comment further down)

src/cloudflare/internal/images-api.ts
ns476: I think it does work with offsets: https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABBBECGUCqMxQBwCCATkWgJ4DOAFGgFyLa6EnkA0iARvY-saWQEpuOXizKIA3gChEyBBSiIiAUwogANooC8iMMoDuDEc340AdOuVgA5lAAWiANScLV23YEBuGUtUaoZhTKUDTsAAxePipqmoHBVBzsaK429pGyKlAgREjR-t4AvlJSKGAKiGiIOjwm5GbARHAAtlQA2gCM7ABM7ADM7AAs7ACsALqRpeUcVUZMfHUNzW0AbOwA7Ox47ACc7O1h494l8nCWFnDWVKXoWMbzlKGcAhMnZ+oXV6gYNffUaOwcQIgDhoMRULrPLxAA

Holding off on the resizable bit as this would be a more sizeable change and it would be much better to use `Buffer` instead

src/cloudflare/internal/images-api.ts
ns476: Error.isError doesn't seem to be in the types for some reason, but made this change anyway

src/cloudflare/internal/images-api.ts
ns476: I guess as it is base64 we won't see any utf8 values if the base64 is valid, and it throws for invalid base64 anyway? Regardless using the Node APIs sounds way better if doable

src/cloudflare/internal/images-api.ts
ns476: Just keeps symmetry with the block above - removed

src/cloudflare/internal/images-api.ts
ns476: This branch should almost never be taken, but replaced it with `new Blob().stream()`

src/cloudflare/internal/images-api.ts
ns476: Removed

src/cloudflare/internal/images-api.ts
ns476: This doesn't type sadly (even though it does what we desire at runtime): https://www.typescriptlang.org/play/?#code/MYewdgzgLgBARgLhgbxgGwKZgOZQBZJgCuAtnBgE4wC+MAPjMWmjALyNHMDcAUDwGZEwwKAEtwMACYgMEMFAAqATwAOGABSIU6LLgIcylGvQ7MAlEjggQmAIZgUPGM5gUMUIhQdwA-ADpMHHwYAD4YAAZeaiA

src/cloudflare/internal/streaming-base64.ts
```diff
@@ -0,0 +1,194 @@
+import base64 from 'cloudflare-internal:base64';
+
+function base64Error(cause: unknown): Error {
+  // @ts-expect-error: Error.isError seems to be missing from types?
```
anonrig, line 5: This is going to be enabled after 5.9 version of Typescript, which is not yet released. Can you add a TODO(soon) here as well, so I can update this once we update TS?

src/cloudflare/internal/streaming-base64.ts
```diff
@@ -0,0 +1,194 @@
...
+  if (isError(cause)) {
+    const e = cause as Error;
+    return new Error(`base64 error: ${e.message}`, { cause });
+  } else {
+    return new Error('unknown base64 error');
+  }
+}
+
+function toBase64(input: Uint8Array): Uint8Array {
+  return new Uint8Array(base64.encodeArray(input));
```
anonrig, line 16: You can use the following since encodeArray returns an ArrayBuffer, which will be faster afaik:

```
new Uint8Array(buffer, byteOffset, length)
```

anonrig: This also applies to other places

src/cloudflare/internal/streaming-base64.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+function fromBase64(input: Uint8Array): Uint8Array {
+  return new Uint8Array(base64.decodeArray(input));
+}
+
+function combineArrays(a: Uint8Array, b: Uint8Array): Uint8Array {
```
anonrig, line 23: FYI: If we had nodejs_compat, we could have used Buffer.concat() from Node.js

src/cloudflare/internal/streaming-base64.ts
```diff
@@ -0,0 +1,194 @@
...
+      : new Uint8Array();
+  const leftover =
+    processableLength !== chunk.length
+      ? chunk.subarray(processableLength)
+      : null;
+
+  return { processable, leftover };
+}
+
+const PADDING_CHAR_CODE = '='.charCodeAt(0);
```
anonrig:
```suggestion
const PADDING_CHAR_CODE = 61;  // '='
```

src/cloudflare/internal/streaming-base64.ts
```diff
@@ -0,0 +1,194 @@
...
+  return chunk.length > 0 && chunk[chunk.length - 1] === PADDING_CHAR_CODE;
+}
+
+export function createBase64EncoderTransformStream(
+  maxEncodeChunkSize: number = 32 * 1024 + 1
+): TransformStream<Uint8Array, Uint8Array> {
+  let leftover: Uint8Array | null = null;
+
+  if (maxEncodeChunkSize % 3 != 0) {
+    // Try to minimise padding
```
anonrig:
```suggestion
    // Try to minimize padding
```

src/cloudflare/internal/streaming-forms.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return new Blob([
+      `content-disposition: form-data; name="${name}"${filenamePart}`,
+      CRLF,
+      CRLF,
+    ]).stream();
+  }
+
+  private multipartBody(stream: ReadableStream): ReadableStream {
```
anonrig: Can we avoid using `private` keyword, and use `#` which is in JS language?

src/cloudflare/internal/test/images/images-api-test.js
```diff
@@ -4,22 +4,62 @@
 // @ts-ignore
 import * as assert from 'node:assert';
 
-function inputStream(body) {
-  return new Blob([body]).stream();
+function inputStream(chunks) {
+  return new ReadableStream({
+    start(controller) {
+      const encoder = new TextEncoder();
```
anonrig: This can be moved to outside of the function

anonrig (approved):

src/cloudflare/internal/streaming-base64.ts
ns476: I think these are the same things - `new Uint8Array(arrayBuffer)` will point at the same memory:
```
> ab = new ArrayBuffer(1)
ArrayBuffer { [Uint8Contents]: <00>, byteLength: 1 }
> u8a = new Uint8Array(ab)
Uint8Array(1) [ 0 ]
> u8a.set([1], 0)
undefined
> u8a
Uint8Array(1) [ 1 ]
> ab
ArrayBuffer { [Uint8Contents]: <01>, byteLength: 1 }
```

src/cloudflare/internal/streaming-forms.ts
ns476: Might be nice to check this with the linter, looks like it can be configured: https://github.com/typescript-eslint/typescript-eslint/issues/1391#issuecomment-1124154589

penalosa (changes requested):
The actual changes look fine, but the type snapshot seems to have gone a bit wrong. This PR is adding over 100,000 lines of code

danlapid:
> The actual changes look fine, but the type snapshot seems to have gone a bit wrong. This PR is adding over 100,000 lines of code

Oh yeah
I suggest deleting all of the type snapshot changes, rebasing on top of workerd main and recreating the snapshot

penalosa (approved):

PR #4142, opened 2025-05-13
enable eval() and new Function on startup

anonrig:
Enables the usage of `eval()` and `new Function()` on startup phase of the main module. Potentially enabling libraries such as Fastify to be used within Cloudflare Workers.

Starting June 1st, 2025, the following code works:

```js
strictEqual(eval('1+1'), 2);
// or even this
strictEqual((await import('module-that-has-eval-in-it)).default, 2);

export default {
    return new Response('Hello World!');
  },
};
```

### DIFF
--- a/src/workerd/api/tests/worker-test.js
+++ b/src/workerd/api/tests/worker-test.js
@@ -1,0 +1,27 @@
+import { throws, rejects, strictEqual } from 'node:assert';
+
+// We allow eval and new Function() during startup (global scope evaluation).
+strictEqual(eval('1+1'), 2); // should work
+const StartupFunction = new Function('return 1+1');
+strictEqual(StartupFunction(), 2); // should work
+await Promise.resolve();
+strictEqual(eval('1+1'), 2); // should still work
+const StartupFunction2 = new Function('return 2+2');
+strictEqual(StartupFunction2(), 4); // should still work
+strictEqual((await import('module-does-eval')).default, 2); // should work
+
+export const testStartupEval = {
+  async test() {
+    throws(() => eval('console.log("This should not work")'), {
+      message: /Code generation from strings disallowed for this context/,
+    });
+
+    throws(() => new Function('console.log("This should not work")'), {
+      message: /Code generation from strings disallowed for this context/,
+    });
+
+    await rejects(() => import('another-module-does-eval'), {
+      message: /Code generation from strings disallowed for this context/,
+    });
+  },
+};
--- a/src/workerd/api/tests/worker-test.wd-test
+++ b/src/workerd/api/tests/worker-test.wd-test
@@ -1,0 +1,20 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "worker-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "worker-test.js"),
+          (name = "module-does-eval", esModule = "export default eval('1+1')"),
+          (name = "another-module-does-eval", esModule = "export default eval('2+2')"),
+        ],
+        compatibilityDate = "2025-05-13",
+        compatibilityFlags = [
+          "nodejs_compat",
+          "allow_eval_during_startup",
+        ]
+      )
+    ),
+  ],
+);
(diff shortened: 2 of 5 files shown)


hoodmane:
This looks helpful.

src/workerd/io/worker.c++
```diff
@@ -1652,6 +1652,11 @@ Worker::Worker(kj::Own<const Script> scriptParam,
             // evaluation is complete.
             TmpDirStoreScope tmpDirStoreScope;
 
+            // We allow eval and new Function() on bootstrap phase, and later disable
+            // them to prevent malicious code from executing.
```
kentonv, line 1656: Nitpick: I don't think we've ever used the term "bootstrap phase" for this before. We just call it "startup" or "global scope execution".

Also the reason we disallow eval is a little more subtle than "malicious code".

```suggestion
            // We allow eval and new Function() during startup, becaues startup time is entirely
            // deterministic, so we can easily reproduce the input to eval() by just running the
            // worker again. We do not allow eval() at runtime because we need to have a record of
            // all code that executes in production for forensic purposes, and at runtime the input
            // to eval() could have come from a remote source on which we don't have a record.
```

src/workerd/api/tests/worker-test.js
```diff
@@ -0,0 +1,19 @@
+import { throws, rejects } from 'node:assert';
+
+// We allow eval and new Function() on bootstrap phase.
```
kentonv, line 3:
```suggestion
// We allow eval and new Function() during startup (global scope evaluation).
```

src/workerd/api/tests/worker-test.js
```diff
@@ -0,0 +1,19 @@
...
+eval('1+1'); // should work
```
kentonv, line 4: We should verify that the eval'd code actually ran. At present this test would pass if eval were implemented as a no-op.

src/workerd/io/worker.c++
```diff
@@ -1652,6 +1652,11 @@ Worker::Worker(kj::Own<const Script> scriptParam,
...
+            js.setAllowEval(true);
```
danlapid, line 1657: compat flag pls

src/workerd/io/worker.c++
anonrig: Thanks you're right!

src/workerd/api/tests/worker-test-module-does-eval.js
```diff
@@ -0,0 +1 @@
+eval(`console.log('I am a module that does eval')`);
```
jasnell, line 1: As Kenton suggests, we should add asserts to ensure the eval ran. 

Also, would be good to include some `new Function(...)` calls in the tests as well, just to be thorough

src/workerd/api/tests/worker-test.wd-test
```diff
@@ -0,0 +1,19 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const unitTests :Workerd.Config = (
+  services = [
+    ( name = "worker-test",
+      worker = (
+        modules = [
+          (name = "worker", esModule = embed "worker-test.js"),
+          (name = "module-does-eval", esModule = embed "worker-test-module-does-eval.js"),
+          (name = "another-module-does-eval", esModule = embed "worker-test-another-module-does-eval.js"),
```
jasnell, line 10: This can be made simpler by just inlining these modules rather than using embed.. something like:

```suggestion
          (name = "module-does-eval", esModule = "export default eval('1+1')"),
          (name = "another-module-does-eval", esModule = "export default eval('2+2')"),
```

Then removing the two additional fixture files.

src/workerd/io/worker.c++
```diff
@@ -1652,6 +1652,16 @@ Worker::Worker(kj::Own<const Script> scriptParam,
...
             TmpDirStoreScope tmpDirStoreScope;
 
+            // We allow eval and new Function() during startup, becaues startup time is entirely
+            // deterministic, so we can easily reproduce the input to eval() by just running the
+            // worker again. We do not allow eval() at runtime because we need to have a record of
+            // all code that executes in production for forensic purposes, and at runtime the input
+            // to eval() could have come from a remote source on which we don't have a record.
+            if (FeatureFlags::get(js).getAllowEvalDuringStartup()) {
+              js.setAllowEval(true);
+              KJ_DEFER(js.setAllowEval(false));
```
danlapid, line 1662: Is that actually working? I would expect that to disable immediately because the scope ends immediately.

anonrig: So, this is why my feature flags check doesn't work...

kentonv (approved):

PR #4143, opened 2025-05-13
Add compile-time serializability checks

npaun:
This PR is my idea on making serialization easier and safer to use. 

The main idea is that you should call `serializer.write` for most cases, and it'll do the right thing:
- `serializer.write(x);` for bools, enums, strings will always use the standard representation we've been using so far.
- `serializer.write(js, typeHandler, x);` for structs.
- I didn't add an overload for integer types because I'm worried about places where we just do int/uint without thinking about size, so that remains as `serializer.writeRawUint32(x)` or serializer.writeRawUint64(x)`

For JSValue, the existing function `serializer.write(js, x)` is now written as  `serializer.writeDynamic(js, x);` to remind callers that we are not checking the contained type at compile time and don't know if it's serializable.

The compile-time serialization check is introduced by adding `SerializedTypeHandler<T>` and a new `writeForSerialize` method in each type wrapper. (The PR only adds it for some types; I will finish the rest once we've reviewed the concept).

There are two "escape hatches" that can be used to wrap a struct field:
- `FakeSerializable<T>` which just ignores the check for this field. Ideally we can get rid of it soon - we just need more validation about what is stored in the CF blob of a request or response.
- `OmitForSerialized<T>` is like `Optional<T>` but always encodes as `undefined` when serializing. This is useful for any struct fields that don't need to be sent.

### DIFF
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -532,7 +532,7 @@ void Headers::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
   // is a common header ID, or the value zero to indicate an uncommon header, which is then
   // followed by a length-delimited name.
 
-  serializer.writeRawUint32(static_cast<uint>(guard));
+  serializer.write(guard);
 
   // Write the count of headers.
   uint count = 0;
@@ -551,9 +551,9 @@ void Headers::serialize(jsg::Lock& js, jsg::Serializer& serializer) {
         serializer.writeRawUint32(c);
       } else {
         serializer.writeRawUint32(0);
-        serializer.writeLengthDelimited(header.name);
+        serializer.write(header.name);
       }
-      serializer.writeLengthDelimited(value);
+      serializer.write(value);
     }
   }
 }
@@ -1285,7 +1285,7 @@ void RequestInitializerDict::validate(jsg::Lock& js) {
 void Request::serialize(jsg::Lock& js,
     jsg::Serializer& serializer,
     const jsg::SerializeTypeHandler<RequestInitializerDict>& initDictHandler) {
-  serializer.writeLengthDelimited(url);
+  serializer.write(url);
 
   // Our strategy is to construct an initializer dict object and serialize that as a JS object.
   // This makes the deserialization end really simple (just call the constructor), and it also
--- a/src/workerd/api/http.c++
+++ b/src/workerd/api/http.c++
@@ -1284,60 +1284,59 @@ void RequestInitializerDict::validate(jsg::Lock& js) {
 
 void Request::serialize(jsg::Lock& js,
     jsg::Serializer& serializer,
-    const jsg::TypeHandler<RequestInitializerDict>& initDictHandler) {
+    const jsg::SerializeTypeHandler<RequestInitializerDict>& initDictHandler) {
   serializer.writeLengthDelimited(url);
 
   // Our strategy is to construct an initializer dict object and serialize that as a JS object.
   // This makes the deserialization end really simple (just call the constructor), and it also
   // gives us extensibility: we can add new fields without having to bump the serialization tag.
-  serializer.write(js,
-      jsg::JsValue(initDictHandler.wrap(js,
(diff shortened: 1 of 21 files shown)


kentonv:
Sorry but I don't think this is the right direction.

Our serializer is primarily designed to serialize JavaScript values, and those values are primarily dynamic. The serializer has a few helper functions for writing integers and raw bytes, but it is not remotely intended to be used for complex data structures. Anything complex should either:
a) Be converted to a JavaScript value, and serialized as such.
b) Be converted to Cap'n Proto, and serialized as an "external".

We do not want to build a third serialization format here, and so we do not need or want to create a framework to define serializers for arbitrary C++ types.

For illustration, notice that the serializer interface defines two different methods that accept `kj::ArrayPtr<const byte>`: `writeLengthDelimited()` and `writeRawBytes()`. These do different things. Which is the correct one for `write()` to defer to? You've chosen `writeLengthDelimited()`, and perhaps there's reasonable arguments for that. However, the more basic reality here is that this is an interface for writing specific bytes sequences; it is not a framework for writing arbitrary values based on type. We _should_ make callers be clear about exactly what format they are trying to write. It should _not_ be inferred from the data type.

Is there an actual problem you ran into that motivated this? Or is it more an exploration of serialization frameworks? I definitely think serialization frameworks are a cool and fun thing to design (I've made a few), but I think it's not what we want in this specific case.

src/workerd/api/http.c++
```diff
@@ -1740,33 +1739,30 @@ jsg::Optional<jsg::JsObject> Response::getCf(jsg::Lock& js) {
...
+        .headers = headers.addRef(),
+        .cf = cf.getRef(js),
+
+        // If a WebSocket is present, we'll try to serialize it. As of this writing, WebSocket
+        // is not serializable, but we could add support for sending it over RPC in the future.
+        //
+        // Note we have to double-Maybe this, so that if no signal is present, the property is absent
+        // instead of `null`.
+        .webSocket = webSocket.map(
+            [](jsg::Ref<WebSocket>& s) -> kj::Maybe<jsg::Ref<WebSocket>> { return s.addRef(); }),
```
npaun, line 1761: @kentonv The motivating example for this proposal is the serialization of Request and Response, which works by creating an initializer dict, wrapping it into a JSValue, then sending it off to be serialized.

In a number of cases, we have fields like `webSocket` with notes like, you can't actually serialize this, but maybe you can in the future. 

What happens if we end up setting websocket and serializing? We won't know until runtime, until we go down this code path. We also probably won't even get a log for it because it manifests as a JS DataCloneError. That's what happened with abortsignal.

So I wanted to "just" make this a compile-time error, without modifying any of our existing serialization implementations.

However, making this happen involved some rather invasive, but I believe fairly trivial changes to the type wrapper. 

Nonetheless, you may well still conclude it's not worth it.

kentonv: I think this doesn't solve the problem.

I knew full well when implementing serialization for `Request` and `Response` that fields like `signal` and `webSocket`, if present, would break the serialization. You've noted my comments saying as much. This was working as intended: If non-serializable fields were present, then Request would fail to serialize, but if they weren't present, then it would work. This is consistent with how serialization works in general: a plain JavaScript object is serializable, unless it contains non-serializable fields.

The problem came when we started making incoming requests have a `signal` by default. This broke code which assumed incoming requests were serializable. This is my fault, in two ways:
1. I should have realized this problem when reviewing the PR.
2. Back when I made `Request` serializable in the first place, I should have written a test case that serializes an incoming request. I unfortunately only wrote a test case that serializes a `Request` constructed from JavaScript. Thus the tests missed the problem later.

But any compile-time checking would not have caught the problem. If a compile-time check had flagged that `Request` had possibly non-serializable fields when I first implemented `Request` serialization, I would have disabled the check, because _I knew that_, and it was working as intended. Your later change to make incoming requests have a `signal` did not change anything about `Request`'s type signature, so would not have been caught by any compile-time checking.

PR #4144, opened 2025-05-13
Fix comments in container.capnp

vicb:


src/workerd/io/container.capnp
```diff
@@ -26,7 +26,7 @@ interface Container @0x9aaceefc06523bca {
     enableInternet @1 :Bool = false;
     # Set true to enable the container to talk directly to the public internet. Otherwise, the
     # public internet will not be accessible -- but it's still possible to intercept connection
-    # attempts and handle them in the DO, using the "listen" methods below.
+    # attempts and handle them in the DO, using the "TcpHandler" interface below.
```
kentonv, line 29:
```suggestion
    # attempts and handle them in the DO, using the `listenTcp()` method below.
```

The interface doesn't do anything on its own. You have to use the method.

Other "listen" methods have been proposed, which is why this was written as "listen methods" instead of naming the specific method that exists today, but sure we can make it more specific.

jasnell (approved):

PR #4145, opened 2025-05-13
Reproduction of issue with `connect()` and `globalOutbound`

penalosa:
Relating to https://github.com/cloudflare/workers-sdk/issues/9238, it seems that using `connect()` in a Worker with a `globalOutbound` configured is not supported, because the `connect()` tries to go via the global outbound, failing because Workers don't support TCP ingress. I'd expect the `connect()` API to bypass `globalOutbound` and go straight to the `internet` service, so that it's possible to use `connect()` in a Worker with `globalOutbound` configured.

To reproduce, run `pnpx workerd samples/connect-with-global-outbound/config.capnp` and `curl http://localhost:8080/connect`. `workerd` will throw an error like the following:

```
workerd/server/server.c++:3872: error: Uncaught exception: workerd/jsg/_virtual_includes/iterator/workerd/jsg/value.h:1479: failed: remote.jsg.TypeError: Incoming CONNECT on a worker not supported
```

Removing the `globalOutbound` setting allows the connection to be established as expected.

### DIFF
--- a/samples/connect-with-global-outbound/config.capnp
+++ b/samples/connect-with-global-outbound/config.capnp
@@ -1,0 +1,21 @@
+using Workerd = import "/workerd/workerd.capnp";
+
+const helloWorldExample :Workerd.Config = (
+  services = [ (name = "main", worker = .helloWorld), (name = "outbound", worker = .outbound) ],
+  sockets = [ ( name = "http", address = "*:8080", http = (), service = "main" ) ]
+);
+
+const helloWorld :Workerd.Worker = (
+  modules = [
+    (name = "worker", esModule = embed "worker.js")
+  ],
+  compatibilityDate = "2023-02-28",
+  globalOutbound = "outbound"
+);
+
+const outbound :Workerd.Worker = (
+  modules = [
+    (name = "worker", esModule = embed "outbound.js")
+  ],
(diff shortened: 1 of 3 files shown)


danlapid:
@kentonv this seems to be because when the outbound is a worker the `connect` api is using the WorkerEntrypoint::connect method.
IIRC in prod for the WfP usecase when we set an outbound worker then `connect` would bypass that and just use the regular outbounds right?
Seems somewhat similar and should likely behave similarly.
Maybe bypassing the outbound worker and going to `internet` if the set outbound is a ServiceDesignator of type worker ?

kentonv:
This is by design.

The main purpose of setting `globalOutbound` is to have another worker intercept all outbound requests to the internet from a worker, possibly to implement controls on what the worker is allowed to access.

Bypassing `globalOutbound` for `connect()` would entirely defeat this purpose and make it impossible to sandbox Workers using this mechanism.

> IIRC in prod for the WfP usecase when we set an outbound worker then connect would bypass that and just use the regular outbounds right?

I certainly hope not! This would be a serious security flaw.

The ideal solution here is to define an API for Workers to receive and proxy raw sockets. Then a worker which has been set as `globalOutbound` can proxy `connect()` as desired. This is a big project, though.

PR #4148, opened 2025-05-14
[o11y] Improve span tracing support, using CompletedSpan internally for now

fhanau:
1st commit is a refactor – should be slightly less efficient (one more branch per loop iteration) but that is negligible.

2nd commit should resolve some of the span timestamp/parent span issues we've seen in testing. No downstream changes are required. As noted in the TODOs, this is needed since we don't have instrumentation at span creation time and span context propagation is incomplete – synthesizing the span events at the end is much easier for now than to build those things and will unblock evaluation on the edge.

### DIFF
--- a/src/workerd/io/trace-stream.c++
+++ b/src/workerd/io/trace-stream.c++
@@ -568,6 +568,10 @@ jsg::JsValue ToJs(jsg::Lock& js, const tracing::TailEvent& event, StringCache& c
     KJ_CASE_ONEOF(spanClose, tracing::SpanClose) {
       obj.set(js, EVENT_STR, ToJs(js, spanClose, cache));
     }
+    KJ_CASE_ONEOF(span, CompleteSpan) {
+      // Spans are being decomposed into SpanOpen|SpanClose|Attributes for now.
+      KJ_UNREACHABLE;
+    }
     KJ_CASE_ONEOF(de, tracing::DiagnosticChannelEvent) {
       obj.set(js, EVENT_STR, ToJs(js, de, cache));
     }
@@ -609,6 +613,10 @@ kj::Maybe<kj::StringPtr> getHandlerName(const tracing::TailEvent& event) {
     }
     KJ_CASE_ONEOF(_, tracing::SpanClose) {
       return SPANCLOSE_STR;
+    }
+    KJ_CASE_ONEOF(_, CompleteSpan) {
+      // Spans are being decomposed into SpanOpen|SpanClose|Attributes for now.
+      KJ_UNREACHABLE;
     }
     KJ_CASE_ONEOF(_, tracing::DiagnosticChannelEvent) {
       return DIAGNOSTICCHANNEL_STR;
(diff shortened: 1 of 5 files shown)


danlapid:
I understand we are quite intentionally making a hack just to make it work but I'm not sure we're doing the right thing or the healthy thing here.
Can I please get some understanding as to how difficult it would be and how long it would take to actually make spans would as they should?

src/workerd/io/trace-stream.c++
```diff
@@ -858,54 +866,64 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
...
+            }
           }
         }
+      };
+
+      KJ_IF_SOME(span, event.event.tryGet<CompleteSpan>()) {
+        // Synthesize sub-events
+        auto open = SpanOpen(span.parentSpanId, kj::str(span.operationName));
+        auto close = SpanClose();
+        kj::Vector<workerd::tracing::Attribute> attr;
```
anonrig, line 909: Can you reserve the size?

src/workerd/io/worker-interface.capnp
```diff
@@ -290,12 +290,13 @@ struct Trace @0x8e8d911203762d34 {
       hibernate @5 :Hibernate;
       spanOpen @6 :SpanOpen;
       spanClose @7 :SpanClose;
-      attribute @8 :List(Attribute);
```
anonrig, line 293: Is this a breaking change?

src/workerd/io/trace-stream.c++
```diff
@@ -858,54 +866,64 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
...
+          // will not wait for each promise to resolve before calling the next iteration.
+          // But we will wait for all promises to settle before returning the resolved
+          // kj promise.
+          auto fn = h.As<v8::Function>();
+          returnValues.push_back(jsg::check(fn->Call(js.v8Context(), h, 1, &eventObj)));
+        } else {
+          // If the handler is an object, then we need to know what kind of events
+          // we have and look for a specific handler function for each.
+          KJ_ASSERT(h->IsObject());
+          jsg::JsObject obj = jsg::JsObject(h.As<v8::Object>());
```
anonrig, line 893: You can move this line to one line below.

src/workerd/io/trace-stream.c++
```diff
@@ -858,54 +866,64 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
...
-            v8::Local<v8::Value> eventObj = ToJs(js, event, stringCache);
-            returnValues.push_back(jsg::check(fn->Call(js.v8Context(), h, 1, &eventObj)));
+        if (h->IsFunction()) {
+          // If the handler is a function, then we'll just pass all of the events to that
+          // function. If the function returns a promise and there are multiple events we
+          // will not wait for each promise to resolve before calling the next iteration.
+          // But we will wait for all promises to settle before returning the resolved
+          // kj promise.
+          auto fn = h.As<v8::Function>();
+          returnValues.push_back(jsg::check(fn->Call(js.v8Context(), h, 1, &eventObj)));
```
anonrig, line 888: I think you should check for the return value here.

fhanau: I'm calling `jsg::check` as before, is that not checking the return value?

fhanau:
@jmorrell-cloudflare this is the PR I mentioned

jmorrell-cloudflare:
For context the timestamp and missing parent span are the two biggest blockers to being able to provide an initial useful tracing experience for internal testers. Getting it in front of internal users is going to be the best way to surface unknown issues and figure out what instrumentation we should prioritize. I'm hopeful that this would allow us to focus on instrumentation and give @fhanau time and space to focus on nailing the correct long-term implementation.

src/workerd/io/worker-interface.capnp
```diff
@@ -290,12 +290,13 @@ struct Trace @0x8e8d911203762d34 {
...
-      return @9 :Return;
-      diagnosticChannelEvent @10 :DiagnosticChannelEvent;
-      exception @11 :Exception;
-      log @12 :Log;
-      link @13 :Link;
+      completedSpan @8 :UserSpanData;
```
danlapid, line 293: Because this is planned to be deleted, move this to the end of the list and add a big comment saying this will be deleted please.
Also explain why we are adding it now and when we can delete it.

fhanau: 👍 updated the capnp definition

danlapid (approved):

PR #4153, opened 2025-05-16
Add WPT streams tests

npaun:
@fhanau recently reminded me that we have redundant copies of Web Platform Tests inside our internal repo. 
This PR enables WPT's streams module under our new test framework. 
This is a first pass over these tests, so I've just made notes where anything wasn't working and have not yet deeply investigated the causes. Of note, one of these tests manages to segfault workerd; I'll prioritize investigating and fixing that one.


src/wpt/harness/harness.ts
```diff
@@ -737,6 +744,27 @@ globalThis.test = (func, name, properties): void => {
...
+ */
+globalThis.step_timeout = (
+  func: UnknownFunc,
+  timeout: number,
+  ...rest: unknown[]
+): ReturnType<typeof setTimeout> => {
+  // eslint-disable-next-line @typescript-eslint/no-this-alias -- Copied as-is from WPT code
+  const outer_this = this;
+
+  return setTimeout(function () {
```
anonrig, line 763: Why don't we just make this an arrow function and get rid of outer_this value?

anonrig (approved):

src/wpt/harness/harness.ts
npaun: I was just doing copy-pasta. There's a few other cases in WPT's harness like it; I can get rid of them all.

anonrig: Sounds good to me

PR #4154, opened 2025-05-16
Pulled two fixes from Ketan's sandboxing branch.

erikcorry:


anonrig (approved):

src/workerd/jsg/setup.c++
```diff
@@ -345,23 +347,22 @@ static v8::Isolate* newIsolate(v8::Isolate::CreateParams&& params, v8::CppHeap*
...
     }
-#if V8_COMPRESS_POINTERS_IN_MULTIPLE_CAGES
-    // Create new isolate group so that isolate is in its own group and not in a shared group. That
-    // way the isolates don't all use the same pointer cage.
-    v8::IsolateGroup group = v8::IsolateGroup::Create();
-    return v8::Isolate::New(group, params);
-#else
-    return v8::Isolate::New(params);
-#endif
+    KJ_IF_SOME(group, maybeGroup) {
```
fhanau, line 350: When would we pass a group vs. no group? Adding a comment would be helpful

src/workerd/jsg/buffersource.h
```diff
@@ -76,7 +76,7 @@ using BufferSourceViewConstructor = v8::Local<v8::Value> (*)(Lock&, BackingStore
 class BackingStore {
  public:
   template <BufferSourceType T = v8::Uint8Array>
-  static BackingStore from(kj::Array<kj::byte> data) {
+  static BackingStore from(Lock& js, kj::Array<kj::byte> data) {
```
fhanau, line 79: BackingStore::from() needing a lock looks unrelated to the IsolateGroup changes, can you add the rationale in a comment or a commit description?

fhanau (approved):

src/workerd/jsg/buffersource.h
erikcorry: I added a comment to explain.  This is because the ArrayBuffer backing needs to be allocated in the correct
sandbox, so we need to make the isolate explicit here.

PR #4156, opened 2025-05-16
Only compare types in the experimental/latest subdirectories

danlapid:
There shouldn't be much value in comparing the type snapshots of every single very-old date. 
It is extremely unlikely that any change to these files will not be in the experimental directory. 
These files create file bloat on PRs and are never really reviewed. 
This commit should clean it up without any effect on our ability to review type changes.


penalosa (approved):

anonrig (approved):

kentonv:
I find it useful to be able to distinguish between experimental and non-experimental API changes...

fhanau:
> I find it useful to be able to distinguish between experimental and non-experimental API changes...

I have to agree, at least maintaining the 2023-07-01 snapshot seems worthwhile to me.

kentonv (commented):
I think I disagree with this change. At least seeing the most-recent compat date informs us if the PR is making non-experimental type changes. In fact I think we should be flagging such PRs for more review.

I suppose it's not that useful to cover every historical compat date, though.

danlapid:
Sure, I'll change it so we have "latest" and "experimental"

penalosa:
> I find it useful to be able to distinguish between experimental and non-experimental API changes...

Are there often experimental-only type changes? The "experimental" entrypoint here is just "turn on all flags", so it will cover the latest non-experimental changes as well

PR #4160, opened 2025-05-17
Add jsg::Lock::current() to replace v8::Isolate::GetCurrent()

jasnell:
Direct use of `v8::Isolate::GetCurrent()` is something that we've tried to avoid but does creep in every so often. Let's provide a more useful utility to avoid it a bit better.

### DIFF
--- a/src/workerd/jsg/jsg.h
+++ b/src/workerd/jsg/jsg.h
@@ -2413,6 +2413,10 @@ class Lock {
     return *reinterpret_cast<Lock*>(v8Isolate->GetData(SET_DATA_LOCK));
   }
 
+  static Lock& current() {
+    return from(v8::Isolate::GetCurrent());
+  }
+
   // RAII construct that reports amount of external memory to be manually attributed to
   // the isolate. When the returned ExtrernalMemoryAdjuster is dropped, the amount will
   // be subtracted from the isolate's external memory accounting. If the adjuster is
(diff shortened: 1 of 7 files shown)


anonrig (commented):
I think this PR should include a custom clang-tidy rule, or it is highly likely that someone can call GetCurrent() directly.

jasnell:
A custom clang-tidy rule could be added separately. Not convinced it's a reason to block on this.

anonrig:
> A custom clang-tidy rule could be added separately. Not convinced it's a reason to block on this.

If it can be circumvented quite easily, then it will eventually become "yet another way of doing X"

jasnell:
Unless I've missed it somewhere, I don't believe we have any clang-tidy rules for workerd. We have them in the internal repo but as far as I recall (and can find in workerd) we don't build any custom clang-tidy rules in this repo. A separate PR can add a rule to the inner repo or can add clang tidy here.

anonrig (approved):

anonrig:
Can you add a todo for clang-tidy?

src/workerd/jsg/jsg.h
```diff
@@ -2413,6 +2413,10 @@ class Lock {
     return *reinterpret_cast<Lock*>(v8Isolate->GetData(SET_DATA_LOCK));
   }
 
+  static Lock& current() {
```
jasnell, line 2416:
```suggestion
  // TODO(someday): A clang-tidy rule to enforce use of Lock::current over
  // v8::Isolate::GetCurrent would be helpful.
  static Lock& current() {
```

PR #4161, opened 2025-05-17
Refactor how RPCs are delivered

kentonv:
This introduces `IoContext::makeReentryCallback()`, which takes a function intended to run under the isolate lock, and returns a function which can be invoked _outside_ the lock, and which will do all the necessary things to enter the isolate. This has proven more complicated than it looks. See the doc comment for details.

I created this for a different use case, but in this PR I went ahead and used it to clean up how JsRpcTarget works, as I was fairly unhappy with the existing code.

This fixes at least one bug with RPC: If you made an RPC call during blockConcurrentlyWhile, and passed it a callback, calls to the callback would be blocked until the end of blockConcurrencyWhile. This goes against the usual semantics of blockConcurrencyWhile, which says that interactions initiated from _inside_ the block are allowed to complete inside the block.

This PR also fixes the warning `Adding task to IoContext with no current IncomingRequest`, but sort of by accident, and there are deeper underlying problems there which aren't actually fixed. I might work on that in a separate PR.

### DIFF
--- a/src/workerd/api/worker-rpc.c++
+++ b/src/workerd/api/worker-rpc.c++
@@ -967,8 +967,7 @@ MakeCallPipeline::Result serializeJsValueWithPipeline(jsg::Lock& js,
 class JsRpcTargetBase: public rpc::JsRpcTarget::Server {
  public:
   JsRpcTargetBase(IoContext& ctx)
-      : weakIoContext(ctx.getWeakRef()),
-        enterIsolateAndCall(ctx.makeReentryCallback<IoContext::TOP_UP>(
+      : enterIsolateAndCall(ctx.makeReentryCallback<IoContext::TOP_UP>(
             [this, &ctx](Worker::Lock& lock,
                 CallContext callContext,
                 kj::Own<capnp::CallContextHook> ownCallContext,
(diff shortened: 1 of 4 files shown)


src/workerd/api/worker-rpc.c++
```diff
@@ -966,12 +966,29 @@ MakeCallPipeline::Result serializeJsValueWithPipeline(jsg::Lock& js,
 // session.
 class JsRpcTargetBase: public rpc::JsRpcTarget::Server {
  public:
-  JsRpcTargetBase(IoContext& ctx)
+  struct MayOutliveIncomingRequest {};
```
ObsidianMinor, line 969: This feels really weird, but I guess there's not really a better way to do it because it's for constructors.

src/workerd/io/io-context.h
```diff
@@ -513,6 +513,54 @@ class IoContext final: public kj::Refcounted, private kj::TaskSet::ErrorHandler
   template <typename T>
   kj::_::ReducePromises<RemoveIoOwn<T>> awaitJs(jsg::Lock& js, jsg::Promise<T> promise);
 
+  // Make a kj::Function which, when called, re-enters this IoContext to run some code.
+  //
+  // `func` is a function with a signature similar to:
```
ObsidianMinor, line 520: Ahh if only we had concepts...

ObsidianMinor (approved):

anonrig (approved):

PR #4164, opened 2025-05-17
Update withinHandleScope to support JsValue and friends

jasnell:
Resolve outstanding todo

### DIFF
--- a/src/workerd/api/actor-state.c++
+++ b/src/workerd/api/actor-state.c++
@@ -122,7 +122,8 @@ ActorObserver& currentActorMetrics() {
 
 jsg::JsRef<jsg::JsValue> listResultsToMap(
     jsg::Lock& js, ActorCacheOps::GetResultList value, bool completelyCached) {
-  return js.withinHandleScope([&] {
+  return js
+      .withinHandleScope([&] {
     auto map = js.map();
     size_t cachedReadBytes = 0;
     size_t uncachedReadBytes = 0;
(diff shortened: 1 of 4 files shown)


anonrig:
I think we should revisit handle scope creation and this function in general. @erikcorry and I take a look at this method, and in most places we don't need an extra handle scope at all (except for places where there is a large loop/iteration, and the original handleScope is not returned asap)

src/workerd/api/actor-state.c++
```diff
@@ -122,7 +122,8 @@ ActorObserver& currentActorMetrics() {
 
 jsg::JsRef<jsg::JsValue> listResultsToMap(
     jsg::Lock& js, ActorCacheOps::GetResultList value, bool completelyCached) {
-  return js.withinHandleScope([&] {
+  return js
+      .withinHandleScope([&] {
```
anonrig, line 126: This change seem unnecessary

jasnell:
Ok, but we can do that separately. In either case this method is likely to stick around and should support `JsValue`

src/workerd/jsg/jsg.h
```diff
@@ -2251,6 +2251,12 @@ class JsMessage;
 JS_TYPE_CLASSES(V)
 #undef V
 
+#define V(Name) || kj::isSameType<T, Js##Name>()
+template <typename T>
+concept IsJsValue =
+    kj::isSameType<T, JsValue>() || kj::isSameType<T, JsMessage>() JS_TYPE_CLASSES(V);
```
anonrig, line 2257: Can't you just use "std::is_base_of"? Something like this

```
template <typename T>
concept IsJsValueDerived = std::is_base_of_v<JsValue, T>;
```

anonrig:
> Ok, but we can do that separately. In either case this method is likely to stick around and should support `JsValue`

Yes, I'm not blocking just sharing some context/knowledge on this.

src/workerd/api/actor-state.c++
jasnell: This was the `just f` formatter for some odd reason. Really not sure why it did this.

src/workerd/jsg/jsg.h
jasnell: Since `kj` is more idiomatic through the code base, I'd prefer to use `kj::isSameType`. Also, the various `Js...` types do not inherit from `JsValue`. They are assignable to `JsValue` but `JsValue` is not the base class.

```
src/workerd/jsg/jsvalue.h:677:15: error: static assertion failed due to requirement 'std::is_base_of_v<workerd::jsg::JsValue, workerd::jsg::JsObject>'
  677 | static_assert(std::is_base_of_v<JsValue, JsObject>);
```

src/workerd/api/actor-state.c++
jasnell: Yeah, just confirmed again, `just f` is doing this and complains about it if I change it back.

anonrig (approved):

PR #4165, opened 2025-05-18
Refactor how IoContext aborts are implemented

kentonv:
IoContext abort logic evolved gradually and, much like evolution, the result was a mess.

This PR attempts to improve this.

Note that this is based on top of #4161; please review that first, and I'll keep this one as a draft in the meantime.

Aside from cleaning things up, this fixes two bugs:
- `IoContext::abortException` was only set for aborts caused by someone explicitly calling `IoContext::abort()`, but many kinds of aborts (such as input gate and output gate failures) did not come in through this path. So, the previous attempt to make `run()` rethrow the abort exception if called post-abort did not always work. Now, all kinds of aborts call `abort()`, so the exception propagates properly.
- In actors, aborting a request did not actually cancel I/O tasks, bit it did cancel the drain() thus causing all IncomingRequests to end. If someone continued to hold open a stub to the actor, I/O tasks could keep running and then generate spurious `expected !incomingRequests.empty()` errors later when they tried to re-enter the isolate.
- There were various race where an aborted context might throw some exception at the client that was not the abort exception, such as throwing "This promise will never resolve" or "The script will never generate a response". Many of these should be fixed now to actually propagate the abort exception.


src/workerd/api/actor-state.c++
```diff
@@ -889,8 +889,14 @@ void DurableObjectState::abort(jsg::Optional<kj::String> reason) {
     s.get()->getActorCacheInterface().shutdown(error);
   }
 
-  IoContext::current().abort(kj::cp(error));
-  kj::throwFatalException(kj::mv(error));
+  IoContext::current().abort(kj::mv(error));
+  js.terminateExecution();
```
ObsidianMinor, line 893: Yeah something I noticed with testing was that you could _catch_ the exception thrown from a call to `DurableObjectState::abort` in JS and do other stuff, but it didn't really seem useful or that terrible of a bug since it broke the output gate which meant you couldn't communicate to the outside world anyway.

src/workerd/api/global-scope.c++
```diff
@@ -86,14 +86,20 @@ void ExecutionContext::abort(jsg::Lock& js, jsg::Optional<jsg::Value> reason) {
   // TODO(someday): Maybe instead of throwing we should TerminateExecution() here? But that
   //   requires some more extensive changes.
```
ObsidianMinor, line 87: Old comment

src/workerd/io/io-context.h
```diff
@@ -1392,19 +1392,6 @@ kj::_::ReducePromises<RemoveIoOwn<T>> IoContext::awaitJs(jsg::Lock& js, jsg::Pro
         fulfiller->reject(JSG_KJ_EXCEPTION(FAILED, Error, "Promise will never complete."));
       }
     }
-
-    kj::Maybe<kj::StringPtr> finalize() override {
-      if (!isDone) {
-        reject();
-        isDone = true;
-        return "A hanging Promise was canceled. This happens when the worker runtime is waiting "
```
ObsidianMinor, line 1397: I wonder if there's a way to have this error still exist?

kentonv: This message was only logged to the devtools console (only in preview). From what I can tell, most people hit by this error never saw this extended description. I think what we really need to do is extend the error in the exception we actually throw. I do want to do that but I kind of want to do it in a separate PR since I think a bunch of people will be interested in bikeshedding it.

kentonv:
The internal build will fail until internal PR 10559 is merged to improve the output gate test. That PR can be merged first; it doesn't depend on this PR, it just enables this PR.

src/workerd/io/io-context.c++
```diff
@@ -1287,7 +1287,9 @@ void IoContext::abortFromHang(Worker::AsyncLock& asyncLock) {
   // Don't bother aborting if limits were exceeded because in that case the abort promise will be
   // fulfilled shortly anyway.
   if (limitEnforcer->getLimitsExceeded() == kj::none) {
-    abort(JSG_KJ_EXCEPTION(FAILED, Error, "The script will never generate a response."));
+    abort(JSG_KJ_EXCEPTION(FAILED, Error,
+        "This request was canceled because the Workers Runtime detected that the Worker is hung "
```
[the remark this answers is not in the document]
kentonv: In case it's hard to read this diff due to this comment splitting it, the new full error text is:

> This request was canceled because the Workers Runtime detected that the Worker is hung and will never return a response.

src/workerd/api/actor-state.c++
```diff
@@ -889,8 +889,14 @@ void DurableObjectState::abort(jsg::Optional<kj::String> reason) {
...
+
+  // terminateExecution() only sets a flag. In order for V8 to notify it, we have to make it do
+  // some work that causes it to check the flag. This has been observed to do the trick.
+  // TODO(cleanup): This appears in ExecutionContext::abort(), DurableObjectState::abort(), and
+  //     processExitImpl() in node/util.c++. Consolidate?
+  jsg::check(v8::JSON::Stringify(js.v8Context(), js.str()));
```
jasnell, line 899: Hmm.. this is a bit ugly. I wonder if a variation on `js.terminateExecution()` could encapsulate this, e.g. `js.terminateExecution({ force: true });` or something similar.

src/workerd/io/io-context.c++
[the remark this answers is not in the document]
jasnell: I think the most likely case would be string matching for logging purposes. I'm ok with changing the message as the new message is a definite improvement but I do think the chance of breakage is small-but-not-zero.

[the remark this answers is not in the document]
kentonv: Do you think that requires a compat flag?

(I lean towards "no" because if the worker is throwing the exception already, it's broken either way. At worst it messes with observability but it's not going to cause outages.)

src/workerd/api/actor-state.c++
kentonv: OK I added a commit to clean this up: 6f4aa13084423f3f745cd29a514fed5a6847c6f8

src/workerd/io/io-context.c++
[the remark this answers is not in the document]
jasnell: Yeah, I agree. #YOLO.

jasnell (approved):

ObsidianMinor (approved):

kentonv:
Just rebased. Unfortunately there's one test in the internal codebase that does depend on that error message so I had to go open a PR for that...

src/workerd/io/io-context.c++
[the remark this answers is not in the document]
irvinebroque: I think fine to change. But if we can two things that help people:

1. Being very explicit that this is an issue with _your_ Worker's code (otherwise people interpret as maybe a Cloudflare problem)
2. Linking to docs for reference / additional context where we have more space to explain

Something like...

```suggestion
        "The Workers runtime canceled this request because it detected that your Worker's code hangs and will never generate a response. Refer to: https://developers.cloudflare.com/workers/observability/errors/"
```

We should update our docs: https://developers.cloudflare.com/workers/observability/errors/#the-script-will-never-generate-a-response-errors with the new message — right now if you search for the existing error message you will get helpful results. If we update our docs with the new message, then one should get our docs in search results.

irvinebroque (commented):
> There were various race where an aborted context might throw some exception at the client that was not the abort exception, such as throwing "This promise will never resolve" or "The script will never generate a response". Many of these should be fixed now to actually propagate the abort exception.

This is awesome — is there a basic example we could use in a changelog entry?

Want to highlight to customers that we are fixing things that result in opaque errors for them.

src/workerd/jsg/jsg.c++
```diff
@@ -258,10 +258,21 @@ void Lock::runMicrotasks() {
...
+void Lock::terminateNextExecution() {
   v8Isolate->TerminateExecution();
 }
 
+[[noreturn]] void Lock::terminateExecutionNow() {
+  terminateNextExecution();
+
+  // HACK: This has been observed to reliably make V8 check the termination flag and raise the
+  //   uncatchable termination exception.
+  jsg::check(v8::JSON::Stringify(v8Context(), str()));
```
balusch, line 270: That's an interesting (and weird, as always in V8) finding. Is there any V8 doc or discussion about it?

kentonv: Nope. It was just the first thing that worked and we didn't spend too much time thinking about it. Probably there are many V8 APIs we could call which should cause it to check the termination flag.

Of course, it would be better to add an explicit API to V8 to ask it to check the flag but that takes time and there's so much to do. :) If this stops working the tests will tell us.

balusch: Yeah, having an explicit API would definitely be ideal, though I imagine it wouldn’t be easy to get into V8. Still, I’m impressed you guys found this workaround — thanks for sharing it!

PR #4170, opened 2025-05-20
avoid string coercion as much as possible

anonrig:
Small optimization to get rid of unnecessary ToString() on a string.

### DIFF
--- a/src/workerd/jsg/value.h
+++ b/src/workerd/jsg/value.h
@@ -481,7 +481,8 @@ class StringWrapper {
       v8::Local<v8::Value> handle,
       kj::String*,
       kj::Maybe<v8::Local<v8::Object>> parentObject) {
-    v8::Local<v8::String> str = check(handle->ToString(context));
+    v8::Local<v8::String> str =
+        handle->IsString() ? handle.As<v8::String>() : check(handle->ToString(context));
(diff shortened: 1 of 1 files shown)


jasnell (approved):
Not that I suspect it actually matters in any practical sense but this is at least theoretically breaking.

```
String.prototype.toString = () => 1;
m = 'hello';
console.log(m); // 'hello'
console.log(m.toString());  // '1'
```

;-)

kentonv:
Does this change actually help? I would expect that the implementation of `ToString()` _already_ checks if the value is a string and, if so, just returns it. So I would expect this change is adding a redundant check which actually makes things slower, not faster. Do you have evidence to the contrary?

jasnell:
> I would expect that the implementation of ToString() already checks if the value is a string and, if so, just returns it...

Yep, I had to go back and check the implementation to verify. I had thought that it would still call the `String.prototype.toString` method but it looks like it does not. Looking at the implementation of `Value::ToString(...)` in `api.cc`, we have `if (i::IsString(*obj)) return ToApiHandle<String>(obj);` so for string values it is not performing any coercion.

anonrig:
Opened a PR to revert @jasnell @kentonv https://github.com/cloudflare/workerd/pull/4174

kentonv:
Thanks for checking!

PR #4171, opened 2025-05-20
Implement connect_pass_through experimental compat flag.

kentonv:
See comments in compatibility-date.capnp for explanation. Miniflare needs this.

cc @penalosa

### DIFF
--- a/src/workerd/io/worker-entrypoint.c++
+++ b/src/workerd/io/worker-entrypoint.c++
@@ -498,7 +498,7 @@ kj::Promise<void> WorkerEntrypoint::connect(kj::StringPtr host,
     kj::AsyncIoStream& connection,
     ConnectResponse& response,
     kj::HttpConnectSettings settings) {
-  TRACE_EVENT("workerd", "WorkerEntrypoint::runScheduled()");
+  TRACE_EVENT("workerd", "WorkerEntrypoint::connect()");
   auto incomingRequest =
       kj::mv(KJ_REQUIRE_NONNULL(this->incomingRequest, "connect() can only be called once"));
   this->incomingRequest = kj::none;
--- a/src/workerd/io/worker-entrypoint.c++
+++ b/src/workerd/io/worker-entrypoint.c++
@@ -498,6 +498,31 @@ kj::Promise<void> WorkerEntrypoint::connect(kj::StringPtr host,
     kj::AsyncIoStream& connection,
     ConnectResponse& response,
     kj::HttpConnectSettings settings) {
+  TRACE_EVENT("workerd", "WorkerEntrypoint::runScheduled()");
+  auto incomingRequest =
+      kj::mv(KJ_REQUIRE_NONNULL(this->incomingRequest, "connect() can only be called once"));
+  this->incomingRequest = kj::none;
+  incomingRequest->delivered();
+  auto& context = incomingRequest->getContext();
+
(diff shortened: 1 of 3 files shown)


kentonv:
(Added test, seems to work.)

src/workerd/io/worker-entrypoint.c++
```diff
@@ -498,6 +498,31 @@ kj::Promise<void> WorkerEntrypoint::connect(kj::StringPtr host,
     kj::AsyncIoStream& connection,
     ConnectResponse& response,
     kj::HttpConnectSettings settings) {
+  TRACE_EVENT("workerd", "WorkerEntrypoint::runScheduled()");
```
jasnell, line 501:
```suggestion
  TRACE_EVENT("workerd", "WorkerEntrypoint::connect()");
```

src/workerd/io/worker-entrypoint.c++
```diff
@@ -498,6 +498,31 @@ kj::Promise<void> WorkerEntrypoint::connect(kj::StringPtr host,
...
+
+    auto next = context.getSubrequestChannelNoChecks(
+        IoContext::NEXT_CLIENT_CHANNEL, false, kj::mv(cfBlobJson));
+
+    // Note: Intentionally return without co_await so that the `incomingRequest` is destroyed,
+    //   because we don't have any need to keep the context around.
+    return next->connect(host, headers, connection, response, settings);
+  }
+
   JSG_FAIL_REQUIRE(TypeError, "Incoming CONNECT on a worker not supported");
```
jasnell, line 526:
```suggestion
  JSG_REQUIRE(context.getWorker().getIsolate().getApi().getFeatureFlags().getConnectPassThrough(),
      TypeError, "Incoming CONNECT on a worker not supported");

  // connect_pass_through feature flag means we should just forward the connect request on to
  // the global outbound.

  auto next = context.getSubrequestChannelNoChecks(
      IoContext::NEXT_CLIENT_CHANNEL, false, kj::mv(cfBlobJson));

  // Note: Intentionally return without co_await so that the `incomingRequest` is destroyed,
  //   because we don't have any need to keep the context around.
  return next->connect(host, headers, connection, response, settings);
```

src/workerd/server/server-test.c++
```diff
@@ -1138,6 +1138,185 @@ KJ_TEST("Server: override globalOutbound") {
   conn.recvHttp200("OK");
 }
 
+KJ_TEST("Server: connect() to default outbound") {
+  TestServer test(singleWorker(R"((
+    compatibilityDate = "2022-08-17",
```
jasnell, line 1143: nit/question: is this intentionally using an older compat date?

jasnell (approved):
Couple minor nits but consider them optional

src/workerd/io/worker-entrypoint.c++
kentonv: I structured it this way intentionally as I assume we'll need to continue supporting this flag if and when we create a real API. When that happens we'll remove the assert at the end of the function and add code there, but not have to touch the block that tests the feature flag and handles it.

src/workerd/server/server-test.c++
kentonv: It's copy/pasted from another test. A compatibility date is required but for this test the date doesn't really matter.

PR #4172, opened 2025-05-20
Add to string variations on jsg::JsString

jasnell:
Support for converting JsString to USVString, DOMString, and ByteString.

Incrementally we should work on converting more of our `kj::String` uses in APIs to `jsg::JsString` to promote lazy conversion of the strings to UTF8., e.g.


src/workerd/jsg/jsvalue.c++
```diff
@@ -300,7 +301,40 @@ size_t JsString::utf8Length(jsg::Lock& js) const {
...
+  auto buf = kj::heapArray<char>(inner->Utf8LengthV2(js.v8Isolate) + 1);
+  inner->WriteUtf8V2(js.v8Isolate, buf.begin(), buf.size(),
+      v8::String::WriteFlags::kNullTerminate | v8::String::WriteFlags::kReplaceInvalidUtf8);
+  return js.accountedUSVString(kj::mv(buf));
+}
+
+jsg::ByteString JsString::toByteString(Lock& js) const KJ_WARN_UNUSED_RESULT {
+  jsg::ByteString result(toString(js));
+
+  if (!simdutf::validate_ascii(result.begin(), result.size())) {
```
anonrig, line 317: We're duplicating the code here. Can we avoid having multiple - exact methods?

src/workerd/jsg/jsvalue.c++
```diff
@@ -300,7 +301,40 @@ size_t JsString::utf8Length(jsg::Lock& js) const {
...
+  }
+
+  return kj::mv(result);
+}
+
+jsg::DOMString JsString::toDOMString(Lock& js) const KJ_WARN_UNUSED_RESULT {
+  auto buf = kj::heapArray<char>(inner->Utf8LengthV2(js.v8Isolate) + 1);
+  inner->WriteUtf8V2(js.v8Isolate, buf.begin(), buf.size(), v8::String::WriteFlags::kNullTerminate);
+  return js.accountedDOMString(kj::mv(buf));
 }
```
anonrig, line 338: There are 4 similar functions in value.h to create kj::String, domString etc. Why don't we make this a little bit generic, and avoid having code duplication?

This will also help us avoid adding new functions whenever we add a new string type.


```C++
template <StringLike ParamType, typename ReturnType>
ReturnType ToGenericType(Lock& js, ParamType& param) const KJ_WARN_UNUSED_RESULT {
  auto buf = kj::heapArray<char>(inner->Utf8LengthV2(js.v8Isolate) + 1);
  inner->WriteUtf8V2(js.v8Isolate, buf.begin(), buf.size(), v8::String::WriteFlags::kNullTerminate);
  if constexpr (kj::isSameType<ReturnType, kj::String>()) {
    return js.accountedDOMString(kj::mv(buf));
  }
  if constexpr (kj::isSameType<ReturnType, jsg::DOMString>()) {
    return js.accountedDOMString(kj::mv(buf));
  }
  // TODO: Implement conversion to other types
}
```

npaun (approved):

src/workerd/jsg/jsvalue.c++
```diff
@@ -300,7 +301,40 @@ size_t JsString::utf8Length(jsg::Lock& js) const {
...
 
 kj::String JsString::toString(jsg::Lock& js) const {
-  return kj::str(inner);
+  return js.accountedKjString(kj::str(inner));
```
kentonv, line 304: I'm concerned that the accounting adds considerable overhead for small strings, do we really want to do it automatically by default? Maybe this should apply only to large strings?

src/workerd/jsg/jsvalue.c++
jasnell: I was just working on consolidating dups in a follow on commit

src/workerd/jsg/jsvalue.c++
jasnell: This PR does not change any of the accounting. If we are seeing a negative impact from the accounting that was added then we can/should address that in a separate PR.

src/workerd/jsg/jsvalue.c++
jasnell: I'd rather not introduce yet another templated method into the mix here. I've added a commit that eliminates the duplicated code paths.

src/workerd/jsg/value.h
```diff
@@ -481,66 +481,32 @@ class StringWrapper {
...
-      if (str->ContainsOnlyOneByte()) {
-        result.warning = ByteString::Warning::CONTAINS_EXTENDED_ASCII;
-      } else {
-        // Storage is two-bytes and it contains two-byte characters.
-        result.warning = ByteString::Warning::CONTAINS_UNICODE;
-      }
-    }
-
-    return kj::mv(result);
+    JsString str(check(handle->ToString(context)));
```
anonrig, line 492: This could benefit from (handle->IsString() checks and by avoiding string coercion

src/workerd/jsg/value.h
```diff
@@ -481,66 +481,32 @@ class StringWrapper {
...
       v8::Local<v8::Value> handle,
       DOMString*,
       kj::Maybe<v8::Local<v8::Object>> parentObject) {
-    v8::Local<v8::String> str = check(handle->ToString(context));
-    v8::Isolate* isolate = context->GetIsolate();
-    auto& js = Lock::from(isolate);
-    auto buf = kj::heapArray<char>(str->Utf8LengthV2(isolate) + 1);
-    str->WriteUtf8V2(isolate, buf.begin(), buf.size(), v8::String::WriteFlags::kNullTerminate);
-    return js.accountedDOMString(kj::mv(buf));
+    JsString str(check(handle->ToString(context)));
```
anonrig, line 508: ditto

src/workerd/jsg/value.h
```diff
@@ -481,66 +481,32 @@ class StringWrapper {
...
+    return str.toDOMString(Lock::from(context->GetIsolate()));
```
anonrig, line 509: What about using a single function but take string type as a template argument?

```suggestion
    return str.toString<DOMString>(Lock::from(context->GetIsolate()));
```

src/workerd/jsg/value.h
jasnell: String coercion is expected in these methods. For instance, if I have a C++ signature like `void foo(kj::String)`, and in JavaScript I pass `foo(1)`, then the expectation is that the the argument will be coerced to a string using the `ToString(...)` algorithm. To avoid coercion we have the `jsg::NonCoercible<T>` utility.

src/workerd/jsg/value.h
jasnell: I'm really not convinced that provides any real benefit here. It's more to type and makes both implementation and compilation more complicated. I'd very much rather just have the separate methods.

src/workerd/jsg/value.h
jasnell: I've updated to check for the IsString to at least avoid calling ToString if the value is already a string and left the remain coercion in place. 

Note that there is a negligible chance this could be breaking but it is exceedingly unlikely. Specifically, if someone was overriding the `String.prototype.toString` method for whatever unlikely reason, this will produce a different result.

```js
String.prototype.toString = () => '1';
m = 'abc';
console.log(m);   // 'abc'
console.log(m.toString());  // '1'
```

Previously with the type mapping here, in this edge case, the typewrapper would yield `"1"` as the value. After this change, it would yield `"abc"` as the value since the String.prototype.toString is no longer being called. The likelihood of anyone actually doing something like that is nearly non-existent as it would be rather pointless... but, have to at least acknowledge the possibility.

jasnell: Actually.. no, this is wrong. `Value::ToString` does not call `toString()` on string values, it just returns the string directly. So, nevermind :-)

src/workerd/jsg/value.h
```diff
@@ -508,66 +507,37 @@ class StringWrapper {
...
-    auto buf = kj::heapArray<char>(str->Utf8LengthV2(isolate) + 1);
-    str->WriteUtf8V2(isolate, buf.begin(), buf.size(), v8::String::WriteFlags::kNullTerminate);
-    auto& js = Lock::from(isolate);
-    return js.accountedKjString(kj::mv(buf));
+    // Note that if handle is already a string, calling ToString will just
+    // return handle without any further coercion. For any other type of
+    // value, v8 will try to coerce it into a string. So there is no need
+    // for us to check if handle is a string here or not, ToString does
+    // that for us.
+    JsString str(check(handle->ToString(context)));
```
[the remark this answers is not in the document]
jasnell: For now, given that the code previously used `ToString(...)` and given the fact that this PR is not meant to change that, I'll leave it as it. It's not entirely apparent that avoiding the handle scope creation will have any meaningful impact on performance with this operation.

src/workerd/jsg/jsvalue.c++
jasnell: We haven't measured in any meaningful way yet but that's something that can be looked at separately.

PR #4173, opened 2025-05-20
Implements wrapper for Python Durable Object fetch.

dom96:
TODO: need to upstream custom_fetcher


hoodmane:
> For now implemented only in 0.26.0a2 because of lack of custom_fetcher support in pyodide's pyfetch

Can you explain a bit more what you mean by this? It seems like it would make sense to implement in 0.27 too?

src/pyodide/internal/_workers.py
```diff
@@ -851,6 +874,27 @@ def __getattr__(self, name):
         setattr(self, name, result)
         return result
 
+    def _fetch_wrapper(self, fetcher, *args, **kwargs):
+        kwargs["fetcher"] = fetcher
+        return fetch(*args, **kwargs)
```
hoodmane: What about:
```suggestion
    def fetch(self, *args, **kwargs):
        return fetch(*args, fetcher=self._binding.fetch, **kwargs)
```

src/pyodide/internal/_workers.py
```diff
@@ -157,13 +159,35 @@ async def blob(self) -> "Blob":
         return Blob(await self.js_object.blob())
 
 
+from js import Object
+
+
+async def _pyfetch(url: str, **kwargs: Any) -> FetchResponse:
```
hoodmane: Maybe call this `_pyfetch_patched`?

src/pyodide/internal/_workers.py
```diff
@@ -157,13 +159,35 @@ async def blob(self) -> "Blob":
         return Blob(await self.js_object.blob())
 
 
+from js import Object
```
hoodmane: Can you move this import to the top of the file?

src/pyodide/internal/_workers.py
```diff
@@ -157,13 +159,35 @@ async def blob(self) -> "Blob":
...
+                    url, to_js(kwargs, dict_converter=Object.fromEntries)
+                ),
+            )
+        except JsException as e:
+            raise OSError(e.message) from None
+    else:
+        return pyfetch(url, **kwargs)
+
+
 async def fetch(
```
hoodmane, line 183: Could we rename this to something other than `fetch`? Since it's quite overloaded? Or this is intended specifically to be a public function?

src/pyodide/internal/_workers.py
hoodmane: It is unfortunate that there are three different things called `fetch` in one place.

src/pyodide/internal/_workers.py
```diff
@@ -851,6 +874,27 @@ def __getattr__(self, name):
...
+        if name == "get":
+            return lambda *args, **kwargs: _FetcherWrapper(attr(*args, **kwargs))
+
+        return attr
+
+    def __getattr__(self, name):
+        result = self._getattr_helper(name)
+        setattr(self, name, result)
+        return result
+
```
hoodmane, line 888:
```suggestion
    def get(self, *args, **kwargs):
        return _FetcherWrapper(self._binding.get(*args, **kwargs))
```

src/pyodide/internal/_workers.py
```diff
@@ -851,6 +874,27 @@ def __getattr__(self, name):
...
+        attr = getattr(self._binding, name)
+        if name == "get":
+            return lambda *args, **kwargs: _FetcherWrapper(attr(*args, **kwargs))
+
+        return attr
+
+    def __getattr__(self, name):
+        result = self._getattr_helper(name)
+        setattr(self, name, result)
+        return result
```
hoodmane:
```suggestion
        return getattr(self._binding, name)
```

dom96:
> Can you explain a bit more what you mean by this? It seems like it would make sense to implement in 0.27 too?

I was hoping we could get https://github.com/pyodide/pyodide/pull/5655 merged in and update 0.27 to include it.

src/pyodide/internal/_workers.py
dom96: Hmm, I'm a bit on the fence on this. But your approach is cleaner. Done.

src/pyodide/internal/_workers.py
dom96: It is intended to be public

hoodmane (approved):
Thanks!

PR #4179, opened 2025-05-21
Expand float16 support

fhanau:
Now that we have float16 support enabled, we can get most tests using float16 to work. Note that this is still incomplete since Node's types do not include float16 in the TypedArray definition yet.

### DIFF
--- a/src/node/internal/internal_comparisons.ts
+++ b/src/node/internal/internal_comparisons.ts
@@ -66,7 +66,7 @@ function areSimilarRegExps(a: RegExp, b: RegExp) {
   );
 }
 
-type FloatArray = Float32Array | Float64Array;
+type FloatArray = Float16Array | Float32Array | Float64Array;
 type AnyArrayBuffer = ArrayBuffer | SharedArrayBuffer;
 
 type Memos = {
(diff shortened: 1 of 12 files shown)


fhanau:
@anonrig I have two (smaller) concerns here:
1) before this PR float16 is available but not fully supported in JSG, resulting in the test failures being addressed here. Could that cause confusion for developers/be a reason to avoid releasing momentarily?
2) We support fp16 more with this change but still not to 100% – there's missing TypedArray support in node types, so I also left our TypedArray definitions as-is to avoid build failures caused by the mismatch. Maybe this still-limited amount of float16 support will again make it more difficult for developers to use, but hopefully this is safe to merge still.

anonrig (approved):

anonrig:
@fhanau we can update node/types in a follow up. I think it's good that we're making these changes.

fhanau:
> What is the missing piece to fully support it?

I got build errors after adding Float16Array in a few more places where Float32Array is present, and I think that was caused by Float16Array not being part of the TypedArray definition in node types. (Based on https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/globals.typedarray.d.ts that might still be the case with latest node?)

PR #4180, opened 2025-05-21
disallow kj::Promise in v8 fast api

anonrig:
Disallow kj::promise for now to understand if it's causing an issue with v8 fast api.

### DIFF
--- a/src/workerd/jsg/fast-api.h
+++ b/src/workerd/jsg/fast-api.h
@@ -16,6 +16,7 @@
 #include <v8-local-handle.h>
 #include <v8-value.h>
 
+#include <kj/async.h>
 #include <kj/common.h>
 
 namespace workerd::jsg {
@@ -28,6 +29,12 @@ constexpr bool isFunctionCallbackInfo = false;
 template <typename T>
 constexpr bool isFunctionCallbackInfo<v8::FunctionCallbackInfo<T>> = true;
 
+template <typename T>
+constexpr bool isKjPromise = false;
+
+template <typename T>
+constexpr bool isKjPromise<kj::Promise<T>> = true;
+
 // These types are passed by fast api as is and do not require wrapping/unwrapping.
 template <typename T>
 concept FastApiPrimitive = kj::isSameType<T, void>() || kj::isSameType<T, bool>() ||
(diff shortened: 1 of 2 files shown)


jasnell (approved):

src/workerd/jsg/fast-api.h
```diff
@@ -48,7 +55,8 @@ concept FastApiPrimitive = kj::isSameType<T, void>() || kj::isSameType<T, bool>(
 
 // Helper to determine if a type can be used as a parameter in V8 Fast API
 template <typename T>
-concept FastApiParam = !isFunctionCallbackInfo<kj::RemoveConst<kj::Decay<T>>>;
+concept FastApiParam = !isFunctionCallbackInfo<kj::RemoveConst<kj::Decay<T>>> &&
+    !isKjPromise<kj::RemoveConst<kj::Decay<T>>>;
```
kentonv, line 59: `RemoveConst<Decay<T>>` is redundant, `Decay` already removes const.

kentonv (commented):
This bans `kj::Promise` parameters but not promises in general, e.g. methods can take `jsg::Promise` or `v8::Local<v8::Promise>` or just `JsValue` that happens to contain a promise.

Hard to think of what's special about promises that would cause a problem with fastapi.

jasnell:
> Hard to think of what's special about promises that would cause a problem with fastapi...

Yep, I think this is mostly experimentation because we're not yet quite sure what exactly was causing the failures in the experimental deployment. This restriction may be entirely unnecessary but tracking down the cause of the crashes that were seen has proven to be rather tricky

anonrig:
> This bans `kj::Promise` parameters but not promises in general, e.g. methods can take `jsg::Promise` or `v8::Local<v8::Promise>` or just `JsValue` that happens to contain a promise.

@kentonv You're correct. We don't have any existing methods that takes a jsg::promise which is also compatible with v8 fast api at the moment.

src/workerd/jsg/fast-api.h
anonrig: Good to know. Will do a follow up. Thanks!

kentonv:
Just keep in mind, if this "fixes" the crashes, it doesn't necessarily mean that the problem was related to promises. It just means the problem was restricted to one or more functions that happened to take Promises as input, but the issue could be something entirely different about those functions.

jasnell:
>  Just keep in mind, if this "fixes" the crashes.....

Yep, I think at this point we're largely just looking for clues to point in the right direction since the stacks on the crashes have been quite useless in helping to pinpoint the issue. I still have a suspicion that the issue has something to do with gc or an invalid local handle but it's difficult to pin down. Hopefully the experiment with this change will yield more information but honest I'm not optimistic.

PR #4181, opened 2025-05-21
Initialise ctx/env fields on Python entrypoint superclass constructors.

dom96:
This is to match behaviour with JS which does initialise these.


anonrig (approved):

dom96:
Thanks for the review Yagiz!

PR #4189, opened 2025-05-22
Verbose logging to debug flaky weakref tests

ketanhwr:
This test timing out is quite weird - presumably happens because of unclean process exit, since the worker doesn't have anything which would cause it to "hang" necessarily. The CI logs also possibly point to the same problem that I'm hypothesizing:

```
==================== Test output for //src/workerd/server/tests/weakref:js-weak-ref-test:
TIMEOUT: //src/workerd/server/tests/weakref:js-weak-ref-test (Summary)
TAP version 13
      /private/var/tmp/_bazel_runner/95172a8db72435c774c2789dc64209fb/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/src/workerd/server/tests/weakref/js-weak-ref-test/test.log
workerd/server/server.c++:3994: info: Inspector is listening
INFO: From Testing //src/workerd/server/tests/weakref:js-weak-ref-test:
workerd/server/server.c++:3994: info: Inspector is listening
# Subtest: JS FinalizationRegistry
ok 1 - JS FinalizationRegistry
  ---
  duration_ms: 130.306625
  ...
-- Test timed out at 2025-05-21 21:17:48 UTC --
/private/var/tmp/_bazel_runner/95172a8db72435c774c2789dc64209fb/sandbox/darwin-sandbox/6104/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/src/workerd/server/tests/weakref/js-weak-ref-test_/js-weak-ref-test.runfiles/_main/src/workerd/server/tests/weakref/js-weak-ref-test_/js-weak-ref-test: line 588: 14806 Terminated: 15          "$JS_BINARY__NODE_WRAPPER" ${JS_BINARY__NODE_OPTIONS[@]+"${JS_BINARY__NODE_OPTIONS[@]}"} -- "$entry_point" ${ARGS[@]+"${ARGS[@]}"} 0<&0
```

For now I asked Claude to improve the child process termination and add some more logging  to ease in debugging further. (unable to reproduce in dev even after `runs_per_test=500`)

### DIFF
--- a/src/workerd/server/tests/server-harness.mjs
+++ b/src/workerd/server/tests/server-harness.mjs
@@ -49,6 +49,7 @@ export class WorkerdServerHarness {
     };
 
     // Start the subprocess.
+    console.log('[HARNESS] Starting workerd with args:', args);
     this.#child = spawn(this.#workerdBinary, args, options);
 
     // Create a promise for every named listen port we were told in our constructor to expect. Parse
@@ -64,6 +65,7 @@ export class WorkerdServerHarness {
         new Promise((resolve, reject) => {
           this.#child.stdio[CONTROL_FD].on('data', (data) => {
             const parsed = JSON.parse(data);
+            console.log('[HARNESS] Control message:', parsed);
             if (parsed.event === 'listen' && parsed.socket === listenPort) {
               resolve(parsed.port);
             }
@@ -77,6 +79,7 @@ export class WorkerdServerHarness {
     this.#listenInspectorPort = new Promise((resolve, reject) => {
       this.#child.stdio[CONTROL_FD].on('data', (data) => {
         const parsed = JSON.parse(data);
+        console.log('[HARNESS] Inspector message:', parsed);
         if (parsed.event === 'listen-inspector') {
           resolve(parsed.port);
         }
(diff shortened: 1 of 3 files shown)


ketanhwr:
I've marked this test as [flaky](https://bazel.build/reference/be/common-definitions) so people don't get blocked while I try to debug this. But I haven't seen this flag used anywhere else though so not sure if it'll work (cc: @fhanau )

```
flaky	
Boolean; nonconfigurable; default is False

Marks test as flaky.

If set, executes the test up to three times, marking it as failed only if it fails each time. By default, this attribute is set to False and the test is executed only once. Note, that use of this attribute is generally discouraged - tests should pass reliably when their assertions are upheld.
```

anonrig (approved):

fhanau:
> I've marked this test as [flaky](https://bazel.build/reference/be/common-definitions) so people don't get blocked while I try to debug this. But I haven't seen this flag used anywhere else though so not sure if it'll work (cc: @fhanau )
> 
> ```
> flaky	
> Boolean; nonconfigurable; default is False
> 
> Marks test as flaky.
> 
> If set, executes the test up to three times, marking it as failed only if it fails each time. By default, this attribute is set to False and the test is executed only once. Note, that use of this attribute is generally discouraged - tests should pass reliably when their assertions are upheld.
> ```

We use the flag extensively internally, that should provide context here.

ketanhwr:
> We use the flag extensively internally, that should provide context here.

Ahh grep fail :facepalm:, thanks for confirming!

PR #4190, opened 2025-05-22
Preload python packages in workerd like in EW.

dom96:


src/workerd/server/workerd-api.h
```diff
@@ -285,4 +287,29 @@ class WorkerdApi final: public Worker::Api {
 kj::Maybe<jsg::Bundle::Reader> fetchPyodideBundle(
     const api::pyodide::PythonConfig& pyConfig, kj::StringPtr version);
 
+// Downloads a package with retry logic (up to 3 attempts with 5-second delays)
```
jasnell, line 290: workerd-api.h is already fairly complex. Is it possible to move these new APIs into a separate pyodide compilation unit or maybe into the existing `api/pyodide.h/c++` ?

src/workerd/server/workerd-api.c++
```diff
@@ -331,6 +333,185 @@ kj::Maybe<jsg::Bundle::Reader> fetchPyodideBundle(
   return pyConfig.pyodideBundleManager.getPyodideBundle(version);
 }
 
+/**
+ * This function matches the implementation of `getPythonRequirements` in Edgeworker. But it
```
jasnell, line 337:
```suggestion
 * This function matches the implementation of `getPythonRequirements` in the internal repo. But it
```

We don't use the internal project name externally

src/workerd/server/workerd-api.c++
```diff
@@ -737,6 +918,14 @@ void WorkerdApi::compileModules(jsg::Lock& lockParam,
       auto bundle = KJ_ASSERT_NONNULL(
           fetchPyodideBundle(impl->pythonConfig, version), "Failed to get Pyodide bundle");
 
+      // Get Python requirements and fetch packages
+      KJ_IF_SOME(a, artifacts) {
+        KJ_IF_SOME(pm, a->packageManager) {
+          auto pythonRequirements = getPythonRequirements(source);
+          fetchPyodidePackages(impl->pythonConfig, pm, pythonRequirements, pythonRelease);
+        }
+      }
```
jasnell, line 927: Will also need to add this to the new module registry setup once https://github.com/cloudflare/workerd/pull/4191 lands... If possible I'd like to get #4191 landed first

src/workerd/server/workerd-api.c++
```diff
@@ -331,6 +333,185 @@ kj::Maybe<jsg::Bundle::Reader> fetchPyodideBundle(
   return pyConfig.pyodideBundleManager.getPyodideBundle(version);
 }
 
+/**
+ * This function matches the implementation of `getPythonRequirements` in Edgeworker. But it
+ * works on the workerd ModulesSource definition rather than the WorkerBundle.
+ */
+kj::Array<kj::String> getPythonRequirements(const Worker::Script::ModulesSource& source) {
```
jasnell, line 340: If all of this were moved into `api/pyodide.h/c++` could it be reused by both the internal repo and workerd? Or is the impl here specific to workerd

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -581,4 +586,69 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
   return kj::str(pyodideRelease.getPyodide(), "_", pyodideRelease.getPyodideRevision(), "_",
       pyodideRelease.getBackport());
 }
+
+namespace api::pyodide {
+
+// Returns a string containing the contents of the hashset, delimited by ", "
+kj::String hashsetToString(kj::HashSet<kj::String>& set) {
+  if (set.size() == 0) {
+    return kj::str("");
```
jasnell, line 595: Returning `kj::String()` instead of `kj::str("")` is more efficient

```suggestion
    return kj::String();
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -581,4 +586,69 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
   return kj::str(pyodideRelease.getPyodide(), "_", pyodideRelease.getPyodideRevision(), "_",
       pyodideRelease.getBackport());
 }
+
+namespace api::pyodide {
+
+// Returns a string containing the contents of the hashset, delimited by ", "
+kj::String hashsetToString(kj::HashSet<kj::String>& set) {
```
jasnell, line 593: Nit... optional 
```suggestion
kj::String hashsetToString(const kj::HashSet<kj::String>& set) {
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -581,4 +586,69 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+  if (set.size() == 0) {
+    return kj::str("");
+  }
+
+  kj::Vector<kj::StringPtr> elems;
+  for (const auto& e: set) {
+    elems.add(e);
+  }
```
jasnell, line 601: Consider using KJ_STACK_ARRAY here to avoid the additional heap allocation when the set it small enough.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -581,4 +586,69 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+      if (maybeRow == kj::none && packageType == "cpython_module") {
+        res.add(kj::mv(fileName));
+      }
+    }
+  }
+
+  if (allRequirements.size() != 0) {
+    JSG_FAIL_REQUIRE(Error,
+        "Requested Python package(s) that are not supported: ", hashsetToString(allRequirements));
+  }
```
jasnell, line 647:
```suggestion
  JSG_REQUIRE(allRequirements.size() == 0, Error,
        "Requested Python package(s) that are not supported: ", hashsetToString(allRequirements));
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -15,6 +15,9 @@
 #include <kj/debug.h>
 #include <kj/string.h>
 
+#include <algorithm>
+#include <atomic>
```
jasnell, line 19: Is `atomic` actually being used? Sorry if I'm overlooking it but I'm not seeing where you're using it.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -581,4 +586,69 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+    }
+
+    // TODO: See above, TL;DR remove this once we release new Python/package version.
+    if (packagesVersion == "20240829.4") {
+      auto packageType = getField(obj, "package_type").getString();
+      if (maybeRow == kj::none && packageType == "cpython_module") {
+        res.add(kj::mv(fileName));
+      }
+    }
+  }
```
jasnell, line 642:
```suggestion
    KJ_IF_SOME(row, allRequirements.find(name)) {
      allRequirements.erase(row);
      res.add(kj::mv(fileName));
    } else {
      // TODO: See above, TL;DR remove this once we release new Python/package version.
      if (packagesVersion == "20240829.4") {
        auto packageType = getField(obj, "package_type").getString();
        if (packageType == "cpython_module") {
          res.add(kj::mv(fileName));
        }
      }
    }
  }
```

src/workerd/server/workerd-api.c++
dom96: Nope, it is workerd-specific because workerd has its own WorkerBundle definition.

src/workerd/api/pyodide/pyodide.c++
dom96: This is only used for debugging so I don't think it's worth these small optimisations. Also this code is copied from our internal repo.

src/workerd/server/workerd-api.c++
jasnell: That PR landed now.

src/pyodide/internal/loadPackage.ts
```diff
@@ -43,57 +31,14 @@ function getPackageMetadata(requirement: string): PackageDeclaration {
...
-      console.log('Retrying fetch in 5 seconds...');
-      await new Promise((r) => setTimeout(r, 5000));
-      return loadBundleFromR2WithRetry(requirement, currRetry + 1);
-    }
-
-    throw new Error(`Could not load '${requirement}' from R2: ${exc}`);
-  }
+// Helper function to get the path to a package in the Python package bucket
+function getPyodidePackagePath(version: string, filename: string): string {
+  return 'python-package-bucket/' + version + '/' + filename;
```
hoodmane, line 36:
```suggestion
  return `python-package-bucket/${version}/${filename}`;
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -12,9 +12,13 @@
 
 #include <kj/array.h>
 #include <kj/common.h>
+#include <kj/compat/gzip.h>
+#include <kj/compat/tls.h>
 #include <kj/debug.h>
 #include <kj/string.h>
 
+#include <algorithm>
```
hoodmane, line 20: What's this header for?

hoodmane: Ah `std::sort` I see. Please add a comment.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+      // Add each package loading promise to our list
+      promises.add(loadPyodidePackage(
+          pyConfig, pyodidePackageManager, packagesVersion, filename, *client, timer, table));
+    }
+
+    // Wait for all package downloads to complete
+    if (!promises.empty()) {
+      kj::joinPromises(promises.releaseAsArray()).wait(io.waitScope);
+    }
+  });
```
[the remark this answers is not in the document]
dom96: This is handled in the same way as fetchPyodideBundle and that seems to work fine.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@
...
+    return;
+  }
+
+  // Get the filenames of packages to fetch
+  auto filenames =
+      getPythonPackageFiles(KJ_ASSERT_NONNULL(pyodideLock), pythonRequirements, packagesVersion);
+
+  // Start a thread for package fetching
+  kj::Thread([&pyConfig, &pyodidePackageManager, filenames = kj::mv(filenames),
+                 packagesVersion = kj::str(packagesVersion)]() mutable {
```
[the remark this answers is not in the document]
dom96: The destructor of kj::Thread joins the thread, so this function will block while the thread runs.

src/pyodide/internal/loadPackage.ts
```diff
@@ -43,57 +31,14 @@ function getPackageMetadata(requirement: string): PackageDeclaration {
...
-      await new Promise((r) => setTimeout(r, 5000));
-      return loadBundleFromR2WithRetry(requirement, currRetry + 1);
-    }
-
-    throw new Error(`Could not load '${requirement}' from R2: ${exc}`);
-  }
+// Helper function to get the path to a package in the Python package bucket
+function getPyodidePackagePath(version: string, filename: string): string {
+  return `python-package-bucket/${version}/${filename}`;
 }
```
hoodmane: I don't think we need this function since it's a one-liner and only called in one place. I do like using the template string over `+`.

src/pyodide/internal/loadPackage.ts
```diff
@@ -43,57 +31,14 @@ function getPackageMetadata(requirement: string): PackageDeclaration {
...
 
 function loadBundleFromArtifactBundler(requirement: string): Promise<Reader> {
```
hoodmane: Since there is only one branch now, make it return a `Reader` and get rid of `Promise.resolve()`.

src/pyodide/internal/loadPackage.ts
```diff
@@ -137,7 +61,7 @@ async function loadPackagesImpl(
     if (VIRTUALIZED_DIR.hasRequirementLoaded(req)) {
       continue;
     }
-    loadPromises.push(loadBundle(req).then((r) => [req, r]));
+    loadPromises.push(loadBundleFromArtifactBundler(req).then((r) => [req, r]));
```
hoodmane: Since loading is all synchronous now, the two for loops  should be combined:
```js
  for (const req of requirements) {
    if (req === 'test') {
      continue; // Skip the test package, it is only useful for internal Python regression testing.
    }
    if (VIRTUALIZED_DIR.hasRequirementLoaded(req)) {
      continue;
    }
    const reader = loadBundleFromArtifactBundler(req);
    const [tarInfo, soFiles] = parseTarInfo(reader);
    const pkg = getPackageMetadata(req);
    VIRTUALIZED_DIR.addSmallBundle(
      tarInfo,
      soFiles,
      req,
      pkg.install_dir
    );
  }
```
You could even inline `loadBundleFromArtifactBundler()` potentially, though I'm also okay with leaving it as a separate function.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -410,6 +407,10 @@ if (IS_WORKERD || IS_TRACING) {
     }
   }
 
+  if (typeof mainModule.test === 'function') {
+    handlers.test = makeHandler('test');
+  }
```
hoodmane: Why are we adding this?

src/pyodide/internal/loadPackage.ts
hoodmane: `loadBundle` also doesn't need to be async anymore.

src/pyodide/internal/loadPackage.ts
```diff
@@ -177,11 +101,6 @@ export async function loadPackages(
...
   }
-  if (IS_WORKERD) {
-    await loadPackagesImpl(Module, pkgsToLoad, (req) =>
-      loadBundleFromR2WithRetry(req, 0)
-    );
-  } else {
-    await loadPackagesImpl(Module, pkgsToLoad, loadBundleFromArtifactBundler);
-  }
+
+  await loadBundle(Module, pkgsToLoad);
```
hoodmane: `loadBundle()` doesn't need to be async anymore so this function doesn't need to be async anymore either. I would probably inline `loadBundle()` into `loadPackages()`.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+  // Sort the elements for consistent output
+  auto array = elems.releaseAsArray();
+  std::sort(array.begin(), array.end());
+
+  return kj::str(kj::delimited(array, ", "_kjc));
+}
+
+kj::String getPyodidePackagePath(kj::StringPtr packagesVersion, kj::StringPtr filename) {
+  return kj::str("python-package-bucket/", packagesVersion, "/", filename);
+}
```
hoodmane: This is also only used once so I'd just inline it.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+    auto name = ent.getName();
+    auto obj = ent.getValue().getObject();
+    auto fileName = kj::str(getField(obj, "file_name").getString());
+
+    auto maybeRow = allRequirements.find(name);
+    KJ_IF_SOME(row, maybeRow) {
+      allRequirements.erase(row);
+      res.add(kj::mv(fileName));
+    } else {
+      // TODO: See above, TL;DR remove this once we release new Python/package version.
```
hoodmane: What does the comment mean? Won't we have to keep this until we stop supporting the oldest version?

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+      if (packagesVersion == "20240829.4") {
```
hoodmane: Could we say `else if (...) {`?

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+        auto packageType = getField(obj, "package_type").getString();
+        if (maybeRow == kj::none && packageType == "cpython_module") {
```
hoodmane: We already know that `maybeRow` is none because we're in the else block of the `KJ_IF_SOME()`...
```suggestion
        if (packageType == "cpython_module") {
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+  // First check if we already have this package in memory
+  if (pyodidePackageManager.getPyodidePackage(path) != kj::none) {
+    co_return;
+  }
+
+  // Then check disk cache
+  KJ_IF_SOME(diskCachePath, pyConfig.packageDiskCacheRoot) {
+    auto parsedPath = kj::Path::parse(path);
+
+    // Check if the file exists before attempting to open it
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+        KJ_LOG(WARNING, "Failed to read or process package from disk cache", path, e);
+      }
+    }
+  }
+
+  // Need to fetch from network
+  kj::String url =
+      kj::str("https://storage.googleapis.com/cloudflare-edgeworker-python-packages/", path);
+
+  // Download the package with retry logic
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+
+    // Now decompress and store in memory
+    kj::ArrayInputStream ais(body);
+    kj::GzipInputStream gzip(ais);
+    auto decompressed = gzip.readAllBytes();
+
+    // Store the decompressed data in memory
+    pyodidePackageManager.setPyodidePackageData(kj::str(path), kj::mv(decompressed));
+  } else {
+    // All retry attempts failed
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+        KJ_LOG(WARNING, "Failed to write package to disk cache", e);
+      }
+    }
+
+    // Now decompress and store in memory
+    kj::ArrayInputStream ais(body);
+    kj::GzipInputStream gzip(ais);
+    auto decompressed = gzip.readAllBytes();
+
+    // Store the decompressed data in memory
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+  co_return;
+}
+
+void fetchPyodidePackages(const PythonConfig& pyConfig,
+    const PyodidePackageManager& pyodidePackageManager,
+    kj::ArrayPtr<kj::String> pythonRequirements,
+    workerd::PythonSnapshotRelease::Reader pythonSnapshotRelease) {
+  auto packagesVersion = pythonSnapshotRelease.getPackages();
+
+  // Get the lock file for this package version
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+  auto pyodideLock = getPyodideLock(pythonSnapshotRelease);
+  if (pyodideLock == kj::none) {
+    KJ_LOG(WARNING, "No lock file found for Python packages version", packagesVersion);
+    return;
+  }
+
+  // Get the filenames of packages to fetch
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+
+    for (const auto& filename: filenames) {
+      // Add each package loading promise to our list
+      promises.add(loadPyodidePackage(
+          pyConfig, pyodidePackageManager, packagesVersion, filename, *client, timer, table));
+    }
+
+    // Wait for all package downloads to complete
+    if (!promises.empty()) {
+      kj::joinPromises(promises.releaseAsArray()).wait(io.waitScope);
```
hoodmane: Are we not allowed to `kj::joinPromises()` an empty list?

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+    // Create a list of promises for each package download
+    kj::Vector<kj::Promise<void>> promises;
+
+    for (const auto& filename: filenames) {
+      // Add each package loading promise to our list
+      promises.add(loadPyodidePackage(
+          pyConfig, pyodidePackageManager, packagesVersion, filename, *client, timer, table));
+    }
+
+    // Wait for all package downloads to complete
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+    auto tlsNetwork = tls->wrapNetwork(network);
+    auto& timer = io.provider->getTimer();
+
+    auto client = kj::newHttpClient(timer, table, network, *tlsNetwork);
+
+    // Create a list of promises for each package download
+    kj::Vector<kj::Promise<void>> promises;
+
+    for (const auto& filename: filenames) {
+      // Add each package loading promise to our list
```
hoodmane:
```suggestion
```

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -603,4 +610,236 @@ kj::String getPythonBundleName(PythonSnapshotRelease::Reader pyodideRelease) {
...
+    options.useSystemTrustStore = true;
+
+    kj::Own<kj::TlsContext> tls = kj::heap<kj::TlsContext>(kj::mv(options));
+    auto& network = io.provider->getNetwork();
+    auto tlsNetwork = tls->wrapNetwork(network);
+    auto& timer = io.provider->getTimer();
+
+    auto client = kj::newHttpClient(timer, table, network, *tlsNetwork);
+
+    // Create a list of promises for each package download
```
hoodmane:
```suggestion
```

hoodmane (approved):
Generally looks good to me, though there are a lot of obvious comments that I think should be removed and a few more other nits.

src/workerd/api/pyodide/pyodide.c++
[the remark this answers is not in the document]
danlapid: The destructor joins synchronously???
I thought it would kill the thread
This sounds bad, it would block the entire event loop until that thread is done?!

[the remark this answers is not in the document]
dom96: I think that's totally fine for workerd, no?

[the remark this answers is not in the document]
danlapid: Not really, not if this suddenly blocks forever for some reason, can we figure out a better way to do this?

Kill the thread, exclusive join its contents with some timeout and async await its join or something else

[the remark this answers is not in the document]
dom96: Why would it suddenly block forever? HttpClient has timeouts afaik and the code has limited retries. If for whatever reason that wasn't the case, then it would also block forever if this was asynchronous, as we cannot run the Worker without its packages being loaded.

The code which loads Pyodide also runs in a thread like this and hasn't had any issues so far.

src/pyodide/python-entrypoint-helper.ts
dom96: Workerd now uses this code path, so we need it for tests.

hoodmane: Thanks.

hoodmane: In other words, I guess, we never use the `test()` functions in edgeworker?

hoodmane: Oh in fact I see now this was just moved. Sorry for the noise.

src/workerd/api/pyodide/pyodide.c++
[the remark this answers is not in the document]
danlapid: The fact that it's done somewhere else doesn't make it right.
AFAICT it's not even needed, why is this even a thread anyway? We could just as well have this be async.
Is it simply a matter of missing a kj::Network& ?
You should be able to pipe that in from server.c++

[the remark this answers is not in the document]
dom96: I think we need to pass the AsyncIoContext through. I just tried this and it's a little bit involved, so after wrangling with it for a while I will separate it out into another PR.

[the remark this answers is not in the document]
danlapid: We don't tend to move the io context around but rather the parts of it that we actually need, do you really need all of it?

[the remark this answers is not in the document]
dom96: I need the `AsyncIoProvider` (to get the network and timer) and the `WaitScope`. So I figure I may as well pass the full AsyncIoContext around.

Anyway, let me split it off and we can discuss in the other PR.

src/workerd/server/tests/python/undefined-handler/server.mjs
```diff
@@ -4,8 +4,7 @@ export default {
   async test(ctrl, env) {
     await rejects(env.pythonWorker.fetch(new Request('http://example.com')), {
       name: 'Error',
-      message:
-        'Python entrypoint "undefined_handler.py" does not export a handler named "on_fetch"',
+      message: 'Handler does not export a fetch() function.',
```
hoodmane: Don't we want to keep this custom message?

dom96: The reason we had that message previously was because we were defining all handlers even if they didn't exist and then throwing this error message from our code. We cannot do this anymore, because if we do so then the validator will assume that all the handlers are defined. This error is now thrown from `ServiceWorkerGlobalScope` which doesn't know whether the worker is written in Python or not.

dom96: Although hmm, I suppose I could check for `IS_WORKERD` and continue to add them all in that case.

PR #4193, opened 2025-05-23
Add addBuiltinModule variant with filter function

jasnell:
Allows us to selectively register built-in modules from a capnp bundle... for instance, to allow us to selectively register certain modules when the experimental flag is enabled.

This can be used, for instance, to allow us to land `node:fs` and `node:http` work-in-progress without that impacting production before we're ready.

### DIFF
--- a/src/workerd/api/rtti.c++
+++ b/src/workerd/api/rtti.c++
@@ -116,6 +116,15 @@ struct EncoderModuleRegistryImpl {
     }
   }
 
+  template <typename Func>
+  void addBuiltinBundleFiltered(jsg::Bundle::Reader bundle, Func filter) {
+    for (auto module: bundle.getModules()) {
+      if (filter(module)) {
+        addBuiltinModule(module);
+      }
+    }
+  }
+
   void addBuiltinModule(jsg::Module::Reader module) {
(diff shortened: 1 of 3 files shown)


src/workerd/api/rtti.c++
```diff
@@ -116,6 +116,15 @@ struct EncoderModuleRegistryImpl {
...
   }
 
+  template <typename Func>
+  void addBuiltinBundleFiltered(jsg::Bundle::Reader bundle, Func filter) {
+    for (auto module: bundle.getModules()) {
+      if (filter(module)) {
+        addBuiltinModule(module);
+      }
+    }
+  }
```
npaun, line 126: Is it intentional that there are two copies of this function? The other is in `modules.h`.

jasnell: Yes. Two different class impls

npaun: Fair nuff but I'm not sure what calls this one because it's in a .c++ file so the caller would need to be there?

npaun (approved):

src/workerd/api/rtti.c++
jasnell: This is part of the type generation mechanism.

PR #4194, opened 2025-05-23
Apply unique_ctx_per_invocation to historical workers too.

kentonv:
See comments for explanation.

### DIFF
--- a/src/workerd/io/compatibility-date.capnp
+++ b/src/workerd/io/compatibility-date.capnp
@@ -662,11 +662,24 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $experimental;
   # Enables delete operations on memory cache if enabled.
 
-  uniqueCtxPerInvocation @74: Bool
+  obsolete74 @74: Bool
       $compatEnableFlag("unique_ctx_per_invocation")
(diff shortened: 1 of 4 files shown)


src/workerd/io/compatibility-date.capnp
```diff
@@ -662,11 +662,24 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+  #
+  # OBSOLETE: We decided to apply this change even to old workers as we've found some old workers
+  #   that assumed this behavior all along, and so this change actually fixes bugs for them.
+  #   It seems unlikely to break anyone because there's no way a Worker could depend on any two
+  #   requests hitting the same isolate, so how could they depend on any two requests having the
+  #   same `ctx` object? At worst they might store some sort of cache on it which becomes
+  #   ineffective, but their Worker would still work, and anyway this would be a very weird thing
+  #   for someone to do.
+
+  reuseCtxAcrossNonclassEvents @92: Bool
```
jasnell, line 679: Nit: not really a fan of the ordinals being out of order but it makes sense here. Might be worth adding a comment at the end of the list so folks don't miss it. Super non-blocking tho since it won't compile if they do.

jasnell (approved):

geelen (approved):

PR #4198, opened 2025-05-23
Implement Approach 2: Container interface with direct Docker API support

petebacondarwin:
- Add DockerClient class for direct Docker API communication via HTTP
- Modify existing Container API to support both RPC and Docker modes
- Add Docker-specific implementations for container lifecycle methods
- Support both production (RPC) and local development (Docker) modes
- Add basic port mapping and TCP connection support for Docker mode

This enables workerd to run containers locally without external services, implementing the Container interface directly with Docker Engine API calls.

### DIFF
--- a/src/workerd/api/BUILD.bazel
+++ b/src/workerd/api/BUILD.bazel
@@ -527,6 +527,7 @@ wd_test(
 )
 
 wd_test(
+    size = "large",
     src = "tests/eventsource-test.wd-test",
     args = ["--experimental"],
     data = ["tests/eventsource-test.js"],
--- a/src/workerd/api/container.h
+++ b/src/workerd/api/container.h
@@ -11,6 +11,8 @@
 #include <workerd/io/io-own.h>
 #include <workerd/jsg/jsg.h>
 
+namespace workerd::io { class DockerClient; }
+
 namespace workerd::api {
 
 class Fetcher;
@@ -21,7 +23,11 @@ class Fetcher;
 // etc.
 class Container: public jsg::Object {
  public:
+  // Constructor for RPC-based containers (production)
   Container(rpc::Container::Client rpcClient, bool running);
+  
+  // Constructor for Docker-based containers (local development)
+  Container(kj::String containerId, kj::String imageTag, io::DockerClient& dockerClient);
 
   struct StartupOptions {
     jsg::Optional<kj::Array<kj::String>> entrypoint;
@@ -60,8 +66,17 @@ class Container: public jsg::Object {
   }
 
  private:
-  IoOwn<rpc::Container::Client> rpcClient;
+  // RPC mode members
+  kj::Maybe<IoOwn<rpc::Container::Client>> rpcClient;
+  
+  // Docker mode members  
+  kj::Maybe<kj::String> containerId;
+  kj::Maybe<kj::String> imageTag;
+  kj::Maybe<io::DockerClient&> dockerClient;
+  kj::HashMap<uint16_t, uint16_t> portMappings; // container -> host port
+  
   bool running;
+  bool isDockerMode = false;
 
   kj::Maybe<jsg::Value> destroyReason;
 
@@ -71,6 +86,8 @@ class Container: public jsg::Object {
 
   class TcpPortWorkerInterface;
   class TcpPortOutgoingFactory;
+  class DockerTcpPortWorkerInterface;
+  class DockerTcpPortOutgoingFactory;
 };
 
 #define EW_CONTAINER_ISOLATE_TYPES api::Container, api::Container::StartupOptions
--- a/src/workerd/api/container.h
+++ b/src/workerd/api/container.h
@@ -11,8 +11,6 @@
 #include <workerd/io/io-own.h>
 #include <workerd/jsg/jsg.h>
 
-namespace workerd::io { class DockerClient; }
-
 namespace workerd::api {
 
 class Fetcher;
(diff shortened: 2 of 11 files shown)


src/workerd/api/container.h
```diff
@@ -21,7 +23,11 @@ class Fetcher;
 // etc.
 class Container: public jsg::Object {
  public:
+  // Constructor for RPC-based containers (production)
   Container(rpc::Container::Client rpcClient, bool running);
+  
+  // Constructor for Docker-based containers (local development)
+  Container(kj::String containerId, kj::String imageTag, io::DockerClient& dockerClient);
```
vicb, line 30: I think Kention mentioned that the generated interface would be the same (even if there is no "transport" when the the server runs in the same process)

src/workerd/io/docker-client.c++
```diff
@@ -0,0 +1,204 @@
...
+  co_await makeDockerRequest(kj::HttpMethod::POST,
+    kj::str("/containers/", containerId, "/kill?signal=", signal));
+}
+
+kj::Promise<void> DockerClient::removeContainer(kj::StringPtr containerId) {
+  co_await makeDockerRequest(kj::HttpMethod::DELETE,
+    kj::str("/containers/", containerId));
+}
+
+kj::Promise<void> DockerClient::buildImage(kj::StringPtr tag, kj::StringPtr context) {
```
vicb, line 118: I don't think we want to handle building images

src/workerd/io/docker-client.c++
```diff
@@ -0,0 +1,204 @@
...
+  co_await makeDockerRequest(kj::HttpMethod::POST,
+    kj::str("/images/create?fromImage=", tag));
+}
+
+kj::Promise<kj::Array<uint16_t>> DockerClient::getExposedPorts(kj::StringPtr imageTag) {
+  try {
+    auto response = co_await makeDockerRequest(kj::HttpMethod::GET,
+      kj::str("/images/", imageTag, "/json"));
+    
+    // Simplified JSON parsing - would need proper JSON parser
```
vicb, line 133: Simplistic?
Don't we have a lib to parse JSON in this repo - maybe ask Claude?

src/workerd/io/docker-client.c++
```diff
@@ -0,0 +1,204 @@
...
+  KJ_IF_SOME(b, body) {
+    auto stream = kj::heap<kj::StringInputStream>(kj::mv(b));
+    request.body = stream.attach(kj::mv(stream));
+  }
+  
+  auto response = co_await request.response;
+  auto responseBody = co_await response.body->readAllText();
+  
+  if (response.statusCode >= 400) {
+    KJ_FAIL_REQUIRE("Docker API error", response.statusCode, responseBody);
```
vicb, line 32: Does that throw?
We probably need a more robust error handling.
I'm not sure if we should look for an existing C++ Docker client or if it makes sense to build one on top of KJ?

vicb: I think the response to my first question is in:

```c++
  try {
    co_await makeDockerRequest(kj::HttpMethod::GET, kj::str("/containers/", containerId, "/json"));
    co_return true;
  } catch (kj::Exception& e) {
    co_return false;
  }
```

kentonv:
Sorry, I think this is the wrong design. I'd prefer that we implement the container.capnp interface in terms of Docker API. This PR directly implements the JS interface in terms of the docker API, skipping the container RPC layer.

FWIW it should be easier (including for Claude) to implement the RPC interface as the code then only needs to consider Cap'n Proto and KJ, not JSG.

petebacondarwin:
> Sorry, I think this is the wrong design. I'd prefer that we implement the container.capnp interface in terms of Docker API. This PR directly implements the JS interface in terms of the docker API, skipping the container RPC layer.
> 
> FWIW it should be easier (including for Claude) to implement the RPC interface as the code then only needs to consider Cap'n Proto and KJ, not JSG.

Yes, I think we all agree with that proposal. If I get a moment I'll ask Claude to rethink the design along those lines.

petebacondarwin:

src/workerd/io/docker-container-client.c++
```diff
@@ -0,0 +1,115 @@
...
+
+  kj::Promise<void> connect(ConnectContext context) override {
+    auto params = context.getParams();
+
+    // Get the downstream ByteStream
+    auto down = params.getDown();
+
+    // TODO: Implement proper ByteStream to AsyncIoStream conversion
+    // For now, this is a placeholder that will need proper implementation
+    KJ_UNIMPLEMENTED("Docker port connection not yet fully implemented");
```
kentonv, line 27: This part is important. `capnp/compat/byte-stream.h` already provides the heavy-lifting.

src/workerd/api/BUILD.bazel
```diff
@@ -527,6 +527,7 @@ wd_test(
 )
 
 wd_test(
+    size = "large",
```
danlapid, line 530: revert?

danlapid:
> @danlapid - the changes to the test config was because (on my MacOS machine) these tests continually timed out (they take ~75s and the timeout was 60s) this was the only way I could find to increase the time out limits. I spent a long time chatting with Claude about this.
> 
> 
> 
> Happy to revert. In fact the better fix would be to change the test that takes forever it is a bit unnecessary as it stands.

I wonder why it takes longer on your machine compared to CI.
You can leave these changes unstaged in your local checkout for your tests to pass but considering it's passing on CI on the macOS machine I'd rather not touch the configuration on master.

petebacondarwin:
Closing as this feature was landed by @anonrig in a completely different way. Thanks for the help Claude!

PR #4199, opened 2025-05-23
R2-3037: Implement `CopyObject` and `UploadPartCopy` for R2 Bindings

helloimalastair:
Also removes `newRuntime` from 2022, as it is no longer necessary.

Hold for R2 release removing `newRuntime` checks.


src/workerd/api/r2-test.js
```diff
@@ -674,6 +760,38 @@ export default {
...
+              uploadedBefore: new Date('0'),
+            },
+          }
+        ),
+        null
+      );
+      const multipart = await env.BUCKET.createMultipartUpload(
+        'onlyIfStrongEtagMultipart'
+      );
+      assert.rejects(
```
npaun: The value returned from `rejects` needs to be awaited: https://nodejs.org/api/assert.html#assertrejectsasyncfn-error-message

src/workerd/api/r2-bucket.h
```diff
@@ -492,6 +522,9 @@ class R2Bucket: public jsg::Object {
         options?: R2PutOptions & { onlyIf: R2BucketConditional | Headers }
       ): Promise<R2Object | null>;
       put(key: string, value: ReadableStream | ArrayBuffer | ArrayBufferView | string | null | Blob, options?: R2PutOptions): Promise<R2Object>;
+
+      copy(key: string, source: R2CopySource, options: R2CopyOptions & { onlyIf: R2BucketConditional | Headers }): Promise<R2Object | null>;
```
bthwaites, line 526: Where does R2BucketConditional come from?

helloimalastair: I'm not entirely sure how, but it is converted to `R2Conditional` when the types are built. I've asked the team to make sure I understand

PR #4200, opened 2025-05-24
Implement fallback service support for new module registry impl

jasnell:
Adds fallback service support, with a new POST based implementation for the fallback service,
Adds extension module support,
Enables registering the module registry for service worker syntax so things like the internal use of `node:util` inspect can be used for improved console output locally just like when using ESM syntax workers.
Moves the fallback service implementation out of server.c++ to a separate compilation unit.

### DIFF
--- a/src/workerd/server/fallback-service.c++
+++ b/src/workerd/server/fallback-service.c++
@@ -172,17 +172,7 @@ ModuleOrRedirect tryResolveV2(ImportType type,
   capnp::JsonCodec json;
   capnp::MallocMessageBuilder moduleMessage;
   auto request = moduleMessage.initRoot<server::config::FallbackServiceRequest>();
-  switch (type) {
-    case ImportType::IMPORT:
-      request.setType("import");
-      break;
-    case ImportType::REQUIRE:
-      request.setType("require");
-      break;
-    case ImportType::INTERNAL:
-      request.setType("internal");
-      break;
-  }
+  request.setType(getMethodFromType(type));
   request.setSpecifier(specifier);
   request.setReferrer(referrer);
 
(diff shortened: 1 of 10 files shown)


src/workerd/server/fallback-service.c++
```diff
@@ -0,0 +1,270 @@
...
+      request.setType("import");
+      break;
+    case ImportType::REQUIRE:
+      request.setType("require");
+      break;
+    case ImportType::INTERNAL:
+      request.setType("internal");
+      break;
+  }
+  request.setSpecifier(specifier);
```
anonrig, line 186: You can use getModuleFromType and simplify this switch/case statement

src/workerd/server/fallback-service.c++
```diff
@@ -0,0 +1,270 @@
...
+  request.setReferrer(referrer);
+
+  if (rawSpecifier != nullptr) {
+    request.setRawSpecifier(rawSpecifier);
+  }
+
+  if (attributes.size() > 0) {
+    auto attrs = request.initAttributes(attributes.size());
+    int n = 0;
```
anonrig, line 195:
```suggestion
    size_t n = 0;
```

src/workerd/server/fallback-service.c++
```diff
@@ -0,0 +1,270 @@
...
+}  // namespace
+
+ModuleOrRedirect tryResolve(Version version,
+    ImportType type,
+    kj::StringPtr address,
+    kj::StringPtr specifier,
+    kj::StringPtr rawSpecifier,
+    kj::StringPtr referrer,
+    const kj::HashMap<kj::StringPtr, kj::StringPtr>& attributes) {
+  return version == Version::V1
```
anonrig, line 265: If you make this type template argument, you can make this branch constexpr

jasnell: It's also worth noting that this is used only in local-dev and then only for certain types of operations. Perf optimizations here are likely to be low-impact.

anonrig (approved):

PR #4201, opened 2025-05-24
[o11y] Expand Streaming Tail Worker support

fhanau:
[o11y] Provide scriptId in Streaming Tail Worker onset event
This is not exposed in the JS API, but still needed to support source maps.

[o11y] Expose executionModel in STW API

[o11y] Update Return event API definition


jasnell (approved):

types/generated-snapshot/experimental/index.ts
```diff
@@ -6524,7 +6525,7 @@ export declare namespace TailStream {
   }
   interface Return {
     readonly type: "return";
-    readonly info?: FetchResponseInfo | Attributes;
```
jmorrell-cloudflare, line 6527: Thank you!

PR #4203, opened 2025-05-26
WPT test harness fixes

npaun:
See the individual commits:

- Wrap test code in an IIFE to prevent `var` declarations from messing the global scope. Some tests intentionally modify it so you can set `global: true` as a test option.
- `promise_test` needs to support `t.add_cleanup()` functions, even though it's mixing different styles of tests.
- This fixes a bunch of tests that were wrongly failed either due to trashed global scope / a lack of cleanup / or both in many cases.
- While I'm at it, commit the nice colored console info/warn/error messages


src/wpt/streams-test.ts
```diff
@@ -781,36 +712,9 @@ export default {
...
-      'testing "{{bogusPartial}}}" (length 1)',
-    ],
-  },
+  'transform-streams/lipfuzz.any.js': {},
   'transform-streams/patched-global.any.js': {
-    comment: 'To be investigated',
-    expectedFailures: [
-      'TransformStream should use the original value of ReadableStream and WritableStream',
-    ],
+    global: true,
```
anonrig, line 717: Can we choose a little bit more verbose key rather than global? It's not really explanatory.

PR #4204, opened 2025-05-27
Move vendor dir in Python search path just before site-packages.

dom96:
We have two functions that adjust the sys path in Python Workers. The first one (adjustSysPath) only runs when not restoring from a snapshot, so it is effectively immortalised in the memory snapshots. To avoid needing to regenerate all the memory snapshots (which for package snapshots would be a problem) we just rearrange the path at the same time as when we add the vendor path to the sys path.

### Test Plan

```
$ bazel run @workerd//src/workerd/server/tests/python:vendor_dir_development@

Python sys.path: ['/session', '/lib/python312.zip', '/lib/python3.12', '/lib/python3.12/lib-dynload', '/session/metadata', '/session/metadata/vendor', '/lib/python3.12/site-packages', '/session/lib/python3.12/site-packages']
```

### DIFF
--- a/src/pyodide/internal/python.ts
+++ b/src/pyodide/internal/python.ts
@@ -50,6 +50,8 @@ function prepareWasmLinearMemory(Module: Module): void {
     // The effects of these are purely in Python state so they only need to be run
     // if we didn't restore a snapshot.
     entropyBeforeTopLevel(Module);
+    // Note that setupPythonSearchPath runs after adjustSysPath and rearranges where
+    // the /session/metadata path is added.
     adjustSysPath(Module);
   }
   if (Module.API.version !== '0.26.0a2') {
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/python.ts
```diff
@@ -65,8 +67,30 @@ function maybeAddVendorDirectoryToPath(pyodide: Pyodide): void {
...
+      # import will pick a module from the vendor path rather than the local path. We've got a test
+      # that reproduces this (vendor_dir).
+      for i, path in enumerate(sys.path):
+        if 'site-packages' in path:
+          sys.path.insert(i, VENDOR_PATH)
+          sys.path.insert(i, SESSION_PATH)
+          break
+      else:
+        # If no site-packages found, fail
+        raise ValueError("No site-packages found in sys.path")
```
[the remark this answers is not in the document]
dom96: This is intended, we want to fail loudly if this happens.

src/pyodide/internal/python.ts
```diff
@@ -65,8 +67,30 @@ function maybeAddVendorDirectoryToPath(pyodide: Pyodide): void {
...
+      # We want to insert before '/lib/python3.12/site-packages' and other site-packages
+      #
+      # We also need the session path to be before the vendor path, if we don't do so then a local
+      # import will pick a module from the vendor path rather than the local path. We've got a test
+      # that reproduces this (vendor_dir).
+      for i, path in enumerate(sys.path):
+        if 'site-packages' in path:
+          sys.path.insert(i, VENDOR_PATH)
+          sys.path.insert(i, SESSION_PATH)
+          break
```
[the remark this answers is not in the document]
dom96: This is checked above.

src/pyodide/internal/python.ts
```diff
@@ -65,8 +67,30 @@ function maybeAddVendorDirectoryToPath(pyodide: Pyodide): void {
       from pathlib import Path
 
       VENDOR_PATH = "/session/metadata/vendor"
-      if Path(VENDOR_PATH).is_dir():
```
hoodmane, line 67: The name of this function is no longer accurate after this change.

src/pyodide/internal/python.ts
```diff
@@ -51,6 +51,8 @@ function prepareWasmLinearMemory(Module: Module): void {
     // The effects of these are purely in Python state so they only need to be run
     // if we didn't restore a snapshot.
     entropyBeforeTopLevel(Module);
+    // Note that maybeAddVendorDirectoryToPath runs after adjustSysPath and rearranges where
+    // the /session/metadata path is added.
     adjustSysPath(Module);
```
hoodmane: Should we only call this here in version 0.26.0a2 and in other cases leave all the path changes to (renamed) `maybeAddVendorDirectoryToPath()`?

dom96: Okay, I've rearranged things a bit. Hoping that doesn't break it.

src/pyodide/internal/python.ts
```diff
@@ -50,22 +50,50 @@ function prepareWasmLinearMemory(Module: Module): void {
...
+      # System paths are typically: ['/session', '/lib/python312.zip', '/lib/python3.12', '/lib/python3.12/lib-dynload']
+      # We want to insert before '/lib/python3.12/site-packages' and other site-packages
+      #
+      # We also need the session path to be before the vendor path, if we don't do so then a local
+      # import will pick a module from the vendor path rather than the local path. We've got a test
+      # that reproduces this (vendor_dir).
+      for i, path in enumerate(sys.path):
+        if 'site-packages' in path:
+          sys.path.insert(i, VENDOR_PATH)
+          sys.path.insert(i, SESSION_PATH)
```
hoodmane: Nit: I'd suggest:
```suggestion
          sys.path[i:i] = [SESSION_PATH, VENDOR_PATH]
```
which is a bit clearer.

hoodmane (approved):
Thanks!

dom96:
@hoodmane Hmm, looks like `adjustSysPath` needs to be called before the memory snapshot is collected (because site-packages need to be in the path for the package imports to work). I just reverted to its original. I guess proper thing to do is to add the site-packages path before memory snapshot is collected, but not the vendor dir. Though I'm not sure there is much point in doing this only in 0.27.7+.

PR #4205, opened 2025-05-27
Implement node:http and node:https module

anonrig:
- Adds ClientRequest, OutgoingMessage and IncomingMessage classes.
- Adds request and get methods.

TODO

- [x] Make a decision about removing/keeping `_http_*` prefixed modules
- [x] Add tests for node:https
- [x] Add more tests for timeout cases

### DIFF
--- a/src/node/internal/internal_http_outgoing.ts
+++ b/src/node/internal/internal_http_outgoing.ts
@@ -57,7 +57,7 @@ export class OutgoingMessage extends Writable {
     return this;
   }
 
-  setHeaders(headers: HeadersInit): this {
+  setHeaders(headers: HeadersInit | null | undefined): this {
     if (this[kHeadersSent]) {
       throw new ERR_HTTP_HEADERS_SENT('set');
     }


src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (this.method !== 'GET' && this.method !== 'HEAD') {
+      body = new Blob(this.#body, {
+        type: this.getHeader('content-type') ?? '',
+      });
+    }
+
+    const headers: [string, string][] = [];
+    for (const name of Object.keys(this)) {
+      const value = this[kOutHeaders][name]?.value;
+      if (value) {
```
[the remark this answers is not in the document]
anonrig: Well it caught a bug on the day I complained about it. What are the chances... @danlapid

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,372 @@
...
+      }
+    }
+
+    const defaultPort = options.defaultPort || 80;
+
+    const optsWithoutSignal = { __proto__: null, ...options };
+
+    const port = (optsWithoutSignal.port = options.port || defaultPort || 80);
+    const host = (optsWithoutSignal.host =
+      validateHost(options.hostname, 'hostname') ||
```
npaun: I feel like this is doing too much. I think it'd be easier to read if the assignments were separated out of this, even if it would be more verbose.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,372 @@
...
+          for (const item of value) {
+            headers.push([name, item as string]);
+          }
+        } else {
+          headers.push([name, value]);
+        }
+      }
+    }
+
+    const header = this.getHeader('host');
```
npaun:
```suggestion
    const host = this.getHeader('host');
```

I was really confused when I saw `http://${header}` later.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,372 @@
...
+    Writable.prototype.end.call(
+      this,
+      data,
+      encoding as BufferEncoding,
+      callback
+    );
+    return this;
+  }
+
+  #resetTimers(done: boolean): void {
```
npaun: nit: Unless this function is matching Node's implementation, I'd suggest using an object like `{done: boolean}` so the call sites are easier to understand. I don't really like bool parameters

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (options.path) {
+      const path = String(options.path);
+      if (INVALID_PATH_REGEX.test(path)) {
+        throw new ERR_UNESCAPED_CHARACTERS('Request path');
+      }
+    }
+
+    const defaultPort = options.defaultPort || 80;
+
+    const optsWithoutSignal = { __proto__: null, ...options };
```
npaun: I'm not quite sure what this variable is for. Debugging?

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    if (options.joinDuplicateHeaders !== undefined) {
+      validateBoolean(
+        options.joinDuplicateHeaders,
+        'options.joinDuplicateHeaders'
+      );
+    }
+
+    this.joinDuplicateHeaders = Boolean(options.joinDuplicateHeaders);
```
npaun: Not yet implemented, right?

anonrig: Yes, not yet implemented.

src/node/internal/internal_http_client.ts
anonrig: This is directly taken from Node.js code. I recommend keeping as it is to easily trace/compare with node.js implementation in the future.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,375 @@
...
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.insecureHTTPParser');
+    }
+
+    // TODO(soon): Support this option.
+    if (options.joinDuplicateHeaders !== undefined) {
+      validateBoolean(
+        options.joinDuplicateHeaders,
+        'options.joinDuplicateHeaders'
+      );
+    }
```
dom96, line 213: Should you throw here since this option isn't implemented yet?

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,375 @@
...
+      }
+    }
+
+    const defaultPort = options.defaultPort || 80;
+
+    const port = (options.port = options.port || defaultPort || 80);
+    const host = (options.host =
+      validateHost(options.hostname, 'hostname') ||
+      validateHost(options.host, 'host') ||
+      'localhost');
```
dom96: I guess Node does this too, but I'm curious why `options.host` and `options.port` is assigned to here.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,375 @@
...
+          }
+        } else {
+          headers.push([name, value]);
+        }
+      }
+    }
+
+    const host = this.getHeader('host');
+    if (!host) {
+      // TODO: Fix this.
```
dom96: Fix what?

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,375 @@
...
+      this.#timer = null;
+    } else if (this.timeout) {
+      this.#timer = setTimeout(() => {
+        this.emit('timeout');
+      }, this.timeout) as unknown as number;
+    }
+  }
+
+  public flushHeaders(): void {
+    // Not implemented
```
dom96: Should this throw?

src/node/internal/internal_http_client.ts
dom96: Agreed on the above to make the meaning of the bool clearer at the call site.

src/node/http.ts
```diff
@@ -8,11 +8,45 @@ import {
...
+): ClientRequest {
+  return new ClientRequest(url, options, cb);
+}
+
+export function get(
+  url: string | URL,
+  options: RequestOptions,
+  cb: (err: Error | null) => void
+): ClientRequest {
+  const req = request(url, options, cb);
```
jasnell, line 32: This should override the `method` property in the `options` to for it to be `GET`.

src/node/http.ts
```diff
@@ -8,11 +8,45 @@ import {
...
 
-export { validateHeaderName, validateHeaderValue, METHODS, STATUS_CODES };
+import type { RequestOptions } from 'node:http';
+
+const { WebSocket } = globalThis;
+
+export function request(
+  url: string | URL,
+  options: RequestOptions,
+  cb: (err: Error | null) => void
```
jasnell: This signature is incorrect. The callback passed in to request/get does not follow the node.js standard pattern. The first argument is the response object, not an error.

src/node/internal/internal_http_client.ts
jasnell: No, I wouldn't expect it to.

src/node/internal/internal_http_client.ts
anonrig: I'm following Node.js implementation and intentionally not touching them to make maintenance easier.

src/node/internal/internal_http_client.ts
jasnell: The underlying fetch implementation is going to handle the headers in the standard defined way regardless of this setting. I would probably just ignore this header entirely if the value is `false` (which is the default) and throw only if the value is `true`

anonrig: I'll update according to @jasnell's comments.

jasnell (approved):

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,383 @@
...
+      validateBoolean(
+        options.joinDuplicateHeaders,
+        'options.joinDuplicateHeaders'
+      );
+
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-boolean-literal-compare
+      if (options.joinDuplicateHeaders === true) {
+        // We do not support joinDuplicateHeaders because the underlying
+        // implementation based on fetch() will handle duplicate headers.
+        throw new ERR_OPTION_NOT_IMPLEMENTED('options.joinDuplicateHeaders');
```
guybedford: We should probably also have option not implemented for (and please correct as I'm missing understanding):

* `agent`
* `createConnection`
* `lookup`
* `insecureHTTPParser`
* `socketPath`

At the very least this would also help to understand the implementation limitations as well.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,383 @@
...
+
+    const headers = options.headers;
+    if (!Array.isArray(headers)) {
+      if (headers != null) {
+        for (const [key, value] of Object.entries(headers)) {
+          this.setHeader(key, value as unknown as string);
+        }
+      }
+
+      if (host && !this.getHeader('host') && setHost) {
```
guybedford, line 231: It seems like it should be straightforward to add `setDefaultHeaders` as another guard here? Alternatively a TODO comment or similar could be useful.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,383 @@
...
+    url.pathname = this.path;
+
+    fetch(url, {
+      method: this.method,
+      headers,
+      body: body ?? null,
+      signal: this.#abortController.signal,
+    })
+      .then(this.#handleFetchResponse.bind(this))
+      .catch(this.#handleFetchError.bind(this));
```
guybedford, line 360: HTTP on fetch has two major design limitations:

1. Content decoding is handled automatically by fetch, but the expectation is that it is not handled in http
2. Redirects are always followed

It would be nice to see some notes about how these are being addressed, for example, if we want to add a custom option to disable content decoding, or if we want to add a warning when a request resulted in a second request to a redirected URL for example (or possibly even recreating the redirect from the followed URL, or possibly even customizing fetch with cf option to not follow redirects?).

src/workerd/api/node/tests/http-client-nodejs-test.js
```diff
@@ -0,0 +1,206 @@
...
+      });
+      res.resume();
+      resolve();
+    });
+    await promise;
+  },
+};
+
+// Test is taken from test/parallel/test-http-contentLength0.js
+export const testHttpContentLength0 = {
```
guybedford, line 148: It would be great to see a content length + compressed content case.

src/node/internal/internal_http_client.ts
anonrig: This won't follow the nodejs implementation then? https://github.com/nodejs/node/blob/main/lib/_http_client.js#L306

src/node/internal/internal_http_client.ts
anonrig: I've added a TODO for now, will address them before removing experimental from this.

src/workerd/api/node/tests/http-client-nodejs-test.js
anonrig: Is there an existing node.js test for this?

src/node/internal/internal_http_agent.ts
```diff
@@ -0,0 +1,99 @@
...
+    return false;
+  }
+
+  public reuseSocket(): void {
+    // Not implemented
+  }
+
+  public destroy(): void {
+    // Not implemented
+  }
```
guybedford: Which of these are expected to have affects, versus which are okay to just ignore?

`destroy()` might have expectations around stream writes failing, `createSocket` would expect a defined return type.

So these might be cases where just outright throwing makes sense. Or maybe destroy does make sense as a no-op.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,420 @@
...
+      }
+      method = this.method = method.toUpperCase();
+    } else {
+      method = this.method = 'GET';
+    }
+
+    const maxHeaderSize = options.maxHeaderSize;
+    if (maxHeaderSize !== undefined) {
+      validateInteger(maxHeaderSize, 'maxHeaderSize', 0);
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.maxHeaderSize');
```
guybedford, line 176: `maxHeaderSize` seems like it could be okay to ignore for compatibility.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    if (options.insecureHTTPParser !== undefined) {
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.insecureHTTPParser');
```
guybedford, line 184: Similarly, this might make sense to ignore for compatibility?

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    if (options.agent !== undefined) {
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.agent');
```
guybedford: Haven't you implemented a simple agent wrapper now?

src/node/internal/internal_http_incoming.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  // @ts-expect-error TS2416 Type inconsistency
+  public get headersDistinct(): Record<string, string> {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument,@typescript-eslint/no-explicit-any
+    return Object.fromEntries(this.#response.headers as any);
+  }
+
+  // @ts-expect-error TS2416 Type inconsistency
+  public get trailers(): Record<string, unknown> {
+    // Not supported.
```
guybedford, line 129: It might make sense to throw unsupported here?

src/workerd/api/node/tests/http-client-nodejs-test.js
guybedford: I'm not sure - perhaps under zlib tests?

src/workerd/api/node/tests/http-client-server.js
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2017-2022 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
```
guybedford, line 3: Maybe add a comment to note that this is running in Node.js, confused me at first not seeing `createServer` implemented.

src/node/internal/internal_http_client.ts
anonrig: Yes I need to update this

src/node/internal/internal_http_incoming.ts
anonrig: This is a getter. If we throw an error it might be too much destruptive. Even on console.log(req), it will throw.

src/node/internal/internal_http_agent.ts
anonrig: I updated the signatures. I think they're all safe to ignore for now, since our implementation does not call any of the Agent methods.

src/node/internal/internal_http_client.ts
anonrig: We can always remove these errors if they're too disruptive but we can not readd them due to backward compatibility limitations.

src/node/internal/internal_http_client.ts
anonrig: We can always remove these errors if they're too disruptive but we can not readd them due to backward compatibility limitations.

src/node/https.ts
```diff
@@ -0,0 +1,51 @@
...
+//     https://opensource.org/licenses/Apache-2.0
+// Copyright Joyent and Node contributors. All rights reserved. MIT license.
+
+import { urlToHttpOptions, isURL } from 'node-internal:internal_url';
+import { ClientRequest } from 'node-internal:internal_http_client';
+import { Agent, globalAgent } from 'node-internal:internal_https_agent';
+import type { IncomingMessageCallback } from 'node-internal:internal_http_util';
+import type { RequestOptions } from 'node:http';
+
+export function request(...args: unknown[]): ClientRequest {
```
jasnell, line 17: Nit: should probably document the actual non-generic signatures as function decls above this.

src/node/https.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (args.length > 0 && typeof args[0] !== 'function') {
+    Object.assign(options, args.shift());
+  }
+
+  // @ts-expect-error TS2740 We don't fully implement all Agent attributes.
+  options._defaultAgent = globalAgent;
+  args.unshift(options);
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  return Reflect.construct(ClientRequest, args);
```
jasnell: `Reflect.construct` can be a bit slow compared to just using `new ClientRequest(...)`, any particular reason not to call a regular constructor?

src/node/internal/internal_http_agent.ts
```diff
@@ -0,0 +1,121 @@
...
+import { validateOneOf, validateNumber } from 'node-internal:validators';
+import { ERR_METHOD_NOT_IMPLEMENTED } from 'node-internal:internal_errors';
+import type {
+  RequestOptions,
+  Agent as _Agent,
+  AgentOptions,
+  ClientRequest,
+  IncomingMessage,
+} from 'node:http';
+import type { Socket } from 'node:net';
```
jasnell, line 16: This could probably use some explanatory comments explaining that we aren't really intending to support the `Agent` API right now beyond providing a very limited stub impl

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,441 @@
...
+} from 'node:http';
+import { IncomingMessage } from 'node-internal:internal_http_incoming';
+import { OutgoingMessage } from 'node-internal:internal_http_outgoing';
+import { Agent, globalAgent } from 'node-internal:internal_http_agent';
+import type { IncomingMessageCallback } from 'node-internal:internal_http_util';
+
+const INVALID_PATH_REGEX = /[^\u0021-\u00ff]/;
+
+function validateHost(host: unknown, name: string): string {
+  if (host !== null && host !== undefined && typeof host !== 'string') {
```
jasnell:
```suggestion
  if (host != null && typeof host !== 'string') {
```

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,441 @@
...
+      throw new ERR_INVALID_ARG_TYPE(
+        'options.agent',
+        ['Agent-like Object', 'undefined', 'false'],
+        agent
+      );
+    }
+    this.agent = agent as unknown as Agent | undefined;
+
+    const protocol = options.protocol || (defaultAgent as Agent).protocol;
+    let expectedProtocol = (defaultAgent as Agent).protocol;
```
jasnell:
```suggestion
    let expectedProtocol = (defaultAgent as Agent).protocol;
    const protocol = options. Protocol || expectedProtocol;
```

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,441 @@
...
+      typeof agent === 'object' &&
+      typeof agent.addRequest !== 'function'
+    ) {
+      throw new ERR_INVALID_ARG_TYPE(
+        'options.agent',
+        ['Agent-like Object', 'undefined', 'false'],
+        agent
+      );
+    }
+    this.agent = agent as unknown as Agent | undefined;
```
jasnell: Should that be `|| undefined`?

As is this will cause `this.agent` to be equal to 0.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,441 @@
...
+
+    const protocol = options.protocol || (defaultAgent as Agent).protocol;
+    let expectedProtocol = (defaultAgent as Agent).protocol;
+    if (this.agent?.protocol) expectedProtocol = this.agent.protocol;
+    const defaultPort = options.defaultPort || options.defaultPort || 80;
```
jasnell: Why is `options.defaultPort` duplicated here?

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,441 @@
...
+      validateInteger(maxHeaderSize, 'maxHeaderSize', 0);
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.maxHeaderSize');
+    }
+
+    if (options.insecureHTTPParser !== undefined) {
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.insecureHTTPParser');
+    }
+
+    if (options.createConnection !== undefined) {
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.createConnection');
```
jasnell, line 191: We should probably at least be validating these are the correct type.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,441 @@
...
+    chunk: Buffer,
+    _encoding: BufferEncoding,
+    callback: VoidFunction
+  ): void {
+    this.#body.push(chunk);
+    callback();
+  }
+
+  public setNoDelay(_noDelay?: boolean): void {
+    // Not implemented
```
jasnell, line 408: We should at least be validating the types.

src/node/internal/internal_http_client.ts
anonrig: Agreed.

vicb:
I would love to see a PR description similar to @guybedford 's #4271 here.

It is helpful both to [document the runtime](developers.cloudflare.com/workers/runtime-apis/nodejs) and update the unenv preset.

Thanks!

src/node/internal/internal_http_outgoing.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    return this;
+  }
+
+  setHeaders(headers: HeadersInit): this {
+    if (this[kHeadersSent]) {
+      throw new ERR_HTTP_HEADERS_SENT('set');
+    }
+
+    if (
```
jasnell, line 65: This does not appear to handle the case where `headers` is `null` correctly

src/node/internal/internal_http_outgoing.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  removeHeader(name: string): void {
+    validateString(name, 'name');
+
+    if (this[kHeadersSent]) {
+      throw new ERR_HTTP_HEADERS_SENT('remove');
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+    delete this[kOutHeaders][name.toLowerCase()];
```
jasnell, line 151: Is Node.js using `delete` for this? Generally we've tried to avoid use of the `delete` keyword because it puts the object immediately into a slow mode within v8. If it's what Node.js is doing, no worries, we can keep it, just curious.

src/node/internal/internal_http_outgoing.ts
```diff
@@ -0,0 +1,185 @@
...
+    // Not implemented
+  }
+
+  get headersSent(): boolean {
+    return this[kHeadersSent];
+  }
+
+  // @ts-expect-error TS2416 Unnecessary type assertion
+  pipe(): void {
+    this.emit('error', new ERR_STREAM_CANNOT_PIPE());
```
jasnell, line 176: This should probably emit the event using `process.nextTick()` to catch the case where the event handler is attached after pipe is called.

src/node/internal/internal_http_incoming.ts
```diff
@@ -0,0 +1,132 @@
+import type { IncomingMessage as _IncomingMessage } from 'node:http';
```
jasnell, line 6: Nit: these all need copyright headers

src/node/internal/internal_http_incoming.ts
```diff
@@ -0,0 +1,132 @@
...
+  httpVersionMinor = 1;
+  httpVersion: string = '1.1';
+
+  // TODO(soon): Get rid of the second argument.
+  constructor(
+    response: Response,
+    resetTimers: (opts: { finished: boolean }) => void
+  ) {
+    if (!(response instanceof Response)) {
+      // TODO(soon): Fully support IncomingMessage constructor
```
jasnell: Can you expand this comment a bit to describe what bits are missing

src/node/internal/internal_http_incoming.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    if (this._readableState) {
+      this._readableState.readingMore = true;
+    }
+
+    this.url = response.url;
+    this.statusCode = response.status;
+    this.statusMessage = response.statusText;
+
+    this.on('end', () => {
```
jasnell: once?

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,454 @@
...
+import { OutgoingMessage } from 'node-internal:internal_http_outgoing';
+import { Agent, globalAgent } from 'node-internal:internal_http_agent';
+import type { IncomingMessageCallback } from 'node-internal:internal_http_util';
+
+const INVALID_PATH_REGEX = /[^\u0021-\u00ff]/;
+
+function validateHost(host: unknown, name: string): string {
+  if (host != null && typeof host !== 'string') {
+    throw new ERR_INVALID_ARG_TYPE(
+      `options.${name}`,
```
jasnell, line 45: This is going to throw if `name` happens to be a `Symbol`. We should probably guard against that.

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,454 @@
...
+      }
+      method = this.method = method.toUpperCase();
+    } else {
+      method = this.method = 'GET';
+    }
+
+    const maxHeaderSize = options.maxHeaderSize;
+    if (maxHeaderSize !== undefined) {
+      validateInteger(maxHeaderSize, 'maxHeaderSize', 0);
+      throw new ERR_OPTION_NOT_IMPLEMENTED('options.maxHeaderSize');
```
jasnell, line 176: It's worth adding comments here about whether these should remain unimplemented and why

src/node/internal/internal_http_client.ts
```diff
@@ -0,0 +1,454 @@
...
+
+    this[kHeadersSent] = true;
+
+    // HTTP on fetch has two major design limitations.
+    //
+    // 1. Content decoding is handled automatically by fetch, but expectation is that it's not handled in http.
+    // 2. Redirects are always followed.
+    //
+    // TODO(soon): Address these concerns and limitations.
+    fetch(url, {
```
jasnell, line 353: We should document that nothing is directly waiting for the fetch promise here. It's up to the user of the http API to arrange for the request to be held open until the fetch completes, typically by passing some promise to `ctx.waitUntil(...)` and resolving that promise when the http request is complete.

src/node/internal/internal_http_outgoing.ts
anonrig: Yes Node.js is using delete as well. This is straight from Node.js code - https://github.com/nodejs/node/blob/c08a1d152b758a1b97b81e9edce6ed60faaf4063/lib/_http_outgoing.js#L812

src/node/internal/internal_http_outgoing.ts
anonrig: This is using the same exact code as Node.js - https://github.com/nodejs/node/blob/c08a1d152b758a1b97b81e9edce6ed60faaf4063/lib/_http_outgoing.js#L1189

src/node/internal/internal_http_incoming.ts
anonrig: I've updated. I'm leaving as not resolved so you can read the actual change and let's make sure you're ok with it.

src/node/internal/internal_http_incoming.ts
anonrig: yeah true.

src/workerd/api/node/tests/http-client-nodejs-test.js
```diff
@@ -0,0 +1,309 @@
...
+          port: env.ASD_SERVER_PORT,
+        },
+        (res) => {
+          strictEqual(res.destroyed, false);
+          res.destroy();
+          strictEqual(res.destroyed, true);
+          res.on('close', resolve);
+        }
+      );
+      await promise;
```
jasnell, line 56: Just noting, it's going to be critical that we document this pattern on adding a promise on the close event that is awaiting in the fetch handler, either directly like this or via the `ctx.waitUntil(...)`. Keep in mind, however, that the `'error'` event is emitted after the `'close'` event, so it's at least theoretically possible that resolving on `close` could cause the error to not be handled. We should test that case if there's not a test in here already.

src/workerd/api/node/tests/http-client-nodejs-test.js
```diff
@@ -0,0 +1,309 @@
...
+            })
+            .resume();
+        }
+      );
+      await promise;
+    }
+  },
+};
+
+// Test is taken from test/parallel/test-http-client-response-timeout.js
```
jasnell, line 81: Why is this test commented out? If we don't currently support it, a TODO comment would be helpful.

src/workerd/api/node/tests/http-nodejs-test.js
```diff
@@ -192,3 +192,39 @@ export const testHttpMethods = {
...
+  async test() {
+    const checkIsHttpToken = httpCommon._checkIsHttpToken;
+    const checkInvalidHeaderChar = httpCommon._checkInvalidHeaderChar;
+
+    // checkIsHttpToken
+    ok(checkIsHttpToken('t'));
+    ok(checkIsHttpToken('tt'));
+    ok(checkIsHttpToken('ttt'));
+    ok(checkIsHttpToken('tttt'));
+    ok(checkIsHttpToken('ttttt'));
```
jasnell, line 208: are we sure this works with actual known http tokens we'd actually use ;-)

src/node/internal/internal_http_client.ts
anonrig: This is taken straight from the node.js implementation - https://github.com/nodejs/node/blob/c08a1d152b758a1b97b81e9edce6ed60faaf4063/lib/_http_client.js#L121

src/workerd/api/node/tests/http-agent-nodejs-test.js
```diff
@@ -0,0 +1,48 @@
+import http from 'node:http';
```
jasnell, line 5: Copyright headers... we're not very consistent on this... I forget them all the time... we probably need a lint rule for it

src/workerd/api/node/tests/http-nodejs-test.js
anonrig: I can add more tests but this is definitely taken from node.js - https://github.com/nodejs/node/blob/c514e8f781b2acedb6a2b42208d8f8f4d8392f09/test/parallel/test-http-common.js, which doesn't have a good test coverage...

src/node/internal/internal_http_client.ts
jasnell: We should fix that in Node.js then too :-)

src/workerd/api/node/tests/http-nodejs-test.js
anonrig: I've added 2 more values...

src/node/internal/internal_http_client.ts
anonrig: Thanks!

anonrig:
Rebased, force pushed and added last 2 commits to address the reviews.

anonrig:
> I would love to see a PR description similar to @guybedford 's #4271 here.
> 
> It is helpful both to [document the runtime](developers.cloudflare.com/workers/runtime-apis/nodejs) and update the unenv preset.
> 
> Thanks!

@vicb This pull-request only enables these methods behind an experimental flag. I'll update the PR that removes the experimental flag so that we can track the changes easier.

Issue #4155, opened 2025-05-16
Using an `AsyncLocalStorage` snapshot across requests


vicb:
Repro: https://github.com/vicb/als-snapshot

When an `AsyncLocalStorage.snapshot()` is used in a different request as the one it was created in, the following error is generated:

```text
[wrangler:inf] GET / 302 Found (6ms)
using snapshot
[wrangler:err] Error: Cannot perform I/O on behalf of a different request. I/O objects (such as streams, request/response bodies, and others) created in the context of one request handler cannot be accessed from a different request's handler. This is a limitation of Cloudflare Workers which allows us to improve overall performance. (I/O type: SpanParent)
    at Object.apply ([...]/strip-cf-connecting-ip-header.js:9:18)
    at null.<anonymous> ([...]/src/index.ts:31:39)
    at Object.fetch ([...]/src/index.ts:31:12)
    at fetchDispatcher ([...]/middleware-loader.entry.ts:54:17)
```

The issue happens when using composable cache in Next with the cloudflare adapter, see https://github.com/opennextjs/opennextjs-cloudflare/pull/666 - Thanks @conico974 for figuring the issue out and submitting a patch.

I discussed this offline with @jasnell and he asked me to create the issue.

Not sure if this is something that we can fix or working as intended.

jasnell:
The Key issue here, I believe, is that the async context is capturing a span context that is specific to the original request, so when we reenter that context from a different request and the internals grab that span context, things go boom. I do think it's possible to fix but we need to carefully consider how to avoid other potential issues.

I wouldn't necessarily label this a bug because it's actually working as intended, it's just surprising behavior that users won't expect and really can't do anything about, so addressing this is a matter of improving DX as opposed to fixing a bug.

jasnell:
We've "fixed" this by making ALS snapshots specific to the individual request in which they were created. It means the error that you get in this case is more specific to the issue rather than the cryptic I/O error.

PR #4206, opened 2025-05-27
Bind AsyncLocalStorage.snapshot/wrap to the IoContext

jasnell:
The AsyncLocalStorage frame can capture values that are bound to the current IoContext. This is not always in the users control since we use the ALS storage frame to propagate internal trace spans as well as user-provided values. This changes introduces a new compat flag that, when set, binds the snapshot / bound functions to the current IoContext and will throw an error if the bound functions are called outside of the IoContext in which they were created. This is a potentially breaking change so the compat flag is required. It will be set to be enabled by default, however.

Unfortunately I believe throwing an error in this case is the only reasonable behavior. It becomes rather complicated to deal with cross-request-bound values, especially when the user may not even be aware that the context frame is capturing such values and can't do anything about it. The specific error thrown will be at least more meaningful than the crypto I/O error message that is thrown now.
 
Refs: https://github.com/cloudflare/workerd/issues/4155

/cc @vicb

### DIFF
--- a/src/workerd/api/node/BUILD.bazel
+++ b/src/workerd/api/node/BUILD.bazel
@@ -68,6 +68,7 @@ wd_cc_library(
         # it, move files to the main node target if needed.
         "//src/workerd/io:compatibility-date_capnp",
         "//src/workerd/jsg",
+        "//src/workerd/io",
         "//src/workerd/util:mimetype",
     ],
 )
@@ -404,4 +405,10 @@ wd_test(
     src = "tests/http-nodejs-test.wd-test",
     args = ["--experimental"],
     data = ["tests/http-nodejs-test.js"],
+)
+
+wd_test(
+    src = "tests/bound-als-test.wd-test",
+    args = ["--experimental"],
+    data = ["tests/bound-als-test.js"],
 )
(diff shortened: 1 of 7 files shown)


conico974:
I was going to comment on the issue, you were faster than me :smile:
> Unfortunately I believe throwing an error in this case is the only reasonable behavior.

Wouldn't it be possible to instead replace the initial `IoContext` with the current one instead. It could also be under a flag that would be off by default. 
I feel like it makes snapshot kind of useless inside workers right now

jasnell:
> Wouldn't it be possible to instead replace the initial IoContext with the current one instead...

Unfortunately no. When an API object is bound to a specific IoContext that binding cannot be switched to a different one.

It is still possible to use a shared snapshot, just not one that is created within a request.. for instance, the following would still work:

```
const als = new AsyncLocalStorage();
let snapshot;
als.run('123', () => snapshot = AsyncLocalStorage.snapshot());

export default {
  fetch(req) {
    return new Response(snapshot(() => als.getStore()));
  }
}
```

Because the `snapshot` was created outside of an IoContext it cannot capture any IoContext-bound values and is therefore safe to use, so this would not throw an error.

jasnell:
> Correct me if i'm wrong, but it also means that we won't be able to do any IO inside of the snapshot callback

That's incorrect. When the snapshot callback is invoked from the correct IoContext then you will be able to perform any i/o you want.

anonrig (approved):

PR #4208, opened 2025-05-28
Stop using deprecated context->GetIsolate.

erikcorry:
This is deprecated in the next version of V8, and
for a while it has merely called v8::Isolate::GetCurrent() so we might as well make it clear what we are doing.

See also https://chromium-review.googlesource.com/c/v8/v8/+/6589896

### DIFF
--- a/src/workerd/jsg/type-wrapper.h
+++ b/src/workerd/jsg/type-wrapper.h
@@ -79,7 +79,7 @@ class V8HandleWrapper {
       v8::Local<v8::Value> handle, v8::Global<v8::type>*,                                          \
       kj::Maybe<v8::Local<v8::Object>> parentObject) {                                             \
     if (handle->Is##type()) {                                                                      \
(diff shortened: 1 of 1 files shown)


npaun (approved):

dom96 (approved):

anonrig (approved):

kentonv:
I'm not against merging this for now, but I think the "right" answer would be to have `wrap()` and `tryUnwrap()` take a `v8::Isolate*` or `jsg::Lock&` parameter. I hate relying on thread-locals like `GetCurrent()`, and also passing the pointer from the caller is probably going to be more efficient. I would expect that in all cases, the isolate pointer is readily available in the caller.

jasnell:
Have to agree on avoiding use of `Isolate::GetCurrent()` ... I think everywhere these are called we have the `v8::Isolate*` available and can grab the `jsg::Lock&` from that. I think it makes the most sense to update these to accept `jsg::Lock&`

PR #4213, opened 2025-05-29
Add compat flag for appropriate handling of unrecognized import attributes

jasnell:
In the original module registry implementation, unknown import attributes were ignored. The spec, however, strongly recommends that unknown attributes be treated as errors. Unfortunately doing so now would break existing workers so we need a compat flag.

### DIFF
--- a/src/workerd/jsg/modules.h
+++ b/src/workerd/jsg/modules.h
@@ -223,7 +223,7 @@ v8::MaybeLocal<v8::Promise> dynamicImportCallback(v8::Local<v8::Context> context
     v8::Local<v8::Data> host_defined_options,
     v8::Local<v8::Value> resource_name,
     v8::Local<v8::String> specifier,
-    v8::Local<v8::FixedArray> import_assertions);
+    v8::Local<v8::FixedArray> import_attributes);
 
 kj::Maybe<kj::OneOf<kj::String, ModuleRegistry::ModuleInfo>> tryResolveFromFallbackService(Lock& js,
     const kj::Path& specifier,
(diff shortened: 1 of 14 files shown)


src/workerd/jsg/modules.h
```diff
@@ -650,6 +635,19 @@ v8::MaybeLocal<v8::Promise> dynamicImportCallback(v8::Local<v8::Context> context
...
+  // is because import attributes can alter the interpretation of a module and
+  // are considered to be part of the unique key for caching a module.
+  // Throwing an error for things we do not understand is the safest thing to do.
+  // However, historically we have not followed this guideline in the spec
+  // and unfortunately there are applications deployed that will break if we
+  // started enforcing that guideline without a compat flag.
+  if (!import_attributes.IsEmpty() && import_attributes->Length() > 0 &&
+      js.getThrowOnUnrecognizedImportAssertion()) {
+    return makeRejected(js.v8Error("Unrecognized import attributes specified"));
+  }
```
guybedford, line 649: Not a request, but a recommendation, feel free to ignore - it might actually be advisable here to provide a more specific message for the `type` attribute, where the `type` attribute is recognized, just not supported. So we could have the error be "Import attribute type 'css' is not currently supported" say, at least allowing a more specific cause to the user. Further it could help to note the module which has the error in the error message in case it was within a dependency. This would help dependency errors on import attributes to be actionable fixes at least.

guybedford (approved):

src/workerd/jsg/modules.h
guybedford: I think the main use case here is running code that uses `import json from './file.json' with { type: 'json' }` and getting an error that doesn't help guide the user to the correction.

PR #4216, opened 2025-05-29
Add TreatErrorsAsHostObjects to v8, customize Error serialization

jasnell:
In v8's ValueSerializer, Errors are handled differently than ordinary objects in that arbitrary own properties are not serialized by default. This commit patches V8 to optionally treat Errors as host objects allowing us to serialize all own properties of Error objects. This also allows us to support additional Error constructors that v8's API does not support by default, such as `AggregateError` and `SuppressedError`, as well as allowing us to serialize the workers specific additional own properties that we attach to some errors.

Critically, this will also allow us to attach a faithful serialization of an error to the `details` property of a kj::Exception for tunneled exceptions.

Creating initially as a draft PR to get some initial feedback to make sure this is the right approach we want to take.

One key question is: should we just have jsg::Serializer *always* enable this (currently it's enabled with an option)

A separate internal PR will be necessary, will open that after we're sure this is the direction we want.

### DIFF
--- a/patches/v8/0027-Implement-additional-Exception-construction-methods.patch
+++ b/patches/v8/0027-Implement-additional-Exception-construction-methods.patch
@@ -1,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: James M Snell <jsnell@cloudflare.com>
+Date: Tue, 1 Jul 2025 17:33:43 -0700
+Subject: Implement additional Exception construction methods
+
+Signed-off-by: James M Snell <jsnell@cloudflare.com>
+
+diff --git a/include/v8-exception.h b/include/v8-exception.h
+index 5441a0ab6a403c566e7b0b6002e720c971480893..b9933027aaf9f7842740d6a6742a2c73da65a46a 100644
+--- a/include/v8-exception.h
++++ b/include/v8-exception.h
+@@ -48,6 +48,14 @@ class V8_EXPORT Exception {
+   static Local<Value> WasmSuspendError(Local<String> message,
+                                        Local<Value> options = {});
+   static Local<Value> Error(Local<String> message, Local<Value> options = {});
++  static Local<Value> URIError(Local<String> message,
++                               Local<Value> options = {});
++  static Local<Value> EvalError(Local<String> message,
++                               Local<Value> options = {});
++  static Local<Value> AggregateError(Local<String> message,
++                                     Local<Value> options = {});
++  static Local<Value> SuppressedError(Local<String> message,
++                                      Local<Value> options = {});
+ 
+   /**
+    * Creates an error message for the given exception.
+diff --git a/src/api/api.cc b/src/api/api.cc
+index 99eb9a5dc075a0e7aa38fe31b0576a652bd12cdf..345bfb3fac0e16ad532179dd915c80b454d0cb88 100644
+--- a/src/api/api.cc
++++ b/src/api/api.cc
+@@ -11281,6 +11281,10 @@ DEFINE_ERROR(WasmCompileError, wasm_compile_error)
+ DEFINE_ERROR(WasmLinkError, wasm_link_error)
+ DEFINE_ERROR(WasmRuntimeError, wasm_runtime_error)
+ DEFINE_ERROR(WasmSuspendError, wasm_suspend_error)
++DEFINE_ERROR(EvalError, eval_error)
++DEFINE_ERROR(URIError, uri_error)
++DEFINE_ERROR(AggregateError, aggregate_error)
++DEFINE_ERROR(SuppressedError, suppressed_error)
+ DEFINE_ERROR(Error, error)
+ 
+ #undef DEFINE_ERROR
+diff --git a/src/logging/runtime-call-stats.h b/src/logging/runtime-call-stats.h
+index 9509646458bd4c7b6e4cdfa7a5109a431b8121c1..03c2063841224df6b043a4edeaea9ee1f568b563 100644
+--- a/src/logging/runtime-call-stats.h
++++ b/src/logging/runtime-call-stats.h
+@@ -218,7 +218,11 @@ namespace v8::internal {
+   V(WeakMap_Delete)                                        \
+   V(WeakMap_Get)                                           \
+   V(WeakMap_New)                                           \
+-  V(WeakMap_Set)
++  V(WeakMap_Set)                                           \
++  V(EvalError_New)                                         \
++  V(URIError_New)                                          \
++  V(AggregateError_New)                                    \
++  V(SuppressedError_New)
+ 
+ #define ADD_THREAD_SPECIFIC_COUNTER(V, Prefix, Suffix) \
+   V(Prefix##Suffix)                                    \
--- a/src/workerd/jsg/ser-test.c++
+++ b/src/workerd/jsg/ser-test.c++
@@ -150,12 +150,43 @@ struct SerTestContext: public ContextGlobalObject {
     return result;
   }
 
+  JsObject roundTripError(Lock& js, JsObject errorObj) {
+    Serializer ser(js,
+        Serializer::Options{
+          .treatErrorsAsHostObjects = true,
+        });
+    ser.write(js, errorObj);
+    auto content = ser.release();
+    Deserializer deser(js, content);
+    auto val = KJ_ASSERT_NONNULL(deser.readValue(js).tryCast<JsObject>());
+
+    auto names = errorObj.getPropertyNames(js, KeyCollectionFilter::OWN_ONLY,
+        PropertyFilter::ALL_PROPERTIES, IndexFilter::SKIP_INDICES);
+    for (size_t n = 0; n < names.size(); n++) {
+      auto before = errorObj.get(js, names.get(js, n));
+      auto after = val.get(js, names.get(js, n));
+      if (before.isArray()) {
+        auto beforeArray = KJ_ASSERT_NONNULL(before.tryCast<JsArray>());
+        auto afterArray = KJ_ASSERT_NONNULL(after.tryCast<JsArray>());
+        KJ_ASSERT(beforeArray.size() == afterArray.size());
+        for (size_t i = 0; i < beforeArray.size(); i++) {
+          KJ_ASSERT(beforeArray.get(js, i).strictEquals(afterArray.get(js, i)));
+        }
+      } else {
+        KJ_ASSERT(before.strictEquals(after));
+      }
+    }
+
+    return val;
+  }
+
   JSG_RESOURCE_TYPE(SerTestContext) {
     JSG_NESTED_TYPE(Foo);
     JSG_NESTED_TYPE(Bar);
     JSG_NESTED_TYPE(Baz);
     JSG_NESTED_TYPE(Qux);
     JSG_METHOD(roundTrip);
+    JSG_METHOD(roundTripError);
   }
 };
 JSG_DECLARE_ISOLATE_TYPE(SerTestIsolate,
@@ -282,6 +313,47 @@ KJ_TEST("serialization") {
                 "obj.bar = bar;\n"
                 "roundTrip(obj).bar.val.bar.val.bar.val.i",
       "number", "321");
+}
+
+KJ_TEST("serialization of errors") {
+  Evaluator<SerTestContext, SerTestIsolate> e(v8System);
+
+  e.expectEval(
+      "e = new Error('a', {cause:'c'}); e.foo = true; roundTripError(e).foo", "boolean", "true");
+  e.expectEval(
+      "roundTripError(new TypeError('a', {cause:'c'})) instanceof TypeError", "boolean", "true");
+  e.expectEval(
+      "roundTripError(new RangeError('a', {cause:'c'})) instanceof RangeError", "boolean", "true");
+  e.expectEval("roundTripError(new ReferenceError('a', {cause:'c'})) instanceof ReferenceError",
+      "boolean", "true");
+  e.expectEval("roundTripError(new SyntaxError('a', {cause:'c'})) instanceof SyntaxError",
+      "boolean", "true");
+  e.expectEval("e = new Error(); Object.defineProperty(e, 'name', {value: 'CustomError'}); "
+               "roundTripError(e).name",
+      "string", "CustomError");
+
+  // Throws due to serializing a cycle. Because we serialize the errors as
+  // host objects, we end up being responsible for ensuring that cycles are
+  // not present in the serialized data. For now, we're just punting on this
+  // to keep things simple, but we end up getting an error when we try to
+  // deserialize. Fortunately this case should always be rare.
+  // TODO(later): Handle cycles in errors as an edge case.
+  e.expectEval(R"(
+    const a = new Error('a');
+    a.cause = a;
+    roundTripError(a);
+  )",
+      "throws", "Error: Unable to deserialize cloned data.");
+
+  e.expectEval(
+      "roundTripError(new URIError('a', {cause:'c'})) instanceof URIError", "boolean", "true");
+  e.expectEval(
+      "roundTripError(new EvalError('a', {cause:'c'})) instanceof EvalError", "boolean", "true");
+  e.expectEval("roundTripError(new AggregateError(['a', 'b'], 'c')) instanceof AggregateError",
+      "boolean", "true");
+  e.expectEval("roundTripError(new SuppressedError('a', 'b', 'c', {cause:'c'})) instanceof "
+               "SuppressedError",
+      "boolean", "true");
 }
 
 }  // namespace
--- a/src/workerd/jsg/ser.c++
+++ b/src/workerd/jsg/ser.c++
@@ -10,6 +10,117 @@
 #include <v8-proxy.h>
 
 namespace workerd::jsg {
+namespace {
+// Keep in sync with the nativeError serialization tag defined in
+// worker-interface.capnp
+constexpr uint32_t SERIALIZATION_TAG_NATIVE_ERROR = 10;
+
+// The error tag is serialized as a uint32_t immediately following
+// the SERIALIZATION_TAG_NATIVE_ERROR tag in order to more efficiently
+// determine the type of error when deserializing so that we can
+// construct the appropriate v8::Exception type on deserialization
+// without having to expensive string comparison on the error name.
+enum class ErrorTag : uint32_t {
+  // The UNKNOWN tag is used when the error name is not recognized.
+  // When this occurs, we will serialize the name of the error, and
+  // when it is deserialized, we will create a generic Error and then
+  // set the name to the stored name.
+  UNKNOWN,
+  ERROR,
+  TYPE_ERROR,
+  RANGE_ERROR,
+  REFERENCE_ERROR,
+  SYNTAX_ERROR,
+  WASM_COMPILE_ERROR,
+  WASM_LINK_ERROR,
+  WASM_RUNTIME_ERROR,
+  WASM_SUSPEND_ERROR,
+  EVAL_ERROR,
+  URI_ERROR,
+  AGGREGATE_ERROR,
+  SUPPRESSED_ERROR,
+};
+
+constexpr ErrorTag getErrorTagFromName(jsg::Lock& js, kj::StringPtr str) {
+  if (str == "Error"_kj) {
+    return ErrorTag::ERROR;
+  } else if (str == "TypeError"_kj) {
+    return ErrorTag::TYPE_ERROR;
+  } else if (str == "RangeError"_kj) {
+    return ErrorTag::RANGE_ERROR;
+  } else if (str == "ReferenceError"_kj) {
+    return ErrorTag::REFERENCE_ERROR;
+  } else if (str == "SyntaxError"_kj) {
+    return ErrorTag::SYNTAX_ERROR;
+  } else if (str == "WasmCompileError"_kj) {
+    return ErrorTag::WASM_COMPILE_ERROR;
+  } else if (str == "WasmLinkError"_kj) {
+    return ErrorTag::WASM_LINK_ERROR;
+  } else if (str == "WasmRuntimeError"_kj) {
+    return ErrorTag::WASM_RUNTIME_ERROR;
+  } else if (str == "WasmSuspendError"_kj) {
+    return ErrorTag::WASM_SUSPEND_ERROR;
+  } else if (str == "EvalError"_kj) {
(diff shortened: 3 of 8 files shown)


patches/v8/0025-Add-more-Error-constructor-methods.patch
```diff
@@ -0,0 +1,48 @@
+From e941ed4d066e60db0a5f7da5022ed8aa463c5d15 Mon Sep 17 00:00:00 2001
+From: James M Snell <jsnell@cloudflare.com>
+Date: Thu, 29 May 2025 15:07:37 -0700
+Subject: [PATCH] Add more Error constructor methods
```
anonrig, line 4: Can this be upstreamed?

jasnell: Maybe? I don't want to worry about up streaming yet. Let's make sure this is what we want first.

patches/v8/0025-Add-more-Error-constructor-methods.patch
```diff
@@ -0,0 +1,48 @@
...
+    * Creates an error message for the given exception.
+diff --git a/src/api/api.cc b/src/api/api.cc
+index c6291d1a7ea..2228cac9cae 100644
+--- a/src/api/api.cc
++++ b/src/api/api.cc
+@@ -11254,6 +11254,10 @@ DEFINE_ERROR(WasmCompileError, wasm_compile_error)
+ DEFINE_ERROR(WasmLinkError, wasm_link_error)
+ DEFINE_ERROR(WasmRuntimeError, wasm_runtime_error)
+ DEFINE_ERROR(WasmSuspendError, wasm_suspend_error)
++DEFINE_ERROR(EvalError, eval_error)
```
kentonv: So V8 already defined these internally, just didn't expose them?

src/workerd/jsg/ser.c++
```diff
@@ -110,6 +150,34 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
...
     jsg::Lock& js = jsg::Lock::from(isolate);
 
+    if (object->IsNativeError()) {
+      // Get the standard name, message, stack, cause, error, errors properties from
+      // the error object.
+      writeRawUint32(SERIALIZATION_TAG_NATIVE_ERROR);
+
+      jsg::JsObject errorObj(object);
+      auto name = errorObj.get(js, "name"_kj);
+      writeLengthDelimited(name.toString(js));
```
kentonv: Instead of this ad hoc serialization, I think we should build a plain JS object containing all the properties and then serialize that. We'd still write the tag first, so that when parsing later, we know to turn the object back into an `Error`.

patches/v8/0025-Add-more-Error-constructor-methods.patch
jasnell: Yep. Specifically they didn't expose them via the C++ API.

src/workerd/jsg/ser.c++
jasnell: I've updated to use a hybrid approach. Just serializing as a plain JS object does not actually save us any complexity and actually introduces a little more. I did update to serialize the js values directly but still using a custom layout. Should work a lot better and be fairly efficient.

src/workerd/jsg/ser-test.c++
```diff
@@ -284,5 +315,46 @@ KJ_TEST("serialization") {
...
+  e.expectEval("e = new Error(); Object.defineProperty(e, 'name', {value: 'CustomError'}); "
+               "roundTripError(e).name",
+      "string", "CustomError");
+
+  // Throws due to serializing a cycle. Because we serialize the errors as
+  // host objects, we end up being responsible for ensuring that cycles are
+  // not present in the serialized data. For now, we're just punting on this
+  // to keep things simple, but we end up getting an error when we try to
+  // deserialize. Fortunately this case should always be rare.
+  // TODO(later): Handle cycles in errors as an edge case.
```
jasnell, line 340: This is perhaps the one key limitation of this approach. If the error that is serialized happens to have a circular reference then we need to explicitly handle it. The current deserialization fails in that case. Could we fix that? It's absolutely an edge case that's unlikely to actually happen but people do weird things....

src/workerd/jsg/ser.c++
jasnell: Actually.. no, I updated to use the plain object but just for the additional properties. I want the name and message to be handled separately still since those are needed more up front.

src/workerd/jsg/ser.c++
```diff
@@ -110,6 +227,57 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
...
+        auto name = names.get(js, n);
+        // The name and message typically come from the prototype and therefore
+        // do not show up in the own properties of the error object. However,
+        // they still might show up in the list if the user has modified them
+        // on the error object itself. Since we already have them serialized
+        // above, we can filter them out here.
+        if (name.strictEquals(nameStr) || name.strictEquals(messageStr)) continue;
+        write(js, name);
+        write(js, errorObj.get(js, name));
+      }
```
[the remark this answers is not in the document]
jasnell: Wow, a good review comment from the AI bot. Thank you robot.

src/workerd/jsg/ser.c++
```diff
@@ -110,6 +227,52 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
...
 
+    if (object->IsNativeError()) {
+      auto nameStr = js.str("name"_kj);
+      auto messageStr = js.str("message"_kj);
+
+      // Get the standard name, message, stack, cause, error, errors properties from
+      // the error object.
+      writeRawUint32(SERIALIZATION_TAG_NATIVE_ERROR);
+
+      // A mix off ad-hoc and regular serialization. We first serialize
```
jasnell:
```suggestion
      // A mix of ad-hoc and regular serialization. We first serialize
```

src/workerd/jsg/ser.c++
```diff
@@ -110,6 +227,52 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
...
+      // A mix of ad-hoc and regular serialization. We first serialize
+      // the error tag, which is an enum that identifies the type of error
+      // for faster/easier deserialization. Then we serialize the message which
+      // usually come from the prototype. Then we grab the own properties,
+      // serializing the number of them followed by each name and value in
+      // sequence.
+
+      jsg::JsObject errorObj(object);
+      auto name = errorObj.get(js, nameStr);
+      auto tag = getErrorTagFromName(js, name);
```
kentonv: Hmm, what if someone constructs a specific error type _and_ overrides the name? I guess we lose the type info?

Would it be possible to derive this tag from the error's type, rather than its name? Maybe by looking at the prototype?

kentonv: Wait actually, what if we did this:

- Serialize the error object using V8's built-in serialization, which presumably retains the type for all built-in types.
- Additionally serialize any own properties as a separate object. (I think this inherently covers `name` if it has been overridden.)
- On the receiving side, just merge them.

jasnell: Well, that would be nice (and is what I tried first) but ..  v8's structured clone implementation does not yet actually properly handled `SuppressedError`, etc
image

jasnell: FWIW, even v8's default serializer behavior loses the type information if someone constructs a specific type and overrides the name... this is in chromium:

image

This is also true in Node.js, Deno, Bun, and Firefox. So not something we should really worry about.

So this impl is already several steps ahead of what v8's implementation is doing. It also kind of feels like we're poking around very uncommon edge cases at this point.

Oh.. and on this question: 
> Would it be possible to derive this tag from the error's type, rather than its name? 

Perhaps? But v8 does not expose the error prototypes for us in C++ so it would actually be more work and less efficient to try to use the prototype. It's worth pointing out that v8's own implementation uses the `name` property in precisely the same way (to determine the serialization tag).

jasnell: I do anticipate that eventually v8's implementation will catch up and will do a better job of handling the new built-in errors and properties, at which point we can likely revise this implementation to depend on it, but given that we know it is currently incomplete and we don't really know what their timeline to complete it is, I think going with this approach is likely best.

kentonv: > at which point we can likely revise this implementation to depend on it

Not really. We can't really change anything about this once it's in the wild since people may have stored data in DO storage based on this serialization. We can introduce a new tag and serialize new errors differently, but we'll have to support the old format forever.

The reason I like the layered approach (use V8 serialization of the error, plus a separate serialization of additional properties) is that it feels like a simple, flexible approach that's easy to reason about and support forever.

On the topic of `AggregateError` and `SuppressedError`, don't these both contain additional fields (nested errors) which are kind of important to serialize as well? Using the layered format, we can potentially support these by including these extra fields into the overlay object -- and then, later, if V8 ever adds explicit support, then we can just stop adding them to the overlay, but not have to worry about back-compat.

jasnell: > Not really. We can't really change anything about this once it's in the wild since people may have stored data in DO storage based on this serialization. We can introduce a new tag and serialize new errors differently, but we'll have to support the old format forever.

This PR does not enable this option by default for DO storage uses. It must be explicitly turned on. The initial intent has been to use it solely for the `kj::Exception` `detail` attribute for tunneled exceptions. If we did want to enable this for DO storage later then adding a new tag is straightforward enough.

The layered approach you're suggesting does not currently work adequately since v8 does *not* maintain the type information either.Specifically, this PR actually takes generally the *same* approach that v8 does... looking at the `name` property and serializing a representative tag. Unfortunately, the v8 implementation currently only looks for a subset of the names and does not otherwise preserve the correct type information.

> On the topic of AggregateError and SuppressedError, don't these both contain additional fields (nested errors) which are kind of important to serialize as well? Using the layered format, we can potentially support these by including these extra fields into the overlay object -- and then, later, if V8 ever adds explicit support, then we can just stop adding them to the overlay, but not have to worry about back-compat.

This implementation *does* support those additional fields. The new `error` / `errors` / `cause` properties are added as own properties to the `SuppressedError` and `AggregateError` objects and are supported by this implementation.

src/workerd/jsg/ser.c++
```diff
@@ -110,6 +227,52 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
...
+      // Get the standard name, message, stack, cause, error, errors properties from
+      // the error object.
+      writeRawUint32(SERIALIZATION_TAG_NATIVE_ERROR);
+
+      // A mix of ad-hoc and regular serialization. We first serialize
+      // the error tag, which is an enum that identifies the type of error
+      // for faster/easier deserialization. Then we serialize the message which
+      // usually come from the prototype. Then we grab the own properties,
+      // serializing the number of them followed by each name and value in
+      // sequence.
```
jasnell, line 246:
```suggestion
      // sequence.
      //
      // Why not just rely on V8's serialization of errors? Good question! Unfortunately V8's
      // serialization does not include enough of the necessary data. For one, it currently only
      // supports a limited subset of the built-in Error types provided by the language and the
      // engine. For instance, the new SuppressedError and AggregateError types are not
      // handled correctly. V8's implementation, like this impl, extracts the `name` property
      // from the error object and uses it to serialize a representative tag representing the 
      // error type, however, it only looks at a handful of names. If we try serializing one of
      // the new SuppressedError, types, for instance, V8 would deserialize that into an Error
      // object that happens to have the name `SuppressedError` but does not satisfy the
      // `deserializedError instanceof SuppressedError` check. It also won't have the `error`,
      // and `suppressed` properties since V8's impl does not preserve any of the additional
      // own properties added to an error other than the `cause`.
      //
      // So why not rely on V8's default error serialization with some supplements for the
      // additional properties? Because V8 still loses necessary information in that translation.
      // V8 currently will not serialize or deserialize the correct type information to ensure that
      // the transformation does not lose essential information. Eventually I suspect it will be
      // updated to provide the correct info (we might even submit a patch to do so), at which
      // time we can update this implementation to rely on that -- likely by introducing another
      // alternative serialization tag -- but for now, this implementation here goes beyond V8's
      // implementation by supporting the full range of built-in error types, special own properties,
      // and our own extension own properties.
```

src/workerd/jsg/ser.c++
```diff
@@ -110,6 +227,52 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
...
+      auto tag = getErrorTagFromName(js, name);
+      writeRawUint32(static_cast<uint32_t>(tag));
+      // We only write the name if it is not one of the known error types.
+      if (tag == ErrorTag::UNKNOWN) {
+        write(js, name);
+      }
+
+      write(js, errorObj.get(js, messageStr));
+
+      auto names = errorObj.getPropertyNames(js, KeyCollectionFilter::OWN_ONLY,
```
jasnell, line 260:
```suggestion
      // This will cover not only the extension properties we add via tunneled exceptions,
      // along with any additional own properties the user may attach to the error object,
      // it also ensures that the standard `cause`, `error`, `errors`, and `suppressed`
      // properties that are defined by the language standard are also captured and
      // serialized. Later, when v8 finally gets around to supporting these by default
      // we can likely filter them out.
      auto names = errorObj.getPropertyNames(js, KeyCollectionFilter::OWN_ONLY,
```

jasnell:
Actually, I remembered one additional necessary todo here.. we need to make sure this supports Node.js' custom errors in `nodejs_compat` mode. The Node.js errors subclass the standard error types and add the `code` property as a prototype property rather than an own property. Passing those through clone here would mean the type information would be lost and the code would not be preserved. So before this can land we should try to support that as much as possible. Really here, tho, we don't need to necessarily support the full type information as it is the `code` property that is the most important.

Update: Actually... no, scratch that... I just verified that `code` is added as an own property so this should actually be fine. Some of the `nodejs_compat` errors will not have their type info preserved, however... namely `AssertionError`, `AbortError`, `DNSError`, etc since those subclass with unrecognized types.

jasnell:
PR updated to account for the recent changes in v8 13.9.

src/workerd/jsg/ser.c++
```diff
@@ -10,6 +10,118 @@
...
+  AGGREGATE_ERROR,
+  SUPPRESSED_ERROR,
+  // The UNKNOWN tag is used when the error name is not recognized.
+  // When this occurs, we will serialize the name of the error, and
+  // when it is deserialized, we will create a generic Error and then
+  // set the name to the stored name.
+  UNKNOWN,
+};
+
+ErrorTag getErrorTagFromName(jsg::Lock& js, const JsValue& name) {
```
anonrig: If you replace this function to take StringPtr, we can make this constexpr

src/workerd/jsg/ser.c++
```diff
@@ -91,22 +205,20 @@ bool Serializer::HasCustomHostObject(v8::Isolate* isolate) {
   // V8 will always call WriteHostObject() for objects that have internal fields. We only need
   // to override IsHostObject() if we want to treat pure-JS objects differently, which we do if
   // treatClassInstancesAsPlainObjects is false.
-  return !treatClassInstancesAsPlainObjects;
+  return !treatClassInstancesAsPlainObjects || treatErrorsAsHostObjects;
```
anonrig, line 207: Can you update the comment for this change?

patches/v8/0027-Implement-additional-Exception-construction-methods.patch
```diff
@@ -0,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
danlapid, line 1: Can we upstream this please?

src/workerd/jsg/ser.c++
```diff
@@ -118,6 +229,58 @@ v8::Maybe<bool> Serializer::WriteHostObject(v8::Isolate* isolate, v8::Local<v8::
   try {
     jsg::Lock& js = jsg::Lock::from(isolate);
 
+    if (object->IsNativeError()) {
```
danlapid, line 232: This probably needs an autogate or a compat flag.
Otherwise we risk sending serialized objects of type nativeError to instances that did not get the new deserialization code.

patches/v8/0027-Implement-additional-Exception-construction-methods.patch
jasnell: @anonrig is already working on that :-) ... I'd prefer if we didn't wait tho as it is likely to take a while to get the upstream changed landed as it may need to work through an html spec change.

src/workerd/jsg/ser.c++
danlapid: I guess we should guard the value of treatErrorsAsHostObjects ?

jasnell: This is not enabled by default on the serializer so nothing will break when landed. That is, there's nothing to gate in this particular PR.

patches/v8/0027-Implement-additional-Exception-construction-methods.patch
anonrig: Spec change: https://github.com/whatwg/html/pull/11425
WPT change: https://github.com/web-platform-tests/wpt/pull/53541
v8 change: https://chromium-review.googlesource.com/c/v8/v8/+/6700488

src/workerd/jsg/ser.c++
```diff
@@ -10,6 +10,117 @@
...
+  WASM_SUSPEND_ERROR,
+  EVAL_ERROR,
+  URI_ERROR,
+  AGGREGATE_ERROR,
+  SUPPRESSED_ERROR,
+  // The UNKNOWN tag is used when the error name is not recognized.
+  // When this occurs, we will serialize the name of the error, and
+  // when it is deserialized, we will create a generic Error and then
+  // set the name to the stored name.
+  UNKNOWN,
```
danlapid: Unknown should not be last, as we add new fields to this enum we will get error type confusion between mismatching versions of the serializer and deserializer.

src/workerd/jsg/ser.c++
```diff
@@ -10,6 +10,117 @@
...
+// Keep in sync with the nativeError serialization tag defined in
+// worker-interface.capnp
+constexpr uint32_t SERIALIZATION_TAG_NATIVE_ERROR = 10;
+
+// The error tag is serialized as a uint32_t immediately following
+// the SERIALIZATION_TAG_NATIVE_ERROR tag in order to more efficiently
+// determine the type of error when deserializing so that we can
+// construct the appropriate v8::Exception type on deserialization
+// without having to expensive string comparison on the error name.
+enum class ErrorTag : uint32_t {
```
danlapid, line 23: Is there a better way to do this that doesn't require a hardcoded list of allowed exception types and possibly supports user-generated types?
Maybe serialize a string of the exception type?

src/workerd/jsg/ser.c++
jasnell:
```suggestion
  // The UNKNOWN tag is used when the error name is not recognized.
  // When this occurs, we will serialize the name of the error, and
  // when it is deserialized, we will create a generic Error and then
  // set the name to the stored name.
  UNKNOWN,
  ERROR,
  TYPE_ERROR,
  RANGE_ERROR,
  REFERENCE_ERROR,
  SYNTAX_ERROR,
  WASM_COMPILE_ERROR,
  WASM_LINK_ERROR,
  WASM_RUNTIME_ERROR,
  WASM_SUSPEND_ERROR,
  EVAL_ERROR,
  URI_ERROR,
  AGGREGATE_ERROR,
  SUPPRESSED_ERROR,
```

src/workerd/jsg/ser.c++
```diff
@@ -10,6 +10,117 @@
...
+      return JsObject(v8::Exception::AggregateError(str).As<v8::Object>());
+    }
+    case ErrorTag::SUPPRESSED_ERROR: {
+      return JsObject(v8::Exception::SuppressedError(str).As<v8::Object>());
+    }
+    case ErrorTag::UNKNOWN: {
+      return JsObject(v8::Exception::Error(str).As<v8::Object>());
+    }
+  }
+  return JsObject(v8::Exception::Error(str).As<v8::Object>());
```
danlapid, line 121: Isn't it better to just not serialize in this case?
Eliding the error type seems wrong?
Does this actually match current behavior?

src/workerd/jsg/ser.c++
jasnell: This is the most compact way and parallels exactly what v8 itself is doing under the covers. I quite intentionally wanted to at least follow the pattern of the approach v8 is taking internally.

src/workerd/jsg/ser.c++
jasnell: Again, this is echoing exactly what v8 is doing internally and follows the same behavior as v8 implements now. It's likely best to serialize/deserialize *something* usable even if the process is lossy.

patches/v8/0027-Implement-additional-Exception-construction-methods.patch
```diff
@@ -0,0 +1,58 @@
...
++++ b/include/v8-exception.h
+@@ -48,6 +48,14 @@ class V8_EXPORT Exception {
+   static Local<Value> WasmSuspendError(Local<String> message,
+                                        Local<Value> options = {});
+   static Local<Value> Error(Local<String> message, Local<Value> options = {});
++  static Local<Value> URIError(Local<String> message,
++                               Local<Value> options = {});
++  static Local<Value> EvalError(Local<String> message,
++                               Local<Value> options = {});
++  static Local<Value> AggregateError(Local<String> message,
```
anonrig, line 20: This seems to be wrong. AggregateError constructor arguments needs to be changed.

patches/v8/0027-Implement-additional-Exception-construction-methods.patch
```diff
@@ -0,0 +1,58 @@
...
++                                     Local<Value> options = {});
++  static Local<Value> SuppressedError(Local<String> message,
```
anonrig, line 22: As same as suppressed error

patches/v8/0027-Implement-additional-Exception-construction-methods.patch
jasnell: Discussed in chat, but for others: These are intentionally not using the JS constructors for these, and instead is following the pattern established for other errors in the v8 api. We attach the additional properties after creating the error. So these are correct as is.

PR #4220, opened 2025-05-30
Make Socket async disposable, and Sql Cursor sync disposable

jasnell:
Start making some api types disposable starting with Socket and Sql Cursor. This is really just to demonstrate how we can start making some of our types supports dispose/async dispose. We don't have many types that actually have a lifecycle in which disposal is relevant but the few we do have can be adapted easily.

example:

```js
await using socket = connect('...');
// socket automatically closes when scope exits.
```


dom96 (approved):
Looks good, but can you add a socket test too?

PR #4223, opened 2025-05-30
update v8 to v13.8

anonrig:
This pull-request includes a new patch to update the BUILD.bazel of v8 to include changes necessary to add dragonbox dependency.


build/BUILD.dragonbox
```diff
@@ -0,0 +1,9 @@
+cc_library(
+    name = "dragonbox",
+    srcs = ["include/dragonbox/dragonbox.h"],
```
npaun, line 3:
```suggestion
```

I don't think you need to list it twice.

jasnell (approved):

build/deps/v8.bzl
```diff
@@ -28,7 +28,7 @@ PATCHES = [
     "0020-Remove-unneded-latomic-linker-flag.patch",
     "0021-Add-methods-to-get-heap-and-external-memory-sizes-di.patch",
     "0022-Remove-DCHECK-from-WriteOneByteV2-to-skip-v8-fatal.patch",
-    "0023-Add-more-sandbox-APIs.patch",
+    "0023-Modify-where-to-look-for-dragonbox.patch",
```
npaun, line 31: Just confirming we no longer need the Add-more-sandbox-APIs patch

npaun (approved):

build/deps/v8.bzl
anonrig: Yes it's upstreamed.

build/BUILD.dragonbox
npaun: Fair, let's leave it then.

WORKSPACE
```diff
@@ -93,6 +93,18 @@ http_archive(
     url = "https://github.com/google/tcmalloc/tarball/cf3dc2d98bd64cb43f4f98db0acaf5028a7b81eb",
 )
 
+git_repository(
+    name = "dragonbox",
+    build_file_content = """cc_library(
+            name = "dragonbox",
+            hdrs = glob(["include/dragonbox/*.h"]),
+            visibility = ["//visibility:public"],
+            include_prefix = "third_party/dragonbox/include/dragonbox",
```
fhanau: @anonrig I think this should just be `third_party/dragonbox` since the `include/dragonbox` path already exists in the dragonbox repo.

anonrig: Thanks for the review. I've updated this.

anonrig: This doesn't seem to work.

```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
external/v8/src/numbers/conversions.cc:26:10: fatal error: 'dragonbox.h' file not found
   26 | #include "dragonbox.h"
      |          ^~~~~~~~~~~~~
1 error generated.
```

What's the correct path for the include statement here?

anonrig: Maybe this will work. I'll try

```
#include "third_party/dragonbox/include/dragonbox/dragonbox.h"
```

.claude/settings.local.json
danlapid, line 1: Don't commit please

WORKSPACE
fhanau: The version you just pushed should work – if you want to go without patching conversions.cc it should be `include_prefix = third_party/dragonbox/src` but either works.

.claude/settings.local.json
anonrig: Dammit.

WORKSPACE
```diff
@@ -78,6 +78,18 @@ http_archive(
     url = "https://github.com/google/tcmalloc/tarball/cf3dc2d98bd64cb43f4f98db0acaf5028a7b81eb",
 )
 
+git_repository(
+    name = "dragonbox",
+    build_file_content = """cc_library(
+            name = "dragonbox",
+            hdrs = glob(["include/dragonbox/*.h"]),
+            visibility = ["//visibility:public"],
+            include_prefix = "third_party/dragonbox",
```
danlapid:
```suggestion
            include_prefix = "third_party/dragonbox/src",
```

If you add src can we drop the include line change in your patch?

anonrig: Nice. This works. I added a new commit with this change.

PR #4225, opened 2025-05-31
NodeJS Compat: more utils tests, experimentalWarning, styleText

guybedford:
Adds more utils tests including:

* test/parallel/test-util-callbackify.js
* test/parallel/test-util-inherits.js
* test/parallel/test-util-promisify.js
* test/parallel/test-util-stripvtcontrolcharacters.js

Then also includes a noop implementation of `util.emitExperimentalWarning`, and adds support for `util.styleText`, porting those test suites as well.

### DIFF
--- a/src/node/internal/internal_inspect.ts
+++ b/src/node/internal/internal_inspect.ts
@@ -450,11 +450,15 @@ const meta = [
 // Adopted from https://github.com/chalk/ansi-regex/blob/HEAD/index.js
 // License: MIT, authors: @sindresorhus, Qix-, arjunmehta and LitoMore
 // Matches all ansi escape code sequences in a string
-const ansiPattern =
+const ansiPattern = new RegExp(
   '[\\u001B\\u009B][[\\]()#;?]*' +
-  '(?:(?:(?:(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]+)*' +
-  '|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)' +
-  '|(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-ntqry=><~]))';
+    '(?:(?:(?:(?:;[-a-zA-Z\\d\\/\\#&.:=?%@~_]+)*' +
+    '|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/\\#&.:=?%@~_]*)*)?' +
+    '(?:\\u0007|\\u001B\\u005C|\\u009C))' +
+    '|(?:(?:\\d{1,4}(?:;\\d{0,4})*)?' +
+    '[\\dA-PR-TZcf-nq-uy=><~]))',
+  'g'
+);
 const ansi = new RegExp(ansiPattern, 'g');
 
 interface Context extends Required<InspectOptionsStylized> {
@@ -661,7 +665,7 @@ inspect.colors = {
   bgMagentaBright: [105, defaultBG],
   bgCyanBright: [106, defaultBG],
   bgWhiteBright: [107, defaultBG],
-};
+} as any as Record<string, [number, number]>;
 
 function defineColorAlias(target: string, alias: string) {
   Object.defineProperty(inspect.colors, alias, {
--- a/src/node/internal/internal_utils.ts
+++ b/src/node/internal/internal_utils.ts
@@ -209,9 +209,12 @@ function callbackifyOnRejected(
   reason: unknown,
   cb: (error?: unknown) => void
 ): void {
+  // `!reason` guard inspired by bluebird (https://github.com/petkaantonov/bluebird/blob/2207fae3572f03b089bc92d3a6cefdd278cff7ab/src/nodeify.js#L30-L43).
+  // Because `null` is a special error value in callbacks which means "no error
+  // occurred", we error-wrap so the callback consumer can distinguish between
+  // "the promise rejected with null" or "the promise fulfilled with undefined".
   if (!reason) {
-    cb(new ERR_FALSY_VALUE_REJECTION(String(reason)));
-    return;
+    reason = new ERR_FALSY_VALUE_REJECTION(reason as any as string);
   }
   cb(reason);
 }
--- a/src/node/internal/streams_util.d.ts
+++ b/src/node/internal/streams_util.d.ts
@@ -1,3 +1,202 @@
+import { Buffer } from 'node:buffer';
+
+// Exported symbols
+export const kDestroyed: unique symbol;
+export const kIsErrored: unique symbol;
+export const kIsReadable: unique symbol;
+export const kIsDisturbed: unique symbol;
+export const kPaused: unique symbol;
+export const kOnFinished: unique symbol;
+export const kDestroy: unique symbol;
+export const kConstruct: unique symbol;
+export const kIsDestroyed: unique symbol;
+export const kIsWritable: unique symbol;
+export const kOnConstructed: unique symbol;
+
+// Stream type definitions
+interface ReadableState {
+  readable?: boolean;
+  ended?: boolean;
+  endEmitted?: boolean;
+  destroyed?: boolean;
+  errored?: Error | null;
+  errorEmitted?: boolean;
+  closed?: boolean;
+  closeEmitted?: boolean;
+  constructed?: boolean;
+  reading?: boolean;
+  autoDestroy?: boolean;
+  emitClose?: boolean;
+  objectMode?: boolean;
+  length?: number;
+}
+
+interface WritableState {
+  writable?: boolean;
+  ended?: boolean;
+  finished?: boolean;
+  destroyed?: boolean;
+  errored?: Error | null;
+  errorEmitted?: boolean;
+  closed?: boolean;
+  closeEmitted?: boolean;
+  constructed?: boolean;
+  finalCalled?: boolean;
+  prefinished?: boolean;
+  ending?: boolean;
+  autoDestroy?: boolean;
+  emitClose?: boolean;
+  objectMode?: boolean;
+  length?: number;
+}
+
+interface NodeStreamLike {
+  _readableState?: ReadableState;
+  _writableState?: WritableState;
+  readable?: boolean;
+  writable?: boolean;
+  readableEnded?: boolean;
+  writableEnded?: boolean;
+  readableFinished?: boolean;
+  writableFinished?: boolean;
+  readableErrored?: Error | null;
+  writableErrored?: Error | null;
+  destroyed?: boolean;
+  closed?: boolean;
+  readableDidRead?: boolean;
+  readableAborted?: boolean;
+  _closed?: boolean;
+  _defaultKeepAlive?: boolean;
+  _removedConnection?: boolean;
+  _removedContLen?: boolean;
+  _sent100?: boolean;
+  _consuming?: boolean;
+  _dumped?: boolean;
+  req?: any;
+  aborted?: boolean;
+  pipe?: Function;
+  on?: Function;
+  pause?: Function;
+  resume?: Function;
+  write?: Function;
+  emit?: Function;
+  once?: Function;
+  removeListener?: Function;
+  destroy?: Function;
+  close?: Function;
+  abort?: Function;
+  listenerCount?: Function;
+  _construct?: Function;
+  _destroy?: Function;
+  setHeader?: Function;
+  socket?: any;
+  [kDestroyed]?: boolean;
+  [kIsErrored]?: boolean;
+  [kIsReadable]?: boolean;
+  [kIsDisturbed]?: boolean;
+}
+
+interface WebReadableStream {
+  pipeThrough: Function;
+  getReader: Function;
+  cancel: Function;
+}
+
+interface WebWritableStream {
+  getWriter: Function;
+  abort: Function;
+}
+
+// Stream detection functions
+export function isReadableNodeStream(obj: any, strict?: boolean): boolean;
+export function isWritableNodeStream(obj: any): boolean;
+export function isDuplexNodeStream(obj: any): boolean;
+export function isNodeStream(obj: any): boolean;
+export function isReadableStream(obj: any): boolean;
+export function isWritableStream(obj: any): boolean;
+export function isIterable(obj: any, isAsync?: boolean): boolean;
+
+// Stream state functions
+export function isDestroyed(stream: NodeStreamLike): boolean | null;
+export function isWritableEnded(stream: NodeStreamLike): boolean | null;
+export function isWritableFinished(
+  stream: NodeStreamLike,
+  strict?: boolean
+): boolean | null;
+export function isReadableEnded(stream: NodeStreamLike): boolean | null;
+export function isReadableFinished(
+  stream: NodeStreamLike,
+  strict?: boolean
+): boolean | null;
+export function isReadable(stream: NodeStreamLike): boolean | null;
+export function isWritable(stream: NodeStreamLike): boolean | null;
+
+interface IsFinishedOptions {
+  readable?: boolean;
+  writable?: boolean;
+}
+
+export function isFinished(
+  stream: NodeStreamLike,
+  opts?: IsFinishedOptions
+): boolean | null;
+export function isWritableErrored(stream: NodeStreamLike): Error | null;
+export function isReadableErrored(stream: NodeStreamLike): Error | null;
+export function isClosed(stream: NodeStreamLike): boolean | null;
+export function isOutgoingMessage(stream: any): boolean;
+export function isServerResponse(stream: any): boolean;
+export function isServerRequest(stream: any): boolean;
+export function willEmitClose(stream: NodeStreamLike): boolean | null;
+export function isDisturbed(stream: NodeStreamLike): boolean;
+export function isErrored(stream: NodeStreamLike): boolean;
+
+// Utility functions
+export const nop: () => void;
+export function once<T extends Function>(callback: T): T;
+
+// High water mark functions
+export function highWaterMarkFrom(
+  options: any,
+  isDuplex: boolean,
+  duplexKey: string
+): number | null;
+export function getDefaultHighWaterMark(objectMode?: boolean): number;
+export function setDefaultHighWaterMark(): never;
+export function getHighWaterMark(
+  state: { objectMode?: boolean },
+  options: any,
+  duplexKey: string,
+  isDuplex: boolean
+): number;
+
+// Abort signal function
+export function addAbortSignal(
+  signal: AbortSignal,
+  stream: NodeStreamLike
+): NodeStreamLike;
+
+// BufferList class
+export class BufferList {
+  head: BufferListNode | null;
+  tail: BufferListNode | null;
+  length: number;
+
+  push(v: Buffer | string): void;
+  unshift(v: Buffer | string): void;
+  shift(): Buffer | string | undefined;
+  clear(): void;
+  join(s: string): string;
+  concat(n: number): Buffer;
+  consume(n: number, hasStrings?: boolean): Buffer | string;
+  first(): Buffer | string;
+  [Symbol.iterator](): IterableIterator<Buffer | string>;
+}
+
+interface BufferListNode {
+  data: Buffer | string;
+  next: BufferListNode | null;
+}
+
 import type { FinishedOptions } from 'node:stream';
 
 type FinishedStream =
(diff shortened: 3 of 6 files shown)


src/node/internal/streams_util.d.ts
```diff
@@ -1,3 +1,202 @@
+import { Buffer } from 'node:buffer';
```
guybedford, line 1: For this file, I pasted the original and had AI generate the typing file.

src/node/util.ts
```diff
@@ -241,8 +249,71 @@ export function getSystemErrorMessage(): void {
...
+    left += escapeStyleCode(formatCodes ? formatCodes[0] : undefined);
+    right = `${escapeStyleCode(formatCodes ? formatCodes[1] : undefined)}${right}`;
+  }
+
+  return skipColorize ? text : `${left}${text}${right}`;
+}
+
+export function emitExperimentalWarning(_msg: string): void {
+  // noop
 }
```
guybedford, line 317: It seems fine to me to just implement this as a noop like this. Further feedback welcome.

src/node/internal/internal_inspect.ts
```diff
@@ -450,11 +450,15 @@ const meta = [
...
-  '(?:(?:(?:(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]+)*' +
-  '|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)' +
-  '|(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-ntqry=><~]))';
+    '(?:(?:(?:(?:;[-a-zA-Z\\d\\/\\#&.:=?%@~_]+)*' +
+    '|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/\\#&.:=?%@~_]*)*)?' +
+    '(?:\\u0007|\\u001B\\u005C|\\u009C))' +
+    '|(?:(?:\\d{1,4}(?:;\\d{0,4})*)?' +
+    '[\\dA-PR-TZcf-nq-uy=><~]))',
+  'g'
+);
```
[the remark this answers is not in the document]
guybedford: This is copied from Node.js source, keeping in sync seems better.

src/node/util.ts
```diff
@@ -241,8 +249,71 @@ export function getSystemErrorMessage(): void {
...
+        stream
+      );
+    }
+
+    // If the stream is falsy or should not be colorized, set skipColorize to true
+    skipColorize =
+      (stream as any)?.isTTY &&
+      (typeof (stream as any)!.getColorDepth === 'function'
+        ? (stream as any)!.getColorDepth() > 2
+        : true);
```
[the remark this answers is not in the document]
guybedford: Fixed.

src/node/util.ts
anonrig: Can you add a more descriptive comment to here? Mentioning we're not deliberately implementing this because XYZ.

src/node/internal/streams_util.d.ts
```diff
@@ -1,3 +1,202 @@
...
+export function setDefaultHighWaterMark(): never;
+export function getHighWaterMark(
+  state: { objectMode?: boolean },
+  options: any,
+  duplexKey: string,
+  isDuplex: boolean
+): number;
+
+// Abort signal function
+export function addAbortSignal(
```
anonrig, line 173: You should use as much as you can from types/node package. Declarations are available at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/stream.d.ts

export { addAbortSignal } from 'node:stream'

src/node/internal/streams_util.d.ts
```diff
@@ -1,3 +1,202 @@
...
+  first(): Buffer | string;
+  [Symbol.iterator](): IterableIterator<Buffer | string>;
+}
+
+interface BufferListNode {
+  data: Buffer | string;
+  next: BufferListNode | null;
+}
+
 import type { FinishedOptions } from 'node:stream';
```
anonrig, line 200: This should be on the top of the file

src/node/internal/streams_util.js
```diff
@@ -116,6 +116,25 @@ export function isNodeStream(obj) {
   );
 }
 
+export function isReadableStream(obj) {
```
anonrig, line 119: We have the same implementation in src/cloudflare/internal/images-api.ts line 308. Can we remove the duplicated one?

src/node/util.ts
```diff
@@ -148,7 +156,7 @@ function pad(n: any): string {
 
 // prettier-ignore
 const months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep',
-                'Oct', 'Nov', 'Dec'];
+  'Oct', 'Nov', 'Dec'];
```
anonrig, line 159: Unnecessary change

src/node/util.ts
```diff
@@ -241,8 +249,71 @@ export function getSystemErrorMessage(): void {
...
 }
 
-export function styleText(): void {
-  throw new Error('node:util styleText is not implemented');
+function escapeStyleCode(code: number | undefined): string {
+  if (code === undefined) return '';
+  return `\u001b[${code}m`;
+}
+
+// TODO(gbedford): We do not yet implement process.stdout, so to ensure the correct
```
anonrig, line 257: We don't use username todos, but always prefer TODO(soon) for things that are prioritized. All other TODOs are from node.js codebase which we deliberately leave as is to keep code difference in minimum

```suggestion
// TODO(soon): We do not yet implement process.stdout, so to ensure the correct
```

src/node/util.ts
```diff
@@ -241,8 +249,71 @@ export function getSystemErrorMessage(): void {
...
+// behaviour, we use a placeholder value internally.
+const stdoutPlaceholder = Object.create(null);
+export function styleText(
+  format: string | string[],
+  text: string,
+  {
+    validateStream = true,
+    stream = stdoutPlaceholder,
+  }: { validateStream?: boolean; stream?: ReadableStream | WritableStream } = {}
```
anonrig, line 266: Can you move the type declaration to the global scope? It improves readability

src/node/util.ts
```diff
@@ -241,8 +249,71 @@ export function getSystemErrorMessage(): void {
...
+): string {
+  validateString(text, 'text');
+  validateBoolean(validateStream, 'options.validateStream');
```
anonrig, line 269: if validateStream !== true, you should run this, since default value is true.

src/node/util.ts
```diff
@@ -241,8 +249,71 @@ export function getSystemErrorMessage(): void {
...
+      throw new ERR_INVALID_ARG_TYPE(
+        'stream',
+        ['ReadableStream', 'WritableStream', 'Stream'],
+        stream
+      );
+    }
+
+    // If the stream is falsy or should not be colorized, set skipColorize to true
+    skipColorize =
+      (stream as any)?.isTTY &&
```
anonrig, line 288: We tend to avoid using any as much as we can. For example, you expect stream to be an object in here, but if you pass `true` it will throw an error

src/workerd/api/node/tests/util-nodejs-test.js
```diff
@@ -25,24 +25,73 @@
...
+  }, exact);
+}
 function assertCalledMustCalls() {
   try {
     for (const error of remainingMustCallErrors) throw error;
   } finally {
     remainingMustCallErrors.clear();
   }
 }
+function commonMustNotCall(msg) {
```
anonrig, line 61: We don't implement common methods in workerd. We use 

```
const { promise, resolve } = Promise.withResolvers()
```

or 

```
const { mock } = require('node:test')
```

to get the call count of a function to validate if it runs multiple times.

src/workerd/api/node/tests/util-nodejs-test.js
```diff
@@ -4894,3 +4943,913 @@ export const testTypes = {
     }
   },
 };
+
+// parallel/test-util-inspect-getters-accessing-this.js
```
anonrig, line 4947: Can you use a github url with a specific commit hash to make sure we have a immutable reference to the test you're porting?

src/workerd/api/node/tests/util-nodejs-test.js
```diff
@@ -4894,3 +4943,913 @@ export const testTypes = {
...
+    // {
+    //   const warningHandler = commonMustNotCall();
+    //   process.on('warning', warningHandler);
+    //   function foo() {}
+    //   foo.constructor = (async () => {}).constructor;
+    //   promisify(foo);
+    //   process.off('warning', warningHandler);
+    // }
+    //
+    // commonExpectWarning(
```
anonrig, line 5456: We don't implement deprecated tests at workerd, since it will eventually be removed from node.js

src/workerd/api/node/tests/util-nodejs-test.js
```diff
@@ -4894,3 +4943,913 @@ export const testTypes = {
...
+    {
+      function fn() {}
+      fn[promisify.custom] = 42;
+      assert.throws(() => promisify(fn), {
+        code: 'ERR_INVALID_ARG_TYPE',
+        name: 'TypeError',
+      });
+    }
+
+    // customPromisifyArgs unsupported
```
anonrig, line 5523: Can you add a TODO(soon) to these methods?

src/workerd/api/node/tests/util-nodejs-test.js
```diff
@@ -4894,3 +4943,913 @@ export const testTypes = {
...
+
+    {
+      const err = new Error();
+
+      const a = promisify((cb) => cb(err))();
+      const b = promisify(() => {
+        throw err;
+      })();
+
+      Promise.all([
```
anonrig, line 5668:
```suggestion
      await Promise.all([
```

src/workerd/api/node/tests/util-nodejs-test.js
```diff
@@ -4894,3 +4943,913 @@ export const testTypes = {
...
+    styleText('red', 'text', { stream: {}, validateStream: false });
+
+    assert.strictEqual(
+      styleText('red', 'test', { validateStream: false }),
+      styled
+    );
+
+    assert.strictEqual(styleText('none', 'test'), 'test');
+
+    // const fd = common.getTTYfd();
```
anonrig, line 5820: If we're going to implement this, can you add a TODO? If not, we should remove this

src/node/internal/internal_inspect.ts
```diff
@@ -661,7 +665,7 @@ inspect.colors = {
   bgMagentaBright: [105, defaultBG],
   bgCyanBright: [106, defaultBG],
   bgWhiteBright: [107, defaultBG],
-};
+} as any as Record<string, [number, number]>;
```
anonrig, line 668: you should assign this object to a value and set the type there, rather than here. This force-casting is not a recommended approach which eliminates type script checker capabilities.

src/node/util.ts
```diff
@@ -241,8 +249,72 @@ export function getSystemErrorMessage(): void {
...
+        ['ReadableStream', 'WritableStream', 'Stream'],
+        stream
+      );
+    }
+
+    // If the stream is falsy or should not be colorized, set skipColorize to true
+    skipColorize = !(
+      (stream as any)?.isTTY &&
+      (typeof (stream as any)!.getColorDepth === 'function'
+        ? (stream as any)!.getColorDepth() > 2
```
jasnell, line 290: I'm waiting for @anonrig to yell at you for using `any` like he yells at me for it :-P lol

src/node/internal/internal_utils.ts
```diff
@@ -209,9 +209,12 @@ function callbackifyOnRejected(
...
   cb: (error?: unknown) => void
 ): void {
+  // `!reason` guard inspired by bluebird (https://github.com/petkaantonov/bluebird/blob/2207fae3572f03b089bc92d3a6cefdd278cff7ab/src/nodeify.js#L30-L43).
+  // Because `null` is a special error value in callbacks which means "no error
+  // occurred", we error-wrap so the callback consumer can distinguish between
+  // "the promise rejected with null" or "the promise fulfilled with undefined".
   if (!reason) {
-    cb(new ERR_FALSY_VALUE_REJECTION(String(reason)));
-    return;
+    reason = new ERR_FALSY_VALUE_REJECTION(reason as any as string);
```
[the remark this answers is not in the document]
anonrig: rather than unknown, we should use string in line 209. Force casting should be avoided. If we can't we can just use ts-expect-error TSXXXXX comment above this line, which will help us in the future (rather than force casting)

src/node/internal/streams_util.d.ts
anonrig: Buffer is global. This line is not needed

src/node/util.ts
```diff
@@ -241,8 +249,71 @@ export function getSystemErrorMessage(): void {
...
+  text: string,
+  {
+    validateStream = true,
+    stream = stdoutPlaceholder,
+  }: { validateStream?: boolean; stream?: ReadableStream | WritableStream } = {}
+): string {
+  validateString(text, 'text');
+  validateBoolean(validateStream, 'options.validateStream');
+
+  let skipColorize;
```
anonrig, line 271:
```suggestion
  let skipColorize: boolean;
```

src/workerd/api/node/tests/util-nodejs-test.js
```diff
@@ -25,24 +25,73 @@
...
   const error = new Error('Expected function to be called');
   remainingMustCallErrors.add(error);
-  return (...args) => {
+  return function (...args) {
     remainingMustCallErrors.delete(error);
-    return f(...args);
+    return f.call(this, ...args);
   };
 }
+function commonMustSucceed(fn, exact) {
```
anonrig, line 48: If we're going to implement these methods, it should be a joint effort. I recommend following the existing test syntax/methodologies.

src/node/internal/internal_utils.ts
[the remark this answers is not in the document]
guybedford: The thing is this is an actual `any` case. I think the best way to avoid force casting would be to just say that `ERR_FALSY_VALUE_REJECTION` is any. Have gone ahead and made that change.

src/node/internal/streams_util.d.ts
guybedford: Sure removed. Note that I generally recommend importing `Buffer` personally, since I do still hold out irrational hope we will deprecate it as a global one day!

src/node/util.ts
guybedford: Suggestions welcome for how to manage this case better!

src/node/internal/streams_util.d.ts
guybedford: I couldn't actually find any others in this file.

src/node/internal/streams_util.js
guybedford: Hmm, they have different implementations though:

cloudflare/internal/images-api.ts
```ts
function isReadableStream(obj: unknown): obj is ReadableStream {
  return !!(
    obj &&
    typeof obj === 'object' &&
    'getReader' in obj &&
    typeof obj.getReader === 'function'
  );
}
```

streams_util.js
```js
export function isReadableStream(obj) {
  return !!(
    obj &&
    !isNodeStream(obj) &&
    typeof obj.pipeThrough === 'function' &&
    typeof obj.getReader === 'function' &&
    typeof obj.cancel === 'function'
  );
}
```

do you feel this is a safe change to unify the Cloudflare implementation with?

src/workerd/api/node/tests/util-nodejs-test.js
guybedford: Turns out this function actually wasn't being used.

src/node/internal/streams_util.d.ts
anonrig: Even if we deprecate it from Node.js, it's highly likely that we will always support it due to cloudflare workers backward compatibility guarantees.

src/workerd/api/node/tests/util-nodejs-test.js
guybedford: This tests is more of a test of `process.on('warning')` than a test of deprecated feature, the deprecation is just a message to test the functionality.

If we don't want to ever implement `process.on('warning')` then I think we should remove this rather only then.

src/node/internal/streams_util.js
anonrig: Hm... That might be a breaking change. You're right. Let's keep it as it is.

src/workerd/api/node/tests/util-nodejs-test.js
guybedford: It really depends on whether or not we are going to allow stdio file descriptors or not. What do you think?

src/node/internal/internal_utils.ts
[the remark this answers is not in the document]
anonrig: Then we should change the type of argument of ERR_FALSY_VALUE_REJECTION. Anything to avoid any...

src/node/internal/internal_inspect.ts
guybedford: It's the `__proto__: null` at the top that breaks this. I've replaced it with a ts-error instead.

src/workerd/api/node/tests/util-nodejs-test.js
guybedford: It definitely makes porting a whole lot easier, as there are some fine-grained behaviours here.

Sure I'll set up a separate common.js.

src/node/internal/streams_util.d.ts
guybedford: Yeah, I'm just still hopeful we can encourage this pattern as opposed to discouraging it in the Node.js ecosystem.

src/node/internal/internal_utils.ts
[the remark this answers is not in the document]
guybedford: It's any falsy value though - so the only other thing we could do is explicitly define Falsy:

```ts
type Falsy = false | 0 | -0 | 0n | "" | null | undefined;
```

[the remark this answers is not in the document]
guybedford: I've gone ahead and added Falsy.

src/node/util.ts
guybedford: I added an `isTTYStream` check function here with typing.

src/workerd/api/node/tests/util-nodejs-test.js
guybedford: I'm happy to also do a more involved translation, I'm just thinking maybe it's easier to port and maintain if we do maintain some of these. And this file was already implementing them so I thought that pattern was okay.

jasnell (approved):

src/node/util.ts
```diff
@@ -241,8 +249,90 @@ export function getSystemErrorMessage(): void {
...
+  isTTY: true;
+  getColorDepth(): number;
+};
+
+function isTTYStream(
+  stream: ReadableStream | WritableStream
+): stream is TTYStream {
+  return (
+    typeof stream === 'object' &&
+    (stream as TTYStream).isTTY &&
```
anonrig:
```suggestion
    'isTTY' in stream &&
    stream.isTTY &&
```

src/workerd/api/node/tests/internal/common.js
```diff
@@ -0,0 +1,72 @@
...
+}
+export function mustSucceed(fn, exact) {
+  return mustCall(function (err, ...args) {
+    assert.ifError(err);
+    if (typeof fn === 'function') return fn.apply(this, args);
+  }, exact);
+}
+export function assertCalledMustCalls() {
+  try {
+    for (const error of remainingMustCallErrors) throw error;
```
anonrig: I believe this will only throw error for the first error?

src/workerd/api/node/tests/internal/common.js
```diff
@@ -0,0 +1,72 @@
...
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+// Partial port of Node.js common testing utilities
+import assert from 'node:assert';
+import { inspect } from 'node:util';
+
+const remainingMustCallErrors = new Set();
+export function mustCall(f) {
+  const error = new Error('Expected function to be called');
```
anonrig: This error isn't really helpful if we have 3-4 different functions. Which one threw? I think we should just not do this right now and use the existing way of testing, and think this more in depth once we decide to do it.

src/workerd/api/node/tests/internal/common.js
guybedford: This was already in the utils file. I've left it as-is for now instead of moving it here.

src/workerd/api/node/tests/internal/common.js
guybedford: Again, this was already in the utils file. I've refectored out this move of the code, and instead inlined the common implementations to avoid relying on these primitives. The only new utility function I've added is invalidArgTypeHelper from the common tests now, which is part of the utils test suite only. We can inline this as well if you prefer too.

PR #4230, opened 2025-06-02
Python: Fix packages that use requests with Pyodide 0.27

hoodmane:
We need to make the ssl import raise an import error for these to work correctly.

### DIFF
--- a/src/pyodide/internal/topLevelEntropy/entropy_import_context.py
+++ b/src/pyodide/internal/topLevelEntropy/entropy_import_context.py
@@ -27,6 +27,8 @@ MODULES_TO_PATCH = [
     "tempfile",
     "aiohttp.http_websocket",
     "aiohttp.connector",
+    "urllib3.util.ssl_",
+    "requests.adapters",
     *RUST_PACKAGES,
 ]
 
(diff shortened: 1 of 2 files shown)


src/pyodide/internal/topLevelEntropy/entropy_import_context.py
```diff
@@ -160,18 +162,52 @@ def patched_Random():
...
+def no_ssl():
+    try:
+        f = NoSslFinder()
+        ssl = sys.modules["ssl"]
+        sys.meta_path.insert(0, f)
+        del sys.modules["ssl"]
+        yield
+    finally:
+        sys.meta_path.remove(f)
+        sys.modules["ssl"] = ssl
```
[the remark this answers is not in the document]
hoodmane: Good bot.

anonrig (approved):

src/pyodide/internal/topLevelEntropy/entropy_import_context.py
```diff
@@ -160,18 +162,47 @@ def patched_Random():
...
+        if ssl:
+            sys.modules["ssl"] = ssl
+
+
 @contextmanager
 def aiohttp_connector_context(module):
     # aiohttp will call ssl.create_default_context() at top level which uses entropy. But it has a
     # work around for Pyodide where it skips this. By setting sys.modules["ssl"] = None temporarily,
     # we exercise the Pyodide workaround and so avoid the entropy calls.
     # After, we put the ssl module back to the normal value.
```
dom96: Nit: move this comment to the no_ssl contextmanager

dom96: also, why is this the workaround? Isn't there a more direct way it can detect it's running in Pyodide? Should we actually fix it so it's not relying on this fragile `ssl` module check?

dom96 (approved):

PR #4233, opened 2025-06-03
Add external memory accounting to commonly used stream API objects

jp4a50:


src/workerd/api/streams/readable.c++
```diff
@@ -494,7 +494,11 @@ jsg::Ref<ReadableStream> ReadableStream::constructor(jsg::Lock& js,
       "To use the new ReadableStream() constructor, enable the "
       "streams_enable_constructors compatibility flag. "
       "Refer to the docs for more information: https://developers.cloudflare.com/workers/platform/compatibility-dates/#compatibility-flags");
-  auto stream = js.alloc<ReadableStream>(newReadableStreamJsController());
+  // We account for the memory usage of the ReadableStream and its controller together because their
+  // lifetimes are identical and memory accounting itself has a memory overhead.
+  auto controller = newReadableStreamJsController();
+  auto stream = js.allocAccounted<ReadableStream>(
+      sizeof(ReadableStream) + controller->jsgGetMemorySelfSize(), kj::mv(controller));
```
jasnell, line 501: Eventually I'd think we want `js.alloc<T>` to just automatically apply the accounting rather than having to switch everything to a new method. I wonder if we can have `js.alloc<T>` check for a method on `T` that provides the size and just avoiding adding this `allocAllocated` now..

src/workerd/api/streams/standard.c++
```diff
@@ -2507,14 +2508,21 @@ void ReadableStreamJsController::setup(jsg::Lock& js,
     JSG_REQUIRE(
         autoAllocateChunkSize > 0, TypeError, "The autoAllocateChunkSize option cannot be zero.");
 
-    state = kj::heap<ByteReadable>(controller.addRef(), *this, autoAllocateChunkSize);
+    // We account for the memory usage of the ByteReadable and its controller together because
+    // their lifetimes are identical (in practice) and memory accounting itself has a memory
+    // overhead. The same applies to ValueReadable below.
+    state = kj::heap<ByteReadable>(controller.addRef(), *this, autoAllocateChunkSize)
+                .attach(js.getExternalMemoryAdjustment(
+                    sizeof(ByteReadable) + sizeof(ReadableByteStreamController)));
```
npaun, line 2516: Just so I understand: This is a fixed size, known at compile-time. Is this change to keep track of the overhead of creating lots of streams?

jasnell: The point is to associate the allocated memory with the isolate so that it counts towards the heap isolate limit.

npaun: James explained this to me in chat.

npaun (approved):

src/workerd/api/streams/readable.c++
jp4a50: Yeah, as I discussed with you a couple of weeks ago, I want to eventually adopt a generalized solution but in the mean time this is a pretty simple change which should have a big impact on the immediate issues we are facing.

PR #4235, opened 2025-06-03
Complete sync and async fs access and exists tests

jasnell:
Couple of cleanups and finishes impl of access and exists APIs with full tests.

Will be working on the completion of small sets of APIs in batches in separate PRs.


src/node/internal/internal_fs_utils.ts
```diff
@@ -140,6 +140,51 @@ export const kWriteFileMaxChunkSize = 512 * 1024;
...
+  mode: number | undefined,
+  type: 'copyFile' | 'access' = 'access'
+): number {
+  let min = kMinimumAccessMode;
+  let max = kMaximumAccessMode;
+  let def = F_OK;
+  if (type === 'copyFile') {
+    min = kMinimumCopyMode;
+    max = kMaximumCopyMode;
+    def = mode || kDefaultCopyMode;
```
guybedford, line 179: Just curious about this - from my best interpretation, `kMinimumCopyMode` and `kMinimumAccesMode` are both `0` (`F_OK === 0`), so it wasn't clear to me what this is achieving in the min case here.

src/node/internal/internal_errors.ts
```diff
@@ -836,11 +827,13 @@ export class NodeSyscallError extends NodeError {
   }
 }
 
-export class ERR_ENOENT extends NodeFsError {
+export class ERR_ENOENT extends NodeSyscallError {
+  path: string;
   constructor(path: string, options: { syscall: string }) {
     super('ENOENT', `No such file or directory: ${path}`, options.syscall);
     this.code = 'ENOENT';
```
guybedford: This is already handled by the parent `super('ENOENT')`.

guybedford (approved):

src/node/internal/internal_fs_utils.ts
jasnell: @anonrig since he wrote this initially. I really didn't go through this bit too much other than to verify it does the correct thing. It can likely be further optimized but I'll do that in another cleanup PR once the main functionality is all fully impl'd and verified.

PR #4238, opened 2025-06-03
add container service for local development

anonrig:
> Work in progress. Not ready to review yet.


src/rust/container/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    Ok(socket)
+}
+
+pub async fn start(
+    container_name: &str,
+    entrypoint: Vec<&str>,
+    env: Vec<&str>,
+) -> Result<String, ContainerError> {
+    // TODO(soon): Persist connection.
+    let socket = connect()?;
```
vicb: It looks like the lib provides a higher level API [start_container](https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.start_container) ([examples](https://docs.rs/bollard/latest/bollard/#examples))

Could it be used instead?

src/rust/container/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+// [x] - kj::Promise<void> start(StartContext context) override;
+// [ ] - kj::Promise<void> monitor(MonitorContext context) override;
+// [x] - kj::Promise<void> destroy(DestroyContext context) override;
+// [x] - kj::Promise<void> signal(SignalContext context) override;
+// [ ] - kj::Promise<void> getTcpPort(GetTcpPortContext context) override;
+// [ ] - kj::Promise<void> listenTcp(ListenTcpContext context) override;
+
+pub fn connect() -> Result<Docker, ContainerError> {
+    let addr = "/var/run/docker.sock";
+    let socket = Docker::connect_with_unix(addr, 30, API_DEFAULT_VERSION)
```
vicb: suggestion
```suggestion
    let docker = Docker::connect_with_unix(addr, 30, API_DEFAULT_VERSION)
```

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,9 @@ struct Worker {
     # workerd uses SQLite to back all Durable Objects, but the SQL API is hidden by default to
     # emulate behavior of traditional DO namespaces on Cloudflare that aren't SQLite-backed. This
     # flag should be enabled when testing code that will run on a SQLite-backed namespace.
+
+    enableDocker @5 :Bool;
```
vicb: Maybe group the config into a `struct DockerConfig` ?
The next thing that you will need to pass is the container name and there should be more config later (List of ports, max instances, ...)

src/rust/container/lib.rs
vicb: It looks like using [connect_with_socket](https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.connect_with_socket) would support both windows and linux?

src/workerd/server/workerd.capnp
```diff
@@ -651,6 +654,10 @@ struct Worker {
   streamingTails @15 :List(ServiceDesignator);
   # List of streaming tail worker services that should receive tail events for this worker.
   # NOTE: This will be deleted in a future refactor, do not depend on this.
+
+  dockerSocketPath @16 :Text;
```
vicb: Note for later:

Maybe we want to make this a `DockerConnection` struct instead of a simple address at some point.

There are [multiple ways](https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.connect_network) to connect to docker, we could also have a version, timeout, ...

Maybe add a todo comment for later. Or create a `struct DockerConnection` already with a single `address: Text` field fow now, that might be better for future extensions.

src/rust/container/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    container_name: &str,
+    entrypoint: Vec<&str>,
+    env: Vec<&str>,
+) -> Result<String, ContainerError> {
+    // TODO(soon): Persist connection.
+    let socket = connect()?;
+    let create_config = CreateExecOptions::default();
+    create_config.cmd = Some(entrypoint);
+    create_config.env = Some(env);
+    // TODO(soon): What to do with enableInternet?
```
emily-shen: discussed this with @mikenomitch and i don't think we need to do anything with this for local dev, at least not urgently. will be a bit difficult to have `enableInternet = false` while still having port mapping to the host, afaict.

deps/rust/cargo.bzl
```diff
@@ -28,4 +28,5 @@ PACKAGES = WORKERD_CXX_PACKAGES | {
     # tokio is huge, let's enable only features when we actually need them.
     "tokio": crate.spec(version = "1", default_features = False, features = ["net", "rt", "rt-multi-thread", "time"]),
     "tracing": crate.spec(version = "0", default_features = False, features = ["std"]),
+    "bollard": crate.spec(version = "0"),
```
danlapid, line 33:
```suggestion
    "bollard": crate.spec(version = "0", default_features = False),
```

This should reduce the dependencies surface and I'm pretty sure you're not uing http or pipe features considering our approach depends on just a custom transport

src/rust/container/lib.rs
danlapid: We won't be able to actually use any of those APIs anyway, this is still in draft stage, we would have to create our own custom transport in order to use KJ's io provider rather than tokio.

src/rust/container/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub fn connect() -> Result<Docker, ContainerError> {
+    let addr = "/var/run/docker.sock";
+    let socket = Docker::connect_with_unix(addr, 30, API_DEFAULT_VERSION)
+        .map_err(|e| ContainerError::ConnectionError(e.to_string()))?;
+    Ok(socket)
+}
+
+pub async fn start(
+    container_name: &str,
```
emily-shen, line 81: should we label the started container with the instance id, to enable reconnecting to the same container?

src/rust/container/lib.rs
anonrig: Thanks! I'll add a comment for future reference.

src/workerd/server/workerd.capnp
anonrig: Yes, that's really good! I'm changing this as you recommended.

src/workerd/server/workerd.capnp
anonrig: @mikenomitch what options do we need to support? Isn't the container name dependent to the durable object id?

src/rust/container/lib.rs
anonrig: I looked at it once again, and start_exec() isn't the correct command.

Start container requires building an image. How are we going to build an image? cc @mikenomitch

src/rust/container/lib.rs
anonrig: instance id, meaning durable object id? I'm not following.

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,14 @@ struct Worker {
...
     # emulate behavior of traditional DO namespaces on Cloudflare that aren't SQLite-backed. This
     # flag should be enabled when testing code that will run on a SQLite-backed namespace.
+
+    docker @5 :DockerOptions;
+    # Optional configuration options for local docker container interaction.
+
+    struct DockerOptions {
+      containerName @0 :Text;
+      # Name of the docker container associated with the current DO namespace.
+    }
```
jasnell: If this is only going to be used here, then this can probably be simplified into a Group rather than struct. But, I'm not the expert on capnp schemas so I'll defer to @kentonv on what's best for the structure here.

src/workerd/io/container-client.h
```diff
@@ -0,0 +1,37 @@
...
+#include <workerd/rust/container/lib.rs.h>
+#include <workerd/rust/cxx-integration/cxx-bridge.h>
+
+#include <stdlib.h>
+
+#include <kj/async-io.h>
+#include <kj/debug.h>
+
+namespace workerd::io {
+struct RustAsyncStream final: public kj::AsyncIoStream {
```
jasnell: heh... what's that you always ~~nag me about~~ remind me to do ;-) ... some code comments in here would be ideal ;-)

src/workerd/io/container-client-test.wd-test
```diff
@@ -0,0 +1,24 @@
...
+        modules = [
+          (name = "worker", esModule = embed "container-client-test.js")
+        ],
+        compatibilityDate = "2025-01-09",
+        compatibilityFlags = ["nodejs_compat"],
+        docker = (address = "/var/run/docker.sock"),
+        durableObjectNamespaces = [
+          ( className = "DurableObjectExample",
+            uniqueKey = "210bd0cbd803ef7883a1ee9d86cce06e",
+            docker = (containerName = "docker-example") ),
```
jasnell: Would definitely want @kentonv's opinion on this structure. Would it actually make more sense to define docker as a kind of service appearing in the `services` collection? ... which would move it into the list alongside the `name = "internet"` and `name = "TEST_TMPDIR"` items below... then then linking this worker to the container can just be a service reference. 

Also, I'm not sure we should (or can) use the name `docker` in here. Docker is a trademarked name and its probably a bad idea to tie ourselves to it. Something generic like `container` is likely better in here.

src/workerd/io/container-client-test.js
```diff
@@ -0,0 +1,7 @@
+export class DurableObjectExample {}
+
+export const test = {
```
jasnell: I assume this is incomplete as it is currently not doing anything :-)

jasnell:
image
 This is payback for me making you review massive PRs, isn't it? ;-)

src/rust/container/lib.rs
jasnell: Is the `Docker` in this a name we chose or is it a third party dependency? If it's a name we control we should go with something more generic. If it's third party dependency then no worry.

anonrig: Docker class comes from the third party library.

src/workerd/server/workerd.capnp
vicb: I see a few problems with a group:
- you would have to add an explicit "enabled"
- the field numbers in groups are from the parent struct
- also I wonder if the docker options really belongs here or to a separate map of DO namespaces to Options - I don't think it's required now but it would be easier to migrate.

Maybe for the last point I would move the struct at top level instead of nested in DO, close to `DockerConnectionOptions`.

src/rust/container/lib.rs
emily-shen: just checking we're all on the same page so i can miniflare set up properly:

so miniflare will build the image and pass workerd the image tag.

then you use that with create_container and start_container
https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.create_container
https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.start_container

looking at those docs, it looks like the options will be generated as sometiing like `bollard::models::ContainerCreateBody` etc. from the openapi spec here:
https://docs.docker.com/reference/api/engine/version/v1.49/#tag/Container/operation/ContainerCreate etc.
so even if its not actually using http the functions here will have the same options I guess? so you can set entrypoint, env vars and port mapping with that.

the container name can be anything, but i guess for readability purposes, can it be whatever gets passed into `DO. idFromString()` or something that makes sense to the end user? 

is that what you were thinking of? thanks for all the work btw 🧡

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn signal(
+        &mut self,
+        params: container::SignalParams,
+        _: container::SignalResults,
+    ) -> Promise<(), capnp::Error> {
+        let container_name = self.container_name.clone();
+        let docker = self.docker.clone();
+        Promise::from_future(async move {
+            let signo = params.get()?.get_signo();
+            let signal_name = match signo {
```
mikenomitch: Looks like there are more standard Linux signals - any reason to not support them all here? -> Does docker not support them?

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+        _: container::MonitorResults,
+    ) -> Promise<(), capnp::Error> {
+        Promise::from_future(async move {
+            Err(capnp::Error::unimplemented(
+                "monitor is not implemented".to_string(),
+            ))
+        })
+    }
+
+    fn listen_tcp(
```
mikenomitch, line 329: Confirming that we don't need this one for now

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    fn monitor(
+        &mut self,
+        _: container::MonitorParams,
+        _: container::MonitorResults,
+    ) -> Promise<(), capnp::Error> {
+        Promise::from_future(async move {
+            Err(capnp::Error::unimplemented(
+                "monitor is not implemented".to_string(),
```
emily-shen: probably use this? https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.wait_container
basically resolve if it exits with 0, reject if the container exits with a non-zero code

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+    pub fn is_write_disconnected(&self) -> bool {
+        self.r#impl
+            .write_shutdown
+            .load(std::sync::atomic::Ordering::Relaxed)
+    }
+}
+
+struct Impl {
+    sender: mpsc::SyncSender<Vec<u8>>,
+    write_shutdown: Arc<std::sync::atomic::AtomicBool>,
```
mikea: this doesn't need to be Arc. Simple AtomicBool would suffice.

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+
+struct Server {
+    docker: Arc<Docker>,
+    container_name: String,
+}
+
+impl Server {
+    fn connect(address: &str, container_name: &str) -> Result<Self, ContainerError> {
+        let docker = Docker::connect_with_socket(address, 120, API_DEFAULT_VERSION)
+            .map_err(|e| ContainerError::ConnectionError(e.to_string()))?;
```
mikea: the whole point of `thiserror` is not to write map_err. This needs to be `connect_with_socket(...)?;`
And the corresponding branch in ContainerError should have from annotation.

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+
+impl AsyncRead for AsyncReader {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        match self.receiver.try_recv() {
+            Ok(data) => {
+                let len = std::cmp::min(buf.len(), data.len());
```
mikea: hm... what will happen to the rest of data? Looks like AsyncReader should have something like `unread: Vec<u8>`? Maybe https://docs.rs/futures/latest/futures/io/struct.BufReader.html source code could be useful here?

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+            ))
+        })
+    }
+}
+
+struct AsyncReader {
+    receiver: mpsc::Receiver<Vec<u8>>,
+}
+
+impl AsyncRead for AsyncReader {
```
mikea: I suggest to create io.rs mod and move AsyncReader & AsyncWriter there. And let's rename them to `MpscReader`/Writer or something?

src/workerd/io/container-client.h
```diff
@@ -0,0 +1,62 @@
...
+    service->shutdown_write();
+  }
+
+  kj::Promise<size_t> tryRead(void* buffer, size_t minBytes, size_t maxBytes) override {
+    // For now, return 0 bytes read (EOF) since this is primarily a write-only stream
+    // In a full implementation, this would read from the container's output
+    (void)buffer;
+    (void)minBytes;
+    (void)maxBytes;
+    return size_t(0);
```
mikea: >since this is primarily a write-only stream
this is bidirectional stream, and we need to plumb messages from MessageCallback to here.

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+        _results: container::ListenTcpResults,
+    ) -> Promise<(), capnp::Error> {
+        Promise::from_future(async move {
+            Err(capnp::Error::unimplemented(
+                "listen_tcp is not implemented".to_string(),
+            ))
+        })
+    }
+
+    fn get_tcp_port(
```
mikenomitch, line 342: Top level interface for this looks like this:

```
  const port = this.ctx.container.getTcpPort(8080);
  const res = await port.fetch("http://container/set-state", { body: initialState, method: 'POST' })
```

I'm not sure exactly what needs to be returned/done at this level to give a fetcher that has a port attached.

I think the challenge here is going to be mapping the given port (8080 in example above) to a local port (defined by Docker?).

So I might call from one DO instance (A):
const portA = this.ctx.container.getTcpPort(8080);

And then from another DO instance (B):
const portB = this.ctx.container.getTcpPort(8080);

And both of these are getting the same image, but a new container, and each container will be mapping some local port to 8080. (maybe port 4000 goes to 8080 for instance A and 4001 goes to 8080 for instance B)

So I think that we need to maintain some mapping of these things when we boot? Or maybe (ideally?) we can get it via some Docker API just when we request the port? Basically we need to select the correct local port and return the right stuff to enable the fetcher above.

I see this function in cloudchamberd - https://gitlab.cfdata.org/cloudflare/cc/cloudchamber/-/blob/master/go/cmd/cloudchamberd/internal/services/durableobject/binding.go#L428 - but TBH I have a hard time translating that to this RPC call specifically

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+                15 => "SIGTERM",
+                18 => "SIGCONT",
+                19 => "SIGSTOP",
+                20 => "SIGTSTP",
+                _ => "SIGTERM", // Default fallback
+            };
+
+            let options = StopContainerOptionsBuilder::default().signal(signal_name);
+            docker
+                .stop_container(&container_name, Some(options.build()))
```
mikenomitch: I assume `stop_container` is just kind of a misnomer from the library we're using?

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+            let env = params
+                .get_environment_variables()?
+                .iter()
+                .map(|s| s.unwrap().to_string().unwrap())
+                .collect::<Vec<String>>();
+            let mut create_config = CreateExecOptions::default();
+            create_config.cmd = Some(entrypoint);
+            create_config.env = Some(env);
+            let result = docker
+                .create_exec(&container_name, create_config)
```
mikenomitch: I believe we want this command: https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.create_container

Since we wont have the container running yet

src/rust/container/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+
+            let mut builder = results.get();
+            if let Some(state) = inspect.state {
+                builder.set_running(state.running.unwrap_or(false));
+            }
+            Ok(())
+        })
+    }
+
+    fn monitor(
```
mikenomitch, line 296: I believe this just blocks until the container exits then surfaces the error if the error code is not 0

src/rust/container/lib.rs
anonrig: I've updated! Thanks.

src/rust/container/lib.rs
anonrig: This was really helpful. Thank you!

src/rust/container/lib.rs
anonrig: Thanks! Added a comment.

src/rust/container/lib.rs
anonrig: Yes, I've updated the implementation. Thanks for the review.

src/rust/container/lib.rs
anonrig: Yes. @emily-shen made an awesome recommendation and I've updated the implementation. Can you double-check if it's what we want?

src/rust/container/lib.rs
anonrig: What should destroy do? should it also remove the container?

anonrig: I've updated the implementation to also call removeContainer(). Let me know if it's incorrect.

src/rust/container/lib.rs
anonrig: I'm not sure what's the best way of converting a signo to signal (string). @mikea any recommendations?

mikenomitch: I think its like ~30 values or so, so hard coded hashmap could work? (like this but bigger)

src/workerd/io/container-client-test.wd-test
anonrig: > Also, I'm not sure we should (or can) use the name docker in here. Docker is a trademarked name and its probably a bad idea to tie ourselves to it. Something generic like container is likely better in here.

I've updated the implementation according to this. Thanks!

src/rust/container/lib.rs
anonrig: I think I can do something like this and avoid this hashmap:

```rust
use std::ffi::CStr;
use std::os::raw::c_char;

extern "C" {
    fn strsignal(sig: i32) -> *const c_char;
}

fn signo_to_string_libc(signo: i32) -> Option<String> {
    unsafe {
        let ptr = strsignal(signo);
        if ptr.is_null() {
            None
        } else {
            CStr::from_ptr(ptr).to_string_lossy().into_owned().into()
        }
    }
}
```

src/rust/container/lib.rs
mikenomitch: LGTM except probably makes sense to match the error message string?

https://gitlab.cfdata.org/cloudflare/cc/cloudchamber/-/blob/master/go/cmd/cloudchamberd/internal/services/durableobject/binding.go#L343

src/workerd/server/server.c++
```diff
@@ -2267,10 +2276,25 @@ class Server::WorkerService final: public Service,
...
 
+        kj::Maybe<rpc::Container::Client> containerClient = kj::none;
+
+        KJ_IF_SOME(config, containerOptions) {
+          KJ_IF_SOME(path, service.containerAddress) {
+            KJ_REQUIRE(config.hasName(), "Container name is required");
+            auto container_name = config.getName();
+            auto stream = io::createContainerStream(path, container_name);
+            capnp::TwoPartyClient client(*stream);
+            containerClient = client.bootstrap().castAs<rpc::Container>();
```
danlapid:
```suggestion
            containerClient = client.bootstrap().castAs<rpc::Container>().attach(kj::mv(stream));
```

danlapid: Or maybe you'll need a membrane?

src/workerd/io/container-client.h
```diff
@@ -0,0 +1,88 @@
...
+class ContainerStreamSharedState {
+ public:
+  ContainerStreamSharedState();
+
+  void enqueueMessage(::rust::Slice<const uint8_t> message);
+  kj::Maybe<kj::Array<kj::byte>> dequeueMessage();
+  kj::Promise<void> waitForMessage();
+
+ private:
+  kj::MutexGuarded<std::queue<kj::Array<kj::byte>>> messageQueue;
```
danlapid: is this shared?

kentonv:
Why does this PR contain a whole lot of Rust code and Rust build file updates that seem unrelated? Can this be cleaned up to something reviewable?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,92 @@
...
+
+  return sharedState->waitForMessage().then(
+      [this, buffer, minBytes, maxBytes]() -> kj::Promise<size_t> {
+    return tryRead(buffer, minBytes, maxBytes);
+  });
+}
+
+kj::Own<ContainerAsyncStream> createContainerStream(
+    kj::StringPtr address, kj::StringPtr containerName) {
+  auto sharedState = kj::heap<ContainerStreamSharedState>();
```
danlapid: why not refcounted?
`kj::Rc<T>` + `kj::rc<T>(...)`

src/workerd/io/container-client.h
danlapid: yes this is, it is shared between the rust tokio thread and the c++ worker thread.

src/workerd/io/container-client.h
```diff
@@ -0,0 +1,88 @@
...
+    if (service->is_write_disconnected()) {
+      return kj::READY_NOW;
+    }
+    return kj::NEVER_DONE;
+  }
+
+ private:
+  ::rust::Box<rust::container::ContainerService> service;
+  kj::Own<ContainerStreamSharedState> sharedState;
+  kj::Maybe<kj::Array<kj::byte>> currentMessage;
```
danlapid: Why do you need currentMessage AND messageQueue?

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,14 @@ struct Worker {
     # workerd uses SQLite to back all Durable Objects, but the SQL API is hidden by default to
     # emulate behavior of traditional DO namespaces on Cloudflare that aren't SQLite-backed. This
     # flag should be enabled when testing code that will run on a SQLite-backed namespace.
+
+    container @5 :ContainerOptions;
+    # Optional configuration options for local development container interaction.
+
+    struct ContainerOptions {
+      name @0 :Text;
+      # Name of the container associated with the current DO namespace.
```
kentonv, line 610: How dose this work? Each DO instance gets its own container. Is this the name of the container _image_? This really needs a lot more commentary.

src/rust/container/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+        })
+    }
+
+    /// TODO(soon): Implement this.
+    fn get_tcp_port(
+        &mut self,
+        _params: container::GetTcpPortParams,
+        _results: container::GetTcpPortResults,
+    ) -> Promise<(), capnp::Error> {
+        Promise::from_future(async move {
```
danlapid, line 347: You need to implement a `rpc::Container::Port::Server` and return an object of that kind.
That object implements `connect` which pumps bytes back and forth between the container tcp port and the incoming bytestream.

danlapid: See example in fake-container-service.c++ internally.

kentonv:
I think we need to back up and write out a document explaining what's going on here. The design here doesn't appear to match any design we've discussed so far and I have no idea what's going on.

src/rust/container/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+        let docker = Docker::connect_with_socket(address, 120, API_DEFAULT_VERSION)?;
+        Ok(Server {
+            docker: docker.into(),
+            container_name: container_name.to_owned(),
+        })
+    }
+}
+
+impl container::Server for Server {
+    fn destroy(
```
mikenomitch, line 184: I believe what happens here is that we send a SIGTERM, wait for 5 min for the container to stop, and if it is not stopped then we send a SIGKILL.

So I think if this were changed to sending a SIGTERM first, then we waited and checked for status and then called `remove_container` conditionally it'd be the right logic.

mikenomitch: I'm not 100% positive that this happens on the `destroy` RPC command though (versus just when we update)... 

@gabivlj - I tracked things to here at least: https://gitlab.cfdata.org/cloudflare/cc/cloudchamber/-/blob/master/go/cmd/cloudchamberd/internal/services/durableobject/binding.go#L362

Does the `container.sidecar.Stop` command do the two signals, or will this just immediately SIGKILL in this case?

src/rust/container/lib.rs
mikenomitch: Commented inline on destroy. - I think this comment was on `signal` not `destroy` FWIW! Hence me asking if stop_container was a misnomer.

src/rust/container/lib.rs
mikenomitch: Oh... maybe stop_container has the behavior we want? (https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.stop_container) and it includes a timeout to wait before it sigkills (https://docs.rs/bollard/latest/bollard/container/struct.StopContainerOptions.html).

So I think:
- If `destroy` is meant to immediately destory, then we call kill_container
- If `destory` does in fact do the timeout behavior, we use stop_container and set the timeout

gabivlj: When `signal()` is called, we just relay that signal to the runtime. When `destroy()` is called, we send a SIGKILL.

src/workerd/server/server.c++
anonrig:
```
src/workerd/server/server.c++:2287:75: error: no member named 'attach' in 'workerd::rpc::Container::Client'                                 2287 |             containerClient = client.bootstrap().castAs<rpc::Container>().attach(kj::mv(stream));
      |                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
      ```

**Comment by anonrig** [comment]
> Why does this PR contain a whole lot of Rust code and Rust build file updates that seem unrelated? Can this be cleaned up to something reviewable?

@kentonv Adding a rust crate as a dependency is causing these changes. I don't know if it's possible but anybody knows a way happy to do it. cc @danlapid @mikea @fhanau

**Comment by kentonv** [comment]
Can you please move all the Rust dependency management into a separate commit? As is my UI is slowing to a crawl and github doesn't seem to want to display most of the files because there's too much stuff in the diff.

**Comment by anonrig** [comment]
> Can you please move all the Rust dependency management into a separate commit? As is my UI is slowing to a crawl and github doesn't seem to want to display most of the files because there's too much stuff in the diff.

@kentonv Yes, I'll do it. As soon as possible.

File: src/workerd/io/container-client.h [hunk 31]
```diff
@@ -0,0 +1,86 @@
...
+// ContainerAsyncStream provides bidirectional communication with a container.
+//
+// This stream implements both reading and writing:
+// - Writing sends data to the container via the Rust service
+// - Reading receives messages from the container via MessageCallback queuing
+//
+// The stream uses shared state to coordinate between the MessageCallback
+// (which receives messages from the container asynchronously) and the
+// tryRead() method (which provides those messages to the C++ side).
+struct ContainerAsyncStream final: public kj::AsyncIoStream {
```
kentonv, line 32: I read the comment but I don't understand the purpose of this class. Why do containers need a special kind of stream?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,80 @@
...
+  auto lockedQueue = messageQueue.lockExclusive();
+  for (auto& byte : message) {
+    lockedQueue->push(byte);
+  }
+
+  auto lockedWaiter = readWaiter.lockExclusive();
+  KJ_IF_SOME(fulfiller, *lockedWaiter) {
+    fulfiller->fulfill();
+    *lockedWaiter = kj::none;
+  }
```
jasnell: Should both locks actually be held simultaneously? Or should each of these be in their own blocks?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,80 @@
...
+}
+
+kj::Maybe<size_t> ContainerStreamSharedState::tryRead(void* buffer, size_t minBytes, size_t maxBytes) {
+  auto lockedQueue = messageQueue.lockExclusive();
+  if (lockedQueue->empty()) {
+    return kj::none;
+  }
+
+  size_t min = kj::min(lockedQueue->size(), maxBytes);
+  KJ_REQUIRE(min > 0, "Should never happen");
```
jasnell, line 38: For these, use `KJ_ASSERT` rather than `KJ_REQUIRE`

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,80 @@
...
+  memcpy(buffer, &lockedQueue->front(), min);
```
jasnell, line 39: Use `kj::ArrayPtr` copyFrom instead for more safety

src/rust/container/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+                .map(|s| s.unwrap().to_string().unwrap())
+                .collect::<Vec<String>>();
+            let env = params
+                .get_environment_variables()?
+                .iter()
+                .map(|s| s.unwrap().to_string().unwrap())
+                .collect::<Vec<String>>();
+
+            let config = ContainerCreateBody {
+                image: Some(container_name.clone()),
```
emily-shen, line 252: it looks like the image ref is the same as container_name? i don't think this is right, because we will have multiple instances running from the same image.

miniflare will give workerd an _image_ name:tag, which should be used here to create the container from the right image. then the _container_ name can be whatever you want. Something readable for the end user preferably, idk if you have access here to the string the DO instance id is created from, but that would be ideal.

then all the other methods here should use that container name, not the image name that miniflare passed in.

maybe we can rename the config option in workerd.capnp to image_tag for clarity?

src/rust/container/io.rs
```diff
@@ -0,0 +1,110 @@
...
+use std::ffi::CStr;
+use std::pin::Pin;
+use std::task::Context;
+use std::task::Poll;
+
+use futures::AsyncRead;
+use futures::AsyncWrite;
+use tokio::sync::mpsc;
+
+unsafe extern "C" {
```
mikea, line 11: @anonrig let restore the original switch. this wouldn't work on win?

PR #4244, opened 2025-06-04
Pass a jsg::Lock to wrap and unwrap so we can get the isolate from it  ￼

erikcorry:


src/workerd/io/promise-wrapper.h
```diff
@@ -30,23 +30,23 @@ class PromiseWrapper {
   // For some reason, this wasn't needed for jsg::V8Ref<T>...
 
   template <typename T>
-  v8::Local<v8::Promise> wrap(v8::Local<v8::Context> context,
+  v8::Local<v8::Promise> wrap(jsg::Lock& js,
+      v8::Local<v8::Context> context,
```
jasnell, line 34: As discussed in the chat, let's change this to pass only the `jsg::Lock&` and not the Lock and the Context.

anonrig: Creating a jsg::Lock& takes 10% of an simple request. It's slow. We can directly pass the isolate rather than the jsg::Lock and we can remove context since context can be accessed through isolate.GetCurrentContext() method.

kentonv: Yeah if we didn't already have the `jsg::Lock&` at all the call sites I would tend to agree we should pass the `v8::Isolate*` instead, and only fetch the lock when needed.

If we have the context at the call sites then we should probably keep passing it too, as it is very commonly needed, and calling `GetCurrentContext()` may take longer than you'd think.

erikcorry: jsg::Lock::from gets fully inlined to

 25e9427:       4c 8b b0 58 02 00 00    mov    0x258(%rax),%r14

when you already have the isolate.  So I'm not too worried about that.

For example:

jsg::Ref<JsRpcPromise> JsRpcProperty::call(const v8::FunctionCallbackInfo<v8::Value>& args) {
  jsg::Lock& js = jsg::Lock::from(args.GetIsolate());

turns into

 25e9420:       48 8b 02                mov    (%rdx),%rax
 25e9423:       48 8b 40 08             mov    0x8(%rax),%rax
 25e9427:       4c 8b b0 58 02 00 00    mov    0x258(%rax),%r14

The first line gets the implicit_args_ from the args.
The second line gets the isolate from the implicit_args_ array.
The third gets the lock using the isolate.

The context is a Local that uses the HandleScope, so GetCurrentContext (and lock.v8Context()) are likely more expensive.  I'll revert the change that takes that out of the arguments.

erikcorry: (Just writing notes to myself.)

The change to remove the context argument to wrap doesn't work currently because the context is not on the isolate during setup.  Worker::setupContext() uses wrapSimpleFunction, which takes a context that it needs to pass to wrap() because it can't be retrieved from the isolate yet.

But I think we agreed to back out this change anyway.

Strack trace is:

1        0x000000000305ae0e <workerd::jsg::Lock::v8Context()+158>
2        0x0000000003a9068d <workerd::jsg::FunctionWrapper<workerd::server::(anonymous namespace)::JsgWorkerdIsolate_TypeWrapper>::wrap<void (v8::FunctionCallbackInfo<v8::Value> const&)>(workerd::jsg::Lock&, kj::Maybe<v8::Local<v8::Object> >, workerd::jsg::Function<void (v8::FunctionCallbackInfo<v8::Value> const&)>&&)::{lambda(workerd::jsg::Ref<workerd::jsg::WrappableFunction<void (v8::FunctionCallbackInfo<v8::Value> const&)> >&)#1}::operator()(workerd::jsg::Ref<workerd::jsg::WrappableFunction<void (v8::FunctionCallbackInfo<v8::Value> const&)> >&) const+253>
3         0x0000000003a9047a <v8::Local<v8::Function> workerd::jsg::Function<void (v8::FunctionCallbackInfo<v8::Value> const&)>::getOrCreateHandle<workerd::jsg::FunctionWrapper<workerd::server::(anonymous namespace)::JsgWorkerdIsolate_TypeWrapper>::wrap<void (v8::FunctionCallbackInfo<v8::Value> const&)>(workerd::jsg::Lock&, kj::Maybe<v8::Local<v8::Object> >, workerd::jsg::Function<void (v8::FunctionCallbackInfo<v8::Value> const&)>&&)::{lambda(workerd::jsg::Ref<workerd::jsg::WrappableFunction<void (v8::FunctionCallbackInfo<v8::Value> const&)> >&)#1}>(v8::Isolate*, workerd::jsg::FunctionWrapper<workerd::server::(anonymous namespace)::JsgWorkerdIsolate_TypeWrapper>::wrap<void (v8::FunctionCallbackInfo<v8::Value> const&)>(workerd::jsg::Lock&, kj::Maybe<v8::Local<v8::Object> >, workerd::jsg::Function<void (v8::FunctionCallbackInfo<v8::Value> const&)>&&)::{lambda(workerd::jsg::Ref<workerd::jsg::WrappableFunction<void (v8::FunctionCallbackInfo<v8::Value> const&)> >&)#1}&&)+90>
4        0x0000000003a903f4 <v8::Local<v8::Function> workerd::jsg::FunctionWrapper<workerd::server::(anonymous namespace)::JsgWorkerdIsolate_TypeWrapper>::wrap<void (v8::FunctionCallbackInfo<v8::Value> const&)>(workerd::jsg::Lock&, kj::Maybe<v8::Local<v8::Object> >, workerd::jsg::Function<void (v8::FunctionCallbackInfo<v8::Value> const&)>&&)+68>
5        0x0000000003a8eeb5 <workerd::jsg::Isolate<workerd::server::(anonymous namespace)::JsgWorkerdIsolate_TypeWrapper>::Lock::wrapSimpleFunction(v8::Local<v8::Context>, workerd::jsg::Function<void (v8::FunctionCallbackInfo<v8::Value> const&)>)+101>
6        0x00000000040259a7 <workerd::Worker::setupContext(workerd::jsg::Lock&, v8::Local<v8::Context>, workerd::Worker::ConsoleMode)::$_0::operator()(char const*, capnp::schemas::Level_a8b8c4ab94a9dea7) const+1175>
7         0x00000000040254a0 <workerd::Worker::setupContext(workerd::jsg::Lock&, v8::Local<v8::Context>, workerd::Worker::ConsoleMode)+576>
8        0x000000000404c282 <workerd::Worker::Isolate::Impl::Lock::setupContext(v8::Local<v8::Context>)+258>
9        0x000000000404be6b <workerd::Worker::Script::Script(kj::Own<workerd::Worker::Isolate const, decltype(nullptr)>, kj::StringPtr, kj::OneOf<workerd::Worker::Script::ScriptSource, workerd::Worker::Script::ModulesSource>, workerd::IsolateObserver::StartType, bool, kj::Maybe<workerd::Worker::ValidationErrorReporter&>, kj::Maybe<kj::Own<workerd::api::pyodide::ArtifactBundler_State, decltype(nullptr)> >)::$_0::operator()(workerd::jsg::V8StackScope&) const::{lambda()#2}::operator()() const+443>

PR #4245, opened 2025-06-04
Create type for a Hello World Binding

penalosa:
Adds a type for a Hello World Binding, which will serve as an explanatory example. It will not work in production and should not be used


jasnell (approved):

swizzlr:
Is this meant to be landing in my generated workers `worker-configuration.d.ts`? Just showed up in wrangler 4.24.0

PR #4246, opened 2025-06-04
Pass a jsg::Lock to wrap and unwrap so we can get the isolate from it

danlapid:


src/workerd/jsg/promise.h
```diff
@@ -675,10 +674,10 @@ class PromiseWrapper {
  private:
   const JsgConfig config;
 
-  static bool isThenable(v8::Local<v8::Context> context, v8::Local<v8::Value> handle) {
+  static bool isThenable(Lock& js, v8::Local<v8::Context> context, v8::Local<v8::Value> handle) {
```
anonrig, line 677: Creating a jsg::Lock& takes 10% of an simple request. It's slow. We can directly pass the isolate rather than the jsg::Lock and we can remove context since context can be accessed through isolate.GetCurrentContext() method.

erikcorry:
> Changes themselves seem simple enough so accepting. But can you explain the context behind why we're doing this?

More context is here: Changes themselves seem simple enough so accepting. But can you explain the context behind why we're doing this?

danlapid:
umm, this was targeted at erik's branch, it should never have targeted master and is not intended to be merged.

PR #4249, opened 2025-06-04
[DRAFT][o11y] Support cfJson for more worker invocation types

fhanau:


fhanau:
Closing this – cfJson should still be available in the STW model, but it is generally only available for fetch so no changes are needed here.

PR #4251, opened 2025-06-04
Add node:fs stat tests

jasnell:


src/node/internal/internal_fs_sync.ts
```diff
@@ -708,15 +713,17 @@ export function statSync(
...
   options: StatOptions = {}
 ): Stats | undefined {
-  validateObject(options, 'options');
-  const { bigint = false, throwIfNoEntry = true } = options;
-  validateBoolean(bigint, 'options.bigint');
-  validateBoolean(throwIfNoEntry, 'options.throwIfNoEntry');
-  const normalizedPath = normalizePath(path);
-  const stat = cffs.stat(normalizedPath, { followSymlinks: true });
+  const {
+    pathOrFd: validatedPath,
```
anonrig, line 717: Any particular reason why validatedPath type is not URL?

anonrig (approved):

src/node/internal/internal_fs_sync.ts
jasnell: It can be number or URL depending on what was passed in to the validator function.

PR #4252, opened 2025-06-04
Separate disabled and omitted tests in WPT reports and coverage

npaun:
This PR splits skipped tests into two categories:

- disabl,edTests: We eventually want to pass these, but right now we can't because either:
   - They crash workerd
   - The harness can't handle them
   - They test a feature we haven't implemented but might implement in the future.
- omittedTests: Do not include in our WPT report or coverage stats, because:
   - They're not reflective of an actual standard (e.g. most tentative tests)
   - They're not relevant server-side
   - We don't intend to support this feature
   - The file isn't a test but is picked up by our globs in bazel anyway
   - etc.

Please let me know if I've misclassified any of our existing tests.


src/wpt/WebCryptoAPI-test.ts
```diff
@@ -213,7 +213,7 @@ export default {
   },
   'idlharness.https.any.js': {
     comment: 'Test file /resources/WebIDLParser.js not found.',
-    skipAllTests: true,
+    skippedTests: 'ALL',
```
anonrig: What is the difference between skipped and excluded tests? It's really ambiguous as a non-native english speaker.

src/wpt/harness/common.ts
```diff
@@ -59,6 +69,10 @@ export class FilterList {
   }
 
   public delete(input: string): boolean {
+    if (this.all) {
```
anonrig: can we use a more verbose variable name? this.all doesn't reflect anything to me (sorry!)

src/wpt/harness/harness.ts
```diff
@@ -69,24 +69,24 @@ type CommonOptions = {
...
-      skipAllTests: true;
-    }
-);
+
+  // Known errors when running the tests
+  expectedFailures?: (string | RegExp)[] | 'ALL';
+
+  // Tests that cannot be run for now, either due to harness limitations, workerd bugs or
+  // features not yet implemented.
+  skippedTests?: (string | RegExp)[] | 'ALL';
```
anonrig: What about?

```suggestion
  fatalTests?: (string | RegExp)[] | true;
```

src/wpt/harness/harness.ts
```diff
@@ -69,24 +69,24 @@ type CommonOptions = {
...
+
+  // Tests that we have decided not to use. These are omitted from results and coverage
+  // calculations.
+  excludedTests?: (string | RegExp)[] | 'ALL';
```
anonrig: What about?

```suggestion
  omittedTests?: (string | RegExp)[] | true;
```

src/wpt/url-test.ts
```diff
@@ -48,76 +49,8 @@ export default {
...
-      'response.formData() with input: %61+%4d%4D=',
-      'request.formData() with input: id=0&value=%',
-      'response.formData() with input: id=0&value=%',
-      'request.formData() with input: b=%2sf%2a',
-      'response.formData() with input: b=%2sf%2a',
-      'request.formData() with input: b=%2%2af%2a',
-      'response.formData() with input: b=%2%2af%2a',
-      'request.formData() with input: b=%%2a',
-      'response.formData() with input: b=%%2a',
+      /request.formData\(\) with input:/,
```
anonrig, line 52: Nice!

src/wpt/harness/harness.ts
npaun: Sure. _omitted_ was the other word I was considering for this.

src/wpt/harness/harness.ts
npaun: I need to be slightly more wishy-washy than fatal. How about disabledTests?

src/wpt/harness/harness.ts
npaun: New contest: Guess how many times @npaun had to run search-and-replace again because he misspelled omitted

src/wpt/harness/harness.ts
anonrig: Skipped vs disabled are somewhat converging. A skipped test can be skipped due to being disabled? But I don't know. I'll leave it to you.

anonrig (approved):

src/wpt/harness/harness.ts
npaun: Will leave disabled for now.

src/wpt/harness/harness.ts
anonrig: 7. Is it 7??

PR #4254, opened 2025-06-04
Add node:fs sync/datasync tests

jasnell:


src/node/internal/internal_fs_callback.ts
```diff
@@ -279,11 +280,10 @@ export function fchown(
 }
 
 export function fdatasync(fd: number, callback: ErrorOnlyCallback): void {
-  // We are not performing any argument validation here because we are
-  // falling through to the synchronous version of fdatasync, which does the
-  // validation itself.
```
npaun, line 284: What happened here? Was the comment incorrect?

npaun (approved):

src/node/internal/internal_fs_callback.ts
jasnell: Changed. I had originally had it set so that input arg validation errors were reported via the callback but Node.js throws those synchronously and only forwards the actual operation errors to the callback. So had to adjust for that.

PR #4261, opened 2025-06-05
bump capnp-es to 0.0.11

vicb:
Bump capnp-es to 0.0.11

A similar bump is pending workers sdk:

- https://github.com/cloudflare/workers-sdk/pull/9495
- https://github.com/cloudflare/workers-sdk/pull/9493

0.0.11 should allow generating ts and js without a patch.
This update is a first step before updating the build rules.


anonrig:
Can you open a PR to EW as well?

vicb:
> Can you open a PR to EW as well?

I'll do when I fix the test failures

vicb:
Did some investigation.

It looks like the issue here is with https://github.com/unjs/capnp-es/blob/main/src/serialization/debug.ts pulling "node:*" which is not available in the worker.

Looks like I need to submit a PR to capnp-es to move this out of the main index.mjs so that we can execute the code on workers.

vicb:
Released capnp-es with a fix to remove dependencies on node packages
Rebased on main

vicb:
@anonrig It looks like EW still uses capnp-ts.

@penalosa do we need to drop the patch before updating to capnp-es there. If that's the root cause, let's merge this first and I will update the bazel rules so that we can drop the patch in a folllow up PR.

/cc @emily-shen

anonrig (approved):

PR #4271, opened 2025-06-06
NodeJS Compat: process features

guybedford:
Here's an initial stab at better process support.

For `process.versions`, we just include the node version maintained through a header file that will need to be manually updated. ~I included the version information for node, v8, icu, ada, brotli, simdutf, sqlite, nbytes, ncrypto as part of the Bazel build process. This exposes internal implementation details which might be considered a security risk. This is also what all Node-like platforms do.~ 

~We can discuss restricting this list down to just node and v8 or even just node. `process.versions.node` is a common compat check though so it's worth bearing that one in mind at least. Happy to remove, not tied to the implementation at all, would value feedback here.~

## Included Features

Properties:
  - process.platform - Returns OS platform (before returned darwin/linux/win32, now always returns linux)
  - process.arch - System architecture (always returns x64)
  - process.version - Node.js version string, hard coded in a header file, and will need to be updated to the running latest stable release as appropriate.
  - process.versions - Version info for node matching process.version
  - process.title - Process title (hardcoded to "workerd")
  - process.features - Feature flags object (as previously)
  - process.argv - Command line arguments array (hardcoded to ["workerd"])
  - process.argv0 - First argument (hardcoded to "workerd")
  - process.execArgv - Node.js specific arguments (empty array)
  - process.pid - Process ID (hardcoded to 1)
  - process.ppid - Parent process ID (hardcoded to 0)
  - ~process.release - Node.js release metadata, hardcoded to Node.js default (this is what Bun & Deno do, open to leaving this out as well)~ (removed)
  - process.config - Build configuration object (empty but just has properties for checking per Node compat)
  - process.allowedNodeEnvironmentFlags - Empty Set for environment flags

Methods:
  - process.getegid/geteuid/getgid/getuid() - Get effective group/user ID (returns 0)
  - process.setegid/seteuid/setgid/setuid() - Set effective group/user ID permission error
  - process.setgroups, getgroups() - permission error
  - process.initgroups - permission error
  - process.setSourceMapsEnabled() - Enable source maps (no-op)
  - process.getSourceMapsSupport() - returning false for all properties
  - process.kill - throws unsupported
  - process.ref - throws unsupported
  - process.unref = throws unsupported
  - process.hrtime - Supported, providing the same time as Date.now()
  - process.uptime - returning 0 currently, may update to be request time based
  - process.chdir - Added as unsupported pending FS work
  - process.cwd - Added as unsupported pending FS work
  - process.umask - Added as unsupported pending FS work
  - process.hrtime - returns the same time as Date.now()
  - process.hrtime.bigint - bigint variation of process.hrtime
  - process.uptime - always returns 0
  - process.emitWarning and process now extends event emitter with "on" support, enabled the former util test.
  - kill() - throws unsupported
  - exit() - as previously implemented
  - abort() - as exit(0)

Events
 - `warning` on `process.emitWarning`
 - `rejectionHandled` per the current global event
 - `unhandledRejection` per the current global event

## Excluded Features

The following features are still entirely omitted and only exported as `undefined`:

- exitCode
- channel
- connected
- binding()
- debugPort - possible for follow-up, would need to track debug port configuration
- dlopen()
- finalization
- getActiveResourcesInfo()
- setUncaughtExceptionCaptureCallback()
- hasUncaughtExceptionCaptureCallback()
- memoryUsage()
- noDeprecation
- permission
- release - deemed unnecessary
- report
- resourceUsage()
- send()
- traceDeprecation - undefined unless set anyway
- throwDeprecation - undefined unless set anyway
- sourceMapsEnabled - experimental and already deprecated, so skipped
- stderr / stdin / stdout - these should probably be implemented, but would be a larger effort
- threadCpuUsage() - added after Node.js 20
- Events: beforeExit, disconnect, exit, message, multipleResolves, workerMessage, uncaughtException, uncaughtExceptionMonitor, worker


src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -62,3 +62,182 @@
...
+    assert.ok(Array.isArray(process.execArgv));
+    assert.strictEqual(process.execArgv.length, 0);
+
+    assert.strictEqual(process.pid, 1);
+    assert.strictEqual(process.ppid, 0);
+
+    // Test release object
+    assert.strictEqual(typeof process.release, 'object');
+    assert.strictEqual(process.release.name, 'node');
+    assert.ok(process.release.sourceUrl.includes('nodejs.org'));
```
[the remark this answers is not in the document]
guybedford: This is a test, ignoring.

build/deps/versions.bzl
```diff
@@ -0,0 +1,182 @@
...
+    doc = "Extracts versions from dependency bzl files and generates C++ headers",
+    attrs = {
+        "node_version": attr.string(
+            mandatory = True,
+            doc = "The Node.js version string",
+        ),
+    },
+)
+
+def extract_versions(node_version):
```
vicb: Question: how is the node_version extracted? 
Is the version of Node in this repo?
If yes, could you expand on how that's related to workerd?

FYI in unenv we set the version to whatever API unenv implements. I honestly have no idea what would be the best value for workerd to use?

vicb:
> `process.versions.node` is a common compat check though so it's worth bearing that one in mind at least

FYI `process.versions.node` has been implemented in `unenv` because quite a few libs rely on that being set. It was released a few weeks back.

I think that was the right thing to do even if we broke other (older) libs along the way, i..e https://github.com/cloudflare/workers-sdk/issues/9352

/cc @mikenomitch @anonrig

src/node/internal/internal_process.ts
```diff
@@ -12,13 +12,163 @@ import { validateObject } from 'node-internal:validators';
...
+export function setegid(_guid: number) {
+  // no-op, since we only support gid 0
+}
+
+export function seteuid(_uid: number) {
+  // no-op, since we only support uid 0
+}
+
+export function setSourceMapsEnabled(_enabled: boolean) {
+  // no-op since we do not support disabling source maps
```
anonrig, line 117: Should we throw ERR_METHOD_NOT_IMPLEMENTED for these functions? Ignoring means we can never throw an error in here, but if we throw, we can remove the error without making it a breaking change.

guybedford:
@vicb am I understanding correctly that your shim caused this to pass in your workers apps, therefore you've effectively already tested "turning on" these feature gates? Are there more cases to worry about that rely on other compat features being available (e.g. should we hold back on this for FS to fully land?). The feedback is definitely good to know in terms of the risk assessment of this PR as well.

build/deps/versions.bzl
guybedford: I shared the NODE_VERSION we define in our main workspace file - https://github.com/cloudflare/workerd/blob/main/WORKSPACE#L21.

It does give a very new meaning to the term though, but seems a relatively good proxy for "the current node version".

We may wish to separate these definitions or more clearly define some process here for keeping this updated. Alternatively we may just always want to reflect the latest LTS instead and have a Bazel step that looks that up and uses that.

Fetching the latest LTS at build time might be the better option here? I can look into that.

src/workerd/api/node/i18n.c++
```diff
@@ -268,6 +268,17 @@ jsg::BufferSource transcode(
...
 
+kj::String getIcuVersion() {
+  // ICU provides version information through U_ICU_VERSION which is defined in uvernum.h
+  // The format is like "75.1" for ICU 75.1
+  UVersionInfo versionArray;
+  u_getVersion(versionArray);
+
+  // Convert the version array to a string
+  // UVersionInfo is an array of 4 uint8_t values: [major, minor, patch, build]
+  return kj::str(versionArray[0], ".", versionArray[1]);
```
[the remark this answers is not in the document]
guybedford: ICU versions are always `X.Y` for some reason.

build/deps/versions.bzl
```diff
@@ -0,0 +1,182 @@
...
+
+    v8_start = v8_bzl_content.find('VERSION = "')
+    if v8_start == -1:
+        fail("Could not find VERSION in v8.bzl")
+
+    v8_start += len('VERSION = "')
+    v8_end = v8_bzl_content.find('"', v8_start)
+    if v8_end == -1:
+        fail("Could not parse VERSION in v8.bzl")
+
```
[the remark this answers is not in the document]
guybedford: Let's first assess what versions we really do want to include, and then refine the code paths after that.

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -62,3 +62,182 @@
...
+    assert.ok(process.release.headersUrl.includes('nodejs.org'));
```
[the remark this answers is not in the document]
guybedford: This is now included.

build/deps/versions.bzl
jasnell: Hmm..the version of Node.js we use for the build is not the same as the version of the Node.js we likely want to claim interop for via the API. I don't think we should be using the build as the source here. We need to think through the version reporting a bit more.

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -62,3 +62,188 @@ export const processEnv = {
...
+
+    // List of expected version properties
+    const expectedVersions = [
+      'ada',
+      'node',
+      'brotli',
+      'simdutf',
+      'sqlite',
+      'nbytes',
+      'ncrypto',
```
jasnell: I'm not sure if we actually want to expose these details.

src/workerd/api/node/process.h
```diff
@@ -0,0 +1,75 @@
...
+#include <workerd/util/mimetype.h>
+
+namespace workerd::api::node {
+
+class ProcessModule final: public jsg::Object {
+ public:
+  ProcessModule() = default;
+  ProcessModule(jsg::Lock&, const jsg::Url&) {}
+
+#ifdef _WIN32
```
jasnell, line 23: Even when workerd is built on win32 and mac, I don't think we actually want to expose that fact. Everything should be expected to act generally the same as is does in production. For instance, in our implementation of `node:path` we always use the posix impl regardless.

src/node/internal/internal_process.ts
```diff
@@ -6,24 +6,180 @@
...
+export const execArgv = [];
+
+export const arch = 'x64';
+
+export const platform = processImpl.processPlatform;
+
+export const release = {
+  name: 'node',
+  sourceUrl: `https://nodejs.org/download/release/v${version}/node-v${version}.tar.gz`,
+  headersUrl: `https://nodejs.org/download/release/v${version}/node-v${version}-headers.tar.gz`,
```
jasnell: I don't think we should expose these properties.

src/workerd/api/node/process.h
```diff
@@ -0,0 +1,75 @@
...
+  jsg::JsObject getVersions(jsg::Lock& js) const {
+    auto versions = js.obj();
+    versions.set(js, "ada"_kj, js.str(workerd::api::node::adaVersion));
+    versions.set(js, "node"_kj, js.str(workerd::api::node::nodeVersion));
+    versions.set(js, "brotli"_kj, js.str(workerd::api::node::brotliVersion));
+    versions.set(js, "simdutf"_kj, js.str(workerd::api::node::simdutfVersion));
+    versions.set(js, "sqlite"_kj, js.str(workerd::api::node::sqliteVersion));
+    versions.set(js, "nbytes"_kj, js.str(workerd::api::node::nbytesVersion));
+    versions.set(js, "ncrypto"_kj, js.str(workerd::api::node::ncryptoVersion));
+    versions.set(js, "v8"_kj, js.str(workerd::api::node::v8Version));
```
jasnell: We generally avoid exposing specific version details for both the runtime and dependencies, largely as a bit of a security-by-obscurity measure. I'd prefer if we didn't expose these but let's get @kentonv's opinion.

guybedford: Yeah I gathered as much, but thought I'd try anyway. The conservative path here would be to do `node` only since it's not even a real version. But compatibility would want to include icu, v8, maybe brotli.

I'm totally fine to just revert to node only - will do that first thing Monday unless a good reason arises to keep the others.

src/node/internal/internal_process.ts
guybedford: It's just that both Bun and Deno implement this, and I was wondering if there are compat paths that use `process.release.name === 'node'` as a compat gate.

Perhaps we just make the URLs undefined to allow compat checks that happen to look like that?

The issue is that the closer we get to Node.js compat, the harder it is to exclude objects which are expected.

src/workerd/api/node/process.h
guybedford: This was actually already there - it's only in the diff since I separated out these process features from util.

It did surprise me as well to see this, it seems it only applies for local workerd workflows. In theory it might be a breaking change for workerd users though?

build/deps/versions.bzl
guybedford: I see two options here:

1. We maintain a Node version header file, tracking a baseline release to which we support its API. We then need to keep this file manually updated.
2. We add a build step to fetch the latest Node.js LTS version and setup the header automatically.

I'm open to either, this is about how we want to maintain it going forward.

danlapid:
process.platform - Returns OS platform (darwin/linux/win32) - [Dan]: should likely always be linux, users dont need to know which OS their workers run on
process.arch - System architecture (x64/arm64/ia32) - [Dan]: should likely always be x64, there might be certain exploits that only work on x64/arm/... machines and attackers might want to target specific production machines, we shouldn't le them.
process.version - Node.js version string - [Dan]: dunno what the right answer is here and afraid upgrading this will be a breaking change but leaving it old will also be bad.
process.versions - Version info for node, v8, icu, ada, brotli, simdutf, sqlite, nbytes, ncrypto - [Dan]: This is all internal details, I don't think we should expose any of it.
process.title - Process title (hardcoded to "workerd")
process.features - Feature flags object
process.argv - Command line arguments array - [Dan]: Where is this populated from? this sounds really bad if in prod all workers will have our command like arguments
process.argv0 - First argument (hardcoded to "node") - [Dan]: why not workerd?
process.execArgv - Node.js specific arguments (empty array)
process.pid - Process ID (hardcoded to 1)
process.ppid - Parent process ID (hardcoded to 0)
process.release - Node.js release metadata, hardcoded to Node.js default (this is what Bun & Deno do, open to leaving this out as well) - [Dan]: same as above, afraid that updating this will be a breaking change.
process.config - Build configuration object (empty but just has properties for checking per Node compat)
process.allowedNodeEnvironmentFlags - Empty Set for environment flags

guybedford:
With the version, `typeof process.versions.node === 'string'` is a very common compat check for Node compat. For this reason alone this PR will likely cause functional changes for many libraries as already mentioned by @vicb.

For the platform - we were already supporting in local workerd providing the native platform information. I've removed this, but in theory that is a breaking change for non edge worker users.

I've reduced the version list to just node now, with a manually maintained version. I've also updated the argv to use `workerd` instead of `node`, which definitely makes more sense to me.

src/workerd/api/node/tests/process-nodejs-test.js
guybedford: Reduced to just `node`.

src/workerd/api/node/process.h
guybedford: I've gone ahead and removed this to always be `linux`, but that is actually a change from what we had previously which is to provide the platform name - https://github.com/cloudflare/workerd/blob/24d5823cfb16684d0e2a46d31a88a456376cd1b0/src/workerd/api/node/util.h#L165.

src/node/internal/internal_process.ts
guybedford: Removed.

src/workerd/api/node/process.h
guybedford: Removed, leaving just a single header file to define the node version.

vicb:
> therefore you've effectively already tested "turning on" these feature gates? 

We have process.versions.node define in OpenNext for probably a couple months. It landed in unenv more recently in Wrangler 4.16.

build/deps/versions.bzl
jasnell: I think option 1 makes the most sense at this point.

src/workerd/api/node/process.h
```diff
@@ -0,0 +1,61 @@
...
+#include <workerd/util/mimetype.h>
+
+namespace workerd::api::node {
+
+class ProcessModule final: public jsg::Object {
+ public:
+  ProcessModule() = default;
+  ProcessModule(jsg::Lock&, const jsg::Url&) {}
+
+  static constexpr kj::StringPtr processPlatform = "linux"_kj;
```
jasnell: Since this is hard coded now it can just be set on the js side and doesn't really need to be in the header, correct?

src/workerd/api/node/node-version.h
```diff
@@ -0,0 +1,15 @@
...
+#include <kj/string.h>
+
+namespace workerd::api::node {
+
+// Node.js version supported by the platform.
+// This represents the most current Node.js version supported by the platform
+// and will change as Node.js release updates ship.
+static constexpr kj::StringPtr nodeVersion = "22.15.1"_kj;
+
+}  // namespace workerd::api::node
```
jasnell:
```suggestion
}  // namespace workerd::api::node

```

jasnell (approved):

src/node/internal/events.ts
```diff
@@ -1020,3 +1028,8 @@ export function on(
     iterator.return();
   }
 }
+
+// Initialize the process global as an event emitter, if process executed first
+if (process._initialized) process._initProcess();
+
+export var _initialized = true;
```
anonrig, line 1037: Why do we export values that are starting with underline? Those are private APIs that are not bound by semver. We avoided implementing those methods to workerd. Cc @jasnell

src/node/internal/internal_process.ts
```diff
@@ -6,24 +6,174 @@
...
+
+export const config = {
+  target_defaults: {},
+  variables: {},
+};
+
+export const pid = 1;
+export const ppid = 0;
+
+export function getegid(): number {
```
anonrig, line 65: Can you add small comments to why we are returning 0, or default values to each of those functions?

src/node/internal/internal_process.ts
```diff
@@ -179,16 +331,14 @@ export const env = new Proxy(getInitialEnv(), {
...
+  return processImpl.getBuiltinModule(id);
 }
 
-export function exit(code: number) {
-  utilImpl.processExitImpl(code);
+export function exit(code: number): void {
+  processImpl.processExitImpl(code);
 }
 
-export const platform = utilImpl.processPlatform;
```
anonrig: I didn't find this. Where did this method go?

src/workerd/api/node/process.h
```diff
@@ -0,0 +1,61 @@
...
+  // This is only added for Node.js compatibility and running OS specific tests
+  kj::StringPtr getProcessPlatform() const {
+    return processPlatform;
+  }
+
+  jsg::JsObject getEnvObject(jsg::Lock& js);
+
+  // Version getters
+  jsg::JsObject getVersions(jsg::Lock& js) const {
+    auto versions = js.obj();
```
anonrig: Can you move implementations to C++ file?

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -3,7 +3,7 @@ import assert from 'node:assert';
 export const processPlatform = {
   test() {
     assert.strictEqual(typeof process.platform, 'string');
-    assert.ok(['darwin', 'win32', 'linux'].includes(process.platform));
+    assert.strictEqual(process.platform, 'linux');
```
anonrig: This is a breaking change. We should not do this. I explicitly added this so we can test platform specific features on the existing test runner.

src/workerd/api/node/util.h
```diff
@@ -230,23 +220,6 @@ class UtilModule final: public jsg::Object {
...
-
-  // This is used in the implementation of process.exit(...). Contrary
-  // to what the name suggests, it does not actually exit the process.
-  // Instead, it will cause the IoContext, if any, and will stop javascript
-  // from further executing in that request. If there is no active IoContext,
-  // then it becomes a non-op.
-  void processExitImpl(jsg::Lock& js, int code);
-
-  // IMPORTANT: This function will always return "linux" on production.
-  // This is only added for Node.js compatibility and running OS specific tests
```
anonrig, line 243: We shouldn't remove this.

src/workerd/api/node/process.h
kentonv: FWIW I am happy with exposing V8 major version (e.g. "13.8"), but would prefer not to expose the full patch number as this isn't useful to applications and would provide TMI about our rollouts.

src/workerd/api/node/process.h
guybedford: Added. @anonrig the suggest here from earlier review was to hard-code `linux` as the platform instead of providing platform variation.

Formally, this is a breaking change from the previous behaviour in supporting platform variations for workerd only.

I'm open to either approach, although I do think with fs exposing a linux style filesystem there's an argument for making that the default. That and that this is a virtual platform to begin with.

@jasnell @anonrig let me know if you want to discuss further. An alternative might be some kind of compat flag / deprecation path for non linux platforms in workerd specifically, but I don't know how important that use case is.

src/workerd/api/node/util.h
guybedford: Let's move discussion to James's comment above.

src/workerd/api/node/tests/process-nodejs-test.js
guybedford: Let's move discussion to James's comment above.

src/node/internal/internal_process.ts
guybedford: It's been moved to the top of the file.

src/node/internal/internal_process.ts
guybedford: Added.

src/node/internal/events.ts
guybedford: This isn't exported publicly - see `src/node/events.ts` and `src/node/process.ts` for the actual public exported modules and the comment there. We need to maintain execution initialization statue between process and events in order to handle their mutual binding. The only other way to do that would be to have a third intermediary module maintaining this state otherwise which I'm open to if it reduces confusion as well.

src/workerd/api/node/process.h
anonrig: I prefer to have an option to test things according to operating system. 

Can we not remove this in this PR, and wait for me to come from holiday to have an in depth discussion?

src/node/internal/internal_process.ts
guybedford: Despite being an experimental function, both Bun and Deno support this so it is therefore part of the baseline compatibility expectations for users that they can call this.

Given that, I think it might be better not to introduce throwing behaviour for compatibility, rather if there is any doubt here it would be better not to implement this function at all.

This function is more like setting an option than calling a function with processing. Typically, in compat, throwing on functions that are unsupported makes sense for functions that are expected to perform operations or do work.

But since this is a configuration function like that is more like a flag or option being set it's probably better not to throw to avoid cases where applications are effectively signalling the expectation but not the requirement for the source maps output.

Also most applications turning this on will also not likely be relying on any effects of the option since it only affects error traces which are more rarely observed although there's always the edge cases.

src/workerd/api/node/process.h
guybedford: If doing this it might make sense to fill in zeroes to match the versioning scheme like `13.8.0.0` say. That said, I think for now it's okay to leave out - rather if this topic comes up due to a real compat question we can re-assess.

src/workerd/api/node/process.h
guybedford: Sure, I'll back out this change, if you're happy to follow-up on the discussion / update here in due course?

For reference original discussion was in https://github.com/cloudflare/workerd/pull/4271/files#diff-73a094ddc8d8c0aaa72486bb34a4225526e5189fbc9dd5fce2584dae589ada0d.

guybedford: Here's the corrected discussion url - https://github.com/cloudflare/workerd/pull/4271#discussion_r2133081484

jasnell: Oy, ok, I had forgotten that @anonrig already had `process.platform`. Well, then, I guess we should avoid the breaking change and allow the value to be set differently. We should just make sure that  in production it always returns `linux`.

src/workerd/api/node/node-version.h
```diff
@@ -0,0 +1,15 @@
...
+//     https://opensource.org/licenses/Apache-2.0
+#pragma once
+
+#include <kj/string.h>
+
+namespace workerd::api::node {
+
+// Node.js version supported by the platform.
+// This represents the most current Node.js version supported by the platform
+// and will change as Node.js release updates ship.
```
jasnell: It's likely worth expanding on the fact that this is an approximation. We do not fully support this specific version (as there are APIs we know are missing) and we may support features that shipped after this version.

src/workerd/api/node/process.h
guybedford: Ok, I've hard-coded:

```h
// We will always set platform: 'linux' for production
#ifdef EW_APPLICATION_NAME
  static constexpr kj::StringPtr processPlatform = "linux"_kj;
```

for now as well.

src/workerd/api/node/process.h
```diff
@@ -0,0 +1,61 @@
...
+namespace workerd::api::node {
+
+class ProcessModule final: public jsg::Object {
+ public:
+  ProcessModule() = default;
+  ProcessModule(jsg::Lock&, const jsg::Url&) {}
+
+  jsg::JsValue getBuiltinModule(jsg::Lock& js, kj::String specifier);
+
+// We will always set platform: 'linux' for edgeworker
```
jasnell:
```suggestion
// We will always set platform: 'linux' for production
```

src/workerd/api/node/node-version.h
guybedford: Added some clarification.

src/workerd/api/node/process.h
anonrig: If we write a test that runs on windows on CI using process.platform, it will never run on EW? This is not we want, do we?

guybedford: `EW_APPLICATION_NAME` is only defined by edgeworker builds.

guybedford: Hmm although I guess header definition ordering isn't going to work out for this. Ok nvm.

jasnell: Not sure what you're asking here @anonrig ... the internal CI builds only ever run on linux machines and the `EW_APPLICATION_NAME` flag is only ever defined on production builds. We want `process.platform` to always be `linux` internally. We can have the windows and mac values available on workerd since that's the only place it is relevant.

anonrig: Now I understand. Thanks folks!

anonrig (approved):

src/workerd/api/node/process.h
```diff
@@ -0,0 +1,59 @@
...
+class ProcessModule final: public jsg::Object {
+ public:
+  ProcessModule() = default;
+  ProcessModule(jsg::Lock&, const jsg::Url&) {}
+
+  jsg::JsValue getBuiltinModule(jsg::Lock& js, kj::String specifier);
+
+// We will always set platform: 'linux' for production
+#ifdef _WIN32
+  static constexpr kj::StringPtr processPlatform = "win32"_kj;
```
kentonv, line 24: We should always report the platform as "linux" even when running locally on non-linux systems. The purpose of running workerd locally is to simulate production. Reporting a value that will never appear in production can only make the simulation unrealistic.

kentonv: Also note that all of our node compat APIs specifically implement compatibility with Node running on Linux, so telling the application that it's not on Linux can only confuse it and make it do the wrong thing.

jasnell: See previous conversation on this. It was already implemented to support these other values in local dev in order to support some needs in web platform test running in CI

kentonv: Why would web platform tests in CI need to know what platform they are running on? They should be performing identical behavior regardless of the host platform, no?

src/workerd/api/node/process.h
kentonv: We do not consider things a "breaking change" if they do not affect production, so it is not a breaking change to make this always return "linux". It should definitely have been doing so all along. There is no reason any tests should need to know about their host platform, because workerd totally hides the host platform, so the behavior being tested should be identical on all platforms. If workerd is implementing any behavior which differs across platforms today, that is a bug which we need to fix.

guybedford: I tend to agree with @kentonv on this, but since this is something that is already implemented, and not a change that this PR is making, can I suggest we land this as-is right now, then get this platform update in as a follow-on?

jasnell: @npaun and @anonrig who were working on these bits of the wpt's

kentonv: I'm fine with merging as-is if the plan is to fix it promptly thereafter.

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -60,5 +60,587 @@ export const processEnv = {
...
+    writeFileSync(validEnvFilePath, validEnv);
+
+    // supports passing path
+    {
+      process.loadEnvFile(validEnvFilePath);
+      assert.strictEqual(process.env.BASIC, 'basic');
+    }
+
+    // TODO(soon): Support once FS issue resolved
+    // supports not-passing a path
```
jasnell, line 695: Fix for the readFileSync issue has been merged.

guybedford: Thanks, seems to be working now.

src/workerd/api/node/process.h
guybedford: After discusing with @anonrig and @npaun it was suggested to implement a test-only compat feature here, so I've gone ahead and added a `unsupported_process_real_platform` compatibility flag which only works under experimental and is never intended to be used by anyone other than them.

Is this a suitable compromise or not? If not, I will split this out into a separate PR to continue the discussion there rather at least that would give us a starting point.

src/node/internal/internal_errors.ts
```diff
@@ -858,6 +858,13 @@ export class ERR_EEXIST extends NodeSyscallError {
   }
 }
 
+export class ERR_EPERM extends NodeSyscallError {
+  constructor(options: { syscall: string }) {
+    super('EPERM', 'Operation not permitted', options.syscall);
+    this.errno = 1; // EPERM
```
jasnell: We might need to double check. `EPERM` is 1 but `UV_EPERM` is -1... is the `errno` supposed to be positive or negative for these? If I call `fs.readFileSync(...)` on a file is not readable, the `errno` property on the resulting error will be `-13` for `UV_EACCES`.

guybedford: I've added `errno` as an option while defaulting to `1`.

guybedford: Specifically, in Node.js, `process.setgid(500)` gives:

```
Uncaught Error: EPERM, Operation not permitted
    at process.setgid (node:internal/bootstrap/switches/does_own_process_state:86:22) {
  errno: 1,
  code: 'EPERM',
  syscall: 'setgid'
}
```

src/node/internal/internal_errors.ts
```diff
@@ -858,6 +858,13 @@ export class ERR_EEXIST extends NodeSyscallError {
   }
 }
 
+export class ERR_EPERM extends NodeSyscallError {
+  constructor(options: { syscall: string; errno?: number }) {
+    super('EPERM', 'Operation not permitted', options.syscall);
+    this.errno = options.errno || 1; // EPERM
+  }
```
jasnell, line 865:
```suggestion
  constructor(options: { syscall: string; errno?: number }) {
    const {
      syscall,
      errno = 1,
    } = options;
    super('EPERM', 'Operation not permitted', syscall);
    this.errno = errno ; // EPERM
  }
```

Passing the `errno` is good but I still think we need to reconcile this, and it might need to be a question for node.js... is the `errno` supposed to reflect UV_EPERM or EPERM?

guybedford: This particular error is `EPERM` - e.g. try `process.setgid(500)` in Node.js.

guybedford:
This PR is ready for final review, with the immediate follow-ups remaining being `chdir`, `cwd`, `stdin`, `stdout` and `stderr`.

src/workerd/api/node/process.h
guybedford: There is now a compat flag implemented for this as mentioned in https://github.com/cloudflare/workerd/pull/4271#discussion_r2141510115.

src/node/internal/internal_errors.ts
jasnell: ugh, we should open an issue to deal with this in Node.js. This needs to be more consistent

jasnell (approved):

src/node/internal/internal_errors.ts
guybedford: I've posted a Node.js issue.

src/workerd/api/node/process.h
guybedford: I believe the new flag has resolved this conflict, since by default the `linux` value is now hard-coded without this flag, but please let me know if there is more to discuss here.

kentonv: I am fine with the compat flag, but I still don't understand why the tests would need to know what platform they are on.

PR #4272, opened 2025-06-06
[cxx] add container service for local development

anonrig:
Despite https://github.com/cloudflare/workerd/pull/4238, this pull-request implements the container service in C++.

In order to run the test you need to pull ubuntu image from docker using: `docker pull ubuntu`.

There are still some todos such as the following, but rest of the functionality works perfectly.

- [x] getTcpPort() implementation
- [x] signal number to signal string (example: SIGINT) conversion
- [x] Uncommenting several tests
- [x] disable the test on CI until we write a mock server that supports docker engine API.

### DIFF
--- a/src/workerd/server/container-client.c++
+++ b/src/workerd/server/container-client.c++
@@ -1,0 +1,411 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "container-client.h"
+
+#include <workerd/io/container.capnp.h>
+#include <workerd/jsg/jsg.h>
+#include <workerd/server/docker-api.capnp.h>
+
+#include <capnp/compat/json.h>
+#include <capnp/message.h>
+#include <kj/async-io.h>
+#include <kj/async.h>
+#include <kj/compat/http.h>
+#include <kj/debug.h>
+#include <kj/encoding.h>
+#include <kj/exception.h>
+#include <kj/string.h>
+
+namespace workerd::server {
+
+namespace {
+kj::StringPtr signalToString(uint32_t signal) {
+  switch (signal) {
+    case 1:
+      return "SIGHUP"_kj;  // Hangup
+    case 2:
+      return "SIGINT"_kj;  // Interrupt
+    case 3:
+      return "SIGQUIT"_kj;  // Quit
+    case 4:
+      return "SIGILL"_kj;  // Illegal instruction
+    case 5:
+      return "SIGTRAP"_kj;  // Trace trap
+    case 6:
+      return "SIGABRT"_kj;  // Abort
+    case 7:
+      return "SIGBUS"_kj;  // Bus error
+    case 8:
+      return "SIGFPE"_kj;  // Floating point exception
+    case 9:
+      return "SIGKILL"_kj;  // Kill
+    case 10:
+      return "SIGUSR1"_kj;  // User signal 1
+    case 11:
+      return "SIGSEGV"_kj;  // Segmentation violation
+    case 12:
+      return "SIGUSR2"_kj;  // User signal 2
+    case 13:
+      return "SIGPIPE"_kj;  // Broken pipe
+    case 14:
+      return "SIGALRM"_kj;  // Alarm clock
+    case 15:
+      return "SIGTERM"_kj;  // Termination
+    case 16:
+      return "SIGSTKFLT"_kj;  // Stack fault (Linux)
+    case 17:
+      return "SIGCHLD"_kj;  // Child status changed
+    case 18:
+      return "SIGCONT"_kj;  // Continue
+    case 19:
+      return "SIGSTOP"_kj;  // Stop
+    case 20:
+      return "SIGTSTP"_kj;  // Terminal stop
+    case 21:
+      return "SIGTTIN"_kj;  // Background read from tty
+    case 22:
+      return "SIGTTOU"_kj;  // Background write to tty
+    case 23:
+      return "SIGURG"_kj;  // Urgent condition on socket
+    case 24:
+      return "SIGXCPU"_kj;  // CPU limit exceeded
+    case 25:
+      return "SIGXFSZ"_kj;  // File size limit exceeded
+    case 26:
+      return "SIGVTALRM"_kj;  // Virtual alarm clock
+    case 27:
+      return "SIGPROF"_kj;  // Profiling alarm clock
+    case 28:
+      return "SIGWINCH"_kj;  // Window size change
+    case 29:
+      return "SIGIO"_kj;  // I/O now possible
+    case 30:
+      return "SIGPWR"_kj;  // Power failure restart (Linux)
+    case 31:
+      return "SIGSYS"_kj;  // Bad system call
+    default:
+      return "SIGKILL"_kj;
+  }
+}
+}  // namespace
+
+template <typename T>
+typename T::Builder decodeJsonResponse(kj::StringPtr response) {
+  capnp::JsonCodec codec;
+  codec.handleByAnnotation<T>();
+  capnp::MallocMessageBuilder message;
+  auto jsonRoot = message.initRoot<T>();
+  codec.decode(response, jsonRoot);
+  return jsonRoot;
+}
+
+ContainerClient::ContainerClient(capnp::ByteStreamFactory& byteStreamFactory,
+    kj::Timer& timer,
+    kj::Network& network,
+    kj::String dockerPath,
+    kj::String containerName,
+    kj::String imageName)
+    : byteStreamFactory(byteStreamFactory),
+      timer(timer),
+      network(network),
+      dockerPath(kj::mv(dockerPath)),
+      containerName(kj::encodeUriComponent(kj::mv(containerName))),
+      imageName(kj::mv(imageName)) {}
+
+// Docker-specific Port implementation that implements rpc::Container::Port::Server
+class ContainerClient::DockerPort final: public rpc::Container::Port::Server {
+ public:
+  DockerPort(ContainerClient& containerClient, kj::String containerHost, uint16_t containerPort)
+      : containerClient(containerClient),
+        containerHost(kj::mv(containerHost)),
+        containerPort(containerPort) {}
+
+  kj::Promise<void> connect(ConnectContext context) override {
+    kj::HttpHeaderTable headerTable;
+    kj::HttpHeaders headers(headerTable);
+
+    // Port mappings might have outdated mappings, we can't know if a connect request
+    // fails because the app hasn't finished starting up or because the mapping is outdated.
+    // To be safe we should inspect the container to get up to date mappings.
+    const auto [_running, portMappings] = co_await containerClient.inspectContainer();
+    auto maybeMappedPort = portMappings.find(containerPort);
+    if (maybeMappedPort == kj::none) {
+      throw JSG_KJ_EXCEPTION(DISCONNECTED, Error,
+          "connect(): Connection refused: container port not found. Make sure you exposed the port in your container definition.");
+    }
+    auto mappedPort = KJ_ASSERT_NONNULL(maybeMappedPort);
+
+    auto address =
+        co_await containerClient.network.parseAddress(kj::str(containerHost, ":", mappedPort));
+    auto connection = co_await address->connect();
+
+    auto upPipe = kj::newOneWayPipe();
+    auto upEnd = kj::mv(upPipe.in);
+    auto results = context.getResults();
+    results.setUp(containerClient.byteStreamFactory.kjToCapnp(kj::mv(upPipe.out)));
+    auto downEnd = containerClient.byteStreamFactory.capnpToKj(context.getParams().getDown());
+    pumpTask =
+        kj::joinPromisesFailFast(kj::arr(upEnd->pumpTo(*connection), connection->pumpTo(*downEnd)))
+            .ignoreResult()
+            .attach(kj::mv(upEnd), kj::mv(connection), kj::mv(downEnd));
+    co_return;
+  }
+
+ private:
+  // ContainerClient is owned by the Worker::Actor and keeps it alive.
+  ContainerClient& containerClient;
+  kj::String containerHost;
+  uint16_t containerPort;
+  kj::Maybe<kj::Promise<void>> pumpTask;
+};
+
+kj::Promise<ContainerClient::Response> ContainerClient::dockerApiRequest(
+    kj::HttpMethod method, kj::StringPtr endpoint, kj::Maybe<kj::StringPtr> body) {
+  kj::HttpHeaderTable headerTable;
+  auto address = co_await network.parseAddress(dockerPath);
+  auto connection = co_await address->connect();
+  auto httpClient = kj::newHttpClient(headerTable, *connection).attach(kj::mv(connection));
+  kj::HttpHeaders headers(headerTable);
+  headers.set(kj::HttpHeaderId::HOST, "localhost");
+
+  KJ_IF_SOME(requestBody, body) {
+    headers.set(kj::HttpHeaderId::CONTENT_TYPE, "application/json");
+    headers.set(kj::HttpHeaderId::CONTENT_LENGTH, kj::str(requestBody.size()));
+
+    auto req = httpClient->request(method, endpoint, headers, requestBody.size());
+    {
+      auto body = kj::mv(req.body);
+      co_await body->write(requestBody.asBytes());
+    }
+    auto response = co_await req.response;
+    auto result = co_await response.body->readAllText();
+    co_return Response{.statusCode = response.statusCode, .body = kj::mv(result)};
+  } else {
+    auto req = httpClient->request(method, endpoint, headers);
+    { auto body = kj::mv(req.body); }
+    auto response = co_await req.response;
+    auto result = co_await response.body->readAllText();
+    co_return Response{.statusCode = response.statusCode, .body = kj::mv(result)};
+  }
+}
+
+kj::Promise<ContainerClient::InspectResponse> ContainerClient::inspectContainer() {
+  // Docker API: GET /containers/{id}/json
+  auto endpoint = kj::str("/containers/", containerName, "/json");
+
+  auto response = co_await dockerApiRequest(kj::HttpMethod::GET, endpoint);
+  // We check if the container with the given name exist, and if it's not,
+  // we simply return false while avoiding an unnecessary error.
+  if (response.statusCode == 404) {
+    co_return InspectResponse{.isRunning = false, .ports = {}};
+  }
+  JSG_REQUIRE(response.statusCode == 200, Error, "Container inspect failed");
+  // Parse JSON response
+  auto jsonRoot = decodeJsonResponse<docker_api::Docker::ContainerInspectResponse>(response.body);
+  kj::HashMap<uint16_t, uint16_t> portMappings;
+  for (auto portMapping: jsonRoot.getNetworkSettings().getPorts().getObject()) {
+    auto port = portMapping.getName();
+    // We need to get "8080" from "8080/tcp"
+    auto rawPort = port.asString().slice(0, KJ_ASSERT_NONNULL(port.asString().find("/")));
+    auto portNumber = kj::str(rawPort).parseAs<uint16_t>();
+    uint16_t number;
+    {
+      // We need to retrieve "HostPort" from the following JSON structure
+      //
+      // "Ports": {
+      // 	"8080/tcp": [
+      // 		{
+      // 			"HostIp": "0.0.0.0",
+      // 			"HostPort": "55000"
+      // 		}
+      // 	]
+      // },
+      //
+      auto array = portMapping.getValue().getArray();
+      JSG_REQUIRE(array.size() > 0, Error, "Malformed ContainerInspect port mapping response");
+      auto obj = array[0].getObject();
+      JSG_REQUIRE(array.size() > 1, Error, "Malformed ContainerInspect port mapping object");
+      auto mappedPort = obj[1].getValue().getString();
+      number = mappedPort.asString().parseAs<uint16_t>();
+    }
+    portMappings.insert(portNumber, number);
+  }
+
+  // Look for Status field in the JSON object
+  JSG_REQUIRE(jsonRoot.hasState(), Error, "Malformed ContainerInspect response");
+  auto state = jsonRoot.getState();
+  JSG_REQUIRE(state.hasStatus(), Error, "Malformed ContainerInspect response");
+  auto status = state.getStatus();
+  bool running = status == "running";
+  co_return InspectResponse{.isRunning = running, .ports = kj::mv(portMappings)};
+}
+
+kj::Promise<void> ContainerClient::createContainer(
+    kj::Maybe<capnp::List<capnp::Text>::Reader> entrypoint,
+    kj::Maybe<capnp::List<capnp::Text>::Reader> environment) {
+  // Docker API: POST /containers/create
+  capnp::JsonCodec codec;
+  codec.handleByAnnotation<docker_api::Docker::ContainerCreateRequest>();
+  capnp::MallocMessageBuilder message;
+  auto jsonRoot = message.initRoot<docker_api::Docker::ContainerCreateRequest>();
+  jsonRoot.setImage(imageName);
+  // Add entrypoint if provided
+  KJ_IF_SOME(ep, entrypoint) {
+    auto jsonCmd = jsonRoot.initCmd(ep.size());
+    for (uint32_t i: kj::zeroTo(ep.size())) {
+      jsonCmd.set(i, ep[i]);
+    }
+  }
+
+  // Add environment variables if provided
+  KJ_IF_SOME(env, environment) {
+    auto jsonEnv = jsonRoot.initEnv(env.size());
+    for (uint32_t i: kj::zeroTo(env.size())) {
+      jsonEnv.set(i, env[i]);
+    }
+  }
+
+  auto hostConfig = jsonRoot.initHostConfig();
+  // We need to publish all ports to properly get the mapped port number locally
+  hostConfig.setPublishAllPorts(true);
+  // We need to set a restart policy to avoid having ambiguous states
+  // where the container we're managing is stuck at "exited" state.
+  hostConfig.initRestartPolicy().setName("on-failure");
+
+  // Encode to JSON string
+  kj::String jsonBody = codec.encode(jsonRoot);
+  const auto endpoint = kj::str("/containers/create?name=", containerName);
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint, jsonBody.asPtr());
+
+  // statusCode 201 refers to "container created successfully"
+  // statusCode 409 refers to "conflict". Occurs when a container with the given name exists.
+  // Both are fine, so long as the container exists. Though we might want to call destroy and
+  // recreate on 409 in the future.
+  if (response.statusCode != 201 && response.statusCode != 409) {
+    JSG_REQUIRE(response.statusCode != 404, Error, "No such image available named ", imageName);
+    JSG_FAIL_REQUIRE(Error, "Create container failed with: ", response.body);
+  }
+}
+
+kj::Promise<void> ContainerClient::startContainer() {
+  // Docker API: POST /containers/{id}/start
+  const auto endpoint = kj::str("/containers/", containerName, "/start");
+  // We have to send an empty body since docker API will throw an error if we don't.
+  kj::StringPtr body = "";
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint, body);
+  // statusCode 204 refers to "no error"
+  // statusCode 304 refers to "container already started"
+  // Both are fine to ignore.
+  JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 304, Error,
+      "Starting container failed with: ", response.body);
+}
+
+kj::Promise<void> ContainerClient::stopContainer() {
+  // Docker API: POST /containers/{id}/stop
+  const auto endpoint = kj::str("/containers/", containerName, "/stop");
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint);
+  // statusCode 204 refers to "no error"
+  // statusCode 304 refers to "container already stopped"
+  // Both are fine to avoid when stop container is called.
+  JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 304, Error,
+      "Stopping container failed with: ", response.body);
+}
+
+kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
+  // Docker API: POST /containers/{id}/kill
+  const auto endpoint =
+      kj::str("/containers/", containerName, "/kill?signal=", signalToString(signal));
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint);
+  // statusCode 409 refers to "container is not running"
+  // We should not throw an error when the container is already not running.
+  JSG_REQUIRE(response.statusCode == 200 || response.statusCode == 409, Error,
+      "Stopping container failed with: ", response.body);
+}
+
+kj::Promise<void> ContainerClient::status(StatusContext context) {
+  const auto [isRunning, _ports] = co_await inspectContainer();
+  context.getResults().setRunning(isRunning);
+}
+
+kj::Promise<void> ContainerClient::start(StartContext context) {
+  const auto params = context.getParams();
+
+  // Get the lists directly from Cap'n Proto
+  kj::Maybe<capnp::List<capnp::Text>::Reader> entrypoint = kj::none;
+  kj::Maybe<capnp::List<capnp::Text>::Reader> environment = kj::none;
+
+  if (params.hasEntrypoint()) {
+    entrypoint = params.getEntrypoint();
+  }
+  if (params.hasEnvironmentVariables()) {
+    environment = params.getEnvironmentVariables();
+  }
+
+  co_await createContainer(entrypoint, environment);
+  co_await startContainer();
+}
+
+kj::Promise<void> ContainerClient::monitor(MonitorContext context) {
+  // Docker API: POST /containers/{id}/wait - wait for container to exit
+  const auto endpoint = kj::str("/containers/", containerName, "/wait");
+  // Monitor is often called right after start but the api layer's start does not await the RPC's
+  // start response. That means that the createContainer call might not have even started yet.
+  // If it hasn't, we'll give it 3 tries before failing.
+  for (int i = 0; i < 3; i++) {
+    auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint);
+    if (response.statusCode == 404) {
+      co_await timer.afterDelay(1 * kj::SECONDS);
+      continue;
+    }
+    JSG_REQUIRE(response.statusCode == 200, Error,
+        "Monitoring container failed with: ", response.statusCode, response.body);
+    // Parse JSON response
+    auto jsonRoot = decodeJsonResponse<docker_api::Docker::ContainerMonitorResponse>(response.body);
+    auto statusCode = jsonRoot.getStatusCode();
+    JSG_REQUIRE(statusCode == 0, Error, "Container exited with unexpected exit code ", statusCode);
+    co_return;
+  }
+  JSG_FAIL_REQUIRE(Error, "Monitor failed to find container");
+}
+
+kj::Promise<void> ContainerClient::destroy(DestroyContext context) {
+  const auto [running, _ports] = co_await inspectContainer();
+  if (running) {
+    co_await stopContainer();
+    auto endpoint = kj::str("/containers/", containerName, "?force=true");
+    auto response = co_await dockerApiRequest(kj::HttpMethod::DELETE, endpoint);
+    // statusCode 204 refers to "no error"
+    // statusCode 404 refers to "no such container"
+    // Both of which are fine for us since we're tearing down the container anyway.
+    JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 404, Error,
+        "Removing a container failed with: ", response.body);
+    {
+      endpoint = kj::str("/containers/", containerName, "/wait");
+      response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint);
+      JSG_REQUIRE(response.statusCode == 200 || response.statusCode == 404, Error,
+          "Waiting for container removal failed with: ", response.statusCode, response.body);
+    }
+  }
+}
+
+kj::Promise<void> ContainerClient::signal(SignalContext context) {
+  const auto params = context.getParams();
+  co_await killContainer(params.getSigno());
+}
+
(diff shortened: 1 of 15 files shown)


src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+kj::Promise<void> ContainerClient::signal(SignalContext context) {
+  auto params = context.getParams();
+  co_await killContainer(params.getSigno());
+}
+
+kj::Promise<void> ContainerClient::getTcpPort(GetTcpPortContext context) {
+  auto params = context.getParams();
+  uint16_t port = params.getPort();
+
+  auto results = context.getResults();
```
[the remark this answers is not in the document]
anonrig: There is a TODO for this...

src/workerd/io/container-client-test.js
```diff
@@ -0,0 +1,281 @@
...
+    for (const badEnv of badEnvs) {
+      try {
+        await container.start({
+          entrypoint: ['/bin/sh'],
+          env: badEnv.value,
+          enableInternet: true,
+        });
+        throw new Error("start should've thrown an error");
+      } catch (err) {
+        assert.equal(err.message, badEnv.err);
```
vicb: nit: use assert.throws ?

src/workerd/io/container-client-test.js
```diff
@@ -0,0 +1,281 @@
...
+      }
+    }
+
+    // Start container with valid configuration
+    await container.start({
```
vicb: `assert.doesNotThrow` ?

src/workerd/io/container-client-test.js
```diff
@@ -0,0 +1,281 @@
...
+        err: "Environment variable names cannot contain '=': B=",
+      },
+    ];
+
+    for (const badEnv of badEnvs) {
+      try {
+        await container.start({
+          entrypoint: ['/bin/sh'],
+          env: badEnv.value,
+          enableInternet: true,
```
vicb: I think that's the only we care about here

```suggestion
          env: badEnv.value,
```

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,17 @@ struct Worker {
     # workerd uses SQLite to back all Durable Objects, but the SQL API is hidden by default to
     # emulate behavior of traditional DO namespaces on Cloudflare that aren't SQLite-backed. This
     # flag should be enabled when testing code that will run on a SQLite-backed namespace.
+
+    container @5 :ContainerOptions;
+    # Optional configuration options for local development container interaction.
+
+    struct ContainerOptions {
```
vicb, line 611: Should this be move to top level?

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,17 @@ struct Worker {
...
+      containerName @0 :Text;
+      # Name of the container that should be created.
+
+      imageTag @1 :Text;
```
vicb: aren't tag local to images (i.e. 2 images can have the same tag).
If that's true the tag is probably not enough?

src/workerd/server/workerd.capnp
```diff
@@ -651,8 +662,17 @@ struct Worker {
   streamingTails @15 :List(ServiceDesignator);
   # List of streaming tail worker services that should receive tail events for this worker.
   # NOTE: This will be deleted in a future refactor, do not depend on this.
+
+  containerService @16 :ServiceDesignator;
```
vicb: Do we need an external service in C++ - I think that's what Kenton mentioned, this could be local.

src/workerd/io/container-client-test.js
```diff
@@ -0,0 +1,281 @@
+import { DurableObject } from 'cloudflare:workers';
+import assert from 'node:assert';
+
+const CONTAINER_NAME = 'container-example';
+const IMAGE_TAG = 'ubuntu';
+
+export class DurableObjectExample extends DurableObject {
```
vicb: Maybe use a more specific name than `DurableObjectExample`

src/workerd/io/container-client-test.js
```diff
@@ -0,0 +1,281 @@
...
+        entrypoint: ['leave-running'],
+      });
+    }
+
+    assert.strictEqual(container.running, true);
+  }
+
+  async checkRunning() {
+    // Check container was started using leaveRunning()
+    let container = this.ctx.container;
```
vicb:
```suggestion
    const container = this.ctx.container;
```

src/workerd/server/server.c++
```diff
@@ -706,10 +708,10 @@ class Server::ExternalHttpService final: public Service, private kj::TaskSet::Er
       wrappedResponse = response;
       if (parent.rewriter->needsRewriteRequest()) {
         auto rewrite = parent.rewriter->rewriteOutgoingRequest(url, headers, metadata.cfBlobJson);
-        return parent.serviceAdapter->request(method, url, *rewrite.headers, requestBody, *this)
-            .attach(kj::mv(rewrite));
+        co_return co_await parent.serviceAdapter->request(
```
npaun: Was there a reason for switching from `return` to `co_return co_await`?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+  }
+}
+
+kj::Promise<void> ContainerClient::startContainer() {
+  // Docker API v1.50: POST /containers/{id}/start
+  auto endpoint = kj::str("/containers/", containerName, "/start");
+  kj::StringPtr body = "";
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint, body);
+  // statusCode 304 refers to "container already started"
+  JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 304, Error,
```
vicb: require 304 ?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+  codec.handleByAnnotation<docker_api::Docker::ContainerInspectResponse>();
+  kj::String jsonBody = codec.encode(jsonRoot);
+
+  auto endpoint = kj::str("/containers/create?name=", containerName);
+
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint, jsonBody.asPtr());
+
+  // statusCode 201 refers to "container created successfully"
+  // statusCode 409 refers to "conflict"
+  if (response.statusCode != 201 && response.statusCode != 409) {
```
vicb: why `response.statusCode != 409` ?
sounds like an error from the comment?

src/workerd/server/server.c++
danlapid: coroutines for life, never return promises, lifetimes get confusing and attachments become annoying.

src/workerd/server/workerd.capnp
danlapid: You have a link to the mention? a service designator can be local, this specific one will point at a unix domain socket, see example.

src/workerd/server/workerd.capnp
danlapid: Why?

src/workerd/server/server.c++
```diff
@@ -760,9 +762,13 @@ class Server::ExternalHttpService final: public Service, private kj::TaskSet::Er
       if (parent.rewriter->needsRewriteResponse()) {
         auto rewrite = headers.cloneShallow();
         parent.rewriter->rewriteResponse(rewrite);
-        return response.send(statusCode, statusText, rewrite, expectedBodySize);
+        auto res = response.send(statusCode, statusText, rewrite, expectedBodySize);
+        wrappedResponse = kj::none;
+        return res;
```
npaun: This doesn't feel related. Does it fix a bug?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+    headers.set(contentType, "application/json");
+    headers.set(contentLength, kj::str(requestBody.size()));
+
+    auto req = httpClient->request(method, endpoint, headers, requestBody.size());
+    {
+      auto body = kj::mv(req.body);
+      co_await body->write(requestBody.asBytes());
+    }
+    auto response = co_await req.response;
+    auto result = co_await response.body->readAllText();
```
vicb: Would it make sense to return JSON (ref l 82)

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+  }
+}
+
+kj::Promise<bool> ContainerClient::isContainerRunning() {
+  // Docker API v1.50: GET /containers/{id}/json
+  auto endpoint = kj::str("/containers/", containerName, "/json");
+
+  auto response = co_await dockerApiRequest(kj::HttpMethod::GET, endpoint);
+  if (response.statusCode == 404) {
+    co_return false;
```
vicb: Should we throw here instead ?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+  }
+  JSG_REQUIRE(response.statusCode == 200, Error, "Container inspect failed");
```
vicb: I might miss something about `JSQ_REQUIRE` but `response.statusCode == 200` sounds wrong?

nit: Clarify the message "Container inspect failed"

src/workerd/io/container-client.c++
danlapid: Of which type?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+
+  // Parse JSON response
+  capnp::JsonCodec codec;
+  codec.handleByAnnotation<docker_api::Docker::ContainerInspectResponse>();
+  capnp::MallocMessageBuilder message;
+  auto jsonRoot = message.initRoot<docker_api::Docker::ContainerInspectResponse>();
```
vicb: Cf my earlier comment.

May dockerApiRequest should be templated and decode the JSON

src/workerd/io/container-client.c++
danlapid: isContainerRunning on a container that wasn't created should be `false` imo.

up to @mikenomitch

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+
+  // Look for Status field in the JSON object
+  if (jsonRoot.hasState()) {
+    auto state = jsonRoot.getState();
+    if (state.hasStatus()) {
+      auto status = state.getStatus();
+      co_return status == "running";
+    }
+  }
+  co_return false;
```
vicb: throw here? 
We should always return on l124?

src/workerd/server/server.c++
```diff
@@ -2267,10 +2293,26 @@ class Server::WorkerService final: public Service,
...
+              "container network needs to be defined in order enable containers on this durable object.");
+          KJ_ASSERT(config.hasContainerName(), "Container name is required");
+          KJ_ASSERT(config.hasImageTag(), "Image tag is required");
+          auto container_name = config.getContainerName();
+          auto image_tag = config.getImageTag();
+          containerClient = kj::heap<io::ContainerClient>(byteStreamFactory,
+              asHttpClient(networkPtr->startRequest({})),
+              asHttpClient(servicePtr->startRequest({})), kj::str(container_name),
+              kj::str(image_tag));
+        }
```
npaun: I'd suggest extracting this stuff into a function

src/workerd/io/docker-api.capnp
vicb, line 1: question: IIUC this file is used to encode he JSON messages with the Docker API? may I suggest to add a file level comment ?

src/workerd/io/container-client.c++
vicb: The template type - suggested than in a comment further down

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+    results.setUp(byteStreamFactory.kjToCapnp(kj::mv(upPipe.out)));
+    auto downEnd = byteStreamFactory.capnpToKj(context.getParams().getDown());
+    co_await kj::joinPromisesFailFast(
+        kj::arr(upEnd->pumpTo(*request.connection), request.connection->pumpTo(*downEnd)));
+  }
+
+ private:
+  capnp::ByteStreamFactory& byteStreamFactory;
+  kj::HttpClient& network;
+  [[maybe_unused]] kj::String containerId;
```
npaun: What is this unused field for?

src/workerd/io/container-client.c++
vicb: Does 404 mean not created?
(The question is better answered in a code comment)

src/workerd/server/server.c++
danlapid: Yeah, this service can only be used for one request without this, a true shame.

danlapid: Though we have a problem that the API wants to use this service for concurrent requests which seems to not work so maybe we'll need another approach anyway, we'll see.

src/workerd/io/container-client.c++
danlapid: ?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,277 @@
...
+      JSG_FAIL_REQUIRE(Error, "Create container failed with: ", response.body);
+    }
+  }
+}
+
+kj::Promise<void> ContainerClient::startContainer() {
+  // Docker API v1.50: POST /containers/{id}/start
+  auto endpoint = kj::str("/containers/", containerName, "/start");
+  kj::StringPtr body = "";
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint, body);
```
npaun: What is the reason for using an empty string as the body? It doesn't seem like this should be necessary?

src/workerd/server/workerd.capnp
vicb: I don't have a link - probably a discussion on chat ? Yagiz probably remembers.

_Agreed that an external service would also work._

src/workerd/io/container-client.c++
anonrig: This is required per docker API. It throws an error without it.

src/workerd/io/container-client.c++
anonrig: Agreed. Updating PR.

src/workerd/io/container-client.c++
anonrig: if it's not 200, we should throw. this is what this function does. it requires the following boolean expression to be truthy, and if not, it throws an error based on 2nd parameter with the 3rd parameter message.

src/workerd/io/container-client.c++
anonrig: Adding a comment now. Thanks

src/workerd/io/container-client.c++
anonrig: I've updated the comment. Thanks!

src/workerd/io/container-client.c++
anonrig: We omit an error when the container is already started.

src/workerd/server/workerd.capnp
anonrig: By default we use latest of an image. If we want to support it in the future, we can definitely support different tags

src/workerd/io/container-client.c++
npaun: Weird; maybe leave a comment about that.

src/workerd/io/container-client-test.js
anonrig: I prefer keeping it since, we also validate these inputs behavior here.

src/workerd/io/container-client-test.js
anonrig: If it throws it will stop the execution and the test will fail. No need to have a doesNotThrow here

npaun (approved):
LGTM. The Docker client is pretty clean and easy to understand.

src/workerd/io/container-client.c++
npaun: +1 I like this idea from a code style POV, as long as it's not too hard to do.

src/workerd/io/container-client-test.js
anonrig: It should be assert.rejects since container returns an async response

src/workerd/io/container-client.c++
danlapid: Kinda forces us to actually decode all of the different response types, we currently don't do that.

src/workerd/server/workerd.capnp
npaun: Sounds like this field is actually the image name?

src/workerd/io/container-client.c++
anonrig: There are endpoints where it doesn't return anything, such as DELETE container endpoint.

src/workerd/server/workerd.capnp
anonrig: yeah imageName might be a better name, agreed.

src/workerd/io/container-client.c++
vicb: Couldn't template specialization help here?
Or at least factor the JSON decoding part in a templated method?

src/workerd/io/container-client.c++
vicb: Oh got what you mean...
If a container already exists it's not an error because you will re-use it.
Is that it?

Maybe add a comment for people that are slow to understand as I am ;)

But what if the env variables or command is different?

src/workerd/server/workerd.capnp
vicb: Mostly extensibility, a container might exist without a DO at some point?
Looks like structs in this file are only nested if the child is a part of the parent.
But that only a suggestion/question.

emily-shen: fwiw, wrangler config doesn't nest container config under DO config, presumably for that reason

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,18 @@ struct Worker {
     # workerd uses SQLite to back all Durable Objects, but the SQL API is hidden by default to
     # emulate behavior of traditional DO namespaces on Cloudflare that aren't SQLite-backed. This
     # flag should be enabled when testing code that will run on a SQLite-backed namespace.
+
+    container @5 :ContainerOptions;
+    # Optional configuration options for local development container interaction.
+
+    struct ContainerOptions {
+      containerName @0 :Text;
```
emily-shen: the container name is per container instance, which is per DO instance - i don't think we can set this here?

src/workerd/io/container-client-test/container-client-test.js
emily-shen, line 1: can we add a test for spinning up multiple container instances from the same image?

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,334 @@
...
+#include <kj/async.h>
+#include <kj/compat/http.h>
+#include <kj/debug.h>
+#include <kj/encoding.h>
+#include <kj/exception.h>
+
+namespace workerd::io {
+
+template <typename T>
+typename T::Builder decodeJsonResponse(kj::String response) {
```
jasnell:
```suggestion
typename T::Builder decodeJsonResponse(kj::StringPtr response) {
```

Passing `response` as a `kj::StringPtr` should be safe here to avoid needing to move at the callsites

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,334 @@
...
+    kj::HttpHeaderTable headerTable;
+    kj::HttpHeaders headers(headerTable);
+
+    // Port mappings might have outdated mappings, we can't know if a connect request
+    // fails because the app hasn't finished starting up or because the mapping is outdated.
+    // To be safe we should inspect the container to get up to date mappings.
+    auto portMappings = kj::get<1>(co_await containerClient.inspectContainer());
+    auto maybeMappedPort = portMappings.find(containerPort);
+    if (maybeMappedPort == kj::none) {
+      throw KJ_EXCEPTION(DISCONNECTED, "connect(): Connection refused: container port not found");
```
jasnell: `FAILED` is probably better in this case than `DISCONNECTED`

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,334 @@
...
+    }
+    auto mappedPort = KJ_ASSERT_NONNULL(maybeMappedPort);
+
+    auto request = containerClient.network->connect(
+        kj::str(containerHost, ":", mappedPort), headers, kj::HttpConnectSettings());
+
+    auto status = co_await request.status;
+    if (status.statusCode != 200) {
+      throw KJ_EXCEPTION(DISCONNECTED, "connect(): Connection refused");
```
jasnell: Ditto... `FAILED` would be better here than `DISCONNECTED` I think.

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,334 @@
...
+};
+
+kj::Promise<ContainerClient::Response> ContainerClient::dockerApiRequest(
+    kj::HttpMethod method, kj::StringPtr endpoint, kj::Maybe<kj::StringPtr> body) {
+  kj::HttpHeaderTable::Builder builder;
+  const auto userAgent = builder.add("User-Agent");
+  const auto accept = builder.add("Accept");
+  const auto contentType = builder.add("Content-Type");
+  const auto contentLength = builder.add("Content-Length");
+  auto headerTable = builder.build();
```
jasnell: Probably better to build this once and reuse it rather than building on each call

src/workerd/io/container-client.c++
```diff
@@ -0,0 +1,334 @@
...
+      co_await body->write(requestBody.asBytes());
+    }
+    auto response = co_await req.response;
+    auto result = co_await response.body->readAllText();
+    co_return Response{.statusCode = response.statusCode, .body = kj::mv(result)};
+  } else {
+    auto req = httpClient->request(method, endpoint, headers);
+    { auto body = kj::mv(req.body); }
+    auto response = co_await req.response;
+    auto result = co_await response.body->readAllText();
```
jasnell: Is the amount of text returned here determined by us or users? If it is determined by users this could be problematic for memory management.

src/workerd/io/docker-api.capnp
```diff
@@ -0,0 +1,292 @@
...
+
+@0xfafd0bdf57acc528;
+using Cxx = import "/capnp/c++.capnp";
+using Json = import "/capnp/compat/json.capnp";
+
+$Cxx.namespace("workerd::docker_api");
+$Cxx.allowCancellation;
+
+
+struct Docker {
```
jasnell: It's best to avoid use of the trademarked name in this. I'd recommend `s/Docker/Container` and rename this file to `container-api.capnp`

src/workerd/io/container-client.c++
danlapid: It needs to be DISCONNECTED for it to match what connect is expecting.

src/workerd/io/container-client.c++
danlapid: It needs to be DISCONNECTED for it to match what connect is expecting.

src/workerd/io/container-client.c++
danlapid: Determined by the docker api. I don't really think we should worry too much about memory management here.
If the user's docker service is returning a response that's bigger than their computer's memory budget it's really their problem.

src/workerd/io/docker-api.capnp
danlapid: This is literally docker's api, there's no getting around that.

jasnell: Well, this part is in our capnp files :-)

danlapid:
@kentonv this PR needs your review as well please

src/workerd/server/workerd.capnp
kentonv: I think the container names should be something like `workerd-${uniqueKey}-${actorId}`, where `uniqueKey` is the existing property of `DurableObjectNamespace` (see above) which is normally used as a stable identifier for storage. It makes sense that container names should similarly be stable.

So I don't think we need any way to configure the container name here -- only the image.

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,18 @@ struct Worker {
     # workerd uses SQLite to back all Durable Objects, but the SQL API is hidden by default to
     # emulate behavior of traditional DO namespaces on Cloudflare that aren't SQLite-backed. This
     # flag should be enabled when testing code that will run on a SQLite-backed namespace.
+
+    container @5 :ContainerOptions;
+    # Optional configuration options for local development container interaction.
```
kentonv: This comment is not very clear. How about:

> If present, Durable Objects in this namespace have attached containers. workerd will talk to Docker running on the same machine to start containers for each Durable Object based on the given image. The Durable Object can access the container via the `ctx.container` API. [link to docs]

src/workerd/io/BUILD.bazel
```diff
@@ -308,6 +308,31 @@ wd_capnp_library(
     ],
 )
 
+wd_capnp_library(src = "docker-api.capnp")
+
+wd_cc_library(
+    name = "container-client",
```
kentonv: Since this library is only used by workerd, I would move it into `workerd/server`. `workerd/io` is meant to be basic isolate lifecycle and I/O management and shared with the internal codebase.

src/workerd/server/server.h
```diff
@@ -102,10 +102,12 @@ class Server final: private kj::TaskSet::ErrorHandler {
     kj::String uniqueKey;
     bool isEvictable;
     bool enableSql;
+    kj::Maybe<config::Worker::DurableObjectNamespace::ContainerOptions::Reader> containerOptions;
   };
   struct Ephemeral {
     bool isEvictable;
     bool enableSql;
+    kj::Maybe<config::Worker::DurableObjectNamespace::ContainerOptions::Reader> containerOptions;
```
kentonv: There's no need to support containers for ephemeral objects.

src/workerd/server/server.c++
```diff
@@ -703,13 +705,16 @@ class Server::ExternalHttpService final: public Service, private kj::TaskSet::Er
...
+      }
       if (parent.rewriter->needsRewriteRequest()) {
         auto rewrite = parent.rewriter->rewriteOutgoingRequest(url, headers, metadata.cfBlobJson);
-        return parent.serviceAdapter->request(method, url, *rewrite.headers, requestBody, *this)
-            .attach(kj::mv(rewrite));
+        co_return co_await parent.serviceAdapter->request(method, url, *rewrite.headers,
+            requestBody, parent.rewriter->needsRewriteResponse() ? *this : response);
       } else {
-        return parent.serviceAdapter->request(method, url, headers, requestBody, *this);
+        co_return co_await parent.serviceAdapter->request(method, url, headers, requestBody,
```
kentonv: There are a bunch of changes in this file which don't seem related to containers. Some of them appear to be converting methods to coroutines for no particular reason.

If these are not strictly needed for the containers change, please revert these changes. If they are needed, please explain why.

If these are just separate refactorings you wanted to do, please make a separate PR for them. Please note, though, replacing `return` with `co_return co_await` hurts performance and may have unintended side effects; we should not do it unless it actually has some other benefit to the code.

src/workerd/server/workerd.capnp
```diff
@@ -651,8 +663,17 @@ struct Worker {
   streamingTails @15 :List(ServiceDesignator);
   # List of streaming tail worker services that should receive tail events for this worker.
   # NOTE: This will be deleted in a future refactor, do not depend on this.
+
+  containerService @16 :ServiceDesignator;
```
kentonv: We don't need to make people configure `containerService` and `containerNetwork` like this. We can just take the `kj::Network` off the `AsyncIoProvider` and use it directly.

We should probably have some way for people to explicitly say that containers should be implemented using docker on the local machine, so we're not just automatically connecting to docker unexpectedly, and we can introduce other container engines later if desired. But this can simply be like:

```
containerEngine :union {
  none @16: void;
  localDocker @17: void;
}
```

(Use a union rather than an enum because some future engine may require more configuration.)

src/workerd/server/workerd.capnp
```diff
@@ -651,6 +663,20 @@ struct Worker {
...
+    # No container engine configured. Container operations will not be available.
+
+    localDocker @17 :DockerConfiguration;
+    # Use local Docker daemon for container operations.
+    # Only used for local development and testing purposes.
+  }
+
+  struct DockerConfiguration {
+    socketPath @0 :Text;
+    # Path to the Docker socket. Defaults to the standard Unix socket path.
```
emily-shen: i don't see this default being set anywhere, did i miss it or is miniflare supposed to be setting this default?

src/workerd/server/container-client.c++
```diff
@@ -0,0 +1,404 @@
...
+    kj::HttpHeaderTable headerTable;
+    kj::HttpHeaders headers(headerTable);
+
+    // Port mappings might have outdated mappings, we can't know if a connect request
+    // fails because the app hasn't finished starting up or because the mapping is outdated.
+    // To be safe we should inspect the container to get up to date mappings.
+    auto portMappings = kj::get<1>(co_await containerClient.inspectContainer());
+    auto maybeMappedPort = portMappings.find(containerPort);
+    if (maybeMappedPort == kj::none) {
+      throw KJ_EXCEPTION(DISCONNECTED, "connect(): Connection refused: container port not found");
```
emily-shen: this is surfaced to the user right? can we add a bit about making sure you've exposed the port in your dockerfile

src/workerd/server/tests/container-client/test.js
```diff
@@ -0,0 +1,269 @@
...
+  async test(_ctrl, env) {
+    for (const CONTAINER of [env.MY_CONTAINER, env.MY_DUPLICATE_CONTAINER]) {
+      const id = CONTAINER.idFromName('testBasics');
+      const stub = CONTAINER.get(id);
+      await stub.testBasics();
+    }
+  },
+};
+
+// Test container persistence across durable object instances
```
emily-shen, line 202: i thought in prod if a DO instance shut down the container would also shut down? (not if it hibernated though, maybe we could test that if you want)
@gabivlj can you confirm? or is this an intentional difference in local dev?

src/workerd/server/tests/container-client/test.js
```diff
@@ -0,0 +1,269 @@
...
+    }
+  },
+};
+
+// Test basic container functionality
+export const testBasics = {
+  async test(_ctrl, env) {
+    for (const CONTAINER of [env.MY_CONTAINER, env.MY_DUPLICATE_CONTAINER]) {
+      const id = CONTAINER.idFromName('testBasics');
+      const stub = CONTAINER.get(id);
```
emily-shen: what about a test that two container instances from one DO namespace, at the same time, works?
(that's what i meant in my earlier comment 😅)

src/workerd/server/workerd.capnp
danlapid: Nope, just bad phrasing, should delete that part of the comment

src/workerd/server/container-client.c++
danlapid: Ah damn this is going to be annoying
This is surfaced to the user as "Network Connection Lost" for reasons of how the stack works.
I need to think about how to do this right

src/workerd/server/workerd.capnp
```diff
@@ -601,6 +601,18 @@ struct Worker {
     # workerd uses SQLite to back all Durable Objects, but the SQL API is hidden by default to
     # emulate behavior of traditional DO namespaces on Cloudflare that aren't SQLite-backed. This
     # flag should be enabled when testing code that will run on a SQLite-backed namespace.
+
+    container @5 :ContainerOptions;
+    # If present, Durable Objects in this namespace have attached containers.
+    # workerd will talk to Docker running on the same machine to start containers for each
```
vicb:
```suggestion
    # workerd will talk to the configured container engine to start containers for each
```

src/workerd/server/container-client.c++
anonrig: If we use `JSG_KJ_EXCEPTION` now it can be caught using try/catch and not always exposed to the user, like KJ_EXCEPTION is.

anonrig: If we use `JSG_KJ_EXCEPTION` now it can be caught using try/catch and not always exposed to the user, like KJ_EXCEPTION is.

danlapid: I don't really see how that helps here

anonrig: This is now exposed as the following:

```
Error: connect(): Connection refused: container port not found. Make sure you exposed the port in your container definition.
    at async DurableObjectExample.testBasics (worker:32:18) {
  retryable: true
}
```

src/workerd/server/container-client.c++
```diff
@@ -0,0 +1,406 @@
...
+    auto downEnd = containerClient.byteStreamFactory.capnpToKj(context.getParams().getDown());
+    pumpTask =
+        kj::joinPromisesFailFast(kj::arr(upEnd->pumpTo(*connection), connection->pumpTo(*downEnd)))
+            .ignoreResult()
+            .attach(kj::mv(upEnd), kj::mv(connection), kj::mv(downEnd));
+    co_return;
+  }
+
+ private:
+  ContainerClient& containerClient;
```
jasnell, line 158: Given that this is a bare reference, a comment here indicating what owns the `ContainerClient` and keeps it alive would be helpful for others coming into this code later.

src/workerd/server/container-client.c++
```diff
@@ -0,0 +1,406 @@
...
+    { auto body = kj::mv(req.body); }
+    auto response = co_await req.response;
+    auto result = co_await response.body->readAllText();
+    co_return Response{.statusCode = response.statusCode, .body = kj::mv(result)};
+  }
+}
+
+kj::Promise<kj::Tuple<bool, kj::HashMap<uint16_t, uint16_t>>> ContainerClient::inspectContainer() {
+  // Docker API: GET /containers/{id}/json
+  auto endpoint = kj::str("/containers/", containerName, "/json");
```
jasnell, line 196: Are we validating that `containerName` does not have any characters that would make the URL invalid? For instance, a `#` or `?` or `/` that is unescaped anywhere in the name.

src/workerd/server/container-client.c++
```diff
@@ -0,0 +1,406 @@
...
+  auto hostConfig = jsonRoot.initHostConfig();
+  // We need to publish all ports to properly get the mapped port number locally
+  hostConfig.setPublishAllPorts(true);
+  // We need to set a restart policy to avoid having ambiguous states
+  // where the container we're managing is stuck at "exited" state.
+  hostConfig.initRestartPolicy().setName("on-failure");
+
+  // Encode to JSON string
+  kj::String jsonBody = codec.encode(jsonRoot);
+  const auto endpoint = kj::str("/containers/create?name=", containerName);
```
jasnell, line 279: Is the `containerName` guaranteed not to have any special characters here? e.g. no `#` or `&`

src/workerd/server/container-client.h
```diff
@@ -0,0 +1,71 @@
...
+
+  struct Response {
+    kj::uint statusCode;
+    kj::String body;
+  };
+
+  // Docker API v1.50 helper methods
+  kj::Promise<Response> dockerApiRequest(
+      kj::HttpMethod method, kj::StringPtr endpoint, kj::Maybe<kj::StringPtr> body = kj::none);
+  kj::Promise<kj::Tuple<bool, kj::HashMap<uint16_t, uint16_t>>> inspectContainer();
```
jasnell: Knowing how @kentonv feels about using tuples, is there a particular reason why this is returning a tuple rather than a struct?

src/workerd/server/container-client.c++
anonrig: Definitely. Nice catch. Thanks @jasnell.

src/workerd/server/container-client.c++
anonrig: You're right. I just added a `kj::encodeUriComponent()` call to the constructor to fix this.

src/workerd/server/container-client.h
anonrig: Updating.

src/workerd/server/tests/container-client/test.js
anonrig: As mentioned in todays meeting, I'll do a follow up once this PR lands.

src/workerd/server/container-client.c++
```diff
@@ -0,0 +1,411 @@
...
+}
+
+kj::Promise<void> ContainerClient::startContainer() {
+  // Docker API: POST /containers/{id}/start
+  const auto endpoint = kj::str("/containers/", containerName, "/start");
+  // We have to send an empty body since docker API will throw an error if we don't.
+  kj::StringPtr body = "";
+  auto response = co_await dockerApiRequest(kj::HttpMethod::POST, endpoint, body);
+  // statusCode 204 refers to "no error"
+  // statusCode 304 refers to "container already started"
```
emily-shen, line 299: if start is called and the container has already started we should error: https://github.com/cloudflare/workerd/blob/main/src/workerd/io/container.capnp#L19

emily-shen (approved):
lgtm, other than that one comment about `start` that can be addressed in a followup along with the proper shutdown behaviour.

thank you for all the work on this!

src/workerd/server/container-client.c++
```diff
@@ -0,0 +1,411 @@
...
+      auto obj = array[0].getObject();
+      JSG_REQUIRE(array.size() > 1, Error, "Malformed ContainerInspect port mapping object");
+      auto mappedPort = obj[1].getValue().getString();
+      number = mappedPort.asString().parseAs<uint16_t>();
+    }
+    portMappings.insert(portNumber, number);
+  }
+
+  // Look for Status field in the JSON object
+  JSG_REQUIRE(jsonRoot.hasState(), Error, "Malformed ContainerInspect response");
```
jasnell, line 237: Not critical given that this is local dev only, but these error messages aren't very actionable for users.

jasnell (approved):
Code LGTM.

anonrig:
Rebasing and squashing.

src/workerd/server/container-client.c++
anonrig: I'll do a follow up. Thanks!

src/workerd/server/container-client.c++
anonrig: I don't think there is anything actionable here if the docker api returns a malformed response.

src/workerd/server/container-client.c++
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  co_await createContainer(entrypoint, environment);
+  co_await startContainer();
+}
+
+kj::Promise<void> ContainerClient::monitor(MonitorContext context) {
+  // Docker API: POST /containers/{id}/wait - wait for container to exit
+  const auto endpoint = kj::str("/containers/", containerName, "/wait");
+  // Monitor is often called right after start but the api layer's start does not await the RPC's
```
gabivlj, line 353: Not a blocker on Cloudchamber's side, but couldn't we wait for any pending start/other rpc calls to exit here? In our service we prevented this by only allowing one in-flight request on the Container capability (with the exception of an inflight Monitor() not blocking other Container RPC calls).

anonrig: Let's discuss this on the group chat. Resolving the comment to land the pull-request.

gabivlj (approved):

PR #4283, opened 2025-06-09
Add prebuilt platform and cross compilation alias

ketanhwr:


src/pyodide/BUILD.bazel
```diff
@@ -16,6 +16,16 @@ alias(
...
+# Cross alias for prebuilt bundle
+alias(
+    name = "pyodide.capnp.bin_cross",
+    actual = select({
+        "@//build/config:prebuilt_binaries_arm64": "@//:bin.arm64/tmp/pyodide/pyodide.capnp.bin.aarch64-linux-gnu",
+        "//conditions:default": "pyodide.capnp.bin",
+    }),
+    visibility = ["//visibility:public"],
+)
+
```
dom96: Hmm, why is this necessary? Doesn't our capnp.bin support all CPU architectures?

ketanhwr: Discussed offline - this isn't to avoid any particular architecture, but to use a prebuilt bundle incase we have a runner which only uses cross-compiled artifacts. Added a comment explaining it

fhanau (approved):

PR #4287, opened 2025-06-09
[o11y] Add tail worker autogates

fhanau:
These are not used yet but may both be needed within workerd.

~Downstream PR to follow.~ Also see the downstream PR.


dom96:
Will these only be used in the downstream PR? If so you may wish to define them there instead of in workerd.

fhanau:
> Will these only be used in the downstream PR? If so you may wish to define them there instead of in workerd.

One of these will definitely be needed in workerd, for the other one it's possible that we'll need it so to avoid future churn it feels better to add both of them here.

danlapid (approved):

PR #4289, opened 2025-06-09
Implement even more node:fs and tests

jasnell:


src/node/internal/internal_fs_sync.ts
```diff
@@ -172,8 +177,20 @@ export function closeSync(fd: number): void {
 export function copyFileSync(
   src: FilePath,
   dest: FilePath,
-  _mode: number
+  mode: number = 0
 ): void {
+  validateOneOf(mode, 'mode', [
+    0,
```
anonrig, line 183: What does 0 stand for?

anonrig (approved):

src/node/internal/internal_fs_sync.ts
jasnell: Means nothing, no flag set.

src/node/internal/internal_fs_callback.ts
```diff
@@ -168,13 +190,7 @@ export function appendFile(
   } else {
     options = optionsOrCallback;
```
guybedford, line 191: In this case do we still need to set the `'a'` flag here?

PR #4290, opened 2025-06-09
Refactor _bazel.yml and WPT ci

npaun:


src/wpt/harness/test.ts
```diff
@@ -242,11 +242,11 @@ class PromiseTest extends Test {
 }
 
 globalThis.promise_test = (func, name, properties): void => {
-  if (maybeAddSkippedTest(name)) {
+  if (maybeAddSkippedTest(name ?? '')) {
     return;
   }
 
-  const testCase = new PromiseTest(name, properties);
+  const testCase = new PromiseTest(name ?? '', properties);
```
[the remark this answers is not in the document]
npaun: This is a temporary fix; WPT actually requires us to later update the test's name to the test's message in some cases if it had no name.

.github/workflows/wpt-report.yml
```diff
@@ -1,27 +1,101 @@
...
+      image: ${{ matrix.image }}
+      os_name: ${{ matrix.os-name }}
+      phase: '-release-'
+      extra_bazel_args: '--strip=always --config=${{matrix.bazel-config}}'
+      arch_name: ${{ matrix.target-arch }}
+      run_tests: false
+      upload_binary: true
+      upload_test_logs: true
+
+  upload-artifacts:
```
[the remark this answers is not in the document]
npaun: Note: This file will be deleted. I'm just using it to check that release.yml will work ok once this PR is applied.

.github/workflows/wpt-report.yml
```diff
@@ -1,27 +1,101 @@
...
+            # As of release v1.20230404.0 the Linux x64 binary after debug_strip is 65.8 MB,
+            # 21.0 MB with gzip and 17.3 MB with brotli -9. Use gzip as a widely supported format
+            # which still produces an acceptable compressed size.
+            gzip -9N -k /tmp/workerd${{ matrix.arch == 'windows-64' && '.exe' || '' }}
+      - run: mv /tmp/workerd${{ matrix.arch == 'windows-64' && '.exe' || '' }}.gz /tmp/workerd-${{ matrix.arch }}.gz
+      # Upload compressed release binaries – one set of artifacts is sufficient with gzip being
+      # widely supported
+      - name: Upload Release Assets
+        id: upload-release-asset
+        run: file /tmp/workerd*
```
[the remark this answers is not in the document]
npaun: Note: This file will be deleted. I'm just using it to check that release.yml will work ok once this PR is applied.

.github/workflows/_wpt.yml
```diff
@@ -0,0 +1,42 @@
...
+          path: testlogs
+      - name: Generate WPT report and stats
+        run: ./tools/cross/wpt_logs.py --print-stats --write-report=wpt-report.json testlogs/ >> $GITHUB_STEP_SUMMARY
+      - name: Upload WPT report
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.report_artifact }}
+          path: wpt-report.json
+
+
```
anonrig: Lots of extra spaces here

.github/workflows/release.yml
```diff
@@ -70,100 +70,52 @@ jobs:
...
-          - os-name: macOS-arm64
-            os: macos-15
+            target-arch: X64
+          - title: macOS-arm64
+            os-name: macOS
+            image: macos-15
             bazel-config: release_macos
             target-arch: ARM64
-          - os-name: windows
-            os: windows-2022
```
anonrig, line 95: Why did we remove these

npaun: It's still the same set of OSs, I just changed some field names.

.github/workflows/_wpt.yml
```diff
@@ -15,6 +15,11 @@ on:
         required: true
         type: string
 
+permissions:
```
fhanau, line 18: What are these needed for?

.github/workflows/_bazel.yml
```diff
@@ -94,6 +110,13 @@ jobs:
       - name: Setup macOS
         if: runner.os == 'macOS'
         run: |
+          # Install lld and link it to /usr/local/bin. We overwrite any existing link, which may
```
fhanau, line 122: This will slow down test CI – installing lld to get smaller binaries is not really relevant to test builds, enabling ICF in particular will slow down linking. This would have to be an opt-in configuration.

.github/workflows/_wpt.yml
npaun: The Github Advanced Security ™️  bot yells at you if you don't add permissions to workflows, now.

.github/workflows/_bazel.yml
```diff
@@ -117,17 +116,13 @@ jobs:
...
         if: inputs.run_tests
         run: |
-          bazel test ${{ inputs.extra_bazel_args }} --config=ci --config=ci-${{ inputs.os_name }}${{ inputs.suffix }} --remote_cache=https://bazel:${{ secrets.BAZEL_CACHE_KEY }}@bazel-remote-cache.devprod.cloudflare.dev ${{ inputs.test_target }}
-      - name: WPT report
-        if: inputs.run_wpt_report
-        run: |
-          tools/cross/wpt_logs.py --print-stats --write-report=wpt-report.json >> $GITHUB_STEP_SUMMARY
-      - name: Upload WPT report
-        if: inputs.run_wpt_report
+          bazel test ${{ inputs.extra_bazel_args }} --config=ci --config=ci-${{ inputs.os_name }}${{ inputs.suffix }} --remote_cache=https://bazel:${{ secrets.BAZEL_CACHE_KEY }}@bazel-remote-cache.devprod.cloudflare.dev --cache_test_results=no ${{ inputs.test_target }}
```
fhanau: Not caching test results: This will make CI a lot slower. I don't think we need it either since for any given PR we are not expecting relevant changes to the logs of cached tests by definition.

npaun: I improved this - now we only skip caching if you enable `upload_test_logs`. So I can put this on only one of the builds (test-linux), so we can do the WPT report.

fhanau: Ok, that's certainly "less bad" – test-linux is not the bottleneck atm, so the impact will be lower. Non-blocking but ideally, we'd just handle the case of missing WPT tests due to caching gracefully and not put out a report in that case, a prior CI run will already have the results.

npaun: Hmm this gave me an idea:
- Allow test caching for regular PRs. The WPT report on a regular PR would only include the tests that are run (or no report if none of the WPT tests ran)
- For the release build, always run the WPT tests without caching to make a full report.

npaun:
There's way too much noise in this PR. I'm going to delete every bot comment.

fhanau (approved):
Great work – I didn't see any major issues despite the size of the change. Did not look closely at the ts/python changes for WPT, hopefully the other reviewers will be able to do so.
- os_name is upper case for release yet lower case for test (e.g. linux vs. Linux), but as far as I can see that's benign
- I'm not sure if the `build:linux --action_env=CC=/usr/lib/llvm-19/bin/clang` block is needed, but it also seems unlikely to be harmful.
- For binary names we're using matrix.target-arch instead of runner.arch now – if that happens to follow a different scheme that could be an issue, I assume you've checked that these are the same as before.

npaun:
>  Did not look closely at the ts/python changes for WPT, hopefully the other reviewers will be able to do so.

No worries, Yagiz looked at those.

> os_name is upper case for release yet lower case for test (e.g. linux vs. Linux), but as far as I can see that's benign
> For binary names we're using matrix.target-arch instead of runner.arch now – if that happens to follow a different scheme that could be an issue, I assume you've checked that these are the same as before.

These all derive from existing inconsistencies between release and test. I decided not to unify them in this PR just to have less worries about breaking stuff.


|                  | os_name     |   architecture |
|-----------|-------------|--------------|
| Test          | Custom list; values don't match Github | `runner.arch` |
| Release   | `runner.os` |  Custom list; values match `runner.arch` |

tools/cross/wpt_logs.py
```diff
@@ -119,20 +139,11 @@ def stats_table(logs: list[Log]) -> str:
...
-    for log in logs:
-        if not log.stats:
-            continue
-
-        cells = " ".join(f"<td align='right'>{value}</td>" for value in log.stats[1:])
-        table += f"<tr><td>{log.stats[0]}</td>{cells}</tr>\n"
-
-    table += """
+    </tr>
+    {cells_html}
```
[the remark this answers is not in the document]
npaun: wat

npaun:
I dropped the check-snapshot changes from this PR as they can be applied independently in another PR. They stopped working for some reason and I couldn't figure it out from a quick glance.

PR #4294, opened 2025-06-10
Add support for readable streams in AI binding

thatsKevinJain:
Workers AI is extending its binding to allow for passing ReadableStreams as inputs,
this will only be supported for a few models and will be fully backwards compatible.

For newer models, users will be able to send bigger media files as raw JS ReadableStream
objects. This will result in faster inference responses.

### DIFF
--- a/src/cloudflare/internal/ai-api.ts
+++ b/src/cloudflare/internal/ai-api.ts
@@ -108,20 +108,13 @@ async function blobToBase64(blob: Blob): Promise<string> {
   return btoa(binary);
 }
 
-export function isReadableStream(value: unknown): boolean {
-  if (value instanceof ReadableStream) {
-    return true;
-  }
-
-  if (value && typeof value === 'object') {
-    const obj = value as Record<string, unknown>;
-    return (
-      typeof obj['getReader'] === 'function' &&
-      typeof obj['locked'] === 'boolean'
-    );
-  }
-
-  return false;
+function isReadableStream(obj: unknown): obj is ReadableStream {
+  return !!(
+    obj &&
+    typeof obj === 'object' &&
+    'getReader' in obj &&
+    typeof obj.getReader === 'function'
+  );
 }
 
 /**
@@ -138,11 +131,9 @@ function findReadableStreamKeys(
       value &&
       typeof value === 'object' &&
       'body' in value &&
-      isReadableStream((value as AiInputReadableStream).body);
+      isReadableStream(value?.body);
 
-    const isDirectReadableStream = isReadableStream(value);
-
-    if (hasReadableStreamBody || isDirectReadableStream) {
+    if (hasReadableStreamBody || isReadableStream(value)) {
       readableStreamKeys.push(key);
     }
   }
@@ -233,6 +224,12 @@ export class Ai {
       const body = (stream as AiInputReadableStream).body;
       const contentType = (stream as AiInputReadableStream).contentType;
 
+      if (options.gateway?.id) {
+        throw new AiInternalError(
+          'AI Gateway does not support ReadableStreams yet.'
+        );
+      }
+
       // Make sure user has supplied the Content-Type
       // This allows AI binding to treat the ReadableStream correctly
       if (!contentType) {
@@ -261,17 +258,15 @@ export class Ai {
       // Append inputs with ai.run options that are passed to the inference request
       const query = {
         ...cleanedOptions,
+        version: '3',
         userInputs: JSON.stringify({ ...userInputs }),
       };
-      const queryParams = new URLSearchParams(query).toString();
-
-      const endpointUrl = `https://workers-binding.ai/run?version=3&${queryParams}`;
-      if (options.gateway?.id) {
-        throw new AiInternalError(
-          'AI Gateway does not support ReadableStreams yet.'
-        );
+      const aiEndpoint = new URL('https://workers-binding.ai/run');
+      for(const [key, value] of Object.entries(query)) {
+        aiEndpoint.searchParams.set(key, value);
       }
-      res = await this.fetcher.fetch(endpointUrl, fetchOptions);
+
+      res = await this.fetcher.fetch(aiEndpoint, fetchOptions);
     }
 
     this.lastRequestId = res.headers.get('cf-ai-req-id');
--- a/src/cloudflare/internal/ai-api.ts
+++ b/src/cloudflare/internal/ai-api.ts
@@ -34,6 +34,11 @@ export type AiOptions = {
   sessionOptions?: SessionOptions;
 };
 
+export type AiInputReadableStream = {
+  body: ReadableStream;
+  contentType: string;
+}
+
 export type ConversionResponse = {
   name: string;
   mimeType: string;
@@ -103,6 +108,26 @@ async function blobToBase64(blob: Blob): Promise<string> {
   return btoa(binary);
 }
 
+export function isReadableStream(value: any): boolean {
+  return (
+    value instanceof ReadableStream ||
+    (value && typeof value?.getReader === "function" && typeof value?.locked === "boolean")
+  );
+}
+
+/**
+ * Find keys in inputs that have a ReadableStream
+ * */
+function findReadableStreamKeys(inputs: Record<string, object>): Array<string> {
+  const readableStreamKeys = [];
+  for (const [key, value] of Object.entries(inputs)) {
+    if (isReadableStream((value as AiInputReadableStream)?.body) || isReadableStream(value instanceof ReadableStream)) {
+      readableStreamKeys.push(key);
+    }
+  }
+  return readableStreamKeys;
+}
+
 export class Ai {
   private readonly fetcher: Fetcher;
 
@@ -135,7 +160,7 @@ export class Ai {
   ): Promise<Response | ReadableStream<Uint8Array> | object | null> {
     this.options = options;
     this.lastRequestId = '';
-
+ 
     // This removes some unwanted options from getting sent in the body
     const cleanedOptions = (({
       prefix,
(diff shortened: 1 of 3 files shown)


src/cloudflare/internal/ai-api.ts
```diff
@@ -103,6 +108,31 @@ async function blobToBase64(blob: Blob): Promise<string> {
...
+}
+
+/**
+ * Find keys in inputs that have a ReadableStream
+ * */
+function findReadableStreamKeys(inputs: Record<string, object>): Array<string> {
+  const readableStreamKeys = [];
+  for (const [key, value] of Object.entries(inputs)) {
+    if (
+      isReadableStream((value as AiInputReadableStream)?.body) ||
```
JoaquinGimenez1: I think this is an unsafe type casting. I think we should use `unknown` instead

src/cloudflare/internal/ai-api.ts
```diff
@@ -103,6 +108,31 @@ async function blobToBase64(blob: Blob): Promise<string> {
...
+      typeof value?.getReader === 'function' &&
+      typeof value?.locked === 'boolean')
+  );
+}
+
+/**
+ * Find keys in inputs that have a ReadableStream
+ * */
+function findReadableStreamKeys(inputs: Record<string, object>): Array<string> {
+  const readableStreamKeys = [];
```
JoaquinGimenez1: Since we are setting the array to empty, it would be nice to add the type too. 
```
const readableStreamKeys: string[] = [];
```

src/cloudflare/internal/ai-api.ts
JoaquinGimenez1: A one line if looks cool but I think we should split the condition into named variables to make it easier to read/understand

src/cloudflare/internal/ai-api.ts
```diff
@@ -103,6 +108,31 @@ async function blobToBase64(blob: Blob): Promise<string> {
...
+    (value &&
+      typeof value?.getReader === 'function' &&
+      typeof value?.locked === 'boolean')
+  );
+}
+
+/**
+ * Find keys in inputs that have a ReadableStream
+ * */
+function findReadableStreamKeys(inputs: Record<string, object>): Array<string> {
```
JoaquinGimenez1: I think this makes more sense`Record<string, unknown>`

src/cloudflare/internal/ai-api.ts
```diff
@@ -144,29 +174,84 @@ export class Ai {
...
-      body: body,
-      headers: {
-        ...this.options.sessionOptions?.extraHeaders,
-        ...this.options.extraHeaders,
-        'content-type': 'application/json',
-        'cf-consn-sdk-version': '2.0.0',
-        'cf-consn-model-id': `${this.options.prefix ? `${this.options.prefix}:` : ''}${model}`,
-      },
-    };
+      const fetchOptions = {
```
JoaquinGimenez1, line 200: We have two almost identical `fetchOptions`, I think we can make merge them and only change what we need instead

src/cloudflare/internal/ai-api.ts
```diff
@@ -144,29 +174,84 @@ export class Ai {
...
+      }
 
-    let endpointUrl = 'https://workers-binding.ai/run?version=3';
-    if (options.gateway?.id) {
-      endpointUrl = 'https://workers-binding.ai/ai-gateway/run?version=3';
-    }
+      res = await this.fetcher.fetch(endpointUrl, fetchOptions);
+    } else if (streamKeys.length > 1) {
+      throw new AiInternalError(
+        `Muliple ReadableStreams [${streamKeys}] are not supported`
```
JoaquinGimenez1: What do you think about  ?
```
  `Multiple ReadableStreams are not supported. Found streams in keys: [${streamKeys.join(', ')}]`

```

src/cloudflare/internal/ai-api.ts
```diff
@@ -144,29 +174,84 @@ export class Ai {
...
+          ...this.options.extraHeaders,
+          'content-type': (inputs[streamKeys[0]!] as AiInputReadableStream)
+            ?.contentType,
+          'cf-consn-sdk-version': '2.0.0',
+          'cf-consn-model-id': `${this.options.prefix ? `${this.options.prefix}:` : ''}${model}`,
+        },
+      };
+
+      // Remove the stream body
+      delete inputs[streamKeys[0]!];
```
JoaquinGimenez1: This is mutating the original input object. I think this could cause issues down the road

src/cloudflare/internal/ai-api.ts
thatsKevinJain: Yeah I like that, updated!

src/cloudflare/internal/ai-api.ts
thatsKevinJain: They are slightly different and I am defining them separately so it's more verbose.

src/cloudflare/internal/ai-api.ts
thatsKevinJain: Updated

thatsKevinJain: Updated

src/cloudflare/internal/ai-api.ts
thatsKevinJain: Updated

src/cloudflare/internal/ai-api.ts
thatsKevinJain: Updated

src/cloudflare/internal/ai-api.ts
```diff
@@ -103,6 +108,48 @@ async function blobToBase64(blob: Blob): Promise<string> {
   return btoa(binary);
 }
 
+export function isReadableStream(value: unknown): boolean {
```
anonrig: src/cloudflare/internal/images-api.ts has a somewhat similar implementation. can you merge the implementations if it doesn't work for you

src/cloudflare/internal/ai-api.ts
```diff
@@ -103,6 +108,48 @@ async function blobToBase64(blob: Blob): Promise<string> {
...
+    // Check if value has a body property that's a ReadableStream
+    const hasReadableStreamBody =
+      value &&
+      typeof value === 'object' &&
+      'body' in value &&
+      isReadableStream((value as AiInputReadableStream).body);
+
+    const isDirectReadableStream = isReadableStream(value);
+
+    if (hasReadableStreamBody || isDirectReadableStream) {
```
anonrig:
```suggestion
    if (hasReadableStreamBody || isReadableStream(value)) {
```

src/cloudflare/internal/ai-api.ts
```diff
@@ -103,6 +108,48 @@ async function blobToBase64(blob: Blob): Promise<string> {
...
+): Array<string> {
+  const readableStreamKeys: Array<string> = [];
+
+  for (const [key, value] of Object.entries(inputs)) {
+    // Check if value has a body property that's a ReadableStream
+    const hasReadableStreamBody =
+      value &&
+      typeof value === 'object' &&
+      'body' in value &&
+      isReadableStream((value as AiInputReadableStream).body);
```
anonrig: isReadableStream takes unknown. I believe this is unnecessary:

```suggestion
      isReadableStream((value).body);
```

src/cloudflare/internal/ai-api.ts
```diff
@@ -144,29 +191,88 @@ export class Ai {
...
+
+      // Construct query params
+      // Append inputs with ai.run options that are passed to the inference request
+      const query = {
+        ...cleanedOptions,
+        userInputs: JSON.stringify({ ...userInputs }),
+      };
+      const queryParams = new URLSearchParams(query).toString();
+
+      const endpointUrl = `https://workers-binding.ai/run?version=3&${queryParams}`;
```
anonrig: Can you just parse the URL using new URL() and later call parsedUrl.searchParams = new URLSearchParams(query); and just pass the URL to fetch method as the first argument?

src/cloudflare/internal/ai-api.ts
```diff
@@ -144,29 +191,88 @@ export class Ai {
...
+      if (options.gateway?.id) {
+        throw new AiInternalError(
+          'AI Gateway does not support ReadableStreams yet.'
+        );
+      }
```
anonrig: I think you should throw this error a lot earlier than where it is right now.

src/cloudflare/internal/test/ai/ai-api-test.js
```diff
@@ -98,6 +98,136 @@ export const tests = {
...
     }
 
+    {
+      // Test one readable stream input
+      const arr = [1, 2, 3];
+      const resp = await env.ai.run('readableStreamIputs', {
+        audio: {
+          body: new ReadableStream({
+            start(controller) {
+              const encoder = new TextEncoder();
```
anonrig: Can you initialize this in line 102?

src/cloudflare/internal/test/ai/ai-api-test.js
```diff
@@ -98,6 +98,136 @@ export const tests = {
...
+
+    {
+      // Test errors from one readable stream input without content-type
+      try {
+        const arr = [1, 2, 3];
+        const resp = await env.ai.run('readableStreamIputs', {
+          audio: {
+            body: new ReadableStream({
+              start(controller) {
+                const encoder = new TextEncoder();
```
anonrig: Can you define this in line 161?

src/cloudflare/internal/test/ai/ai-api-test.js
```diff
@@ -98,6 +98,136 @@ export const tests = {
...
+          version: '3',
+        },
+        requestUrl:
+          'https://workers-binding.ai/run?version=3&userInputs=%7B%22detect_language%22%3Atrue%2C%22prompt%22%3A%22test+prompt%22%7D',
+      });
+    }
+
+    {
+      // Test errors from one readable stream input without content-type
+      try {
```
anonrig: This whole for try/catch can be eliminated by calling 

```
await assert.rejects()
```

src/cloudflare/internal/test/ai/ai-mock.js
```diff
@@ -80,7 +88,17 @@ export default {
...
-    const data = await request.json();
+    const reqContentType = request.headers.get('content-type');
+
+    let data = {};
+    if (reqContentType === 'application/json') {
+      data = await request.json();
+    } else {
+      data = {
+        inputs: request.body,
+        options: extractQueryParams(url.searchParams),
```
anonrig: You don't need this function

```suggestion
        options: Object.fromEntries(url.searchParams),
```

src/cloudflare/internal/ai-api.ts
thatsKevinJain: I wasn't able to import the function defined in images-api to use in ai-api, but I modified the existing implementation to match the one in images-api

src/cloudflare/internal/ai-api.ts
thatsKevinJain: Updated

src/cloudflare/internal/ai-api.ts
thatsKevinJain: 'searchParams' is a read-only property so I was unable to assign anything to it directly, I slightly modified the implementation to use searchParams.set() instead.

src/cloudflare/internal/test/ai/ai-api-test.js
thatsKevinJain: Updated

src/cloudflare/internal/test/ai/ai-api-test.js
thatsKevinJain: Updated

src/cloudflare/internal/test/ai/ai-api-test.js
thatsKevinJain: TIL, it's much cleaner, updated!

src/cloudflare/internal/test/ai/ai-mock.js
thatsKevinJain: TIL as well, updated

src/cloudflare/internal/ai-api.ts
thatsKevinJain: Nice idea, updated!

src/cloudflare/internal/ai-api.ts
anonrig: What error did you see? Can you add a todo to merge these 2 functions?

thatsKevinJain: IIRC it was unable to import from images-api, it said module not found, even though LSP was able to confirm that the method is being imported from the right file

thatsKevinJain: Added a todo on the file, will try to debug more in future, thanks!

anonrig (approved):

PR #4296, opened 2025-06-10
fix bench-fast-api.c++ benchmark

anonrig:
Fixes benchmark.


npaun:
Why remove this test?

anonrig:
> Why remove this test?

Because `fastAddWithLock` doesn't exist anymore and it crashes when run

PR #4301, opened 2025-06-10
Pass a jsg::Lock to wrap and unwrap so we can get the isolate from it…

erikcorry:
… (#4244)

* Pass a jsg::Lock to wrap and unwrap so we can get the isolate from it

In #4208 we reduced the use of context->GetCurrent since is is only a wrapper around v8::Isolate::GetCurrent.

However, discussion there raised the issue that v8::Isolate::GetCurrent is potentially slow and messy since it gets the isolate from thread-local storage. It was agreed that it would be better to pass the jsg::Lock (which has the isolate) to those functions that need it. This change implements that suggestion.

### DIFF
--- a/src/workerd/io/promise-wrapper.h
+++ b/src/workerd/io/promise-wrapper.h
@@ -30,23 +30,23 @@ class PromiseWrapper {
   // For some reason, this wasn't needed for jsg::V8Ref<T>...
 
   template <typename T>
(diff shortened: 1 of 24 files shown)


erikcorry:
Reland of https://github.com/cloudflare/workerd/pull/4244

ketanhwr (approved):

anonrig (approved):

PR #4302, opened 2025-06-10
Implement fs.promises.FileHandle details and tests

jasnell:


src/workerd/api/filesystem.c++
```diff
@@ -1227,15 +1227,41 @@ kj::Array<FileSystemModule::DirEntHandle> FileSystemModule::readdir(
...
+  fdHandle = kj::none;
+}
+
+FileFdHandle::~FileFdHandle() noexcept {
+  if (fdHandle != kj::none) {
+    // We can safely close the file descriptor without an explicit lock
+    // because we are in the destructor of a jsg::Object which should
+    // only be destroyed when the isolate lock is held per the rules of
+    // the jsg::Ref<T> holder and the deferred destruction queue.
+    fdHandle = kj::none;
```
guybedford, line 1246: Does it make sense to have at least a debug assert for something like this?

src/workerd/io/worker-fs.c++
```diff
@@ -886,16 +905,36 @@ class VirtualFileSystemImpl final: public VirtualFileSystem {
...
 
+  kj::Own<void> wrapFd(jsg::Lock& js, int fd) const override {
+    return kj::heap<FdHandle>(weakThis.addRef(), fd);
+  }
+
  private:
+  // All operations on the VFS are expected to be performed while holding the
+  // isolate lock even tho the VFS itself does not depend directly on the
+  // lock. This is to ensure that the VFS operations are thread-safe without
+  // incurring additional locking overhead.
```
guybedford, line 916: Again is it possible / useful to add debug asserts for this?

src/workerd/io/worker-fs.h
```diff
@@ -586,6 +588,13 @@ class VirtualFileSystem {
   // API and posix in general.
   virtual void closeFd(jsg::Lock& js, int fd) const = 0;
 
+  // Returns an opaque RAII handle that wraps a file descriptor.
+  // When it is dropped, the file descriptor will be closed. The handle
+  // holds a weak reference to the VirtualFileSystem so that, on the off
+  // chance the VirtualFileSystem is destroyed while the handle is still
+  // alive, destruction of the handle will become a no-op.
+  virtual kj::Own<void> wrapFd(jsg::Lock& js, int fd) const KJ_WARN_UNUSED_RESULT = 0;
```
guybedford, line 596: Is the `Own<void>` pattern idiomatic KJ for opaque pointers? What is the benefit of making this opaque?

guybedford (approved):

src/workerd/api/filesystem.c++
jasnell: To ensure that it's within the isolate lock? We have other tests in various places that verify that jsg::Object destructors are run within the isolate lock so I think that would likely be redundant.

src/workerd/io/worker-fs.h
jasnell: Yeah we use this pattern in a number of places and I would consider it idiomatic for kj. The benefit is really just reducing the external API surface area when it's not needed. In this case I want it to be clear that this is purely an RAII handle. The only thing you can do with it is hold it then drop it.

PR #4304, opened 2025-06-10
Fix deprecation warning in async-context.

danlapid:
GetContinuationPreservedEmbedderData and SetContinuationPreservedEmbedderData 
Were replaced by v8 versions that generalized the solution for any v8::Data type.

Since we are only placing v8::Value the conversion `.As<v8::Value>` is safe.


danlapid:
@anonrig whenever you merge 13.8, or maybe just inline this into your branch

anonrig:
I cherry-picked your commit on https://github.com/cloudflare/workerd/pull/4223

PR #4309, opened 2025-06-11
CLAUDE.md

mikea:
result of /init + some manual prompts.


CLAUDE.md
```diff
@@ -0,0 +1,111 @@
...
+**workerd** is Cloudflare's JavaScript/WebAssembly server runtime that powers Cloudflare Workers. It's an open-source implementation of the same technology used in production at Cloudflare, designed for self-hosting applications, local development, and programmable HTTP proxy functionality.
+
+## Build System & Commands
+
+### Primary Build System: Bazel
+- Main build command: `bazel build //src/workerd/server:workerd`
+- Binary output: `bazel-bin/src/workerd/server/workerd`
+- Use `bazel sync --configure` or `bazel clean --expunge` to fix toolchain issues after installing dependencies
+
+### Just Commands (recommended for development)
```
kentonv, line 22: Really? I've never used or heard of any of these. Did I miss it?

CLAUDE.md
```diff
@@ -0,0 +1,111 @@
...
+- Run tests with `just test` before submitting PRs
+
+### Rust Development
+- `just update-rust <package>` - Update Rust dependencies (equivalent to `cargo update`)
+- `just clippy <package>` - Run clippy linting on Rust code
+- `just _rust-analyzer` - Generate rust-project.json for rust-analyzer
+
+## Package Management
+
+- Uses **pnpm** (version 10.10.0) for TypeScript/JavaScript dependencies
```
anonrig:
```suggestion
- Uses **pnpm** for TypeScript/JavaScript dependencies
```

CLAUDE.md
```diff
@@ -0,0 +1,111 @@
...
+- Root package.json contains development dependencies
+- Platform-specific npm packages in `npm/` directory
```
anonrig:
```suggestion
```

CLAUDE.md
```diff
@@ -0,0 +1,111 @@
...
+## Development Workflow
+
+### Code Style & Formatting
+- Automatic formatting via clang-format (enforced in CI)
+- Run `just format` before committing
+- Uses KJ C++ style guide (see Cap'n Proto project)
+
+### Language Server Support
+- Run `just compile-commands` to generate `compile_commands.json` for clangd
+- VSCode integration documented in `docs/vscode.md`
```
anonrig: Zed?

CLAUDE.md
```diff
@@ -0,0 +1,111 @@
...
+## Project Overview
+
+**workerd** is Cloudflare's JavaScript/WebAssembly server runtime that powers Cloudflare Workers. It's an open-source implementation of the same technology used in production at Cloudflare, designed for self-hosting applications, local development, and programmable HTTP proxy functionality.
+
+## Build System & Commands
+
+### Primary Build System: Bazel
+- Main build command: `bazel build //src/workerd/server:workerd`
+- Binary output: `bazel-bin/src/workerd/server/workerd`
+- Use `bazel sync --configure` or `bazel clean --expunge` to fix toolchain issues after installing dependencies
```
danlapid: I wouldn't actually suggest it to use `bazel clean --expunge` I feel like it would do it too often.

kentonv:
I'm not sure if using `/init` is the best way to initialize this file. We need to think about what things we always want to initialize the context with. Most of what's in this file now doesn't seem like the most important stuff.

CLAUDE.md
danlapid: Let's not over-document Zed, we really don't want to officially support more than one editor.
Regardless, I don't think the IDE is relevant to claude anyway.

CLAUDE.md
mikea: yes you did :)

kentonv:
FWIW I wrote the claude.md on the internal codebase and there I just thought about everything I'd tell to a new engineer on the team. A lot of text there is really about workerd and could be copied out here.

But I'm not necessarily sure if that was the right approach. It doesn't seem like the usual way to use claude.md. But intuitively it seems right to me -- initialize the context with general high-level knowledge that you always need.

mikea:
> Most of what's in this file now doesn't seem like the most important stuff.

Depends on what do you think the most important stuff is? Currently it is basically set of commands to run and code layout. I trimmed some extra bits.

mikea:
> FWIW I wrote the claude.md on the internal codebase and there I just thought about everything I'd tell to a new engineer on the team. A lot of text there is really about workerd and could be copied out here.
> 
> But I'm not necessarily sure if that was the right approach. It doesn't seem like the usual way to use claude.md. But intuitively it seems right to me -- initialize the context with general high-level knowledge that you always need.

Firstly I have no idea what I am doing: my first attempt to use the tool :) But my plan was to start building on this and add additional architecture information after. This was not supposed to be "the final" version of CLAUDE.md but just something to start working with.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Architecture
+
+- **Cap'n Proto source code** available in `external/capnp-cpp` - contains KJ C++ base library and
+capnproto RPC library. Consult it for all questions about `kj/` and `capnproto/` includes and
+`kj::` and `capnp::` namespaces.
+
+### Core Directory Structure (`src/workerd/`)
+- **`api/`** - Runtime APIs (HTTP, crypto, streams, WebSocket, etc.)
+  - Contains both C++ implementations and TypeScript definitions
```
jasnell: hmm... I think it would be more accurate to say that the TypeScript definitions are derived from the C++ implementation. The actual TypeScript definitions don't live in `api/*` with exception to the overrides.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+- **`src/pyodide/`** - Python runtime support via Pyodide
+- **`src/rust/`** - Rust integration components
+
+- TypeScript definitions generated in `types/` directory
+
+### Configuration System
+- Uses **Cap'n Proto** for configuration files (`.capnp` format)
+- Main schema: `src/workerd/server/workerd.capnp`
+- Sample configurations in `samples/` directory
+- Configuration uses capability-based security model
```
jasnell, line 78: Should we specifically make mention of compatibility flags and autogates. It would be interesting to see if claude could automatically make recommendations about when either may be needed based on whether it detects a possible breaking change.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Development Workflow
+
+### Code Style & Formatting
+- Automatic formatting via clang-format (enforced in CI)
+- Run `just format` before committing
+- Uses KJ C++ style guide (see Cap'n Proto project)
```
jasnell: This is not entirely true. We *mostly* follow the KJ style guide but with exceptions... for isntance, the more idiomatic comment placement. My fear is that this will lead to claude making recommendation that go against this style.

CLAUDE.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Contributing
+- Strong backwards compatibility commitment - features cannot be removed once deployed
```
jasnell: Not sure if this is nuanced enough. https://github.com/cloudflare/workerd/pull/4309/files#r2140805070

CLAUDE.md
```diff
@@ -0,0 +1,99 @@
...
+## Architecture
+
+- **Cap'n Proto source code** available in `external/capnp-cpp` - contains KJ C++ base library and
+capnproto RPC library. Consult it for all questions about `kj/` and `capnproto/` includes and
+`kj::` and `capnp::` namespaces.
+
+### Core Directory Structure (`src/workerd/`)
+- **`api/`** - Runtime APIs (HTTP, crypto, streams, WebSocket, etc.)
+  - Contains both C++ implementations
+  - Node.js compatibility layer in `api/node/`
```
jasnell:
```suggestion
  - C++ portions of the Node.js compatibility layer are in `api/node/`, while the JavaScript and TypeScript implementations live in `src/node/`
```

anonrig:
@danlapid CI is not run because we're only changing a markdown file. CI jobs are ignored, and this is an existing bug. This can only be landed by a person who has access to the repo as an admin. (cc @jasnell)

jasnell:
The only thing I'm apparently allowed to do even with admin rights is enable auto merge. It's not giving me the option to override.

mikea:
> @danlapid CI is not run because we're only changing a markdown file. CI jobs are ignored, and this is an existing bug. This can only be landed by a person who has access to the repo as an admin. (cc @jasnell)

heh. let me touch some source :)

PR #4311, opened 2025-06-11
Explicitly mark glob as unsupported

jasnell:
The implementation of `node:fs` glob functionality will be provided in the future. For now, it is explicitly marked as unsupported.

### DIFF
--- a/src/node/internal/internal_fs_callback.ts
+++ b/src/node/internal/internal_fs_callback.ts
@@ -84,6 +84,9 @@ import {
 import type {
   BigIntStatsFs,
   CopySyncOptions,
+  GlobOptions,
+  GlobOptionsWithFileTypes,
+  GlobOptionsWithoutFileTypes,
   MakeDirectoryOptions,
   OpenDirOptions,
   ReadAsyncOptions,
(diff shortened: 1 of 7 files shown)


src/node/internal/internal_fs_callback.ts
```diff
@@ -1247,6 +1250,21 @@ export function createWriteStream(): void {
...
+    | GlobOptions
+    | GlobOptionsWithFileTypes
+    | GlobOptionsWithoutFileTypes,
+  _callback: ErrorOnlyCallback
+): void {
+  // We do not yet implement the globSync function. In Node.js, this
+  // function depends heavily on the third party minimatch library
+  // which is not yet available in the workers runtime. This will be
+  // explored for implementation separately in the future.
+  throw new ERR_UNSUPPORTED_OPERATION();
```
anonrig, line 1265: Why not throw `ERR_METHOD_NOT_IMPLEMENTED`?

jasnell: Because I wanted to differentiate between a method that has not been implemented yet vs. one that we are explicitly not supporting yet.

guybedford (approved):

PR #4312, opened 2025-06-11
Python: Add baselineSnapshot to pyodide_extra

hoodmane:
So that the name of the baseline snapshot is available directly to C++.


dom96:
Why do we need the name in the C++?

hoodmane:
For edgeworker tests.

hoodmane:
In any case it's a relatively small amount of extra data to embed.

dom96:
True, but I just had in my head the idea that we might be able to get rid of pyodide_extra someday. Maybe not.

dom96 (approved):

PR #4314, opened 2025-06-11
Python: Add a script to automate snapshot generation

hoodmane:


dom96 (approved):
Nice! The only thing that would make this better is if we could test it in CI to ensure it doesn't break. But I won't block you on that :)

PR #4317, opened 2025-06-12
[nfc] always run ci jobs

mikea:
These jobs are required and if we don't run them we can't land docs-only changes.

Undos https://github.com/cloudflare/workerd/pull/3552


npaun:
Not a blocker but we can put an `if: ` statement on individual steps to avoid running them if code wasn't changed.

fhanau:
Thanks for landing this, I suggested the initial change but it clearly didn't work as intended.

PR #916, opened 2023-07-25
Revert "fix v8 deprecation warning (#909)"

jasnell:
This reverts commit ad65da8e4d7a1bbf5b800de35d13cc5a7eef73f0.

For some reason I have not yet determined, this change appears to be persistently breaking a couple of internal tests. Let's back it out for now while we investigate further.

PR #714, opened 2023-05-30
Fix deprecation warning for v::BasicTracedReference::operator->()

ohodson:
Replace uses of operator->() with Get() per the deprecation guidance.

Test: bazel build //...

PR #4328, opened 2025-06-13
Use Get() instead of deprecated arrow operator.

erikcorry:
In preparation for V8 13.9 where the operator is
gone.


kentonv:
Is this the same change we've tried to make several times before and it caused crashes in production every time but no one ever debugged it?

anonrig:
> Is this the same change we've tried to make several times before and it caused crashes in production every time but no one ever debugged it?

Yes. I've tried it twice at least

kentonv:
First attempt: #714
Reverted: #716
Second attempt: #909
Reverted: #916

fhanau:
I wasn't sure but yes it is. Same as #714, currently we patch in the old version in `patches/v8/0011-Revert-TracedReference-deref-API-removal.patch`.

kentonv:
Both reverts only say it caused trouble in tests, though.

I thought there was a time it got to prod and caused crashes but maybe I'm misremembering.

anonrig:
Fifth attempt: https://github.com/cloudflare/workerd/pull/4122

kentonv:
Probably the problem is that `handle.Get(isolate)` creates a new local handle, but this callback is called during garbage collection, which is a bad time for creating new handles. We need some other way to get an the internal pointers, without any side effects.

erikcorry:
It's sort of dumb to make handles during a GC, since they are roots, and the number of roots should not change during GC, but the new API requires it.  However I think it's enough to just create a HandleScope so that the newly created handles are quickly destroyed again before GC continues.  Let's see if that fixes it (or was that also already tried)?

I'm also trying to upstream https://chromium-review.googlesource.com/c/v8/v8/+/6652167 to make this footgun less subtle.

erikcorry:
I checked the linked attempts here, and none of them added the HandleScope.

PR #4330, opened 2025-06-13
[o11y] Fix disabling tracing for test() event

fhanau:
So far, we only disabled this for tests with entrypoints named "test", which is not the case for all tests.

This will help us support span tracing in production. Two traces that were produced indirectly using a test event (test() -> tail() and test() -> tail() -> tailStream()) have been dropped, for the regular streaming tail worker we were already not producing traces from tests since test() does not produce an onset event.


fhanau:
Ended up taking the opposite approach in #4298: Only test() events create spans in the tail-worker tests we currently have, so to avoid writing new tests I'm simply adding support for tracing test(). Based on already disabling tracing for "test" entrypoints, we have less bloat test output bloat than expected.

PR #4331, opened 2025-06-13
destroy container on actor shutdown

anonrig:
Let's destroy the container on actor shutdown.


src/workerd/server/container-client.h
```diff
@@ -61,8 +65,11 @@ class ContainerClient final: public rpc::Container::Server {
   };
 
   // Docker API v1.50 helper methods
-  kj::Promise<Response> dockerApiRequest(
-      kj::HttpMethod method, kj::StringPtr endpoint, kj::Maybe<kj::StringPtr> body = kj::none);
+  static kj::Promise<Response> dockerApiRequest(kj::Network& network,
+      kj::String dockerPath,
+      kj::HttpMethod method,
+      kj::String endpoint,
+      kj::Maybe<kj::StringPtr> body = kj::none);
```
danlapid:
```suggestion
      kj::Maybe<kj::String> body = kj::none);
```

danlapid (approved):

src/workerd/server/server.c++
```diff
@@ -2132,6 +2134,7 @@ class Server::WorkerService final: public Service,
       kj::Maybe<kj::Promise<void>> shutdownTask;
       kj::Maybe<kj::Promise<void>> onBrokenTask;
       kj::Maybe<kj::Exception> brokenReason;
+      kj::TaskSet& waitUntilTasks;
```
danlapid: Don't we already have some server task set? I'm pretty sure we do

danlapid: Yeah, use the WorkerService waitUntilTasks

anonrig: Yeah, you're right. Nice catch.

danlapid (approved):

anonrig:
Rebased and force pushed

src/workerd/server/container-client.c++
```diff
@@ -106,13 +106,21 @@ ContainerClient::ContainerClient(capnp::ByteStreamFactory& byteStreamFactory,
...
       timer(timer),
       network(network),
       dockerPath(kj::mv(dockerPath)),
       containerName(kj::encodeUriComponent(kj::mv(containerName))),
-      imageName(kj::mv(imageName)) {}
+      imageName(kj::mv(imageName)),
+      waitUntilTasks(waitUntilTasks) {}
+
+ContainerClient::~ContainerClient() noexcept(false) {
+  waitUntilTasks.add(dockerApiRequest(network, kj::str(dockerPath), kj::HttpMethod::DELETE,
```
kentonv, line 120: It looks like this `dockerApiRequest()` will keep accessing members of `ContainerClient` after it has been destroyed. I think it needs to be refactored into a static function that can take ownership of the things it needs to complete itself.

anonrig: Right now, it only accesses kj::network. And we can't transfer ownership since, worker service is the owner, and will live longer than ContainerClient. Any suggestions?

danlapid: It was refactored to a static function though.

kentonv: Ugh sorry, I think I didn't look hard enough before commenting.

kentonv: Specifically, I looked at this change, and I somehow didn't look at all of the diffs below it, but instead looked over at the old version of the file in my editor to see that `dockerApiRequest()` was a non-static method.

Issue #4338, opened 2025-06-16
nodejs compat `process` regression


petebacondarwin:
https://github.com/cloudflare/workerd/pull/4271 has caused a [regression with the Cloudflare Vite plugin](https://github.com/cloudflare/workers-sdk/actions/runs/15656283865/job/44107762797?pr=9591) (and maybe even with currently deployed nodejs applications?).

## Reproduction

To reproduce this you can checkout the workerd bump PR branch and run the vite playground:

```
git clone https://github.com/cloudflare/workers-sdk
cd workers-sdk
git switch dependabot/npm_and_yarn/workerd-and-workers-types-5ef7e5cec4
pnpm i
pnpm build -F @cloudflare/vite-plugin
cd packages/vite-plugin-cloudflare/playground/node-compat
pnpm run basic:dev
```

If you want to debug the Worker code itself then modify the Vite config: `packages/vite-plugin-cloudflare/playground/node-compat/vite.config.worker-basic.ts`, removing the following line:

```
inspectorPort: false,
```

Then when you run `pnpm run basic:dev` you can open the browser at `http://localhost:5173/__debug`, which will open up a dev tools console attached to the worker.

## Analysis

In that PR we have the [following code](https://github.com/cloudflare/workerd/pull/4271/files#diff-88108dcbc39ce5c312cac49ec7a85ea4a0fa0d831318a83caa878546b0639e4aR442-R445):

```ts
// TODO(soon): Support via experimental FS
export function cwd(): string {
  throw new ERR_UNSUPPORTED_OPERATION();
}
```

But in the `pathe` npm library, [there is this code](https://github.com/unjs/pathe/blob/main/src/_path.ts#L91C1-L96C2):

```ts
function cwd() {
  if (typeof process !== "undefined" && typeof process.cwd === "function") {
    return process.cwd().replace(/\\/g, "/");
  }
  return "/";
}
```

Which is used by the Vite module-runner: https://github.com/vitejs/vite/blob/main/packages/vite/src/module-runner/utils.ts#L39.

Previously the `process.cwd` method did not exist so the pathe function simply returned `/`. Now it tries to call it and we get  the `ERR_UNSUPPORTED_OPERATION` error.

Here is an example callstack:

```
Uncaught (in response) Error: The requested operation is unsupported
    at EventEmitter.cwd (node-internal:internal_process:319:11)
    at cwd (index.js:78:77)
    at resolve (index.js:84:51)
[...]

PR #4345, opened 2025-06-16
Remove polyfill-breaking throwing stubs for cwd(), chdir(), umask()

guybedford:
The `process.cwd()` throwing stub caused a break when this was being polyfilled.

This removes the throwing stub for all of `cwd`, `chdir` and `umask`, until these are correctly implemented.

In the mean time tests are added for undefined implementations as distinct from throwing stubs.

Resolves https://github.com/cloudflare/workerd/issues/4338.


petebacondarwin (approved):

guybedford:
Will re-land with the new gated version.

PR #4346, opened 2025-06-16
Special handling for WorkflowEntrypoint

Caio-Nogueira:
This PR adds `WorkflowEntrypoint` as a special class to be executed. It also wraps the existing workflows (`WorkflowStep`) API to be used in a more Python friendly way. With these changes, the Python environment should be able to declare a workflow and have it work just like in Javascript (same context under the hood)

- wraps the `stubStep` on the entrypoint
- tests RPC calling between js and Python, just like it happens for the workflows' control plane when calling into a user worker.


src/workerd/server/tests/python/workflow-entrypoint/workflow.py
```diff
@@ -0,0 +1,25 @@
...
+
+from workers import WorkflowEntrypoint
+from js import console
+
+class WorkflowEntrypointExample(WorkflowEntrypoint):
+    def __init__(self, state, env):
+        self.state = state
+        self.counter = 0
+
+    async def on_run(self, event, step):
```
dom96, line 13: So based on the code, am I understanding correctly that this can also be defined as:

```
async def on_run(self, step)
```

Looking at the JS it doesn't seem that is supported there. Why are you adding this support here?

src/pyodide/internal/_workers.py
```diff
@@ -942,6 +984,39 @@ def wrapped_init(self, *args, **kwargs):
...
+    )
+
+    if not is_workflow_subclass:
+        # Not a workflow subclass, so don't wrap `on_run`.
+        return
+
+    if asyncio.iscoroutinefunction(run_fn):
+        async def wrapped_run(self, *args, **kwargs):
+            if len(args) > 1:
+                args = (python_from_rpc(args[0]), _WorkflowStepWrapper(args[1])) + args[2:]
```
dom96: You're calling python_from_rpc on `self` here (args[0]), right? I don't think you should be doing that.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -258,7 +259,10 @@ function makeEntrypointProxyHandler(
       const isKnownDoHandler =
         isDurableObject && SPECIAL_DO_HANDLER_NAMES.includes(prop);
       const isFetch = prop == 'fetch';
-      if (isKnownHandler || isKnownDoHandler) {
+      // TODO: find a better workaround for this. We can't keep considering any worker that implements a run method
+      // as a workflow. What consequences do we have here?
+      const isWorkflowHandler = isWorkflow || prop == 'run';
+      if (isKnownHandler || isKnownDoHandler || isWorkflowHandler) {
```
dom96: Why isn't it enough to check for `isWorkflow`?

src/pyodide/internal/_workers.py
```diff
@@ -942,6 +984,39 @@ def wrapped_init(self, *args, **kwargs):
 
     cls.__init__ = wrapped_init
 
+    run_fn = getattr(cls, "on_run", None)
```
dom96, line 991: Can you move the code you added here to a separate function and call it from here?

src/pyodide/python-entrypoint-helper.ts
Caio-Nogueira: `isWorkflow` is only true when the class name is `WorkflowEntrypoint`. When we rely on metadata from the runtime, we don't distinguish between `WorkerEntrypoint` and `WorkflowEntrypoint`. The workaround that I found was considering the `run` method as well, although I could see it being problematic. Lmk if you have another approach that might apply here.

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
Caio-Nogueira: Yeah that was wrong 😓 . `on_run` should take `event` and `step`. I updated the arguments handling to reflect this

src/pyodide/internal/_workers.py
```diff
@@ -936,44 +937,49 @@ class _WorkflowStepWrapper:
...
-
-                return result
-            return wrapper
-
-        return decorator
+    def do(self, name, callback=None, config=None):
+        if callback is None:
+            # Return a decorator that will execute the function when called
+            def decorator(func):
+                async def wrapper(*args, **kwargs):
```
hoodmane:
```suggestion
                @functools.wraps(callback)
                async def wrapper(*args, **kwargs):
```

src/pyodide/internal/_workers.py
Caio-Nogueira: `args[0]` seems to be `event` here. I think `self` should be captured by the first positional argument (?)

src/pyodide/internal/_workers.py
```diff
@@ -936,44 +937,49 @@ class _WorkflowStepWrapper:
...
+            return result
+
+        if config is None:
+            result = self._js_step.do(name, _callback)
+        else:
+            result = self._js_step.do(name, to_js(config), _callback)
 
-    # JS <-> PY already converts casing
-    # No further work needed for sleep/sleep_until since APIs are identical
+        return python_from_rpc(result)
```
hoodmane: Let's factor this out into a separate private `_do_call()` top level function. Better not to do a method in case people define their own method with the same name on a subclass.

src/pyodide/python-entrypoint-helper.ts
hoodmane: I think it would be okay to consider anything with an `on_run` method to be a workflow. But we should not break existing `run` methods. If the name is `run` but it doesn't inherit `WorkflowEntrypoint`, first check for a `run` method and if it exists call that normally. If no such exists, check for an `on_run` method and do the workflow stuff.

Or just only allow `WorkflowEntrypoint` subclasses.

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -292,6 +300,7 @@ function makeEntrypointClass(
   methods: string[]
 ): any {
   const isDurableObject = classKind.name === 'DurableObject';
+  const isWorkflow = classKind.name === 'WorkflowEntrypoint';
```
hoodmane: Maybe we should pass `classKind.name` down into `makeEntrypointProxyHandler()` and perform these comparisons in there.

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
```diff
@@ -0,0 +1,32 @@
...
+
+        await workflow_step()
+
+        async def step_with_callback():
+            self.counter += 1
+            console.log(f"hello from python {self.counter}")
+
+            return event["foo"]
+
+        return step.do("my_second_step", step_with_callback)
```
hoodmane: Does the return value mean anything? Maybe:
```suggestion
        await step.do("my_second_step", step_with_callback)
```

src/pyodide/internal/_workers.py
```diff
@@ -942,6 +990,36 @@ def wrapped_init(self, *args, **kwargs):
...
+    run_fn = getattr(cls, "on_run", None)
+    if run_fn is None:
+        return
+
+    # Only patch `on_run` for subclasses of WorkflowEntrypoint.
+    is_workflow_subclass = any(
+        base.__name__ == "WorkflowEntrypoint" for base in cls.__mro__[1:]
+    )
+
+    if not is_workflow_subclass:
```
hoodmane:
```suggestion
    if not issubclass(cls, WorkflowEntrypoint):
```

src/pyodide/internal/_workers.py
```diff
@@ -942,6 +990,36 @@ def wrapped_init(self, *args, **kwargs):
...
+        # Not a workflow subclass, so don't wrap `on_run`.
+        return
+
+    if asyncio.iscoroutinefunction(run_fn):
+        async def wrapped_run(self, *args, **kwargs):
+            if len(args) > 1:
+                args = (python_from_rpc(args[0]), _WorkflowStepWrapper(args[1])) + args[2:]
```
hoodmane:
```suggestion
        async def wrapped_run(self, event = None, step = None, /, *args, **kwargs):
            if event is not None:
                event = python_from_rpc(args[0]) 
            if step is not None:
                step = _WorkflowStepWrapper(step)
            return await run_fn(self, event, step, *args, **kwargs)
```

src/pyodide/internal/_workers.py
```diff
@@ -942,6 +990,36 @@ def wrapped_init(self, *args, **kwargs):
...
+
+            return await run_fn(self, *args, **kwargs)
```
hoodmane: To reduce duplication, maybe we can always define an async wrapper and do the following:
```suggestion
            result = run_fn(self, *args, **kwargs)
            if iscoroutine(result):
                result = await result
            return result
```

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
```diff
@@ -0,0 +1,32 @@
...
+class WorkflowEntrypointExample(WorkflowEntrypoint):
+    def __init__(self, state, env):
+        self.state = state
+        self.counter = 0
+
+    async def on_run(self, event, step):
+        @step.do("my_first_step")
+        async def workflow_step():
+            self.counter += 1
+            console.log(f"hello from python {self.counter}")
```
hoodmane: Why not:
```suggestion
            print(f"hello from python {self.counter}")
```

src/pyodide/python-entrypoint-helper.ts
Caio-Nogueira: Yeah this is a tricky one. I don't really see a way that doesn't introduce (at least the possibility) of breaking behavior. I agree with this approach though.

> Or just only allow WorkflowEntrypoint subclasses.

Unfortunately I think this doesn't work for all cases at the moment

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
Caio-Nogueira: Heh I could just remove that. Not the point of the test anyway

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
Caio-Nogueira: We need a return here because that's what the JS worker is expecting. We do something similar on workflows: users can return from inside `step.do`, and have the result be stored by the platform.

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,47 @@ def wrapper(*args, **kwargs):
...
+    def __init__(self, js_step):
+        self._js_step = js_step
+
+    def do(self, name, callback=None, config=None):
+        if callback is None:
+            # Return a decorator that will execute the function when called
+            def decorator(func):
+                @functools.wraps(func)
+                async def wrapper(*args, **kwargs):
+                    return self._do_call(name, config, func)
```
hoodmane: Let's move `_do_call` to a top level function instead of a method and call it like:
```suggestion
                    return _do_call(self, name, config, func)
```

src/pyodide/internal/_workers.py
```diff
@@ -943,6 +985,31 @@ def wrapped_init(self, *args, **kwargs):
...
+    run_fn = getattr(cls, "on_run", None)
+    if run_fn is None:
+        return
+
+    # Only patch `on_run` for subclasses of WorkflowEntrypoint.
+    if not issubclass(cls, WorkflowEntrypoint):
+        # Not a workflow subclass, so don't wrap `on_run`.
+        return
+
+    async def wrapped_run(self, event=None, step=None, /, *args, **kwargs):
```
hoodmane, line 1001:
```suggestion
    @functools.wraps(run_fn)
    async def wrapped_run(self, event=None, step=None, /, *args, **kwargs):
```

src/pyodide/internal/_workers.py
```diff
@@ -943,6 +985,31 @@ def wrapped_init(self, *args, **kwargs):
...
+        if event is not None:
+            event = python_from_rpc(event)
+        if step is not None:
+            step = _WorkflowStepWrapper(step)
+
+        result = run_fn(self, event, step, *args, **kwargs)
+
+        if inspect.iscoroutine(result):
+            return python_from_rpc(await result)
```
hoodmane:
```suggestion
            result = await result
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -244,40 +244,68 @@ const SPECIAL_DO_HANDLER_NAMES = [
 
 function makeEntrypointProxyHandler(
   pyInstancePromise: Promise<PyModule>,
-  isDurableObject: boolean
+  className: string
 ): ProxyHandler<any> {
+  const compatibilityFlags: any =
+    (globalThis as any)?.Cloudflare?.compatibilityFlags ?? {};
+  const workflowsEnabled: boolean = !!compatibilityFlags.python_workflows;
```
hoodmane: Interesting... Can we add a typescript type for this?

Caio-Nogueira: Sure. I'm also not entirely sure if this is the best way to verify the compat flag 😄 .

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -242,42 +242,81 @@ const SPECIAL_DO_HANDLER_NAMES = [
...
+  python_workflows?: boolean;
+  [key: string]: boolean | undefined;
+}
+
+interface CloudflareGlobal {
+  Cloudflare?: {
+    compatibilityFlags?: CompatibilityFlags;
+  };
+}
+
```
dom96: Move this and the code reading the workflowsEnabled field below into metadata.ts

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -242,42 +242,81 @@ const SPECIAL_DO_HANDLER_NAMES = [
...
+        if (
+          targetProp == 'run' &&
+          typeof pyInstance['on_run'] === 'function' &&
+          typeof pyInstance['run'] !== 'function'
+        ) {
+          // unfortunately we can't solely rely on the subclass
+          // we're also considering everything that implements an `on_run` method to be a workflow
+          // and handling it appropriately on the `run` prop
+          // Note(caio): if a normal run method exists, we handle it normally to avoid breaking changes
+          // (this is not ideal)
```
dom96: We'll be supporting `fetch` as well as `on_fetch` soon, I wonder if we should only support `run` here instead. Doing this would also mean you can simplify this code.

CC @danlapid

Caio-Nogueira: It would definitely make the logic simpler. I'm concerned though with breaking `WorkerEntrypoint` with a `run` method defined.

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
```diff
@@ -0,0 +1,58 @@
...
+            await workflow_step()
+        except MyCustomException as e:
+            print(f"successfully caught {e} with decorator")
+            pass
+
+        @step.do("my_third_step")
+        async def workflow_step_success():
+            return event["foo"]
+
+        try:
```
Talador12, line 50: It's good to have all of these custom exceptions, but we may want to put this try catch inside of the step itself. That would make this file much easier to read

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,49 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+class _WorkflowStepWrapper:
+    def __init__(self, js_step):
+        self._js_step = js_step
```
sidharthachatterjee, line 936: Not necessarily blocking but let's figure out before we make available to external folks: Can userland code access _js_step? We shouldn't leak internals

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,49 @@ def wrapper(*args, **kwargs):
...
+        async def wrapper(*args, **kwargs):
+            return _do_call(self, name, config, callback)
+
+        return wrapper()
+
+
+def _do_call(entrypoint, name, config, callback):
+    async def _callback():
+        result = callback()
+        if inspect.isawaitable(result):
```
sidharthachatterjee: What's the difference between isawaitable and iscoroutine

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,49 @@ def wrapper(*args, **kwargs):
...
+            result = await result
+        return result
+
+    if config is None:
```
sidharthachatterjee, line 969: Sorry for making you need to do this 😅

src/pyodide/python-entrypoint-helper.ts
sidharthachatterjee: 🦆 typing 🎉

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -244,40 +245,64 @@ const SPECIAL_DO_HANDLER_NAMES = [
...
       return async function (...args: any[]): Promise<any> {
         // Check if the requested method exists and if so, call it.
         const pyInstance = await pyInstancePromise;
-        if (typeof pyInstance[prop] !== 'function') {
-          throw new TypeError(`Method ${prop} does not exist`);
+        let targetProp = prop;
+
+        if (
+          targetProp == 'run' &&
+          typeof pyInstance['on_run'] === 'function' &&
```
sidharthachatterjee: Correct me if I'm wrong: 

This is a breaking change for Durable Objects because a DO authored in Python with an `on_run` fn will be run as a Workflow?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -244,40 +245,64 @@ const SPECIAL_DO_HANDLER_NAMES = [
...
+          typeof pyInstance['run'] !== 'function'
```
sidharthachatterjee: Check `className === 'WorkerEntrypoint'`?

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) 2025 Cloudflare, Inc.
+# Licensed under the Apache 2.0 license found in the LICENSE file or at:
+#     https://opensource.org/licenses/Apache-2.0
+
+from workers import NonRetryableError, WorkflowEntrypoint
```
sidharthachatterjee, line 5: In JavaScript, we import `NonRetryableError` from `workflows`

Not the end of the world but consistency would be nice

sidharthachatterjee (commented):
Left a few comments, looks good broadly

Please confirm DO case, thanks!

src/pyodide/python-entrypoint-helper.ts
sidharthachatterjee: @dom96 Maybe you know the answer off the top of your head

Talador12:
For the exception example workflow:

This would be more "pythonic" and it's easier to see what each step is doing this way
```py
# Copyright (c) 2025 Cloudflare, Inc.
# Licensed under the Apache 2.0 license found in the LICENSE file or at:
#     https://opensource.org/licenses/Apache-2.0

from workers import NonRetryableError, WorkflowEntrypoint


class MyCustomException(Exception):
    pass


class WorkflowEntrypointExample(WorkflowEntrypoint):
    async def on_run(self, event, step):
        async def await_step(fn):
            try:
                return await fn()
            except Exception as e:
                print(f"Successfully caught {type(e).__name__}: {e}")

        @step.do("my_first_step")
        async def step_one():
            print("Doing my_first_step... About to raise TypeError")
            raise TypeError("Test error")

        await await_step(step_one)

        @step.do("my_second_step")
        async def step_two():
            print("Doing my_second_step... About to raise NonRetryableError")
            raise NonRetryableError("Test error")

        await await_step(step_two)

        @step.do("my_errored_step")
        async def step_three():
            print("Doing my_errored_step... About to raise MyCustomException")
            raise MyCustomException("Test error")

        await await_step(step_three)

        @step.do("my_third_step")
        async def step_success():
            print("Doing my_third_step... Should succeed")
            return event["foo"]

        return await await_step(step_success)


async def test(ctrl, env, ctx):
    pass
 ```

**Comment by Talador12** [comment]
Second idea - a happy path "DAG diamond" example

     step_1
       /     \
 step_2   step_3
       \     /
     step_4
 
 
 Something like this shows parallel steps, dependent steps, and accepting results from prior steps. May need to work out how the results are returned to other steps, but this is generally the idea:
```py
 # Copyright (c) 2025 Cloudflare, Inc.
# Licensed under the Apache 2.0 license found in the LICENSE file or at:
#     https://opensource.org/licenses/Apache-2.0

from workers import NonRetryableError, WorkflowEntrypoint


class MyCustomException(Exception):
    pass


class WorkflowEntrypointExample(WorkflowEntrypoint):
    async def on_run(self, event, step):
        async def await_step(fn):
            try:
                return await fn()
            except Exception as e:
                print(f"Successfully caught {type(e).__name__}: {e}")

        @step.do("step_1")
        async def step_1():
            print("Executing step 1")
            return {"foo": event.get("foo", "default_foo")}

        @step.do("step_2")
        async def step_2():
            print("Executing step 2 (depends on step 1)")
            result = await step_1()
            print(f"step_2 got: {result}")
            raise TypeError("Intentional error in step 2")

        @step.do("step_3")
        async def step_3():
            print("Executing step 3 (depends on step 1)")
            result = await step_1()
            print(f"step_3 got: {result}")
            return {"bar": result["foo"] + "_bar"}

        @step.do("step_4")
        async def step_4():
            print("Executing step 4 (depends on steps 2 and 3)")
            result2 = await step_2()
            result3 = await step_3()
            print(f"step_4 got: {result2}, {result3}")
            return {"combined": (result2, result3)}

        return await await_step(step_4)


async def test(ctrl, env, ctx):
    pass
```

src/pyodide/internal/_workers.py
Caio-Nogueira: `isawaitable` is broader than `iscoroutine`. You can await for coroutines and other things like Futures. In this case I think we can narrow down the check if that's what you're saying.

src/pyodide/python-entrypoint-helper.ts
Caio-Nogueira: I'll remove this for now. I'm assuming that traced requests will not work though. Since this is experimental I think we're fine

src/pyodide/internal/_workers.py
hoodmane: They can. But leaking internals is kind of the Python way.

src/pyodide/python-entrypoint-helper.ts
hoodmane: Durable object support hasn't been around very long and it's all beta so I think we're okay with small breaking changes like this.

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
Caio-Nogueira: I see what you're saying. Let's just not block on this for now and figure it out later as it is lower priority imo

src/pyodide/internal/_workers.py
Caio-Nogueira: The underscore is the convention to specify this intent. I think it should be sufficient. Specially because Python doesn't seem to offer any alternatives to truly make some field private 😓

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
Caio-Nogueira: I was testing try...catching the step itself to make sure that we can capture the exceptions that cross the language barrier are correctly captured by the runtime (after Pyodide translation). We allow try.. catching steps in JS Workflows https://developers.cloudflare.com/workflows/build/sleeping-and-retrying/

If a user raises and handles the exception inside the callback (i.e.: the method running inside `step.do`), then the interpreter should have no problems in handling that - no type translation or serialization is required.

src/pyodide/python-entrypoint-helper.ts
Caio-Nogueira: > Durable object support hasn't been around very long and it's all beta so I think we're okay with small breaking changes like this.

We don't need to break DOs here. If I leave this condition on (just to solve the tracing code path), then we're worried about `WorkerEntrypoint`, right? Assuming we check that `classkind.name` is `WorkerEntrypoint`.

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,55 @@ def wrapper(*args, **kwargs):
...
+        async def wrapper(*args, **kwargs):
+            return _do_call(self, name, config, callback)
+
+        return wrapper()
+
+    def sleep(self, *args, **kwargs):
+        return self._js_step.sleep(*args, **kwargs)
+
+    def sleep_until(self, *args, **kwargs):
+        return self._js_step.sleep_until(*args, **kwargs)
```
hoodmane: Is it spelled `sleep_until` in JavaScript?

src/pyodide/python-entrypoint-helper.ts
dom96: I'd definitely recommend checking for `WorkflowEntrypoint` and avoiding the on_run heuristic, even if it means tracing is broken.

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,55 @@ def wrapper(*args, **kwargs):
...
+        # we need to wrap the callback so that we can handle exceptions:
+        # JS promises should be awaited in a Python async context so that Pyodide
+        # can handle the promise resolution/rejection
+        # If the async context where we await these promises is further down the call stack
+        # pyodide won't map the JS exceptions to Python equivalents properly
+        @functools.wraps(callback)
+        async def wrapper(*args, **kwargs):
+            return _do_call(self, name, config, callback)
+
+        return wrapper()
```
hoodmane: Since we immediately call this we don't need `functools.wraps()` or the varargs:
```suggestion
        async def wrapper():
            return _do_call(self, name, config, callback)

        return wrapper()
```
Also, since this branch has less nesting, let's move it first to get our deepest nesting down to 4 from 5:
```py
        if callback:
            async def wrapper():
                return _do_call(self, name, config, callback)

            return wrapper()
```

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -244,40 +245,50 @@ const SPECIAL_DO_HANDLER_NAMES = [
...
+      const isDurableObject = className === 'DurableObject';
+      const isWorkflow = className === 'WorkflowEntrypoint';
 
       // Proxy calls to `fetch` to methods named `on_fetch` (and the same for other handlers.)
       const isKnownHandler = SPECIAL_HANDLER_NAMES.includes(prop);
       const isKnownDoHandler =
         isDurableObject && SPECIAL_DO_HANDLER_NAMES.includes(prop);
       const isFetch = prop == 'fetch';
-      if (isKnownHandler || isKnownDoHandler) {
+      const isWorkflowHandler = isWorkflow && prop == 'run';
```
hoodmane: We should be using `===` unless there is a specific reason to use `==`.

hoodmane (approved):

src/pyodide/python-entrypoint-helper.ts
Caio-Nogueira: You're right! Should I also update `prop === 'fetch'`?

hoodmane: Yeah please do.

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,50 @@ def wrapper(*args, **kwargs):
...
+
+            async def wrapper():
+                return _do_call(self, name, config, callback)
+
+            return wrapper()
+
+        # Return a decorator that will execute the function when called
+        def decorator(func):
+            @functools.wraps(func)
+            async def wrapper(*args, **kwargs):
```
hoodmane: We seem to be ignoring the arguments. What are the arguments we would be getting here? I think maybe this should also be:
```suggestion
            async def wrapper():
```
Or would that not work for some reason?

Caio-Nogueira: Yeah we can omit the args here

src/pyodide/internal/_workers.py
```diff
@@ -930,6 +931,50 @@ def wrapper(*args, **kwargs):
...
+        if callback:
+
+            async def wrapper():
+                return _do_call(self, name, config, callback)
+
+            return wrapper()
+
+        # Return a decorator that will execute the function when called
+        def decorator(func):
+            @functools.wraps(func)
```
hoodmane: We should get rid of `@functools.wraps()` now since it is for decorators that don't change the arguments:
```suggestion
```

PR #4347, opened 2025-06-16
Stop using kj::Thread in fetchPyodideBundle.

dom96:
We pass kj::AsyncIoContext around to fetchPyodideBundle. There are cases where an AsyncIoContext is unavailable, like server-test which uses an EventLoop. If there is a better approach, I'd love your thoughts.


kentonv (changes requested):
AsyncIoContext is not intended to be passed around like this.

AsyncIoContext exists as a convenience to set up common top-level stuff. Not all programs use it. The internal codebase, for example, does not use it (green threads set up their event loop differently). Passing it deep into the codebase like this would make it hard to change the top-level setup to work differently later.

Also, AsyncIoContext contains a `waitScope` which absolutely should not be made available to anything outside the top level.

Normally, you should instead pass around the specific components of the AsyncIoContext hat you need -- like UnixEventPort, or AsyncIoProvider, etc.

danlapid:
Agreed with Kenton and that's what I responded originally on the previous PR.
You only need network and timer, just pass those.

kentonv:
Indeed, `Network` and `Timer` are already being passed down, if that's all you need then just use those.

Another advantage here is that `Network` and `Timer` are abstract interfaces, so they can be mocked for test purposes. `server-test`, for example, mocks both! `AsyncIoContext` cannot be mocked.

Incidentally `server-test` does not use `AsyncIoContext`, as you probably noticed since you are passing `kj::none` there. It would be unfortunate if some parts of the server don't work in the case of a missing `AsyncIoContext`, and therefore cannot be tested in `server-test`.

src/workerd/server/workerd-api.c++
```diff
@@ -295,39 +295,33 @@ kj::Maybe<jsg::Bundle::Reader> fetchPyodideBundle(
...
 
-      auto req = client->request(kj::HttpMethod::GET, url.asPtr(), headers);
+  auto req = client->request(kj::HttpMethod::GET, url.asPtr(), headers);
 
-      auto res = req.response.wait(io.waitScope);
-      KJ_ASSERT(res.statusCode == 200,
-          kj::str(
-              "Request for Pyodide bundle at ", url, " failed with HTTP status ", res.statusCode));
-      auto body = res.body->readAllBytes().wait(io.waitScope);
+  auto res = req.response.wait(io.waitScope);
```
danlapid, line 317: Once you convert this function to a promise you won't need the waitScope, you could just await.

dom96:
In that case I need to make `fetchPyodideBundle` async, how do I run it without the `waitScope`? Do I need to make all its parents async too and bubble up until I am at the level where a `waitScope` is available? I'm also not sure how to handle cases like the `fetchPyodideBundle` call in `WorkerdApi::Impl` which is a constructor, can I run an async function via the jsg Lock somehow?

src/workerd/server/workerd-api.c++
kentonv: Oh wow, just looked closer at this. This whole hack needs to be refactored -- we shouldn't be blocking on I/O in a synchronous function, either using a thread or using `.wait()`.

I guess it happens to work today only because workerd constructs all `WorkerdApi` objects at startup anyway, so this only blocks startup. But that won't be true forever. I'm working on an API right now for dynamically loading isolates, which will need to construct new `WorkerdApi` instances at runtime.

I think that Python dependencies need to be fetched at a higher level, and then passed down into `WorkerdApi`.

danlapid:
> constructor

Yeah you would just make everything async, if you need the information in a constructor then call it beforehand and just pass the results into the constructor.

dom96:
Closing in favour of https://github.com/cloudflare/workerd/pull/4548

PR #4350, opened 2025-06-16
NodeJS Compat Flag for process versions

guybedford:
This adds a new compat flag for populating `process.versions.node`, just like we had the compat flag for `process.env` population.

This is because feature detection paths usually check `process.versions?.node` in Node.js checks, which with this flag will be properly guarded and enabled only within the Node.js compat feature.

Review would be very welcome I am understanding the flag application here correctly with regards to compat date, and Node.js compat flag interactions.


src/workerd/io/compatibility-date.capnp
```diff
@@ -814,7 +814,7 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
 
   bindAsyncLocalStorageSnapshot @93 :Bool
       $compatEnableFlag("bind_asynclocalstorage_snapshot_to_request")
-      $compatDisableFlag("do_not_bind_asynclocalstorage_snapshot_to-request")
+      $compatDisableFlag("do_not_bind_asynclocalstorage_snapshot_to_request")
```
anonrig: Can you revert this change?

src/workerd/io/compatibility-date.capnp
```diff
@@ -838,4 +838,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $experimental;
   # Enables support for process.platform to expose the actual system platform
   # This will never ever be supported and is a WPT test path only.
+
+  populateProcessVersions @96 :Bool
+      $compatEnableFlag("nodejs_compat_populate_process_versions")
+      $compatDisableFlag("nodejs_compat_do_not_populate_process_versions")
+      $impliedByAfterDate(name = "nodeJsCompat", date = "2025-06-16");
```
anonrig, line 845: Can you give at least 2 weeks of time for internal systems to propagate?

src/workerd/io/compatibility-date.capnp
guybedford: Done

src/workerd/api/node/process.c++
```diff
@@ -92,9 +92,12 @@ namespace {
 
 jsg::JsObject ProcessModule::getVersions(jsg::Lock& js) const {
   auto versions = js.obj();
-  // Node.js version - represents the most current Node.js version supported
-  // by the platform, as defined in node-version.h
-  versions.set(js, "node"_kj, js.str(nodeVersion));
+
+  if (FeatureFlags::get(js).getPopulateProcessVersions()) {
```
anonrig, line 96: This only populates the node.js version, but the compat name refers to populating all process.versions?

src/workerd/api/node/process.c++
danlapid, line 1: I'm worried about the severing approach.
I'm worried we'll end up reverting and re-merging the process PR many times.
Can we explore adding the compat flag on the entire process change somehow?

guybedford:
Yeah let's gate the entire feature in the next round rather.

Issue #2240, opened 2024-06-08
bug: `DurableObjectId::id.name` doesn't work


threepointone:
(maybe this is a feature request?)

When creating a durable object with `.idFromName(name)`, I would expect `this.ctx.id.name` _inside_ a durable object. to reflect the same. The types imply that it should it work, was it just missed? 

We use this name extensively, but currently have to hack it in by passing it in with an explicit method, but it's super cumbersome to do so without a bunch of hacks and making sure it's passed before any subsequent requests/methods are called. It would simplify a lot if workerd/workers/wrangler could provide it by default. Please and thank you!

threepointone:
The "hack"/workaround we have also doesn't work for alarms, so fixing this would fix that bug as well.

threepointone:
I was meaning to ask, why is that pattern even around? I kinda remember it being in the docs, but can't find it there anymore (and rewrote some of my non-released code to not have that intermediate step anymore). 

I think we can simply message the conditions under which that name will exist. 
- It won't exist if using newUniqueId()
- it won't exist if using the above pattern (or rather, it'll return the hex id instead)
- it's only `ns.get(ns.idForName(...))`. This might exclude any existing DOs that rewrite to this pattern, but that's a hit we should take 


> I guess it would get initialized the first time a name-bearing ID is used?)

Yes 

> his gets weird when you consider that DOs are implicitly deleted if they shut down while their storage is empty. So the name has to be forgotten at that point. But only if the storage is empty? Is that weird?

Oh hmm. Does that mean it won't be available in alarms? In which case, can we store the name everywhere the ID is stored? 
Or does setting an alarm mean that storage isn't empty? In which case, it's fine, since it'll get the name again when it comes back to life.

kentonv:
The next alarm time counts as a stored value, so when an alarm is present then storage is not empty.

There are other built-in features planned which are likely to store hex IDs and where we wouldn't really want those systems to have to think about storing names alongside them.

But perhaps we can be reasonably confident that future features similar to alarms that store a hex ID only are never going to be waking up an empty object? And therefore the only way an object might not know its name is if application code wakes up an empty object by ID? Maybe that's OK?

Another weird issue: Technically someone could be passing in a multi-megabyte string as a name, which works fine if it's just getting hashed down but not so fine if we have to transmit that name over the network implicitly. Maybe there's some threshold at which we refuse to automatically store the name for them?

threepointone:
[...]

PR #4351, opened 2025-06-16
Fix ctx.id.name for Durable Objects to match production.

kentonv:
In prdouction, ctx.id.name is not available. We wish it were, but there are reasons why it's not easy to do. See #2240

A recent refactoring of workerd accidentally made `ctx.id.name` available in local testing. This was not intended, as it doesn't match production. This commit removes the name, to match production behavior.


src/workerd/server/server.c++
```diff
@@ -2023,6 +2023,14 @@ class Server::WorkerService final: public Service,
     }
 
     kj::Own<IoChannelFactory::ActorChannel> getActorChannel(Worker::Actor::Id id) {
+      KJ_IF_SOME(doId, id.tryGet<kj::Own<ActorIdFactory::ActorId>>()) {
+        // To emulate production, we have to recreate this ID.
+        ActorIdFactoryImpl::ActorIdImpl* idImpl =
+            dynamic_cast<ActorIdFactoryImpl::ActorIdImpl*>(doId.get());
+        KJ_ASSERT(idImpl != nullptr, "Unexpected ActorId type?");
```
anonrig, line 2030:
```suggestion
        KJ_ASSERT_NONNULL(idImpl, "Unexpected ActorId type?");
```

kentonv: KJ_ASSERT_NONNULL is meant to be used with kj::Maybe values. The comparison here is on a pointer, so using this macro is inappropriate.

We should arguably rename KJ_ASSERT_NONNULL to KJ_ASSERT_SOME.

kentonv: Also FWIW even if this were comparing on a kj::Maybe, I think this would be an example of kind of thing that's not worth commenting on.

During coder reviews, it's extremely tempting to comment on every little thing that is not exactly the way the reviewer would have written it, but often it's not worth the time. If the code has a bug, is confusing, or contains maintainability pitfalls, then comment. Or if the reviewee might not be aware that there's a better way to write the code, then it might be worth commenting to help them learn. Otherwise, it may not be worth commenting -- better to spend our limited time on important things.

(FWIW I get this wrong a lot, have to remind myself during every code review.)

anonrig: > We should arguably rename KJ_ASSERT_NONNULL to KJ_ASSERT_SOME.

TIL. Yes I agree.

> During coder reviews, it's extremely tempting to comment on every little thing that is not exactly the way the reviewer would have written it, but often it's not worth the time. If the code has a bug, is confusing, or contains maintainability pitfalls, then comment. 

I agree in principal but most of the time, such comments are educational and beneficial for reviewers more than the author itself.

PR #4353, opened 2025-06-16
Python: Fix httpx content-encoding problem with Pyodide 0.27

hoodmane:
This fixes the content encoding problem with async httpx requests. The commits applied are here:
https://github.com/hoodmane/httpx/tree/pyodide


danlapid (approved):

danlapid:
Is it already published and available for vendoring users?

hoodmane:
Unfortunately not yet. However, Pyodide's release cycle is going to get faster for packages with the unvendoring work. I could release it in Pyodide 0.27.8 possibly.

hoodmane:
To pass the internal build, it requires an update to the httpx package snapshot.

hoodmane:
Or I guess it doesn't... hmm.

anonrig (approved):

PR #4354, opened 2025-06-16
EW-9298 [o11y] Base user span tracing off of WorkerTracer

fhanau:
This will make these systems work much more smoothly together, the lifetimes of top-level "worker" span and WorkerTracer are now aligned.

- Do not report span attributes when the set of attributes is empty.

Also see the downstream PR.


src/workerd/io/tracer.h
```diff
@@ -174,10 +179,16 @@ class WorkerTracer: public BaseTracer {
       tracing::EventInfo&& info) override;
   void setFetchResponseInfo(tracing::FetchResponseInfo&& info) override;
   void setOutcome(EventOutcome outcome, kj::Duration cpuTime, kj::Duration wallTime) override;
+  SpanParent getUserRequestSpan() override;
+  // Allow setting the user request span after the tracer has been created so its observer can
+  // reference the tracer. This can only be set once.
+  void setUserRequestSpan(SpanBuilder&& span);
```
mar-cf, line 185: My preference would be set this in a constructor. It would avoid a possible condition of getting the user request span before setting the user request span, or potentially missing setting this.

We tracked the patterns where this is set and get currently and don't believe this would be an issue - but it's a pattern we should clean up later.

src/workerd/io/tracer.c++
```diff
@@ -328,6 +334,10 @@ void WorkerTracer::setOutcome(EventOutcome outcome, kj::Duration cpuTime, kj::Du
   trace->cpuTime = cpuTime;
   trace->wallTime = wallTime;
 
+  // All spans must have wrapped up by now. Report the top-level "worker" span and deallocate the
+  // span builder which holds a reference to this workerTracer.
+  userRequestSpan.end();
```
mar-cf: Remove.

src/workerd/io/tracer.h
```diff
@@ -174,10 +179,16 @@ class WorkerTracer: public BaseTracer {
...
 
  private:
   PipelineLogLevel pipelineLogLevel;
   kj::Own<Trace> trace;
+  // The root span for the new tracing format.
+  SpanBuilder userRequestSpan;
```
mar-cf, line 191: Consider clearly establishing ownership of user request span by wrapping w/ kj::Own

src/workerd/io/tracer.c++
mar-cf, line 1: It will help to add context about what event was reported after the outcome seen event.

Add annotation about the tail event to the log. If it's not disruptive, consider an assert?

mar-cf (approved):
Generally LGTM. Offline, we spoke about the dependency cycle that results in the set/get pattern that could be avoided.

Without having additional information about the tail event that was reported, we can't know with confidence if this will solve the issue.

src/workerd/io/tracer.c++
fhanau: Added some additional information to be logged here, and a TODO to support stringify for the kj::OneOf.
An assert would be disruptive – the given issue indicates a program bug, but this may surface again in different ways and could also indicate a minor issue (e.g. a log missing based on being reported too late). In that case, we do not want to assert and force the request to fail.

src/workerd/io/tracer.h
fhanau: Note that the SpanBuilder being a class variable of WorkerTracer already defines ownership – the SpanBuilder is alive for as long as WorkerTracer is alive, it is not being co-owned by anyone else and ownership is not transferred. This is also what we have been doing previously, and how we are using SpanBuilder elsewhere. Making this kj::Own would introduce a memory allocation that is not needed.

src/workerd/io/tracer.c++
fhanau: 👍

src/workerd/io/tracer.h
fhanau: I believe this is possible by moving observer construction into the WorkerTracer constructor, but we'll likely need a new constructor for each of several code paths (for different observers) and it may end up being more verbose. How to do so will be more clear after integrating a follow-up PR that changes how the observer is configured internally.

src/workerd/io/tracer.h
fhanau: Note that this is in contrast with SpanParent, which does not indicate exclusive ownership and can only be used to create children for a span. But for that we don't use kj::Own either.

jasnell (approved):

PR #4355, opened 2025-06-16
NodeJS Compat: process with a compat flag

guybedford:
This reinstates https://github.com/cloudflare/workerd/pull/4271 exactly as in that PR, but with the exception that all new process properties, methods are exported as `undefined` unless the new experimental `nodejs_compat_process` flag is set. In addition the `process` global will not extend from event emitter unless this flag is set.

This way all user-facing changes on this PR can be gated behind a compat date.

The goal for removing experimental here would be to complete the stdio implementations ideally as well since these are often expected to find on process, but I would like to add those as a secondary follow-on PR still. Then we can look at removing experimental and enabling the compat date at some point next month ideally.


src/workerd/io/compatibility-date.capnp
```diff
@@ -832,4 +832,15 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # throw an error when unknown import attributes are encountered. In the new module registry
   # implementation the recommended behavior is what is implemented. With this compat flag
   # enabled, the original module registry implementation will follow the recommended behavior.
+
+  unsupportedProcessRealPlatform @95 :Bool
+      $compatEnableFlag("unsupported_process_real_platform")
```
vicb: Maybe clarify what "unsupported" means?

src/workerd/io/compatibility-date.capnp
```diff
@@ -832,4 +832,15 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $experimental;
+  # Enables support for process.platform to expose the actual system platform
+  # This will never ever be supported and is a WPT test path only.
+
+  nodeJsCompatProcess @96 :Bool
+      $compatEnableFlag("nodejs_compat_process")
+      $experimental;
+  # Enables full process compat support, instead of the former partial implementation.
```
vicb, line 847: I would argue that "Enables full process compat support, instead of the former partial implementation." is that super helpful.

- What is added or changed?
- Where will future bug fixes/changes go? (flag enabled/disabled)

One thing I have in mind here is that the unenv behavior will need to check/adapt to the value of this flag and w/o deeper details on the current and future shape of the process API, it will be very hard.

unenv is only one example of client code, customers will face the same problem.

Edit: one other thing that come to mind is how do we test dependencies between flags. When @anonrig added `node:timers`, the first version of the PR failed when `nodejs_compat` was not enabled. See #3377.

src/node/events.ts
```diff
@@ -3,5 +3,20 @@
 //     https://opensource.org/licenses/Apache-2.0
 //
 
-export * from 'node-internal:events';
+// Explicit list used to hide _initialized private export
+export {
```
anonrig, line 7: Can we move these somewhat unrelated changes to a different pull-request?

src/node/internal/events.ts
```diff
@@ -495,26 +497,34 @@ function _addListener(
...
+            'emitter.setMaxListeners() to increase limit'
+        ),
+        {
+          name: 'MaxListenersExceededWarning',
+          emitter: target,
+          type: type,
+          count: existing.length,
+        }
+      );
+      process.emitWarning?.(w);
```
anonrig: Can you put this warning generation into a if statement and check if process.emitWarning is enabled or not? And can you also add a comment stating how and under which conditions process.emitWarning can be defined/undefined?

src/node/internal/internal_utils.ts
```diff
@@ -373,3 +373,90 @@ export function callbackify<T extends (...args: unknown[]) => Promise<unknown>>(
   // @ts-expect-error
   return callbackified;
 }
+
+export function parseEnv(content: string): Record<string, string> {
```
anonrig, line 377: This file is for regular utility functions, but this function is for "node:util". Can we seperate these 2 declarations into separate files?

src/node/internal/process.d.ts
```diff
@@ -0,0 +1,7 @@
+export const versions: Record<string, string>;
+export function getEnvObject(): Record<string, string>;
+export function getBuiltinModule(id: string): object;
+export function processExitImpl(code: number): void;
+export const processPlatform: string;
```
anonrig: This file is already named "process", this prefixes seems to be redundant.

```suggestion
export function exitImpl(code: number): void;
export const platform: string;
```

src/node/process.ts
```diff
@@ -3,5 +3,72 @@
...
+  send,
+  traceDeprecation,
+  throwDeprecation,
+  sourceMapsEnabled,
+  stdin,
+  stdout,
+  stderr,
+  threadCpuUsage,
+} from 'node-internal:internal_process';
+export { default } from 'node-internal:internal_process';
```
anonrig, line 75: Is this the same as the following?

```js
import { default as defaultImpl } from 'node-internal:internal_process'
export default defaultImpl;
```

src/workerd/api/node/process.c++
```diff
@@ -0,0 +1,115 @@
...
+
+jsg::JsObject ProcessModule::getVersions(jsg::Lock& js) const {
+  auto versions = js.obj();
+  // Node.js version - represents the most current Node.js version supported
+  // by the platform, as defined in node-version.h
+  versions.set(js, "node"_kj, js.str(nodeVersion));
+  return versions;
+}
+
+void ProcessModule::processExitImpl(jsg::Lock& js, int code) {
```
anonrig:
```suggestion
void ProcessModule::exitImpl(jsg::Lock& js, int code) {
```

src/workerd/api/node/i18n.c++
```diff
@@ -268,6 +268,17 @@ jsg::BufferSource transcode(
       "Unable to transcode buffer");
 }
 
+kj::String getIcuVersion() {
```
anonrig: we should cache this

src/workerd/io/compatibility-date.capnp
```diff
@@ -832,4 +832,15 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # throw an error when unknown import attributes are encountered. In the new module registry
   # implementation the recommended behavior is what is implemented. With this compat flag
   # enabled, the original module registry implementation will follow the recommended behavior.
+
+  unsupportedProcessRealPlatform @95 :Bool
```
anonrig: I think we should have a better naming for this, such as:

```suggestion
  exposeActualProcessPlatform @95 :Bool
```

src/workerd/io/compatibility-date.capnp
```diff
@@ -832,4 +832,15 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatEnableFlag("unsupported_process_real_platform")
+      $experimental;
+  # Enables support for process.platform to expose the actual system platform
+  # This will never ever be supported and is a WPT test path only.
```
anonrig:
```suggestion
  # By default, Workerd will always expose "linux" as the process.platform
  # This flag ensures the support to enable internal use-cases that validates
  # the underlying operating system.
```

src/workerd/io/compatibility-date.capnp
anonrig: I agree. Something like `enableNodejsProcessMethods` and with a description of why these are added behind a compat flag would be beneficial for others.

src/node/util.ts
```diff
@@ -32,14 +32,14 @@ import {
...
-import { callbackify } from 'node-internal:internal_utils';
+import { callbackify, parseEnv } from 'node-internal:internal_utils';
 import {
   isReadableStream,
   isWritableStream,
   isNodeStream,
 } from 'node-internal:streams_util';
 export { inspect, format, formatWithOptions, stripVTControlCharacters };
-export { callbackify } from 'node-internal:internal_utils';
+export { callbackify, parseEnv } from 'node-internal:internal_utils';
```
anonrig, line 42: You can merge line 42 with line 41

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+  send = undefined,
+  traceDeprecation = undefined,
+  throwDeprecation = undefined,
+  sourceMapsEnabled = undefined,
+  threadCpuUsage = undefined,
+  kill = undefined,
+  ref = undefined,
+  unref = undefined;
+
+interface Process extends EventEmitter {
```
anonrig, line 636: Can you move this TS declaration to process.d.ts?

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+
+function setuid(id: number | string): void {
+  if (id !== 0) throw new ERR_EPERM({ syscall: 'setuid' });
+}
+
+function initgroups(
+  _user: number | string,
+  _extractGroup: number | string
+): void {
+  throw new ERR_EPERM({ syscall: 'initgroups' });
```
anonrig, line 589: Why do we throw an error, can you add comments to each one of these methods that explicitly throws an error?

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+  Object.assign(process.env, parsed);
+}
+
+// On the virtual filesystem, we only support group 0
+function getegid(): number {
+  return 0;
+}
+
+function getgid(): number {
+  return 0;
```
anonrig, line 544: Can you add a comment to each one of these magical numbers?

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+export function exit(code: number): void {
+  processImpl.processExitImpl(code);
+}
+
+function abort(): void {
+  processImpl.processExitImpl(1);
+}
+
+function uptime(): number {
+  return 0;
```
anonrig, line 522: Why do we return 0 uptime, can you add a comment?

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+  ERR_UNSUPPORTED_OPERATION,
+  NodeError,
+} from 'node-internal:internal_errors';
+
+import { default as processImpl } from 'node-internal:process';
+import {
+  _initialized as eventsInitialized,
+  default as EventEmitter,
+} from 'node-internal:events';
+import type { Buffer } from 'node:buffer';
```
anonrig, line 29: You can just use NodeJS.Buffer, rather than importing this.

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+  argv: string[] | undefined,
+  argv0: string | undefined,
+  execArgv: string[] | undefined,
+  arch: string | undefined,
+  config: object | undefined,
+  pid: number | undefined,
+  ppid: number | undefined,
+  allowedNodeEnvironmentFlags: Set<string> | undefined;
+
+export let hrtime:
```
anonrig, line 57: You can use `types/node` package for most of these things.

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+
+// TODO(soon): Implement stdio along with TTY streams.
+export const stdin = undefined;
+export const stdout = undefined;
+export const stderr = undefined;
+
+// TODO(soon): Implement along with FS work.
+export const chdir = undefined;
+export const cwd = undefined;
+export const umask = undefined;
```
anonrig: Are we going to implement these before removing experimental? If yes, let's keep it as it is, if not, I recommend (not blocking) to remove these.

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+
+  arch = 'x64';
+
+  config = {
+    target_defaults: {},
+    variables: {},
+  };
+
+  pid = 1;
+  ppid = 0;
```
anonrig, line 154: Can you add comments for these magical numbers?

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+    enabled: false,
+    nodeModules: false,
+    generatedCode: false,
+  };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+export function nextTick(cb: Function, ...args: unknown[]): void {
+  queueMicrotask(() => {
+    // eslint-disable-next-line
```
anonrig: Can you add the specific eslint rule that you're disabling?

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+    cb(...args);
+  });
+}
+
+interface EmitWarningOptions {
```
anonrig: Can we move these types to process.d.ts?

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+    (err as ErrorWithDetail).detail = detail;
+  }
+
+  // Capture stack trace using the provided constructor or emitWarning itself
+  // This excludes the constructor (and frames above it) from the stack trace
+  Error.captureStackTrace(err, ctor || emitWarning);
+
+  // Emit the warning event on the process object
+  // Use nextTick to ensure the warning is emitted asynchronously
+  nextTick(() => {
```
anonrig:
```suggestion
  queueMicrotask(() => {
```

src/node/internal/internal_process.ts
```diff
@@ -0,0 +1,810 @@
...
+  // A boolean value that is true if the current Node.js build includes the inspector.
+  inspector: false,
+  // A boolean value that is true if the current Node.js build supports loading ECMAScript modules using require().
+  // TODO(soon): Update this when we support ESM modules through require().
+  require_module: false,
+  // A boolean value that is true if the current Node.js build includes support for TLS.
+  tls: true,
+});
+
+// Unsupported features - implemented as 'undefined' so they can still be imported
```
anonrig, line 608: This is technically a breaking change. 

```
'exitCode' in process
```

will return true, with this declaration, but not if we don't include them. I'm fine with processing as it is.

cc @danlapid fyi

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -1,15 +1,103 @@
...
+  'throwDeprecation',
+  'title',
+  'traceDeprecation',
+  'umask',
+  'unref',
+  'uptime',
+  'version',
+  'versions',
+];
+export const processKeys = {
```
anonrig, line 83: Can you also add a test where you call getBuiltinModule() and directly call await import() etc.

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -60,5 +148,651 @@ export const processEnv = {
...
+  test() {
+    // TODO(soon): Implement these!
+    assert.strictEqual(process.cwd, undefined);
+    assert.strictEqual(process.chdir, undefined);
+    assert.strictEqual(process.umask, undefined);
+    assert.strictEqual(process.stdin, undefined);
+    assert.strictEqual(process.stdout, undefined);
+    assert.strictEqual(process.stderr, undefined);
+
+    // This may be implemented in future
```
anonrig:
```suggestion
    // These might be implemented in future
```

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -60,5 +148,651 @@ export const processEnv = {
...
+        `process.versions.${versionKey} should be a string`
+      );
+    }
+  },
+};
+
+export const processFeatures = {
+  test() {
+    assert.strictEqual(typeof process.features, 'object');
+    assert.ok(process.features !== null);
```
anonrig:
```suggestion
    assert.ok(process.features != null);
```

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -60,5 +148,651 @@ export const processEnv = {
...
+        order.push('nextTick3');
+      },
+      'hello',
+      'world'
+    );
+
+    order.push('sync');
+
+    // Wait for microtasks to complete
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
anonrig:
```suggestion
    await scheduler.wait(0);
```

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -60,5 +148,651 @@ export const processEnv = {
...
+      const diff = process.hrtime([0, 1e9 - 1]);
+      assert(diff[1] >= 0); // https://github.com/nodejs/node/issues/4751
+    }
+
+    // process.hrtime.bigint
+    // https://github.com/nodejs/node/blob/5f7dbf45a3d3e3070d5f58f9a9c2c43dbecc8672/test/parallel/test-process-hrtime-bigint.js
+    {
+      const start = process.hrtime.bigint();
+
+      await new Promise((resolve) => setTimeout(resolve, 1000));
```
anonrig: Can you also change other await new Promises as well?

```suggestion
      await scheduler.wait(1000);
```

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -60,5 +148,651 @@ export const processEnv = {
...
+    //     process.chdir(originalCwd);
+    //   }
+    // }
+  },
+};
+
+export const processRejectionListeners = {
+  async test() {
+    const e = new Error();
+    const unhandledPromise = new Promise((resolve) =>
```
anonrig: Can you replace these with

src/node/events.ts
guybedford: This is completely related, due to the cycle introduced by making process inherit from EventEmitter, so can't be separated.

src/node/process.ts
guybedford: Yes almost, but not quite - `export default defaultImpl` creates a new binding, while `export { default }` ensures a live binding.

src/workerd/api/node/i18n.c++
guybedford: Removed as it's unused.

src/workerd/api/node/process.c++
guybedford: Added

src/workerd/io/compatibility-date.capnp
guybedford: The `unsupported` is important to indicate this feature will never move out of experimental. I've replaced `real` with `actual`.

src/workerd/io/compatibility-date.capnp
guybedford: Where the current process support is minimal and partial, this implements a comprehensive process polyfill, but there are still methods that are explicitly unsupported and still left for polyfilling. As such the best way to name this is simply "process polyfill v2". I've updated the flag name to `enable_nodejs_process_v2`.

src/node/internal/events.ts
guybedford: I've added a comment clarifying this behaviour.

src/node/internal/internal_utils.ts
guybedford: `callbackify` is a utils function that is defined in internal_utils. We need a place to put `parseEnv` that can be used by both process and utils, general utils seems like the right place? I'd be weary of adding a new file or having `node:process` directly import from `node:util`, as we might hit more cycle issues down the line.

src/node/internal/process.d.ts
guybedford: Added.

src/node/internal/internal_utils.ts
anonrig: My perspective was around having "internal_NODEJS_MODULE_NAME" syntax across the internal folder. But not blocking.

src/workerd/io/compatibility-date.capnp
guybedford: Added a variant of this suggestion retaining the note that this will never be formally supported.

src/node/util.ts
guybedford: The bindings on line 41 come from `node-internal:internal_inspect`, not `node-internal:internal_utils`.

src/node/internal/internal_process.ts
guybedford: `process.d.ts` defines the `process.c++` binding and not the `internal_process.ts` module.

src/node/internal/internal_process.ts
guybedford: Added

src/node/internal/internal_process.ts
guybedford: Added.

src/node/internal/internal_process.ts
guybedford: Added.

src/node/internal/internal_process.ts
guybedford: Yes we should implement all of these before removing the experimental status.

src/node/internal/internal_process.ts
guybedford: Added.

src/node/internal/internal_process.ts
guybedford: Why does this matter?

src/node/internal/internal_process.ts
guybedford: Added.

src/node/internal/internal_process.ts
guybedford: Added.

src/node/internal/internal_process.ts
guybedford: Yes, as discussed this is the nature of fully implementing the module interface. Properties exist but are undefined, unless using the compat flag in which case they are defined. Every other user facing aspect of the compat change is otherwise gated.

src/workerd/api/node/tests/process-nodejs-test.js
guybedford: Added.

src/workerd/io/compatibility-date.capnp
guybedford: Added some clarification that it means it will never be non-experimental.

src/node/internal/internal_process.ts
guybedford: I was unable to easily use `types/node` here. If there's a follow-on simplification then certainly we can add that.

src/workerd/api/node/tests/process-nodejs-test.js
guybedford: Can you clarify what you mean by this?

src/node/internal/internal_utils.ts
guybedford: With the current naming of the `internal_` modules this isn't clear today because so many are already named after Node.js modules. For `util` specifically adding a note that it is a generic utility file and not necessarily specific to Node.js might be beneficial to clarify that.

anonrig (approved):

src/workerd/api/node/tests/process-nodejs-test.js
guybedford: I added a test here for `getBuiltinModule('process')` as well.

PR #4356, opened 2025-06-17
Extend existing KV test with instrumentation assertions

jmorrell-cloudflare:
This sets up a streaming tail worker for the existing KV tests, and then asserts against the telemetry that they emit today. Copies the approach in the [tail worker tests](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/tail-worker-test.js).

There is no KV implementation in workerd so the existing KV tests [mock each response](https://github.com/cloudflare/workerd/blob/main/src/workerd/api/kv-test.js#L9). I could duplicate this logic into a separate test file to not disturb the existing tests, but thought that keeping everything together makes sense. As we extend the instrumentation logic this will likely encourage us to improve the coverage in the existing tests.

### DIFF
--- a/src/workerd/api/kv-instrumentation-test.js
+++ b/src/workerd/api/kv-instrumentation-test.js
@@ -16,7 +16,7 @@ export default {
           break;
         case 'attributes': {
           let span = spans.get(event.spanId);
-          for (let {name, value} of event.event.info) {
+          for (let { name, value } of event.event.info) {
             span[name] = value;
           }
           spans.set(event.spanId, span);
@@ -41,7 +41,9 @@ export const test = {
 
     // Recorded streaming tail worker events, in insertion order,
     // filtering spans not associated with KV
-    let received = Array.from(spans.values()).filter(span => span.name !== 'jsRpcSession');
+    let received = Array.from(spans.values()).filter(
+      (span) => span.name !== 'jsRpcSession'
+    );
 
     // spans emitted by kv-test.js in execution order
     let expected = [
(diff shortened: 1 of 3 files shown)


src/workerd/api/kv-test.wd-test
```diff
@@ -10,8 +10,18 @@ const unitTests :Workerd.Config = (
...
+    (name = "tail", worker = .tailWorker, ),
   ],
 );
+
+const tailWorker :Workerd.Worker = (
+  modules = [
+    (name = "worker", esModule = embed "kv-instrumentation-test.js")
+  ],
+  compatibilityDate = "2024-10-14",
+  compatibilityFlags = ["experimental", "nodejs_compat", "streaming_tail_worker"],
```
fhanau: Nit: Currently the STW compat flag is only needed on tailed workers, dropping it here would make that more clear (of course we're aiming to do the opposite but that'll take time).

src/workerd/api/kv-instrumentation-test.js
```diff
@@ -0,0 +1,190 @@
...
+          break;
+        }
+      }
+    };
+  },
+};
+
+export const test = {
+  async test() {
+    // HACK: The prior tests terminates once the scheduled() invocation has returned a response, but
```
fhanau: Nit: Comment is cargo-culted, either update it to not refer to scheduled() or drop it.

src/workerd/api/kv-instrumentation-test.js
```diff
@@ -0,0 +1,190 @@
...
+
+    // Recorded streaming tail worker events, in insertion order,
+    // filtering spans not associated with KV
+    let received = Array.from(spans.values()).filter(
+      (span) => span.name !== 'jsRpcSession'
+    );
+
+    // spans emitted by kv-test.js in execution order
+    let expected = [
+      {
```
fhanau, line 65: The first four entries are identical – I wonder if compressing the output by using `{<first entry>} * 4 + [...]` would make this easier or harder for humans to read and maintain?
Overall the structured test output we compare to here is more verbose than the stringified output used in the existing tail worker test, but that's mostly a stylistic choice, this is certainly prettier.

fhanau (approved):

src/workerd/api/kv-instrumentation-test.js
jmorrell-cloudflare: These are currently the same, but as we improve the instrumentation and add any missing cases I expect that will change quickly. This formatting is better for reviewing the upcoming changes.

src/workerd/api/kv-instrumentation-test.js
jmorrell-cloudflare: I replaced the `wait` with a set of promises which should be a little less hacky https://github.com/cloudflare/workerd/pull/4356/files#diff-0d469bf60cca1c210a45ffb535cf9b7eba6c7b8189312c817322f2b9944830d8R55

PR #4357, opened 2025-06-17
[build] 06-17-25 Manual deps updates

fhanau:
- ~Roll Bazel 8.2.1:8.3.0~ this breaks CI entirely, possibly a bug within Bazel
- Update rules_cc toolchain
- Transition boringssl to bzlmod and update to latest

[nfc] Remove dh-primes.h
    We no longer support boringssl versions where these would need to be backported,
    and handling for this has moved to ncrypto anyway.


.bazelversion
```diff
@@ -1 +1 @@
-8.2.1
+8.3.0rc1
```
fhanau: - [ ] Wait for 8.3.0 proper to be released before merging.

PR #4360, opened 2025-06-17
Various refactoring for dynamic worker loading

kentonv:
This is just some simple refactoring needed for my upcoming dynamic worker loading feature. Figured I'd put these up and get them merged before conflicts arise (especially the first commit, which moves a bunch of code).

### DIFF
--- a/src/workerd/io/compatibility-date.c++
+++ b/src/workerd/io/compatibility-date.c++
@@ -95,8 +95,8 @@ kj::String currentDateStr() {
   return CompatDate::today().toString();
 }
 
-void compileCompatibilityFlags(kj::StringPtr compatDate,
-    capnp::List<capnp::Text>::Reader compatFlags,
+static void compileCompatibilityFlags(kj::StringPtr compatDate,
+    kj::HashSet<kj::String> flagSet,
     CompatibilityFlags::Builder output,
     Worker::ValidationErrorReporter& errorReporter,
     bool allowExperimentalFeatures,
@@ -124,14 +124,6 @@ void compileCompatibilityFlags(kj::StringPtr compatDate,
     case CompatibilityDateValidation::FUTURE_FOR_TEST:
       // No validation.
       break;
-  }
-
-  kj::HashSet<kj::String> flagSet;
-  flagSet.reserve(compatFlags.size());
-  for (auto flag: compatFlags) {
-    flagSet.upsert(kj::str(flag), [&](auto& existing, auto&& newValue) {
-      errorReporter.addError(kj::str("Compatibility flag specified multiple times: ", flag));
-    });
   }
 
   auto schema = capnp::Schema::from<CompatibilityFlags>();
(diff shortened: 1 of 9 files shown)


src/workerd/io/compatibility-date.c++
```diff
@@ -252,6 +244,42 @@ void compileCompatibilityFlags(kj::StringPtr compatDate,
...
+void compileCompatibilityFlags(kj::StringPtr compatDate,
+    capnp::List<capnp::Text>::Reader compatFlags,
+    CompatibilityFlags::Builder output,
+    Worker::ValidationErrorReporter& errorReporter,
+    bool allowExperimentalFeatures,
+    CompatibilityDateValidation dateValidation) {
+  kj::HashSet<kj::String> flagSet;
+  flagSet.reserve(compatFlags.size());
+  for (auto flag: compatFlags) {
+    flagSet.upsert(kj::str(flag), [&](auto& existing, auto&& newValue) {
```
anonrig, line 256: I'm not sure about the lifecycles here but can we get away from making an extra copy for these flags and convert set to StringPtr?

src/workerd/server/server.c++
```diff
@@ -1792,7 +1787,19 @@ struct ErrorReporter: public Worker::ValidationErrorReporter {
...
   }
 };
-}  // namespace
+
+// Implementation of ErrorReporter specifically for reporting errors in the top-level workerd
+// config.
+struct Server::ConfigErrorReporter final: public ErrorReporter {
+  ConfigErrorReporter(Server& server, kj::StringPtr name): server(server), name(name) {}
+
+  Server& server;
```
anonrig, line 1796: Would you mind adding a comment about the ownership and the lifecycle expectancy for this?

src/workerd/server/server.c++
```diff
@@ -3502,6 +3533,75 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
     errorReporter.addError(kj::str("Worker must specify compatibilityDate."));
   }
 
+  kj::Vector<FutureSubrequestChannel> subrequestChannels;
+  kj::Vector<FutureActorChannel> actorChannels;
+  kj::Vector<FutureActorClassChannel> actorClassChannels;
+
+  auto confBindings = conf.getBindings();
+  kj::Vector<WorkerdApi::Global> globals(confBindings.size());
```
anonrig, line 3541: Does this vector grow? Can we use heapArray?

src/workerd/server/server.c++
```diff
@@ -3686,20 +3775,18 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
     { auto drop = kj::mv(ctxExportsHandle); }
   });
 
-  auto linkCallback = [this, name, conf, subrequestChannels = kj::mv(subrequestChannels),
-                          actorChannels = kj::mv(actorChannels),
-                          actorClassChannels = kj::mv(actorClassChannels),
-                          &localActorConfigs](WorkerService& workerService) mutable {
+  auto linkCallback = [this, def = kj::mv(def)](WorkerService& workerService,
```
anonrig, line 3778: Question for education purposes: why don't we use jsg visitable lambda here?

src/workerd/server/server.c++
```diff
@@ -3713,26 +3800,26 @@ kj::Own<Server::Service> Server::makeWorker(kj::StringPtr name,
...
     // Link the ctx.exports self-referential channels. Note that it's important these are added
     // in exactyl the same order as the channels were allocated earlier when we compiled the
     // ctx.exports bindings.
     if (workerService.hasDefaultEntrypoint()) {
-      services.add(KJ_ASSERT_NONNULL(workerService.getEntrypoint(kj::none, kj::none)));
+      services.add(KJ_ASSERT_NONNULL(workerService.getEntrypoint(kj::none, {})));
     }
     for (auto& ep: entrypointNames) {
-      services.add(KJ_ASSERT_NONNULL(workerService.getEntrypoint(ep, kj::none)));
+      services.add(KJ_ASSERT_NONNULL(workerService.getEntrypoint(ep, {})));
```
anonrig: Can you add a comment next to the {} variable for the variable name?

src/workerd/server/server.c++
kentonv: Two lines down, we only add to this vector conditionally if `createBinding()` returns non-null. We don't know the actual size until then, though it usually ends up being `confBindings.size()`.

In any case this code is not new, just moved.

src/workerd/server/server.c++
kentonv: Because this isn't a JavaScript object.

src/workerd/server/server.c++
kentonv: The `Server` is the top-level object of the entire process. `ConfigErrorReporter` is a nested class of `Server`.

It's very common for nested objects to have references back to their parent, and it's normal to assume that it's the parent's responsibility to ensure that it doesn't allow instances of the nested object to outlive itself. I think commenting on that here would add more noise than it is worth.

src/workerd/io/compatibility-date.c++
kentonv: Maybe, but I only moved this code from above. Optimizing it is not part of my objective here.

src/workerd/server/server.c++
kentonv: Sure: e26c3d1d5

But I wonder if such comments make sense anymore with modern IDEs that can be configured to display the param names automatically.

src/workerd/server/server.c++
```diff
@@ -1761,21 +1761,16 @@ class NullIsolateLimitEnforcer final: public IsolateLimitEnforcer {
...
 };
 
-struct ErrorReporter: public Worker::ValidationErrorReporter {
-  ErrorReporter(Server& server, kj::StringPtr name): server(server), name(name) {}
-
-  Server& server;
-  kj::StringPtr name;
+}  // namespace
 
+// Shared ErrorReporter base implemnetation. The logic to collect entrypoint information is the
```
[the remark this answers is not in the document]
kentonv: Sorry bot but I just don't care.

PR #4361, opened 2025-06-17
First pass at instrumenting kv.list

jmorrell-cloudflare:
~~Note: depends on merging https://github.com/cloudflare/workerd/pull/4356 first~~

- Modifies some of the plumbing around getting an HTTP client so that `TraceContext` can be passed along, and modifies `kv.list` to use the new path.
- Adds more attributes for `kv.list` plus tests


src/workerd/api/kv.c++
```diff
@@ -60,12 +67,55 @@ static void parseListMetadata(
...
     obj.set(js, "cacheStatus"_kjc, cacheStatus.orDefault(js.null()));
+    KJ_IF_SOME(cs, cacheStatus) {
+      traceContext.userSpan.setTag("cloudflare.kv.response.cache_status"_kjc, cs.toString(js));
+    }
   });
 }
 
 constexpr auto FLPROD_405_HEADER = "CF-KV-FLPROD-405"_kj;
 
+kj::Own<kj::HttpClient> KvNamespace::getHttpClient(IoContext& context,
```
jmorrell-cloudflare, line 94: The other `KvNamespace::getHttpClient` creates spans for every method all at once. This works only if you do not want to introspect the response or capture errors during execution.

This implementation allows us more flexibility with instrumentation. I will move each KV method over to it as I instrument them and then delete the other.

src/workerd/api/kv.c++
```diff
@@ -40,15 +40,22 @@ static void validateKeyName(kj::StringPtr method, kj::StringPtr name) {
...
+  static constexpr auto CURSOR = "cursor"_kjc;
+  static constexpr auto LIST_COMPLETE = "list_complete"_kjc;
+  static constexpr auto EXPIRATION = "expiration"_kjc;
 
   js.withinHandleScope([&] {
     auto obj = KJ_ASSERT_NONNULL(listResponse.tryCast<jsg::JsObject>());
     KJ_IF_SOME(keysArr, obj.get(js, KEYS).tryCast<jsg::JsArray>()) {
       auto length = keysArr.size();
+      traceContext.userSpan.setTag(
+          "cloudflare.kv.response.returned_rows"_kjc, static_cast<int64_t>(length));
```
jmorrell-cloudflare, line 58: This would be a little cleaner if we supported more `int` types as tag values

@fhanau Would you be open to a PR adding int32 and uint32 to this type? https://github.com/cloudflare/workerd/blob/main/src/workerd/io/trace.h#L571

fhanau: adding more types for tags would add lots of boilderplate code to places like spanTagClone(). I don't think it really matters here – since you added the cast I assume the compiler refuses to automatically cast here, but a simple cast seems acceptable here too.

src/workerd/api/kv.c++
```diff
@@ -424,23 +473,34 @@ jsg::Promise<jsg::JsRef<jsg::JsValue>> KvNamespace::list(
...
       }
       KJ_IF_SOME(maybePrefix, o.prefix) {
         KJ_IF_SOME(prefix, maybePrefix) {
+          traceContext.userSpan.setTag("cloudflare.kv.query.prefix"_kjc, kj::str(prefix));
           url.query.add(kj::Url::QueryParam{kj::str("prefix"), kj::str(prefix)});
         }
       }
       KJ_IF_SOME(maybeCursor, o.cursor) {
         KJ_IF_SOME(cursor, maybeCursor) {
+          traceContext.userSpan.setTag("cloudflare.kv.query.cursor"_kjc, true);
```
fhanau, line 503: Doesn't this overlap with the existing `cloudflare.kv.query.parameter.cursor` tag? Same applies for some related tags.

src/workerd/api/kv.c++
```diff
@@ -424,23 +473,34 @@ jsg::Promise<jsg::JsRef<jsg::JsValue>> KvNamespace::list(
     jsg::Lock& js, jsg::Optional<ListOptions> options) {
   return js.evalNow([&] {
     auto& context = IoContext::current();
+    auto traceSpan = context.makeTraceSpan("kv.list"_kjc);
+    auto userSpan = context.makeUserTraceSpan("kv.list"_kjc);
+    TraceContext traceContext(kj::mv(traceSpan), kj::mv(userSpan));
+
+    traceContext.userSpan.setTag("db.system"_kjc, kj::str("cloudflare.kv"_kjc));
```
fhanau, line 480: I don't think this should be set within list itself – see the existing getHttpClient(), where we set this attribute for a number of operations at once.
The value of the tag is not matching what we have there (cloudflare.kv vs. cloudflare-kv, for the operation name we have kv.list vs. kv_list)

src/workerd/api/kv.c++
```diff
@@ -424,23 +473,34 @@ jsg::Promise<jsg::JsRef<jsg::JsValue>> KvNamespace::list(
...
+    traceContext.userSpan.setTag("db.operation.name"_kjc, kj::str("list"_kjc));
+    traceContext.userSpan.setTag("cloudflare.binding_type"_kjc, kj::str("KV"_kjc));
```
fhanau, line 482: If we add this we'd want to have it for each binding, right? Not opposed to adding it per se, but that's something that should be discussed internally first so that we agree that it's needed/what the right tag name is, and it should be added across bindings in a different change set.

src/workerd/api/kv.c++
jmorrell-cloudflare: See my other comment. The existing code trying to add all the instrumentation code in one place is less flexible given that it cannot observe anything about the response. The idea is to lift all of the instrumentation to live alongside the code it instruments.

src/workerd/api/kv.c++
jmorrell-cloudflare: Yes, this modifies the existing instrumentation and extends it quite a bit: https://wiki.cfdata.org/display/WOBS/June+2025%3A+Next+steps+for+Runtime+Instrumentation

There are no dependencies on this existing instrumentation that I'm aware of. Am I missing something?

src/workerd/api/kv.c++
jmorrell-cloudflare: I'm certainly open to renaming it, but we absolutely want some equivalent of this attribute for all of our bindings

fhanau: My point is that this is a KV instrumentation PR – adding "global" tags is out of scope here.

src/workerd/api/kv.c++
fhanau: The naming scheme should always be consistent, everything else is likely to cause confusion, especially when renaming things gets forgotten until code is used in production. I'm not opposed to changing the naming scheme per se, but this should be done for all KV operations, in a commit or PR either before or after adding new tags. Everything else is difficult to review.
We also need to be mindful of keeping a consistent source of truth for this – both the old source of truth and the new document you added use db.system:cloudflare-kv, the tag implementation needs to be consistent with its specification.

PR #4363, opened 2025-06-18
[o11y] Fix STW string cache memory management

fhanau:


src/workerd/io/trace-stream.c++
```diff
@@ -501,7 +501,10 @@ jsg::JsValue ToJs(jsg::Lock& js, const tracing::Exception& ex, StringCache& cach
 }
 
 jsg::JsValue ToJs(jsg::Lock& js, const LogLevel& level, StringCache& cache) {
-  return cache.get(js, toLower(enumToStr<LogLevel>(level)));
+  // This must be assigned to a variable to avoid the string being implicitly converted to
+  // StringPtr in the cache.get() call, leading to a use-after-free.
+  kj::String levelStr = toLower(enumToStr<LogLevel>(level));
+  return cache.get(js, levelStr);
```
dom96: Looking at the implementation of `get`, it seems to store the kj::StringPtr in the cache so won't this still end up with a use-after-free? Can `get` be changed to take a kj::String instead?

fhanau: I asked about this internally and think so too. I'll probably end up only using string constants here instead.

src/workerd/util/strings.h
```diff
@@ -77,10 +77,10 @@ constexpr bool isAlphaLower(const kj::byte c) noexcept {
 }
 
 // Convert ASCII alpha characters in the given string to lowercase in place.
-kj::String toLower(kj::String&& str);
+kj::String toLower(kj::String&& str KJ_LIFETIMEBOUND);
```
fhanau: @danlapid you suggested that adding KJ_LIFETIMEBOUND is sufficient to fix this? There is no compile error or warning with the old code in trace-stream after adding KJ_LIFETIMEBOUND, does the compiler just extend the lifetime on its own?

src/workerd/io/trace-stream.c++
fhanau: After internal discussion, we believe that this is the only issue going on here. Addressed in updated commit.

mikea (approved):

PR #4364, opened 2025-06-18
Don't report disk usage in Windows build.

kentonv:
For some reason this step takes 16 minutes on Windows. It only takes a couple seconds on Mac and Linux.

I assume these usage stats are not actually that important to collect on every build, so let's disable it.


kentonv:
I just checked the last five successful builds and saw:

1m6s, 10m52s, 16m13s, 1m5s, 52s

So yes, I guess it often takes only a minute, but also often takes 10+ minutes.

kentonv:
Internal build failure obviously spurious, bypassing.

Issue #4362, opened 2025-06-18
🐛 Bug Report — container status is incorrect after destroying container without monitor


emily-shen:
To reproduce:

1. start a container
2. await container.destroy();
wait for it actually destroy. checked with docker ps -a, definitely not there anymore
3. check container.running - it returns true. Trying `getTCPPort().connect()` correctly errors with `connect(): Connection refused: container port not found. Make sure you exposed the port in your container definition.`

if you also add `await monitor` after destroying, we correctly get `container.running` as false.

@anonrig

emily-shen:
seems to happen in prod too 🤔

anonrig:
Container::destroy() doesn't set internal state "running" to false. That's why you're seeing this error. This is happening because we're storing a "running" state in the runtime, but in reality, it can be started/stopped by external factors, etc.

```
jsg::Promise<void> Container::destroy(jsg::Lock& js, jsg::Optional<jsg::Value> error) {
  if (!running) return js.resolvedPromise();

  if (destroyReason == kj::none) {
    destroyReason = kj::mv(error);
  }

  return IoContext::current().awaitIo(
      js, rpcClient->destroyRequest(capnp::MessageSize{4, 0}).sendIgnoringResult());
}
```

anonrig:
> if you also add await monitor after destroying, we correctly get container.running as false.

Monitor works because it actually sets `running = false`.

PR #4365, opened 2025-06-18
update internal state on container.destroy

anonrig:
Fixes https://github.com/cloudflare/workerd/issues/4362


mikenomitch (commented):
This seems correct to me (talked it thru with Yagiz too) but @gabivlj should take a quick look as well.

emily-shen (approved):

kentonv:
Guys we've discussed this before and this isn't correct.

PR #4368, opened 2025-06-18
do not set running to false on container monitor fail

anonrig:
Currently, we set running to false even if the container API returned a failing response. This pull-request ensures that we're updating "running" correctly.

### DIFF
--- a/src/workerd/api/container.c++
+++ b/src/workerd/api/container.c++
@@ -65,7 +65,6 @@ jsg::Promise<void> Container::monitor(jsg::Lock& js) {
       js.throwException(kj::mv(error));
     }
   }, [this](jsg::Lock& js, jsg::Value&& error) {
-    running = false;
     destroyReason = kj::none;
     js.throwException(kj::mv(error));
   });
--- a/src/workerd/server/container-client.c++
+++ b/src/workerd/server/container-client.c++
@@ -372,15 +372,19 @@ kj::Promise<void> ContainerClient::monitor(MonitorContext context) {
       co_await timer.afterDelay(1 * kj::SECONDS);
       continue;
     }
-    JSG_REQUIRE(response.statusCode == 200, Error,
+    JSG_REQUIRE(response.statusCode == 200 || response.statusCode == 404, Error,
         "Monitoring container failed with: ", response.statusCode, response.body);
-    // Parse JSON response
-    auto jsonRoot = decodeJsonResponse<docker_api::Docker::ContainerMonitorResponse>(response.body);
-    auto statusCode = jsonRoot.getStatusCode();
-    JSG_REQUIRE(statusCode == 0, Error, "Container exited with unexpected exit code ", statusCode);
+
+    // If container is not found, we ignore it to set "running" correctly in the container orchestrator.
+    if (response.statusCode == 200) {
+      auto jsonRoot =
+          decodeJsonResponse<docker_api::Docker::ContainerMonitorResponse>(response.body);
+      auto statusCode = jsonRoot.getStatusCode();
+      JSG_REQUIRE(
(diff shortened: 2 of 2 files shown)


src/workerd/api/container.c++
```diff
@@ -65,7 +65,6 @@ jsg::Promise<void> Container::monitor(jsg::Lock& js) {
       js.throwException(kj::mv(error));
     }
   }, [this](jsg::Lock& js, jsg::Value&& error) {
-    running = false;
```
danlapid, line 68: This will change behavior even in prod, are we aligned with CC team?

anonrig: cc @mikenomitch

src/workerd/server/container-client.c++
```diff
@@ -372,15 +372,19 @@ kj::Promise<void> ContainerClient::monitor(MonitorContext context) {
...
     }
-    JSG_REQUIRE(response.statusCode == 200, Error,
+    JSG_REQUIRE(response.statusCode == 200 || response.statusCode == 404, Error,
         "Monitoring container failed with: ", response.statusCode, response.body);
-    // Parse JSON response
-    auto jsonRoot = decodeJsonResponse<docker_api::Docker::ContainerMonitorResponse>(response.body);
-    auto statusCode = jsonRoot.getStatusCode();
-    JSG_REQUIRE(statusCode == 0, Error, "Container exited with unexpected exit code ", statusCode);
+
+    // If container is not found, we ignore it to set "running" correctly in the container orchestrator.
```
danlapid, line 378: I don't understand this code change, can you explain what's going on? What's the bug? what are you trying to fix?

anonrig: Before this change:

- Container API returns 500 or 404, we set running to false.

With this change:

- If container API returns 404, we succeed so that container class can set running=false
- If container API returns 500, we don't set running=false on container class, and keep the current state of running attribute.

kentonv (changes requested):
```
  monitor @2 ();
  # Waits for the container to shut down.
  #
  # If the container shuts down because the root process exited with a success status, or because
  # the client invoked `destroy()`, then `monitor()` completes without an error. If it shuts down
  # for any other reason, `monitor()` throws an exception describing what happened. (This exception
  # may or may not be a JSG exception depending on whether it is an application error or a system
  # error.)
```

Per the docs, it is correct `running = false` when `monitor()` throws.

src/workerd/api/container.c++
gabivlj: When monitor fails I think we should keep it that running should be false _for now_  as monitor might return an error if there is an exit code that is not zero. I think we want Monitor() to not throw an error and return an exit code, and make workerd throw an error with an exitCode property:
https://github.com/cloudflare/workerd/pull/4369#discussion_r2155076343

anonrig: If monitor returns 500 (which is a container engine api error), we still change it to running = false, which is false behavior.

anonrig:
> Per the docs, it is correct running = false when monitor() throws.

@kentonv What is your take on when engine api returns 404 on monitor request? Should it throw or succeed?

src/workerd/api/container.c++
gabivlj: in the case of a 500, should we just try to destroy the resources so the user tries to start() again? On our implementation, if there is an internal error it usually means a disconnect, which we will ask for another container instance.

kentonv:
It's an error to call `monitor()` when the container is not running. In fact, that will fail on the client side without even making the RPC, I think.

But as soon as `start()` has been called, then `monitor()` should succeed, and only return/throw when the started container exits.

I do not know under what conditions Docker returns 404 exactly.

PR #4369, opened 2025-06-18
containers: Make sure the exit code error pattern matches production

gabivlj:


src/workerd/server/container-client.c++
```diff
@@ -377,7 +377,8 @@ kj::Promise<void> ContainerClient::monitor(MonitorContext context) {
     // Parse JSON response
     auto jsonRoot = decodeJsonResponse<docker_api::Docker::ContainerMonitorResponse>(response.body);
     auto statusCode = jsonRoot.getStatusCode();
-    JSG_REQUIRE(statusCode == 0, Error, "Container exited with unexpected exit code ", statusCode);
+    // NOTE: the following error pattern matches the containers service in production.
+    JSG_REQUIRE(statusCode == 0, Error, "container exited with unexpected exit code: ", statusCode);
```
kentonv: JSG errors are supposed to be complete sentences starting with a capital letter, so arguably production is using the wrong style here.

Are people string-matching this? Why does the capitalization matter?

gabivlj: I think people will string-match an exception thrown by monitor to get the exit code if we don't add the exitCode as a property on the error. 

We could change the monitor capnp schema to not throw an error and always return the exit code, so its easy for us on workerd to retrieve the exit code and throw the specific error pattern (with the exitCode property set). 

For now, I think I will change the pattern on production to be capitalised, I think the ':' should be kept for ease of reading.

kentonv: The ':' is fine, yes.

kentonv (approved):

PR #4370, opened 2025-06-18
Revert "update internal state on container.destroy"

kentonv:
Reverts cloudflare/workerd#4365

This is not the correct solution, per internal conversation.


anonrig:
Opened a PR to address this: https://github.com/cloudflare/workerd/pull/4371

PR #4371, opened 2025-06-18
call monitor on container destroy

anonrig:
Addresses https://github.com/cloudflare/workerd/issues/4362 from a different perspective


kentonv:
> What we actually need to do is make a call to monitor() automatically internally, so that we can keep the running boolean accurate. The JS monitor() method will have to piggy-back on the internal call's result.

To clarify, we should call `monitor()` automatically in `start()` (or in the constructor, if the container is already running).

src/workerd/api/container.c++
```diff
@@ -48,24 +48,38 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
   running = true;
 }
 
 jsg::Promise<void> Container::monitor(jsg::Lock& js) {
   JSG_REQUIRE(running, Error, "monitor() cannot be called on a container that is not running.");
 
+  monitoring = true;
+
   return IoContext::current()
       .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).sendIgnoringResult())
```
kentonv: I think instead of making a second monitor request, we should piggy-back on the existing background request, i.e. return a copy of the same promise.

src/workerd/api/container.c++
```diff
@@ -48,24 +48,38 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
   }
 
   IoContext::current().addTask(req.sendIgnoringResult());
+  IoContext::current().addTask(rpcClient->monitorRequest(capnp::MessageSize{4, 0})
```
kentonv: We also need to automatically set up monitoring in the constructor if the container is already running.

src/workerd/api/container.c++
anonrig: kj::Promise<void> doesn't have a copy constructor. Any other suggestions?

anonrig: Promises (kj::Promise) are not copyable. Am I missing something?

src/workerd/api/container.c++
```diff
@@ -48,17 +48,45 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+  IoContext::current().addTask(kj::mv(req));
 }
 
 jsg::Promise<void> Container::monitor(jsg::Lock& js) {
   JSG_REQUIRE(running, Error, "monitor() cannot be called on a container that is not running.");
 
+  monitoringExplicitly = true;
+
   return IoContext::current()
       .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).send())
```
kentonv: I think we should piggy-back on the existing monitor request rather than make a new one? Can we create a promise-resolver, and put the resolver in a list to be resolved when the background monitor completes?

kentonv: Or perhaps, instead of a promise-resolver queue, store the final promise from the background monitor call into a `jsg::MemoizedIdentity` which you return whenever `monitor()` is called from JS.

src/workerd/api/container.c++
```diff
@@ -48,18 +48,74 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+                           .then([](auto&& results) -> uint8_t {
+      return results.getExitCode();
+    }).fork();
+  }
+
+  auto req = KJ_ASSERT_NONNULL(monitorKjPromise)
+                 .addBranch()
+                 .then([this, self = JSG_THIS](uint8_t exitCode) {
+    running = false;
+    monitoring = false;
```
jasnell: You can likely get away with using an `.attach(kj::defer([this, self = JSG_THIS]() { running = false; monitoring = false; }))` to the promise rather than explicitly handling it in both the then and error handler to eliminate the duplication here.

src/workerd/api/container.h
```diff
@@ -62,15 +62,24 @@ class Container: public jsg::Object {
...
   IoOwn<rpc::Container::Client> rpcClient;
   bool running;
+  bool monitoring;
+  bool monitoringExplicitly;
 
   kj::Maybe<jsg::Value> destroyReason;
+  void monitorOnBackgroundIfNeeded();
+
+  // Store the shared monitor promise and its KJ counterpart
+  kj::Maybe<kj::ForkedPromise<uint8_t>> monitorKjPromise;
```
jasnell: I believe the `kj::ForkedPromise<uint8_t>` here may need to be held in an `IoOwn`? @kentonv could confirm for sure.

src/workerd/api/container.c++
anonrig: Nice. Thanks for the recommendation.

anonrig:
@kentonv This is ready for another round of reviews. Would you mind taking a look?

src/workerd/api/actor-state.c++
```diff
@@ -945,7 +945,9 @@ DurableObjectState::DurableObjectState(jsg::Lock& js,
       exports(kj::mv(exports)),
       storage(kj::mv(storage)),
       container(container.map([&](rpc::Container::Client& cap) {
-        return js.alloc<Container>(kj::mv(cap), containerRunning);
+        auto container = js.alloc<Container>(kj::mv(cap), containerRunning);
+        container->monitorOnBackgroundIfNeeded();
```
kentonv: Can this be called from `Container`'s constructor, rather than make it the caller's responsibility? Then you can remove the `friend class DurableObjectState` declaration in `Container`.

src/workerd/api/container.c++
```diff
@@ -48,18 +48,64 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+  monitorOnBackgroundIfNeeded();
+}
+
+void Container::monitorOnBackgroundIfNeeded() {
+  if (monitoring || !running) {
+    return;
+  }
+  monitoring = true;
+
+  if (monitorKjPromise == kj::none) {
```
kentonv: I think there's a bug here if we're on the _second_ run of the container: `monitorKjPromise` may be filled in with a leftover promise from the first run, which is never nulled out even when it completes.

Please add a test that starts a container, monitors it, shuts it down, starts it again, and monitors that. The second monitor() call should monitor the second run of the container, but I think with the current code it'll return immediately because it returns a Promise created during the first run.

I think the best fix is to simply delete this if() statement, and also delete the repeated code in `monitor()` below that creates this promise, instead having `monitor()` rely on `monitorOnBackgroundIfNeeded()` to create the promise. (See comment there.)

src/workerd/api/container.c++
```diff
@@ -48,18 +48,64 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
-      .then(js, [this](jsg::Lock& js, capnp::Response<rpc::Container::MonitorResults> results) {
+  monitoringExplicitly = true;
+
+  KJ_IF_SOME(memoized, monitorJsPromise) {
+    return memoized;
+  }
+
+  if (monitorKjPromise == kj::none) {
+    auto responsePromise =
+        rpcClient->monitorRequest(capnp::MessageSize{4, 0})
```
kentonv: Why do you need to repeat the `monitorRequest()` code here? Couldn't you call `monitorOnBackgroundIfNeeded()` first, and then rely on it having created `monitorKjPromise`?

src/workerd/api/container.c++
```diff
@@ -48,18 +48,64 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
-jsg::Promise<void> Container::monitor(jsg::Lock& js) {
+jsg::MemoizedIdentity<jsg::Promise<void>>& Container::monitor(jsg::Lock& js) {
   JSG_REQUIRE(running, Error, "monitor() cannot be called on a container that is not running.");
 
-  return IoContext::current()
-      .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).send())
-      .then(js, [this](jsg::Lock& js, capnp::Response<rpc::Container::MonitorResults> results) {
+  monitoringExplicitly = true;
+
+  KJ_IF_SOME(memoized, monitorJsPromise) {
```
kentonv, line 205: This, too, might return a stale `monitorJsPromise` -- you'll need to null that out on container shutdown.

src/workerd/api/container.c++
```diff
@@ -73,9 +119,16 @@ jsg::Promise<void> Container::monitor(jsg::Lock& js) {
...
   }, [this](jsg::Lock& js, jsg::Value&& error) {
     running = false;
+    monitoringExplicitly = false;
     destroyReason = kj::none;
     js.throwException(kj::mv(error));
   });
+
+  monitorJsPromise = jsg::MemoizedIdentity<jsg::Promise<void>>(kj::mv(jsPromise));
+
+  monitorOnBackgroundIfNeeded();
```
kentonv: I think you actually want to call this in `start()`, not in `monitor()`? Otherwise, as currently written, if the container is not running at startup, and the application calls `start()` explicitly, but never calls `monitor()`, we won't get the automatic `abort()` on failure.

Please add a test for this.

src/workerd/api/container.c++
```diff
@@ -48,18 +48,64 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+    // If not, we can abort the current context and error.
+    if (!monitoringExplicitly) {
+      IoContext::current().abort(kj::mv(error));
+    }
+  }).attach(kj::defer([this, self = JSG_THIS]() {
+    running = false;
+    monitoring = false;
+  }));
+
+  IoContext::current().addTask(kj::mv(req));
```
kentonv: Adding a task to the `IoContext` here will prevent the DO from hibernating, as hibernation is only allowed when there are no tasks left. So, we should probably avoid this.

There is no need to add the task in any case. You can simply hold the `Promise` in the `Container` object directly (using `.eagerlyEvaluate(nullptr)` to make sure it actually executes to completion). If the `Container` object is destroyed, that implies the DO has shut down, in which case it's fine for us to cancel this promise -- there's no longer an IoContext to abort anyway.

src/workerd/api/container.c++
```diff
@@ -48,18 +48,64 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+    }).fork();
+  }
+
+  auto req = KJ_ASSERT_NONNULL(monitorKjPromise)
+                 .addBranch()
+                 .then([](uint8_t exitCode) {}, [this, self = JSG_THIS](kj::Exception&& error) {
+    // Check if there is an existing monitor request made by the user.
+    // If not, we can abort the current context and error.
+    if (!monitoringExplicitly) {
+      IoContext::current().abort(kj::mv(error));
```
kentonv: Since this is a KJ promise continuation, when it runs, the isolate lock will not be held, and there will be no current `IoContext`, so this throws an exception.

You need to capture `IoContext::current().getWeakRef()` instead, and use that to call `abort()` (which can be called without holding the lock). I would also recommend putting the `monitoringExplicitly` flag into a separate refcounted object so that you aren't reaching into a JSG object here -- that way you don't need to capture the `self` reference, which isn't safe to access without the lock held. (Note that the separate refcounted flag object would need to be held by the `Container` using an `IoOwn`.)

src/workerd/api/container.h
kentonv: Yes, it needs to be.

src/workerd/api/container.c++
```diff
@@ -48,18 +48,64 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+    }
+  }).attach(kj::defer([this, self = JSG_THIS]() {
+    running = false;
```
kentonv: Here, too, you don't have the isolate lock, so accessing the JSG object is unsafe. Perhaps these flags can also be moved to the separate refcounted object.

Or, alternatively, perhaps updating these flags can be done in a continuation of the JSG promise returned by `monitor()`, instead of here. Note that if `monitor()` wasn't called, then we will have aborted the `IoContext` anyway, in which case updating these flags is not important.

src/workerd/api/container.c++
```diff
@@ -48,18 +48,64 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+        rpcClient->monitorRequest(capnp::MessageSize{4, 0})
+            .send()
+            .then([](capnp::Response<rpc::Container::MonitorResults>&& results) -> uint8_t {
+      return results.getExitCode();
+    });
+    monitorKjPromise = responsePromise.fork();
+  }
+
+  auto jsPromise = IoContext::current()
+                       .awaitIo(js, KJ_ASSERT_NONNULL(monitorKjPromise).addBranch())
```
kentonv: This is not a new problem in your PR, so I'm not asking you to fix it, but it occurs to me that this `awaitIo()` will block the DO from hibernating until the `monitor()` call completes. I guess this will be complicated to fix, as if we do want to allow hibernation, then we also need to make sure to wake the DO back up when the container shuts down. I think we'd actually have to lie to the DO, claiming that the container is still running, up until the point the DO calls `monitor()`, at which point we would immediately inform it of what actually happened. Ugh, kind of complicated. (Again, not for this PR.)

kentonv (changes requested):
I don't see any tests in this PR nor a PR to the internal codebase modifying the test there. Where is the new functionality tested?

src/workerd/api/container.c++
jasnell: > ... Adding a task to the IoContext here will prevent the DO from hibernating,...

FWIW I've seen this pop up a few times over the past few months. Are the things that prevent hibernation documented anywhere? If not we likely should. Not needed in this PR, tho, of course.

kentonv: Any tasks added by `addTask()` or `addWaitUntil()` (which are synonymous in actors) will block hibernation.

src/workerd/api/container.h
anonrig: You're right. Added.

src/workerd/api/actor-state.c++
anonrig: Updated. Called from the Container's constructor

src/workerd/api/container.c++
anonrig: Removed duplication.

src/workerd/api/container.c++
anonrig: Called reset() before new monitoring cycle.

src/workerd/api/container.c++
anonrig: Adding a test...

src/workerd/api/container.c++
anonrig: Seems interesting. I'll try to tackle this problem here

src/workerd/api/container.h
```diff
@@ -93,8 +93,54 @@ class Container: public jsg::Object {
...
+  kj::Maybe<IoOwn<kj::ForkedPromise<uint8_t>>> monitorKjPromise;
+
+  // The memoized JS promise returned by monitor(). Allows multiple calls to monitor()
+  // to return the same promise.
+  kj::Maybe<jsg::MemoizedIdentity<jsg::Promise<void>>> monitorJsPromise;
+
+  // The background monitor promise, held to keep it alive. Uses eagerlyEvaluate()
+  // to ensure it runs to completion. Held here (rather than via addTask()) to avoid
+  // preventing DO hibernation.
+  kj::Maybe<IoOwn<kj::Promise<void>>> backgroundMonitor;
```
jasnell, line 135: Eventually, it might be nice to bundle all the `IoOwn`s into a struct so that there's just *one* `IoOwn` in use. Non-blocking for this tho.

src/workerd/api/container.c++
```diff
@@ -59,8 +64,62 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+void Container::monitorOnBackgroundIfNeeded() {
+  if (backgroundMonitor != kj::none || !running) return;
+
+  auto& ioContext = IoContext::current();
+
+  // Create the shared forked monitor promise. Both the background monitor and the
+  // JS monitor() method will get branches from this single RPC request.
+  auto forked = rpcClient->monitorRequest(capnp::MessageSize{4, 0})
+                    .send()
+                    .then([](auto&& results) -> uint8_t {
```
jasnell: **[HIGH]** The capnp schema defines `monitor()` as returning `exitCode: Int32`, but this truncates to `uint8_t` — silently losing information for exit codes outside 0–255. While Unix convention uses 0–255, the schema intentionally chose `Int32`.

This type mismatch propagates to several locations:
- `ForkedPromise<uint8_t>` (here and `container.h` `monitorKjPromise` declaration)
- `MonitorState::exitCode` (`container.h`)
- `handleExitCode` lambda parameter (line 176)
- Success branch lambda parameter (line 104)

```suggestion
                    .then([](auto&& results) -> int32_t {
```

The same `uint8_t` → `int32_t` change is needed in the other locations listed above.

src/workerd/api/container.c++
```diff
@@ -59,8 +64,62 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+
+  // Reset monitoring state for the new run. Any stale promises from the previous
+  // run are discarded so that a subsequent monitor() call gets fresh results.
+  monitorKjPromise = kj::none;
+  monitorJsPromise = kj::none;
+  backgroundMonitor = kj::none;
+  monitorState->monitoringExplicitly = false;
+  monitorState->finished = false;
+  monitorState->exitCode = 0;
+  monitorState->exception = kj::none;
```
jasnell: **[HIGH]** Potential stale data race on restart: the old background monitor continuation holds `kj::addRef(*monitorState)` — a reference to this *same* `MonitorState` object. If the old monitor RPC completes after `start()` resets these fields, the old continuation's `.then()` writes `finished=true` with stale exit code/error data, corrupting the new run's monitoring state.

Creating a fresh `MonitorState` instead of resetting fields in-place would make the old continuation write to an orphaned object:

```suggestion
  monitorState = IoContext::current().addObject(kj::refcounted<MonitorState>());
```

src/workerd/api/container.c++
```diff
@@ -59,8 +64,62 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
...
+    // Store the error so monitor() can return immediately without awaitIo().
+    state->finished = true;
+    state->exception = kj::cp(error);
+
+    // The container exited with an error. If the user hasn't called monitor()
+    // explicitly, abort the IoContext so the DO is notified of the failure.
+    if (!state->monitoringExplicitly) {
+      weakIoContext->runIfAlive([&error](IoContext& ctx) { ctx.abort(kj::mv(error)); });
+    }
+  }).eagerlyEvaluate(nullptr);
```
jasnell: **[MEDIUM]** `eagerlyEvaluate(nullptr)` silently swallows any exception that escapes the error handler above. If there's an unexpected exception type or a bug in the handler, it disappears without a trace.

```suggestion
  }).eagerlyEvaluate([](kj::Exception&& e) { KJ_LOG(ERROR, "unexpected error in container background monitor", e); });
```

src/workerd/api/container.c++
```diff
@@ -106,14 +165,24 @@ jsg::Promise<void> Container::interceptAllOutboundHttp(jsg::Lock& js, jsg::Ref<F
...
 
-  return IoContext::current()
-      .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).send())
-      .then(js, [this](jsg::Lock& js, capnp::Response<rpc::Container::MonitorResults> results) {
+  monitorState->monitoringExplicitly = true;
+
+  KJ_IF_SOME(memoized, monitorJsPromise) {
+    return memoized;
+  }
+
```
jasnell, line 208: **[MEDIUM]** Both `handleExitCode` and `handleError` capture `this` (the `Container*`). This is safe because the resulting promise is stored in `monitorJsPromise`, which is traced by `visitForGc()`, preventing the Container from being GC'd while the promise is alive — but that invariant is non-obvious and easy to break in future refactors. A brief comment would help:

```suggestion
  // Safe to capture `this`: the resulting promise is stored in monitorJsPromise,
  // which is traced by visitForGc(), preventing GC while the promise is alive.
  auto handleExitCode = [this](jsg::Lock& js, uint8_t exitCode) {
```

src/workerd/server/container-client.c++
```diff
@@ -679,32 +745,35 @@ kj::Promise<void> ContainerClient::createSidecarContainer(
...
   uint32_t idx = 0;
   cmd.set(idx++, "--http-egress-port");
   cmd.set(idx++, kj::str(egressPort));
   cmd.set(idx++, "--docker-gateway-cidr");
   cmd.set(idx++, networkCidr);
-  if (!ipv6Enabled) {
-    cmd.set(idx++, "--disable-ipv6");
-  }
+  cmd.set(idx++, "--address");
+  cmd.set(idx++, address);
```
gabivlj, line 757: this is going to break in some MacOS docker desktops as they dont have v6 enabled altogether. Thats why I decided against creating a network ourselves as it's complicated to accommodate all use-cases,

src/workerd/api/container.h
```diff
@@ -138,6 +134,12 @@ class Container: public jsg::Object {
   // preventing DO hibernation.
   kj::Maybe<IoOwn<kj::Promise<void>>> backgroundMonitor;
 
+  // Promise that resolves when the current start() RPC completes. The background
+  // monitor chains after this to avoid racing with the start RPC (which creates/
+  // starts the container). Without this, the monitor RPC's Docker /wait could see
+  // a stale container from a previous run and return its old exit code.
```
gabivlj, line 140: I think this is an impl. detail of the container service. I think it should be up to the container capnp service to hold the monitor inflight if the start is already inflight.

src/workerd/server/container-client.c++
```diff
@@ -365,48 +390,82 @@ class EgressHttpService final: public kj::HttpService {
...
-      auto config = ipamConfig[0];
+    kj::Maybe<kj::String> v4Gateway;
+    kj::Maybe<kj::String> v4Subnet;
+    kj::Maybe<kj::String> v6Gateway;
+    kj::Maybe<kj::String> v6Subnet;
+
+    for (auto config: ipamConfig) {
+      auto subnet = kj::str(config.getSubnet());
+      auto gateway = kj::str(config.getGateway());
+      if (subnet.findFirst(':') != kj::none) {
```
jasnell, line 442: Nit: consider checking the `findFirst` first, then doing to allocation/copy for the `subnet` and `gateway`

src/workerd/server/container-client.c++
```diff
@@ -486,7 +545,7 @@ kj::Promise<ContainerClient::InspectResponse> ContainerClient::inspectContainer(
   // We check if the container with the given name exist, and if it's not,
   // we simply return false while avoiding an unnecessary error.
   if (response.statusCode == 404) {
-    co_return InspectResponse{.isRunning = false, .ports = {}};
+    co_return InspectResponse{.isRunning = false, .ports = {}, .hostsPath = kj::none};
```
jasnell, line 548:
```suggestion
    co_return InspectResponse{
      .isRunning = false,
      .ports = {},
      .hostsPath = kj::none,
    };
```

src/workerd/server/container-client.c++
```diff
@@ -531,13 +590,18 @@ kj::Promise<ContainerClient::InspectResponse> ContainerClient::inspectContainer(
   // perspective, a restarting container is still "alive" and should be treated as running
   // so that start() correctly refuses to start a duplicate and destroy() can clean it up.
   bool running = status == "running" || status == "restarting";
-  co_return InspectResponse{.isRunning = running, .ports = kj::mv(portMappings)};
+  auto hostsPath = getHostsPathFromInspectResponse(response.body);
+
+  co_return InspectResponse{
+    .isRunning = running, .ports = kj::mv(portMappings), .hostsPath = kj::mv(hostsPath)};
```
jasnell, line 596:
```suggestion
    .isRunning = running,
    .ports = kj::mv(portMappings),
    .hostsPath = kj::mv(hostsPath),
  };
```

src/workerd/api/container.c++
```diff
@@ -58,9 +63,96 @@ void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions)
 
   req.setCompatibilityFlags(flags);
 
-  IoContext::current().addTask(req.sendIgnoringResult());
+  auto& ioContext = IoContext::current();
+
+  // Store the start promise so monitorOnBackgroundIfNeeded() can chain after it.
+  // This prevents the monitor RPC from racing ahead of the start RPC — without
```
kentonv, line 69: This seems like the container service's responsibility. Cap'n Proto RPC maintains ordering, so the service at the other end should be able to recognize when start() was recently called and wait for it if needed before proceeding with monitoring. Waiting on the client end forces an extra round trip and creates more opportunities for race conditions.

jasnell:
@anonrig ... what's the status on this one?

PR #4374, opened 2025-06-18
containers: allow setting the exit code from a container runtime, and throw an error instead from jsg Container

gabivlj:
Allows the user doing:
```javascript
this.ctx.container.monitor().catch(err => console.log(err.exitCode)); // exitCode will be defined if the error is due to the container exiting
```

### DIFF
--- a/src/workerd/api/container.c++
+++ b/src/workerd/api/container.c++
@@ -56,13 +56,20 @@ jsg::Promise<void> Container::monitor(jsg::Lock& js) {
   JSG_REQUIRE(running, Error, "monitor() cannot be called on a container that is not running.");
 
   return IoContext::current()
-      .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).sendIgnoringResult())
-      .then(js, [this](jsg::Lock& js) {
+      .awaitIo(js, rpcClient->monitorRequest(capnp::MessageSize{4, 0}).send())
+      .then(js, [this](jsg::Lock& js, capnp::Response<rpc::Container::MonitorResults> results) {
     running = false;
+    auto exitCode = results.getExitCode();
     KJ_IF_SOME(d, destroyReason) {
       jsg::Value error = kj::mv(d);
       destroyReason = kj::none;
       js.throwException(kj::mv(error));
+    }
+
+    if (exitCode != 0) {
+      auto err = js.error(kj::str("Container exited with unexpected exit code: ", exitCode));
+      KJ_ASSERT_NONNULL(err.tryCast<jsg::JsObject>()).set(js, "exitCode", js.num(exitCode));
+      js.throwException(err);
     }
   }, [this](jsg::Lock& js, jsg::Value&& error) {
     running = false;
--- a/src/workerd/server/tests/container-client/test.js
+++ b/src/workerd/server/tests/container-client/test.js
@@ -3,6 +3,29 @@ import assert from 'node:assert';
 import { scheduler } from 'node:timers/promises';
 
 export class DurableObjectExample extends DurableObject {
+  async testExitCode() {
+    const container = this.ctx.container;
+    if (container.running) {
+      let monitor = container.monitor().catch((_err) => {});
+      await container.destroy();
+      await monitor;
+    }
+    assert.strictEqual(container.running, false);
+
+    // Start container with valid configuration
+    await container.start({
+      entrypoint: ['node', 'nonexistant.js'],
+    });
+
+    let exitCode = undefined;
+    await container.monitor().catch((err) => {
+      exitCode = err.exitCode;
+    });
+
+    assert.strictEqual(typeof exitCode, 'number');
+    assert.notEqual(0, exitCode);
+  }
+
   async testBasics() {
     const container = this.ctx.container;
     if (container.running) {
@@ -194,6 +217,15 @@ export const testBasics = {
       const stub = CONTAINER.get(id);
       await stub.testBasics();
     }
+  },
+};
+
+// Test exit code monitor functionality
+export const testExitCode = {
+  async test(_ctrl, env) {
+    const id = env.CONTAINER.idFromName('testExitCode');
+    const stub = env.CONTAINER.get(id);
+    await stub.testExitCode();
   },
 };
 
--- a/src/workerd/io/container.capnp
+++ b/src/workerd/io/container.capnp
@@ -36,7 +36,7 @@ interface Container @0x9aaceefc06523bca {
     # The container runtime should validate the environment variables input.
   }
 
-  monitor @2 ();
+  monitor @2 () -> (exitCode: Int32);
   # Waits for the container to shut down.
   #
   # If the container shuts down because the root process exited with a success status, or because
--- a/src/workerd/server/container-client.c++
+++ b/src/workerd/server/container-client.c++
@@ -362,6 +362,7 @@ kj::Promise<void> ContainerClient::monitor(MonitorContext context) {
   // Monitor is often called right after start but the api layer's start does not await the RPC's
   // start response. That means that the createContainer call might not have even started yet.
   // If it hasn't, we'll give it 3 tries before failing.
+  auto results = context.getResults();
   for (int i = 0; i < 3; i++) {
     // Docker API: POST /containers/{id}/wait - wait for container to exit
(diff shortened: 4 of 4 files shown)


src/workerd/server/tests/container-client/test.js
```diff
@@ -3,6 +3,29 @@ import assert from 'node:assert';
...
+    await container.start({
+      entrypoint: ['node', 'nonexistant.js'],
+    });
+
+    let exitCode = undefined;
+    await container.monitor().catch((err) => {
+      exitCode = err.exitCode;
+    });
+
+    assert.strictEqual(typeof exitCode === 'number', true);
```
anonrig:
```suggestion
    assert.strictEqual(typeof exitCode, 'number');
```

src/workerd/server/tests/container-client/test.js
```diff
@@ -197,6 +220,17 @@ export const testBasics = {
   },
 };
 
+// Test exit code monitor functionality
+export const testExitCode = {
+  async test(_ctrl, env) {
+    for (const CONTAINER of [env.MY_CONTAINER, env.MY_DUPLICATE_CONTAINER]) {
```
anonrig: I think it's ok to run this only on env.MY_CONTAINER. The test already takes 35 seconds and redundant tests significantly increases the test duration.

src/workerd/api/container.c++
```diff
@@ -9,6 +9,15 @@
 
 namespace workerd::api {
 
+void setExitCodeError(jsg::Lock& js, jsg::JsValue& a, int exitCode) {
```
anonrig: Recommend a better variable naming for this

```suggestion
void setExitCodeError(jsg::Lock& js, jsg::JsValue& error, int exitCode) {
```

kentonv:
I don't think I agree with the design of attaching the exit code to the exception. This is going down the path of making exceptions be part of the API, which JavaScript generally avoids -- most JS exceptions aren't really designed to be caught, except for the purpose of logging or displaying the error.

In chat I had suggested a different approach, which is that the application could opt into getting the exit code as a return instead of an exception, like:

```
let exitCode = await ctx.container.monitor({onNonZeroExit: "return"});
```

I'm not attached to this particular API, but we should come up with some API that doesn't involve catching an exception. (But the default behavior should still be an exception -- this has to be opt-in.)

gabivlj:
> I don't think I agree with the design of attaching the exit code to the exception. This is going down the path of making exceptions be part of the API, which JavaScript generally avoids -- most JS exceptions aren't really designed to be caught, except for the purpose of logging or displaying the error.
> 
> In chat I had suggested a different approach, which is that the application could opt into getting the exit code as a return instead of an exception, like:
> 
> ```
> let exitCode = await ctx.container.monitor({onNonZeroExit: "return"});
> ```
> 
> I'm not attached to this particular API, but we should come up with some API that doesn't involve catching an exception. (But the default behavior should still be an exception -- this has to be opt-in.)

I agree that most exceptions in JS should be designed to just rendering them. I also agree that if the user wants to handle the exit code as a value they could opt-in to do so.
However, I think it's not a bad idea to include the exit code on the exception, as some users could still consider their container exiting in that path as an "error", and they might want to know if it was because of a non-zero exit code error. In that case having the property "exitCode" is really useful.
For example, in NodeJS "child_process" includes the exit code on the exception when your process exits with a non-zero for ease of retrieving it (it's one of the reasons I added this).
I will also say that I don't have a strong opinion on including the value on the exception, as we can just tell them "to opt for marking non-zero as non-error to get it", but maybe they want to retrieve the exit code that caused an error easily without doing some string pattern matching.

gabivlj:
> > I don't think I agree with the design of attaching the exit code to the exception. This is going down the path of making exceptions be part of the API, which JavaScript generally avoids -- most JS exceptions aren't really designed to be caught, except for the purpose of logging or displaying the error.
> > In chat I had suggested a different approach, which is that the application could opt into getting the exit code as a return instead of an exception, like:
> > ```
> > let exitCode = await ctx.container.monitor({onNonZeroExit: "return"});
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not attached to this particular API, but we should come up with some API that doesn't involve catching an exception. (But the default behavior should still be an exception -- this has to be opt-in.)
> 
> I agree that most exceptions in JS should be designed to just rendering them. I also agree that if the user wants to handle the exit code as a value they could opt-in to do so. However, I think it's not a bad idea to include the exit code on the exception, as some users could still consider their container exiting in that path as an "error", and they might want to know if it was because of a non-zero exit code error. In that case having the property "exitCode" is really useful. For example, in NodeJS "child_process" includes the exit code on the exception when your process exits with a non-zero for ease of retrieving it (it's one of the reasons I added this). I will also say that I don't have a strong opinion on including the value on the exception, as we can just tell them "to opt for marking non-zero as non-error to get it", but maybe they want to retrieve the exit code that caused an error easily without doing some string pattern matching.

Another example as well of exec() APIs having this as default behaviour: https://bun.sh/docs/runtime/shell#error-handling
I think we should try to stay close to these (and potentially doing the same for  `ctx.container.exec()`?).

mikenomitch:
My two cents (with less context than either of you): Feels like if a user wants the errorCode, not having them change the way they call the methods/error handle is nice. And as is the alignment with nodejs conventions. And downside of adding to the error we already throw seems very low.

kentonv:
> For example, in NodeJS "child_process" includes the exit code on the exception when your process exits with a non-zero for ease of retrieving it (it's one of the reasons I added this).

A wide range of Node APIs include a `.code` property. The usage is actually inconsistent: in most cases it is a string name of an errno constant, like "ENOENT", but in the case of `child_prcoess` it is apparently a numeric exit code, which seems weird and inconsistent. But in any case, the `.code` property is a broad convention in node APIs, but doesn't appear in web APIs, and we generally follow web API design for our non-standard APIs. Also, the property you've suggested here is `exitCode`, so it really isn't following the Node convention anyway.

Yes, Bun also has their own non-standard APIs but I don't think we take design cues from Bun. We look to web APIs.

With all that said, I don't necessarily have a problem with adding properties to an error object in order to describe the error so that it can be more cleanly reported. But what I do not want is to force people to write code that catches exceptions as part of normal operation. If people are returning a particular exit code not to signal an error, but rather to signal some instruction to the DO about what it should do next, that needs to be communicated in a way that doesn't involve throwing exceptions. (But at the same time, if the calling code does not intend to handle error codes explicitly, we need to throw an exception, so that errors aren't silently dropped.)

src/workerd/api/container.c++
```diff
@@ -56,14 +65,21 @@ jsg::Promise<void> Container::monitor(jsg::Lock& js) {
...
+    auto exitCode = results.getExitCode();
     KJ_IF_SOME(d, destroyReason) {
       jsg::Value error = kj::mv(d);
       destroyReason = kj::none;
       js.throwException(kj::mv(error));
     }
+
+    if (exitCode != 0) {
+      auto err = js.error(kj::str("Container exited with unexpected exit code: ", exitCode));
+      setExitCodeError(js, err, exitCode);
```
kentonv: I think this can be replaced with:

```suggestion
      KJ_ASSERT_NONNULL(err.tryCast<jsg::JsObject>()).set("exitCode", js.num(exitCode));
```

(I'm pretty sure `jsg.error()` _always_ returns an object, but unfortunately the underlying V8 APIs are declared only as returning `Value`.)

gabivlj: thanks, I really complicated setting a key value on this thing for no reason :sweat_smile:

src/workerd/server/tests/container-client/test.js
```diff
@@ -197,6 +220,15 @@ export const testBasics = {
   },
 };
 
+// Test exit code monitor functionality
+export const testExitCode = {
+  async test(_ctrl, env) {
+    const id = env.CONTAINER.idFromName('testExitCode');
+    const stub = env.CONTAINER.get(id);
```
anonrig, line 227: This is actually wrong. The test should be as following.

env.MY_CONTAINER rather than env.CONTAINER

kentonv: We really need to actually run this test in CI. Can we set it up with a mock backend?

PR #4375, opened 2025-06-18
remove unnecessary eslint rules and enable on more files

anonrig:
Removes unnecessary eslint rule that forces to add "public" qualifier to all methods in classes. This is unnecessary and just a typescript sugar. Doesn't make any difference to add public or not, and just bloats the code.

I also enabled eslint on couple of more files with these changes.


dario-piotrowicz (approved):
lgtm 🙂

shrima-cf (approved):

PR #4377, opened 2025-06-19
suppress unnecessary microsoft-unqualified-friend warning

anonrig:
This suppresses the following warning on Windows

```
bazel-out/x64_windows-opt/bin/src/workerd/io/_virtual_includes/io\workerd/api\blob.h(115,16): warning: unqualified friend declaration referring to type outside of the nearest enclosing namespace is a Microsoft extension; add a nested name specifier [-Wmicrosoft-unqualified-friend]
  115 |   friend class File;
      |                ^
      |                ::workerd::
1 warning generated.
```


kentonv:
Wouldn't it have been easier to fix the warning by adding the qualifier?

PR #4378, opened 2025-06-19
Rename KV span attributes to match new schema

jmorrell-cloudflare:
This renames all the existing KV span attributes to match the new schema before we start the work of expanding them.

Note that we are not bringing in the binding name for the span name here yet, but I will update https://github.com/cloudflare/workerd/pull/4372 to do that for all bindings


fhanau (approved):
LGTM. Lmk if you need help with updating the downstream test

PR #4383, opened 2025-06-20
Experimental feature: Dynamic worker loading

kentonv:
This is an experimental-only feature and will probably remain so until we play around with it a bit. Similar to facets (#4123), the interface is likely to change based on testing. This is also strictly an implementation in workerd -- the production runtime would need a very different implementation.

A Worker loader binding allows you to start an isolate from code provided at runtime, and then talk to that isolate.

```js
let id = "foo";
let worker = env.LOADER.get(id, () => {
  // This callback is called if the given worker ID is not loaded yet.
  // (If a Worker with the same ID is already loaded by this loader, the
  // existing worker is returned without calling the callback.)

  // The callback returns the Worker's code. It can be async.
  return {
    compatibilityDate: "2025-06-01",

    // Specify the main module, whose exports are the entrypoints.
    mainModule: "foo.js",

    // Specify a map of all modules. This is an object mapping module
    // names to source code.
    modules: {
      // Normal ES modules are specified just as strings.
      "foo.js":
        "export default {\n" +
        "  fetch(req, env, ctx) { return new Response('Hello'); }\n" +
        "}\n",
      // Non-ES module types are supported too, by specifying an object
      // with one property that indicates the type.
      "some.wasm": { wasm: WASM_DATA },
    },

    // The `env` is specified as an arbitrary object. This gets serialized and
    // passed into the Worker.
    env: {
      SOME_ENV_VAR: 123
    },

    // You can override the global outbound. When the worker calls global
    // `fetch()` (or `connect()`), it will be directed to this service binding.
    globalOutbound: env.MY_OUTBOUND_WORKER,
  };
});

// Now you can get its entrypoint.
let defaultEntrypoint = worker.getEntrypoint();
await defaultEntrypoint.fetch(...);

// You can get non-default entrypoints as well.
let someEntrypoint = worker.getEntrypoint("SomeEntrypointClass");

// You can also get DO classes. These can be used with facets (#4123).
let someClass = worker.getDurableObjectClass("SomeDoClass");
```

[The test](https://github.com/cloudflare/workerd/pull/4383/files#diff-c49794f41387a0c828c7ccbcc621b25f6e4346b8a1346bec6a8d3ca88dd71958) has more examples.

TODO (this PR):
* Add a bunch more tests.

TODO (later PRs, probably):
* Eviction if idle workers.
* Extend `env` so you can actually provide resource bindings. Currently it can only contain simple serializable values.
* Support hooking up tails and cache API.

### DIFF
--- a/src/workerd/api/actor-state.c++
+++ b/src/workerd/api/actor-state.c++
@@ -823,12 +823,12 @@ void DurableObjectTransaction::maybeRollback() {
 class FacetOutgoingFactory final: public Fetcher::OutgoingFactory {
  public:
   FacetOutgoingFactory(Worker::Actor::FacetManager& facetManager,
-      uint actorClassChannel,
+      kj::Own<IoChannelFactory::ActorClassChannel> actorClass,
       kj::String name,
       Worker::Actor::Id id)
       : facetManager(facetManager),
-        actorClassChannel(actorClassChannel),
         name(kj::mv(name)),
+        actorClass(kj::mv(actorClass)),
         id(kj::mv(id)) {}
 
   kj::Own<WorkerInterface> newSingleUseClient(kj::Maybe<kj::String> cfStr) override {
@@ -844,7 +844,7 @@ class FacetOutgoingFactory final: public Fetcher::OutgoingFactory {
       if (actorChannel == kj::none) {
         actorChannel = facetManager.getFacet(name,
             {
-              .actorClassChannel = actorClassChannel,
+              .actorClass = kj::mv(actorClass),
               .id = kj::mv(id),
             });
       }
@@ -859,9 +859,11 @@ class FacetOutgoingFactory final: public Fetcher::OutgoingFactory {
 
  private:
   Worker::Actor::FacetManager& facetManager;
-  uint actorClassChannel;
   kj::String name;
-  Worker::Actor::Id id;  // moved away when `actorChannel` is initialized
+
+  // These are moved away when `actorChannel` is initialized.
+  kj::Own<IoChannelFactory::ActorClassChannel> actorClass;
+  Worker::Actor::Id id;
 
   kj::Maybe<kj::Own<IoChannelFactory::ActorChannel>> actorChannel;
 };
@@ -885,8 +887,10 @@ jsg::Ref<Fetcher> DurableObjectFacets::get(jsg::Lock& js, kj::String name, GetOp
     id = ioCtx.getActorOrThrow().cloneId();
   }
 
+  auto actorClass = ioCtx.getIoChannelFactory().getActorClass(options.$class->getChannel());
+
   kj::Own<Fetcher::OutgoingFactory> factory =
-      kj::heap<FacetOutgoingFactory>(fm, options.$class->getChannel(), kj::mv(name), kj::mv(id));
+      kj::heap<FacetOutgoingFactory>(fm, kj::mv(actorClass), kj::mv(name), kj::mv(id));
 
   auto requiresHost = FeatureFlags::get(js).getDurableObjectFetchRequiresSchemeAuthority()
       ? Fetcher::RequiresHostAndProtocol::YES
--- a/src/workerd/api/actor-state.c++
+++ b/src/workerd/api/actor-state.c++
@@ -887,7 +887,7 @@ jsg::Ref<Fetcher> DurableObjectFacets::get(jsg::Lock& js, kj::String name, GetOp
     id = ioCtx.getActorOrThrow().cloneId();
   }
 
-  auto actorClass = ioCtx.getIoChannelFactory().getActorClass(options.$class->getChannel());
+  auto actorClass = options.$class->getChannel(ioCtx);
 
   kj::Own<Fetcher::OutgoingFactory> factory =
       kj::heap<FacetOutgoingFactory>(fm, kj::mv(actorClass), kj::mv(name), kj::mv(id));
--- a/src/workerd/api/worker-loader-test.js
+++ b/src/workerd/api/worker-loader-test.js
@@ -137,21 +137,31 @@ export let facets = {
 // Test isolate uniqueness when loading the same name multiple times.
 export let isolateUniqueness = {
   async test(ctrl, env, ctx) {
-    let workerCode = {
-      compatibilityDate: '2025-01-01',
-      mainModule: 'foo.js',
-      modules: {
-        'foo.js': `
-          import {WorkerEntrypoint} from "cloudflare:workers";
-          let i = 0;
-          export default class extends WorkerEntrypoint {
-            increment() {
-              return i++;
+    let loadCount = 0;
+    let loadCodeCallback = () => {
+      ++loadCount;
+      return {
+        compatibilityDate: '2025-01-01',
+        mainModule: 'foo.js',
+        modules: {
+          'foo.js': `
+            import {WorkerEntrypoint} from "cloudflare:workers";
+            let i = 0;
+            export default class extends WorkerEntrypoint {
+              increment() {
+                return i++;
+              }
             }
-          }
-        `,
-      },
+          `,
+        },
+      };
     };
+
+    let assertNotCalled = () => {
+      throw new Error(
+          "Code loader callback called when it should have reused an existing isolate.");
+    };
+
     let name = 'isolateUniqueness';
     let name2 = 'isolateUniquenessOtherName';
 
@@ -159,17 +169,17 @@ export let isolateUniqueness = {
     // - Two loaders with the same ID.
     // - One with a unique ID.
     // - One with no ID (anonymous).
-    let shared1 = env.sharedLoader1.get(name, () => workerCode).getEntrypoint();
-    let shared2 = env.sharedLoader2.get(name, () => workerCode).getEntrypoint();
-    let unique = env.uniqueLoader.get(name, () => workerCode).getEntrypoint();
-    let anonymous = env.loader.get(name, () => workerCode).getEntrypoint();
+    let shared1 = env.sharedLoader1.get(name, loadCodeCallback).getEntrypoint();
+    let shared2 = env.sharedLoader2.get(name, assertNotCalled).getEntrypoint();
+    let unique = env.uniqueLoader.get(name, loadCodeCallback).getEntrypoint();
+    let anonymous = env.loader.get(name, loadCodeCallback).getEntrypoint();
 
     // Also try loading a different name from various loaders.
     let anonymousOtherName = env.loader
-      .get(name2, () => workerCode)
+      .get(name2, loadCodeCallback)
       .getEntrypoint();
     let sharedOtherName = env.sharedLoader1
-      .get(name2, () => workerCode)
+      .get(name2, loadCodeCallback)
       .getEntrypoint();
 
     // Same name from loaders with the same ID should go to the same isolate.
@@ -191,8 +201,10 @@ export let isolateUniqueness = {
     assert.strictEqual(await sharedOtherName.increment(), 1);
 
     // Load the same name from the same loader -> same isolate.
-    let anonymousAgain = env.loader.get(name, () => workerCode).getEntrypoint();
+    let anonymousAgain = env.loader.get(name, assertNotCalled).getEntrypoint();
     assert.strictEqual(await anonymousAgain.increment(), 2);
     assert.strictEqual(await anonymousAgain.increment(), 3);
+
+    assert.strictEqual(loadCount, 5);
   },
 };
--- a/src/workerd/api/worker-loader-test.js
+++ b/src/workerd/api/worker-loader-test.js
@@ -57,3 +57,36 @@ export let passEnv = {
     assert.strictEqual(await resp.text(), 'env.hello = 123');
   },
 };
+
+export let testOutbound = {
+  async fetch(req, env, ctx) {
+    return new Response('hello from testOutbound');
+  },
+};
+
+// Test overriding globalOutbound
+export let overrideGlobalOutbound = {
+  async test(ctrl, env, ctx) {
+    let worker = env.loader.get('overrideGlobalOutbound', () => {
+      return {
+        compatibilityDate: '2025-01-01',
+        mainModule: 'foo.js',
+        modules: {
+          'foo.js': `
+            export default {
+              fetch(req, env, ctx) {
+                return fetch(req);
+              },
+            }
+          `,
+        },
+
+        // Set globalOutbound to redirect to our own `testOutbound` entrypoint.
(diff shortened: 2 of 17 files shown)


kentonv:
FYI @irvinebroque @threepointone @rohinlohe

jasnell:
Definitely happy to see this progressing but I definitely think it's rather unfortunate for us to have introduce a bespoke syntax/API for this. It would be nice if there were alignment here with either/both Node.js worker threads and web workers API.

kentonv:
> It would be nice if there were alignment here with either/both Node.js worker threads and web workers API.

The API design here is intended to directly expose how our platform works. We could consider keeping it private, while building compatibility wrappers for specific standard APIs.

I suspect, though, that there will be a lot of friction trying to apply various standard APIs that were designed for different use cases. Both Node worker threads and Web Workers are primarily intended to allow parallelization, offloading work to a background thread. The dynamic worker loader API in this PR explicitly does _not_ do that -- when you call into the worker, it runs in the _same_thread. Instead, our feature is more about sandboxing and dynamic code loading.

Moreover, it's an explicit design goal that the dynamic worker loader should be able to load any real Worker code that you could otherwise deploy with wrangler, etc. That pretty much necessitates that our API include concepts like bindings and compat dates, which obviously don't exist in any standard worker-loading API.

irvinebroque:
> it runs in the _same_thread

Wonder if there is a mode where it shouldn't? What if I'm building a platform?

src/workerd/server/workerd.capnp
```diff
@@ -419,6 +419,20 @@ struct Worker {
         limits @25 :MemoryCacheLimits;
       }
 
+      workerLoader :group {
+        # A binding representing the ability to dynamically load Workers from code presented at
+        # runtime.
+        #
+        # An Worker loader is not just a function that loads a Worker, but also serves as a
```
jasnell:
```suggestion
        # A Worker loader is not just a function that loads a Worker, but also serves as a
```

src/workerd/server/workerd.capnp
```diff
@@ -419,6 +419,20 @@ struct Worker {
...
+        # cache of Workers, automatically unloading Workers that are not in use. To that end, each
+        # Worker must have a name, and if a Worker with that name already exists, it'll be reused.
+
+        id @27 :Text;
+        # The identifier associated with this Worker loader. Multiple Workers can bind to the
+        # same ID in order to access the same loader, so that if they request the same name from
+        # it, they'll end up sharing the same loaded Worker.
```
jasnell: Hmm, is this something we also would expect in production or just here in workerd? I'm hesitant to introduce another shared resource as it could end up being used for sharing state.

src/workerd/server/server.c++
```diff
@@ -227,8 +227,17 @@ class Server::Service {
...
+class Server::ActorClass: public IoChannelFactory::ActorClassChannel {
  public:
+  // The caller must call this before calling newActor(). If it returns a promise, then the
+  // caller must await the promise before calling other methods.
+  //
+  // In particular, this is needed with dynamically-loaded workers. The isolate may still be
+  // loading when the caller calls `getDurableObjectClass()` on it.
+  virtual kj::Maybe<kj::Promise<void>> whenReady() {
+    return kj::none;
+  }
```
jasnell, line 239: Nit: Not a big fan of maybe-its-a-promise-maybe-its-not type returns. Would it be possible just to have this always return a promise and have it return `kj::READY_NOW`?

src/workerd/server/server.c++
```diff
@@ -986,7 +992,7 @@ kj::Own<Server::Service> Server::makeNetworkService(config::Network::Reader conf
     tlsNetwork = ownedTlsContext->wrapNetwork(*restrictedNetwork).attach(kj::mv(ownedTlsContext));
   }
 
-  return kj::heap<NetworkService>(globalContext->headerTable, timer, entropySource,
+  return kj::refcounted<NetworkService>(globalContext->headerTable, timer, entropySource,
```
jasnell, line 995: Any reason not to use `kj::Rc<...>` and `kj::rc<...>` with this?

src/workerd/api/worker-loader.h
```diff
@@ -0,0 +1,128 @@
...
+ public:
+  // Create a WorkerLoader backed by the given I/O channel.
+  //
+  // `compatDateValidation` will differ between workerd vs. production.
+  explicit WorkerLoader(uint channel, CompatibilityDateValidation compatDateValidation)
+      : channel(channel),
+        compatDateValidation(compatDateValidation) {}
+
+  struct Module {
+    // Exactly one must be filled in.
```
jasnell, line 55: bleh, I understand why you've structured it this way but it still feels yucky not being able to use a kj::OneOf here.

src/workerd/api/worker-loader.h
```diff
@@ -0,0 +1,128 @@
...
+    // `Fetcher` (e.g. service binding) representing the loaded worker's global outbound.
+    //
+    // If omitted, inherit the current worker's global outbound.
+    //
+    // If `null`, block the global outbound (all requests throw errors).
+    jsg::Optional<kj::Maybe<jsg::Ref<Fetcher>>> globalOutbound;
+
+    // TODO(someday): cache API outbound?
+
+    // TODO(someday): Tails?
```
jasnell: I'm assuming this means tail workers for observability? If so, it would be worth expanding the comment just a bit to make it clearer

src/workerd/api/worker-loader.c++
```diff
@@ -0,0 +1,219 @@
...
+          .content = Worker::Script::EsModule{.body = text},
+        };
+      }
+      KJ_CASE_ONEOF(module, Module) {
+        uint fieldCount = (module.js != kj::none) + (module.cjs != kj::none) +
+            (module.text != kj::none) + (module.data != kj::none) + (module.json != kj::none);
+        JSG_REQUIRE(fieldCount == 1, TypeError,
+            "Each module must contain exactly one of 'js', 'cjs', 'text', 'data', or 'json'. "
+            "Module '",
+            entry.name, "' contained ", fieldCount, " properties.");
```
jasnell, line 157: Again, not being able to use kj::OneOf here makes me sad.

src/workerd/api/worker-loader-test.js
```diff
@@ -0,0 +1,135 @@
+import { DurableObject } from 'cloudflare:workers';
+import assert from 'node:assert';
+
+export let basics = {
+  async test(ctrl, env, ctx) {
+    let worker = env.loader.get('basics', () => {
```
jasnell, line 6: these tests can be expanded.

* The loader function return value is actually converted into a promise internally but these are all sync functions. Having a test variation that explicitly returns a promise that awaits some i/o (simple timer) would be good so that we're not always testing with synchronously resolved promises.
* None of the current test cases appear to test what happens if the loader callback throws or returns invalid data.
* Having a case where a worker tries loading a worker twice just to verify it works as expected ... e.g. `let w1 = env.loader.get('basics', ...); let w2 = env.loader.get('basics', ...); await Promise.all([w2.getEntrypoint().greet('...'), w1.getEntrypoint().greet('...')]);` (note that flipping the order of the getEntrypoint calls is intentional there).

src/workerd/api/worker-loader.h
```diff
@@ -0,0 +1,128 @@
...
+
+    JSG_STRUCT(compatibilityDate,
+        compatibilityFlags,
+        allowExperimental,
+        mainModule,
+        modules,
+        env,
+        globalOutbound);
```
penalosa, line 103: I'm concerned that this is _yet another_ way to specify a Worker. We already have:
- The API format (`metadata.json`)
- Wrangler's file format (`wrangler.json`), and API format (`StartDevWorkerInput`)
- `workerd.capnp`
- Miniflare's format

which are all similar but different in subtle ways. Do we need a new format here or can we re-use an existing one?

src/workerd/api/worker-loader.c++
```diff
@@ -0,0 +1,219 @@
...
+ public:
+  kj::Own<WorkerInterface> startRequest(IoChannelFactory::SubrequestMetadata metadata) override {
+    JSG_FAIL_REQUIRE(Error,
+        "This worker is not permitted to access the internet via global functions like fetch(). "
+        "It must use capabilities (such as bindings in 'env') to talk to the outside world.");
+  }
+};
+
+jsg::Ref<WorkerStub> WorkerLoader::get(
+    jsg::Lock& js, kj::String name, jsg::Function<jsg::Promise<WorkerCode>()> getCode) {
```
penalosa, line 95: What about providing the `name`/`id` to the `getCode` callback? That'd make it easier to dynamically load a worker from external storage.

src/workerd/server/workerd.capnp
kentonv: This is a lot like the situation with MemoryCache. A WorkerLoader essentially implements an "isolate cache" -- getting the same worker name multiple times rapidly will return the same isolate.

Just like with MemoryCache, it may make sense in some cases for multiple callers to share the same cache. However, the vast majority of the time, they probably won't, and this ID will just be random so as to avoid any sharing. That said, I suppose the id could be optional -- if null, then the cache is not shared at all?

src/workerd/server/server.c++
kentonv: I wanted to avoid spinning the KJ event loop in the common case. It doesn't add too much complexity here.

(Unfortunately KJ cannot always optimize away a `co_await` on an already-resolved promise. In particular, the first `co_await` in a coroutine cannot be optimized away because the application may assume that the remainder of the function runs asynchronously after the first `co_await`.)

src/workerd/server/server.c++
kentonv: Mentioned in a commit message:

> I realize it would be even better to go all the way to `kj::Rc` here, but that would create a lot more noise, also requiring a coordinated internal change, so I'm punting that for the moment.

src/workerd/api/worker-loader.h
kentonv: Yeah unfortunately without this the type of module is not distinguishable from the source code's JS type, since that's always just `string` or `ArrayBuffer`.

I'm open to other ideas here, but I didn't really want to introduce a bunch of classes people have to import from some built-in module.

src/workerd/api/worker-loader.c++
kentonv: You're saying instead of capturing the name into the callback like:

```
env.loader.get(name, () => load(name));
```

You want to be able to write:

```
env.loader.get(name, load);
```

Is it really a big deal, though? I think it would actually perform worse: converting the name back from C++ to JavaScript would be slower than creating the lambda that captures it. (Though the difference is negligible regardless.)

src/workerd/api/worker-loader.h
kentonv: I am not sure any of the other formats would work here.

There's three big components of what's specified here:
* Source code.
* Bindings (and other capabilities, like `globalOutbound`)
* Other "settings", like compat date/flags.

The "settings" part is already identical between all the formats (mostly because it's so trivial, what else would you do?).

For source code, both `metadata.json` and `wrangler.json` expect the code to come in the form of separate files rather than inline strings. `wrangler.json` doesn't even explicitly refer to the files; they are found by walking imports. `workerd.capnp` specifies source code as inline strings (taking advantage of capnp's `embed` directive), but as a Cap'n Proto schema, it is the least JavaScript-like (e.g. it uses a list of name/value pairs rather than a map/dict). It's also not a format that most users interact with directly anyway. So I'm not sure it's worth trying to match.

The biggest difference, though, is bindings. In all other config formats, bindings must be specified as static configuration. In this worker loader API, the calling worker can specify an arbitrary object, which just gets serialized. This object can contain capabilities. This is extremely powerful, and certainly not worth giving up just to conform with some other config format! Moreover, converting binding configuration into live bindings normally requires talking to EWC, which we want to avoid here. So I think trying to match the existing formats just doesn't make sense for bindings.

Given all this, I think we do indeed need to introduce a new format here.

src/workerd/api/worker-loader.h
kentonv: fbea3a06d

src/workerd/server/workerd.capnp
kentonv: 1969b212a

src/workerd/api/actor-state.c++
```diff
@@ -885,8 +887,10 @@ jsg::Ref<Fetcher> DurableObjectFacets::get(jsg::Lock& js, kj::String name, GetOp
     id = ioCtx.getActorOrThrow().cloneId();
   }
 
+  auto actoClass = options.$class->getChannel(ioCtx);
```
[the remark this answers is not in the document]
kentonv: c9e691301

kentonv:
@irvinebroque 

> > it runs in the _same_thread
>
> Wonder if there is a mode where it shouldn't? What if I'm building a platform?

I think there's a lot of room for us to add features to support parallelism, but they are probably orthogonal to this feature...

src/workerd/server/workerd.capnp
kentonv: Added the ability to omit the ID: e6cad27ad

kentonv:
More tests added:
* 8cf3cac3c2ee38761df9c709019887d62e5895d7
* f574ed6ad94076bf9b862218538c5f6abbace926

kentonv:
Fixed cyclic reference bug introduced by anonymous loaders: b4a7f4813

Then rebased on main.

src/workerd/api/worker-loader.c++
```diff
@@ -0,0 +1,219 @@
...
+      return channel->startRequest({.cfBlobJson = kj::mv(cfStr), .tracing = tracing});
+    },
+        {.inHouse = true,
+          .wrapMetrics = true,
+          .operationName = kj::ConstString("dynamic_worker_subrequest"_kjc)}));
+  }
+
+ private:
+  kj::Own<IoChannelFactory::SubrequestChannel> channel;
+};
```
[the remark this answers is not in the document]
kentonv: Wow that's actually a good observation.

However, I plan to add this method to all the various `OutgoingFactory` implementations in a later PR.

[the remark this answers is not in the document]
jasnell: Ok, that one is actually impressive lol

jasnell:
Grr... thanks Github

image

kentonv:
> Grr... thanks Github

That's weird, it lets me load it in the browser (after a click).

jasnell:
I switched to github's new preview UX that they are testing and apparently they reduced the threshold on size of the diff. Switching back to the original UX and it works.

PR #4387, opened 2025-06-20
Fixes regression with SDK imports in Python search path.

dom96:
The recent change to our search path caused an existing worker which had a `workers.py` file in their bundle to fail. This happened because our own code imports `workers.py` (which is also what our SDK entrypoint is), so it attempts to import the user's module instead of our SDK module.

I have changed our code to not move the path where the user's code is before the site-packages path (where the SDK is). This change also now means that if a worker places a module that shares the name with their main module in the vendor directory, it will attempt to import the file in the vendor directory. I think that what we want to do is just fail validation in that case and that's what I implemented here too.

I have implemented a validation test in the parent repo for this too.

### Test Plan

```
$ bazel run @workerd//src/workerd/server/tests/python:vendor_dir_development@
Python sys.path: ['/session', '/lib/python312.zip', '/lib/python3.12', '/lib/python3.12/lib-dynload', '/session/metadata/vendor', '/lib/python3.12/site-packages', '/session/metadata', '/session/lib/python3.12/site-packages']
```

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -421,6 +421,31 @@ kj::Own<PyodideMetadataReader::State> PyodideMetadataReader::State::clone() {
   return kj::heap<PyodideMetadataReader::State>(*this);
 }
 
+void PyodideMetadataReader::State::verifyNoMainModuleInVendor() {
+  // Verify that we don't have module named after the main module in the `vendor` subdir.
+  // mainModule includes the .py extension, so we need to extract the base name
+  kj::ArrayPtr<const char> mainModuleBase = mainModule;
+  if (mainModule.endsWith(".py")) {
+    mainModuleBase = mainModuleBase.slice(0, mainModuleBase.size() - 3);
+  }
+
+  for (auto& name: moduleInfo.names) {
+    if (name.startsWith(kj::str("vendor/", mainModule))) {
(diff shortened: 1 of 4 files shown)


src/workerd/api/pyodide/pyodide.c++
```diff
@@ -421,6 +421,16 @@ kj::Own<PyodideMetadataReader::State> PyodideMetadataReader::State::clone() {
...
 }
 
+void PyodideMetadataReader::State::verifyNoMainModuleInVendor() {
+  // Verify that we don't have module named after the main module in the `vendor` subdir.
+  for (auto& name: moduleInfo.names) {
+    if (name.startsWith(kj::str("vendor/", mainModule))) {
+      JSG_FAIL_REQUIRE(
+          Error, kj::str("Python module vendor/", mainModule, " clashes with main module"));
+    }
+  }
```
hoodmane: If the main module is called `blah.py` they could still have `vendor/blah/__init__.py` and that would also shadow. Or `vendor/blah.so`. I don't know if we care about these cases.

hoodmane (approved):
Generally makes sense, but see my comment.

src/workerd/api/pyodide/pyodide.c++
dom96: Good point, I think we do care about these too.

PR #4394, opened 2025-06-23
add codspeed benchmark runner

anonrig:
Completes the work started by @art049 at https://github.com/cloudflare/workerd/pull/4137


.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,78 @@
...
+          sudo ./llvm.sh 19
+          # keep in sync with build/ci.bazelrc
+          sudo apt-get install -y -qq --no-install-recommends clang-19 lld-19 libunwind-19 libc++abi1-19 libc++1-19 libc++-19-dev libclang-rt-19-dev llvm-19
+          sudo ln -s /usr/bin/llvm-symbolizer-19 /usr/bin/llvm-symbolizer
+          echo "build:linux --action_env=CC=/usr/lib/llvm-19/bin/clang" >> .bazelrc
+          echo "build:linux --host_action_env=CC=/usr/lib/llvm-19/bin/clang" >> .bazelrc
+
+      - name: Build benchmarks
+        run: |
+          bazel build ${{ env.BAZEL_ARGS }} //src/workerd/tests:all_benchmarks
```
danlapid, line 47: This means we need to build another workerd binary for benchmarks, we should add the benchmarks as an extra step on top of one of our already running jobs using its artifacts rather than another full build.
There's existing examples in test.yml for generated types and wpt reports etc.

danlapid:
> ## [CodSpeed Performance Report](https://codspeed.io/cloudflare/workerd/branches/yagiz%2Fenable-codspeed)
> ### Congrats! CodSpeed is installed 🎉
> `🆕` **10 new benchmarks were detected.**
> 
> You will start to see performance impacts in the reports once the benchmarks are run from your default branch.
> 
> ### Detected benchmarks
> * `ParseAndSerialize` (17.7 µs)
> * `Serialize` (4.4 µs)
> * `constructor[ApiHeaders]` (27.9 µs)
> * `SlowAPIWithLock[FastMethodFixture]` (3.7 ms)
> * `SlowAPI[FastMethodFixture]` (4 ms)
> * `request[GlobalScopeBenchmark]` (331 µs)
> * `Test_JSON_DEC` (74.4 µs)
> * `Test_JSON_ENC` (43.7 µs)
> * `Parse[KjHeaders]` (8.8 µs)
> * `request[RegExpBenchmark]` (7.2 ms)

These numbers are likely too small to be used for performance comparisons, I think aiming for 5-10ms per benchmark test is likely a good starting point, microseconds benchmarks likely need more iterations

.github/workflows/codspeed.yml
anonrig: We're building it with custom flags (such as --config=benchmark), if we don't have a custom artifact, it will increase the duration of the runner. Are we ok with that?

anonrig: If we want to reduce the duration, we should add a large runner as well: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/managing-larger-runners#adding-a-larger-runner-to-an-organization

@danlapid I believe you are the only one who can add this.

anonrig:
> These numbers are likely too small to be used for performance comparisons, I think aiming for 5-10ms per benchmark test is likely a good starting point, microseconds benchmarks likely need more iterations

Agreed. I think we should make an update to all benchmarks after this pull-request lands. Most of those benchmarks are not needed, but I'm pretty sure it's up to the discussion.

.github/workflows/codspeed.yml
danlapid: The benchmark flag is not really changing the compile flags, does it?
It just enables compilation of a few extra files?

anonrig: Ideally it changes compile flags. It should run proper optimization flags etc.

danlapid: So is it the same flags as the release build?

anonrig: Unfortunately not.

```
# configuration used for performance profiling
build:profile --config=thin-lto
build:profile --copt="-fno-omit-frame-pointer" --copt="-mno-omit-leaf-frame-pointer"
build:profile --config=limited-dbg-info
build:profile --strip=never

# configuration used for performance benchmarking is the same as profiling for consistency
build:benchmark --config=profile
```

build/ci.bazelrc
```diff
@@ -43,7 +43,7 @@ build:ci-test --config=v8-codegen-opt
 # LLVM produces a bit more debug info on macOS by default to facilitate debugging with
 # LLDB. This is not needed for this use case, so disable it using -fno-standalone-debug –
 # this is already the default for Linux/Windows.
-build:ci-limit-storage --build_tag_filters=-off-by-default,-benchmark
+build:ci-limit-storage --build_tag_filters=-off-by-default,+workerd-benchmark
```
fhanau: Tag filters mean that "Only those targets will be built that contain at least one included tag and do not contain any excluded tags." so this change would change the configuration to exclusively run benchmark tests, which is not what we want. Just having `-off-by-default,-workerd-benchmark` would keep the same behavior as before right?

anonrig: You're right. Reverted this.

danlapid (approved):

PR #4395, opened 2025-06-23
add a composite action for preparing runners

anonrig:
It seems we're duplicating the same exact code multiple times. Let's use a composite workflow action

### DIFF
--- a/.github/actions/setup-runner/action.yml
+++ b/.github/actions/setup-runner/action.yml
@@ -1,0 +1,46 @@
+name: 'Setup runner environment'
+description: 'Sets up runner environment with proper toolchain for building workerd'
+runs:
+  using: 'composite'
+  steps:
+    - name: Setup Linux
+      shell: bash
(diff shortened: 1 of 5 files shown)


anonrig:
This is ready to land.

.github/workflows/lint.yml
```diff
@@ -20,9 +20,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           show-progress: false
-      - name: Configure git hooks
-        # Configure git to quell an irrelevant warning for runners (they never commit / push).
-        run: git config core.hooksPath githooks
```
fhanau: Nit: This job does not need to install the LLVM toolchain, so adding the shared config will only slow it down, can we leave the file as-is?

fhanau (approved):
LGTM otherwise.

.github/workflows/lint.yml
anonrig: Reverted.

PR #4401, opened 2025-06-24
feat(types): allow cmd-click into rpc props

mmkal:
Fixes https://github.com/cloudflare/workers-sdk/issues/9682

this change means you can cmd-click into RPC classes accessed through bindings via worker-configuration.d.ts:

Before (cmd-click on the rpc method in cursor/vscode does nothing):

https://github.com/user-attachments/assets/9d81c5da-98d3-4a8a-9609-3d790902f2d1

After (cmd-click goes to the rpc method implementation):

https://github.com/user-attachments/assets/0b4cb826-db06-4132-acb4-20a705511c46

I couldn't get bazel build working for me locally so thought I'd try opening here to see if CI tells me if something's wrongly formatted or typo'd. If a build goes red will push a fix. 

I didn't add any tests. I noticed you're using expect-type for testing types (I happen to be the author of that library!) so there's nothing really to add - the types before and after this change are equivalent, and the existing expect-type assertions should cover this/ensure no regressions. The only change is in IDEs - see the videos on the linked issue. 

Cc @penalosa since you suggested I tag you!

### DIFF
--- a/src/workerd/api/performance.c++
+++ b/src/workerd/api/performance.c++
@@ -5,7 +5,7 @@
 #include "performance.h"
 
 #include <workerd/io/io-util.h>
-#include <workerd/io/worker.h>
+#include <workerd/io/limit-enforcer.h>
 
 #include <kj/encoding.h>
 
--- a/src/workerd/io/BUILD.bazel
+++ b/src/workerd/io/BUILD.bazel
@@ -278,7 +278,7 @@ wd_cc_library(
     hdrs = ["limit-enforcer.h"],
     visibility = ["//visibility:public"],
     deps = [
-        ":observer",
+        ":outcome_capnp",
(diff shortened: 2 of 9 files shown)


mmkal:

sam-goodwin:
Can we get an update here? The rpc types are broken.

mmkal:
@penalosa another ping - `Run internal build` told me to ask the auto-assigned reviewers but it doesn't look like there are any.

petebacondarwin:
I think the CI failures are due to @mmkal not being a member of the Cloudflare org.
Not sure what the protocol for landing this is, in that case.

southpolesteve:

ketanhwr:
I've kicked off the internal build in a mirror branch: #5576 - check-snapshot should hopefully pass this time

PR #4403, opened 2025-06-24
Add non-top level tracing span context to js rpc calls.

mar-cf:


fhanau:
@jasnell we believe this could fix errors for tail worker events being reported before the onset event – possibly because we did not report an onset for non-top level JSRPC so far. I think this is worth trying

src/workerd/api/worker-rpc.c++
```diff
@@ -1511,7 +1511,10 @@ class TransientJsRpcTarget final: public JsRpcTargetBase {
   }
 
   void addTrace(jsg::Lock& js, IoContext& ioctx, kj::StringPtr methodName) override {
-    // TODO(someday): Trace non-top-level calls?
+    KJ_IF_SOME(t, mapAddRef(ioctx.getWorkerTracer())) {
```
fhanau, line 1514: Is the mapAddRef() call needed here?

fhanau:
> @jasnell we believe this could fix errors for tail worker events being reported before the onset event – possibly because we did not report an onset for non-top level JSRPC so far. I think this is worth trying

I think as long as we guard this behind the streaming tail worker compat flag, there is less risk for breakage – should be something like `ioCtx.getApi().getFeatureFlags().getStreamingTailWorker()`. But I don't understand the JSRPC implementation well, we need to avoid adding the onset twice.

fhanau:
@mar-cf Did some local testing using src/workerd/api/tests/js-rpc-test.js – this doesn't work as-is. It tries to add onsets to existing events, resulting in `failed: expected trace->eventInfo == kj::none [(can't stringify) == (can't stringify)]; tracer can only be used for a single event` errors.

PR #4404, opened 2025-06-24
handle native errors on value serializer

anonrig:
This is required for downstream v8 13.9 change.

### DIFF
--- a/src/workerd/jsg/ser.c++
+++ b/src/workerd/jsg/ser.c++
@@ -99,6 +99,14 @@ v8::Maybe<bool> Serializer::IsHostObject(v8::Isolate* isolate, v8::Local<v8::Obj
   KJ_ASSERT(!treatClassInstancesAsPlainObjects);
   KJ_ASSERT(!prototypeOfObject.IsEmpty());
 
+  // After v8 13.9, native errors needs to be handeld as a non-host objects.
+  // Meaning, we need to let default v8 serializer to handle native errors, rather than our
+  // own.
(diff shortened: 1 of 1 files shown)


src/workerd/jsg/ser.c++
```diff
@@ -99,6 +99,14 @@ v8::Maybe<bool> Serializer::IsHostObject(v8::Isolate* isolate, v8::Local<v8::Obj
   KJ_ASSERT(!treatClassInstancesAsPlainObjects);
   KJ_ASSERT(!prototypeOfObject.IsEmpty());
 
+  // After v8 13.9, native errors needs to be handeld as a non-host objects.
+  // Meaning, we need to let default v8 serializer to handle native errors, rather than our
+  // own.
+  // Ref: https://github.com/v8/v8/commit/e3df60f3f5abe85f819ff2fad6a41d0709e30a61
+  if (object->IsNativeError()) {
+    return v8::Just(false);
+  }
```
jasnell, line 108: Are we certain this is not a breaking change? Admittedly, it's unlikely but I think we need to be sure. 

 Specifically, if I have a class FooError that subclasses Error, this code will have previously caused it to be treated as a host object, correct?

anonrig: No, take a look at the referenced v8 change. We had to add this check because the just recently errors have started to act like host objects. But in reality default value serializer can handle them correctly, which is what happens when we return false from here.

PR #4405, opened 2025-06-24
Ensure SQLite KV put/putMultiple behavior matches actor-cache

jqmmes:
In actor-cache we validate that put / putMultiple values and keys are within the supported limits before doing any storage operations. We were not doing the same in actor-sqlite. 

This MR adds a preliminary check for key and value size in actor-sqlite put / putMultiple, implementing the same behavior we see in actor-cache.

### DIFF
--- a/src/workerd/io/actor-sqlite-test.c++
+++ b/src/workerd/io/actor-sqlite-test.c++
@@ -5,6 +5,7 @@
 #include "actor-sqlite.h"
 #include "io-gate.h"
 
+#include <workerd/util/autogate.h>
 #include <workerd/util/capnp-mock.h>
 #include <workerd/util/test.h>
 
@@ -147,6 +148,16 @@ struct ActorSqliteTest final {
   auto put(kj::StringPtr key, kj::StringPtr value, ActorCache::WriteOptions options = {}) {
     return actor.put(kj::str(key), kj::heapArray(value.asBytes()), options);
   }
+  auto putMultiple(
+      kj::Array<ActorCache::KeyValuePair> pairs, ActorCache::WriteOptions options = {}) {
+    return actor.put(kj::mv(pairs), options);
+  }
+  auto putMultipleExplicitTxn(
+      kj::Array<ActorCache::KeyValuePair> pairs, ActorCache::WriteOptions options = {}) {
+    auto txn = actor.startTransaction();
+    txn->put(kj::mv(pairs), options);
+    return txn->commit();
+  }
   auto setAlarm(kj::Maybe<kj::Date> newTime, ActorCache::WriteOptions options = {}) {
     return actor.setAlarm(newTime, options);
   }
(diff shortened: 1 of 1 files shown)


src/workerd/io/actor-sqlite.c++
```diff
@@ -379,6 +379,16 @@ void ActorSqlite::maybeDeleteDeferredAlarm() {
   }
 }
 
+void ActorSqlite::checkKeyValSize(Key& key, Value& value) {
```
jasnell:
```suggestion
void ActorSqlite::checkKeyValSize(const Key& key, const Value& value) {
```
?

src/workerd/io/actor-sqlite.c++
```diff
@@ -451,13 +461,20 @@ kj::OneOf<ActorCacheOps::GetResultList, kj::Promise<ActorCacheOps::GetResultList
...
+  checkKeyValSize(key, value);
+
   kv.put(key, value);
   return kj::none;
 }
 
 kj::Maybe<kj::Promise<void>> ActorSqlite::put(kj::Array<KeyValuePair> pairs, WriteOptions options) {
   requireNotBroken();
 
+  for (auto& pair: pairs) {
```
jasnell:
```suggestion
  for (const auto& pair: pairs) {
```

src/workerd/io/actor-sqlite.h
```diff
@@ -244,6 +244,9 @@ class ActorSqlite final: public ActorCacheInterface, private kj::TaskSet::ErrorH
   // Called when DeferredAlarmDeleter is destroyed, to delete alarm if not reset or cancelled
   // during handler.
   void maybeDeleteDeferredAlarm();
+
+  // Checks that the key and value sizes are bellow the SQLITE_LIMIT_LENGTH limit.
+  void checkKeyValSize(Key& key, Value& value);
```
jasnell:
```suggestion
  void checkKeyValSize(const Key& key, const Value& value);
```

src/workerd/util/sqlite-kv.h
```diff
@@ -57,6 +57,8 @@ class SqliteKv: private SqliteDatabase::ResetListener {
 
   uint deleteAll();
 
+  uint32_t getLimitLength();
```
jasnell:
```suggestion
  uint32_t getLimitLength() const;
```

src/workerd/util/sqlite-kv.c++
```diff
@@ -87,6 +87,10 @@ uint SqliteKv::deleteAll() {
   return count;
 }
 
+uint32_t SqliteKv::getLimitLength() {
```
jasnell:
```suggestion
uint32_t SqliteKv::getLimitLength() const {
```

Less sure about this one being const but ideally it would be. Not critical tho

src/workerd/io/actor-sqlite.c++
```diff
@@ -379,6 +379,16 @@ void ActorSqlite::maybeDeleteDeferredAlarm() {
...
 }
 
+void ActorSqlite::checkKeyValSize(Key& key, Value& value) {
+  auto limitLength = kv.getLimitLength() - 1;
+  JSG_REQUIRE(key.size() <= limitLength, RangeError,
+      kj::str("Keys cannot be larger than ", limitLength, " bytes. ", "A key of size ", key.size(),
+          " was provided."));
+  JSG_REQUIRE(value.size() <= limitLength, RangeError,
+      kj::str("Values cannot be larger than ", limitLength, " bytes. ", "A value of size ",
+          value.size(), " was provided."));
```
justin-mp: This calculation is incorrect.  SQLITE_LIMIT_LENGTH is "[t]he maximum size of any string or BLOB or table row, in bytes."  The table row includes both the key and the value and some overhead.

You could make this more accurate by checking that the length of the key + the length of the value < the limit length, but it would still be wrong in a bunch of cases.

The problem that we have here is that if you `putMultiple` and some value is above the limit, KV-backed DOs will not commit anything while SQLite-backed DOs will still commit some values.  I thought we were going to put everything into a transaction and have SQLite do the check for us, and if the check failed, then we roll back the transaction.

jqmmes: Dropped this approach all together.  The current approach rolls back the transaction if a single put fails.

src/workerd/io/actor-sqlite.c++
```diff
@@ -474,9 +474,18 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::put(Key key, Value value, WriteOptions
 kj::Maybe<kj::Promise<void>> ActorSqlite::put(kj::Array<KeyValuePair> pairs, WriteOptions options) {
   requireNotBroken();
 
-  for (auto& pair: pairs) {
-    kv.put(pair.key, pair.value);
+  for (const auto& pair: pairs) {
+    try {
+      kv.put(pair.key, pair.value);
+    } catch (kj::Exception e) {
+      KJ_IF_SOME(txn, currentTxn.tryGet<ImplicitTxn*>()) {
```
justin-mp: This doesn't seem correct to me.  If you did a single put, then this multi-put, and this multi-put fails, this rollback would rollback the single put.

What would be better would be to create your own explicit transaction here and either commit or rollback that transaction.

jqmmes: Yeah, I wasn't too sure what transaction would work best. I noticed since we always have an implicit transaction that could work, but I guess that, if we don't commit a transaction at the end of each kv operation, then it can cause unwanted rollbacks

jqmmes: Updated the PR. 
- If we are not in a transaction, wrap it in an explicit transaction and call put on it (which will circle back to the put implementation, but inside a ExplictiTxn instead). 
- If we are in a transaction, set a SAVEPOINT and try to do puts. If a put fails, ROLLBACK TO the SAVEPOINT

These changes pass all internal tests.

jqmmes:
Updated PR to wrap new behavior in an autogate for safer rollout.

src/workerd/io/actor-sqlite.c++
```diff
@@ -473,9 +474,33 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::put(Key key, Value value, WriteOptions
 
 kj::Maybe<kj::Promise<void>> ActorSqlite::put(kj::Array<KeyValuePair> pairs, WriteOptions options) {
   requireNotBroken();
-
-  for (auto& pair: pairs) {
-    kv.put(pair.key, pair.value);
+  if (util::Autogate::isEnabled(util::AutogateKey::SQL_KV_PUT_MULTIPLE_TRANSACTION)) {
```
justin-mp, line 478: It's probably worth leaving a cleanup comment in this block that this code is basically stolen from `DurableObjectStorage::transactionSync`.

src/workerd/io/actor-sqlite.c++
```diff
@@ -473,9 +474,33 @@ kj::Maybe<kj::Promise<void>> ActorSqlite::put(Key key, Value value, WriteOptions
...
+    if (currentTxn.is<NoTxn>()) {
```
justin-mp, line 479: Please test three branches of `currentTxn` (No, Explicit, Implicit) in a test in workerd.  You can probably do this in actor-sqlite-test.c++.

justin-mp (commented):
I think the code is fine, but we could use some tests.

src/workerd/io/actor-sqlite.c++
jqmmes: Will do

jqmmes: Added testing in actor-sqlite-test

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -168,6 +179,284 @@ KJ_TEST("can set and get alarm") {
...
+  kj::Vector<ActorCache::KeyValuePair> putKVs;
+
+  // Add some regular key-value pairs that we know are supported
+  putKVs.add(ActorCache::KeyValuePair{kj::str("foo"), kj::heapArray(kj::str("bar").asBytes())});
+  putKVs.add(ActorCache::KeyValuePair{kj::str("foo2"), kj::heapArray(kj::str("bar2").asBytes())});
+  putKVs.add(ActorCache::KeyValuePair{kj::str("foo3"), kj::heapArray(kj::str("bar3").asBytes())});
+
+  // Now create a key that's too large. Should fail with  string or blob too big: SQLITE_TOOBIG
+  auto tooLongKey = kj::heapString(2200000);
+  for (int x = 0; x < 2200000; x++) {
```
justin-mp: You can do `tooLongKey.asArray().fill('a');` which seems slightly cleaner because you don't have to repeat the constant.

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -168,6 +179,295 @@ KJ_TEST("can set and get alarm") {
...
```
justin-mp: Same comment as above.

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -168,6 +179,295 @@ KJ_TEST("can set and get alarm") {
...
+    KJ_ASSERT(!test.actor.isCommitScheduled());
+    try {
+      test.putMultiple(putKVs.releaseAsArray());
+      // We should fail with correct error before reaching here.
+      KJ_UNREACHABLE;
+    } catch (kj::Exception e) {
+      KJ_ASSERT(
+          e.getDescription() == "expected false; jsg.Error: string or blob too big: SQLITE_TOOBIG");
+    }
+    // We need this check for the @all-autogates-enabled test variant.
```
justin-mp: Rather than doing this, it might be easier to think about to check for the autogate at the top of the test, and if it's set, emit a log saying you're skipping this test because the autogate is set and then return.

jqmmes: Addressed in latest update

src/workerd/io/actor-sqlite-test.c++
jqmmes: That's helpful! Applied in the latest update

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -168,6 +179,270 @@ KJ_TEST("can set and get alarm") {
...
+    KJ_ASSERT(KJ_ASSERT_NONNULL(expectSync(test.get("foo3"))) == kj::str("bar3").asBytes());
+    commitFulfiller->fulfill();
+  }
+}
+
+KJ_TEST("check put multiple wraps operations in a transaction and does not rollbacks on error") {
+  ActorSqliteTest test;
+
+  if (util::Autogate::isEnabled(util::AutogateKey::SQL_KV_PUT_MULTIPLE_TRANSACTION)) {
+    // We should skip this test as it will expect a different behavior when
```
justin-mp, line 286: nit: you should log something so it at least appears in the test logs.

src/workerd/io/actor-sqlite-test.c++
```diff
@@ -168,6 +179,270 @@ KJ_TEST("can set and get alarm") {
...
+    test.putMultiple(putKVs.releaseAsArray());
+
+    auto commitFulfiller = kj::mv(test.pollAndExpectCalls({"commit"})[0]);
+    KJ_ASSERT(KJ_ASSERT_NONNULL(expectSync(test.get("baz"))) == kj::str("bat").asBytes());
+    KJ_ASSERT(KJ_ASSERT_NONNULL(expectSync(test.get("foo3"))) == kj::str("bar3").asBytes());
+    commitFulfiller->fulfill();
+  }
+}
+
+KJ_TEST("check put multiple wraps operations in a transaction and does not rollbacks on error") {
```
justin-mp: s/rollbacks/rollback/

PR #4406, opened 2025-06-24
fix arm64 builds caused by v8 13.8 update

anonrig:
Let's not wait for v8 13.9 to fix the arm64 builds.


anonrig:
This is ready to land, and fixes the error on arm64 debug builds.

PR #4408, opened 2025-06-24
enable clang-tidy and fix json indentations

anonrig:
This enables clang-tidy on vscode and zed folders so that our IDEs can detect clang-tidy errors. This doesn't change anything yet, but in the future we might want to add a github workflow to detect clang-tidy errors/regressions.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -8,8 +8,6 @@ Checks: >
   bugprone-move-forwarding-reference,
   bugprone-return-const-ref-from-parameter,
   bugprone-use-after-move,
-  clang-analyzer-optin.cplusplus.VirtualCall,
-  clang-analyzer-security.insecureAPI.strcpy,
   misc-header-include-cycle,
   readability-duplicate-include,
 WarningsAsErrors: '*'


.clang-tidy
```diff
@@ -0,0 +1,21 @@
...
+# TODO: We currently disable default checks with -*. While many checks provide little value or
+# produce false positives, try to incrementally enable most of them.
+# TODO: these checks are in progress of cleaning up
+Checks: >
+  -*,
+  bugprone-capturing-this-in-member-variable,
+  bugprone-move-forwarding-reference,
+  bugprone-return-const-ref-from-parameter,
+  bugprone-use-after-move,
+  clang-analyzer-cplusplus.PureVirtualCall,
```
fhanau: nit: We use these checks downstream, but clang static analyzer checks can be excruciatingly slow – might be something we can't afford to run automatically by the IDE

anonrig: Happy to remove them. I'm not sure which ones are extremely slow and which ones aren't. Can you recommend github code change?

jclee: FWIW, the `clang-analyzer-*` ones, specifically, are extremely slow.

fhanau: Yup, dropping just those should be sufficient here (clang static analyzer can be useful, despite being really slow and having many false positives, we should find another place to enable them).

fhanau (approved):
LGTM otherwise.

.clang-tidy
anonrig:
```suggestion
```

.clang-tidy
```diff
@@ -0,0 +1,21 @@
...
+  clang-analyzer-optin.cplusplus.VirtualCall,
+  clang-analyzer-security.insecureAPI.strcpy,
```
anonrig:
```suggestion
```

PR #4413, opened 2025-06-25
[build] Roll Bazel -> 8.3.1

fhanau:


fhanau:
Bazel skylib published a new release right after I added a git_override() to get the latest commit🙄  at least the configuration is cleaner now that we have the release.

anonrig (approved):

Issue #4326, opened 2025-06-13
[Feature] Container creation and execution


vicb:
## Starting of the container

We do start container using a [start command](https://github.com/cloudflare/workerd/blob/6a10e29237a9824051a142bfcd8dab1a1138f01e/src/workerd/server/container-client.c%2B%2B#L292)

I believe this will re-start an existing container if it already exists. That's not the same as starting a new container when the container is stateful.

Does that match what we do in prod? 

## Creation of the container

If a container of the same name already exists ([error 409](https://github.com/cloudflare/workerd/blob/6a10e29237a9824051a142bfcd8dab1a1138f01e/src/workerd/server/container-client.c%2B%2B#L286)) we will re-use that.

That could be a problem when the existing container had a different environment/entrypoint as it will not be updated.

One way to overcome that would be to hash `{environment, entrypoint}` and attach that to the container (via the name or a label). And re-create the container if something changed.

/cc @mikenomitch @gabivlj @emily-shen @anonrig

anonrig:
> I believe this will re-start an existing container if it already exists. That's not the same as starting a new container when the container is stateful.

@vicb If the container is already started, we throw an error: https://github.com/cloudflare/workerd/blob/6a10e29237a9824051a142bfcd8dab1a1138f01e/src/workerd/server/container-client.c%2B%2B#L299

vicb:
> [@vicb](https://github.com/vicb) If the container is already started, we throw an error:

Oh sorry I should have been clearer by "re-start" I do not mean "start" -> "start" but "stop" -> "start" (i.e. start a stopped container).

gabivlj:
In production, when we stop a container, we start all-over with a brand new rootfs. The new configuration from start() always has precedence. In summary, there is nothing stateful across restarts.

emily-shen:
so it looks like we want to destroy and recreate here if we get a 409? https://github.com/cloudflare/workerd/blob/6a10e29237a9824051a142bfcd8dab1a1138f01e/src/workerd/server/container-client.c%2B%2B#L284C53-L284C54

PR #4416, opened 2025-06-26
destroy on recreate container if it's already created

anonrig:
Fixes https://github.com/cloudflare/workerd/issues/4326


src/workerd/server/container-client.c++
```diff
@@ -286,17 +286,22 @@ kj::Promise<void> ContainerClient::createContainer(
...
-  // recreate on 409 in the future.
+  // In that case we destroy and re-create the container.
+  if (response.statusCode == 409) {
+    co_await destroyContainer();
+    response = co_await dockerApiRequest(network, kj::str(dockerPath), kj::HttpMethod::POST,
+        kj::str("/containers/create?name=", containerName), codec.encode(jsonRoot));
+  }
+
+  // statusCode 201 refers to "container created successfully"
   if (response.statusCode != 201 && response.statusCode != 409) {
```
vicb: Should `!= 409` still be there?

anonrig: I'm not sure. If even after we destroy and recreate, and it still says it's already running, I assume that's a bug with docker that we can ignore.

src/workerd/server/container-client.c++
```diff
@@ -335,6 +340,25 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
...
+kj::Promise<void> ContainerClient::destroyContainer() {
+  auto endpoint = kj::str("/containers/", containerName, "?force=true");
+  auto response = co_await dockerApiRequest(
+      network, kj::str(dockerPath), kj::HttpMethod::DELETE, kj::mv(endpoint));
+  // statusCode 204 refers to "no error"
+  // statusCode 404 refers to "no such container"
+  // Both of which are fine for us since we're tearing down the container anyway.
+  JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 404, Error,
+      "Removing a container failed with: ", response.body);
+  {
```
vicb: question for my understanding: why do you create a new scope here?

src/workerd/server/container-client.c++
```diff
@@ -286,17 +286,22 @@ kj::Promise<void> ContainerClient::createContainer(
...
+  auto response = co_await dockerApiRequest(network, kj::str(dockerPath), kj::HttpMethod::POST,
+      kj::str("/containers/create?name=", containerName), codec.encode(jsonRoot));
 
-  // statusCode 201 refers to "container created successfully"
   // statusCode 409 refers to "conflict". Occurs when a container with the given name exists.
-  // Both are fine, so long as the container exists. Though we might want to call destroy and
-  // recreate on 409 in the future.
+  // In that case we destroy and re-create the container.
+  if (response.statusCode == 409) {
+    co_await destroyContainer();
```
vicb, line 295: would calling `co_await destroyContainer();` on l 288 and removing this `if` block make a diff?

src/workerd/server/container-client.c++
anonrig: No need. I don't remember why I did that in the first place. Probably code evolved so that this is not needed anymore. 

I'll remove it.

src/workerd/server/container-client.c++
```diff
@@ -335,6 +340,25 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
       "Stopping container failed with: ", response.body);
 }
 
+// This method assumes that the container is already running.
```
vicb: I'm not sure to understand why?

src/workerd/server/container-client.c++
```diff
@@ -335,6 +340,25 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
...
+  auto endpoint = kj::str("/containers/", containerName, "?force=true");
+  auto response = co_await dockerApiRequest(
+      network, kj::str(dockerPath), kj::HttpMethod::DELETE, kj::mv(endpoint));
+  // statusCode 204 refers to "no error"
+  // statusCode 404 refers to "no such container"
+  // Both of which are fine for us since we're tearing down the container anyway.
+  JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 404, Error,
+      "Removing a container failed with: ", response.body);
+  {
+    endpoint = kj::str("/containers/", containerName, "/wait");
```
vicb: I should check the doc... but I guess that is waiting something on the container.
So do we need to call this when the previous responses 404?

src/workerd/server/container-client.c++
vicb: > I assume that's a bug with docker that we can ignore.

I was under the same impression (i.e. it should not happen) but I had a different conclusion: may we should `JSG_REQUIRE(response.statusCode != 409)` ?

src/workerd/server/container-client.c++
anonrig: Avoiding this assumption requires a request for inspect path. If we don't assume then, we have to make an unnecessary request on destroy() function.

src/workerd/server/container-client.c++
anonrig: Good catch. Updated.

src/workerd/server/container-client.c++
anonrig: In a happy path, we now send only 1 request, but with your proposed changes we will always send 2 request on a happy path. I think it makes more sense to optimize for happy path rather than for people calling "start()" on a container that is already started.

kentonv:
Probably best for @gabivlj to look at this and say whether this matches production behavior.

It seems a bit surprising to me -- do we really want to recreate containers instead of connecting to the already-running one?

src/workerd/server/container-client.c++
vicb: Could you add that to the comment?

anonrig:
@gabivlj can you review? this is ready to land.

anonrig:
This requires an additional review from runtime team. @danlapid can you review?

src/workerd/server/container-client.c++
```diff
@@ -335,6 +341,27 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
       "Stopping container failed with: ", response.body);
 }
 
+// This method assumes that the container is already running to avoid sending
+// unnecessary requests for checking the status of the container (via /inspect).
```
vicb: Could you add what happens if the container is not already running?

vicb: I'm asking because it doesn't look to me that when this is called on l295 the assumption  is enforced/true.

src/workerd/server/container-client.c++
```diff
@@ -335,6 +341,27 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
...
+  // statusCode 204 refers to "no error"
+  // statusCode 404 refers to "no such container"
+  // Both of which are fine for us since we're tearing down the container anyway.
+  JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 404, Error,
+      "Removing a container failed with: ", response.body);
+  // Do not send a wait request if container doesn't exist. This avoids sending an
+  // unnecessary request.
+  if (response.statusCode == 204) {
+    response = co_await dockerApiRequest(network, kj::str(dockerPath), kj::HttpMethod::POST,
+        kj::str("/containers/", containerName, "/wait"));
```
vicb: It looks like the condition is [`not-running` by default](https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Container/operation/ContainerWait) - do we want `remove` here instead?
At least I look to me that this is what we want l295 and also l361

Also maybe clarify the behavior in the method comment?

src/workerd/server/container-client.c++
```diff
@@ -335,6 +341,27 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
       "Stopping container failed with: ", response.body);
 }
 
+// This method assumes that the container is already running to avoid sending
+// unnecessary requests for checking the status of the container (via /inspect).
+kj::Promise<void> ContainerClient::destroyContainer() {
+  auto endpoint = kj::str("/containers/", containerName, "?force=true");
```
vicb, line 348: I don't see this endpoint documented at `https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Container`.

Should it be `/json` instead, i.e.  `auto endpoint = kj::str("/containers/", containerName, "/json?force=true");`

i.e. https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Container/operation/ContainerInspect

If that's the case, should we improve the test coverage?

vicb (changes requested):
If I'm not mistaken 2 things need to be changed:

- ~the endpoint address in `destroyContainer`~
- the wait behavior in the same method.

Also would be nice to add more details in the comment.

All of that is covered in more details in inline comments

Edit (June30):

> ~the endpoint address in `destroyContainer`~

was my mistake, no change needed here

> the wait behavior in the same method.

has been fixed

src/workerd/server/container-client.c++
vicb: Oops no it's https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Container/operation/ContainerDelete

src/workerd/server/container-client.c++
anonrig: Please take a look at the comment above that line

```
  // statusCode 409 refers to "conflict". Occurs when a container with the given name exists.
  // In that case we destroy and re-create the container.
```

response.statusCode == 409 will only be thrown if the container already exists. And if the container exists, you can remove it. I'll add one more line for the comment and share the reference link.

src/workerd/server/container-client.c++
anonrig: Sounds good to me. Applying the changes.

danlapid (approved):

src/workerd/server/container-client.c++
vicb: What I see is "when a container with the given name exists" which is not the same as "assumes that the container is already running".

Assuming the assumption should be "exists" instead of "running" then I don't think it's needed as deleting the container that does not exist will return a 404 in this method when the container does not exist.

src/workerd/server/container-client.c++
```diff
@@ -335,6 +341,29 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
       "Stopping container failed with: ", response.body);
 }
 
+// This method assumes that the container is already running to avoid sending
+// unnecessary requests for checking the status of the container (via /inspect).
```
vicb:
```suggestion
// Destroys the container.
// No-op when the container does not exist.
// Wait for the container to actually be stopped and removed when it exists.
```

src/workerd/server/container-client.c++
```diff
@@ -335,6 +341,30 @@ kj::Promise<void> ContainerClient::killContainer(uint32_t signal) {
       "Stopping container failed with: ", response.body);
 }
 
+// Destroys the container.
+// No-op when the container does not exist.
+// Wait for the container to actually be stopped and removed when it exists.
+// if the container doesn't exist it will return 404.
+// Ref: https://docs.docker.com/reference/api/engine/version/v1.50/#tag/Container/operation/ContainerDelete
```
vicb: This returns kj::Promise\<void\>

```suggestion
```

anonrig:
Rebased and force pushed.

src/workerd/server/container-client.c++
```diff
@@ -387,22 +415,7 @@ kj::Promise<void> ContainerClient::monitor(MonitorContext context) {
...
-    JSG_REQUIRE(response.statusCode == 204 || response.statusCode == 404, Error,
-        "Removing a container failed with: ", response.body);
-    {
-      endpoint = kj::str("/containers/", containerName, "/wait");
-      response = co_await dockerApiRequest(
-          network, kj::str(dockerPath), kj::HttpMethod::POST, kj::mv(endpoint));
-      JSG_REQUIRE(response.statusCode == 200 || response.statusCode == 404, Error,
-          "Waiting for container removal failed with: ", response.statusCode, response.body);
-    }
+    co_await destroyContainer();
```
vicb, line 418: Question (Sorry I missed that during my earlier reviews)

I'm wondering if there is a need to have both `destroyContainer` and `destroy`?

Couldn't the current implementation of `destroyContainer` be used here (and then we do not need to add that new method)?

`container.capnp` has:

```
  destroy @3 ();
  # Immediately and abruptly stops the container and tears it down. The application is not given
  # any warning, it simply stops immediately. Upon successful return from destroy(), the container
  # is no longer running. If a call to `monitor()` is waiting when `destroy()` is invoked,
  # `monitor()` will also return (with no error). If the container is not running when `destroy()`
  # is invoked, `destroy()` silently returns with no error.
```

But with the current implementation, if the container is not running (i.e. `signal()` was called  or the app ended) then it will not be tear down. So it looks like not tearing down the container when the container is not running might lead to an unexpected state?

/cc @gabivlj @emily-shen  @mikenomitch

anonrig: destoyContainer is called from 2 different places. I'll prefer landing this right now, and have this discussion afterwards.

vicb: > destoyContainer is called from 2 different places

Yes what I'm suggesting above is that I think it's not needed and might lead to a bug?

I'll create a separate issue.

mikenomitch: Yeah sorry @vicb I'm not following that well TBH, so to the extent you can explain it like I'm 5 in the issue, will be helpful!

vicb: I created #4440 but basically:

```ts
kj::Promise<void> ContainerClient::destroy(DestroyContext context) {
  const auto [running, _ports] = co_await inspectContainer();
  if (running) {
    co_await destroyContainer();
  }
}
```

is a no-op if the container is not running.

but the comment says it should (implicitly always) tear down the container.

So I think the impl should rather be:

```ts
kj::Promise<void> ContainerClient::destroy(DestroyContext context) {
  co_await destroyContainer();
}
```

At this point `destroy` only calls `destroyContainer` so the 2 fn should be merged.

anonrig: Ah I see what you mean now. You're right.

PR #4417, opened 2025-06-26
containers: Abort when the monitor exception is DISCONNECT

gabivlj:
The issue that we are trying to fix is the capability throwing DISCONNECTED and us aborting the DO if that happens.

maybe not the best way to do it? How would I go about it? I just came up with catching the monitor exception and aborting in that case.

### DIFF
--- a/src/workerd/api/container.c++
+++ b/src/workerd/api/container.c++
@@ -14,7 +14,26 @@ namespace workerd::api {
 
 Container::Container(rpc::Container::Client rpcClient, bool running)
     : rpcClient(IoContext::current().addObject(kj::heap(kj::mv(rpcClient)))),
-      running(running) {}
+      running(running) {
+}
+
+void catchAndAbortOnDisconnect(jsg::Lock& js, jsg::Value err) {
+  jsg::JsValue handle = jsg::JsValue(err.getHandle(js));
+  auto kjException = js.exceptionToKj(handle);
+
+  // Check if the error is around disconnect
+  if (kjException.getType() == kj::Exception::Type::DISCONNECTED) {
+    auto e = JSG_KJ_EXCEPTION(FAILED, Error, "The client has disconnected");
+
+    // And abort the DO if it's due to a disconnect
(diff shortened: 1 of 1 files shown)


src/workerd/api/container.c++
```diff
@@ -75,7 +94,7 @@ jsg::Promise<void> Container::monitor(jsg::Lock& js) {
     running = false;
```
danlapid, line 94: You can use this block which is already the error handler 😄 

Would really help deduplicate some code.

src/workerd/api/container.c++
```diff
@@ -14,7 +14,26 @@ namespace workerd::api {
...
+void catchAndAbortOnDisconnect(jsg::Lock& js, jsg::Value err) {
+  jsg::JsValue handle = jsg::JsValue(err.getHandle(js));
+  auto kjException = js.exceptionToKj(handle);
+
+  // Check if the error is around disconnect
+  if (kjException.getType() == kj::Exception::Type::DISCONNECTED) {
+    auto e = JSG_KJ_EXCEPTION(FAILED, Error, "The client has disconnected");
+
+    // And abort the DO if it's due to a disconnect
+    IoContext::current().abort(kj::mv(e));
```
danlapid, line 29: Do we really want to abort? (honest question).
Shouldn't we just reconnect to CC?

gabivlj: How would it look like reconnecting in EW?

I think aborting could make sense because it would be like a script update (disconnects are infrequent)

kentonv:
I guess the reason you want to abort rather than throw to the caller is because they likely cannot start the container again, because it'll just throw the same exception. We need to restart the object in order to get a new connection to the container engine.

However, I think matching on DISCONNECTED for this is too coarse.

We probably need to monitor for disconnect of the actual connection inside the supervisor (in the internal codebase), and break the actor when we see a disconnect there.

That said, isn't the connection over a UDS socket? Why would it disconnect? (Maybe this discussion should move to an internal thread, though...)

PR #4419, opened 2025-06-27
[o11y] nfc: Refactor span tracing, avoid synthesizing span events in backend

fhanau:
Working towards fully decomposing spans, providing spanId/parentSpanId as intended in design.

- Move SpanBuilder constructor into source file


src/workerd/io/worker-interface.capnp
```diff
@@ -289,10 +289,6 @@ struct Trace @0x8e8d911203762d34 {
       exception @11 :Exception;
       log @12 :Log;
       link @13 :Link;
-      # While invocation span context (EW-8821) is not fully implemented, send completed spans as
-      # events so that we can provide timestamps and parent span definitions properly. Can be
-      # removed once that is done and span API is finalized.
-      completedSpan @14 :UserSpanData;
```
mar-cf, line 295: Non-backwards compatible change, breakages OK'd here at this stage.

fhanau:
Latest push is rebase only.

src/workerd/io/tracer.c++
```diff
@@ -35,7 +35,7 @@ void TailStreamWriter::report(const InvocationSpanContext& context, TailEvent::E
   // been closed.
   // This could be an assert, but just log an error in case this is prevalent in some edge case.
   if (outcomeSeen) {
-    KJ_LOG(ERROR, "reported tail stream event after stream close", event.is<CompleteSpan>(),
+    KJ_LOG(ERROR, "reported tail stream event after stream close", event.is<SpanOpen>(),
```
mar-cf, line 38: Add check for attribute and span close

mar-cf:
Following up will include removing CompleteSpan entirely - which will require WorkerTracer to consume composed span events instead of CompleteSpan.

PR #4420, opened 2025-06-27
Add base64 internal module and use for AI binding

ns476:
This will also be used for base64 support in the Images binding (hence the two functions that appear unused).


src/workerd/api/base64.c++
```diff
@@ -0,0 +1,45 @@
...
+  return buf.slice(0, out_size).attach(kj::mv(buf));
+}
+
+kj::String Base64Module::encodeArrayToString(kj::Array<kj::byte> input) {
+  auto size = simdutf::base64_length_from_binary(input.size()) + 1;
+  auto buf = kj::heapArray<kj::byte>(size);
+  auto out_size = simdutf::binary_to_base64(
+      input.asChars().begin(), input.size(), buf.asChars().begin(), simdutf::base64_default);
+  KJ_ASSERT(out_size + 1 <= size);
+  buf[out_size] = 0;
```
anonrig: What's the purpose of adding this?

src/workerd/api/base64.c++
```diff
@@ -0,0 +1,45 @@
...
+
+  // As we've base64 encoded, `buf` is guaranteed to contain valid UTF-8
+  return kj::String(buf.releaseAsChars());
```
anonrig: This makes an unnecessary copy. I think this function should just return jsg::JsString and can use

```
return js.str(buf.first(out_size);
```

You can use the following code as an example

```
jsg::JsString ServiceWorkerGlobalScope::btoa
```

types/defines/base64.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module "cloudflare:base64" {
```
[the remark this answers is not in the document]
anonrig: This recommendation is correct

types/defines/base64.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module "cloudflare-internal:base64" {
```
anonrig: Are you sure this file is required?

ns476: Ah yes, this is a private module, so we don't want this

src/workerd/api/base64.c++
```diff
@@ -0,0 +1,43 @@
+#include "base64.h"
+
+#include <kj/debug.h>
+
```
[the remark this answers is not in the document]
dario-piotrowicz: I think this comment is valid? 🤔

dario-piotrowicz (commented):
Hey @ns476 I can see windows build failures, could you look into addressing those? 🙏

ns476:
> Hey @ns476 I can see windows build failures, could you look into addressing those? 🙏

I've tried a different approach, let's see if that works

src/workerd/api/base64.c++
```diff
@@ -0,0 +1,45 @@
+#include "base64.h"
+
+#include "simdutf.h"
+
+#include <kj/debug.h>
+
+namespace workerd::api {
+
+kj::StringPtr simd_error_to_string(simdutf::error_code err) {
```
anonrig: This needs to return kj::String or you will see a use after free error in asan builds. Can you add a test that triggers this path?

src/workerd/api/base64.c++
```diff
@@ -0,0 +1,40 @@
+#include "base64.h"
+
+#include "simdutf.h"
+
+#include <kj/debug.h>
+
+namespace workerd::api {
+
+kj::Array<kj::byte> Base64Module::decodeArray(kj::Array<kj::byte> input) {
+  auto size = simdutf::maximal_binary_length_from_base64((const char*)input.begin(), input.size());
```
dario-piotrowicz, line 10:
```suggestion
  auto size = simdutf::maximal_binary_length_from_base64((const char*) input.begin(), input.size());
```
nit: are we missing a space here? 🤔

dario-piotrowicz (approved):
LGTM 🙂

src/workerd/api/base64.c++
ns476: clang-format yells with a space:
```
src/workerd/api/base64.c++:10:71: error: code should be clang-formatted [-Wclang-format-violations]
  auto size = simdutf::maximal_binary_length_from_base64((const char*) input.begin(), input.size());
                                                                      ^
ERROR:root:Code has linting issues. Fix with python ./tools/cross/format.py

ERROR: changes staged for commit have lint. Pass '--no-verify' or '-n' to skip.

To fix lint:
  just format
```

dario-piotrowicz: oh

ok sorry, my bad 🙇

kentonv:
Hey folks, this broke the internal build. It looks like the internal build CI action was skipped for some reason.

jasnell:
This was reverted and will need to be redone.  Looks like the original feature branch was deleted so we can't reopen this.

PR #4421, opened 2025-06-27
run container tests on CI

anonrig:
Let's run container tests on CI


build/ci.bazelrc
```diff
@@ -89,6 +89,11 @@ test:ci-linux --test_size_filters=
 test:ci-macOS --test_size_filters=
 test:ci-windows --test_size_filters=
 
+# Enable container tests that require container engine on Linux CI runs only
+# This ensures tests tagged with "requires-container-engine" run only in CI, not locally
+# The filter excludes "off-by-default" tests UNLESS they also have "requires-container-engine"
+test:ci-linux --test_tag_filters=-off-by-default+requires-container-engine
```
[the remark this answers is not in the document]
danlapid: Bro, that's the point

src/workerd/server/tests/container-client/BUILD.bazel
```diff
@@ -1,9 +1,75 @@
 load("//:build/wd_test.bzl", "wd_test")
 
+# TODO(soon): Replace this with https://github.com/bazelbuild/rules_docker
```
anonrig: This means we also have to install this dependency in downstream repo. @fhanau any recommendations on how to make the following ugly code better?

fhanau: Shouldn't using `off-by-default` be sufficient to not have this run downstream?

anonrig: Yes, the goal is to run this on GitHub workflow.

build/deps/gen/dep_rules_docker.bzl
danlapid, line 1: remove?

src/workerd/server/tests/container-client/BUILD.bazel
```diff
@@ -1,9 +1,75 @@
...
+    name = "build_container_image",
+    srcs = [
+        "Dockerfile",
+        "simple-node-app.js",
+    ],
+    outs = ["container_image_built.txt"],
+    cmd = """
+        # Set Docker config to a writable location
+        export DOCKER_CONFIG=/tmp/.docker
+        mkdir -p $$DOCKER_CONFIG
```
danlapid: Use Bazel's TEST_TMPDIR ?

.github/actions/setup-runner/action.yml
```diff
@@ -25,6 +25,9 @@ runs:
         sudo ln -s /usr/bin/llvm-symbolizer-19 /usr/bin/llvm-symbolizer
         echo "build:linux --action_env=CC=/usr/lib/llvm-19/bin/clang" >> .bazelrc
         echo "build:linux --host_action_env=CC=/usr/lib/llvm-19/bin/clang" >> .bazelrc
+    - name: Setup Docker on Linux
+      if: runner.os == 'Linux'
+      uses: docker/setup-docker-action@v4
```
mikea: I think best security practice nowadays is to put commit ids

src/workerd/server/tests/container-client/BUILD.bazel
```diff
@@ -1,9 +1,75 @@
...
+# TODO(soon): Replace this with https://github.com/bazelbuild/rules_docker
+# Build the Docker image required for the container test
+genrule(
+    name = "build_container_image",
+    srcs = [
+        "Dockerfile",
+        "simple-node-app.js",
+    ],
+    outs = ["container_image_built.txt"],
+    cmd = """
```
mikea: I vote rules_docker or rules_oci, maintaining this will be terrible.

anonrig: rules_oci requires buildx which we need to manually copy to the codebase: https://github.com/bazel-contrib/rules_oci/blob/ac44bb6e1f577ca0d3529646de850e9dea200a2d/examples/dockerfile/buildx.bzl

build/ci.bazelrc
```diff
@@ -58,6 +58,7 @@ build:ci-linux-common --action_env=CC=/usr/lib/llvm-19/bin/clang
 build:ci-linux-common --host_action_env=CC=/usr/lib/llvm-19/bin/clang
 
 build:ci-linux --config=ci-linux-common --remote_download_regex=".*src/workerd/server/workerd.*"
+build:ci-linux --build_tag_filters=-off-by-default,+requires-container-engine
```
fhanau: You should update test_tag_filters instead – otherwise the test is being built and still filtered out. All targets that are test targets or their dependencies will be built if they are not excluded by the test filter, even if they would be filtered out by the build filter.
However, even with test_tag_filters we'd run into trouble: If we use the filter as written here we would only run the container test and no other test since "Only those test targets will be found that contain at least one included tag and do not contain any excluded tags". https://bazel.build/reference/command-line-reference#build-flag--test_tag_filters

danlapid: Maybe a solution similar to enable-python is more appropriate?

fhanau: Two ways to address this:
1) Mark the test as ["off-by-default", "requires-container-engine"] as written. Would require running `bazel test` several times, the first time with -off-by-default filter as before and the second time with +requires-container-engine
2) Mark test as just ["requires-container-engine"]. Avoids having to run Bazel twice, but would require adding -requires-container-engine to all filters except for the one used when we want to run this test, including the ones downstream.

I think this also means the test did not run in CI yet?

danlapid:
We should probably revive this PR rather soon, this is important coverage.

gpanders:
@anonrig @danlapid Container tests (and associated images) are now built and run by Bazel in Linux CI. Please re-review when you're able.

MODULE.bazel
```diff
@@ -84,3 +84,21 @@ include("//build/deps:deps.MODULE.bazel")
 include("//build/deps:nodejs.MODULE.bazel")
 include("//build/deps:python.MODULE.bazel")
 include("//build/deps:v8.MODULE.bazel")
+
+# rules_oci and rules_multirun are used to build Docker images
```
anonrig: @fhanau would you mind taking a look at the bazel changes?

package.json
```diff
@@ -14,7 +14,8 @@
     "esbuild": "^0.25.10",
     "eslint": "^9.37.0",
     "expect-type": "^1.2.2",
-    "typescript-eslint": "^8.46.1"
+    "typescript-eslint": "^8.46.1",
+    "ws": "^8.18.3"
```
anonrig: This shouldn't be necessary. What's the reason for this change?

src/workerd/server/tests/container-client/README.md
```diff
@@ -1,33 +1,33 @@
...
 
-- Run the test
+5. Run the test
 
-```shell
-just stream-test //src/workerd/server/tests/container-client:test
-```
\ No newline at end of file
+   ```shell
+   just stream-test //src/workerd/server/tests/container-client
```
anonrig, line 32: Nit: It would be helpful if we could add a command to justfile that runs every single one of these steps

package.json
gpanders: `ws` is used by the container client test image. We need the dependency in the pnpm lock file so that Bazel includes it as a source when building the image.

anonrig: Can we move it to dockerfile creation and just install using a single line "npm install ws@8.18.3 --save-exact" or something?

gpanders: Bazel doesn't use Dockerfiles. We use the `oci_image` rule to build the image, which, like everything in Bazel, requires all of the inputs to be explicitly specified to keep the build hermetic.

If you don't want to include the dependency here, I think we can use pnpm workspaces to have multiple different `package.json` files throughout the monorepo, still represented in a single lock file (which Bazel can automatically parse and use to download all required dependencies).

anonrig: > If you don't want to include the dependency here, I think we can use pnpm workspaces to have multiple different package.json files throughout the monorepo, still represented in a single lock file (which Bazel can automatically parse and use to download all required dependencies).

My thinking was that most developers will not run this test, and rely on CI, and for those people we can avoid installing a dependency.

gpanders: I moved it to a separate package in 76a11ca8. IIUC that should prevent the dependency being downloaded unless the target is referenced directly (which it should not be by default because it has the `requires-container-engine` build tag).

gpanders:
> @gpanders This looks like it should be several PRs – one making test changes, one adding container tests to CI. Easier to review that way. 

This PR only does the latter. The only changes to the tests are those that were required to get them to work when run by Bazel.

> The changes look good for the most part (only took a brief look so far). Do you have numbers for how much overhead this adds? As I understand you already only run this in the test-linux job, but keep in mind that the workers-sdk-test job depends on test-linux and runs after it, so test-linux may be in the critical path for CI duration already.

Using https://github.com/cloudflare/workerd/actions/runs/19862763581/job/56917088984 as a reference, the "Build and load container images step" adds 14s. "Bazel build" took 3m13s and "Bazel test" took 1m5s. Compared to https://github.com/cloudflare/workerd/actions/runs/19832826061/job/56822887495 (which is the last commit on main that this PR is based on right now), where "Bazel build" took 4m7s and "Bazel test" took 1m46s. The large difference is probably due to caching. But it doesn't seem like this PR adds any significant time.

I'll add too that I'm working on this PR at @anonrig's request, who asked that this be done before https://github.com/cloudflare/workerd/pull/5491 is merged. I'll leave it up to you all whether or not you want to run these tests in CI, as long as we can unblock #5491.

danlapid:
> > @gpanders This looks like it should be several PRs – one making test changes, one adding container tests to CI. Easier to review that way.
> 
> This PR only does the latter. The only changes to the tests are those that were required to get them to work when run by Bazel.
> 
> > The changes look good for the most part (only took a brief look so far). Do you have numbers for how much overhead this adds? As I understand you already only run this in the test-linux job, but keep in mind that the workers-sdk-test job depends on test-linux and runs after it, so test-linux may be in the critical path for CI duration already.
> 
> Using [cloudflare/workerd/actions/runs/19862763581/job/56917088984](https://github.com/cloudflare/workerd/actions/runs/19862763581/job/56917088984) as a reference, the "Build and load container images step" adds 14s. "Bazel build" took 3m13s and "Bazel test" took 1m5s. Compared to [cloudflare/workerd/actions/runs/19832826061/job/56822887495](https://github.com/cloudflare/workerd/actions/runs/19832826061/job/56822887495) (which is the last commit on main that this PR is based on right now), where "Bazel build" took 4m7s and "Bazel test" took 1m46s. The large difference is probably due to caching. But it doesn't seem like this PR adds any significant time.
> 
> I'll add too that I'm working on this PR at @anonrig's request, who asked that this be done before #5491 is merged. I'll leave it up to you all whether or not you want to run these tests in CI, as long as we can unblock #5491.

We definitely want these tests in CI.
This PR does not need to be split, it is perfectly reasonable as it is.
We can accept the additional test time required to run these tests.

images/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
+load("@rules_multirun//:defs.bzl", "command", "multirun")
```
fhanau, line 1: We have just a single command being run with multirun() at the moment. I assume the intention is that there will be more such commands in the future which would provide a benefit from using multirun?

build/deps/oci.MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+# rules_oci and rules_multirun are used to build Docker images
+bazel_dep(name = "rules_oci", version = "2.2.6")
```
fhanau: nit: both of these have never versions now. We'll hopefully have auto-update support soon but when we first land these they should have the current version

build/deps/oci.MODULE.bazel
```diff
@@ -0,0 +1,17 @@
...
+bazel_dep(name = "rules_multirun", version = "0.10.0")
+
+oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
+
+# Pull any required base images used to build Docker images
+oci.pull(
+    name = "node_25_slim",
```
fhanau, line 10: How was this image chosen? What does the `linux/arm64/v8` platform mean, like why does one platform say v8 while the other does not?
Not opposed to using this image, but a one-line comment with the rationale for using it would be nice.

images/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
...
+        command = target,
+    )
+    for (name, target) in IMAGES.items()
+]
+
+multirun(
+    name = "load_all",
+    commands = IMAGES.keys(),
+    tags = [
+        "no-remote",
```
fhanau: Why is this no-remote? If no-remote is needed, it should be applied to the underlying target instead, if it is applied here but not to the load command the load command can still be remote cached so it wouldn't have the intended effect.

images/container-client-test/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
+load("@aspect_rules_js//js:defs.bzl", "js_binary", "js_image_layer")
+load("@npm//:defs.bzl", "npm_link_all_packages")
+load("@rules_oci//oci:defs.bzl", "oci_image", "oci_load")
+
+npm_link_all_packages(name = "node_modules")
+
+js_binary(
+    name = "app",
+    chdir = package_name(),
```
fhanau: tiny nit: We don't use chdir anywhere else for js_binary, it seems that running in the top-level directory instead is the more idiomatic approach. Is it needed here?

images/BUILD.bazel
gpanders: Yes that's right. It's not strictly necessary, we could instead run `bazel run //images/container-client-test:load`. But this has the advantage of being a bit more "generic" and won't require changes to the GitHub workflow files if another image is added.

images/container-client-test/BUILD.bazel
gpanders: Doesn't appear to be. I followed the upstream docs for this but will stick with the repo convention and remove it. f99a4c34

images/BUILD.bazel
gpanders: You're right, good catch. I moved it to the `oci_image` target in 5841a497. We do the same thing in the Cloudchamber repo because we found that caching these large tar images isn't worth it (they can be quite large and uploading/downloading them is usually slower than just building them). I can remove the tag completely if you prefer (your call).

build/deps/oci.MODULE.bazel
gpanders: >What does the linux/arm64/v8 platform mean, like why does one platform say v8 while the other does not?

The platform names map to the platforms in the container manifest. If you run `docker manifest inspect node:25-slim` you will see in the list of manifests different platforms, including

```json
"platform": {
  "architecture": "amd64",
  "os": "linux"
}
```

and

```json
"platform": {
  "architecture": "arm64",
  "os": "linux",
  "variant": "v8"
}
```

There are no variants or versions for the AMD 64 architecture, which is why there is no version associated with it.


> Not opposed to using this image, but a one-line comment with the rationale for using it would be nice.

I added a comment in 63dfef95 explaining that this image is used as the base image for `container-client-test`, which uses a NodeJS test application, so this is the smallest Node image that meets all of the necessary requirements for the test.

build/deps/oci.MODULE.bazel
gpanders: Updated in 569726c1

images/BUILD.bazel
fhanau: Thank you for the explanation – I think using no-remote makes sense here, we can go ahead with having it on the oci_image target.

build/deps/oci.MODULE.bazel
fhanau: Sounds good

PR #4423, opened 2025-06-30
Adds web_socket argument to Response in Python SDK.

dom96:
Requested on Discord


danlapid (approved):

dom96:
Ran into an interesting bug here: https://gist.github.com/dom96/831dbf17f2509e47fe0b19831f29463c "SystemError: error return without exception set". It only seems to affect Pyodide 0.26 though. Leaving this here for future reference in case we see it again.

PR #4425, opened 2025-06-30
Internal Build: Handle forks from Cloudflare org members

ketanhwr:


ketanhwr:
> Looks good, but will we have a way to still merge PRs on forks if we verify that internal CI passes manually?

How about if someone adds a label 'Internal Build Passed', it'll pass? From Github docs, looks like only people with 'triage' permissions can assign/remove labels so we should be good?

danlapid:
> > Looks good, but will we have a way to still merge PRs on forks if we verify that internal CI passes manually?
> 
> 
> 
> How about if someone adds a label 'Internal Build Passed', it'll pass? From Github docs, looks like only people with 'triage' permissions can assign/remove labels so we should be good?

Seeing as this is a temporary measure until we can actually run CI on forks I don't think this is worth the effort

PR #4427, opened 2025-06-30
Add base64 internal module and use for AI binding

ns476:
Fixed version of #4420, which broke the internal build as I missed an include


ketanhwr:
(Merging since the internal build passed manually)

PR #4428, opened 2025-06-30
add codecov for code coverage

anonrig:
Requires CODECOV_TOKEN to be set.


build/kj_test.bzl
```diff
@@ -48,20 +47,99 @@ def kj_test(
...
-        srcs = ["//build/fixtures:kj_test.sh"],
-        data = [cross_alias] + data,
-        args = ["$(location " + cross_alias + ")"],
+        binary = cross_alias,
+        data = data,
         tags = tags,
+        env = {"WORKERD_ALL_AUTOGATES": "1"},
+    )
+
+def _kj_test_impl(ctx):
```
fhanau: Why do we need a custom rule for coverage? We've worked with coverage data before and seemed to be able to go without it right?

justfile
```diff
@@ -142,8 +142,8 @@ eslint:
     //types:types_lib@eslint
 
 coverage path="//...":
-  bazel coverage {{path}}
-  genhtml --branch-coverage --output coverage "$(bazel info output_path)/_coverage/_coverage_report.dat"
+  bazel coverage {{path}} --combined_report=lcov --nocache_test_results
```
fhanau: --combined_report defaults to lcov, no need to add the flag. Why do we need to disable caching here

.bazelrc
```diff
@@ -400,16 +400,24 @@ build:windows --per_file_copt=external/boringssl@-Wno-unused-command-line-argume
...
-build:coverage --instrumentation_filter="^//,-//external"
+build:coverage --instrumentation_filter="^//src[:/]"
 build:coverage --instrument_test_targets
+# KJ uses _exit() by default which bypasses atexit handlers and prevents LLVM profile runtime
+# from writing coverage data. KJ_CLEAN_SHUTDOWN forces use of normal exit() instead.
+build:coverage --test_env=KJ_CLEAN_SHUTDOWN=1
+# Coverage processing requires more time, so allow enormous tests
+test:coverage --test_size_filters=
+test:coverage --test_tag_filters=-off-by-default,-requires-fuzzilli,-requires-container-engine,-lint
+test:coverage --test_timeout=60,300,900,3600
```
fhanau: I don't think we should ever allow tests to run for an hour. Downstream has test timeouts of at most 240 seconds, we shouldn't allow more than that.

.bazelrc
```diff
@@ -400,16 +400,24 @@ build:windows --per_file_copt=external/boringssl@-Wno-unused-command-line-argume
...
 build:coverage --experimental_generate_llvm_lcov
 build:coverage --experimental_fetch_all_coverage_outputs
 build:coverage --collect_code_coverage
-build:coverage --instrumentation_filter="^//,-//external"
+build:coverage --instrumentation_filter="^//src[:/]"
 build:coverage --instrument_test_targets
+# KJ uses _exit() by default which bypasses atexit handlers and prevents LLVM profile runtime
+# from writing coverage data. KJ_CLEAN_SHUTDOWN forces use of normal exit() instead.
+build:coverage --test_env=KJ_CLEAN_SHUTDOWN=1
+# Coverage processing requires more time, so allow enormous tests
```
fhanau: tiny nit: the first part of this statement is true, and the second part sounds reasonable too, but 1) doesn't imply 2).

.github/workflows/_bazel.yml
```diff
@@ -133,9 +139,22 @@ jobs:
           docker version
           bazel run --remote_cache=https://bazel:${{ secrets.BAZEL_CACHE_KEY }}@bazel-remote-cache.devprod.cloudflare.dev --config=ci ${{ inputs.extra_bazel_args }} //images:load_all
       - name: Bazel test
-        if: inputs.run_tests
+        if: inputs.run_tests && !inputs.upload_coverage
```
fhanau: nit: I don't think this line needs to change, as long as the coverage build avoids setting run_tests.

.github/workflows/test.yml
```diff
@@ -93,9 +93,11 @@ jobs:
       upload_test_logs: true
       upload_binary: true
       build_container_images: true
+      upload_coverage: true
```
fhanau: The coverage build requires rebuilding everything as it adds additional copts. I don't think it should run as part of the test-linux job, it should be a separate job. Otherwise, we'd expect test-linux (which is already in the critical path frequently) to be significantly slower.

codecov.yml
```diff
@@ -0,0 +1,12 @@
+codecov:
+  require_ci_to_pass: true
+
+coverage:
+  precision: 2
```
fhanau, line 7: What do these settings/values mean? This needs to be documented for it to be maintainable.

.bazelrc
```diff
@@ -400,16 +400,24 @@ build:windows --per_file_copt=external/boringssl@-Wno-unused-command-line-argume
 # set the correct value. We currently don't check __cplusplus, but some dependencies do.
 build:windows --cxxopt='/Zc:__cplusplus' --host_cxxopt='/Zc:__cplusplus'
 
-build:coverage --action_env=GCOV=llvm-profdata-19
 build:coverage --action_env=BAZEL_USE_LLVM_NATIVE_COVERAGE=1
+build:coverage --action_env=GCOV=llvm-profdata-19
+build:coverage --action_env=BAZEL_LLVM_COV=llvm-cov-19
```
fhanau: nit: This assumes that LLVM 19 is available locally, which isn't documented as a requirement and doesn't seem to be available on CI (easy to install though). Maybe it's better to just depend on llvm-cov and expect users to create a softlink to the specific version they are using?
Also it would be nice to document what these defines refer to – BAZEL_LLVM_COV is being used via rules_cc, but developers could easily miss out on that information if they just search the bazel documentation or the bazel code base, they might assume that these variables are unused.

.bazelrc
jasnell: Is the intent to run this on every PR or only nightly on `main` etc? If it's on every PR, I 100% agree with @fhanau that we should never allow for an hour timeout. If we're running coverage in CI only once per night on `main` or only selectively running it locally, then an hour timeout is probably ok?

fhanau: Even if we found a 1h timeout to be acceptable, it would be difficult for CI to finish in time, CI jobs have a maximum duration of 6 hours so a couple such tests (including test variants) could push us over the limit.
I'd suggest the following instead:
1) Let's enable optimizations – we use --copt="-Os" downstream based on disk space constraints, if that's not an issue here I'd hope that we still get good coverage quality with -O1/-O2.
2) If there's still a bad slowdown let's disable this on enormous tests for now, i.e. `--test_size_filters=-enormous --test_timeout=60,240,240`

build/wd_test.bzl
```diff
@@ -117,6 +117,11 @@ WINDOWS_TEMPLATE = """
 @echo off
 setlocal EnableDelayedExpansion
 
+REM Set up coverage profiling for workerd subprocess
```
fhanau: nit: It is commendable that we try to support Windows here, but we probably don't have the resources to regularly test it/lack the Windows expertise to actively maintain coverage support on Windows right? If you want to keep it that's ok with me but I wouldn't worry about Windows support in this case.

types/BUILD.bazel
```diff
@@ -42,6 +42,8 @@ js_run_binary(
         "//:node_modules/prettier",
         "//src/workerd/server:workerd",
     ],
+    # Discard coverage data from workerd subprocess - it's not a test target
+    env = {"LLVM_PROFILE_FILE": "/dev/null"},
```
fhanau: nit: This is just to avoid generating coverage data we don't need right? Would it be useful anywhere else?

src/workerd/api/tests/BUILD.bazel
```diff
@@ -43,6 +43,7 @@ wd_test(
         "jsrpc-timing-test.js",
         "jsrpc-timing-test-tail.js",
     ],
+    tags = ["no-coverage"],
```
fhanau, line 46: nit: I assume this test is just too slow, or is there another reason to exclude it?

README.md
```diff
@@ -121,6 +121,29 @@ version of workerd:
 bazel build --config=thin-lto //src/workerd/server:workerd
 ```
 
+### Code Coverage
+
+To generate code coverage reports, you need LLVM coverage tools installed:
+
+* On Linux, install LLVM 19 and create symlinks:
```
fhanau: nit: A version other than 19 should work too – if we do not reference a specific version here or perhaps say "at least 19" there's no maintenance overhead. It looks like Debian Trixie now defaults to LLVM 19 (https://packages.debian.org/trixie/llvm) – perhaps we can just put `sudo apt-get install llvm` and then the symlinks won't be needed? (We might still need them in CI, where we install llvm a bit differently)

justfile
```diff
@@ -142,8 +142,8 @@ eslint:
     //types:types_lib@eslint
 
 coverage path="//...":
-  bazel coverage {{path}}
-  genhtml --branch-coverage --output coverage "$(bazel info output_path)/_coverage/_coverage_report.dat"
+  bazel coverage {{path}} --nocache_test_results
```
fhanau: nit: I'm still convinced that --nocache_test_results is not needed here, coverage data will still be regenerated if there were test changes/if we previously compiled workerd without coverage enabled.

types/BUILD.bazel
anonrig: Yes this is just for discarding

src/workerd/api/tests/BUILD.bazel
anonrig: It's just too slow

fhanau (approved):

PR #3144, opened 2024-11-20
Instantiate Emscripten Runtime for python workers separately

danlapid:

PR #4434, opened 2025-07-01
Revert most of the work for pyodide isolate pools.

danlapid:
While this work made sense at the time it is growing to be more of a maintenance burden than it's worth. Our runtime is not really built to support isolates with multiple contexts.

Work was never done to actually make use of this in production because we are not currently focusing on performance of python cold starts.

This commit can be reverted in the future if we want to revisit this approach, but for now, it's better to reduce the tech debt by getting rid of it.

Because of other features we are currently developing it is likely that we'll have a better solution to python cold starts that doesn't require multiple contexts and would eliminate the need for this anyway.

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -544,12 +544,11 @@ void DiskCache::put(jsg::Lock& js, kj::String key, kj::Array<kj::byte> data) {
 
 jsg::JsValue SetupEmscripten::getModule(jsg::Lock& js) {
   js.installJspi();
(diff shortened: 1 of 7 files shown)


dom96 (approved):
Looks good. Not in this PR, but I wonder if we should follow up by removing SetupEmscripten (and EmscriptenRuntime) and go back to calling `initialiseEmscriptenModule` directly (like we used to here: https://github.com/cloudflare/workerd/pull/3144/files#diff-b4a6e1839f4e02b6251f531344daaa2660caa17b28950343783d4f7c1bb88123L96). i.e. the less C++ the better.

PR #4435, opened 2025-07-01
add a test for dynamic import in handler with the new registry

vicb:
As discussed offline

### DIFF
--- a/src/workerd/api/tests/new-module-registry-test.js
+++ b/src/workerd/api/tests/new-module-registry-test.js
@@ -208,6 +208,37 @@ export const wasmModuleTest = {
   },
 };
 
+const testDynamicImportInHandler = {
+  async test(_ctrl, { handler }) {
+    const res = await handler.fetch('http://test.me/dynamic-import');
+    strictEqual(await res.text(), 'threw');
+  },
+  async doTest() {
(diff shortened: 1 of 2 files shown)


src/workerd/api/tests/new-module-registry-test.js
```diff
@@ -208,6 +208,37 @@ export const wasmModuleTest = {
...
+  async doTest() {
+    try {
+      // Modules are not allowed to do top level I/O even when imported inside an handler
+      await import('bad');
+      return new Response('imported');
+    } catch {
+      return new Response('threw');
+    }
+  },
+};
```
jasnell, line 225: This test already has coverage for this:

```js
export const noTopLevelIo = {
  async test() {
    await rejects(import('bad'), {
      message: /^Disallowed operation called within global scope/,
    });
  },
};
```

src/workerd/api/tests/new-module-registry-test.js
```diff
@@ -208,6 +208,37 @@ export const wasmModuleTest = {
...
+
+/**
+ * Run tests inside a request handler.
+ */
```
jasnell, line 229: All tests already run within an IoContext.

PR #4436, opened 2025-07-01
Fixup node:crypto signature lengths

jasnell:
Refs: https://github.com/axefinaxe/cfbug

### DIFF
--- a/src/workerd/api/node/tests/crypto_sign-test.js
+++ b/src/workerd/api/node/tests/crypto_sign-test.js
@@ -3,11 +3,12 @@ import {
   createVerify,
   createPrivateKey,
   createPublicKey,
+  generateKeyPairSync,
   sign,
   verify,
 } from 'node:crypto';
 
-import { strictEqual, throws } from 'node:assert';
+import { ok, strictEqual, throws } from 'node:assert';
 
 const rsaSig =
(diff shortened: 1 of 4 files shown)


src/workerd/api/node/tests/crypto_sign-test.js
```diff
@@ -144,3 +145,37 @@ export const dsaSignVerifyObjects = {
...
+        type: 'spki',
+        format: 'pem',
+      },
+      privateKeyEncoding: {
+        type: 'pkcs8',
+        format: 'pem',
+      },
+    });
+
+    for (let n = 0; n < 1000; n++) {
```
vicb, line 166: isn't this testing 1000 times the same thing?
_(should `const message =...` be moved inside the loop)_

vicb: Alternatively should we find **offline** a single value (/a few values) that are known to fail and hardcode them in the test?

anonrig (approved):

src/workerd/api/node/tests/crypto_sign-test.js
jasnell: Doesn't really matter. The signature length can vary regardless of the message... that is, it's not deterministic. With identical input, when DER output is selected, the signature can vary from between 68-72 bytes in length

Issue #4236, opened 2025-06-03
WorkerEntrypoint::test fails to print abort message


npaun:
If a worker is aborted due to a hanging promise, `IoContext::abortFromHang()` is invoked. This causes the error "The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response."

Most of the time, this error bubbles up as expected. For instance, it definitely works in `WorkerEntrypoint::fetch()` but for some reason it's not working right in `WorkerEntrypoint::test()`. This causes tests to appear as failed, but with no explanation of why. There are currently three affected tests in the WPT streams module.

PR #4437, opened 2025-07-01
propagate abort errors correctly

anonrig:
On a test case like this:

```js
export const whyGodWhy = {
  async test(_ctrl, env, ctx) {
    const { promise } = Promise.withResolvers();
    return promise;
  },
};
```

We will have the following error from now on:

```
workerd/server/server.c++:5194: debug: [ TEST ] leak-fetch-test:whyGodWhy
workerd/io/io-context.c++:377: info: uncaught exception; exception = workerd/io/io-context.c++:1296: failed: jsg.Error: The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response. Refer to: https://developers.cloudflare.com/workers/observability/errors/
stack: src/workerd/server/workerd@36a7350
decltype(reducePromiseType((workerd::RemoveIoOwn_<void>::Type*)(nullptr), false)) workerd::IoContext::awaitJs<void>(workerd::jsg::Lock&, workerd::jsg::Promise<void>)::'lambda'()::operator()() const
bazel-out/k8-dbg/bin/src/workerd/io/_virtual_includes/io/workerd/io/io-context.h:1435:0


workerd/server/server.c++:5208: debug: [ FAIL ] leak-fetch-test:whyGodWhy (2.408259593s)
```

Fixes https://github.com/cloudflare/workerd/issues/4236

### DIFF
--- a/src/workerd/io/worker-entrypoint.c++
+++ b/src/workerd/io/worker-entrypoint.c++
@@ -729,6 +729,21 @@ kj::Promise<bool> WorkerEntrypoint::test() {
       [](IoContext& context, kj::Own<IoContext::IncomingRequest> request) -> kj::Promise<bool> {
     TRACE_EVENT("workerd", "WorkerEntrypoint::test() waitForFinished()");
     auto result = co_await request->finishScheduled();
+
+    if (result == IoContext_IncomingRequest::FinishScheduledResult::ABORTED) {
+      // If the test handler throws an exception (without aborting - just a regular exception),
+      // then `outcome` ends up being EventOutcome::EXCEPTION, which causes us to return false.
+      // But in that case we are separately relying on the exception being logged as an uncaught
+      // exception, rather than throwing it.
+      // This is why we don't rethrow the exception but rather log it as an uncaught exception.
+      try {
+        co_await context.onAbort();
+      } catch (...) {
+        auto exception = kj::getCaughtExceptionAsKj();
+        KJ_LOG(ERROR, exception);
+      }
+    }
+
     bool completed = result == IoContext_IncomingRequest::FinishScheduledResult::COMPLETED;
     auto outcome = completed ? context.waitUntilStatus() : EventOutcome::EXCEEDED_CPU;
     co_return outcome == EventOutcome::OK;


src/workerd/io/worker-entrypoint.c++
```diff
@@ -729,6 +729,19 @@ kj::Promise<bool> WorkerEntrypoint::test() {
...
+
+    if (result == IoContext_IncomingRequest::FinishScheduledResult::ABORTED) {
+      // The test was aborted. Since onAbort() should already be rejected at this point,
+      // we can directly try to get the exception from it.
+      try {
+        co_await context.onAbort();
+        // If we reach here, onAbort() wasn't rejected, which is unexpected
+        KJ_FAIL_ASSERT("test() handler aborted but onAbort() was not rejected");
+      } catch (kj::Exception& e) {
+        KJ_EXCEPTION(FAILED, kj::mv(e));
```
kentonv: This line doesn't do anything. KJ_EXCEPTION evaluates to an exception but doesn't throw it.

Why not remove the try/catch entirely, and just let the abort exception flow through?

anonrig: This has a bigger stack trace, but yeah, the code is simpler.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -729,6 +729,11 @@ kj::Promise<bool> WorkerEntrypoint::test() {
       [](IoContext& context, kj::Own<IoContext::IncomingRequest> request) -> kj::Promise<bool> {
     TRACE_EVENT("workerd", "WorkerEntrypoint::test() waitForFinished()");
     auto result = co_await request->finishScheduled();
+
+    if (result == IoContext_IncomingRequest::FinishScheduledResult::ABORTED) {
+      co_await context.onAbort();
```
kentonv: I would add a comment:

> Force the abort exception to be rethrown from the test handler. This is important because the abort exception is not otherwise logged as an uncaught exception elsewhere, so if we don't rethrow it, the test log won't actually say why the test failed (although the test will still fail, by returning `false` below).

src/workerd/io/worker-entrypoint.c++
```diff
@@ -729,6 +729,11 @@ kj::Promise<bool> WorkerEntrypoint::test() {
...
+    }
+
     bool completed = result == IoContext_IncomingRequest::FinishScheduledResult::COMPLETED;
     auto outcome = completed ? context.waitUntilStatus() : EventOutcome::EXCEEDED_CPU;
```
kentonv, line 748: Hmm, I guess if the test handler throws an exception (without aborting -- just a regular exception), then `outcome` ends up being `EventOutcome::EXCEPTION`, which causes us to return false... but in that case we are separately relying on the exception being logged as an uncaught exception, rather than rethrowing it.

So it's a bit weird that for _abort_ exceptions specifically, we rethrow the exception, but for regular thrown exceptions, we log the exception elsewhere and return false?

This sort of suggests that, for consistency, we should either:
- Don't rethrow the abort exception, but rather log it as an uncaught exception.
- Try to rethrow regular exceptions, rather than rely on them being thrown elsewhere.

With all that said, I guess it doesn't _really_ matter than much. As long as the test fails when it should, and the error gets logged somehow, then we're good. It's just sorta weird.

So I guess I'll approve this as-is...

anonrig: > Don't rethrow the abort exception, but rather log it as an uncaught exception.

I think this is a better approach. I'll update my implementation. Thank you for your review.

src/workerd/io/worker-entrypoint.c++
anonrig: I've added a comment taken from your review response.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -729,6 +729,20 @@ kj::Promise<bool> WorkerEntrypoint::test() {
...
+
+    if (result == IoContext_IncomingRequest::FinishScheduledResult::ABORTED) {
+      // If the test handler throws an exception (without aborting - just a regular exception),
+      // then `outcome` ends up being EventOutcome::EXCEPTION, which causes us to return false.
+      // But in that case we are separately relying on the exception being logged as an uncaught
+      // exception, rather than throwing it.
+      // This is why we don't rethrow the exception but rather log it as an uncaught exception.
+      try {
+        co_await context.onAbort();
+      } catch (kj::Exception& e) {
```
kentonv: This is not the correct way to catch a KJ exception. You need to write:

```
} catch (...) {
  auto e = kj::getCaughtExceptionAsKj();
```

Otherwise, stack traces will be wrong, and std exceptions won't be caught.

src/workerd/io/worker-entrypoint.c++
```diff
@@ -729,6 +729,21 @@ kj::Promise<bool> WorkerEntrypoint::test() {
...
+      // If the test handler throws an exception (without aborting - just a regular exception),
+      // then `outcome` ends up being EventOutcome::EXCEPTION, which causes us to return false.
+      // But in that case we are separately relying on the exception being logged as an uncaught
+      // exception, rather than throwing it.
+      // This is why we don't rethrow the exception but rather log it as an uncaught exception.
+      try {
+        co_await context.onAbort();
+      } catch (...) {
+        auto e = kj::getCaughtExceptionAsKj();
+        LOG_EXCEPTION("test", e);
```
kentonv: I guess arguably this should be `KJ_LOG(ERROR` -- `LOG_EXCEPTION` specifically logs exceptions formatted for sentry, but this isn't intended for sentry.

anonrig: Thanks! Updated.

PR #4444, opened 2025-07-02
Implements support for jaeger tracing at top-level for Python.

dom96:
Python Workers need to be initialised at the top-level, but the IoContext is not available in that case. The Jaeger tracing grabs the current request's span via the IoContext. This PR (and the upstream PR) implements an alternative path of grabbing a span, by passing it through to InternalJaegerTrace when it is created. This enables Python to log spans at the top-level.

This enables us to remove the last special code path for Python initialisation (which had a number of limitations) and simplifies Python init significantly.

### DIFF
--- a/src/pyodide/internal/jaeger.ts
+++ b/src/pyodide/internal/jaeger.ts
@@ -1,10 +1,11 @@
 import { default as internalJaeger } from 'pyodide-internal:internalJaeger';
+import { IS_TRACING } from 'pyodide-internal:metadata';
 
 /**
  * Used for tracing via Jaeger.
  */
 export function enterJaegerSpan<T>(span: string, callback: () => T): T {
-  if (!internalJaeger.traceId) {
+  if (!IS_TRACING || !internalJaeger.traceId) {
     // Jaeger tracing not enabled or traceId is not present in request.
     return callback();
   }
--- a/src/pyodide/internal/metadata.ts
+++ b/src/pyodide/internal/metadata.ts
@@ -39,8 +39,6 @@ export const TRANSITIVE_REQUIREMENTS =
 
 // Entrypoints
 export const MAIN_MODULE_NAME = MetadataReader.getMainModule();
-export const DURABLE_OBJECT_CLASSES = MetadataReader.getDurableObjectClasses();
-export const WORKER_ENTRYPOINT_CLASSES = MetadataReader.getEntrypointClasses();
 
 export interface CompatibilityFlags {
   python_workflows?: boolean;
--- a/src/pyodide/python-entrypoint-helper.ts
+++ b/src/pyodide/python-entrypoint-helper.ts
@@ -6,14 +6,11 @@ import { loadPyodide } from 'pyodide-internal:python';
 import { enterJaegerSpan } from 'pyodide-internal:jaeger';
 import { patchLoadPackage } from 'pyodide-internal:setupPackages';
 import {
-  IS_TRACING,
   IS_WORKERD,
   LOCKFILE,
   TRANSITIVE_REQUIREMENTS,
   MAIN_MODULE_NAME,
   WORKERD_INDEX_URL,
-  DURABLE_OBJECT_CLASSES,
-  WORKER_ENTRYPOINT_CLASSES,
   SHOULD_SNAPSHOT_TO_DISK,
   workflowsEnabled,
 } from 'pyodide-internal:metadata';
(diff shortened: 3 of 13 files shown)


src/pyodide/internal/metadata.ts
```diff
@@ -39,5 +39,6 @@ export const TRANSITIVE_REQUIREMENTS =
 
 // Entrypoints
 export const MAIN_MODULE_NAME = MetadataReader.getMainModule();
+// TODO: Remove the below.
```
hoodmane: Can you explain the reason they should be removed here? Not obvious to me.

hoodmane: Are they still used in Edgeworker but not workerd?

src/pyodide/python-entrypoint-helper.ts
```diff
@@ -385,57 +382,28 @@ export async function initPython(): Promise<PythonInitResult> {
     workflowEntrypoints: [],
   };
 
-  // Do not setup anything to do with Python in the global scope when tracing. The Jaeger tracing
```
hoodmane, line 398: So much beautiful red. The diff should really swap the two colors and make added code red and removed code green =)

src/pyodide/internal/jaeger.ts
```diff
@@ -1,13 +1,14 @@
...
  */
 export function enterJaegerSpan<T>(span: string, callback: () => T): T {
-  if (!internalJaeger.traceId) {
+  if (!IS_TRACING || !internalJaeger.traceId) {
     // Jaeger tracing not enabled or traceId is not present in request.
     return callback();
   }
 
-  return internalJaeger.enterSpan(span, callback);
+  return internalJaeger.toplevelEnterSpan(span, callback);
```
hoodmane: Does this work when not at top level?

src/pyodide/internal/metadata.ts
dom96: They are no longer used by anything

src/pyodide/internal/jaeger.ts
dom96: Yep, it falls back to `enterSpan`

src/pyodide/internal/metadata.ts
hoodmane: Why not delete them instead of adding a TODO?

dom96: Because I haven't got around to it yet :)

dom96: Now done

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -409,13 +409,7 @@ PyodideMetadataReader::State::State(const State& other)
       isWorkerdFlag(other.isWorkerdFlag),
       isTracingFlag(other.isTracingFlag),
       snapshotToDisk(other.snapshotToDisk),
-      createBaselineSnapshot(other.createBaselineSnapshot),
-      memorySnapshot(other.memorySnapshot.map(
```
danlapid, line 413: Was memorySnapshot deleted on purpose?

src/workerd/io/worker.h
```diff
@@ -341,7 +343,8 @@ class Worker::Isolate: public kj::AtomicRefcounted {
       IsolateObserver::StartType startType,
       bool logNewScript = false,
       kj::Maybe<ValidationErrorReporter&> errorReporter = kj::none,
-      kj::Maybe<kj::Own<api::pyodide::ArtifactBundler_State>> artifacts = kj::none) const;
+      kj::Maybe<kj::Own<api::pyodide::ArtifactBundler_State>> artifacts = kj::none,
+      SpanParent parentSpan = nullptr) const;
```
danlapid: I don't think the code actually handles this being a nullptr correctly and we might end up with a null ptr deref.

Are we just giving it a default value because of its placement in the argument list even though it should never be null?
if so, let's move it's location.

danlapid:
LFG

src/workerd/api/pyodide/pyodide.c++
dom96: good catch

src/workerd/io/worker.h
dom96: Ahh yeah, I was meaning to change this so it doesn't have a default value. I updated each call site to explicitly pass a SpanParent, in some cases `SpanParent(nullptr)` is the right value to pass in though and that is supported explicitly by `SpanParent`.

danlapid (approved):

PR #4445, opened 2025-07-02
Fixes Durable Object test on Pyodide 0.27.7

dom96:


dom96:
@hoodmane for this to work we need to update Pyodide to include my patch which adds custom fetcher support for pyfetch.

PR #4446, opened 2025-07-02
NodeJS Compat: Process Compat via Separate Implementation

guybedford:
This reincludes the unrevert of the original PR from https://github.com/cloudflare/workerd/pull/4271 and all of its features, along with the unrevert of the compatibility flag introduced in https://github.com/cloudflare/workerd/pull/4350.

Then to ensure compatibility, instead of exporting undefined named exports for process features when not using the `enable_nodejs_process_v2` compatibility flag, two entirely separate process implementations are provided as internal modules - `node-internal:public_process` for the new implementation and `node-internal:legacy_process` for the former implementation. Implementation internals are then shared as suitable through the internal process module as well.

Then in the resolver, we detect `node:process` and `process` imports and direct those to either `node-internal:public_process` or `node-internal:legacy_process` depending on the compat flag. `process.getBuiltinModule` also has a similar switch logic along with the global getter function for `globalThis.process`.

Tests are added for the legacy process module to ensure it does not contain unexpected properties. In this way, full compatibility is maintained for all users not using the new process implementation.

New process features remain identical to the original summary in https://github.com/cloudflare/workerd/pull/4271.

Stdio and FS integrations remain pending for removing experimental.

While working on this I noticed the Node.js LTS version already bumped a minor - Claude was able to cook up a script to bump this automatically for us which is also added here as `just update-reported-node-version` for convenience.


src/node/internal/public_process.ts
```diff
@@ -0,0 +1,408 @@
...
+}
+
+export function initgroups(
+  _user: number | string,
+  _extractGroup: number | string
+): void {
+  throw new ERR_EPERM({ syscall: 'initgroups' });
+}
+
+interface Process extends EventEmitter {
```
anonrig, line 239: Can you use the type from types/node package rather than defining our own?

guybedford: If I do that, I get this error:

```
Argument of type 'Process' is not assignable to parameter of type 'EventEmitter'.
  Property '[kRejection]' is missing in type 'Process' but required in type 'EventEmitter'.ts(2345)
events.ts(87, 3): '[kRejection]' is declared here.
```

because our event emitter has a non Node.js property typed.

src/node/internal/internal_errors.ts
```diff
@@ -862,6 +862,13 @@ export class ERR_EEXIST extends NodeSyscallError {
   }
 }
 
+export class ERR_EPERM extends NodeSyscallError {
+  constructor(options: { syscall: string; errno?: number }) {
+    super('EPERM', 'Operation not permitted', options.syscall);
+    this.errno = options.errno || 1; // EPERM
```
anonrig:
```suggestion
    this.errno = options.errno ?? 1; // EPERM
```

src/workerd/io/compatibility-date.capnp
```diff
@@ -838,4 +839,20 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
...
+      $compatEnableFlag("unsupported_process_actual_platform")
+      $experimental;
+  # By default, Workerd will always expose "linux" as the process.platform.
+  # This flag enables support for process.platform to expose the actual system platform.
+  # This is unsupported, as this feature will never ever be supported as non-experimental and is a
+  # temporary WPT test path only.
+
+  enableNodeJsProcessV2 @97 :Bool
+      $compatEnableFlag("enable_nodejs_process_v2")
+      $experimental;
```
jasnell, line 853: I think it's likely safe for this one to have a default enable date and likely a `impliedByAfterDate` so that it's just the default when `nodejs_compat` is enabled. Typically we'd select a date that is at least a few weeks out from when we expect this to hit production.

guybedford: I'd like to aim make this update once we've finished up the FS related (chdir etc as commented) and stdio support side here. Will post the stdio PR soon, then perhaps we can figure out a timeline for the FS related methods and get the dates together.

PR #4452, opened 2025-07-02
improve performance of recursiveFreeze

anonrig:
Use iteration instead of recursion with a queue to avoid stack overflow and improve cache locality.

### DIFF
--- a/src/workerd/jsg/util.c++
+++ b/src/workerd/jsg/util.c++
@@ -575,7 +575,6 @@ void recursivelyFreeze(v8::Local<v8::Context> context, v8::Local<v8::Value> valu
   v8::HandleScope outerScope(isolate);
   kj::Vector<v8::Local<v8::Value>> queue;
   queue.reserve(128);
-  kj::HashSet<int> visitedHashes;
   queue.add(value);
 
   while (!queue.empty()) {
@@ -584,10 +583,6 @@ void recursivelyFreeze(v8::Local<v8::Context> context, v8::Local<v8::Value> valu
 
     if (item->IsArray()) {
       auto arr = item.As<v8::Array>();
-      int hash = arr->GetIdentityHash();
-      if (visitedHashes.contains(hash)) continue;
-      visitedHashes.insert(hash);
-
       uint32_t length = arr->Length();
 
       // Process array elements in batches to reduce V8 API call overhead
@@ -607,22 +602,6 @@ void recursivelyFreeze(v8::Local<v8::Context> context, v8::Local<v8::Value> valu
       check(arr->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
     } else if (item->IsObject()) {
       auto obj = item.As<v8::Object>();
-
-      int hash = obj->GetIdentityHash();
-      if (visitedHashes.contains(hash)) continue;
-      visitedHashes.insert(hash);
-
-      // Fast path: skip if object is already frozen
-      v8::PropertyAttribute attrs;
-      if (obj->GetRealNamedPropertyAttributes(
-                 context, v8::String::NewFromUtf8Literal(isolate, "__proto__"))
-              .To(&attrs)) {
-        if ((attrs & v8::DontDelete) && (attrs & v8::ReadOnly)) {
-          // Likely already frozen, verify with a sample property check
-          continue;
-        }
-      }
-
       auto names = check(obj->GetPropertyNames(context, v8::KeyCollectionMode::kOwnOnly,
           v8::ALL_PROPERTIES, v8::IndexFilter::kIncludeIndices));
       if (!names.IsEmpty()) {
--- a/src/workerd/jsg/util.c++
+++ b/src/workerd/jsg/util.c++
@@ -570,58 +570,39 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
   }
 }
 
-void recursivelyFreeze(v8::Local<v8::Context> context, v8::Local<v8::Value> value) {
-  auto isolate = v8::Isolate::GetCurrent();
-  v8::HandleScope outerScope(isolate);
-  kj::Vector<v8::Local<v8::Value>> queue;
-  queue.reserve(128);
-  queue.add(value);
+void recursivelyFreeze(jsg::Lock& js, v8::Local<v8::Value> value) {
+  v8::HandleScope outerScope(js.v8Isolate);
+  v8::LocalVector<v8::Value> queue(js.v8Isolate);
+  queue.push_back(value);
 
   while (!queue.empty()) {
     auto item = queue.back();
-    queue.removeLast();
+    queue.pop_back();
 
     if (item->IsArray()) {
       auto arr = item.As<v8::Array>();
       uint32_t length = arr->Length();
-
-      // Process array elements in batches to reduce V8 API call overhead
-      constexpr uint32_t BATCH_SIZE = 32;
-      for (uint32_t i = 0; i < length;) {
-        uint32_t batchEnd = kj::min(i + BATCH_SIZE, length);
-        queue.reserve(queue.size() + (batchEnd - i));
-
-        for (; i < batchEnd; ++i) {
-          auto element = check(arr->Get(context, i));
-          if (!element->IsNullOrUndefined()) {
-            queue.add(element);
-          }
+      for (uint32_t i = 0; i < length; i++) {
+        auto prop = check(arr->Get(js.v8Context(), i));
+        if (prop->IsArray() || prop->IsObject()) {
+          queue.push_back(prop);
         }
       }
 
-      check(arr->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
+      check(arr->SetIntegrityLevel(js.v8Context(), v8::IntegrityLevel::kFrozen));
     } else if (item->IsObject()) {
       auto obj = item.As<v8::Object>();
-      auto names = check(obj->GetPropertyNames(context, v8::KeyCollectionMode::kOwnOnly,
+      auto names = check(obj->GetPropertyNames(js.v8Context(), v8::KeyCollectionMode::kOwnOnly,
           v8::ALL_PROPERTIES, v8::IndexFilter::kIncludeIndices));
-      if (!names.IsEmpty()) {
-        uint32_t length = names->Length();
-
-        constexpr uint32_t BATCH_SIZE = 16;
-        for (uint32_t i = 0; i < length;) {
-          uint32_t batchEnd = std::min(i + BATCH_SIZE, length);
-          queue.reserve(queue.size() + (batchEnd - i));
-
-          for (; i < batchEnd; ++i) {
-            auto propValue = check(obj->Get(context, check(names->Get(context, i))));
-            if (!propValue->IsNullOrUndefined()) {
-              queue.add(propValue);
-            }
-          }
+      uint32_t length = names->Length();
+      for (uint32_t i = 0; i < length; i++) {
+        auto prop = check(obj->Get(js.v8Context(), check(names->Get(js.v8Context(), i))));
+        if (prop->IsArray() || prop->IsObject()) {
+          queue.push_back(prop);
         }
       }
 
-      check(obj->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
+      check(obj->SetIntegrityLevel(js.v8Context(), v8::IntegrityLevel::kFrozen));
     }
 
     // Primitive types don't need freezing
--- a/src/workerd/jsg/util.c++
+++ b/src/workerd/jsg/util.c++
@@ -10,6 +10,7 @@
 #include <openssl/rand.h>
 
 #include <kj/debug.h>
+#include <kj/hash.h>
 
 #include <cstdlib>
 #include <set>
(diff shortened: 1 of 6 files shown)


src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,81 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
+      // Fast path: skip if object is already frozen
+      v8::PropertyAttribute attrs;
+      if (obj->GetRealNamedPropertyAttributes(
+                 context, v8::String::NewFromUtf8Literal(v8::Isolate::GetCurrent(), "__proto__"))
+              .To(&attrs)) {
+        if ((attrs & v8::DontDelete) && (attrs & v8::ReadOnly)) {
+          // Likely already frozen, verify with a sample property check
+          continue;
+        }
+      }
```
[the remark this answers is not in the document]
anonrig: >       v8::Local<v8::Boolean> isFrozen = check(v8::Object::IsFrozen(obj));

Thank you AI for randomly creating functions that doesn't exist in v8.

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,81 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
+      // Fast path: skip if object is already frozen
+      v8::PropertyAttribute attrs;
+      if (obj->GetRealNamedPropertyAttributes(
+                 context, v8::String::NewFromUtf8Literal(isolate, "__proto__"))
+              .To(&attrs)) {
+        if ((attrs & v8::DontDelete) && (attrs & v8::ReadOnly)) {
+          // Likely already frozen, verify with a sample property check
+          continue;
+        }
+      }
```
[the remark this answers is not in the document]
anonrig: There is no method called IsExtensible

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,81 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
+      auto obj = item.As<v8::Object>();
+
+      int hash = obj->GetIdentityHash();
+      if (visitedHashes.contains(hash)) continue;
+      visitedHashes.insert(hash);
+
+      // Fast path: skip if object is already frozen
+      v8::PropertyAttribute attrs;
+      if (obj->GetRealNamedPropertyAttributes(
+                 context, v8::String::NewFromUtf8Literal(isolate, "__proto__"))
```
kentonv: Use `obj->GetPrototypeV2()` here, it'll be much faster than constructing the string `__proto__` and then matching it to a property.

kentonv: Oh wait you are getting the _attributes_ of the property... maybe I don't understand this.

kentonv:
> Use iteration instead of recursion with a queue to avoid stack overflow and improve cache locality.

Is it actually faster? My guess would be it is slower. Did you measure?

Can we just enforce a recursion depth limit on this instead?

src/workerd/jsg/util.c++
anonrig: Actually __proto__ is not a reliable way of detecting frozenness. I'll just remove it.

kentonv:
I'm particularly concerned that this algorithm may create unlimited handles within a single HandleScope.

anonrig:
> Is it actually faster? My guess would be it is slower. Did you measure?

I'll update this PR with a benchmark.

anonrig:
> Can we just enforce a recursion depth limit on this instead?

This PR also fixes cyclic references in the same object.

kentonv:
> This PR also fixes cyclic references in the same object.

AFAICT we use this function is one single place in the codebase and it's on an object that is intended to be JSON-serialized later, hence cycles are not allowed anyway.

anonrig:
> > This PR also fixes cyclic references in the same object.
> 
> AFAICT we use this function is one single place in the codebase and it's on an object that is intended to be JSON-serialized later, hence cycles are not allowed anyway.

You're right. No need to check for hashes.

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
+      auto arr = item.As<v8::Array>();
+      uint32_t length = arr->Length();
+
+      // Process array elements in batches to reduce V8 API call overhead
+      constexpr uint32_t BATCH_SIZE = 32;
+      for (uint32_t i = 0; i < length;) {
+        uint32_t batchEnd = kj::min(i + BATCH_SIZE, length);
+        queue.reserve(queue.size() + (batchEnd - i));
+
+        for (; i < batchEnd; ++i) {
```
erikcorry: How does this reduce overhead?

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
 }
 
 void recursivelyFreeze(v8::Local<v8::Context> context, v8::Local<v8::Value> value) {
-  if (value->IsArray()) {
-    // Optimize array freezing (Array is a subclass of Object, but we can iterate it faster).
-    v8::HandleScope scope(v8::Isolate::GetCurrent());
```
erikcorry, line 575: If there is an improvement, then I guess it mainly comes from removing HandleScopes?

src/workerd/jsg/util.c++
anonrig: We don't increment the capacity of the queue one by one, but rather in batches.

anonrig:
It seems codspeed has a bug.

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
-    auto arr = value.As<v8::Array>();
-
-    for (auto i: kj::zeroTo(arr->Length())) {
-      recursivelyFreeze(context, check(arr->Get(context, i)));
-    }
+  auto isolate = v8::Isolate::GetCurrent();
+  v8::HandleScope outerScope(isolate);
+  kj::Vector<v8::Local<v8::Value>> queue;
```
jasnell: This really ought to be using `v8::LocalVector`

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
+        }
+      }
 
-    check(arr->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
-  } else if (value->IsObject()) {
-    v8::HandleScope scope(v8::Isolate::GetCurrent());
-    auto obj = value.As<v8::Object>();
-    auto names = check(obj->GetPropertyNames(context, v8::KeyCollectionMode::kOwnOnly,
-        v8::ALL_PROPERTIES, v8::IndexFilter::kIncludeIndices));
+      check(arr->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
```
jasnell: @erikcorry ... do you happen to know if `Array::Iterate` (https://v8docs.nodesource.com/node-24.1/d3/d32/classv8_1_1_array.html#aa56a09c66abbd70a384070d00bbdc832) would be possible to use here? I'm not sure if the `SetIntegrityLevel` and `IsNullOrUndefined` calls would violate the restrictions on that.  Mostly just curious because I haven't actually had occasion to use `Array::Iterate` yet.

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
 }
 
 void recursivelyFreeze(v8::Local<v8::Context> context, v8::Local<v8::Value> value) {
```
jasnell: I think all uses of this go through the `jsg::Lock` now, or at least should be. Let's modify this to accept the `v8::Isolate*` as the first argument and have the call routed in through `jsg::Lock` pass in the `v8Isolate`. that'll save us from having to do the additional `Isolate::GetCurrent()`.

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
-    }
+  auto isolate = v8::Isolate::GetCurrent();
+  v8::HandleScope outerScope(isolate);
+  kj::Vector<v8::Local<v8::Value>> queue;
+  queue.reserve(128);
+  queue.add(value);
+
+  while (!queue.empty()) {
+    auto item = queue.back();
+    queue.removeLast();
```
jasnell: I think this might be a problem with the original too, but how is this going to handle circular references?

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
+
+    if (item->IsArray()) {
```
jasnell, line 582: An Array can also have named properties, do we care? Looks like the original version ignored those.

src/workerd/jsg/util.c++
anonrig: It won't handle circular references. The util.h has a comment saying that it's ok to ignore circular references. This method is called on a data that is parsed and stringified, hence eliminating circular references.

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
-    v8::HandleScope scope(v8::Isolate::GetCurrent());
-    auto obj = value.As<v8::Object>();
-    auto names = check(obj->GetPropertyNames(context, v8::KeyCollectionMode::kOwnOnly,
-        v8::ALL_PROPERTIES, v8::IndexFilter::kIncludeIndices));
+      check(arr->SetIntegrityLevel(context, v8::IntegrityLevel::kFrozen));
+    } else if (item->IsObject()) {
+      auto obj = item.As<v8::Object>();
+      auto names = check(obj->GetPropertyNames(context, v8::KeyCollectionMode::kOwnOnly,
+          v8::ALL_PROPERTIES, v8::IndexFilter::kIncludeIndices));
+      if (!names.IsEmpty()) {
```
jasnell: Can `names.IsEmpty()` be false here? That would would suggest that an exception was scheduled which should have been caught by `check()` above.

anonrig: If it's an empty object, I don't see why not...

src/workerd/jsg/util.c++
```diff
@@ -570,29 +571,60 @@ kj::Array<kj::byte> asBytes(v8::Local<v8::ArrayBufferView> arrayBufferView) {
...
+
+        constexpr uint32_t BATCH_SIZE = 16;
+        for (uint32_t i = 0; i < length;) {
+          uint32_t batchEnd = std::min(i + BATCH_SIZE, length);
+          queue.reserve(queue.size() + (batchEnd - i));
+
+          for (; i < batchEnd; ++i) {
+            auto propValue = check(obj->Get(context, check(names->Get(context, i))));
+            if (!propValue->IsNullOrUndefined()) {
+              queue.add(propValue);
```
jasnell: Since we only actually care about `Array` and `Object`, it would likely be more efficient to filter out non-Array, and non-Object values when adding to the queue, that way we can avoid unnecessary additional allocations via the queue reservations.

src/workerd/jsg/util.c++
kentonv: `IsEmpty()` doesn't mean the object is empty, it means the handle is null.

kentonv: It makes me nervous that you're writing this code without enough understanding of how the V8 API works.

Can we please just add a simple recursion limit on top of the existing code, and not otherwise change it?

anonrig: > It makes me nervous that you're writing this code without enough understanding of how the V8 API works.

It's a simple mix up. It doesn't mean anything.

> Can we please just add a simple recursion limit on top of the existing code, and not otherwise change it?

I'll get to it, I'm trying to setup Codspeed (benchmarking) correctly so we can compare it with the current implementation. Only force pushed to get the changes in main branch.

jasnell: Specifically, the implementation of `jsg::check(...)` will check whether the returned `MaybeLocal<Array>` from `GetPropertyNames(...)` is empty, arranging to throw/propagate the exception if it is, and it returns the `Local<Array>`, which should never actually be `IsEmpty()` here or there's a deeper fundamental problem in v8.

anonrig: Yes, there's no need for .IsEmpty() check. I thought `names` was a std::vector of some sort and mixed up the function names/variable types.

src/workerd/jsg/util.c++
jasnell: that's what I figured. Let's add a comment in here mentioning that.

src/workerd/jsg/util.c++
jasnell: I think the same argument for circular references applies here too. If we're getting this from a parsed json structure then no named properties will matter. A comment in here along those lines is helpful for future generations.

src/workerd/jsg/util.c++
kentonv: The input could have circular references. Such input is _invalid_, because it will later attempt to be serialized to JSON. So it's OK if we throw an exception in that case, but we can't completely ignore it.

jasnell: Well, currently our only use of this is an object that comes straight out of a JSON.parse, and the current impl doesn't handle circular references either. Still, would be better to handle if we want this to be generic.

kentonv: Oops, I guess I misremembered where this is used, sorry.

src/workerd/jsg/util.c++
anonrig: @dario-piotrowicz used it on a nodejs pr quite recently!

dario-piotrowicz: Yeah here: https://github.com/nodejs/node/pull/58782/files#diff-62fb3389674420a22b14191b1eb3b65a173f6c21657cf3e8a44c442244115f0fR621 :slightly_smiling_face:

anonrig:
Rebased and force-pushed.

PR #4454, opened 2025-07-03
add recursivelyFreeze benchmark

anonrig:
Ref: https://github.com/cloudflare/workerd/pull/4452

### DIFF
--- a/build/wd_cc_benchmark.bzl
+++ b/build/wd_cc_benchmark.bzl
@@ -12,7 +12,13 @@ def wd_cc_benchmark(
     native.cc_binary(
         name = name,
         defines = ["WD_IS_BENCHMARK"],
-        linkstatic = 1,
+        # Use shared linkage for benchmarks, matching the approach used for tests. Unfortunately,
+        # bazel does not support shared linkage on macOS and it is broken on Windows, so only
+        # enable this on Linux.
+        linkstatic = select({
+            "@platforms//os:linux": 0,
+            "//conditions:default": 1,
+        }),
         linkopts = linkopts + select({
             "@//:use_dead_strip": ["-Wl,-dead_strip", "-Wl,-no_exported_symbols"],
             "//conditions:default": [""],
--- a/build/wd_cc_benchmark.bzl
+++ b/build/wd_cc_benchmark.bzl
@@ -12,13 +12,7 @@ def wd_cc_benchmark(
     native.cc_binary(
         name = name,
         defines = ["WD_IS_BENCHMARK"],
-        # Use shared linkage for benchmarks, matching the approach used for tests. Unfortunately,
-        # bazel does not support shared linkage on macOS and it is broken on Windows, so only
-        # enable this on Linux.
-        linkstatic = select({
-            "@platforms//os:linux": 0,
-            "//conditions:default": 1,
-        }),
+        linkstatic = 1,
         linkopts = linkopts + select({
             "@//:use_dead_strip": ["-Wl,-dead_strip", "-Wl,-no_exported_symbols"],
             "//conditions:default": [""],
--- a/src/workerd/tests/bench-util.c++
+++ b/src/workerd/tests/bench-util.c++
@@ -1,0 +1,79 @@
+// Copyright (c) 2023 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include <workerd/jsg/jsg.h>
+#include <workerd/jsg/setup.h>
+#include <workerd/jsg/util.h>
+#include <workerd/tests/bench-tools.h>
+
+namespace workerd {
+namespace {
+
+// Ref: https://developers.cloudflare.com/workers/runtime-apis/request/#the-cf-property-requestinitcfproperties
+static constexpr kj::StringPtr cfProperty = R"DATA({
+  "apps": false,
+  "cacheEverything": false,
+  "cacheKey": "my-cache-key",
+  "cacheTags": ["production", "development"],
+  "cacheTtl": 3600,
+  "cacheTtlByStatus": { "200-299": 86400, "404": 1, "500-599": 0 },
+  "image": null,
+  "mirage": true,
+  "polish": "lossless",
+  "scrapeShield": true,
+  "webp": false
+})DATA"_kjc;
+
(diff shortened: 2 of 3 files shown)


anonrig:
@fhanau @mikea any idea why we're getting this error now?

bazel-out/k8-opt/bin/src/workerd/tests/bench-api-headers: error while loading shared libraries: libsrc_Sworkerd_Stests_Slibtest-fixture.so: cannot open shared object file: No such file or directory

jasnell (approved):

fhanau:
> @fhanau @mikea any idea why we're getting this error now?
> 
> bazel-out/k8-opt/bin/src/workerd/tests/bench-api-headers: error while loading shared libraries: libsrc_Sworkerd_Stests_Slibtest-fixture.so: cannot open shared object file: No such file or directory

No idea – this seems like an issue that is unrelated to the change here since we are not changing bench-api-headers or its dependencies, we're only adding other targets in the commit where the failure is encountered.
I assume that this is a Bazel bug of some sorts, maybe related to caching. Have you tried running it again? Dynamic linkage for test binaries is valuable and always worked before in my experience.
Edit: To be specific, dynamically linking test and bench binaries is one optimization helping us to avoid running out of disk space on Linux CI runners despite installing LLVM and building lots of code, benchmark binaries in particular would be pretty large otherwise. Remote cache downloads would also be slower if we statically linked binaries.

anonrig:
> dependencies, we're only adding other targets in the commit where the failure is encountered.

@fhanau we change all_benchmarks target which is built on benchmarks ci. my two cents: it appears that this is indeed a caching issue that reveals an actual problem.

fhanau:
> > dependencies, we're only adding other targets in the commit where the failure is encountered.
> 
> @fhanau we change all_benchmarks target which is built on benchmarks ci. my two cents: it appears that this is indeed a caching issue that reveals an actual problem.

Yes, but we're only adding more targets to filegroup, we're not doing anything that would changing the bench-api-headers target in any way. I think we should remove the second commit, if CI fails with that again we might need to wipe the remote cache. I see no indication that benchmark changes or dynamic linkage are the problem here per se.

fhanau:
> ## [CodSpeed Performance Report](https://codspeed.io/cloudflare/workerd/branches/yagiz%2Fadd-benchmark-for-bench-util)
> ### Merging #4454 will **improve performances by 19.06%**
> Comparing `yagiz/add-benchmark-for-bench-util` ([906e630](https://github.com/cloudflare/workerd/commit/906e6305ea44130fc9e05c72f629c0151100b379)) with `main` ([0bf8308](https://github.com/cloudflare/workerd/commit/0bf83083571c3c52cd4b8d06dc72007d77665bb2))
> 
> ### Summary
> `⚡ 5` improvements `✅ 5` untouched benchmarks `🆕 1` new benchmarks
> 
> ### Benchmarks breakdown
> Benchmark	`BASE`	`HEAD`	Change
> ⚡	`ParseAndSerialize`	17.7 µs	16 µs	+10.88%
> ⚡	`Serialize`	4.4 µs	3.9 µs	+11.1%
> ⚡	`constructor[ApiHeaders]`	27.7 µs	23.3 µs	+19.06%
> ⚡	`request[GlobalScopeBenchmark]`	328.9 µs	289.7 µs	+13.57%
> ⚡	`Parse[KjHeaders]`	8.8 µs	7.5 µs	+17.36%
> 🆕	`recursivelyFreeze[UtilFixture]`	N/A	117 ms	N/A

false positives 😬

anonrig:
@fhanau linking statically will improve performance wouldn't it?

src/workerd/tests/bench-util.c++
```diff
@@ -0,0 +1,79 @@
...
+struct UtilFixture: public benchmark::Fixture {
+  virtual ~UtilFixture() noexcept(true) {}
+
+  void SetUp(benchmark::State& state) noexcept(true) override {}
+
+  void TearDown(benchmark::State& state) noexcept(true) override {}
+};
+
+BENCHMARK_F(UtilFixture, recursivelyFreeze)(benchmark::State& state) {
+  UtilIsolate isolate(system, kj::heap<jsg::IsolateObserver>(), {});
```
mikea, line 55: I am not super familiar with google benchmark, but wouldn't this include isolate setup in measurements? I imagined everything that is not needed to be benchmarked needs to be in SetUp/TearDown?

mikea:
> linking statically will improve performance wouldn't it?

it shouldn't unless we measure wrong things?

build/wd_cc_benchmark.bzl
```diff
@@ -12,13 +12,7 @@ def wd_cc_benchmark(
...
         name = name,
         defines = ["WD_IS_BENCHMARK"],
-        # Use shared linkage for benchmarks, matching the approach used for tests. Unfortunately,
-        # bazel does not support shared linkage on macOS and it is broken on Windows, so only
-        # enable this on Linux.
-        linkstatic = select({
-            "@platforms//os:linux": 0,
-            "//conditions:default": 1,
-        }),
+        linkstatic = 1,
```
anonrig:
```suggestion
        # Use shared linkage for benchmarks, matching the approach used for tests. Unfortunately,
        # bazel does not support shared linkage on macOS and it is broken on Windows, so only
        # enable this on Linux.
        linkstatic = select({
            "@platforms//os:linux": 0,
            "//conditions:default": 1,
        }),
```

anonrig:
> > linking statically will improve performance wouldn't it?
> 
> 
> 
> it shouldn't unless we measure wrong things?

I'll open another PR to update and improve all benchmarks

src/workerd/tests/bench-util.c++
anonrig: I'll do a follow up since this is a common pattern in all existing benchmarks.

fhanau:
> @fhanau linking statically will improve performance wouldn't it?

If you have a benchmarking framework at your hands, that question should be easy to answer 😉

anonrig:
@fhanau the build seems to fail without static linking change

fhanau:
Looks like CI is passing after flushing the cache!

src/workerd/tests/bench-util.c++
danlapid: This is definitely right and we still need to do the follow up to increase iteration count to a statistically significant benchmark duration

PR #4459, opened 2025-07-03
Implement node:fs cp APIs

jasnell:
Implements the `node:fs` `fs.cpSync()`, `fs.cp()`, `fsPromises.cp()` APIs

### DIFF
--- a/src/workerd/api/filesystem.c++
+++ b/src/workerd/api/filesystem.c++
@@ -1204,6 +1204,615 @@ kj::Array<FileSystemModule::DirEntHandle> FileSystemModule::readdir(
   }
 }
 
+namespace {
+
+using MaybeFsNode = kj::Maybe<
+    kj::OneOf<kj::Rc<workerd::Directory>, kj::Rc<workerd::File>, kj::Rc<workerd::SymbolicLink>>>;
+
+MaybeFsNode getNodeOrError(jsg::Lock& js,
+    const workerd::VirtualFileSystem& vfs,
+    const jsg::Url& url,
+    const FileSystemModule::CpOptions& options) {
+  KJ_IF_SOME(node, vfs.resolve(js, url, {.followLinks = options.deferenceSymlinks})) {
+    KJ_SWITCH_ONEOF(node) {
+      KJ_CASE_ONEOF(file, kj::Rc<workerd::File>) {
+        return MaybeFsNode(kj::mv(file));
+      }
+      KJ_CASE_ONEOF(dir, kj::Rc<workerd::Directory>) {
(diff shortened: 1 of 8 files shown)


src/workerd/api/node/tests/fs-cp-test.js
```diff
@@ -0,0 +1,2135 @@
...
+  symlinkSync,
+  cpSync,
+  cp,
+  existsSync,
+  readFileSync,
+  lstatSync,
+  statSync,
+  promises as fsPromises,
+} from 'node:fs';
+
```
jasnell, line 13: Heads up to reviewers: Tests were mostly generated by Claude

src/workerd/api/filesystem.c++
```diff
@@ -1204,6 +1204,615 @@ kj::Array<FileSystemModule::DirEntHandle> FileSystemModule::readdir(
   }
 }
 
+namespace {
+
+using MaybeFsNode = kj::Maybe<
+    kj::OneOf<kj::Rc<workerd::Directory>, kj::Rc<workerd::File>, kj::Rc<workerd::SymbolicLink>>>;
```
guybedford, line 1210: Still picking up on the kj patterns, but could this be a `kj::Own` ref that you move instead of an Rc?

src/workerd/api/filesystem.c++
```diff
@@ -1204,6 +1204,615 @@ kj::Array<FileSystemModule::DirEntHandle> FileSystemModule::readdir(
...
+          // Now we can recursively copy the contents of the source directory into the new one.
+          handleCpDir(js, vfs, kj::mv(dir), kj::mv(newDir), options);
+        }
+      }
+      KJ_CASE_ONEOF(link, kj::Rc<workerd::SymbolicLink>) {
+        KJ_IF_SOME(existing,
+            dest->tryOpen(js, kj::Path({name}),
+                {
+                  .followLinks = options.deferenceSymlinks,
+                })) {
```
guybedford, line 1510: Since this is in every path it could be simplified by hoisting out with a special case for the directory recovery path.

src/workerd/api/filesystem.c++
jasnell: We introduced `kj::Rc` in order to be able to move away from always using `kj::Own` giving the ambiguous semantics.

src/workerd/api/filesystem.c++
jasnell: Possibly. I'll defer this tho. Before I move `node:fs` out of experimental I plan to make a cleanup pass through the code and will revisit this then

PR #4460, opened 2025-07-03
use bazel run directly for benchmarking

anonrig:
Without this change we get an error 

Ref: https://github.com/cloudflare/workerd/actions/runs/16059498845/job/45321954837


anonrig:
> Definitely fine to try this if it works. I am not sure exactly why it's needed though.

My assumption is, we're running those binaries directly rather than using "bazel run" which prepares the shared libraries for us upon execution.

npaun (approved):

PR #4461, opened 2025-07-04
Refine streaming tail worker interface types

jmorrell-cloudflare:
An attempt at fixing the overbroad types I detailed here: https://wiki.cfdata.org/display/~jmorrell/May+2025+Streaming+Tail+Worker+Feedback?focusedCommentId=1166311884#comment-1166311884

### DIFF
--- a/types/defines/trace.d.ts
+++ b/types/defines/trace.d.ts
@@ -175,19 +175,43 @@ interface Attributes {
   readonly info: Attribute[];
 }
 
-interface TailEvent {
-  readonly traceId: string;
+type EventType =
+  | Onset
+  | Outcome
+  | Hibernate
+  | SpanOpen
+  | SpanClose
+  | DiagnosticChannelEvent
+  | Exception
+  | Log
+  | Return
+  | Link
+  | Attributes;
+
+interface TailEvent<Event extends EventType> {
   readonly invocationId: string;
   readonly spanId: string;
   readonly timestamp: Date;
   readonly sequence: number;
-  readonly event: Onset | Outcome | Hibernate | SpanOpen | SpanClose | DiagnosticChannelEvent | Exception | Log | Return | Link | Attributes;
+  readonly event: Event;
 }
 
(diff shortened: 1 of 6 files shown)


types/defines/trace.d.ts
```diff
@@ -175,19 +175,43 @@ interface Attributes {
...
-  readonly event: Onset | Outcome | Hibernate | SpanOpen | SpanClose | DiagnosticChannelEvent | Exception | Log | Return | Link | Attributes;
-}
+  readonly event: Event;
+}
+
+type TailEventHandler<Event extends EventType = EventType> = (
+  event: TailEvent<Event>
+) => void | Promise<void>;
+
+type TailEventHandlerObject = {
```
jmorrell-cloudflare, line 203: Note that this does not contain an `onset` handler since this is only returned in response to an `onset` event.

types/defines/trace.d.ts
```diff
@@ -175,19 +175,43 @@ interface Attributes {
...
+  | SpanOpen
+  | SpanClose
+  | DiagnosticChannelEvent
+  | Exception
+  | Log
+  | Return
+  | Link
+  | Attributes;
+
+interface TailEvent<Event extends EventType> {
```
fhanau, line 191: What do the `<Event extends EventType>` here and the extends below mean? I assume it's used to enforce that we can specialize TailEvent as `TailEvent<Type>` so that the event field can be coerced to be one specific type, but only one of the types that are part of EventType.
LGTM otherwise.

fhanau (approved):

types/defines/trace.d.ts
jmorrell-cloudflare: > I assume it's used to enforce that we can specialize TailEvent as TailEvent<Type> so that the event field can be coerced to be one specific type, but only one of the types that are part of EventType.

Yep, that's it. It means that any `Event` should match one of the interfaces in the `EventType` union.

jasnell (approved):

petebacondarwin (approved):

Issue #4453, opened 2025-07-02
🐛 Bug Report — Runtime APIs -- EventListeners on an EventTarget are unbinded to the parent class


ProgrammerIn-wonderland:
Workerd incorrectly does not bind an EventListener on an EventTarget. 

For the example 
```js
class ExampleEventTarget extends EventTarget {
    constructor() {super()}
}
const instance = new ExampleEventTarget()

function instanceListener () {
    console.log(this)
 }
instance.addEventListener("eventName", instanceListener)
instance.dispatchEvent(new CustomEvent("eventName"))
```
nodejs and browsers will return `ExampleEventTarget {}` while cloudflare workers (and other workerd applications) return` ServiceWorkerGlobalScope {}` due to the incorrect binding behavior. 

Screenshots:

![Image](https://github.com/user-attachments/assets/701b31e3-d8a6-458b-8739-e1ddb667e7cd)
![Image](https://github.com/user-attachments/assets/f13b4838-297d-4ee0-bc25-1c217d70af5f)
![Image](https://github.com/user-attachments/assets/901cb453-1251-4a2b-a9c4-e11747584f56)

### Workaround
For applications currently affected by this incorrect behavior I have found the following patch to work for my code and presumably other applications which expect  the parent class to be bound to the function attached to an eventlistener
```js
const __cfEventTarget = EventTarget;
globalThis.EventTarget = class EventTarget extends __cfEventTarget {
    constructor(...args) {
        super(...args)
    }
    addEventListener(type, listener, options) {
        super.addEventListener(type, listener.bind(this), options);
    }
}
```

jelveh:
Could you please help us? This is causing serious issues for our service workers.

anonrig:
Cc @jasnell

PR #4468, opened 2025-07-07
Set `this` correctly in EventTarget event handlers with compat flag

jasnell:
Fixes: https://github.com/cloudflare/workerd/issues/4453

### DIFF
--- a/src/workerd/api/basics.c++
+++ b/src/workerd/api/basics.c++
@@ -221,7 +221,8 @@ kj::Array<kj::StringPtr> EventTarget::getHandlerNames() const {
 void EventTarget::addEventListener(jsg::Lock& js,
     kj::String type,
     kj::Maybe<jsg::Identified<Handler>> maybeHandler,
-    jsg::Optional<AddEventListenerOpts> maybeOptions) {
+    jsg::Optional<AddEventListenerOpts> maybeOptions,
+    const jsg::TypeHandler<jsg::Ref<EventTarget>>& eventTargetHandler) {
   if (warnOnSpecialEvents && isSpecialEventType(type)) {
     js.logWarning(kj::str("When using module syntax, the '", type,
         "' event handler should be "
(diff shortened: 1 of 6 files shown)


src/workerd/io/compatibility-date.capnp
```diff
@@ -855,4 +855,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
   # "getBuiltinModule", "platform" and "features" property implementations, to the full-featured
   # Node.js-compatibile process implementation with all process properties either stubbed or
   # implemented. Implies nodejs_compat + nodejs_compat_v2 (which sets the process global).
+
+  setEventTargetThis @98 :Bool
+      $compatEnableFlag("set_event_target_this")
+      $compatDisableFlag("no_set_event_target_this")
+      $compatEnableDate("2025-08-01");
```
anonrig, line 862: Is this going to fix any existing WPT? If yes, than after August 1st tests will start to fail

jasnell: Running the WPTs with the flag set manually does not reveal any new WPT failures.

image

src/workerd/api/basics.c++
```diff
@@ -236,9 +237,15 @@ void EventTarget::addEventListener(jsg::Lock& js,
       HandlerFunction handlerFn = ([&]() {
         KJ_SWITCH_ONEOF(handler.unwrapped) {
           KJ_CASE_ONEOF(fn, HandlerFunction) {
+            if (FeatureFlags::get(js).getSetEventTargetThis()) {
+              fn.setReceiver(js.v8Ref(eventTargetHandler.wrap(js, JSG_THIS)));
```
KernelDeimos, line 241: I'm not very familiar with v8-interfacing C++ code, but is `fn.setReceiver` here equivalent to calling `.bind()` in javascript? If it is, this effectively mutates the function because `.bind` can only be called once on a function, and the equivalent of `.call()` should be used instead.

jasnell: No, it's not the same. This makes it so that our wrapper (`jsg::Function`) calls the function explicitly with the set receiver. It does not mutate the state of the js function itself.

KernelDeimos: Perfect! Thanks for clarifying

PR #4470, opened 2025-07-07
Python Workflows API improvements

Caio-Nogueira:
This PR includes breaking changes on the `step` API. Workflows are still experimental and under active development
- bugfix `step.do` to await for Js promise rejection
- new error translation layer between Python exceptions and Js errors (for now custom exceptions won't be catchable)
- new DAG optional params - users are now able to define a list of dependencies (other steps they depend on) and have them be executed (in parallel or sequentially). See [this diff](https://github.com/cloudflare/workerd/compare/caio/python-workflows-2?expand=1#diff-5a01b2a830d4f6b351d55e7100a3762882d709d39e2a5cdb89f1398d80a8b34b)
- removes the option to call a step with a callable as the second param - makes the API more uniform
- adds coverage for these additions in the existing test

### DIFF
--- a/src/pyodide/internal/_workers.py
+++ b/src/pyodide/internal/_workers.py
@@ -6,6 +6,7 @@ import functools
 import http.client
 import inspect
 import json
+from asyncio import create_task, gather
 from collections.abc import Generator, Iterable, MutableMapping
 from contextlib import ExitStack, contextmanager
 from enum import StrEnum
@@ -198,6 +199,30 @@ def _to_python_exception(exc: JsException) -> Exception:
         return TypeError(exc.message)
     else:
         return exc
+
+
+class NonRetryableError(Exception):
+    # This is a marker exception used to signal that a workflow step should not be retried.
+    # This is a special exception used by workflows
+    pass
+
+
+def _from_js_error(exc: JsException) -> Exception:
+    # convert into Python exception after a full round trip
+    # Python - JS - Python
+    if not exc.message or not exc.message.startswith("PythonError"):
+        return _to_python_exception(exc)
+
+    # extract the Python exception type from the traceback
+    error_message_last_line = exc.message.split("\n")[-2]
+    if error_message_last_line.startswith("TypeError"):
+        return TypeError(error_message_last_line)
+    elif error_message_last_line.startswith("ValueError"):
+        return ValueError(error_message_last_line)
+    elif error_message_last_line.startswith("_workers.NonRetryableError"):
+        return NonRetryableError(error_message_last_line)
+    else:
+        return _to_python_exception(exc)
 
 
 @contextmanager
@@ -937,20 +962,24 @@ def handler(func):
 class _WorkflowStepWrapper:
     def __init__(self, js_step):
         self._js_step = js_step
+        self._memoized_dependencies = {}
+        self._in_flight = {}
 
-    def do(self, name, callback=None, config=None):
-        if callback:
-
-            async def wrapper():
-                return _do_call(self, name, config, callback)
-
-            return wrapper()
-
-        # Return a decorator that will execute the function when called
+    def do(self, name, depends=None, concurrent=False, config=None):
         def decorator(func):
             async def wrapper():
-                return _do_call(self, name, config, func)
+                if concurrent:
+                    results = await gather(
+                        *[self._resolve_dependency(dep) for dep in depends or []]
+                    )
+                else:
+                    results = [
+                        await self._resolve_dependency(dep) for dep in depends or []
+                    ]
+                python_results = [python_from_rpc(result) for result in results]
+                return await _do_call(self, name, config, func, *python_results)
 
+            wrapper._step_name = name
             return wrapper
 
         return decorator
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/_workers.py
```diff
@@ -200,6 +201,31 @@ def _to_python_exception(exc: JsException) -> Exception:
         return exc
 
 
+class NonRetryableError(Exception):
+    # This is a marker exception used to signal that a workflow step should not be retried.
+    # This is a special exception used by workflows
+    pass
+
+
+def _from_js_error(exc: JsException) -> Exception:
```
dom96, line 210: Any reason not to just put this in `_to_python_exception` instead of defining a new function for it?

src/pyodide/internal/_workers.py
```diff
@@ -200,6 +201,31 @@ def _to_python_exception(exc: JsException) -> Exception:
...
+    # convert into Python exception after a full round trip
+    # Python - JS - Python
+    if not exc.message or not exc.message.startswith("PythonError"):
+        return _to_python_exception(exc)
+
+    # extract the Python exception type from the traceback
+    error_message_last_line = exc.message.split("\n")[-2]
```
dom96, line 217: It would be nice if we had a way to avoid parsing the traceback here. Are there any fields on the exception that give us this metadata? @hoodmane any ideas?

hoodmane: Should be in `exc.type` unless that got deleted when crossing the rpc.

src/pyodide/internal/_workers.py
```diff
@@ -200,6 +201,31 @@ def _to_python_exception(exc: JsException) -> Exception:
...
+    if error_message_last_line.startswith("TypeError"):
+        return TypeError(error_message_last_line)
+    elif error_message_last_line.startswith("ValueError"):
+        return ValueError(error_message_last_line)
+    elif error_message_last_line.startswith("_workers.NonRetryableError"):
+        return NonRetryableError(error_message_last_line)
+    # TODO(caio): Add support for timeout error from engine
+    else:
+        return _to_python_exception(exc)
```
hoodmane: Probably something like this:
```suggestion
    import builtins
    exc_type = getattr(builtins, exc.type, None)
    if exc.type == "NonRetryableError":
       exc_type = NonRetryableError
    if exc_type is not None:
       return exc_type(exc.message)
    return _to_python_exception(exc)
```
And agreed with @dom96  that it would make sense to make this change directly in `_to_python_exception()` rather than adding separate logic.

src/pyodide/internal/_workers.py
```diff
@@ -938,18 +964,16 @@ class _WorkflowStepWrapper:
...
-
-            async def wrapper():
-                return _do_call(self, name, config, callback)
-
-            return wrapper()
-
-        # Return a decorator that will execute the function when called
+    def do(self, name, depends=None, concurrent=False, config=None):
         def decorator(func):
+            @functools.wraps(func)
```
hoodmane: Again I think this doesn't belong here because `@functools.wraps()` is for when we're not changing the arguments but the wrapper always takes zero arguments.

src/pyodide/internal/_workers.py
```diff
@@ -938,18 +964,16 @@ class _WorkflowStepWrapper:
...
             async def wrapper():
-                return _do_call(self, name, config, func)
+                if concurrent:
+                    results = await gather(*[dep() for dep in depends or []])
```
hoodmane: Shouldn't concurrent be the default?

Caio-Nogueira: It's riskier making the dependent steps run concurrently by default. This is a userland method and it's not explicit to the user that these will run concurrently. Doing so by default could pass some race conditions on to the users without them being aware of them

src/pyodide/internal/_workers.py
Caio-Nogueira: I think `_to_python_exception` won't work for this. This is a general error thrown from the Js side. As such `exc.name` will just be `Error`, and the information about the error type is included in the `message` (that includes the traceback).

An alternative to this would be moving this added logic to the `_to_python_exception` to make it handle this use cases where errors are caught after a full round trip.

src/pyodide/internal/_workers.py
Caio-Nogueira: > Probably something like this:
> 
> And agreed with @dom96 that it would make sense to make this change directly in `_to_python_exception()` rather than adding separate logic.

I think `type` may be lost across language or rpc barrier:
```
  File "/lib/python3.12/site-packages/_workers.py", line 212, in _from_js_error
    exc_type = getattr(builtins, exc.type, None)
                                 ^^^^^^^^
AttributeError: type
```

Talador12:
Do you have any auto linting going into each of these PRs? I do not think it's the scope of this PR to solve it, but I think it would be good to format these with Black, Ruff (Flake8)

Sometimes you see things like this all in one project:
```
black
pyright
isort
ssort
darglint
pydocstyle
ruff
```

src/pyodide/internal/_workers.py
LuisDuarte1: >  unless that got deleted when crossing the rpc.

that's exactly what happens :sweat_smile:  (`step`,in JS-land, its just an `RpcTarget`)

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
```diff
@@ -2,56 +2,57 @@
...
-            await workflow_step()
-        except MyCustomException as e:
-            print(f"successfully caught {e} with decorator")
-            pass
+        # `step_1` and `step_2` run serially
+        @step.do("step_4", depends=[step_1, step_2], concurrent=False)
+        async def step_4(result1, result2):
+            print("Executing step 4 (depends on step 1 and step 2)")
+            assert result1["foo"] == "foo"
+            assert result2["bar"] == "bar"
```
LuisDuarte1: `step_2` will run two times in this case (as step names can be repeated, and we do an internal count internally) - since users are expecting this behavior, when using a DAG-like syntax? i would personally also expect it to run only once and only use the memoized result on the second dependent.

Caio-Nogueira:
> Do you have any auto linting going into each of these PRs? I do not think it's the scope of this PR to solve it, but I think it would be good to format these with Black, Ruff (Flake8)
> 
> Sometimes you see things like this all in one project:
> 
> ```
> black
> pyright
> isort
> ssort
> darglint
> pydocstyle
> ruff
> ```

We use the `ruff` linter/formatter

src/workerd/server/tests/python/workflow-entrypoint/workflow.py
Caio-Nogueira: I see. The issue here is that if we declare some step as a dependency multiple times, it will run multiple times. The correct behavior here (or at least most intuitive) is to have the dependent step run once and then memoize its result. When some other step uses it again for its dependency tree, we just use this memoized result.

@Talador12 wdyt here?

src/pyodide/internal/_workers.py
hoodmane: I wonder if we can teach the RPC to serialize these fields.

LuisDuarte1 (approved):

PR #4472, opened 2025-07-07
DO NOT MERGE: bench impact of accounted*

anonrig:
Do not merge.


jasnell:
While benchmarking this, do not forget that we cannot eliminate the memory accounting. We have to have a reasonable approach that accounts for the memory usage of the `Headers` object. The key performance bottleneck here, as has been suspected, is the creation and attachment of the external memory adjuster on each of the things. We can likely eliminate that if we just simply had the `Headers` object store the JavaScript strings themselves directly and only lazily converted to `kj::String` when the `Headers` object is consumed and copied into a `kj::Headers`. By storing JS strings rather than `kj::String` we get the memory accounting for free. However, that may have a performnace impact of its own. bottom line is that raw performance on this is not the only concern.

/cc @jp4a50

mikea:
> While benchmarking this, do not forget that we cannot eliminate the memory accounting.

at this point we're just establishing the base line: a number that can be reached by removing all "unecessary" stuff, and then carefully re-adding all features back, while trying really hard to stay close to the base line.

src/workerd/api/http.c++
```diff
@@ -90,7 +90,7 @@ jsg::ByteString normalizeHeaderValue(jsg::Lock& js, jsg::ByteString value) {
   if (slice.size() == value.size()) {
     return kj::mv(value);
   }
-  return js.accountedByteString(kj::str(slice));
+  return jsg::ByteString(kj::str(slice));
```
jasnell, line 93: FWIW, part of the main motivation for putting the accounting logic behind the `js.accountedByteString(...)` utility methods was so that we could easily and quickly disable it within those functions ;-) ... would have been a lot easier, I think, to just disable the bits in the `js.accounted...String` methods ;-)

src/workerd/api/http.c++
```diff
@@ -161,14 +161,14 @@ Headers::Headers(jsg::Lock& js, const Headers& other): guard(Guard::NONE) {
 }
 
 Headers::Headers(jsg::Lock& js, const kj::HttpHeaders& other, Guard guard): guard(Guard::NONE) {
-  other.forEach([this](kj::StringPtr name, kj::StringPtr value) {
-    // The variation of toLower we use here creates a copy.
-    auto key = jsg::ByteString(toLower(name));
-    KJ_IF_SOME(header, headers.find(key)) {
-      header.values.add(jsg::ByteString(kj::str(value)));
+  other.forEachHeader([this](kj::HttpHeaders::Header header) {
+    auto key = jsg::ByteString(kj::str(header.lowcaseName));
```
jasnell: It would be ideal if we could avoid the additional allocation (`kj::str(...)`) and wrapping it in `jsg::ByteString` here.

anonrig: Byte string is a kj::string, and kj::string constructor just moves the value. So technically there is no unnecessary allocation here.

jasnell: I was referring to the additional allocation caused by the `kj::str(...)` call, which copies the `kj::StringPtr`. It would be awesome if we could eliminate that allocation unless we end up in the else block.

PR #4473, opened 2025-07-07
Make import_from_javascript work at top level with Pyodide 0.27

hoodmane:


src/workerd/server/tests/python/importable-env/worker.py
```diff
@@ -9,24 +9,50 @@
...
+try:
+    vectorize = import_from_javascript("cloudflare:vectorize")
+except ImportError:
+    pass
+
+if __version__ == "0.26.0a2":
+    # Should have thrown throw "ImportError: Failed to import 'cloudflare:vectorize': Only 'cloudflare:workers' and 'cloudflare:sockets' are available until the next python runtime version."
+    assert vectorize is None, (
+        'import_from_javascript("cloudflare:vectorize") not expected to work in the global scope in 0.26.0a2'
+    )
```
dom96: This will pass if `import_from_javascript` happens to return `None` without throwing. We should assert on it throwing.

src/workerd/server/tests/python/importable-env/worker.py
```diff
@@ -9,24 +9,50 @@
...
+else:
+    assert {"DistanceMetric", "KnownModel", "MetadataRetrievalLevel"}.issubset(
+        dir(vectorize)
+    )
+
+from js import setTimeout
```
dom96, line 30: Should this live in a separate test file? Can you also add a comment that describes what this tests?

src/pyodide/internal/python.ts
```diff
@@ -114,6 +115,24 @@ function validatePyodideVersion(pyodide: Pyodide): void {
...
+): number {
+  // Redirect top level setTimeout(cb, 0) to queueMicrotask().
+  // If we don't know how to handle it, call normal setTimeout() to force failure.
+  if (typeof handler === 'string') {
+    return origSetTimeout(handler, timeout);
+  }
+  if (timeout) {
+    return origSetTimeout(handler, timeout);
+  }
+  queueMicrotask(handler as () => void);
```
anonrig: Can you just change the type of handler to VoidFunction rather than this force cast?

src/pyodide/internal/python.ts
```diff
@@ -176,3 +200,8 @@ export function loadPyodide(
     reportError(e as Error);
   }
 }
+
+export function beforeRequest(Module: Module): void {
```
anonrig, line 203:
```suggestion
export function beforeRequest(module: Module): void {
```

src/pyodide/internal/python.ts
hoodmane: Well I did it this way so that `setTimeoutTopLevelPatch` has type `typeof setTimeout`. The alternative is I can cast it like `setTimeoutTopLevelPatch as typeof setTimeout`, which is perhaps a bit cleaner since we could remove one lint disable and one cast in trade for one cast.

src/pyodide/internal/python.ts
hoodmane: We have this all over the place so if we're going to change it it should be consistent. I think it would be better to rename the type to `ModuleType` personally, since Emscripten calls it `Module` with a capital `M`.

PR #4474, opened 2025-07-07
fix formatters path for zed and vscode

anonrig:
When running benchmarks, bazel-bin folder doesn't include formatters. This fixes this particular problem.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -18,9 +18,9 @@
     "--query-driver=**",
     "--clang-tidy"
   ],
-  "clang-format.executable": "${workspaceRoot}/bazel-bin/build/deps/formatters/clang-format",
-  "bazel.buildifierExecutable": "${workspaceRoot}/bazel-bin/build/deps/formatters/buildifier",
-  "ruff.path": ["${workspaceRoot}/bazel-bin/build/deps/formatters/ruff"],
+  "clang-format.executable": "${workspaceRoot}/bazel-out/k8-fastbuild/bin/build/deps/formatters/clang-format",
+  "bazel.buildifierExecutable": "${workspaceRoot}/bazel-out/k8-fastbuild/bin/build/deps/formatters/buildifier",
+  "ruff.path": ["${workspaceRoot}/bazel-out/k8-fastbuild/bin/build/deps/formatters/ruff"],
   "rust-analyzer.workspace.discoverConfig": {
     "command": ["just", "_rust-analyzer"],
     "progressLabel": "generating rust analyzer config",
--- a/.zed/settings.json
+++ b/.zed/settings.json
@@ -7,7 +7,7 @@
     "C++": {
       "formatter": {
         "external": {
-          "command": "bazel-bin/build/deps/formatters/clang-format",
+          "command": "bazel-out/k8-fastbuild/bin/build/deps/formatters/clang-format",
           "args": ["-i", "{buffer_path}"]
         }
       }
@@ -15,7 +15,7 @@
     "Python": {
       "formatter": {
         "external": {
-          "command": "bazel-bin/build/deps/formatters/ruff",
+          "command": "bazel-out/k8-fastbuild/bin/build/deps/formatters/ruff",
           "arguments": ["format", "--stdin-filename", "{buffer_path}"]
         }
       }
@@ -23,7 +23,7 @@
     "Starlark": {
       "formatter": {
         "external": {
-          "command": "bazel-bin/build/deps/formatters/buildifier",
+          "command": "bazel-out/k8-fastbuild/bin/build/deps/formatters/buildifier",
           "arguments": ["--lint=fix", "--path", "{buffer_path}"]
         }
       }


.vscode/settings.json
```diff
@@ -18,9 +18,9 @@
     "--query-driver=**",
     "--clang-tidy"
   ],
-  "clang-format.executable": "${workspaceRoot}/bazel-bin/build/deps/formatters/clang-format",
-  "bazel.buildifierExecutable": "${workspaceRoot}/bazel-bin/build/deps/formatters/buildifier",
-  "ruff.path": ["${workspaceRoot}/bazel-bin/build/deps/formatters/ruff"],
+  "clang-format.executable": "${workspaceRoot}/bazel-out/k8-fastbuild/bin/build/deps/formatters/clang-format",
+  "bazel.buildifierExecutable": "${workspaceRoot}/bazel-out/k8-fastbuild/bin/build/deps/formatters/buildifier",
+  "ruff.path": ["${workspaceRoot}/bazel-out/k8-fastbuild/bin/build/deps/formatters/ruff"],
```
npaun, line 23: k8 is bazel's weird term for amd64. We'll need to find a platform independent approach.

anonrig: bazel-bin is supposed to be that, but unfortunately if you run "just b mimetype" bazel-bin doesn't have any build folder...

npaun: Maybe we could make the formatter installation create that folder if it doesn't already exist?

danlapid: The way format.py works this directory should always exist, I don't understand why you don't get that with just but please don't change those lines

anonrig:
> Please understand why that directory is missing under just

@npaun @danlapid 

When running specific targets that doesn't have a dependency to any formatter binary, they don't get symlinked into bazel-bin folder. For example, `just bench util` will create bazel-out/k8-opt because it's an optimized build. After completion bazel-bin will point to bazel-out/k8-opt. And since k8-opt does not include build folder (because benchmarks doesn't depend on any formatting targets), formatters will stop working unless building it once again using (bazel build //...) which symlinks bazel-bin to bazel-out/k8-dbg. 

So, unless we separate this binary from the build flow, we need to make assumptions towards finding the correct path for the formatters to work.

PS: Folder being missing is not related to just at all.

.vscode/settings.json
anonrig: If this file doesn't exist, format.py runs "bazel run ..." which literally bypasses this path and runs the executable.

danlapid:
> > Please understand why that directory is missing under just
> 
> 
> 
> @npaun @danlapid 
> 
> 
> 
> When running specific targets that doesn't have a dependency to any formatter binary, they don't get symlinked into bazel-bin folder. For example, `just bench util` will create bazel-out/k8-opt because it's an optimized build. After completion bazel-bin will point to bazel-out/k8-opt. And since k8-opt does not include build folder (because benchmarks doesn't depend on any formatting targets), formatters will stop working unless building it once again using (bazel build //...) which symlinks bazel-bin to bazel-out/k8-dbg. 
> 
> 
> 
> So, unless we separate this binary from the build flow, we need to make assumptions towards finding the correct path for the formatters to work.
> 
> 
> 
> PS: Folder being missing is not related to just at all.

I don't understand the problem, if you haven't run format.py then the formatters are not available and they shouldn't be.

Yes, running format.py is a prerequisite

anonrig:
> I don't understand the problem, if you haven't run format.py then the formatters are not available and they shouldn't be.

@danlapid Please run the following

1. Run `tools/cross/format.py`
2. Validate that bazel-bin/build/deps/formatters/clang-format exist
3. Run `bazel run //src/workerd/tests:bench-util --config=benchmark`
4. Validate that bazel-bin/build folder is now removed.
5. Validate that bazel-out/k8-dbg/bin/build/deps/formatters/clang-format exist
6. Try to format using "format" functionality in the IDE, it will fail

danlapid:
> > I don't understand the problem, if you haven't run format.py then the formatters are not available and they shouldn't be.
> 
> @danlapid Please run the following
> 
> 1. Run `tools/cross/format.py`
> 2. Validate that bazel-bin/build/deps/formatters/clang-format exist
> 3. Run `bazel run //src/workerd/tests:bench-util --config=benchmark`
> 4. Validate that bazel-bin/build folder is now removed.
> 5. Validate that bazel-out/k8-dbg/bin/build/deps/formatters/clang-format exist

Ok so the problem you have is that the `--config=benchmark` is overriding your bazel-bin and replaces the symlink from k8-dbg to k8-benchmark or whatever.

That's a problem specific to that flow and not general/common.

The way to fix that is not by breaking everyone else but rather by simply moving `--config=benchmark` to your `~/.bazelrc` when working on benchmark config.
That way, even the formatters will be built into the `--config=benchmark` config directory

anonrig:
> Ok so the problem you have is that the --config=benchmark is overriding your bazel-bin and replaces the symlink from k8-dbg to k8-benchmark or whatever.

The problem is not specific to benchmarks, if you don't build all targets, you'll never have the formatters, unless you run format.py script. And in that scenario, whenever you run a non-debug mode, you'll always encounter this issue.

For example,

1. Clone the repository
2. Run `bazel build //src/workerd/server:workerd`
3. Validate that bazel-bin/build doesn't exist

Or

1. Run `bazel build //src/workerd/server:workerd --config=asan`
2. Validate that bazel-bin/build doesn't exist

Or

1. Run `bazel build //src/workerd/server:workerd --config=release`
2. Validate that bazel-bin/build doesn't exist

danlapid:
> 2\. Validate that bazel-bin/build doesn't exist

Yep, as I said, if you want the formatters to be available after using a non-debug config, put the config into `~/.bazelrc`

PR #4476, opened 2025-07-07
improve typescript types for nodejs_compat

anonrig:


src/node/internal/internal_http_constants.ts
```diff
@@ -3,45 +3,47 @@
...
+    'TRACE',
+    'UNBIND',
+    'UNLINK',
+    'UNLOCK',
+    'UNSUBSCRIBE',
+  ].toSorted()
+);
 
-export const STATUS_CODES: Record<string, string> = {
+export const STATUS_CODES: Record<string, string> = Object.freeze({
```
guybedford: Node.js STATUS_CODES _are mutable_, so I wonder if we might hit issues with this?

guybedford (approved):

src/node/internal/internal_http_constants.ts
jasnell: +1... we should not freeze these. There is code in the ecosystem that mutates this.

src/node/internal/internal_http_constants.ts
```diff
@@ -3,45 +3,47 @@
...
-  'SOURCE',
-  'SUBSCRIBE',
-  'TRACE',
-  'UNBIND',
-  'UNLINK',
-  'UNLOCK',
-  'UNSUBSCRIBE',
-].toSorted();
+export const METHODS = Object.freeze(
+  [
```
jasnell: There is code in the ecosystem that mutates `METHODS`, this should not be frozen

PR #4477, opened 2025-07-07
Structured Logging

danlapid:
This commit adds structured logging support to workerd. This is disabled by default but can be enabled with the structuredLogging configuration option.

When enabled workerd will output json log messages to stdout. The two sources of logs are c++ logs/errors and js errors.

### DIFF
--- a/src/node/internal/internal_inspect.ts
+++ b/src/node/internal/internal_inspect.ts
@@ -2971,11 +2971,29 @@ export function stripVTControlCharacters(str: string): string {
 
 // Called from C++ on `console.log()`s to format values
 export function formatLog(
-  ...args: [...values: unknown[], colors: boolean]
+  ...args: [
+    ...values: unknown[],
+    colors: boolean,
+    structuredLogging: boolean,
+    level: string,
+  ]
 ): string {
-  const inspectOptions: InspectOptions = { colors: args.pop() as boolean };
+  const level = args.pop() as string;
+  const structuredLogging = args.pop() as boolean;
+  const colors = args.pop() as boolean;
+  const inspectOptions: InspectOptions = { colors };
+
   try {
-    return formatWithOptions(inspectOptions, ...args);
+    const message = formatWithOptions(inspectOptions, ...args);
+    if (structuredLogging) {
+      return JSON.stringify({
+        timestamp: Date.now(),
+        level,
+        message,
+      });
+    } else {
+      return message;
+    }
   } catch (err) {
     return `<Formatting threw (${isError(err) ? err.stack : String(err)})>`;
   }
--- a/src/workerd/io/worker.c++
+++ b/src/workerd/io/worker.c++
@@ -61,6 +61,23 @@ namespace workerd {
 
 namespace {
 
+constexpr kj::StringPtr logLevelToString(LogLevel level) {
+  switch (level) {
+    case LogLevel::DEBUG_:
+      return "debug";
+    case LogLevel::INFO:
+      return "info";
+    case LogLevel::LOG:
+      return "log";
+    case LogLevel::WARN:
+      return "warn";
+    case LogLevel::ERROR:
+      return "error";
+    default:
+      return "log";
+  }
+}
+
 void headersToCDP(const kj::HttpHeaders& in, capnp::JsonValue::Builder out) {
   std::map<kj::StringPtr, kj::Vector<kj::StringPtr>> inMap;
   in.forEach([&](kj::StringPtr name, kj::StringPtr value) {
@@ -559,6 +576,7 @@ struct Worker::Isolate::Impl {
           oldCurrentApi(currentApi),
           limitEnforcer(isolate.getLimitEnforcer()),
           consoleMode(isolate.consoleMode),
+          structuredLogging(isolate.structuredLogging),
           lock(isolate.api->lock(stackScope)) {
       WarnAboutIsolateLockScope::maybeWarn();
 
@@ -606,7 +624,7 @@ struct Worker::Isolate::Impl {
         i.get()->contextCreated(
             v8_inspector::V8ContextInfo(context, 1, jsg::toInspectorStringView("Worker")));
       }
-      Worker::setupContext(*lock, context, consoleMode);
+      Worker::setupContext(*lock, context, consoleMode, structuredLogging);
     }
 
     void disposeContext(jsg::JsContext<api::ServiceWorkerGlobalScope> context) {
@@ -642,6 +660,10 @@ struct Worker::Isolate::Impl {
     const IsolateLimitEnforcer& limitEnforcer;  // only so we can call getIsolateStats()
 
     ConsoleMode consoleMode;
+
+    // When structuredLogging is true AND consoleMode is STDOUT js logs will be emitted to STDOUT
+    // as newline separated json objects.
+    bool structuredLogging;
 
    public:
     kj::Own<jsg::Lock> lock;
@@ -984,12 +1006,14 @@ Worker::Isolate::Isolate(kj::Own<Api> apiParam,
     kj::StringPtr id,
     kj::Own<IsolateLimitEnforcer> limitEnforcerParam,
     InspectorPolicy inspectorPolicy,
-    ConsoleMode consoleMode)
+    ConsoleMode consoleMode,
+    bool structuredLogging)
     : metrics(kj::mv(metricsParam)),
       id(kj::str(id)),
       limitEnforcer(kj::mv(limitEnforcerParam)),
       api(kj::mv(apiParam)),
       consoleMode(consoleMode),
+      structuredLogging(structuredLogging),
       featureFlagsForFl(makeCompatJson(decompileCompatibilityFlagsForFl(api->getFeatureFlags()))),
       impl(kj::heap<Impl>(*api, *metrics, *limitEnforcer, inspectorPolicy)),
       weakIsolateRef(WeakIsolateRef::wrap(this)),
@@ -1454,8 +1478,10 @@ void setWebAssemblyModuleHasInstance(jsg::Lock& lock, v8::Local<v8::Context> con
       module->DefineOwnProperty(context, v8::Symbol::GetHasInstance(lock.v8Isolate), function));
 }
 
-void Worker::setupContext(
-    jsg::Lock& lock, v8::Local<v8::Context> context, Worker::ConsoleMode consoleMode) {
+void Worker::setupContext(jsg::Lock& lock,
+    v8::Local<v8::Context> context,
+    Worker::ConsoleMode consoleMode,
+    bool structuredLogging) {
   // Set WebAssembly.Module @@HasInstance
   setWebAssemblyModuleHasInstance(lock, context);
 
@@ -1471,9 +1497,9 @@ void Worker::setupContext(
         lock.v8Isolate, jsg::check(console->Get(context, methodStr)).As<v8::Function>());
 
     auto f = lock.wrapSimpleFunction(context,
-        [consoleMode, level, original = kj::mv(original)](
+        [consoleMode, level, structuredLogging, original = kj::mv(original)](
             jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>& info) {
-      handleLog(js, consoleMode, level, original, info);
+      handleLog(js, consoleMode, level, structuredLogging, original, info);
     });
     jsg::check(console->Set(context, methodStr, f));
   };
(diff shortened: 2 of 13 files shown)


danlapid:
This PR is ready for review, it's in draft because I'm not sure if my tests will compile on Windows but other than that the PR is fully reviewable.

npaun:
Cool! I love structured logging

src/workerd/io/worker.c++
```diff
@@ -61,6 +61,23 @@ namespace workerd {
 
 namespace {
 
+const char* logLevelToString(LogLevel level) {
```
jasnell:
```suggestion
constexpr kj::StringPtr logLevelToString(LogLevel level) {
```

src/workerd/io/worker.c++
```diff
@@ -1928,9 +1952,12 @@ void Worker::handleLog(jsg::Lock& js,
     KJ_ASSERT(formatLogVal->IsFunction());
     auto formatLog = formatLogVal.As<v8::Function>();
 
+    const char* levelStr = logLevelToString(level);
```
jasnell: Might be easiest to just pass the number value through to JS and do the conversion to string there.

src/workerd/io/worker.c++
```diff
@@ -1928,9 +1952,12 @@ void Worker::handleLog(jsg::Lock& js,
...
     args[length] = v8::Boolean::New(js.v8Isolate, colors);
+    args[length + 1] = v8::Boolean::New(js.v8Isolate, structuredLogging);
+    args[length + 2] = jsg::v8StrIntern(js.v8Isolate, levelStr);
```
jasnell, line 1965: You could likely get away with just passing `levelStr` conditionally. If the value is passed as `undefined`, then assume `structuredLogging` is false, otherwise we assume `structuredLogging` is true.

src/workerd/io/worker.c++
danlapid: Is it easier? 😅

src/workerd/io/worker.c++
danlapid: Yeah I could but that sounds ugly and odd doesn't it? why not convey intent directly?

src/workerd/io/worker.c++
danlapid: The enum is defined in C++, it'll be easier to figure out where it is used and where should be updated if it is expanded if we keep it in c++, plus, I really don't think it'll be prettier code in JS

src/workerd/io/worker.c++
jasnell: Why odd? It's fairly idiomatic in JavaScript.

```c++
args[length + 1] = js.undefined();
if (structuredLogging) {
  args[length + 1] = js.strIntern(levelStr);
}
```

Then in JS

```js
if (level !== undefined) {
  // structured logging enabled
} else {
  // structured logging disabled
}
```

src/workerd/io/worker.c++
jasnell: Well, it's faster at least. With this, even with the interning, you're incurring an additional cost for the lookup of the interned string on every call. If you move that out to JS you skip that entirely and optimize it down to a simple switch statement with zero string comparisons. The enum is exceedingly unlikely to change so some minor duplication out in JS land isn't going to be that significant. Entirely optional tho.

src/workerd/server/workerd.capnp
```diff
@@ -84,6 +84,11 @@ struct Config {
   # A list of gates which are enabled.
   # These are used to gate features/changes in workerd and in our internal repo. See the equivalent
   # config definition in our internal repo for more details.
+
+  structuredLogging @5 :Bool = false;
+  # If true, logs will be emitted as JSON for structured logging.
+  # When false, logs use the traditional human-readable format.
+  # This affects the format of logs from KJ_LOG and exception reporting as well as js logs.
```
jasnell, line 91: It should probably be called out that this won't work for service worker syntax workers that are using the original module registry implementation.

src/workerd/io/worker.c++
danlapid: I just meant that it's kinda odd to use whether or not level is defined to convey whether or not to do structured logging.

Seems besides the point and confusing to a random reader.

Why work so hard to optimize away the additional argument? It's not like this is performance critical code that's running in production. Seems more right to prefer readability over a slight of performance.

src/workerd/server/server-test.c++
```diff
@@ -4914,5 +4916,138 @@ KJ_TEST("Server: Durable Object facets") {
...
+          modules = [
+            ( name = "main.js",
+              esModule =
+                `export default {
+                `  async fetch(request, env, ctx) {
+                `    console.log("This is a log message", { key: "value" });
+                `    console.info("This is an info message");
+                `    console.warn("This is a warning message");
+                `    console.error("This is an error message");
+                `    console.debug("This is a debug message");
```
jasnell, line 4972: It would be helpful if the tests also covered output like `console.log({a: 1})`.

src/workerd/server/server-test.c++
```diff
@@ -4914,5 +4916,138 @@ KJ_TEST("Server: Durable Object facets") {
...
+                `
+                `    try {
+                `      throw new Error("Test exception for structured logging");
+                `    } catch (e) {
+                `      console.error("Caught exception:", e.message);
+                `    }
```
jasnell, line 4979: Would be good to test the outputting the error itself... `console.error(e);`

src/node/internal/internal_inspect.ts
```diff
@@ -2971,11 +2971,29 @@ export function stripVTControlCharacters(str: string): string {
...
+    colors: boolean,
+    structuredLogging: boolean,
+    level: string,
+  ]
 ): string {
-  const inspectOptions: InspectOptions = { colors: args.pop() as boolean };
+  const level = args.pop() as string;
+  const structuredLogging = args.pop() as boolean;
+  const colors = args.pop() as boolean;
+  const inspectOptions: InspectOptions = { colors };
```
jasnell, line 2984: Mutating the `args` is unnecessary and incurs a perf penalty. Since we always pass these we could likely get by with a simple...

```js
const level = args[args.length - 1] as string;
const structuredLogging = args[args.length - 2] as boolean;
const colors = args[args.length - 3] as boolean;
```

src/node/internal/internal_inspect.ts
```diff
@@ -2971,11 +2971,29 @@ export function stripVTControlCharacters(str: string): string {
...
+
   try {
-    return formatWithOptions(inspectOptions, ...args);
+    const message = formatWithOptions(inspectOptions, ...args);
+    if (structuredLogging) {
+      return JSON.stringify({
+        timestamp: Date.now(),
+        level: level,
+        message: message,
+      });
```
jasnell:
```suggestion
      return JSON.stringify({
        timestamp: Date.now(),
        level,
        message,
      });
```

src/node/internal/internal_inspect.ts
danlapid: This was adapted from the way colors already worked and expanded for two more arguments, I didn't really start this.
The reason it was done that way before is because `args` is then expanded into 
`formatWithOptions`
in
```
const message = formatWithOptions(inspectOptions, ...args);
```

src/node/internal/internal_inspect.ts
danlapid: Sure, will change, seems like a missing linter rule though

src/workerd/io/worker.c++
```diff
@@ -642,6 +660,7 @@ struct Worker::Isolate::Impl {
     const IsolateLimitEnforcer& limitEnforcer;  // only so we can call getIsolateStats()
 
     ConsoleMode consoleMode;
+    bool structuredLogging;
```
anonrig, line 666: Can you add a comment to this line?

src/workerd/io/worker.c++
```diff
@@ -1805,12 +1828,13 @@ void Worker::processEntrypointClass(jsg::Lock& js,
...
     ConsoleMode consoleMode,
     LogLevel level,
+    bool structuredLogging,
     const v8::Global<v8::Function>& original,
     const v8::FunctionCallbackInfo<v8::Value>& info) {
   // Call original V8 implementation so messages sent to connected inspector if any
   auto context = js.v8Context();
   int length = info.Length();
-  v8::LocalVector<v8::Value> args(js.v8Isolate, length + 1);
+  v8::LocalVector<v8::Value> args(js.v8Isolate, length + 3);
```
anonrig, line 1843: Can you add a comment about this magical number change?

src/workerd/server/json-logger-test.c++
```diff
@@ -0,0 +1,91 @@
...
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <capnp/compat/json.h>
+#include <capnp/message.h>
+#include <kj/async-unix.h>
+#include <kj/io.h>
+
+#include <cstdio>
+#endif
```
anonrig:
```suggestion
#endif  // __linux__
```

src/workerd/server/json-logger-test.c++
```diff
@@ -0,0 +1,91 @@
...
+  codec.decode(jsonOutput, logEntry);
+
+  // Validate the parsed JSON structure
+  KJ_EXPECT(logEntry.getLevel() == log_schema::LogEntry::LogLevel::ERROR);
+  KJ_EXPECT(logEntry.getMessage() == "Test JSON message");
+  KJ_EXPECT(logEntry.getTimestamp() > 0);
+  auto source = logEntry.getSource();
+  KJ_EXPECT(kj::StringPtr(source.begin(), source.size()).endsWith("json-logger-test.c++:49"));
+}
+#endif
```
anonrig: Ditto

src/workerd/server/BUILD.bazel
```diff
@@ -270,3 +285,11 @@ kj_test(
         "@capnp-cpp//src/kj",
     ],
 )
+
+kj_test(
+    src = "json-logger-test.c++",
```
npaun, line 290: Maybe plop in a target_compatible_with linux only?

src/workerd/server/json-logger.c++
```diff
@@ -0,0 +1,86 @@
...
+  // Prevent infinite recursion if logging code itself logs
+  if (loggingInProgress) {
+    return;
+  }
+  loggingInProgress = true;
+  KJ_DEFER(loggingInProgress = false);
+
+  auto json = buildJsonLogMessage(severity, file, line, contextDepth, text);
+
+  fprintf(stdout, "%s\n", json.cStr());
```
npaun:
```suggestion
 puts(json.cStr());
```

src/workerd/server/server-test.c++
```diff
@@ -4914,5 +4916,145 @@ KJ_TEST("Server: Durable Object facets") {
       kj::Path({"3652ef6221834806dc8df802d1d216e27b7d07e0a6b7adf6cfdaeec90f06459a.4.sqlite"})));
 }
 
+#if __linux__
+// This test uses pipe2 and dup2 to capture stdout which is far easier on linux.
+#include <unistd.h>
```
npaun, line 4921: I feel weird about in an `include` in the middle of a file like this. I know it's legal...

npaun (approved):

src/workerd/server/BUILD.bazel
danlapid: I wanted to do that but can't because the definition of kj_test already has a target_compatible_with 😢

npaun: lol thanks kj_test

src/workerd/server/server-test.c++
kentonv: No, actually, it's not legal!

Includes work by literally replacing the include with the content of the file, but we're inside a namespace block here, so this is declaring all the contents inside our namespace instead of the global namespace. So the declarations do not actually match the C library.

I think the reason it seems to work is because:
1. Since it's a C header, all the symbols will be wrapped in an `extern "C"` block. This changes the symbol names for linking purposes to effectively drop the namespace prefix. So the linker ends up finding the symbols after all.
2. We don't happen to have any conflicting symbol names in our namespace. (Otherwise there would be compiler errors presumably.)

But the fact that this works is entirely coincidental and we shouldn't be relying on it.

kentonv: The correct thing to do is to place the include at the top of the file, in a separate `#if __linux__` guard.

npaun: Huh, for some reason I thought @danlapid had already changed this one.  Nice catch, @kentonv.

PR #4480, opened 2025-07-08
NodeJS Compat: Add process.stdin, stdout and stderr

guybedford:
This implements the new `enable-nodejs-process-v2` `stdin`, `stdout` and `stderr` compat.

We build off of the approach in https://github.com/cloudflare/workerd/pull/4486 where `process.stdout.write('foo')` and `writeFileSync(process.stdout.fd, 'foo')` are treated equally by making this a part of the virtual filesystem. One consequence of that is that this makes the process implementation dependent on the virtual filesystem, but I think when the FS flag is added it would be fine to make that imply the process flag and vice versa.

Without a refactoring of the logging system, the easiest approach was to call back into `console.log`. I'm not by any means very happy with this approach, but it avoids getting too deep into the weeds and lets us discuss and test the coherent design before considering further refactorings.

1. The main goal is to have sensible behaviour when the inspector is attached for edgeworker support, without introducing a whole new logging pipeline. (although perhaps we should?)
2. Therefore `process.stdout.write('foo\n')` with a newline call calls console log as if you did`console.log('foo')`.
3. `process.stdout` goes to `console.log`, while `process.stderr` goes to `console.error`
4. The buffer limit for a single `process.stdout.write()` call is 16KiB, over which we only write the first 16KiB per standard stream backpressure techniques in Node.js. We do not split newlines for large buffer writes, although we could split them up into separate log calls.
5. The buffer limit for a single line without newlines is 4KiB, over which the line will be truncated to the most recent contents (effectively ring buffer behaviour, although I did not implement an actual ring buffer for now).
6. Tracer attachment is supported naturally.
7. Structured logging when enabled, it is thus fully enabled for `process.stdout.write` providing JSON timestamps and log types for each line
8. When in inspector debug mode, the console.log calls log as kj log as console logs.
9. If there is no newline, we buffer until a newline is found and then call the console log. This way `process.stdout.write('a'); process.stdout.write('b'); process.stdout.write('\n')` will log like `console.log('ab')`.
10. When there is content in the line buffer, a microtask is scheduled to clear the line buffer if no logs with a newline are posted synchronously. This will then force a newline. `process.stdout.write('a'); await 0; process.stdout.write('b'); await 0;` is thus the same as `console.log('a'); console.log('b');` causing a split.
11. If there are multiple newlines like `process.stdout.write('a\nb\nc\n')` we split them up into separate calls


src/workerd/api/node/process.c++
```diff
@@ -137,4 +139,11 @@ void ProcessModule::exitImpl(jsg::Lock& js, int code) {
   js.logWarning(kj::str(err.get(js, "stack"_kj)));
 }
 
+void ProcessModule::writeAndFlush(jsg::Lock& js, int fd, jsg::BufferSource data) {
+  FILE* file = (fd == 1) ? stdout : stderr;
+  auto ptr = data.asArrayPtr();
+  fwrite(ptr.begin(), 1, ptr.size(), file);
+  fflush(file);
```
jasnell: This definitely is not going to work for the production environment. It's generally fine for workerd and local dev but given the sandbox and the way the production environment works, we can't -- and don't want -- to write user code logging out to the real stdout and stderr. Accordingly, I don't think we can land this as is. Instead, we likely need to have an internal callback that is configured specific to each environment much like we do with `console.log` output and other logs emitted via `jsg::Lock`.

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -786,3 +783,123 @@ export const processRejectionListeners = {
...
+    assert.strictEqual(
+      windowSize.length,
+      2,
+      'getWindowSize should return array of length 2'
+    );
+
+    // Test file descriptor properties (Node.js semantics)
+    assert.strictEqual(process.stdin.fd, 0, 'process.stdin.fd should be 0');
+    assert.strictEqual(process.stdout.fd, 1, 'process.stdout.fd should be 1');
+    assert.strictEqual(process.stderr.fd, 2, 'process.stderr.fd should be 2');
```
jasnell, line 956: We will need to account for these in the `node:fs` implementation, which makes use of a virtual `fd` table. Specifically, we'll want to be able to incorporate these into the VFS as virtual files in the internal fd table so that they can be written to / read from via the `node:fs` APIs the way we can in Node.js. This will also need to be addressed before this lands, I think.

guybedford: Ahh nice, would be great to see this implemented. I can definitely aim to add this to this PR once we've figured out the console hook correctly.

src/node/internal/public_process.ts
```diff
@@ -32,10 +34,103 @@ import {
...
+  setRawMode(_mode: boolean): this {
+    return this;
+  }
+}
+
+export const stdin = new ReadStream({
+  read(): void {
+    this.push(null);
+  },
+});
```
jasnell: In Node.js, `process.stdin`, `process.stdout`, and `process.stderr` are all *both* `stream.Readable` and `stream.Writable` instances. It is possible (weirdly and not-without-odd-side-effects) to attach `data` events to `process.stdin` and to call `process.stdin.write(...)`, etc.

src/node/internal/public_process.ts
```diff
@@ -32,10 +34,103 @@ import {
...
+
+function chunkToBuffer(
+  chunk: Buffer | ArrayBufferView | DataView | string,
+  encoding: BufferEncoding
+): Uint8Array {
+  if (typeof chunk === 'string') {
+    return new Uint8Array(Buffer.from(chunk, encoding));
```
jasnell: This is going to copy twice, once to create the `Buffer` and again to copy into the `Uint8Array`.

```suggestion
    const buf = Buffer.from(chunk, encoding);
    return new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength);
```

src/node/internal/public_process.ts
```diff
@@ -32,10 +34,103 @@ import {
...
+    this.fd = fd;
+  }
+  getColorDepth(_env?: object): number {
+    return 1;
+  }
+  hasColors(_count?: number, _env?: object): boolean {
+    return false;
+  }
+  clearLine(_dir: -1 | 0 | 1, callback?: () => void): boolean {
+    if (callback) nextTick(callback);
```
jasnell: These should implement the appropriate type checking. In Node.js, the `callback` is type checked before dispatching the `nextTick`

src/node/internal/public_process.ts
```diff
@@ -32,10 +34,103 @@ import {
...
+class ReadStream extends Readable {
+  fd: number = 0;
+  isTTY: boolean = false;
+  isRaw: boolean = false;
+  setRawMode(_mode: boolean): this {
+    return this;
+  }
+}
+
+export const stdin = new ReadStream({
```
jasnell: these would need to be added to the `_process` object at the end of this file too, correct?

src/workerd/api/node/process.c++
danlapid: Agreed that this can't work as is but,
Why can't it just be implemented in JS as literally just console.log calls?
Sure every write will be logged as a full line but I think that's fine, it's not like we need this to be perfect.

src/node/internal/public_process.ts
jasnell: Not to contradict myself ;-) ... but, it is a bit more complicated than that, of course https://nodejs.org/docs/latest/api/process.html#processstdout

> The process.stdout property returns a stream connected to stdout (fd 1). It is a [net.Socket](https://nodejs.org/docs/latest/api/net.html#class-netsocket) (which is a [Duplex](https://nodejs.org/docs/latest/api/stream.html#duplex-and-transform-streams) stream) unless fd 1 refers to a file, in which case it is a [Writable](https://nodejs.org/docs/latest/api/stream.html#writable-streams) stream.

src/workerd/api/node/process.c++
jasnell: The chief difficulty is in the fact that `console.log(...)` does not correctly handle partial multibyte sequences. For instance, compare:

```
process.stdout.write(Buffer.from([0xe2])); process.stdout.write(Buffer.from([0x82])); process.stdout.write(Buffer.from([0xac]));
// Which prints a single Euro character to stdout
```

vs.

```
console.log(Buffer.from([0xe2]).toString()); console.log(Buffer.from([0x82]).toString()); console.log(Buffer.from([0xac]).toString());
// Which prints garbage...
```

jasnell: Another difference here, like I mention in my [other comment](https://github.com/cloudflare/workerd/pull/4480#discussion_r2193494337) we would need the `node:fs` impl to correctly handle a case like `fs.write(process.stdout.fd, ...)` implemented at the native layer, which makes it a bit more complicated to have a pure-JS implementation on this.

src/node/internal/public_process.ts
guybedford: Yes, because these were previously defined as undefined it's already in there as well from previously.

src/node/internal/public_process.ts
guybedford: Good catch thanks.

src/node/internal/public_process.ts
guybedford: I've gone with the `SyncWriteStream` approach instead here now which aligns with a `nohup node app.js` workflow.

src/node/internal/public_process.ts
guybedford: Because we are now using a `SyncWriteStream` and no longer a TTY stream, none of these methods are needed anymore.

src/workerd/api/node/process.c++
guybedford: This is now using the file descriptor outputs.

guybedford:
I've updated up the update here with `console.log` and `console.error` diverting from the stdio stream on a newline split.

It's not elegant by any means but it now works. I've described the behaviours in detail in the main PR description and also suggested possible areas for discussion.

guybedford:
Added another change here to support writing buffers larger than 16KiB, but using the standard Node.js stream backpressure convention returning the buffer limit length as what was written and writing no more than that.

Also improved tests for line buffer overflow and buffer limits.

guybedford:
One outstanding question here is whether we should split up `process.stdout.write('foo\nbar')` as `consoleLog('foo'); consoleLog('bar')` as opposed to `consoleLog('foo\nbar')`. Currently we perform a single log call, only handling missing `\n` as forming a line buffer, as opposed to splitting on `\n`.

The output would be more consistent if we split on `\n`, but it's potentially a more costly approach? Perhaps better to be consistent?

src/workerd/io/worker-fs.c++
```diff
@@ -1632,12 +1632,42 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
+  auto context = js.v8Context();
+  auto console =
+      jsg::check(context->Global()->Get(context, jsg::v8StrIntern(js.v8Isolate, "console")));
+  KJ_ASSERT(console->IsObject());
+  auto consoleObj = console.As<v8::Object>();
+
+  // Choose the appropriate console method based on stdio type
+  const char* methodName = (type == VirtualFileSystem::Stdio::OUT) ? "log" : "error";
+  auto method = jsg::check(consoleObj->Get(context, jsg::v8StrIntern(js.v8Isolate, methodName)));
+  KJ_ASSERT(method->IsFunction());
```
jasnell: Probably not a great idea to use `KJ_ASSERT` here. The user code might override `console.log` or `console.error` and this would result in an "internal error". I mean, it's exceedingly unlikely but possible. Let's make this a JSG_REQUIRE instead?

src/workerd/io/worker-fs.c++
```diff
@@ -1632,12 +1632,42 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
+// - structured logging
+// - stdio output otherwise
+void writeStdio(jsg::Lock& js, VirtualFileSystem::Stdio type, kj::ArrayPtr<const kj::byte> bytes) {
+  auto chars = bytes.asChars();
+  size_t endPos = chars.size();
+  if (chars[endPos - 1] == '\n') endPos--;
+  auto context = js.v8Context();
+  auto console =
+      jsg::check(context->Global()->Get(context, jsg::v8StrIntern(js.v8Isolate, "console")));
+  KJ_ASSERT(console->IsObject());
```
jasnell: JSG_REQUIRE instead.

src/workerd/io/worker-fs.c++
```diff
@@ -1738,6 +1796,46 @@ class StdioFile final: public File, public kj::EnableAddRefToThis<StdioFile> {
...
+    microtaskScheduled = true;
+
+    // Create ephemeral callback with weak reference for safety
+    auto callback = js.wrapSimpleFunction(js.v8Context(),
+        [weakThis = weakThis->addRef()](jsg::Lock& js, const v8::FunctionCallbackInfo<v8::Value>&) {
+      weakThis->runIfAlive([&](StdioFile& self) {
+        self.microtaskScheduled = false;
+
+        if (self.stdoutLineBuffer.size() > 0) {
+          if (IoContext::hasCurrent()) {
```
jasnell: Just to simplify things slightly, we can probably get by with just a single `IoContext::hasCurrent()` check and do both writes within it/

src/workerd/api/node/tests/process-stdio-nodejs-test.js
```diff
@@ -0,0 +1,235 @@
...
+      process.stderr instanceof Writable,
+      'stderr should be instance of Writable'
+    );
+  },
+};
+
+export const processStdioWriteTest = {
+  async test() {
+    process.stdout.write('Test string write to stdout\n');
+    process.stderr.write('Test string write to stderr\n');
```
jasnell, line 66: Perhaps have a test that includes multiple `\n` instances in a single write?

src/workerd/io/worker-fs.c++
```diff
@@ -1681,41 +1711,69 @@ class StdioFile final: public File, public kj::EnableAddRefToThis<StdioFile> {
...
+            lineBuffer.addAll(remaining);
+          } else {
+            // New data alone exceeds limit, replace entire line buffer
+            if (remaining.size() >= MAX_LINE_BUFFER_SIZE) {
+              lineBuffer.clear();
+              lineBuffer.addAll(remaining.slice(remaining.size() - MAX_LINE_BUFFER_SIZE));
+            } else {
+              // Combined size exceeds limit, remove oldest data
+              auto toRemove = totalSize - MAX_LINE_BUFFER_SIZE;
+              kj::Vector<kj::byte> newBuffer;
```
jasnell: Nit: reserve the size?

src/workerd/io/worker-fs.c++
```diff
@@ -1738,6 +1796,46 @@ class StdioFile final: public File, public kj::EnableAddRefToThis<StdioFile> {
  private:
   VirtualFileSystem::Stdio type;
   mutable kj::Maybe<kj::String> maybeUniqueId;
+
+  static constexpr size_t MAX_LINE_BUFFER_SIZE = 4096;
+  static constexpr size_t MAX_WRITE_SIZE = 16 * 1024;
+  mutable kj::Vector<kj::byte> stdoutLineBuffer;
+  mutable kj::Vector<kj::byte> stderrLineBuffer;
```
jasnell: do we need both in each instance? We're creating one of these for each of stdout and stderr correct? And `getLineBuffer()` only ever returns one.

src/workerd/io/worker-fs.c++
```diff
@@ -1632,12 +1632,42 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
+  KJ_ASSERT(console->IsObject());
+  auto consoleObj = console.As<v8::Object>();
+
+  // Choose the appropriate console method based on stdio type
+  const char* methodName = (type == VirtualFileSystem::Stdio::OUT) ? "log" : "error";
+  auto method = jsg::check(consoleObj->Get(context, jsg::v8StrIntern(js.v8Isolate, methodName)));
+  KJ_ASSERT(method->IsFunction());
+  auto methodFunc = method.As<v8::Function>();
+
+  v8::Local<v8::Value> args[] = {jsg::v8StrIntern(js.v8Isolate, kj::str(chars.slice(0, endPos)))};
```
jasnell: Does `chars.first(endPos)` work here?

jasnell: Also, you should be able to do `js.str(chars.slice(0, endPos))` I believe, and avoid the extra allocation (`kj::str` heap allocates)

src/workerd/io/worker-fs.c++
```diff
@@ -1632,12 +1632,42 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
+// - inspector reporting
+// - structured logging
+// - stdio output otherwise
+void writeStdio(jsg::Lock& js, VirtualFileSystem::Stdio type, kj::ArrayPtr<const kj::byte> bytes) {
+  auto chars = bytes.asChars();
+  size_t endPos = chars.size();
+  if (chars[endPos - 1] == '\n') endPos--;
+  auto context = js.v8Context();
+  auto console =
+      jsg::check(context->Global()->Get(context, jsg::v8StrIntern(js.v8Isolate, "console")));
```
jasnell: `js.str("console")` should work here and be more compact

src/workerd/io/worker-fs.c++
```diff
@@ -1632,12 +1632,42 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
+
 // An StdioFile is a special file implementation used to represent stdin,
 // stdout, and stderr outputs. Writes are always forwarded to the underlying
 // logging mechanisms. Reads always return EOF (0-byte reads).
 class StdioFile final: public File, public kj::EnableAddRefToThis<StdioFile> {
  public:
-  StdioFile(VirtualFileSystem::Stdio type): type(type) {}
+  StdioFile(VirtualFileSystem::Stdio type)
+      : type(type),
+        weakThis(kj::rc<WeakRef<StdioFile>>(kj::Badge<StdioFile>(), *this)) {}
```
jasnell, line 1721: Is the `WeakRef` really necessary? When we create an StdioFile it will be kept alive until the `VirtualFileSystem` instance is destroyed which doesn't happen until the Isolate is destroyed. It's fine if we want to be extra safe but I always have to second guess the use of WeakRefs as a general rule :-)

jasnell:
> The output would be more consistent if we split on \n, but it's potentially a more costly approach? Perhaps better to be consistent?

I'd suggest consistency is best -- and easier to explain.

src/workerd/io/worker-fs.c++
```diff
@@ -1632,12 +1632,42 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
 
+// Write stdio via console.log. Somewhat convoluted, but this then supports:
+// - inspector reporting
+// - structured logging
+// - stdio output otherwise
+void writeStdio(jsg::Lock& js, VirtualFileSystem::Stdio type, kj::ArrayPtr<const kj::byte> bytes) {
+  auto chars = bytes.asChars();
+  size_t endPos = chars.size();
+  if (chars[endPos - 1] == '\n') endPos--;
+  auto context = js.v8Context();
```
jasnell: You can use `js.global()` here to get the `Global` as a `jsg::JsObject` that'll be a bit nicer to use in here.

src/workerd/io/worker-fs.c++
guybedford: I actually did hit this believe it or not - becuase we use a scheduled task to flush the line buffer it's possible to have liveness issues.

jasnell: got it.. ok. can I ask you to leave a TODO comment in here to remind me to revisit this. I'd like to see if there's a way we can do this without requiring the weakref.

src/workerd/io/worker-fs.c++
guybedford: I replaced all these with if statements rather simplified on jsg idioms instead.

src/workerd/io/worker-fs.c++
guybedford: Added thanks.

src/workerd/io/worker-fs.c++
guybedford: Well spotted.

src/workerd/io/worker-fs.c++
guybedford: Added, thanks.

src/workerd/api/node/tests/process-stdio-nodejs-test.js
guybedford: Thanks, added along with structured logging tests of the same.

src/workerd/io/worker-fs.c++
```diff
@@ -1690,12 +1690,39 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
+// Write stdio via console.log. Somewhat convoluted, but this then supports:
+// - inspector reporting
+// - structured logging
+// - stdio output otherwise
+void writeStdio(jsg::Lock& js, VirtualFileSystem::Stdio type, kj::ArrayPtr<const kj::byte> bytes) {
+  auto chars = bytes.asChars();
+  size_t endPos = chars.size();
+  if (chars[endPos - 1] == '\n') endPos--;
+
+  KJ_IF_SOME(console, js.global().get(js, js.str("console"_kj)).tryCast<jsg::JsObject>()) {
```
jasnell: This should work...

```suggestion
  KJ_IF_SOME(console, js.global().get(js, "console"_kj).tryCast<jsg::JsObject>()) {
```

src/workerd/io/worker-fs.c++
```diff
@@ -1690,12 +1690,39 @@ class DevRandomFile final: public File, public kj::EnableAddRefToThis<DevRandomF
...
+    auto method = console.get(js, type == VirtualFileSystem::Stdio::OUT ? "log"_kj : "error"_kj);
+    v8::Local<v8::Value> methodVal(method);
+    if (methodVal->IsFunction()) {
```
jasnell: you can do `method.isFunction(...)` and do the conversion to the `v8::Local<v8::Value>` within the if block body.

guybedford: Sure, added.

PR #4482, opened 2025-07-09
Lift spans in KV

jmorrell-cloudflare:
I'm proposing this change as an alternative to https://github.com/cloudflare/workerd/pull/4378 and https://github.com/cloudflare/workerd/pull/4356

The current code instruments spans in the `KvNamespace::getHttpClient` method, however while this cuts down on the amount of repetitive code needed, it ties our hands a bit because the span is no longer in scope after the result comes back, so we cannot annotate any thing about the result or any errors.

To fix this and prepare for more extensive instrumentation this lifts the existing span declarations into each of the exposed KV methods (get, getBulk, getWithMetadata, list, put, delete) while keeping the emitted spans the same to minimize disruption.


src/workerd/api/kv-instrumentation-test.js
```diff
@@ -67,7 +67,6 @@ export const test = {
       { closed: true, name: 'worker' },
       { closed: true, name: 'worker' },
       { closed: true, name: 'worker' },
-      { closed: true, name: 'worker' },
```
jmorrell-cloudflare, line 70: I'm not quite sure why the ordering changed here

src/workerd/api/kv.h
```diff
@@ -247,15 +257,15 @@ class KvNamespace: public jsg::Object {
...
-  // the limiter for that op type to be checked. If a string is used, that's used as the operation
-  // name for the HttpClient without any limiter enforcement.
+  // the limiter for that op type to be checked. If a string is used, there isn't any limiter
+  // enforcement.
   // NOTE: The urlStr is added to the headers as a non-owning reference and thus must outlive
   // the usage of the headers.
   kj::Own<kj::HttpClient> getHttpClient(IoContext& context,
       kj::HttpHeaders& headers,
-      kj::OneOf<LimitEnforcer::KvOpType, kj::LiteralStringConst> opTypeOrName,
+      kj::OneOf<LimitEnforcer::KvOpType, kj::LiteralStringConst> opTypeOrUnknown,
```
fhanau: Nit: This variable should keep its name right?

jmorrell-cloudflare: Yeah, good catch.

Since this is only used to enforce limits I think I'd like to refactor it out into its own `enforceLimits(opType)` method so that this is only dealing with the http client work. I'll do that in a follow-up PR.

src/workerd/api/kv-instrumentation-test.js
fhanau: I think this happens because the lifetime of the attached spans changed – instead of being owned by the worker interface they get attached to in getSubrequestNoChecks(), they are now owned by whatever function originally created the trace context (since it is only passed in as a reference and it stays alive there until it goes out of scope). This results in a longer lifetime than that of the core KV subrequest – to fix it, try to propagate the TraceContext itself instead of a reference to it (using kj::mv() to move it instead of passing a reference), and always attach observed spans in getSubrequestNoChecks() for both existing and newly created trace context.
Getting the lifetimes right is important so that the span duration does not suddenly change. Happy to elaborate on how to change to the non reference-based approach, it shouldn't be any more complex really.

jmorrell-cloudflare: That makes sense, though I'm not very familiar with thinking through lifetimes yet. Maybe we could pair on this in our 1:1 today?

fhanau: Based on discussing this further, we accept that the lifetime (and thus order of spans) changes slightly here. This is hard to avoid when trying to add span tags after a request a span is based on has completed.

PR #4484, opened 2025-07-09
add coverage and profile commands to justfile

anonrig:
Not yet fully complete, since it doesn't detect all tests.

image

### DIFF
--- a/build/wd_rust_binary.bzl
+++ b/build/wd_rust_binary.bzl
@@ -10,7 +10,8 @@ def wd_rust_binary(
         visibility = None,
         tags = [],
         cxx_bridge_src = None,
-        cxx_bridge_deps = []):
+        cxx_bridge_deps = [],
+        test_size = "small"):
     """Define rust binary.
 
     Args:
@@ -82,4 +83,5 @@ def wd_rust_binary(
             "@//build/config:no_build": ["@platforms//:incompatible"],
             "//conditions:default": [],
         }),
+        size = test_size,
     )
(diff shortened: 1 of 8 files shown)


build/wd_rust_binary.bzl
```diff
@@ -82,4 +82,5 @@ def wd_rust_binary(
             "@//build/config:no_build": ["@platforms//:incompatible"],
             "//conditions:default": [],
         }),
+        size = "large",
```
fhanau: We should not be unconditionally marking all rust tests as large

mikea (approved):

fhanau:
This is intriguing, I didn't know we had coverage support in workerd at all. Ideally, we'd implement support for the `bazel coverage` command instead, but that would be some more work (it may or may not work well outside of Google3) https://bazel.build/configure/coverage.

mikea:
> This is intriguing, I didn't know we had coverage support in workerd at all. Ideally, we'd implement support for the `bazel coverage` command instead, but that would be some more work (it may or may not work well outside of Google3) https://bazel.build/configure/coverage.

but this does use bazel coverage, doesn't it?

fhanau:
> > This is intriguing, I didn't know we had coverage support in workerd at all. Ideally, we'd implement support for the `bazel coverage` command instead, but that would be some more work (it may or may not work well outside of Google3) https://bazel.build/configure/coverage.
> 
> but this does use bazel coverage, doesn't it?

It does indeed! Last time I tried it seemed hard to work with, glad we have it working now

build/wd_rust_binary.bzl
anonrig: Agreed. Updating now. Thanks for the review Felix

anonrig:
> It does indeed! Last time I tried it seemed hard to work with, glad we have it working now

The output is not working 100% correctly but we can iterate from here.

PR #4486, opened 2025-07-09
Include support for virtual "stdio" file descriptors

jasnell:
Reserve fd's 0, 1, and 2 for stdin, stdout, and stderr virtual files.

These are intended to serve as the underlying basis for `process.stdin`, `process.stdout`, and `process.stderr`

### DIFF
--- a/src/workerd/api/node/tests/fs-access-test.js
+++ b/src/workerd/api/node/tests/fs-access-test.js
@@ -15,6 +15,11 @@ import {
   exists,
   promises,
   constants,
+  fstatSync,
+  closeSync,
+  openSync,
+  writeSync,
+  readSync,
 } from 'node:fs';
 
 const { F_OK, R_OK, W_OK, X_OK } = constants;
@@ -291,4 +296,43 @@ export const existsCallbackTest = {
   },
 };
 
+export const stdioFds = {
+  test() {
+    let stat1 = fstatSync(0);
+    let stat2 = fstatSync(1);
+    let stat3 = fstatSync(2);
+    ok(stat1.isFile());
+    ok(stat2.isFile());
+    ok(stat3.isFile());
+    closeSync(0);
+    closeSync(1);
+    closeSync(2);
+    // After closing the stdio file descriptors, they can still be used.
+    // That is, we don't actually close them.
+    stat1 = fstatSync(0);
+    stat2 = fstatSync(1);
+    stat3 = fstatSync(2);
+    ok(stat1.isFile());
+    ok(stat2.isFile());
+    ok(stat3.isFile());
+
+    // Writing to the stdio file descriptors is permitted, always in append mode.
+    // Position is ignored.
+    writeSync(0, 'Hello, stdin!\n', 1000);
+    writeSync(0, 'test', 500);
+    writeSync(1, 'Hello, stdin!\n', 1000);
+    writeSync(1, 'test', 500);
+    writeSync(2, 'Hello, stdin!\n', 1000);
+    writeSync(2, 'test', 500);
+
+    stat1 = fstatSync(0);
+    strictEqual(stat1.size, 0);
+
+    const buf = Buffer.alloc(20);
+    const bytesRead = readSync(0, buf, { position: 1000 });
+
+    const fd = openSync('/tmp/test.txt', 'w');
+    strictEqual(fd, 3);
+  },
+};
 // There is no promise version of exists in Node.js, so no test for it here.
--- a/src/node/internal/internal_fs_utils.ts
+++ b/src/node/internal/internal_fs_utils.ts
@@ -372,23 +372,21 @@ export function validateWriteArgs(
   if (isArrayBufferView(buffer)) {
     if (typeof offsetOrOptions === 'object' && offsetOrOptions != null) {
       ({
-        offset = 0,
-        length = buffer.byteLength - offset,
+        offset = buffer.byteOffset,
+        length = buffer.byteLength,
         position = null,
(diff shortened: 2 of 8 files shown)


anonrig (approved):

jasnell:
The key downside with this approach is that any data written to the files here will accumulate when it really ought to work more like a pipe. I think this works well enough to get started but we'll likely want to ensure that these don't endlessly accumulate data when written to.

guybedford:
Hmm I am a bit worried about the unbounded memory growth here as well. Never underestimate how much can be logged...

It would be ideal if we could consume the pipe to avoid memory bloat I think.

So I'm actually not sure if this is the right approach, or if we should rather just have a buffered streaming primitive that can also be assigned a file descriptor in the virtual filesystem.

jasnell:
The underlying VFS can easily support a "FIFO" type, I'd just need to implement it. Wouldn't take much. I'll update this PR to handle that.

danlapid:
Isn't it the case that all of this is accounted memory anyway? so it's not really unbounded.

Either way I don't think we really need this for stdio

guybedford (approved):
So long as we can eventually see this as a FIFO that is fine to me, I just wanted to be sure we have the right sort of streaming primitive. While I agree with Dan that we don't really need this stream today, structuring it as a FIFO stream would be beneficial to treat that as the primitive that does the line ending separated batching to the log calls on the other side, and knowing that is being done correctly for Node.js apps with massive amounts of logs. So if we don't do the FIFO work now, at least agreeing that is the general idea architecturally would help. Will leave that up to @jasnell, but otherwise has my approval.

jasnell:
Updated so that the stdio files act like FIFO's. See the updated test.

jasnell:
> Either way I don't think we really need this for stdio...

It's trivial to support and there's a fair amount of code out in the Node.js ecosystem that expects to be able to do things like `fs.write(process.stdout.fd, 'some data', cb)` ... This gives us that kind of primitive.

jasnell:
Also, yes, the files report to the external memory accounter so will count towards the isolate limit. They are also bound to an absolute max of `MAX(uint32_t)` bytes. Folks should generally avoid just writing blindly the stdio fds without consuming them, and we might consider a mode in which the i/o is fully transient so the data does not accumulate.

We can also consider possibly using a more constrained ring-buffer type of queue... specifically, set a max buffer size, it can always be added to but the equivalent number of bytes are always dropped from the front when the buffer is full.

Note that we have other options here also. The way these virtual files are set up, we don't have to accumulate the data in a buffer at all if we don't want. We can forward the writes on to any location we want. The operations are all called under the isolate lock so we can even just forward the writes out to JS if that's easier, or push those off to the `handleLog(...)` function just like we do with `console.log`... the point is we have options here and don't need to go with exactly what is in this PR... this PR just shows that it is trivially possible for us to support the write-via-node-fs-api case that is fairly common in the ecosystem

src/workerd/api/node/tests/fs-access-test.js
```diff
@@ -291,4 +296,45 @@ export const existsCallbackTest = {
...
 };
 
+export const stdioFds = {
+  test() {
+    let stat1 = fstatSync(0);
+    let stat2 = fstatSync(1);
+    let stat3 = fstatSync(2);
+    ok(stat1.isCharacterDevice());
+    ok(stat2.isCharacterDevice());
+    ok(stat3.isCharacterDevice());
```
guybedford: Shouldn't this be `isFIFO()`?

jasnell: I thought so originally also.. but 
image

guybedford: You need to explicitly do the test with a pipe - `node test.js | cat`. The results vary depending on REPL / `node test.js` / `node test.js > out` and `node test.js | cat`. Also been doing a bunch of these investigations today!

jasnell: Yes, I know it varies based on how it's called. I was aiming to align with the default. We have options here, tho. I think we probably need to write up the expected behaviors here.

guybedford: Right, I can get behind the default being `isCharacterDevice()`.

guybedford (approved):

jasnell:
@anonrig @guybedford ... the entire impl has been reworked based on the internal chat. Specifically, `FifoFile` was replaced with `StdioFile` which directs the output either directly to the `WorkerTracer` or to `KJ_LOG(...)` without any buffering. When going through it the buffering just added additional complexity that really didn't seem worth it. In workerd, the output will end up logging to the console. In both workerd and production, if there is a tail worker attached, the logging output will be directed to the tail worker as regular log entries. I'm currently not doing anything to deal with partial utf8 sequences. Those will just end up as mangled log outputs which, is likely fine for now. We can continue to tweak this as necessary, but it at least gives us something to build on.

@guybedford ... since `node:fs` is not enabled by default, if someone tries to output to `process.stdout` and `process.stderr` when `node:fs` is disabled, you can either choose to have that be non-op or redirect those as `console.log(...)` statements in JavaScript. When `node:fs` is enabled tho, it'll be best to direct those out as `node:fs` writes to the fd's.

fd 0 (stdin) will act as a /dev/null. All writes will just do nothing and all reads will return 0.

fd 1 and 2 will log writes and always return 0 on reads.

guybedford (approved):
Looks like the right foundation to me. Happy to work off this further for the process PR, and handle any further feedback there.

PR #4487, opened 2025-07-09
Rename LogLevel::DEBUG to avoid clashing with DEBUG macro

npaun:
The macro `DEBUG` is often defined during debug builds. For example, it's defined for anyone developing locally on a Mac. Because macros are a mess, this means that we can't use `DEBUG` as an enum constant, as it would clash. I added an ugly `L_` prefix to avoid the clash.

### DIFF
--- a/src/workerd/server/log-schema.capnp
+++ b/src/workerd/server/log-schema.capnp
@@ -28,7 +28,8 @@ struct LogEntry {
   # Context depth for nested operations (optional, only included if > 0)
 
   enum LogLevel {
-    debug @0 $Json.name("debug");
+    debug @0 $Cxx.name("debug_") $Json.name("debug");
+    # in C++, the constant will be DEBUG_ to avoid clashing with the DEBUG macro
     info @1 $Json.name("info");
     warning @2 $Json.name("warning");
     error @3 $Json.name("error");
--- a/src/workerd/server/json-logger.c++
+++ b/src/workerd/server/json-logger.c++
@@ -26,7 +26,7 @@ log_schema::LogEntry::LogLevel severityToLogLevel(kj::LogSeverity severity) {
     case kj::LogSeverity::FATAL:
       return log_schema::LogEntry::LogLevel::FATAL;
     case kj::LogSeverity::DBG:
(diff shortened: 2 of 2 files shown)


src/workerd/server/log-schema.capnp
```diff
@@ -28,10 +28,11 @@ struct LogEntry {
...
-    debug @0 $Json.name("debug");
-    info @1 $Json.name("info");
-    warning @2 $Json.name("warning");
-    error @3 $Json.name("error");
-    fatal @4 $Json.name("fatal");
+    lDebug @0 $Json.name("debug");
+    lInfo @1 $Json.name("info");
+    lWarning @2 $Json.name("warning");
+    lError @3 $Json.name("error");
+    lFatal @4 $Json.name("fatal");
```
jasnell: Minor nit: Does it require renaming all of them? generally we've tried to limit these kinds of things to just the individual conflicts. The naming scheme here is a bit awkward also.

danlapid:
Should probably use `debug @1 $Cxx.name("debug_");` similar to worker-interface.capnp Level enum

npaun:
@jasnell @danlapid 
I fixed the PR

src/workerd/server/log-schema.capnp
```diff
@@ -28,7 +28,8 @@ struct LogEntry {
   # Context depth for nested operations (optional, only included if > 0)
 
   enum LogLevel {
```
fhanau, line 30: I wonder if we even need the new enum here and can't just go with the existing and largely identical `rpc::Trace::Log::Level` in `src/workerd/io/worker-interface.capnp`, where we already apply the Cxx.name trick? We can conveniently refer to it with a `using LogLevel = rpc::Trace::Log::Level;` alias, as done in trace.h.

danlapid: Almost identical, but not quite, the log level in worker-interface is based on JS style errors rather than c++ style errors.

JS has "log" C++ doesn't.
C++ has "fatal" JS doesn't.

danlapid (approved):

PR #4488, opened 2025-07-09
Add log for hibernation failures

danlapid:


anonrig (approved):

src/workerd/io/hibernation-manager.c++
```diff
@@ -146,7 +146,11 @@ void HibernationManagerImpl::acceptWebSocket(
 
   // Finally, we initiate the readloop for this HibernatableWebSocket and
   // give the task to the HibernationManager so it lives long.
-  readLoopTasks.add(handleReadLoop(refToHibernatable));
+  readLoopTasks.add(handleReadLoop(refToHibernatable).catch_([](kj::Exception&& e) {
+    if (isInterestingException(e)) {
+      LOG_EXCEPTION("HibernationManagerImpl::handleReadLoop", e);
+    }
+  }));
```
jclee, line 153: This works fine... I suppose an alternate approach would be to put the implementation in the `DisconnectHandler onDisconnect` that we pass as `readLoopTasks`'s error handler -- which doesn't currently do anything (and has a comment claiming it doesn't need to).

PR #4489, opened 2025-07-09
EW-9285 Fix STW stream cancellation warnings further

fhanau:
The recent drain() fix causes us to drain properly when there is an exception, but also caused (largely benign) stream cancellation warnings on RPC-based requests (where we need to await some TailStreamTarget promises before returning) to reappear. finishScheduled() allows us to get the proper behavior in either case.

Essentially a follow-up to https://github.com/cloudflare/workerd/pull/4415, which solved some problems but wasn't a perfect solution either.


fhanau:
Obsoleted by https://github.com/cloudflare/workerd/pull/4564.

PR #4494, opened 2025-07-10
Refactor headers and improve performance

jasnell:
This is a refactor of the Headers implementation that improves construction benchmark performance.... The revised impl passes all WPT's and other tests and reduces per-header memory usage and memory accounting overhead while we're at it. Offering this as input to https://github.com/cloudflare/workerd/pull/4472


anonrig:
Can you move to the new file on a different PR? It's not possible to see what has changed when you move the whole implementation to a new file.

jasnell:
The move to a different file is in a different commit already. That said, pretty much the entire impl is changed through these series of commits.

src/workerd/jsg/value.h
```diff
@@ -529,7 +529,8 @@ class StringWrapper {
     if constexpr (kj::isSameType<kj::String, T>()) {
       return js.accountedKjString(kj::mv(buf));
     } else if constexpr (kj::isSameType<ByteString, T>()) {
-      return js.accountedByteString(kj::mv(buf));
+      //return js.accountedByteString(kj::mv(buf));
```
anonrig:
```suggestion
```

PR #4500, opened 2025-07-11
[nfc] Add WD_STRONG_BOOL() macro

harrishancock:
This commit defines a macro which allows us to define "strong boolean" types. In particular, these types never implicitly convert to or from `bool`, or each other. This allows for greater type safety, especially in places where functions accept multiple `bool` parameters right next to each other. The call sites of such functions are notoriously difficult to maintain, as switching the order of two `bool` parameters will change the semantics of every call site, but will not introduce any compilation failures to alert you to the change.

A common way to define a strong boolean is to use an `enum class Foo: bool { NO, YES }` type. That pattern is a big improvement in terms of type safety over standard booleans. It has a couple flaws, however:

- Contextual boolean conversion is impossible
- Uninitialized values remain possible

The lack of contextual boolean conversion is a significant pain point preventing our adoption -- no one wants to write `if (foo == Foo::YES)` where they would normally just write `if (foo)`. The uninitialized value issue is not a disadvantage over regular booleans, but is unfortunate.

This commit introduces a macro, WD_STRONG_BOOL(T), which defines a more ergonomic strong bool than the `enum class` pattern can provide. The resulting types provide:

- Two nested constants, `NO` and `YES`, to be used as "literals"
- Contextual boolean conversion
- Explicit conversion to/from booleans
- Protection against uninitialized values
- Type-preserving logical operators
- Comparison operators via the spaceship operator
- KJ stringification support

### DIFF
--- a/src/workerd/util/strong-bool-test.c++
+++ b/src/workerd/util/strong-bool-test.c++
@@ -1,0 +1,180 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the Apache 2.0 license found in the LICENSE file or at:
+//     https://opensource.org/licenses/Apache-2.0
+
+#include "strong-bool.h"
+
+#include <kj/test.h>
+
+namespace workerd {
+
+WD_STRONG_BOOL(Strongbad);
+WD_STRONG_BOOL(Burninator);
+
+constexpr Strongbad giveStrongbad() {
+  return Strongbad::NO;
+}
+constexpr Burninator giveBurninator() {
+  return Burninator::YES;
+}
+constexpr void takeStrongbad(Strongbad strongbadValue) {}
+constexpr void takeBurninator(Burninator burninatorValue) {}
+
+// TODO(soon): We don't have negative compile tests that I know of :(
+KJ_TEST("WD_STRONG_BOOL compile failures") {
+  [[maybe_unused]] Strongbad strongbadValue{Strongbad::NO};
+  [[maybe_unused]] Burninator burninatorValue{Burninator::YES};
+  [[maybe_unused]] bool booleanValue = false;
+  [[maybe_unused]] int integerValue = 123;
+
+  // No uninitialized values
+  //Strongbad uninitialized;
+
+  // No implicit conversion from bool or int
+  //strongbad = false;
+  //strongbad = 123;
+  //strongbad = booleanValue;
+  //strongbad = integerValue;
+  //Strongbad s1 = false;
+  //Strongbad s2 = 123;
+
+  // No implicit conversion to bool or int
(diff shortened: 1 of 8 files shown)


src/workerd/util/strong-bool-test.c++
```diff
@@ -0,0 +1,180 @@
...
+
+KJ_TEST("WD_STRONG_BOOL can be stringified") {
+  constexpr Strongbad strongbadNo{Strongbad::NO};
+  constexpr Strongbad strongbadYes{Strongbad::YES};
+  constexpr auto stringNo = kj::_::STR * strongbadNo;
+  constexpr auto stringYes = kj::_::STR * strongbadYes;
+  static_assert(stringNo == "Strongbad::NO"_kjc);
+  static_assert(stringYes == "Strongbad::YES"_kjc);
+
+  static_assert(kj::_::STR * Burninator::NO == "Burninator::NO"_kjc);
```
harrishancock, line 176: @sohpeach

src/workerd/util/strong-bool-test.c++
```diff
@@ -0,0 +1,180 @@
...
+  static_assert(kj::isSameType<decltype(Strongbad::YES), const Strongbad>());
+
+  // ... until you use it in an explicitly boolean context.
+  static_assert(kj::isSameType<decltype(!strongbadNo), bool>());
+  static_assert(kj::isSameType<decltype(strongbadNo && burninatorYes), bool>());
+  static_assert(kj::isSameType<decltype(strongbadNo || burninatorYes), bool>());
+
+  // These are all contextually converted to `bool`s.
+  if (strongbadNo) {}
+  if (strongbadNo && burninatorYes) {}
```
harrishancock, line 94: @danlapid @ketanhwr this syntax works.

src/workerd/util/strong-bool.h
```diff
@@ -0,0 +1,60 @@
...
+// - Contextual boolean conversion: `if (strongBool)`, `while (strongBool)`, `!strongBool`
+//
+// `StrongBool` supports the full suite of comparison and logical operators. Note that ! is
+// supported via contextual conversion (`explicit operator bool()`), rather than `operator!()`.
+//
+// Logical operators (&& and ||) preserve the type of their operands when the operands are the same
+// `StrongBool` type. Otherwise, contextual boolean conversion applies, and the operands will be
+// converted to `bool` before the operator is invoked.
+#define WD_STRONG_BOOL(Type)                                                                       \
+  class Type {                                                                                     \
```
jasnell:
```suggestion
  class Type final {                                                                               \
```

danlapid (approved):

PR #4502, opened 2025-07-11
Enforce a smaller max file size for vfs files

jasnell:


src/workerd/api/node/tests/fs-cp-test.js
```diff
@@ -1915,7 +1915,7 @@ export const deepDirectoryStructureTestsCallback = {
       `${sourceRoot}/level1/duplicate_file.txt`,
       'source level1 duplicate content'
     );
-
+    console.log('.......');
```
danlapid: Nit

src/workerd/io/worker-fs.c++
```diff
@@ -1228,8 +1247,9 @@ class DevNullFile final: public File, public kj::EnableAddRefToThis<DevNullFile>
     };
   }
 
-  kj::Rc<File> clone(jsg::Lock&) override {
-    return addRefToThis();
+  kj::OneOf<FsError, kj::Rc<File>> clone(jsg::Lock&) override {
+    kj::Rc<File> ref = addRefToThis();
```
danlapid, line 1251: Nit
```suggestion
    return kj::Rc<File>(addRefToThis());
```

src/workerd/io/worker-fs.c++
```diff
@@ -695,13 +703,22 @@ class FileImpl final: public File {
     }
   }
 
-  kj::Rc<File> clone(jsg::Lock&) override {
+  kj::OneOf<FsError, kj::Rc<File>> clone(jsg::Lock& js) override {
+    auto maxSize = Worker::Isolate::from(js).getLimitEnforcer().getBlobSizeLimit();
     KJ_SWITCH_ONEOF(ownedOrView) {
       KJ_CASE_ONEOF(owned, kj::Array<kj::byte>) {
-        return kj::rc<FileImpl>(kj::heapArray<kj::byte>(owned));
+        if (owned.size() > maxSize) [[unlikely]] {
```
danlapid, line 710: I don't get this case, if is is bigger than maxSize then the isolate is condemned and presumably it's not bigger by a lot as that would cause termination, feels unnecessary to add this check here.

src/workerd/api/node/tests/fs-cp-test.js
jasnell:
```suggestion
```

src/workerd/io/worker-fs.c++
jasnell: It generally should never happen (which is why it's marked explicitly `[[unlikely]]`... but good as an additional check in case regressions are introduced at some time.

src/workerd/io/worker-fs.c++
jasnell: I actually prefer the way I have with the implicit cast and return.

src/workerd/io/worker-fs.c++
danlapid: I really don't like how much uglier it makes the API, all of these ONEOF and having clone be fallible feels really wrong imo.
Esp when it doesn't provide much value.
We could honestly just have this be a JSG_REQUIRE considering how unlikely this is.
But I'll approve to not block.

danlapid (approved):

src/workerd/io/worker-fs.c++
jasnell: the node.js and WebFS APIs have different expectations for errors. We cannot really just use JSG_REQUIRE's in these lower bits as that makes ensuring the appropriate errors are thrown via each of the APIs rather more cumbersome.

PR #4504, opened 2025-07-11
http headers eliminate key field

anonrig:
This replaces the separate key field with on demand case insensitive comparison and hashcode. This significantly reduces the number of allocations per every headers object at the cost of a small CPU overhead. The overhead is not present during lookups or inserts, because the old version used toLower. The only potential performance degredation is in iteration functions that needs to sort keys now. However, the number of keys are usually very low and sorting will require just a few comparisons. All of this is completely outbalanced by kj::HashMap performance gains comparing to std::map.

Co-Authored-By: @mikea


guybedford (approved):

src/workerd/api/http.c++
```diff
@@ -139,8 +139,44 @@ jsg::Optional<kj::StringPtr> getCacheModeName(Request::CacheMode mode) {
   KJ_UNREACHABLE;
 }
 
+constexpr char toLowerAscii(unsigned char c) {
+  // Create a mask that's 0x20 for uppercase letters, 0 otherwise
+  unsigned char is_upper = ((c - 'A') <= ('Z' - 'A'));
```
npaun: I feel like this thing will return true for a number like `5`. 
e.g. in this sloppy Python translation `(ord('5') - ord('A')) <= (ord('Z') - ord('A'))`

anonrig: I don't want to add this but this works:

```
static_assert(toLowerAscii('5') == '5', "toLowerAscii('5') should return '5'");
static_assert(toLowerAscii('A') == 'a', "toLowerAscii('A') should return 'a'");
static_assert(toLowerAscii('Z') == 'z', "toLowerAscii('Z') should return 'z'");
static_assert(toLowerAscii('a') == 'a', "toLowerAscii('a') should return 'a'");
static_assert(toLowerAscii('z') == 'z', "toLowerAscii('z') should return 'z'");
```

guybedford: It seems this works because char is unsigned and wraps around flipping the sign of the subtraction.

jasnell:
We have a couple of PRs that are touching this same code. Before merging this it would be good to know what approach you plan to take on the refactor. For instance, you signed off on my PR but this wouldn't apply in that case. Do you have an idea of what your end state is? Should I close my other PR?

anonrig:
> We have a couple of PRs that are touching this same code. Before merging this it would be good to know what approach you plan to take on the refactor. For instance, you signed off on my PR but this wouldn't apply in that case. Do you have an idea of what your end state is? Should I close my other PR?

@jasnell My goal was to split the changes into multiple pull-requests to land them easier. TBH moving the implementation to a different file made things a little bit harder, and I'm not sure if we should land this first, or your pull-request. I'll spend next week splitting https://github.com/cloudflare/workerd/pull/4472 into multiple ones that are easier to review&land. Your pull-request has changes that are beneficial and doesn't optimize the things we (@mikea) did, so we should definitely land all of them.

Edit: To summarize, I'll leave the decision of landing this or your PR first to you. Both of them works for me.

jasnell:
The test failures suggest that the ordering of the iterators (keys, values, and entries) are incorrect on Windows or the header names are being lowercased incorrectly. That's just a guess tho as I haven't dug in

jasnell:
Ok, let me go ahead and land mine then since it's the larger set of changes. The rest appears to be fairly tactical and incremental. I do plan on continuing to refine the memory accounting to reduce more of the overhead there.

PR #4505, opened 2025-07-11
update workerd-cxx

mikea:
Recent version doesn't have warnings


anonrig (approved):

anonrig:
We might need to increase the iteration count of this particular benchmark...

PR #4510, opened 2025-07-12
ActorSqlite: treat moved-later alarm rescheduling exceptions as not breaking output gate

jclee:
ActorSqlite captures the first exception thrown by an output gate promise, so that it can continue to throw the exception on subsequent storage operations. The capture is done via a TaskSet error handler, with the assumption that all its tasks came via the output gate.

However, the TaskSet had an additional source of tasks that did not come from the output gate: When we reschedule an alarm later, we intentionally do not take an output gate lock, since we don't need to hold the gate open for it to complete.  If an alarm rescheduling task failed, the error handler would set ActorSqlite's "broken" exception.  This would cause subsequent storage operations throw exceptions -- but the actor's output gate would not be broken, so subsequent requests would not automatically go to a new instance of the actor, allowing it to remain in the degraded state until the instance was reset for other reasons (inactivity, runtime resets, etc.).

This change moves the alarm rescheduling tasks to a separate TaskSet, such that they will not affect the tracking of the "broken" exception.

### DIFF
--- a/src/workerd/io/actor-sqlite.c++
+++ b/src/workerd/io/actor-sqlite.c++
@@ -36,7 +36,8 @@ ActorSqlite::ActorSqlite(kj::Own<SqliteDatabase> dbParam,
       hooks(hooks),
       kv(*db),
       metadata(*db),
-      commitTasks(*this) {
+      commitTasks(*this),
+      alarmLaterTasks(alarmLaterErrorHandler) {
   db->onWrite(KJ_BIND_METHOD(*this, onWrite));
   db->onCriticalError(KJ_BIND_METHOD(*this, onCriticalError));
   lastConfirmedAlarmDbState = metadata.getAlarm();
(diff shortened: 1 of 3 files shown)


src/workerd/io/actor-sqlite.c++
```diff
@@ -202,6 +203,8 @@ void ActorSqlite::onCriticalError(
     kj::StringPtr errorMessage, kj::Maybe<kj::Exception> maybeException) {
   // If we have already experienced a terminal exception, no need to replace it
   if (broken == kj::none) {
+    // TODO(someday): If we are setting the broken exception, do we also need to explicitly break
+    // the output gate?
```
jclee, line 207: I noticed that this "critical error" handler was another place where the ActorSqlite "broken" exception gets set, but it wasn't obvious to me that this would necessarily result in the output gate breaking or the actor getting reset.

Looking through the caller, I guess this only fires for certain [sqlite error conditions during a transaction](https://sqlite.org/c3ref/get_autocommit.html)?

jqmmes (approved):
LGTM

PR #4511, opened 2025-07-12
Improve/simplify memory accounting in worker-fs FileImpl

jasnell:


src/workerd/io/worker-fs.c++
```diff
@@ -780,26 +783,39 @@ class FileImpl final: public File {
...
+  }
+
+  kj::ArrayPtr<kj::byte> writableView() {
+    return KJ_REQUIRE_NONNULL(ownedOrView.tryGet<Owned>()).data.asPtr();
   }
 
-  kj::Array<kj::byte>& writableView() {
-    return KJ_REQUIRE_NONNULL(ownedOrView.tryGet<kj::Array<kj::byte>>());
+  jsg::ExternalMemoryAdjustment& getAdjustment() {
+    return KJ_REQUIRE_NONNULL(ownedOrView.tryGet<Owned>()).adjustment;
```
guybedford, line 809: Is the file definitely writable if this is called?

jasnell: Yep, if you look at the callsites, we always call `isWritable()` before calling this.

guybedford (approved):

PR #4520, opened 2025-07-14
[o11y] Expand set of supported span tags in Cache API

fhanau:
Downstream PR adding tests to follow.

### DIFF
--- a/src/workerd/api/cache.c++
+++ b/src/workerd/api/cache.c++
@@ -538,12 +538,42 @@ kj::Own<kj::HttpClient> Cache::getHttpClient(IoContext& context,
   auto userSpan = context.makeUserTraceSpan(operationName);
 
   userSpan.setTag("url.full"_kjc, kj::str(url));
-  // TODO(o11y): We add a tag for must-revalidate as a proof-of-concept for now (see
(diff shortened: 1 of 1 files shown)


src/workerd/api/cache.c++
```diff
@@ -538,12 +538,41 @@ kj::Own<kj::HttpClient> Cache::getHttpClient(IoContext& context,
   auto userSpan = context.makeUserTraceSpan(operationName);
 
   userSpan.setTag("url.full"_kjc, kj::str(url));
-  // TODO(o11y): We add a tag for must-revalidate as a proof-of-concept for now (see
-  // https://developers.cloudflare.com/cache/concepts/cache-control/#revalidation). Parse
-  // cacheControl to support other revalidation, cacheability and expiration directives too.
+  // TODO(o11y): Can we parse cacheControl more cleanly? For example, if tags are duplicated in the
```
fhanau, line 541: Open to suggestions here... I couldn't find any dedicated HTTP header parsing code in the codebase. I guess in this case all we'd need would be to have something like header.split(",").trimLeadingAndTrailingWhitespace() and then process the tokens in order but there's not really a split() function for KJ strings

PR #4522, opened 2025-07-14
[nfc] Add StrongBool::toBool()

harrishancock:
I belatedly recalled that I shouldn't be using functional-style casts, since they implicitly perform a const_cast. In this case, it's not a big risk, but in theory someone could change a strong bool type to some other type with a non-const `operator bool()`, causing a safety issue.

The type already supports static_cast, of course, but no one's going to want to use that.

### DIFF
--- a/src/workerd/util/strong-bool-test.c++
+++ b/src/workerd/util/strong-bool-test.c++
@@ -108,19 +108,19 @@ KJ_TEST("WD_STRONG_BOOL can be explicitly converted to and from `bool`") {
   Strongbad strongbadNo{Strongbad::NO};
   Strongbad strongbadYes{Strongbad::YES};
 
-  auto booleanValue = bool(strongbadNo);
+  auto booleanValue = strongbadNo.toBool();
   static_assert(kj::isSameType<decltype(booleanValue), bool>());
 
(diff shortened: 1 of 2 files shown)


jasnell (approved):

jasnell:
@anonrig ... unrelated to this PR, but I've noticed the SlowAPIWithLock benchmark appears to have fairly massive swings from one run to the next in some of these PRs. This PR, for instance, shouldn't really have impacted the benchmark at all but it's reporting an 18.9% improvement here. Likely something worth looking into at some point as I'm not sure how reliable that benchmark actually is given the variance.

anonrig:
> @anonrig ... unrelated to this PR, but I've noticed the SlowAPIWithLock benchmark appears to have fairly massive swings from one run to the next in some of these PRs. This PR, for instance, shouldn't really have impacted the benchmark at all but it's reporting an 18.9% improvement here. Likely something worth looking into at some point as I'm not sure how reliable that benchmark actually is given the variance.

@jasnell The problem should been fixed with https://github.com/cloudflare/workerd/pull/4525 which landed on main but this PR was not up to date with main. I think the iteration count is not reliable, and the PR should do some real work to avoid variance.

PR #4523, opened 2025-07-14
fix EventEmitter types

anonrig:
This fixes almost all typescript errors related to our EventEmitter implementation.

### DIFF
--- a/src/node/internal/events.ts
+++ b/src/node/internal/events.ts
@@ -35,8 +35,8 @@ import {
   ERR_UNHANDLED_ERROR,
 } from 'node-internal:internal_errors';
 import type {
-  EventEmitter as _EventEmitter,
   EventEmitterAsyncResource as _EventEmitterAsyncResource,
+  EventEmitter as _EventEmitter,
 } from 'node:events';
 import {
   validateAbortSignal,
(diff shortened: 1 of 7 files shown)


src/node/internal/events.ts
```diff
@@ -745,7 +760,9 @@ export function listenerCount(emitter: any, type: string | symbol) {
 }
 
 EventEmitter.prototype.eventNames = function eventNames() {
-  return this._eventsCount > 0 ? Reflect.ownKeys(this._events) : [];
+  return this._eventsCount > 0 && this._events !== undefined
+    ? Reflect.ownKeys(this._events)
+    : [];
```
jasnell: Nit: I believe you could simplify this a bit with...

```suggestion
  return this._eventsCount > 0 
    ? Reflect.ownKeys(this._events || {})
    : [];
```

anonrig: Agreed

src/node/internal/events.ts
```diff
@@ -745,7 +760,9 @@ export function listenerCount(emitter: any, type: string | symbol) {
 }
 
 EventEmitter.prototype.eventNames = function eventNames() {
-  return this._eventsCount > 0 ? Reflect.ownKeys(this._events) : [];
+  return this._eventsCount > 0 
+    ? Reflect.ownKeys(this._events || {})
+    : [];
```
anonrig:
```suggestion
  return this._eventsCount > 0 ? Reflect.ownKeys(this._events || {}) : [];
```

PR #4529, opened 2025-07-14
improve test coverage

anonrig:
Simplifies test coverage


build/wd_test.bzl
```diff
@@ -173,17 +174,39 @@ def _wd_test_impl(ctx):
...
         ),
+        instrumented_files_info,
     ]
 
 _wd_test = rule(
     implementation = _wd_test_impl,
     test = True,
     attrs = {
+        # Implicit dependencies used by Bazel to generate coverage reports.
+        "_lcov_merger": attr.label(
```
fhanau, line 202: Nit: Not sure if they help here, can Bazel synthesize these on its own?

fhanau (approved):

build/wd_test.bzl
anonrig: If I understood the documentation in here correctly, yes: https://bazel.build/extending/rules#test_rules

fhanau: Sounds good

.bazelrc
```diff
@@ -395,14 +395,25 @@ build:windows --copt='-Wno-unused-command-line-argument' --host_copt='-Wno-unuse
...
+build:coverage --combined_report=lcov
+build:coverage --strategy=TestRunner=remote,sandboxed,local
+build:coverage --strategy=CoverageReport=sandboxed,local
+build:coverage --experimental_use_llvm_covmap
+build:coverage --experimental_generate_llvm_lcov
+build:coverage --experimental_fetch_all_coverage_outputs
+build:coverage --collect_code_coverage
+build:coverage --instrumentation_filter="^//,-//external"
+build:coverage --remote_download_minimal
+build:coverage --define=tcmalloc=gperftools
```
fhanau, line 410: Why are we setting this? We don't necessarily build with tcmalloc, and never use the ancient gperftools version. If it is relevant for coverage, there should be a comment explaining such.

PR #4530, opened 2025-07-14
[o11y] STW API cleanup

fhanau:
- Drop cloudflare.r2.key
- Provide deseralized log messages, as with legacy tail worker
- Deserialize cfJson, only provide it when not empty

Downstream PR to follow.


src/workerd/io/trace-stream.c++
```diff
@@ -517,7 +519,7 @@ jsg::JsValue ToJs(jsg::Lock& js, const tracing::Log& log, StringCache& cache) {
   auto obj = js.obj();
   obj.set(js, TYPE_STR, cache.get(js, LOG_STR));
   obj.set(js, LEVEL_STR, ToJs(js, log.logLevel, cache));
-  obj.set(js, MESSAGE_STR, js.str(log.message));
+  obj.set(js, MESSAGE_STR, jsg::JsValue(js.parseJson(log.message).getHandle(js)));
```
jasnell, line 523: Nit.. for later... we should probably have a version of `js.parseJson` that just returns a `jsg::JsValue`

types/defines/trace.d.ts
```diff
@@ -149,7 +149,7 @@ interface Exception {
 interface Log {
   readonly type: "log";
   readonly level: "debug" | "error" | "info" | "log" | "warn";
-  readonly message: string;
+  readonly message: object;
```
jasnell, line 152: Are we certain that parsing the `message` will always produce an object? As far as I see in the `parseJson(...)` login in `trace-stream` we're not verifying that it is indeed an object that is the result of the parse?

src/workerd/io/trace-stream.c++
```diff
@@ -189,7 +189,9 @@ jsg::JsValue ToJs(jsg::Lock& js, const tracing::FetchEventInfo& info, StringCach
   obj.set(js, TYPE_STR, cache.get(js, FETCH_STR));
   obj.set(js, METHOD_STR, cache.get(js, kj::str(info.method)));
   obj.set(js, URL_STR, js.str(info.url));
-  obj.set(js, CFJSON_STR, js.str(info.cfJson));
+  if (info.cfJson.size() > 0) {
+    obj.set(js, CFJSON_STR, js.str(info.cfJson));
```
mar-cf: Still needs to be deserialized

src/workerd/io/trace-stream.c++
fhanau: Yup, figuring out how to do this took a while since in the equivalent code for LTW we don't need to convert to JsValue

types/defines/trace.d.ts
fhanau: Good question – in the LTW (see api/trace.c++ getTraceLogMessage()) we cast the return value of js.parseJson() to v8::Object, does that enforces that we're returning an object?

src/workerd/io/trace-stream.c++
fhanau: 👍  implemented in latest push

mar-cf (approved):

jasnell (approved):

src/workerd/io/trace-stream.c++
```diff
@@ -517,7 +519,8 @@ jsg::JsValue ToJs(jsg::Lock& js, const tracing::Log& log, StringCache& cache) {
   auto obj = js.obj();
   obj.set(js, TYPE_STR, cache.get(js, LOG_STR));
   obj.set(js, LEVEL_STR, ToJs(js, log.logLevel, cache));
-  obj.set(js, MESSAGE_STR, js.str(log.message));
+  // TODO(o11y): Check that we are always returning an object here
+  obj.set(js, MESSAGE_STR, jsg::JsValue(js.parseJson(log.message).getHandle(js)));
```
petebacondarwin, line 523: Is the `_STR` postfix still accurate here since this property is now an object?

petebacondarwin (commented):
This is a breaking change for users who currently expect a string for these properties, right?
What is the status of this STW feature? Are we allowed to break users?

src/workerd/io/trace-stream.c++
petebacondarwin: Similar question for the `CFJSON_STR` prop above.

fhanau:
> This is a breaking change for users who currently expect a string for these properties, right? What is the status of this STW feature? Are we allowed to break users?

This feature is still in development and currently gated behind the experimental "streaming-tail-worker" compatibility flag, so it is not accessible to external users.

src/workerd/io/trace-stream.c++
fhanau: MESSAGE_STR and similar parameters refer to a constant that is used to cache the name of the field to make V8 memory management more efficient, in this case it simply represents the string "message". Since the name of the variable (which is now an object) is still "message", this can remain as-is.

types/defines/trace.d.ts
fhanau: Added a TODO for now.

PR #4533, opened 2025-07-15
Only log ActorCache::deleteAll errors to sentry if they're interesting

ObsidianMinor:
Lots of disconnects in here.


fhanau (approved):

kentonv:
Can we please get review from the Durable Objects team on things like this?

I think we should not be seeing disconnects here as disconnects should have been retried upstream. DOs in general should not be getting reset due to disconnecting from storage, so this is actually a bug that should be addressed.

PR #4535, opened 2025-07-15
increase test coverage of node:http

anonrig:
Increases the test coverage of node:http modules, and adds IncomingMessage to node:http (reported by @vicb)


justfile
```diff
@@ -36,7 +36,7 @@ stream-test *args:
 
 # e.g. just node-test zlib
 node-test test_name *args:
-  just stream-test //src/workerd/api/node:tests/{{test_name}}-nodejs-test {{args}}
+  just stream-test //src/workerd/api/node/tests:{{test_name}}-nodejs-test {{args}}
```
npaun, line 39: Doesn't this break some other node tests?

npaun: Maybe we should make sure they're all named consistently.

anonrig: I added a build bazel file in another pr and forgot to update this.

npaun (approved):

PR #4540, opened 2025-07-15
Block completion of io context until tail stream target completes. Catch tail stream JS exception and set event outcome. Synchronously drain io context.

mar-cf:


src/workerd/io/trace-stream.c++
```diff
@@ -12,6 +12,9 @@
 namespace workerd::tracing {
 namespace {
 
+// Uniquely identifies js tail session failures
+constexpr kj::Exception::DetailTypeId TAIL_STREAM_JS_FAILURE = 0xfa110000bad0e0e0;
```
danlapid: you can use `capnp id` to generate a int that's a bit better

src/workerd/io/trace-stream.c++
```diff
@@ -946,7 +949,9 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
       if (doFulfill) {
         p = p.then(js, [&](jsg::Lock& js) { doneFulfiller->fulfill(); },
             [&](jsg::Lock& js, jsg::Value&& value) {
-          doneFulfiller->reject(KJ_EXCEPTION(DISCONNECTED, "Streaming tail session canceled"));
+          kj::Exception exception = KJ_EXCEPTION(FAILED, "Streaming tail session exception");
```
danlapid: Why not serialize value and attach it to the exception?
tbf, the JS exception really should just be the exception here, no?

src/workerd/io/trace-stream.c++
```diff
@@ -985,32 +990,19 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::run
...
-    // waitUntil() should allow extending execution on the server side even when the client
-    // disconnects.
-    waitUntilTasks.add(incomingRequest->drain().attach(kj::mv(incomingRequest)));
+  auto eventOutcome = co_await donePromise.exclusiveJoin(ioContext.onAbort()).then([]() {
+    return EventOutcome::OK;
+  }, [](kj::Exception&& e) {
+    if (e.getDetail(TAIL_STREAM_JS_FAILURE) != kj::none) {
+      return EventOutcome::EXCEPTION;
+    }
+    kj::throwRecoverableException(kj::mv(e));
```
danlapid, line 931: If we hit this case we now no longer drain, that feels wrong 🤔

danlapid:
Instead of the exception detail Id why not make the doneFulfiller just fulfill with an outcome object directly?
Also, we need to make sure this exception reaches the STW's tail worker.

src/workerd/io/trace-stream.c++
```diff
@@ -946,7 +949,9 @@ class TailStreamTarget final: public rpc::TailStreamTarget::Server {
       if (doFulfill) {
         p = p.then(js, [&](jsg::Lock& js) { doneFulfiller->fulfill(); },
             [&](jsg::Lock& js, jsg::Value&& value) {
-          doneFulfiller->reject(KJ_EXCEPTION(DISCONNECTED, "Streaming tail session canceled"));
+          kj::Exception exception = KJ_EXCEPTION(FAILED, "Streaming tail session exception");
+          exception.setDetail(TAIL_STREAM_JS_FAILURE, kj::heapArray<kj::byte>(0));
+          doneFulfiller->reject(kj::mv(exception));
```
fhanau, line 878: Does rejecting here with an exception other than DISCONNECTED still cause an exception to be logged to Sentry? If so we should avoid that since user-level exceptions (e.g. a faulty user script being deployed) should not lead to exceptions on our side.

fhanau: I'm also not sure if rejecting the doneFulfiller is the right action here since it only relates to the TailStreamTarget. Can you see the exception appear anywhere where the user would be able to take note of it?

src/workerd/io/trace-stream.c++
```diff
@@ -985,32 +990,19 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::run
...
-  // early, the donePromise should be fulfilled or rejected. Below we arrange for
-  // the IoContext for the tail stream request to remain alive until the donePromise
-  // is settled. We also block completion of the call to run on the same condition.
-  //
-  // Attaching the registerPendingEvent() to the promise is necessary to keep the
-  // IoContext alive during the times our tail worker is idle waiting for more
-  // events to be delivered.
-  kj::ForkedPromise<void> forked = donePromise.fork();
-  ioContext.addWaitUntil(forked.addBranch().attach(ioContext.registerPendingEvent()));
+  donePromise = donePromise.attach(ioContext.registerPendingEvent());
```
fhanau, line 917: 1) We should keep most of the comments above since they are still relevant.
2) I don't understand the reason for forking here that well – someone else should see if we can get rid of it safely (as proposed here)

src/workerd/io/trace-stream.c++
```diff
@@ -1039,6 +1031,7 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
   this->capFulfiller->fulfill(kj::mv(cap));
 
   try {
+    // TODO consume result
```
fhanau: I think to achieve this we need to extend the EventDispatcher RPC surface, otherwise the send() call won't return the result... I have filed #4564 which handles this. It overlaps somewhat with this PR, we should see if we can combine them effectively.

src/workerd/io/trace-stream.c++
```diff
@@ -985,32 +990,19 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::run
...
+    if (e.getDetail(TAIL_STREAM_JS_FAILURE) != kj::none) {
+      return EventOutcome::EXCEPTION;
+    }
+    kj::throwRecoverableException(kj::mv(e));
+    KJ_UNREACHABLE;
   });
-
-  co_await forked.addBranch().exclusiveJoin(ioContext.onAbort());
-  co_return WorkerInterface::CustomEvent::Result{.outcome = EventOutcome::OK};
+  co_await incomingRequest->drain();
```
fhanau: Open question based on our discussion yesterday: We need to determine if asynchronous draining (as ported from JSRPC and currently used) is needed so that all scheduled promises and the capability are cleaned up properly in case we have an exception. #4489 would handle that and ensure we drain with and without an exception.

src/workerd/io/trace-stream.c++
fhanau: Tested this locally – this approach looks sane.

fhanau:
I think with the sheer complexity of this piece of code, this should be split into several parts:
1) Converting donePromise to not be forked (I agree that this should be possible, or if it is needed I don't see why)
2) Propagating exceptions from JS promises and returning EventOutcome::EXCEPTION if appropriate
3) Changing how draining happens (we've discussed several approaches for this before, and I'm not sure if there's a great solution here)
4) Returning the right outcome code (this is perhaps better addressed by #4564, which also reports other outcomes and propagates the result via RPC)

I think the best approach here is to handle 4) with #4564, and address 1) and 2) in separate commits within this PR after rebasing on top of #4564 if that gets approved.

PR #4545, opened 2025-07-16
Update dependencies in samples/hello-wasm to support latest Rust

cp-20:
Hello maintainers!   I found a minor build issue in the hello-wasm sample, which can be easily fixed.

The original hello-wasm sample depended on an outdated version of wasm-bindgen that is no longer compatible with the latest Rust toolchain.

```
error: older versions of the `wasm-bindgen` crate are incompatible with current versions of Rust; please update to `wasm-bindgen` v0.2.88
```

This PR updates the dependencies to fix the issue and also includes a few minor type corrections.

### DIFF
--- a/samples/hello-wasm/Cargo.lock
+++ b/samples/hello-wasm/Cargo.lock
@@ -13,24 +13,15 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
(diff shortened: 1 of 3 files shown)


cp-20:

jasnell (approved):

PR #4546, opened 2025-07-16
Forbid use of TypeScript private keyword

ns476:


anonrig:
Amazing! I wanted to do this... Can we use https://typescript-eslint.io/rules/explicit-member-accessibility/ somehow or https://typescript-eslint.io/rules/parameter-properties/?

ns476:
> Amazing! I wanted to do this... Can we use https://typescript-eslint.io/rules/explicit-member-accessibility/ somehow or https://typescript-eslint.io/rules/parameter-properties/?

AFAICT the approach in this PR seems to be the recommended way to enforce use of the hash syntax (even though it's a bit ugly). Those rules do slightly different things but it could be a good idea to turn them on too.

src/wpt/harness/common.ts
```diff
@@ -31,16 +31,21 @@ export type PromiseTestFn = () => Promise<unknown>;
 
 export class FilterList {
   // Matches any input
+  // TODO(soon): Can we use the # syntax here?
+  // eslint-disable-next-line no-restricted-syntax
```
anonrig, line 35: We can replace everything private with #. They're literally the same thing but in Typescript vs. Javascript native syntax.

ns476: Discussed in chat and as they are not quite 100% the same, fine to merge with TODOs and fix incrementally

src/cloudflare/internal/ai-api.ts
```diff
@@ -128,13 +128,17 @@ function findReadableStreamKeys(
...
 export class Ai {
+  // TODO(soon): Can we use the # syntax here?
+  // eslint-disable-next-line no-restricted-syntax
   private readonly fetcher: Fetcher;
 
   /*
    * @deprecated this option is deprecated, do not use this
    */
   // @ts-expect-error this option is deprecated, do not use this
+  // eslint-disable-next-line no-restricted-syntax
```
dario-piotrowicz, line 140: no TODO here?

src/cloudflare/internal/ai-api.ts
```diff
@@ -128,13 +128,17 @@ function findReadableStreamKeys(
...
   private logs: Array<string> = [];
+  // eslint-disable-next-line no-restricted-syntax
```
dario-piotrowicz, line 143: no TODO here?

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -97,6 +97,7 @@ export class PipelineTransformImpl<
 
   // called by the dispatcher to validate that run is properly implemented by the subclass
   // @ts-expect-error thinks ping is never used
+  // eslint-disable-next-line no-restricted-syntax
```
dario-piotrowicz, line 100: no TODO here?

src/cloudflare/internal/pipeline-transform.ts
```diff
@@ -114,6 +115,7 @@ export class PipelineTransformImpl<
   // the reason this is typescript private and not javascript private is that this must be
   // able to be called by the dispatcher but should not be called by the class implementer
   // @ts-expect-error _run is called by rpc
+  // eslint-disable-next-line no-restricted-syntax
```
dario-piotrowicz, line 118: no TODO here?

ns476: The comment documents the reason for TypeScript private

src/cloudflare/internal/pipeline-transform.ts
ns476: Similarly here, I think this one needs to be typescript private

PR #4547, opened 2025-07-16
Add support for process cwd, chdir and umask in process v2

guybedford:
Updates the process v2 implementation to support `process.cwd()`, `process.chdir()` and `process.umask`.

`cwd()` is integrated into the virtual filesystem where the default cwd is `/bundle`, and it is stored as a property of the tmpfs as that matches the same FS context (in theory turning virtual tmpdir into a virtual fs state).

`process.umask` is just supported as a non-failing stub supporting umask validations but never actually applying the umask as we don't have a vfs with fine-grained permissions, but it still makes sense not to throw on these operations for FS-based code.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:22
+FROM mcr.microsoft.com/vscode/devcontainers/base:dev-bookworm
 
 # Install dependencies, including clang via through LLVM APT repository. Note that this
 # will also install lldb and clangd alongside dependencies.
@@ -11,6 +11,3 @@ ENV PATH /usr/lib/llvm-${LLVM_VERSION}/bin:$PATH
 
 # Install Bazel (via Bazelisk)
 RUN npm install -g @bazel/bazelisk
-
-# Install Just
-RUN npm install -g rust-just
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/vscode/devcontainers/base:dev-bookworm
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:22
 
 # Install dependencies, including clang via through LLVM APT repository. Note that this
 # will also install lldb and clangd alongside dependencies.
@@ -11,3 +11,6 @@ ENV PATH /usr/lib/llvm-${LLVM_VERSION}/bin:$PATH
 
 # Install Bazel (via Bazelisk)
 RUN npm install -g @bazel/bazelisk
+
+# Install Just
+RUN npm install -g rust-just
--- a/src/node/internal/public_process.ts
+++ b/src/node/internal/public_process.ts
@@ -30,6 +30,7 @@ import {
   features,
   _setEventsProcess,
 } from 'node-internal:internal_process';
+import { validateString } from './validators';
 
 export { platform, nextTick, emitWarning, env, features };
 
@@ -39,15 +40,11 @@ export const stdout = undefined;
 export const stderr = undefined;
 
 export function chdir(path: string | Buffer | URL): void {
-  if (typeof path !== 'string') {
-    throw new ERR_INVALID_ARG_TYPE('directory', 'string', path);
-  }
+  validateString(path, 'directory');
   processImpl.setCwd(path);
 }
 
-export function cwd(): string {
-  return processImpl.getCwd();
-}
+export const cwd = processImpl.getCwd.bind(processImpl);
 
 // We do not support setting the umask as we do not have a fine-grained
 // permissions on the VFS. Instead we only support validation of input
--- a/src/node/internal/public_process.ts
+++ b/src/node/internal/public_process.ts
@@ -11,6 +11,7 @@ import { default as EventEmitter } from 'node-internal:events';
 import {
   ERR_EPERM,
   ERR_INVALID_ARG_TYPE,
+  ERR_INVALID_ARG_VALUE,
   ERR_OUT_OF_RANGE,
   ERR_UNSUPPORTED_OPERATION,
 } from 'node-internal:internal_errors';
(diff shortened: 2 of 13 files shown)


src/node/internal/public_process.ts
```diff
@@ -37,10 +38,45 @@ export const stdin = undefined;
...
 export const stderr = undefined;
 
-// TODO(soon): Implement along with FS work (and as a requirement for removing experimental).
-export const chdir = undefined;
-export const cwd = undefined;
-export const umask = undefined;
+export function chdir(path: string | Buffer | URL): void {
+  if (typeof path !== 'string') {
+    throw new ERR_INVALID_ARG_TYPE('directory', 'string', path);
+  }
```
anonrig: Why not call validateString method?

src/node/internal/public_process.ts
```diff
@@ -37,10 +38,45 @@ export const stdin = undefined;
...
+  processImpl.setCwd(path);
+}
+
+export function cwd(): string {
+  return processImpl.getCwd();
+}
```
anonrig:
```suggestion
export const cwd = processImpl.getCwd.bind(processImpl);
```

src/workerd/io/worker-fs.h
```diff
@@ -694,8 +694,17 @@ class TmpDirStoreScope final {
     return dir.addRef();
   }
 
+  kj::PathPtr getCwd() const {
+    return kj::PathPtr(cwd);
+  }
+
+  void setCwd(kj::Path newCwd) const {
```
anonrig: Why do we have const in here?

```suggestion
  void setCwd(kj::Path newCwd) {
```

src/workerd/io/worker-fs.h
```diff
@@ -694,8 +694,17 @@ class TmpDirStoreScope final {
...
+    cwd = kj::mv(newCwd);
+  }
+
  private:
   mutable kj::Rc<Directory> dir;
+  mutable kj::Path cwd;
```
anonrig:
```suggestion
  kj::Path cwd;
```

src/workerd/io/worker-fs.h
guybedford: Went ahead and removed the const usage.

src/node/internal/public_process.ts
guybedford: Added thanks.

guybedford:
I've updated the approach here per discussion to set the cwd to `/bundle` during static initialization and `/tmp` during request handling.

@jasnell I wasn't sure how to easily test the static initialization / non IO context case here, would value your input on that.

src/workerd/io/worker-fs.c++
```diff
@@ -13,6 +13,36 @@ namespace workerd {
...
+// Returns the Cwd if available, otherwise returns an empty path (root).
+kj::PathPtr getCurrentWorkingDirectory() {
+  if (IoContext::hasCurrent()) {
+    return IoContext::current().getTmpDirStoreScope().getCwd();
+  }
+  if (TmpDirStoreScope::hasCurrent()) {
+    return TmpDirStoreScope::current().getCwd();
+  }
+  // Default to "bundle" if no context is available
+  // this way bundle is used during init
```
jasnell:
```suggestion
  // Default to "tmp" if no context is available
  // this way bundle is used during init
```

src/workerd/io/worker-fs.c++
```diff
@@ -1146,7 +1176,8 @@ TmpDirStoreScope& TmpDirStoreScope::current() {
 }
 
 TmpDirStoreScope::TmpDirStoreScope(kj::Maybe<kj::Badge<TmpDirStoreScope>> guard)
-    : dir(Directory::newWritable()) {
+    : dir(Directory::newWritable()),
+      cwd(kj::Path::parse("tmp")) {  // Initialize to tmp path
```
jasnell: I believe you could probably just do `cwd(kj::Path({"path"}))` I think? Parse is a bit more expensive.

src/workerd/api/node/tests/process-nodejs-test.js
```diff
@@ -786,3 +775,139 @@ export const processRejectionListeners = {
     }
   },
 };
+
+// TODO: figure out how to test static IO context
+// const staticCwd = process.cwd();
```
jasnell: If the need is to test what the cwd is outside of an IoContext, then you can just put the assert here in the top level scope.
```suggestion
// const staticCwd = process.cwd();
assert.strictEqual(process.cwd(), '/bundle');
```

src/workerd/api/node/process.c++
```diff
@@ -137,4 +139,56 @@ void ProcessModule::exitImpl(jsg::Lock& js, int code) {
   js.logWarning(kj::str(err.get(js, "stack"_kj)));
 }
 
+kj::String ProcessModule::getCwd(jsg::Lock& js) {
+  auto cwd = getCurrentWorkingDirectory();
+  if (cwd == nullptr) {
+    return kj::str("/");
+  }
+  return kj::str("/", cwd.toString());
```
jasnell: `cwd` here is a `kj::PathPtr` correct? If so, I believe you can just `return cwd.toString(true)` here

src/workerd/api/node/process.c++
```diff
@@ -137,4 +139,56 @@ void ProcessModule::exitImpl(jsg::Lock& js, int code) {
...
+
+  if (path.size() == 0) {
+    throwUVException(js, UV_ENOENT, "chdir"_kj);
+  }
+
+  auto& vfs = VirtualFileSystem::current(js);
+  auto url = KJ_REQUIRE_NONNULL(jsg::Url::tryParse(path, "file:///"_kj));
+
+  // Resolve the path against current working directory if it's relative
+  kj::Path resolvedPath = [&]() {
```
jasnell, line 162: You're parsing the path above into a url, then again as a path here. The below here should probably work off the url's `pathname` instead for consistency with the rest of the fs apis.

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/vscode/devcontainers/base:dev-bookworm
```
jasnell: Are the Dockerfile changes here intentional? Doesn't appear immediately relevant this change. If this is necessary, can this be moved to a separate commit?

src/workerd/io/worker-fs.c++
guybedford: It's supposed to be bundle for no context? Only tmp once the io context is initialized?

src/workerd/api/node/tests/process-nodejs-test.js
guybedford: Okay maybe I'm not implementing it correctly because the test isn't passing for checking it's bundle.

.devcontainer/Dockerfile
guybedford: Opened https://github.com/cloudflare/workerd/pull/4553.

src/workerd/api/node/process.c++
guybedford: Thanks, that was from a previous iteration where I was doing the URL approach, this line just got left behind.

src/workerd/io/worker-fs.c++
guybedford: Ahh this was a typo in the implementation not the comment - should have been `"bundle"`.

guybedford:
I've pushed up a fix for the cwd handling so that it should be correctly scoped to the isolate or iocontext vfs now. `process.chdir()` is allowed during initialization with an initial directory of `/bundle`, but the `cwd` is then updated at iocontext initialization time to reflect the `/tmp` directory.

jasnell (approved):

PR #4548, opened 2025-07-16
Fetch Python deps asynchronously in the main thread.

dom96:
Moves where Python deps (bundle and built-in packages) are downloaded so that the download can be accomplished asynchronously in the main workerd thread.

### Test Plan

```
$ bazel run @workerd//src/workerd/server/tests/python:durable-object_0.26.0a2@
$ bazel run @workerd//src/workerd/server/tests/python:import/urllib3_0.27.7@ --test_tag_filters=-none
$ bazel run @workerd//src/workerd/server:workerd -- pyodide-lock
```

### DIFF
--- a/src/workerd/api/pyodide/pyodide.c++
+++ b/src/workerd/api/pyodide/pyodide.c++
@@ -734,12 +734,10 @@ kj::Promise<void> loadPyodidePackage(const PythonConfig& pyConfig,
     const PyodidePackageManager& pyodidePackageManager,
     kj::StringPtr packagesVersion,
     kj::StringPtr filename,
-    kj::HttpClient& client,
-    kj::Timer& timer,
-    kj::HttpHeaderTable& table) {
+    kj::Network& network,
+    kj::Timer& timer) {
 
   auto path = kj::str("python-package-bucket/", packagesVersion, "/", filename);
-
   // First check if we already have this package in memory
   if (pyodidePackageManager.getPyodidePackage(path) != kj::none) {
     co_return;
@@ -770,10 +768,17 @@ kj::Promise<void> loadPyodidePackage(const PythonConfig& pyConfig,
   }
 
   // Need to fetch from network
-  kj::String url =
-      kj::str("https://storage.googleapis.com/cloudflare-edgeworker-python-packages/", path);
+  kj::HttpHeaderTable table;
+  kj::TlsContext::Options tlsOptions;
+  tlsOptions.useSystemTrustStore = true;
+  kj::Own<kj::TlsContext> tlsContext = kj::heap<kj::TlsContext>(kj::mv(tlsOptions));
 
-  auto maybeBody = co_await downloadPackageWithRetry(client, timer, table, url, path);
+  auto tlsNetwork = tlsContext->wrapNetwork(network);
+  auto client = kj::newHttpClient(timer, table, network, *tlsNetwork);
+
+  kj::String url = kj::str(PYTHON_PACKAGES_URL, path);
+
+  auto maybeBody = co_await downloadPackageWithRetry(*client, timer, table, url, path);
   KJ_IF_SOME(body, maybeBody) {
     // Successfully downloaded the package
     // Save the compressed data to disk cache (if enabled)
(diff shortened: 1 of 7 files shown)


src/workerd/api/pyodide/pyodide.c++
```diff
@@ -801,41 +807,107 @@ kj::Promise<void> loadPyodidePackage(const PythonConfig& pyConfig,
...
-      promises.add(loadPyodidePackage(
-          pyConfig, pyodidePackageManager, packagesVersion, filename, *client, timer, table));
-    }
+  kj::Vector<kj::Promise<void>> promises;
+  for (const auto& filename: filenames) {
+    promises.add(loadPyodidePackage(
+        pyConfig, pyodidePackageManager, packagesVersion, filename, network, timer));
+  }
+
+  co_await kj::joinPromises(promises.releaseAsArray());
```
danlapid:
```suggestion
  co_await kj::joinPromisesFailFast(promises.releaseAsArray());
```

src/workerd/api/pyodide/pyodide.c++
danlapid, line 1: Somewhat related to this PR though it existed before.
A lot of this code needs to move from here to "src/workerd/server".
For example "src/workerd/server/pyodide.c++/h".

This file is being included from the internal repo and used from there, we definitely do NOT want the internal repo to be creating HttpClients like that with direct kj::Network and a tlsContext and just fetching stuff.

Of course I realize that this code is not used in the internal repo but we try to keep everything that is workerd specific to the server directory.

src/workerd/api/pyodide/pyodide.h
```diff
@@ -500,15 +501,28 @@ kj::Promise<void> loadPyodidePackage(const PythonConfig& pyConfig,
...
+    workerd::PythonSnapshotRelease::Reader pythonSnapshotRelease,
+    kj::Network& network,
+    kj::Timer& timer);
+
+// Preload Pyodide bundle during workerd startup
+kj::Promise<kj::Maybe<jsg::Bundle::Reader>> fetchPyodideBundle(
+    const PythonConfig& pyConfig, kj::String version, kj::Network& network, kj::Timer& timer);
+
+// Helper functions for bundle file operations
+kj::Path getPyodideBundleFileName(kj::StringPtr version);
```
danlapid: nit: any function that isn't really used by including files doesn't need to be in the header.

danlapid (commented):
Looks really good

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -770,10 +768,18 @@ kj::Promise<void> loadPyodidePackage(const PythonConfig& pyConfig,
...
+  kj::HttpHeaderTable table;
+  kj::TlsContext::Options tlsOptions;
+  tlsOptions.useSystemTrustStore = true;
+  kj::Own<kj::TlsContext> tlsContext = kj::heap<kj::TlsContext>(kj::mv(tlsOptions));
+
+  auto tlsNetwork = tlsContext->wrapNetwork(network);
+  auto client = kj::newHttpClient(timer, table, network, *tlsNetwork);
+
   kj::String url =
       kj::str("https://storage.googleapis.com/cloudflare-edgeworker-python-packages/", path);
```
jasnell: Nit: I know this isn't new here but would really prefer this to be moved to a constant... and ideally made configurable. Hard coding URLs into the source really isn't my favorite approach.

src/workerd/server/server.c++
```diff
@@ -4923,13 +4917,59 @@ uint startInspector(
       [](const uint& port) { return port; });
 }
 
-void Server::startServices(jsg::V8System& v8System,
+kj::Promise<void> Server::preloadPython(config::Config::Reader config) {
+  for (auto serviceConf: config.getServices()) {
+    if (serviceConf.isWorker()) {
+      ConfigErrorReporter errorReporter(*this, serviceConf.getName());
```
kentonv: Will this cause errors to be double-reported, as we'll attempt to compile the compat flags again later on? Perhaps we should use an ErrorReporter here that just discards all errors instead?

src/workerd/server/server.c++
```diff
@@ -4923,13 +4917,59 @@ uint startInspector(
...
+kj::Promise<void> Server::startServices(jsg::V8System& v8System,
     config::Config::Reader config,
     kj::HttpHeaderTable::Builder& headerTableBuilder,
     kj::ForkedPromise<void>& forkedDrainWhen) {
   // ---------------------------------------------------------------------------
   // Configure services
   TRACE_EVENT("workerd", "startServices");
+
+  // Pre-load Python bundles for all Python workers before creating services
+  co_await preloadPython(config);
```
kentonv: Doing all the Python preloading in an earlier phase like this seems to make it harder to support dynamic loading of Python in the future. Dynamic worker loading doesn't support Python yet, but we likely want it to at some point. But, dynamically-loaded workers obviously aren't in the config, so won't be available during this preload step.

I think what we'd ideally do instead is have `Server::makeWorkerImpl()` (and its callers) become async, so that it can load Python stuff if necessary. This method is also used for dynamically-loaded workers. This probably also lets you avoid double-compiling the compat flags, as you could load the python artifacts _after_ compat flags are compiled normally.

You've already made `startServices()` async, so you would just also have to make `makeService()`, `makeWorker()`, and `makeWorkerImpl()` be async, I think.

src/workerd/api/pyodide/pyodide.c++
dom96: Okay, to avoid a confusing diff I'll do this in a separate PR.

dom96:
Ready for another review

src/workerd/server/server.c++
```diff
@@ -5012,7 +5034,7 @@ void Server::startServices(jsg::V8System& v8System,
   // Second pass: Build services.
   for (auto serviceConf: config.getServices()) {
     kj::StringPtr name = serviceConf.getName();
-    auto service = makeService(serviceConf, headerTableBuilder, config.getExtensions());
+    auto service = co_await makeService(serviceConf, headerTableBuilder, config.getExtensions());
```
kentonv, line 5037: What if there are multiple python workers in the config. Should we be fetching the artifacts concurrently for faster startup? If they rely on the same artifacts, will we end up fetching them multiple times or is there a cache somewhere?

dom96: There is a cache (PyodideBundleManager and PyodidePackageManager), so the same artifacts won't be fetched multiple times.

I looked into implementing concurrent fetches here, it looks like we have this implemented in EW already (using a CrossThreadWaitList). I think it would make sense to move that list to workerd and reuse it here, but I'd like to do so in a separate PR as it will require quite a bit more refactoring (I think merging the two manager classes may make sense too and I'd like to do it at the same time).

kentonv: Fair enough.

kentonv:
I'm happy with the changes in server.c++ but don't know enough about the Python fetching code to adequately review that.

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -801,41 +806,107 @@ kj::Promise<void> loadPyodidePackage(const PythonConfig& pyConfig,
...
+  auto tlsNetwork = tls->wrapNetwork(network);
+  auto client = kj::newHttpClient(timer, table, network, *tlsNetwork);
+
+  kj::HttpHeaders headers(table);
+
+  auto req = client->request(kj::HttpMethod::GET, url.asPtr(), headers);
+
+  auto res = co_await req.response;
+  KJ_ASSERT(res.statusCode == 200,
+      kj::str("Request for Pyodide bundle at ", url, " failed with HTTP status ", res.statusCode));
```
jasnell, line 902: Super nit: since this is an internal only error message, it seems unfortunate to allocate the string (kj::str(...)) entirely. This could just be...

```suggestion
  KJ_ASSERT(res.statusCode == 200,
            "Request for Pyodide bundle failed",
            url, res.statusCode);
```

And we'd avoid the additional string allocation.

Obviously extremely minor but still ;-)

src/workerd/api/pyodide/pyodide.c++
```diff
@@ -801,41 +806,107 @@ kj::Promise<void> loadPyodidePackage(const PythonConfig& pyConfig,
...
-    auto tlsNetwork = tlsContext->wrapNetwork(network);
-    auto& timer = ioContext.provider->getTimer();
-
-    auto client = kj::newHttpClient(timer, table, network, *tlsNetwork);
-    kj::Vector<kj::Promise<void>> promises;
-    for (const auto& filename: filenames) {
-      promises.add(loadPyodidePackage(
-          pyConfig, pyodidePackageManager, packagesVersion, filename, *client, timer, table));
-    }
+  kj::Vector<kj::Promise<void>> promises;
```
jasnell, line 826: Nit... we know how many `filenames` there are, we could reserve the space in advance here on the vector.

src/workerd/api/pyodide/pyodide.c++
dom96: I'll take care of these in a follow up PR.

justin-mp: If we're playing assert golf, I think:

```
KJ_ASSERT(res.statusCode == 200, "Request for Pyodide bundle failed", url)
```
will print something like:
```
assertion failed: res.statusCode == 200 [404 == 200], Request for Pyodide bundle failed, url = http://...
```
since the comparison operators inside asserts print both sides if they can.

src/workerd/api/pyodide/pyodide.c++
dom96: https://github.com/cloudflare/workerd/pull/4596

src/workerd/api/pyodide/pyodide.c++
kentonv: In fact, the difference is more than just superficial. As written currently, Sentry will produce a different error fingerprint for every different value of `url` and `res.statusCode`. Whereas with @jasnell's or @justin-mp's version, it'll all be fingerprinted into one issue.

In general, you should almost never use `kj::str` to manually stringify a message for a KJ_LOG/KJ_ASSERT.

PR #4549, opened 2025-07-16
introduce a way to hook into fetch events

anonrig:
This PR introduces an experimental way of hooking into the fetch handler and listening and handling events outside of the fetch handler (to be used only for nodejs)

This PR is not ready to land. It just demonstrates the possibilities.

Here is an example where a client returns `Hello world`

```js
import { registerFetchEvents } from 'cloudflare:nodejs_compat';

const listener = (event) => {
  const response = new Response('Hello, World!');
  const responseEvent = new CustomEvent(`nodejs.fetch-${event.clientId}`, {
    detail: response,
  });
  dispatchEvent(responseEvent);
};
addEventListener('nodejs.fetch', listener);

export default registerFetchEvents();
```


kentonv:
Is `addEventListener('nodejs.fetch', listener);` an actual node API, or is this something invented for this PR?

If the latter, what is the motivation for introducing this API?

You mention supporting createServer(), but for that I'd expect something like:

```js
import * as http from 'node:http';

let server = http.createServer((req, res) => {
  res.writeHead(200, { 'Content-Type': 'application/json' });
  res.end(JSON.stringify({
    data: 'Hello World!',
  }));
});

export default server.asWorkerEntrypoint();
```

Or maybe we can even say that calling `server.listen(80)` during startup works as an alternative to exporting a default entrypoint. (I'm not sure I like this, but it's worth considering.)

Either way, I don't see why we need to involve `addEventListener` here.

guybedford:
I was also thinking of something more like:

app.js
```js
import http from 'node:http';

http.createServer((req, res) => {
  // ...
}).listen(8080);
```

cloudflare-wrapper.js
```js
import { createNodeJSServer } from 'cloudflare:nodejs-compat';
import './app.js';
export default createNodeJSServer({
  portFilter: '*' // or, say, 8080
});
```

anonrig:
addEventListener is not the final implementation but merely a way of achieving PoC without writing too much code. This won't work in production since anybody can sniff into the requests through the public api. 

My goal was to not mutate/introduce new methods to Node.js api. 

Tomorrow I'll update the pull request in a way that registerHttpMethods takes a port number parameter, and triggers a server callback associated with the port number. Later, nodejs server class will convert Request to Node.js server request instance and exposes it for others to consume through public nodejs api. 

Your example will not work because there are libraries such as express and fastify that abstracts the createServer flow in a way that it's not possible to retrieve the server instance and just return our custom function in the default export.

jasnell:
@anonrig and I discussed an alternative approach in chat as I don't think this proposed approach is ideal or workable.

The alternative suggestion leverages the fact that all the frameworks (whether it is node:http, express, fastify, etc) all ultimately end up calling down to `server.listen(...)` where `server` is an instance of `node:http` `Server`. The call to `server.listen(...)` always either accepts a port or assigns a port (if `port` is passed as `0`). All of the frameworks also rely on emitting of the `'request'` event on the `Server` instance when a new request is received. Given that, a more workable approach here is to have an *internal* "Port Mapper" type mechanism that is essentially a map of "port" (not a real port but an analogue) and a `Server` instance... Our implementation of `server.listen(...)` would associate the `Server` instance with the port in that map.... so something in pseudo-code like:

```js
// in our node:http impl

import { portMapper } from 'cloudflare-internal:http-support';

class Server {
  // ...
  listen(port, ...) {
    // ...
    portMapper.add(port, this);
  }
}
```

Then, in the worker entrypoint, we would have something like...

```js
import { createServer } from 'node:http';
import { portMapper } from 'cloudflare:workers';

const server = createServer((req, res) => {
  // ...
});
server.listen(123);  // causes this server to be associated with the port

export default {
  async fetch(req, env, ctx) {
    // Where the return value of get(...) here is a Fetcher that dispatches to the
    // node:http Server registered by the earlier listen call. The portMapper here
    // could also be exposed via env as a binding... that doesn't matter so much.
    return await portMapper.get(123).fetch(req, env, ctx);
  }
}
```

The fetcher in this model would adapt the standard Request, Response objects to Node.js and would cause the `'request'` event to be emitted on the `node:http` `Server` object in exactly the same way happens in Node.js. Done this way, frameworks like express and fastify really won't have to do anything special to be supported and we won't have to rely on any event handling hacks or special event names.

The globalThis.addEventListener(...) approach has the further issue that any request can add handlers to the globalThis so it would actually make it more likely that requests could spy on each other, which we absolutely don't want. The fetcher based approach is safer, simpler, requires the fewest hacks, and can be entirely implemented in JS if we want.

Info references:
* Request event on the Node.js http.Server https://nodejs.org/docs/latest/api/http.html#event-request
* Express https://github.com/expressjs/express/blob/master/lib/application.js#L599-L605
* Fastify https://github.com/fastify/fastify/blob/main/lib/server.js#L296

kentonv:
@guybedford's and @jasnell's ideas both make sense to me.

anonrig:
Closing in favor of https://github.com/cloudflare/workerd/pull/4591

PR #4557, opened 2025-07-16
GitHub workflows to check actor over author

guybedford:
Posting as a draft and for discussion, if it would be useful for third-party contributions to be able to have the internal build triggered by having the workflow run by a team member.

Perhaps a stronger security guarantee is needed for this check though?

### DIFF
--- a/.github/workflows/internal-build.yml
+++ b/.github/workflows/internal-build.yml
@@ -21,7 +21,7 @@ jobs:
         if: github.event.pull_request.head.repo.fork
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


guybedford:
//cc @ketanhwr

.github/workflows/internal-build.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         if: github.event.pull_request.head.repo.fork
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          AUTHOR: ${{ github.event.pull_request.user.login }}
+          AUTHOR: ${{ github.actor }}
```
ketanhwr, line 24: I think `github.triggering_actor` might be better suited here? So that it checks the permission each time..

Issue #4558, opened 2025-07-17
Event.currentTarget is set when it's expected to be null


npaun:
In the test, `EventTarget-constructible.any.js`, we fail the subtest `A constructed EventTarget implements dispatch correctly`, because `event.currentTarget` is set even after `target.dispatchEvent(event)`. It is expected to be `null`.

I am not that experienced with the DOM, but I noticed this in the documentation of `currentTarget` on MDN:

> Note that the value of currentTarget is only available in a handler for the event. Outside an event handler it will be null. This means that, for example, if you take a reference to the Event object inside an event handler and then access its currentTarget property outside the event handler, its value will be null.
- https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

This makes me think that we need to do something to make sure you can only use that property inside an event handler?

jasnell:
Confirmed that we're inconsistent with the spec here.

Node.js -> Clears currentTarget
Deno -> Clears currentTarget
Firefox -> Clears currentTarget
Chrome -> Does not clear currentTarget

jasnell:
I think this is one we ought to safely be able to do without a compat flag.

Issue #4556, opened 2025-07-16
Event.isTrusted needs to be an own property not a prototype property


npaun:
WPT test is `Event-isTrusted.any.js`

PR #4559, opened 2025-07-17
Correctly handle the event.currentTarget property

jasnell:
Per the spec, the currentTarget property of an Event should be null if the event is not being dispatched.

Fixes: https://github.com/cloudflare/workerd/issues/4558
Fixes: https://github.com/cloudflare/workerd/issues/4556

~~@danlapid @kentonv ... I'm hoping this can be something we don't need a compat flag for. Whatcha think?~~

Adds a new `pedantic_wpt` compat flag for these kinds of fixes. Additional changes will be made incrementally using `implied_by` compat flags so that'll bundle all these types of changes together.

/cc @npaun

### DIFF
--- a/src/workerd/api/basics.h
+++ b/src/workerd/api/basics.h
@@ -129,8 +129,8 @@ class Event: public jsg::Object {
 
   // The currentTarget is the EventTarget on which the Event is being
   // dispatched. This will be set every time dispatchEvent() is called
-  // successfully and will remain set after dispatching is completed.
-  jsg::Optional<jsg::Ref<EventTarget>> getCurrentTarget();
+  // successfully and will be null after dispatchEvent returns.
+  kj::Maybe<jsg::Ref<EventTarget>> getCurrentTarget();
 
   // Because we don't support hierarchical EventTargets, this function
   // will always return the same value as getCurrentTarget().
@@ -155,11 +155,23 @@ class Event: public jsg::Object {
       JSG_READONLY_PROTOTYPE_PROPERTY(cancelable, getCancelable);
       JSG_READONLY_PROTOTYPE_PROPERTY(defaultPrevented, getDefaultPrevented);
       JSG_READONLY_PROTOTYPE_PROPERTY(returnValue, getReturnValue);
-      JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getCurrentTarget);
+      if (flags.getPedanticWpt()) {
+        JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getCurrentTarget);
+      } else {
+        // The original implementation had getTarget simply deferring to
+        // getCurrentTarget, the new impl moves the original impl into
+        // getTarget here so having currentTarget point to getTarget
+        // preserves the original behavior.
+        JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getTarget);
+      }
       JSG_READONLY_PROTOTYPE_PROPERTY(target, getTarget);
-      JSG_READONLY_PROTOTYPE_PROPERTY(srcElement, getCurrentTarget);
+      JSG_READONLY_PROTOTYPE_PROPERTY(srcElement, getTarget);
       JSG_READONLY_PROTOTYPE_PROPERTY(timeStamp, getTimestamp);
-      JSG_READONLY_PROTOTYPE_PROPERTY(isTrusted, getIsTrusted);
+      if (flags.getPedanticWpt()) {
+        JSG_READONLY_INSTANCE_PROPERTY(isTrusted, getIsTrusted);
+      } else {
+        JSG_READONLY_PROTOTYPE_PROPERTY(isTrusted, getIsTrusted);
+      }
 
       JSG_PROTOTYPE_PROPERTY(cancelBubble, getCancelBubble, setCancelBubble);
     } else {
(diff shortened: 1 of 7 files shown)


npaun (approved):
Implementation is very clear and simple.

jasnell:
Crap, I forgot that this would change the generated types. I think based on that alone we'll need a compat flag here... ugh.

kentonv:
I don't think changing the generated types in itself is a reason for a compat flag -- the types don't exist in production, after all, and compat flags are about protecting workers in production.

But whether the API change itself requires a compat flag, I don't know.

We could add logging to see if anyone actually calls this when `!isBeingDispatched`. If anyone does, then I think a compat flag is probably needed.

Though that doesn't really answer whether the undefined->null change in itself might require a compat flag, which I guess is harder to tell.

kentonv:
I wonder if we should create a new compat flag called `wpt_pedantic` or something, and accumulate all the little irrelevant changes needed to fix WPT tests into it. Once we have all the fixes done, only then do we set a default-on date.

That way we don't create a massive number of compat flags for little things nobody actually cares about...

cc @npaun

jasnell:
That would make sense I think

npaun:
@kentonv Yeah I've considered creating one a few times now

jasnell:
I'll move this back to draft, add a compat flag, and will bundle multiple WPT fixes under this PR. We can likely do this in batches with several compat flags, one for each batch.

jasnell:
PR updated to add the compat flag and also handle the `isTrusted` property pendantically. PTAL

src/workerd/api/basics.h
```diff
@@ -155,11 +155,19 @@ class Event: public jsg::Object {
       JSG_READONLY_PROTOTYPE_PROPERTY(cancelable, getCancelable);
       JSG_READONLY_PROTOTYPE_PROPERTY(defaultPrevented, getDefaultPrevented);
       JSG_READONLY_PROTOTYPE_PROPERTY(returnValue, getReturnValue);
-      JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getCurrentTarget);
+      if (flags.getPedanticWpt()) {
+        JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getCurrentTarget);
+      } else {
+        JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getTarget);
```
jasnell, line 165: Review note: the original implementation had `getTarget` simply deferring to `getCurrentTarget`, the new impl moves the original impl into `getTarget` here so having `currentTarget` point to `getTarget` preserves the original behavior.

src/workerd/io/compatibility-date.capnp
```diff
@@ -876,4 +876,11 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("disable_nodejs_http_modules")
       $impliedByAfterDate(name = "nodeJsCompat", date = "2025-08-15");
   # Enables Node.js http related modules such as node:http and node:https
+
+  pedanticWpt @101 :Bool
+      $compatEnableFlag("pedantic_wpt")
+      $compatDisableFlag("non_pedantic_wpt");
```
jasnell, line 882: Review note: this intentionally has no disable enable date for now.

src/workerd/api/basics.h
npaun: I'd paste in this comment in the code. Would definitely save a headscratcher  later on.

src/wpt/BUILD.bazel
```diff
@@ -66,6 +66,7 @@ wpt_test(
 
 wpt_test(
     name = "dom/events",
+    compat_flags = ["pedantic_wpt"],
```
npaun, line 69: Should we maybe put this on for every wpt test? That'd be in `build/wpt_test.bzl`.

src/workerd/api/basics.h
```diff
@@ -155,11 +155,19 @@ class Event: public jsg::Object {
       JSG_READONLY_PROTOTYPE_PROPERTY(cancelable, getCancelable);
       JSG_READONLY_PROTOTYPE_PROPERTY(defaultPrevented, getDefaultPrevented);
       JSG_READONLY_PROTOTYPE_PROPERTY(returnValue, getReturnValue);
-      JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getCurrentTarget);
+      if (flags.getPedanticWpt()) {
+        JSG_READONLY_PROTOTYPE_PROPERTY(currentTarget, getCurrentTarget);
+      } else {
```
jasnell, line 160:
```suggestion
      } else {
        // The original implementation had getTarget simply deferring to
        // getCurrentTarget, the new impl moves the original impl into
        // getTarget here so having currentTarget point to getTarget
        // preserves the original behavior.
```

src/wpt/dom/events-test.ts
```diff
@@ -45,9 +45,7 @@ export default {
   'EventTarget-addEventListener.any.js': {},
   'EventTarget-constructible.any.js': {
     comment: 'Should be null, not EventTarget',
-    expectedFailures: [
```
npaun, line 48: Shouldn't there be a change relating to `isTrusted` here as well?

jasnell: I thought so too but it passed locally without it.. so... I was confused lol... wanted to see what the main CI did.

src/wpt/BUILD.bazel
jasnell: I think eventually yes, but forcing it to be explicit on each one will likely help us find issues. When we've caught up, then probably putting it on all of them makes sense.

npaun (approved):

npaun (approved):

PR #4564, opened 2025-07-17
[o11y] Return STW outcome properly

fhanau:
Provide the right outcome for the Streaming Tail Worker event, instead of always returning OK. This includes support for propagating the result via RPC.
Overlaps with #4540, should try to stitch these together cleanly.

### DIFF
--- a/src/workerd/io/trace-stream.c++
+++ b/src/workerd/io/trace-stream.c++
@@ -982,7 +982,11 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::run
   });
 
   co_await forked.addBranch().exclusiveJoin(ioContext.onAbort());
-  co_return WorkerInterface::CustomEvent::Result{.outcome = EventOutcome::OK};
+
+  // Return the outcome. Synchronous draining is not required.
+  co_return WorkerInterface::CustomEvent::Result{
+    .outcome = ioContext.waitUntilStatus(),
+  };
 }
 
 kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sendRpc(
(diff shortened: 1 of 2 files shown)


src/workerd/io/trace-stream.c++
```diff
@@ -1020,16 +1023,17 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
   this->capFulfiller->fulfill(kj::mv(cap));
 
   try {
-    co_await sent.ignoreResult().exclusiveJoin(kj::mv(completionPaf.promise));
+    EventOutcome outcome = co_await sent.then([](auto resp) {
+      return resp.getResult();
+    }).exclusiveJoin(kj::mv(completionPaf.promise).then([]() { return EventOutcome::OK; }));
```
danlapid: Can we get a comment here on why completionPaf.promise should be OK?
Intuitively it feels more like "cancelled" to me but I can be convinced otherwise by a good comment 😄

fhanau: Addressed in the latest push, which is (finally) also informed by a proper understanding of why the stream cancellation warning above happens. With this approach the completion promise should only fulfill first if there's an actual cancellation, confirmed this locally with the two downstream PRs adding tests for stream cancellation and for the outcome event being reported properly.

jasnell (approved):

src/workerd/io/trace-stream.c++
```diff
@@ -1010,17 +1014,28 @@ kj::Promise<WorkerInterface::CustomEvent::Result> TailStreamCustomEventImpl::sen
...
   try {
-    co_await sent.ignoreResult().exclusiveJoin(kj::mv(completionPaf.promise));
+    EventOutcome outcome = co_await sent.then([](auto resp) {
+      return resp.getResult();
+    }).exclusiveJoin(forked.addBranch().then([]() { return EventOutcome::CANCELED; }));
+
+    // If the sent promise returned first, we still need to wait for the parent process to drop the
+    // capability (which should happen right after it receives the stop signal) so that no
+    // capabilities remain in an incomplete state when we return.
+    co_await forked.addBranch();
```
mar-cf, line 1030: It's unfortunate this this has been merged because I believe, with my limited understanding, that this is a workaround that suppresses a warning that prevents sending over the tail stream session messages that will never be processed. Most of that was said in offline discussions - I'll repeat it here.

The completion membrane and scope guarded warning ensure that messages sent over the tail stream event will be sent to an a live remote context that will process them. If the custom event returns and we have not fulfilled the completion membrane, meaning the client is still active, we could report a tail stream events that will never be processed. A purpose of the completion membrane is that when the remote context completes, and the event outcome is returned, we have already shut down the medium to send tail stream events. Awaiting the completion here doesn't serve a purpose other than to suppress the scope guarded warning above. There may be another pipeline capability related reason to await closure, but if the purpose of the completion membrane is guarding against sending messages over a client that closed on the remote side, this circumvents that protection.

@kentonv

fhanau: As discussed with Mar, I think that the approach taken here is correct based on my understanding. This change was tested rigorously (see the two PRs adding downstream tests) and is described in detail in the comments.
Based on https://github.com/cloudflare/workerd/pull/4030#pullrequestreview-2816509032, we must wait for all capability to be dropped before returning from this function so that all capabilities are cleaned up as enforced by the membranes. If the sent promise in the exclusiveJoin fulfills first, we have received all tail stream events, did all required work in the tail Stream handler and have indicated to the client that the TailStreamTarget capability can be dropped, but the client may not have dropped the capability yet. This causes the cancellation warning to appear based on the membrane not being fulfilled in time – the message was and is misleading as all tail session event processing is done, we are not at risk of losing any events.

PR #4570, opened 2025-07-18
[nfc] Add no-op method to LimitEnforcer

fhanau:
This allows us to perform some build cleanup in the downstream build (see downstream PR).

### DIFF
--- a/src/workerd/io/limit-enforcer.h
+++ b/src/workerd/io/limit-enforcer.h
@@ -153,6 +153,9 @@ class LimitEnforcer {
 
   // Report resource usage metrics to the given request metrics object.
   virtual void reportMetrics(RequestObserver& requestMetrics) = 0;
+
+  // Only used downstream for internal metrics.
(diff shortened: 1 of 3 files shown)


src/workerd/io/limit-enforcer.h
```diff
@@ -153,6 +153,9 @@ class LimitEnforcer {
 
   // Report resource usage metrics to the given request metrics object.
   virtual void reportMetrics(RequestObserver& requestMetrics) = 0;
+
+  // Only used downstream for internal metrics.
+  virtual kj::Duration consumeTimeElapsedForPeriodicLogging() = 0;
```
jasnell, line 158: If we implement this method in the internal impl but not here, we can simplify this by just having this virtual method `return 0 * kj::SECONDS;` rather than making it purely virtual... unless you specifically *want* it to be purely virtual.

fhanau: In this case, the whole class is pure virtual so far. Not sure if it matters but making this pure virtual allows us to keep that property for now.

PR #4572, opened 2025-07-18
add waitUntil method to cloudflare:workers module

anonrig:
Starting from today, the following code works:

```js
import { waitUntil } from 'cloudflare:workers';

let globalWaitUntilPromise = null;

export default {
  async fetch() {
    globalWaitUntilPromise = null;
    let resolve;
    globalWaitUntilPromise = new Promise((r) => {
      resolve = r;
    });
    
    waitUntil(
      (async () => {
        await scheduler.wait(100);
        resolve();
      })()
    );

    return new Response('hello world');
  }
}
```

### DIFF
--- a/src/workerd/api/tests/js-rpc-test.js
+++ b/src/workerd/api/tests/js-rpc-test.js
@@ -1,4 +1,5 @@
 import assert from 'node:assert';
+import { waitUntil } from 'cloudflare:workers';
 import {
   WorkerEntrypoint,
   DurableObject,
@@ -449,6 +450,23 @@ export class MyService extends WorkerEntrypoint {
     });
 
     this.ctx.waitUntil(
+      (async () => {
+        await scheduler.wait(100);
+        resolve();
+      })()
+    );
+  }
+
+  testImportedWaitUntil() {
+    // Initialize globalWaitUntilPromise to a promise that will be resolved in a waitUntil task
+    // later on. We'll perform a cross-context wait to verify that the waitUntil task actually
+    // completes and resolves the promise.
+    let resolve;
+    globalWaitUntilPromise = new Promise((r) => {
+      resolve = r;
+    });
+
+    waitUntil(
       (async () => {
         await scheduler.wait(100);
         resolve();
@@ -1179,11 +1197,40 @@ export let crossContextSharingDoesntWork = {
 
 export let waitUntilWorks = {
   async test(controller, env, ctx) {
-    globalWaitUntilPromise = null;
-    await env.MyService.testWaitUntil();
+    // Tests ctx.waitUntil
+    {
+      globalWaitUntilPromise = null;
+      await env.MyService.testWaitUntil();
 
-    assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
-    await globalWaitUntilPromise;
+      assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
+      await globalWaitUntilPromise;
+    }
+
+    // Tests `import { waitUntil } from 'cloudflare:workers` on DO
+    {
+      globalWaitUntilPromise = null;
+      await env.MyService.testImportedWaitUntil();
+
+      assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
+      await globalWaitUntilPromise;
+    }
+
+    // Tests `import { waitUntil } from 'cloudflare:workers` directly
+    {
+      globalWaitUntilPromise = null;
+
+      let resolve;
+      globalWaitUntilPromise = new Promise((r) => {
+        resolve = r;
+      });
+
+      waitUntil(
+        (async () => {
+          await scheduler.wait(100);
+          resolve();
+        })()
+      );
+    }
   },
 };
 
--- a/src/workerd/api/worker-rpc.h
+++ b/src/workerd/api/worker-rpc.h
@@ -554,6 +554,11 @@ class EntrypointsModule: public jsg::Object {
   EntrypointsModule() = default;
   EntrypointsModule(jsg::Lock&, const jsg::Url&) {}
 
+  void waitUntil(kj::Promise<void> promise) {
+    JSG_REQUIRE(IoContext::hasCurrent(), Error, "waitUntil requires an active request");
+    IoContext::current().addWaitUntil(kj::mv(promise));
+  }
+
   JSG_RESOURCE_TYPE(EntrypointsModule) {
     JSG_NESTED_TYPE(WorkerEntrypoint);
     JSG_NESTED_TYPE(WorkflowEntrypoint);
@@ -562,6 +567,8 @@ class EntrypointsModule: public jsg::Object {
     JSG_NESTED_TYPE_NAMED(JsRpcProperty, RpcProperty);
     JSG_NESTED_TYPE_NAMED(JsRpcStub, RpcStub);
     JSG_NESTED_TYPE_NAMED(JsRpcTarget, RpcTarget);
+
+    JSG_METHOD(waitUntil);
   }
 };
 
--- a/src/cloudflare/internal/workers.d.ts
+++ b/src/cloudflare/internal/workers.d.ts
@@ -24,3 +24,5 @@ export class RpcStub {
 }
 
 export class RpcTarget {}
+
+export function waitUntil(promise: Promise<unknown>): void;
--- a/src/cloudflare/workers.ts
+++ b/src/cloudflare/workers.ts
@@ -89,3 +89,5 @@ export const env = new Proxy(
     },
   }
 );
+
+export const waitUntil = entrypoints.waitUntil.bind(entrypoints);


src/workerd/api/worker-rpc.h
```diff
@@ -554,6 +554,11 @@ class EntrypointsModule: public jsg::Object {
   EntrypointsModule() = default;
   EntrypointsModule(jsg::Lock&, const jsg::Url&) {}
 
+  void waitUntil(kj::Promise<void> promise) {
```
kentonv: Nitpick: This method body, while it may look small, is likely far too complex for the compiler to actually inline, so it should be placed in the .c++ file.

src/workerd/api/worker-rpc.h
```diff
@@ -554,6 +554,11 @@ class EntrypointsModule: public jsg::Object {
...
+    JSG_REQUIRE(IoContext::hasCurrent(), Error, "waitUntil requires an active request");
```
guybedford: It might be worthwhile having a test for this error as well.

guybedford (approved):

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -455,6 +456,23 @@ export class MyService extends WorkerEntrypoint {
...
+    globalWaitUntilPromise = new Promise((r) => {
+      resolve = r;
+    });
+
+    waitUntil(
+      (async () => {
+        await scheduler.wait(100);
+        resolve();
+      })()
+    );
```
jasnell: nit: I believe you could simplify this down to...
```suggestion
    waitUntil(scheduler.wait(100).then(resolve));
```

src/workerd/api/worker-rpc.h
kentonv: You don't really need this line, `current()` will already throw a JSG exception if there's no active request.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1179,11 +1197,40 @@ export let crossContextSharingDoesntWork = {
...
 export let waitUntilWorks = {
   async test(controller, env, ctx) {
-    globalWaitUntilPromise = null;
-    await env.MyService.testWaitUntil();
+    // Tests ctx.waitUntil
+    {
+      globalWaitUntilPromise = null;
+      await env.MyService.testWaitUntil();
+
+      assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
```
jasnell: nit...
```suggestion
      assert.ok(globalWaitUntilPromise instanceof Promise);
```

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1179,11 +1197,40 @@ export let crossContextSharingDoesntWork = {
...
+      await globalWaitUntilPromise;
+    }
+
+    // Tests `import { waitUntil } from 'cloudflare:workers` on DO
+    {
+      globalWaitUntilPromise = null;
+      await env.MyService.testImportedWaitUntil();
+
+      assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
```
jasnell:
```suggestion
      assert.ok(globalWaitUntilPromise instanceof Promise);
```

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -455,6 +456,23 @@ export class MyService extends WorkerEntrypoint {
...
+  testImportedWaitUntil() {
+    // Initialize globalWaitUntilPromise to a promise that will be resolved in a waitUntil task
+    // later on. We'll perform a cross-context wait to verify that the waitUntil task actually
+    // completes and resolves the promise.
+    let resolve;
+    globalWaitUntilPromise = new Promise((r) => {
+      resolve = r;
+    });
+
+    waitUntil(
```
kentonv: ~This isn't really testing anything: If you deleted the `waitUntil()` call (but kept the parameter computation) the test would still pass, because Durable Objects keep running for 1 minute after the last request.~

~The only thing that seems worth testing here is that `waitUntil()` doesn't, like, throw an exception.~

kentonv: Wait I was confused, I thought this was a Durable Object.

src/workerd/api/worker-rpc.h
anonrig: Thanks. Yes, I'll update.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1179,11 +1197,40 @@ export let crossContextSharingDoesntWork = {
...
+      globalWaitUntilPromise = new Promise((r) => {
+        resolve = r;
+      });
+
+      waitUntil(
+        (async () => {
+          await scheduler.wait(100);
+          resolve();
+        })()
+      );
```
jasnell: bit more compact and readable ...

```suggestion
      const { promise, resolve } = Promise.withResolvers();
      waitUntil(scheduler.wait(100).then(resolve));
      await promise;
```

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1179,11 +1197,40 @@ export let crossContextSharingDoesntWork = {
...
+    // Tests ctx.waitUntil
+    {
+      globalWaitUntilPromise = null;
+      await env.MyService.testWaitUntil();
+
+      assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
+      await globalWaitUntilPromise;
+    }
+
+    // Tests `import { waitUntil } from 'cloudflare:workers` on DO
```
kentonv: I'm confused, the code below does not test on a DO AFAICT?

But I don't think such a test is worth it anyway.

(This test as written tests waitUntil() on a stateless worker, which makes sense.)

src/workerd/api/worker-rpc.h
jasnell: Nit: the error message here could be made a bit more useful. Not critical now tho.

src/workerd/api/tests/js-rpc-test.js
anonrig: Yes, my comment was wrong. Updating it.

src/workerd/api/worker-rpc.h
anonrig: Removed this per Kenton's recommendation.

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1,4 +1,5 @@
 import assert from 'node:assert';
+import { waitUntil } from 'cloudflare:workers';
```
jasnell, line 2: The test should likely also test the error condition of calling `waitUntil` at the global scope / outside of an IoContext.

src/workerd/api/worker-rpc.h
anonrig: Done.

src/workerd/api/tests/js-rpc-test.js
anonrig: Added a test in the last commit

jasnell: ha! I refreshed then saw that you had that :-) .. nevermind!

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1179,11 +1197,40 @@ export let crossContextSharingDoesntWork = {
...
+      globalWaitUntilPromise = null;
+      await env.MyService.testImportedWaitUntil();
+
+      assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
+      await globalWaitUntilPromise;
+    }
 
-    assert.strictEqual(globalWaitUntilPromise instanceof Promise, true);
-    await globalWaitUntilPromise;
+    // Tests `import { waitUntil } from 'cloudflare:workers` directly
```
kentonv: I'm not sure what's the value of this test compared to the previous block. It doesn't look like this block actually testes anything as nobody actually waits on the `globalWaitUntilPromise` to make sure it resolves. But the previous block already covers that anyway, so I think this block can simply be deleted?

src/workerd/api/tests/js-rpc-test.js
anonrig: My comment was wrong, which led to this confusion. I've updated the test. This was actually testing WorkerEntrypoint.

src/workerd/api/tests/js-rpc-test.js
anonrig: Agreed. removed it with the last commit.

jasnell (approved):

src/workerd/api/tests/js-rpc-test.js
```diff
@@ -1179,11 +1197,30 @@ export let crossContextSharingDoesntWork = {
...
-    await globalWaitUntilPromise;
+      assert.ok(globalWaitUntilPromise instanceof Promise);
+      await globalWaitUntilPromise;
+    }
+
+    // Tests `import { waitUntil } from 'cloudflare:workers` directly
+    {
+      const { promise, resolve } = Promise.withResolvers();
+      waitUntil(scheduler.wait(100).then(resolve));
+      await promise;
```
kentonv: This doesn't test anything interesting. This promise would resolve even without the `waitUntil()` call above. `waitUntil()` only makes a difference if the promise is still running when the event ends, but that's not the case here.

kentonv (approved):

src/workerd/api/tests/js-rpc-test.js
kentonv: (Oh it was already removed.)

anonrig:
@cloudflare/wrangler @dario-piotrowicz can you review?

edmundhung (approved):

PR #4576, opened 2025-07-18
disallow jsg::Promise in v8 fast api

anonrig:
We disallow kj::Promise but not jsg::Promise. Let's disallow it.


jasnell (changes requested):
As mentioned, needs a test added to fast-api-test.c++

jasnell (approved):

anonrig:
(Rebased and force pushed)

PR #4580, opened 2025-07-21
[build] Split workerd-api from server target

fhanau:
This allows us to only have test-fixture depend on workerd-api. This allows us to shrink the size of some of our largest binaries and reduce thinLTO link times, test-fixture test shrinks by 6%.


anonrig:
It seems building benchmarks are now failing

fhanau:
> It seems building benchmarks are now failing

The errors look unrelated to this PR, but concerning nevertheless:
- thinLTO is enabled for the codspeed benchmarks (I don't necessary think it shouldn't be but in the past we generally considered thinLTO to be too expensive for the workerd CI environment; that may no longer be the case today but if we use thinLTO we should apply the optimizations we have for it internally as well)
- We link with the (very slow) bfd system linker, despite specifying `-fuse-ld=lld`
- bfd uses a LLVM backend based on LLVM 13??

fhanau:
Merging this since the same error appears on main, but I'll still try to see if I can find the cause...

PR #4582, opened 2025-07-21
Implements default entrypoint support for Python.

dom96:
Implements support for default WorkerEntrypoint in Python workers:

```python
from workers import Response, WorkerEntrypoint

class default(WorkerEntrypoint):
  async def on_fetch(self, req):
    return Response("hello")
```


danlapid (approved):

anonrig (commented):
Can we have a test?

dom96:
We have multiple in the EW repo

anonrig (approved):

PR #4583, opened 2025-07-21
[ci] Always use lld in CI on Linux

fhanau:
This should have already been working, but the benchmark run now appears to be broken due to trying to link with the system linker.

Trying to fix benchmark build, which appears to be broken suddenly.


fhanau:
PR is passing with this – I'm not sure if this actually solved the issue or just caused the bazel cache to be invalidated and that somehow improved things, but at worst this introduces some redundant definitions so let's go with it.

PR #4584, opened 2025-07-21
More refinement of webfs impl

jasnell:
Reworks the Webfs implementation to be more correct and pass more of the WPT's.

Simplifies things a bit by removing the sync fs api for now (we can add it back in later)

### DIFF
--- a/src/workerd/api/filesystem.c++
+++ b/src/workerd/api/filesystem.c++
@@ -2019,17 +2019,106 @@ jsg::Ref<jsg::DOMException> fsErrorToDomException(jsg::Lock& js, workerd::FsErro
 }
 }  // namespace
 
+FileSystemHandle::FileSystemHandle(
+    const workerd::VirtualFileSystem& vfs, jsg::Url&& locator, jsg::USVString name)
+    : vfs(vfs),
+      locator(kj::mv(locator)),
+      name(kj::mv(name)) {}
+
+jsg::Promise<kj::StringPtr> FileSystemHandle::getUniqueId(
+    jsg::Lock& js, const jsg::TypeHandler<jsg::Ref<jsg::DOMException>>& deHandler) {
+  KJ_IF_SOME(item, vfs.resolve(js, getLocator(), {})) {
+    KJ_SWITCH_ONEOF(item) {
+      KJ_CASE_ONEOF(file, kj::Rc<workerd::File>) {
+        return js.resolvedPromise(file->getUniqueId(js));
+      }
+      KJ_CASE_ONEOF(dir, kj::Rc<workerd::Directory>) {
+        return js.resolvedPromise(dir->getUniqueId(js));
+      }
+      KJ_CASE_ONEOF(link, kj::Rc<workerd::SymbolicLink>) {
+        return js.resolvedPromise(link->getUniqueId(js));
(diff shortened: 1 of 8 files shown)


jasnell:
Looks like, at least on mac, the wpt changes need more work and the snapshots need to be updated. Will make the additional edits tomorrow.

jasnell:
mac osx wpt failure was an unrelated flake...

src/wpt/fs-test.ts
```diff
@@ -64,14 +64,13 @@ export default {
...
-      'remove() on an already removed file should fail',
       'remove() to remove an empty directory',
       'remove() on an already removed directory should fail',
       'remove() on a non-empty directory should fail',
       'remove() on a directory recursively should delete all sub-items',
-      'remove() on a file should ignore the recursive option',
       'remove() while the file has an open writable fails',
+
+      // Intentionally unsupported
       'can remove the root of a sandbox file system',
```
npaun, line 74: You can put this under `omittedTests` to get it off our stats.

jasnell: I will once it's entirely updated. I'm keeping it here for now just in case I accidentally change something that causes the expectation to fail.

jasnell:
Btw, the biggest change here is that the WebFs classes no longer directly hold the `kj::Rc<...>` for the underlying fs node and instead will look it up each time an operation is called... which is what the spec actually expects. Per the spec, and per the tests, it is actually possible to acquire a handle for a node, have that node be deleted and recreated, and the handle would still be viable/usable. The prior implementation failed that expectation.

This also still isn't perfect, as there are a few other bits in here that need to be fixed, but I'm trying to go through these in smaller, more consumable/reviewable chunks rather than all at once.

src/workerd/api/filesystem.c++
```diff
@@ -2019,17 +2019,106 @@ jsg::Ref<jsg::DOMException> fsErrorToDomException(jsg::Lock& js, workerd::FsErro
...
+  auto recursive = opts.recursive.orDefault(false);
+  KJ_IF_SOME(parent, vfs.resolve(js, relative.base, workerd::VirtualFileSystem::ResolveOptions{})) {
+    KJ_SWITCH_ONEOF(parent) {
+      KJ_CASE_ONEOF(parentDir, kj::Rc<workerd::Directory>) {
+        // Webfs requires that the entry exists before we try to remove it.
+        kj::Path path({name});
+        if (parentDir->stat(js, path) == kj::none) {
+          auto ex = js.domException(kj::str("NotFoundError"), kj::str("The entry was not found."));
+          return js.rejectedPromise<void>(deHandler.wrap(js, kj::mv(ex)));
+        }
```
npaun, line 2078: This looks the same as: https://github.com/cloudflare/workerd/pull/4584/files#diff-f82d13fe2be1bc6edb903a4f830ad17b7209d8da05cb7b45eb1e8e224cbc96beR2085

I'm not sure I understand the distinction between the two cases.

anonrig (approved):

src/workerd/api/filesystem.c++
jasnell: Just two different paths to ending at the same result.

PR #4585, opened 2025-07-21
[o11y] Specify "trace" event type for Streaming Tail Worker

fhanau:
This matches the existing tail worker.
For the event info provided in the trace onset, use CustomEventInfo as with TraceEventInfo we'd always end up reporting an empty array.


fhanau:
Discussed this with Dan – we'll keep the "trace" type for the Onset event info, so this PR reduces to providing the right event type for the CustomEvent.

PR #4588, opened 2025-07-22
feat(types): allow Service<T> to handle non-WorkerEntrypoint default export

edmundhung:
This PR updates `Service<T>` so that:

- Returns `Fetcher<T>` when `T` is a `WorkerEntrypoint` class (Current behavior)
- Returns `Fetcher<InstanceType<T>>` when `T` is a `WorkerEntrypoint` constructor
- Returns  `never` when `T` is a `DurableObject` or `WorkflowEntrypoint`
- Returns `Fetcher` otherwise

This will enable `wrangler types` to infer the RPC methods from the default export with `Service<typeof import("./worker").default>` next.

Ref: https://github.com/cloudflare/workers-sdk/pull/9897

### DIFF
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -436,10 +436,17 @@ class Fetcher: public JsRpcClientProvider {
       });
     }
     JSG_TS_DEFINE(
-      type Service<T = undefined> = Fetcher<
-          T extends new (...args: any[]) => infer EntrypointClass
-              ? EntrypointClass extends Rpc.WorkerEntrypointBranded ? EntrypointClass : undefined
-              : T extends Rpc.WorkerEntrypointBranded ? T : undefined
+      type Service<
+        T extends
+          | (new (...args: any[]) => Rpc.WorkerEntrypointBranded)
+          | Rpc.WorkerEntrypointBranded
+          | ExportedHandler<any, any, any>
+          | undefined = undefined,
+      > = Fetcher<
+        T extends new (...args: any[]) => infer EntrypointClass ? EntrypointClass
+        : T extends Rpc.WorkerEntrypointBranded ? T
+        : T extends Exclude<Rpc.EntrypointBranded, Rpc.WorkerEntrypointBranded> ? never
+        : undefined
       >;
     );
 
--- a/src/workerd/api/http.h
+++ b/src/workerd/api/http.h
@@ -436,7 +436,11 @@ class Fetcher: public JsRpcClientProvider {
       });
     }
     JSG_TS_DEFINE(
-      type Service<T extends Rpc.WorkerEntrypointBranded | undefined = undefined> = Fetcher<T>;
+      type Service<T = undefined> = Fetcher<
+          T extends new (...args: any[]) => infer EntrypointClass
+              ? EntrypointClass extends Rpc.WorkerEntrypointBranded ? EntrypointClass : undefined
+              : T extends Rpc.WorkerEntrypointBranded ? T : undefined
+      >;
     );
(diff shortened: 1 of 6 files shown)


edmundhung:

types/test/types/rpc.ts
```diff
@@ -406,11 +406,10 @@ class TestNaughtyObject extends DurableObject {
 interface Env {
   REGULAR_SERVICE: Service;
   RPC_SERVICE: Service<TestEntrypoint>;
+  TYPEOF_RPC_SERVICE: Service<typeof TestEntrypoint>;
   NAUGHTY_SERVICE: Service<TestNaughtyEntrypoint>;
   // @ts-expect-error `BoringClass` isn't an RPC capable type
   __INVALID_RPC_SERVICE_1: Service<BoringClass>;
-  // @ts-expect-error `TestEntrypoint` is a `DurableObject`, not a `WorkerEntrypoint`
-  __INVALID_RPC_SERVICE_2: Service<TestObject>;
```
edmundhung, line 413: I tried to restrict the type accepted but the DurableObject class matches the ExportedHandler interface here 🙃 

As a workaround, I have added additional logic to resolve DurableObject and WorkflowEntrypoint to `never`.

src/workerd/api/http.h
```diff
@@ -436,7 +436,18 @@ class Fetcher: public JsRpcClientProvider {
       });
     }
     JSG_TS_DEFINE(
-      type Service<T extends Rpc.WorkerEntrypointBranded | undefined = undefined> = Fetcher<T>;
+      type Service<
+        T extends
+          | (new (...args: any[]) => Rpc.WorkerEntrypointBranded)
+          | Rpc.WorkerEntrypointBranded
+          | ExportedHandler<any, any, any>
+          | undefined = undefined,
```
penalosa, line 444: What are these two for?

types/test/types/rpc.ts
penalosa: Could you expand on this?

src/workerd/api/http.h
edmundhung: We accept `undefined` before this change, so this just make sure it is non-breaking.

`ExportedHandler<any, any, any>` is for case we receive an object with the fetch handler (non worker entrypoint)

We can remove all these constraints if we don't mind it accept any generics. I am just trying to follow the same pattern we are doing on the Fetcher type to restrict what we accept.

types/test/types/rpc.ts
edmundhung: Before my change, type check fail here because we restrict the generic type to match the `Rpc.WorkerEntrypointBranded` type, which the DurableObject type doesn't implement. But now that we accept `ExportedHandler<any, any, any>`, which is effectively this:

```ts
{
  fetch(request: Request): Promise<Response>;
  // ...
}
```

I couldn't find any way to restrict such case but still accepting `ExportedHandler`. So I have added this condition in the Service type:

```ts
T extends Exclude<Rpc.EntrypointBranded, Rpc.WorkerEntrypointBranded>
  ? never
  : undefined`
```

If user provides a DurableObject class to the `Service` type, typescript won't fail right away, but the result type will become `never`.

src/workerd/api/http.h
penalosa: Sounds good!

types/test/types/rpc.ts
penalosa: Sounds good!

harrishancock (approved):

PR #4590, opened 2025-07-22
Add startup exception detail and clean up extra log line

danlapid:


src/workerd/io/worker.c++
```diff
@@ -302,8 +302,10 @@ void reportStartupError(kj::StringPtr id,
               message.addJsStackTrace(js, lines);
               auto trace = kj::strArray(lines, "; ");
               auto description = KJ_ASSERT_NONNULL(permanentException).getDescription();
-              KJ_LOG(ERROR, "script startup threw exception", id, description, trace);
-              KJ_FAIL_REQUIRE("script startup threw exception");
+              auto e =
+                  KJ_EXCEPTION(FAILED, "script startup threw exception", id, description, trace);
+              e.setDetail(STARTUP_EXCEPTION_DETAIL_ID, kj::heapArray<kj::byte>(0));
```
jasnell, line 307: FWIW, we have planned (at least eventually) to use `setDetail(...)` to attach a serialized JS exception to `kj::Exception`. Doesn't mean this needs to change but just as an FYI in case we run into competing cases for this additional field in the future.

jasnell (approved):

src/workerd/io/worker.c++
danlapid: Could always add it on another detail id

jasnell: Yep, just throwing in a reminder as use of this starts to expand. Want to make sure that folks are keeping the various use cases in mind ;-)

danlapid:
Closing in favor of another option. Although I would like to merge those log messages one day.

PR #4592, opened 2025-07-22
Update sidecar tests to use SIDECAR_HOST environment variable

npaun:
## Summary
- Updated all sidecar-based tests to use `SIDECAR_HOST` environment variable for dynamic IP assignment
- Enhanced server logging to display full IP:port addresses for better debugging
- Added SIDECAR_HOST support to WPT (Web Platform Tests) system

## Background
The sidecar test framework runs auxiliary server processes alongside tests to provide realistic network endpoints. Previously, these sidecars used hardcoded localhost/127.0.0.1 addresses with random ports, which could still cause conflicts when multiple test suites ran in parallel on the same machine.

The sidecar-supervisor.mjs was recently enhanced to assign random IP addresses in the 127.x.x.x range (stored in `SIDECAR_HOST`) in addition to random ports. This provides better test isolation by ensuring each test suite gets its own unique IP address space, preventing network conflicts and improving reliability in CI environments and local parallel test execution.

## Changes Made
1. **Node.js API Tests**: Updated all `.wd-test` files to include SIDECAR_HOST environment binding
2. **Server Files**: Modified all sidecar server files to listen on `process.env.SIDECAR_HOST` instead of default host
3. **Test Files**: Updated client connections to use `SIDECAR_HOST` instead of hardcoded localhost
4. **Logging**: Enhanced `reportPort()` functions to show full `IP:port` format instead of just port numbers
5. **WPT System**: Updated build/wpt_test.bzl and harness.ts to support SIDECAR_HOST with localhost fallback

## Benefits
- **Better Test Isolation**: Each test suite gets its own IP address space
- **Reduced Conflicts**: Eliminates potential network binding conflicts in parallel execution
- **Improved CI Reliability**: More stable test execution in containerized environments
- **Enhanced Debugging**: Full IP:port logging makes network issues easier to diagnose

## Test Coverage
- Updated 24 files across Node.js API tests, regular API tests, and WPT system
- All sidecar-based tests now support dynamic IP assignment
- Maintains backward compatibility in WPT harness with localhost fallback

## Test Plan
- [x] Run Node.js API tests: `just node-test net`, `just node-test http`, `just node-test tls`
- [x] Run WebSocket tests: `bazel test //src/workerd/api:tests/websocket-client-error-test`
- [x] Run WPT tests to verify harness changes
- [x] Verify server logs show full IP:port addresses
- [x] Ensure no hardcoded localhost/127.0.0.1 references remain in sidecar tests


npaun:
Claude's explanation is rather prolix.

 TL;DR: We're going to avoid port clash by also assigning a random IP in 127.0.0.0/8. This works as-is on Linux, but on macOS you need to add the IP with `ifconfig`, so I'm going to only do that in CI. Locally, we can probably get away with `127.0.0.1`.

npaun:
Bonus: I asked Claude to do some first year university math and check what value we could use for `--runs_per_test`. Like a good vibe coder, I am neglecting to actually check its work.

⏺ I'll solve this using the birthday paradox, which is perfect for this type of collision probability problem.

  For the network 127.0.0.0/8:
  - Network has 24 host bits (32 - 8 = 24)
  - Total possible IPs: 2^24 = 16,777,216

  Using the birthday paradox formula, the probability that at least two processes select the same IP is:

  P(collision) = 1 - P(all different)

  Where P(all different) for n processes is:
  P(all different) = (N/N) × ((N-1)/N) × ((N-2)/N) × ... × ((N-n+1)/N)

  For large N and moderate n, this approximates to:
  P(collision) ≈ 1 - e^(-n²/(2N))

  Setting this equal to 0.5 and solving for n:
  0.5 = 1 - e^(-n²/(2N))
  e^(-n²/(2N)) = 0.5
  -n²/(2N) = ln(0.5) = -0.693
  n² = 2N × 0.693 = 1.386N
  n = √(1.386N) = √(1.386 × 16,777,216) ≈ 4,823

  Therefore, approximately 4,823 processes would be needed before the probability of two processes selecting the same IP exceeds 50%.

  This follows the classic birthday paradox result where you need roughly √(πN/2) items for a 50% collision probability in a space of size N.

src/workerd/api/node/tests/http-client-nodejs-server.js
```diff
@@ -18,12 +19,16 @@ const pongServer = http.createServer((req, res) => {
     res.end('pong');
   });
 });
-pongServer.listen(process.env.PONG_SERVER_PORT, () => reportPort(pongServer));
+pongServer.listen(process.env.PONG_SERVER_PORT, process.env.SIDECAR_HOST, () =>
+  reportPort(pongServer)
+);
```
anonrig: If we make these 3 lines a method, it will be a lot more readable, I assume.

src/workerd/api/node/tests/http-nodejs-test.js
```diff
@@ -346,6 +370,7 @@ export const testHttpDontSetDefaultHeadersWithSetHost = {
     http
       .request({
         method: 'POST',
+        hostname: env.SIDECAR_HOST,
```
anonrig: I believe the name should end with HOSTNAME

```suggestion
        hostname: env.SIDECAR_HOSTNAME,
```

src/workerd/api/node/tests/http-nodejs-test.js
```diff
@@ -379,7 +407,11 @@ export const testHttpRequestEndTwice = {
 export const testHttpSetTimeout = {
   async test(_ctrl, env) {
     const { promise, resolve, reject } = Promise.withResolvers();
-    const request = http.get({ port: env.TIMEOUT_SERVER_PORT, path: '/' });
+    const request = http.get({
+      hostname: env.SIDECAR_HOST,
+      port: env.TIMEOUT_SERVER_PORT,
```
anonrig: These changes would be a lot less if we change TIMEOUT_SERVER_PORT to TIMEOUT_SERVER_HOST where it is equal to `localhost:8080`

src/workerd/api/node/tests/sidecar-supervisor.mjs
```diff
@@ -60,13 +63,53 @@ function runSidecar(cmd) {
...
+  if (process.platform === 'linux') {
+    child_process.spawnSync(
+      'sudo',
+      ['/usr/sbin/ip', 'route', 'show', 'table', 'all'],
+      { stdio: 'inherit' }
+    );
+    // We can just start using the IP
+    return getRandomLoopbackHost();
+  } else if (process.platform === 'darwin') {
+    if (process.env.CI === '1') {
```
anonrig: Can you assign this to a variable on top rather than having this check here

src/workerd/api/node/tests/sidecar-supervisor.mjs
```diff
@@ -60,13 +63,53 @@ function runSidecar(cmd) {
...
+  return `127.${crypto.randomInt(2, 255)}.${crypto.randomInt(2, 255)}.${crypto.randomInt(2, 255)}`;
+}
+
+function assignLoopbackHost() {
+  if (process.platform === 'linux') {
+    child_process.spawnSync(
+      'sudo',
+      ['/usr/sbin/ip', 'route', 'show', 'table', 'all'],
+      { stdio: 'inherit' }
+    );
```
anonrig: What's the purpose of this call here?

src/workerd/api/node/tests/sidecar-supervisor.mjs
```diff
@@ -60,13 +63,53 @@ function runSidecar(cmd) {
   return { promise, proc };
 }
 
+function getRandomLoopbackHost() {
+  // Pick a random address in 127.0.0.0/8.
+  // Range is chosen not to use network address, gateway address, or broadcast address.
+  return `127.${crypto.randomInt(2, 255)}.${crypto.randomInt(2, 255)}.${crypto.randomInt(2, 255)}`;
```
anonrig, line 121: What happens if this loopback host is taken? We don't check it anywhere

src/workerd/api/node/tests/sidecar-supervisor.mjs
```diff
@@ -60,13 +63,53 @@ function runSidecar(cmd) {
...
+        ['/sbin/ifconfig', 'lo0', 'alias', randomHost],
+        { stdio: 'inherit' }
+      );
+      return randomHost;
+    } else {
+      // If not running in CI, we will just use 127.0.0.1 to avoid bothering devs with sudo
+      return '127.0.0.1';
+    }
+  } else {
+    // TODO(soon): I don't even want to think about adding extra IPs on Windows
```
anonrig:
```suggestion
    // TODO: Randomize host number on Windows, if needed.
```

anonrig: Can you add an assertion here that process.platform is win32

src/workerd/api/tests/websocket-client-error-sidecar.js
```diff
@@ -68,8 +68,14 @@ function sendMessage(socket, message) {
   socket.write(frame);
 }
 
-// Get the port from environment variable
+// Get the port and host from environment variables
 const port = process.env.BIG_MESSAGE_SERVER_PORT || 9000;
+const host = process.env.SIDECAR_HOST || 'localhost';
```
anonrig: Why is this optional?

src/workerd/api/node/tests/sidecar-supervisor.mjs
npaun: It's not working properly in CI so I have to debug like this. This stuff will go away soon.

src/workerd/api/node/tests/sidecar-supervisor.mjs
npaun: If a host is taken, we'll get some random ports on it anyway. This will increase the likelihood of collision but will otherwise work.

anonrig: Can we detect collision, and improve the case where a collusion occurs?

npaun:
@anonrig I think Claude does explain the motivation reasonably well. Please let me know if anything seems missing or unclear in this:

> Background
> The sidecar test framework runs auxiliary server processes alongside tests to provide realistic network endpoints. Previously, these sidecars used hardcoded localhost/127.0.0.1 addresses with random ports, which could still cause conflicts when multiple test suites ran in parallel on the same machine.

> The sidecar-supervisor.mjs was recently enhanced to assign random IP addresses in the 127.x.x.x range (stored in SIDECAR_HOST) in addition to random ports. This provides better test isolation by ensuring each test suite gets its own unique IP address space, preventing network conflicts and improving reliability in CI environments and local parallel test execution.

npaun:
Marking as draft while I fight with routing tables on the CI runners.

src/workerd/api/node/tests/sidecar-supervisor.mjs
npaun: Maybe... it's going to be OS-specific. Let's see what comes out of my investigation on why it's not working on CI.

anonrig:
@npaun This PR is amazing, and is on the right track of improving things. The motivation is clear, and I have no blocker regarding the implementation. I just think we should have a documentation written in the repository and not in pull-request description. It'll get harder and harder to keep track of what is supported and what's not in the test runner (over time).

src/workerd/api/node/tests/http-nodejs-test.js
npaun: I thought so too, and decided to try it, but there are lots of methods in the Node API (like `server.listen`) that want to be passed host and port separately.

anonrig: Oof. I expected more from Node.js

src/workerd/api/node/tests/sidecar-supervisor.mjs
npaun: My conclusion is that it'd be really ugly, and would involve parsing the output of some CLI tools. I'll add a TODO though indicating that that IP address collisions aren't checked.

src/workerd/api/tests/websocket-client-error-sidecar.js
npaun: Dunno. The existing code was like this, but it's gone now.

src/workerd/api/node/tests/sidecar-supervisor.mjs
npaun: I'm not sure which value I need to assign to a variable here. Can you tell me which?

npaun: This code's changed a lot anyway. I think it should be clear now.

jasnell (approved):

PR #4595, opened 2025-07-23
Add rpc/serialization support for TraceItem

ketanhwr:
I ideally wanted to make all subtypes (eg: `FetchEventInfo`, `TraceLog`, etc) serializable as well, and then have a method `toCapnp()` in each of them which I could reuse in serializing `TraceItem`. But the constructors of most of these types need a `Trace&` which makes it a bit challenging from the point of view of deserializing.

Is there any particular reason passing `Trace&` in the constructor for these types, given it's not used at all? (Is it because we don't want to allow construction of these types unless it's part of a `TraceItem`?)

Do I need a compat flag for this? If yes - how exactly do I feature-flag `JSG_SERIALIZABLE`?


edmundhung:
cc: @Ankcorn

danlapid:
definitely needs a compat flag, should be experimental as well until it is fully rolled out

You can look abortSignalRpc.

src/workerd/api/tail-worker-rpc-test.wd-test
```diff
@@ -0,0 +1,100 @@
...
+  services = [
+    # Embedded tests we get traces from
+    ( name = "http-test",
+      worker = (
+        modules = [
+          ( name = "worker", esModule = embed "http-test.js" )
+        ],
+        bindings = [
+          ( name = "SERVICE", service = "http-test" ),
+          ( name = "CACHE_ENABLED", json = "false" ),
```
danlapid, line 13: unused?

src/workerd/api/trace.c++
```diff
@@ -310,6 +323,306 @@ uint TraceItem::getWallTime() {
...
+            valueBuilder.setFloat64(d);
+          }
+          KJ_CASE_ONEOF(i, int64_t) {
+            valueBuilder.setInt64(i);
+          }
+        }
+      }
+    }
+  }
+  auto words = capnp::messageToFlatArray(traceMessage);
```
kentonv, line 605: Hmm, it's sort of ugly to be embedding a capnp messages into v8-serialized JS objects. The v8-serialized object itself is just going to be embedded into another capnp message.

I think ideally we'd actually serialize this as a straight JS object. That sort of makes sense here, as TraceItem is essentially just a plain JS object (and trace workers commonly reserialize it as JSON, even). However, I guess there isn't a convenient way to actually achieve such serialization from here, because we don't actually have that JS object, we have a C++ wrapped object...

The other thing worth exploring is whether we should use an "external" here. See `JsValue::External` in `worker-interface.capnp`. You could extend that with `traceItem :Trace`. This would mean the serialization code here would _only_ work with Workers RPC, though. You'd start the serialize() function with code like:

```
  auto& handler = JSG_REQUIRE_NONNULL(serializer.getExternalHandler(), DOMDataCloneError,
      "Traces can only be serialized for RPC.");
  auto externalHandler = dynamic_cast<RpcSerializerExternalHandler*>(&handler);
  JSG_REQUIRE(externalHandler != nullptr, DOMDataCloneError,
      "Traces can only be serialized for RPC.");
```

But I dunno, maybe the turducken approach is fine and easiest to implement?

ketanhwr: TIL about turducken.

jasnell: I have *ALWAYS* really disliked the fact that these were defined as jsg::Objects from the beginning. They should have been regular objects from the start and the direction for streaming tail workers, unless it's changed, is that those would be regular objects. I think maybe, tho, we can still serialize these as ordinary objects within the custom serialization.

jasnell: It's likely going to be easier/better for us to just construct an internal plain-ole-javascript-object representation here and serialize that instead. Or we might just be able to get by with JSON stringifying the object, serializing that, and parsing it on the deserialize side, using the result to reconstruct the `TraceItem`.

ketanhwr: AFAIK, @Ankcorn is already doing the JSON stringifying thing, and it's taking _a lot_ of cpu time. He's eager about this change especially to see the CPU usage go down. I went the "turducken" approach since it makes the deserialization quite trivial.

kentonv: Hmm, if the goal is to save CPU time then we'd better be doing some benchmarks to show that this actually does that, because I could easily imagine that it doesn't.

ketanhwr: Hmm fair enough - should've measured it properly beforehand (although I assumed that JSON conversion in javascript land will always be slower).

I've created a simple multi-worker benchmark setup here which measures time taken by both the C++ serialization as well as JSON parse/stringify method: https://github.com/cloudflare/workerd/commit/a492fc373c9c5e30bdbe8549c1ac3e5b920e9b31

The main part is in `proxy-worker.js`:

```js
    // Measure JSON.stringify serialization time
    const jsonStart = performance.now();
    await this.env.SERVICE.tail(JSON.parse(JSON.stringify(events)));
    const jsonEnd = performance.now();
    const jsonTime = jsonEnd - jsonStart;

    // Measure C++ serialization time (direct object passing)
    const cppStart = performance.now();
    await this.env.SERVICE.tail(events);
    const cppEnd = performance.now();
    const cppTime = cppEnd - cppStart;
```

(I updated `Performance.now()` to return microseconds instead of milliseconds for higher precision). After a few thousand samples here's the output:

```
Returning performance stats: Performance Statistics (10007 samples):
C++ Serialization: 443.6219ms average
JSON.stringify: 761.9454ms average
Speedup factor: 1.72x
```

This does include serialization + sending it across RPC. Let me know if I'm misinterpreting the output somehow or missing something in the benchmark - but looks like a decent enough improvement?

ketanhwr: FYI - I tried to change the order or JS vs C++ and also tried them individually to check if some connection re-use/other RPC optimizations behind the scenes didn't cause the performance diff - but still saw roughly ~1.7x speedup in all cases.

kentonv: Is 1.7x actually that big an improvement? It's not a paradigm shift, and moving this responsibility to C++ adds a fair amount of inflexibility and maintenance burden.

I wonder if there are ways we could optimize the JSON serialization to get 70% faster. For example, currently the V8 JSON serializer is being forced to walk these wrapped objects that are fairly slow to access, since each property access is going to go through JSG. If it were walking native JS objects, would it be faster? Probably this API should have been native JS objects all along, and I wonder if we could change it (with a compat flag?). Alternatively, what if these types had native `toJSON` methods, so that V8 doesn't have to walk them -- we build the JSON in pure C++ code. How much faster would that be?

ketanhwr: This calculation unfortunately involves the time taken for rest of the RPC lifecycle as well - I tried to sample times more accurate by just calling `structuredClone` instead and noticed ~3x speed up in that case. But yeah - better but not a _huge_ improvement.

Also - wouldn't implementing a `toJSON` method be quite similar to implementing and maintaining a `serialize()` function? As for implementing this natively in JS behind a compat flag, I can certainly take a look but I'm not sure if it impacts/overlaps with the work @fhanau has been doing lately

kentonv: > Also - wouldn't implementing a toJSON method be quite similar to implementing and maintaining a serialize() function?

The difference is:
* It would immediately speed up existing use cases that are JSON-serializing these objects, with no changes on their end.
* It does not introduce any new protocols that we need to maintain forever.

I think the first point may make it worth doing regardless. JSON-serializing these objects is pretty common in tail workers, and in some cases cannot be replaced with RPC since the JSON is being sent to things that aren't workers.

Maybe once we know how fast that is we can decide if it's worth introducing a new RPC format as well.

kentonv: And to reiterate, I would expect the reason JSON serialization of these objects is so slow today is not so much because of JSON, but because V8 has to walk the tree of wrapped objects, calling getter callbacks for every property. Implementing toJSON() would avoid that.

PR #4600, opened 2025-07-23
Moar webfs excitement

jasnell:
More incremental work...

### DIFF
--- a/src/workerd/api/filesystem.h
+++ b/src/workerd/api/filesystem.h
@@ -500,6 +500,11 @@ class FileSystemDirectoryHandle final: public FileSystemHandle {
     JSG_METHOD(values);
     JSG_METHOD(forEach);
     JSG_ASYNC_ITERABLE(entries);
+
+    JSG_TS_OVERRIDE({
+      entries(): AsyncIterableIterator<[string, FileSystemHandle]>;
+      [Symbol.asyncIterator]() : AsyncIterableIterator<[string, FileSystemHandle]>;
+    });
   }
 
  private:
--- a/src/wpt/fs-test.ts
+++ b/src/wpt/fs-test.ts
@@ -84,9 +84,9 @@ export default {
     expectedFailures: [],
   },
   'FileSystemDirectoryHandle-resolve.https.any.js': {
-    comment: '...',
-    expectedFailures: [
-      // TODO(node-fs): Fix these tests
+    comment:
+      'We currently do not implement the resolve() method on directories',
+    disabledTests: [
       'Resolve returns empty array for same directory',
       'Resolve returns correct path',
       'Resolve returns correct path with non-ascii characters',
--- a/types/generated-snapshot/experimental/index.d.ts
+++ b/types/generated-snapshot/experimental/index.d.ts
@@ -3179,7 +3179,7 @@ declare abstract class FileSystemDirectoryHandle extends FileSystemHandle {
   ): Promise<void>;
   /* [MDN Reference](https://developer.mozilla.org/docs/Web/API/FileSystemDirectoryHandle/resolve) */
   resolve(possibleDescendant: FileSystemHandle): Promise<string[]>;
-  entries(): AsyncIterableIterator<(string | FileSystemHandle)[]>;
+  entries(): AsyncIterableIterator<[string, FileSystemHandle]>;
   keys(): AsyncIterableIterator<string>;
   values(): AsyncIterableIterator<FileSystemHandle>;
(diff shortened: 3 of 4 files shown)


src/workerd/api/tests/webfs-test.js
```diff
@@ -125,7 +126,7 @@ export const asyncIteratorTest = {
 
     let count = 0;
     for await (const node of dir) {
-      ok(node.value instanceof FileSystemDirectoryHandle);
+      ok(node[1] instanceof FileSystemDirectoryHandle);
```
anonrig, line 131: Can you add a comment to what [0] and [1] refers to? It would be easier to debug if one of the tests fail in the future.

src/wpt/fs-test.ts
```diff
@@ -69,38 +69,22 @@ export default {
   },
   'FileSystemDirectoryHandle-getDirectoryHandle.https.any.js': {
     comment: '...',
-    expectedFailures: [
-      // TODO(node-fs): Fix these tests
-      'getDirectoryHandle(create=false) with a path separator when the directory exists',
-      'getDirectoryHandle(create=true) with a path separator',
-    ],
+    expectedFailures: [],
```
anonrig, line 72: We can just remove all contents inside the parent object

src/workerd/api/filesystem.c++
```diff
@@ -1965,7 +1965,8 @@ jsg::JsValue createUVException(jsg::Lock& js,
 
 namespace {
 constexpr bool isValidFileName(kj::StringPtr name) {
-  return name.size() > 0 && name != "."_kj && name != ".."_kj && name.find("/"_kj) == kj::none;
+  return name.size() > 0 && name != "."_kj && name != ".."_kj && name.find("/"_kj) == kj::none &&
```
anonrig, line 1968: Can you add a comment to this function explaining what a valid file name consist of?

src/workerd/api/filesystem.h
```diff
@@ -374,18 +374,11 @@ class FileSystemFileHandle final: public FileSystemHandle {
...
-    EntryType(jsg::USVString key, jsg::Ref<FileSystemHandle> value)
-        : key(kj::mv(key)),
-          value(kj::mv(value)) {}
-  };
-
  private:
-  using EntryIteratorType = EntryType;
+  // The actual entry type is an array of two elements, the first being the key,
+  // the second being the value.
+  using EntryType = kj::OneOf<jsg::JsRef<jsg::JsString>, jsg::Ref<FileSystemHandle>>;
```
npaun, line 374: Why replace that explicit EntryType struct with a comment explaining what that struct looks like?

jasnell: Not sure what you mean. The struct was wrong. The entry type here is an array whose first element is a string and whose second element is the handle.

src/wpt/fs-test.ts
jasnell: Yep, this is temporary while I'm working through these... I'm using it as a visual reminder that I've already looked at these. It'll all end up with good formatting when all done.

src/workerd/api/filesystem.c++
jasnell: I'm going to be coming back to this function and doing some perf optimization on it. I'll add the comment then.

types/generated-snapshot/experimental/index.d.ts
```diff
@@ -3190,7 +3190,9 @@ declare abstract class FileSystemDirectoryHandle extends FileSystemHandle {
     ) => void,
     thisArg?: any,
   ): void;
-  [Symbol.asyncIterator](): AsyncIterableIterator<FileSystemDirectoryHandleEntryType>;
+  [Symbol.asyncIterator](): AsyncIterableIterator<
+    (string | FileSystemHandle)[]
```
guybedford: Shouldn't this be `[string, FileSystemHandle]`?

jasnell: Yes, this is what the automatic JSG-based type generation produces. I'm going to need to put in an override. I plan on adding the overrides as a follow up... i guess i could do it now tho lol

jasnell: done

src/workerd/api/filesystem.h
npaun: Ah. I see.

npaun (approved):

guybedford (approved):

PR #4610, opened 2025-07-24
Extend hrtime check for flakes

guybedford:
//cc @jasnell I can't reproduce these locally, but perhaps it may help again.


npaun (approved):
I'm not James, but the change LGTM.

PR #4612, opened 2025-07-25
Pyodide 0.28

hoodmane:


src/pyodide/internal/pool/builtin_wrappers.ts
```diff
@@ -206,11 +206,16 @@ function prepareStackTrace(
   try {
     const funcName = stack[2].getFunctionName();
     const fileName = stack[2].getFileName();
-    if (fileName !== 'pyodide-internal:generated/emscriptenSetup') {
+    if (fileName !== 'pyodideRuntime-internal:emscriptenSetup') {
```
dom96: Why has this been renamed?

hoodmane: I'm expecting to revert this before it's ready to be merged. But it's because I commented out the code in
`src/workerd/server/workerd-api.c++` and loaded the module from JS instead of from C++. This requires several bazel changes to work which have a consequence of changing the name of the file.

hoodmane: Reverted.

hoodmane:
Remaining failing tests: `import/soupsieve`, `importable-env`, `python-rpc`, and `durable-object`. I managed to break `durable-object` even on 0.26.0a2.

src/pyodide/helpers.bzl
```diff
@@ -258,6 +258,11 @@ def _python_bundle(version, *, pyodide_asm_wasm = None, pyodide_asm_js = None, p
             "function _PyEM_CountFuncParams(func){",
             "function _PyEM_CountFuncParams(func){ return patched_PyEM_CountFuncParams(Module, func);",
         ],
+        [
+            "var tableBase=metadata.tableSize?wasmTable.length:0;",
+            "var tableBase=metadata.tableSize?wasmTable.length:0;" +
+            "Module.snapshotDebug && console.log('loadWebAssemblyModule', libName, memoryBase, tableBase);",
+        ],
```
dom96: Should this be removed?

src/pyodide/internal/_workers.py
```diff
@@ -163,10 +163,10 @@ async def blob(self) -> "Blob":
         return Blob(await self.js_object.blob())
 
 
-async def _pyfetch_patched(url: str, **kwargs: Any) -> FetchResponse:
-    # TODO: This is copied from https://github.com/pyodide/pyodide/blob/d3f99e1d/src/py/pyodide/http.py
+if pyodide_version == "0.26.0a2":
 
-    if pyodide_version == "0.26.0a2":
+    async def pyfetch(url: str, **kwargs: Any) -> FetchResponse:
```
dom96: I guess this is overriding the imported `pyfetch`? A little confusing, definitely preferable if you're explicit here:

```suggestion
    async def _pyfetch_patched(url: str, **kwargs: Any) -> FetchResponse:
```

src/pyodide/internal/pool/sentinel.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const sentinelWasm = decodeBase64(
+  'AGFzbQEAAAABDANfAGAAAW9gAW8BfwMDAgECByECD2NyZWF0ZV9zZW50aW5lbAAAC2lzX3NlbnRpbmVsAAEKEwIHAPsBAPsbCwkAIAD7GvsUAAs'
+);
+
+export async function getSentinelImport() {
+  const module: WebAssembly.Module = new WebAssembly.Module(sentinelWasm);
+  const instance = await WebAssembly.instantiate(module);
+  return instance.exports;
+}
```
dom96, line 50: Add some comments explaining what this sentinel is for

src/pyodide/internal/pool/sentinel.ts
```diff
@@ -0,0 +1,41 @@
...
+    const chunk1 = base64Lookup[input.charCodeAt(inIdx++)];
+    const chunk2 = base64Lookup[input.charCodeAt(inIdx++)];
+    const chunk3 = base64Lookup[input.charCodeAt(inIdx++)];
+    const chunk4 = base64Lookup[input.charCodeAt(inIdx++)];
+    output[outIdx++] = (chunk1 << 2) | (chunk2 >> 4);
+    output[outIdx++] = (chunk2 << 4) | (chunk3 >> 2);
+    output[outIdx++] = (chunk3 << 6) | chunk4;
+  }
+  return output;
+}
```
dom96: Can we use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64 here?

src/workerd/api/pyodide/setup-emscripten.c++
```diff
@@ -82,6 +82,8 @@ EmscriptenRuntime EmscriptenRuntime::initialize(
       pyodideAsmWasmReader = module.getData();
     }
   }
+  js.setAllowEval(true);
+  KJ_DEFER(js.setAllowEval(false));
```
dom96: Why is this necessary? Can you add a comment? Also can you add a guard somewhere to assert that `allowEval` is false before executing the user's Python code? Just in case we mess up in cases like this.

src/workerd/server/tests/python/durable-object/worker.py
```diff
@@ -105,7 +105,8 @@ async def test(ctrl, env, ctx):
     assert third_resp_data == "hello from python 3"
 
     # Wait for alarm to get triggered.
-    while True:
+    for n in range(20):
```
dom96: revert this change

src/workerd/server/tests/python/py_wd_test.bzl
```diff
@@ -3,7 +3,7 @@ load("//:build/wd_test.bzl", "wd_test")
 
 FEATURE_FLAGS = {
     "0.26.0a2": [],
-    "0.27.7": ["python_workers_20250116"],
+    "0.28.0": ["python_workers_20250116"],
```
dom96: Should this be renamed to `python_workers_20250808`?

src/workerd/server/tests/python/sdk/worker.py
```diff
@@ -423,7 +428,7 @@ async def response_unit_tests(env):
 
     response_none = Response(None, status=204)
     assert response_none.status == 204
-    assert response_none.body is None
+    assert response_none.body is jsnull
```
dom96: Hmm. This is the SDK which shouldn't leak `jsnull`, does the sdk need to be updated to transform jsnull into `None`?

dom96 (commented):
There are a few things I'd like to see changed so not approving just yet. But overall great stuff, thanks!

src/workerd/server/tests/python/durable-object/worker.py
hoodmane: I think this is better though. It needs another change to make it fail if it didn't succeed in 20 tries, but it should fail in a finite amount of time if the test isn't working rather than hanging.

src/pyodide/internal/pool/sentinel.ts
hoodmane: No, it's in spider monkey but not in v8 yet.

src/pyodide/helpers.bzl
hoodmane: No I think we should keep it since we will likely have to debug this more in the future.

src/workerd/server/tests/python/sdk/worker.py
hoodmane: I guess so.

src/workerd/server/tests/python/durable-object/worker.py
hoodmane: Split into #4740.

src/pyodide/helpers.bzl
hoodmane: Split into #4739.

src/pyodide/internal/pool/sentinel.ts
hoodmane: > Per their followup to your comment it will hit stable in Chrome on Sept 2.

https://github.com/tc39/proposal-arraybuffer-base64/issues/51#issuecomment-3010378969

So we can get rid of this in ~ one month.

src/workerd/api/pyodide/setup-emscripten.c++
hoodmane: Seems like it might not be necessary. I removed it, lets see if CI passes.

src/workerd/server/tests/python/py_wd_test.bzl
hoodmane: Okay.

src/workerd/server/tests/python/durable-object/worker.py
dom96: Yeah, that makes more sense. Without that change it would just succeed without actually potentially completing the test.

src/pyodide/internal/pool/sentinel.ts
dom96: There is also `atob`, though not sure it's available in Workers. But worth trying? In any case, please at least add a TODO comment in here so we know to get rid of this later.

src/pyodide/internal/_workers.py
```diff
@@ -122,7 +133,7 @@ def body(self) -> "js.ReadableStream":
         """
         Returns the body as a JavaScript ReadableStream from the JavaScript Response instance.
         """
-        return self.js_response.body
+        return _jsnull_to_none(self.js_response.body)
```
dom96, line 136: Nice, thanks!

dom96 (approved):

src/pyodide/internal/pool/sentinel.ts
hoodmane: `atob` returns a string, we need a `Uint8Array`. cf the explanation of why they are adding `Uint8Array.fromBase64()`.

hoodmane: Will add a comment.

PR #4614, opened 2025-07-25
Changes `run` method and fixes `sleep_until`

Caio-Nogueira:
- Changes the workflow handler method to `run` instead of the `on_run`, as the platform moves away from the prefixes
- Fixes `sleep_until`: users can now provide either a string (respecting rules of `WorkflowDelayDuration`), or a `datetime.datetime` object which is then converted into a JS `Date` object

### DIFF
--- a/src/pyodide/internal/_workers.py
+++ b/src/pyodide/internal/_workers.py
@@ -985,10 +985,14 @@ class _WorkflowStepWrapper:
         return decorator
 
     def sleep(self, *args, **kwargs):
+        # all types should be primitives - no need for explicit translation
         return self._js_step.sleep(*args, **kwargs)
 
-    def sleep_until(self, *args, **kwargs):
-        return self._js_step.sleepUntil(*args, **kwargs)
+    def sleep_until(self, name, timestamp):
+        if not isinstance(timestamp, str):
+            timestamp = python_to_rpc(timestamp)
+
+        return self._js_step.sleepUntil(name, timestamp)
 
     def wait_for_event(self, name, event_type, /, timeout="24 hours"):
         return self._js_step.waitForEvent(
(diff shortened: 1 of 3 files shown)


src/pyodide/internal/_workers.py
```diff
@@ -869,6 +869,24 @@ def python_to_rpc(value) -> JsProxy:
     return result
 
 
+def python_to_js_object(value) -> JsProxy:
```
Caio-Nogueira: This is meant to be another user facing method. Just like `python_to_rpc`, but with a different `dict_converter`. Does its naming reflect it?

LuisDuarte1: why does `python_to_rpc` convert it to `Map`s? I would expect it to convert it as an object - since everything in js uses an object

Caio-Nogueira: Maybe a `Map` would be the most faithful translation to a Python dictionary, since it would preserve insertion order, and may be less lossy.

Maybe @hoodmane could have more context into this

dom96: It's because that's how we pass Python dictionaries through the RPC layer. `Map` is explicitly supported through RPC.

So I'm a little on the fence of whether this function should be exported and whether it's even needed. You're mixing RPC and ordinary JS FFI here. Why are you doing this? Is what you really want just `to_js` without the default_converter being set to `_python_to_rpc_default_converter`?

Caio-Nogueira: > It's because that's how we pass Python dictionaries through the RPC layer. Map is explicitly supported through RPC.

The goal here is to have a helper method that explicitly converts to JS objects instead of `Map`. IIUC the `default_converter` doesn't change how dictionaries are translated.

dom96: Pyodide already provides this helper method, you should use it directly and without the `default_converter` being set (unless there is a good reason you're doing this, if so what is it?)

Caio-Nogueira: I'm okay with removing this. It means that users must use the `to_js` method directly (and specify the `dict_converter` accordingly)

PR #4616, opened 2025-07-25
Adds compat flag which disables global Python handlers.

dom96:
As a team we decided to remove global handlers in Python Workers and instead standardise on a `Default` class which defines the handler methods. This PR implements a new compat flag which enforces the usage of this class.

This PR also renames the default WorkerEntrypoint to start with a capital letter rather than lowercase.

### Test Plan

```
bazel run @workerd//src/workerd/server/tests/python:sdk_development@
```

### DIFF
--- a/src/pyodide/internal/_workers.py
+++ b/src/pyodide/internal/_workers.py
@@ -808,6 +808,11 @@ def python_from_rpc(obj: "JsProxy"):
     if not hasattr(obj, "constructor"):
         return obj
 
+    if obj.constructor.name == "TestController":
+        # This object currently has no methods defined on it. If this changes we should
+        # implement a Python wrapper for it, but for now we'll just pass in None.
+        return None
+
     result = obj.to_py(default_converter=_python_from_rpc_default_converter)
 
     return result
--- a/src/pyodide/internal/metadata.ts
+++ b/src/pyodide/internal/metadata.ts
@@ -42,7 +42,7 @@ export const MAIN_MODULE_NAME = MetadataReader.getMainModule();
 
 export interface CompatibilityFlags {
   python_workflows?: boolean;
-  [key: string]: boolean | undefined;
+  python_no_global_handlers?: boolean;
 }
 
 export interface CloudflareGlobal {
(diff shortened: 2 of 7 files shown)


src/workerd/io/compatibility-date.capnp
```diff
@@ -889,4 +889,10 @@ struct CompatibilityFlags @0x8f8c1b68151b6cef {
       $compatDisableFlag("no_expose_global_message_channel")
       $compatEnableDate("2025-08-15");
   # Enables exposure of the MessagePort and MessageChannel classes on the global scope.
+
+  pythonNoGlobalHandlers @103 :Bool
+      $compatEnableFlag("python_no_global_handlers")
```
danlapid, line 906: We probably need a disable flag as well

danlapid (approved):

src/pyodide/internal/_workers.py
```diff
@@ -808,6 +808,11 @@ def python_from_rpc(obj: "JsProxy"):
     if not hasattr(obj, "constructor"):
         return obj
 
+    if obj.constructor.name == "TestController":
```
hoodmane, line 811: What's `TestController`?

src/pyodide/internal/metadata.ts
```diff
@@ -42,6 +42,7 @@ export const MAIN_MODULE_NAME = MetadataReader.getMainModule();
 
 export interface CompatibilityFlags {
   python_workflows?: boolean;
+  python_no_global_handlers?: boolean;
   [key: string]: boolean | undefined;
```
hoodmane: Maybe we should get rid of the `[key: string]` signature and just add new flags here as we use them.

hoodmane: No action necessary.

src/pyodide/internal/metadata.ts
```diff
@@ -54,3 +55,5 @@ export interface CloudflareGlobal {
 const compatibilityFlags: CompatibilityFlags =
   (globalThis as CloudflareGlobal)?.Cloudflare?.compatibilityFlags ?? {};
 export const workflowsEnabled: boolean = !!compatibilityFlags.python_workflows;
+export const pythonNoGlobalHandlers: boolean =
+  !!compatibilityFlags.python_no_global_handlers;
```
hoodmane: Can we remove a layer of negation? e.g.,
```suggestion
export const legacyGlobalHandlers: boolean =
  !compatibilityFlags.python_no_global_handlers;
```
I don't like the double negative of ` !pythonNoGlobalHandlers`.

src/pyodide/internal/_workers.py
dom96: It's the argument that gets passed to the `test` handler https://github.com/cloudflare/workerd/blob/main/src/workerd/api/global-scope.h#L226

PR #4617, opened 2025-07-25
Some fixes for sidecar IP randomization

npaun:
* Make sure randomization is used unless it's explicitly disabled
* Make the supervisor assign _all_ of those weird ports WPT uses, even if we don't currently need them for tests.

### DIFF
--- a/build/wd_test.bzl
+++ b/build/wd_test.bzl
@@ -151,7 +151,7 @@ def _wd_test_impl(ctx):
         runtest = runtest,
         supervisor = ctx.file.sidecar_supervisor.short_path if ctx.file.sidecar_supervisor else "",
         port_bindings = ",".join(ctx.attr.sidecar_port_bindings),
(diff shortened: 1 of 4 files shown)


build/wd_test.bzl
```diff
@@ -151,7 +151,7 @@ def _wd_test_impl(ctx):
         runtest = runtest,
         supervisor = ctx.file.sidecar_supervisor.short_path if ctx.file.sidecar_supervisor else "",
         port_bindings = ",".join(ctx.attr.sidecar_port_bindings),
-        randomize_ip = str(ctx.attr.sidecar_randomize_ip),
+        randomize_ip = "true" if ctx.attr.sidecar_randomize_ip else "false",
```
anonrig, line 154: Can't we just make this a boolean rather than a string?

npaun: We're passing this as an environment variable in the generated shell scripts of wd_test, so everything is a string.

PR #4618, opened 2025-07-25
Avoid passing ownership of WorkerSource to newScript(...)

jasnell:


src/workerd/tests/test-fixture.c++
```diff
@@ -338,9 +338,10 @@ TestFixture::TestFixture(SetupParams&& params)
           scriptId,
           kj::heap<MockIsolateLimitEnforcer>(),
           Worker::Isolate::InspectorPolicy::DISALLOW)),
+      workerSource(server::WorkerdApi::extractSource(mainModuleName, config, *errorReporter)),
       workerScript(kj::atomicRefcounted<Worker::Script>(kj::atomicAddRef(*workerIsolate),
           scriptId,
-          server::WorkerdApi::extractSource(mainModuleName, config, *errorReporter),
+          workerSource,
```
kentonv: I think this change is not actually necessary. The temporary value returned by `extractSource()` will live through the constructor of `workerScript`, and isn't needed after that.

PR #4620, opened 2025-07-26
Support Structured Logging at startup.

danlapid:
Many startup and config errors are logged through ProcessContext. These logs up until now were logged in plain text even when structuredLogging was enabled.

This change adds support for structured logging at startup. This change is still limited as any logs and errors during parsing of the configuration will still be logged as plain text.

But this should minimize those to the absolute minimum.

One important thing to note, in order to not duplicate any logic from TopLevelProcessContext, error logs from StructuredLoggingProcessContext go to STDERR instead of STDOUT.

### DIFF
--- a/src/workerd/server/json-logger.h
+++ b/src/workerd/server/json-logger.h
@@ -6,6 +6,7 @@
 
 #include <kj/exception.h>
 #include <kj/function.h>
+#include <kj/main.h>
 #include <kj/string.h>
 
 namespace workerd::server {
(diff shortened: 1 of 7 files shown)


src/workerd/server/tests/structured-logging/structured-logging-test.mjs
```diff
@@ -0,0 +1,73 @@
...
+  );
+  assert(
+    typeof logEntry.timestamp === 'string',
+    `timestamp should be string, got: ${typeof logEntry.timestamp}`
+  );
+
+  // Verify timestamp is a valid ISO date
+  assert.doesNotThrow(() => {
+    new Date(logEntry.timestamp);
+  }, `timestamp should be valid date: ${logEntry.timestamp}`);
```
dario-piotrowicz, line 62: very very nit:

```suggestion
  }, `timestamp should be valid date, got: ${logEntry.timestamp}`);
```

For consistency with the above messages

src/workerd/server/tests/structured-logging/structured-logging-json.wd-test
```diff
@@ -0,0 +1,27 @@
...
+  sockets = [
+    (
+      name = "http",
+      address = "*:8080",
+      http = (),
+      service = "main"
+    )
+  ],
+  structuredLogging = true
+);
```
dario-piotrowicz, line 27:
```suggestion
);

```

src/workerd/server/json-logger.h
```diff
@@ -28,4 +29,31 @@ class JsonLogger: public kj::ExceptionCallback {
...
+  // A ProcessContext implementation that supports both plain text and structured JSON logging.
+  // This context wraps TopLevelProcessContext and adds the ability to emit log messages in
+  // JSON format when structured logging is enabled.
+
+ public:
+  explicit StructuredLoggingProcessContext(kj::StringPtr programName);
+
+  // Enable structured JSON logging. This can only be called once and cannot be reversed.
+  // When enabled: Log messages are formatted as JSON and sent to stdout
+  // When disabled: Log messages are sent as plain text to stderr (like TopLevelProcessContext)
```
dario-piotrowicz: Are these comments accurate? because when structured logs are enabled the log messages can make it into stdout but also stderr now, no?

Also when structured logs are disabled log messages are not all sent to stderr (but to both stderr and stdout)

Or am I missing something?

dario-piotrowicz (approved):
I've left a couple of very nit comments

I am not really familiar with this codebase but ballpark the changes look good to me 🙂 

I also tested these changes with wrangler and they seem to work as expected there.

src/workerd/server/json-logger.h
danlapid: This should be stderr, not sure how or when it changed, it was stderr in previous commits.
Amended the comment to be a bit clearer.

PR #4625, opened 2025-07-28
reproduce v8 fast api bug

anonrig:
Here's a repro we saw couple of weeks ago:

Findings 

- Passing a v8 promise to a C++ method that has a non-void return value might trigger GC
- FastOneByteString's should be addressed before any GC occurs
- v8 promise triggering GC will corrupt FastOneByteString

Reproduction

- Force trigger GC while unwrapping jsg::Promise (v8::Function::New can trigger GC)
- Access v8 fast one byte string value will show corruption

Fix (alternatives)

- One alternative is to remove v8::FastOneByteString
- Second alternative is to unwrap FastOneByteString before any other parameters


jasnell:
+1 to removing `FastOneByteString`

PR #4626, opened 2025-07-28
Add logging for out of order TailEvents.

mar-cf:


src/workerd/io/trace.c++
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
 
+#include "kj/string.h"
```
fhanau: Nit: VSCode/clangd likes adding these, but they are generally not needed (if one wanted to include them still to strictly follow IWYU, it should be with angled brackets). Adding `--header-insertion=never` to the clangd command should prevent these from spawning again.

src/workerd/io/trace.c++
```diff
@@ -251,6 +253,50 @@ kj::String KJ_STRINGIFY(const InvocationSpanContext& context) {
   return kj::str(context.getTraceId(), "-", context.getInvocationId(), "-", context.getSpanId());
 }
 
+kj::String KJ_STRINGIFY(const tracing::TailEvent::Event& event) {
+  KJ_SWITCH_ONEOF(event) {
+    KJ_CASE_ONEOF(onset, tracing::Onset) {
+      return kj::str("Onset");
```
fhanau, line 257: Nit: Shouldn't we also stringify onset here? For solving the issue at hand this does not seem necessary (we've only seen SpanOpen/SpanClose cause trouble so far) but there should be at least a comment here in that case.

mar-cf: SG. I'll open another PR for onset debug logs.

PR #4627, opened 2025-07-28
Add name to container StartParams

gpanders:
Also add it as a readonly property to the container object.

The name is passed into the containers runtime and made available via the API so that users can view container names in the Dash (rather than only DO actor IDs).

### DIFF
--- a/src/workerd/api/container.c++
+++ b/src/workerd/api/container.c++
@@ -14,7 +14,8 @@ namespace workerd::api {
 
 Container::Container(rpc::Container::Client rpcClient, bool running)
     : rpcClient(IoContext::current().addObject(kj::heap(kj::mv(rpcClient)))),
-      running(running) {}
+      running(running),
+      name() {}
 
 void Container::start(jsg::Lock& js, jsg::Optional<StartupOptions> maybeOptions) {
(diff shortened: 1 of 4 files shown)


gpanders:

kentonv:
Do we really want to add this API just for containers? It seems like people using DO without containers also need this, just as much. I don't want to solve problems specifically for containers when we could be solving it for DOs in general.

Of course, for DOs in general, the ideal solution would be to pass down the name from `idFromName()`, but there are some technical details that make this harder than expected. It's possible, though, it's just a matter of engineering.

If we feel we don't have bandwidth to do said engineering, a compromise could be to have, say, a `ctx.setName()` API (or maybe just a property with a setter) -- independent of the container. We could then pass that name down to the container, if it is available.

Technical comments on the code in this PR:
* If the container is already running when the DO starts up, then `start()` will never be called and `ctx.contianer.name` will stay undefined. This includes when the DO hibernates and starts back up. I would probably remove the property. The app can just as easily store the name itself as a property of the DO instance or something.
* Should there be a size limit enforced on the name? Someone could set it to a 128MB string...

gpanders:
Discussed internally, we have an alternative approach that we'll go with instead.

PR #4630, opened 2025-07-29
Replace the obsolete internal compat flags module

jasnell:
Use the globalThis.Cloudflare instead.

As a follow-up todo for later, we should look at automatically generating types from the compat flag definitions so that `globalThis.Cloudflare` can be more than just a `Record<string, boolean | undefined>` /cc @penalosa


anonrig:
Please don't merge this before http server pull request. It will be extremely hard to fix a git conflict on 20+ commits.

jasnell:
I will update this once that lands. Just queuing it up so I don't forget.

jasnell:
This will make it easier https://github.com/cloudflare/workerd/pull/4631

anonrig:
> This will make it easier https://github.com/cloudflare/workerd/pull/4631

Extremely helpful. Thank you!

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -1,7 +1,14 @@
 // Copyright (c) 2023 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
-import { default as flags } from 'workerd:compatibility-flags';
+declare global {
+  const Cloudflare: {
+    readonly compatibilityFlags: Record<string, boolean | undefined>;
```
jasnell:
```suggestion
    readonly compatibilityFlags: Record<string, boolean>;
```

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -1,7 +1,14 @@
 // Copyright (c) 2023 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
-import { default as flags } from 'workerd:compatibility-flags';
+declare global {
```
anonrig: Can we move this to internal/globals.d.ts file and include it in Cloudflare/tsconfig so we don't have to import it on every place

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -1,7 +1,14 @@
...
+  const Cloudflare: {
+    readonly compatibilityFlags: Record<string, boolean>;
+  };
+}
+
+const queryMetadataOptional =
+  !!Cloudflare.compatibilityFlags['vectorize_query_metadata_optional'];
```
anonrig, line 6: Why do we need !!? The type says it's already a boolean?

src/cloudflare/internal/vectorize-api.ts
```diff
@@ -1,7 +1,14 @@
 // Copyright (c) 2023 Cloudflare, Inc.
 // Licensed under the Apache 2.0 license found in the LICENSE file or at:
 //     https://opensource.org/licenses/Apache-2.0
-import { default as flags } from 'workerd:compatibility-flags';
+declare global {
+  const Cloudflare: {
+    readonly compatibilityFlags: Record<string, boolean>;
```
anonrig: Can we explicitly define these values that we use?

jasnell: I'd prefer that we find a way of auto-generating the list from the source capnp file. Let's explore that first rather than adding another thing to manually update. I'm not in a hurry to land this.

src/cloudflare/internal/vectorize-api.ts
jasnell: Not strictly needed but I prefer it. Going to keep it with the !! for now.

src/cloudflare/internal/global.d.ts
```diff
@@ -0,0 +1,3 @@
+declare namespace Cloudflare {
+  const compatibilityFlags: Record<string, boolean>;
+}
```
jasnell, line 3: @penalosa ... it should be possible to generate this directly from the compat flags capnp definitions, both for internal and external use. If you can help guide I can do the work ;-) 

the structure is simple...

```
declare namespace Cloudflare {
  const compatibilityFlags: {
    readonly foo: boolean,
    //...
  };
}
```

Where there is one `readonly foo: boolean` entry for every enable flag in the capnp.

anonrig: What about adding a js script, that uses capnp-es, to tools folder that updates flags.d.TS and we import that file in here?

jasnell: Yeah something like that should work. I don't think it's worth blocking on it tho. This is something we can incrementally improve on.

src/cloudflare/internal/vectorize-api.ts
jasnell: done

anonrig (approved):

src/cloudflare/internal/global.d.ts
jasnell: Going to mark this resolved so I can merge but will unresolve it after so we don't forget ;-)

jamesopstad (approved):

PR #4632, opened 2025-07-29
Implements support for Python in dynamic workers.

dom96:
### Test Plan

```
bazel run @workerd//src/workerd/api:worker-loader-test
```

### DIFF
--- a/build/python_metadata.bzl
+++ b/build/python_metadata.bzl
@@ -72,6 +72,7 @@ def _make_bundle_version_info(versions):
         if name != "development":
             entry["id"] = _bundle_id(**entry)
         entry["feature_flags"] = [entry["flag"]]
+        entry["feature_string_flags"] = [entry["enable_flag_name"]]
         if "packages" in entry:
             entry["packages"] = entry["packages"]["info"]["tag"]
         _add_integrity(entry)
@@ -92,6 +93,7 @@ BUNDLE_VERSION_INFO = _make_bundle_version_info([
         "backport": "63",
         "integrity": "sha256-xrG65VJvao9GYH07C73Uq2jA9DW7O1DP16fiZo36Xq0=",
         "flag": "pythonWorkers",
+        "enable_flag_name": "python_workers",
         "emscripten_version": "3.1.52",
         "python_version": "3.12.1",
         "baseline_snapshot": "baseline-d13ce2f4a.bin",
@@ -121,6 +123,7 @@ BUNDLE_VERSION_INFO = _make_bundle_version_info([
         "backport": "2",
         "integrity": "sha256-04qtaf3jr6q7mixWrpeASgYzTW1WHb9NEILBGl8M9hk=",
         "flag": "pythonWorkers20250116",
+        "enable_flag_name": "python_workers_20250116",
         "emscripten_version": "3.1.58",
         "python_version": "3.12.7",
         "baseline_snapshot": "baseline-59fa311f4.bin",
@@ -149,5 +152,6 @@ BUNDLE_VERSION_INFO = _make_bundle_version_info([
         "pyodide_date": "dev",
         "id": "dev",
         "flag": "pythonWorkersDevPyodide",
+        "enable_flag_name": "python_workers_development",
     },
 ])
--- a/src/workerd/api/worker-loader-test.js
+++ b/src/workerd/api/worker-loader-test.js
@@ -32,6 +32,39 @@ export let basics = {
   },
 };
 
+export let pythonBasics = {
+  async test(ctrl, env, ctx) {
+    let worker = env.loader.get('pythonBasics', () => {
+      return {
+        compatibilityDate: '2025-01-01',
+        mainModule: 'foo.py',
+        compatibilityFlags: ['python_workers', 'python_no_global_handlers'],
+        modules: {
+          'foo.py': `
+from workers import WorkerEntrypoint
+class Default(WorkerEntrypoint):
+  async def greet(self, name):
+    return "Hello, " + name
+class alternate(WorkerEntrypoint):
+  async def greet(self, name):
+    return "Welcome, " + name
+          `,
+        },
+      };
+    });
+
+    {
+      let result = await worker.getEntrypoint().greet('Stacey');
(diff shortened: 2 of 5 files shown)


danlapid:
Can we have a workerd test as well please?
The two implementations are a bit different and we want to make sure both work as intended long term

kentonv:
Yeah it's possible some changes are actually needed in server.c++ and/or workerd-api.c++ to make this work in workerd, even though it apparently worked without further changes in the edge runtime.

src/workerd/api/worker-loader.h
```diff
@@ -59,8 +59,9 @@ class WorkerLoader: public jsg::Object {
     jsg::Optional<kj::Array<const byte>> data;  // byte blob, imports as ArrayBuffer
     jsg::Optional<jsg::Value> json;             // arbitrary JS value, will be serialized to JSON
                                                 // and then parsed again when imported
+    jsg::Optional<kj::String> py;               // Python module
```
kentonv, line 62: What about `pythonRequirement`? (I don't actually know what it does, but notice it's not represented here...)

danlapid: It's used for runtime-provided python packages.
We're moving away from it short term as it creates odd conflicts and provides little actual value outside of some small memory sharing opportunity.
We intend to revisit it later after we're done with UX and performance.

All that to say, because we are no longer suggesting it for regular python workers, I don't think we need to support it for dynamic python worker loading at the moment.

Using the vendoring approach is better for the users at the moment.

dom96: `pythonRequirement` allows the Worker to ask for built-in Python packages to be loaded, though I don't think we should expose it in this API for two reasons:

* Package snapshots are tricky to generate here. Currently they get generated in the validator, but since these don't go through the validator they won't be generated here.
* We likely want to provide a more higher-level API for package support, one which automatically determines whether to vendor the package or use a built-in one (like we intend to do in the pywrangler tool).

src/workerd/api/worker-loader-test.js
```diff
@@ -32,6 +32,41 @@ export let basics = {
...
+  async test(ctrl, env, ctx) {
+    let worker = env.loader.get('pythonBasics', () => {
+      return {
+        compatibilityDate: '2025-01-01',
+        mainModule: 'foo.py',
+        compatibilityFlags: ['python_workers'],
+        modules: {
+          'foo.py': `
+from workers import WorkerEntrypoint
+# TODO: workerd bug with default entrypoint
```
dom96: This will be fixed by https://github.com/cloudflare/workerd/pull/4616

danlapid (approved):

src/workerd/api/worker-loader-test.js
kentonv: Looks like that was merged, can this be updated?

src/workerd/api/worker-loader.c++
```diff
@@ -170,11 +180,11 @@ Worker::Script::Source WorkerLoader::extractSource(jsg::Lock& js, WorkerCode& co
     KJ_UNREACHABLE;
   };
 
+  bool isPython = code.mainModule.endsWith(".py"_kj);
```
kentonv, line 189: We have other "isPython" checks which return true if the bundle contains _any_ Python modules, but here you're only checking the main module. I imagine in practice a Python worker must have a Python main module, but I worry the inconsistency could lead to bugs if people specify a worker containing a weird mix of modules.

* What happens if your main module is JS, but you have Python modules?
* What happens if your main module is Python, but you have JS modules?

Do either of these two configurations make sense today? If not, can we explicitly enforce that you not do that? That both avoids bugs and frees our hand to decide to support one of these configurations in a more thoughtful way in the future.

src/workerd/api/worker-loader.c++
```diff
@@ -135,16 +135,24 @@ Worker::Script::Source WorkerLoader::extractSource(jsg::Lock& js, WorkerCode& co
   auto modules = KJ_MAP(entry, code.modules.fields) -> Worker::Script::Module {
     KJ_SWITCH_ONEOF(entry.value) {
       KJ_CASE_ONEOF(text, kj::String) {
+        if (entry.name.endsWith(".py"_kj)) {
```
kentonv, line 138: Hmm. If we're going to use file extensions to auto-detect types then I feel like we should also require that implicitly-typed JavaScript files end with `.js`, otherwise each time we add a language we're possibly breaking people if, like, they had JS files with a `.py` extension before.

build/python_metadata.bzl
```diff
@@ -72,6 +72,7 @@ def _make_bundle_version_info(versions):
         if name != "development":
             entry["id"] = _bundle_id(**entry)
         entry["feature_flags"] = [entry["flag"]]
+        entry["feature_string_flags"] = [entry["enable_flag_name"]]
```
kentonv, line 75: Are these bazel changes actually related to this PR?

dom96: Yes, I need them in the EW PR.

src/workerd/api/worker-loader.c++
dom96: Good point. I'll disallow this mixing.

src/workerd/api/worker-loader.c++
dom96: Done

src/workerd/api/worker-loader.c++
```diff
@@ -135,16 +135,28 @@ Worker::Script::Source WorkerLoader::extractSource(jsg::Lock& js, WorkerCode& co
...
+        }
+
+        if (entry.name.endsWith(".js"_kj)) {
+          return {
+            .name = entry.name,
+            .content = Worker::Script::EsModule{.body = text},
+          };
+        }
+
+        JSG_FAIL_REQUIRE(TypeError, "Module name must end with '.js' or '.py'. Got: ", entry.name);
```
kentonv:
```suggestion
        JSG_FAIL_REQUIRE(TypeError, "Module name must end with '.js' or '.py' (or the content must be an object indicating the type explicitly). Got: ", entry.name);
```

PR #4635, opened 2025-07-29
move httpServerHandler to cloudflare:node module

anonrig:
Forgot to add this with the node http server pull-request.

### DIFF
--- a/src/node/internal/internal_fs_callback.ts
+++ b/src/node/internal/internal_fs_callback.ts
@@ -929,6 +929,7 @@ export function readFile(
   } else {
     options = optionsOrCallback;
   }
+  // TODO(node-fs): Validate options more. Specifically the encoding option
   callWithSingleArgCallback(() => fssync.readFileSync(path, options), callback);
 }
 
--- a/src/workerd/api/filesystem.c++
+++ b/src/workerd/api/filesystem.c++
@@ -4,6 +4,7 @@
 #include "url-standard.h"
 #include "url.h"
 
+#include <workerd/api/node/exceptions.h>
 #include <workerd/api/streams/standard.h>
 
 namespace workerd::api {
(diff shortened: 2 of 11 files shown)


dario-piotrowicz (approved):

types/defines/rpc.d.ts
```diff
@@ -295,4 +295,18 @@ declare module 'cloudflare:workers' {
...
 
   export const env: Cloudflare.Env;
+
+  export interface DefaultHandler {
+    fetch?(request: Request): Response | Promise<Response>;
+    tail?(events: TraceItem[]): void | Promise<void>;
+    trace?(traces: TraceItem[]): void | Promise<void>;
+    scheduled?(controller: ScheduledController): void | Promise<void>;
+    queue?(batch: MessageBatch<unknown>): void | Promise<void>;
+    test?(controller: TestController): void | Promise<void>;
```
vicb: It looks like email is missing.

Could this type be retrieved from existing types to avoid missing fields when a prop is added?

anonrig: I copied the contents of WorkerEntrypoint. So, I'm not sure which is correct.

kentonv: `nodeCompatHttpServerHandler` only actually implements `fetch()`, right? Why not just declare it as such?

types/defines/rpc.d.ts
```diff
@@ -295,4 +295,18 @@ declare module 'cloudflare:workers' {
...
+  }
+
+  export function nodeCompatHttpServerHandler(
```
kentonv: Why is this exported from `cloudflare:workers` and not some node module, or maybe `cloudflare:node`, or something? It doesn't seem appropriate to put node compat stuff into the `cloudflare:workers` module?

types/defines/rpc.d.ts
jasnell: We want to make sure that other kinds of entry points are supported. That is, it should be possible to have this only provide the implementation for `fetch()` but to allow the handler to have any of our other entrypoints also defined on it optionally.

anonrig: I've created a ticket for email handler. Will follow up on a separate PR - https://github.com/cloudflare/workerd/issues/4652

src/cloudflare/workers.ts
```diff
@@ -92,41 +91,3 @@ export const env = new Proxy(
...
-  if (handlers == null || typeof handlers !== 'object') {
-    throw new Error(
-      'Handlers parameter passed to nodeCompatHttpServerHandler method must be an object'
-    );
-  }
-
-  return Object.assign(handlers, {
-    // We intentionally omitted ctx and env variables. Users should use
-    // importable equivalents to access those values. For example, using
-    // import { env, waitUntil } from 'cloudflare:workers
```
jasnell, line 117: Side note... in the docs for the compat flag and the examples, please be sure to mention that the `disallow_importable_env` compat flag should *not* be set if you are using this and need access to the `env` since that will prevent access.

types/defines/rpc.d.ts
anonrig: I've updated the PR and moved the function declaration to cloudflare:node as recommended.

jasnell (approved):

types/defines/rpc.d.ts
kentonv: @jasnell I don't understand what you mean.

`nodeCompatHttpServerHandler()` only provides `fetch`, right? How would "have any of our other entrypoints also defined on it optionally"?

anonrig: The second parameter can be used to pass an object with keys to different handlers

